Issue #2186, opened 2021-08-04
Simplify Server/Database Startup


tustvold:
**Problem:**

Over the last weeks we've seen a large number of issues related to the current server and database startup/shutdown logic contained within server, and the structure of the server crate in general:

* https://github.com/influxdata/influxdb_iox/issues/2146
* https://github.com/influxdata/influxdb_iox/issues/2162
* https://github.com/influxdata/influxdb_iox/issues/2174
* https://github.com/influxdata/influxdb_iox/issues/2091
* https://github.com/influxdata/influxdb_iox/issues/2134
* https://github.com/influxdata/influxdb_iox/issues/2139
* https://github.com/influxdata/influxdb_iox/issues/2065
* https://github.com/influxdata/influxdb_iox/issues/2053
* https://github.com/influxdata/influxdb_iox/issues/1990
* https://github.com/influxdata/influxdb_iox/pull/2168

Additionally there are the following problems which currently lack filed issues:

* A database cannot be shutdown, which blocks adding database deletion
* The status of a database's startup is only externally visible once it finishes initialization or errors
* There is a large amount of highly coupled code across lib.rs, init.rs, config.rs that is non-trivial to understand
* The separation of responsibilities between `Server` and `Db` is unclear

**Proposal:**

The full proposal can be found [here](https://docs.google.com/document/d/1sSRODNDcqQ0jncMZGK43zpD5RmvZIXCT_yyoiUfCx-I/edit?usp=sharing) but the major changes are:

* Create a `Database` abstraction that contains the subsystems necessary to implement a single `DatabaseRules`
* `Database` handles its startup/shutdown state machine, and background loop
* `Db` becomes an instance-local, queryable, possibly persisted, and possibly mutable data store
* `LifecyclePolicy`, `WriteBuffer`, etc... is pulled out of `Db` onto `Database`
* `Server` stores a collection of `Database`, and has no involvement in their initialization process
* A `/initialized` HTTP endpoint is added that returns 200 when the `Server` and all `Database` are initialized - for use as a kubernetes startup probe
* The `Server` and `Database` startup state machines remain largely the same, but the error states are explicit, with explicit operator initiated recovery transitions

**Potential Future Improvements**

* Separate `Db` into its own crate
* Ability to delete databases - https://github.com/influxdata/influxdb_iox/issues/1881
* Ability to shutdown databases with an API call

alamb:
Maybe a better name for `Db` is `Store` or `DataStore`

[...]

Issue #2139, opened 2021-07-28
Binary Write Protocol Skips Routing/Sharding


tustvold:
The binary write protocol added in https://github.com/influxdata/influxdb_iox/pull/1842/files#diff-64aa9b484d4b3759cc0dd469329aa5a6bab87366c6c55eb61d9a1d1129201fcdR588 always writes to the local DB, skipping the routing configuration if any. This should be fixed.

As a general point, the code in the PR is not obviously wrong. I'm currently working on trying to cleanup Server a bit, and will see if I can remove this footgun.

PR #2203, opened 2021-08-05
feat: add Database abstraction (#2186)

tustvold:
This starts the process of rejigging the server crate to reduce the tight coupling between server, db and the database startup sequence. Unfortunately I could not find a way to do this as a small incremental change, as the code was so tightly coupled together, and so this is an unfortunately large PR.

This reworks `Server` to be a collection of `Database` which each manage their startup independently. The result is most of the logic that used to live under init.rs now lives in database.rs, and much of the coordination logic that used to live in config.rs is no longer necessary. 

In addition to fixing the issues identified in #2186, the eventual "vision", of which this is the first set of changes, is for a slightly less monolithic design where everything has to be plumbed through the entire object hierarchy. 

To put this in concrete terms, an API to rollover a partition in a `Db` should not imo need to be plumbed through as member functions on `Server` or `Database`, or an API to recover from a broken catalog need to be exposed as a member function on `Server`, instead a service implementation might fetch a `Db` or a `Database` from the hierarchy and make calls directly to that. I at least find it easier to reason about smaller components responsible for a couple of things, than large aggregating objects responsible for everything beneath them.

There are a large number of TODOs still in the code, not all the error propagation is hooked up, not all the logic sits in its final position, etc... but I'd like to get these in as subsequent, smaller PRs.


src/influxdb_ioxd/http.rs
```diff
@@ -643,8 +636,10 @@ async fn query<M: ConnectionManager + Send + Sync + Debug + 'static>(
     let db_name = DatabaseName::new(&db_name_str).context(DatabaseNameError)?;
     debug!(uri = ?req.uri(), %q, ?format, %db_name, "running SQL query");
 
+    // TODO: Propagate error
```
tustvold, line 639: This is no worse than it was before, but should probably be fixed

server/src/database.rs
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    // TODO: Drain database jobs
+
+    info!(db_name=%shared.config.name, "database worker finished");
+}
+
+/// Try to drive the database to `DatabaseState::Initialized` returns when
+/// this is achieved or the shutdown signal is triggered
+async fn initialize_database(shared: &DatabaseShared) {
```
tustvold, line 266: This is the method that drives the Database initialization

server/src/database.rs
```diff
@@ -0,0 +1,540 @@
...
+                            info!(%db_name, "init transaction already in progress");
+                            None
+                        }
+                    }
+                }
+                // Operator intervention required
+                DatabaseState::RulesLoadError(_, e)
+                | DatabaseState::CatalogLoadError(_, e)
+                | DatabaseState::ReplayError(_, e) => {
+                    error!(%db_name, error=%e, "database in error state - operator intervention required");
```
tustvold, line 293: This will log the error repeatedly, I think this is a good thing

server/src/database.rs
```diff
@@ -0,0 +1,540 @@
...
+            Some((state, handle)) => (state, handle),
+            None => {
+                info!(%db_name, "backing off initialization");
+                tokio::time::sleep(INIT_BACKOFF).await;
+                continue;
+            }
+        };
+
+        // Try to advance to the next state
+        let next_state = match state {
```
tustvold, line 310: It occurs to me on reading this `advance` should probably short-circuit on shutdown - something for a future PR

server/src/lib.rs
```diff
@@ -449,183 +457,161 @@ where
...
+
+        Ok(Arc::clone(db))
     }
 
-    /// Current database init state.
-    pub fn database_state(&self, name: &str) -> Option<DatabaseStateCode> {
-        let db_name = DatabaseName::new(name).ok()?;
-        let config = self.config().ok()?;
-        config.db_state(&db_name)
+    /// Returns an initialized `Db` by name
```
tustvold, line 542: This method feels a bit off to me, but it reduced code churn to keep it, at least for now

server/src/lib.rs
```diff
@@ -883,20 +876,11 @@ where
         NoRemoteReachable { errors }.fail()
     }
 
-    pub async fn write_entry(&self, db_name: &str, entry_bytes: Vec<u8>) -> Result<()> {
```
tustvold, line 886: I removed this method as it was confusingly named, which possibly contributed to #2139 and doesn't seem to really add anything significant above `write_entry_local`

server/src/lib.rs
```diff
@@ -1007,99 +982,59 @@ where
         })
     }
 
-    /// Wipe preserved catalog of specific DB.
+    /// Recover database that has failed to load its catalog by wiping it
     ///
-    /// The DB must not yet exist within this server for this to work! This is done to prevent race conditions between
-    /// DB jobs and this command.
+    /// The DB must exist in the server and have failed to load the catalog for this to work
+    /// This is done to prevent race conditions between DB jobs and this command
```
tustvold, line 988: Longer term I'd like to remove this method from server, and have a variety of recovery transitions exposed on `Database`. The only database APIs I'd expect on server are creation, deletion and retrieval.

server/src/lib.rs
```diff
@@ -2013,66 +2068,56 @@ mod tests {
...
-        assert!(!PreservedCatalog::exists(
-            application.object_store().as_ref(),
-            server_id,
-            &db_name_non_existing.to_string()
-        )
-        .await
-        .unwrap());
-        assert!(server.error_database(&db_name_non_existing).is_none());
-        assert!(server.db(&db_name_non_existing).is_none());
+        assert_eq!(
```
tustvold, line 2071: Wipe preserved catalog is now only valid when the catalog has failed to load, if you wish to wipe the catalog of a database in a different state, it should go through the database deletion process.

server/src/lib.rs
```diff
@@ -2127,30 +2173,6 @@ mod tests {
         assert!(matches!(err, Error::CannotCreatePreservedCatalog { .. }));
     }
 
-    #[tokio::test]
```
tustvold, line 2130: It isn't possible to write this test anymore, as you can't inject a custom `WriteBuffer` into `Database`. I plan to address this in a follow up

FWIW this is a test that highlights one of the reasons I don't like plumbing everything through member functions on successively large objects - the error propagation gets gnarly, as the responsibilities of the object balloon.

src/commands/run.rs
```diff
@@ -270,6 +270,15 @@ Possible values (case insensitive):
     )]
     pub max_http_request_size: usize,
 
+    /// Automatically wipe the preserved catalog on error
+    #[structopt(
```
tustvold, line 274: I needed to plumb this through so that I could write end-to-end tests of the catalog wiping functionality, with this set to true it automatically wipes the catalog so you can't get it into a state where the operator would :sweat_smile:

server/src/database.rs
```diff
@@ -0,0 +1,540 @@
...
+                | DatabaseState::CatalogLoadError(_, e)
+                | DatabaseState::ReplayError(_, e) => return Err(Arc::clone(e)),
+            }
+
+            notify.await;
+        }
+    }
+
+    /// Recover from a CatalogLoadError by wiping the catalog
+    pub fn wipe_preserved_catalog(&self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
```
tustvold, line 172: It occurs to me that we need the ability to wipe a loaded, but not replayed catalog as well - will fix tomorrow

tustvold, line 172: The slightly strange signature is because we want to do the planning up-front, so we can surface "expected" errors to the client synchronously, but want to spawn the actual work as a background task. This is very similar to the API for lifecycle tasks

server/src/init.rs
```diff
@@ -1,432 +0,0 @@
...
-}
-
-pub type Result<T, E = Error> = std::result::Result<T, E>;
-
-/// Loads the database configurations based on the databases in the
-/// object store. Any databases in the config already won't be
-/// replaced.
-///
-/// Returns a Vec containing the results of loading the contained databases
-pub(crate) async fn initialize_server(
```
tustvold, line 96: This method got reworked into server as list_databases

server/src/init.rs
```diff
@@ -1,432 +0,0 @@
...
-                .await;
-                (db_name, result)
-            })
-        })
-        .collect();
-
-    Ok(futures::future::join_all(handles).await)
-}
-
-async fn initialize_database(
```
tustvold, line 140: The equivalent of this function, and those it calls, can now be found in database.rs

server/src/lib.rs
```diff
@@ -449,183 +457,161 @@ where
...
-            ServerStage::Initializing { last_error, .. } => last_error.clone(),
+    pub fn server_init_error(&self) -> Option<Arc<InitError>> {
+        match &**self.shared.state.read() {
+            ServerState::InitError(_, e) => Some(Arc::clone(e)),
             _ => None,
         }
     }
 
-    /// List all databases with errors in sorted order.
-    pub fn databases_with_errors(&self) -> Vec<String> {
```
tustvold, line 536: These methods are no longer exposed on `Server`, rather the service implementation can fetch a `Database` (or multiple) and ask them directly for their status

server/src/lib.rs
```diff
@@ -449,183 +457,165 @@ where
...
+            (
+                initialized.server_id,
+                Arc::clone(&initialized.create_database_lock),
+            )
+        };
 
-        // Reserve name before expensive IO (e.g. loading the preserved catalog)
-        let mut db_reservation = config.create_db(rules.name.clone())?;
+        // Prevent concurrent database creation
+        let guard = create_database_lock.lock().await;
```
tustvold, line 572: I don't like this, this all stems from what is a current limitation of Freezable. If a database or state transition is in progress we want to wait until it isn't, not just return an error. I have some ideas on how to address this, but I didn't want to block this on them

tests/end_to_end_cases/scenario.rs
```diff
@@ -561,3 +561,58 @@ pub async fn list_chunks(fixture: &ServerFixture, db_name: &str) -> Vec<ChunkSum
 
     chunks.into_iter().map(|c| c.try_into().unwrap()).collect()
 }
+
+/// Creates a database with a broken catalog
+pub async fn fixture_broken_catalog(db_name: &str) -> ServerFixture {
```
tustvold, line 566: Wipe preserved catalog is now exclusively a recovery action for a broken catalog. If you want to actually delete a database, I'd expect this to go through the delete database API @carols10cents is working on

tests/end_to_end_cases/persistence.rs
```diff
@@ -196,13 +197,12 @@ async fn test_update_late_arrival() {
...
     // fixtures
     let fixture = ServerFixture::create_single_use().await;
-    let server_id = 42;
     let db_name = rand_name();
 
     // set server ID
     let mut management_client = fixture.management_client();
     management_client
-        .update_server_id(server_id)
+        .update_server_id(DEFAULT_SERVER_ID)
```
tustvold, line 205: This is a bit of a quirk of the test harness - calling restart, restarts the server and sets this server ID. I think this is a tad surprising and might fix as a follow on PR

src/influxdb_ioxd/rpc/error.rs
```diff
@@ -47,6 +45,14 @@ pub fn default_server_error_handler(error: server::Error) -> tonic::Status {
         }
         .into(),
         Error::RemoteError { source } => tonic::Status::unavailable(source.to_string()),
+        Error::WipePreservedCatalog { source } => match source {
```
tustvold, line 48: I expect we'll end up with a default_database_error_handler method eventually

server/src/lib.rs
```diff
@@ -1115,6 +1054,7 @@ where
     }
 }
 
+/// TODO: Revisit this trait's API
```
tustvold, line 1057: I'm not really sure what is using this API and why, this is more a TODO for myself to look into it

server/src/database.rs
```diff
@@ -0,0 +1,554 @@
...
+/// TODO: move to object_store crate
+async fn get_store_bytes(
+    store: &ObjectStore,
+    location: &Path,
+) -> Result<bytes::Bytes, object_store::Error> {
+    use futures::stream::TryStreamExt;
+
+    let stream = store.get(location).await?;
+    let bytes = stream
+        .try_fold(BytesMut::new(), |mut acc, buf| async move {
```
tustvold, line 547: This method used to be in init.rs - this makes the additional change to avoid copying each byte range twice. In practice I don't imagine this will matter all that much, as we likely won't get a streaming reply for the rules file, but :shrug:

tests/common/server_fixture.rs
```diff
@@ -226,13 +231,20 @@ impl ServerFixture {
         }
     }
 
-    /// Wait until server is initialized. Databases should be loaded.
-    pub async fn wait_server_initialized(&self) {
+    /// Wait until server is initialized and all databases have initialized or errored.
```
tustvold, line 234: Previously the server was only initialized when it had attempted to initialize all databases. I'm not sure any tests actually relied on this, but I figured it was better to eliminate this potential source of test flakiness at source

server/src/lib.rs
```diff
@@ -449,183 +457,165 @@ where
...
+    /// Returns an initialized `Db` by name
+    pub fn db(&self, name: &DatabaseName<'_>) -> Result<Arc<Db>> {
+        self.database(name)?
+            .initialized_db()
+            .ok_or(Error::DatabaseNotInitialized {
+                db_name: name.to_string(),
+            })
     }
 
     /// Tells the server the set of rules for a database.
```
alamb, line 551:
```suggestion
    /// Tells the server the set of rules for a database.
    /// resolves only after the database has been successfully initialized or an error has occurred
```

server/src/lib.rs
```diff
@@ -449,183 +457,165 @@ where
...
-        let mut data = BytesMut::new();
-        encode_database_rules(rules, &mut data).context(ErrorSerializingRulesProtobuf)?;
+        let database = {
+            let mut state = self.shared.state.write();
+
+            // Have exclusive lock on state - can drop database creation lock
+            std::mem::drop(guard);
+
+            let state = state.get_mut().expect("no transaction in progress");
+            let database = match state {
```
alamb, line 594: I wonder if you could cut down a few levels of the indentation by using
```rust
        let initialized = state.initialized()?;
```

server/src/lib.rs
```diff
@@ -449,183 +457,165 @@ where
...
+                ServerState::Initialized(initialized) => {
+                    match initialized.databases.entry(db_name.clone()) {
```
alamb, line 596: Maybe you could cut down even more indent by just calling `insert` rather than using the entry API:

```rust
let new_db = Arc::new(Database::new(.....));
let existing_db initialized.databases.insert(db_name.clone(), );
assert!(existing_db.is_none())
```

server/src/lib.rs
```diff
@@ -636,76 +626,92 @@ where
...
-                        config: Arc::clone(&config),
-                        wipe_catalog_on_error,
-                        last_error,
-                    };
-                    (wipe_catalog_on_error, skip_replay_and_seek_instead, config)
+        if self.initialized() {
+            return;
+        }
+
+        let (init_ready, handle) = {
```
alamb, line 636: In order to understand the state of server startup from the logs, can you please add INTO logging here that says "starting server initialization"

server/src/lib.rs
```diff
@@ -979,23 +966,15 @@ where
     /// background job, dropping when complete.
     pub fn close_chunk(
```
alamb, line 967: maybe also the plan is to push this into Database?

server/src/lib.rs
alamb: I think the tests in the storage RPC handler might be doing so.

The trait definitely doesn't need to be in `query` however, and I had a todo for a long time to put it in some other crate

alamb (approved):
I pulled this code down, reviewed it in my editor. I like the new structure a lot -- I have some logging / testing questions and suggestions, but the core structure of this PR is very nice.

Well done @tustvold 🏅 

😅  now I need to go for a walk to clear my mind

server/src/database.rs
tustvold: I wanted to be able to use the ? operator, which is why it is structured this way. Otherwise you end up with lots of duplicated map_err

src/commands/run.rs
tustvold: I'm not sure I follow - this is a boolean flag?

server/src/database.rs
```diff
@@ -0,0 +1,554 @@
...
+                .state
+                .read()
+                .get_initialized()
+                .expect("expected initialized")
+                .db,
+        );
+
+        info!(db_name=%shared.config.name, "database finished initialization - starting Db loop");
+
+        // TODO: Pull this out of `Db`
```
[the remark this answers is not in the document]
tustvold: Will clarify, I'm referring to the background_worker

server/src/database.rs
```diff
@@ -0,0 +1,554 @@
...
+    Initialized(DatabaseStateInitialized),
+
+    RulesLoadError(DatabaseStateKnown, Arc<InitError>),
+    CatalogLoadError(DatabaseStateRulesLoaded, Arc<InitError>),
+    ReplayError(DatabaseStateCatalogLoaded, Arc<InitError>),
+}
+
+impl DatabaseState {
+    fn state_code(&self) -> DatabaseStateCode {
+        // TODO: Update DatabaseStateCode
```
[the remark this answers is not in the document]
tustvold: It was missing some states - will just fix

src/influxdb_ioxd/rpc/management.rs
```diff
@@ -418,33 +397,32 @@ where
...
-                .map(|db_name| {
-                    let error = self.server.error_database(&db_name).map(|e| ProtobufError {
-                        message: e.to_string(),
-                    });
-
-                    let state: database_status::DatabaseState =
-                        self.server.database_state(&db_name).into();
+        let mut database_statuses: Vec<_> = self
+            .server
+            .databases()
```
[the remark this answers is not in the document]
tustvold: This is actually intentional, it isn't an error if you ask for the status of an initializing database - will make it more clear, but nice spot :eagle: :eye:

server/src/lib.rs
```diff
@@ -393,34 +368,67 @@ pub struct Server<M: ConnectionManager> {
...
+    Initialized(ServerStateInitialized),
+}
+
+impl ServerState {
+    fn initialized(&self) -> Result<&ServerStateInitialized> {
+        match self {
+            ServerState::Startup(_) => Err(Error::IdNotSet),
+            ServerState::InitReady(state) | ServerState::InitError(state, _) => {
+                Err(Error::ServerNotInitialized {
+                    server_id: state.server_id,
```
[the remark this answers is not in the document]
tustvold: This is the case, in both cases it will continue to keep trying to initialize. This is different from Database, where all error are currently assumed to be terminal.

server/src/lib.rs
tustvold: I'd need to add an initialized_mut method... This is the downside of borrow mutability being part of the function signature

server/src/lib.rs
tustvold: If the assertion does fire, I would expect it to not have mutated the collection, but perhaps I'm being overly paranoid...

server/src/lib.rs
tustvold: There are logs literally a couple of lines below, that will print continually every INIT_BACKOFF until init succeeds, including one that says it is attempting to init?

server/src/lib.rs
tustvold: Or Db directly tbh

server/src/database.rs
tustvold: I mean there is one in Database::new and at the start of background_loop, but I guess more logging can't hurt :sweat_smile:

server/src/config.rs
```diff
@@ -1,932 +0,0 @@
...
-    use super::*;
-
-    fn make_config() -> Config {
-        let store = Arc::new(ObjectStore::new_in_memory());
-        let server_id = ServerId::try_from(1).unwrap();
-        Config::new(Arc::new(ApplicationState::new(store, Some(1))), server_id)
-    }
-
-    #[tokio::test]
-    async fn create_db() {
```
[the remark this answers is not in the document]
tustvold: They're tests of the state machine and synchronisation logic that no longer exist, although I guess I should probably add some unit tests of their replacement even if they are extensively covered by other tests (most tests start a database :grin: )

server/src/lib.rs
```diff
@@ -393,34 +368,67 @@ pub struct Server<M: ConnectionManager> {
...
+            ServerState::InitReady(state) => Some(state.server_id),
+            ServerState::InitError(state, _) => Some(state.server_id),
+            ServerState::Initialized(state) => Some(state.server_id),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct ServerStateStartup {
+    wipe_catalog_on_error: bool,
```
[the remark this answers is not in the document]
tustvold: I was trying to avoid bringing RemoteTemplate along for the ride, will revisit if we end up with more

src/influxdb_ioxd/http.rs
tustvold: I'll just fix the error handling :laughing: or at least this bit of it. The HTTP API error handling could probably do with some TLC tbh

server/src/init.rs
```diff
@@ -1,432 +0,0 @@
...
-}
-
-#[cfg(test)]
-mod tests {
-    use object_store::path::ObjectStorePath;
-
-    use super::*;
-
-    #[tokio::test]
-    async fn test_get_database_config_bytes() {
```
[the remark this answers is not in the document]
tustvold: I actually didn't port this logic, it was previously performing a list operation prior to the get operation in order to provide this error code. The object store crate does not make it easy to determine the nature of the error

tests/common/server_fixture.rs
```diff
@@ -226,13 +231,20 @@ impl ServerFixture {
...
-                break;
+            let status = client.get_server_status().await.unwrap();
+            if status.initialized
+                && status.database_statuses.iter().all(|status| {
+                    status.error.is_some()
+                        || DatabaseState::from_i32(status.state).unwrap()
+                            == DatabaseState::Initialized
+                })
+            {
+                return status;
```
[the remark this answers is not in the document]
tustvold: It doesn't check the database initialized successfully, just that it did reach a terminal state. This preserves the behaviour from before. There are cases that very much expect a database to fail to initialize :grin:

server/src/config.rs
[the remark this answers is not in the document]
tustvold: Actually this PR is monstrous enough, and I'll be iterating on the Database in subsequent PRs. Unless you feel strongly, I'd like to get this in and add those later

src/commands/run.rs
alamb: I was trying to say that there seemed to be two different ways to encode a boolean flag (using "yes" / "no" and "true"/false) which seemed redundant and confusing to me

server/src/lib.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/2219

server/src/lib.rs
tustvold: Fix in #2224

Issue #2193, opened 2021-08-05
Relative paths in `parquet_file` 


crepererum:
# Use case:
Makes it easier to copy, backup/restore, and move catalogs including parquet files from one server to another. #2177 might benefit from easy copies as well.

# Current behaviour:
Currently all paths in the `parquet_file` crate start with the following prefix:

```text
<server_id>/<db_name>/
```

With #1881 that prefix might even grow to include a generation number.

# Proposal:
Avoid constructing the server-db prefix (see "Current behaviour") in `parquet_file` and instead use one of the following:

- a root path passed from outside
- an object store wrapper that adds prefixes

All paths emitted from `parquet_file` (either as return value or as stored value in the preserved catalog) should be relative to the root.

# Alternatives considered:
Not changing it.

PR #2204, opened 2021-08-05
WIP refactor: Extract a database persister type that wraps object store

carols10cents:
Connects to #2193.

This is not compiling yet, but this is the direction I'm going in. I decided to make a new type for a database (and eventually for a particular database generation) that wraps an object store so that we don't miss anyplace that we need to change the path.

Bonus, this should eliminate a *lot* of current duplication where we're currently passing around an object store, server id, and database name by encapsulating those together!

### DIFF
--- a/parquet_file/src/catalog.rs
+++ b/parquet_file/src/catalog.rs
@@ -14,9 +14,9 @@ use std::{
 use crate::metadata::IoxParquetMetaData;
 use bytes::Bytes;
 use chrono::{DateTime, Utc};
-use futures::TryStreamExt;
+use futures::{StreamExt, TryStreamExt};
 use generated_types::influxdata::iox::catalog::v1 as proto;
-use iox_object_store::IoxObjectStore;
+use iox_object_store::{IoxObjectStore, RelativePath};
 use object_store::{
     path::{parsed::DirsAndFileName, parts::PathPart, ObjectStorePath, Path},
     ObjectStore, ObjectStoreApi,
@@ -268,7 +268,7 @@ pub struct PreservedCatalog {
 impl PreservedCatalog {
     /// Checks if a preserved catalog exists.
     pub async fn exists(iox_object_store: &IoxObjectStore) -> Result<bool> {
-        Ok(!list_files(iox_object_store).await?.is_empty())
+        Ok(!iox_object_store.catalog_transactions().await.context(Read)?.next().await.is_some())
     }
 
     /// Find last transaction-start-timestamp.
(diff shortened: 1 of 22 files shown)


internal_types/src/persister.rs
```diff
@@ -0,0 +1,119 @@
...
+    where
+        S: Stream<Item = io::Result<Bytes>> + Send + Sync + 'static,
+    {
+        unimplemented!();
+    }
+
+    pub async fn list<'a>(
+        &'a self,
+        _prefix: Option<&'a Path>,
+    ) -> Result<BoxStream<'a, Result<Vec<Path>>>> {
```
crepererum, line 71: I am wondering if `Path` is the right type here of if something like `RelativePath` would be better. Or at least if we could have some functionality to calculate relative paths to the `data_path`, so that we can use these in the catalog protobuf files (aka if you copy the whole thing over to a new DB generation or server ID, that the relative paths preserved in the catalog would also point to the copied data files).

parquet_file/src/catalog.rs
```diff
@@ -245,36 +245,24 @@ pub struct PreservedCatalog {
     previous_tkey: RwLock<Option<TransactionKey>>,
     transaction_semaphore: Semaphore,
 
-    object_store: Arc<ObjectStore>,
-    server_id: ServerId,
-    db_name: String,
+    persister: Arc<Persister>,
```
crepererum, line 248: :heart:

internal_types/Cargo.toml
```diff
@@ -8,9 +8,13 @@ readme = "README.md"
 
 [dependencies]
 arrow = { version = "5.0", features = ["prettyprint"] }
+bytes = "1.0"
```
alamb, line 11: I would like to suggest putting the database persistor in its own crate to avoid all these new dependencies on `internal_types`

internal_types/src/persister.rs
alamb: What about wrapping. `Path` with the specific type of object that is stored there. Like

```rust
enum IOxPath {
  DBRules { db_name: String },
  ParquetFile { id: String, table_name: String }
  CheckpointFile { id: String, ... }
}

impl IOxPath {
  fn path(&self) -> Path
}
```


This would then centralize the code for the mapping from "IOx specific thing" to "object store path" and all your APIs could be in terms of `IOxPath`s and avoid all path manipulation / creation in other crates

internal_types/src/persister.rs
```diff
@@ -0,0 +1,119 @@
...
+use futures::{stream::BoxStream, Stream};
+use object_store::{
+    path::{ObjectStorePath, Path},
+    ListResult, ObjectStore, ObjectStoreApi, Result,
+};
+use std::{io, sync::Arc};
+
+/// Handles persistence of data for a particular database. Writes within its directory/prefix.
+#[derive(Debug)]
+pub struct Persister {
```
alamb, line 12: I wonder if calling this `IOxObjectStore` would get across the meaning that it added IOx specific semantics on to the more general ObjectStore interface?

alamb (commented):
I really like the direction this is heading @carols10cents 👍

internal_types/Cargo.toml
tustvold: `object_store` in particular is not a lightweight dependency

carols10cents:
I decided to submit part of this as #2263, and the `IoxPath` will be in a future PR, so I'm going to close this.

PR #2205, opened 2021-08-06
fix: only warn when partition ckpt and DB ckpt mins are out-of-sync

### LINKED ISSUE #2185: DB went down with replay error when hitting a likely datafusion bug
It seems a Datafusion bug while execute a query lead to IOx replay error

__Steps to reproduce:__
Run the following query in DB `844910ece80be8bc_3c0bd4c89186ca89`

```
SELECT Row_number() 
         OVER ( 
           partition BY Date_trunc('minute', time), bucket_id, partition_id 
           ORDER BY time DESC)    AS rn, 
       gauge, 
       Date_trunc('minute', time) AS time_bucket, 
       env 
FROM   storage_usage_bucket_cardinality 
WHERE  Cast(time AS BIGINT) > Cast(Now() AS BIGINT) - 5 * 60 * 1000000000 
       AND time <= Now(); 
```

We will see the following Datafusion internal error:
`Error running remote query: status: Internal, message: "Internal error reading points from database 844910ece80be8bc_3c0bd4c89186ca89: Internal error executing plan: Arrow error: External error: Arrow error: External error: Arrow error: External error: Execution error: Internal error: Unsupported data type in hasher. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 03 Aug 2021 18:55:08 GMT"} }`
Data Ticket: https://github.com/apache/arrow-datafusion/issues/821

Then `844910ece80be8bc_3c0bd4c89186ca89` also goes down with following error:

```
./scripts/grpcurl -plaintext localhost:1234 influxdata.iox.management.v1.ManagementService.GetServerStatus
.....
{
        "error": {
          "message": "cannot load catalog: Cannot build replay plan: Minimum sequence number in partition checkpoint (123964015) is lower than database-wide number (183235313) for partition http_api_request_duration_seconds:2021-08-02 08:00:00 and sequencer 0"
        },
        "state": "DATABASE_STATE_RULES_LOADED"
      },
....
```

Update: the DF fusion bug and the replay error may not be related. After I wiped the catalog and ran the query again, I hit the same DF bug but the DB is still up running.

crepererum:
There are currently a few bugs and semi-understood edge cases that can
lead to this case. So instead of bailing out, just issue a warning.

Closes #2185.

### DIFF
--- a/server/src/db/replay.rs
+++ b/server/src/db/replay.rs
@@ -1713,7 +1713,7 @@ mod tests {
                 Step::Ingest(vec![TestSequencedEntry {
                     sequencer_id: 0,
                     sequence_number: 1,
-                    lp: "table_1,tag_partition_by=a bar=10 0",
+                    lp: "table_1,tag_partition_by=a,tag=1 bar=10 10",
                 }]),
                 Step::Await(vec![Check::Query(
                     "select max(bar) as bar from table_1",
@@ -1723,7 +1723,8 @@ mod tests {
                 Step::Ingest(vec![TestSequencedEntry {
                     sequencer_id: 0,
                     sequence_number: 2,
-                    lp: "table_1,tag_partition_by=a bar=20 0",
+                    // same time as first entry
+                    lp: "table_1,tag_partition_by=a,tag=2 bar=20 10",
                 }]),
                 Step::Await(vec![Check::Query(
                     "select max(bar) as bar from table_1",
@@ -1733,7 +1734,7 @@ mod tests {
                 Step::Ingest(vec![TestSequencedEntry {
                     sequencer_id: 0,
                     sequence_number: 3,
-                    lp: "table_1,tag_partition_by=b bar=30 0",
+                    lp: "table_1,tag_partition_by=b,tag=3 bar=30 30",
                 }]),
                 Step::Await(vec![Check::Query(
                     "select max(bar) as bar from table_1",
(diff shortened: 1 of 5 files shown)


server/src/db/replay.rs
```diff
@@ -1838,15 +1821,103 @@ mod tests {
...
-                Step::Assert(vec![Check::Partitions(vec![
-                    ("table_1", "tag_partition_by_a"),
-                    ("table_1", "tag_partition_by_b"),
-                ])]),
+                Step::Assert(vec![
+                    Check::Partitions(vec![
+                        ("table_1", "tag_partition_by_a"),
+                        ("table_1", "tag_partition_by_b"),
+                    ]),
+                    Check::Query(
```
tustvold, line 1829: I don't understand what happened to "table_1,tag_partition_by=a bar=10 0" and "table_1,tag_partition_by=b bar=30 0" .

As far as I understand the test they should have been persisted?

tustvold (approved):
Tentatively approved, but I am very confused by the tests - they look wrong to me?

server/src/db/replay.rs
crepererum: same timestamp, aka override / dup

tustvold: Oh - brain still booting up - yes we discussed this yesterday :facepalm:

crepererum: simplified

crepererum:
Test got simplified and should now be easier to understand.

tustvold (approved):
Much clearer now :smile:

PR #2219, opened 2021-08-06
refactor: Move DatabaseStore out of query crate

alamb:
refactor: Move DatabaseStore into server crate

# Rationale:
As noted by @tustvold on https://github.com/influxdata/influxdb_iox/pull/2203/files#r684070187 this trait isn't by the query crate, it is used to mock out the server in gRPC service tests.

# Changes:
1. Move DatabaseStore trait into Server
2. Move TestDatabaseStore into tests of storage/service.rs  where it is used

There is no intended change in functionality


server/src/lib.rs
```diff
@@ -259,6 +259,31 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+#[async_trait]
+/// Storage for `Databases` which can be retrieved by name
+pub trait DatabaseStore: std::fmt::Debug + Send + Sync {
```
tustvold, line 264: Perhaps this should be QueryDatabaseStore to make clear it is for use by queries?

tustvold (approved):
This looks fine to me, FWIW I was fine with this living in query, my comment was more that the API should be revisited to reflect the changes to the server crate. db() returning an Option for instance, using DatabaseName, etc...

server/src/lib.rs
alamb: Interestingly it isn't used by the query crate (which was the reason to move it out of query in the first place).

It is  used to write mocked tests of the gRPC storage interface. 

```
/Users/alamb/Software/influxdb_iox/server/src/lib.rs
1144: impl<M> DatabaseStore for Server<M>
/Users/alamb/Software/influxdb_iox/src/influxdb_ioxd/rpc/management.rs
12: use server::{ApplicationState, ConnectionManager, DatabaseStore, Error, Server};
/Users/alamb/Software/influxdb_iox/src/influxdb_ioxd/rpc/storage.rs
19: use server::DatabaseStore;
24: struct StorageService<T: DatabaseStore> {
29: pub fn make_server<T: DatabaseStore + 'static>(
/Users/alamb/Software/influxdb_iox/src/influxdb_ioxd/rpc/storage/service.rs
  32: use server::DatabaseStore;
 215:     T: DatabaseStore + 'static,
 862:     T: DatabaseStore + 'static,
 901:     T: DatabaseStore + 'static,
 956:     T: DatabaseStore + 'static,
1008:     T: DatabaseStore + 'static,
1065:     T: DatabaseStore + 'static,
1131:     T: DatabaseStore + 'static,
2900:     impl DatabaseStore for TestDatabaseStore {
```

PR #2135, opened 2021-07-28
feat: enable row timestamp metrics with environment variable

tustvold:
These metrics are very high-cardinality, 60 per table, which can causes issues for some monitoring systems. This PR makes it so that they are only enabled on a per-table basis with an environment variable

PR #2101, opened 2021-07-23
feat: add row timestamp metrics

tustvold:
This adds a "histogram" metric for the distribution of row timestamps ingested per-table. This will help better understand the distribution of the data that is landing, which in turn will help improve the lifecycle policy.

Issue #1355, opened 2021-04-29
Catalog Lock Contention Metrics


tustvold:
The catalog contains a number of RwLocks that are used to mediate access to the chunks within a database.

Critically writes to the MUB take place under the catalog chunk's write lock, and snapshots take place under the catalog chunk's read lock. Neither of these operations are necessarily very fast, and therefore there is a high chance of lock contention, particularly as for high-throughput tables there are likely multiple concurrent writes to the same table and therefore potentially the same chunk.

As these locks are not async (which is probably still correct) lock contention can potentially starve out tokio worker threads, which would be bad. 

We should get some metrics on how much time is spent waiting to acquire read and write locks. This could take the form of two AtomicUsize for each lock, storing the number of milliseconds spent waiting to acquire the lock as read and write respectively. These could then be exposed as monotonic gauges to prometheus, or something similar.

alamb:
This is a good idea -- I think it may be worth gathering this as part of the "handle tools write workload" epic

e-dard:
This makes sense to do but it pains me that we have to write code to do it rather than being able to simply take a blocking profile of the program :-)

tustvold:
> simply take a blocking profile of the program

They are OS-level locks, so would show up in a standard perf profile, however, the advantage to instrumenting would be that we could have the metrics all the time :grinning:

PR #2222, opened 2021-08-06
docs: document Metrics

NGA-TRAN:
Closes #


NGA-TRAN:
@alamb @carols10cents  @e-dard @tustvold @crepererum @mkmik @jacobmarble 
These are metrics I believe you have created. Can you help review them?
I use a few of these metrics to create I[Ox Metrics dashboard](https://us-west1-1.gcp.cloud2.influxdata.com/orgs/2f5ec5d98a8d6526/dashboards/07f20d557feae000?lower=now%28%29%20-%203h) that I think will be useful for performance measurement as well as discovering issues

docs/metrics.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Chunks
+| Metric name |  Code Name | Description |
+| --- | --- | --- |
+| catalog_chunks_mem_usage_bytes | memory_metrics | Total memory usage by chunks (MUB, RUB, OS statistics) |
+| catalog_loaded_chunks | chunk_storage | Total number of chunks (MUB, RUB, RUBandOS) for each table |
+| catalog_loaded_rows | row_count | Total number of rows (MUB, RUB, RUBandOS) for each table |
+| catalog_lock_total | ? | ? |
+| catalog_lock_wait_seconds_total | ? | ? |
+| ? | table_lock_tracker| ? |
```
tustvold, line 34: I don't think this is a metric? It should being reported under `catalog_lock_total` and `catalog_lock_wait_seconds_total`.

These are metrics on the number of each type of lock acquired, and the total time spent blocked on acquiring them. They were added in #1355

docs/metrics.md
```diff
@@ -0,0 +1,69 @@
...
+| ? | partition_lock_tracker | ? |
+| ? | chunk_lock_tracker | ? |
+| ? | timestamp_histogram| ? |
```
tustvold, line 37: This is a breakdown of the timestamp distribution. It is only published for tables contained in a comma separate environment variable `INFLUXDB_IOX_ROW_TIMESTAMP_METRICS`.

It was added in #2101 and the environment flag was added in #2135

docs/data_management.md
```diff
@@ -89,13 +89,19 @@ From drawings/data_lifecycle.monopic
                  └─────────────────────┘   └───────────────────────┘    └────────────────────┘
                             │                                                      │
                             │                                                      │
-         ┌──────────────────┼───────────────────┐                      ┌───────────┴────────────┐
```
tustvold, line 92: This section probably needs a more concrete update in light of the Kafka-ification, the segment store no longer exists, but out of scope for this PR

tustvold (approved):
Thank you for doing this @NGA-TRAN , I've added comments on the metrics that I'm familiar with, mainly those under the Chunks heading. Hopefully we can keep this document up to date as we add/remove metrics

docs/metrics.md
```diff
@@ -0,0 +1,69 @@
...
+| catalog_chunks_total | chunk_state | Number of chunks in different states (open, closed, compacting, compacted, writing_os, rub_and_os, ...)) |
```
tustvold, line 38: FWIW this metric is probably going to be removed - it is superceded by catalog_loaded_chunks which is also easier to interpret

PR #2224, opened 2021-08-08
feat: blocking Freezable

tustvold:
Whilst writing #2203 an annoying limitation of Freezable came up, namely it didn't support waiting for a handle.

This resolves this, by using a tokio::sync::Mutex in place of the Arc shenanigans. There is likely an incredibly minor additional performance cost of the additional synchronisation, but this is almost certainly irrelevant.

As an added bonus, freeze() no longer requires exclusive write access.


internal_types/src/freezable.rs
```diff
@@ -64,49 +64,128 @@ use std::sync::{Arc, Weak};
...
     fn deref(&self) -> &Self::Target {
-        &self.0
+        &self.payload
+    }
+}
+
+/// The `WriteGuard` provides mutable access to the `Freezable` data whilst ensuring
+/// that it remains unfrozen for the duration of the mutable access
+#[derive(Debug)]
+pub struct WriteGuard<'a, T> {
```
tustvold, line 168: It is now possible to freeze a Freezable without a borrow (mutable or otherwise) on Freezable, by virtue of awaiting the future returned by `Freezable::freeze`. This is necessary to make the future actually useful, as we need to be able to await it without holding locks.

Unfortunately this means we now need a write guard to ensure that Freezable isn't frozen during a mutable borrow.

internal_types/src/freezable.rs
```diff
@@ -64,49 +64,128 @@ use std::sync::{Arc, Weak};
...
 
     /// Returns a `FreezeHandle` that prevents modification
     /// of the contents of `Freezable` until it is dropped
     ///
     /// Returns None if the object is already frozen
-    pub fn try_freeze(&mut self) -> Option<FreezeHandle<T>> {
-        // Verify exclusive
-        self.get_mut()?;
-        Some(FreezeHandle(Arc::downgrade(&self.0)))
+    pub fn try_freeze(&self) -> Option<FreezeHandle> {
```
tustvold, line 117: No longer need to rely on &mut to provide exclusivity :tada:

internal_types/Cargo.toml
```diff
@@ -13,6 +13,8 @@ indexmap = "1.6"
 itertools = "0.10.1"
 observability_deps = { path = "../observability_deps" }
 snafu = "0.6"
+tokio = { version = "1.0", features = ["sync"] }
```
tustvold, line 16: I'm not sure if it is better to just use the same feature flags on all modules, or the minimal set on each... The former saves local recompilation, but the latter is more correct. Perhaps @carols10cents has some thoughts?

internal_types/src/freezable.rs
```diff
@@ -64,49 +64,128 @@ use std::sync::{Arc, Weak};
...
+        let guard = Arc::clone(&self.lock).try_lock_owned().ok()?;
```
alamb, line 118: I always get confused when I see `.ok()?` and have to mentally remind myself that it returns the `None` if `is_none`, rather than silently ignoring the error

internal_types/src/freezable.rs
```diff
@@ -64,49 +64,128 @@ use std::sync::{Arc, Weak};
...
+        Some(FreezeHandle(guard))
+    }
+
+    /// Returns a future that resolves to a FreezeHandle
+    pub fn freeze(&self) -> impl std::future::Future<Output = FreezeHandle> {
```
alamb, line 123: this is cool

internal_types/src/freezable.rs
alamb: I actually think the use of a `WriteGuard` is good and makes it more explicit what is happening

alamb (approved):
The code looks reasonable. Is the idea to use this new feature in `Database` et al as a follow on PR?

I also wonder if there are some tests in order

tustvold:
Added some further tests, along with the ability to create a `FreezeHandle` from a `WriteGuard`, to allow multi-stage transactions

internal_types/src/freezable.rs
```diff
@@ -127,4 +210,96 @@ impl<T: std::fmt::Display> std::fmt::Display for Freezable<T> {
...
+
+        let mut f = Freezable::new(1);
+
+        let freeze_fut = f.freeze();
+        pin_mut!(freeze_fut);
+
+        let write_guard = f.get_mut().unwrap();
+
+        // Shouldn't resolve whilst write guard active
+        assert!(freeze_fut.poll_unpin(&mut cx).is_pending());
```
alamb, line 234: this is fancy

internal_types/src/freezable.rs
```diff
@@ -127,4 +210,96 @@ impl<T: std::fmt::Display> std::fmt::Display for Freezable<T> {
...
+                    let mut guard = locked.unfreeze(handle);
+
+                    assert_eq!(*guard, 0);
+                    *guard = i;
+
+                    guard.freeze()
+                };
+
+                // Do async work
+                tokio::time::sleep(tokio::time::Duration::from_nanos(10)).await;
```
alamb, line 290: making this sleep some random number of nanos is also a classic way to augment fuzz testing with extra entropy

alamb (commented):
The changes look very nice

PR #2227, opened 2021-08-09
feat: remove create_database_lock

tustvold:
This was a hack introduced in #2203 to workaround the inability to wait for a FreezeHandle. This functionality was added in #2224 and so the hack can be removed. This has the advantage of making the code more future-proof, as if we ever add transitions that de-initialize a server, it will continue to work correctly


server/src/lib.rs
```diff
@@ -552,7 +547,11 @@ where
         let db_name = rules.name.clone();
         let object_store = self.shared.application.object_store().as_ref();
 
-        let (server_id, create_database_lock) = {
+        // Wait for exclusive access to mutate server state
+        let handle_fut = self.shared.state.read().freeze();
```
tustvold, line 551: These must be separate so that the lock is not held across the await point (which results in a compile error)

server/src/lib.rs
```diff
@@ -586,10 +579,7 @@ where
         let database = {
             let mut state = self.shared.state.write();
 
-            // Have exclusive lock on state - can drop database creation lock
-            std::mem::drop(guard);
-
-            let mut state = state.get_mut().expect("no transaction in progress");
+            let mut state = state.unfreeze(handle);
```
alamb, line 582:
```suggestion
            // Have exclusive lock on state - can drop database creation lock
            let mut state = state.unfreeze(handle);
```

alamb (approved):
Very cool

server/src/lib.rs
tustvold: The database creation lock no longer exists. `unfreeze()` here is returning a WriteGuard that preserves the exclusivity

PR #2060, opened 2021-07-20
fix: update PersistenceWindows on rules update (#2018)

tustvold:
Closes #2018

PR #2229, opened 2021-08-09
fix: always remember max seen sequ. numbers during replay

### LINKED ISSUE #2215: Replay Restore Fully Persisted Sequencer
It is assumed that consecutive PartitionCheckpoints are monotonically increasing that is:

* Later PartitionCheckpoints have the same or larger min/max seqno for each sequencer id than previous
* Later PartitionCheckpoints have the same or more sequencer ids than previous

This last point is currently not necessarily the case. Consider a scenario with two sequencer ids 0 and 1, and a partition that is fully persisted in sequencer id 0. On replay only writes from sequencer id 1 will be replayed to the partition. If the partition continues to not receive writes from sequencer id 0, it may persist a PartitionCheckpoint having never "seen" a write from sequencer id 0.

At the end of replay, all persistence windows should be "seeded" with the maximum values from their PartitionCheckpoint to ensure that future checkpoints are consistent. (NB: The DatabaseCheckpoint might contain sequencer ids that have never written to this partition, and so it must be the PartitionCheckpoint that is used)

crepererum:
Do not forget max seen sequence numbers for partition-sequencer
combinations that can be skipped during replay.

Fixes #2215.

### DIFF
--- a/server/src/db/replay.rs
+++ b/server/src/db/replay.rs
@@ -7,7 +7,10 @@ use std::{
 use entry::{Sequence, TableBatch};
 use futures::TryStreamExt;
 use observability_deps::tracing::info;
-use persistence_windows::{checkpoint::ReplayPlan, min_max_sequence::OptionalMinMaxSequence};
+use persistence_windows::{
+    checkpoint::ReplayPlan, min_max_sequence::OptionalMinMaxSequence,
+    persistence_windows::PersistenceWindows,
+};
 use snafu::{ResultExt, Snafu};
 use write_buffer::config::WriteBufferConfig;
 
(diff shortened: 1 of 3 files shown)


server/src/db/replay.rs
```diff
@@ -231,6 +234,32 @@ pub async fn perform_replay(db: &Db, replay_plan: &ReplayPlan) -> Result<()> {
...
+                    .expect("replay plan inconsistent");
+
+                match partition.persistence_windows_mut() {
+                    Some(windows) => {
+                        windows.mark_seen_and_persisted(partition_checkpoint);
+                    }
+                    None => {
+                        let mut windows = PersistenceWindows::new(
+                            partition.addr().clone(),
+                            late_arrival_window,
```
alamb, line 254: Would it cause a problem if the `db_rules` change between when a database is shutdown and when replay is performed? If this type of out of band change would cause a problem, I don't think IOx has to gracefully handle it, but it would be nice if IOx could detect the problem an error quickly (rather than some time later when some other inconsistency surfaces)

server/src/db/replay.rs
```diff
@@ -1925,6 +1954,136 @@ mod tests {
...
+                            "| bar | tag_partition_by | time                           |",
+                            "+-----+------------------+--------------------------------+",
+                            "| 10  | b                | 1970-01-01T00:00:00.000000010Z |",
+                            "| 20  | a                | 1970-01-01T00:00:00.000000020Z |",
+                            "| 30  | a                | 1970-01-01T00:00:00.000000030Z |",
+                            "| 40  | b                | 1970-01-01T00:00:00.000000040Z |",
+                            "+-----+------------------+--------------------------------+",
+                        ],
+                    ),
+                    // no additional chunk for partition a was created
```
alamb, line 2068: 👍  test makes sense to me

alamb (approved):
The framework for writing replay tests keeps on showing its value 👍  @crepererum

server/src/db/replay.rs
crepererum: The persistence windows are a in-memory concept (i.e. they are not persisted). So updating during shutdown isn't a problem. Updating while the DB is online was done in #2060.

PR #2231, opened 2021-08-09
refactor: do not construct replay plan when skipping replay

crepererum:
Up until now we only skipped the execution of the replay plan, not its
construction. The replay plan construction has some bugs left, so let's
move this part behind the toggle as well.


tustvold (approved):
LGTM

alamb (commented):
👍

PR #2232, opened 2021-08-09
feat: Server own background worker

tustvold:
Rejigs `Server` to own its own background worker, to match `Database`.

Also renamed background loop to background worker, as it isn't actually a single loop and we weren't being consistent with terminology (likely entirely my fault).

Relates to #2186


server/src/lib.rs
```diff
@@ -635,90 +689,6 @@ where
...
-    /// Loads the database configurations based on the databases in the
-    /// object store. Any databases in the config already won't be
-    /// replaced.
-    ///
-    /// This requires the serverID to be set.
-    ///
-    /// It will be a no-op if the configs are already loaded and the server is ready.
-    ///
-    /// TODO: Make server own background loop and make this private
-    pub async fn maybe_initialize_server(&self) {
```
tustvold, line 647: This method is just moved lower down in the file, with only very minor modification

server/src/lib.rs
```diff
@@ -527,6 +550,37 @@ where
         self.shared.state.read().initialized().is_ok()
     }
 
+    /// Triggers shutdown of this `Server`
+    pub fn shutdown(&self) {
```
tustvold, line 554: This is pretty much an identical setup to in `Database`

server/src/lib.rs
```diff
@@ -1724,49 +1781,33 @@ mod tests {
             "expected:\n{:#?}\n\nactual:{:#?}\n\n",
             expected, actual
         );
-
-        // ensure that we don't leave the server instance hanging around
```
tustvold, line 1728: We aren't consistently waiting for server shutdown in most tests, so I think I'll just follow this up with a Drop method on Server and Database, that warn and trigger shutdown if it hasn't already been

server/src/lib.rs
```diff
@@ -360,6 +364,9 @@ pub struct Server<M: ConnectionManager> {
 
     metrics: Arc<ServerMetrics>,
 
+    /// Future that resolves when the background worker exits
```
crepererum, line 367: It took me a while to realize that we have a server-level background worker now (I've missed that change).

server/src/lib.rs
```diff
@@ -369,9 +376,17 @@ pub struct Server<M: ConnectionManager> {
 
 #[derive(Debug)]
 struct ServerShared {
+    /// A token that is used to trigger shutdown of the background worker
+    shutdown: CancellationToken,
```
crepererum, line 380: Is this specific token only used for the server-level worker or will it be cloned and also be used for the DB-level workers?

crepererum (approved):
Clean design :+1:

server/src/lib.rs
tustvold: We've always had one, it's what drains the tracker registry, etc... but it was spawned by the main influxdb_iox method so it was perhaps not obvious how it plumbed in

server/src/lib.rs
tustvold: Only the server - this was a complication I removed as in practice it wouldn't make a huge difference. The server background worker is the only thing that waits for the Database workers to finish up, and so it might as well trigger the shutdown of the Databases and avoid the complexity of plumbing through CancellationTokens (which I thought made for a bit of a meh API). If the server background worker is dead, nothing is going to wait on the Database's to shutdown anyway, regardless of whether they're told to shutdown or not.

server/src/database.rs
```diff
@@ -100,7 +100,7 @@ impl Database {
             state_notify: Default::default(),
         });
 
-        let handle = tokio::spawn(background_loop(Arc::clone(&shared)));
+        let handle = tokio::spawn(background_worker(Arc::clone(&shared)));
```
alamb, line 103: I like the consistent terminology

alamb: I found a few more remnants of `background loop`: 

```grep
-*- mode: grep; default-directory: "~/Software/influxdb_iox/" -*-
Grep started at Mon Aug  9 13:45:57

rg -n -H -i --no-heading -e 'background loop' $(git rev-parse --show-toplevel || pwd)
/Users/alamb/Software/influxdb_iox/server/src/utils.rs:81:        // make background loop spin a bit faster for tests
/Users/alamb/Software/influxdb_iox/server/src/db.rs:1684:        // We need to do that BEFORE shutting down the background loop because gauges would be dropped and resetted otherwise
/Users/alamb/Software/influxdb_iox/server/src/db/replay.rs:422:        /// Background loop is started as well but neither persistence nor write buffer reads are allowed until
/Users/alamb/Software/influxdb_iox/server/src/db/replay.rs:443:        /// Wait that background loop generates desired state (all checks pass).

```

server/src/lib.rs
alamb: Is it just the shutdown of the background worker, or could we say that this `CancellationToken` is used to trigger shutdown of the entire datatabase?

If it is related to just the background worker, perhaps we should call it `background_worker_shutdown` or something more specific to the background worker

server/src/lib.rs
```diff
@@ -1024,49 +994,140 @@ where
...
+/// Background worker function for the server
+async fn background_worker(shared: Arc<ServerShared>) {
+    info!("started server background worker");
 
-        let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(1));
+    let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(1));
 
-        // TODO: Move out of Server background loop
-        let job_registry = self.shared.application.job_registry();
+    // TODO: Move out of Server background worker
```
alamb, line 1006: to where?

server/src/lib.rs
```diff
@@ -1024,49 +994,140 @@ where
...
+        job_registry.reclaim();
 
-            // TODO: Trigger termination if database worker finishes
+        // TODO: Trigger termination if database worker finishes
 
-            tokio::select! {
-                _ = interval.tick() => {},
-                _ = shutdown.cancelled() => break
-            }
+        tokio::select! {
```
alamb, line 1015: should this also check the `notify` for state changes too? Or is it assumed that this loop is the only thing that will change `state`s of databases and thus doesn't need to wait for its own notification? Or maybe I am misunderstanding

alamb (approved):
Looks like a solid improvement to me

server/src/lib.rs
tustvold: I guess it could have its own background loop, but yeah, probably not worth it

server/src/lib.rs
tustvold: We could use the notify to edge trigger this, will add to a follow on PR

tustvold:
Going to get this in and address feedback in a follow on PR, I have a PR that builds on top of this one I'd like to get up

PR #2233, opened 2021-08-09
fix: incorrect speculation of post-persist sequence ranges

### LINKED ISSUE #2206: FlushHandle Generates Incorrect PartitionCheckpoint
As part of a persist operation, a `FlushHandle` is obtained from the `PersistenceWindow`, this in turn is used to generate the `PartitionCheckpoint` that is written as part of this operation. This must describe the state of the system speculatively, i.e. what the state of the system will be after the persist operation, but must defer the actual modification of the `PersistenceWindow` until after the persist operation completes, at which point `FlushHandle::flush` is called.

On flush all closed windows that were present at the time of the `FlushHandle`'s creation, have their min row timestamp updated to be one nanosecond greater than the flush timestamp (as those rows must have been persisted). 

The bug concerns what happens when the chosen flush timestamp exceeds the max of a closed (but not persistable) window, in this case `FlushHandle::flush` drops the window entirely. 

Whilst correct this can cause a mismatch between the state of the system described by the PartitionCheckpoint, and the actual state after the persistence. Namely it is possible for the PartitionCheckpoint to think there are remaining unpersisted writes when there aren't. 

Subsequent persist operations to other partitions, will encode this in the DatabaseCheckpoint that they generate. This then causes a sanity check to fail on restart, as the incorrect PartitionCheckpoint will say that writes to that partition remain to be persisted, but the last DatabaseCheckpoint will say we've fully persisted up to a min sequence number that is greater than the min in the PartitionCheckpoint.

When generating the sequence number for the PersistenceWindow, the system should take into account that any "covered" closed windows will also be flushed entirely.

crepererum:
This fixes an edge case where the speculated sequence ranges that can be
obtained from flush handles do not account for overlapping windows. The
symptom being that the resulting partition checkpoint marked sequence
numbers as unpersisted that where already persisted.

Fixes #2206.

### DIFF
--- a/persistence_windows/src/persistence_windows.rs
+++ b/persistence_windows/src/persistence_windows.rs
@@ -2,11 +2,12 @@
 use std::{
     collections::{btree_map::Entry, BTreeMap, VecDeque},
     num::NonZeroUsize,
+    ops::Deref,
     sync::Arc,
     time::{Duration, Instant},
 };
 
-use chrono::{DateTime, TimeZone, Utc};
+use chrono::{DateTime, Utc};
 
 use data_types::{partition_metadata::PartitionAddr, write_summary::WriteSummary};
 use entry::Sequence;
@@ -238,10 +239,15 @@ impl PersistenceWindows {
         }
     }
 
+    /// Returns the sequence number range of unpersisted writes described by this instance.
+    pub fn sequencer_numbers(&self) -> BTreeMap<u32, OptionalMinMaxSequence> {
+        self.sequencer_numbers_inner(false)
+    }
+
     /// Returns the sequence number range of unpersisted writes described by this instance
     ///
-    /// Can optionally skip the persistable window if any
-    pub fn sequencer_numbers(
+    /// Can optionally skip the persistable window if any.
+    fn sequencer_numbers_inner(
         &self,
         skip_persistable: bool,
     ) -> BTreeMap<u32, OptionalMinMaxSequence> {
@@ -249,9 +255,9 @@ impl PersistenceWindows {
             Default::default()
         }
 
-        let skip = match skip_persistable {
-            true if self.persistable.is_some() => 1,
-            _ => 0,
+        let (skip, min_time) = match (skip_persistable, self.persistable.deref()) {
+            (true, Some(persistable)) => (1, Some(new_min_time_from_persistable(persistable))),
+            _ => (0, None),
         };
 
         self.max_sequence_numbers
@@ -261,7 +267,14 @@ impl PersistenceWindows {
                 let window = self
                     .windows()
                     .skip(skip)
-                    .filter_map(|window| window.sequencer_numbers.get(sequencer_id))
+                    .filter_map(|window| {
+                        if let Some(min_time) = min_time {
+                            if window.max_time < min_time {
+                                return None;
+                            }
+                        }
+                        window.sequencer_numbers.get(sequencer_id)
+                    })
                     .next();
 
                 let min = window.map(|window| {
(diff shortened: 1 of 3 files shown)


persistence_windows/src/persistence_windows.rs
```diff
@@ -398,6 +411,10 @@ impl PersistenceWindows {
     }
 }
 
+fn new_min_time_from_persistable(persistable: &Window) -> DateTime<Utc> {
```
alamb, line 414: Can we add a doc comment please (explaining the rationale for the return value of this function)?

persistence_windows/src/persistence_windows.rs
```diff
@@ -238,20 +239,25 @@ impl PersistenceWindows {
         }
     }
 
+    /// Returns the sequence number range of unpersisted writes described by this instance.
```
alamb, line 242: I like how the public API is now less confusing (can't pass in `skip`)

persistence_windows/src/persistence_windows.rs
```diff
@@ -1472,4 +1491,41 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    fn test_regression_2206() {
```
alamb, line 1496: "regression 2206" -- sounds like a science fiction (or maybe horror) movie title 😆

persistence_windows/src/persistence_windows.rs
```diff
@@ -1472,4 +1491,41 @@ mod tests {
...
+            now,
+        );
+
+        // persist
+        let handle = w.flush_handle(now).unwrap();
+        let ckpt = handle.checkpoint();
+        w.flush(handle);
+
+        // speculated checkpoint should be correct
+        let ckpt_sequencer_numbers: BTreeMap<_, _> = ckpt.sequencer_numbers_iter().collect();
```
alamb, line 1528: 👍

server/src/db/replay.rs
```diff
@@ -1933,11 +1924,6 @@ mod tests {
         }
         .run()
         .await;
-
-        assert_contains!(
```
alamb, line 1937: Maybe we should change these asserts to `assert_not_contains!`? But on the other hand, given there is a regression test now specifically for this case, perhaps it is not needed.

alamb (approved):

persistence_windows/src/persistence_windows.rs
```diff
@@ -238,20 +239,25 @@ impl PersistenceWindows {
...
     ) -> BTreeMap<u32, OptionalMinMaxSequence> {
         if self.is_empty() {
             Default::default()
         }
 
-        let skip = match skip_persistable {
-            true if self.persistable.is_some() => 1,
-            _ => 0,
+        let (skip, min_time) = match (skip_persistable, self.persistable.deref()) {
+            (true, Some(persistable)) => (1, Some(new_min_time_from_persistable(persistable))),
```
tustvold, line 259: FWIW it would also be correct to just call this flush_time, and drop any windows with a max time <= it, instead of the +1 dance which only exists for updating the min time in the windows. Doesn't matter hugely though

tustvold (approved):
LGTM :+1:

persistence_windows/src/persistence_windows.rs
crepererum: done

server/src/db/replay.rs
crepererum: done

persistence_windows/src/persistence_windows.rs
crepererum: done

PR #2234, opened 2021-08-09
feat: background worker panic handling (#2091)

### LINKED ISSUE #2091: Compaction Panic Stalls Lifecycle
It was observed that a database suffering from #2090 would cease to perform any further compactions. This would suggest that the background worker loop aborted.

There are a couple of takeaways from this:

* The main loop should print a prominent error and exit if the server loop exits (it already does, but it should be an error not info)
* The server background loop should monitor the loops of its databases to make sure they haven't exited - it should print an error and try to restart it if so
* The Db background loop should spawn potentially panic-ing work, to reduce the risk of it bringing down the loop

tustvold:
Closes #2091

This adds logic to trigger shutdown on a background worker panic, along with some tests :tada:

The way these tests work is not especially elegant, but it was the simplest way I could think to inject panics.


server/src/database.rs
```diff
@@ -113,8 +113,8 @@ impl Database {
     }
 
     /// Waits for the background worker of this `Database` to exit
-    pub async fn join(&self) -> Result<(), Arc<JoinError>> {
-        self.join.clone().await
+    pub fn join(&self) -> impl Future<Output = Result<(), Arc<JoinError>>> {
```
tustvold, line 116: This just means that we can "await" on this completing in a task without having to keep an Arc<Db> around

server/src/lib.rs
```diff
@@ -463,6 +463,50 @@ struct ServerStateInitialized {
...
+                config,
+            ))),
+            hashbrown::hash_map::Entry::Occupied(_) => {
+                return Err(Error::DatabaseAlreadyExists {
+                    db_name: config.name.to_string(),
+                })
+            }
+        };
+
+        // Spawn a task to monitor the Database and trigger server shutdown if it fails
```
tustvold, line 488: This is the new logic that "watches" the background workers, and triggers termination if they fail

server/src/lib.rs
```diff
@@ -463,6 +463,50 @@ struct ServerStateInitialized {
...
+        let fut = database.join();
+        let shutdown = shared.shutdown.clone();
+        let _ = tokio::spawn(async move {
+            match fut.await {
+                Ok(_) => info!(%db_name, "server observed clean shutdown of database worker"),
```
tustvold, line 493: Currently this only occurs on server shutdown, but given we may in the future have per-Database shutdown it made sense for this not to be an error

server/src/utils.rs
```diff
@@ -194,3 +194,30 @@ pub fn count_object_store_chunks(db: &Db) -> usize {
         })
         .count()
 }
+
+static PANIC_DATABASE: once_cell::race::OnceBox<parking_lot::Mutex<hashbrown::HashSet<String>>> =
```
tustvold, line 198: This is the slightly meh way that panics can be induced in supposedly panic-free codepaths

src/influxdb_ioxd.rs
```diff
@@ -101,29 +101,7 @@ async fn wait_for_signal() {
...
-    // Install custom panic handler and forget about it.
-    //
-    // This leaks the handler and prevents it from ever being dropped during the
-    // lifetime of the program - this is actually a good thing, as it prevents
-    // the panic handler from being removed while unwinding a panic (which in
-    // turn, causes a panic - see #548)
-    let f = SendPanicsToTracing::new();
-    std::mem::forget(f);
-
+async fn make_application(config: &Config) -> Result<Arc<ApplicationState>> {
```
tustvold, line 104: None of this logic is altered, at least not intentionally, but it is broken up to make it easier to test

src/influxdb_ioxd.rs
```diff
@@ -207,7 +228,7 @@ pub async fn main(config: Config) -> Result<()> {
     .fuse();
     info!(bind_address=?bind_addr, "HTTP server listening");
 
-    info!(git_hash, "InfluxDB IOx server ready");
+    info!("InfluxDB IOx server ready");
```
tustvold, line 231: This is already being logged on startup so it seemed redundant to plumb it through

src/influxdb_ioxd.rs
```diff
@@ -133,22 +111,28 @@ pub async fn main(config: Config) -> Result<()> {
...
+) -> Arc<AppServer<ConnectionManager>> {
     let server_config = ServerConfig {
-        remote_template: config.remote_template.map(RemoteTemplate::new),
+        remote_template: config.remote_template.clone().map(RemoteTemplate::new),
         wipe_catalog_on_error: config.wipe_catalog_on_error.into(),
         skip_replay_and_seek_instead: config.skip_replay_and_seek_instead.into(),
     };
 
-    if config.grpc_bind_address == config.http_bind_address {
+    if config.grpc_bind_address == config.http_bind_address && config.grpc_bind_address.port() != 0
```
tustvold, line 134: I'm not sure why you would ever use a port of 0 outside of a test, but it is very useful for writing said tests :grin: 

(it is a random OS assigned port)

server/src/database.rs
```diff
@@ -260,6 +260,13 @@ async fn background_worker(shared: Arc<DatabaseShared>) {
 
         info!(db_name=%shared.config.name, "database finished initialization - starting Db worker");
 
+        crate::utils::panic_test(|| {
```
alamb, line 263: 👍  this is good

server/src/utils.rs
alamb: I will accept "meh" as a valid description when I see a more "elegant" want to accomplish this :)

I have seen things similar to this that insert waits (basically a Barrier) to line up hard to reproduce race conditions as well.

server/src/utils.rs
```diff
@@ -194,3 +194,30 @@ pub fn count_object_store_chunks(db: &Db) -> usize {
...
+}
+
+#[inline]
+/// Used to induce panics in tests
+///
+/// This needs to be public and not test-only to allow usage by other crates, unfortunately
+/// this means it currently isn't stripped out of production builds
+///
+/// The runtime cost if no keys are registered is that of a single atomic load, and so
+/// it is deemed not worth the feature flag plumbing necessary to strip it out
```
alamb, line 214: 💯 

Another potential rational for stripping this code out via feature flag is potential security / denial of service attacks, but that seems unlikely given how it is used / triggered.

src/influxdb_ioxd.rs
```diff
@@ -620,4 +642,158 @@ mod tests {
...
+        pin_mut!(serve_fut);
+
+        futures::select! {
+            _ = serve_fut => panic!("serve shouldn't finish"),
+            _ = tokio::time::sleep(tokio::time::Duration::from_millis(10)).fuse() => {}
+        }
+
+        server.shutdown();
+
+        futures::select! {
```
alamb, line 665: do I read this correctly that this basically says `serve_fut` must resolve within 5 seconds otherwise the test will panic?

src/influxdb_ioxd.rs
```diff
@@ -620,4 +642,158 @@ mod tests {
...
+            "--grpc-bind",
+            "127.0.0.1:0",
+        ]);
+        let application = make_application(&config).await.unwrap();
+        let server = make_server(Arc::clone(&application), &config);
+
+        let serve_fut = serve(config, application, Arc::clone(&server)).fuse();
+        pin_mut!(serve_fut);
+
+        futures::select! {
```
alamb, line 686: I don't understand this test -- it looks to me like it simply ensures that the server hasn't come up within 10ms but doesn't actually prevent it from doing so. That seems like a race condition but I admit to not being fully versed in the futures logic here. 

I wonder why it is even important to wait on `serve_fut` -- wouldn't it make sense to not even try to serve so as to ensure it never resolves?

src/influxdb_ioxd.rs
```diff
@@ -620,4 +642,158 @@ mod tests {
...
+            "127.0.0.1:0",
+        ]);
+        let application = make_application(&config).await.unwrap();
+        let server = make_server(Arc::clone(&application), &config);
+        server.wait_for_init().await.unwrap();
+
+        let serve_fut = serve(config, application, Arc::clone(&server)).fuse();
+        pin_mut!(serve_fut);
+
+        futures::select! {
```
alamb, line 722: I would have thought this test would wait until `serve_fut` was actually started so that the test modeled a panic in the background worker of a running database

src/influxdb_ioxd.rs
```diff
@@ -620,4 +642,158 @@ mod tests {
...
+            .unwrap();
+
+        let other_db = server.database(&other_db_name).unwrap();
+
+        let serve_fut = serve(config, Arc::clone(&application), Arc::clone(&server)).fuse();
+        pin_mut!(serve_fut);
+
+        futures::select! {
+            _ = serve_fut => panic!("serve shouldn't finish"),
+            _ = tokio::time::sleep(tokio::time::Duration::from_millis(10)).fuse() => {}
```
alamb, line 763: same comment as above

server/src/lib.rs
```diff
@@ -463,6 +463,50 @@ struct ServerStateInitialized {
...
+                return Err(Error::DatabaseAlreadyExists {
+                    db_name: config.name.to_string(),
+                })
+            }
+        };
+
+        // Spawn a task to monitor the Database and trigger server shutdown if it fails
+        let fut = database.join();
+        let shutdown = shared.shutdown.clone();
+        let _ = tokio::spawn(async move {
```
alamb, line 491: Is there any reason for `let _ = ...` rather than just `...`?

server/src/lib.rs
```diff
@@ -1010,6 +1049,11 @@ async fn background_worker(shared: Arc<ServerShared>) {
         maybe_initialize_server(shared.as_ref()).await;
         job_registry.reclaim();
 
+        crate::utils::panic_test(|| {
+            let server_id = shared.state.read().initialized().ok()?.server_id;
+            Some(format!("server background worker: {}", server_id))
+        });
+
         // TODO: Trigger termination if database worker finishes
```
alamb, line 1057: I think this TODO may be complete now

alamb (approved):
I have some questions about the tests though that may be mostly related to my (mis)understanding of how the serving futures work. 

Thank you for the comments on the PR -- it makes it much simpler to review

👍

src/influxdb_ioxd.rs
tustvold: Yes

src/influxdb_ioxd.rs
tustvold: It does, this is what `server.wait_for_init()` in the lines above does. This here, ensure that the future has been polled a bit, and not completed, to be sure that the cause of the exit is the panic below and not say some port collision

src/influxdb_ioxd.rs
tustvold: The server id is never set and so it should never initialize - I will add a comment tomorrow

server/src/lib.rs
tustvold: tokio::spawn returns a JoinHandle, which is a future, and at least at some point clippy or the compiler used to complain about how it must be used

src/influxdb_ioxd.rs
alamb: Thank you -- sorry for my confusion

src/influxdb_ioxd.rs
alamb: ah i missed the point that the server id was not set. makes sense

PR #2235, opened 2021-08-09
fix: Remove some unneeded lifetimes

carols10cents:
I can't remember why I added these, but with just a few little clones in
the disk-based object store, this works great!

This is interfering with the object store wrapper for relative paths I'm working on.

### DIFF
--- a/object_store/src/disk.rs
+++ b/object_store/src/disk.rs
@@ -147,32 +147,33 @@ impl ObjectStoreApi for File {
         Ok(())
     }
 
-    async fn list<'a>(
-        &'a self,
-        prefix: Option<&'a Self::Path>,
-    ) -> Result<BoxStream<'a, Result<Vec<Self::Path>>>> {
+    async fn list(
+        &self,
+        prefix: Option<&Self::Path>,
+    ) -> Result<BoxStream<'static, Result<Vec<Self::Path>>>> {
         let root_path = self.root.to_raw();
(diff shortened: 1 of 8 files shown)


alamb (approved):
seems fine to me 👍

carols10cents:
LOL except for that i wasn't actually compiling the object stores and it doesn't actually work yet... one sec 😅

carols10cents:
hoookay this exposed a bug in the cloud_storage crate

object_store/src/disk.rs
```diff
@@ -147,32 +147,33 @@ impl ObjectStoreApi for File {
...
-    async fn list<'a>(
-        &'a self,
-        prefix: Option<&'a Self::Path>,
-    ) -> Result<BoxStream<'a, Result<Vec<Self::Path>>>> {
+    async fn list(
+        &self,
+        prefix: Option<&Self::Path>,
+    ) -> Result<BoxStream<'static, Result<Vec<Self::Path>>>> {
         let root_path = self.root.to_raw();
+        let prefix = prefix.cloned();
```
crepererum, line 155: So the only side-effect is that we need to clone the prefix to move it into the future/stream? That seems fine, considering all the other string IO/copying that will be done during object store IO.

crepererum (approved):

carols10cents:
I think I'm going to work around this for now instead.

PR #2239, opened 2021-08-10
feat: drain database jobs on shutdown

tustvold:
This makes the Database worker drain wait for its jobs to finish before "completing", this allows cleanly shutting down a Database - which will be important for Database deletion among other things.

In order to test this, I added the ability DummyJob to be spawned with a database name, which comprises the majority of this PR's changes


server/src/database.rs
```diff
@@ -581,3 +607,63 @@ async fn get_store_bytes(
...
+        );
+
+        // Should have failed to load (this isn't important to the test)
+        let err = database.wait_for_init().await.unwrap_err();
+        assert!(matches!(err.as_ref(), InitError::RulesFetch { .. }));
+
+        // Database should be running
+        assert!(database.join().now_or_never().is_none());
+
+        let db_tracker = application
```
crepererum, line 641: Can we call this `job_tracker`? I'm pretty confused by "database", "database.join" and "db_tracker".

server/src/database.rs
```diff
@@ -581,3 +607,63 @@ async fn get_store_bytes(
...
+        // Expect timeout to not complete
+        tokio::time::timeout(tokio::time::Duration::from_millis(1), database.join())
+            .await
+            .unwrap_err();
+
+        // Database task shouldn't have finished yet
+        assert!(!db_tracker.is_complete());
+
+        // Wait for database to finish
+        database.join.await.unwrap();
```
crepererum, line 661: why is this additional "join" necessary, shouldn't this be done by `database.shutdown`?

tustvold: Shutdown doesn't block, it just signals shutdown, perhaps it should be named something else?

server/src/database.rs
tustvold: I renamed it, hopefully it is less confusing now?

server/src/database.rs
crepererum: Then maybe there should be two functions then: `signal_shutdown` and `wait_shutdown_completed`? And then the join handle can be private.

tustvold: Oh! The join handle is private, there is a member function called join(), this is just accidentally using the member directly - will fix

crepererum (approved):
Now it's clearer, thanks!

PR #2241, opened 2021-08-10
feat: shutdown Database and Server on drop

tustvold:
This is mainly important for tests, most of which don't do this. It shouldn't occur in non-test code


server/src/database.rs
```diff
@@ -223,6 +223,20 @@ impl Database {
...
 
+impl Drop for Database {
+    fn drop(&mut self) {
+        let db_name = &self.shared.config.name;
+        if !self.shared.shutdown.is_cancelled() {
+            warn!(%db_name, "database dropped without calling shutdown()");
+            self.shared.shutdown.cancel();
+        }
+
+        if self.join.clone().now_or_never().is_none() {
```
alamb, line 234: very clever

alamb (approved):

crepererum (approved):

Issue #2217, opened 2021-08-06
Panic in `CatalogChunk::new_rub_chunk`:  'inconsistent min/max expected String got NULL', read_buffer/src/table.rs:709:25


alamb:
This is the same reproducer as described by @e-dard  on https://github.com/influxdata/influxdb_iox/issues/2112 but seems to be a differnet underlying cause, so opening a new ticket: 

The following Flux is causing IOx to error in our internal deployment:

```
from(bucket: "apps")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "http_query_response_bytes")
  |> aggregateWindow(every: v.windowPeriod, fn: count, createEmpty: false)
```

This results in the `read_window_aggregate` physical operation being issues, which then needs a `Grouped SeriesSet` plan to execute in order to get the data in the correct series-key based format.

The query never works and Flux displays `An internal error has occurred`.
I can see the following in the IOx logs:

```
level=error msg="thread 'tokio-runtime-worker' panicked at 'UNKNOWN', read_buffer/src/table.rs:709:25" target="panic_logging" location="
panic_logging/src/lib.rs:110" time=1628596458858162477
thread 'tokio-runtime-worker' panicked at 'inconsistent min/max expected String got NULL', read_buffer/src/table.rs:709:25
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: std::panicking::begin_panic_fmt
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:457:5
   2: read_buffer::table::MetaData::to_summary::{{closure}}
   3: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/ops/function.rs:280:13
   4: core::option::Option<T>::map
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/option.rs:489:29
   5: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::next
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/iter/adapters/map.rs:101:9
[...]

PR #2245, opened 2021-08-10
docs: Add/update doccomments in the read_buffer

alamb:
I noticed  some comments while workng on the fix for https://github.com/influxdata/influxdb_iox/issues/2217 so I figured I would put it up as its own PR for review

### DIFF
--- a/read_buffer/src/table.rs
+++ b/read_buffer/src/table.rs
@@ -540,8 +540,10 @@ impl Table {
     }
 }
 
-// TODO(edd): reduce owned strings here by, e.g., using references as keys.
-// Does not implement `Default`; must contain some data.
+/// Table level MetaData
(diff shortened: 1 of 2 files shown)


read_buffer/src/table.rs
```diff
@@ -540,8 +540,10 @@ impl Table {
     }
 }
 
-// TODO(edd): reduce owned strings here by, e.g., using references as keys.
-// Does not implement `Default`; must contain some data.
+/// Table level MetaData
```
alamb, line 543: There are several structs called `Metadata` and my editor shows the first sentence of the docstring when I hover over it, so use the docstring to try and distinguish between them quickly

e-dard (approved):

Issue #2124, opened 2021-07-26
Statistics count mismatch: IOx panicking on failed non-null value assertion


e-dard:
https://github.com/influxdata/influxdb_iox/pull/2103 aligned the tracking of all value counts within columns and also all non-null counts, across the various execution engines (MUB, RUB, Parquet).

This has now been deployed internally and I have seen the panic below.

It is panicking here:

https://github.com/influxdata/influxdb_iox/blob/6ae1ba57b91647ff57708f8b4a26a5ef480a5902/data_types/src/partition_metadata.rs#L155-L166

And given the specific column (`os`) is known to be reasonably sparse I suspect it has something to do with the NULL values in there.

```
level=error msg="thread 'tokio-runtime-worker' panicked at 'UNKNOWN', data_types/src/partition_metadata.rs:156:13" span=549755813910 target="panic_logging" location="panic_logging/src/lib.rs:110" time=1627333382210685572
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `87389`,
 right: `100194`: Mismatch counts in table http_api_request_duration_seconds column os, expected 100194 got 87389', data_types/src/partition_metadata.rs:156:13
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at ./rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/panicking.rs:92:14
   2: core::panicking::assert_failed::inner
   3: core::panicking::assert_failed
             at ./rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/panicking.rs:143:5
   4: data_types::partition_metadata::TableSummary::total_count
             at ./home/rust/project/data_types/src/partition_metadata.rs:156:13
   5: server::db::lifecycle::compact::compact_chunks::{{closure}}
             at ./home/rust/project/server/src/db/lifecycle/compact.rs:49:27
   6: core::iter::adapters::map::map_try_fold::{{closure}}
             at ./rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/core/src/iter/adapters/map.rs:89:28
   7: core::iter::traits::iterator::Iterator::try_fold
[...]

PR #2249, opened 2021-08-10
fix(mutable_buffer): Statistics incorrect when subset of columns is inserted

alamb:
Part of https://github.com/influxdata/influxdb_iox/issues/2124

# Rationale
When inserting batches into the MUB that do have all columns, the values were padded out to the correct row counts with nulls, but the statistics were not.

# Changes
1. Pad statisics as well
2. Tests for same


mutable_buffer/src/chunk.rs
```diff
@@ -502,6 +503,86 @@ mod tests {
...
+        let lp4 = "cpu new_field=200 4";
+
+        let mut chunk = write_lp_to_new_chunk(lp1).unwrap();
+        write_lp_to_chunk(&lp2, &mut chunk).unwrap();
+        write_lp_to_chunk(&lp3, &mut chunk).unwrap();
+        write_lp_to_chunk(&lp4, &mut chunk).unwrap();
+
+        let summary = chunk.table_summary();
+
+        let expected_column_summaries = vec![
```
alamb, line 530: There were some great values in here originally (like total_count = 3 for some columns and total_count = 4 for others)

NGA-TRAN: Nice to see all columns in same chunk have same count.

mutable_buffer/src/column.rs
```diff
@@ -266,12 +268,30 @@ impl Column {
...
+                stats.update_for_nulls(delta as u64);
+            }
+            ColumnData::Bool(data, stats) => {
+                data.append_unset(delta);
+                stats.update_for_nulls(delta as u64);
+            }
+            ColumnData::Tag(data, _dict, stats) => {
+                data.resize(len, INVALID_DID);
+                stats.update_for_nulls(delta as u64);
+            }
```
NGA-TRAN, line 294: nice

NGA-TRAN (approved):
LGTM

mutable_buffer/src/column.rs
alamb: 😆  yes that seems like an improvement for sure

PR #2250, opened 2021-08-10
fix: Convert warn --> panic  on statistics mismatch

alamb:
Related PRs fixing the underlying issue
- [x] https://github.com/influxdata/influxdb_iox/pull/2249 
- [x] https://github.com/influxdata/influxdb_iox/pull/2251

Resolves #2124 

# Rationale:
1. In https://github.com/influxdata/influxdb_iox/pull/2125 we converted an internal assertion to a warning until we could deal with it
2. We are fixing the root cause (see list above)
3. Converting back to an assert is important because incorrect statistics will lead to incorrect results

# Changes:
1. Revert  commit 7415fa0d05ec35fff4d555812732e11214dfe0bf.

PR #2251, opened 2021-08-10
fix: Handle all null columns in ColumnSummary::update_from

alamb:
Another fix related to https://github.com/influxdata/influxdb_iox/issues/2124 (found with https://github.com/influxdata/influxdb_iox/pull/2250).

# Rationale
1. When combining `TableSummary`s that do have all columns, the statistics are not updated to take into account the missing columns

# Changes
1. Pad statistics for missing columns


data_types/src/partition_metadata.rs
```diff
@@ -971,7 +1006,7 @@ mod tests {
         let col = table_a.column("float").unwrap();
         assert_eq!(
             col.stats,
-            Statistics::F64(StatValues::new_non_null(Some(1.3), Some(9.1), 2))
+            Statistics::F64(StatValues::new(Some(1.3), Some(9.1), 4, 2))
```
alamb, line 1009: columns in the same logical table should have the same number of rows (4 in this case)! 🤕

data_types/src/partition_metadata.rs
```diff
@@ -994,8 +1029,75 @@ mod tests {
...
+        ));
+
+        let expected_int_stats = Statistics::I64(StatValues::new(
+            Some(5),
+            Some(5),
+            2,
+            0, // no nulls
+        ));
+
+        // update table 1 with table 2
```
alamb, line 1080: this test explicitly ensures that the order of updating stats does not affect the output statistics

data_types/src/partition_metadata.rs
```diff
@@ -1035,10 +1137,11 @@ mod tests {
         let col = t.column("string").unwrap();
         assert_eq!(
             col.stats,
-            Statistics::String(StatValues::new_non_null(
+            Statistics::String(StatValues::new(
                 Some("bar".to_string()),
                 Some("foo".to_string()),
-                2,
+                3, // no entry for "string" column in table_a_2
```
alamb, line 1143: again, in this test the other column has three rows, and so should the `"string"` column

e-dard (approved):

PR #2255, opened 2021-08-10
refactor: remove logging

e-dard:
Tidies up some logging we were using for debugging compaction issues.

### DIFF
--- a/read_buffer/src/chunk.rs
+++ b/read_buffer/src/chunk.rs
@@ -8,7 +8,7 @@ use arrow::record_batch::RecordBatch;
 use data_types::{chunk_metadata::ChunkColumnSummary, partition_metadata::TableSummary};
 use internal_types::{schema::builder::Error as SchemaError, schema::Schema, selection::Selection};
 use metrics::{Gauge, KeyValue};
-use observability_deps::tracing::info;
+use observability_deps::tracing::debug;
 use snafu::{ResultExt, Snafu};
 use std::{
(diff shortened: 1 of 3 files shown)


read_buffer/src/chunk.rs
```diff
@@ -286,54 +287,11 @@ impl Chunk {
...
-    let raw_rb_comp = format!(
-        "{:.2}%",
-        (1.0 - (rg_size as f64 / raw_size_null as f64)) * 100.0
-    );
-
-    info!(%rows, %columns, rg_size, mub_rb_size, %mub_rb_comp, raw_size_null, raw_size_no_null, %raw_rb_comp, ?table_name, ?compressing_took, "row group added");
-
+    let row_group = RowGroup::from(rb);
+    debug!(rows=row_group.rows(), columns=row_group.columns(), size_bytes=row_group.size(), 
+        raw_size_null=row_group.size_raw(true), raw_size_no_null=row_group.size_raw(true), table_name=?table_name, compressing_took=?now.elapsed(), "row group added");
```
alamb, line 294: > raw_size_null=row_group.size_raw(true), raw_size_no_null=row_group.size_raw(true),

Is it intended that the call to `row_group.size_raw(true)` is the same for both parameters?

alamb (approved):
not sure about the double call to `row_group.size_raw(true)` but otherwise this looks good

NGA-TRAN (approved):
LGTM

PR #2256, opened 2021-08-11
feat: manual skip replay, part 1/2

crepererum:
I want to be able to manually get databases out of a failed replay (e.g. when Kafka already deleted data that we would need for replay). The easiest thing would be to manually trigger "skip replay" in the same way we do that via a env flag at the moment (`INFLUXDB_IOX_SKIP_REPLAY`).

This PR implements the Rust-API-level part of it (`Server::skip_replay`, similar to `Server::wipe_preserved_catalog`). A follow-up PR will do the gRPC and CLI wiring including an end2end test.

This very PR here turned out to be a bit larger because the mocking that is required to test this. Namely we must be able to have some in-memory mocked write buffer that survives server restarts. Since the write buffer is not passed directly (in contrast to the object store) but only as a connection string, some (not too messy) config handling is required for that.

### DIFF
--- a/server/src/database.rs
+++ b/server/src/database.rs
@@ -60,6 +60,12 @@ pub enum Error {
         db_name: String,
         source: Box<parquet_file::catalog::Error>,
     },
+
+    #[snafu(display("failed to skip replay for database ({}): {}", db_name, source))]
+    SkipReplay {
+        db_name: String,
+        source: Box<InitError>,
+    },
 }
 
 /// A `Database` represents a single configured IOx database - i.e. an entity with a corresponding
@@ -216,6 +222,50 @@ impl Database {
             {
                 let mut state = shared.state.write();
                 *state.unfreeze(handle) = DatabaseState::RulesLoaded(current_state);
+            }
+
+            Ok(())
+        })
+    }
+
+    /// Recover from a ReplayError by skipping replay
+    pub fn skip_replay(&self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
+        let db_name = &self.shared.config.name;
+        let (mut current_state, handle) = {
+            let state = self.shared.state.read();
+            let current_state = match &**state {
+                DatabaseState::ReplayError(rules_loaded, _) => rules_loaded.clone(),
+                _ => {
+                    return Err(Error::InvalidState {
+                        db_name: db_name.to_string(),
+                        state: state.state_code(),
+                        transition: "SkipReplay".to_string(),
+                    })
+                }
+            };
+
+            let handle = state.try_freeze().ok_or(Error::TransitionInProgress {
+                db_name: db_name.to_string(),
+                state: state.state_code(),
+            })?;
+
+            (current_state, handle)
+        };
+
+        let shared = Arc::clone(&self.shared);
+
+        Ok(async move {
+            let db_name = &shared.config.name;
+            current_state.replay_plan = Arc::new(None);
+            let current_state = current_state
+                .advance()
+                .await
+                .map_err(Box::new)
+                .context(SkipReplay { db_name })?;
+
+            {
+                let mut state = shared.state.write();
+                *state.unfreeze(handle) = DatabaseState::Initialized(current_state);
             }
 
             Ok(())
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -155,6 +155,9 @@ pub enum Error {
     #[snafu(display("error wiping preserved catalog: {}", source))]
     WipePreservedCatalog { source: database::Error },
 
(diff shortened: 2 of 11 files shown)


tustvold (approved):
I think this is probably fine, but I think it would be possible to write an equivalent test without needing to resort to global state by just constructing a `Database` in the CatalogLoaded state. 

I'm not sure why `Server` needs to be involved at all in a database replay test, and using it necessitates a strange dance to do dependency injection into the Database.

crepererum:
For the record: I'm trying out a less invasive mocking as @tustvold suggested, will see how that goes.

server/src/database.rs
```diff
@@ -221,6 +227,50 @@ impl Database {
...
         })
     }
+
+    /// Recover from a ReplayError by skipping replay
+    pub fn skip_replay(&self) -> Result<impl Future<Output = Result<(), Error>>, Error> {
+        let db_name = &self.shared.config.name;
+        let (mut current_state, handle) = {
+            let state = self.shared.state.read();
+            let current_state = match &**state {
+                DatabaseState::ReplayError(rules_loaded, _) => rules_loaded.clone(),
```
alamb, line 237: 👍

server/src/lib.rs
```diff
@@ -1034,6 +1037,25 @@ where
 
         Ok(tracker)
     }
+
+    /// Recover database that has failed to replay by skipping replay
+    ///
+    /// The DB must exist in the server and have failed to replay for this to work.
+    /// This is done to prevent race conditions between DB jobs and this command.
+    pub fn skip_replay(&self, db_name: &DatabaseName<'static>) -> Result<TaskTracker<Job>> {
```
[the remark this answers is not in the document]
alamb: if it is a background task do we have to worry about two concurrent requests? Or maybe the second to run would just fail?

write_buffer/src/config.rs
```diff
@@ -23,24 +27,150 @@ impl WriteBufferConfig {
...
         let name = rules.db_name();
 
-        // Right now, the Kafka producer and consumers ar the only production implementations of the
-        // `WriteBufferWriting` and `WriteBufferReading` traits. If/when there are other kinds of
-        // write buffers, additional configuration will be needed to determine what kind of write
-        // buffer to use here.
         match rules.write_buffer_connection.as_ref() {
             Some(WriteBufferConnection::Writing(conn)) => {
-                let kafka_buffer = KafkaBufferProducer::new(conn, name)?;
+                if let Some(conn) = conn.strip_prefix(PREFIX_MOCK) {
```
alamb, line 32: this will likely be a useful debugging / testing aid in the future. 👍

write_buffer/src/config.rs
```diff
@@ -23,24 +27,150 @@ impl WriteBufferConfig {
...
+                } else {
+                    let kafka_buffer = KafkaBufferProducer::new(conn, name)?;
 
-                Ok(Some(Self::Writing(Arc::new(kafka_buffer) as _)))
+                    Ok(Some(Self::Writing(Arc::new(kafka_buffer) as _)))
+                }
             }
             Some(WriteBufferConnection::Reading(conn)) => {
-                let kafka_buffer = KafkaBufferConsumer::new(conn, server_id, name).await?;
+                if let Some(conn) = conn.strip_prefix(PREFIX_MOCK) {
```
alamb, line 48: I wonder if there is some way to avoid the copy/paste here?

write_buffer/src/mock.rs
```diff
@@ -91,6 +112,54 @@ impl MockBufferSharedState {
...
+}
+
+impl Drop for MockBufferSharedState {
+    fn drop(&mut self) {
+        self.entries.take();
+
+        if let Some(global_state) = STATES.get() {
+            let mut guard = global_state.write();
+
+            guard.retain(|_id, entries| Weak::upgrade(entries).is_some());
```
alamb, line 160: so the point of this is so that a second `MockBufferForWriting` can be created and still have the entries previously written? Nice

server/src/lib.rs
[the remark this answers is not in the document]
crepererum: Will move away from using a task here, but since @alamb asked: you can have two concurrent tasks, but one of them will fail w/ an exception stating that the DB is not in right state to perform this operation. Aka the tasks are serialized. That's the `state.read()` and `state.try_freeze()` stuff in `Database::skip_replay`.

crepererum:
Closing, since the write buffer mocking has a new design now and the interface should be `Database`-only (aka not reflected in `Server`) and doesn't require a background job. Will file a new version later today.

PR #2258, opened 2021-08-11
feat: row-level pruning during replay

crepererum:
Only filed for very early review.

### DIFF
--- a/mutable_buffer/src/chunk.rs
+++ b/mutable_buffer/src/chunk.rs
@@ -41,6 +41,9 @@ pub enum Error {
 
     #[snafu(display("Column not found: {}", column))]
     ColumnNotFound { column: String },
+
+    #[snafu(display("Mask had {} rows, expected {}", expected, actual))]
+    IncorrectMaskLength { expected: usize, actual: usize },
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
@@ -88,7 +91,13 @@ pub struct MBChunk {
 impl MBChunk {
     /// Create a new batch and write the contents of the [`TableBatch`] into it. Chunks
     /// shouldn't exist without some data.
-    pub fn new(metrics: ChunkMetrics, batch: TableBatch<'_>) -> Result<Self> {
+    ///
+    /// If `mask` is provided, only entries that are marked w/ `true` are written.
+    pub fn new(
+        metrics: ChunkMetrics,
+        batch: TableBatch<'_>,
+        mask: Option<&[bool]>,
+    ) -> Result<Self> {
         let table_name = Arc::from(batch.name());
 
         let mut chunk = Self {
@@ -99,15 +108,21 @@ impl MBChunk {
         };
 
         let columns = batch.columns();
-        chunk.write_columns(columns)?;
+        chunk.write_columns(columns, mask)?;
 
         Ok(chunk)
     }
 
     /// Write the contents of a [`TableBatch`] into this Chunk.
     ///
+    /// If `mask` is provided, only entries that are marked w/ `true` are written.
+    ///
     /// Panics if the batch specifies a different name for the table in this Chunk
-    pub fn write_table_batch(&mut self, batch: TableBatch<'_>) -> Result<()> {
+    pub fn write_table_batch(
+        &mut self,
+        batch: TableBatch<'_>,
+        mask: Option<&[bool]>,
+    ) -> Result<()> {
         let table_name = batch.name();
         assert_eq!(
             table_name,
@@ -115,7 +130,7 @@ impl MBChunk {
             "can only insert table batch for a single table to chunk"
         );
 
-        self.write_columns(batch.columns())?;
+        self.write_columns(batch.columns(), mask)?;
 
         // Invalidate chunk snapshot
         *self
@@ -273,10 +288,28 @@ impl MBChunk {
 
     /// Validates the schema of the passed in columns, then adds their values to
     /// the associated columns in the table and updates summary statistics.
-    fn write_columns(&mut self, columns: Vec<entry::Column<'_>>) -> Result<()> {
+    ///
+    /// If `mask` is provided, only entries that are marked w/ `true` are written.
+    fn write_columns(
+        &mut self,
+        columns: Vec<entry::Column<'_>>,
+        mask: Option<&[bool]>,
+    ) -> Result<()> {
         let row_count_before_insert = self.rows();
         let additional_rows = columns.first().map(|x| x.row_count).unwrap_or_default();
-        let final_row_count = row_count_before_insert + additional_rows;
+        let masked_values = if let Some(mask) = mask {
+            ensure!(
+                additional_rows == mask.len(),
+                IncorrectMaskLength {
+                    expected: additional_rows,
+                    actual: mask.len(),
+                }
+            );
+            mask.iter().filter(|x| !*x).count()
+        } else {
+            0
+        };
+        let final_row_count = row_count_before_insert + additional_rows - masked_values;
 
         // get the column ids and validate schema for those that already exist
         columns.iter().try_for_each(|column| {
@@ -313,7 +346,7 @@ impl MBChunk {
                 })
                 .1;
 
-            column.append(&fb_column).context(ColumnError {
+            column.append(&fb_column, mask).context(ColumnError {
                 column: fb_column.name(),
             })?;
 
@@ -353,7 +386,7 @@ pub mod test_helpers {
             );
 
             for batch in table_batches {
-                chunk.write_table_batch(batch)?;
+                chunk.write_table_batch(batch, None)?;
             }
         }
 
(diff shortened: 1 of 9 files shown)


mutable_buffer/src/column.rs
```diff
@@ -379,31 +422,122 @@ impl Column {
...
             );
 
-            Ok(data
-                .iter()
-                .map(|x| {
-                    // Currently the bit mask is backwards
-                    !x.reverse_bits()
-                })
-                .collect())
+            let res: Vec<u8> = if let Some(inclusion_mask) = inclusion_mask {
```
crepererum, line 481: this bit-mangling needs some more testing!

mutable_buffer/src/chunk.rs
```diff
@@ -264,10 +272,20 @@ impl MBChunk {
...
+        &mut self,
+        columns: Vec<entry::Column<'_>>,
+        mask: Option<&[bool]>,
+    ) -> Result<()> {
         let row_count_before_insert = self.rows();
         let additional_rows = columns.first().map(|x| x.row_count).unwrap_or_default();
-        let final_row_count = row_count_before_insert + additional_rows;
+        let masked_values = if let Some(mask) = mask {
+            assert_eq!(additional_rows, mask.len());
+            mask.iter().map(|x| !x as usize).sum::<usize>()
```
tustvold, line 284: `.filter().len()` might be a more obvious way to write this?

mutable_buffer/src/chunk.rs
```diff
@@ -264,10 +272,20 @@ impl MBChunk {
 
     /// Validates the schema of the passed in columns, then adds their values to
     /// the associated columns in the table and updates summary statistics.
-    fn write_columns(&mut self, columns: Vec<entry::Column<'_>>) -> Result<()> {
+    fn write_columns(
```
tustvold, line 275: Appreciate this is a draft, but some docs on what the mask is would be good :laughing:

mutable_buffer/src/column.rs
```diff
@@ -125,15 +125,25 @@ impl Column {
...
+        } else {
+            0
+        };
+        let row_count = entry.row_count - masked_values;
         if row_count == 0 {
             return Ok(());
         }
 
-        let mask = construct_valid_mask(entry)?;
+        let valid_mask = construct_valid_mask(entry, row_count, inclusion_mask)?;
```
tustvold, line 146: The Entry data structure doesn't store NULLs and so if you mask out a row in this mask, you'll actually just end up shifting the values. Unless you're compensating for this somewhere, I don't think this is correct

mutable_buffer/src/column.rs
```diff
@@ -148,21 +158,26 @@ impl Column {
                 col_data.append_unset(row_count);
 
                 let initial_total_count = stats.total_count;
-                let to_add = entry_data.len();
-                let null_count = row_count - to_add;
+                let mut added = 0;
 
-                for (idx, value) in iter_set_positions(&mask).zip(entry_data) {
+                for (idx, value) in iter_set_positions(&valid_mask)
```
tustvold, line 163: Assuming my understanding of inclusion_mask is correct - I'd expect this to simply do the normal iter_set_positions(&valid_mask) and then lookup idx in inclusion_mask and skip if masked out. I guess I'm not following why MaskedIter is necessary

mutable_buffer/src/column.rs
```diff
@@ -237,23 +276,27 @@ impl Column {
...
                 assert_eq!(
                     stats.total_count - initial_total_count - null_count as u64,
-                    to_add as u64
+                    added as u64
                 );
             }
         };
 
-        self.valid.append_bits(entry.row_count, &mask);
+        self.valid.append_bits(row_count, &valid_mask);
```
tustvold, line 299: Here I'd expect a if / else. If there is no inclusion mask it should do this, otherwise it would

Call append_unset with `row_count` and then iterate through valid_mask zipped with inclusion_mask and set the bit where they both are set.

server/src/db/replay.rs
```diff
@@ -256,20 +256,27 @@ fn filter_entry(
...
+                    (false, None)
                 }
                 SequenceNumberSection::PartiallyPersisted => {
                     // TODO: implement row filtering, for now replay the entire batch
-                    true
+                    let maybe_mask = table_batch.timestamps().ok().map(|timestamps| {
+                        let ts_flush = ts_flush.timestamp_nanos();
+                        timestamps
+                            .into_iter()
+                            .map(|ts_row| ts_row >= ts_flush)
```
tustvold, line 272: Has the change been made to make this the flush timestamp, or is it still the min_unpersisted. If so I think this should be > as it it 1 nanosecond more than the flush timestamp?

tustvold (commented):
I didn't look in great detail at the exact code, let me know if you would like me to, but I think there is a lot more additional complexity in the MUB than I would perhaps expect

server/src/db/replay.rs
crepererum: Good catch

mutable_buffer/src/column.rs
crepererum: The problem is that we later use `valid_mask` to construct the "where do the non-null data points go to"-index iterator, aka: `data=[a, b], mask=0b1010` is mapped to `a=>3, b=>1` (note that the mask is read lowest to highest bit). So if you apply the masked iterator to the Zip of these two, we would need to correct the indices. Instead of doing that I found it easier to just filter the null/valid-mask according to the inclusion-mask.

mutable_buffer/src/column.rs
crepererum: See me comment above.

mutable_buffer/src/column.rs
crepererum: Maybe I should rename stuff:

- the `row_count` here is the "row count AFTER applying the inclusion mask" aka "number of included elements"
- the `valid_mask` it this point is already filtered through the inclusion mask and therefore contains that many bits (+ the byte padding)

mutable_buffer/src/column.rs
tustvold: I'm not suggesting using a zip, that is precisely what I'm saying is really confusing. I'm saying something like

```
for (idx, value) in iter_set_positions(&valid_mask) {
    if !inclusion_mask[idx] {
        continue
    }
}
```

This avoids needing to modify valid_mask at all, and avoids custom iterator types

mutable_buffer/src/column.rs
tustvold: Sorry I should have been clearer, I would expect to not be modifying the valid_mask - this was in the context of the comments above

mutable_buffer/src/column.rs
crepererum: `idx` is also used to write data at `offset + idx` (`offset` being the number of entries prior to this append operation). So if you filter out rows like this, you get an out-of-bounds issue (because we might not even want that many rows according to the inclusion mask). What I could do however is to keep a counter every time the `continue` in your example is hit and write to `offset + idx - n_excluded_already`.

tustvold: Oh yeah, because you're skipping rows the offsets change :facepalm:

I personally think not conflating the mask which is necessary for correctly interpreting the flatbuffer, and the inclusion mask which is skipping rows from this flatbuffer is worth doing if it is possible. This logic could perhaps be encapsulated in an iterator that takes a flatbuffer and an inclusion mask and just yields the insertion offsets and values... I dunno, I think was already somewhat non-trivial to understand, and it is now even more so...

server/src/db/replay.rs
crepererum: renamed the variable. in that case I think the comparison is correct? (aka all rows that are greater or equal the min unpersisted ts should be processed)

mutable_buffer/src/column.rs
crepererum: @tustvold done, now everything is based on the masked iterator which is tested in isolation

mutable_buffer/src/chunk.rs
crepererum: done

mutable_buffer/src/column.rs
```diff
@@ -399,6 +449,102 @@ impl Column {
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(mask) = self.mask.as_mut() {
+            for (set_position, value) in &mut self.it {
+                loop {
+                    let (idx_mask, included) = mask.next().expect("inclusion mask too short");
+                    if !included {
+                        self.exluded_count += 1;
+                    }
+                    if idx_mask == set_position {
```
tustvold, line 527: Purely a stylistic thing you can take or leave. You could reduce the nesting and potentially make this easier to read by changing this to be if `idx_mask < set_position { continue }`

tustvold (approved):
Extremely minor nit but otherwise looks great! Sorry for taking so long to get back to this

mutable_buffer/src/column.rs
crepererum: fixed

PR #2259, opened 2021-08-11
refactor: move database creation code around

crepererum:
Now all the code that is required to create a new database lives under
`server::database`, so it can easily be used for tests that don't
involve a server.


alamb (approved):
Looks like a nice cleanup to me. Less stuff in `lib` sounds like a good idea to me

PR #2262, opened 2021-08-11
refactor: Extract a Transaction type to manage transaction paths

carols10cents:
Connects to #2193. This is an attempt to submit changes that aren't refactoring the *entire* world, just a bit of it :)

### DIFF
--- a/parquet_file/src/catalog.rs
+++ b/parquet_file/src/catalog.rs
@@ -456,7 +456,7 @@ impl PreservedCatalog {
                 &object_store,
                 server_id,
                 &db_name,
-                &tkey,
+                tkey,
                 &mut state,
                 &last_tkey,
                 file_type,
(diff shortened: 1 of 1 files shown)


parquet_file/src/catalog.rs
```diff
@@ -579,53 +582,87 @@ fn file_path(
...
-    // pad number: `u64::MAX.to_string().len()` is 20
-    path.push_dir(format!("{:0>20}", tkey.revision_counter));
+    transaction.file_path()
+}
 
-    let file_name = format!("{}.{}", tkey.uuid, file_type.suffix());
-    path.set_file_name(file_name);
+struct Transaction {
+    /// catalog_root will be stored elsewhere when all parquet file paths are relative
+    catalog_root: Option<Path>,
```
alamb, line 602: I don't fully understand the use case for having `catalog_root` or `file_path` ever be `None` but from your comments below it sounds like it will become clear in a follow on PR

carols10cents: Yeah, right now it's more like an enum, there'll be one or the other... but in a follow on PR, the catalog root will be handled by the iox object store for a particular database and this will only ever have a relative path akin to file path.

PR #2263, opened 2021-08-11
feat: Crate that wraps object store for a particular database

carols10cents:
Connects to #2193.

This is a step towards making parquet files relative to the database directory so that we can easily change the database directory.

The weirdest part of this is the `tokio::spawn` with the channel for listing from the object store... this is because in the next step, I'm going to change the iox_object_store's API to only use and return `IoxPath`s that are created within this crate. The current object_store trait API has lifetime restrictions that the output stream's lifetime is tied to the input.... which I don't think is actually needed, but is proving tricky to resolve (I think the cloud_storage crate has some overly restrictive lifetimes, and Rust might need generators to understand what I'm trying to tell it...)

### DIFF
--- a/iox_object_store/src/lib.rs
+++ b/iox_object_store/src/lib.rs
@@ -1,0 +1,198 @@
+#![deny(broken_intra_doc_links, rustdoc::bare_urls, rust_2018_idioms)]
+#![warn(
+    missing_copy_implementations,
+    missing_debug_implementations,
+    missing_docs,
+    clippy::explicit_iter_loop,
+    clippy::future_not_send,
+    clippy::use_self,
+    clippy::clone_on_ref_ptr
+)]
+
+//! Wraps the object_store crate with IOx-specific semantics.
+
+// TODO: Create an IoxPath type and only take/return paths of those types, and wrap in the
+// database's root path before sending to the underlying object_store.
+
+use bytes::Bytes;
+use data_types::{server_id::ServerId, DatabaseName};
+use futures::{stream::BoxStream, Stream, StreamExt};
+use object_store::{
+    path::{parsed::DirsAndFileName, ObjectStorePath, Path},
+    ObjectStore, ObjectStoreApi, Result,
+};
+use std::{io, sync::Arc};
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+/// Handles persistence of data for a particular database. Writes within its directory/prefix.
+#[derive(Debug)]
+pub struct IoxObjectStore {
+    store: Arc<ObjectStore>,
+    server_id: ServerId,
+    database_name: String, // TODO: use data_types DatabaseName?
+    root_path: RootPath,
+}
+
+impl IoxObjectStore {
+    /// Create a database-specific wrapper. Takes all the information needed to create the
+    /// root directory of a database.
+    pub fn new(
+        store: Arc<ObjectStore>,
+        server_id: ServerId,
+        database_name: &DatabaseName<'_>,
+    ) -> Self {
+        let root_path = RootPath::new(store.new_path(), server_id, database_name);
+        Self {
+            store,
+            server_id,
+            database_name: database_name.into(),
+            root_path,
+        }
+    }
+
+    /// The name of the database this object store is for.
+    pub fn database_name(&self) -> &str {
+        &self.database_name
+    }
+
+    /// Path where transactions are stored.
+    ///
+    /// The format is:
+    ///
+    /// ```text
+    /// <server_id>/<db_name>/transactions/
+    /// ```
+    // TODO: avoid leaking this outside this crate
+    pub fn catalog_path(&self) -> Path {
+        let mut path = self.store.new_path();
+        path.push_dir(self.server_id.to_string());
+        path.push_dir(&self.database_name);
+        path.push_dir("transactions");
+        path
+    }
+
+    /// Location where parquet data goes to.
+    ///
+    /// Schema currently is:
+    ///
+    /// ```text
+    /// <server_id>/<db_name>/data/
+    /// ```
+    // TODO: avoid leaking this outside this crate
+    pub fn data_path(&self) -> Path {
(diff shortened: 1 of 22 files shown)


iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    ObjectStore, ObjectStoreApi, Result,
+};
+use std::{io, sync::Arc};
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+/// Handles persistence of data for a particular database. Writes within its directory/prefix.
+#[derive(Debug)]
+pub struct IoxObjectStore {
+    store: Arc<ObjectStore>,
```
alamb, line 31: I wonder if calling this `inner` is a more common name in Rust when wrapping some other type

iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+use futures::{stream::BoxStream, Stream, StreamExt};
+use object_store::{
+    path::{parsed::DirsAndFileName, ObjectStorePath, Path},
+    ObjectStore, ObjectStoreApi, Result,
+};
+use std::{io, sync::Arc};
+use tokio::sync::mpsc::channel;
+use tokio_stream::wrappers::ReceiverStream;
+
+/// Handles persistence of data for a particular database. Writes within its directory/prefix.
```
alamb, line 28:
```suggestion
/// Handles persistence of data for a particular database. Writes within its directory/prefix.
///
/// This wrapper on top of an `ObjectStore` maps IOx specific concepts to ObjectStore locations
```

iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// Location where parquet data goes to.
+    ///
+    /// Schema currently is:
+    ///
+    /// ```text
+    /// <server_id>/<db_name>/data/
+    /// ```
+    // TODO: avoid leaking this outside this crate
```
alamb, line 82: 👍  I like the TODOs

iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Store this data in this database's object store.
+    pub async fn put<S>(&self, location: &Path, bytes: S, length: Option<usize>) -> Result<()>
+    where
+        S: Stream<Item = io::Result<Bytes>> + Send + Sync + 'static,
+    {
+        self.store.put(location, bytes, length).await
+    }
+
+    /// List all the catalog transaction files in object storage for this database.
+    pub async fn catalog_transaction_files(&self) -> Result<BoxStream<'static, Result<Vec<Path>>>> {
```
alamb, line 100: Is the intention that this will eventually return `stream goop ... IoxPath` rather than raw `Path`s?

iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+        let store = Arc::clone(&self.store);
+        let prefix = prefix.cloned();
+
+        // This is necessary because of the lifetime restrictions on the ObjectStoreApi trait's
+        // methods, which might not actually be necessary but fixing it involves changes to the
+        // cloud_storage crate that are longer term.
+        tokio::spawn(async move {
+            match store.list(prefix.as_ref()).await {
+                Err(e) => {
+                    tx.send(Err(e)).await.expect("sending over channel failed");
```
alamb, line 119: I think it is ok if sending on the `tx` fails and should not panic (both this line and that below). A failure to send simply means that whatever code called list() did not fully consume the results which seems fine to me

iox_object_store/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Creates a new in-memory object store. These tests rely on the `Path`s being of type
+    /// `DirsAndFileName` and thus using object_store::path::DELIMITER as the separator
+    fn make_object_store() -> Arc<ObjectStore> {
+        Arc::new(ObjectStore::new_in_memory())
+    }
+
+    #[test]
+    fn catalog_path_is_relative_to_db_root() {
+        let server_id = make_server_id();
+        let database_name = DatabaseName::new("clouds").unwrap();
```
alamb, line 183: 🌧️

parquet_file/src/catalog.rs
```diff
@@ -258,35 +261,25 @@ pub struct PreservedCatalog {
...
-    pub async fn exists(
-        object_store: &ObjectStore,
-        server_id: ServerId,
-        db_name: &str,
-    ) -> Result<bool> {
-        Ok(!list_files(object_store, server_id, db_name)
-            .await?
-            .is_empty())
+    pub async fn exists(iox_object_store: &IoxObjectStore) -> Result<bool> {
+        Ok(!list_files(iox_object_store).await?.is_empty())
```
alamb, line 270: its already looking better (more encapsulated) 👍

parquet_file/src/storage.rs
```diff
@@ -439,32 +432,11 @@ impl TryClone for MemWriter {
...
 }
 
-/// Location where parquet data goes to.
-///
-/// Schema currently is:
-///
-/// ```text
-/// <writer_id>/<database>/data
-/// ```
-pub(crate) fn data_location(
```
alamb, line 449: ❤️

alamb (approved):
🏅 Wow -- this is a pretty epic PR @carols10cents . Nicely done

I also think this is coming together nicely (once all object store access is via IoxObjectStore then adding a generation number to the catalog will be simple. 

I won't say I was able to scrutinize every line of this PR, but I skimmed it all and the parts I read carefully looked really good

Delete database here we come!

iox_object_store/src/lib.rs
carols10cents: Yup!

iox_object_store/src/lib.rs
carols10cents: 🤷🏻‍♀️ I'm into it, I don't know that there's a convention written down anywhere but it certainly is common.

PR #2264, opened 2021-08-12
feat: use heappy implemented by mkm

NGA-TRAN:
Closes #


heappy/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+enable_heap_profiler = [ "jemalloc_shim" ]
+measure_free = []
+
+[dependencies]
+backtrace = "0.3.59"
+bytes = "1.0.1"
+lazy_static = "1.4.0"
+libc = { version = "^0.2.8", default-features = false }
+pprof = {version = "0.4.3", features = [ "flamegraph", "protobuf" ] }
+prost = "0.7"
```
NGA-TRAN, line 19: In the top level Cargo.toml, Prost version is 0.8

src/influxdb_ioxd/http.rs
```diff
@@ -838,6 +857,41 @@ async fn pprof_profile<M: ConnectionManager + Send + Sync + Debug + 'static>(
...
+    {
+        report.flamegraph(&mut body);
+        if body.is_empty() {
+            return EmptyFlamegraph.fail();
+        }
+    }
+    // } else {
+    //     let profile = report.pprof();
+    //     profile.encode(&mut body).context(Prost)?; // prost version 0.8 does not support encode
+    // }
```
NGA-TRAN, line 890: Because of new Prost 0.8 version, I am not sure how to make this work yet

NGA-TRAN:
I got test influxdb_ioxd::tests::test_server_shutdown ... FAILED
I will look into it tomorrow

src/influxdb_ioxd/http.rs
mkmik: that's why I created this helper:

```
    pub fn write_pprof<W: Write>(&self, writer: &mut W) -> std::io::Result<()> {
        let mut buf = vec![];
        self.pprof().encode(&mut buf)?;
        writer.write_all(&buf)
    }
```

but I see the "complex" example didn't use it

Cargo.toml
```diff
@@ -59,6 +60,7 @@ datafusion = { path = "datafusion" }
 data_types = { path = "data_types" }
 entry = { path = "entry" }
 generated_types = { path = "generated_types" }
+heappy = { path = "heappy", features = ["enable_heap_profiler"] }
```
mkmik, line 63: I intend to merge heappy upsteam to `pprof-rs`; for now you can just refer to it as:

```
heappy = { git = "https://github.com/mkmik/heappy", rev = "01e902a7c930d05f76ab777f7d7fb032a8528691" }
```

heappy/Cargo.toml
mkmik: this is not a problem, crate does handle multiple versions just fine.

The problem you had below was the attempt to use a structure produced from a crate that depended on prost 0.7 from another crate that has been compiled with prost 0.8; you then tried to call `.encode` on that struct, but since the version of the trait didn't match, it appeared as if there was no such method. (in other words, you thought you brought Message trait in scope with `use prost::Message` but you didn't bring the same `prost::Message` trait that was required by the concrete return type of `report()`.

To obviate for this kind of problem, I included `write_pprof` helper in heappy, so we don't have to leak the dependency to the protobuf library since most of the times end users just want to save this to a file or stream.

mkmik: that said, I also updated the pprof dep in heappy, so now it also depends on prost 0.8

Cargo.toml
mkmik: I'd like to work on heappy on a separate repo because the size of the iox repo is making it impossible for me to work on faster personal M1 mac

NGA-TRAN: Yeah, I like to idea to make it separate, too.

NGA-TRAN (commented):
Thanks for your review @mkmik. I will remove heavy director and use yours instead. When everything works, I will send to the group for full review

Cargo.toml
```diff
@@ -59,6 +59,8 @@ datafusion = { path = "datafusion" }
 data_types = { path = "data_types" }
 entry = { path = "entry" }
 generated_types = { path = "generated_types" }
+heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384" }
```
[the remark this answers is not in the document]
alamb: Here are the available features: https://github.com/mkmik/heappy/blob/main/Cargo.toml#L16

Issue #2257, opened 2021-08-11
Read Buffer memory usage optimistic


e-dard:
In the same vein as #1565 there are a couple of issues with the size of chunks reported by the read buffer (if you want to use this data for determining how much main-memory they are consuming).

- [x] The reported `size` method (which is used for metrics and table summary) uses length for containers not capacity. #2267
- [ ] The bitmaps associated with, for example, RLE columns, use an estimation of heap size because there isn't a readily available API for getting their allocated size.

The rationale for using `len` instead of `capacity` for containers is that I find it useful to determine the type of compression we can achieve with the read buffer ignoring the nuances of how the allocator decides to pre-allocate. 

I plan to add an API for both the existing length-based and a capacity-based sizing.

PR #2267, opened 2021-08-12
feat: Add API to encodings to get allocated buffer sizes

e-dard:
Towards: https://github.com/influxdata/influxdb_iox/issues/2257

This PR extends the `size` on all physical Read Buffer encodings to enable one to get either the minimal size the encoding would take up in memory, or the actual size it takes up in memory (which will include any pre-allocated buffers that are not being used by the encoding's contents yet).

I haven't changed any of the API higher up in the crate - that will be a follow on PR or two. My plan is to expose both of these sizes as Read Buffer metrics, and then update the Chunk memory tracking metrics to use the capacity based version.


e-dard:
I had to rebase #2268 to be able to get the tests updated locally.

data_types/src/instant.rs
```diff
@@ -29,6 +29,14 @@ pub fn to_approximate_datetime(instant: Instant) -> DateTime<Utc> {
     }
 }
 
+// *NOTE*: these tests currently fail on (at least) aarch64 architectures
```
alamb, line 32: I think these were included in https://github.com/influxdata/influxdb_iox/pull/2268 as well

read_buffer/src/column/encoding/bool.rs
```diff
@@ -42,8 +43,12 @@ impl Bool {
 
     /// Returns an estimation of the total size in bytes used by this column
     /// encoding.
```
alamb, line 45: I reommend updating the docstring to document what `buffers` means

read_buffer/src/column/encoding/bool.rs
```diff
@@ -42,8 +43,12 @@ impl Bool {
...
-    pub fn size(&self) -> usize {
-        std::mem::size_of::<BooleanArray>() + self.arr.get_array_memory_size()
+    pub fn size(&self, buffers: bool) -> usize {
+        size_of::<Self>()
+            + match buffers {
+                true => self.arr.get_array_memory_size(), // includes buffer capacities
```
alamb, line 49: TLDR is I think we should always use `self.arr.get_array_memory_size()` --the distinction between these two calls this is different than the distinction between `Vec::len()` and `Vec::capacity()`

The distinction in the `arrow` crate is that:

`get_buffer_memory_size()` includes just the size of the raw underlying arrow buffers (aka the slabs of contiguous memory that hold the data, and the accounting structures for that Data).

`get_array_memory_size()` includes both everything in `get_buffer_memory_size` as well as the actual Rust struct that points to the buffers (e.g. the ArrayData itself). Here is part of the implementation that I think illustrates the difference:  https://github.com/apache/arrow-rs/blob/master/arrow/src/array/data.rs#L341

I think the read buffer should be accounting for the overhead (array_memory_size)

If you are trying to get the equivalent of `capacity` I don't think that is yet exposed at the array data level -- you would have to implement something like `get_buffer_memory_size()` that uses `Buffer::len()` rather than `Buffer::capacity()`

read_buffer/src/column/encoding/scalar.rs
```diff
@@ -18,8 +18,10 @@ pub trait ScalarEncoding<L>: Debug + Display + Send + Sync {
     /// A useful name for the encoding, likely used in instrumentation.
     fn name(&self) -> &'static str;
 
-    /// The total size in bytes to store encoded data in memory.
-    fn size(&self) -> usize;
+    /// The total size in bytes to store encoded data in memory. If `buffers`
+    /// is true then the returned size should account for any allocated buffers
+    /// within the contained encoding structures.
+    fn size(&self, buffers: bool) -> usize;
```
alamb, line 24: I personally found this api a bit confusing. I wonder if calling the parameter `use_capacity` would make the intent clearer.

I actually think using `capacity` in the API name might make things even clearer, as that concept is already exposed and well understood in Rust (and in C++). Perhaps two methods called `size_bytes()` and `capacity_bytes()`?

read_buffer/src/column/encoding/scalar/rle.rs
```diff
@@ -70,7 +70,7 @@ where
             f,
             "[{}] size: {:?} rows: {:?} nulls: {} runs: {} ",
             self.name(),
-            self.size(),
+            self.size(false),
```
alamb, line 73: I think it might be confusing if the numbers reported to the user are (effectively) `size(true)` but the internal debugging reports `size(false)`. Maybe renaming size as suggested above to `size_bytes()` and `capacity_bytes()` would make this clearer

read_buffer/src/column/encoding/scalar/fixed.rs
```diff
@@ -425,6 +429,19 @@ mod test {
         (Fixed::new(values, Arc::clone(&mock)), mock)
     }
 
+    #[test]
+    fn size() {
+        let (v, _) = new_encoding(vec![22_i64, 1, 18]);
+        // Self if 32 bytes and there are 3 * 8b values
```
alamb, line 435:
```suggestion
        // Self is 32 bytes and there are 3 * 8b values
```

alamb (commented):
Is the rationale for  reporting capacity vs size so you can see "theoretical minimum memory that could be used" for use as a ReadBuffer diagnostic tool to pinpoint where the actual deviates significantly from the reality and try to reduce the delta?

It seems like "Actual Memory Used" is probably the one that is the most important for operating IOx and to report and should always include all memory allocated .

read_buffer/src/column/encoding/scalar.rs
e-dard: These are all private APIs deeper in the Read Buffer; I just decided to make the `size` method have different behaviour rather than duplicating methods everywhere. The `buffers` parameter won't be exposed higher up in the crate, instead there will be two methods exposed outside of the crate:

```
fn size{_bytes} -> usize{
     self.foo.size(false)
}
```

```
fn capacity{_bytes} -> usize{
     self.foo.size(true)
}
```

read_buffer/src/column/encoding/scalar/rle.rs
e-dard: This work is planned for the follow on PRs. This PR is about adding new features without changing the existing behaviour (aside from some minor accounting fixes). The follow on PR will more clearly show which type of size is exposed where.

read_buffer/src/column/encoding/bool.rs
e-dard: Hmm that is confusing! What is the point of the different methods then?

What I want is to know the minimal size in memory necessary to store the data in an Arrow array (I thought that would be `get_buffer_memory_size`) and then the _actual_ size including any pre-allocated buffers (as in, the capacity of all the vectors or anything else that lets you allocated ahead), which would be `get_array_memory_size`.

If there is no API in Arrow for that I'll go with your suggestion.

e-dard:
@alamb the rationale is two-fold:

1) to expose more accurate metrics about memory actually taken up by structures in the Read Buffer. This will be exposed to other crates and will be in things like system tables / metrics. I have *not* updated/exposed this new accouning behaviour in this PR; that's for the follow on PRs.
2) to still maintain a method for me (as someone who works on the Read Buffer) to have a consistent sizing API to know the minimum number of bytes things are taking up. This will mainly be used in testing (and also i think is useful in metrics and development of new features/changes). This API is unlikely to be of interest outside of the Read Buffer, which is one of the reasons why I didn't duplicate a bunch of similar methods everywhere and just made the output of the `size` API variable.

One other thing about (2). It's useful to be able to assert things using size without having to worry about how much the runtime has quietly pre-allocated behind your back. That way you can think about all the inputs to your data structure, count them up and assert you got that size back.

Hope that makes some sense.

read_buffer/src/column/encoding/bool.rs
NGA-TRAN: +1

read_buffer/src/column/encoding/scalar.rs
NGA-TRAN: I like to have these clear 2 methods that expose outside

NGA-TRAN (approved):
Very nice and more accurate API. Will you also modify metrics to have one for use and one fore capacity? I think they will be useful for us to measure and improve the allocation in arrow datafusion if possible.

read_buffer/src/column/encoding/bool.rs
e-dard: I'm going to make this change in the follow on PR.

read_buffer/src/column/encoding/scalar.rs
e-dard: Yes I plan to add them in the external API and integrate them into our metrics in the next PR :-)

read_buffer/src/column/encoding/bool.rs
e-dard: Resolved in next PR

read_buffer/src/column/encoding/bool.rs
alamb: The *actual* size is what is exposed in arrow as I can tell.

I do not know the usecase of the two different methods n arrow

read_buffer/src/column/encoding/scalar.rs
alamb: > These are all private APIs deeper in the Read Buffer; I just decided to make the size method have different behaviour rather than duplicating methods everywhere

I think two methods would help the readability even within the read buffer

PR #2268, opened 2021-08-12
test: ignore failing tests on aarch64

e-dard:
Unfortunately the tests in this PR are failing locally (consistently) on Apple M1 machines. Perhaps related to the following Rust issue: https://github.com/rust-lang/rust/issues/87906.

Ignoring them for now to 🍏 up the suite locally.

### DIFF
--- a/data_types/src/instant.rs
+++ b/data_types/src/instant.rs
@@ -29,6 +29,14 @@ pub fn to_approximate_datetime(instant: Instant) -> DateTime<Utc> {
     }
 }
 
+// *NOTE*: these tests currently fail on (at least) aarch64 architectures
+// such as an Apple M1 machine.
+//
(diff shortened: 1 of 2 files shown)


tustvold:
Can we just ignore them for mac aarch64 targets? If we have these issues on Linux I'd like to know

e-dard:
@tustvold realised this as soon as I pushed. Fixing it now :)

alamb (approved):
Makes sense to me. Might be worth a ticket in IOx to record the fact there may be issues if IOx is run on aarch (as the assumptions that may be baked into timestamps / persistence widows don't hold)

e-dard:
Added in https://github.com/influxdata/influxdb_iox/issues/2270

PR #2269, opened 2021-08-12
test: restore `write_buffer_errors_propagate`

crepererum:
This was removed in #2203 due to insufficient mocking capabilities.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -2235,7 +2235,7 @@ mod tests {
     #[tokio::test]
     async fn write_buffer_errors_propagate() {
         let mut factory = WriteBufferConfigFactory::new();
-        factory.register_alway_fail_mock("my_mock".to_string());
+        factory.register_always_fail_mock("my_mock".to_string());
         let application = Arc::new(ApplicationState::with_write_buffer_factory(
             Arc::new(ObjectStore::new_in_memory()),
             Arc::new(factory),
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -1279,9 +1279,11 @@ mod tests {
     use data_types::{
         chunk_metadata::ChunkAddr,
         database_rules::{
-            HashRing, LifecycleRules, PartitionTemplate, ShardConfig, TemplatePart, NO_SHARD_CONFIG,
+            HashRing, LifecycleRules, PartitionTemplate, ShardConfig, TemplatePart,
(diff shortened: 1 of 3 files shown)


server/src/lib.rs
```diff
@@ -2234,6 +2239,44 @@ mod tests {
         assert!(matches!(err, Error::CannotCreateDatabase { .. }));
     }
 
+    #[tokio::test]
+    async fn write_buffer_errors_propagate() {
```
alamb, line 2243: 👍

write_buffer/src/config.rs
```diff
@@ -42,17 +51,29 @@ impl WriteBufferConfigFactory {
...
     /// When mock with identical name is already registered.
     pub fn register_mock(&mut self, name: String, state: MockBufferSharedState) {
+        self.set_mock(name, Mock::Normal(state));
+    }
+
+    /// Registers new mock that always fail.
+    ///
+    /// # Panics
+    /// When mock with identical name is already registered.
+    pub fn register_alway_fail_mock(&mut self, name: String) {
```
alamb, line 61: There is a typo here:

```suggestion
    pub fn register_always_fail_mock(&mut self, name: String) {
```

write_buffer/src/config.rs
```diff
@@ -247,14 +282,113 @@ mod tests {
...
         let mut factory = WriteBufferConfigFactory::new();
 
         let state = MockBufferSharedState::empty_with_n_sequencers(1);
         let mock_name = "some_mock";
         factory.register_mock(mock_name.to_string(), state.clone());
         factory.register_mock(mock_name.to_string(), state);
     }
+
+    #[test]
+    #[should_panic(expected = "Mock with the name 'some_mock' already registered")]
```
alamb, line 364: I wonder if these new three tests add much coverage? Is the code path for registering the different mock types different enough to warrant testing all 4 permutations?

It seems to me a single test would be sufficient.

alamb (approved):
Thank you

write_buffer/src/config.rs
crepererum: I can live w/ a single test as well, maybe I'm just overdoing it :D

write_buffer/src/config.rs
crepererum: fixed

write_buffer/src/config.rs
crepererum: changed to a single test

PR #2271, opened 2021-08-12
 feat: manual skip replay, part 1/2 (2nd try)

crepererum:
In contrast to the original PR (#2256):

- improved mocking (made possible by #2266)
- more "local" testing (doesn't require a full `Server` anymore, a `Database` just works)
- no background job (since the skip should be very quick)


alamb (approved):
looking much cleaner

PR #2223, opened 2021-08-08
feat: add fmt::Display implementation for path data structs

sunng87:
Closes #2181 

Added `fmt::Display` implementations for `Path` and related types. For `Path`, prefix like `s3:`/`file:` was added to distinguish its backend.

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

Issue #2181, opened 2021-08-03
Improve display of object store path names


alamb:
__Use case:__
While logging, such as https://github.com/influxdata/influxdb_iox/pull/2180 it would be nice to have human readable display paths in the logs to make it easier to understand

__Proposal:__
Implement `std::fmt::Display` for `object_store::Path`:

__Current behaviour:__
Paths in the logs look like this
```
location=File(FilePath { inner: Parsed(DirsAndFileName { directories: [PathPart("42")], file_name: None })
```

__Desired behaviour:__
```
location="/42"
```

PR #2272, opened 2021-08-12
refactor: remove display methods, use fmt::Display instead.

sunng87:
This is follow up of #2223  that removes obsolete `display` methods and use `fmt::Display` instead.

### DIFF
--- a/object_store/src/memory.rs
+++ b/object_store/src/memory.rs
@@ -80,7 +80,7 @@ impl ObjectStoreApi for InMemory {
             .get(location)
             .cloned()
             .context(NoDataInMemory {
-                location: format!("{}", location),
+                location: location.to_string(),
             })?;
 
         Ok(futures::stream::once(async move { Ok(data) }).boxed())
--- a/object_store/src/memory.rs
+++ b/object_store/src/memory.rs
@@ -1,8 +1,6 @@
 //! This module contains the IOx implementation for using memory as the object
 //! store.
-use crate::{
-    path::parsed::DirsAndFileName, ListResult, ObjectMeta, ObjectStoreApi, ObjectStorePath,
-};
+use crate::{path::parsed::DirsAndFileName, ListResult, ObjectMeta, ObjectStoreApi};
 use async_trait::async_trait;
 use bytes::Bytes;
 use chrono::Utc;
@@ -82,7 +80,7 @@ impl ObjectStoreApi for InMemory {
             .get(location)
             .cloned()
             .context(NoDataInMemory {
-                location: location.display(),
+                location: format!("{}", location),
             })?;
 
         Ok(futures::stream::once(async move { Ok(data) }).boxed())
--- a/object_store/src/path.rs
+++ b/object_store/src/path.rs
@@ -41,11 +41,6 @@ pub trait ObjectStorePath:
 
     /// Push a bunch of parts as directories in one go.
     fn push_all_dirs<'a>(&mut self, parts: impl AsRef<[&'a str]>);
-
-    /// Like `std::path::Path::display, converts an `ObjectStorePath` to a
-    /// `String` suitable for printing; not suitable for sending to
-    /// APIs.
-    fn display(&self) -> String;
 }
 
 /// Defines which object stores use which path logic.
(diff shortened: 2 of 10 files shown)


object_store/src/memory.rs
```diff
@@ -82,7 +80,7 @@ impl ObjectStoreApi for InMemory {
             .get(location)
             .cloned()
             .context(NoDataInMemory {
-                location: location.display(),
+                location: format!("{}", location),
```
alamb, line 83:
```suggestion
                location: location.to_string(),
```
Since location now implements `Display` I think you can use `location.to_string()` instead here

object_store/src/path.rs
```diff
@@ -101,16 +96,6 @@ impl ObjectStorePath for Path {
             Self::MicrosoftAzure(path) => path.push_all_dirs(parts),
         }
     }
-
-    fn display(&self) -> String {
```
alamb, line 105: 🎉

alamb (approved):
Thanks @sunng87  ! Since I had your branch pulled down anyways, I took the liberty of touching up this PR and will push a commit shortly that updates a few more places

alamb:
As this might conflict with https://github.com/influxdata/influxdb_iox/pull/2272 from @carols10cents  I think we should wait until that is merged before merging this one

object_store/src/memory.rs
sunng87: Good catch. Using `to_string()` here is better.

sunng87:
No problem. I will rebase when #2278 is ready.

object_store/src/path.rs
```diff
@@ -41,11 +41,6 @@ pub trait ObjectStorePath:
 
     /// Push a bunch of parts as directories in one go.
     fn push_all_dirs<'a>(&mut self, parts: impl AsRef<[&'a str]>);
-
-    /// Like `std::path::Path::display, converts an `ObjectStorePath` to a
-    /// `String` suitable for printing; not suitable for sending to
-    /// APIs.
```
carols10cents, line 47: Um, so this comment plus the comment on line 31 in this file explains why having a `display` method rather than implementing `Display` was a deliberate design decision, to match [`std::path::Path::display`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.display): you might be tempted to do `object_storage.get(format!("{}", object_store_path)) but that's not appropriate, and with the prefix added to the display string, that definitely wouldn't work. The extra call to `display` is meant to warn you that something weird is going on and perhaps you should investigate.

Perhaps that convention isn't clear enough, and perhaps the API of the different object storage types are ergonomic enough and enough of a precedence that changes/additions would be likely to use the `to_raw` methods as they should. If you think that's the case, I'm happy to be overruled here @alamb, but I did want to explain my reasoning.

I apologize that I didn't see #2181 soon enough to have this discussion before work started!

alamb: 🤔  I definitely found the lack of display on Path puzzling in https://github.com/influxdata/influxdb_iox/issues/2181. I am sorry I should have tagged you on that ticket for discussion @carols10cents 

I thought the primary reason for `std::path::Path::display` was that paths do not always have a Utf-8 representable form. This is not the case for object store `Path`s that I know of. 

Given that `object_storage(object_store_path)` works, I am not sure how often someone might be tempted to turn a path into a string via `format!()` but not also do the same thing using `display()`

The fact that several tests were relying on the `display()` result for a `Path` and the `display()` result for a `DirsAndFileName` being the same is not ideal either and at least now the tests compare strings created from the same object type 

🤔  I need to think more on this

alamb: Perhaps another reason for `std::path::Path::display` is that it is common for other file based APIs (e.g all the `libc` functions) to take a string as a path, which would perhaps be confusing.

Perhaps another similar API that we can draw inspiration from is `Uri` from the http crate which does implement `Display`: https://docs.rs/http/0.2.4/http/uri/struct.Uri.html#impl-Display

alamb: @carols10cents  and I caught up -- the plan I believe we agreed to is to merge this PR as is (thank you again @sunng87) and I'll make a follow on PR that removes stale  comments about display not being implemented

alamb: in https://github.com/influxdata/influxdb_iox/pull/2302

Issue #2273, opened 2021-08-12
IOx Tracing


tustvold:
It should be possible to use distributed tracing (e.g. Jaeger) to observe IOx.

This has at least two pieces:
1. Respect the incling request headers to connect spans made in IOx to the overall trace id
2. For Query execution, add visibility into which operators spend what time.

For query execution visibility, we hope to see something like the following in our Jaeger instance:
<img width="689" alt="Screen Shot 2021-08-12 at 11 14 33 AM" src="https://user-images.githubusercontent.com/490673/129233137-0f08660b-2aeb-4f00-bb1c-8573da4a7ec7.png">

More details [here](https://docs.google.com/document/d/1oCWaiEAGPHpAFXagh3sBsu3PjshwfwEgn-qrBSGvg5A/edit?usp=sharing)

alamb:
I think this one is done now -- @tustvold  let me know if that is not correct.

tustvold:
Making a note of https://github.com/influxdata/influxdb_iox/issues/2549 and https://github.com/influxdata/influxdb_iox/issues/2367 which track further work in this space

Issue #2192, opened 2021-08-05
Make compiling opentelementry-* depedencies optional


alamb:
__Use case:__
During normal development of IOx as a developer I want faster compile / link times so I can iterate on changes and be more productive. The more crates IOx depends on the longer each iteration cycle takes

The `opentelemetry-*` set of crates adds many useful features (like sending metrics to jaeger and open telemetry collectors) that are important for production use but are not used during the course of normal development. These crates have many additional dependencies (e.g. they all use tonic/prost) and some are using different versions - e.g. in https://github.com/influxdata/influxdb_iox/pull/2172

Also, the plan is for our internal system, `conductor` to use IOx’s observability crates (trogging, observability_deps, metrics) so making it easy to choose to include support for opentelemetry or not will be helpful

__Proposal:__
Follow the model for the object store functionality where AWS / S3 / Azure support is not enabled by default but must be compiled in via a feature flag. 

THe proposal is to  feature gate as many of the “opentelemetry” crates as possible under a new `opentelemetry` feature in both IOx and the dependent crates and this

The `opentelemetry` feature flag would default to false, but we would enable it for production builds

Without `opentelemtry`  enabled, only logging to stderr would be supported. Any attempt to use other open telemetry features will result in an error telling the user that IOx needs to be recompiled with the `opentelemetry` feature 

I will try and consolidate stuff in the observability / trogging / metrics crate (edited) 

__Current behaviour:__
Compile open telemetry all the time

__Desired behaviour:__
Only compile the opentelemetry crates when we want to use them

__Alternatives considered:__
Describe other solutions or features you considered.

tustvold:
>  logging to stderr

Why not stdout? It would be nice to keep tokio-tracing in debug builds, as I use its filtering frequently, but just strip out the jaeger/otel stuff?

> like sending metrics to jaeger and open telemetry collectors

Definitely supportive of this - would be good to keep prometheus support though. For one thing the tests need it, but also because it is a useful tool for local development (it's like a more detailed system table), and should be a very lightweight dependency.

alamb:
> Why not stdout? It would be nice to keep tokio-tracing in debug builds, as I use its filtering frequently, but just strip out the jaeger/otel stuff?

Yes, I should have said "keep doing the same thing as we do to day when `LOGGING_LEVEL` is set or `-v` / `-vv` are passed on the command line

tustvold:
> @tustvold how do you configure / enable logging during development?

[...]

PR #2274, opened 2021-08-12
feat: add trace collection plumbing (#2273)

tustvold:
This adds the initial plumbing for distributed tracing within IOx. Currently this is just used to create a single root span for a gRPC request, and "publish" it by logging it to stdout. As this is just a tower layer, that instruments http requests, it should be fairly straightforward to inject for the HTTP API as well.

This is the first step of a multi-step process to get jaeger traces within IOx, and will allow confirming that trace context is reaching IOx in our production environments before moving forwards.

Subsequent steps will:

* Hook up opentelemetry-jaeger as a TraceCollector
* Propagate SpanContext onto IOxExecutionContext
* Collect DataFusion metrics within IOxExecutionContext and construct corresponding Spans

This PR notably does not use tokio-tracing, and if approved, will move IOx away from using tokio tracing for distributed tracing. 

There are a couple of reasons for this decision:

* A number of different team members have spent significant lengths of time trying to get it working and have been, as far as I can tell, unsuccessful
* The crate is extremely difficult to understand - not least because current master renames a number of key concepts
* Most in the wild uses of tokio-tracing appear to be focused around logging, with the design of its filter stack, concept of levels, etc... fitting this sort of use case. Unsurprisingly this is where the library shines, and I would like to keep using it for this
* I couldn't see an obvious way to hook into this system to publish artificial, metrics-derived span
* Despite tower-http having a trace layer, it doesn't appear to have the ability to extract trace context from [inbound requests](https://github.com/tower-rs/tower-http/blob/master/tower-http/src/trace/make_span.rs#L58) - which is a key part of implementing distributed tracing (I'm not sure if I'm missing something here)

The TLDR is, rather than trying to fight an already fairly complex crate to do something that doesn't appear to have been an initial design goal, we can get something working quicker by just implementing something custom. It is possible the ecosystem will evolve, and we can adopt whatever consensus is reached when that happens, but right now I don't think the upstream tooling is where we need it to be.

Another point worth noting is that the crate purposefully does not depend on OpenTelemetry, this is to fit in with the eventual goal in #2192 of reducing the coupling of IOx to large instrumentation libraries. The exposed data model is extremely similar, and should be easy to feed into [opentelemetry-jaeger](https://docs.rs/opentelemetry-jaeger/0.15.0/opentelemetry_jaeger/struct.Exporter.html) or any other OTEL exporter.


trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
...
+        ))
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct SpanId(NonZeroU64);
+
+impl SpanId {
+    pub fn gen() -> Self {
+        // Should this be a UUID?
```
tustvold, line 71: This is what the upstream OTEL crate does so :shrug:

trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
...
+            },
+            start: None,
+            end: None,
+            status: SpanStatus::Unknown,
+            metadata: Default::default(),
+            events: Default::default(),
+        }
+    }
+
+    pub fn from_headers(
```
tustvold, line 123: This is what extracts the [Span Context](https://github.com/openzipkin/b3-propagation#overview) from the request headers

trace/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use crate::span::Span;
+use observability_deps::tracing::info;
+
+pub mod ctx;
+pub mod span;
+pub mod tower;
+
+/// A TraceCollector is a sink for completed `Span`
+pub trait TraceCollector: std::fmt::Debug + Send + Sync {
```
tustvold, line 18: This is the place where different collectors can be injected. Currently there is only a dummy log one, but I'd envisage implementations to write to jaeger, buffer up a list for testing, etc...

trace/src/span.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// A `Span` is a representation of a an interval of time spent performing some operation
+///
+/// A `Span` has a name, metadata, a start and end time and a unique ID. Additionally they
+/// have relationships with other Spans that together comprise a Trace
+///
+/// On Drop a `Span` is published to the registered collector
+///
+#[derive(Debug)]
+pub struct Span<'a> {
```
tustvold, line 23: This is what the entire crate is about, producing and exporting this data model

trace/src/tower.rs
```diff
@@ -0,0 +1,174 @@
...
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut request: Request<ReqBody>) -> Self::Future {
+        let span = match SpanContext::from_headers(&self.collector, request.headers()) {
+            Ok(Some(ctx)) => {
+                let span = ctx.child("IOx");
+
+                // Add context to request for use by service handlers
+                request.extensions_mut().insert(span.ctx.clone());
```
tustvold, line 64: Service implementations will be able to extract this field from the Request they are passed and optionally create more spans, hand it off to the IOxExecutionContext, etc...

This trades off a bit more plumbing, compared to using something like thread-local storage, for slightly less magic. Especially as queries cross thread pools, having the explicit plumbing to ensure the context propagates is worthwhile IMO

trace/src/tower.rs
```diff
@@ -0,0 +1,174 @@
...
+    type Response = Response<TraceBody<ResBody>>;
+    type Error = S::Error;
+    type Future = TraceFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut request: Request<ReqBody>) -> Self::Future {
+        let span = match SpanContext::from_headers(&self.collector, request.headers()) {
```
tustvold, line 59: This is where we extract the trace context from the request headers

trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(Some(SpanContext {
+            trace_id: required_header(headers, B3_TRACE_ID_HEADER)?,
+            parent_span_id: decoded_header(headers, B3_PARENT_SPAN_ID_HEADER)?,
+            span_id: required_header(headers, B3_SPAN_ID_HEADER)?,
+            collector: Arc::clone(&collector),
+        }))
+    }
+}
+
+fn get_header<'a>(
```
tustvold, line 152: I will write some tests for this decode logic before merging

trace/src/span.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl<'a> Drop for Span<'a> {
+    fn drop(&mut self) {
+        self.ctx.collector.export(self)
+    }
+}
+
+/// Values that can be stored in a Span's metadata and events
+#[derive(Debug, Clone, Copy)]
+pub enum Meta<'a> {
```
alamb, line 58: Maybe calling this `MetaValue` would be more precise. I don't feel strongly and we can always change it if needed

trace/src/span.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl<'a> EnteredSpan<'a> {
+    pub fn new(mut span: Span<'a>) -> Self {
+        span.start = Some(SystemTime::now());
+        Self { span }
+    }
+}
+
+impl<'a> Drop for EnteredSpan<'a> {
+    fn drop(&mut self) {
```
alamb, line 110: This is cool / clever use of drop to automatically update timestamps and publish the spans

src/influxdb_ioxd/rpc.rs
```diff
@@ -84,7 +87,8 @@ where
         .build()
         .context(ReflectionError)?;
 
-    let mut builder = tonic::transport::Server::builder();
+    let builder = tonic::transport::Server::builder();
+    let mut builder = builder.layer(trace::tower::TraceLayer::new(trace_collector));
```
alamb, line 91: The interface / wiring certainly looks nice

trace/src/ctx.rs
alamb: Maybe we could use a global counter (`AtomicU64`) type thing with fetch_add to ensure no collisions

The likelihood of that this will cause a problem is probably close to zero, so I agree with the 🤷  and it likely doesn't matter

trace/src/ctx.rs
alamb:
```suggestion
    /// Create a  SpanContext for the trace described in the request's headers
    pub fn from_headers(
```

alamb: Maybe it is also worth mentioning that it looks for headers that conform to  B3  (or whatever that convention is called) and add a link to https://github.com/openzipkin/b3-propagation

trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
...
+        .get(header)
+        .map(|value| {
+            value
+                .to_str()
+                .map_err(|source| ContextError::InvalidUtf8 { header, source })
+        })
+        .transpose()
+}
+
+fn decoded_header<T: FromStr<Err = DecodeError>>(
```
alamb, line 166: I was confused about what was being decoded here -- it looks like this is doing something more like `parsed_header` (as it is invoking `from_str`)

trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
...
+        .transpose()
+        .map_err(|source| ContextError::HeaderDecodeError { source, header })
+}
+
+fn required_header<T: FromStr<Err = DecodeError>>(
+    headers: &HeaderMap,
+    header: &'static str,
+) -> Result<T, ContextError> {
+    decoded_header(headers, header)?.ok_or(ContextError::Missing { header })
+}
```
alamb, line 181: I would think this module could easily have some unit tests that show creating spans from headers. I think that might be worth investing in so that if/when we have to change this code, we don't have to test in production that the tracing is working as expected

Perhaps that is what you meant by "write some tests for this decode logic"

trace/src/lib.rs
alamb: 👍

trace/src/tower.rs
```diff
@@ -0,0 +1,174 @@
+use crate::{ctx::SpanContext, span::EnteredSpan, TraceCollector};
```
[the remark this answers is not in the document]
alamb: I think this is valuable context to have for future readers reading this code -- perhaps you can copy/paste it into a doc comment for hte module

trace/src/ctx.rs
```diff
@@ -0,0 +1,181 @@
+use std::num::{NonZeroU128, NonZeroU64, ParseIntError};
```
alamb, line 1: Is there ever a usecase for generating spans traces or spans independently ? Like for IOx to create the traceID and root span id?

trace/src/tower.rs
```diff
@@ -0,0 +1,173 @@
...
+                Some(EnteredSpan::new(span))
+            }
+            Ok(None) => None,
+            Err(e) => {
+                error!(%e, "error extracting trace context from request");
+                None
+            }
+        };
+
+        TraceFuture {
```
alamb, line 75: Maybe calling this Trace**d**Future would make my 5th grade English teacher happy? Though I suppose I shouldn't be offering English suggestions to the English 😆

trace/src/tower.rs
alamb: Another reason to make it explicit is that  thread local storage is unlikely to work correctly. As I understand the `async`/`await` model, after any `await` the code may be running on a different thread (if any of the futures involved in `await` do some sort of async operation).

trace/src/tower.rs
```diff
@@ -0,0 +1,173 @@
...
+    ResBody: http_body::Body,
+{
+    type Output = Result<Response<TraceBody<ResBody>>, Error>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result = ready!(self.as_mut().project().inner.poll(cx));
+        if let Some(span) = self.as_mut().project().span.as_mut() {
+            match &result {
+                Ok(_) => span.event("request processed"),
+                Err(_) => span.error("error processing request"),
```
alamb, line 104: I may take a friendly look at (copy) this code for doing something similar to trace the output of DataFusion plans

trace/src/tower.rs
```diff
@@ -0,0 +1,173 @@
...
+            Ok(response) => Poll::Ready(Ok(response.map(|body| TraceBody {
+                span: self.as_mut().project().span.take(),
+                inner: body,
+            }))),
+            Err(e) => Poll::Ready(Err(e)),
+        }
+    }
+}
+
+/// `TraceBody` wraps a `http_body::Body` and instruments it
```
alamb, line 118: Likewise. maybe Trace**d**Body would be a better name. A "niggle" (love that word) to be sure

trace/src/tower.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        }
+        Poll::Ready(Some(result))
+    }
+
+    fn poll_trailers(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<Option<http::header::HeaderMap>, Self::Error>> {
+        // TODO: Classify response status and set SpanStatus
```
alamb, line 154: Is this still a todo? It seems like it is done

alamb (approved):
I think this looks great @tustvold  nice work. I think the abstractions are clean and using this in IOx will be straightforward. 

I suggest testing:
1. The HTTP header parsing code
2. an end-to-end test that sends a request with appropriate headers and validates that the trace and span contexts were extracted correctly

I suggest someone that is more familiar with how tracing in other languages and our internal systems at InfluxData (perhaps @jacobmarble  and/or @e-dard  / @mkmik ) review this too as my review wouldn't likely catch edge cases or be able to review the actual contents of what is being produced and consumed. 

> we can get something working quicker by just implementing something custom. It is possible the ecosystem will evolve, and we can adopt whatever consensus is reached when that happens, but right now I don't think the upstream tooling is where we need it to be.

This decision and the rationale makes a lot of sense to me. 

> Another point worth noting is that the crate purposefully does not depend on OpenTelemetry, this is to fit in with the eventual goal in #2192 of reducing the coupling of IOx to large instrumentation libraries. 

I am also a large fan of this approach -- have an IOx specific interface for spans / etc that we can plug various backends into. It may take a little more initial development time but will likely make it easier to integrate with different systems as well as keeping the code easier to understand IMHO. 💯 ``

src/influxdb_ioxd/rpc.rs
```diff
@@ -76,6 +76,9 @@ pub async fn serve<M>(
 where
     M: ConnectionManager + Send + Sync + Debug + 'static,
 {
+    // TODO: Replace this with a jaeger collector
```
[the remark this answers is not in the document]
tustvold: Yeah - this TODO is referring to the concrete implementation being used, not the fact there is a trace collector

trace/src/ctx.rs
tustvold: If you have two IOx's in the same trace, there would be a high(er) chance of them generating a the same span id with an AtomicU64. I think it needs to be random, but you could construct the randomness to make it even less likely for a collision.

trace/src/ctx.rs
tustvold: Yup - tests added

trace/src/ctx.rs
tustvold: Renamed - and added docstrings

trace/src/tower.rs
tustvold: > thread local storage is unlikely to work correctly.

Yes, this is why tokio::tracing has to instrument and wrap futures, in order to ensure that the thread-local "span stack" is pushed on future poll, and popped on return. Manual plumbing avoids all of this complexity

trace/src/tower.rs
tustvold: This doesn't look at the actual status code returned in the trailer - e.g. 200, 404, 503, etc... which it should but currently doesn't

tustvold:
> The HTTP header parsing code

Tests for this, and some other things, have been added PTAL

> an end-to-end test that sends a request with appropriate headers and validates that the trace and span contexts were extracted correctly

To write a true end-to-end test of this I need #2283 , currently there isn't a way to make a request to the IOx API with custom headers :sweat_smile:

tustvold:
(experimenting with just using serde_json instead of a load of custom display impl)

trace/src/ctx.rs
```diff
@@ -0,0 +1,277 @@
...
+        assert_eq!(
+            SpanContext::from_headers(&collector, &headers)
+                .unwrap_err()
+                .to_string(),
+            "error decoding header 'X-B3-SpanId': value decode error: invalid digit found in string"
+        );
+
+        headers.insert(B3_SPAN_ID_HEADER, HeaderValue::from_static("0"));
+
+        assert_eq!(
```
alamb, line 270: tests look good

alamb (approved):

PR #2276, opened 2021-08-12
refactor: gRPC connection in single location

tustvold:
As part of #2273 I want to add the ability to add arbitrary headers to requests sent by the CLI, this will allow sending requests with trace information set. 

This PR rejigs the CLI plumbing so that the gRPC connection is created in one place, to allow the headers, and any future connection options, to be plumbed in a single place


alamb (approved):
looks like a nice cleanup to me

alamb:
Note at least one of the tests shows a less useful error message

```
---- end_to_end_cases::sql_cli::test_error_connecting stdout ----
thread 'end_to_end_cases::sql_cli::test_error_connecting' panicked at 'Unexpected stderr, failed var.contains("Error connecting to http://hope_this_addr_does_not_exist")
└── var as str: Error connecting to IOx: Connection error: transport error
```

It would be nice if the error continued to have the address that could not be connect to

PR #2278, opened 2021-08-12
refactor: Extract different types for the paths that databases use in object storage

### LINKED ISSUE #2193: Relative paths in `parquet_file`

carols10cents:
Fixes #2193.

This centralizes all knowledge of what the paths are for different kinds of files for a database to the iox_object_store crate. The *contents* of the files are still managed by other parts of the code, but if we want to change how paths are constructed, now there's one spot.


iox_object_store/src/lib.rs
```diff
@@ -151,16 +188,172 @@ impl IoxObjectStore {
...
+            chunk_id,
+            uuid,
+        })
+    }
+
+    // Deliberately private; this transformation should only happen within this crate
+    fn from_absolute(absolute_path: Path) -> Result<Self, ParquetFilePathParseError> {
+        let absolute_path: DirsAndFileName = absolute_path.into();
+
+        let mut absolute_dirs = absolute_path.directories.iter();
```
shepmaster, line 312: Should `.fuse()` when we aren't checking the result of `next()`

parquet_file/src/chunk.rs
```diff
@@ -112,22 +110,20 @@ pub struct ParquetChunk {
...
         table_name: Arc<str>,
         partition_key: Arc<str>,
         metrics: ChunkMetrics,
     ) -> Result<Self> {
-        let schema = parquet_metadata.read_schema().context(SchemaReadFailed {
-            path: &file_location,
-        })?;
+        let schema = parquet_metadata
+            .read_schema()
+            .context(SchemaReadFailed { path: path.clone() })?;
```
shepmaster, line 123: Ideally `path` implements `Into` and this works:

```suggestion
            .context(SchemaReadFailed { path })?;
```

Otherwise:

```suggestion
            .with_context(|| SchemaReadFailed { path: path.clone() })?;
```

iox_object_store/src/lib.rs
```diff
@@ -103,6 +92,54 @@ impl IoxObjectStore {
...
+    pub async fn parquet_files(&self) -> Result<BoxStream<'static, Result<Vec<ParquetFilePath>>>> {
+        Ok(self
+            .list(Some(&self.data_path.inner))
+            .await?
+            .map_ok(move |list| {
+                list.into_iter()
+                    .map(|path| ParquetFilePath::from_absolute(path))
+                    // This `flatten` ignores any filename in the data_path we couldn't parse as
+                    // a ParquetFilePath
+                    .flatten()
```
shepmaster, line 105: I think you just moved this, but `flat_map` is more concise.

carols10cents:
It is indeed ready for review. Thank you!

iox_object_store/src/paths/transaction_file.rs
```diff
@@ -0,0 +1,390 @@
...
+/// File suffix for transaction files in object store.
+const TRANSACTION_FILE_SUFFIX: &str = "txn";
+
+/// File suffix for checkpoint files in object store.
+const CHECKPOINT_FILE_SUFFIX: &str = "ckpt";
+
+/// Location of a catalog transaction file within a database's object store.
+/// The exact format is an implementation detail and is subject to change.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct Path {
```
alamb, line 18: FWIW I personally find the re-use of a common struct name like `Path` within different modules confusing -- even though the names are distinct within their module namespaces, I find it makes reading code somewhere that is not an editor that can look up the definition for you (aka the github UI) very challenging for me. 

The same applies to re-aliasing of `use Path as FooPath` -- I find  mentally challenging as my mind sees `FooPath` and thinks "ok, this must be some new concept I haven't seen before" rather than "oh that is a different name for Path in this context) -- aka my mind doesn't keep up with aliases like it used to (or maybe it never did 😆 )

But I realize this is a matter of preference / style / mental fitness 🦾 

Also while on my soapbox, I wonder why follow different conventions for `RootPath` and `transaction_file::Path` -- so like why not `root::Path` instead of `RootPath`?

iox_object_store/src/paths.rs
```diff
@@ -0,0 +1,170 @@
...
+                .file_name
+                .expect("Transaction file paths have filenames")
+                .to_string(),
+        );
+        result
+    }
+}
+
+/// A database-specific object store path for all data files. This should not be leaked outside
+/// this crate.
```
alamb, line 80: if it shouldn't be leaked outside the crate, perhaps it could be marked `pub(crate)` ?

server/src/database.rs
```diff
@@ -104,7 +99,7 @@ impl Database {
     ///
     /// This is backed by an existing database, which was [created](Self::create) some time in the past.
     pub fn new(application: Arc<ApplicationState>, config: DatabaseConfig) -> Self {
-        info!(db_name=%config.name, store_prefix=%config.store_prefix.to_string(), "new database");
+        info!(db_name=%config.name, "new database");
```
alamb, line 102: it might still be useful to log the root path of the object store here so the operator knows where in object store contains the transactions and data

alamb (approved):
I think this is really cool @carols10cents  / @shepmaster  👨‍🍳 👌  

I won't say I was able to review every last line of this PR with incredible detail, but I did read it as carefully as I could and I thought it looked great.  🎉

parquet_file/src/cleanup.rs
```diff
@@ -195,42 +179,31 @@ mod tests {
...
             };
             transaction.add_parquet(&info).unwrap();
             transaction.remove_parquet(&info.path);
-            let path_string = iox_object_store
-                .path_from_dirs_and_filename(info.path.clone())
-                .to_string();
-            paths_keep.push(path_string);
-
-            // not a parquet file => keep
-            let mut path = info.path;
```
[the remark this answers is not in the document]
carols10cents: Correct.

iox_object_store/src/paths.rs
carols10cents: It's actually effectively already `pub(crate)` because the `paths` module is private and `RootPath` is only `use`d in lib.rs, not `pub use`d, but I agree `pub(crate)` conveys that information more clearly in this spot :)

iox_object_store/src/paths/transaction_file.rs
carols10cents: I'm totally up for changing the names, I see what you're saying once you're on the reading side of this code (which we all will be for much longer than the time it's taken me to write this ;)) My brain doesn't like keeping a lot of stuff in RAM either. 

I mostly did it this way because I wanted to separate out the more complicated paths into their own files and `transaction_file::TransactionFilePath` felt a bit redundant. The module name can still be an implementation detail though, so I'm cool with making this change!

parquet_file/src/catalog.rs
```diff
@@ -2582,36 +2463,6 @@ mod tests {
             .unwrap();
     }
 
-    #[tokio::test]
```
crepererum, line 2585: what happened to this test?

parquet_file/src/catalog.rs
```diff
@@ -1617,56 +1555,6 @@ mod tests {
         assert!(option.is_none());
     }
 
-    #[tokio::test]
```
crepererum, line 1620: what happened to this test?

parquet_file/src/catalog.rs
```diff
@@ -32,19 +35,17 @@ use uuid::Uuid;
 /// For breaking changes, this will change.
 pub const TRANSACTION_VERSION: u32 = 11;
```
crepererum, line 36: sorry for being late, but we should probably bump this version, otherwise we now have existing catalogs that have absolute paths but we now think the that the paths are relative.

PR #2279, opened 2021-08-12
fix: Add "heappy" feature to enable heap profiling, new test

NGA-TRAN:
Closes #

See https://github.com/influxdata/influxdb_iox/pull/2279#issuecomment-898449111

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -192,8 +192,8 @@ jobs:
           name: Build benches
           command: cargo test --workspace --benches --no-run
       - run:
-          name: Build with object store + exporter support
-          command: cargo build --features="aws,gcp,azure,jaeger,otlp"
+          name: Build with object store + exporter support + HEAP profiling
+          command: cargo build --features="aws,gcp,azure,jaeger,otlp,heappy"
       - cache_save
 
   # Lint protobufs.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,27 @@ jobs:
           command: cargo test --workspace
       - cache_save
 
+  # end to end tests with Heappy (heap profiling enabled)
+  test_heappy:
+    docker:
+      - image: quay.io/influxdb/rust:ci
+    resource_class: xlarge # use of a smaller executor tends crashes on link
+    environment:
+        # Disable incremental compilation to avoid overhead. We are not preserving these files anyway.
+        CARGO_INCREMENTAL: "0"
+        # Disable full debug symbol generation to speed up CI build
+        # "1" means line tables only, which is useful for panic tracebacks.
+        RUSTFLAGS: "-C debuginfo=1"
+        RUST_BACKTRACE: "1"
+    steps:
+      - checkout
+      - rust_components
+      - cache_restore
+      - run:
+          name: End to end tests (with heappy)
+          command: cargo test --features=heappy end_to_end
+      - cache_save
+
   test_kafka_integration:
     machine: true
     resource_class: xlarge
@@ -314,6 +335,7 @@ workflows:
       - lint
       - protobuf-lint
       - test
+      - test_heappy
       - test_kafka_integration
       - test_influxdb2_client
       - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,10 +273,10 @@ jobs:
       - cache_restore
       - run:
           name: Print rustc target CPU options
-          command: cargo run --release --features="aws,gcp,azure,jaeger,otlp" --bin print_cpu
+          command: cargo run --release --features="aws,gcp,azure,jaeger,otlp,heappy" --bin print_cpu
       - run:
           name: Cargo release build with target arch set for CRoaring
-          command: cargo build --release --features="aws,gcp,azure,jaeger,otlp"
+          command: cargo build --release --features="aws,gcp,azure,jaeger,otlp,heappy"
       - run: |
           echo sha256sum after build is
           sha256sum target/release/influxdb_iox
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ datafusion = { path = "datafusion" }
 data_types = { path = "data_types" }
 entry = { path = "entry" }
 generated_types = { path = "generated_types" }
-heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384" , features = ["enable_heap_profiler"]}
+heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384", features = ["enable_heap_profiler", "jemalloc_shim"], optional = true}
 
 influxdb_iox_client = { path = "influxdb_iox_client", features = ["format"] }
 influxdb_line_protocol = { path = "influxdb_line_protocol" }
@@ -147,3 +147,5 @@ gcp = ["object_store/gcp"] # Optional GCP object store support
 aws = ["object_store/aws"] # Optional AWS / S3 object store support
 jaeger = ["trogging/jaeger"] # Enable optional jaeger tracing support
 otlp = ["trogging/otlp"] # Enable optional open telemetry collector
+# heappy is also an optional feature; Not on by default as it
+# runtime overhead on all allocations (calls to malloc)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -110,7 +110,7 @@ serde_urlencoded = "0.7.0"
 snafu = "0.6.9"
 structopt = "0.3.21"
 thiserror = "1.0.23"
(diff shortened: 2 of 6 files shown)


NGA-TRAN:
@alamb @mkmik and @tustvold This PR turn heappy on. Can you have a look?

mkmik (approved):

src/influxdb_ioxd/http.rs
```diff
@@ -790,7 +796,8 @@ async fn dump_rsprof(seconds: u64, frequency: i32) -> pprof::Result<pprof::Repor
     guard.report().build()
 }
 
-async fn dump_heappy_rsprof(seconds: u64, frequency: i32) -> HeapReport {
+#[cfg(feature = "heappy")]
```
alamb, line 799: I think we may also need to add `heappy` as a feature flag and also mark `optional=true`

alamb: As written this PR I think will set heappy on for all builds

mkmik: I thought that was the purpose of this PR 

I meant to say "no let's make this optional" but then I thought that perhaps we could just turn it on and see if it does cause any problems (consume more CPU etc)

NGA-TRAN: @alamb and @mkmik : 
- I am not clear what extra things I have to do here with optional. Should I do anything extra?
- I also see the [global_allocator](#[global_allocator]). DO I have to do any thing extra to disable it? Seems like no need, right?

mkmik: basically this is what I did:

`Cargo.lock`:

```
-tikv-jemallocator = {version = "0.4.0", features = ["unprefixed_malloc_on_supported_platforms"] }
+tikv-jemallocator = {version = "0.4.0" }
```

```
-heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384" }
+heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384" , features = ["enable_heap_profiler"]}
```


`src/main.rs`:

```
-#[global_allocator]
-static GLOBAL: Jemalloc = Jemalloc;
```

mkmik: I _think_ we can safely remove the `static GLOBAL: Jemalloc = Jemalloc;` unconditionally (I'll file a separate PR for that).

The tricky bit is make the aforementioend changes conditional on a feature flag (or to just enable heappy in our builds and disable it later if it proves to slow us down)

alamb:
Ok, I just pushed a commit that makes heappy optional and I think makes it work as expected

If you run IOx like this (default, no features):
```shell
cargo run -- run -v  --object-store=file --data-dir=$HOME/.influxdb_iox --server-id=42
```

Going to `http://localhost:8080/debug/pprof/allocs` results in:
Screen Shot 2021-08-13 at 7 59 45 AM


Then if you run with the `heappy` feature (note the `--features=heappy` flag):
```shell
cargo run --features=heappy -- run -v  --object-store=file --data-dir=$HOME/.influxdb_iox --server-id=42
```

Going to `http://localhost:8080/debug/pprof/allocs` results in:

Screen Shot 2021-08-13 at 8 08 21 AM

alamb:
Note that I didn’t test on Linux — we should probably do so prior to merging this PR so that we don’t break the production deployment

src/influxdb_ioxd/http.rs
```diff
@@ -897,6 +891,16 @@ async fn pprof_heappy_profile<M: ConnectionManager + Send + Sync + Debug + 'stat
     Ok(Response::new(Body::from(body)))
 }
 
+//  Return error if heappy not enabled
+#[cfg(not(feature = "heappy"))]
+#[tracing::instrument(level = "debug")]
+async fn pprof_heappy_profile<M: ConnectionManager + Send + Sync + Debug + 'static>(
+    req: Request<Body>,
+) -> Result<Response<Body>, ApplicationError> {
+    #[cfg(not(feature = "heappy"))]
```
NGA-TRAN: So we need to have `#[cfg(not(feature = "heappy"))]` on top of the function and right before return error?

src/influxdb_ioxd/http.rs
```diff
@@ -897,6 +891,16 @@ async fn pprof_heappy_profile<M: ConnectionManager + Send + Sync + Debug + 'stat
     Ok(Response::new(Body::from(body)))
 }
 
+//  Return error if heappy not enabled
+#[cfg(not(feature = "heappy"))]
+#[tracing::instrument(level = "debug")]
```
NGA-TRAN, line 896: This means we check the feature and only allows it in debug mode?

.circleci/config.yml
```diff
@@ -131,6 +131,27 @@ jobs:
...
+        RUST_BACKTRACE: "1"
+    steps:
+      - checkout
+      - rust_components
+      - cache_restore
+      - run:
+          name: End to end tests (with heappy)
+          command: cargo test --features=heappy end_to_end
+      - cache_save
+
```
NGA-TRAN, line 154: 👍

NGA-TRAN (commented):
Thanks @alamb

src/influxdb_ioxd/http.rs
alamb: No the second one in the body was a left over (I have removed it)

Basicaly the way feature flags works is that

```
    #[cfg(not(feature = "heappy"))]
```

Has a similar meaning to  `#ifndef FEATURE_HEAPPY` C++. In Rust it means "skip the entire next statement if the "heappy" feature is not enabled. So if it is on a module the module isn't compiled. If it is on a function definition the entire function isn't compiled

src/influxdb_ioxd/http.rs
alamb: no, the `#[cfg(not(feature = "heappy"))]` has the meaning as above.

The 
```rust
#[tracing::instrument(level = "debug")]
```

is some sort of macro magic from the tracing crate that prints out a log message every time this function is called

src/influxdb_ioxd/http.rs
```diff
@@ -1221,7 +1224,7 @@ mod tests {
         let response = client
             .get(&format!(
                 "{}/iox/api/v1/databases/MyOrg_MyBucket/query?q={}&format=json",
-                server_url, "select%20*%20from%20h2o_temperature"
+                server_url, "select%20*%20from%20h2o_temperature%20order%20by%20surface_degrees"
```
NGA-TRAN, line 1227: Add `order by` to have deterministic output to avoid flaky test failures

Cargo.toml
```diff
@@ -59,7 +59,7 @@ datafusion = { path = "datafusion" }
 data_types = { path = "data_types" }
 entry = { path = "entry" }
 generated_types = { path = "generated_types" }
-heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384" }
+heappy = { git = "https://github.com/mkmik/heappy", rev = "39530c987aecda0b2d76503e8b5bb33d8a93f384", features = ["enable_heap_profiler", "jemalloc_shim"], optional = true}
```
mkmik, line 62: enable_heap_profiler already pulls in the jemalloc_shim

Cargo.toml
```diff
@@ -110,7 +110,7 @@ serde_urlencoded = "0.7.0"
 snafu = "0.6.9"
 structopt = "0.3.21"
 thiserror = "1.0.23"
-tikv-jemallocator = {version = "0.4.0", features = ["unprefixed_malloc_on_supported_platforms"] }
+tikv-jemallocator = {version = "0.4.0" }
```
mkmik, line 113: We should remove `features = ["unprefixed_malloc_on_supported_platforms"]` only when heappy is enabled.

Otherwise our memory tracking metrics (which depend on jemalloc statistics) will be off because they won't be capturing croaring bitmaps

.circleci/config.yml
```diff
@@ -273,10 +273,10 @@ jobs:
       - cache_restore
       - run:
           name: Print rustc target CPU options
-          command: cargo run --release --features="aws,gcp,azure,jaeger,otlp" --bin print_cpu
+          command: cargo run --release --features="aws,gcp,azure,jaeger,otlp,heappy" --bin print_cpu
```
mkmik, line 276: @NGA-TRAN @alamb PTAL

mkmik (approved):

Cargo.toml
alamb: somehow this comment got missed prior to merging... filed https://github.com/influxdata/influxdb_iox/issues/2288 to track

.circleci/config.yml
mkmik: @alamb this is what should close #2288 . It's brittle but Nga and I decided at this stage it was worth getting the heap profiler out for the staging environment while we're not yet cutting public releases.

PR #2280, opened 2021-08-13
refactor: isolate CLI handling for object store

crepererum:
The object store handling is a large part of our CLI and kinda inlined
into the rest of the setup routines. This commit moves the object store
related code into its own module. As a bonus it can now be reused by
other CLI commands (e.g. low-level debug tools).

There are no functional changes here except for slight changes in the
error structure.


mkmik (approved):
thanks!

PR #2282, opened 2021-08-13
feat: manual skip replay, part 2/2

crepererum:
This contains:

- gRPC interface
- CLI
- all the stuff required for the end2end test


docker/integration_test.sh
```diff
@@ -3,4 +3,5 @@
 set -euxo pipefail
 
 cargo test -p write_buffer kafka -- --nocapture
+cargo test -p influxdb_iox --test end_to_end skip_replay -- --nocapture
```
alamb, line 6: 👍

src/influxdb_ioxd/rpc/error.rs
```diff
@@ -45,22 +45,15 @@ pub fn default_server_error_handler(error: server::Error) -> tonic::Status {
...
         Error::RemoteError { source } => tonic::Status::unavailable(source.to_string()),
-        Error::WipePreservedCatalog { source } => match source {
-            server::database::Error::TransitionInProgress { .. } => tonic::Status::unavailable("recovery operation already in progress"),
-            server::database::Error::InvalidState { .. } => tonic::Status::failed_precondition(source.to_string()),
-            error => {
-                error!(?error, "Unexpected error wiping catalog");
-                InternalError {}.into()
-            }
-        }
+        Error::WipePreservedCatalog { source } => default_database_error_handler(source),
```
alamb, line 48: Should this also handle `ReplayError` and/or `ReplayPlanError`?

tests/end_to_end_cases/management_cli.rs
```diff
@@ -692,6 +693,49 @@ async fn test_wipe_persisted_catalog_error_db_exists() {
         .stderr(predicate::str::contains(&expected_err));
 }
 
+#[tokio::test]
+async fn test_skip_replay() {
```
alamb, line 697: 👍

tests/end_to_end_cases/scenario.rs
```diff
@@ -620,3 +634,134 @@ pub async fn fixture_broken_catalog(db_name: &str) -> ServerFixture {
...
+                .unwrap(),
+            0,
+        )
+        .await
+        .unwrap();
+
+    // wait for ingest
+    let t_0 = Instant::now();
+    loop {
+        // use later partition here so that we can implicitely wait for both entries
```
alamb, line 720: given there are three entries (and two for `partition_by=b` is this potentially a race condition where the final entry (second for `partition_by=b`) may not be stored?

Maybe that doesn't matter

tests/end_to_end_cases/scenario.rs
```diff
@@ -620,3 +634,134 @@ pub async fn fixture_broken_catalog(db_name: &str) -> ServerFixture {
...
+        .unwrap();
+
+    // Try to replay and error
+    let fixture = fixture.restart_server().await;
+
+    let status = fixture.wait_server_initialized().await;
+    assert_eq!(status.database_statuses.len(), 1);
+
+    let load_error = &status.database_statuses[0].error.as_ref().unwrap().message;
+    assert_contains!(load_error, "error during replay: Cannot replay");
```
alamb, line 764: 👨‍🍳 👌  -- very nice test setup

alamb (approved):
This looks great to me. Nice work @crepererum

tests/end_to_end_cases/scenario.rs
crepererum: No race condition because we have a single kafka topic w/ a single partition and read + progress from a single sequencers are serial.

src/influxdb_ioxd/rpc/error.rs
crepererum: In theory we could handle this here but it cannot occur in practice. These errors here can only occur as a reaction to a gRPC call. The only DB-init things that you can do via gRPC are:

- create a fresh DB: no replay, no catalog, no replay plan
- recover a DB via wiping the catalog: then there's no catalog left and no replay plan and no replay
- recover a DB via skipping replay: there was a replay plan constructed already, but we "skip replay" says "don't use it". replay error cannot occur because we would get a `SkipReplay` error in this particular case (e.g. when the kafka client cannot seek).

Overall, that's obviously a lot of specific knowledge and I somewhat dislike our error handling here because whenever we forget to wire up a specific case the user only receives an "internal error" which is not helpful. Maybe we should make all these matches exhaustive and w/o a default case. If we decide so, I would do that in a separate PR though.

tests/end_to_end_cases/scenario.rs
alamb: right but my point was since the data is read entry at a time from the sequencer and there are 2 entries for `partition_by=b` the `loop` that is waiting may terminate after only the first entry is read

src/influxdb_ioxd/rpc/error.rs
alamb: yeah, I agree an improvement for another time

tests/end_to_end_cases/scenario.rs
crepererum: Good catch: https://github.com/influxdata/influxdb_iox/pull/2298

PR #2283, opened 2021-08-13
feat: set header on gRPC requests

tustvold:
This PR is mostly plumbing, but adds the ability to specify headers from the CLI to include in gRPC requests made to IOx. It does this by adding a tower layer that injects headers into the request path.

I'm not entirely sure how best to test this functionality, I could add an echo endpoint or something, but this feels a bit meh... FWIW the distributed tracing tests will exercise this functionality, and so perhaps that will be sufficient.

NB: This will not compile if you have code from pre-0.5.2 tonic-build, you may need to do a clean build. It relies on the relaxation of the types that can be fed to a gRPC service definition.


influxdb_iox_client/src/connection.rs
```diff
@@ -73,18 +77,17 @@ impl Builder {
     {
         let endpoint = Endpoint::from(dst.try_into()?)
             .user_agent(self.user_agent)?
+            .connect_timeout(self.connect_timeout)
             .timeout(self.timeout);
 
-        // Manually construct connector to workaround https://github.com/hyperium/tonic/issues/498
-        let mut connector = hyper::client::HttpConnector::new();
```
alamb, line 79: Is it intentional that this code no longer sets the connect timeout to `self.connect_timeout`?

influxdb_iox_client/src/connection.rs
```diff
@@ -73,18 +77,17 @@ impl Builder {
...
-        connector.set_connect_timeout(Some(self.connect_timeout));
+        let channel = endpoint.connect().await?;
 
-        // Defaults from from tonic::channel::Endpoint
```
alamb, line 82: The theory is that these defaults are set somewhere in tower now?

influxdb_iox_client/src/connection.rs
```diff
@@ -95,6 +98,13 @@ impl Builder {
         }
     }
 
+    /// Sets a header to be included on all requests
```
alamb, line 101:
```suggestion
    /// Adds an additional header to be included on all requests
```

influxdb_iox_client/src/tower.rs
```diff
@@ -0,0 +1,60 @@
...
+    type Future = S::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut request: Request<ReqBody>) -> Self::Future {
+        let headers = request.headers_mut();
+        for (name, value) in self.headers.iter() {
+            headers.entry(name).or_insert_with(|| value.clone());
```
alamb, line 56: My reading of the docs suggests this code will not overwrite existing headers. I think this might be surprising if someone sets a HTTP header override and then it is ignored (silently).

I suggest either erroring if overriding an existing header with a different value or perhaps better to actually override the header with the specified value

src/main.rs
```diff
@@ -322,3 +334,35 @@ unsafe fn set_signal_handler(signal: libc::c_int, handler: unsafe extern "C" fn(
...
+    K: FromStr,
+    V: FromStr,
+    K::Err: std::fmt::Display,
+    V::Err: std::fmt::Display,
+{
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        use itertools::Itertools;
+        match s.split(':').collect_tuple() {
```
alamb, line 356: TIL `collect_tuple`

alamb (approved):
Approving this PR but I think the comments about connection timeout seemingly being lost as well as the behavior when overriding existing headers should be addressed before merge

> I'm not entirely sure how best to test this functionality, I could add an echo endpoint or something, but this feels a bit meh... FWIW the distributed tracing tests will exercise this functionality, and so perhaps that will be sufficient.

I agree that will be sufficient

influxdb_iox_client/src/connection.rs
tustvold: Yes sorry, forgot to highlight this - https://github.com/hyperium/tonic/issues/498 has been fixed

influxdb_iox_client/src/connection.rs
tustvold: These were only ever here as a temporary hack around https://github.com/hyperium/tonic/issues/498 which has since been fixed - I have confirmed the defaults in tonic are still the same

influxdb_iox_client/src/connection.rs
tustvold: I've changed it to always set the header

influxdb_iox_client/src/tower.rs
tustvold: I've changed it to always set the header

influxdb_iox_client/src/connection.rs
```diff
@@ -73,18 +77,17 @@ impl Builder {
     {
         let endpoint = Endpoint::from(dst.try_into()?)
             .user_agent(self.user_agent)?
+            .connect_timeout(self.connect_timeout)
```
tustvold, line 80: This didn't used to be available

PR #2284, opened 2021-08-13
refactor: include all allocated buffers in Read Buffer memory usage

e-dard:
Towards #2257

This PR contains two things and follows on from #2267. Firstly it ensures that the capacity of all allocated buffers in the Read Buffer is taken into account and exposed via the crate's `size()` method. This API is used by things like the catalog and system tables to describe how much on-heap memory the Read Buffer chunks are taking up.

The second aspect to this PR is to add a new metric, which is primarily of use to devlopment of the Read Buffer. This metric `read_buffer_column_required_bytes` describes the minimal required bytes to store encoded data in according to the encoding/datatype etc. It does not account for any pre-allocated buffers and it does not include data structures that exist for performance optimisation such as roaring bitmaps. Effectively it's a way to track compression characteristics.

So to summarise - as far as the rest of IOx is concerned the Read Buffer will now include all pre-allocated buffers when something asks the Read Buffer how big a column/table/chunk is.


read_buffer/src/column.rs
```diff
@@ -51,8 +51,17 @@ impl Column {
     //  Meta information about the column
     //
 
-    /// The estimated size in bytes of the column that is held in memory.
+    /// The estimated size in bytes of all data in the column, including all
+    /// allocated but unfilled buffers. An estimation of the on-heap memory usage
+    // of this column.
     pub fn size(&self) -> usize {
+        self.size_bytes(true)
```
e-dard, line 58: This ensures that anything (catalog/other crates etc) that wants to know the in-memory size of a chunk gets a value that includes everything allocated on the heap.

read_buffer/src/column/encoding/bool.rs
```diff
@@ -43,12 +43,8 @@ impl Bool {
...
     /// Returns an estimation of the total size in bytes used by this column
     /// encoding.
-    pub fn size(&self, buffers: bool) -> usize {
-        size_of::<Self>()
-            + match buffers {
-                true => self.arr.get_array_memory_size(), // includes buffer capacities
-                false => self.arr.get_buffer_memory_size(),
-            }
+    pub fn size(&self) -> usize {
+        size_of::<Self>() + self.arr.get_array_memory_size()
```
e-dard, line 47: As suggested in https://github.com/influxdata/influxdb_iox/pull/2267#discussion_r687655417

NGA-TRAN:
I am starting reviewing this

read_buffer/src/column.rs
```diff
@@ -1382,15 +1391,16 @@ impl Iterator for RowIDsIterator<'_> {
...
 pub(crate) struct Statistics {
     pub enc_type: Cow<'static, str>, // The encoding type
     pub log_data_type: &'static str, // The logical data-type
     pub values: u32,                 // Number of values present (NULL and non-NULL)
     pub nulls: u32,                  // Number of NULL values present
-    pub bytes: usize,                // Total size of data
-    pub raw_bytes: usize,            // Estimated "uncompressed" size
-    pub raw_bytes_no_null: usize,    // Estimated "uncompressed" size ignoring NULL values
+    pub allocated_bytes: usize, // Total size of all data including allocated but unused buffers
+    pub required_bytes: usize, // Total minimum required bytes needed for the column (excluding capacity bytes)
```
NGA-TRAN, line 1401: I am not sure what `(excluding capacity bytes)` mean and whether it is useful in the comment here. The `required_bytes` seems self-explained already

read_buffer/src/column/encoding/string/rle.rs
```diff
@@ -1007,11 +1016,10 @@ mod test {
 
         // * Self: 24 + 24 + 24 + 1 + (padding 3b) + 4 = 80b
         // * index entries: (4) are is (24*4) + 14 == 110
-        // * index row ids: (bitmaps) is (4 * 4) + (108b for bitmaps) == 124
+        // * index row ids: (bitmaps) not included in calc
```
NGA-TRAN, line 1019: Maybe a silly question. Why this does not include in calculation? Is it because it was considered in row_ids_bitmap_size?

NGA-TRAN (approved):
Nice to see we capture allocated memory. I can't wait to see your sophisticated metrics to compare used and allocated :)

read_buffer/src/column/encoding/string/rle.rs
e-dard: It's because bitmaps are not strictly necessary to do the encoding, rather they're a way of making the query evaluation faster. So I like to separate them out as not "required" but certainly taking up size (reflected in the `size()` API.

PR #2285, opened 2021-08-13
feat: add end-to-end tracing test

tustvold:
I'm not sure if this is technically an end-to-end test or an integration test of the server binary, but it tests that the trace instrumentation is working correctly.

Most of this PR is plumbing as it needs to support a type of test we don't currently have - we currently have integration tests of the server crate, and end-to-end tests of the binary, but no integration tests of the service handlers themselves. In order to test tracing, I need to inject a non-standard trace collector into the application, which necessitates an integration test. I suspect this sort of high-level integration test will be useful for more than just this - particularly w.r.t getting the server into unusual states.


src/influxdb_ioxd.rs
```diff
@@ -160,7 +153,39 @@ pub async fn main(config: Config) -> Result<()> {
...
+        application,
+        grpc_listener,
+        http_listener,
+        trace_collector,
+        app_server,
+    )
+    .await
+}
+
+async fn grpc_listener(addr: SocketAddr) -> Result<tokio::net::TcpListener> {
```
tustvold, line 171: The tests below bind port 0, splitting out the result of the bind operation allows the test to determine the concrete port that was bound and connect to that

src/influxdb_ioxd.rs
```diff
@@ -170,42 +195,39 @@ pub async fn main(config: Config) -> Result<()> {
...
         frontend_shutdown.clone(),
         config.initial_serving_state.into(),
     )
     .fuse();
 
-    info!(bind_address=?grpc_bind_addr, "gRPC server listening");
-
-    let bind_addr = config.http_bind_address;
-    let addr = AddrIncoming::bind(&bind_addr).context(StartListeningHttp { bind_addr })?;
+    info!("gRPC server listening");
```
tustvold, line 218: The listener creation now logs the address

src/influxdb_ioxd.rs
```diff
@@ -160,7 +153,39 @@ pub async fn main(config: Config) -> Result<()> {
...
+    )
+    .await
+}
+
+async fn grpc_listener(addr: SocketAddr) -> Result<tokio::net::TcpListener> {
+    let listener = tokio::net::TcpListener::bind(addr)
+        .await
+        .context(StartListeningGrpc { addr })?;
+
+    match listener.local_addr() {
```
tustvold, line 176: I'm not entirely sure how this can fail, perhaps unix domain sockets or something?? Either way we fallback if it fails

src/influxdb_ioxd.rs
```diff
@@ -160,7 +153,39 @@ pub async fn main(config: Config) -> Result<()> {
     let application = make_application(&config).await?;
     let app_server = make_server(Arc::clone(&application), &config);
 
-    serve(config, application, app_server).await
+    let grpc_listener = grpc_listener(config.grpc_bind_address).await?;
```
tustvold, line 156: This is where the new code is plumbed into influxdb_ioxd (as opposed to the rest which is test plumbing)

src/influxdb_ioxd.rs
```diff
@@ -476,4 +511,87 @@ mod tests {
             tracker::TaskResult::Success
         )
     }
+
+    #[tokio::test]
+    async fn test_tracing() {
```
tustvold, line 516: These tests are currently fairly verbose to write, I suspect as we end up with more of these style tests we may want to extract some sort of TestFixture like we have in other modules

src/influxdb_ioxd.rs
```diff
@@ -476,4 +511,87 @@ mod tests {
...
+
+        let mut client = influxdb_iox_client::management::Client::new(client);
+
+        client.list_databases().await.unwrap();
+
+        assert_eq!(trace_collector.spans().len(), 0);
+
+        let tracing_client = influxdb_iox_client::connection::Builder::default()
+            .header(
+                HeaderName::from_static("x-b3-sampled"),
```
tustvold, line 553: This method panics if you give it a string containing upper case characters (as internally everything is stored lower-case)

src/influxdb_ioxd.rs
```diff
@@ -476,4 +511,87 @@ mod tests {
...
+        assert_eq!(spans[1].name, "IOx");
+        assert_eq!(spans[1].ctx.parent_span_id.unwrap().0.get(), 111111);
+        assert_eq!(spans[1].ctx.trace_id.0.get(), 999999999);
+        assert!(spans[1].start.is_some());
+        assert!(spans[1].end.is_some());
+
+        assert_ne!(spans[0].ctx.span_id, spans[1].ctx.span_id);
+        server.shutdown();
+        join.await.unwrap().unwrap();
+    }
```
NGA-TRAN, line 596: The test looks good

NGA-TRAN (approved):

PR #2293, opened 2021-08-16
feat: support jaeger context propagation format

tustvold:
It turns out that use of B3 propagation headers is currently _aspirational_, and currently our environments use the jaeger trace propagation format. This PR therefore adds support for decoding them

Relates to #2273


src/main.rs
```diff
@@ -354,7 +354,7 @@ where
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         use itertools::Itertools;
-        match s.split(':').collect_tuple() {
+        match s.splitn(2, ':').collect_tuple() {
```
tustvold, line 357: This allows writing `cargo run -- server set 232 --header uber-trace-id:3459495:30434:0:1`

trace/src/ctx.rs
```diff
@@ -43,10 +46,20 @@ pub enum DecodeError {
...
     ValueDecodeError { source: ParseIntError },
 
+    #[snafu(display("Expected \"trace-id:span-id:parent-span-id:flags\""))]
+    InvalidJaegerTrace,
+
     #[snafu(display("value cannot be 0"))]
     ZeroError,
 }
 
+impl From<ParseIntError> for DecodeError {
```
tustvold, line 56: I'm not sure if this is something Snafu would consider a feature request for? @shepmaster ?

trace/src/ctx.rs
```diff
@@ -229,10 +319,9 @@ mod tests {
             SpanContext::from_headers(&collector, &headers)
                 .unwrap_err()
                 .to_string(),
-            "header 'X-B3-TraceId' not found"
+            "header 'X-B3-SpanId' not found"
```
tustvold, line 322: The trace id header is now used to determine the codec, so this test had to be slightly re-arranged

trace/src/ctx.rs
```diff
@@ -130,29 +135,113 @@ impl SpanContext {
...
+
+struct JaegerCtx {
+    trace_id: TraceId,
+    span_id: SpanId,
+    parent_span_id: Option<SpanId>,
+    flags: u8,
+}
+
+impl FromStr for JaegerCtx {
+    type Err = DecodeError;
```
tustvold, line 203: Really this should have its own error enumeration, to allow adding context to which of parent_span_id, flags, etc... the error was encountered - I can add this if people feel strongly, but it seemed a tad overkill

src/main.rs
e-dard: What does that above do? If it's something operational can we document it?

e-dard (approved):
LGTM 👍

src/main.rs
tustvold: `--header` is just a mechanism for setting headers on gRPC requests, it was added in https://github.com/influxdata/influxdb_iox/pull/2283

alamb: > What does that above do? If it's something operational can we document it?

I think the point of @tustvold 's comment was the change from `split` to `splitn` means that the CLI will now properly handle header values with `:` in it

src/influxdb_ioxd.rs
```diff
@@ -565,13 +565,27 @@ mod tests {
...
+        let mut b3_tracing_client = influxdb_iox_client::management::Client::new(b3_tracing_client);
 
-        tracing_client.list_databases().await.unwrap();
-        tracing_client.get_server_status().await.unwrap();
+        b3_tracing_client.list_databases().await.unwrap();
+        b3_tracing_client.get_server_status().await.unwrap();
+
+        let jaeger_tracing_client = influxdb_iox_client::connection::Builder::default()
+            .header(
+                HeaderName::from_static("uber-trace-id"),
```
alamb, line 575: the jaeger tracing client uses `uber-trace-id` as the header name. Man -- definitely sounds like there is some need for standardization lol

trace/src/ctx.rs
alamb: I don't understand the comment about "no context and a custom message" -- isn't that exactly what `ZeroError` [link](https://github.com/influxdata/influxdb_iox/pull/2293/files#diff-c0eb3f0dae31d88b6f712346c617b1d04d52f5be286c4397999ac6f10201b5c3R52) a few lines up does?

```rust
    #[snafu(display("value cannot be 0"))]
    ZeroError,
```

trace/src/ctx.rs
alamb: I think this is fine -- and we can revisit if we ever need more detail / context in the errors

trace/src/ctx.rs
tustvold: Snafu has the ability to automatically define the From conversion, but doesn't then let you specify a custom message - https://docs.rs/snafu/0.6.10/snafu/guide/attributes/index.html#controlling-context. With say thiserror it is possible to write something like and it will automatically define the from conversion.

```
enum Error {
   #[error("value decode error: {}", source)]
   ValueDecodeError{ #[from] source: ParseIntError
}
```

If one is following a pattern of defining small-ish error enumerations, as is imo a good idea, you often end up in a situation where there isn't really meaningful error context to provide

src/influxdb_ioxd.rs
tustvold: Good thing there are 3 standards :laughing: 

B3 is definitely the one that seems to have the greatest adoption though, but it's one of those things where there isn't enough pressure to force standardisation as instrumentation is such an ad-hoc thing anyway. Each organization does there own thing, and it doesn't really matter

PR #2295, opened 2021-08-16
fix: disallow dropping unpersted chunks from persisted DB

### LINKED ISSUE #2291: disallow dropping unpersisted chunks when the database is persisting data
# Abstract
It is very challenging engineering wise to recover to a “consistent” state (during replay) if arbitrary chunks can be dropped and the complexity is not worth the feature at this time.

# Technical Background
We currently allow dropping individual chunks no matter if persistence is enabled or no:

- When **persistence is disabled**, this is not an issue at all.
- When **persistence is enabled**,
  - **dropping already persisted chunks** also works fine since we just need to store a "this parquet file is not of the catalog any longer" in the transaction log. Also while the chunk in question was originally persisted we also recorded the fact that the write buffer sequence number(s) were consumed (`PartitionCheckpoint`).
  - **dropping an unpersisted chunks** does not require to "forget" a parquet file, but at the same time we currently don't have a way to record the write buffer progress for that chunk, since `PartitionCheckpoint`s are currently only preserved when parquet files are written.

While solving the illustrated issue is not impossible, it creates additional engineering effort. At the moment this effort does not seem to pay off.

# Proposed Changes
If the user tries to delete a chunk when persist is on, the request will be rejected w/ an error.

crepererum:
It doesn't play well w/ replay at the moment since we would forget which
sequence numbers we've already seen.

Fixes #2291.


server/src/db/lifecycle/drop.rs
```diff
@@ -34,6 +34,16 @@ pub fn drop_chunk(
         chunk: guard.addr().clone(),
     });
 
+    // check if we're dropping an unpersisted chunk in a persisted DB
```
alamb, line 37: 💯  for the references to the rationale. This is the kind of code that is easy to see *what* it does but not the *why* 🏅

alamb: I thought the plan was to disallow dropping chunks at all if `lifecycle_rules.persist` though if it is easy to support drops once persisted that is also fine.

alamb (approved):

PR #2296, opened 2021-08-16
feat: add temporary logging of request header keys

tustvold:
Adds a temporary log line to hopefully gain some insight into what tracing payloads are reaching IOx


trace/src/tower.rs
```diff
@@ -69,6 +69,9 @@ where
     }
 
     fn call(&mut self, mut request: Request<ReqBody>) -> Self::Future {
+        use itertools::Itertools;
+        // TEMPORARY
```
crepererum, line 73: you might wanna create a "remove it"-ticket and link it here, but I don't wanna be too picky.

crepererum (approved):

trace/src/tower.rs
tustvold: https://github.com/influxdata/influxdb_iox/issues/2297

PR #2308, opened 2021-08-16
feat: log value of sample header

tustvold:
The logging added in #2296 shows that IOx is receiving trace context, but isn't doing anything with it, as it isn't logging an error this suggests it isn't understanding its sampling decision. This PR adds some more logging to help diagnose this, and removes the logging that has now served its purpose.


crepererum (approved):

trace/src/ctx.rs
```diff
@@ -176,14 +177,28 @@ fn decode_b3(
...
+        trace_id: required_header(headers, B3_TRACE_ID_HEADER)?,
+        parent_span_id: parsed_header(headers, B3_PARENT_SPAN_ID_HEADER)?,
+        span_id: required_header(headers, B3_SPAN_ID_HEADER)?,
+        collector: Some(Arc::clone(collector)),
+    };
+
+    // TEMPORARY (#2297)
+    info!(?ctx, "tracing request");
+
     Ok(Some(SpanContext {
```
alamb, line 202: maybe replace here with `Ok(Some(ctx))` ?

alamb (approved):

trace/src/tower.rs
```diff
@@ -69,9 +69,6 @@ where
     }
 
     fn call(&mut self, mut request: Request<ReqBody>) -> Self::Future {
-        use itertools::Itertools;
-        // TEMPORARY
-        info!(headers=%request.headers().keys().join(","), "gRPC request with headers");
```
alamb, line 74: maybe you should also just log values here temporarily to get all possible information you might need

trace/src/ctx.rs
tustvold: Oops, that's what I meant to do :sweat_smile:

PR #2310, opened 2021-08-16
feat: drop entire partitions

### LINKED ISSUE #2292: Add a drop partition call
# Rationale
We do need *some* way to get data out of IOx (e.g. as it ages out, we want to clean up S3, etc) but want this mechanism to avoid causing necessary complication in Replay (see #2291).

# Proposed Changes
Add a new “delete partition”  command that will drop all data in a partition. Initially, this may generate error if the partition is not completely persisted and persistence is enabled (again see #2291 for the rational).

crepererum:
Closes #2292

### DIFF
--- a/generated_types/src/job.rs
+++ b/generated_types/src/job.rs
@@ -53,6 +53,11 @@ impl From<Job> for management::operation_metadata::Job {
                 table_name: chunk.table_name.to_string(),
                 chunk_id: chunk.chunk_id,
             }),
+            Job::DropPartition { partition } => Self::DropPartition(management::DropPartition {
+                db_name: partition.db_name.to_string(),
+                partition_key: partition.partition_key.to_string(),
+                table_name: partition.table_name.to_string(),
+            }),
         }
     }
 }
@@ -133,6 +138,17 @@ impl From<management::operation_metadata::Job> for Job {
                     table_name: Arc::from(table_name.as_str()),
                     partition_key: Arc::from(partition_key.as_str()),
                     chunk_id,
+                },
+            },
+            Job::DropPartition(management::DropPartition {
+                db_name,
+                partition_key,
+                table_name,
+            }) => Self::DropPartition {
+                partition: PartitionAddr {
+                    db_name: Arc::from(db_name.as_str()),
+                    table_name: Arc::from(table_name.as_str()),
+                    partition_key: Arc::from(partition_key.as_str()),
                 },
             },
         }
--- a/influxdb_iox_client/src/client/management.rs
+++ b/influxdb_iox_client/src/client/management.rs
@@ -328,7 +328,7 @@ pub enum DropPartitionError {
     #[error("Server unavailable: {}", .0.message())]
     Unavailable(tonic::Status),
 
-    /// Server indicated that it is not (yet) available
+    /// Server indicated some other action was active for this partition
     #[error("Cannot perform operation due to wrong chunk lifecycle state: {}", .0.message())]
     LifecycleError(tonic::Status),
 
--- a/influxdb_iox_client/src/client/management.rs
+++ b/influxdb_iox_client/src/client/management.rs
@@ -317,6 +317,26 @@ pub enum SkipReplayError {
     ServerError(tonic::Status),
 }
 
+/// Errors returned by [`Client::drop_partition`]
+#[derive(Debug, Error)]
+pub enum DropPartitionError {
+    /// Database not found
+    #[error("Not found: {}", .0)]
+    NotFound(String),
+
+    /// Server indicated that it is not (yet) available
+    #[error("Server unavailable: {}", .0.message())]
+    Unavailable(tonic::Status),
+
+    /// Server indicated that it is not (yet) available
+    #[error("Cannot perform operation due to wrong chunk lifecycle state: {}", .0.message())]
+    LifecycleError(tonic::Status),
+
+    /// Client received an unexpected error from the server
+    #[error("Unexpected server error: {}: {}", .0.code(), .0.message())]
+    ServerError(tonic::Status),
+}
+
 /// An IOx Management API client.
 ///
 /// This client wraps the underlying `tonic` generated client with a
@@ -774,6 +794,34 @@ impl Client {
                 tonic::Code::FailedPrecondition => SkipReplayError::FailedPrecondition(status),
                 tonic::Code::InvalidArgument => SkipReplayError::InvalidArgument(status),
                 _ => SkipReplayError::ServerError(status),
+            })?;
+
+        Ok(())
+    }
+
+    /// Drop partition from memory and (if persisted) from object store.
+    pub async fn drop_partition(
+        &mut self,
+        db_name: impl Into<String> + Send,
+        table_name: impl Into<String> + Send,
+        partition_key: impl Into<String> + Send,
+    ) -> Result<(), DropPartitionError> {
+        let db_name = db_name.into();
+        let partition_key = partition_key.into();
+        let table_name = table_name.into();
+
+        self.inner
+            .drop_partition(DropPartitionRequest {
+                db_name,
+                partition_key,
+                table_name,
+            })
+            .await
+            .map_err(|status| match status.code() {
+                tonic::Code::NotFound => DropPartitionError::NotFound(status.message().to_string()),
+                tonic::Code::Unavailable => DropPartitionError::Unavailable(status),
+                tonic::Code::FailedPrecondition => DropPartitionError::LifecycleError(status),
+                _ => DropPartitionError::ServerError(status),
             })?;
 
         Ok(())
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -609,6 +609,25 @@ impl Db {
         fut.await.context(TaskCancelled)?.context(LifecycleError)
     }
 
+    /// Drops the specified partition from the catalog and all storage systems
+    pub async fn drop_partition(
+        self: &Arc<Self>,
+        table_name: &str,
+        partition_key: &str,
+    ) -> Result<()> {
+        // Use explicit scope to ensure the async generator doesn't
+        // assume the locks have to possibly live across the `await`
+        let fut = {
+            let partition = self.partition(table_name, partition_key)?;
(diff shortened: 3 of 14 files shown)


generated_types/src/job.rs
```diff
@@ -135,6 +140,17 @@ impl From<management::operation_metadata::Job> for Job {
                     chunk_id,
                 },
             },
+            Job::DropPartition(management::DropPartition {
+                db_name,
+                partition_key,
+                table_name,
+            }) => Self::DropPartition {
+                partition: PartitionAddr {
+                    db_name: Arc::from(db_name.as_str()),
```
alamb, line 149: It does not matter at all, but I think you can also make `Arc<str>`s like this:

```suggestion
                    db_name: db_name.as_str().into(),
```

influxdb_iox_client/src/client/management.rs
```diff
@@ -317,6 +317,26 @@ pub enum SkipReplayError {
...
+pub enum DropPartitionError {
+    /// Database not found
+    #[error("Not found: {}", .0)]
+    NotFound(String),
+
+    /// Server indicated that it is not (yet) available
+    #[error("Server unavailable: {}", .0.message())]
+    Unavailable(tonic::Status),
+
+    /// Server indicated that it is not (yet) available
```
alamb, line 331:
```suggestion
    /// Server indicated some other action was active for this partition
```

server/src/db.rs
```diff
@@ -4251,6 +4270,47 @@ mod tests {
...
+        // See https://github.com/influxdata/influxdb_iox/issues/2291
+        let err = db.drop_partition("cpu", partition_key).await.unwrap_err();
+        assert!(matches!(
+            err,
+            Error::LifecycleError {
+                source: super::lifecycle::Error::CannotDropUnpersistedChunk { .. }
+            }
+        ));
+
+        // once persisted drop should work
```
alamb, line 4300: 👍  good test

server/src/db/lifecycle/drop.rs
```diff
@@ -80,3 +80,110 @@ pub fn drop_chunk(
...
+
+    // get locks for all chunks
+    let lockable_chunks: Vec<_> = partition
+        .chunks()
+        .map(|chunk| LockableCatalogChunk {
+            db: Arc::clone(&db),
+            chunk: Arc::clone(chunk),
+        })
+        .collect();
+    let mut guards: Vec<_> = lockable_chunks
```
alamb, line 107: I wonder if we need to ensure these lock guards are acquired in the same order to avoid deadlocks with concurrent `drop_partition` calls? 

Perhaps sorting by `chunk_id` prior to acquiring the guards would be sufficient?

server/src/db/lifecycle/drop.rs
```diff
@@ -80,3 +80,110 @@ pub fn drop_chunk(
...
+
+        // collect parquet files that we need to remove
+        let mut paths = vec![];
+        for chunk in &chunks {
+            let chunk_read = chunk.read();
+
+            if let ChunkStage::Persisted { parquet, .. } = chunk_read.stage() {
+                let path: DirsAndFileName = parquet.path().into();
+                paths.push(path);
+            }
```
alamb, line 161: I would suggest something like

```suggestion
            }
            else if db.rules().lifecycle_rules.persist {
              unreachable!()
            }
```

to both:
1. Make it clear to future readers that the assumption of stage is valid as it was checked above
2. If the check above is ever changed, will assert if they don't also update this code (which relies on the assumption)

server/src/db/lifecycle/drop.rs
```diff
@@ -80,3 +80,110 @@ pub fn drop_chunk(
...
+        // only create catalog transaction when there's anything to do
+        if !paths.is_empty() {
+            let mut transaction = preserved_catalog.open_transaction().await;
+            for path in paths {
+                transaction.remove_parquet(&path);
+            }
+            transaction.commit().await.context(CommitError)?;
+        }
+
+        let mut partition = partition.write();
```
alamb, line 173: I am probably missing something here, but it looks like this code iterated on `for chunk in &chunks` twice? We shouldn't be dropping the chunks out of the partition more than once -- I think this code may end up trying to drop the chunk `N^2` times.

The tests wouldn't hit this as it only tries to drop a single chunk

server/src/db/replay.rs
```diff
@@ -505,6 +505,13 @@ mod tests {
         /// The partitions are by table name and partition key.
         Persist(Vec<(&'static str, &'static str)>),
 
+        /// Drop partitions.
+        ///
+        /// Note that this only works for fully persisted partitions.
```
alamb, line 510:
```suggestion
        /// Note that this only works for fully persisted partitions if
        /// the database is configured for persistence.
```

server/src/db/replay.rs
```diff
@@ -2337,6 +2365,108 @@ mod tests {
         .await
     }
 
+    #[tokio::test]
+    async fn replay_after_drop() {
```
alamb, line 2369: 🥇

server/src/db/replay.rs
```diff
@@ -2337,6 +2365,108 @@ mod tests {
...
+                            "| 20  | b                | 1970-01-01T00:00:00.000000020Z |",
+                            "+-----+------------------+--------------------------------+",
+                        ],
+                    ),
+                ]),
+                Step::MakeWritesPersistable,
+                Step::Persist(vec![("table_1", "tag_partition_by_b")]),
+                Step::Drop(vec![("table_1", "tag_partition_by_b")]),
+                Step::Assert(vec![
+                    // note that at the moment the dropped partition still exists but has no chunks
```
alamb, line 2405: 🤔  I wonder if there is some reason not to drop the entire partition itself from the catalog?

server/src/db/replay.rs
```diff
@@ -2337,6 +2365,108 @@ mod tests {
...
+                        ],
+                    ),
+                ]),
+                Step::Restart,
+                Step::Assert(vec![
+                    // partition also exists after restart but still has no data
+                    Check::Partitions(vec![("table_1", "tag_partition_by_b")]),
+                    // Note that currently this edge case also leads to "no columns" because the query executor does not
+                    // return any record batches.
+                    Check::Query("select * from table_1 order by bar", vec!["++", "++"]),
```
alamb, line 2427: what is the reason to ingest the new data prior to calling replay? I don't see anything wrong with doing this it just confused me

server/src/db/replay.rs
```diff
@@ -2337,6 +2365,108 @@ mod tests {
...
+                        vec![
+                            "+-----+------------------+--------------------------------+",
+                            "| bar | tag_partition_by | time                           |",
+                            "+-----+------------------+--------------------------------+",
+                            "| 10  | a                | 1970-01-01T00:00:00.000000010Z |",
+                            "+-----+------------------+--------------------------------+",
+                        ],
+                    ),
+                ]),
+                // new data can still be ingested into the dropped partition
```
alamb, line 2451: 👍

alamb (dismissed):
Overall this looks really good @crepererum  👍  

I am a little confused about the double chunk loop in `drop.rs` so I don't feel comfortable clicking approve just yet. But otherwise this PR looks very solid

generated_types/src/job.rs
crepererum: I think that `.into` also works, but it's a pattern used within the same file. I can do a cleanup in a follow-up PR.

server/src/db/lifecycle/drop.rs
crepererum: I've added some docs explaining why we get away with this code.

server/src/db/lifecycle/drop.rs
crepererum: Done (same for `drop_chunk`).

server/src/db/lifecycle/drop.rs
crepererum: I've tried to add some more docs, but I think you might be misreading the code here. The loops are NOT nested (hence there is no `N^2`). The two loops are (now explained w/ some docs):

1. first loop: collect parquet files
2. if there are any files: commit transaction (that might fail due to IO)
3. second loop: perform in-memory change (aka: drop chunks)

Note that while we're doing this, nothing else can perform any changes on the set of chunks that we've collected (there might be new chunks but that's not an issue), since while holding the chunk locks we set a special lifecycle action named "dropping".

server/src/db/replay.rs
crepererum: I normally do that in all replay tests just to make sure that we don't "overshoot" while replaying, aka that we don't read any data that we're not supposed to read yet.

server/src/db/replay.rs
crepererum: Created #2315.

crepererum:
Note that dropping data has under some circumstances the edge case that it can re-appear: #2318. I nevertheless will merge this PR because fixing it requires work on a very different part of the stack and also doesn't have highest priority.

server/src/db/lifecycle/drop.rs
alamb: I read the comments -- thanks for that. I agree that holding the partition level write lock prevents deadlocks with two concurrent invocations of `drop_partition`

The more I think about it, I am not sure that holding a partition write lock is sufficient to avoid deadlocks when locking chunks because all the partition lock allows is getting `LockableChunks` -- the actual locking of the chunks can happen once the partition is released. 

I reviewed the code for Partition::chunks -- and it indeed returns chunks in sorted order. That was not obvious to me. I will add some comments perhaps

alamb (approved):
Thank you

server/src/db/lifecycle/drop.rs
crepererum: > The more I think about it, I am not sure that holding a partition write lock is sufficient to avoid deadlocks when locking chunks because all the partition lock allows is getting LockableChunks -- the actual locking of the chunks can happen once the partition is released.

That's correct, we currently don't ensure that technically / type-wise.

alamb: https://github.com/influxdata/influxdb_iox/pull/2323 for some additional docs + tests

PR #2316, opened 2021-08-17
chore: fix dockerignore (#2300)

### LINKED ISSUE #2300: Incorrect perf_image build context
The perf image docker build context is a lot larger than it should probably be, and this significantly slows down the image build. This is likely due to an incorrect/non-existent .dockerignore file resulting in it copying both the code and a load of temporary build artifacts from /target

image

tustvold:
Closes #2300


.circleci/config.yml
```diff
@@ -286,6 +286,7 @@ jobs:
           # Disabling for now, and tracked further investigations
           # in https://github.com/influxdata/k8s-idpe/issues/3038
           docker_layer_caching: false
+          version: 19.03.14
```
tustvold, line 289: https://github.com/moby/buildkit/pull/901 which contains the logic to understand dockerignore overrides was included in [19.03 release](https://docs.docker.com/engine/release-notes/19.03/#19030) of docker. Previously we were using 17.09.0-ce which is the default and so the dockerignore was not being used.

mkmik (approved):

.circleci/config.yml
jacobmarble: Interesting, thanks for this note.

PR #2319, opened 2021-08-17
feat: support sinking traces to an OTEL SpanExporter

tustvold:
This PR adds the ability to sink traces to any OpenTelemetry::SpanExporter. A future PR will use this to hook up Jaeger export.

As part of this PR the lifetimes were removed from Span and friends, as there isn't a "safe" way to convert a &'a to Cow<'static, str> without always allocating a new string even when 'a: 'static. As the common case is for static strings, it makes sense to optimise for this case


trace/src/span.rs
```diff
@@ -65,93 +67,101 @@ impl<'a> Span<'a> {
             }
         }
     }
-}
 
-impl<'a> Drop for Span<'a> {
-    fn drop(&mut self) {
-        if let Some(collector) = &self.ctx.collector {
+    /// Exports this `Span` to its registered collector if any
+    pub fn export(mut self) {
```
tustvold, line 72: This is a change that results from the removal of borrows, we want to be able to move the potentially owned string data out of the span, and so TraceCollector::export no longer takes a reference. 

The implication is that a Span can't publish itself on Drop (partial moves are not supported on types implementing Drop). Fortunately this was a bit of a strange behavior anyway, and so this publication logic is just moved into EnteredSpan and an explicit export method added to Span

trace/src/span.rs
```diff
@@ -65,93 +67,101 @@ impl<'a> Span<'a> {
...
         Self::Int(v)
     }
 }
 
 /// Updates the start and end times on the provided Span
 #[derive(Debug)]
-pub struct EnteredSpan<'a> {
-    span: Span<'a>,
+pub struct EnteredSpan {
+    /// Option so we can take out of it
```
alamb, line 122:
```suggestion
    /// Option so we can take out of it on drop / publish
```

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+        Self {
+            join,
+            shutdown,
+            sender,
+        }
+    }
+
+    /// Triggers shutdown of this `OtelExporter`
+    pub fn shutdown(&self) {
+        info!("jaeger exporter shutting down");
```
alamb, line 64:
```suggestion
        info!("otel exporter shutting down");
```

It is confusing to me to intermix the terms jaeger and otel. I don't really care which is chosen but I think they should be consistent

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+        self.shutdown.cancel()
+    }
+
+    /// Waits for the background worker of OtelExporter to finish
+    pub fn join(&self) -> impl Future<Output = Result<(), Arc<JoinError>>> {
```
alamb, line 69: I don't understand why this isn't `pub async fn join(&self) -> Result<(), Arc<JoinError>>` (aka why clone the future rather than waiting directly?) Is it so that `&mut self` is not required?

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+impl TraceCollector for OtelExporter {
+    fn export(&self, span: Span) {
+        use mpsc::error::TrySendError;
+
+        match self.sender.try_send(span.into()) {
+            Ok(_) => {
+                //TODO: Increment some metric
+            }
+            Err(TrySendError::Full(_)) => {
+                warn!("exporter cannot keep up, dropping spans")
```
alamb, line 83: will this warn! itself try and send a span too?

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+
+async fn background_worker<T: SpanExporter + 'static>(
+    shutdown: CancellationToken,
+    exporter: T,
+    receiver: mpsc::Receiver<SpanData>,
+) {
+    tokio::select! {
+        _ = exporter_loop(exporter, receiver) => {
+            // Don't expect this future to complete
+            error!("jaeger exporter loop completed")
```
alamb, line 100:
```suggestion
            error!("otel exporter loop completed")
```

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+        }
+        _ = shutdown.cancelled() => {}
+    }
+    info!("jaeger exporter shut down")
```
alamb, line 104:
```suggestion
    info!("otel exporter shut down")
```

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+/// An opentelemetry::SpanExporter that sinks writes to a tokio mpsc channel.
+///
+/// Intended for testing ONLY
+///
+/// Note: There is a similar construct in opentelemetry behind the testing feature
```
alamb, line 111: 👍

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+        // TODO: Batch export spans
+        if let Err(e) = exporter.export(vec![span]).await {
+            error!(%e, "error exporting span")
+        }
+    }
+
+    warn!("sender-side of jaeger exporter dropped without waiting for shut down")
+}
+
+impl From<Span> for SpanData {
```
alamb, line 149: I like that the transforming of the `IOx` trace data model to `otel` data model is isolated in a module 

I would love to have a similar split for metrics so there are `IOx` metrics and then explicit code that converts from IOx metric to/from prometheus / otel rather than having the otel/prometheus model bleed throughout the codebase

trace/src/otel.rs
```diff
@@ -0,0 +1,369 @@
...
+
+            ret.attributes.insert(KeyValue::new(key, value))
+        }
+        ret
+    }
+}
+
+impl<'a> From<&'a SpanContext> for observability_deps::opentelemetry::trace::SpanContext {
+    fn from(ctx: &'a SpanContext) -> Self {
+        Self::new(
```
alamb, line 192: Should this conversion also try to connect up `ctx.parent_span_id` as well?

alamb (approved):
looks good to me. I am not sure if @jacobmarble  wants to take a look too

trace/src/otel.rs
tustvold: Definitely, this thing use to be called JaegerExporter and I renamed it, I must have missed these :+1:

trace/src/otel.rs
tustvold: It is so that the returned future does not have a lifetime coupled to that of the borrow of self. The same pattern is used in Server and Database.

`pub async fn foo(&self)` effectively desugars to `pub fn foo(&self) -> impl Future<Output=()> + '_`

trace/src/otel.rs
tustvold: Nope - there is no integration between tokio::trace and the distributed tracing provided by this PR. There is perhaps a legitimate point that if the exporter can't keep up because the span throughput is too high, logging might be a little noisy, but I suspect this is more likely to trip because the exporter has stalled than because too many spans are being produced.

trace/src/otel.rs
tustvold: I may have plans in this space :grin:

trace/src/otel.rs
tustvold: This is a slight mismatch, OTEL sticks this on the Span itself for some reason. The conversion is above on line 156

trace/src/span.rs
```diff
@@ -22,9 +23,9 @@ pub enum SpanStatus {
 ///
 /// On Drop a `Span` is published to the registered collector
```
tustvold, line 24:
```suggestion

```

Issue #2317, opened 2021-08-17
Track row group distribution in Read Buffer


e-dard:
Recently we did some optimisation work to remove redunant sorts from compaction plans (where possible). This means that the output of `SortPreservingMerge` may be fed directly into the Read Buffer. Since the output can be multiple record batches that means that multiple row groups can be created in the Read Buffer chunk.

It would be good to track the distribution of row groups in Read Buffer chunks via metrics.

PR #2320, opened 2021-08-17
feat: add metric to track total row groups

e-dard:
Towards #2317

IOx is now utilising a feature of the Read Buffer that allows a chunk to contain multiple row groups. This needs to be observable via metrics. Whilst tracking the distribution of row groups in chunks would be ideal (outlined in #2317) I am starting with just a gauge to track overall number of row groups in the Read Buffer.

### DIFF
--- a/read_buffer/src/chunk.rs
+++ b/read_buffer/src/chunk.rs
@@ -301,8 +301,14 @@ impl std::fmt::Debug for Chunk {
     }
 }
 
+/// The collection of metrics exposed by the Read Buffer. Note: several of these
+/// be better represented as distributions, but the histogram story in IOx is not
+/// yet figured out.
 #[derive(Debug)]
 pub struct ChunkMetrics {
+    /// The total number of row groups in the chunk.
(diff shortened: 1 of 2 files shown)


read_buffer/src/chunk.rs
```diff
@@ -301,8 +301,14 @@ impl std::fmt::Debug for Chunk {
     }
 }
 
+/// The collection of metrics exposed by the Read Buffer. Note: several of these
+/// be better represented as distributions, but the histogram story in IOx is not
+/// yet figured out.
 #[derive(Debug)]
 pub struct ChunkMetrics {
+    /// The total number of row groups in the chunk.
+    row_groups_total: Gauge,
```
tustvold, line 310: If you aren't going to use tags, you could just store GaugeValue, but up to you. Providing the labels up-front, or lack thereof, saves some string encoding when recording. For this code path completely irrelevant but thought I'd point it out

tustvold (approved):

PR #2323, opened 2021-08-17
docs: Add some doc comments about sortedness in catalog Partition chunks

alamb:
# Rationale:
There is code that relies on the order of chunks to avoid a deadlock. See  https://github.com/influxdata/influxdb_iox/pull/2310/files#r690125197 for one such example.

The guarantee that `Partition::chunks()` are always returned sorted by chunk is important but not documented or tested.

While I admit this may seem like a quixotic exercise, I personally don't want to be tracking down deadlocks that magically appear if we change some of these assumptions unwittingly

# Changes
1. Add docs and tests


crepererum (approved):
Thanks!

PR #2324, opened 2021-08-17
refactor: remove drop_non_persisted

### LINKED ISSUE #2212: Remove Drop Unpersisted Lifecycle
There are two flags in the LifecyclePolicy, `drop_unpersisted` and `persist`.

If they are both false the system will just fill up to the hard buffer limit and stop working

If they are both true, persistence won't be particularly meaningful as it will just be irretrievably dropping random chunks of data

Following discussion, `drop_unpersisted` should be removed and instead inferred as the negation of `persist`

crepererum:
Closes #2212 .

### DIFF
--- a/data_types/src/database_rules.rs
+++ b/data_types/src/database_rules.rs
@@ -131,10 +131,6 @@ pub struct LifecycleRules {
     /// rejecting writes
     pub buffer_size_hard: Option<NonZeroUsize>,
 
-    /// Allow dropping data that has not been persisted to object storage
-    /// once the database size has exceeded the configured limits
-    pub drop_non_persisted: bool,
-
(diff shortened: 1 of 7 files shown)


data_types/src/database_rules.rs
```diff
@@ -131,10 +131,6 @@ pub struct LifecycleRules {
     /// rejecting writes
     pub buffer_size_hard: Option<NonZeroUsize>,
 
-    /// Allow dropping data that has not been persisted to object storage
```
alamb, line 134: I wonder if we need to remove these settings from the k8s-idpe repo as well. Will send link via DM

alamb (approved):

PR #2325, opened 2021-08-17
feat: fix log crate output

tustvold:
A number of crates we make use of, including rust-rdkafka use the log crate instead of tokio-tracing. Currently these logs are just silently dropped, this is not ideal.

This PR installs a shim to forward log crate output to tokio-tracing, where it feeds into the filter pipeline as normal.

It also removes tracing_subscriber from observability_deps as it is not a small dependency, and does not need to be included in every crate. This should help with build times.


trogging/src/lib.rs
```diff
@@ -457,6 +462,7 @@ where
     pub fn install_global(self) -> Result<TracingGuard> {
         let subscriber = self.build()?;
         tracing::subscriber::set_global_default(subscriber)?;
+        tracing_log::LogTracer::init()?;
```
tustvold, line 465: This is the actual change

src/influxdb_ioxd.rs
```diff
@@ -229,7 +229,8 @@ async fn serve(
     .fuse();
     info!("HTTP server listening");
 
-    info!("InfluxDB IOx server ready");
+    // Purposefully use log not tokio-tracing to ensure correctly hooked up
+    log::info!("InfluxDB IOx server ready");
```
tustvold, line 233: This is then used by an end-to-end test to ensure that this pipeline is correctly hooked up

mkmik (approved):
how is the log level for that log crate governed?

tustvold:
> how is the log level for that log crate governed?

By tokio-tracing, it just feeds into the same filter stack

trogging/src/lib.rs
alamb: TIL https://docs.rs/tracing-log/0.1.2/tracing_log/ 👍

alamb (approved):
As DataFusion also uses the `log` crate this PR will likely make DataFusion logs available to IOx as well. 👍  Thanks @tustvold

PR #2326, opened 2021-08-17
fix: include underlying error message in connection error

alamb:
# Rationale:
I made a mistake and got a very unhelpful error message. I would have preferred to be told more clearly what my blunder was so that I could fix it quickly

For some reason, the `Display` impl of `tonic::transport::Error does not include any source details... So I added it

# Change
1. Include underlying error information along with `tonic::transport::Error` details

# Before this change:
```
./target/debug/influxdb_iox -vv --host localhost:1234 sql
Error connecting to localhost:1234: Connection error: transport error
```
👎 

# After this change:

```
target/debug/influxdb_iox --host 'localhost:1234' sql
Error connecting to localhost:1234: Connection error: transport error (error trying to connect: invalid URL, scheme is missing)
```
👍


tustvold (approved):
I'm not sure about this one, the ecosystem in general appears to be moving away from including the source in the Display output, instead preferring to walk `source()` now that this is stabilised. I'm not sure where Snafu stands on this tbh... More context can be found [here](https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html#duplicate-information-issue)

FWIW this was explicitly removed from tonic - https://github.com/hyperium/tonic/issues/632

Approving this PR as it fits with what we're doing elsewhere, but it feels like we may want to keep an eye on where the ecosystem is going with respect to error handling in general, and have a bit of a think about how we want to integrate with it.

alamb:
> I'm not sure about this one, the ecosystem in general appears to be moving away from including the source in the Display output, instead preferring to walk source() now that this is stabilised. I'm not sure where Snafu stands on this tbh... More context can be found here

@tustvold  can you help me understand what the preferred method for getting an error message that contains the source is? I don't have any religion on how the code looks I simply want the client to print out the underlying error message

https://github.com/hyperium/tonic/issues/632 says:

>  I think a better approach is to include the source in Error::source so its up to users how much information they want to show.

And then https://github.com/hyperium/tonic/pull/633#issue-632568148 says:

> At Embark we have a little helper function that converts a &dyn std::error::Error into a String by walking the full chain of sources (with std::error::Error::source) and joining them into a String.

Seems like a good idea.. but something that would require us to change all places where we want to see error messages (e.g. in the `info!` logs) to use a similar helper.

tustvold:
Sorry TBC I agree with what this PR is doing :+1:

I was just taking the opportunity to bang on my favourite drum of, "something isn't quite right with the way we are doing error handling in IOx" :laughing:, that it seems to be at odds with where the ecosystem is headed, and trying to provide context on why tonic isn't providing a display impl that renders the source. It certainly isn't priority number 1 for IOx right now, but was just making a note that at some point we probably should properly revisit this.

Sorry for any confusion

alamb:
@tustvold  provided the following link which is also insightful about where Rust is heading:

https://blog.rust-lang.org/inside-rust/2021/07/01/What-the-error-handling-project-group-is-working-towards.html#2-error-reporter

PR #2327, opened 2021-08-17
chore: enable object stores and tracing impls in Dockerfile

jacobmarble:
Make the image generated by Dockerfile very useful for users who don't
have to build it themselves. Enable all object store backends and
tracing destinations.

This will be useful for the image that I auto-publish at https://hub.docker.com/r/jacobmarble/influxdb_iox


alamb (approved):

alamb:
Can you remind me again what uses this dockerfile (as in I wonder why we don't use this dockerfile to create the images that are then deployed to tools). cc @mkmik

jacobmarble:
This Dockerfile is meant to be a simple `docker build` target for people who are used to that. It is referenced in the README under `### Build a Docker image (optional)`

PR #2330, opened 2021-08-17
fix: Make allocs profile record all allocations

mkmik:
Rename `frequency` to `interval` and set the default to `1`, which means that
every allocation will be counted.

Before this PR, we were incorrectly reusing the same "frequency" parameter (and related default of 99)
which meant that we were skipping small allocations that.

The sampling interval is a number of bytes that have to cumulatively allocated for a sample to be taken.
For example if the sampling interval is 99, and you're doing a million of 40 bytes allocations, the allocations profile will account for 16MB instead of 40MB. Heappy will adjust the estimate for sampled recordings, but now that feature is not yet implemented.

### DIFF
--- a/src/influxdb_ioxd/http.rs
+++ b/src/influxdb_ioxd/http.rs
@@ -778,7 +778,7 @@ async fn pprof_home<M: ConnectionManager + Send + Sync + Debug + 'static>(
     );
     let allocs_cmd = format!(
         "/debug/pprof/allocs?seconds={}",
-        PProfArgs::default_seconds()
+        PProfAllocsArgs::default_seconds()
     );
     Ok(Response::new(Body::from(format!(
         r#"<a href="{}">http://{}{}</a><br><a href="{}">http://{}{}</a>"#,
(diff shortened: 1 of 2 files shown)


src/influxdb_ioxd/http.rs
```diff
@@ -821,6 +821,26 @@ impl PProfArgs {
     }
 }
 
+#[derive(Debug, Deserialize)]
+struct PProfAllocsArgs {
+    #[serde(default = "PProfAllocsArgs::default_seconds")]
+    seconds: u64,
+    // Sampling interval
```
alamb, line 828: I think adding the explanation from the PR description as a doc comment here would be super helpful as I would definitely be confused about what units the interval was in and probably would not guess "cumulatively sampled bytes"

> The sampling interval is a number of bytes that have to cumulatively allocated for a sample to be taken.
For example if the sampling interval is 99, and you're doing a million of 40 bytes allocations, the allocations profile will account for 16MB instead of 40MB. Heappy will adjust the estimate for sampled recordings, but now that feature is not yet implemented.

alamb (approved):

src/influxdb_ioxd/http.rs
NGA-TRAN: +1

NGA-TRAN (approved):

src/influxdb_ioxd/http.rs
mkmik: doc updated; thanks

PR #2333, opened 2021-08-17
feat: improve size estimation of roaring bitmaps

### LINKED ISSUE #2257: Read Buffer memory usage optimistic

e-dard:
Closes #2257

This PR improves the estimation of the total memory required by roaring bitmaps in the Read Buffer. It uses statistics about the composition of the bitmaps as an estimation of the amount of memory allocated.

For more information on the bitmap statistics see the paper [Roaring Bitmaps: Implementation of an Optimized Software
Library](https://arxiv.org/pdf/1709.07821.pdf).

In fact, I used the examples in the paper as test cases for the size estimation. Specifically:

Screenshot 2021-08-17 at 22 47 14

In general I think the previous estimation was probably pretty accurate, but this one will be better.

As a follow on I plan to run the bitmap storage optimisation once we have created columns within the Read Buffer as this will be a one-time operation and should reduce memory usage in some cases.


read_buffer/src/column.rs
```diff
@@ -1421,6 +1431,41 @@ mod test {
         assert_eq!(row_ids.to_vec(), vec![2, 3, 4]);
     }
 
+    #[test]
+    fn row_ids_size() {
```
e-dard, line 1435: Here is a test for bitmap size that uses examples from the linked paper to verify correctness. Funnily enough it looks like the paper got the sizing wrong for the RLE container 😄

read_buffer/src/column.rs
```diff
@@ -1320,6 +1324,12 @@ impl RowIDs {
         }
     }
 
+    pub fn optimise_storage(&mut self) {
```
e-dard, line 1327: I will optimise storage of all bitmaps in a follow on PR.

alamb (approved):
makes sense to me

PR #2335, opened 2021-08-18
feat: management API to persist partitions

### LINKED ISSUE #2334: Management API to persist a partition
Wire the `persist_partition` API introduced in #1972 into the gRPC and CLI layer.

This is an alternative for #1935 which is not doable w/o breaking the persistence windows (see [this comment](https://github.com/influxdata/influxdb_iox/issues/1935#issuecomment-900420473)).

crepererum:
Closes #2334.


src/influxdb_ioxd/rpc/error.rs
```diff
@@ -120,6 +120,18 @@ pub fn default_db_error_handler(error: server::db::Error) -> tonic::Status {
             ),
         }
         .into(),
+        Error::CannotFlushPartition {
```
alamb, line 123: I suggest we  make the terminology consistent -- aka use `Persist` everywhere rather than  also sometimes `flush` (I realize the existing code was not consistent on this point either, so we don't have to fix it in this PR).

At least perhaps we can use the term `persist` in the message below even if we leave the enum names the same

tests/end_to_end_cases/management_api.rs
```diff
@@ -1226,3 +1226,103 @@ async fn test_drop_partition_error() {
...
+    .await;
+
+    let chunks = management_client
+        .list_chunks(&db_name)
+        .await
+        .expect("listing chunks");
+    assert_eq!(chunks.len(), 1);
+    let partition_key = &chunks[0].partition_key;
+
+    let err = management_client
```
alamb, line 1323: why can't this persist the partition?

tests/end_to_end_cases/management_api.rs
```diff
@@ -1226,3 +1226,103 @@ async fn test_drop_partition_error() {
...
+        .persist_partition(&db_name, "data", &partition_key[..])
+        .await
+        .unwrap_err();
+    assert_contains!(err.to_string(), "Cannot flush partition");
```
alamb, line 1327: I definitely think this should say "Cannot persist partition" rather than "Cannot **flush** partition"

alamb (approved):
Looks good -- thanks for doing this plumbing @crepererum  💯 

I think it is worth updating the error message, but otherwise this is good to go in my opinion.

tests/end_to_end_cases/management_api.rs
crepererum: The type is a `PersistPartitionError::LifecycleError`. The whole IOx client error handling does NOT add the action (like "do X") as a prefix. If we wanna do that, I suggest we do that in another PR.

Issue #2243, opened 2021-08-10
Pull WriteBuffer out of Db onto Database


tustvold:
There are a couple of reasons for wanting to do this:

* I think it makes for a more intelligible abstraction if `Db` is instance-local, with a write to a `Db` being a write to this instance
* A WriteBufferConsumer writes into a `Db` and so it is a little surprising for it to be owned by it
* It will remove some logic in Db to handle database startup/replay/tests
* It decouples the write buffer from the query components

As per #2242 I plan to do this when I have some idle time, as it isn't urgent right now

Issue #2242, opened 2021-08-10
Pull LifecyclePolicy out of Db into Database


tustvold:
There are a couple of reasons for wanting to do this:

* It removes the need for some `Weak<Db>` shenanigans
* It better fits with the intended design of the LifecyclePolicy operating on `Db`, rather than being owned by it
* It will remove some logic in Db to handle database startup/replay
* It makes it easier to construct tests that need to fake timestamps

PR #2337, opened 2021-08-18
feat: split db worker loop

tustvold:
This is a precursor to #2242 and #2243 and splits up the Db background worker more explicitly. 

As a happy accident this not only makes the code simpler, at least imo, but also fixes what is sort of a bug in that if one of the background loop exits prematurely it previously wouldn't cause termination of the worker as a whole (and by extension server shutdown).


server/src/db.rs
```diff
@@ -821,81 +821,83 @@ impl Db {
...
+                    futures.push(fut);
                 }
-            },
-        );
+
+                futures::future::join_all(futures).await;
+            }
+        };
+
+        tokio::select! {
```
tustvold, line 895: This is fine as all of the futures are fine with being cancelled by being dropped - i.e. they don't need to perform drain logic on shutdown

server/src/db.rs
```diff
@@ -821,81 +821,85 @@ impl Db {
...
+                        metrics,
+                    );
+                    futures.push(fut);
                 }
-            },
-        );
+
+                futures::future::join_all(futures).await;
+            } else {
+                futures::future::pending::<()>().await;
```
tustvold, line 893: This is "meh" but #2243 will remove it anyway so :shrug:

alamb: As I read the docs, `pending()` effectively never resolves (aka never returns) 

Is this needed because if any of the background loops resolves it triggers a server shutdown?

server/src/db.rs
alamb:
```suggestion
        // None of the futures need to perform drain logic on shutdown.
        // When the first one finishes, all of them are dropped
        tokio::select! {
```

alamb (approved):
Makes sense to me.

server/src/db.rs
tustvold: Correct :+1: It's a bit of a hack, but is simpler than making than the future itself conditional

PR #2344, opened 2021-08-19
refactor: remove mem-qe

tustvold:
This isn't used, but is getting built, linted, etc... If we need it again we can restore it from git history


alamb:
It is a good idea in my opinion but I think @e-dard  should get the final say as in the past he has wanted to keep mem_qe in the repo

e-dard:
I was actually doing this last week before I headed off for a break 😄. I have got the crate on a fork so I can easily access the source code, which I find useful sometimes.

alamb (approved):

PR #2348, opened 2021-08-19
feat: jaeger span export (#2273) 

tustvold:
This PR adds the ability to export spans to jaeger


trace_exporters/src/lib.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::num::NonZeroU16;
+use std::sync::Arc;
+use structopt::StructOpt;
+use trace::TraceCollector;
+
+pub mod otel;
+
+/// CLI config for distributed tracing options
+#[derive(Debug, StructOpt, Clone)]
+pub struct TracingConfig {
```
tustvold, line 20: This is copied pretty much verbatim from the trogging crate, but with non-jaeger or no longer relevant options removed. I intend to remove the now orphaned code from trogging in a subsequent PR.

alamb (commented):
FYI @domodwyer  as I think I remember you saying conductor does / will use `trogging` for tracing / logging. 

also cc @mkmik  re the planned changes to `trogging` (maybe `tracing` belongs in `trogging`?)

trace_exporters/Cargo.toml
```diff
@@ -12,8 +12,15 @@ chrono = { version = "0.4" }
 futures = "0.3"
 observability_deps = { path = "../observability_deps" }
 opentelemetry = "0.15"
+opentelemetry-jaeger = { version = "0.14", optional = true }
```
alamb, line 15: > optional = true

💯

alamb (approved):
looks good to me

tustvold:
> (maybe tracing belongs in trogging?)

`trace-exporter` and `trogging` do indeed fill a similar role, however, I wanted to highlight the separation between this crate and tokio-tracing. I.e. `trogging` would continue to be used to configure tokio-tracing and logging, but distributed tracing would be separate

alamb:
image

Issue #2288, opened 2021-08-14
Heap usage statistics / metrics likely broken when heappy is not enabled


alamb:
@mkmik  points out: https://github.com/influxdata/influxdb_iox/pull/2279/files#r688568347 we have likely broken our memory usage statistics when adding heappy support

__Expected behaviour:__
We should remove `features = ["unprefixed_malloc_on_supported_platforms"]` only when heappy is enabled.

Otherwise our memory tracking metrics (which depend on jemalloc statistics) will be off because they won't be capturing croaring bitmaps

__Actual behaviour:__
`unprefixed_malloc_on_supported_platforms` has been removed entirely

alamb:
FYI @mkmik and @NGA-TRAN

mkmik:
We have enabled heappy (or at least we intended to) in the perf build target of circleci

alamb:
https://github.com/influxdata/influxdb_iox/pull/2279#discussion_r689004773 is the location

PR #2350, opened 2021-08-19
feat: memory allocator selection and valgrind support

crepererum:


src/main.rs
```diff
@@ -19,6 +19,8 @@ use observability_deps::tracing::warn;
 use crate::commands::tracing::TracingGuard;
 use influxdb_iox_client::connection::Builder;
 use std::str::FromStr;
+
+#[cfg(all(not(feature = "heappy"), feature = "jemalloc"))]
```
mkmik, line 23: actually we don't need this anymore.

we should always have `unprefixed_malloc_on_supported_platforms` unless we enable heappy.

we removed `unprefixed_malloc_on_supported_platforms` because we enabled heappy but we should really fix the feature selection code so that we don't accidentally get misleading memory metrics, see https://github.com/influxdata/influxdb_iox/issues/2288

mkmik:
yay valgrind

src/influxdb_ioxd.rs
```diff
@@ -127,18 +127,41 @@ fn make_server(
...
+        .read()
+        .unwrap()
+        .to_string()
+}
+
+#[cfg(all(feature = "heappy", feature = "jemalloc"))]
+fn build_malloc_conf() -> String {
+    error!("must use exactly one memory allocator")
+}
+
```
mkmik, line 153: I don't think this logic is correct.

Heappy is using jemalloc underneath, it just a wrapper in front of it (and TODO other allocators)

domodwyer (approved):
Very cool - nice wrapper script too!

src/main.rs
mkmik: let's do this cleanup first:

mkmik: https://github.com/influxdata/influxdb_iox/pull/2352

crepererum: But #2288 isn't fixed, so I suspect we still need the global memory in the no-heappy case? Cargo doesn't allow you to deselect features (like `unprefixed_malloc_on_supported_platforms`) when other features (like `heappy`) are enabled.

mkmik: can't we make this a static build error instead?

mkmik (approved):

PR #2351, opened 2021-08-19
feat: make catalog state test deterministic (#2349)

### LINKED ISSUE #2349: Intermittent failures / flaky db::load::tests::test_catalog_state on ci
Reported by @tustvold 

The test sometimes fails on CI and then will pass again if rerun even with no code changes

__Steps to reproduce:__
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/9427/workflows/722e6239-9fbb-427b-bae5-741b347a56ae/jobs/53833

https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/9418/workflows/5992f027-056d-43d2-ba14-730e84833a2d/jobs/53733

__Expected behaviour:__
pass all the time

__Actual behaviour:__
fails some of the time 

cc @mkmik @crepererum

tustvold:
Closes #2349

This test was previously highly non-deterministic, what files got removed at what stages was up to the order the Hashmap decided by the hashmap. This therefore changes the test to instead add/remove chunks based on chunk ID, so that it adds/removes the same chunk each time. In so doing this should resolve the flakiness, as previously there were various scenarios where it could fail if it happened to remove the "wrong" chunk.

Edit: To expand on the exact cause of the failure

```
let path = expected.keys().next().unwrap();
```

As expected is a Hashmap what that yields is up to a combination of the hash functions random seed and the hash implementation for the type.

This would therefore remove a "random" chunk from the state

```
// remove, add, remove in same transaction
        {
            let path = expected.keys().next().unwrap().clone();
            let metadata = expected.get(&path).unwrap();
            state.remove(&path).unwrap();
            state
                .add(
                    Arc::clone(&iox_object_store),
                    CatalogParquetInfo {
                        path: path.clone(),
                        file_size_bytes: 33,
                        metadata: Arc::clone(metadata),
                    },
                )
                .unwrap();
            state.remove(&path).unwrap();
            expected.remove(&path);
        }
```

If this happened to remove chunk with id 0, the following would fail as `Loader` doesn't actually check that the file exists, but rather that the chunk ID is unique. 

```
let path = expected.keys().next().unwrap();
            let (_, metadata) = make_metadata(&iox_object_store, "fail", chunk_addr(0)).await;
            let err = state
                .add(
                    Arc::clone(&iox_object_store),
                    CatalogParquetInfo {
                        path: path.clone(),
                        file_size_bytes: 33,
                        metadata: Arc::new(metadata),
                    },
                )
                .unwrap_err();
            assert!(matches!(err, Error::ParquetFileAlreadyExists { .. }));
```

Currently a draft as I fix TestCatalogState as currently it does not check that chunk IDs are unique, which the updated formulation of this test now explicitly checks for.


parquet_file/src/catalog.rs
```diff
@@ -1337,36 +1331,34 @@ pub mod test_helpers {
...
-            expected.remove(&path);
         }
         assert_checkpoint(&state, &f, &expected);
 
         // error handling, no real opt
         {
-            // already exists (should also not change the metadata)
-            let path = expected.keys().next().unwrap();
-            let (_, metadata) = make_metadata(&iox_object_store, "fail", chunk_addr(0)).await;
+            // TODO: Error handling should disambiguate between chunk collision and filename collision
```
tustvold, line 1358: Currently the errors don't distinguish between, chunk already exists and parquet file already exists. This results in a slightly confusing situation where you can try to insert a chunk with a path and get back ParquetFileAlreadyExists, but then try to remove said file and get ParquetFileDoesNotExist as whilst a chunk with same ID exists it isn't at that specific path.

parquet_file/src/catalog.rs
```diff
@@ -1379,46 +1371,43 @@ pub mod test_helpers {
...
-            let path = expected.keys().next().unwrap();
-            let (_, metadata) = make_metadata(&iox_object_store, "fail", chunk_addr(0)).await;
+            let (_, metadata) = expected.get(&0).unwrap();
             let err = state
                 .add(
                     Arc::clone(&iox_object_store),
                     CatalogParquetInfo {
-                        path: path.clone(),
+                        // Intentionally "incorrect" path
+                        path: ParquetFilePath::new(&chunk_addr(10)),
```
tustvold, line 1389: This is another example of the slightly confusing error handling, the error here isn't that a parquet file exists at that path, but rather a chunk with the same ID, as determined from the metadata, already exists

parquet_file/src/catalog.rs
```diff
@@ -1182,31 +1227,38 @@ pub mod test_helpers {
...
         }
 
         fn remove(&mut self, path: &ParquetFilePath) -> Result<()> {
-            match self.parquet_files.entry(path.clone()) {
-                Occupied(o) => {
-                    o.remove();
-                }
-                Vacant(v) => {
-                    return Err(Error::ParquetFileDoesNotExist { path: v.into_key() });
+            let partitions = self
```
tustvold, line 1234: This is a bit wild but reflects what `Loader` does, perhaps a subsequent PR should use the information in the ParquetFilePath but currently this isn't exposed, and I wanted to avoid making that change as part of this PR

parquet_file/src/catalog.rs
```diff
@@ -1148,32 +1148,77 @@ impl<'c> Debug for CheckpointHandle<'c> {
...
+        pub chunks: HashMap<u32, CatalogParquetInfo>,
+    }
 
     /// In-memory catalog state, for testing.
-    #[derive(Clone, Debug)]
+    #[derive(Clone, Debug, Default)]
     pub struct TestCatalogState {
         /// Map of all parquet files that are currently pregistered.
-        pub parquet_files: HashMap<ParquetFilePath, CatalogParquetInfo>,
+        pub tables: HashMap<Arc<str>, Table>,
```
tustvold, line 1167: These changes are necessary because TestCatalogState previously did not verify that chunk IDs are unique, making it behave differently from the real catalog.

carols10cents (approved):
Looks great to me! I was definitely confused by these tests when I changed them; this is better 👍🏻

carols10cents:
I'm adding the automerge because I've hit rerun on a few test failures today and I'd like to stop having to do that ;)

alamb:
> I'm adding the automerge because I've hit rerun on a few test failures today and I'd like to stop having to do that ;)


Thank you @carols10cents

parquet_file/src/catalog.rs
alamb: Shouldn't it return `ChunkIdAlreadyExists`?

alamb (approved):
Looks good to me. it would probably be good to have @crepererum  also review this posthumously to make sure we didn't miss anything substantial.

parquet_file/src/catalog.rs
tustvold: Or something similar, yes - this was my point :grin:

PR #2352, opened 2021-08-19
fix: Remove confusing and unnecessary rust global allocator

mkmik:
The jemalloctor rust global allocator only intercepts allocations coming from rust.
We used to enabled the `unprefixed_malloc_on_supported_platforms` on the jemallocator crate
which effectivelly installs jemalloc as a replacement for system malloc so that both rust and C libraries
would use jemalloc. This not only ensures that all libraries use jemalloc for the sake of using
a better allocator, but also allows us to use the jemalloc metrics (exposed by jemalloc-ctl) to figure
out how much memory we are consuming; if we don't enable `unprefixed_malloc_on_supported_platforms` and only install the rust global allocator everything works just fine, except the memory metrics return a lower number because memory allocated by e.g. croaring bitmaps wouldn't be tracked.

However since heappy is is incompatible with `unprefixed_malloc_on_supported_platforms` but effectively plays the same role (by installing a replacement of malloc that ultimately invokes jemalloc) and we
were in a rush last week, we turned on heappy and turned off `unprefixed_malloc_on_supported_platforms`
without doing the full work of figuring out how to tell Cargo that those flags have that mutually-exclusive logic; the followup work has been put in a #2288.

This PR removes this rust global allocator since it's a source of confusion. Now if we'll accidentally
disable jemalloc for C we'll also disable it for rust which will be immediately visible since we won't
have any memory metrics at all.


alamb (approved):
I'll trust you on this one; I would have thought that this change would mean that when compiled without `heappy` we will not use Jemalloc at all.

mkmik:
> I'll trust you on this one; I would have thought that this change would mean that when compiled without `heappy` we will not use Jemalloc at all.

Yes that's why we have the other PR that adds the noheappy feature that deals with that case.

But it any case, this global allocator is not doing anything except confusing developers

alamb:
> But it any case, this global allocator is not doing anything except confusing developers

I am for sure confused :)

mkmik:
> > But it any case, this global allocator is not doing anything except confusing developers
> 
> I am for sure confused :)

Ideally I'd like to find a better way to hook library 

Unfortunately cargo doesn't let us use the right tool which is the linker's "-fwrap" argument

PR #2356, opened 2021-08-19
fix: Make heappy and nonheappy mutually exclusive

### LINKED ISSUE #2288: Heap usage statistics / metrics likely broken when heappy is not enabled

mkmik:
Closes #2288

Layered on top of #2352

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,6 @@ datafusion = { path = "datafusion" }
 data_types = { path = "data_types" }
 entry = { path = "entry" }
 generated_types = { path = "generated_types" }
-heappy = { git = "https://github.com/mkmik/heappy", rev = "82a383128e484039cc2f31476e6289bed48a6701", features = ["enable_heap_profiler", "jemalloc_shim", "measure_free"], optional = true }
 
 influxdb_iox_client = { path = "influxdb_iox_client", features = ["format"] }
 influxdb_line_protocol = { path = "influxdb_line_protocol" }
@@ -117,9 +116,7 @@ serde_urlencoded = "0.7.0"
 snafu = "0.6.9"
 structopt = "0.3.21"
 thiserror = "1.0.23"
-# remember to put "unprefixed_malloc_on_supported_platforms" if you disable heappy
-tikv-jemallocator = { version = "0.4.0" }
-tikv-jemalloc-ctl = "0.4.0"
+tikv-jemalloc-ctl = { version = "0.4.0" }
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "parking_lot", "signal"] }
 tokio-stream = { version = "0.1.2", features = ["net"] }
 tokio-util = { version = "0.6.3" }
@@ -127,6 +124,11 @@ tonic = "0.5.0"
 tonic-health = "0.4.0"
 tonic-reflection = "0.2.0"
 uuid = { version = "0.8", features = ["v4"] }
+
+# jemalloc-sys with unprefixed_malloc_on_supported_platforms feature and heappy are mutually exclusive
+tikv-jemalloc-sys = { version = "0.4.0", optional = true, features = ["unprefixed_malloc_on_supported_platforms"]}
+heappy = { git = "https://github.com/mkmik/heappy", rev = "82a383128e484039cc2f31476e6289bed48a6701", features = ["enable_heap_profiler", "jemalloc_shim", "measure_free"], optional = true }
+
 
 [dev-dependencies]
 # Workspace dependencies, in alphabetical order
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -154,6 +156,10 @@ gcp = ["object_store/gcp"] # Optional GCP object store support
...
-# runtime overhead on all allocations (calls to malloc)
-# pprof is also an optional feature for pprof support
\ No newline at end of file
+# pprof is an optional feature for pprof support
+
+# heappy is an optional feature; Not on by default as it
+# runtime overhead on all allocations (calls to malloc).
+# Cargo cannot currently implement mutually exclusive features so let's force every build
+# to pick either heappy or noheappy feature at least until we figure out something better.
+noheappy = ["tikv-jemalloc-sys"]
```
alamb, line 165: it seems like this should be a "default" feature (aka it would be nice to get this without having to do `--features=noheappy` 🤔

Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN \
   --mount=type=cache,id=influxdb_iox_git,sharing=locked,target=/usr/local/cargo/git \
   --mount=type=cache,id=influxdb_iox_target,sharing=locked,target=/influxdb_iox/target \
   du -cshx /usr/local/cargo/registry /usr/local/cargo/git /influxdb_iox/target && \
-  cargo build --target-dir /influxdb_iox/target --release --features azure,gcp,aws,jaeger,otlp,pprof && \
+  cargo build --target-dir /influxdb_iox/target --release --features azure,gcp,aws,jaeger,otlp,pprof,heappy && \
```
alamb, line 16: FYI @jacobmarble  -- not sure if you have any opinions about heappy being in the dockerfile (possibly slower but also then you can have heap profiling)

src/main.rs
```diff
@@ -51,6 +46,12 @@ static VERSION_STRING: Lazy<String> = Lazy::new(|| {
     )
 });
 
+#[cfg(not(any(feature = "heappy", feature = "noheappy")))]
+compile_error!("you need to pick either heappy or noheappy feature, cargo can't pick a default out of a mutually exclusive set");
```
alamb, line 50: ❤️

alamb (approved):
looks good to me other than potentially adding the `noheappy` as a default option

Cargo.toml
carols10cents: Yeah I would rather have CI have to have `--no-default-features --features=heappy` than to locally have to do `--features=noheappy`

carols10cents:
This does look like the best way that's possible to do this right now, aside from making a default choice for ease of development. Also I wonder if we should switch to [the new feature resolver](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver), which as I understand it, won't *directly* help with this problem, exactly, but would help if dev, build, or target-specific dependencies mess with the `unprefixed_malloc_on_supported_platforms` flag? More details: https://rust-lang.github.io/rfcs/2957-cargo-features2.html

Dockerfile
mkmik: Fwiw, we didn't notice any major slowdowns. 

I still need to compare our benchmarks with and without the heappy linked in.

Cargo.toml
crepererum: I agree w/ the default feature.

Since `noheappy` just implies "use jemalloc unprefixed", why not call the feature `jemalloc_global` or `jemalloc_unprefixed` instead? Then the compile error in main could also explain why they both clash.

Also, we're now effectively back to what #2350 proposed feature-wise, just w/ different names.

crepererum (changes requested):

mkmik:
The main difference is the compile time error vs runtime error

crepererum:
> The main difference is the compile time error vs runtime error

Oh, I meant to use `compile_error` in my PR :sweat_smile:

mkmik:
done:

now by default we enable the `jemalloc_replacing_malloc` (long but explicit name) and use `--no-default-features` everywhere we enable heappy. 

@crepererum PTAL

src/main.rs
```diff
@@ -46,6 +46,12 @@ static VERSION_STRING: Lazy<String> = Lazy::new(|| {
     )
 });
 
+#[cfg(not(any(feature = "heappy", feature = "jemalloc_replacing_malloc")))]
+compile_error!("you need to pick either heappy or jemalloc_replacing_malloc feature, cargo can't pick a default out of a mutually exclusive set");
```
crepererum, line 50: I'll probably remove this particular error because for valgrind we need neither of them (aka use the default system allocator).

crepererum (approved):

src/main.rs
mkmik: ok, feel free to change this in the PR that enables valgrind. I'd like to make sure we're not accidentally losing memory stats

PR #2357, opened 2021-08-19
refactor: move data generator to IOx repo and fix build

pauldix:
Just a lift and shift from the other repo along with a bunch of little changes to fix imports and builld.

PR #2358, opened 2021-08-19
feat: create initial performance test

jacobmarble:
Create a new environment for performance tests. Test by simply running `perf/perf.py all`

Remaining:
- ~generate data per some spec; see https://github.com/influxdata/influxdb_iox/pull/2357~
- ~execute some list of SQL queries~
  - ~record duration~
  - ~compare results to reference for correctness~
- ~output report~
- ~handle ctrl-C better~
- ~document this thing~


crepererum:
Could you try a clean rebase? It seems your branch picked up commits of two PRs (or at least GitHub thinks that).

perf/perf.py
```diff
@@ -0,0 +1,608 @@
...
+            docker_run_zookeeper(dc)
+            docker_run_kafka(dc)
+        else:
+            docker_run_redpanda(dc)
+        if args.object_store == 's3':
+            docker_run_minio(dc)
+        docker_run_jaeger(dc)
+        processes['iox_router'] = exec_iox(1, 'iox_router', debug=args.debug, object_store=args.object_store)
+        processes['iox_writer'] = exec_iox(2, 'iox_writer', debug=args.debug, object_store=args.object_store)
+        # time.sleep(0.1)  # TODO is this necessary?
```
crepererum, line 78: technically you need to wait here until the server is ready. There is a short window when the gRPC service is online but the server is not ready. Use this to figure out if the server is ready:

https://github.com/influxdata/influxdb_iox/blob/6ba25728e10c9503befda8874a62cab0f6f71236/generated_types/protos/influxdata/iox/management/v1/service.proto#L77-L78

Technical background: during that time the server looks for existing DBs and will load preserved data and executes Kafka replays. Since you don't have any existing DBs, the time window is short enough to sleep over it here, but if the users system is quite slow (or under load), this might be an issue.

perf/perf.py
```diff
@@ -0,0 +1,608 @@
...
+import docker
+import grpc_requests
+import minio
+
+ioxperf_name = "ioxperf"
+ioxperf_labels = {ioxperf_name: None}
+ioxperf_filters = {'label': ioxperf_name}
+org_name = 'myorg'
+bucket_name = 'mybucket'
+db_name = '%s_%s' % (org_name, bucket_name)
```
crepererum, line 29: I prefer modern f-stings (amongst many other things like type annotations), but since this isn't run in production I think it's fine.

perf/perf.py
```diff
@@ -0,0 +1,608 @@
...
+
+        print('running test "%s"' % name)
+        time_start = time.time()
+        params = {'q': sql, 'format': 'csv'}
+        response = requests.get(url=query_url, params=params)
+        time_delta = '%dms' % math.floor((time.time() - time_start) * 1000)
+
+        if not response.ok:
+            print(response.reason)
+            print(response.content.decode('utf-8'))
```
crepererum, line 573:
```suggestion
            print(response.text)
```

Same pattern is used in other places as well.

perf/perf.py
```diff
@@ -0,0 +1,608 @@
...
+import subprocess
+import threading
+import time
+import urllib.parse
+
+import requests
+import toml
+import urllib3
+
+import docker
```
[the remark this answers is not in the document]
jacobmarble: TIL `requirements.txt`. Fixed.

perf/perf.py
jacobmarble: FWIW, I do this via the CLI (`server wait-server-initialized`), but `create_database.rs` occasionally requires this sleep. So far `grpc_create_database()` seems to be slow enough, so I'll drop the comment.

I'll switch from checking via CLI to gRPC.

perf/perf.py
jacobmarble: TIL f-strings 👍 . For now, I'll keep these legacy formatters in place.

perf/perf.py
crepererum: The CLI   uses the same gRPC interface. I'm a bit surprised that the sleep is required.

Proposal: You can leave it as it is and we can merge it. I'll have a look once it's merged and see how / why this is broken.

jacobmarble: That was more involved that I thought, but it's done now.

crepererum (approved):

Issue #1871, opened 2021-07-01
Implement `delete` database management API


alamb:
__Use case:__
Cleaning up for distributed database creation after partial failure. As @domodwyer  says, 

> if one server does not accept a create_db request it’ll try and revert the change on the others that did accept it by deleting [the database on the server that it worked]

__Proposal:__
Implement a drop_database gRPC management Service API (client and server)

__Current behaviour:__
Can not delete

__Desired behaviour:__
able to delete databases

__Implementation designed in #1881:__

- Add a delete database API that takes a database name. When database `foo_bar` gets deleted,
  - The current database knows its generation number. Add the tombstone file to that generation subdirectory and return success. 
  - If the tombstone file already exists in that generation directory, don't recreate the file, but return success. This could happen as a race condition with another call to this API for the same database but the way delete is implemented, it's not an error - the database ends up marked as deleted, which is what we requested.
  - If that generation subdirectory does not exist, return not found because we can't find any database with that name to delete. This could happen as a race condition with the hard delete operation.

PR #2359, opened 2021-08-19
WIP: yes i am using CI because i need a new computer, i'm waiting til the new macbooks in october

carols10cents:
Connects to #1871.


carols10cents:
Actually I think this will be clearer as part of the other PR i'm working on

PR #2360, opened 2021-08-19
feat: Add batching to the data generator

pauldix:
Adds batch_size to the data genrator to optionally gather multiple calls to generate for each agent. For example, if you have a sampling interval of 10 seconds and start at some point back in time with a batch size of 3, it gather 3 samplings before writing to the points writer. For runs against a server API, this will batch them together in a single API call.

### DIFF
--- a/iox_data_generator/src/agent.rs
+++ b/iox_data_generator/src/agent.rs
@@ -126,11 +126,20 @@ impl<T: DataGenRng> Agent<T> {
 
     /// Generate and write points in batches until `generate` doesn't return any
     /// points. Meant to be called in a `tokio::task`.
-    pub async fn generate_all(&mut self, mut points_writer: PointsWriter) -> Result<usize> {
+    pub async fn generate_all(
+        &mut self,
+        mut points_writer: PointsWriter,
+        batch_size: u16,
+    ) -> Result<usize> {
         let mut total_points = 0;
 
         let mut points = self.generate().await?;
+        let mut batches = 1;
         while !points.is_empty() {
+            while batches < batch_size {
+                points.append(&mut self.generate().await?);
+                batches += 1;
+            }
             info!("[agent {}] sending {} points", self.name, points.len());
             total_points += points.len();
             points_writer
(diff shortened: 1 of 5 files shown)


iox_data_generator/src/agent.rs
```diff
@@ -126,18 +126,28 @@ impl<T: DataGenRng> Agent<T> {
 
     /// Generate and write points in batches until `generate` doesn't return any
     /// points. Meant to be called in a `tokio::task`.
-    pub async fn generate_all(&mut self, mut points_writer: PointsWriter) -> Result<usize> {
+    pub async fn generate_all(
+        &mut self,
+        mut points_writer: PointsWriter,
+        batch_size: u16,
```
alamb, line 132: I suggest using a  `usize` for `batch_size` -- it would avoid potentially nasty surprises for anyone who did like `batch_size=100000`

iox_data_generator/src/agent.rs
```diff
@@ -126,18 +126,28 @@ impl<T: DataGenRng> Agent<T> {
 
     /// Generate and write points in batches until `generate` doesn't return any
     /// points. Meant to be called in a `tokio::task`.
```
alamb, line 128: it might be worth explaining in this docstring that `batches` means the number of times `generate_points` will be invoked per write.

alamb (approved):
looks good to me -- I suggest using a `usize` rather than `u16` (mostly as I don't see any benefit to using a u16 and I can see some potential uses where that might be confusing)

iox_data_generator/src/agent.rs
```diff
@@ -125,19 +125,30 @@ impl<T: DataGenRng> Agent<T> {
...
+    ) -> Result<usize> {
         let mut total_points = 0;
 
         let mut points = self.generate().await?;
+        let mut batches = 1;
         while !points.is_empty() {
+            while batches < batch_size {
+                points.append(&mut self.generate().await?);
+                batches += 1;
+            }
```
shepmaster, line 143:
```suggestion
            for _ in 0..batch_size {
                points.append(&mut self.generate().await?);
            }
```

shepmaster (commented):
Minor thing

PR #2361, opened 2021-08-20
fix: The path to db rules.pb shouldn't end in a slash

carols10cents:
Sorry about that @jacobmarble, but this should fix your problem with minio. @shepmaster's [fault for writing the bug](https://github.com/influxdata/influxdb_iox/commit/63111d9d9adf16557141a204ab218e34f212d675), my fault for not catching that he didn't write any tests 😛

And I actually think there should be more tests that I'm happy to write in the morning!

### DIFF
--- a/iox_object_store/src/lib.rs
+++ b/iox_object_store/src/lib.rs
@@ -21,7 +21,7 @@ use futures::{
     Stream, StreamExt, TryStreamExt,
 };
 use object_store::{
-    path::{parsed::DirsAndFileName, Path},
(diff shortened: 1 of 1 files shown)


jacobmarble (approved):
Sweeeeet! Thank you

carols10cents:
I'm going to merge this in because I actually think stuff will be really broken next time this gets deployed and the catalog wiped, and I don't want to have everything broken in Europe morning. 

I promise more tests are coming!

PR #2290, opened 2021-08-15
docs: Remove the description of unused commands in cli

capkurmagati:
Describe your proposed changes here.

The commands `convert`, `meta` and `stats` have been removed by #1639. 
This PR removes the description in the command line. 

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #2364, opened 2021-08-20
docs: Update system table schema in docs/sql.md

capkurmagati:
- Column `count` is renamed to `row_count`  via #1937
- System tables splits into smaller ones  #2061
- This PR updates the sql to reflect those changes.

### DIFF
--- a/docs/sql.md
+++ b/docs/sql.md
@@ -198,8 +198,8 @@ LIMIT 20;
 ### Total row count by table
 
 ```sql
-SELECT database_name, table_name, max(count) as total_rows
-FROM columns
+SELECT database_name, table_name, max(row_count) as total_rows
+FROM chunk_columns
 GROUP BY database_name, table_name
 ORDER BY total_rows DESC
 LIMIT 20;
@@ -208,9 +208,9 @@ LIMIT 20;
 ### Total row count by partition and table
 
 ```sql
-SELECT database_name, partition_key, table_name, max(count) as total_rows
-FROM columns
-GROUP BY database_name, partition_key, table_name
+SELECT database_name, table_name, max(row_count) as total_rows
+FROM chunk_columns
+GROUP BY database_name, table_name
 ORDER BY total_rows DESC
 LIMIT 20;
 ```
@@ -229,9 +229,9 @@ total_rows
 from
 (select table_name,
         column_name,
-        sum(count) as total_rows,
+        sum(row_count) as total_rows,
         max(cast(max_value as double)) - min(cast(min_value as double)) as range
- from system.chunk_columns
+ from chunk_columns
  where column_name = 'time'
  group by table_name, column_name
 )
--- a/docs/sql.md
+++ b/docs/sql.md
@@ -198,8 +198,12 @@ LIMIT 20;
 ### Total row count by table
 
 ```sql
-SELECT database_name, table_name, max(row_count) as total_rows
-FROM chunk_columns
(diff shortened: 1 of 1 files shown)


docs/sql.md
```diff
@@ -198,8 +198,8 @@ LIMIT 20;
 ### Total row count by table
 
 ```sql
-SELECT database_name, table_name, max(count) as total_rows
-FROM columns
+SELECT database_name, table_name, max(row_count) as total_rows
```
capkurmagati, line 201: This query does not actually return the `Total row count by table`.
Instead, it returns the row count of the largest partition by table.
I think we might need something like:
```
SELECT database_name, table_name, sum(total_rows) as total_rows
FROM (
  SELECT database_name, table_name, max(row_count) as total_rows
  FROM chunk_columns
  GROUP BY database_name, partition_key, table_name
)
GROUP BY database_name, table_name
ORDER BY total_rows DESC;
```

alamb: yes I agree -- nice catch @capkurmagati 👍

docs/sql.md
```diff
@@ -208,9 +208,9 @@ LIMIT 20;
 ### Total row count by partition and table
```
alamb, line 208: Given the query has been changed, I think we should also change the section title

```suggestion
### Total row count by table
```

alamb (commented):
Thank you @capkurmagati . This is looking good

docs/sql.md
capkurmagati: Ah, I accidentally removed the `partition_key`. Let me put it back.

docs/sql.md
capkurmagati: Thanks! I will update the pr and rebase against main branch.

capkurmagati:
@alamb PR updated. PTAL.

alamb (approved):
Looks great -- thank you @capkurmagati !

alamb:
For some reason circle CI checks don't appear to be running for this PR.

I think this repo is setup correctly according to https://circleci.com/docs/2.0/oss/#build-pull-requests-from-forked-repositories (has "build forked pull requests"). However that page also says

> Note: If a user submits a pull request to your repository from a fork, but no pipeline is triggered, then the user most likely is following a project fork on their personal account rather than the project itself of CircleCi, causing the jobs to trigger under the user’s personal account and not the organization account. To resolve this issue, have the user unfollow their fork of the project on CircleCI and instead follow the source project. This will trigger their jobs to run under the organization when they submit pull requests.

Does this make sense to you @capkurmagati ? If not, I will just pull your changes into another PR and we'll get it merged that way but I figured I would ask if you had the time

capkurmagati:
@alamb Thanks for the information.
I'm not sure if there is any step I've missed, but I couldn't find where the source project is.
I can only choose my org(the forked project is in there and I'm not following it). 
image
I wonder if my last PR #2290 triggered the CI.

alamb:
Thanks @capkurmagati  --

> I wonder if my last PR #2290 triggered the CI.

It did indeed -- I wonder if it was related to when I rebased the PR. Let me add some minimal commit to this one and see if I can trigger the CI

alamb:
Ok, the checks are running 🎉  I expect this PR to be merged imminently 🤞

alamb:
🎉

capkurmagati:
@alamb Thanks for the help. Much appreciated!

alamb:
Thank you @capkurmagati  for the contribution

Issue #2076, opened 2021-07-21
Clean up metrics API


crepererum:
The `metrics` crate is a bit of an API mess at the moment:

- **clone:** For some metrics types `Clone` is implemented, for others it isn't
- **labels by ref / by value:** Some constructors take `&[KeyValue]`, others `Vec<KeyValue>`
- **drop:** While counters can be dropped w/o consequence, gauges reset their value on drop
- **`u64`/`usize`:** Some metrics accept `usize` as a value, others `u64`
- **labels:** make it easier to reason about the final labels (see [this comment](https://github.com/influxdata/influxdb_iox/pull/2077#discussion_r673847959))

e-dard:
FWIW I am not a fan of gauges being decremented when objects are dropped. I find it surprising, and I don't like surprising :-) 

I'm happy for gauges to decrement on drop but I think it would be explicitly done via a `Drop` implementation in the code that is being instrumented.

tustvold:
The reason for gauges decrementing on drop is conceptually they are an observation of the current state (they're actually an aggregation of a number of observations). If the value no longer exists to be observed I would expect the value to no longer include it. This is the key thing that distinguishes gauges from a counter, one is an observation of the current state, the other an accumulation of past states. If you leave the drop to be manual you then end up with orphaned metrics, which imo is even more confusing...

Otherwise I agree entirely with the points raised. Personally I'd like to stabilise on `&[(&str, &str)]` as the KeyValue API is obtuse, but I seem to remember at the time this was non-trivial for whatever reason

e-dard:
@tustvold I completely agree on the semantics of gauges, merely that from a software engineering perspective I (and I'm sure future committers) will find it confusing that metrics have to be explicitly created and wired in etc, yet some of them (some gauges) are _implicitly_ decremented when objects are dropped. That's my only point really.

tustvold:
> (some gauges)

It should be all gauges? Both Gauge and GaugeValue should decrement on drop?

e-dard:
> It should be all gauges? Both Gauge and GaugeValue should decrement on drop? If not that's a bug...
what about gauges that are tracking something orthogonal to the lifetime of the object doing the tracking? For example a gauge that tracks the last time something was processed. If the processor (that's managing the gauge metric) drops I wouldn't expect the value of the gauge to change. 

Perhaps I'm still misunderstanding the design :-)

tustvold:
> tracks the last time something was processed

[...]

PR #2365, opened 2021-08-20
feat: add new metrics abstraction (#2076)

tustvold:
_This needs a load more tests, and some further implementation, but I wanted get a draft PR up to get feedback._

The current metrics abstraction, provided by the `metrics` crate, has slowly grown over time as we worked around limitations, bugs, added new functionality, etc... However, this evolution has now resulted in a crate that is trying to provide a metrics abstraction that fits the requirements of IOx, on top of a fairly opinionated metrics abstraction with a different set of design goals. Additionally as we've used the abstraction provided by the `metrics` crate, we have identified some aspects of it we would perhaps like to change.

This PR therefore proposes an API that fits the requirements so far as I understand them. It should:

* Require minimal additional dependencies to instrument a given crate (#2192)
* Decouple metric recording from metric export
* Be easy to reason about what labels are associated with a given metric (#2076)
* Be easy to grep for a given metric name
* Allow amortizing any label manipulation over multiple metric records
* Individual metric recording should be as cheap as possible
* Allow exposing the recorded metrics to internal systems (e.g. lifecycle, system tables, etc...)
* It should be possible to hook up alternative metric sinks
* Test instrumentation directly without relying on a, potentially extremely large, prometheus dump

Major changes this PR makes worth highlighting:

* The concept of a metrics domain is now implicit in the metrics name - this means a metric name prefix should always appear in the codebase as a literal string
* There is a version of Gauge that does not decrement on drop, and is not aggregated
* What was previously called `Gauge` is now called `CumulativeGauge`
* OpenTelemetry is no longer a dependency and `prometheus` is only a dependency for the exporter crate

I believe everything from `RedMetric` to `LockTracker` could easily be ported to use these new abstractions.


metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+pub use duration::*;
+pub use gauge::*;
+pub use histogram::*;
+
+/// A `Registry` stores a map of metric names to `Metric<T>`
+///
+/// It allows retrieving them by name, registering new metrics and generating
+/// reports of all registered metrics
+#[derive(Debug, Default)]
+pub struct Registry {
```
tustvold, line 34: The idea is there would be a single global `Arc<Registry>` which would get passed around

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    ///
+    pub fn register_with_options<T: MetricObserver, F: FnOnce() -> T::Options>(
+        &self,
+        name: &'static str,
+        description: &'static str,
+        options: F,
+    ) -> Metric<T> {
+        let mut instruments = self.instruments.lock();
+        let metric = match instruments.entry(name) {
+            Entry::Occupied(o) => match o.get().as_any().downcast_ref::<Metric<T>>() {
```
tustvold, line 84: Metrics should generally be registered once and then kept around, so the cost of this dynamic cast is likely irrelevant

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn report(&self, reporter: &mut dyn Reporter);
+
+    /// Returns the type as [`Any`](std::any::Any) so that it can be downcast to
+    /// it underlying type
+    fn as_any(&self) -> &dyn Any;
+}
+
+/// `Reporter` is the trait that should be implemented by anything that wants to
+/// extract the state of all metrics within a `Registry` and export them
+pub trait Reporter {
```
tustvold, line 129: Without GATs it is very hard to write something that uses types to prevent calling methods in an illegal sequence, or at least I couldn't work out how to do it. I think this is probably good enough, `futures::Stream` has the same problem

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+/// A `Registry` stores a map of metric names to `Metric<T>`
+///
+/// It allows retrieving them by name, registering new metrics and generating
+/// reports of all registered metrics
+#[derive(Debug, Default)]
+pub struct Registry {
+    /// A list of instruments indexed by metric name
+    ///
+    /// A BTreeMap is used to provide a consistent ordering
+    instruments: Mutex<BTreeMap<&'static str, Box<dyn Instrument>>>,
```
tustvold, line 38: Technically prometheus doesn't require this, but given we can provide this it makes sense to do so. Any cost of a BTreeMap is likely far outweighed by the cost of constructing `Attributes`. For hot paths the `Recorder` should be cached and reused

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn report(&self, reporter: &mut dyn Reporter) {
+        let instruments = self.instruments.lock();
+        for instrument in instruments.values() {
+            instrument.report(reporter)
+        }
+    }
+}
+
+/// `Instrument` is a type-erased version of `Metric<T>` that is stored within `Registry`
+trait Instrument: std::fmt::Debug {
```
tustvold, line 113: This is just an implementation detail of how typed `Metric<T>` are stored in `Registry`

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Finish recording a given metric
+    ///
+    /// Must not be called without a prior call to start_metric with
+    /// no intervening call to finish_metric
+    fn finish_metric(&mut self);
+}
+
+/// Identifies the type of `Observation` reported by this `Metric`
+#[derive(Debug, Copy, Clone)]
```
tustvold, line 155: I couldn't work out how to avoid the duplication between `MetricKind` and `Observation`

metric/src/metric.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// let registry = Registry::new();
+    /// let metric: Metric<U64Gauge> = registry.register("metric_name", "description");
+    ///
+    /// metric.recorder(&[("foo", "bar")]).set(34);
+    ///
+    /// let recorder = metric.recorder(&[("foo", "biz")]);
+    /// recorder.set(34);
+    ///
+    /// ```
+    pub fn recorder(&self, attributes: impl Into<Attributes>) -> T::Recorder {
```
tustvold, line 63: The `T::Recorder` returned can be cached and used to record observations without needing to perform string manipulation

metric/src/metric.rs
```diff
@@ -0,0 +1,130 @@
...
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+/// Types that wish to be used with `Metric` must implement this trait
+/// that exposes the necessary reporting API
+///
+/// `Metric` maintains a distinct `MetricObserver` for each unique set of `Attributes`
+pub trait MetricObserver: MakeMetricObserver + std::fmt::Debug + 'static {
```
tustvold, line 94: This is the trait that different types of metric recorder must implement. It is currently focused on observers that match prometheus' metric model, but there is no particular reason you couldn't write an Observer that retains a list of observations and reports some custom aggregation. I believe this was something @pauldix was interested in

metric_exporters/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+            MetricKind::U64Counter => {
+                (format!("{}_total", metric_name), MetricType::COUNTER)
+            }
+            MetricKind::U64Gauge => {
+                (format!("{}", metric_name), MetricType::GAUGE)
+            }
+            MetricKind::U64Histogram => {
+                (format!("{}", metric_name), MetricType::HISTOGRAM)
+            }
+            MetricKind::DurationCounter => {
```
tustvold, line 53: This is why Duration is treated differently - there is a convention in prometheus that durations are reported as floating point seconds. I don't know why, but this formulation keeps this as an implementation detail of the exporter. Other exporters could use the actual recorded data which is integer nanoseconds.

metric_exporters/src/lib.rs
```diff
@@ -0,0 +1,148 @@
...
+impl<'a, W: Write> metric::Reporter for PrometheusTextEncoder<'a, W> {
+    fn start_metric(
+        &mut self,
+        metric_name: &'static str,
+        description: &'static str,
+        kind: MetricKind,
+    ) {
+        assert!(self.metric.is_none(), "metric already in progress");
+
+        let (name, metric_type) = match kind {
```
tustvold, line 43: This is why Duration is treated differently - there is a convention in prometheus that durations are reported as floating point seconds. I don't know why, but this formulation keeps this as an implementation detail of the exporter. Other exporters could use the actual recorded data which is integer nanoseconds.

alamb (commented):
Structurally this interface looks really nice to me and seems like it could be used for all the uses of Metrics I have seen in the IOx . I don't have strong religion or experience about metrics but this API looks good to me.

I for one am excited

metric/src/metric.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use super::{Attributes, Instrument, MetricKind, Observation, Reporter};
+
+/// A `Metric` collects `Observation` for each unique set of `Attributes`
+///
+/// It is templated by `T: MetricObserver` which determines the type of
+/// `Observation` made by this `Metric` along with its semantics
+#[derive(Debug)]
+pub struct Metric<T: MetricObserver> {
+    name: &'static str,
```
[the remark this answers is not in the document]
tustvold: Yes, it is a specific goal that the metric name has to appear "in the plain"

metric/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    DurationCounter(std::time::Duration),
+    DurationGauge(std::time::Duration),
+    U64Histogram(Vec<ObservationBucket<u64>>),
+    DurationHistogram(Vec<ObservationBucket<std::time::Duration>>),
+}
+
+/// A bucketed observation
+///
+/// Stores the number of values that were less than `le` and
+/// greater than the `le` of the previous bucket
```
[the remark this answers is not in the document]
tustvold: This should read less than or equal to `le` and strictly greater than `le` of the previous bucket - will fix

metric/src/metric.rs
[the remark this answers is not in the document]
alamb: probably a good idea to include that rationale as doc comments (you did spell it out very nicely in this PR's rationale)

sunng87:
Just for your reference, I have been working on a metrics library [metriki](https://github.com/sunng87/metriki) that follows design of Java's dropwizard metrics, which is used in database like cassandra and many other Java projects. It has clear separation of metrics itself and reporter/exporters.

alamb:
@sunng87  -- https://github.com/sunng87/metriki looks very cool 👍

tustvold:
Thank you @sunng87, metriki definitely looks cool and bears a lot of implementation similarity to what is in this PR which is :ok_hand:. That being said it doesn't appear to have a notion of metric attributes/labels/tags, with each metric name instead associated with a single measurement. This is likely to be a deal breaker for us.

Typically we would want to reuse the same metric name with different sets of tags, to allow computing aggregations at different granularities at query time (e.g. per-host, per-table, etc...). Prometheus calls this concept labels, opentelemetry attributes, and Datadog and InfluxDB tags, but they're all the same thing.

_FWIW I did some digging into DropWizard's support for these concepts and found https://github.com/dropwizard/metrics/issues/1272 and https://github.com/dropwizard/metrics/issues/1175, which I think were intended to be in a 5.0 release, but this appears to have been abandoned https://github.com/dropwizard/metrics/issues/1704_

sunng87:
@tustvold @alamb Thank you for your investigation. I will be looking into this PR to see if I can merge some features into metriki. And perhaps I can help with future development of this metrics module.

e-dard:
This looks good to me @tustvold. Great job. Here are some things that I think are super important (I think you have covered them by the way).

 - we should be able to export prometheus histograms with whatever bucketing system we want.
 - metrics need to be testable in isolation without worrying about what other crates/modules etc are doing.

The first doesn't work in the current `metrics` crate due to the issues with OT and the latter it looks like you have got a similar ability in this crate so I think it's great.

Issue #2370, opened 2021-08-22
Add OTLP Trace Exporter


tustvold:
The otlp trace exporter has been identified as destination we would like to be able to send trace data to. However, currently `trace-exporters` only has support for Jaeger. This should be rectified.

PR #2371, opened 2021-08-22
chore: remove orphaned code from trogging

tustvold:
This code is no longer used by IOx, and so I think it can go?


.circleci/config.yml
```diff
@@ -273,10 +273,10 @@ jobs:
       - cache_restore
       - run:
           name: Print rustc target CPU options
-          command: cargo run --release --no-default-features --features="aws,gcp,azure,jaeger,otlp,heappy" --bin print_cpu
+          command: cargo run --release --no-default-features --features="aws,gcp,azure,jaeger,heappy" --bin print_cpu
```
tustvold, line 276: I've created #2370 to track re-adding otlp. It should be very straightforward to add, but I didn't think urgent?

trogging/src/lib.rs
```diff
@@ -427,53 +242,37 @@ where
...
         tracing_log::LogTracer::init()?;
-        Ok(TracingGuard)
+        Ok(TroggingGuard)
     }
 }
 
-/// A RAII guard. On Drop, tracing and OpenTelemetry are flushed and shut down.
+/// A RAII guard. On Drop, ensures all events are flushed
+///
+/// Note: This is currently unnecessary but has been kept in case we choose to
```
tustvold, line 269: I think moving to tracing-appender might allow removing some of the other complexity in this module around write buffering, but I didn't want to do this as part of this PR

Cargo.lock
```diff
@@ -4749,35 +4678,6 @@ dependencies = [
  "serde",
 ]
 
-[[package]]
-name = "tonic"
```
tustvold, line 4753: We no longer have mixed tonic versions :tada:

.circleci/config.yml
mkmik: it's not urgent, it's more just something we want to do to partecipate in the otlp ecosystem

mkmik (approved):

mkmik:
so to summarize, we're no longer going to use the  `tracing`  (formerly known as tokio tracing) library to produce distributed traces because we're also using it for logging and instead of fighting the two use cases we are going to use our own `trace` library to create spans directly?

tustvold:
> because we're also using it for logging and instead of fighting the two use cases we are going to use our own trace library to create spans directly?

Not really, the fact we are using tokio-tracing for logging is not really a factor, it is more that tokio-tracing does not at the current time support distributed tracing in any meaningful way. I wrote up the full rationale for this decision on the original PR here - https://github.com/influxdata/influxdb_iox/pull/2274#issue-711236406

PR #2373, opened 2021-08-22
refactor: reuse IOxExecutionContext

tustvold:
As part of adding query tracing (#2273) I need to associate a trace context with a specific query. Ideally the context would be extracted from the request extension, and bundled into the IOxExecutionContext used to run the queries.

Unfortunately this is currently not possible:

* IOxExecutionContext is private to the query crate
* Planners create there own contexts on-demand
* Execution doesn't use the same context as planning
* Execution may use different contexts for different plans within the same "query"

This PR tries to address by moving the execution methods off `Executor` and onto `IOxExecutionContext`. This allows callers to construct a single execution context from the `Executor`, and then use it for the entirety of query execution. Perhaps unsurprisingly this creates a fair bit of churn, but it is mostly mechanical.


query/src/exec.rs
```diff
@@ -151,157 +76,6 @@ impl Executor {
         &mut self.config
     }
 
-    /// Executes this plan on the query pool, and returns the
-    /// resulting set of strings
-    pub async fn to_string_set(&self, plan: StringSetPlan) -> Result<StringSetRef> {
```
tustvold, line 156: These methods are moved into IOxExecutionContext

query/src/exec/context.rs
```diff
@@ -189,15 +199,13 @@ impl IOxExecutionContext {
     }
 
     /// registers a catalog with the inner context
-    pub fn register_catalog(&mut self, name: impl Into<String>, catalog: Arc<dyn CatalogProvider>) {
+    pub fn register_catalog(&self, name: impl Into<String>, catalog: Arc<dyn CatalogProvider>) {
```
tustvold, line 202: There isn't actually a reason for this to be `&mut self` as `ExecutionContext` has interior mutability. Removing &mut self makes it slightly clearer that this will also modify all other `Clone` of this `IOxExecutionContext`

query/src/exec/context.rs
```diff
@@ -155,6 +164,7 @@ impl IOxExecutionConfig {
 /// Eventually we envision this also managing additional resource
 /// types such as Memory and providing visibility into what plans are
 /// running
+#[derive(Clone)]
```
tustvold, line 167: This still feels a bit funky to me, but it matches what is done by DataFusion. Mixing Clone and interior mutability I at least find exceedingly confusing as it isn't immediately clear if a modification made to one Clone will also impact the others, whereas if you have say `Arc<InteriorMutable>` it is clear. I may revisit this

query/src/exec/context.rs
```diff
@@ -253,11 +259,170 @@ impl IOxExecutionContext {
...
+                        .convert(
+                            table_name,
+                            tag_columns,
+                            field_columns,
+                            num_prefix_tag_group_columns,
+                            it,
+                        )
+                        .await
+                        .map_err(|e| {
+                            Error::Execution(format!(
```
tustvold, line 314: I opted to kept IOxExecutionContext returning a DatafusionError, the user-facing behaviour should be the same (or better).

FWIW this sort of error propagation is where something like anyhow really shines...

query/src/exec/task.rs
```diff
@@ -154,6 +156,7 @@ impl DedicatedExecutor {
     /// Only the first all to `join` will actually wait for the
     /// executing thread to complete. All other calls to join will
     /// complete immediately.
+    #[allow(dead_code)] // https://github.com/influxdata/influxdb_iox/issues/2372
```
tustvold, line 159: I don't know why clippy only just decided to moan about this but :shrug:

query/src/frontend/sql.rs
```diff
@@ -1,70 +1,7 @@
 use std::sync::Arc;
 
-use snafu::{ResultExt, Snafu};
-
-use crate::exec::{context::DEFAULT_CATALOG, Executor, ExecutorType};
-use datafusion::{
-    catalog::catalog::CatalogProvider, error::DataFusionError, physical_plan::ExecutionPlan,
-};
-
-#[derive(Debug, Snafu)]
```
tustvold, line 10: All bar one of these variants would never be returned, so I just opted to remove this error enumeration. I think it dates from before DataFusion had catalog support

query/src/frontend/sql.rs
```diff
@@ -84,10 +21,9 @@ impl SqlQueryPlanner {
         &self,
         database: Arc<D>,
         query: &str,
-        executor: &Executor,
+        ctx: &IOxExecutionContext,
     ) -> Result<Arc<dyn ExecutionPlan>> {
-        let mut ctx = executor.new_context(ExecutorType::Query);
         ctx.register_catalog(DEFAULT_CATALOG, database);
-        ctx.prepare_sql(query).context(Preparing)
+        ctx.prepare_sql(query)
```
tustvold, line 27: This SQLQueryPlanner does feel a little bit redundant, now that most of the logic is contained in Datafusion...

src/influxdb_ioxd/planner.rs
```diff
@@ -85,17 +58,13 @@ impl Planner {
...
-                        .table_names(database.as_ref(), predicate)
-                        .context(InfluxRpc)
-                },
-                ExecutorType::Query,
-            )
+        self.ctx
+            .run(async move {
+                planner
+                    .table_names(database.as_ref(), predicate)
+                    .map_err(|e| Error::Plan(format!("table_names error: {}", e)))
```
tustvold, line 65: Similar to IOxExecutionContext I opted to just use DatafusionError so I could reuse the `IOxExecutionContext::run` in practice this actually gives better error context than Snafu and is less code....

query_tests/src/pruning.rs
```diff
@@ -70,14 +70,9 @@ async fn chunk_pruning_sql() {
...
 
-    let executor = db.executor();
-    let physical_plan = SqlQueryPlanner::default()
-        .query(db, query, &executor)
-        .unwrap();
-    let batches = executor
-        .collect(physical_plan, ExecutorType::Query)
-        .await
-        .unwrap();
+    let ctx = db.executor().new_context(ExecutorType::Query);
```
alamb, line 73: It is also kind of strange to me that some of the operations (like sql planning) take a `ctx` parameter, while some of them function on the context itself.

I wonder if you could avoid the strange clone semantics above by simply adding a `IOxExecutionContext` parameter to `collect()`:

So rather than
```rust
    let batches = ctx.collect(physical_plan).await.unwrap();
```

```rust
    let batches = executor.collect(physical_plan, &ctx).await.unwrap();
```

I think that is actually closer to the DataFusion API and it doesn't intermix query execution logic with the context state.

server/src/lib.rs
```diff
@@ -2178,13 +2178,9 @@ mod tests {
...
-        let executor = db.executor();
+        let ctx = db.executor().new_context(ExecutorType::Query);
 
-        let physical_plan = planner.query(db, query, &executor).unwrap();
-
-        executor
-            .collect(physical_plan, ExecutorType::Query)
-            .await
-            .unwrap()
+        let physical_plan = planner.query(db, query, &ctx).unwrap();
```
alamb, line 2183: 👍  for using the same context to plan the query as to `collect` it 👍

alamb (approved):
So I am clicking approve as I think this PR does improve and fix the issue it sets out to.

However,  I think we might be able to get basically the same effect, and avoid the `clone` nonsense if we simply added a `ctx: &IoxExecutionContext` parameter to `Executor::collect()`, and other functions rather than creating such contexts directly. The intermixing of context and function that executes plans seems awkward to me

query/src/exec/context.rs
tustvold: It still needs to be cloneable to allow for task fanout, tokio requires tasks to be 'static, i.e. they can't borrow. and so you have to move the context into the future, which can only be done once without a clone

query/src/exec.rs
```diff
@@ -436,13 +158,16 @@ mod tests {
...
+    use crate::plan::stringset::StringSetPlan;
+    use arrow::record_batch::RecordBatch;
 
     #[tokio::test]
     async fn executor_known_string_set_plan_ok() {
         let expected_strings = to_set(&["Foo", "Bar"]);
         let plan = StringSetPlan::Known(Arc::clone(&expected_strings));
 
-        let executor = Executor::new(1);
+        let executor = Executor::new(1).new_context(ExecutorType::Query);
```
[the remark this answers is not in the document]
tustvold: I think this was just me being lazy - will fix :+1:

query_tests/src/pruning.rs
tustvold: Tbh I think the functions could just be free, the IOxExecutionContext already has a reference to the executor?

tustvold:
I spoke with @alamb about this, we both feel something isn't quite right with this as it conflates the execution context with the actual execution itself. However, as this isn't a change introduced by this PR, collect, execute, etc... were already member functions, we're going to move forward with this PR largely as-is, with some additional docstrings, and potentially look to iterate on the design in subsequent PRs

PR #2379, opened 2021-08-23
fix: shrink RUB-string-RLE keys capacity to fit

crepererum:
We were underestimating the size of a RUB string-RLE column depending on
how the data came into existence. A well-placed debug assert proved
that.


e-dard:
@crepererum and I looked through this carefully and concluded that this is a good change to make but that this issue shouldn't normally get hit in production.

 - in the test suite we often use `to_string` to put stuff into RLE column dictionaries, and `to_string` can over-allocate (minimum string allocation looks to be 8 bytes).
 - in production we use a different code-path that would use `into` to go from `&str` to `String` and `into` does not over-allocate.

Regardless, this is a good change to add.

e-dard (approved):

PR #2380, opened 2021-08-23
refactor: improve string storage and sizing metrics for Read Buffer columns

e-dard:
Following on from https://github.com/influxdata/influxdb_iox/pull/2379

This is a belt and braces type of PR.

This PR updates the capacity-based size metrics to include any excess capacity in allocated strings stored in dictionary encodings. It also ensures strings are minimally sized when using some APIs.

### DIFF
--- a/read_buffer/src/column/encoding/string/dictionary.rs
+++ b/read_buffer/src/column/encoding/string/dictionary.rs
@@ -71,7 +71,10 @@ impl Dictionary {
             .entries
             .iter()
             .map(|k| match k {
-                Some(v) => v.len(),
+                Some(v) =>  match buffers {
+                    true => v.capacity(),
+                    false => v.len(),
+                },
                 None => 0,
             } + size_of::<Option<String>>())
             .sum::<usize>();
--- a/read_buffer/src/column/encoding/string/dictionary.rs
+++ b/read_buffer/src/column/encoding/string/dictionary.rs
@@ -55,9 +55,15 @@ impl Dictionary {
     /// Callers are not required to provide a logical NULL value as part of the
     /// dictionary. The encoding already reserves a representation for that.
     pub fn with_dictionary(dictionary: BTreeSet<String>) -> Self {
-        let mut _self = Self::default();
-        _self.entries.extend(dictionary.into_iter().map(Some));
-        _self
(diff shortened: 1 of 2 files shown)


read_buffer/src/column/encoding/string/dictionary.rs
```diff
@@ -55,9 +55,15 @@ impl Dictionary {
...
     /// dictionary. The encoding already reserves a representation for that.
     pub fn with_dictionary(dictionary: BTreeSet<String>) -> Self {
-        let mut _self = Self::default();
-        _self.entries.extend(dictionary.into_iter().map(Some));
-        _self
+        let mut enc_dict = Self::default();
+
+        enc_dict.entries.reserve_exact(dictionary.len());
+        for mut entry in dictionary {
+            entry.shrink_to_fit(); // ensure entry is minimally sized
```
e-dard, line 62: Similar to what @crepererum did in #2379

read_buffer/src/column/encoding/string/dictionary.rs
```diff
@@ -71,7 +77,10 @@ impl Dictionary {
             .entries
             .iter()
             .map(|k| match k {
-                Some(v) => v.len(),
+                Some(v) =>  match buffers {
```
e-dard, line 80: Since it could be possible for overly sized strings to get into the encoding we should just account for them in the sizing.

read_buffer/src/column/encoding/string/rle.rs
```diff
@@ -91,9 +91,9 @@ impl RLE {
         index_entries_size += self
             .index_entries
             .iter()
-            .map(|k| {
-                debug_assert_eq!(k.len(), k.capacity());
-                k.len()
+            .map(|k| match buffers {
```
e-dard, line 94: @crepererum I figured rather than having the assertion in the sizing method we could just accurately account for it. Then we can (as you have done) try and shrink down strings on the way in.

crepererum:
For other potential reviewers: `shrink_to_fit` is a "free" (aka just a check) if the capacity and the size are already equal. So we don't need to worry about copying strings around too often.

crepererum (approved):

Issue #2381, opened 2021-08-23
panic on "select count(*) from system.columns"


mkmik:
(on query-3000 on our internal staging env)

```
; ./target/debug/influxdb_iox database query 844910ece80be8bc_3c0bd4c89186ca89 'select count(*) from system.columns;'
Error querying: status: InvalidArgument, message: "Error while planning query: Internal executor error while planning query: Error joining execution task: channel closed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 23 Aug 2021 15:02:14 GMT"} }
```

mkmik:
```
hread 'IOx Query Executor Thread' panicked at 'must contain at least one element', /home/rust/project/data_types/src/partition_metadata.rs:73:33
level=error msg="thread 'IOx Query Executor Thread' panicked at 'UNKNOWN', /home/rust/project/data_types/src/partition_metadata.rs:73:33" target="panic_logging" location="panic_logging/src/lib.rs:110" time=1629731020865554395
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   2: core::option::expect_failed
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/option.rs:1243:5
   3: core::option::Option<T>::expect
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/option.rs:351:21
   4: <data_types::partition_metadata::TableSummary as core::iter::traits::collect::FromIterator<data_types::partition_metadata::TableSummary>>::from_iter
             at ./home/rust/project/data_types/src/partition_metadata.rs:73:21
   5: data_types::partition_metadata::PartitionSummary::from_table_summaries
             at ./home/rust/project/data_types/src/partition_metadata.rs:58:20
   6: server::db::catalog::partition::Partition::summary
             at ./home/rust/project/server/src/db/catalog/partition.rs:282:9
   7: server::db::catalog::table::Table::partition_summaries::{{closure}}
             at ./home/rust/project/server/src/db/catalog/table.rs:99:42
   8: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
[...]

PR #2383, opened 2021-08-23
fix: do not error if partition has no chunks

alamb:
Fixes https://github.com/influxdata/influxdb_iox/issues/2381

# Rationale:
Querying `system.columns` on a table with a partition that has been dropped or otherwise stripped of its Chunks results in a `panicked at 'must contain at least one element', /home/rust/project/data_types/src/partition_metadata.rs:73:33`

# Changes
1. Handle partition with no chunks specially


server/src/db/catalog/partition.rs
```diff
@@ -277,14 +277,19 @@ impl Partition {
...
-        PartitionSummary::from_table_summaries(
-            self.addr.partition_key.to_string(),
-            self.chunks
-                .values()
-                .map(|x| x.read().table_summary().as_ref().clone()),
-        )
+    /// Return a PartitionSummary for this partition. If the partition
+    /// has no chunks, returns None.
+    pub fn summary(&self) -> Option<PartitionSummary> {
+        if self.chunks.is_empty() {
```
alamb, line 283: here is the fix -- need to check for no chunks.

I believe @crepererum  said there was a good reason why dropping a partition from the catalog didn't remove the `Partition` object which I can't remember now

NGA-TRAN (approved):
LGTM

server/src/db/catalog/partition.rs
crepererum: Technically because we wanna keep the persistence windows associated w/ that partition.

PR #2384, opened 2021-08-23
refactor: rework exec module

tustvold:
Both I and @alamb were pretty under-whelmed by #2373, this PR aims to achieve a state of whelm. 

Specifically it does the following:

* Reworks Executor and IOxExecutionContext to be immutable
* Alters the SQL query path to no longer mutate the context to set the default catalog
* Adjusts IOxExecutionConfig into a builder like datafusion's ExecutionConfig


server/src/db.rs
```diff
@@ -478,6 +479,16 @@ impl Db {
         Arc::clone(&self.exec)
     }
 
+    /// Returns a new execution context suitable for running queries
```
tustvold, line 482: Perhaps a future PR could remove the member function above that exposes the raw Executor, to encourage consistency, but I didn't want to add to the churn

query/src/exec.rs
```diff
@@ -66,21 +81,14 @@ impl Executor {
...
 
-    /// returns the config of this executor
-    pub fn config(&self) -> &IOxExecutionConfig {
-        &self.config
-    }
-
-    /// returns a mutable reference to this executor's config
-    pub fn config_mut(&mut self) -> &mut IOxExecutionConfig {
-        &mut self.config
+    pub fn new_execution_config(&self, executor_type: ExecutorType) -> IOxExecutionConfig {
```
tustvold, line 84: By exposing this as a separate method it allows callers to override options at creation time, e.g. set the default catalog, etc...

query/src/exec/context.rs
```diff
@@ -145,13 +135,55 @@ impl fmt::Debug for IOxExecutionConfig {
...
+    /// Set the default catalog provider
+    pub fn with_default_catalog(self, catalog: Arc<dyn CatalogProvider>) -> Self {
+        Self {
+            default_catalog: Some(catalog),
+            ..self
+        }
+    }
+
+    /// Create an ExecutionContext suitable for executing DataFusion plans
+    pub fn build(self) -> IOxExecutionContext {
```
alamb, line 163: this is a much nicer pattern. 🏅

query/src/exec/context.rs
```diff
@@ -186,25 +218,11 @@ impl fmt::Debug for IOxExecutionContext {
 }
 
 impl IOxExecutionContext {
```
alamb, line 220: is it worth a docstring addition to `IOxExecutionContext` that says "You create an `IOxExecutionContext` from an executor?"

query/src/frontend/sql.rs
```diff
@@ -17,13 +17,7 @@ impl SqlQueryPlanner {
...
     /// When the plan is executed, it will run on the query executor
     /// in a streaming fashion.
-    pub fn query<D: CatalogProvider + 'static>(
-        &self,
-        database: Arc<D>,
-        query: &str,
-        ctx: &IOxExecutionContext,
-    ) -> Result<Arc<dyn ExecutionPlan>> {
-        ctx.register_catalog(DEFAULT_CATALOG, database);
+    pub fn query(&self, query: &str, ctx: &IOxExecutionContext) -> Result<Arc<dyn ExecutionPlan>> {
```
alamb, line 20: this is much nicer.

The docstring probably needs to be updated to say "plans a SQL query against any tables defined in a `IOxExecutionContext`" rather than "database"

query_tests/src/pruning.rs
```diff
@@ -70,8 +70,8 @@ async fn chunk_pruning_sql() {
     ];
     let query = "select * from cpu where bar < 3.0";
 
-    let ctx = db.executor().new_context(ExecutorType::Query);
-    let physical_plan = SqlQueryPlanner::default().query(db, query, &ctx).unwrap();
+    let ctx = db.new_query_context();
```
alamb, line 73: ❤️

server/src/db.rs
alamb: I like this API, for what it is worth

alamb (approved):
Thank you @tustvold . I am *definitely* whelmed by this PR 👍  

image

I think this API is much easier to understand the relationship between config and contexts

Issue #2367, opened 2021-08-20
Support custom tracing header names via JAEGER_TRACE_CONTEXT_HEADER_NAME and JAEGER_DEBUG_HEADER


alamb:
__Use case:__
Sometimes environments (like InfluxData's internal clouds) use their own custom tracing header names

__Proposal:__
In addition to supporting B3 and jaeger headers, IOx should also support headers defined by configuration/environment variable. Specifically it should support environment variables named thusly

```
JAEGER_TRACE_CONTEXT_HEADER_NAME
JAEGER_DEBUG_HEADER
```

__Current behaviour:__
IOx can only interpret  "standard" `X-B3-*` or jaeger (`uber-trace-id`) headers -- see https://github.com/influxdata/influxdb_iox/blob/main/trace/src/ctx.rs#L18-L24

__Desired behaviour:__
The names of headers that contain the trace context should be configurable via environment variable

__Alternatives considered:__
Describe other solutions or features you considered.

cc @BondAnthony

alamb:
per @BondAnthony Note: `jarger-debug-id` is a forced tracing header engineers can use to force a trace. https://www.jaegertracing.io/docs/1.16/client-features/

PR #2388, opened 2021-08-24
refactor: split out trace http

tustvold:
This PR splits the HTTP and tower goop out of the primary trace crate, into trace_http. This allows crates, such as query, to depend on the `trace` functionality without bringing in a load of other dependencies.

I also removed the serde_json support as this was actually a workaround from when `Span` had an associated lifetime, it no longer does and so RingBufferTraceCollector can just store spans directly


trace/src/ctx.rs
```diff
@@ -1,68 +1,15 @@
 use std::borrow::Cow;
-use std::num::{NonZeroU128, NonZeroU64, ParseIntError};
```
tustvold, line 2: Most of this logic is moved into trace_http

trace/src/span.rs
```diff
@@ -197,43 +179,15 @@ mod tests {
     fn test_span() {
         let collector = Arc::new(RingBufferTraceCollector::new(5));
 
-        let mut span = make_span(Arc::<RingBufferTraceCollector>::clone(&collector));
```
tustvold, line 200: This was testing JSON formatting which is no longer functionality provided, it was largely a hack to workaround when Span was not `'static`

trace_http/src/ctx.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+fn parse_span(s: &str) -> Result<SpanId, DecodeError> {
+    Ok(SpanId(
+        NonZeroU64::new(u64::from_str_radix(s, 16)?).ok_or(DecodeError::ZeroError)?,
+    ))
+}
+
+/// Create a SpanContext for the trace described in the request's headers
+pub fn parse_span_ctx(
```
tustvold, line 72: This may become a member function as part of #2367 but that is a separate PR

trace_http/src/ctx.rs
```diff
@@ -0,0 +1,369 @@
+use std::num::{NonZeroU128, NonZeroU64, ParseIntError};
```
tustvold, line 1: This code is pretty much verbatim, but can no longer use FromStr due to orphan rules.

trace/src/ctx.rs
alamb: ❤️

trace/src/span.rs
```diff
@@ -3,13 +3,10 @@ use std::collections::HashMap;
...
 
 use chrono::{DateTime, Utc};
-use serde::{Deserialize, Serialize};
-
-use observability_deps::tracing::error;
 
 use crate::ctx::SpanContext;
 
-#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
+#[derive(Debug, Copy, Clone)]
```
alamb, line 9: the change here is to remove the (non trivial) `serde` dependency right? So that everything that needs `trace!` macro doesn't need serde as well?

alamb (approved):
Looks good to me 👍

trace/src/span.rs
tustvold: > So that everything that needs trace! macro doesn't need serde as well?

There isn't a `trace!` macro, (well there is but that's logging/tokio-tracing), but in terms of reducing the dependencies necessary to create a span, yes.

I also removed it as it was only ever there as a hack, which is no longer needed. We don't really want to expose/define a custom span serialization format if we can possibly avoid it. There are enough standards as it is :laughing:

tustvold:
Accidentally pushed another branch on top of this one :sweat_smile:

PR #2390, opened 2021-08-24
feat: wait for Executor shutdown

### LINKED ISSUE #2372: Wait for Executor shutdown
Query tasks are run on background threadpools, however, these are never actually shutdown. Whilst likely harmless, this should be fixed.

crepererum:
Closes #2372.

### DIFF
--- a/query/src/exec.rs
+++ b/query/src/exec.rs
@@ -81,13 +81,17 @@ impl Executor {
         }
     }
 
-    /// Return a new execution config, suitable for executing a new query or system task
+    /// Return a new execution config, suitable for executing a new query or system task.
+    ///
+    /// Note that this context (and all its clones) will be shut down once `Executor` is dropped.
     pub fn new_execution_config(&self, executor_type: ExecutorType) -> IOxExecutionConfig {
         let exec = self.executor(executor_type).clone();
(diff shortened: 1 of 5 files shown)


query/src/exec.rs
```diff
@@ -176,7 +197,8 @@ mod tests {
         let expected_strings = to_set(&["Foo", "Bar"]);
         let plan = StringSetPlan::Known(Arc::clone(&expected_strings));
 
-        let ctx = Executor::new(1).new_context(ExecutorType::Query);
+        let exec = Executor::new(1);
```
alamb, line 200: this change (to name `exec`) is to ensure that `exec` lives to the end of this block, right?

src/influxdb_ioxd.rs
```diff
@@ -341,6 +341,9 @@ async fn serve(
 
     info!("server completed shutting down");
 
+    application.join();
+    info!("application completed shutting down");
```
alamb, line 345:
```suggestion
    info!("shared application state completed shutting down");
```

alamb (approved):
Looks great -- thank you @crepererum

query/src/exec.rs
crepererum: yes.

PR #2391, opened 2021-08-24
 feat: query tracing (#2273) 

tustvold:
This PR plumbs distributed tracing into the query path, in particular creating spans for the major parts of query execution (e.g. plan, execution, etc...). In future we can hopefully extend this with further spans generated from the DataFusion metrics @alamb is adding


src/influxdb_ioxd/http.rs
```diff
@@ -648,8 +648,8 @@ async fn query<M: ConnectionManager + Send + Sync + Debug + 'static>(
 
     let db = server.db(&db_name)?;
 
-    let ctx = db.new_query_context();
-    let physical_plan = Planner::new(ctx.clone()).sql(&q).await.context(Planning)?;
+    let ctx = db.new_query_context(req.extensions().get().cloned());
```
tustvold, line 651: This is how the SpanContext that is added by the tower layer, is pulled out of the Request.

query/src/exec/context.rs
```diff
@@ -201,7 +214,6 @@ impl IOxExecutionConfig {
 ///
 /// An IOxExecutionContext is created directly from an Executor, or from
 /// an IOxExecutionConfig created by an Executor
-#[derive(Clone)]
```
tustvold, line 204: Here today, gone tomorrow :laughing: 

This was recently added by #2373 and is now replaced by `IOxExecutionContext::create_child`

trace/src/span.rs
```diff
@@ -99,52 +98,67 @@ impl From<i64> for MetaValue {
     }
 }
 
-/// Updates the start and end times on the provided Span
-#[derive(Debug)]
-pub struct EnteredSpan {
-    /// Option so we can take out of it on drop / publish
+/// Updates the start and end times on the provided Span if any
+#[derive(Debug, Default)]
+pub struct SpanRecorder {
```
tustvold, line 103: I opted to rework this as plumbing `Option<EnteredSpan>` all over the place was decidedly meh.

query/Cargo.toml
```diff
@@ -31,6 +31,7 @@ regex = "1"
 snafu = "0.6.3"
 tokio = { version = "1.0", features = ["macros"] }
 tokio-stream = "0.1.2"
+trace = { path = "../trace" }
 tracker = { path = "../tracker" }
 observability_deps = { path = "../observability_deps" }
```
alamb, line 36: is the "longer" term plan to remove `observability_deps` as a dependency?

query/src/exec/context.rs
```diff
@@ -161,6 +166,11 @@ impl IOxExecutionConfig {
         }
     }
 
+    /// Set the span context from which to create spans for this query
+    pub fn with_span_context(self, span_ctx: Option<SpanContext>) -> Self {
+        Self { span_ctx, ..self }
```
alamb, line 171: this is a cool syntax. I will be stashing that away for the future

query/src/exec/context.rs
```diff
@@ -451,4 +473,13 @@ impl IOxExecutionContext {
             .await
             .unwrap_or_else(|e| Err(Error::Execution(format!("Join Error: {}", e))))
     }
+
+    /// Returns a IOxExecutionContext with a SpanRecorder that is a child of the current
+    pub fn child_ctx(&self, name: &'static str) -> Self {
```
alamb, line 478: What would you think about calling this `child_span_ctx` to emphasize the fact the "child" relationship is related to tracing / spans rather than some other parent/child relationship?

query/src/exec/context.rs
```diff
@@ -161,6 +166,11 @@ impl IOxExecutionConfig {
         }
     }
 
+    /// Set the span context from which to create spans for this query
```
alamb, line 169:
```suggestion
    /// Set the span context from which to create distributed tracing spans for this query
```

query/src/exec/context.rs
```diff
@@ -182,9 +192,12 @@ impl IOxExecutionConfig {
             inner.register_catalog(DEFAULT_CATALOG, default_catalog);
         }
 
+        let maybe_span = self.span_ctx.map(|ctx| ctx.child("IOxExecutionContext"));
```
alamb, line 195: The string `"IOxExecutionContext"` will appear in distributed traces now, right? Should we call this something more human understandable like "Query Execution"?

src/influxdb_ioxd.rs
```diff
@@ -643,32 +655,67 @@ mod tests {
...
+                span.ctx.parent_span_id == Some(parent.ctx.span_id) && span.name == name
+            })
+        };
+
+        let ctx_span = child(query_span, "IOxExecutionContext").unwrap();
+
+        let planner_span = child(ctx_span, "Planner").unwrap();
+        let sql_span = child(planner_span, "sql").unwrap();
+        let prepare_sql_span = child(sql_span, "prepare_sql").unwrap();
+        child(prepare_sql_span, "prepare_plan").unwrap();
```
alamb, line 708: This is very cool

src/influxdb_ioxd/http.rs
alamb: The way it all fits together is a thing of beauty

trace/src/span.rs
alamb: Yes, this is much nicer

alamb (approved):
looks great to me

query/Cargo.toml
tustvold: I think we might rename it to something like `logging` but we are still using tokio-tracing, just for logging and not tracing

query/src/exec/context.rs
tustvold: I think I would expect a method called child_span_ctx to return a SpanContext... I think it is fair to call it a child context, it just happens that this only means anything for spans at the moment, although this I could see other things hooking into the child-concept in future - e.g. cancellation

query/Cargo.toml
alamb: My mind still ends up in a small knot everytime I hear `tokio-tracing` is for logging 🤯

PR #2400, opened 2021-08-25
feat: make `RLE` a bit smaller by capacity-based allocation

crepererum:
For some demo data this reduced the overall chunk size from

195049367 bytes
to
191088095 bytes


read_buffer/src/column/encoding/string/rle.rs
```diff
@@ -60,7 +60,13 @@ impl RLE {
...
     /// ordinal relationship will exist between the encoded values.
     pub fn with_dictionary(dictionary: BTreeSet<String>) -> Self {
-        let mut _self = Self::default();
+        let mut index_entries = Vec::with_capacity(dictionary.len() + 1);
+        index_entries.push(String::from(""));
+
+        let mut _self = Self {
+            index_entries,
+            ..Self::default()
+        };
```
e-dard, line 69: This is a little more complicated than necessary. Here is an alternative:

```suggestion
        let mut _self = Self::default();
        _self.index_entries.reserve_exact(dictionary.len());
```

e-dard (approved):
Just the small simplification otherwise LGTM

PR #2404, opened 2021-08-25
feat: add storage client to influxdb_iox_client

tustvold:
As part of adding query tracing (#2273) I want to be able to write an integration test of the tracing functionality for queries made to the storage RPC API.

Using the storage API directly, however, is non-trivial. This PR therefore promotes the StorageClientWrapper found within the storage RPC tests into a "standard" client in `influxdb_iox_client`. This is pretty much just moved, but with some minor modifications to make it more inline with the other clients in the crate.


generated_types/src/lib.rs
```diff
@@ -119,6 +119,7 @@ pub fn protobuf_type_url_eq(url: &str, protobuf_type: &str) -> bool {
     }
 }
 
+// TODO: Remove these
```
tustvold, line 122: This confuses the interface of the module as you have generated_types::Aggregate, etc... at the root level, this dates from when generated_types solely contained the storage API and has been kept for backwards compat. I will remove this in a subsequent PR.

influxdb_iox_client/src/client/storage.rs
```diff
@@ -0,0 +1,313 @@
...
+            .map(tag_key_bytes_to_strings)
+            .collect::<Vec<_>>();
+
+        strings.sort();
+
+        strings
+    }
+}
+
+/// Converts bytes representing tag_keys values to Rust strings,
```
tustvold, line 303: I _think_ this is a method that is part of the public storage API, and not just some test helper and therefore belongs here, but I'm not actually sure...

influxdb_iox_client/src/client/storage.rs
```diff
@@ -0,0 +1,314 @@
...
+pub struct OrgAndBucket {
+    org_id: NonZeroU64,
+    bucket_id: NonZeroU64,
+    db_name: String,
+}
+
+impl OrgAndBucket {
+    /// Create a new `OrgAndBucket` from the provided `org_id` and `bucket_id`
+    pub fn new(org_id: NonZeroU64, bucket_id: NonZeroU64) -> Self {
+        let db_name = format!("{:016x}_{:016x}", org_id, bucket_id);
```
tustvold, line 31: This does mean we have the notion of how to convert org_id + bucket_id to database in two places, but I think this is probably fine for now.

influxdb_iox_client/src/client/storage.rs
e-dard: I'm not quite sure what your defintion of "part of the public storage API" is but there isn't a dedicated endpoint for this conversion. However the implementation of how measurement names and field keys are represented differs between TSM storage (where they're stored as bytes `0x00` and `0xff` and the query tier where they're represented as `_m` and `_f` or `_measurement` and `_field` (I think depending on where in the stack you are).

Anyway, we don't store the explicit tag keys for measurements and fields in IOx (we store the tag values for those things as column names) so I guess that this method is here to satisfy some end to end testing of some sort? Maybe @alamb knows.

e-dard (approved):
Seems fine to me if it will help with testing.

tustvold:
I've split the storage client into its own crate, this necessitated splitting the connection logic into a separate crate so that it could shared this. PTAL

client_util/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+//! An InfluxDB IOx API client.
```
alamb, line 1:
```suggestion
//! Shared InfluxDB IOx API client functionality.
```

generated_types/src/lib.rs
alamb: if you are going to leave the `//TODO` in the code I suggest you also add the rationale (copy this comment from the PR as well) so that future readers who encounter it will have some context

influxdb_storage_client/src/lib.rs
```diff
@@ -0,0 +1,332 @@
...
+            .try_collect()
+            .await?;
+
+        let data_frames_count = responses
+            .into_iter()
+            .flat_map(|r| r.frames)
+            .flat_map(|f| f.data)
+            .count();
+
+        let s = format!("{} group frames", data_frames_count);
```
alamb, line 276: and here

alamb (approved):

PR #2405, opened 2021-08-25
refactor: use sorted vector for RLE dictionary entry collection

### LINKED ISSUE #2402: Reduce size of RLE encoding by swapping out BTreeMap
Right now the RLE encoding using a `BTreeMap` to maintain an ordered collection of encoded physical values to the logical string values for the column.

The ordering is vital so that we can work directly on the encoded data without having to materialise string values. However, since the RLE encoding only supports constuction via an ordered set of distinct values (e.g., an ordered set) then the `BTreeMap` doesn't need to do any ordering for us. Instead we can switch it out for a `Vec<RowIDs>`, which will probably save us a little memory.

Thanks to @crepererum for bringing this to my attention 👍

e-dard:
Closes #2402

This PR swaps out the `BTreeMap` used for storing an ordered mapping of all the row IDS belonging to each encoded value in an RLE encoded column. Rather than using a `BTreeMap` it uses a vector, which allows us to potentially size the memory usage more accurately (because there are internals to a `BTreeMap` which can't easily be sized).

The change doesn't really impact performance according to a run of benchmarks (I included them in a commit).

Finally, during this refactor I realised I had not accounted for an `enum` container previously. I suspect this means that RUB chunks would have been undereporting about `32b` per RLE entry. The error would have accumulated more for higher cardinality columns than lower cardinality ones (generally).


read_buffer/src/column.rs
```diff
@@ -379,7 +379,7 @@ impl Column {
         }
     }
 
-    pub fn grouped_row_ids(&self) -> Either<Vec<&RowIDs>, Vec<RowIDs>> {
+    pub fn grouped_row_ids(&self) -> Either<&Vec<RowIDs>, Vec<RowIDs>> {
```
e-dard, line 382: FYI - I am going to try and tidy up (remove) this `Either` in a follow on PR.

crepererum (approved):
Thanks!

read_buffer/src/column/encoding/string/rle.rs
```diff
@@ -129,12 +122,12 @@ impl RLE {
     ///
     pub fn size_raw(&self, include_nulls: bool) -> usize {
         let mut total_size = 0;
-        for (idx, rows) in &self.index_row_ids {
+        for (idx, rows) in self.index_row_ids.iter().enumerate() {
             // the length of the string value x number of times it appears
             // in the column.
-            total_size += self.index_entries[*idx as usize].len() * rows.len();
+            total_size += self.index_entries[idx as usize].len() * rows.len();
```
alamb, line 128: Should this function also account the heap used by `rows` itself? Something like  `total_size += rows.capacity()` as well? 

I couldn't find where that memory would be accounted for

PR #2407, opened 2021-08-25
feat: propagate span context into storage RPC queries

tustvold:
_Draft as builds on top of #2404_ 

This adds the plumbing to trace storage RPC requests


src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -2,7 +2,7 @@
 //! implemented in terms of the [`QueryDatabase`](query::QueryDatabase) and
 //! [`DatabaseStore`]
 
-use std::{collections::HashMap, sync::Arc};
+use std::collections::HashMap;
```
tustvold, line 5: I think there might be ways to reduce the duplication within this module, but something for a future PR

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -857,9 +881,10 @@ fn get_database_name(input: &impl GrpcInputs) -> Result<DatabaseName<'static>, S
 /// Gathers all measurement names that have data in the specified
 /// (optional) range
 async fn measurement_name_impl<T>(
-    db_store: Arc<T>,
+    db_store: &T,
```
tustvold, line 884: I took the opportunity to remove the Arc from this as it wasn't actually being used

query/src/lib.rs
```diff
@@ -71,6 +72,9 @@ pub trait QueryDatabase: Debug + Send + Sync {
 
     /// Return a summary of all chunks in this database, in all partitions
     fn chunk_summaries(&self) -> Result<Vec<ChunkSummary>, Self::Error>;
+
+    /// Returns a new execution context suitable for running queries
+    fn new_query_context(self: &Arc<Self>, span_ctx: Option<SpanContext>) -> IOxExecutionContext;
```
alamb, line 77: it seems like an alternate is to add a `IoxExecutionContext` to each call to the gRPC planner, as is done for the `SQLPlanner`:

https://github.com/influxdata/influxdb_iox/blob/main/query/src/frontend/sql.rs#L17

I would stylistically prefer to avoid adding the notion of a `Span` into the `QueryDatabase` trait API, though I have no actual rational reason for objecting to it and I am fine with adding it.

server/src/db.rs
```diff
@@ -1403,6 +1400,10 @@ impl QueryDatabase for Db {
     fn table_schema(&self, table_name: &str) -> Option<Arc<Schema>> {
         self.catalog_access.table_schema(table_name)
     }
+
+    fn new_query_context(self: &Arc<Self>, span_ctx: Option<SpanContext>) -> IOxExecutionContext {
```
alamb, line 1404: Is there some reason to implement `new_query_context` outside of the trait impl as well? (aka why not just only implement `QueryDatabase::new_query_context`)?

server/src/db/access.rs
```diff
@@ -68,6 +71,9 @@ impl AccessMetrics {
 /// (and DataFusion) to access the contents of the IOx catalog.
 #[derive(Debug)]
 pub(crate) struct QueryCatalogAccess {
+    /// Executor for running queries
+    exec: Arc<Executor>,
```
alamb, line 75: this seems strange to me -- that catalog access has a back pointer to the executor. I wonder if this introduces a ref cycle (so that something points at an executor which then points at the catalog access?) It is probably fine but I feel like the executor doesn't really belong on something called "catalog access" as it doesn't have to do with accessing the catalog

alamb (approved):
I reviewed https://github.com/influxdata/influxdb_iox/pull/2407/commits/3f26d798f27af21ba1df7317dabcb030a0834857 (the commit that seems to be the meat of this PR)

Looks good - I had some suggestions for how some of the code got moved around that I don't think are required for correctness and I don't have any objection to merging this PR in as is

PR #2424, opened 2021-08-26
feat: Add an iox_object_store API for writing the tombstone file

### LINKED ISSUE #1871: Implement `delete` database management API

carols10cents:
This PR adds a management gRPC API to delete a database, which creates a tombstone file. The code that looks for active generations now checks to see if the tombstone file exists in a generation directory, and if it does exist, that generation is marked as deleted.

I feel like I'm probably doing some snafu things wrong; I'd like to have @shepmaster take a look at it.

Closes #1871.

### DIFF
--- a/server/src/database.rs
+++ b/server/src/database.rs
@@ -72,6 +72,15 @@ pub enum Error {
 
     #[snafu(display("cannot persisted updated rules: {}", source))]
     CannotPersistUpdatedRules { source: crate::rules::Error },
+
+    #[snafu(display("cannot mark database deleted: {}", source))]
+    CannotMarkDatabaseDeleted {
+        db_name: String,
+        source: object_store::Error,
+    },
+
+    #[snafu(display("no active database named {} to delete", db_name))]
+    NoActiveDatabaseToDelete { db_name: String },
 }
 
 /// A `Database` represents a single configured IOx database - i.e. an
@@ -134,9 +143,15 @@ impl Database {
     ) -> Result<(), InitError> {
         let db_name = provided_rules.db_name();
         let iox_object_store = Arc::new(
-            IoxObjectStore::new(Arc::clone(application.object_store()), server_id, db_name)
+            match IoxObjectStore::new(Arc::clone(application.object_store()), server_id, db_name)
                 .await
-                .context(IoxObjectStoreError)?,
+            {
+                Ok(ios) => ios,
+                Err(iox_object_store::IoxObjectStoreError::DatabaseAlreadyExists { name }) => {
+                    return Err(InitError::DatabaseAlreadyExists { name })
+                }
+                Err(source) => return Err(InitError::IoxObjectStoreError { source }),
+            },
         );
 
         provided_rules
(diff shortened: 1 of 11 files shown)


server/src/database.rs
```diff
@@ -147,6 +162,50 @@ impl Database {
...
+            state.try_freeze().context(TransitionInProgress {
+                db_name,
+                state: state.state_code(),
+            })?
+        };
+
+        // If there is an object store for this database, write out a tombstone file.
+        // If there isn't an object store, something is wrong and we shouldn't switch the
+        // state without being able to write the tombstone file.
+        self.iox_object_store()
```
alamb, line 184: BTW this is a good pattern where if the write to object store fails, no changes are made to the `DatabaseState` (which is correct in my opinion)

server/src/database.rs
```diff
@@ -147,6 +162,51 @@ impl Database {
         Ok(())
     }
 
+    /// Mark this database as deleted.
+    pub async fn delete(&self) -> Result<(), Error> {
+        let db_name = &self.shared.config.name;
+
+        let handle = {
```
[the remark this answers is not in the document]
alamb: after the work on https://github.com/influxdata/influxdb_iox/pull/2430 with @tustvold  I am more confident that this sequence is correct

tests/end_to_end_cases/management_api.rs
```diff
@@ -270,6 +270,16 @@ async fn test_create_get_update_database() {
...
+
+    client
+        .delete_database(&db_name)
+        .await
+        .expect("delete database failed");
+
+    client
+        .get_database(&db_name)
+        .await
+        .expect_err("get database should have failed but didn't");
```
crepererum, line 282: you should check the error here since the error propagation through gRPC was often broken and ended up in some "internal error". It would basically act as a test for your changes in `src/influxdb_ioxd/rpc/error.rs` or as a safety net for anyone who ever wants to modify the error handling. It would also test the client side where I think (!) that one of the two error cases is missing or doesn't map to nicely.

server/src/database.rs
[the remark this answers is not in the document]
crepererum: I'm certainly not the expert when it comes to this construct, but to me it looks solid.

crepererum (approved):
I only have a tiny nitpick regarding the end2end test and how it covers the non-happy path. Rest looks solid :)

NGA-TRAN (approved):

tests/end_to_end_cases/management_api.rs
carols10cents: You were totally right, the wrong error was being returned :) Good call!!

PR #2430, opened 2021-08-27
feat: Store only the database rules sent by the client (do not store default values)

### LINKED ISSUE #2053: Can modify database rules without persisting
It is currently possible to mutate a database's DatabaseRules without persisting the changes. This is because Server::update_db_rules is what handles persistence, but the DatabaseRules themselves are actually stored on Db.

#2052 started work to address this, by pushing the modification of the rules into Db, but it left the persistence logic on Server to avoid code churn. This should be addressed.

### LINKED ISSUE #2409: Store only the database rules sent by the client (do not store default values)
__Use case:__
As described on an internal ticket https://github.com/influxdata/conductor/issues/468, the current behavior of IOx on database creation or a rules update is to fill in any missing values with their (current) defaults and save that configuration file to object store.

Not only is this behavior confusing, it makes the logic to keep configurations synchronized across multiple IOx nodes more challenging. 

__Proposal:__
1. IOx will store in object store exactly the configuration supplied by the client, 
2. The API to get database rules will have two modes:
Mode 1: Return the current configuration that is being used by the server, with all default values filled
Mode 2: Return the persisted configuration (aka only what was supplied when the database was created or last modified)

__Current behaviour:__
Given this chain of events:
1. Send a configuration to a server
2. Update the server software
3. Send the *exact same* configuration to the server running the new software

The effective configuration (the values used by the server's operation) can change (if the software has new default values). 

__Desired behaviour:__
If a  user sends the same config to a server, updates the server software and then send the same config, the server's effective configuration (the values it uses at runtime) will be the same.

__Alternatives considered:__
1. Require the user to send the entire configuration and if any fields are missing reject the request. This would likely require some additional "upgrade and persist upgraded configuration" type logic
2. Do nothing -- complicates conductor and any other software that is trying to manage the configuration of IOx, especially when trying to keep the configurations in sync

alamb:
Resolves #2409 
Resolves #2053

# Rationale:
As described on an internal ticket https://github.com/influxdata/conductor/issues/468, the current behavior of IOx on database creation or a rules update is to fill in any missing values with their (current) defaults and save that configuration file to object store.

Not only is this behavior confusing, it makes the logic to keep configurations synchronized across multiple IOx nodes more challenging.

# Changes:
1. the database_rules are store in object store exactly the configuration supplied by the client,
2. New argument added to get_database: `omit_defaults` that defaults to `false` (aka the current behavior is the default)

# Behavior Change
This will mean that the effective configuration (the values used during the database's operation) can change (if the software has new default values for fields that were not specified in the origina file).

# Design:
The goal was to keep the notion of "what user provided" as isolated as possible and thus there is no change to `internal_types::database_rules::DatabaseRules`. 

The `Database` (not to be confused with `Db`, the in memory thing) now has a ProvidedDatabaseRules which has both the original encoded version of the `management::DatabaseRules` used to create the database, as well as a `internal_types::DatabaseRules` struct.  `db` and lower modules are not modified as suggested by @tustvold 

## Example use:

Create a new database:
```shell
./target/debug/influxdb_iox database create my_db
```

Edit the rules:
```shell
./scripts/edit_db_rules localhost:8082 my_db
```

The change output so it only specifies part of the rules:

```json
{
  "rules": {
    "name": "my_db",
    "workerCleanupAvgSleep": "500s"
  }
}

```

Now you can see both what was sent (note the difference in display is due to the fact that the IOx CLI prints out tonic structs differently than grpcur, which is what `edit_db_rules` uses).

```shell
./target/debug/influxdb_iox database get --omit-defaults my_db
{
  "name": "my_db",
  "partitionTemplate": null,
  "lifecycleRules": null,
  "workerCleanupAvgSleep": {
    "seconds": 500,
    "nanos": 0
  },
  "routingRules": null,
  "writeBufferConnection": null
}
```

And what is being used (no `--omit-defaults` flag):
```shell
./target/debug/influxdb_iox database get my_db
{
  "name": "my_db",
  "partitionTemplate": {
    "parts": []
  },
  "lifecycleRules": {
    "bufferSizeSoft": 0,
    "bufferSizeHard": 0,
    "persist": false,
    "immutable": false,
    "workerBackoffMillis": 1000,
    "catalogTransactionsUntilCheckpoint": 100,
    "lateArriveWindowSeconds": 300,
    "persistRowThreshold": 1000000,
    "persistAgeThresholdSeconds": 1800,
    "mubRowThreshold": 100000,
    "parquetCacheLimit": 0,
    "maxActiveCompactionsCfg": {
      "maxActiveCompactionsCpuFraction": 1.0
    }
  },
  "workerCleanupAvgSleep": {
    "seconds": 500,
    "nanos": 0
  },
  "routingRules": null,
  "writeBufferConnection": null
}
```


server/src/database.rs
```diff
@@ -179,8 +189,77 @@ impl Database {
...
+            );
+
+            let handle = state.try_freeze().context(RulesNotUpdateable {
+                db_name: new_provided_rules.db_name(),
+                state: state_code,
+            })?;
+            (handle, iox_object_store)
+        };
+
+        let shared = Arc::clone(&self.shared);
```
tustvold, line 232: I'm not sure I follow why this Arc::clone is necessary?

server/src/database.rs
```diff
@@ -179,8 +189,77 @@ impl Database {
...
+
+        // Attempt to persist to object store, if that fails, roll
```
tustvold, line 234: I like that this tries the fallible part first, so rollback is simply bailing out and dropping the freeze handle :+1:

server/src/database.rs
```diff
@@ -179,8 +189,77 @@ impl Database {
...
+        let mut state = state.unfreeze(handle);
+
+        if let DatabaseState::Initialized(DatabaseStateInitialized { db, provided_rules }) =
+            state.deref_mut()
+        {
+            db.update_rules(Arc::clone(new_provided_rules.rules()));
+            *provided_rules = new_provided_rules;
+            Ok(())
+        } else {
+            // The fact we have a handle should prevent any changes to the database state
```
tustvold, line 260: :+1:

tustvold (approved):
LGTM

server/src/database.rs
```diff
@@ -179,8 +189,77 @@ impl Database {
...
+        new_provided_rules: Arc<ProvidedDatabaseRules>,
+    ) -> Result<(), Error> {
+        // get a handle to signal our intention to update the state
+        let (handle, iox_object_store) = {
+            // scope so we drop the read lock
+            let state = self.shared.state.read();
+            let state_code = state.state_code();
+
+            // a handle to the object store  so we can update the rules, so we don't try hold the mut writer
+            // across a thread. While this is a
```
tustvold, line 208: Incomplete comment?

server/src/database.rs
```diff
@@ -179,8 +189,77 @@ impl Database {
...
+            // we get the lock
+            ensure!(
+                state_code == DatabaseStateCode::Initialized,
+                RulesNotUpdateable {
+                    db_name: new_provided_rules.db_name(),
+                    state: state_code,
+                }
+            );
+
+            let handle = state.try_freeze().context(RulesNotUpdateable {
```
tustvold, line 225: FWIW you can avoid this error by calling `freeze()` which returns a future you can await (once you've dropped the locks). In this case an error is perfectly reasonable, but thought I'd mention it isn't the only option, you can wait for the lock. The freezable.rs module doc has an example of this

server/src/database.rs
alamb: I think it is left over from a previous implementation -- I will try and remove it

server/src/database.rs
alamb: Got it

server/src/database.rs
alamb: fixed

server/src/database.rs
alamb: I plan to merge this PR in and then create a follow on one with better semantics (aka don't error)

alamb: follow on https://github.com/influxdata/influxdb_iox/pull/2433/files

PR #2433, opened 2021-08-28
feat: Less errors on update_database_rules

alamb:
# Rationale
As coded, #2430 has the potentially annoying behavior that if two clients attempt to update the database rules at ~ the same time, one of them will get an error about database rules can not be updated.

@tustvold  suggests an alternate approach on https://github.com/influxdata/influxdb_iox/pull/2430#discussion_r697845991 which would allow the requests to queue up and execute in some order

# Changes
1. `await` on the freeze handle (exclusive intent to write) rather than error


server/src/database.rs
```diff
@@ -223,20 +223,20 @@ impl Database {
                 }
             );
 
-            let handle = state.try_freeze().context(RulesNotUpdateable {
-                db_name: new_provided_rules.db_name(),
-                state: state_code,
-            })?;
+            let handle = state.freeze();
```
alamb, line 226: this is the change suggested. However, since the lock has to be dropped to `await` on this, handle, it seems like nothing prevents the state from changing to a value other than `DatabaseStateCode::Initialized` prior to writing to the object store (the lock is only reacquired after that write, and the lock can't be held across the write)

Thus I am not convinced this is a correct change

tustvold: It isn't, you will need to do something like

```
let handle = self.shared.state.read().freeze();
handle.await;

let iox_object_store = {
    let state = self.shared.state.read();
    ....
}

new_provided_rules.persist(...)
```

server/src/database.rs
```diff
@@ -199,54 +199,54 @@ impl Database {
...
             );
-
-            let handle = state.try_freeze().context(RulesNotUpdateable {
-                db_name: new_provided_rules.db_name(),
-                state: state_code,
-            })?;
-            (handle, iox_object_store)
-        };
+            iox_object_store
+        }; // drop read lok
```
tustvold, line 234:
```suggestion
        }; // drop read lock
```

server/src/database.rs
```diff
@@ -199,54 +199,54 @@ impl Database {
         new_provided_rules: Arc<ProvidedDatabaseRules>,
     ) -> Result<(), Error> {
         // get a handle to signal our intention to update the state
-        let (handle, iox_object_store) = {
-            // scope so we drop the read lock
+        let handle = self.shared.state.read().freeze();
+
+        // wait for the freeze handle. Only one thread can ever have
+        // it at any time so we know past this point no other thread
+        // chan change the DatabaseState (even though this code
```
tustvold, line 206:
```suggestion
        // can change the DatabaseState (even though this code
```

server/src/database.rs
```diff
@@ -199,54 +199,54 @@ impl Database {
...
+        }; // drop read lok
 
         // Attempt to persist to object store, if that fails, roll
         // back the whole transaction (leave the rules unchanged).
         //
-        // The fact the freeze handle is held here ensures only one
-        // one task can ever be in this code at any time. Another
-        // concurrent attempt would error in the `try_freeze()`
-        // call above.
+        // Even though we don't hold a lock here, the freeze handle is
```
tustvold, line 239:
```suggestion
        // Even though we don't hold a lock here, the freeze handle
```

tustvold (approved):
LGTM :+1:

carols10cents:
I just rebased to pick up #2442

Issue #2416, opened 2021-08-26
Metrics for Operations


crepererum:
__Use case:__
Being able to see operations over time and reason about their effects on other metrics (like memory/CPU usage, chunk creations etc.).

__Proposal:__
Expose most of the data that we currently have within this system table as metrics as well:

https://github.com/influxdata/influxdb_iox/blob/c81a65040904050156187ca768005fce895d7ac4/server/src/db/system_tables/operations.rs#L46-L63

__Current behaviour:__
Users have to use the system table, but that's somewhat messy because you cannot easily fold that into a dashboard + the table is gone after restarts.

mkmik:
Poor man's approach:

```
./target/debug/influxdb_iox operation list | jq '.[]|select(.job.PersistChunks)|.wall_time.secs' | datamash perc:50 1 perc:90 1 perc:95 1 perc:97 1 perc:99 1 perc:100 1
0	106.8	112.7	125.4	142.18	150
```

PR #2435, opened 2021-08-28
feat: migrate jemalloc metrics to `metric` crate

tustvold:
This migrates the jemalloc metric reporting to be through the new `metric` crate. This showcases how to use the raw Instrument API for pull-based metrics, as opposed to instrumenting code with `Metric<T>`. I expect this to be a niche use-case, the only other example I can think of is possibly the tracker registry (#2416), but it is something that we need to support.


server/Cargo.toml
```diff
@@ -47,7 +47,6 @@ serde = "1.0"
 serde_json = "1.0"
 snafu = "0.6"
 snap = "1.0.0"
-tikv-jemalloc-ctl = "0.4.0"
```
tustvold, line 50: There is no particular reason that I can see for this reporting logic to live in the server crate, and so I pulled it out into the main binary crate with the rest of the custom allocator code

crepererum (approved):

PR #2443, opened 2021-08-30
test: use random free port for end2end tests

### LINKED ISSUE #786: chore: use random port for test server
See https://github.com/influxdata/influxdb_iox/issues/745#issuecomment-774092141

### LINKED ISSUE #2403: Flaky Test: Server ID Conflict
When running tests locally on current main, I'm consistently seeing these two tests failing

```
failures:
    end_to_end_cases::management_api::test_wipe_preserved_catalog
    end_to_end_cases::write_pb::test_write_pb
```

The errors appear to be test races, but I've not looked into it yet

test_wipe_preserved_catalog

```
Server already has an ID (32); possibly a stray/orphan server from another test run.
```

test_write_pb

```
Server already has an ID (42); possibly a stray/orphan server from another test run.
```

crepererum:
Fixes #786.
Fixes #2403.

### DIFF
--- a/tests/common/server_fixture.rs
+++ b/tests/common/server_fixture.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::net::SocketAddrV4;
 use std::sync::Arc;
 use std::time::Duration;
@@ -41,25 +42,30 @@ impl BindAddresses {
         &self.grpc_bind_addr
     }
 
-    fn get_free_port() -> SocketAddrV4 {
+    fn get_free_ports(n: usize) -> Vec<SocketAddrV4> {
         let mut rng = rand::thread_rng();
         let ip = std::net::Ipv4Addr::new(127, 0, 0, 1);
 
-        loop {
+        let mut results = BTreeSet::new();
+
+        while results.len() < n {
             let port = rng.gen_range(8000..9000);
             let addr = SocketAddrV4::new(ip, port);
             if std::net::TcpListener::bind(addr).is_ok() {
-                return addr;
+                results.insert(addr);
             }
         }
+
+        results.into_iter().collect()
     }
 }
 
 impl Default for BindAddresses {
     /// return a new port assignment suitable for this test's use
     fn default() -> Self {
-        let http_addr = Self::get_free_port();
-        let grpc_addr = Self::get_free_port();
+        let mut addrs = Self::get_free_ports(2);
+        let http_addr = addrs.pop().unwrap();
+        let grpc_addr = addrs.pop().unwrap();
 
         let http_base = format!("http://{}", http_addr);
         let iox_api_v1_base = format!("http://{}/iox/api/v1", http_addr);
--- a/tests/common/server_fixture.rs
+++ b/tests/common/server_fixture.rs
@@ -1,5 +1,5 @@
-use std::collections::BTreeSet;
 use std::net::SocketAddrV4;
+use std::sync::atomic::{AtomicU16, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{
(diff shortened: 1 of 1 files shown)


tests/common/server_fixture.rs
```diff
@@ -47,24 +40,34 @@ impl BindAddresses {
...
+
+    fn get_free_port() -> SocketAddrV4 {
+        let mut rng = rand::thread_rng();
+        let ip = std::net::Ipv4Addr::new(127, 0, 0, 1);
+
+        loop {
+            let port = rng.gen_range(8000..9000);
+            let addr = SocketAddrV4::new(ip, port);
+            if std::net::TcpListener::bind(addr).is_ok() {
+                return addr;
```
shepmaster, line 52: Does anything prevent another process from grabbing this port between here and when the server actually starts?

tests/common/server_fixture.rs
```diff
@@ -47,24 +40,34 @@ impl BindAddresses {
...
 impl Default for BindAddresses {
     /// return a new port assignment suitable for this test's use
     fn default() -> Self {
-        let http_port = NEXT_PORT.fetch_add(1, SeqCst);
-        let grpc_port = NEXT_PORT.fetch_add(1, SeqCst);
-
-        let http_bind_addr = format!("127.0.0.1:{}", http_port);
-        let grpc_bind_addr = format!("127.0.0.1:{}", grpc_port);
+        let http_addr = Self::get_free_port();
+        let grpc_addr = Self::get_free_port();
```
shepmaster, line 62: It’s possible that these two ports are the same, due to randomness, correct?

or even two parallel tests grabbing the same random value(s)?

tests/common/server_fixture.rs
```diff
@@ -47,24 +40,34 @@ impl BindAddresses {
     pub fn grpc_bind_addr(&self) -> &str {
         &self.grpc_bind_addr
     }
+
+    fn get_free_port() -> SocketAddrV4 {
+        let mut rng = rand::thread_rng();
+        let ip = std::net::Ipv4Addr::new(127, 0, 0, 1);
+
+        loop {
+            let port = rng.gen_range(8000..9000);
```
shepmaster, line 49: Isn’t there a dedicated range for ephemeral ports?

tests/common/server_fixture.rs
crepererum: No, but I don't know if there's anything at all we can do about it (except for making the server more complicated and support passing an FD around).

tests/common/server_fixture.rs
crepererum: Uhh, good catch! We should probably hold the first port while scanning for the second.

tests/common/server_fixture.rs
crepererum: Does this count as ephemeral though? It's not really the client-side here, but we're looking for a server port that we're potentially connect to multiple times.

tests/common/server_fixture.rs
crepererum: fixed

NGA-TRAN (approved):
LGTM

tests/common/server_fixture.rs
```diff
@@ -41,25 +42,35 @@ impl BindAddresses {
...
         let mut rng = rand::thread_rng();
         let ip = std::net::Ipv4Addr::new(127, 0, 0, 1);
 
-        loop {
+        let mut results = BTreeSet::new();
+
+        while results.len() < n {
             let port = rng.gen_range(8000..9000);
             let addr = SocketAddrV4::new(ip, port);
+            if results.contains(&addr) {
```
shepmaster, line 54: The check is redundant since you have a set and the loop is on the length of the set.

crepererum: fixed

crepererum:
I need another round on this tomorrow. This doesn't work w/ multi-threaded tests (which we use by default). So we need a hybrid between the new and the old solution (aka using some process-global explicit or implicit set but also check if the port is usable).

shepmaster:
> a hybrid between the new and the old solution

I was going to suggest something like that. The old version with the monotonically increasing ports that skips a port if it can detect that the port is currently in use. We don't actually _desire_ randomness for this.

PR #2447, opened 2021-08-30
test: Database startup error when there are multiple active generations

### LINKED ISSUE #2196: Changes to server startup to support database deletion
Design from #1881:

Check for tombstone files within generation subdirectories before adding a database to the catalog
  - [ ] If there are more than one generation subdirectories missing tombstone files, this is a bad state. Mark the database as being in an error state and continue startup.
  - [x] If there aren't any generation subdirectories missing tombstone files, ~~don't add that database to the catalog at all.~~ Add it to the catalog in a state that indicates there are no active generation directories. Done in 6d0959fbc.

carols10cents:
Fixes #2196.

Most of the server startup work was done in https://github.com/influxdata/influxdb_iox/commit/6d0959fbc3c0bfb7420caa9e92b0ba7e250f7d6b but this adds a test for the expected error state of a database when there's more than one generation directory that doesn't have a tombstone file.

### DIFF
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -1257,10 +1257,14 @@ mod tests {
         },
     };
     use entry::test_helpers::lp_to_entry;
+    use futures::TryStreamExt;
     use influxdb_line_protocol::parse_lines;
     use iox_object_store::IoxObjectStore;
     use metrics::TestMetricRegistry;
-    use object_store::{path::ObjectStorePath, ObjectStore, ObjectStoreApi};
+    use object_store::{
+        path::{parsed::DirsAndFileName, ObjectStorePath},
+        ObjectStore, ObjectStoreApi,
+    };
     use parquet_file::catalog::{api::PreservedCatalog, test_helpers::TestCatalogState};
     use query::{exec::ExecutionContextProvider, frontend::sql::SqlQueryPlanner, QueryDatabase};
     use std::{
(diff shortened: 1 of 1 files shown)


NGA-TRAN (approved):
LGTM. I am wondering whether we plan to add in existing tests that with this parttern:
1. Create DB
2. Use that DB for some loads and queries
3. Delete DB
4. Use it and get error 
5. Create the same DB again
6. Use it again (loads and queries)

server/src/lib.rs
```diff
@@ -2050,6 +2054,83 @@ mod tests {
...
+            .flatten()
+            .collect();
+
+        // Delete all tombstone files
+        for file in database_files {
+            let parsed: DirsAndFileName = file.clone().into();
+            if parsed.file_name.unwrap().to_string() == "DELETED" {
+                application.object_store().delete(&file).await.unwrap();
+            }
+        }
```
crepererum, line 2107:
```suggestion
        // Delete all tombstone files
        let mut deleted_something = false;
        for file in database_files {
            let parsed: DirsAndFileName = file.clone().into();
            if parsed.file_name.unwrap().to_string() == "DELETED" {
                application.object_store().delete(&file).await.unwrap();
                deleted_something = true;
            }
        }
        assert!(deleted_something);
```

If the tombstone pattern ever changes, we should ensure that this test still deletes them. I'm using a similar pattern when messing around w/ the catalog. While for sure it's not perfect it's at least some safety net.

crepererum (approved):
Apart from the additional safety net I would like to see, that's a solid test :+1:

carols10cents:
@NGA-TRAN Great idea! I filed https://github.com/influxdata/influxdb_iox/issues/2454 to track that :) Thanks!

Issue #2396, opened 2021-08-24
Support Soft Delete


NGA-TRAN:
# Rationale
This is a part of https://github.com/influxdata/influxdb_iox/issues/2395. Refer to [this delete design document ](https://docs.google.com/document/d/1AKJYEvarjRxpnRx4q5XwcrhVpQtlVcLO0ltPjg_5aoU/edit#) for use cases, design ,  and related features.

# Implementation  Check List 
There are three main actions for soft delete, each has its own checklist

1. Issue delete

- [x]  Implement gRPC management API: `delete(db_name, table_name, predicate)` to issue the delete
- [x] #2463, #2667 and  #2567 
- [x] Parse specific range of timestamps
- [x] #2478
- [x] Add `delete_predicates` field into chunk catalog. Open MUB chunks must be closed/frozen with this event.
- [x] #2493
- [x] #2518 
- [x]  #2550
- [x] #2703 
- [x] Current we delete [start, stop). Need to make stop inclusive [start, stop] 
- [x] @e-dard will make [delete time range](https://github.com/influxdata/influxdb_iox/blob/d50f41c4271127606202fe5d495c7294b02bf206/predicate/src/predicate.rs#L176) and [query time range](https://github.com/influxdata/influxdb_iox/blob/c77f1ef4e290da66ab63396254ccb84db4bdf4cf/data_types/src/timestamp.rs#L1-L11) consistent #2734 
- [x] #2627 -> @crepererum 

2. Query none-deleted-data
- [x] Push delete predicates down to RUB and OS to pruning data as early as possible. Note that MUB has not yet supported predicate pushdown/pruning. OS only supports pruning. RUB  supports delete predicates push down.  This will be done at [`read_filter`](https://github.com/influxdata/influxdb_iox/blob/f7792aafe68edcb3bce01778bcf5f85bc844e64d/server/src/db.rs#L2504). Add  the following signature to be able to apply `select_predicate` and `delete_predicates` separately if needed.

```
pub fn read_filter(
   &self,
   select_predicate: Predicate,
   select_columns: Selection<'_>,
   delete_predicate: Predicate, 
)
```
- [x] At scan step, apply delete predicates on appropriate chunks
- [x] Optimize scan for chunks with no delete predicates #2603
- [x] Add delete_predicates as a parameter to influxrpc's operators #2751 &br #2548. -> @NGA-TRAN
- [x] Not throw error when all of rows of a OS chunk are soft deleted #2546 

[...]

PR #2448, opened 2021-08-30
feat: API and framework for Delete parsing and testing

NGA-TRAN:
Second check box of #2396 

Done in this PR
- [x] Clearer management API for parsing the delete time range and predicate
- [x] Parse timestamp
- [x] Framework for end-to-end test

Not that Parse Predicate is not done yet


server/src/db/catalog/chunk.rs
```diff
@@ -74,6 +75,9 @@ pub struct ChunkMetadata {
 
     /// The schema for the table in this Chunk
     pub schema: Arc<Schema>,
+
+    /// Delete predicates of this chunk
+    pub delete_predicates: Arc<Vec<Predicate>>,
```
NGA-TRAN, line 80: @crepererum Even though this is just a draft, can you help see if this is the right place to store the delete_predicates with the chunk catalog?

crepererum: I wonder if we need to store them on a chunk-level instead of on a partition- or even table-level. We're likely going to store them (in the preserved catalog) on a more coarse-grained level anyway.

NGA-TRAN: We do want to store them on a chunk-level because them only effect on chunks before their delete time. I do not think table-level is what we want because we will unexpectedly apply them on chunks of the table that created after deletes.
Let me dig where the chunk-level info stores

NGA-TRAN: @crepererum To follow up our discussion, here is is[ the link](https://github.com/influxdata/influxdb_iox/blob/317a83fc50d0c9e884fe90c88676ec1cb7e2f5c5/data_types/src/partition_metadata.rs#L175t )hat defines the supported predicate and its time range. I need to convert these 3 strings into them into a [Predicate](https://github.com/influxdata/influxdb_iox/blob/7602bde850976f3447d18fa1af558d7bc2c79f79/query/src/predicate.rs#L53) that includes predicate expressions and time ranges

influxdb_line_protocol/src/lib.rs
```diff
@@ -711,12 +728,113 @@ fn integral_value_signed(i: &str) -> IResult<&str, &str> {
...
+        || input.contains(" OR ")
+        || input.contains(" Or ")
+        || input.contains(" oR ")
+    {
+        return Err(nom::Err::Error(Error::OrInDeleteExpression {
+            value: input.to_string(),
+        }));
+    }
+
+    // This is very silly and expensive
```
e-dard, line 758: I don't think this will work correctly for all valid cases. There are some relevant line-protocol considerations:

 - tag keys can contain whitspace;
 - tag keys can contain `AND`;
 - tag keys are case sensitive.

Therefore it is possible to write the following into Cloud 2:

```
cpu,server\ AND\ abc=foo temp=100
cpu,server\ and\ abc=foo temp=300
```

This will produce two distinct time-series. In IOx we would store them as a table like this:

```
"server AND abc | server and abc | temp   | time
         foo   |          NULL  |   100  |  12345667
         NULL  |          foo   |   100  |  12345667
```

Finally, if we issue a delete with the following predicate:

```
"server AND abc" = foo
```

The **only** one row shoudl be deleted from the table.

However, I think that this logic here would result in a broken expression being built because the input would be split halfway into a tag key.

influxdb_line_protocol/src/lib.rs
```diff
@@ -711,12 +728,113 @@ fn integral_value_signed(i: &str) -> IResult<&str, &str> {
...
+///
+/// Example: city = "Boston" AND building != 12 and month = 10
+///
+/// This function is currently used to support Delete's gPRC API only.
+/// In the future, when we support all kinds of delete predicates supported in
+/// Datafusion, we may want to reuse the parser used by Datafusion to keep
+/// things consistent
+pub fn parse_delete_predicate(input: &str) -> IResult<&str, Vec<BinaryExpr>> {
+    // Or expression is invalid
+    if input.contains(" or ")
```
crepererum, line 748: You probably want something like `input.to_lowercase()` here and some lines down as well (the `AND` check).

influxdb_line_protocol/src/lib.rs
```diff
@@ -711,12 +728,113 @@ fn integral_value_signed(i: &str) -> IResult<&str, &str> {
...
+///
+/// Note that we currently not allow white spaces in the middle of the expression
+///
+/// Examples of valid expressions:
+///    " col=1000  "
+///    r#"col!="abc""#  -- Not there must be double quotes around abc
+///
+/// Examples of invalid expressions:
+///    "col =100"  -- white space after col
+///    "col= 100"  -- white space before 100
```
e-dard, line 804: The above two are both valid expressions in InfluxDB Cloud

influxdb_line_protocol/src/lib.rs
```diff
@@ -711,12 +728,113 @@ fn integral_value_signed(i: &str) -> IResult<&str, &str> {
...
+///    "col>100"  -- greater
+///    "col<=100"  -- less and equal
+///    "col=abc"   -- abc is not embedded in a double quote
```
e-dard, line 807: tag values do not need to be surrounded by double quotes. For example the following predicate is valid:

inserting the following
```
cpu,server=a temp 22.3
```

and then deleting with this predicate
```
server=a
```

Should result in the data being deleted.

In actual fact, if you provided the following predicate:

```
server="a"
```

Then nothing should get deleted because the `"` characters get interpreted as literals (that is, you're saying please delete any rows with the literal value `"a"` in the column `server`.

crepererum (changes requested):
I think the parser needs a tiny bit more work.

TBH I'm not a huge fan of these deeply nested hand-rolled parsers and would prefer having something that forces a more structured approach (like [rowa](https://github.com/rust-analyzer/rowan) or [logos](https://github.com/maciejhirsz/logos)) or even something that works with a freestanding grammar (like [ungrammar](https://github.com/rust-analyzer/ungrammar), [pest](https://github.com/pest-parser/pest), [lalrop](https://github.com/lalrpop/lalrpop)), but I think that's something we can do in some future.

influxdb_line_protocol/src/lib.rs
```diff
@@ -2209,4 +2357,135 @@ her"#,
 
         assert_eq!(vals[0].tag_value("asdf"), None);
     }
+
+    #[test]
+    fn test_parse_delete_expression() {
+        // Positive tests: valid expressions
+        let input = "col=123";
```
e-dard, line 2364: It might be good to DRY things up a bit here and make these into table driven tests.

influxdb_line_protocol/src/lib.rs
```diff
@@ -711,12 +728,113 @@ fn integral_value_signed(i: &str) -> IResult<&str, &str> {
...
+        let expr = parse_delete_expression(expr_str)?;
+        exprs.push(expr.1);
+    }
+    Ok((input, exprs))
+}
+
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct BinaryExpr {
+    pub column_name: String,
+    pub op: String,      // either "=" or "!="
```
e-dard, line 780: I agree with @crepererum that it might be good to make this an `enum`.

The other thing to consider here is that we need to decide where to translate the special `_measurement` and `_field` column names into the correct table name and column name respectively.

e-dard (changes requested):
I think overall this is looking good @NGA-TRAN! 🎉 

However, I think it's worth getting the parsing correct before we move forwards. I will send you an internal link with some test cases for the current parser.

influxdb_line_protocol/src/lib.rs
NGA-TRAN: I did convert it to lower case and discover it will make the case like this "city=Boston" wrong so I removed it :)

influxdb_iox_client/src/client/management.rs
```diff
@@ -932,13 +946,33 @@ impl Client {
...
-        let _delete_predicate = delete_predicate.into();
+        let db_name = db_name.into();
+        let table_name = table_name.into();
+        let delete_predicate = delete_predicate.into();
+        let start_time = start_time.into();
+        let stop_time = stop_time.into();
 
-        // todo
+        // NGA question: Currently, the "parse and validate start_time, stop_time, and delete_predicate" is
+        // done in the server's delete function. Should do it here at the client side instead?
```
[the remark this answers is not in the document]
NGA-TRAN: I think so, too. Let me move the parser call here

influxdb_line_protocol/src/lib.rs
```diff
@@ -69,6 +70,15 @@ pub enum Error {
         value: String,
     },
 
+    #[snafu(display(r#"Unable to parse delete expression with white space '{}'"#, value))]
```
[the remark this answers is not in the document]
NGA-TRAN: I am not happy with this either. Let me rewrite the delete parser not using line protocol one

influxdb_line_protocol/src/lib.rs
NGA-TRAN: This is a very good input. Let me play with the tests you pointed me to and rewrite this parser

influxdb_line_protocol/src/lib.rs
NGA-TRAN: Yeah, the `_measurement` and `_field` need more thinking

influxdb_line_protocol/src/lib.rs
NGA-TRAN: Yeah, I need to rewrite the parser

influxdb_line_protocol/src/lib.rs
NGA-TRAN: This is a good input, Edd

influxdb_line_protocol/src/lib.rs
NGA-TRAN: Currently I make different unit tests and this one is for single expression but I think the test might be organized differently after I rewrite a new parser

influxdb_line_protocol/src/lib.rs
NGA-TRAN: @e-dard I have read the [Flux Delete Syntax](https://docs.influxdata.com/influxdb/v2.0/reference/syntax/delete-predicate/), for string data type, we have to provide double quote. Here is examples:
 
Valid:
    ```
 city="Boston" -- string embedded in double quotes
 city = "Boston" -- white spaces supported
 temp=45.7  -- number no need quotes
temp = 45.6 -- white spaces supported
```

File: data_types/src/job.rs [hunk 10]
```diff
@@ -42,9 +42,6 @@ pub enum Job {
 
     /// Wipe preserved catalog
     WipePreservedCatalog { db_name: Arc<str> },
-
-    /// Delete data from a table
-    Delete { delete_info: DeleteInfo },
```
NGA-TRAN, line 47: We do not need the delete job (yet) so remove it make the code cleaner

data_types/src/partition_metadata.rs
```diff
@@ -163,38 +163,6 @@ impl TableSummary {
     }
 }
 
-/// Delete information
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
```
NGA-TRAN, line 167: Since no Delete job is not needed, its DeleteInfo no is not needed either

influxdb_iox_client/src/client/management.rs
```diff
@@ -360,7 +360,21 @@ pub enum DropPartitionError {
...
+    #[error("Server returned an empty response")]
+    EmptyResponse,
+
+    /// Server indicated that it is not (yet) available
+    #[error("Server unavailable: {}", .0.message())]
+    Unavailable(tonic::Status),
+
+    /// Client received an unexpected error from the server
+    #[error("Unexpected server error: {}: {}", .0.code(), .0.message())]
+    ServerError(tonic::Status),
```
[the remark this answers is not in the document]
NGA-TRAN: Yes, I will add more as I go. I think I will need to move some Error I out other places here, too.

NGA-TRAN (commented):
@e-dard and @crepererum 
I have refactored the code to create a clear structures to add parser at client side, as well as converting code to management API and vice versa. I have also added framework for ed-to-end tests.
There are a lot to do but I am happy this is clear enough to get reviewed and merged

influxdb_line_protocol/src/lib.rs
crepererum: Wait, but what if the input contains `city=" or "`? Do you bail out?

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -0,0 +1,253 @@
+//! parse string that should be done at client before sending to server
+
+#![deny(rustdoc::broken_intra_doc_links, rustdoc::bare_urls)]
```
alamb, line 3: I wonder if we have considered (re)using any of the code from the `sqlparser` / data fusion expression parser here?



Perhaps we could do something like rewrite the delete predicates to:

```
DELETE FROM <foo> WHERE <predicate>
```

And then run the https://github.com/sqlparser-rs/sqlparser-rs parser.

This would let us be sure the same syntax for timestamps and expressions were used. 

Even if we don't do this as part of the initial PR, I think consolidating the predicate parsing / handling in a future PR would save us both confusing user behavior when delete predicate support was inconsistent with query predicates as well as a second code path to maintain. I would be happy to help work on this

influxdb_line_protocol/src/lib.rs
NGA-TRAN: The parser will be redone next. I have removed all the code regarding this

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: @alamb The purpose of this refactor is exactly for that (to reuse the available parser). Beside the range timestamps that @e-dard suggested to keep them separated in the management API to avoid user to delete everything by accident, the delete predicate is not yet parsed yet and I have planned to ask you about using some available one in DF. This is just a place to plug the others in.

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -0,0 +1,253 @@
...
+pub struct ProvidedDeleteBinaryExpr {
+    pub column: String,
+    pub op: ProvidedDeleteOp,
+    pub value: String, // NGA Todo: should be enum like FieldValue
+}
+
+/// Delete Operator which either "=" or "!="
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy, Serialize, Deserialize)]
+pub enum ProvidedDeleteOp {
+    /// represent "="
```
e-dard, line 56: The doc string here has `///` but the fields are not public. Did you intend the fields to be public?

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -0,0 +1,253 @@
...
+    Eq,
+    /// represet "!="
+    NotEq,
+}
+
+impl ProvidedDeleteOp {
+    /// Return a str representation of this DeleteOp
+    pub fn as_str(&self) -> &'static str {
```
e-dard, line 64: You could also implement `Display for ProvidedDeleteOp`.

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -0,0 +1,253 @@
...
+    pub fn new(start_time: i64, stop_time: i64, predicate: Vec<ProvidedDeleteBinaryExpr>) -> Self {
+        Self {
+            start_time,
+            stop_time,
+            predicate,
+        }
+    }
+
+    /// Parse and convert the delete grpc API into ProvidedParseDelete to send to server
+    pub fn parse_delete(start: &str, stop: &str, predicate: &str) -> Result<Self> {
```
e-dard, line 83: I think it would be more idiomatic to call this fallable constructor `try_new` ?

query/src/test.rs
```diff
@@ -897,6 +897,11 @@ impl QueryChunkMeta for TestChunk {
     fn schema(&self) -> Arc<Schema> {
         Arc::clone(&self.schema)
     }
+
+    fn delete_predicates(&self) -> Arc<Vec<Predicate>> {
```
e-dard, line 901: Does this need a TODO somewhere or is it redundant?

e-dard (commented):
Looking good.

I agree with figuring out how we might leverage the other SQL parser for handling the non-time based predicates.

This PR mainly covers the management API and some initial parsing work, but it looks like it also touches on some plumbing around query nodes actually processing deletes. I think it's important to remember that when we run IOx in Cloud we will be processing deletes on the same Kafka consumption path, and those deletes will be encapsulated in Flat Buffer messages. I'm assuming that that work be handled in follow on PRs?

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: Will do

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: Good idea

query/src/test.rs
NGA-TRAN: This is trait function. I will add comment.

NGA-TRAN:
> when we run IOx in Cloud we will be processing deletes on the same Kafka consumption path, and those deletes will be encapsulated in Flat Buffer messages. I'm assuming that that work be handled in follow on PRs?

@e-dard : this is a good reminder especially this is my first time to look into creating new management API and Flat Buffer. I will add this in the todo list and might need some input.

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: Good catch. Yes, public.

NGA-TRAN: hah, I wonder why I used it as a public field somewhere already and it compiled. This is an enum and we do not need pub, @e-dard  :)

e-dard (approved):
👍

crepererum (approved):

PR #2450, opened 2021-08-31
test: simplify testing around preserved catalog

crepererum:
Preparation for the "dump catalog in text form" feature.

### DIFF
--- a/parquet_file/src/catalog/api.rs
+++ b/parquet_file/src/catalog/api.rs
@@ -977,7 +977,7 @@ mod tests {
     use crate::catalog::test_helpers::{
         assert_catalog_state_implementation, break_catalog_with_weird_version, TestCatalogState,
     };
-    use crate::test_utils::{chunk_addr, make_iox_object_store, make_metadata};
+    use crate::test_utils::{chunk_addr, make_iox_object_store, make_metadata, TestSize};
 
     #[tokio::test]
     async fn test_create_empty() {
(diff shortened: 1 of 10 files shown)


parquet_file/src/test_utils.rs
```diff
@@ -466,61 +493,65 @@ pub fn make_record_batch(
...
-        &mut schema_builder,
-    );
-    create_column_field_string(
-        &format!("{}_field_string_null_all", column_prefix),
-        vec![vec![None], vec![None], vec![None, None]],
-        &mut arrow_cols,
-        &mut summaries,
-        &mut schema_builder,
-    );
+    if test_size.is_full() {
```
NGA-TRAN, line 521: I wonder why we need the exact if condition here? Why we do not use the if above for this?
This function is so long, maybe we want to refactor it in the future?

NGA-TRAN (approved):

parquet_file/src/catalog/api.rs
```diff
@@ -238,6 +239,9 @@ pub struct PreservedCatalog {
     transaction_semaphore: Semaphore,
 
     iox_object_store: Arc<IoxObjectStore>,
+
+    fixed_uuid: Option<Uuid>,
+    fixed_timestamp: Option<DateTime<Utc>>,
```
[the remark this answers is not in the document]
crepererum: done

parquet_file/src/test_utils.rs
crepererum: Good call, I've split the function to make it easier to read.

PR #2451, opened 2021-08-31
feat: expose job metrics

### LINKED ISSUE #2416: Metrics for Operations

crepererum:
Closes #2416.

### DIFF
--- a/server/src/application.rs
+++ b/server/src/application.rs
@@ -6,7 +6,7 @@ use observability_deps::tracing::info;
 use query::exec::Executor;
 use write_buffer::config::WriteBufferConfigFactory;
 
-use crate::JobRegistry;
+use crate::{job::JobRegistryMetrics, JobRegistry};
 
 /// A container for application-global resources
 /// shared between server and all DatabaseInstances
@@ -28,13 +28,20 @@ impl ApplicationState {
         let num_threads = num_worker_threads.unwrap_or_else(num_cpus::get);
         info!(%num_threads, "using specified number of threads per thread pool");
 
+        let metric_registry_v2 = Arc::new(metric::Registry::new());
+
+        let job_registry = Arc::new(JobRegistry::new());
+        metric_registry_v2.register_instrument("job_registry_metrics", || {
+            JobRegistryMetrics::new(Arc::clone(&job_registry))
+        });
+
         Self {
             object_store,
             write_buffer_factory: Arc::new(Default::default()),
             executor: Arc::new(Executor::new(num_threads)),
-            job_registry: Arc::new(JobRegistry::new()),
+            job_registry,
             metric_registry: Arc::new(metrics::MetricRegistry::new()),
-            metric_registry_v2: Arc::new(Default::default()),
+            metric_registry_v2,
         }
     }
 
--- a/server/src/application.rs
+++ b/server/src/application.rs
@@ -6,7 +6,7 @@ use observability_deps::tracing::info;
 use query::exec::Executor;
 use write_buffer::config::WriteBufferConfigFactory;
 
-use crate::{job::JobRegistryMetrics, JobRegistry};
+use crate::JobRegistry;
 
 /// A container for application-global resources
 /// shared between server and all DatabaseInstances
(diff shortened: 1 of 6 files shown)


crepererum:
Still a draft, wanna add the wall and CPU times as increasing counters as well.

server/src/application.rs
```diff
@@ -28,13 +28,20 @@ impl ApplicationState {
         let num_threads = num_worker_threads.unwrap_or_else(num_cpus::get);
         info!(%num_threads, "using specified number of threads per thread pool");
 
+        let metric_registry_v2 = Arc::new(metric::Registry::new());
+
+        let job_registry = Arc::new(JobRegistry::new());
+        metric_registry_v2.register_instrument("job_registry_metrics", || {
+            JobRegistryMetrics::new(Arc::clone(&job_registry))
+        });
```
NGA-TRAN, line 36: I like you make it clearer here. I wonder if we name  `metric_registry_v2` more meaningful such as `job_registry_metric_registry`?

server/src/lib.rs
```diff
@@ -2351,4 +2351,44 @@ mod tests {
...
+
+        let wait_nanos = 1000;
+        let job = application
+            .job_registry()
+            .spawn_dummy_job(vec![wait_nanos], None);
+
+        job.join().await;
+
+        let mut reporter = metric::RawReporter::default();
+        application.metric_registry_v2().report(&mut reporter);
```
NGA-TRAN, line 2368: If you plan to modify the name in comment above, maybe modify the corresponding name here, too.

NGA-TRAN (approved):
LGTM

server/src/application.rs
crepererum: I think it's an intermediate stage as long as we have two metric systems.

crepererum:
Changed the implementation following a hint from @tustvold and included wall and CPU time histograms (only for completed jobs, should be sufficient for now). Now ready for final review.

server/src/job.rs
```diff
@@ -69,7 +71,96 @@ impl JobRegistry {
...
+        mut_ref.registry.tracked_len()
+    }
+}
+
+#[derive(Debug)]
+struct JobRegistryMetrics {
+    active_gauge: metric::Metric<metric::U64Gauge>,
+    cpu_time_histogram: metric::Metric<metric::DurationHistogram>,
+    wall_time_histogram: metric::Metric<metric::DurationHistogram>,
+    reclaimed_accu: BTreeMap<metric::Attributes, u64>,
```
tustvold, line 87: Is there a reason this isn't a `metric::Metric<metric::U64Counter>`?

server/src/job.rs
```diff
@@ -69,7 +71,96 @@ impl JobRegistry {
...
+            }
+            if let Some(nanos) = status.wall_nanos() {
+                self.wall_time_histogram
+                    .recorder(attr)
+                    .record(std::time::Duration::from_nanos(nanos as u64));
+            }
+        }
+
+        // scan current jobs
+        let mut accumulator: BTreeMap<metric::Attributes, u64> = self.reclaimed_accu.clone();
```
tustvold, line 134: Why does this clone reclaimed_accu - doesn't this mean the active gauge will report active and past jobs?

server/src/job.rs
```diff
@@ -69,7 +71,96 @@ impl JobRegistry {
...
+    cpu_time_histogram: metric::Metric<metric::DurationHistogram>,
+    wall_time_histogram: metric::Metric<metric::DurationHistogram>,
+    reclaimed_accu: BTreeMap<metric::Attributes, u64>,
+}
+
+impl JobRegistryMetrics {
+    fn new(metric_registry_v2: Arc<metric::Registry>) -> Self {
+        Self {
+            active_gauge: metric_registry_v2
+                .register_metric("influxdb_iox_job_count", "Number of known jobs"),
```
tustvold, line 94:
```suggestion
                .register_metric("influxdb_iox_job_count", "Number of running jobs"),
```

tustvold (approved):
Approved, but I'm not sure that the value reported by active_gauge is correct? Or perhaps I misunderstand what the gauge is. Currently it is a gauge but I don't see how it could ever decrease...

server/src/job.rs
crepererum: Yes:

- active (aka non-finished jobs) are counted as they currently exist
- finished / reclaimed jobs are counted up

This way the 2nd case doesn't depend on the history size of the tracker. This is useful because the sum of both cases can be used as any other counter (you could even use it calculate "jobs per second" or something).

server/src/job.rs
crepererum: That's not exactly true, since this gauge also includes finished jobs (just by a different attribute/label), see [this comment](https://github.com/influxdata/influxdb_iox/pull/2451#discussion_r702772645).

server/src/job.rs
crepererum: Because later we fold that number into a single gauge that has a `(job description, status)` granularity. One of the status variants is `finished`.

server/src/job.rs
tustvold: Aah - then I think this code has a race condition if a job completes between reclaim and update - specifically the running count will decrease but the finished count won't increase

crepererum: @tustvold fixed, PTAL

tustvold (approved):
LGTM :+1:

PR #2453, opened 2021-09-01
feat: "dump catalog" debug CLI

crepererum:


src/main.rs
```diff
@@ -176,7 +179,6 @@ fn main() -> Result<(), std::io::Error> {
 
         match config.command {
             Command::Database(config) => {
-                let _tracing_guard = handle_init_logs(init_simple_logs(log_verbose_count));
```
carols10cents, line 179: Is removing log initialization from the `database` commands intentional? This change didn't look related to dumping the catalog to me...

carols10cents (approved):
One teeny tiny question, otherwise LGTM!

src/main.rs
NGA-TRAN: +1

NGA-TRAN (approved):
LGTM

src/main.rs
crepererum: fixed

PR #2456, opened 2021-09-01
feat: correctly account MUB sizes

### LINKED ISSUE #1565: MUB Memory Accounting
The memory accounting within the MUB is currently quite optimistic:

- [ ] It doesn't account for the size of the Chunk structures/enumerations themselves only the data they contain
- [ ] It only counts used data, not allocated - i.e. it uses len() not capacity()
- [ ] It doesn't account for the costs of snapshot storage

~Currently the lifecycle uses a threshold on a chunk's memory usage to decide when to close it, points two and three, however, will seriously confuse this logic. I therefore think we need to hold off on this change until a move to driving the lifecycle off something like row count. This will likely fall out of the work @pauldix is doing on compaction policy~ (fixed)

crepererum:
Fixes #1565.


server/src/db/catalog/chunk.rs
```diff
@@ -880,9 +880,7 @@ impl CatalogChunk {
         self.set_lifecycle_action(ChunkLifecycleAction::Dropping, registration)?;
 
         // set memory metrics to 0 to stop accounting for this chunk within the catalog
-        self.metrics.memory_metrics.mutable_buffer.set(0);
-        self.metrics.memory_metrics.read_buffer.set(0);
-        self.metrics.memory_metrics.object_store.set(0);
+        self.metrics.memory_metrics.set_to_zero();
```
carols10cents, line 883: 👍🏻 In addition to preventing race conditions, I like this abstraction!

query_tests/src/sql.rs
```diff
@@ -312,8 +312,8 @@ async fn sql_select_from_system_chunks() {
         "+----+---------------+------------+-------------------+--------------+-----------+",
         "| id | partition_key | table_name | storage           | memory_bytes | row_count |",
         "+----+---------------+------------+-------------------+--------------+-----------+",
-        "| 0  | 1970-01-01T00 | h2o        | OpenMutableBuffer | 213          | 3         |",
-        "| 0  | 1970-01-01T00 | o2         | OpenMutableBuffer | 177          | 2         |",
+        "| 0  | 1970-01-01T00 | h2o        | OpenMutableBuffer | 1639         | 3         |",
+        "| 0  | 1970-01-01T00 | o2         | OpenMutableBuffer | 1635         | 2         |",
```
carols10cents, line 316: 🤣 Yeahhhhh just a bit of an accounting error!!

carols10cents (approved):
LGTM!

server/src/db/catalog/metrics.rs
```diff
@@ -214,82 +214,122 @@ impl PartitionMetrics {
 ///
 /// This can then be used within each `CatalogChunk` to record its observations for
 /// the different storages
-#[derive(Debug)]
 pub struct StorageGauge {
-    pub(super) mutable_buffer: GaugeValue,
-    pub(super) read_buffer: GaugeValue,
-    pub(super) object_store: GaugeValue,
+    inner: Mutex<StorageGaugeInner>,
```
tustvold, line 218: This Mutex doesn't actually prevent an observer from getting an inconsistent snapshot of the gauges - the reported aggregate totals are shared and therefore. I'll need to have a think about how to support this with the new metric API (which has the same limitation) if this is even something we particularly care about.

mutable_buffer/src/column.rs
```diff
@@ -319,24 +319,33 @@ impl Column {
         }
     }
 
-    /// The approximate memory size of the data in the column. Note that
-    /// the space taken for the tag string values is represented in
+    /// The approximate memory size of the data in the column.
+    ///
+    /// Note that the space taken for the tag string values is represented in
```
tustvold, line 324: This comment is out-of-date, there is no longer a per-chunk dictionary. The size reported here is the size of the column

mutable_buffer/src/chunk.rs
```diff
@@ -118,31 +118,28 @@ impl MBChunk {
...
-        snapshot
+        (snapshot, true)
     }
 
-    /// Returns a queryable snapshot of this chunk
+    /// Returns a queryable snapshot of this chunk and an indicator if the snapshot was just cached.
     #[cfg(feature = "nocache")]
     pub fn snapshot(&self) -> Arc<ChunkSnapshot> {
-        Arc::new(ChunkSnapshot::new(self))
+        (Arc::new(ChunkSnapshot::new(self)), false)
```
tustvold, line 142: The type signature needs to be updated for this variant also

mutable_buffer/src/chunk.rs
```diff
@@ -118,31 +118,28 @@ impl MBChunk {
         self.write_columns(batch.columns())?;
 
         // Invalidate chunk snapshot
-        *self
```
tustvold, line 121: Why did you remove this? This method is still &mut self and therefore this should never fail?

tustvold (approved):
Looks good to me, I think there is a general raciness issue with the way the metrics for these chunks are being recorded (which this PR didn't introduce) but the change to accounting looks like it should do the trick :+1:

server/src/db/catalog/metrics.rs
crepererum: The mutex here is more for "inner mutability" instead of exactness, because the query access that only holds a immutable chunk ref can create a MUB snapshot which then needs a metrics update. Up to this point metric updates required a mutable ref though.

mutable_buffer/src/chunk.rs
crepererum: Leftover from debugging. Reverted.

mutable_buffer/src/chunk.rs
crepererum: done

mutable_buffer/src/column.rs
crepererum: fixed by removing the paragraph.

tests/end_to_end_cases/management_api.rs
```diff
@@ -686,7 +686,7 @@ async fn test_list_partition_chunks() {
         id: 0,
         storage: ChunkStorage::OpenMutableBuffer.into(),
         lifecycle_action: ChunkLifecycleAction::Unspecified.into(),
-        memory_bytes: 100,
+        memory_bytes: 1016,
```
alamb, line 689: well, that was a fairly large mismatch. 👍  for making this more consistent

PR #2457, opened 2021-09-01
feat: Add list_deleted_databases API

### LINKED ISSUE #2198: Add a list soft-deleted databases API to support database deletion
Design from #1881:

Add a list restoration candidates API that takes a database name (or lists all candidates you have access to?) Name suggestions: `restore_list`, `list_restore_candidates`, `list_deleted_databases` ✅  ?
  - If there are any generation subdirectories that don't contain tombstone files, return an error because you can't restore a database while there's a current database with that name
  - Return all existing generation subdirectories with helpful metadata like the time they were deleted (which would be the creation time of the tombstone file) or the amount of data in them so the user can pick which one to restore

carols10cents:
Ok this is done!

Closes #2198.

If you've created and deleted some databases locally, you can now do:

```
$ cargo run -- database list --deleted
Deleted at                      | Generation ID | Name
--------------------------------+---------------+--------
2021-09-02 18:57:02.496932 UTC   0               mydb
2021-09-02 18:57:02.496987 UTC   1               mydb
```

This PR also includes:

- A test for `list_possible_databases` that I realized was missing
- A private-for-now `list_all_databases` API that does more object storage calls than `list_possible_databases` but also returns the generation metadata (both deleted and active generations), I'm not sure if this API will be useful outside the crate or not, but it was at least an intermediate step to `list_deleted_databases` that felt like it made sense to have
- A tiny fix to an incorrect assumption in the in-memory object storage that's only a problem if you have a database whose name is a prefix of another database's name


pauldix (approved):
LGTM. Maybe the only other thing I'd add is an option to list `--detailed` on the databases command that includes existing and deleted all in one go with the generation information.

carols10cents:
> LGTM. Maybe the only other thing I'd add is an option to list `--detailed` on the databases command that includes existing and deleted all in one go with the generation information.

Sounds good! I've opened https://github.com/influxdata/influxdb_iox/issues/2462 for that and I'm going to merge this.

Issue #2455, opened 2021-09-01
Write Buffer's Kafka implementation should create topic


e-dard:
Currently when you push a write buffer config (for either a reader or writer) that defines Kafka as the log to read from/write to then there is an assumption that the topic for the database already exists.

This complicates the running of IOx in this way in a production setting.

It would be preferable if IOx created the topic when it first processed the pushed config. It should be possible to create the topic from either the production side (IOx router) or the consumer side (IOx query node).

There is some discussion of the how the current approach is confusion here: https://github.com/influxdata/influxdb_iox/issues/2189

### Definition of Done

 - when a config is pushed to an IOx router, the router will check if a topic for the database exists via the provided Kafka cluster connection.
 - if the topic does not exist the router will create it.
 - if the topic exists the router will correctly handle the response (and not do anything further).
 - the same steps above should apply for an IOx query node.

PR #2458, opened 2021-09-02
refactor: more flexible writer buffer config

crepererum:
This allows:

- different types (instead of guessing through the connection URL)
- sequencer counts (not used yet but will be by #2455)
- extensible configs (e.g. to configure Kafka in a more granular way,
  not wired up yet)
- future extensions (since we use a message now instead of a single
  string)

**BREAKING: This requires changes for deployed systems / existing DBs!**


pauldix (approved):
Overall looks good on a first read. It's hard to see exactly how the `creation_config` and `connection_config` will be used without actual examples, but I assume you have a plan there.

PR #2461, opened 2021-09-03
perf: Don't open the output file each time we write.

shepmaster:
This improves performance of the the file output mode, which should
make it easier to improve the performance of the core generation
logic.

Benchmarked via:

```
time \
./target/release/iox_data_generator \
--spec iox_data_generator/schemas/fully-supported.toml \
--output /tmp/out \
--start '1 month ago'
```

Before:

```
Submitted 271608 total points

real	10.912	10911567us
user	3.129	3129032us
sys	6.257	6257340us
cpu	86%
mem	7152 KiB
```

After:

```
Submitted 271588 total points

real	2.291	2291364us
user	1.969	1969357us
sys	0.058	58030us
cpu	88%
mem	7104 KiB
```

That's 21.0% of the previous time.

### DIFF
--- a/iox_data_generator/src/lib.rs
+++ b/iox_data_generator/src/lib.rs
@@ -77,6 +77,15 @@ pub enum Error {
         /// Underlying `agent` module error that caused this problem
         source: agent::Error,
     },
+
+    /// Error that may happen when constructing an agent's writer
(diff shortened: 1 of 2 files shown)


carols10cents (approved):
AFAIK, Paul is using the data generator to write into IOx, not to a file, so this won't impact that work either positively or negatively. This looks good though, so merging!

PR #2464, opened 2021-09-03
feat: Add an API for restoring a database that was marked deleted

### LINKED ISSUE #2199: Add a restore database API to support database deletion undo
Design from #1881: 

Add a restore database API that takes a database name and generation ID gotten from the previous list restore candidates (if this is done in a UI the end user won't necessarily need to see or worry about the generation ID at all, but we need to know which one we're restoring)
  - If there are any generation subdirectories that don't contain tombstone files, return an error because you can't restore a database while there's a current database with that name (prevent race conditions between running list restore candidates and restoring a candidate)
  - Delete the tombstone file in the specified generation subdirectory and return success

carols10cents:
Ok, I think I've figured out how to restart the state machine to re-initialize a database when it's restored... by introducing more Mutexes. Please check this and let me know if you see a better way to do it!

Closes #2199


server/src/database.rs
```diff
@@ -216,6 +219,36 @@ impl Database {
...
+            })?
+        };
+
+        let shared = Arc::clone(&self.shared);
+
+        {
+            // Reset the state to initializing with the given iox object storage
+            let mut state = shared.state.write();
+            *state.unfreeze(handle) =
+                DatabaseState::DatabaseObjectStoreFound(DatabaseStateDatabaseObjectStoreFound {
```
carols10cents, line 244: Here I'm resetting the state to `DatabaseObjectStoreFound`, and then I'd like to restart the background worker to redo initialize_database, but `self` isn't `mut` and I'm not sure if it makes sense to... change `self.join` to a Mutex? So that it can be replaced with a new `Future` here? 

Is it possible that the old future I want to replace here would still be doing work, and what would we do here if it was?

tustvold (dismissed):
I think the gRPC and object store plumbing look good, I think there might be a way to address the race and avoid needing to change Database at all. I'll have a play later today and see what I can come up with, and possibly raise a PR targeting this one. I _think_ it should simply be a case of doing the same thing Server::create_database does, and swapping in a new Database.

Edit: TBC this wasn't me suggesting you couldn't do this, but I don't want to send you off on a wild goose chase if it doesn't work :laughing:

server/src/database.rs
```diff
@@ -212,20 +218,67 @@ impl Database {
...
+                    iox_object_store: Arc::new(iox_object_store),
+                });
+            info!(%db_name, "set database state to object store found");
+        }
+
+        // Restart the background worker to re-initialize the database
```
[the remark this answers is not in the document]
alamb: In case it is not clear, @tustvold proposal is in https://github.com/influxdata/influxdb_iox/pull/2530

iox_object_store/src/lib.rs
```diff
@@ -363,6 +372,58 @@ impl IoxObjectStore {
...
+        ensure!(
+            active.is_none(),
+            ActiveDatabaseAlreadyExists {
+                name: database_name
+            }
+        );
+
+        let restore_candidate = generations
+            .iter()
+            .find(|g| g.id == generation_id && !g.is_active())
```
alamb, line 401: FWIW given the check above for `g.is_active()` it is probably redundant to check for `is_active()` here as well. Though it won't hurt either

server/src/database.rs
```diff
@@ -212,20 +218,67 @@ impl Database {
...
+            info!(%db_name, "set database state to no active database");
         }
 
+        self.shutdown();
+
+        Ok(())
+    }
+
+    /// Mark this database as restored.
+    pub async fn restore(&self, iox_object_store: IoxObjectStore) -> Result<(), Error> {
```
alamb, line 230: I think with https://github.com/influxdata/influxdb_iox/pull/2530, this method can simply transition the database from `DatabaseState::NoActiveDatabase` back to `DatabaseState::Known` and the background worker will take over from there. 

@tustvold  does that sound correct?

tustvold: Correct, I think this PR still needs to hold a FreezeHandle across the actual object store operation to prevent recover races. 

This method could therefore do the object store operation much like the in the delete case, but it could also just hold the Server level FreezeHandle and continue to do the object store transaction there. It doesn't really matter, so long as you can't run two recover operations on the same database concurrently

iox_object_store/src/lib.rs
carols10cents: This is checking for NOT active though.

carols10cents:
This was much easier with @tustvold 's changes!! Ready for rereview :)

server/src/database.rs
```diff
@@ -244,6 +258,47 @@ impl Database {
...
+                inner: generation_id,
+            },
+        )
+        .await
+        .context(CannotRestoreDatabaseInObjectStorage)?;
+
+        let shared = Arc::clone(&self.shared);
+
+        {
+            // Reset the state
```
alamb, line 292: 👨‍🍳 😗

server/src/database.rs
```diff
@@ -1291,22 +1344,7 @@ mod tests {
...
-
-        os.delete(&path).await.unwrap();
-
-        {
-            let mut state = database.shared.state.write();
-            let mut state = state.get_mut().unwrap();
-            *state = DatabaseState::Known(DatabaseStateKnown {});
-            database.shared.state_notify.notify_waiters();
-        }
+        database.restore(0).await.unwrap();
```
alamb, line 1347: 👍

alamb (approved):
I think this looks great @carols10cents  -- great work

server/src/database.rs
```diff
@@ -244,6 +258,47 @@ impl Database {
...
+        let db_name = &self.shared.config.name;
+        info!(%db_name, %generation_id, "restoring database");
+
+        let handle = {
+            let state = self.shared.state.read();
+
+            // Can't restore an already active database.
+            ensure!(!state.is_active(), CannotRestoreActiveDatabase { db_name });
+
+            state.try_freeze().context(TransitionInProgress {
```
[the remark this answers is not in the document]
carols10cents: Oooh nice! That looks way better.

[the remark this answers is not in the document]
carols10cents: We decided to make this change here and in other places in a separate PR so that we can merge this in but other folks can review the other change at a later time.

PR #2465, opened 2021-09-03
test: Ensure that a deleted and recreated database returns new data

### LINKED ISSUE #2454: Add integration test for database deletion and querying
Great idea from @NGA-TRAN [here](https://github.com/influxdata/influxdb_iox/pull/2447#pullrequestreview-742057051):

1. Create DB
2. Use that DB for some loads and queries
3. Delete DB
4. Use it and get error 
5. Create the same DB again
6. Use it again (loads and queries)

shepmaster:
Closes #2454

### DIFF
--- a/tests/end_to_end_cases/deletes.rs
+++ b/tests/end_to_end_cases/deletes.rs
@@ -1,0 +1,98 @@
+use super::scenario::rand_name;
+use crate::common::server_fixture::ServerFixture;
+use arrow_util::assert_batches_sorted_eq;
+use generated_types::influxdata::iox::management::v1::DatabaseRules;
+
+type Result<T, E = Box<dyn std::error::Error>> = std::result::Result<T, E>;
+
+#[tokio::test]
+async fn test_querying_deleted_database() {
+    let fixture = ServerFixture::create_shared().await;
+    let mut management_client = fixture.management_client();
+    let mut write_client = fixture.write_client();
+    let mut flight_client = fixture.flight_client();
+
+    let db_name = rand_name();
+    let rules = DatabaseRules {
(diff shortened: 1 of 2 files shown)


NGA-TRAN (approved):
Great tests.

tests/end_to_end_cases/deletes.rs
```diff
@@ -0,0 +1,98 @@
...
+    // ensure the data is returned.
+
+    management_client
+        .create_database(rules.clone())
+        .await
+        .expect("create database failed");
+    let num_lines_written = write_client
+        .write(&db_name, "cpu,region=west user=12.3 100")
+        .await
+        .expect("cannot write");
```
NGA-TRAN, line 31: I think here the expect is something like "successful write", right? Or the "expect" happens at timeout and we have not received anything yet?

shepmaster: I'm not 100% sure I understand your question...

`expect` can sometimes appear a bit backwards, but you generally put the text you want shown to the user:

```rust
let a: Result<u8, _> = Err("Database is deleted");
a.expect("could not write");
```

```
thread 'main' panicked at 'could not write: "Database is deleted"', src/main.rs:3:7
```

I don't fully know all the possible reasons that writing lines might fail. I assume things like network connections flaking, the server being down, the write being rejected, etc. Any of those could cause this particular `write` call to be an `Err`.

NGA-TRAN: I asked because I saw `.expect("successful write)` in other places with the same context. I think I understand  now the `expect` happens when something goes wrong.

Issue #1075, opened 2021-03-30
Use proper jsonpb encoding for our protos


mkmik:
Our CLI needs to be able to accept configuration as JSON and render configuration as JSON.

Protobufs technically have an official JSON encoding rule called `jsonpb` but prost doesn't
offer native support for it yet.

The feature may land upstream (See https://github.com/danburkert/prost/issues/75, https://github.com/danburkert/prost/pull/255, https://github.com/danburkert/prost/issues/277, https://github.com/danburkert/prost/pull/336); we may help and/or we may find some workarounds until it does.

In #1069 we started by just adding `serde` derives, but there are still major divergences from `jsonpb`:

- [ ] Enums are rendered numerically (jsonpb renders as string but accepts both string or number)
- [ ] Fields with default values are emitted (jsonpb omits a field if its value is `0`, `""`, `[]`, `null`)
- [ ] Well Known Types (such as `Any` and `Duration`) may have specific json renderings.

jacobmarble:
@mkmik please update this issue. Can it be closed now?

mkmik:
there may be some things missing (like encoding of well known types such as Any) but now that responsibility has moved to the https://github.com/influxdata/pbjson repo (e.g. https://github.com/influxdata/pbjson/issues/2) so I think we can close this one

Issue #908, opened 2021-03-02
Improved CLI Output


tustvold:
In order to facilitate scripting of the CLI (and testing), it should be possible to request JSON-formatted output, enable/disable ANSI escape codes, etc... 

There should also be stability guarantees associated with the structure of this output. Depending on when JSON support lands in prost, it might be possible to build on this - see [here](https://github.com/danburkert/prost/issues/75).

Creating this issue largely as a placeholder that will need further fleshing out as part of an implementation effort.

alamb:
There was some additional discussion here: https://github.com/influxdata/influxdb_iox/pull/939#discussion_r589411347

about the use of `println!` vs `eprintln!` and possibly not printing anything on success at all

tustvold:
Much of the CLI is now making use of pbsjon to provide stable output

PR #2468, opened 2021-09-05
feat: add pbjson support

tustvold:
Relates to https://github.com/influxdata/influxdb_iox/issues/908 and https://github.com/influxdata/influxdb_iox/issues/1075

**Background:**

Large amounts of IOx make use of protobuf, for API and persisted data. This is beneficial as it gives a strongly-typed data representation, with a well developed story for schema evolution, multi-language compatibility, etc... Unfortunately, being a binary representation, it is not human readable nor editable. This presents a problem for the CLI and other tooling that involves user interaction.

Unfortunately despite there being a standard for mapping Protobuf to JSON, prost does not currently support it - https://github.com/tokio-rs/prost/issues/75 and https://github.com/tokio-rs/prost/issues/277.

We currently work around this in a couple of ways

1. We JSON serialize/deserialize the data_types representations
2. We debug print the protobuf
3. We JSON serialize/deserialize the prost generated struct with autoderived serde annotations

Workaround 1. is problematic as we don't really want to maintain two separate wire-compatible serialization APIs if we can avoid it, and these types are defaulted and don't have the optional-ity, by design, that is required to make API evolution possible

Workaround 2. is problematic as it isn't possible to deserialize this representation and has no stability guarantees

Workaround 3. is problematic as it has limited stability guarantees, there are non-breaking protobuf changes that would result in a breaking change to it, it encodes enumerations solely as integers, and is best described as a pragmatic hack

This PR therefore adds code that can generate serde Deserialize and Serialize implementations that are compliant with the protobuf JSON mapping. I'm not sure this code should live in IOx long-term, but if we're going to use it in IOx, which is itself a question, I think it should go through the standard code review process.

**Approach**

I initially starting thinking this would be a moderately straightforward modification of the prost proc-macro. Unfortunately it quickly became clear that additional information is necessary for the conversion, which would necessitate additional non-trivial modifications to prost-build and prost itself. It also became clear that the generated code, and the code to generate it, is relatively complex, and I thought a tough sell to get included in prost.

I therefore decided to write this as a separate code generation step that can be run on the descriptor set that prost-build outputs (and also itself uses). I decided not to use the proc-macro approach that prost uses, because I felt it just added additional complexity. With prost I believe there is a desire for the code generation to be optional and therefore there is a desire for a proc-macro to do the heavy lifting, I don't anticipate this being required functionality for this.

**Limitations**

* This will never generate "correct" serialization implementations for the well-known-types such as Any, Duration, etc... These will need to be hand-crafted but there aren't too many of them, they are relatively simple, and they are highly unlikely to change
* Bytes are currently not supported as I stumbled across a bug in the prost macro whilst implementing them
* The generated logic is NOT optimal w.r.t collections, a future iteration could/should implement streaming per-element conversion if this becomes a problem
* Limited effort has been put into avoiding name collisions - if a message has a field called "v" or "GeneratedVisitor" I expect it to cause issues (FWIW prost-build also has this problem)
* Serde is complicated, code generating code is complicated, protobuf's descriptor format is wild, all together the result is rather :exploding_head: - I've tried my best but apologise in advance...


pbjson/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[doc(hidden)]
+pub mod private {
+    use std::str::FromStr;
+    /// Used to parse a number from either a string or its raw representation
+    #[derive(Debug, Copy, Clone, PartialOrd, PartialEq, Hash, Ord, Eq)]
+    pub struct NumberDeserialize<T>(pub T);
+
+    #[derive(serde::Deserialize)]
+    #[serde(untagged)]
```
tustvold, line 18: There isn't actually a way to write this without using a "private" serde API which is what the proc macro uses - https://github.com/serde-rs/serde/issues/1947

pbjson/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
```
tustvold, line 1: Much like serde or prost, there is a crate containing the dependencies for the generated code.

google_types/build.rs
```diff
@@ -16,16 +17,18 @@ fn main() -> Result<()> {
...
-            ".google",
-            "#[derive(serde::Serialize,serde::Deserialize)] #[serde(rename_all = \"camelCase\")]",
-        )
         .bytes(&[".google"])
         .compile_protos(&proto_files, &[root])?;
 
+    let descriptor_set = std::fs::read(descriptor_path)?;
+    pbjson_build::Builder::new()
+        .register_descriptors(&descriptor_set)?
+        .build(&[".google"])?;
```
tustvold, line 31: Note: this will derive incorrect serializations for the well-known-types - this can be fixed in a later PR

pbjson_build/src/generator.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        for _ in 0..self.0 {
+            write!(f, "    ")?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+pub struct Config {
```
tustvold, line 28: I expect more options to find there way here over time

pbjson_build/src/generator/enumeration.rs
```diff
@@ -0,0 +1,138 @@
+//! This module contains the code to generate Serialize and Deserialize
```
tustvold, line 1: I recommend reading https://serde.rs/impl-serialize.html, https://serde.rs/impl-deserialize.html to understand what this is generating

pbjson_build/src/generator/message.rs
```diff
@@ -0,0 +1,754 @@
+//! This module contains the code to generate Serialize and Deserialize
```
tustvold, line 1: https://serde.rs/deserialize-struct.html may help understand the generated code better

tests/end_to_end_cases/management_cli.rs
```diff
@@ -112,7 +112,7 @@ async fn test_create_database() {
                 .and(predicate::str::contains(format!(r#""name": "{}"#, db)))
                 // validate the defaults have been set reasonably
                 .and(predicate::str::contains("%Y-%m-%d %H:00:00"))
-                .and(predicate::str::contains(r#""bufferSizeHard": 104857600"#))
+                .and(predicate::str::contains(r#""bufferSizeHard": "104857600""#))
```
tustvold, line 115: U64 are encoded as strings because all numbers in JSON are f64

pbjson_build/src/generator/message.rs
```diff
@@ -0,0 +1,754 @@
...
+            writeln!(writer, "?;")?;
+            writeln!(
+                writer,
+                "{}struct_ser.serialize_field(\"{}\", &v)?;",
+                Indent(indent),
+                field.json_name()
+            )
+        }
+        FieldType::Map(_, value_type)
+            if matches!(
```
tustvold, line 249: Unfortunately box destructuring support seems to have stalled out slightly - https://github.com/rust-lang/rust/issues/29641

google_types/build.rs
mkmik: yeah, they are wrong anyway with the current hack

e-dard (commented):
I haven't started reviewing this yet but I'm wondering given the significant amount of code, which I assume is generally useful and not specific to IOx if this should be in a separate repo/project. If you think that would make sense @tustvold let me know and I can help on the influx side of things.

mkmik:
I also think this should go in a separate project eventually; I interpreted what @tustvold said as that we could first put it here since we have a core review workflow setup in this repo and possibly split it up later

pbjson_build/src/generator.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+    writeln!(writer, "{}];", Indent(indent))?;
+    writeln!(writer)
+}
+
+fn write_serialize_start<W: Write>(indent: usize, rust_type: &str, writer: &mut W) -> Result<()> {
+    writeln!(
+        writer,
+        r#"{indent}impl serde::Serialize for {rust_type} {{
+{indent}    #[allow(deprecated)]
```
mkmik, line 89: I wonder if this code would be cleaner with something like https://crates.io/crates/indent_write

tustvold:
@e-dard I don't have strong opinions about where this code lives, I'd be happy to split it out into a new public repository if we'd rather that. I was just trying to avoid adding a lot of git dependencies to IOx, as I don't think this is quite ready just yet to be published to crates.io. Provided there is a code review flow I'm cool with wherever :grin:

pbjson_build/src/message.rs
```diff
@@ -0,0 +1,275 @@
+//! The raw descriptor format is not very easy to work with, a fact not aided
+//! by prost making almost all members of proto2 syntax message optional
```
mkmik, line 2: isn't the protobuf JSON canonical mapping only a json3 thing anyway?

mkmik (approved):
LGTM

pbjson_build/src/message.rs
tustvold: It is, but the descriptor format is proto2.

As an aside this generator does support generating serde implementations for proto2, but as you rightly point out it is not a standard and different libraries may do different things - especially w.r.t defaults which I only now realise how borked they are

crepererum:
> I was just trying to avoid adding a lot of git dependencies to IOx, as I don't think this is quite ready just yet to be published to crates.io.

I honestly don't understand why people are so shy about publishing a crate. `0.x` releases are meant for exactly this situation. Publish early, publish often. If people really use a `0.x` crate and later complain about bugs or API breakage, it's their fault.

alamb:
One of the benefits of putting this up as a crate on crates.io is that if you give the community a bit of attention, it is likely they will find and fix issues and battle test it more, which might save us the trouble within IOx

pauldix:
If we're going to use this in IOx, I think for right now it makes sense to have it in this repo, just like we do with the influxdbv2 and infludb_iox clients. However, I think it would also be good to publish it as its own Crate. The beauty of Rust is that we can have the code in the same repo without it completely messing up people that might want to use the Crate on it's own without pulling in IOx. I think Edd has PTSD from working in Go.

mkmik:
I need this in conductor; so we can merge it here and then move it elsewhere

alamb:
Let's merge it!

yusdacra:
This is really great work! I tried using `pbjson` on a project and it seems to work well. Would really appreciate if it was released to crates.io so that we don't have to pull all of this repo :stuck_out_tongue:

tustvold:
@yusdacra I'll look into doing this next week, I still need to finish the serialisation of well known types, but then it should be good to go :+1:

tustvold:
@yusdacra split out to https://github.com/influxdata/pbjson and published to crates.io

PR #2098, opened 2021-07-22
fix: Print all timestamps using RFC3339 format

alamb:
# Rationale
Closes https://github.com/influxdata/influxdb_iox/issues/2071 ("predicate bug")

See rationale on https://github.com/influxdata/influxdb_iox/pull/2096

# Changes:
1. PR switches from the pretty printer in arrow to an IOx specifific pretty printer introduced in https://github.com/influxdata/influxdb_iox/pull/2096
2. Updates tests (so many tests!)

# Plan
- [x] Add a pretty printer which prints timestamps in RFC3339 format (https://github.com/influxdata/influxdb_iox/pull/2096)
- [x] Use the pretty printer everywhere in IOx + update tests (this PR)

PR #2475, opened 2021-09-06
build: feature-gate arrow dep for mgmt client

domodwyer:
This was forcing Conductor to pull in and compile all of the `arrow` crate, adding 22 compilation units - we spent more time compiling arrow than `libconductor`!

Looks like an oversight added in #2098  @alamb ?

---

* build: feature-gate arrow dep for mgmt client (dd88b5ed)

      Avoid pulling in & compiling arrow for consumers of the managemnt client.

      PR #2098 added arrow_util as a dependency of this crate which in turn pulls in
      all of arrow, however it is only used in the flight client.

      This commit feature-gates arrow_util behind "flight" with the rest of the
      arrow deps.

### DIFF
--- a/influxdb_iox_client/Cargo.toml
+++ b/influxdb_iox_client/Cargo.toml
@@ -5,12 +5,12 @@ authors = ["Dom Dwyer <dom@itsallbroken.com>"]
 edition = "2018"
 
 [features]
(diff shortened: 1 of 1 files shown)


mkmik (approved):

e-dard (approved):

alamb:
definitely an oversight. Sorry @domodwyer . Thanks for the fix

PR #2479, opened 2021-09-07
fix: perf broken after Protobuf changes

jacobmarble:
Broken at #2458

Also adding CI check to prevent regression


perf/perf.py
```diff
@@ -461,7 +466,15 @@ def grpc_create_database(router_id, writer_id):
             },
             'routing_config': {'sink': {'kafka': {}}},
             'worker_cleanup_avg_sleep': '500s',
-            'writing': '127.0.0.1:9093',
+            'write_buffer_connection': {
+                'direction': 1,
```
crepererum, line 470: IIRC you could also use a string here:

```suggestion
                'direction': 'DIRECTION_WRITE',
```

perf/perf.py
```diff
@@ -594,11 +615,11 @@ def run_test_battery(battery_name, router_id, writer_id, debug=False):
 
         if not response.ok:
             print(response.reason)
-            print(response.content.text)
+            print(response.content.decode('UTF-8'))
```
crepererum, line 618:
```suggestion
            print(response.text)
```

perf/perf.py
```diff
@@ -594,11 +615,11 @@ def run_test_battery(battery_name, router_id, writer_id, debug=False):
...
             print()
             continue
 
-        got = response.content.text.strip()
+        got = response.content.decode('UTF-8').strip()
```
crepererum, line 622:
```suggestion
        got = response.text
```

.circleci/config.yml
```diff
@@ -152,6 +152,18 @@ jobs:
           command: cargo test --no-default-features --features=heappy end_to_end
       - cache_save
 
+  test_perf:
+    machine: true
+    resource_class: xlarge
+    steps:
```
crepererum, line 158: From looking at the CI results: I think we need a newer python version here. `python3` seems to default to version 3.5 which is too old. Could we at least get a 3.8?

crepererum (changes requested):

perf/perf.py
jacobmarble: I reverted this because it was failing with some version of Python 3.

.circleci/config.yml
jacobmarble: The CircleCI image has up to 3.7.0. I'm switching to that in the current iteration. If that doesn't work, then I'll have it install 3.9.x, which will take more time on every execution.

jacobmarble: So many things. Now using Python 3.8, as packaged with Ubuntu 20.04.

perf/perf.py
jacobmarble: Python 3.8, as installed on the most recent CircleCI/Ubuntu 20.04 image, yields this error:
```
'bytes' object has no attribute 'text'
```

crepererum (approved):

PR #2480, opened 2021-09-08
fix: order chunks correctly during query processing

### LINKED ISSUE #1963: Chunk Ordering
## Problem

When resolving duplicates, deletes and upserts, etc... it is necessary to establish an unambiguous ordering of which came later.

In single instance mode this order is established by the order data is written to the open chunk in the partition.

In multiple instance mode, there is a requirement for all replicas to agree on the same order. For this reason an ordering broker, such as Kafka, provides a total ordering (it is assumed potentially overlapping writes are not spread across Kafka partitions). Data is then written to the open chunk in this order.

Currently the assumption made by the catalog/query is that the chunk IDs establish an ordering of writes, with later chunks having higher chunk IDs. With the introduction of compaction and split-based persistence this is no longer the case:

```
Chunk 0: Open
---
Chunk 0: Frozen
Chunk 1: Open
---
Chunk 0: Compacting
Chunk 1: Open
---
Chunk 1: Open
Chunk 2: Compacted Chunk 0
```

```
Chunk 0: Open
---
Chunk 0: Frozen
Chunk 1: Open
---
Chunk 0: Compacting
Chunk 1: Open
---
Chunk 0: RUB
Chunk 1: Open
---
Chunk 0: Split for persist
Chunk 1: Frozen
---
Chunk 1: Frozen
Chunk 2: (Split from 0) Persisting
Chunk 3: RUB
```

## Proposal

* Associate an integer generation with each chunk
* Each new open chunk within a partition is assigned a generation greater than any previous chunk in the partition
* When compacting/splitting, etc... the new chunk(s) should have the minimum generation of the input chunk(s)
* When deduplicating order chunks based on their generation
* Compaction/persistence must not skip over generations, i.e. cannot compact chunks with generations 1 and 5 if there is an unpersisted chunk with generation 3

This relies on the property that compaction and persistence must deduplicate, and it is therefore sufficient that they preserve the order of their inputs w.r.t to the chunks in the partition not included in the operation.

crepererum:
Closes #1963.


server/src/db/lifecycle/compact.rs
```diff
@@ -58,6 +64,8 @@ pub(crate) fn compact_chunks(
                 .map(|prev_last| prev_last.max(candidate_last))
                 .or(Some(candidate_last));
 
+            min_order = min_order.min(chunk.order());
```
alamb, line 67: Does assigning the `min` of the chunks that are compacted produce the correct result

For example, if the the following scenario is possible it would not be correct:

```
chunk order 1 <-- chosen to compact
chunk order 2 
chunk order 3 <-- chosen to compact
```

After compaction, the order would be
```
chunk order 1 <-- contains deduplicated data from 1 and 3
chunk order 2 
```

Which will have effectively made data in chunk 3 look as though it comes before data in chunk 2.

If compaction never "skips" chunks like this then it will be ok. 

cc  @tustvold

server/src/db/system_tables/chunks.rs
```diff
@@ -184,17 +192,18 @@ mod tests {
...
-            "| 0  | p1            | table1     | OpenMutableBuffer |                              | 23754        |                    | 11        |                      | 1970-01-01T00:00:10Z | 1970-01-01T00:00:10Z |             |",
-            "| 1  | p1            | table1     | OpenMutableBuffer | Persisting to Object Storage | 23455        |                    | 22        | 1970-01-01T00:12:34Z | 1970-01-01T00:01:20Z | 1970-01-01T00:01:20Z |             |",
-            "| 2  | p1            | table1     | ObjectStoreOnly   |                              | 1234         | 5678               | 33        | 1970-01-01T00:00:05Z | 1970-01-01T00:01:40Z | 1970-01-01T00:03:20Z |             |",
-            "+----+---------------+------------+-------------------+------------------------------+--------------+--------------------+-----------+----------------------+----------------------+----------------------+-------------+",
+            "+----+---------------+------------+-------------------+------------------------------+--------------+--------------------+-----------+----------------------+----------------------+----------------------+-------------+-------+",
```
alamb, line 200: ❤️  for adding to the system tables

server/src/db/lifecycle/compact.rs
tustvold: > If compaction never "skips" chunks like this then it will be ok.

Yes, this is required for this to be correct and is not currently the case. Persistence, however, should be fine as by formulation it must capture all potentially overlapping rows (it flushes all less than a given timestamp)

crepererum: I take care of that...

NGA-TRAN:
I start reviewing this. Sorry it was late yesterday when I finished the delete parsing

query_tests/cases/in/chunk_order.expected
```diff
@@ -0,0 +1,7 @@
+-- Test Setup: ChunkOrder
```
NGA-TRAN, line 1: Since this is a chunk ordering test, should we have more data in the table to have at least 2 chunks?

lifecycle/src/policy.rs
```diff
@@ -231,18 +231,35 @@ where
             return;
         }
 
-        let mut chunks = LockablePartition::chunks(&partition);
-        // Sort by chunk ID to ensure a stable lock order
-        chunks.sort_by_key(|x| x.0);
+        let chunks = LockablePartition::chunks(&partition);
+
+        // Also sort chunks by:
+        // 1. order: to ensure we compact closed order ranges during persistence
```
alamb, line 237: maybe something like `order: ensure compacting chunks with any potential updates together`

query_tests/src/scenarios.rs
```diff
@@ -1110,3 +1111,70 @@ where
...
+            .clear_lifecycle_action()
+            .unwrap();
+
+        let scenario = DbScenario {
+            scenario_name: "chunks where chunk ID alone cannot be used for ordering".into(),
+            db,
+        };
+
+        vec![scenario]
+    }
```
NGA-TRAN, line 1179: I may miss something but I cannot see how the chunk order helps in this test. Maybe a comment where it helps?

NGA-TRAN (approved):
Very nice.

lifecycle/src/policy.rs
crepererum: fixed

lifecycle/src/policy.rs
```diff
@@ -231,18 +231,35 @@ where
...
+        // 2. ID: for a stable lock order
+        let mut chunks: Vec<_> = chunks
+            .into_iter()
+            .map(|(id, chunk)| {
+                let order = chunk.read().order();
+                (order, id, chunk)
```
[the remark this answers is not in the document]
crepererum: If you persist a single chunk, it will be split row-wise based on the flush timestamp:

- the data that is old enough and will be persisted
- the data that is not old enough yet and that we have to keep in-memory (this chunk is optional since there might be no data here)

These two chunks are new and get new, separate IDs. However their order is the same as the original chunks (in case you're wondering: if you persist multiple chunks the minimum order will be used). So we need to sort by both: the order (to establish the order) and the ID (to establish determinism and a stable lock order).

query_tests/cases/in/chunk_order.expected
crepererum: It has two (look at the asserts in the scenario): one "mutable buffer" and one "read buffer and object store".

query_tests/src/scenarios.rs
crepererum: I've added a comment, let me know if that helps.

lifecycle/src/policy.rs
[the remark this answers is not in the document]
tustvold: Not reviewed the rest of the PR yet, I will do later today, but acquiring locks and holding them in any order other than chunk ID could result in deadlock, and so I think this code is potentially not quite right

lifecycle/src/policy.rs
```diff
@@ -387,10 +404,20 @@ where
...
+            })
+            .collect();
+        chunks.sort_by_key(|(order, id, _chunk)| (*order, *id));
```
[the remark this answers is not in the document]
crepererum: done

lifecycle/src/policy.rs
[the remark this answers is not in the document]
crepererum: Let me sort them twice then...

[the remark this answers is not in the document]
crepererum: Now everything is sorted by `order, ID` (last commit).

query_tests/src/scenarios.rs
```diff
@@ -1110,3 +1111,85 @@ where
...
+
+        // create chunk 1: data->MUB
+        write_lp(&db, "cpu,region=west user=2 100").await;
+        assert_eq!(count_mutable_buffer_chunks(&db), 1);
+        assert_eq!(count_read_buffer_chunks(&db), 1);
+        assert_eq!(count_object_store_chunks(&db), 0);
+
+        // prevent chunk 1 from being part of the persistence
+        // NOTE: In "real life" that could happen when writes happen while a persistence is in progress, but it's easier
+        //       to trigger w/ this tiny locking trick.
```
tustvold, line 1146: Nice :ok_hand:

query_tests/src/scenarios.rs
```diff
@@ -1110,3 +1111,85 @@ where
...
+        let lockable_chunk = db.lockable_chunk(table_name, partition_key, 1).unwrap();
+        lockable_chunk
+            .chunk
+            .write()
+            .set_dropping(&Default::default())
+            .unwrap();
+
+        // transform chunk 0 into chunk 2 by persisting
```
tustvold, line 1154: You could assert the IDs of the returned chunks just in case we change the way the id assignment, etc... works

lifecycle/src/policy.rs
```diff
@@ -1597,9 +1653,6 @@ mod tests {
                 MoverEvents::Persist(vec![4, 5]),
                 MoverEvents::Persist(vec![11, 12]),
                 MoverEvents::Persist(vec![15, 16]),
-                // 17 is the resulting chunk from the persist split above
-                // This is "quirk" of TestPartition operations being instantaneous
-                MoverEvents::Compact(vec![14, 17])
```
tustvold, line 1602: I'm not sure why this has dissappeared and that concerns me?

server/src/db/catalog.rs
```diff
@@ -156,20 +156,20 @@ impl Catalog {
             })
     }
 
-    /// Get a specific chunk returning an error if it can't be found
+    /// Get a specific chunk and its order returning an error if it can't be found
```
tustvold, line 159: I wonder if perhaps there should be a chunk_with_order, and this method left unchanged and just proxying to that. I imagine the vast majority of callsites won't need/want the chunk order

server/src/db/catalog/partition.rs
```diff
@@ -135,35 +140,50 @@ impl Partition {
...
         time_of_last_write: DateTime<Utc>,
         schema: Arc<Schema>,
-    ) -> Arc<RwLock<CatalogChunk>> {
-        let chunk_id = self.next_chunk_id;
-        assert_ne!(self.next_chunk_id, u32::MAX, "Chunk ID Overflow");
-        self.next_chunk_id += 1;
+        chunk_order: u32,
+    ) -> (u32, Arc<RwLock<CatalogChunk>>) {
+        let chunk_id = Self::pick_next(&mut self.next_chunk_id, "Chunk ID Overflow");
+        assert!(
```
tustvold, line 181: :ok_hand:

server/src/db/catalog/partition.rs
```diff
@@ -250,31 +297,42 @@ impl Partition {
...
 
-    /// Return a iterator over chunks in this partition with their
-    ///  ids.
+    /// Return chunks in this partition with their order and ids.
     ///
-    /// Note that chunks are guaranteed ordered by chunk ID.
-    pub fn keyed_chunks(&self) -> impl Iterator<Item = (u32, &Arc<RwLock<CatalogChunk>>)> {
-        self.chunks.iter().map(|(a, b)| (*a, b))
+    /// Note that chunks are guaranteed ordered by chunk order and ID.
+    pub fn keyed_chunks(&self) -> Vec<(u32, u32, &Arc<RwLock<CatalogChunk>>)> {
```
tustvold, line 328: Definitely does not need to be in this PR, but I wonder if we should be introducing newtypes for chunk id and order, just to avoid mixups

server/src/db/catalog/partition.rs
```diff
@@ -250,31 +297,42 @@ impl Partition {
...
+        let mut chunks: Vec<_> = self
+            .chunks
+            .iter()
+            .map(|(id, (order, chunk))| (*id, *order, chunk))
+            .collect();
+        chunks.sort_by_key(|(id, order, _chunk)| (*order, *id));
```
tustvold, line 334: It seems a touch unfortunate to be sorting the chunks every time they're accessed, and also having to buffer everything in Vec. 

Perhaps we could key the BTreeMap by chunk id and sort order, and then have an additional HashMap to get the chunk order for a given chunk id... Perhaps something for a future PR though

server/src/db/lifecycle.rs
```diff
@@ -64,6 +64,8 @@ pub struct WeakDb(pub(super) Weak<Db>);
 pub struct LockableCatalogChunk {
     pub db: Arc<Db>,
     pub chunk: Arc<RwLock<CatalogChunk>>,
+    pub id: u32,
```
tustvold, line 67: As far as I can tell there is a lot of additional plumbing being added to this PR, to return chunk orders, etc... that largely serves to be able to populate these fields. However, these fields don't appear to be necessary, as the actual sorting is performed in `keyed_chunks` within the catalog, and the actual lifecycle actions will have acquired the locks and so can just ask the CatalogChunk directly?

lifecycle/src/lib.rs
```diff
@@ -185,6 +191,22 @@ pub trait LifecycleChunk {
...
+    fn order(&self) -> u32;
+}
+
+/// Sort chunks for linear processing.
+///
+/// Sort chunks by:
+///
+/// 1. order: ensure compacting chunks with any potential updates together
+/// 2. ID: for a stable lock order
+pub fn sort_chunks<C>(mut chunks: Vec<C>) -> Vec<C>
```
tustvold, line 204: This seems to only be used in TestPartition? Given this perhaps it could just be inlined there? Otherwise I think this perhaps needs a less ambiguous name, sort_chunks might to the idle observer seem to suggest something to do with the chunks actual data ordering

tustvold (approved):
This looks good to me, I like that the catalog produces chunks in the "correct" order as I think this will avoid a whole host of silly bugs. That being said I have some comments that I think could reduce the "volume" of this PR

lifecycle/src/lib.rs
crepererum: this is a leftover from an intermediate state, fixed.

query_tests/src/scenarios.rs
crepererum: done

server/src/db/lifecycle.rs
crepererum: I found it useful to have a lock-free ID and order attached to the lockable chunk, because it would allow API users to re-sort the chunks if they want (aka if there's an intermediate step that messes up the ordering), or if they wanna log information w/o locking the chunk. The same argument BTW also holds for ID which isn't really used.

server/src/db/catalog.rs
crepererum: Will do that as part of a future PR that also implements the btree-map idea you had (aka do the sorting via a btree-map and use a hashmap for ID->order).

lifecycle/src/policy.rs
crepererum: @tustvold easy: chunk 13 is not persisted but has "compacting" lifecycle action attached (created by the test fixture itself, not created by the policy). because we don't wanna "jump" the order, we then refuse to check any other chunks for compaction. Aka: the check was overly strict, that's now fixed by the last commit.

server/src/db/catalog.rs
tustvold: I went ahead and did this as part of https://github.com/influxdata/influxdb_iox/pull/2634

PR #2481, opened 2021-09-08
feat: migrate lock metrics to metric crate

tustvold:
This PR is mostly plumbing, but migrates the lock tracking to use the new metric crate. A follow up PR will migrate the rest of the catalog


server/src/db/catalog/table.rs
```diff
@@ -197,20 +197,18 @@ impl<'a> TableSchemaUpsertHandle<'a> {
...
     fn test_handle_no_change() {
-        let lock_tracker = LockTracker::default();
         let table_schema_orig = SchemaBuilder::new()
             .measurement("m1")
             .influx_column("tag1", InfluxColumnType::Tag)
             .influx_column("tag2", InfluxColumnType::Tag)
             .build()
             .unwrap();
-        let table_schema = lock_tracker.new_lock(Arc::new(table_schema_orig.clone()));
+        let table_schema = RwLock::new(Arc::new(table_schema_orig.clone()));
```
tustvold, line 211: No reason to register these locks with a lock tracker for these tests

server/src/db/load.rs
```diff
@@ -159,6 +166,7 @@ impl LoaderEmptyInput {
         db_name: &str,
         server_id: ServerId,
         metrics_registry: Arc<MetricRegistry>,
+        metrics_registry_v2: Arc<metric::Registry>,
```
tustvold, line 169: Having two different metrics_registry is hopefully a temporary thing

metric/src/lib.rs
```diff
@@ -439,6 +458,19 @@ where
     }
 }
 
+impl<const N: usize> From<[(&'static str, Cow<'static, str>); N]> for Attributes {
```
tustvold, line 461: This allows writing something like `Attributes::from([("foo", string_value.into())])`

metric/src/lib.rs
```diff
@@ -420,16 +437,18 @@ impl Attributes {
...
+        self.0.insert(key, value.into());
+    }
 }
 
-impl<'a, I> From<I> for Attributes
-where
-    I: IntoIterator<Item = &'a (&'static str, &'static str)>,
-{
-    fn from(iterator: I) -> Self {
+impl<'a, const N: usize> From<&'a [(&'static str, &'static str); N]> for Attributes {
```
tustvold, line 447: I had to write this explicitly as the previous blanket impl would prevent any other From derives

tracker/src/lock.rs
```diff
@@ -13,134 +12,97 @@ pub type MappedRwLockReadGuard<'a, T> = lock_api::MappedRwLockReadGuard<'a, RawR
...
-
-    pub fn shared_count(&self) -> u64 {
-        self.inner.shared_count.load(Ordering::Relaxed)
-    }
-
-    pub fn upgradeable_count(&self) -> u64 {
-        self.inner.upgradeable_count.load(Ordering::Relaxed)
-    }
+impl LockMetrics {
+    pub fn new(registry: &metric::Registry, attributes: impl Into<Attributes>) -> Self {
```
tustvold, line 28: I'm not a massive fan of passing this a set of attributes, but I can't see any alternatives so :shrug:

server/src/db.rs
```diff
@@ -3831,75 +3831,70 @@ mod tests {
...
-            .has_metric_family("catalog_lock_total")
-            .with_attributes(&[
+        let shared_chunk = reporter
+            .metric("catalog_lock")
+            .unwrap()
+            .observation(&[
                 ("db_name", "lock_tracker"),
                 ("lock", "chunk"),
                 ("table", "cpu"),
-                ("svr_id", "10"),
```
tustvold, line 3883: This did not seem worth the additional plumbing to get it through to the catalog. All metrics will have the pod ID included as a tag anyway, and so I think this is probably sufficient

metric/src/lib.rs
crepererum: Since you take the array by ref, do you really need to specify the size? Aka couldn't you implement this for `&[...]` instead for `&[...; N]`?

metric/src/lib.rs
crepererum: Here the const generic makes sense :+1:

server/src/db.rs
crepererum: Mmh, that basically pushes this information to our infra and requires metric queries to dissect the pod labels. I'm more a fan of using the information that is available.

server/src/db.rs
```diff
@@ -3831,75 +3831,70 @@ mod tests {
...
                 ("lock", "chunk"),
-                ("svr_id", "10"),
                 ("table", "cpu"),
                 ("access", "shared"),
             ])
-            .counter()
-            .gt(0.07)
             .unwrap();
+        assert!(
+            matches!(shared_chunk_wait, Observation::DurationCounter(d) if d > &Duration::from_millis(70))
```
crepererum, line 3896: why is that condition required?

tracker/src/lock.rs
crepererum: I actually like it: the caller can construct the attributes w/ all the details they like and re-use the same attribute set for multiple metrics.

metric/src/lib.rs
```diff
@@ -420,16 +437,18 @@ impl Attributes {
     pub fn iter(&self) -> std::collections::btree_map::Iter<'_, &'static str, Cow<'static, str>> {
         self.0.iter()
     }
+
+    /// Sets the given key, overriding it if already set
+    pub fn insert(&mut self, key: &'static str, value: impl Into<Cow<'static, str>>) {
```
crepererum, line 442: what's the use case of this public function?

metric/src/lib.rs
```diff
@@ -359,6 +369,13 @@ impl Reporter for RawReporter {
 }
 
 impl RawReporter {
+    /// Returns the observation set for a given metric name if any
+    pub fn metric(&self, metric_name: &str) -> Option<&ObservationSet> {
```
crepererum, line 373: :+1: thank you for simplifying testing!

metric/src/lib.rs
tustvold: That's what I thought also, but I couldn't get it to work... I'll have a further play

metric/src/lib.rs
tustvold: It is used to build up Attributes collections - you can see it being used in both the lock tracker and the catalog metrics. It just avoids duplicating keys

server/src/db.rs
tustvold: It's part of the test, it is testing that it recorded some time spent waiting for the lock

server/src/db.rs
tustvold: I guess my point was that this tag is redundant in the presence of the pod id, both are just ways to identify a single workload with no special semantic value to them. You might as well also include database generation if you're going to include this information, I dunno it just seemed a bit pointless and a non-trivial amount of additional plumbing

metric/src/lib.rs
tustvold: So this is because Rust does not perform the conversion from `&[T;N]` to `&[T]` automatically. So if I remove the const parameter it goes from

```
Attributes::from(&[("foo","bar")])
```

to


```
Attributes::from(&[("foo","bar")][..])
```

It might be worthwhile adding a conversion from a slice but I'd prefer to do that when I have a need for it

crepererum (approved):

server/src/db.rs
alamb: How about double check if any of the existing dashboards for IOx use `svr_id` and if not, we keep this PR as is and file a ticket to re-add `svr_id` if there is a use for it?

server/src/db.rs
alamb: This is the new form of the `gt(0.7)` condition in the previous code, right?

An alternate could be to implement `PartialOrd` on `Observations`, but that might be messy when the types are not the same (e.g. comparing `DurationCounter` to a `U64Counter`)

tracker/src/lock.rs
```diff
@@ -215,7 +175,7 @@ unsafe impl<R: lock_api::RawRwLock + Sized> lock_api::RawRwLock for InstrumentRa
 
     /// Acquires an exclusive lock, blocking the current thread until it is able to do so.
     fn lock_exclusive(&self) {
-        match &self.shared {
+        match &self.metrics {
             Some(shared) => {
```
alamb, line 179: nitpick might be to rename this local from `shared` to `metrics` to match the field name both here and several instances below.

PR #2482, opened 2021-09-08
fix: job metrics

crepererum:
Make the job metrics actually useful.

~**This PR sits on top of #2481 and needs a rebase once that one is merged.**~ (done)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2206,6 +2206,7 @@ dependencies = [
  "metric",
  "observability_deps",
  "prometheus",
+ "test_helpers",
 ]
 
 [[package]]
--- a/metric_exporters/Cargo.toml
+++ b/metric_exporters/Cargo.toml
@@ -11,3 +11,4 @@ metric = { path = "../metric" }
 prometheus = "0.12"
 
 [dev-dependencies] # In alphabetical order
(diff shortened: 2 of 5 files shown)


metric/src/duration.rs
```diff
@@ -133,10 +133,18 @@ impl Default for DurationHistogramOptions {
                 Duration::from_millis(100),
                 Duration::from_millis(250),
                 Duration::from_millis(500),
-                Duration::from_millis(1000),
-                Duration::from_millis(2500),
-                Duration::from_millis(5000),
-                Duration::from_millis(10000),
+                Duration::from_millis(1_000),
```
tustvold, line 136: I'm not sure the defaults should include 5000 seconds? This seems a tad extreme, perhaps the jobs could/should override with a custom DurationHistogramOptions?

tustvold (approved):
I think I'd be wary about increasing the default buckets to such an extreme extent, as they're a good way to dramatically increase our metric cardinality. I don't think we expect requests to be measured in 100s of seconds, and with this change we'd end up a large number of additional endpoints for every HTTP method path

metric/src/duration.rs
crepererum: fixed (moved from default to job-specific option)

PR #2486, opened 2021-09-08
feat: migrate catalog timestamp summary to metric crate

tustvold:
This migrates the timestamp summary within the catalog to the new metric crate


tests/end_to_end_cases/metrics.rs
```diff
@@ -33,7 +33,7 @@ pub async fn test_row_timestamp() {
     let db_name_attribute = format!("db_name=\"{}\"", db_name);
 
     // Should only be enabled for the system table
-    assert_eq!(lines.len(), 60);
+    assert_eq!(lines.len(), 61);
```
alamb, line 36: why did it increase to 61?

server/src/db/catalog/chunk.rs
```diff
@@ -242,7 +242,7 @@ pub struct ChunkMetrics {
     pub(super) memory_metrics: StorageGauge,
 
     /// Track ingested timestamps
-    pub(super) timestamp_histogram: TimestampHistogram,
+    pub(super) timestamp_histogram: Option<TimestampHistogram>,
```
alamb, line 245: why did this need to become optional?

alamb (approved):

tests/end_to_end_cases/metrics.rs
tustvold: Prometheus histograms have to include +Inf. Previously it was actually incorrect :sweat_smile: It is now a legitimate histogram :tada:

server/src/db/catalog/chunk.rs
tustvold: I thought it was less confusing than the "optional" being internal to the thing called TimestampHistogram

crepererum (approved):

Issue #2463, opened 2021-09-03
Delete Predicate API


NGA-TRAN:
This is a part of Support Soft Delete #2396 that defines what Delete API we will support in IOx. The goal is to try to make it compatible with [Influx Delete](https://docs.influxdata.com/influxdb/v2.0/reference/cli/influx/delete/) as much as possible. However, since Flux needs something that IOx does not provide such as `bucket` and `org`, Flux Delete API and IOx Delete API are not equivalent. 

### [IOx Delelete API](https://github.com/influxdata/influxdb_iox/blob/fdfec8fa4f1c91653631a7f0ff016ddf7f0bb64c/generated_types/protos/influxdata/iox/management/v1/jobs.proto#L154) includes:

1. `db_name` 
2. `table_name` :  equivalent to Flux's `measurement` but Flux specifies measurement in the `predicate` below instead of providing it in a separated field. @e-dard I am debating where to keep this as a separated field  or we extract it from the `predicate`. What do you think?
3. `delete_predicate` :  equivalent to Flux's `predicate` without measurement (or we can make it with measurement, too). Currently, Flux only supports conjunctive predicate of binary expressions of `=` and `!=` and IOx will stick with this.
4. `start_time` : equivalent to Flux's start
5. `stop_time` : equivalent to Flux's stop

Comparing with Flux's syntax, here is some IOx valid examples:

#### Delete everything from table named `example-measurement`
```
db_name:  your_db_name
table_name: example-measurement   (or we do not have this field at all)
delete_predicate: either empty   or _measurement="example-measurement" depending on our decision to go with separated table_name or not
start_time: 1970-01-01T00:00:00Z
stop_time: $(date +"%Y-%m-%dT%H:%M:%SZ")
```

#### Delete points in a measurement with a specific tag value
```
db_name:  your_db_name
table_name: example-measurement
delete_predicate:  host="old-host"  (AND _measurement="example-measurement") -- the one in parentheses depend on design
start_time: 1970-01-01T00:00:00Z
stop_time: $(date +"%Y-%m-%dT%H:%M:%SZ"
```
#### Delete all points within a specified time frame for all tables
```
db_name:  your_db_name
table_name:  empty
delete_predicate: 
start_time: 2020-03-01T00:00:00Z
stop_time: 2020-11-14T00:00:00Z
```

#### Delete all points within a specified time frame for a specified table and its tag and field
```
db_name:  your_db_name
table_name:  example-measurement
[...]

PR #2488, opened 2021-09-08
feat: Finally have the delete predicate parsed and invoked from client

### LINKED ISSUE #2478: Use sqlparser-rs to parse delete predicate
# Rationale
Even though IOx has its own [Delete API](https://github.com/influxdata/influxdb_iox/issues/2463) to delete data that does not need to go through SQL Delete, its `delete predicate` has the same syntax with SQL Delete and this ticket is to use https://github.com/sqlparser-rs/sqlparser-rs to parse IOx delete predicate.

In the first version, we will support all syntax supported in Flux defined in #2463

NGA-TRAN:
Closes #2478 #2463 

Another task of #2396 

Use sqlparser-rs to parse the delete predicate


generated_types/src/parse_delete.rs
```diff
@@ -1,103 +0,0 @@
-//! Conversion code to management Delete API structures and vice versa
```
NGA-TRAN, line 1: Since I decided, instead of parsing string at client side, I do it at at server side to avoid extra post-parser work to convert already parsed string to Expression at server side, all the pluming in this file no longer needed

influxdb_iox_client/src/client/management.rs
```diff
@@ -967,44 +960,30 @@ impl Client {
...
         let start_time = start_time.into();
         let stop_time = stop_time.into();
+        let predicate = predicate.into();
 
-        // parse the time range and predicate
-        let provided_parse_delete_result = ProvidedParseDelete::try_new(
-            start_time.as_str(),
-            stop_time.as_str(),
-            predicate.as_str(),
-        );
```
NGA-TRAN, line 985: This now will be done at server side

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -9,11 +9,20 @@
     clippy::clone_on_ref_ptr
```
NGA-TRAN, line 9: My plan right after this PR is merged is to move all content of this file into int [predicate.rs](https://github.com/influxdata/influxdb_iox/blob/29b10413051f8c4a2193e8633aa133e45b0e505a/query/src/predicate.rs#L1)

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -29,83 +38,188 @@ pub enum Error {
...
 
 /// Parser for Delete predicate and time range
-#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Serialize, Deserialize)]
-pub struct ProvidedParseDelete {
+#[derive(Debug, Clone)]
+pub struct ParseDeletePredicate {
     pub start_time: i64,
     pub stop_time: i64,
     // conjunctive predicate of binary expressions of = or !=
-    pub predicate: Vec<ProvidedDeleteBinaryExpr>,
```
NGA-TRAN, line 43: Since I reuse DF's Expr, I do not need these structs anymore

src/influxdb_ioxd/rpc/management.rs
```diff
@@ -581,49 +580,51 @@ where
...
-        let provided_parse_delete: ProvidedParseDelete = parse_delete
-            .try_into()
-            .map_err(|e: FieldViolation| e.scope("parse_delete"))?;
-
         // Validate that the database name is legit
         let db_name = DatabaseName::new(db_name).field("db_name")?;
         let db = self
             .server
             .db(&db_name)
             .map_err(default_server_error_handler)?;
```
NGA-TRAN, line 593: This is the parse code done in server

crepererum:
I'm a bit worried about the approach here: To me it looks that this PR moves the parsing issue from the client to the server and now uses two opaque types (a string during the client->server communication and a datafusion `Expr` withing the server) instead of a hand rolled, transparent expression type. However at one point we need to have a stable, IOx-controlled serialization format for these expressions, because we need to store them within the preserved catalog. So my main question is the following: can a datafusion `Expr`

1. easily be converted into a protobuf message
2. easily constructed by hand (out of our catalog protobufs)?
3. would these two operations easily survive datafusion refactors and upgrades?

influxdb_line_protocol/src/delete_parser.rs
alamb: Yes - I agree with this plan.  I don't think this code belongs in the `influxdb_line_protocol` crate at all (and the line protocol crate shouldn't have to depend on sqlparser or datafusion)

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -29,83 +38,188 @@ pub enum Error {
...
+    /// "column_name = literal" or "column_name != literal"
+    ///
+    /// The split expressions will be converted into data fusion expressions
+    pub fn split_members(
+        table_name: &str,
+        predicate: &SqlParserExpr,
+        predicates: &mut Vec<Expr>,
+    ) -> bool {
+        // Th below code built to be compatible with
+        //  https://github.com/influxdata/idpe/blob/master/influxdbv2/predicate/parser_test.go
```
alamb, line 150: FWIW I don't think idpe is a public repo (so we should perhaps not have references to it in a public repo). 

Perhaps https://github.com/influxdata/influxdb/blob/master/predicate/parser_test.go would be a better reference

NGA-TRAN:
@crepererum 
> I'm a bit worried about the approach here: To me it looks that this PR moves the parsing issue from the client to the server and now uses two opaque types (a string during the client->server communication and a datafusion Expr withing the server) instead of a hand rolled, transparent expression type. However at one point we need to have a stable, IOx-controlled serialization format for these expressions, because we need to store them within the preserved catalog. So my main question is the following: can a datafusion Expr

>easily be converted into a protobuf message
>easily constructed by hand (out of our catalog protobufs)?
>would these two operations easily survive datafusion refactors and upgrades?

The parsing was built completely at client side on Tuesday but I discovered it was a draw back doing so and spent yesterday to convert it back to server side. Here are the reasons:
1. The parsing is very lightweight and I bet the SELECT parsing  also done at server or even after that down in DF. @alamb can you confirm?
2. If we want to run query at server only, we need to parse it at server anyway. We do not want to have duplicate code here.
3. The management API to communicate between client and server is way a lot more complicated with a lot of code that already a lot and I need to add more to keep the messages to convert back to corresponding structs at server side. 
4. Related to 4 that double data structures need to build maintain at both client and server. Furthermore, DF has Expr enum and all related structs that is ready for us to use.

All those 4 reasons made me have a strong opinion to spend a day to redo the work I did the previous day to keep things simple and reuse a lot of code without satisfying performance.

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -29,83 +38,188 @@ pub enum Error {
...
+                };
+
+                // verify if right is a literal
+                let value = match &**right {
+                    SqlParserExpr::Identifier(Ident {
+                        value,
+                        quote_style: _,
+                    }) => Expr::Literal(ScalarValue::Utf8(Some(value.to_string()))),
+                    SqlParserExpr::Value(Value::DoubleQuotedString(value)) => {
+                        Expr::Literal(ScalarValue::Utf8(Some(value.to_string())))
```
alamb, line 191: FWIW you can use `use datafusion::prelude::lit` to write this code more concisely like

```suggestion
                        lit(value.to_string())
```

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -29,83 +38,188 @@ pub enum Error {
...
+                    }) => Expr::Column(Column {
+                        relation: Some(table_name.to_string()),
+                        name: value.to_string(),
+                    }),
+                    _ => return false, // not a column name
+                };
+
+                // verify if right is a literal
+                let value = match &**right {
+                    SqlParserExpr::Identifier(Ident {
```
alamb, line 186: this code also allows right to be an identifier (e.g. `colA = colB`) I think, but the comment says it is verifying that right is a literal

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -29,83 +38,188 @@ pub enum Error {
...
+                    SqlParserExpr::Value(Value::NationalStringLiteral(value)) => {
+                        Expr::Literal(ScalarValue::Utf8(Some(value.to_string())))
+                    }
+                    SqlParserExpr::Value(Value::HexStringLiteral(value)) => {
+                        Expr::Literal(ScalarValue::Utf8(Some(value.to_string())))
+                    }
+                    SqlParserExpr::Value(Value::Number(v, _)) => {
+                        Expr::Literal(ScalarValue::Float64(Some(v.parse().unwrap())))
+                    }
+                    // NGA todo: how to now this is an integer?
```
alamb, line 205: I don't understand this comment -- a boolean will not be an integer I don't think

query/Cargo.toml
```diff
@@ -29,6 +29,7 @@ metrics = { path = "../metrics" }
 parking_lot = "0.11.2"
 regex = "1"
 snafu = "0.6.3"
+thiserror = "1.0.28"
```
alamb, line 32: I wonder why this was added? I don't think it is needed

src/influxdb_ioxd/rpc/management.rs
alamb: 🎉

crepererum:
>  Furthermore, DF has Expr enum and all related structs that is ready for us to use and do not need to build both of them.

That alone doesn't provide any long-term stability that we need for preservation.

NGA-TRAN:
>> Furthermore, DF has Expr enum and all related structs that is ready for us to use and do not need to build both of them.

> That alone doesn't provide any long-term stability that we need for preservation.

@crepererum Sorry I was not clear. What I meant was DF Expr was used everywhere in IOx and if we want to modify Delete Expr, we want to modify DF Expr instead. So the whole point here is instead doing all conversion between protobuf to new structs, we do conversion from protobuf to DF Expr, nd vice versa as needed.

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: Good catch

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: This is a great catch. I need to add tests to see how this work because:
1. From my test of "how to use sqlparser" below, the literal "Boston" was stored in an Identifier, and
2. I am not sure we want to support `colA= colB` in delete. Need to verify with Influx delete case.

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: @alamb This comment is for a missing line of integer case. I do not know how to do it yet. If you know, let me know.

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -189,76 +306,233 @@ mod test {
...
     #[test]
     fn test_parse_timestamp_out_of_range() {
         let input = r#"99999999999999999999999999999999"#;
-        let time = ProvidedParseDelete::parse_time(input);
+        let time = ParseDeletePredicate::parse_time(input);
         assert!(time.is_err());
     }
+
+    #[test]
+    fn test_sqlparser() {
```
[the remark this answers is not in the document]
NGA-TRAN: Will remove it before merging

query/Cargo.toml
NGA-TRAN: I might have removed code that used it. Will remove this dependency. I will go over my previous PRs that added dependency at client and line protocol crates that no longer needed after the code is remove by this PR and remove them.

influxdb_line_protocol/src/delete_parser.rs
alamb: @NGA-TRAN  the way DataFusion does it is to try and parse the number as an `i64` first and if that fails fall back to trying to parse as a `f64`:  https://github.com/apache/arrow-datafusion/blob/master/datafusion/src/sql/planner.rs#L1053

```rust
            SQLExpr::Value(Value::Number(n, _)) => match n.parse::<i64>() {
                Ok(n) => Ok(lit(n)),
                Err(_) => Ok(lit(n.parse::<f64>().unwrap())),
            },
```

influxdb_line_protocol/src/delete_parser.rs
NGA-TRAN: I have fixed the comment to allow both identifier and literal. And they are consistent with influx delete parser tests

PR #2489, opened 2021-09-09
feat: remove deprecated catalog metrics

tustvold:
`catalog_chunks_total` and `catalog_chunk_creation_size_bytes` date from the very first pass at adding catalog size instrumentation, and with the changes to a partition-based as opposed to chunk-based lifecycle are now very difficult to interpret, and have been largely superseded by `catalog_loaded_chunks` and `catalog_chunk_mem_usage`. They were kept for backwards compatibility but rather than migrating them to the new `metric` crate I think they should just be removed, and some potential confusion eliminated.

`catalog_chunks_total` records the total the number of chunks that have moved through certain phases in the lifecycle. As we now compact and upsert a new chunk, this is of limited utility. `catalog_loaded_chunks` and `catalog_loaded_rows` provide a way to debug the lifecycle, that is easier to interpret (they are gauges that reflect the current state).

`catalog_chunk_creation_size_bytes` records the size of chunks being created in the catalog as a histogram. As the lifecycle will iteratively build successively larger chunks until it persists them, this is also of limited utility. Most of its remaining functionality of ascertaining where memory is being used is provided by `catalog_chunk_mem_usage`.


server/src/db.rs
```diff
@@ -2594,37 +2521,18 @@ mod tests {
...
-            .has_metric_family("catalog_chunk_creation_size_bytes")
-            .with_attributes(&[
-                ("db_name", "placeholder"),
-                ("state", "moved"),
-                ("svr_id", "1"),
-            ])
-            .histogram()
-            .sample_sum_eq(3834.0)
-            .unwrap();
+        catalog_chunk_size_bytes_metric_eq(&test_db.metric_registry, "mutable_buffer", 0);
```
tustvold, line 2534: This is a good example of why this metric is really hard to interpret with the lifecycle as it is at the moment. I honestly have no idea where 5085 is coming from...

e-dard:
I have a couple of questions, which may or may not have been answerable with the previous metrics, but would be good to answer with the replacements.

1) How best would we currently measure the rate at which chunks are moving through the system from one location to the next?
2) How do we assess the distribution of chunk sizes over time? For example if I wanted to know how the p90 chunk size is changing over a time period would I be able to do that with the observabilty we're adding back in?

tustvold:
> How do we assess the distribution of chunk sizes over time? For example if I wanted to know how the p90 chunk size is changing over a time period would I be able to do that with the observabilty we're adding back in?

I think it would help if I understood the question this observability is supposed to answer? The size distribution of chunks as a whole is a bit meaningless, as they may contain vastly different rows with different data distributions (the histogram is aggregated DB wide). I could understand the value of more granular reporting within the RBChunk or ParquetChunk, but at the catalog level I'm not really sure what this would tell us?

> How best would we currently measure the rate at which chunks are moving through the system from one location to the next?

catalog_loaded_rows tells you the number of rows in each storage location, so you could, for example, compute the rate of increase in the persisted state. You could do the same with `catalog_loaded_chunks`, but I'm not sure how useful this would be given the chunks could be vastly different sizes.

tustvold:
I think I would like to move forward with this PR unless there are strong objections, as there are other PRs getting backed up on it. If there are gaps in the observability provided by the catalog metrics, I'd prefer we design and implement a fix for that as a separate piece of work, as I am confident the metrics being removed cannot easily be used to gain insights into IOx with a fairly good understanding of what they are measuring and how the lifecycle works, and even then I'm unconvinced

alamb (approved):
> > If there are gaps in the observability provided by the catalog metrics, I'd prefer we design and implement a fix for that as a separate piece of work, 

Sounds like a good plan to me. If we discover it was critical for some reason we can always add them back in

PR #2490, opened 2021-09-09
feat: migrate remaining catalog metrics to new crate

tustvold:
~~Draft as builds on top of #2489 and #2486~~ this PR migrates the remaining catalog metrics to the new crate


server/src/db/catalog/chunk.rs
```diff
@@ -195,7 +199,9 @@ pub struct CatalogChunk {
     lifecycle_action: Option<TaskTracker<ChunkLifecycleAction>>,
 
     /// The metrics for this chunk
-    metrics: ChunkMetrics,
+    ///
+    /// Wrapped in a mutex to allow updating metrics without exclusive access to CatalogChunk
+    metrics: Mutex<ChunkMetrics>,
```
tustvold, line 204: I pulled the Mutex up a level so as to make it clearer that it is to give CatalogChunk interior mutability, not to provide synchronisation of metric updates (as it doesn't provide this)

server/src/db/catalog/metrics.rs
```diff
@@ -113,20 +104,17 @@ impl CatalogMetrics {
     }
 }
 
+/// Metrics that are collected and aggregated per-table within a database
 #[derive(Debug)]
 pub struct TableMetrics {
-    /// Metrics domain
-    metrics_domain: Arc<metrics::Domain>,
+    catalog_metrics: Arc<CatalogMetrics>,
```
tustvold, line 110: I changed the metrics to reference the parent, as opposed to copying the instruments along. I think this makes it clearer at what level each of the metrics is aggregated, but feedback welcome.

This won't result in reference cycles as the metrics don't have a pre-existing ownership relationship with each other

server/src/db/load.rs
```diff
@@ -177,11 +176,8 @@ impl LoaderEmptyInput {
             KeyValue::new("db_name", db_name.to_string()),
             KeyValue::new("svr_id", format!("{}", server_id)),
         ];
-        let domain =
```
tustvold, line 180: :wave:

metric/src/metric.rs
```diff
@@ -64,12 +64,19 @@ impl<T: MetricObserver> Metric<T> {
...
-            .or_insert_with(|| T::create(&self.shared.options))
-            .recorder()
+        self.observer(attributes).recorder()
+    }
+
+    /// Retrieves the observer for a given set of attributes
+    ///
+    /// If this is the first time this method has been called with this set of attributes,
+    /// it will initialize the corresponding `MetricObserver` with the default observation
+    pub fn observer(&self, attributes: impl Into<Attributes>) -> MappedMutexGuard<'_, T> {
```
crepererum, line 74: What is the difference between an observer and a recorder?

crepererum (approved):
PR looks good, I just start to loose track of all the terminology.

metric/src/metric.rs
tustvold: In most cases, they're the same type. `Metric` stores and is typed by `MetricObserver`. This then has a method `recorder()` which returns a recorder for that metric type. In almost all cases this is the same type, i.e. `MetricObserver::Recoder = Self`, an exception is `CumulativeGauge` where the returned recorder is instead `CumulativeRecorder` which has the drop semantics.

PR #2494, opened 2021-09-09
feat: add RecorderCollection

tustvold:
I think the docstring should be clear, but I want to use this to migrate the metrics within the read buffer. Effectively `metric::MetricRecorder<CumulativeGauge` is the equivalent of `metrics::Gauge` _(note the different crate)_


alamb (commented):
The docstring is 💯  thank you for that.

My biggest suggestion is to try and find a less confusing (at least to me)  name for `MetricRecorder` (as creating a `MetricRecorder` which must then have call `recorder()` called to give a `T::Recorder`(which is different than `MetricRecorder`) is somewhat  🤯 

How about something like `StatefulMetric` or `StatefulRecorder`, or perhaps a more generic `MetricHandle`?

alamb (approved):

crepererum (approved):
Change looks good, docstring is awesome. I however share the naming concern (and confusion about `MetricRecorder::recorder`) w/ @alamb and would also like to see some less confusing name.

tustvold:
@alamb @crepererum What do people thing of `RecorderCollection` as an alternative name?

alamb:
> @alamb @crepererum What do people think of RecorderCollection as an alternative name?


Much better for me 👍

PR #2495, opened 2021-09-09
refactor: remove process clock

crepererum:
The process clock is a leftover from the pre-Kafka writer buffer design
and is no longer required.


domodwyer (approved):
👋

PR #2496, opened 2021-09-09
chore: fix perf again, after another protobuf change

jacobmarble:
After #2474, this broke. Also, CI did not require `test_perf` to pass, so Kodiak merged a broken test.


domodwyer (approved):
Looks legit 👍

crepererum (approved):

PR #2497, opened 2021-09-09
feat: distribute writes over mulitple write buffer sequencers

crepererum:


entry/src/entry.rs
```diff
@@ -1114,6 +1217,112 @@ impl<'a> Column<'a> {
             _ => panic!("unknown fb values type"),
         }
     }
+
+    fn deep_copy<'fbb>(
```
crepererum, line 1221: why is flatbuffer so horrible that I have to manually implement this? :sob:

carols10cents: 😬  https://github.com/google/flatbuffers/issues/264

carols10cents (commented):
Some questions!

entry/src/entry.rs
crepererum: > This issue is stale because it has been open 6 months with no activity. Please comment or this will be closed in 14 days.

I really hate auto-close bots. They encourage spam comments (aka "still an issue", "any progress?!") and make it hard to find active issues (you now have to read through the entire thread to figure out if that issue is fixed or was actively forgotten).

server/src/db.rs
```diff
@@ -1144,11 +1144,21 @@ impl Db {
                 // the write buffer, and it's not an error. We ignore the returned metadata; it
                 // will get picked up when data is read from the write buffer.
 
-                // TODO: be smarter than always using sequencer 0
-                let _ = write_buffer
-                    .store_entry(&entry, 0)
-                    .await
-                    .context(WriteBufferWritingError)?;
+                // split writes into many
```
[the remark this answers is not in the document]
crepererum: Good catch :+1:

entry/src/entry.rs
```diff
@@ -301,6 +307,88 @@ fn build_table_write_batch<'a>(
     ))
 }
 
+pub fn distribute_write_entry(entry: &Entry, n: NonZeroUsize) -> Vec<Option<Entry>> {
+    assert_eq!(entry.fb().operation_type(), entry_fb::Operation::write);
```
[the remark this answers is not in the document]
crepererum: 1. The problem is that `Entry` is somewhat of an container type, which makes typing a bit harder.
2. I think I should probably use `unimplemented` in this case, because we'll have "delete" messages soon and these we probably should distribute in a different way (maybe to all sequencers?!).

PR #2498, opened 2021-09-09
feat: enable tracing for http API (#2273)

tustvold:
This requires a bit of a tower dance, but the benefit is that most of the actual logic can be shared with gRPC services :tada:


src/influxdb_ioxd/http.rs
```diff
@@ -950,15 +953,16 @@ pub async fn serve<M>(
...
     shutdown: CancellationToken,
     max_request_size: usize,
+    trace_collector: Option<Arc<dyn TraceCollector>>,
 ) -> Result<(), hyper::Error>
 where
     M: ConnectionManager + Send + Sync + Debug + 'static,
 {
     let router = router(application, server, max_request_size);
-    let service = RouterService::new(router).unwrap();
+    let new_service = tower::MakeService::new(router, trace_collector);
```
alamb, line 962: this is fairly beautiful

src/influxdb_ioxd/http/tower.rs
```diff
@@ -0,0 +1,71 @@
...
+        collector: Option<Arc<dyn TraceCollector>>,
+    ) -> Self {
+        Self {
+            inner: RouterService::new(router).unwrap(),
+            trace_layer: TraceLayer::new(collector),
+        }
+    }
+}
+
+impl tower::Service<&AddrStream> for MakeService {
```
alamb, line 39: 🎩  to you for figuring this out -- I find the maze of traits very hard to understand when working with tower.

src/influxdb_ioxd/http/tower.rs
```diff
@@ -0,0 +1,71 @@
...
+    type Response = Service;
+    type Error = Infallible;
+    type Future = MakeServiceFuture;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, conn: &AddrStream) -> Self::Future {
```
alamb, line 48: This appears to be the equivalent of `async_trait`, is that correct?

In other words, if using `async_traint` this might look something like 

```rust
async call(&mut self, conn: &AddrStream) -> Result<Self::Response, Self::Error>
```

? (I am just checking my understanding)

alamb (approved):
Looks good to me

> This requires a bit of a tower dance, but the benefit is that most of the actual logic can be shared with gRPC services 🎉

I didn't see any sharing of logic with gRPC services. Perhaps I misunderstand the comment or you plan the work for a subsequent PR

src/influxdb_ioxd/http/tower.rs
tustvold: It's taken me about a year of bashing my head against them, they now are actually starting to make sense to me :laughing:

tustvold:
> I didn't see any sharing of logic with gRPC services. Perhaps I misunderstand the comment or you plan the work for a subsequent PR

The trace logic is already wired up into the gRPC services, the fact you can't see it is why I was so pleasantly surprised - I didn't have to make any changes - it just worked :grin:

src/influxdb_ioxd/http/tower.rs
tustvold: Sort of, what async_trait generates is closer to

```
fn call<'a>(&'a mut self, conn: &AddrStream) -> Pin<Box<dyn Future<Output=Result<Self::Response, Self::Error>>> + 'a>
```

Or equivalently


```
fn call<'a>(&'a mut self, conn: &AddrStream) -> futures::future::BoxFuture<'a, Result<Self::Response, Self::Error>>
```

Unlike async_trait, this does not result in any additional boxing or dynamic dispatch (beyond what is forced by routerify)

src/influxdb_ioxd/http/tower.rs
alamb: 🧙‍♂️

src/influxdb_ioxd/http/tower.rs
alamb: yes, sorry if I was not clear. I am not suggesting this PR try to use `async_trait` -- I am working on my mental model of how rust handles / implements futures and `async`. Thank you for the explination

PR #2499, opened 2021-09-09
feat: Add pre-generated values and tag sets to data generator

pauldix:
This adds the ability to pre-generate values and tag sets in the data generator. This makes it easy to have tags that depend on other tag values (like buckets in an org) and have tag values that have one associated tag (like if something is in production or staging environment). Follow on work will add the actual generation to the agent spec. An added bonus of these pre-generated values is that generating samples won't require any sort of template evaluation for all of the tags in the tag sets. Only unique values (like trace_id or span_id) would need to be generated during sampling generation.

At this point I think I have a specification that is nice and tag data that looks like how I want it to look (see the tests in `tag_set.rs`. Next up I need to update `MeasurementSpec` to be able to take advantage of these tag sets. I'll also optimize how the line protocol is generated so that we can in many cases avoid handlebars evaluation during generation of samples. Likely require a bit of a refactor.

### DIFF
--- a/iox_data_generator/src/specification.rs
+++ b/iox_data_generator/src/specification.rs
@@ -94,7 +94,7 @@ pub struct ValuesSpec {
     /// The name of the collection of values
     pub name: String,
     /// The handlebars template to create each value in the collection
-    pub value: String,
+    pub template: String,
     /// How many of these values should be generated. If belongs_to is
     /// specified, each parent will have this many of this value. So
     /// the total number of these values generated would be parent.len() * self.cardinality
@@ -125,9 +125,10 @@ impl ValuesSpec {
 pub struct TagSetsSpec {
     /// The name of the tag set spec
     pub name: String,
-    /// An array of the values to loop through. To reference parent belongs_to or has_one
-    /// values, the parent should come first and then the has_one or child next. See the
-    /// doc for `ForEachValueTag` and its `value` for more detail.
+    /// An array of the `ValuesSpec` to loop through. To reference parent belongs_to or has_one
+    /// values, the parent should come first and then the has_one or child next. Each successive
+    /// entry in this array is a nested loop. Multiple has_one and a belongs_to on a parent can
+    /// be traversed.
     pub for_each: Vec<String>,
 }
 
--- a/iox_data_generator/src/specification.rs
+++ b/iox_data_generator/src/specification.rs
@@ -52,6 +52,19 @@ pub struct DataSpec {
     /// future configurations if reproducing a particular set of sequences
     /// is desired.
     pub base_seed: Option<String>,
+    /// Specifies values that are generated before agents are created. These values
+    /// can be used in tag set specs, which will pre-create tag sets that can then be
+    /// used by the agent specs.
+    #[serde(default)]
+    pub values: Vec<ValuesSpec>,
+    /// Specifies collections of tag sets that can be referenced by agents. These
+    /// pre-generated tag sets are an efficient way to have many tags without
+    /// re-rendering their values on every agent generation. They can also have
+    /// dependent values, making it easy to create high cardinality data sets
+    /// without running through many handlebar renders while having a well defined
+    /// set of tags that appear.
+    #[serde(default)]
+    pub tag_sets: Vec<TagSetsSpec>,
     /// The specification for the data-generating agents in this data set.
     pub agents: Vec<AgentSpec>,
 }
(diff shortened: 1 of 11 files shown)


iox_data_generator/src/specification.rs
```diff
@@ -73,6 +86,51 @@ impl FromStr for DataSpec {
...
+/// set of tags in.
+#[derive(Deserialize, Debug)]
+#[cfg_attr(test, derive(Default))]
+#[serde(deny_unknown_fields)]
+pub struct TagSetsSpec {
+    /// The name of the tag set spec
+    pub name: String,
+    /// An array of the values to loop through. To reference parent belongs_to or has_one
+    /// values, the parent should come first and then the has_one or child next. See the
+    /// doc for `ForEachValueTag` and its `value` for more detail.
```
alamb, line 130:
```suggestion
    /// doc for `ValuesSpec ` and its `value` for more detail.
```

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+
+#[derive(Debug, Default)]
+pub struct ParentToHasOnes {
+    // each parent id will have its has_ones stored in this map. The map within
+    // maps the has_one name to its generated value
+    id_to_has_ones: BTreeMap<usize, BTreeMap<Arc<String>, Arc<GeneratedValue>>>,
+}
+
+#[derive(Debug)]
+pub struct TagSetKey {
```
alamb, line 85: this struct doesn't seem to be used anywhere... I wonder if it can be deleted

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug)]
+pub struct GeneratedValueCollection {
+    name: String,
+    values: Vec<GeneratedValue>,
+}
+
+#[derive(Debug)]
+pub struct GeneratedValue {
```
alamb, line 60: I am having a bit of a hard time following what values are for what entity -- for example in this struct `TagPair` contains a `key=value` but then `GeneratedValue` also contains another `value` which I am not sure what it is used for

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+    value: Arc<String>,
+    id: usize,
+    tag_pair: Arc<TagPair>,
+}
+
+#[derive(Debug, Default)]
+pub struct GeneratedTagSets {
+    values: BTreeMap<String, Vec<Arc<GeneratedValue>>>,
```
alamb, line 68: it might help to specify what the `String` keys for each of these maps is for (aka what does the key for `values` mean (as there is no `name` field on `GeneratedValue`).

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+    // each parent-child will have its children stored in this map. The children map
+    // the id of the parent to the collection of its children values
```
alamb, line 70:
```suggestion
    /// each parent-child will have its children stored in this map. The children map
    /// the id of the parent to the collection of its children values
```

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+    child_values: BTreeMap<String, BTreeMap<usize, Vec<Arc<GeneratedValue>>>>,
+    // each parent-has_one will have its has_ones stored in this map
```
alamb, line 72:
```suggestion
    /// each parent-has_one will have its has_ones stored in this map
```

iox_data_generator/src/tag_set.rs
```diff
@@ -0,0 +1,641 @@
...
+name = "f1"
+i64_range = [0, 23]
+"#;
+
+        let spec = DataSpec::from_str(toml).unwrap();
+        let tag_sets = GeneratedTagSets::from_spec(&spec).unwrap();
+        let testage = tag_sets.sets_for("testage").unwrap();
+        let sets = testage.iter().map(|t| t.to_string()).join("\n");
+        let expected = r#"
+foo=1#foo
```
alamb, line 514: I can't say I went over all the code in this module in detail, but I think these tests are really helpful to understand how `values`, `tag_sets` and `belongs_to` are used. 

When the feature is done (or maybe even as part of this PR) a brief example, perhaps cribbed from these tests) showing how to use these settings, including part of the output, would likely really help

alamb (approved):
The code looks good to me. I had some small "niggles" as our friends across the Pond would say but all in all 👍  I think this PR could be merged as is. 

I can't say I fully grok how the `has_some` and `belongs_to` is supposed to work, but I suspect that will get fleshed out in further PRs and docs. 

This is quite cool stuff @pauldix

iox_data_generator/src/tag_set.rs
pauldix: The TOML file included should serve as that example. I'm going to flesh that out a bit more (along with comments) to show how this all hangs together.

iox_data_generator/src/tag_set.rs
pauldix: I added the `TagPair` member along the way to optimize re-using previous data. I can probably remove the `value` here as it's just a dupe of the tag pair's value.

PR #2500, opened 2021-09-09
refactor: move ParseDeletePredicate to predicate.rs

NGA-TRAN:
This is a follow-up of #2488 and a part of #2396 to move the delete predicate to the right place


influxdb_line_protocol/Cargo.toml
```diff
@@ -5,15 +5,10 @@ authors = ["Paul Dix <paul@pauldix.net>"]
 edition = "2018"
 
 [dependencies] # In alphabetical order
-chrono = "0.4"
```
alamb, line 8: ❤️  -- much better!

query/src/predicate.rs
```diff
@@ -11,12 +11,50 @@ use std::{
 use data_types::timestamp::TimestampRange;
 use datafusion::{
     error::DataFusionError,
-    logical_plan::{col, Expr, Operator},
+    logical_plan::{col, lit, Column, Expr, Operator},
```
NGA-TRAN, line 14: The new code here was moved from the deleted delete_parser.rs plus converting Error to Snafu

influxdb_line_protocol/src/delete_parser.rs
```diff
@@ -1,461 +0,0 @@
-//! parse string that should be done at client before sending to server
```
NGA-TRAN, line 1: The content here is moved to predicate.rs

query/Cargo.toml
```diff
@@ -24,16 +24,19 @@ datafusion = { path = "../datafusion" }
 datafusion_util = { path = "../datafusion_util" }
 futures = "0.3"
 hashbrown = "0.11"
+influxdb_line_protocol = { path = "../influxdb_line_protocol" }
```
alamb, line 27: it is unfortunate to bring in this dependency, but it makes sense that we need to parse timestamps consistently, so I don't have a better idea at this point

query/src/predicate.rs
```diff
@@ -393,6 +431,212 @@ impl PredicateBuilder {
...
+    pub fn new(start_time: i64, stop_time: i64, predicate: Vec<Expr>) -> Self {
+        Self {
+            start_time,
+            stop_time,
+            predicate,
+        }
+    }
+
+    /// Parse and convert the delete grpc API into ParseDeletePredicate to send to server
+    pub fn try_new(table_name: &str, start: &str, stop: &str, predicate: &str) -> Result<Self> {
```
alamb, line 454: FWIW the other gRPC requests send their start/stop times in as integers (rather than strings), but perhaps that is because they handle parsing in Flux / InfluxQL

alamb (approved):

query/src/predicate.rs
NGA-TRAN: Yeah, we have 2 options either parse timestamps in the client and send integers to server or parse timestamp with the predicate in server and send all strings. We choose the latter now which can change to former if you think it is a better choice

alamb: I can't think of any reason for one way or the other at this time.

PR #2501, opened 2021-09-09
fix: checksum rustup in CI

jacobmarble:
Best to verify the checksum of anything fetched from the internet.


crepererum (approved):

domodwyer (approved):
I trust your hashes too.

scripts/install_rustup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
```
alamb, line 2:
```suggestion

# Use rustup to install rust tool chain after first
# comparing to a known checksum
```

mkmik:
> I trust your hashes too.

I wish rustup served a .sha256 or a SHA256SUMS like in the good old days; so it would be very cheap to verify during review

scripts/install_rustup.sh
```diff
@@ -0,0 +1,25 @@
...
+set -eu -o pipefail
+
+RUSTUP_VERSION=1.24.3
+RUSTUP_HASH=a3cb081f88a6789d104518b30d4aa410009cd08c3822a1226991d6cf0442a0f8
+
+# Notice --location to follow redirects -- important on Ubuntu 20.04
+curl --proto '=https' --tlsv1.2 -sSf --location --max-redirs 1 \
```
mkmik, line 9: I wish curl could just add this verify sha functionality built in.

mkmik: fwiw, in the past I've been using this https://github.com/mkmik/runck

NGA-TRAN:
Great to have this after the security training

shepmaster:
> I wish rustup served a .sha256 or a SHA256SUMS 

Are the ones listed on https://rust-lang.github.io/rustup/installation/other.html what you are looking for?

mkmik:
yeah, but if they were in a standard location they could be used by a tool like https://getsum.pub/

mkmik:
anyway, not all transparent log checksum checkers require that file, for example https://github.com/transparencylog/tl doesn't. I architected getsum so that a random attacker cannot just squat a bunch of urls with some bad hashes preventing the legitimate owner of domains to use some filenames.

mkmik:
> > I wish rustup served a .sha256 or a SHA256SUMS
> 
> Are the ones listed on https://rust-lang.github.io/rustup/installation/other.html what you are looking for?

@shepmaster the problem with those paths is that they don't encode the version name in the path, e.g.:

* https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init
* https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init.sha256

carols10cents:
> @shepmaster the problem with those paths is that they don't encode the version number in the path, e.g.:
>
> *    static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init
> *   static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init.sha256

Further down on https://rust-lang.github.io/rustup/installation/other.html it says:

> You can fetch an older version from `https://static.rust-lang.org/rustup/archive/{rustup-version}/{target-triple}/rustup-init[.exe]`

mkmik:
ah, but that's the sha for the binary, not for the shell script itself.

I.e. right now we're downloading https://raw.githubusercontent.com/rust-lang/rustup/1.24.3/rustup-init.sh and we'd want a reviewer to easily verify the checksum for that

PR #2502, opened 2021-09-10
feat: response classification (#2273)

tustvold:
This adds the ability to correctly ascertain whether a request succeeded or failed, and include this information in the produced spans. Aside from reporting error/success status to jaeger this same logic will be used for generating http api metrics.

I was unable to devise a simple test case for the streaming response failure (where the trailer indicates an error) but the logic is at the very least exercised by the non-streaming test case


src/influxdb_ioxd.rs
```diff
@@ -700,6 +701,16 @@ mod tests {
             .await
             .unwrap();
 
+        flight
```
crepererum, line 704: Maybe add a doc comment that these errors here should lead to traces describing them. I was about to complain that you don't check the error type here but later figured out these lines are just triggers for the actual test.

crepererum (approved):

PR #2504, opened 2021-09-10
docs: Add usecase to freezable doc string

alamb:
Try and encode my learnings from working with `Freezable` on https://github.com/influxdata/influxdb_iox/pull/2433 into some docstrings


internal_types/src/freezable.rs
```diff
@@ -2,7 +2,13 @@ use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 
 /// A wrapper around a type `T` that can be frozen with `Freezable::try_freeze`, preventing
-/// modification of the contained `T` until the returned `FreezeHandle` is dropped
+/// modification of the contained `T` until the returned `FreezeHandle` is dropped.
+///
+/// A common usecase for Freezable is preventing other threads from
```
alamb, line 7: This content is partially duplicative with the description after the code example, https://github.com/influxdata/influxdb_iox/pull/2504/files#diff-c0e65cadc0940055296ff6642bfa5bacc87c19c21f1bb09d5b4672117b89374cR37

But I think it helps readability to try and explain the usecase up front (so readers know if/why they should care about this code)

internal_types/src/freezable.rs
```diff
@@ -2,7 +2,13 @@ use std::ops::{Deref, DerefMut};
...
+/// modifying some object for an extended period of time prior to
+/// actually modifying it, while allowing other threads to read the
+/// object during that time period except during the actual
```
tustvold, line 10: I think I'd remove the "except during the actual modification"

tustvold (approved):

PR #2508, opened 2021-09-10
feat: migrate http metrics to metric crate

tustvold:
This migrates the HTTP metrics from using `metrics::RedMetric` to using the new metrics abstraction. Instead of a one-to-one port I opted to instead encode this in the tower middleware that is used for tracing instrumentation.

This has a couple of advantages:

* Blanket coverage of all current and future HTTP and gRPC endpoints 
* No potentially error-prone manual error recording
* Correctly handles streaming responses
* Ensures tracing and metrics are consistent
* Ensures a consistent relationship between the status code returned to the user and recorded in the metrics
* Can easily be extended with response size metrics, among other things

This does, however, introduce three **breaking changes**:

* The gRPC metrics are now at the HTTP path and not a custom string, i.e. `path="/influxdata.platform.storage.Storage/ReadWindowAggregate"` vs `operation="read_window_aggregate"`
* Additional metric context, in particular db_name, is not present anymore
* The server errors are reported under `status=server_error` vs `status=error` to distinguish them from `status=client_error` and make it clear they are specifically server errors and not all errors

The first of these I imagine is fairly uncontroversial. However, the second I will concede is not ideal and I intend to rectify in a follow up PR. It is not particularly complex, simply including a header in the response that gets stripped out by the trace middleware, but I'd prefer to get this in and iterate on it unless people have strong objections.


src/influxdb_ioxd/http.rs
```diff
@@ -485,20 +484,13 @@ async fn write<M>(req: Request<Body>) -> Result<Response<Body>, ApplicationError
...
 {
-    let path = req.uri().path().to_string();
     let Server {
         app_server: server,
         max_request_size,
     } = req.data::<Server<M>>().expect("server state");
     let max_request_size = *max_request_size;
     let server = Arc::clone(server);
 
-    // TODO(edd): figure out best way of catching all errors in this observation.
```
tustvold, line 496: This is auto-magically handled for us by the middleware

src/influxdb_ioxd/http.rs
```diff
@@ -1038,8 +986,28 @@ mod tests {
 
         let data = response.text().await.unwrap();
 
-        assert!(data.contains(&"\nhttp_requests_total{path=\"/metrics\",status=\"ok\"} 1\n"));
```
tustvold, line 1041: This was actually a somewhat dubious prior behavior - it would record the metric before the request had actually completed

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -1229,21 +1068,28 @@ mod tests {
...
-        fixture
+        count: u64,
+    ) {
+        let metric = fixture
             .test_storage
-            .metrics_registry
-            .has_metric_family("gRPC_requests_total")
-            .with_attributes(&[
-                ("operation", operation),
-                ("db_name", "000000000000007b_00000000000001c8"),
```
tustvold, line 1242: As mentioned in the PR description, we lose the `db_name` granularity but I intend to re-introduce this in a follow on PR

trace_http/src/classify.rs
```diff
@@ -0,0 +1,84 @@
...
+    Ok,
+    /// The request was to an unrecognised path
+    PathNotFound,
+    /// The request was unsuccessful but it was not the fault of the service
+    ClientErr,
+    /// The request was unsuccessful and it was the fault of the service
+    ServerErr,
+}
+
+pub fn classify_response<B>(response: &http::Response<B>) -> (Cow<'static, str>, Classification) {
```
tustvold, line 16: This code is moved from tower.rs and extended with `Classification`

trace_http/src/tower.rs
```diff
@@ -126,19 +145,34 @@ where
...
 
-        let recorder = self.as_mut().project().recorder;
+        let projected = self.as_mut().project();
+        let span_recorder = projected.span_recorder;
+        let mut metrics_recorder = projected.metrics_recorder.take().unwrap();
         match &result {
             Ok(response) => match classify_response(response) {
-                Ok(_) => recorder.event("request processed"),
-                Err(e) => recorder.error(e),
+                (_, Classification::Ok) => match response.body().is_end_stream() {
```
tustvold, line 153: This was actually a mistake in #2502 as it made the incorrect assumption that poll_trailers would always be called, this now handles the case where there is no body or there are no trailers

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -1851,7 +1697,7 @@ mod tests {
 
         assert_contains!(response_string, "Sugar we are going down");
 
-        grpc_request_metric_has_count(&fixture, "measurement_tag_values", "ok", 1).unwrap();
```
tustvold, line 1854: I think this was previously a bug? This is an error and yet was being recorded as a success

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -1762,7 +1608,7 @@ mod tests {
             "Error converting tag_key to UTF-8 in tag_values request"
         );
 
-        grpc_request_metric_has_count(&fixture, "tag_values", "client_error", 1).unwrap();
```
tustvold, line 1765: I think this is another case of the error-prone manual error recording being incorrect. This test exercises two error cases, and so the count should be two

src/influxdb_ioxd/http.rs
```diff
@@ -958,8 +904,9 @@ pub async fn serve<M>(
 where
     M: ConnectionManager + Send + Sync + Debug + 'static,
 {
+    let metric_registry = Arc::clone(application.metric_registry_v2());
     let router = router(application, server, max_request_size);
-    let new_service = tower::MakeService::new(router, trace_collector);
+    let new_service = tower::MakeService::new(router, trace_collector, metric_registry);
```
alamb, line 909: 👍

src/influxdb_ioxd/http.rs
```diff
@@ -1038,8 +986,28 @@ mod tests {
...
+            .get(&format!("{}/metrics", server_url))
+            .send()
+            .await
+            .unwrap();
+
+        let data = response.text().await.unwrap();
+
+        // Should include previous metrics scrape but not the current one
+        assert!(data.contains(&"\nhttp_requests_total{path=\"/metrics\",status=\"ok\"} 1\n"));
+        // Should include 404 but not encode the path
```
alamb, line 1009: maybe also
```suggestion
        // Should include 404 but not encode the path
        assert!(!data.contains(&"nonexistent"));
```

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -300,21 +273,12 @@ where
...
 
-        let group = expr::convert_group_type(group)
-            .context(ConvertingReadGroupType {
-                aggregate_string: &aggregate_string,
-            })
-            .map_err(|e| {
-                ob.client_error_with_attributes(attributes);
-                e
-            })?;
+        let group = expr::convert_group_type(group).context(ConvertingReadGroupType {
```
alamb, line 276: If the middleware adds the metrics correctly this certainly is a nice improvement 👍

src/influxdb_ioxd/rpc/storage/service.rs
alamb: yes I think db_name is quite important -- perhaps adding a ticket to track would be helpful. However you are quite good about not forgetting things so perhaps it is not necessary either :)

src/influxdb_ioxd/rpc/storage/service.rs
```diff
@@ -1353,7 +1199,7 @@ mod tests {
             expected_predicate
         );
 
-        grpc_request_metric_has_count(&fixture, "measurement_names", "ok", 2).unwrap();
+        grpc_request_metric_has_count(&fixture, "MeasurementNames", "ok", 2);
```
alamb, line 1202: it might be worth also calling out that the operation name will change from snake case to camel case (more closely matching ProtoBuf, which is a good think in my opinion)

trace_http/src/classify.rs
```diff
@@ -0,0 +1,84 @@
...
+    let status = response.status();
+    match status {
+        http::StatusCode::OK | http::StatusCode::CREATED | http::StatusCode::NO_CONTENT => {
+            classify_headers(Some(response.headers()))
+        }
+        http::StatusCode::BAD_REQUEST => ("bad request".into(), Classification::ClientErr),
+        // Note: This is potentially over-zealous but errs on the side of caution to
```
alamb, line 23: Maybe it is worth noting the usecase / rationale here (aka to avoid adding non-existent paths as metric attribute values), even if sometimes the non-existent thing represented by `404` could be some entity even if the URL was existent.

trace_http/src/classify.rs
```diff
@@ -0,0 +1,84 @@
...
+                6 => ("already exists".into(), Classification::ClientErr),
+                7 => ("permission denied".into(), Classification::ClientErr),
+                8 => ("resource exhausted".into(), Classification::ServerErr),
+                9 => ("failed precondition".into(), Classification::ClientErr),
+                10 => ("aborted".into(), Classification::ClientErr),
+                11 => ("out of range".into(), Classification::ClientErr),
+                12 => ("unimplemented".into(), Classification::ServerErr),
+                13 => ("internal".into(), Classification::ServerErr),
+                14 => ("unavailable".into(), Classification::ServerErr),
+                15 => ("data loss".into(), Classification::ServerErr),
```
alamb, line 74: somewhat terrifying that there is a expected error for this class (rather than just internal error)

trace_http/src/metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+use parking_lot::{MappedMutexGuard, Mutex, MutexGuard};
+use std::sync::Arc;
+use std::time::Instant;
+
+/// `MetricsCollection` is used to retrieve `MetricsRecorder` for instrumenting http requests
+#[derive(Debug)]
+pub struct MetricsCollection {
+    is_grpc: bool,
+    metric_registry: Arc<metric::Registry>,
+    metrics: Mutex<HashMap<Option<String>, Metrics>>,
```
alamb, line 13: maybe worth mentioning what the key represents: the path of the request's uri?

trace_http/src/metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+                    (path, metrics)
+                });
+            request_metrics
+        })
+    }
+}
+
+/// The request metrics for a specific set of attributes (e.g. path)
+#[derive(Debug)]
+struct Metrics {
```
alamb, line 55: this is very cool

trace_http/src/metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        attributes.insert("status", "ok");
+        let ok = counter.recorder(attributes.clone());
+        let duration_ok = duration.recorder(attributes.clone());
+
+        attributes.insert("status", "client_error");
+        let client_error = counter.recorder(attributes.clone());
+        let duration_client_error = duration.recorder(attributes.clone());
+
+        attributes.insert("status", "error");
```
alamb, line 89: why not use `server_error` to be more explicit (and follow the pattern from `client_error`)?

```suggestion
        attributes.insert("status", "server_error");
```

trace_http/src/metrics.rs
```diff
@@ -0,0 +1,152 @@
...
+
+impl MetricsRecorder {
+    /// Sets the classification of this request if not already set
+    pub fn set_classification(&mut self, classification: Classification) {
+        if matches!(classification, Classification::PathNotFound) {
+            // Don't want to pollute metrics with invalid paths
+            self.path = None
+        }
+
+        if self.classification.is_none() {
```
alamb, line 125: I wonder if this should also check (perhaps `debug_assert!` or something) that if `self.classification` was previously set it should match `classification`?

trace_http/src/tower.rs
```diff
@@ -12,30 +12,42 @@
...
+use observability_deps::tracing::error;
 use trace::{span::SpanRecorder, TraceCollector};
 
+use crate::classify::{classify_headers, classify_response, Classification};
+use crate::ctx::parse_span_ctx;
+use crate::metrics::{MetricsCollection, MetricsRecorder};
+
 /// `TraceLayer` implements `tower::Layer` and can be used to decorate a
 /// `tower::Service` to collect information about requests flowing through it
 #[derive(Debug, Clone)]
```
alamb, line 35: maybe it is worth updating this docstring to mention that this `TraceLayer` is used to provide generic distributed tracing and RED metric reporting

trace_http/src/tower.rs
alamb: it used to handle the case in the sense that a trace was emitted, right? It seems like the difference now is that there is a more specific event used.

trace_http/src/tower.rs
```diff
@@ -110,7 +128,8 @@ where
 #[pin_project]
 #[derive(Debug)]
 pub struct TracedFuture<F> {
-    recorder: SpanRecorder,
+    span_recorder: SpanRecorder,
```
alamb, line 131: All we have to do is to add logging (ala `info!`, `error!` etc and we'll have the observability trifecta!)

trace_http/src/tower.rs
```diff
@@ -168,10 +203,21 @@ impl<B: http_body::Body> http_body::Body for TracedBody<B> {
...
+            Ok(_) => match projected.inner.is_end_stream() {
+                true => {
+                    metrics_recorder.set_classification(Classification::Ok);
+                    span_recorder.ok("returned body data and no trailers")
+                }
+                false => span_recorder.event("returned body data"),
+            },
+            Err(_) => {
+                metrics_recorder.set_classification(Classification::ServerErr);
+                span_recorder.error("eos getting body");
```
alamb, line 219: is it always "eos" or maybe we should say "error getting body"?

alamb (approved):
FYI @mkmik  -- if the metrics dashboards break/start doing something wonky maybe this PR is related. 

I think this is looking great. Nice work @tustvold.  Can't wait to see the old metrics crate go the way of the 🦤

trace_http/src/classify.rs
tustvold: :laughing: I _think_ it is used to indicate that the service knows it has lost data and an operator needs to intervene with out of band backups or something.

trace_http/src/tower.rs
tustvold: Correct, in all cases it would emit a span. All that would happen is that in certain cases it wouldn't know if it was successful or not - i.e. `SpanStatus::Unknown`

trace_http/src/metrics.rs
tustvold: I was trying to be backwards compatible, but I think it is less confusing to be explicit - will update

trace_http/src/metrics.rs
tustvold: So the problematic scenario is PathNotFound as to all intents and purposes the request is actually successful and it is later recorded as such in TracedBody. I've instead gone for adding Ord to Classification and taking the max here. I think this captures the spirit of the debug_assert, avoiding issues getting lost, whilst allowing some "slop".

src/influxdb_ioxd/rpc/storage/service.rs
tustvold: Created https://github.com/influxdata/influxdb_iox/issues/2515

e-dard:
The changes look good, but I didn't review them:

I have a question:
 
>  Instead of a one-to-one port I opted to instead encode this in the tower middleware that is used for tracing instrumentation.

Does this mean there is no longer going to be a `REDMetric` helper type for folks to use in the project? It wasn't clear to me whether the requests, errors and duration aspect of that type had been hard coded only into the gRPC/HTTP middleware or whether there is a new `REDMetric` abstraction elsewhere?

The only reason I ask is that it might be useful to have that abstraction for use elsewhere other than the `gRPC/HTTP` boundary. Just a thought.

PR #2510, opened 2021-09-10
feat: migrate read buffer metrics to metric crate

tustvold:
Migrates the read buffer to use the new metric crate


metric/src/cumulative.rs
```diff
@@ -52,6 +52,14 @@ pub struct CumulativeRecorder {
 }
 
 impl CumulativeRecorder {
+    /// Gets a new unregistered recorder
```
tustvold, line 55: This is also added in #2490 but didn't want to make the PRs depend on each other

read_buffer/src/chunk.rs
```diff
@@ -712,124 +697,111 @@ mod test {
...
-        r#"read_buffer_column_values{db="mydb",encoding="FIXED",log_data_type="f64",null="true"} 0"#,
-        r#"read_buffer_column_values{db="mydb",encoding="FIXEDN",log_data_type="bool",null="false"} 6"#,
-        r#"read_buffer_column_values{db="mydb",encoding="FIXEDN",log_data_type="bool",null="true"} 0"#,
-        r#"read_buffer_column_values{db="mydb",encoding="RLE",log_data_type="string",null="false"} 6"#,
-        r#"read_buffer_column_values{db="mydb",encoding="RLE",log_data_type="string",null="true"} 0"#,
-        "# HELP read_buffer_row_group_total The number of row groups within the Read Buffer",
-        "# TYPE read_buffer_row_group_total gauge",
-        r#"read_buffer_row_group_total{db="mydb"} 2"#,
-        "",
+        let expected_observations = vec![
```
tustvold, line 700: I could have used metric-exporter, but I didn't want to introduce a dependency on prometheus

read_buffer/src/chunk.rs
```diff
@@ -712,124 +697,111 @@ mod test {
...
+            ObservationSet {
+                metric_name: "read_buffer_column_allocated_bytes",
+                description: "The number of bytes used by all data in the Read Buffer including allocated by unused buffers",
+                kind: MetricKind::U64Gauge,
+                observations: vec![
+                    (Attributes::from(&[("db_name", "mydb"), ("encoding", "BT_U32-FIXED"), ("log_data_type", "i64")]), Observation::U64Gauge(192)),
```
tustvold, line 706: "db_name" is the "correct" attribute and is what is currently exposed by the server binary, previously this test used "db" but the ChunkMetric constructor forces consistency (which I think is a good thing)

server/src/db.rs
```diff
@@ -2163,7 +2163,7 @@ mod tests {
 
         // verify chunk size updated (chunk moved from closing to moving to moved)
         catalog_chunk_size_bytes_metric_eq(&test_db.metric_registry, "mutable_buffer", 0);
-        let expected_read_buffer_size = 1922;
+        let expected_read_buffer_size = 1706;
```
tustvold, line 2166: `read_buffer::ChunkMetrics` is now smaller (although the actual memory usage might be higher)

alamb (approved):
Looks like a very solid change to me 💯

alamb:
FYI @e-dard

PR #2511, opened 2021-09-10
feat: save delete predicates in chunks

NGA-TRAN:
Another task of #2396 


tests/end_to_end_cases/management_api.rs
```diff
@@ -1489,6 +1489,7 @@ async fn test_delete() {
         .await
         .unwrap();
     // todo: The delete function  above just parses the input, nothing deleted in DB yet
+    // todoL: should add different tests for different stages of chunks, too
```
NGA-TRAN, line 1492: The end-to-end tests will be a lot more complicated when I apply delete predicate at query time which will be my next step

server/src/db/catalog/chunk.rs
```diff
@@ -1086,6 +1105,75 @@ mod tests {
...
+        let pred = &del_preds[1];
+        if let Some(range) = pred.range {
+            assert_eq!(range.start, 20); // start time
+            assert_eq!(range.end, 50); // start time
+        } else {
+            panic!("No time range set for delete predicate");
+        }
+        assert_eq!(pred.exprs, expected_exprs2);
+    }
+
```
NGA-TRAN, line 1176: These (integration) tests have proved the delete predicates are applied to chunks. I will add  complete end-to-end tests when the delete predicates are applied at query time for all kinds of chunk types.

server/src/db/catalog/chunk.rs
```diff
@@ -1086,6 +1105,75 @@ mod tests {
...
+        // chunk must be in frozen stage now
+        assert_eq!(chunk.stage().name(), "Frozen");
+        // chunk must have a delete predicate
+        let del_preds = chunk.delete_predicates();
+        assert_eq!(del_preds.len(), 1);
+        // verify delete predicate value
+        let pred = &del_preds[0];
+        if let Some(range) = pred.range {
+            assert_eq!(range.start, 1); // start time
+            assert_eq!(range.end, 100); // start time
```
NGA-TRAN, line 1139: Will make the comment to stop time before merging

server/src/db/catalog/chunk.rs
```diff
@@ -684,7 +700,7 @@ impl CatalogChunk {
         // This ensures the closing logic is consistent but doesn't break code that
         // assumes a chunk can be moved from open
         if matches!(self.stage, ChunkStage::Open { .. }) {
-            self.freeze()?;
+            self.freeze(&PredicateBuilder::default().build())?;
```
alamb, line 703: As I mentioned above, the code change in this PR might be able to be reduced if you added a new `freeze_with_predicate` instead of changing the signature of `freeze`

server/src/db/lifecycle/compact.rs
```diff
@@ -58,6 +62,8 @@ pub(crate) fn compact_chunks(
                 .map(|prev_last| prev_last.max(candidate_last))
                 .or(Some(candidate_last));
 
+            delete_predicates = chunk.delete_predicates();
```
alamb, line 65: As written this code appears to use the delete predicates for the last chunk (rather than all the delete predicates for all the chunks) when compacting the list. 

Perhaps the code should be appending the delete predicates of the chunk (rather than overwriting them)? Something like

```suggestion
            delete_predicates.append(chunk.delete_predicates().clone());
```

server/src/db/lifecycle/persist.rs
```diff
@@ -64,6 +68,8 @@ pub fn persist_chunks(
             .map(|prev_last| prev_last.max(candidate_last))
             .or(Some(candidate_last));
 
+        delete_predicates = chunk.delete_predicates();
```
alamb, line 71: I think the same comment from above applies here -- I think the persisted chunk should have *all* the delete predicates from the input chunks, rather than just the last one

alamb (approved):
Nice work 👍 

The idea looks good to me --  I think the delete predicates created during compact/persist should be reviewed before merge. Everything else is a style comment

server/src/db.rs
```diff
@@ -551,7 +555,9 @@ impl Db {
         info!(%table_name, %partition_key, found_chunk=chunk.is_some(), "rolling over a partition");
         if let Some(chunk) = chunk {
             let mut chunk = chunk.write();
-            chunk.freeze().context(FreezingChunk)?;
+            chunk
+                .freeze(&PredicateBuilder::default().build())
```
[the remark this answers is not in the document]
NGA-TRAN: This is a good idea because freeze is invoked many places without the need of delete predicates. Will do.

server/src/db.rs
```diff
@@ -653,9 +659,11 @@ impl Db {
             let partition = partition.write();
             let chunks = partition.chunks();
             for chunk in chunks {
-                // NGA todo: verify where to close MUB before adding the predicate
+                // save the delete predicate in the chunk
```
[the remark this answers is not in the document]
NGA-TRAN: The functions that are invoked by `add_delete_predicate` check for MUB and close it. Throwing an error is a good idea to ensure we do not miss any cases. I will find a place either here or a bit lower level but high enough to throw this error

server/src/db/catalog/chunk.rs
```diff
@@ -295,12 +295,13 @@ impl CatalogChunk {
         time_of_last_write: DateTime<Utc>,
         schema: Arc<Schema>,
         metrics: ChunkMetrics,
+        delete_predicates: Arc<Vec<Predicate>>,
```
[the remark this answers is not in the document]
NGA-TRAN: delete_predicate is copied a lot to different chunks is the reason but I have no strong opion on this

server/src/db/catalog/chunk.rs
NGA-TRAN: Agree. Will do

server/src/db/lifecycle/compact.rs
NGA-TRAN: This is a great catch

server/src/db/lifecycle/persist.rs
NGA-TRAN: Yeah. Bug here 😄

server/src/db/catalog/chunk.rs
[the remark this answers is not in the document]
alamb: me neither -- this looks fine to me

PR #2516, opened 2021-09-13
feat: migrate chunk pruning metrics

tustvold:
Migrates the pruning metrics to the new metric crate


server/src/db/chunk.rs
```diff
@@ -213,8 +213,8 @@ impl DbChunk {
     }
 
     /// Return the name of the table in this chunk
-    pub fn table_name(&self) -> Arc<str> {
-        Arc::clone(&self.table_name)
+    pub fn table_name(&self) -> &Arc<str> {
```
tustvold, line 216: The only reason to return a pre-cloned Arc is if there is a lock that prevents propagating the borrow. This is not the case here, and so imo returning a pre-cloned Arc is an anti-pattern that results in additional atomic increments and prevents borrow propagation (the returned Arc str is a stack-local temporary)

query_tests/src/pruning.rs
```diff
@@ -129,27 +125,22 @@ async fn chunk_pruning_influxrpc() {
...
-        .with_attributes(&[
-            ("db_name", "placeholder"),
-            ("table_name", "cpu"),
-            ("svr_id", "1"),
-        ])
-        .counter()
-        .eq(3.0)
-        .unwrap();
+    let pruned_rows = metrics_registry_v2
+        .get_instrument::<Metric<U64Counter>>("query_access_pruned_rows")
```
tustvold, line 140: The prometheus `_total` suffix is a detail added by the prometheus exporter, the registry stores the actual metric name

query_tests/src/pruning.rs
```diff
@@ -76,29 +77,24 @@ async fn chunk_pruning_sql() {
...
+        .fetch();
+    assert_eq!(pruned_chunks, 1);
 
     // Validate that the chunk was pruned using the metrics
-    metric_registry
-        .has_metric_family("query_access_pruned_rows_total")
-        .with_attributes(&[
-            ("db_name", "placeholder"),
-            ("table_name", "cpu"),
-            ("svr_id", "1"),
```
tustvold, line 97: As is par for the course in this migration we lose svr_id as it is no longer an attribute associated with the top-level domain. If there is a need for this metric, the more idiomatic way to expose it in prometheus would be a "label gauge" - i.e. a gauge with a single value of 1 that can then be joined with the other metrics. I may in fact just do this...

query/Cargo.toml
```diff
@@ -26,7 +26,6 @@ futures = "0.3"
 hashbrown = "0.11"
 influxdb_line_protocol = { path = "../influxdb_line_protocol" }
 internal_types = { path = "../internal_types" }
-metrics = { path = "../metrics" }
```
tustvold, line 29: This didn't actually need to be here

query_tests/src/pruning.rs
alamb: I think db_name is far more important for the time being. svr_id could also be added as part of the telegraf scraping config perhaps

server/src/db/access.rs
```diff
@@ -34,37 +36,63 @@ use query::{
 /// Metrics related to chunk access (pruning specifically)
 #[derive(Debug)]
 struct AccessMetrics {
+    /// The database name
+    db_name: Arc<str>,
     /// Total number of chunks pruned via statistics
-    pruned_chunks: Counter,
+    pruned_chunks: Metric<U64Counter>,
```
alamb, line 42: Is there a reason to use the aggregated values (given that `TableAccessMetrics` has the same information with a finer granularity)?

server/src/db/chunk.rs
alamb: 👍  I think letting the caller decide if they need their own `Arc` is a good pattern too

alamb (approved):
Thank you. 👍

server/src/db/access.rs
tustvold: Nope, they just are "cached" here to avoid having to go back to the central Registry every time a new table is encountered.

query_tests/src/pruning.rs
tustvold: Indeed - or even exposed as a label gauge

PR #2520, opened 2021-09-13
fix: MockBufferForReading waker registration

tustvold:
Previously the streams returned by MockBufferForReading would return Poll::Pending without storing the waker. The result would be that even if further entries were written to the shared state, the streams would not necessarily ever be polled again and consequently the write buffer would lock up.

I ran into this whilst working on #2243 

I initially tried to fix this using tokio::sync::Notify, however, the returned futures (`Notified`) unregister on drop, which makes them tricky to use in the context of stream::poll_fn - you end up needing to manually implement a future state machine which is onerous.

I then experimented with using tokio::sync::broadcast but this then also didn't quite fit because of get_messages() and some details w.r.t lagged consumers, etc...

In the end this PR goes with the dumbest possible implementation, a `Vec<Waker>`. No attempt is made to handle waker cancellation, as not only is this only used for tests, but in general there should only be one stream consumer anyway so this is almost certainly good enough


write_buffer/src/mock.rs
```diff
@@ -749,4 +768,44 @@ mod tests {
...
+            stream.next().await.unwrap().unwrap();
+            stream.next().await.unwrap().unwrap();
+        });
+
+        // Wait for consumer to read first entry
+        while playback_state.lock().get(&0).unwrap().vector_index < 1 {
+            tokio::time::sleep(Duration::from_millis(1)).await;
+        }
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
alamb, line 798: what is the purpose of the 1 second `sleep` in the test. Can it be smaller (10ms for example?) Or conditionalized on the a signal / barrier in the consumer it is in the correct state?

alamb (approved):
Looks good to me. While not a futures wizard the implementation looks straightforward enough to me 👍  for the simple implementation

I worry about the `sleep` in the tests being annoying in the future, so it would be cool if that could be removed, but otherwise looks good to me

write_buffer/src/mock.rs
tustvold: Oh this can go, it was from when I was testing something

PR #2522, opened 2021-09-13
feat: push delete predicate down to RUB and OS chunks to eliminate deleted data at query time

NGA-TRAN:
A task of #2396 


NGA-TRAN:
@crepererum This is a new PR I have just started working on this afternoon my time. It is not ready for review but I just want to let you know I touch the Predicate again. I am not sure how much conflict it will be if you move Predicate to its own crate at this point.

server/src/db/catalog/chunk.rs
```diff
@@ -714,7 +731,7 @@ impl CatalogChunk {
     }
 
     pub fn freeze(&mut self) -> Result<()> {
-        self.freeze_with_predicate(&PredicateBuilder::default().build())
```
NGA-TRAN, line 717: This was a bug I introduced last PR that adds an empty delete predicate into the chunk which tells the chunk to delete everything because no delete time range is specified

server/src/db/load.rs
```diff
@@ -269,7 +269,7 @@ impl CatalogState for Loader {
             .map_err(|e| Box::new(e) as _)
             .context(SchemaError { path: info.path })?;
 
-        let delete_predicates: Arc<Vec<Predicate>> = Arc::new(vec![]); // NGA todo: After Marco save delete predicate into the catalog, it will need to extract into this variable
+        let delete_predicates: Arc<Vec<Predicate>> = Arc::new(vec![]); // NGA todo: After Marco saves delete predicate into the catalog, it will need to get extracted into this variable
```
NGA-TRAN, line 272: #2550

predicate/src/predicate.rs
```diff
@@ -161,6 +161,28 @@ impl Predicate {
     pub fn is_empty(&self) -> bool {
         self == &EMPTY_PREDICATE
     }
+
+    /// Add each range [start, stop] of the delete_predicates into the predicate in
+    /// the form "time < start OR time > stop" to eliminate that range from the query
+    pub fn add_delete_ranges(&mut self, delete_predicates: &[Self]) {
```
e-dard, line 167: I'm wondering why you used a disjunctive expression here, instead of a conjunctive one (`time > x AND time < y`).

The Read Buffer does not support disjunctive expressions right now so this delete predicate would not be able to be pushed down to the Read Buffer.

e-dard (commented):
I will finish review tomorrow but I had a question.

server/src/db/chunk.rs
```diff
@@ -314,18 +327,36 @@ impl QueryChunk for DbChunk {
...
         // when possible for performance gain
 
         debug!(?predicate, "Input Predicate to read_filter");
         self.access_recorder.record_access_now();
 
+        debug!(?delete_predicates, "Input Delete Predicates to read_filter");
+
+        // add negated deleted ranges to the predicate
+        let mut pred_with_deleted_ranges = predicate.clone();
+        pred_with_deleted_ranges.add_delete_ranges(delete_predicates);
```
NGA-TRAN, line 346: This is needed for both RUB and OS.
@e-dard I do not add delete time range in the negated predicate (see below) so I include the not(time range) in the normal predicate here. Is it the expected behavior for RUB?

server/src/db/chunk.rs
```diff
@@ -314,18 +327,36 @@ impl QueryChunk for DbChunk {
...
+        debug!(
+            ?pred_with_deleted_ranges,
+            "Input Predicate plus deleted ranges"
+        );
+
+        // add negated deleted predicates
+        let mut pred_wth_deleted_exprs = pred_with_deleted_ranges.clone();
+        pred_wth_deleted_exprs.add_delete_exprs(delete_predicates);
```
NGA-TRAN, line 354: This is only needed for OS chunks. RUB chunks will accept negated predicates directly. See below on how to build negated predicate

predicate/src/predicate.rs
NGA-TRAN: Since RUB's negated_predicate only accept `=` and `!=`, this time range must be added in the normal predicate and it needs to get converted to `not(start<time and time<stop)` which means `time< start OR time>stop`

predicate/src/predicate.rs
```diff
@@ -161,6 +161,28 @@ impl Predicate {
...
+        }
+    }
+
+    /// Add a list of not(a delete expr)
+    pub fn add_delete_exprs(&mut self, delete_predicates: &[Self]) {
+        for pred in delete_predicates {
+            for exp in &pred.exprs {
+                self.exprs.push(exp.clone().not());
+            }
+        }
```
NGA-TRAN, line 184: I think there is a bug here. I am fixing this.

predicate/src/predicate.rs
NGA-TRAN: I have looked into this and this is not needed for RUB. It only needs for OS so I will resolve this conversation.

server/src/db/chunk.rs
NGA-TRAN: This is no longer needed for RUB so I will resolve the conversation

predicate/src/predicate.rs
alamb: I really think it would make for a better design if there was no special case ReadBuffer handling anywhere except the final call to `ReadBuffer::read_filter` at which point the predicates can be broken down / rewritten to take advantage of all the special RUB filtering logic.

If we try and special case deletes / RUB any higher than that I think the code is going to get messy

query/src/provider/physical.rs
```diff
@@ -116,14 +116,18 @@ impl<C: QueryChunk + 'static> ExecutionPlan for IOxReadFilterNode<C> {
...
 
-        let stream = chunk.read_filter(&self.predicate, selection).map_err(|e| {
-            DataFusionError::Execution(format!(
-                "Error creating scan for table {} chunk {}: {}",
-                self.table_name,
-                chunk.id(),
-                e
-            ))
-        })?;
+        let del_preds = chunk.delete_predicates();
```
alamb, line 119: I wonder if it would be ok here to merge `delete_predicates` and `self.predicate` into one list so that the chunk just saw one unified list? I think isolating the knowledge of deletes to as few places as possible will help us avoid a "we forgot deletes" for certain types of queries

alamb (commented):
I started reviewing this PR but ran out of time -- I will review it more carefully in the morning

predicate/src/predicate.rs
NGA-TRAN: I think that is what it is done in this PR. This is for general case that we need to merge delete predicate into the select  predicate. What I can improve here is to put two functions `add_delete_ranges` and `add_delete_exprs` into only one function and name it `merge_delete_predicates_into_select_predicate`

query/src/provider/physical.rs
NGA-TRAN: Let me think about this. The reason it is kept at this low level because we need it to send to RUB as is.

alamb: In some ways, combining the list will require "undoing" the combination before sending to the RUB

However, the upside is that by "undoing" the combination before sending it to the RUB we may well find non delete predicates that can be sent to the RUB's fast processing path as well

server/src/db.rs
```diff
@@ -3782,4 +3784,262 @@ mod tests {
...
+            .build();
+        db.delete("cpu", &pred).await.unwrap();
+        // When the above delete is issued, the open mub chunk is frozen with the delete predicate added
+        // Verify there is MUB but no RUB no OS
+        assert!(!mutable_chunk_ids(&db, partition_key).is_empty());
+        assert!(read_buffer_chunk_ids(&db, partition_key).is_empty());
+        assert!(read_parquet_file_chunk_ids(&db, partition_key).is_empty());
+        // Let query again
+        // NGA todo: since MUB does not support predicate push down, the filtering needs to be done at scan step
+        let batches = run_query(Arc::clone(&db), "select * from cpu").await;
```
alamb, line 3834: I think this test might be able to be ported / written in the `query_tests` crate -- specifically, it is checking that the correct query results are produced in various MUB/RUB chunk setups, which is exactly what `query_tests` exists to do. 

One thing that this test does cover that a simple `query_test` would not is that a delete predicate added to a `MUB` chunk is properly passed on to a `RUB` or `Parquet` chunk. In order to cover that case you would have to put the delete in the setup of `DbScenario`

server/src/db.rs
```diff
@@ -3782,4 +3784,262 @@ mod tests {
...
+        assert_batches_sorted_eq!(expected, &batches); // this fails when the todo is done
```
alamb, line 3835: Just so I am clear -- this line shows an incorrect result (that will be resolved in a future PR, right)?

alamb (approved):
I think this PR moves the delete project forward. While there (as always!) more work to do I think we can fix it up in follow on PRs

query/src/provider/physical.rs
NGA-TRAN: @alamb This is a good idea for @crepererum and me to consider. I think we will be able to determine more pros and con in the next few days when we finish this project up and decide how to make this clearer and easier to handle and maintenance

server/src/db.rs
NGA-TRAN: Hah. My least satisfaction of this PR was these 2 complicated tests and you hit the right spot :)
These are great suggestions. Right after merging this PR, I will work on moving these tests to `query_tests` and add more stuff in DbScenario to build more effective test framework for delete.

server/src/db.rs
NGA-TRAN: Right. Since MUB does not do predicate push down, we do not see deleted data eliminated from query yet.

PR #2523, opened 2021-09-13
fix: emit traces to Jaeger in perf.py

jacobmarble:
The `perf.py` tool wasn't actually emitting (or collecting) traces in Jaeger. Now, battery queries emit traces.

It will be good to trace everything, once the sampling config is working in IOx.


perf/perf.py
```diff
@@ -636,7 +660,11 @@ def run_test_battery(battery_name, router_id, writer_id, debug=False):
         print('running test "%s"' % name)
         time_start = time.time()
         params = {'q': sql, 'format': 'csv'}
-        response = requests.get(url=query_url, params=params)
+        headers = {}
+        if do_trace:
+            # TODO remove this after IOx can be configured to sample 100% of traces
+            headers['uber-trace-id'] = '%x:%x:0:1' % (random.randrange(0, 2 ** 64), random.randrange(0, 2 ** 64))
```
tustvold, line 666: FWIW we support 128-bit trace ids, but this will also work

tustvold (approved):

.circleci/config.yml
```diff
@@ -176,7 +183,7 @@ jobs:
             - cargo-lock-{{ checksum "Cargo.lock" }}
       - run:
           name: Prime Rust build cache
-          command: cargo build --package influxdb_iox --features jaeger --bin influxdb_iox --package iox_data_generator --bin iox_data_generator
+          command: cargo build --package influxdb_iox --bin influxdb_iox --package iox_data_generator --bin iox_data_generator
```
alamb, line 186: What is the rationale for omitting jaeger? Is it so that this cache can be reused by other (potentially non-perf) jobs?

perf/perf.py
```diff
@@ -403,14 +425,15 @@ def exec_iox(id, service_name, debug=False, object_store='memory', print_only=Fa
...
-        'TRACES_SAMPLER': 'always_on',
         'RUST_BACKTRACE': '1',
         'LOG_FILTER': 'debug,lifecycle=info,rusoto_core=warn,hyper=warn,h2=warn',
     }
+    if do_trace:
+        env['TRACES_EXPORTER'] = 'jaeger'
+        env['TRACES_EXPORTER_JAEGER_AGENT_HOST'] = 'localhost'
+        env['TRACES_EXPORTER_JAEGER_AGENT_PORT'] = '6831'
+        env['TRACES_EXPORTER_JAEGER_SERVICE_NAME'] = service_name
+        env['TRACES_SAMPLER'] = 'always_on'
```
alamb, line 436: as @tustvold  has noted elsewhere, I believe `TRACES_SAMPLER`  is not currently hooked to anything

alamb (approved):

.circleci/config.yml
jacobmarble: Faster build times in CI

perf/perf.py
jacobmarble: That's fine, I hope it gets added soon ish?

jacobmarble: Or am I wrong about that?

alamb: Wrong about your hope that it gets added soon? I am not sure what you mean :)

jacobmarble: Yes, I'm just allowing for the possibility that the sampling flag is low-priority, or that I have the env var name wrong.

PR #2524, opened 2021-09-14
feat: prune old transactions from preserved catalog

crepererum:
Through checkpoints, the preserved catalog is currently growing quadratic since each new checkpoint lists all files of the previous one plus the new files since then. Since we probably only wanna "go back in time" for a limited time span, let's just delete old transactions.

In the future, we could extend the pruning to be more sophisticated, e.g. keep checkpoints in different granularity depending on the age (e.g. hourly for checkpoints aged >1d, daily for checkpoints aged >1W, ...). I deem this simple approach here good enough for now though.

### DIFF
--- a/parquet_file/src/catalog/prune.rs
+++ b/parquet_file/src/catalog/prune.rs
@@ -1,0 +1,260 @@
+//! Tooling to remove parts of the preserved catalog that are no longer needed.
+use std::{collections::BTreeMap, sync::Arc};
+
+use chrono::{DateTime, Utc};
+use futures::TryStreamExt;
+use iox_object_store::{IoxObjectStore, TransactionFilePath};
+use object_store::{ObjectStore, ObjectStoreApi};
+use snafu::{ResultExt, Snafu};
+
+use crate::catalog::{
+    api::{ProtoIOError, ProtoParseError},
+    internals::{proto_io::load_transaction_proto, proto_parse::parse_timestamp},
+};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Error during store read operation: {}", source))]
+    Read {
+        source: <ObjectStore as ObjectStoreApi>::Error,
+    },
+
+    #[snafu(display("Error during store delete operation: {}", source))]
+    Delete {
+        source: <ObjectStore as ObjectStoreApi>::Error,
+    },
+
+    #[snafu(display("Error during protobuf IO: {}", source))]
+    ProtobufIOError { source: ProtoIOError },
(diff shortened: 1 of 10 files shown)


parquet_file/src/catalog/prune.rs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            for file in files {
+                if (file.revision_counter < earliest_keep) || !is_checkpoint_or_zero(&file) {
+                    iox_object_store
+                        .delete_catalog_transaction_file(&file)
+                        .await
+                        .context(Delete)?;
+                }
+            }
```
mkmik, line 118: Isn't there a simple construct to collect the futures and then wait for all of them while they run in parallel?

server/src/db.rs
```diff
@@ -894,23 +895,42 @@ impl Db {
...
+                            self.iox_object_store(),
+                            Utc::now() - catalog_transaction_prune_age,
+                        )
+                        .await
+                        {
+                            error!(%e, "error while pruning catalog transactions");
+                        }
+                    }
+                    Err(e) => {
+                        error!(%e, "cannot convert `catalog_transaction_prune_age`, skipping transaction pruning");
```
mkmik, line 928: perhaps?

```suggestion
                        error!(%e, "cannot parse `catalog_transaction_prune_age`, skipping transaction pruning");
```

mkmik (approved):
modulo minor comments

parquet_file/src/catalog/prune.rs
crepererum: We currently do IO ops NOT in parallel (same goes for deleting unreferenced parquet files). I guess that would require a connection or (in case of http2/3) a stream limit, otherwise we just  DoS the object store provider. Let's leave it with a sequential approach for now and I check the parallel connection limit for `object_store` later.

parquet_file/src/catalog/prune.rs
```diff
@@ -0,0 +1,260 @@
...
+/// | T2        | Transaction               |
+/// | T3, C3    | Transaction + Checkpoint  |
+/// | T4        | Transaction               |
+/// | `before`  |                           |
+/// | T5        | Transaction               |
+/// | T6, C6    | Transaction + Checkpoint  |
+/// | T7        | Transaction               |
+/// | T9, C9    | Transaction + Checkpoint  |
+///
+/// This will delete the following content: C1, T2, and T3. C3 and T4 cannot be deleted because it is required to
```
alamb, line 53: this is a nice example

PR #2525, opened 2021-09-14
feat: pull WriteBuffer consumer out of Db and onto Database (#2243)

tustvold:
This PR pulls the write buffer consumer out of Db and into a WriteBufferConsumer that is stored on Database directly. Aside from reducing the amount of code in db.rs, I think this makes things a bit easier to reason about. 

Most of the code is just moved around


server/src/database.rs
```diff
@@ -888,24 +905,27 @@ impl DatabaseStateRulesLoaded {
...
-        let write_buffer = shared
-            .application
-            .write_buffer_factory()
-            .new_config(
-                shared.config.server_id,
-                self.provided_rules.rules().as_ref(),
-            )
-            .await
-            .context(CreateWriteBuffer)?;
+        let rules = self.provided_rules.rules();
```
tustvold, line 908: Creation of the read_config is deferred to when performing replay, not only does this avoid the need for additional synchronisation primitives but I think also avoids a situation where the state from failed replays is kept around (although at the moment I don't think this matters)

server/src/db/replay.rs
```diff
@@ -81,64 +81,59 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 /// operation fails. In that case some of the sequencers in the write buffers might already be seeked and others not.
 /// The caller must NOT use the write buffer in that case without ensuring that it is put into some proper state, e.g.
 /// by retrying this function.
-pub async fn seek_to_end(db: &Db) -> Result<()> {
```
tustvold, line 84: It might be easier to read without whitespace https://github.com/influxdata/influxdb_iox/pull/2525/files?w=1

server/src/db/replay.rs
```diff
@@ -2588,16 +2612,10 @@ mod tests {
         // create write buffer w/ sequencer 0 and 1
         let write_buffer_state =
             MockBufferSharedState::empty_with_n_sequencers(NonZeroU32::try_from(2).unwrap());
-        let write_buffer = MockBufferForReading::new(write_buffer_state, None).unwrap();
+        let mut write_buffer = MockBufferForReading::new(write_buffer_state, None).unwrap();
 
         // create DB
-        let db = TestDb::builder()
```
tustvold, line 2594: TestDb no longer has any involvement in the creation and/or management of write buffer consumers

server/src/db.rs
```diff
@@ -418,7 +307,10 @@ pub(crate) struct DatabaseToCommit {
     pub(crate) preserved_catalog: PreservedCatalog,
     pub(crate) catalog: Catalog,
     pub(crate) rules: Arc<DatabaseRules>,
-    pub(crate) write_buffer: Option<WriteBufferConfig>,
+
+    /// TODO: Move onto Database
+    pub(crate) write_buffer_producer: Option<Arc<dyn WriteBufferWriting>>,
```
tustvold, line 312: I hope to remove this as a follow on PR, but this PR was already fairly large

server/src/db.rs
```diff
@@ -1781,61 +1465,37 @@ mod tests {
             .unwrap();
         metrics
             .has_metric_family("write_buffer_read_bytes_total")
-            .with_attributes(&[
-                ("db_name", "placeholder"),
-                ("svr_id", "1"),
```
tustvold, line 1786: Note svr_id has gone, this will happen as part of the metrics migration anyway so I wasn't too concerned about plumbing this through

server/src/database.rs
```diff
@@ -931,19 +951,44 @@ struct DatabaseStateCatalogLoaded {
...
+                let mut consumer = write_buffer_factory
+                    .new_config_read(
+                        shared.config.server_id,
+                        shared.config.name.as_str(),
+                        connection,
+                    )
+                    .await
+                    .context(CreateWriteBuffer)?;
+
+                db.perform_replay(self.replay_plan.as_ref().as_ref(), consumer.as_mut())
```
tustvold, line 976: This member function feels off, but it was a fair amount of code-churn to address, perhaps a follow on PR

server/src/db.rs
```diff
@@ -915,210 +800,15 @@ impl Db {
...
+            _ = object_store_cleanup_loop => error!("object store cleanup exited - db worker bailing out"),
+            _ = shutdown.cancelled() => info!("db worker shutting down"),
         }
 
-        info!("finished background worker");
-    }
-
-    /// This is used to take entries from a `Stream` and put them in the mutable buffer, such as
-    /// streaming entries from a write buffer.
-    async fn stream_in_sequenced_entries<'a>(
```
tustvold, line 961: This code is moved pretty much verbatim to WriteBufferConsumer

server/src/write_buffer/metrics.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Updates:
+    ///
+    /// - bytes read
+    /// - last sequence number
+    /// - lag
+    /// - min ts
+    /// - max ts
+    /// - ingest ts
+    pub fn record_write(&mut self, sequenced_entry: &entry::SequencedEntry, watermark: u64) {
+        let entry = sequenced_entry.entry();
```
tustvold, line 111: This is moved out of stream_in_sequenced_entries for better readability

server/src/database.rs
```diff
@@ -931,30 +952,66 @@ struct DatabaseStateCatalogLoaded {
...
             db,
+            write_buffer_consumer,
             provided_rules: Arc::clone(&self.provided_rules),
         })
     }
 }
 
 #[derive(Debug, Clone)]
-struct DatabaseStateInitialized {
+pub struct DatabaseStateInitialized {
```
tustvold, line 999: It was always a design intent that these struct be public with private members, just didn't have a motivating use-case. I now do :tada:

write_buffer/src/mock.rs
```diff
@@ -49,8 +52,8 @@ impl EntryResVec {
 
     /// Register a waker to be notified when a new entry is pushed
     pub fn register_waker(&mut self, waker: &Waker) {
-        for waker in &self.wait_list {
-            if waker.will_wake(waker) {
+        for wait_waker in &self.wait_list {
```
tustvold, line 55: Stupid bug from #2520

server/src/write_buffer.rs
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+            watermark_last_updated = Some(now);
+        }
+
+        std::mem::drop(red_observation);
+
+        metrics.record_write(&sequenced_entry, watermark)
+    }
+}
```
crepererum, line 186: could we move the write buffer consumer tests as well?

crepererum (approved):
Overall good, this might need a few follow-ups though.

server/src/db.rs
```diff
@@ -400,11 +295,6 @@ pub struct Db {
     /// - to keep the lifecycle state (e.g. the number of running compactions) around
     lifecycle_policy: tokio::sync::Mutex<Option<::lifecycle::LifecyclePolicy<WeakDb>>>,
 
-    /// Flag to stop background worker from reading from the write buffer.
-    ///
-    /// TODO: Move write buffer read loop out of Db.
-    no_write_buffer_read: AtomicBool,
```
alamb, line 406: This removal seems like an improvement to me

alamb (commented):
Looks like a good change to me. I think @crepererum 's comments / review is more important to mine, but what I read also looks good 👍

write_buffer/src/config.rs
```diff
@@ -77,30 +78,7 @@ impl WriteBufferConfigFactory {
             .ok_or_else::<WriteBufferError, _>(|| format!("Unknown mock ID: {}", name).into())
     }
 
-    /// Create new config.
```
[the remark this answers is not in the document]
tustvold: It's partly a quirk of not moving the producer out of DB in this PR, but at least currently I need to create the Writer and Reader at different stages in the initialization process. The Writer needs to be created when Db is created before Replay, whereas the Reader needs to be created as part of running Replay (to avoid lock shenanigans).

Longer-term I thought it would be simpler to reason about them separately and avoid some impossible code-paths, where we already know we are creating a consumer, etc...

server/src/db/replay.rs
```diff
@@ -14,7 +14,7 @@ use persistence_windows::{
     persistence_windows::PersistenceWindows,
```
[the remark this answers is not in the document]
tustvold: Agreed, I actually did this at one point, but it was using some non-public bits of Db and this PR was already a chonker

server/src/database.rs
```diff
@@ -931,30 +952,66 @@ struct DatabaseStateCatalogLoaded {
...
+        &self,
+        shared: &DatabaseShared,
+    ) -> Result<DatabaseStateInitialized, InitError> {
         let db = Arc::clone(&self.db);
 
-        db.perform_replay(self.replay_plan.as_ref().as_ref())
-            .await
-            .context(Replay)?;
-
         // TODO: Pull write buffer and lifecycle out of Db
```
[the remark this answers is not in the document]
tustvold: Yeah, the producer still needs to get pulled out though

write_buffer/src/config.rs
```diff
@@ -77,30 +78,7 @@ impl WriteBufferConfigFactory {
...
-                ))),
-            },
-            None => None,
-        };
-
-        Ok(cfg)
-    }
-
-    async fn new_config_write(
+    pub async fn new_config_write(
```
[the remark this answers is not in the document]
tustvold: Will fix

server/src/write_buffer.rs
tustvold: Good shout :+1:

PR #2527, opened 2021-09-14
refactor: move `predicate` into its own crate

crepererum:
Two reasons:

1. I wanna decouple `parquet_file` from `query` (nearly done, needs a
   small follow-up PR).
2. `predicate` will have more and more features (like serialization)
   which justifies a new home


NGA-TRAN (approved):
I like this isolated PR that focuses to only move Predicate and regex to their own crate.

PR #2528, opened 2021-09-14
feat: compress encoded parquet metadata

crepererum:
Depending on the number of columns, this should safe between 60% and 75%.


tustvold (approved):

Cargo.lock
```diff
@@ -2911,6 +2911,7 @@ dependencies = [
  "tokio-stream",
  "tracker",
  "uuid",
+ "zstd",
```
tustvold, line 2914: Looks like this isn't a new dependency which is :ok_hand:

mkmik (approved):

parquet_file/src/metadata.rs
```diff
@@ -228,6 +228,12 @@ pub enum Error {
         expected
     ))]
     IoxMetadataVersionMismatch { actual: u32, expected: Vec<u32> },
+
+    #[snafu(display("Cannot encode ZSTD message: {}", source))]
```
alamb, line 232: It might make sense to note in this error that the problem was encoding parquet metadata (as the use of zstd is likely an implementation detail that could be confusing)

```suggestion
    #[snafu(display("Cannot encode parquet metadata: {}", source))]
```

The same comment applies for the decoding failure.

This is likely not a big deal as either of these errors doesn't seem recoverable so there isn't much a user is going to be able to do about them

alamb (approved):

crepererum:
Tests are failing because I need to adjust a few more sizes, but nothing fundamental.

Cargo.lock
crepererum: right, zstd is already used by parquet and a somewhat well established and performant format. that's why I chose it.

parquet_file/src/metadata.rs
crepererum: fixed

PR #2530, opened 2021-09-14
fix: reset Database background worker on deletion

tustvold:
#2424 added the ability to delete a Database by writing a tombstone file. However, this would leave the background worker in a zombie state as it wouldn't trigger the Database background worker in any way. In practice this would mean that the `Db` would continue to exist and be running, until the server was shut down. Whilst not ideal, where this really becomes a problem is when trying to re-initialize the Database as encountered in #2464.

This PR instead makes it so that deleting a Database resets the worker back to the "initialization-phase". This means that if a subsequent action updates the DatabaseState to the point where it could be initialized, the worker will just pick it up and continue where it left off.

Unfortunately the way `Db` currently handles its background worker, or more specifically the way it doesn't, makes this a bit cumbersome than it would otherwise be, but it isn't too terrible.


server/src/database.rs
```diff
@@ -519,8 +526,47 @@ async fn background_worker(shared: Arc<DatabaseShared>) {
...
+                info!(db_name=%shared.config.name, "database no longer initialized");
+                break;
+            }
+
+            let shutdown = shared.shutdown.cancelled().fuse();
+            futures::pin_mut!(shutdown);
+
+            // We must use `futures::select` as opposed to the often more ergonomic `tokio::select`
+            // Because of the need to "re-use" the background worker future
+            // TODO: Make Db own its own background loop (or remove it)
```
tustvold, line 554: This would eliminate all the pin_mut and fuse nonsense

server/src/database.rs
```diff
@@ -519,8 +526,47 @@ async fn background_worker(shared: Arc<DatabaseShared>) {
...
-        // TODO: Pull background_worker out of `Db`
-        db.background_worker(shared.shutdown.clone()).await
+        let db_shutdown = CancellationToken::new();
+        let db_worker = db.background_worker(db_shutdown.clone()).fuse();
+        futures::pin_mut!(db_worker);
+
+        // This inner loop runs until either:
+        //
+        // - Something calls `Database::shutdown`
+        // - The Database transitions away from `DatabaseState::Initialized`
```
alamb, line 536:
```suggestion
        // - The Database transitions to a state where it is no longer initialized 
        //  (no longer with [`DatabaseState::Initialized`])
```

server/src/database.rs
```diff
@@ -498,9 +499,15 @@ struct DatabaseShared {
 async fn background_worker(shared: Arc<DatabaseShared>) {
     info!(db_name=%shared.config.name, "started database background worker");
 
-    initialize_database(shared.as_ref()).await;
+    // The background loop runs until `Database::shutdown` is called
+    while !shared.shutdown.is_cancelled() {
+        initialize_database(shared.as_ref()).await;
```
alamb, line 504: i think logging here (or perhaps moving the `info!()` from above down into this loop would help -- so that we can tell from the logs when the database went back to try re-initializing.

alamb (approved):
This looks fine to me and makes sense intellectually. I think without a test it is hard for me to be sure. Is there some way to verify that database re-initialization would work upon deletion prior to @carols10cents  trying to hook it up in the `delete-database` code?

tustvold:
I added a database delete restore test in https://github.com/influxdata/influxdb_iox/pull/2530/commits/c6807431cc4fb054f25a10f6c03a50faaf0115a5 that might be useful for @carols10cents and also confirms this functionality at least appears to work correctly. Note this currently hacks around the functionality added in #2464 as I didn't want to make this PR dependent, but should be migrated across once the functionality exists

server/src/database.rs
tustvold: I think this is provided by initialize_database which will log "database initialization started"

alamb: Ah, makes sense

PR #2536, opened 2021-09-15
feat: migrate write buffer metrics

tustvold:
This migrates the write buffer metrics instrumentation to the new `metric` crate. It preserves the somewhat inconsistent naming, but it is a lot more obvious with the new crate and can be addressed in a follow on PR if we wish to.

This also added a utility method `Entry::table_batches` to facilitate iterating over the table batches in an entry. This will be useful in a number of other places, and aside from being less verbose, avoids a load of unnecessary buffering.


entry/src/entry.rs
```diff
@@ -761,6 +761,35 @@ impl Entry {
             None => None,
         }
     }
+
+    /// Returns an iterator over the PartitionWrite in this entry
+    ///
+    /// TODO: Replace partition_writes with this
```
tustvold, line 767: There is no reason for partition_writes to continue to exist, but I wanted to avoid churn as part of this PR

entry/src/entry.rs
```diff
@@ -811,13 +840,13 @@ pub struct PartitionWrite<'a> {
 }
 
 impl<'a> PartitionWrite<'a> {
-    pub fn key(&self) -> &str {
+    pub fn key(&self) -> &'a str {
```
tustvold, line 843: The returned data is valid for the lifetime of the entry, i.e. `'a` not the lifetime of the borrow of self. This allows borrowing out of `PartitionWrite` for longer than the `PartitionWrite` itself lives

server/src/write_buffer.rs
```diff
@@ -333,63 +323,92 @@ mod tests {
             tokio::time::sleep(Duration::from_millis(100)).await;
         }
 
-        // check: metrics
-        // We need to do that BEFORE shutting down the background loop because gauges would be dropped and resetted otherwise
```
tustvold, line 337: This limitation is no longer the case, we have true Gauges :tada:

entry/src/entry.rs
alamb: makes sense

server/src/write_buffer.rs
```diff
@@ -333,63 +323,92 @@ mod tests {
...
-            .unwrap();
-
         // do: stop background task loop
         shutdown.cancel();
         join_handle.await.unwrap();
 
         consumer.shutdown();
         consumer.join().await.unwrap();
 
+        let metrics = test_db.metrics_registry_v2;
```
alamb, line 333: I double checked the same values come out 👍

server/src/write_buffer.rs
```diff
@@ -34,16 +33,11 @@ impl WriteBufferConsumer {
...
     ) -> Self {
         let shutdown = CancellationToken::new();
 
-        let ingest_domain = registry.register_domain_with_attributes(
-            "write_buffer",
-            vec![KeyValue::new("db_name", db.rules().name.to_string())],
-        );
-
-        let ingest_metrics = metrics::WriteBufferIngestMetrics::new(Arc::new(ingest_domain));
+        let ingest_metrics = WriteBufferIngestMetrics::new(registry, db.rules().name.to_string());
```
alamb, line 40:
```suggestion
        let ingest_metrics = WriteBufferIngestMetrics::new(registry, &db.rules().name);
```

probably doesn't matter but I think the argument is `impl Into<String>`

trace_http/src/metrics.rs
```diff
@@ -58,14 +58,14 @@ impl MetricsCollection {
...
-    ok: U64Counter,
-    client_error: U64Counter,
-    server_error: U64Counter,
-    aborted: U64Counter,
-
-    duration_ok: DurationHistogram,
-    duration_client_error: DurationHistogram,
-    duration_server_error: DurationHistogram,
+    /// Counts of un-aborted requests
+    request_count: ResultMetric<U64Counter>,
```
alamb, line 62: 👍  for refactoring into `ResultMetric`

FYI @e-dard I think `ResultMetric` is the new `RED` ! 🔴

alamb (approved):
nice cleanup -- I checked the counter values in the tests and they all look good to me

PR #2537, opened 2021-09-15
feat: `Predicate` serialization

### LINKED ISSUE #2493: Serialize related Delete Expression to store in Catalog
# Rationale 
As part of #2396  we need to store delete predicate in the catalog for us to eliminate data when reading and compacting.  After a discussion, we decide to serialize the Delete Expr to protobuf messages/structures before storing to avoid backward incompatibility in the future. 

The Delete Expr is subset of DF Expr and implemented in this PR #2488

crepererum:
Closes #2493.

### DIFF
--- a/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
+++ b/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
@@ -74,8 +74,7 @@ message Scalar {
   oneof value {
     bool value_bool = 1;
     int64 value_i64 = 2;
-    uint64 value_u64 = 3;
-    double value_f64 = 4;
-    string value_string = 5;
+    double value_f64 = 3;
+    string value_string = 4;
   }
 }
--- a/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
+++ b/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
@@ -1,0 +1,81 @@
+syntax = "proto3";
+package influxdata.iox.catalog.v1;
+
+// Represents a parsed predicate for evaluation by the InfluxDB IOx query engine.
+message Predicate {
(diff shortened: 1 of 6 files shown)


generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
```diff
@@ -0,0 +1,81 @@
...
+  // Inequality (`!=`).
+  OP_NE = 2;
+}
+
+// Scalar value of a certain type.
+message Scalar {
+  oneof value {
+    bool value_bool = 1;
+    int64 value_i64 = 2;
+    uint64 value_u64 = 3;
```
NGA-TRAN, line 77: FYI. DF parser seems not generate u64. Currently, we only generate i64 and f64

predicate/src/serialize.rs
```diff
@@ -0,0 +1,263 @@
...
+            .table(table_name)
+            .timestamp_range(parse_delete_pred.start_time, parse_delete_pred.stop_time);
+
+        for expr in parse_delete_pred.predicate {
+            del_predicate_builder = del_predicate_builder.add_expr(expr);
+        }
+
+        del_predicate_builder.build()
+    }
+}
```
NGA-TRAN, line 263: Maybe add a negative test that error happens because of unsupported string?

NGA-TRAN (approved):
LGTM

predicate/src/serialize.rs
crepererum: done

generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
crepererum: removed that variant, we can easily add it back later

PR #2541, opened 2021-09-15
test: even more time-related lifting

crepererum:
Lift a few `Utc::now()` calls further and narrow down checks in tests.
Also avoid a few `<` comparisons which might not always hold.

### DIFF
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -1332,7 +1332,7 @@ mod tests {
         assert_store_sequenced_entry_failures,
         db::{
             catalog::chunk::ChunkStage,
-            test_helpers::{run_query, try_write_lp, write_lp},
+            test_helpers::{run_query, try_write_lp, write_lp, write_lp_with_time},
         },
         utils::{make_db, TestDb},
     };
@@ -1876,9 +1876,8 @@ mod tests {
         let test_db = make_db().await;
         let db = Arc::new(test_db.db);
 
-        let time0 = Utc::now();
-        write_lp(db.as_ref(), "cpu bar=1 10").await;
-        let time1 = Utc::now();
+        let t_write1 = Utc::now();
+        write_lp_with_time(db.as_ref(), "cpu bar=1 10", t_write1).await;
 
         let partition_key = "1970-01-01T00";
         let mb_chunk = db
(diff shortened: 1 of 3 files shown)


server/src/db.rs
```diff
@@ -2494,72 +2495,35 @@ mod tests {
...
-            summary
-        );
-        assert!(
-            summary.time_closed.unwrap() < after_close,
-            "summary; {:#?}",
-            summary
-        );
+        assert_eq!(summary.time_of_first_write, t_first_write);
+        assert_eq!(summary.time_of_last_write, t_second_write);
+        assert_eq!(summary.time_closed.unwrap(), t_close);
```
tustvold, line 2518: I could be missing something but the vast majority of the changes in this PR appear to be just for this single assertion. This feels a little hard to justify? Perhaps we could just check the time_closed is greater or equal to time_of_last_write and matches what is recorded on the chunk and call it a day?

crepererum: @tustvold done, PTAL

server/src/db/lifecycle/compact.rs
```diff
@@ -158,15 +158,13 @@ mod tests {
...
-
-        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=1 10").await;
-        write_lp(db.as_ref(), "cpu,tag1=asfd,tag2=foo bar=2 20").await;
-        write_lp(db.as_ref(), "cpu,tag1=bingo,tag2=foo bar=2 10").await;
-        write_lp(db.as_ref(), "cpu,tag1=bongo,tag2=a bar=2 20").await;
-        write_lp(db.as_ref(), "cpu,tag1=bongo,tag2=a bar=2 10").await;
-
-        let time1 = Utc::now();
+        let t_first_write = Utc::now();
+        write_lp_with_time(db.as_ref(), "cpu,tag1=cupcakes bar=1 10", t_first_write).await;
```
tustvold, line 162: FYI time_of_first_write is actually the min write timestamp, so this could fail if Utc::now() goes backwards. This is almost certainly fine for a test, but thought I'd mention it

server/src/db.rs
```diff
@@ -2484,72 +2480,35 @@ mod tests {
...
-            summary
-        );
-        assert!(
-            summary.time_closed.unwrap() < after_close,
-            "summary; {:#?}",
-            summary
-        );
+        assert_eq!(summary.time_of_first_write, t_first_write);
+        assert_eq!(summary.time_of_last_write, t_second_write);
+        assert!(t_close_before <= summary.time_closed.unwrap());
```
tustvold, line 2502: There are scenarios where this could fail, e.g. the clock goes backwards but I'd wager we won't see that in practice for a test suite

tustvold (approved):
LGTM - thank you :smile:

server/src/db.rs
crepererum: Let's see if this fails and adjust it if that really happens.

PR #2542, opened 2021-09-15
feat: migrate http ingest metrics

tustvold:
Migrates the ingest metrics to the new metrics crate. This moves the HTTP ingest metrics into the http endpoint where they originate, and moves the entry ingest metrics into Database.

I think the naming of these metrics is a little bit confusing, given they are recorded at different parts of the stack (something clarified but not introduced by this PR), but we can easily revisit this later.


server/src/lib.rs
```diff
@@ -786,85 +728,40 @@ where
         default_time: i64,
     ) -> Result<()> {
         let db = self.db(db_name)?;
+        let rules = db.rules();
```
tustvold, line 731: I took the opportunity to cleanup this method, as for one thing the locks it refers to no longer exist

server/src/lib.rs
```diff
@@ -936,49 +833,27 @@ where
 
     /// Write an entry to the local `Db`
     ///
-    /// TODO: Push this out of `Server` into `Database`
+    /// TODO: Remove this and migrate callers to `Database::write_entry`
```
tustvold, line 836: I plan to do this in a follow up PR

src/influxdb_ioxd/http.rs
```diff
@@ -522,74 +532,31 @@ where
...
+    match server.write_lines(&db_name, &lines, default_time).await {
+        Ok(_) => {
+            lp_metrics.record_write(&db_name, lines.len(), num_fields, body.len(), true);
+            Ok(Response::builder()
+                .status(StatusCode::NO_CONTENT)
+                .body(Body::empty())
+                .unwrap())
+        }
+        Err(server::Error::DatabaseNotFound { .. }) => {
+            debug!(%db_name, %num_lines, "database not found");
```
tustvold, line 544: This is a change, but I think is correct - we don't want to pollute our metrics with a load of typos

data_types/src/database_rules.rs
```diff
@@ -444,7 +443,7 @@ pub struct ShardConfig {
     pub ignore_errors: bool,
     /// Mapping between shard IDs and node groups. Other sharding rules use
     /// ShardId as targets.
-    pub shards: Arc<HashMap<ShardId, Sink>>,
+    pub shards: HashMap<ShardId, Sink>,
```
alamb, line 446: is this just a drive by cleanup?

server/src/database.rs
```diff
@@ -462,6 +509,53 @@ impl Database {
...
+    /// - write it to a write buffer
+    /// - write it to a local `Db`
+    ///
+    pub async fn write_entry(
+        &self,
+        entry: entry::Entry,
+        time_of_write: DateTime<Utc>,
+    ) -> Result<(), WriteError> {
+        let bytes = entry.data().len();
+        let maybe_db = {
```
alamb, line 524: it might make sense to note as a comment here that no `return` is allowed because the error metrics would not be updated.

Otherwise perhaps some sort of RAAI "success counter" abstraction would be valuable (much like `IngestRecorder` from https://github.com/influxdata/influxdb_iox/pull/2536/files#diff-837be3c421bc1306afa1c49dab6941f1cc5359866b0843a2e221064c292ade15R124)

src/influxdb_ioxd/http.rs
```diff
@@ -1169,13 +1212,36 @@ mod tests {
...
+
+        // Perform an invalid write
+        client
+            .post(&post_url)
+            .body(incompatible_lp_data)
+            .send()
+            .await
             .unwrap();
+
+        // This currently results in an InternalServerError which is correctly recorded
```
alamb, line 1235: I reviewed the values in these tests and they appear to be consistent with the prior values as well

src/influxdb_ioxd/http/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+    ingest_bytes_ok: U64Counter,
+
+    /// The number of LP bytes ingested unsuccessfully
+    ingest_bytes_error: U64Counter,
+}
+
+impl LineProtocolMetrics {
+    pub fn new(registry: &metric::Registry) -> Self {
+        Self {
+            ingest_lines_total: registry
```
alamb, line 46: There is a discrepancy between the field name (ingest_**lines**)  and the metric name (ingest_**points**)

I believe the field name is correct and the metric name should be changed to `ingest_lines` and `"total LP lines ingested"`

src/influxdb_ioxd/http/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+                .register_metric("ingest_points", "total LP points ingested"),
+            ingest_fields_total: registry
+                .register_metric("ingest_fields", "total LP field values ingested"),
+            ingest_bytes_total: registry
+                .register_metric("ingest_points_bytes", "total LP points bytes ingested"),
```
alamb, line 51:
```suggestion
                .register_metric("ingest_points_bytes", "total LP bytes ingested"),
```

alamb (approved):
looks like there is a test failure and I think the naming of `LineProtocolMetrics::ingest_lines_total` should be reviewed prior to merge. Otherwise 🚀 

Nice work

data_types/src/database_rules.rs
tustvold: Well the cleanup of the write_lines removed the need for it

src/influxdb_ioxd/http/metrics.rs
tustvold: Yeah, this is what I meant by "naming of the metrics a bit confusing" :laughing: - I will just change it even though it is technically a breaking change

server/src/database.rs
tustvold: I was being lazy :sweat_smile: - done :+1:

Issue #2539, opened 2021-09-15
Create new proto DatabaseStates for `DatabaseObjectStoreLookupError` and  `NoActiveDatabase`


carols10cents:
Right now, [both `DatabaseObjectStoreLookupError` (an error state) and `NoActiveDatabase` (the soft-deleted state) are mapping to `Known`](https://github.com/influxdata/influxdb_iox/blob/27b0c46f79338793bbdc723e7018997079f22db1/server/src/database.rs#L812-L813). `Known` is a non-terminal state but these two are terminal states, so they should have their own return value in the proto.

carols10cents:
This is now fixed by #2543 and #2547.

PR #2543, opened 2021-09-15
fix: report correct DatabaseStateCode

tustvold:
The original cut of the server status API didn't include the error states as discrete states, this was supposed to be changed in #2203 but I copy-pasta'ed the conversion logic and didn't notice that I hadn't updated the tests :facepalm: 

This PR fixes this and also adds the enumerations for the states added since then.

I also opted to remove the manually defined enumeration in data_types as its only purpose seemed to be to define an impl, which can just be defined in generated_types and some duplication avoided.


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -386,6 +389,9 @@ message DatabaseStatus {
 
     // Error during replay
     DATABASE_STATE_REPLAY_ERROR = 7;
+
+    // Error encountered identifying active generation
+    DATABASE_STATE_DATABASE_OBJECT_STORE_LOOKUP_ERROR = 9;
```
tustvold, line 394: This enumeration name makes me irrationally sad, but I couldn't think of a better name. We can always change it later

generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -177,16 +177,16 @@ message DeleteDatabaseResponse {}
...
 // This resource represents a deleted database.
 message DeletedDatabase {
-	// The generation ID of the deleted database.
+  // The generation ID of the deleted database.
   uint64 generation_id = 1;
 
-	// The UTC datetime at which this database was deleted.
-	google.protobuf.Timestamp deleted_at = 2;
+  // The UTC datetime at which this database was deleted.
+  google.protobuf.Timestamp deleted_at = 2;
```
tustvold, line 189: Perhaps we need a format check for protobuf :thinking:

alamb: maybe the `buf` thing we use could be pressed into service

generated_types/protos/influxdata/iox/management/v1/service.proto
alamb: its A DATABASE! A DATABASE!

alamb (approved):

carols10cents (approved):
When I started looking into this, I started adding some cases to ensure we could get some of the other failing database states-- so normally I *would* ask you to add that, but I have it halfway done anyway so I'll add it in a new PR :)

carols10cents:
Connects to #2539

Issue #2549, opened 2021-09-15
Support "sampling" distributed traces to jaeger even if requests do not have a trace id


alamb:
__Use case:__
Sometimes I would like to get detailed execution trace information into Jaeger from IOx while it is processing a request, even the request itself did not have a trace / span id (e.g. the request did not include an HTTP header such as `uber-trace-id:4459495:30434:0:1`) so that I can explore how IOx is working even when the upstream service didn't request tracing.

In other words I would like to trace some arbitrary requests to IOx, not just those that explicitly requested them. 

__Proposal:__
Add a new command line argument `--traces-*` and corresponding environment variable  `TRACES_` that defines some sampling percentage.

TBD what this configuration option should be called and what options it should allow for. I was hoping for something standard but I could not find any such thing quickly.

I doubt we need to support all the sophistication of Jaeger as described in https://www.jaegertracing.io/docs/1.26/sampling/#client-sampling-configuration but it would be cool not to re-invent our own sampling mechanism too.

__Current behaviour:__
Here are the supported options for configuring tracing today `./target/debug/influxdb_iox run --help`:

```
...
       --traces-exporter <traces-exporter>
            Tracing: exporter type
            
            Can be one of: none, jaeger [env: TRACES_EXPORTER=]  [default: none]
        --traces-exporter-jaeger-agent-host <traces-exporter-jaeger-agent-host>
            Tracing: Jaeger agent network hostname
            
            Protocol is Thrift/Compact over UDP.
            
            Only used if `--traces-exporter` is "jaeger". [env: TRACES_EXPORTER_JAEGER_AGENT_HOST=]  [default: 0.0.0.0]
        --traces-exporter-jaeger-agent-port <traces-exporter-jaeger-agent-port>
            Tracing: Jaeger agent network port
            
            Protocol is Thrift/Compact over UDP.
            
            Only used if `--traces-exporter` is "jaeger". [env: TRACES_EXPORTER_JAEGER_AGENT_PORT=]  [default: 6831]
        --traces-exporter-jaeger-service-name <traces-exporter-jaeger-service-name>
            Tracing: Jaeger service name.
            
            Only used if `--traces-exporter` is "jaeger". [env: TRACES_EXPORTER_JAEGER_SERVICE_NAME=]  [default: iox-
            conductor]
...
```

__Context__
[...]

PR #2529, opened 2021-09-14
feat: Translate DataFusion execution metrics to IOx Spans

alamb:
Resolves https://github.com/influxdata/influxdb_iox/issues/2505

# Rationale
This PR connects DataFusion's metrics added in https://github.com/apache/arrow-datafusion/issues/679 to our distributed tracing (e.g. Jaeger) system.

# Changes
1. Translate the DataFusion `metrics` for a particular plan to IOx spans if there is a tracing context
2. Tests for same

## Notes:
1. Currently the metrics are not hooked up for lifecycle actions (as there are no span contexts available). It might be nice to have the ability to see how execution progresses for the lifecycle actions as well. @tustvold and I will sort out how to do so if it turns out to be needed.
2. Several common operators (e.g. `ProjectionExec` and `LimitExec`) use the start/stop timestamps of their parents as they have not yet been annotated correctly. I need to finish that under the aegis of https://github.com/apache/arrow-datafusion/issues/866
3. ~There is no end-to-end test for this functionality (I didn't see one for tracing yet)~

# Examples:

You can test this out yourself by following the instructions [here](https://github.com/influxdata/influxdb_iox/blob/main/CONTRIBUTING.md#running-jaeger--tracing-locally).

Here are some screenshots of traces sent to Jaeger locally:

Screen Shot 2021-09-13 at 12 05 54 PM

Among other things, it also exports how many rows come out of each operator as well as the amount of "compute time" is spent (as the spans show start/stop timestamps).

Screen Shot 2021-09-13 at 12 07 52 PM

PR #2551, opened 2021-09-15
refactor: Add SpanContext:new() to make a new span

alamb:
# Rationale:
As suggested by @tustvold on #2529 in https://github.com/influxdata/influxdb_iox/pull/2529/files#r709426141, rather than duplicating this code it would be better to have a function that can create root spans to:
1. Reduce code duplication
2. Prepare to support sampling (as described in https://github.com/influxdata/influxdb_iox/issues/2549)

# Changes:
1. Add a `SpanContext::new_root()` function which creates a new root span with random trace id and span id.
2. Remove other duplicated code


trace/src/ctx.rs
```diff
@@ -55,6 +55,22 @@ pub struct SpanContext {
 }
 
 impl SpanContext {
+    /// Create a new root span context, sent to to `collector`. The
```
tustvold, line 58:
```suggestion
    /// Create a new root span context, sent to `collector`. The
```

tustvold (approved):

PR #2553, opened 2021-09-15
fix: Capture query execution traces for storage gRPC queries as well

alamb:
# Rationale:
As spotted by @tustvold  on #2529 in https://github.com/influxdata/influxdb_iox/pull/2529/files#r709428990 detailed query execution information is not included when running series sets plans (aka storage gRPC queries)

# Changes
1. renames execute --> execute_stream and execute_partitioned to execute_stream_partitioned to match DataFusion names
1. Route all executions through `execute_stream`
3. Wrap results in a `TracingStream` that reports metrics on drop

This also has the upside that callers can't forget to record the metrics.

I couldn't think of a good way to do a proper end-to-end test of storage gRPC because the storage gRPC tests aren't actually hooked up to a real database (they use a mock).

At some point perhaps we can wire up a proper end-to-end test that gets traces out of the influxdb_ioxd server to jaeger or something similar, and we could test it there.

Note: tests for https://github.com/influxdata/influxdb_iox/pull/2529 cover this case as it now goes through the (unified) codepath


query/src/exec/query_tracing.rs
```diff
@@ -1,15 +1,67 @@
...
+    DisplayFormatType, ExecutionPlan, RecordBatchStream, SendableRecordBatchStream,
 };
+use futures::StreamExt;
 use observability_deps::tracing::debug;
-use trace::span::Span;
+use trace::span::{Span, SpanRecorder};
+
+/// Stream wrapper that records DataFusion `MetricSets` into IOx
+/// [`Span`]s when it is dropped.
+pub(crate) struct TracedStream {
```
alamb, line 18: This is the major "innovation" -- when this stream is drop, all DataFusion metrics get reported

query/src/exec/context.rs
```diff
@@ -272,45 +273,65 @@ impl IOxExecutionContext {
     /// Executes the logical plan using DataFusion on a separate
     /// thread pool and produces RecordBatches
     pub async fn collect(&self, physical_plan: Arc<dyn ExecutionPlan>) -> Result<Vec<RecordBatch>> {
-        let ctx = self.child_ctx("collect");
         debug!(
             "Running plan, physical:\n{}",
             displayable(physical_plan.as_ref()).indent()
         );
+        let ctx = self.child_ctx("collect");
+        let stream = ctx.execute_stream(physical_plan).await?;
```
alamb, line 281: route this code path through `execute_stream` as well so we can get metrics for all of them

NGA-TRAN (approved):
The code does what it was described in the "Changes"  section. I am not familiar with the tracing area and as pointed out that there is no clear end-to-end tests for storage gPRC for reviewer to verify, I think @tustvold may want to review this

src/influxdb_ioxd.rs
```diff
@@ -751,11 +751,14 @@ mod tests {
         let sql_span = child(planner_span, "sql").unwrap();
         let prepare_sql_span = child(sql_span, "prepare_sql").unwrap();
         child(prepare_sql_span, "prepare_plan").unwrap();
+        println!("spans are {:#?}", spans);
```
tustvold, line 754: Did you intend to keep this?

tustvold (approved):
Big fan of this :+1: 

Minor nits that maybe are me caring about something that doesn't really matter, but I still wince a little when I see unnecessarily boxed and dyn-ified streams/futures :laughing:

query/src/exec/context.rs
```diff
@@ -11,7 +11,8 @@ use datafusion::{
     logical_plan::{LogicalPlan, UserDefinedLogicalNode},
     physical_plan::{
         coalesce_partitions::CoalescePartitionsExec,
-        collect, displayable,
+        common::collect,
```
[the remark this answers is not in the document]
alamb: Probably b/c we didn't know any better (I was following the model in DataFusion). I will update IOx and then submit a patch upstream as well to DataFusion

src/influxdb_ioxd.rs
alamb: I personally found it very handy for debugging when this test failed, but it is probably a good idea to remove it

query/src/exec/context.rs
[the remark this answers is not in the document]
alamb: Actually after some thought and porting IOx to use `try_collect` and `err_into()`, I think it makes sense to keep `collect` in DataFusion to make the code more "ergonomic" (as it skips the error conversion which is annoying and very non obvious to Rust newcomers, in my opinion)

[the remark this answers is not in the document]
alamb: Added in a95952aa for IOx

alamb:
> there is no clear end-to-end tests for storage gPRC for reviewer to verify,

I have a proposal in https://github.com/influxdata/influxdb_iox/pull/2594

PR #2555, opened 2021-09-15
fix: Ensure byte trimmed floats can handle arbitrary float literals

e-dard:
This PR addresses a defect I identified when working on a new timestamp compression feature.

As far as I'm aware we have not hit upon this defect during internal testing yet, likely because we primarily test via the gRPC API, and with that API you generally do not filter on field columns.

The defect applies to `f64` columns in the Read Buffer that have been encoded using the Byte Trimming encoder. Recall that this encoding can store a column of `f64` values as integers with a smaller physical footprint, subject to the input values all being natural numbers.

For example: `[1.0, 20.0, 30.0] -> [1_u8, 2, 30]`.

The defect is as follows: prior to this change, if you provided a predicate along these lines to the read buffer `WHERE my_col >= 20.4` the column encoding would first convert the logical representation of the literal (an `f64` with a value of `20.4`) into the encoded physical representation (a `u8` with a value of `20`). The values `[20.0, 30.0]` would then satisfy this predicate. This is clearly incorrect. 

Realising that this will also be a problem for some future encodings I am working on I decided the best course of action would be to add a new method to the `transcoder` trait that would allow a transcoder to describe how to encode a literal used for a predicate. This method is called `encode_comparable`.

For the `FloatByteTrimmer` encoding in question the implementation of `encode_comparable` considers the comparison operator and returns a physical representation and comparison operator that ensures that the correct values are evaluated on the column. This allows us to continue working direclty on the compressed represenation. In the above example of `WHERE my_col >= 20.4` the encoding would convert the predicate operator and literal to `> 20_u8` ensuring the correct values satisfy the predicate (`[30.0]`).

### DIFF
--- a/read_buffer/src/column.rs
+++ b/read_buffer/src/column.rs
@@ -2068,6 +2068,40 @@ mod test {
     }
 
     #[test]
+    fn row_ids_filter_float_trimmed() {
+        let input = &[100.0, 200.0, 300.0, 2.0, 22.0, 30.0];
+
+        let col = Column::from(&input[..]);
+        let mut row_ids = col.row_ids_filter(
+            &cmp::Operator::Equal,
+            &Value::from(200.0),
+            RowIDs::new_bitmap(),
+        );
+        assert_eq!(row_ids.unwrap().to_vec(), vec![1]);
+
+        row_ids = col.row_ids_filter(
+            &cmp::Operator::LT,
+            &Value::from(64000.0),
+            RowIDs::new_bitmap(),
+        );
+        assert!(matches!(row_ids, RowIDsOption::All(_)));
+
+        row_ids = col.row_ids_filter(
+            &cmp::Operator::GTE,
+            &Value::from(-1_000_000.0),
+            RowIDs::new_bitmap(),
+        );
+        assert!(matches!(row_ids, RowIDsOption::All(_)));
+
+        row_ids = col.row_ids_filter(
+            &cmp::Operator::NotEqual,
+            &Value::from(1_000_000.3),
+            RowIDs::new_bitmap(),
+        );
+        assert!(matches!(row_ids, RowIDsOption::All(_)));
+    }
+
+    #[test]
     fn row_ids_range() {
         let input = &[100_i64, 200, 300, 2, 200, 22, 30];
 
--- a/read_buffer/src/column/encoding/scalar/fixed.rs
+++ b/read_buffer/src/column/encoding/scalar/fixed.rs
@@ -7,6 +7,7 @@
 //! allow results to be emitted as some logical type `L` via a transformation
 //! `T`.
 use either::Either;
+use observability_deps::tracing::debug;
 use std::cmp::Ordering;
 use std::fmt::Debug;
 use std::marker::PhantomData;
@@ -373,14 +374,36 @@ where
         Some(self.transcoder.decode(max))
     }
 
-    fn row_ids_filter(&self, value: L, op: &cmp::Operator, dst: RowIDs) -> RowIDs {
-        let value = self.transcoder.encode(value);
+    fn row_ids_filter(&self, value: L, op: &cmp::Operator, mut dst: RowIDs) -> RowIDs {
+        debug!(value=?value, operator=?op, encoding=?ENCODING_NAME, "row_ids_filter");
+        let (value, op) = match self.transcoder.encode_comparable(value, *op) {
+            Some((value, op)) => (value, op),
+            None => {
+                // The value is not encodable. This can happen with the == or !=
+                // operator. In the case of ==, no values in the encoding could
+                // possible satisfy the expression. In the case of !=, all
+                // values would satisfy the expression.
+                dst.clear();
+                return match op {
+                    cmp::Operator::Equal => dst,
+                    cmp::Operator::NotEqual => {
+                        dst.add_range(0, self.num_rows());
+                        dst
+                    }
+                    op => panic!("operator {:?} not expected", op),
+                };
+            }
+        };
+        debug!(value=?value, operator=?op, encoding=?ENCODING_NAME, "row_ids_filter encoded expr");
+
+        // N.B, the transcoder may have changed the operator depending on the
+        // value provided.
         match op {
             cmp::Operator::GT => self.row_ids_cmp_order(&value, PartialOrd::gt, dst),
             cmp::Operator::GTE => self.row_ids_cmp_order(&value, PartialOrd::ge, dst),
             cmp::Operator::LT => self.row_ids_cmp_order(&value, PartialOrd::lt, dst),
             cmp::Operator::LTE => self.row_ids_cmp_order(&value, PartialOrd::le, dst),
-            _ => self.row_ids_equal(&value, op, dst),
+            _ => self.row_ids_equal(&value, &op, dst),
         }
     }
 
--- a/read_buffer/src/column/encoding/scalar/fixed.rs
+++ b/read_buffer/src/column/encoding/scalar/fixed.rs
@@ -413,8 +413,16 @@ where
         right: (L, &cmp::Operator),
         dst: RowIDs,
     ) -> RowIDs {
-        let left = (self.transcoder.encode(left.0), left.1);
-        let right = (self.transcoder.encode(right.0), right.1);
+        debug!(left=?left, right=?right, encoding=?ENCODING_NAME, "row_ids_filter_range");
+        let left = self
+            .transcoder
+            .encode_comparable(left.0, *left.1)
+            .expect("transcoder must return Some variant");
+        let right = self
+            .transcoder
+            .encode_comparable(right.0, *right.1)
+            .expect("transcoder must return Some variant");
+        debug!(left=?left, right=?right, encoding=?ENCODING_NAME, "row_ids_filter_range encoded expr");
 
         match (&left.1, &right.1) {
             (cmp::Operator::GT, cmp::Operator::LT)
@@ -425,8 +433,8 @@ where
             | (cmp::Operator::LT, cmp::Operator::GTE)
             | (cmp::Operator::LTE, cmp::Operator::GT)
             | (cmp::Operator::LTE, cmp::Operator::GTE) => self.row_ids_cmp_range_order(
-                (&left.0, Self::ord_from_op(left.1)),
-                (&right.0, Self::ord_from_op(right.1)),
+                (&left.0, Self::ord_from_op(&left.1)),
+                (&right.0, Self::ord_from_op(&right.1)),
                 dst,
             ),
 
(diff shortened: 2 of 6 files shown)


read_buffer/src/column.rs
```diff
@@ -2067,6 +2067,40 @@ mod test {
         assert!(matches!(row_ids, RowIDsOption::All(_)));
     }
 
+    #[test]
+    fn row_ids_filter_float_trimmed() {
```
e-dard, line 2071: This test validates it is impossible to provide a predicate literal outside of the bounds of the physical type used to store the compressed columnar data.

read_buffer/src/column/encoding/scalar/fixed.rs
```diff
@@ -373,14 +374,36 @@ where
         Some(self.transcoder.decode(max))
     }
 
-    fn row_ids_filter(&self, value: L, op: &cmp::Operator, dst: RowIDs) -> RowIDs {
-        let value = self.transcoder.encode(value);
+    fn row_ids_filter(&self, value: L, op: &cmp::Operator, mut dst: RowIDs) -> RowIDs {
```
e-dard, line 377: Instead of just encoding the provided literal (`L` stands for logical type) into a physical type using `encode`, which in the case of the `FloatByteTrimmer` would convert e.g., `30.3` into `30.0`, we instead use `encode_comparable` and then provide that literal and comparison operator to the column predicate evaluator.

read_buffer/src/column/encoding/scalar/fixed.rs
```diff
@@ -373,14 +374,36 @@ where
...
+        debug!(value=?value, operator=?op, encoding=?ENCODING_NAME, "row_ids_filter");
+        let (value, op) = match self.transcoder.encode_comparable(value, *op) {
```
e-dard, line 379: There is no sensible value to encode something like `!= 22.37` where `22.37` is an invalid value for the encoding. So instead we return `None` and handle it as a special case.

read_buffer/src/column/encoding/scalar/fixed_null.rs
```diff
@@ -227,6 +228,43 @@ where
         }
         dst
     }
+
+    // Identify all row IDs that contain a non-null value.
+    fn all_non_null_row_ids(&self, mut dst: RowIDs) -> RowIDs {
```
e-dard, line 233: As a special case for evaluating a predicate where we know all non-null rows will match, but where we were unable to prove this using meta-data alone, we need to return row IDs for all non-null values. The implementation for adding ranges of satisfying rows (rather than rows one at a time) is a common pattern I have developed for the Read Buffer.

read_buffer/src/column/encoding/scalar/rle.rs
```diff
@@ -288,18 +289,35 @@ where
                 let left_cmp_result = next.partial_cmp(left.0);
                 let right_cmp_result = next.partial_cmp(right.0);
 
-                // TODO(edd): eurgh I still don't understand how I got this to
-                // be correct. Need to revisit to make it simpler.
                 let left_result_ok =
-                    !(left_cmp_result == Some(left_op.0) || left_cmp_result == Some(left_op.1));
+                    left_cmp_result == Some(left_op.0) || left_cmp_result == Some(left_op.1);
```
e-dard, line 293: Revisting this logic it was a lot clearer to me 😄. I was probably tired when I originally did this...

read_buffer/src/column/encoding/scalar/transcoders.rs
```diff
@@ -13,7 +14,17 @@ use std::{
...
 pub trait Transcoder<P, L>: Debug + Display {
+    /// A function that encodes a logical value into a physical representation.
     fn encode(&self, _: L) -> P;
+
+    /// A function that attempts to encode a logical value, within the context
+    /// of a comparison operator, into a physical representation.
+    ///
+    /// Implementation should return a suitable operator for the physical
+    /// representation, which may differ from the provided operator.
+    fn encode_comparable(&self, _: L, _: Operator) -> Option<(P, Operator)>;
```
e-dard, line 25: The new method for transcoders - describe how to encode a provided logical type and operator into a physical type and operator.

read_buffer/src/column/encoding/scalar/transcoders.rs
```diff
@@ -91,6 +110,54 @@ macro_rules! make_float_trimmer {
                 v as $type
             }
 
+            fn encode_comparable(&self, v: f64, op: Operator) -> Option<($type, Operator)> {
```
e-dard, line 113: This is basically the fix for the `FloatByteTrimmer`. Hopefully the comments clarify how the implemenation works.

read_buffer/src/column/encoding/scalar/transcoders.rs
```diff
@@ -91,6 +110,54 @@ macro_rules! make_float_trimmer {
...
+                assert!(v <= <$type>::MAX as f64);
```
e-dard, line 114: This assertion must be guaranteed by the user of the encoding, in the Read Buffer's case the column implemenation in `column.rs`. I have added test coverage for this assertion.

read_buffer/src/column/float.rs
```diff
@@ -754,4 +754,303 @@ mod test {
             //assert_eq!(dst.unwrap_vector(), &exp, "example '{} {:?}' failed", op, v);
         }
     }
+
+    #[test]
+    fn row_ids_filter_float_trimmer() {
```
e-dard, line 759: These tests cover all encodings for a Float column, ensuring that they behave correctly regardless of the literal and comparison operator provided.

read_buffer/src/column/encoding/scalar/fixed_null.rs
mkmik: I was surprised that rust didn't come with an more "off-the-shelf" way to express this common idiom, so I played around a bit with it and landed on:

```rust
// Given an iterator of bools, record in `row_ids` all (non-overlapping) runs of "true" values present in the iterator.
fn runs<'a>(mut row_ids: RowIDs, input: impl Iterator<Item = bool> + 'a) -> RowIDs {
    use itertools::Itertools;

    for r in input
        .group_by(|i| *i)
        .into_iter()
        .map(|(_, i)| i.count())
        .scan(0, |acc, n| {
            *acc += n;
            Some(*acc - n..*acc)
        })
        .step_by(2)
    {
        row_ids.add_range(r.start, r.end);
    }
    row_ids
}
```

(full code in https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=7725fe499b1d9bb9026f7afed449291f)

(which is meant to be fed with something like `(0..self.num_rows()).map(|i| self.arr.is_null(i))`, but I didn't typecheck that)

I wonder how far rust zero-cost abstraction can go; didn't compare the emitted asm.

mkmik (commented):
(tomorrow I'll do a proper review, tonight I was just playing with something that caught my eye)

read_buffer/src/column/float.rs
```diff
@@ -754,4 +754,303 @@ mod test {
...
+
+        for enc in cases {
+            _row_ids_filter_float_trimmer(enc)
+        }
+    }
+
+    fn _row_ids_filter_float_trimmer(enc: Box<dyn ScalarEncoding<f64>>) {
+        // [100.0, 200.0, 100.0, 300.0, 400.0]
+        let cases = vec![
+            (100.0, Operator::Equal, vec![0, 2]),          // 100.0, 100.0
```
mkmik, line 792: It's very hard to ensure that those comments are and stay correct (and if they are not they may confuse future maintainer of this test table)

what about doing something like this:

```rust
fn _row_ids_filter_float_trimmer(data: &[f64], enc: Box<dyn ScalarEncoding<f64>>) {
       let cases = vec![
            (100.0, Operator::Equal, vec![0, 2], vec![100.0, 100.0]), 
            (100.0, Operator::NotEqual, vec![1, 3, 4], vec![200.0, 300.0, 400.0]),
....
        for (v, op, exp, vals) in cases {
            let dst = enc.row_ids_filter(v, &op, RowIDs::new_vector());
            assert_eq!(
                dst.unwrap_vector(),
                &exp,
                "example '{} {:?}' failed for {:?}",
                op,
                v,
                enc.name()
            );
            assert_eq!(
                dst.unwrap_vector().iter().map(|i| data[i]),
                &vals,
                "example '{} {:?}' failed for {:?}",
                op,
                v,
                enc.name()
            );
        }
```

?

read_buffer/src/column/float.rs
```diff
@@ -754,4 +754,303 @@ mod test {
...
+            )),
+        ];
+
+        for enc in cases {
+            _row_ids_filter_float_trimmer_with_nulls(enc)
+        }
+    }
+
+    fn _row_ids_filter_float_trimmer_with_nulls(enc: Box<dyn ScalarEncoding<f64>>) {
+        // [100.0, NULL, NULL, 200.0]
```
mkmik, line 878: same here

read_buffer/src/column/float.rs
```diff
@@ -754,4 +754,303 @@ mod test {
...
+            )),
+        ];
+
+        for enc in cases {
+            _row_ids_filter_range_float_trimmer(enc)
+        }
+    }
+
+    fn _row_ids_filter_range_float_trimmer(enc: Box<dyn ScalarEncoding<f64>>) {
+        // [100.0, 200.0, 100.0, 300.0, 400.0]
```
mkmik, line 964: and here

read_buffer/src/column/float.rs
```diff
@@ -754,4 +754,303 @@ mod test {
...
+            )),
+        ];
+
+        for enc in cases {
+            _row_ids_filter_range_float_trimmer_with_nulls(enc)
+        }
+    }
+
+    fn _row_ids_filter_range_float_trimmer_with_nulls(enc: Box<dyn ScalarEncoding<f64>>) {
+        // [100.0, NULL, NULL, 200.0, NULL]
```
mkmik, line 1033: and here

mkmik (approved):
minor comment on test table

read_buffer/src/column/encoding/scalar/fixed_null.rs
e-dard: This is very cool @mkmik. Thanks. I will try and play with it :-)

read_buffer/src/column/encoding/scalar/fixed_null.rs
```diff
@@ -227,6 +228,43 @@ where
...
+            if self.arr.is_null(i) && found {
+                // add the non-null range
+                let (min, max) = (i as u32 - count, i as u32);
+                dst.add_range(min, max);
+                found = false;
+                count = 0;
+                continue;
+            } else if self.arr.is_null(i) {
+                continue;
+            }
```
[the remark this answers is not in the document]
e-dard: 👨‍🍳

read_buffer/src/column/float.rs
e-dard: That's fair, and I'm happy to go through all my tests and change them to use this pattern in a follow on 👍

e-dard: Filed https://github.com/influxdata/influxdb_iox/issues/2558

alamb:
This is a very cool PR, FWIW.

PR #2556, opened 2021-09-16
refactor: split preserved catalog `api` into `core` and `interface`

crepererum:
This makes it clearer which traits and functions users of the preserved
catalog must implement. This also splits the error types into smaller
enums that are easier to understand.

This change should make it easier to implement new functionality (like
capturing delete predicates).

### DIFF
--- a/parquet_file/src/catalog/interface.rs
+++ b/parquet_file/src/catalog/interface.rs
@@ -1,0 +1,99 @@
+//! Abstract interfaces to make different users work with the perserved catalog.
+use std::{collections::HashMap, sync::Arc};
+
+use iox_object_store::{IoxObjectStore, ParquetFilePath};
+use snafu::Snafu;
+
+use crate::metadata::IoxParquetMetaData;
+
+/// Struct containing all information that a catalog received for a new parquet file.
+#[derive(Debug, Clone)]
+pub struct CatalogParquetInfo {
+    /// Path within this database.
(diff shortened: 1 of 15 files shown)


parquet_file/src/catalog/interface.rs
```diff
@@ -0,0 +1,99 @@
...
+
+/// Abstraction over how the in-memory state of the catalog works.
+pub trait CatalogState {
+    /// Input to create a new empty instance.
+    ///
+    /// See [`new_empty`](Self::new_empty) for details.
+    type EmptyInput: Send;
+
+    /// Create empty state w/o any known files.
+    fn new_empty(db_name: &str, data: Self::EmptyInput) -> Self;
```
e-dard, line 74: Does the implementation detail (`_empty`) about the returned `Self` need to be part of the signature if it's the only way to create a `CatalogState`?

Could we just call this `new` given there are no other constructors?

e-dard (approved):
Overall this LGTM. I just had one style suggestion.

parquet_file/src/catalog/interface.rs
crepererum: Yeah, I was thinking about that as well. I'll probably remove the "new empty" thing in a future PR (this very PR here doesn't change the interface, it just moves code around and adjusts the error type a bit).

PR #2557, opened 2021-09-16
feat: remove legacy http list partitions endpoint

tustvold:
This endpoint appears to be an orphan from before we had the management API, among other things it is using the org and bucket mechanism to reference a database which we've since moved away from. I think it can be safely removed


alamb (approved):
I think this was one of the earliest endpoints that was added during initial development of the write path (and when we were still figuring out whether to use protobuf or http endpoints). I agree it can be safely removed

Nice cleanup

cc @pauldix

PR #2559, opened 2021-09-16
feat: alter http query endpoint

tustvold:
I could be wrong on this front, but I don't believe there is an intention to provide a REST HTTP API for IOx anymore. It therefore seems strange, at least to me, to have the query endpoint at a "REST-like" location but without anything else around it.

As such I wonder if we might simplify the query URL, and pass the database name as a query argument. Aside from simplifying the API, it will also allow us to remove routerify as a dependency and archive our fork, as we can just match the HTTP path.

See also #2557 which removes the list partitions endpoint


alamb (approved):
I think this is a fine change until if/when we decide we want a proper REST API. We have so much invested in gRPC I think we should just keep pushing in that direction and remove `routerify` until we have better need for it

alamb:
again, cc @pauldix  who I think found routerify in the first place

PR #2560, opened 2021-09-16
feat: add pbjson bytes support

tustvold:
The [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) defines that bytes fields should be encoded/decoded as base64. This implements that functionality.

Unfortunately there is an outstanding bug in prost that prevents bytes from functioning correctly in maps (see https://github.com/tokio-rs/prost/issues/531) and so I've left out support for that admittedly niche use-case for now


pbjson_test/src/lib.rs
```diff
@@ -197,12 +197,30 @@ mod tests {
         verify_decode(&decoded, "{}");
 
         // Bytes support currently broken
```
crepererum, line 199:
```suggestion
```

pbjson/src/lib.rs
```diff
@@ -38,4 +42,41 @@ pub mod private {
...
+    impl<'de, T> Deserialize<'de> for BytesDeserialize<T>
+    where
+        T: From<Vec<u8>>,
+    {
+        fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+        where
+            D: serde::Deserializer<'de>,
+        {
+            let str: &str = Deserialize::deserialize(deserializer)?;
+            let decoded = base64::decode(str).map_err(serde::de::Error::custom)?;
```
crepererum, line 58:
```suggestion
            let s: &str = Deserialize::deserialize(deserializer)?;
            let decoded = base64::decode(s).map_err(serde::de::Error::custom)?;
```

Could we please not shadow builtins or use types as variable names?

crepererum (approved):
Looks good, at least as far as I can judge the code :D

pbjson_build/src/generator/message.rs
```diff
@@ -717,6 +728,10 @@ fn write_deserialize_field<W: Write>(
                     )?;
                     "v.0"
                 }
+                FieldType::Scalar(ScalarType::Bytes) => {
+                    // https://github.com/tokio-rs/prost/issues/531
```
alamb, line 732: 👍

alamb (approved):

PR #2561, opened 2021-09-16
refactor: `Arc<Vec<...>>` => `Vec<Arc<...>>` for del predicates

crepererum:
The motivations are:

1. The API uses a SINGLE predicate and adds that to many chunks. With
   `Arc<Vec<...>>` you gain nothing, with `Vec<Arc<...>>` the predicate
   is only stored once (in many vectors)
2. While we currently add predicates blindly to all chunks, we can be way
   smarter in the future and prune out tables, partitions or even single
   chunks (based on statistics). With that, it will be rare that many
   chunks share the exact same set of predicates.
3. It would be nice if we could de-duplicate predicates when writing them
   to the preserved catalog without needing to repeat the pruning
   discussed in point 2. This is way easier to implement whan chunks
   exists in `Arc`s.
4. As a side-note: the `Arc<Vec<...>>` wasn't really cloned around but
   instead was created many time. So the new version should be more
   memory efficient out of the box.

### DIFF
--- a/server/src/db/catalog/chunk.rs
+++ b/server/src/db/catalog/chunk.rs
@@ -80,7 +80,7 @@ pub struct ChunkMetadata {
     pub schema: Arc<Schema>,
 
     /// Delete predicates of this chunk
-    pub delete_predicates: Arc<Vec<Predicate>>,
+    pub delete_predicates: Vec<Arc<Predicate>>,
 }
 
 /// Different memory representations of a frozen chunk.
@@ -307,14 +307,14 @@ impl CatalogChunk {
         time_of_last_write: DateTime<Utc>,
         schema: Arc<Schema>,
         metrics: ChunkMetrics,
-        delete_predicates: Arc<Vec<Predicate>>,
+        delete_predicates: Vec<Arc<Predicate>>,
         order: ChunkOrder,
     ) -> Self {
         let stage = ChunkStage::Frozen {
             meta: Arc::new(ChunkMetadata {
                 table_summary: Arc::new(chunk.table_summary()),
                 schema,
-                delete_predicates: Arc::clone(&delete_predicates),
+                delete_predicates,
             }),
             representation: ChunkStageFrozenRepr::ReadBuffer(Arc::new(chunk)),
         };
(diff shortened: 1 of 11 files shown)


server/src/db/catalog/chunk.rs
```diff
@@ -469,45 +469,35 @@ impl CatalogChunk {
...
-    pub fn add_delete_predicate(&mut self, delete_predicate: &Predicate) -> Result<()> {
+    pub fn add_delete_predicate(&mut self, delete_predicate: Arc<Predicate>) -> Result<()> {
         match &mut self.stage {
             ChunkStage::Open { mb_chunk: _ } => {
                 // Freeze/close this chunk and add delete_predicate to its frozen one
-                self.freeze_with_predicate(delete_predicate)?;
+                self.freeze_with_predicate(Some(delete_predicate))?;
             }
-            ChunkStage::Frozen { meta, .. } => {
+            ChunkStage::Frozen { meta, .. } | ChunkStage::Persisted { meta, .. } => {
```
NGA-TRAN, line 478: Great way to remove duplicate code

server/src/db/catalog/chunk.rs
```diff
@@ -692,7 +685,9 @@ impl CatalogChunk {
                 let metadata = ChunkMetadata {
                     table_summary: Arc::new(mb_chunk.table_summary()),
                     schema: s.full_schema(),
-                    delete_predicates: Arc::new(vec![delete_predicate.clone()]),
+                    delete_predicates: delete_predicate
+                        .map(|delete_predicate| vec![delete_predicate])
+                        .unwrap_or_default(),
```
NGA-TRAN, line 690: This `map` and `unwrap` are good things for me to learn from this PR

server/src/db/catalog/chunk.rs
```diff
@@ -714,7 +709,7 @@ impl CatalogChunk {
     }
 
     pub fn freeze(&mut self) -> Result<()> {
-        self.freeze_with_predicate(&PredicateBuilder::default().build())
+        self.freeze_with_predicate(None)
```
NGA-TRAN, line 712: Yeah, the previous line was a bug and you fix it with this change. This will conflict with today merge that also fix this in different way but I think you will resolve it easily

server/src/db/lifecycle/compact.rs
```diff
@@ -66,8 +66,7 @@ pub(crate) fn compact_chunks(
                 .map(|prev_last| prev_last.max(candidate_last))
                 .or(Some(candidate_last));
 
-            let mut preds = (*chunk.delete_predicates()).clone();
-            delete_predicates.append(&mut preds);
+            delete_predicates.extend(chunk.delete_predicates().iter().cloned());
```
NGA-TRAN, line 69: Nice. Another new thing, extend, for me to learn

NGA-TRAN (approved):
Very nice refactor. A few things for me to learn from this PR including the description and the code

PR #2562, opened 2021-09-16
feat: migrate operations CLI to use pbjson

tustvold:
A significant amount of complexity and plumbing existed to workaround the lack of serialization support for the returned protobufs. With #2468 this is no longer necessary and can be expunged :tada:.

This is a bit fiddly because the google operations API makes use of `google::protobuf::Any` which doesn't really have a defined JSON serialization that is user-readable. I therefore introduced an `IoxOperation` type that is just the `Operation` paired with the decoded metadata. There is some redundancy here, but this is what tools like grpcurl do, so I'm inclined to consider it good enough.

With #2560 which fixes the bytes encoding the output looks like

```
[
{
    "operation": {
      "name": "54076",
      "metadata": {
        "typeUrl": "type.googleapis.com/influxdata.iox.management.v1.OperationMetadata",
        "value": "CN7thqkBEMKSo6MDGAFSYQohODQ0OTEwZWNlODBiZThiY18zYzBiZDRjODkxODZjYTg5EhMyMDIxLTA5LTE2IDEzOjAwOjAwGgTPAtACIiFodHRwX2FwaV9yZXF1ZXN0X2R1cmF0aW9uX3NlY29uZHNoAQ=="
      },
      "done": true,
      "response": {
        "typeUrl": "type.googleapis.com/google.protobuf.Empty"
      }
    },
    "metadata": {
      "cpuNanos": "354531038",
      "wallNanos": "879282498",
      "totalCount": "1",
      "successCount": "1",
      "compactChunks": {
        "partitionKey": "2021-09-16 13:00:00",
        "tableName": "http_api_request_duration_seconds",
        "chunks": [
          335,
          336
        ]
      }
    }
  }
]⏎    
```


crepererum:
Is `operation.metadata.value` in your example above a protobuf message embedded as bytes within a protobuf message?

inception

tustvold:
> Is operation.metadata.value in your example above a protobuf message embedded as bytes within a protobuf message?

Yes, that is what google::protobuf::Any is

tests/end_to_end_cases/management_cli.rs
```diff
@@ -739,18 +738,16 @@ async fn test_close_partition_chunk() {
...
 
-    let expected_job = Job::CompactChunk {
-        chunk: ChunkAddr {
-            db_name: Arc::from(db_name.as_str()),
-            table_name: Arc::from("cpu"),
-            partition_key: Arc::from("cpu"),
-            chunk_id: 0,
-        },
-    };
+    let expected_job = Job::CloseChunk(CloseChunk {
```
alamb, line 741: 🤔  not sure why this was previously expecting `CompactChunk`, but getting back a `CloseChunk` in response to a `close-chunk` request seems very reasonable to me 👍

tests/end_to_end_cases/operations_api.rs
```diff
@@ -28,20 +28,15 @@ async fn test_operations() {
         .expect("list operations failed");
 
     assert_eq!(running_ops.len(), 1);
-    assert_eq!(running_ops[0].name(), operation.name);
+    assert_eq!(running_ops[0].operation.name, operation.operation.name);
```
alamb, line 31: so many operations! I do wonder if having an `operation` field on an `operation` is overly meta (or as @crepererum would say "you're waiting for a train"). I wonder could it be inlined?

tests/end_to_end_cases/operations_cli.rs
```diff
@@ -45,33 +46,33 @@ async fn test_start_stop() {
     .expect("expected JSON output");
 
     assert_eq!(operations.len(), 1);
-    match &operations[0].job {
-        Some(Job::Dummy { nanos, .. }) => {
+    match &operations[0].metadata.job {
+        Some(Job::Dummy(Dummy { nanos, .. })) => {
```
alamb, line 50: I wonder if we could get away with less `Dummy` repetition? Like can we possibly keep it as 
```rust
        Some(Job::Dummy { nanos, .. }) => {
```
?

generated_types/src/google.rs
```diff
@@ -21,6 +26,44 @@ pub mod longrunning {
...
+        }
+    }
+
+    /// Groups together an `Operation` with a decoded `OperationMetadata`
+    ///
+    /// When serialized this will serialize the encoded Any field on `Operation` along
+    /// with its decoded representation as `OperationMetadata`
+    #[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+    pub struct IoxOperation {
+        pub operation: Operation,
```
alamb, line 54: what do you think about calling this field something like `inner` rather than `operation` so as to avoid `operation.operation` type calls?

alamb (approved):
I took it for a spin: 

```shell
](arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox operation list | python -m json.tool
[
    {
        "operation": {
            "name": "0",
            "metadata": {
                "typeUrl": "type.googleapis.com/influxdata.iox.management.v1.OperationMetadata",
                "value": [
                    42,
                    0
                ]
            },
            "done": true,
            "response": {
                "typeUrl": "type.googleapis.com/google.protobuf.Empty"
            }
        },
        "metadata": {
            "dummy": {}
        }
    }
]
```

Looks very cool

tests/end_to_end_cases/operations_cli.rs
tustvold: Unfortunately this is what prost generates, the outer enumeration is the oneof variant, the inner is the name of the message

tests/end_to_end_cases/operations_api.rs
tustvold: I could rename it to iox_operation to make it clearer that they're different types?

generated_types/src/google.rs
tustvold: I _think_ that would be more confusing as Operation technically contains OperationMetadata. So if anything it should be outer...

alamb: It is `inner` to `IOxOperation` though....

tustvold: I think I'm going to leave it as is and we can revisit if it causes confusion/people feel strongly

Issue #2518, opened 2021-09-13
Persist delete predicates with chunk


NGA-TRAN:
# Rationale
As a apart of #2396, we need to persist the `delete_predicates` stored in the in-memory chunks when the chunk is persisted

Issue #2397, opened 2021-08-24
Support Hard Delete


NGA-TRAN:
# Rationale
This is a part of https://github.com/influxdata/influxdb_iox/issues/2395. Refer to [this delete design document ](https://docs.google.com/document/d/1AKJYEvarjRxpnRx4q5XwcrhVpQtlVcLO0ltPjg_5aoU/edit#) for use cases, design ,  and related features.

# Implementation Check List

NGA-TRAN:
This one will be done as a part of #3089 and its future lifecycle ones

NGA-TRAN:
Since the Compact OS chunks, #3089, that this hard delete relies on  is done except integration tests, I declare this is done

PR #2564, opened 2021-09-16
feat: delete catalog pres. catalog <=> in-mem catalog API

crepererum:
First step towards #2518. Creates the Rust API to communicate delete
predicates between the preserved catalog and the in-memory catalog and
adds tests ensuring that the in-mem catalog produces the wanted errors
as well as correct checkpoints (similar to how this is done for the
parquet file tracking already).

**This does NOT contain the actual preservation!**


parquet_file/src/catalog/core.rs
```diff
@@ -2159,13 +2157,4 @@ mod tests {
                 .is_some()
         );
     }
-
-    #[tokio::test]
-    async fn test_catalog_state() {
```
crepererum, line 2164: This test was moved into `test_helpers` because it all the code it tests lives there, or in other words: it tests that the mocked in-memory catalog is sane.

parquet_file/src/catalog/interface.rs
```diff
@@ -82,6 +119,15 @@ pub trait CatalogState {
...
     /// Remove parquet file from state.
     fn remove(&mut self, path: &ParquetFilePath) -> Result<(), CatalogStateRemoveError>;
+
+    /// Register new predicate to delete data.
+    ///
+    /// The delete predicate will only be applied to the given chunks (by table name, partition key, and chunk ID).
+    fn delete_predicate(
+        &mut self,
+        predicate: Arc<Predicate>,
+        chunks: Vec<ChunkAddrWithoutDatabase>,
```
crepererum, line 129: This will be the storage format: the protobuf created in #2537 + a list of affected chunks. This way:

1. we can blindly re-use the predicate -> chunk distribution/pruning logic from `Db` (currently it just assigns all chunks, but this will be improved)
2. we don't need  to be careful about not applying predicates to new chunks
3. we don't need to re-compute predicate -> chunk assignment (point 1) while loading the catalog (we however trade in some storage for that)

parquet_file/src/catalog/cleanup.rs
```diff
@@ -137,6 +141,15 @@ impl CatalogState for TracerCatalogState {
         // Do NOT remove the file since we still need it for time travel
         Ok(())
     }
+
+    fn delete_predicate(
+        &mut self,
+        _predicate: Arc<Predicate>,
+        _chunks: Vec<ChunkAddrWithoutDatabase>,
+    ) -> Result<(), CatalogStateDeletePredicateError> {
+        // No need to track delete predicates.
```
alamb, line 150: it might help explain the rationale in the comments (the code is pretty clear that the predicate is not tracked, but the *why* it is not tracked is less clear to me)

parquet_file/src/catalog/interface.rs
alamb: This storage format makes sense to me. What was the alternative? Storing the predicate directly on the chunk?

Why not use `ChunkAddr` (why introduce a `ChunkAddrWithoutDatabase`)? Is it to handle potential future db renames?

parquet_file/src/catalog/test_helpers.rs
```diff
@@ -60,10 +69,38 @@ impl TestCatalogState {
...
+                        predicates
+                            .entry(addr)
+                            .and_modify(|(_predicate, v)| v.push(pred_chunk()))
+                            .or_insert_with(|| (Arc::clone(predicate), vec![pred_chunk()]));
+                    }
+                }
+            }
+        }
+
+        predicates.into_values().collect()
```
alamb, line 101: I wonder if sorting the return result would be helpful to avoid some sort of intermittent discrepancies as the order of the output seems like it may depend on the order of iteration over hashmap keys

alamb (commented):
Looks good to me -- I am going to defer to @NGA-TRAN  to actually approve as I am not as familiar with this code or the intended design as she is. 👍

parquet_file/src/catalog/test_helpers.rs
```diff
@@ -60,10 +69,38 @@ impl TestCatalogState {
...
+        let mut predicates: HashMap<usize, (Arc<Predicate>, Vec<ChunkAddrWithoutDatabase>)> =
+            Default::default();
+
+        for (table_name, table) in &self.tables {
+            for (partition_key, partition) in &table.partitions {
+                for (chunk_id, chunk) in &partition.chunks {
+                    for predicate in &chunk.delete_predicates {
+                        let predicate_ref: &Predicate = predicate.as_ref();
+                        let addr = (predicate_ref as *const Predicate) as usize;
+                        let pred_chunk = || ChunkAddrWithoutDatabase {
```
NGA-TRAN, line 87: Rust question: What "= ||" means?

parquet_file/src/catalog/cleanup.rs
NGA-TRAN: I went all the way down and if In understand correctly, this interface is for adding  delete predicates to corresponding chunks. Maybe a comment to tell what this function will help

parquet_file/src/catalog/interface.rs
```diff
@@ -96,4 +142,10 @@ pub struct CheckpointData {
     ///
     /// If a file was once added but later removed it MUST NOT appear in the result.
     pub files: HashMap<ParquetFilePath, CatalogParquetInfo>,
+
+    /// List of active delete predicates together with their chunks (by table name, partition key, and chunk ID).
+    ///
+    /// This must only contains chunks that are still present in the catalog. Predicates that do not have any chunks
+    /// attached should be left out.
+    pub delete_predicates: Vec<(Arc<Predicate>, Vec<ChunkAddrWithoutDatabase>)>,
```
NGA-TRAN, line 150: It makes sense that you store this structure in the catalog checkpoint. My question is beside this, do you save delete predicate with the chunk data too or not?

NGA-TRAN (approved):
I do not know the catalog code well but the changes to add delele_predicates make sense

parquet_file/src/catalog/interface.rs
crepererum: > Why not use `ChunkAddr` (why introduce a `ChunkAddrWithoutDatabase`)? Is it to handle potential future db renames?

The preserved catalog doesn't care about DB name (because it's per DB anyway) and for the reason of future DB renames I also have no plan changing that.

parquet_file/src/catalog/test_helpers.rs
crepererum: You might have an helpful idea, because I don't: The issue is that `Predicate` doesn't implement an ordering (also because the datafusion `Expr`s it contains don't). The pointer address we use for deduplication here is also somewhat random. So what shall we use for ordering? The chunks (the second part of the tuples) are not sufficient to establish a total ordering because two predicates can have the exact same chunk set.

parquet_file/src/catalog/test_helpers.rs
crepererum: Uh, I see how this syntax is really confusing: It creates a closure (like a "local function"). So this:

```rust
some_option.unwrap_or_else(|| Foo::new(...))
```

and this:

```rust
let f = || Foo::new(...);
some_option.unwrap_or_else(f)
```

and this:

```rust
let f = || {
    Foo::new(...)
};
some_option.unwrap_or_else(f)
```

are identical. In the latter case however you could reuse `f` (that's what I do in the code snipped above).

parquet_file/src/catalog/interface.rs
crepererum: > do you save delete predicate with the chunk data too or not?

It's extra (aka not with the chunk data), because:

1. **new delete predicate is added to unpersisted chunk:** We don't need to store that, because the data that is deleted here is also not preserved yet.
2. **new delete predicate is added to persisted chunk:** We need to store that but we cannot change the chunk metadata any longer because it is already embedded in the parquet file and we don't wanna rewrite existng files. So we just say "for this chunk that we already preserved, here is a new delete predicate".
3. **chunk is persisted:** Here we could just store the chunk and store the delete predicate (still side-by-side, aka predicates are extra) within the same partition. However this is only required if a new delete predicate arrives while we're running the compaction and the split, because otherwise we can materialize the predicates already and don't need to store them at all.

parquet_file/src/catalog/cleanup.rs
crepererum: I've added a comment that might help. Note that this implementation here is only for the parquet file cleanup.

parquet_file/src/catalog/interface.rs
crepererum: > What was the alternative? Storing the predicate directly on the chunk?

Storing within the chunk isn't really an alternative, since we must be able to attach delete predicates to chunks that are already persisted and we don't wanna rewrite existing files. We could store the predicate next to the parquet file path, but next to it is also the serialized parquet metadata and if we need to repeat all of that, storing delete predicates gets kinda costly (in the worst case it's basically identical to triggering a catalog checkpoint).

Another option would be to store the predicate per DB (like it is initially received via API or write buffer) and let the `Db` code re-assign them to the right chunks during catalog loading. Then however we also need to store the maximum chunk ID that is affected by the predicate (to not delete data that arrived after the predicate) and together with locking / timing and the lifecycle policy that creates more chunks while we're initially assigning the predicate to chunks, that asks for some subtle desync bugs that I don't wanna risk. Aka: using the predicate -> chunk assignment that is present in the in-mem catalog and store it is easier to keep in-sync and we can blame a single piece of code when this assignment is wrong.

alamb: I agree storing the chunk ids affected by the predicate sounds better than either of the alternate options you describe 👍  Thank you for explaining

parquet_file/src/catalog/test_helpers.rs
alamb: maybe calling the variable `pred_chunk_closure` could hint to readers what is going on

parquet_file/src/catalog/test_helpers.rs
alamb: One hacky thought would be to order by the chunks first and on ties compare the Debug representation of the Exprs.

We could also define `PartialOrd` for Exprs upstream in DataFusion. I'll give that a try. 

I don't think this is critical at the moment, now should it block merging this PR

alamb: https://github.com/apache/arrow-datafusion/pull/1015 <-- upstream datafusion PR

parquet_file/src/catalog/interface.rs
NGA-TRAN: Make senses. These make me think there is overlap between the soft delete #2396 and hard delete #2397 where we will delete data during compaction. Anyway, we have to do step by step to avoid overwhelming ourselves :)

parquet_file/src/catalog/test_helpers.rs
NGA-TRAN: Nice @alamb

parquet_file/src/catalog/test_helpers.rs
NGA-TRAN: Thanks for the explanation @crepererum

crepererum: done (renamed)

parquet_file/src/catalog/test_helpers.rs
crepererum: I'll fix that in a follow-up PR (since it also requires a datafusion upgrade) so I can continue to implement this feature and it's not super critical for

crepererum: Actually, the tests are flaky w/o sorting, so I'll to address this before merging this PR.

alamb: DF changes included in https://github.com/influxdata/influxdb_iox/pull/2592

PR #2565, opened 2021-09-16
test: tests for delete

NGA-TRAN:
Move tests as well as prepare framework for delete tests


query_tests/Cargo.toml
```diff
@@ -10,19 +10,19 @@ description = "Tests of the query engine against different database configuratio
 [dependencies]
 async-trait = "0.1"
 data_types = { path = "../data_types" }
+datafusion = { path = "../datafusion" }
 once_cell = { version = "1.4.0", features = ["parking_lot"] }
+predicate = { path = "../predicate" }
```
NGA-TRAN, line 15: These 2 are needed in scenarios.rs and its module is public. I do not know any better solution

server/src/db.rs
```diff
@@ -3741,262 +3741,4 @@ mod tests {
             .await
             .unwrap();
     }
-
```
NGA-TRAN, line 3744: The below deleted 2 tests are a part of the above added tests

query_tests/src/sql.rs
```diff
@@ -808,3 +808,68 @@ async fn sql_select_all_different_tags_chunks() {
...
+        "| 2   | 1970-01-01T00:00:00.000000020Z |",
+        "+-----+--------------------------------+",
+    ];
+
+    // Data deleted when it is in MUB, and then moved to RUB and OS
+    run_sql_test_case!(
+        DeleteFromMubOneMeasurementOneChunk {},
+        "SELECT * from cpu",
+        &expected
+    );
```
NGA-TRAN, line 827: This is equivalent with the removing `test_delete_on_different_chunk_types`

query_tests/src/sql.rs
```diff
@@ -808,3 +808,68 @@ async fn sql_select_all_different_tags_chunks() {
...
+
+    // Data deleted when it is in RUB, and then moved OS
```
NGA-TRAN, line 829: This is new

query_tests/src/sql.rs
```diff
@@ -808,3 +808,68 @@ async fn sql_select_all_different_tags_chunks() {
...
+    run_sql_test_case!(
+        DeleteFromRubOneMeasurementOneChunk {},
+        "SELECT * from cpu",
+        &expected
+    );
+
+    // Data deleted when it is in OS
```
NGA-TRAN, line 836: this is new

query_tests/src/sql.rs
```diff
@@ -808,3 +808,68 @@ async fn sql_select_all_different_tags_chunks() {
...
+    run_sql_test_case!(
+        DeleteFromOsOneMeasurementOneChunk {},
+        "SELECT * from cpu",
+        &expected
+    );
+}
+
+#[tokio::test]
+async fn sql_select_with_deleted_data_from_multi_exprs() {
```
NGA-TRAN, line 845: similarly to above function, the first case is equivalent to the removing `test_delete_on_different_chunk_types_multi_exprs` and the last 2 are new.

query_tests/Cargo.toml
alamb: I don't think these are actually new "real" dependencies for `query_test` - they are already used transitively by `query` so I don't think this appreciable change in overall dependencies

query_tests/src/scenarios.rs
```diff
@@ -1,12 +1,14 @@
 //! This module contains testing scenarios for Db
-
+#[allow(unused_imports, dead_code, unused_macros)]
```
alamb, line 2: why is this needed?

query_tests/src/scenarios.rs
```diff
@@ -1198,3 +1200,648 @@ impl DbSetup for ChunkOrder {
...
+
+async fn make_delete_mub(lp_lines: Vec<&str>, pred: &Predicate) -> DbScenario {
+    let db = make_db().await.db;
+    // create an open MUB
+    write_lp(&db, &lp_lines.join("\n")).await;
+    // One open MUB, no RUB, no OS
+    assert_eq!(count_mutable_buffer_chunks(&db), 1);
+    assert_eq!(count_read_buffer_chunks(&db), 0);
+    assert_eq!(count_object_store_chunks(&db), 0);
+    // delete one row
```
alamb, line 1510:
```suggestion
```

I don't think that comment is valid anymore (it may be deleting more than one row) -- likewise below there are several "delete one row" type comments which I think are stale

query_tests/src/scenarios.rs
alamb: I was able to remove them without issue

```diff
diff --git a/query_tests/src/scenarios.rs b/query_tests/src/scenarios.rs
index 5e6ac585..e8ce9cff 100644
--- a/query_tests/src/scenarios.rs
+++ b/query_tests/src/scenarios.rs
@@ -1,5 +1,4 @@
 //! This module contains testing scenarios for Db
-#[allow(unused_imports, dead_code, unused_macros)]
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
diff --git a/server/src/db.rs b/server/src/db.rs
index 99f5dc13..e1ff050e 100644
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -1301,9 +1301,7 @@ mod tests {
     use arrow::record_batch::RecordBatch;
     use bytes::Bytes;
     use chrono::{DateTime, TimeZone};
-    use datafusion::logical_plan::{col, lit};
     use futures::{stream, StreamExt, TryStreamExt};
-    use predicate::predicate::PredicateBuilder;
     use tokio_util::sync::CancellationToken;
 
     use ::test_helpers::assert_contains;
```

alamb (approved):
This is pretty epic @NGA-TRAN ! I think it would be good to remove the `#[allow(unused_imports, dead_code, unused_macros)]` prior to merging as I find those warnings have been quite helpful detecting bugs in other places

query_tests/src/scenarios.rs
NGA-TRAN: Nice. I will remove them

query_tests/src/scenarios.rs
NGA-TRAN: Good catch. Copy and paste from old test :) I will remove them.

PR #2566, opened 2021-09-16
feat: Detailed database list command

### LINKED ISSUE #2462: Add `--detailed` flag to `database list` CLI
From @pauldix on https://github.com/influxdata/influxdb_iox/pull/2457#pullrequestreview-746194929:

Add an option to `influxdb_iox database list --detailed` that:

- includes existing and deleted all in one go 
- includes the generation information.

carols10cents:
Closes #2462.

This adds a `--detailed` flag to `database list` that will list both active and deleted databases with their generation. Deleted databases will have the datetime at which they were deleted; active databases have nothing in that column.

Here's some sample output of `influxdb_iox database list --detailed` for an active database foo, deleted database aaa, database moo that has been created and deleted twice then its generation 0 restored, and database zoom that has been created, deleted, created:

```
Deleted at                       | Generation ID | Name
---------------------------------+---------------+--------
2021-09-17 15:11:14.667568 UTC    0               aaa
                                  0               foo
                                  0               moo
2021-09-17 15:11:14.667861 UTC    1               moo
2021-09-17 15:11:14.667982 UTC    0               zoom
                                  1               zoom
```

Is this what you were imagining, @pauldix ?


src/commands/database.rs
```diff
@@ -244,20 +249,24 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
...
                         .deleted_at
                         .and_then(|t| {
                             let dt: Result<DateTime<Utc>, _> = t.try_into();
                             dt.ok().map(|d| d.to_string())
                         })
-                        .unwrap_or_else(|| String::from("Unknown"));
+                        .unwrap_or_else(String::new);
                     println!(
-                        "{:<33}{:<16}{}",
+                        "{:<34}{:<16}{}",
```
alamb, line 269: Another way to format such tabubar output would be with the `prettytable` crate as is done when printing remotes:

https://github.com/influxdata/influxdb_iox/blob/cba03e9c478d2224a849e19fa799f777c62229a7/src/commands/server_remote.rs#L53-L66

alamb (approved):
Looks great @carols10cents . I also tried it out locally and it worked well 👍

src/commands/database.rs
carols10cents: OMG it looks so much nicer, both in code and output 😻 

```
+--------------------------------+---------------+------+
| Deleted at                     | Generation ID | Name |
+--------------------------------+---------------+------+
| 2021-09-17 19:41:58.897 UTC    | 0             | aaa  |
|                                | 0             | foo  |
|                                | 0             | moo  |
| 2021-09-17 19:41:58.897280 UTC | 1             | moo  |
| 2021-09-17 19:41:58.897396 UTC | 0             | zoom |
|                                | 1             | zoom |
+--------------------------------+---------------+------+
```

PR #2570, opened 2021-09-17
fix: Await on freeze handles instead of try_freeze/returning Transition errors

carols10cents:
As suggested by @alamb [on 2464](https://github.com/influxdata/influxdb_iox/pull/2464#discussion_r711128973) to match the change made in https://github.com/influxdata/influxdb_iox/pull/2433.

I'm pretty confident about making this change for `delete` and `restore`. I'm slightly less confident about making this change for `skip_replay`.

I also tried making this change for `wipe_preserved_catalog` but in the process discovered that function should probably stay the way it is: it freezes and checks for the error state synchronously first, then returns a future that gets executed as a tracked future at some other time, moving the frozen handle into the future. Changing that whole function to be async to enable awaiting on the freeze handle would change the semantics and move some of the errors to happen after the future is `tokio::spawn`ed. This behavior is encoded [in this test](https://github.com/influxdata/influxdb_iox/blob/cbf166c035c5c28f7d359896ed07e806d3a2a48e/server/src/lib.rs#L2164-L2172) that fails if I make this change for `wipe_preserved_catalog`.

This isn't urgent, more of a consistency thing and small improvement in relatively rare cases, so this is fine to leave open until @tustvold returns and can review :)


alamb:
This code also may improve the user experience -- instead of (in very rare cases) operations like "delete" and "restore" returning an error (which might go away if the operation is retried by the client), now the request will wait until the previous operation is complete.

tustvold (approved):
Looks good to me, thank you for this :+1:

I think perhaps the TransitionInProgress error variant could also go now, but not checked.

alamb:
I checked and it turns out there are a few more instances of `TransitionInProgress` so it isn't dead code yet

```
rg -n -H --no-heading -e 'TransitionInProgress' $(git rev-parse --show-toplevel || pwd)
/Users/alamb/Software/influxdb_iox/src/influxdb_ioxd/rpc/error.rs:113:        Error::TransitionInProgress { .. } => {
/Users/alamb/Software/influxdb_iox/server/src/database.rs:42:    TransitionInProgress {
/Users/alamb/Software/influxdb_iox/server/src/database.rs:241:                TransitionInProgress {
/Users/alamb/Software/influxdb_iox/server/src/database.rs:467:            let handle = state.try_freeze().context(TransitionInProgress {
```

PR #2571, opened 2021-09-17
feat: Add tag_set and tag_pairs to measurements in Data Generator

pauldix:
This adds the ability to specify a tag_set and a collection of tag_pairs to measurements in the data generator. Tag pairs are evaluated once when the generator is created. This avoids re-running handlebars evaluations while generating data for tags that don't change value.

This commit also fixes an issue when printing the generation output to stdout while generating from more than one agent. Previously it would be garbled together.

Follow on PRs will update the tag generation code in measurement specs to be more consistent and optimized for performance. I'll be removing the restriction of using different options while using tag_set and tag_pairs. I wanted to get this in first to show the structure of what is output.

Running this from the project root will show the example output for the included schema that uses this new feature:
```
 cargo run -p iox_data_generator -- --spec ./iox_data_generator/schemas/storage_cardinality_example.toml --print
```

And a snippet of that output looks like this:
```
storage_usage_bucket_cardinality,bucket_id=1_2Q2iaNH0Y6qpCHo,env=whatever-environment-1,host=storage-0,hostname=0,node_id=0,org_id=1_00nRAqX0VxfOBEl,partition_id=1,role=storage,url=http://127.0.0.1:6060/metrics/usage gauge=721566i 1631909529018737000
storage_usage_bucket_cardinality,bucket_id=1_2Q2iaNH0Y6qpCHo,env=whatever-environment-1,host=storage-0,hostname=0,node_id=0,org_id=1_00nRAqX0VxfOBEl,partition_id=2,role=storage,url=http://127.0.0.1:6060/metrics/usage gauge=1415953i 1631909529018737000
storage_usage_bucket_cardinality,bucket_id=1_2Q2iaNH0Y6qpCHo,env=whatever-environment-1,host=storage-0,hostname=0,node_id=0,org_id=1_00nRAqX0VxfOBEl,partition_id=3,role=storage,url=http://127.0.0.1:6060/metrics/usage gauge=5762156i 1631909529018737000
```

### DIFF
--- a/iox_data_generator/src/measurement.rs
+++ b/iox_data_generator/src/measurement.rs
@@ -8,10 +8,15 @@ use crate::{
     DataGenRng, RandomNumberGenerator,
 };
 
+use crate::substitution::{FormatNowHelper, RandomHelper};
+use crate::tag_set::{GeneratedTagSets, TagPair, TagSet};
+use handlebars::Handlebars;
 use influxdb2_client::models::DataPoint;
 use itertools::Itertools;
-use snafu::{ResultExt, Snafu};
+use serde_json::json;
+use snafu::{OptionExt, ResultExt, Snafu};
 use std::fmt;
+use std::sync::{Arc, Mutex};
 
 /// Measurement-specific Results
 pub type Result<T, E = Error> = std::result::Result<T, E>;
@@ -78,6 +83,41 @@ pub enum Error {
         /// Underlying `tag` module error that caused this problem
         source: crate::tag::Error,
     },
+
+    /// Error generated if tag_set and tag_pair are used with any other options
+    #[snafu(display("If using either tag_set or tag_pair in a measurement spec, other options are not currently supported"))]
+    TagSetTagPairOnly,
+
+    /// Error generated if specifying a tag_set that wasn't generated
+    #[snafu(display(
+        "Tag set {} referenced not found for measurement {}",
+        tag_set,
+        measurement
+    ))]
+    GeneratedTagSetNotFound {
+        /// The name of the tag set
+        tag_set: String,
+        /// The name of the measurement it is being associated with
+        measurement: String,
+    },
+
+    /// Error that may happen when compiling a template from the values specification
+    #[snafu(display("Could not compile template `{}`, caused by:\n{}", template, source))]
+    CantCompileTemplate {
+        /// Underlying Handlebars error that caused this problem
+        source: handlebars::TemplateError,
+        /// Template that caused this problem
+        template: String,
+    },
+
+    /// Error that may happen when rendering a template with passed in data
+    #[snafu(display("Could not render template `{}`, caused by:\n{}", template, source))]
+    CantRenderTemplate {
+        /// Underlying Handlebars error that caused this problem
+        source: handlebars::RenderError,
+        /// Template that caused this problem
+        template: String,
+    },
 }
 
 /// A set of `count` measurements that have the same configuration but different
@@ -98,6 +138,7 @@ impl<T: DataGenRng> MeasurementGeneratorSet<T> {
         parent_seed: impl fmt::Display,
         static_tags: &[Tag],
         execution_start_time: i64,
+        generated_tag_sets: &GeneratedTagSets,
     ) -> Result<Self> {
         let count = spec.count.unwrap_or(1);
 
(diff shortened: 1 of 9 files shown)


alamb (commented):
I think @shepmaster  or @carols10cents  are the best to review this change as they are most familiar with the code. 

BTW it is a follow on to https://github.com/influxdata/influxdb_iox/pull/2499 which I reviewed, so I am also happy to review if they don't have time.

iox_data_generator/src/specification.rs
```diff
@@ -28,11 +28,14 @@ type Result<T, E = Error> = std::result::Result<T, E>;
...
-    /// Every point generated from this configuration will contain a tag
-    /// `data_spec=[this value]` to identify what generated that data. This
-    /// name can also be used in string replacements by using the
-    /// placeholder `{{data_spec}}`.
+    /// If `include_spec_tag` is set to true, eveyr point generated from this
+    /// configuration will contain a tag `data_spec=[this value]` to identify
+    /// what generated that data. This name can also be used in string replacements
+    /// by using the placeholder `{{data_spec}}`.
     pub name: String,
+    /// If set to true, all samples generated will include a `data_spec=[name]`
```
carols10cents, line 36: What if `include_spec_tag` is set to `Some(false)` or `None`? I actually don't see any use of `include_spec_tag` anywhere, is it something that's going to be used in the future or is it something that ended up not being needed?

This caught my eye because `Option<bool>` is fairly rare... it does happen when the semantics of "unset" are different from the semantics of "explicitly set to true" or "explicitly set to false", but it looks like this could be `bool` that defaults to `false` if not specified by using `#[serde(default)]`?

iox_data_generator/src/specification.rs
```diff
@@ -28,11 +28,14 @@ type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct DataSpec {
-    /// Every point generated from this configuration will contain a tag
-    /// `data_spec=[this value]` to identify what generated that data. This
-    /// name can also be used in string replacements by using the
-    /// placeholder `{{data_spec}}`.
+    /// If `include_spec_tag` is set to true, eveyr point generated from this
```
carols10cents, line 31:
```suggestion
    /// If `include_spec_tag` is set to true, every point generated from this
```

iox_data_generator/src/measurement.rs
```diff
@@ -286,7 +404,49 @@ impl<T: DataGenRng> MeasurementGenerator<T> {
...
+            .map(|ts| {
+                let mut point = DataPoint::builder(&self.name);
+
+                for tag in &ts.tags {
+                    point = point.tag(tag.key.to_string(), tag.value.to_string());
+                }
+
+                for tag in &pregen.tag_pairs {
+                    point = point.tag(tag.key.to_string(), tag.value.to_string());
+                }
```
carols10cents, line 425: Yeeeaahhh I would bet that all this string allocation is an impact on the performance... and I see `DataPoint` stores owned `String`s. I wonder if that can be changed to be `impl AsRef<str>`, which might also need a `+ 'static`, so that `DataPoint` could hold either `String` or `Arc<str>`?

iox_data_generator/src/tag_set.rs
```diff
@@ -461,8 +460,8 @@ impl std::fmt::Display for TagSet {
 
 #[derive(Debug, PartialEq, PartialOrd)]
 pub struct TagPair {
-    key: Arc<String>,
-    value: Arc<String>,
+    pub key: Arc<String>,
+    pub value: Arc<String>,
```
carols10cents, line 464: Because `String`s are a pointer to bytes on the heap, and `Arc`s are a pointer to the heap, this is adding an extra layer of indirection. This can be:

```suggestion
    pub key: Arc<str>,
    pub value: Arc<str>,
```

and where you construct `TagPair`s, if you have a `String` you can do `string.as_str().into()` (deref coercion isn't being as helpful as I wish it would be here) and if you have a `&str` you can do `string_slice.into()` to get `Arc<str>`s.

I don't know if or how much this is impacting the performance, but making the change certainly can't *hurt*.

carols10cents (changes requested):
Some fixes, some guesses about what might help with the performance!

carols10cents:
I think this is superseded now by #2989, so I'm going to close it!

PR #2573, opened 2021-09-18
feat: disable Snafu futures feature

tustvold:
This eliminates an old version of pin-project from the dependencies, and inconsistent feature flags across the workspace which can lead to recompiling code when switching build targets. Subjectively I also think it is clearer, but expect opinions on that may differ.


alamb (approved):
Removing old dependencies is worth a few slightly more verbose callsites in my opinon

PR #2574, opened 2021-09-18
feat: remove parse_duration

tustvold:
[parse_duration](https://github.com/zeta12ti/parse_duration) is no longer maintained, has a known [security vulnerability](https://github.com/zeta12ti/parse_duration/issues/21) - [RUSTSEC-2021-0041](https://rustsec.org/advisories/RUSTSEC-2021-0041.html), and also has some outdated dependencies. We already have a time parsing dependency, used by env_logger, etc... called humantime so we might as well just use that. 

This does mean we lose the ability to write things like "1 day -1 hour" but I imagine this isn't a deal breaker for us


alamb (approved):
looks like a nice change to me

Issue #2392, opened 2021-08-24
Batch Span Export


tustvold:
Currently `trace_export::OtelExporter` performs no batching of spans. For current IOx deployments where the spans are sent as UDP packets on localhost, this is almost certainly perfectly fine, but this is potentially not always going to be the case. The OTELExporter should instead probably batch based on some linger period, so that spans produced at similar times get merged into the same packet. This could potentially use the stream batching I implemented [here](https://github.com/tustvold/rust-playground/blob/master/lib/stream/src/batch.rs#L34)

PR #2576, opened 2021-09-18
feat: don't use opentelemetry async trace exporter

tustvold:
The `trace_exporters::OtelExporter` already has a background worker that does the publication, we don't need a further asynchronous queue. This appears to have been unintentionally changed as part of #2305 https://github.com/influxdata/influxdb_iox/pull/2440/commits/7ff28405ea6e6adf2d1e7ab380458f6c75aba048

It is perhaps worth noting that OtelExporter does not currently batch spans for export, the is almost certainly irrelevant for sending traces over UDP to a jaeger-agent, but might matter if we add a more expensive trace export. https://github.com/influxdata/influxdb_iox/issues/2392 tracks adding this. It is fairly trivial just not currently implemented.


alamb (approved):
I trust your judgement on this one -- as long as the traces are still exported it would be good to me

alamb (approved):

tustvold:
Tested working locally

PR #2577, opened 2021-09-18
feat: remove human format

tustvold:
The human_format crate appeared to only be using to provide thousands grouping for size types, this PR therefore removes it in favor of an incredibly simple method to achieve the same thing.

The padding in the formatted output was fairly inconsistent, it was present in some places and not others. I opted to remove it all for consistency, but happy to restore it if preferred.

Longer-term I would very much like to remove the packers crate, as it duplicates functionality found elsewhere, but at least we can eliminate this for now.


alamb:
Removing the packers crate is a good idea (a natural time will be when we figure out how to do bulk tsm -> parquet ingest)

alamb (approved):

PR #2586, opened 2021-09-20
feat: remove routerify

tustvold:
The HTTP API is now simple enough that we can just use a plain match block. This eliminates a fair bit of complexity, a dependency, and a custom fork.


crepererum:
This also fixes #1811.

crepererum (approved):

PR #2588, opened 2021-09-20
refactor: introduce `ChunkId` type

crepererum:


data_types/src/chunk_metadata.rs
```diff
@@ -196,6 +199,32 @@ impl ChunkSummary {
...
+impl ChunkId {
+    pub fn new(id: u32) -> Self {
+        Self(id)
+    }
+
+    pub fn get(&self) -> u32 {
+        self.0
+    }
+
+    pub fn next(&self) -> Self {
```
alamb, line 217: It might be woth a doc comment here that it is the caller's responsibility to ensure this is called on the current maximum chunk_id. I worry it may be easy to misuse (as nothing about this API prevent duplicates chunk ids)

alamb (approved):
Seems like a straightforward change 👍 

What is the rationale for the change? I am not opposed to adding a wrapper around a `u32` but I also wonder how much value that wrapper is adding.

Perhaps you have plans for a future PR that will add additional logic to `ChunkId`?

crepererum:
> What is the rationale for the change? 

1. That was brought up while introducing the chunk order. Originally the order also had no newtype wrapper so there were two u32 values per chunk which was pretty confusing. For the order I since fixed that by adding a newtype wrapper, so at least the two cannot be confused any longer.
2. But I still feel that some return values and data structures could benefit from the ID wrapper.
3. This change could allow us to switch the underlying type more easily (e.g. to an UUID). We might need the latter when we wanna read from multiple catalogs at the same time in some midterm future (either due to cluster re-balancing where  one server merges the catalogs of two previous servers or for read-only replicas that tail a preserved catalog while also reading from Kafka).

data_types/src/chunk_metadata.rs
crepererum: done

PR #2589, opened 2021-09-20
feat: teach preserved catalog to handle delete predicates

crepererum:
~Based on #2572 because this PR here uses the fact that predicates are not carrying table names any longer.~

For #2518. There will be one last PR on top of this that adds the `Db` => `PreservedCatalog` wiring + testing.


NGA-TRAN (approved):
Code and tests look good

PR #2591, opened 2021-09-20
feat: Add support for custom tracing header name via JAEGER_TRACE_CONTEXT_HEADER_NAME

alamb:
# Rationale
The InfluxData cloud system has migrated to use a custom JAEGER header (rather than `uber-trace-id`) for passing around span contexts. Thus implement `JAEGER_TRACE_CONTEXT_HEADER_NAME` as described in https://pkg.go.dev/k8s.io/ingress-nginx/internal/ingress/controller/config

# Changes
1. Allow specifying the tracing header via the standard environment variable (rather than hard coded string)

# Current behavior
1. Looks for header `uber-trace-id` , if found parse as Jaeger context
2. if not present, looks for `b3` formatted headers `X-B3-*`

# New behavior
1 ,If env / option is set, look for that value, otherwise look for `uber-trace-id`
2. if not present, looks for `b3` formatted headers `X-B3-*`

This is mostly plumbing and a small amount of actual code. The next PR in the series is https://github.com/influxdata/influxdb_iox/pull/2600


src/influxdb_ioxd.rs
```diff
@@ -236,12 +237,16 @@ async fn serve(
     // Construct a token to trigger shutdown of API services
     let frontend_shutdown = tokio_util::sync::CancellationToken::new();
 
+    let trace_header_parser = TraceHeaderParser::new()
```
alamb, line 240: this is where the plumbing of the new command line / env variables are connected to the actual parsing

crepererum (approved):

alamb:
Related to https://github.com/influxdata/influxdb_iox/issues/2367

PR #2594, opened 2021-09-20
test: Add end to end tracing test with mock Jaeger agent

alamb:
# Rationale
I need the ability to test all the environment variable plumbing is hooked up for custom trace headers in https://github.com/influxdata/influxdb_iox/pull/2591. I also wanted the ability to write a test for gRPC tracing in https://github.com/influxdata/influxdb_iox/pull/2553 but did not have a harness to do so

# Changes:
1. Add a way to configure`TestServer` / `ServerFixture` to include some arbitrary header with all client requests
2. Add a server that captures UDP messages
3. Add a simple test that verifies we get UDP messages from the jager exporter with some expected content


pauldix (approved):
Lot of plumbing but looks pretty straight forward. When is the Jaeger feature enabled and when isn't it?

tests/common/server_fixture.rs
```diff
@@ -279,8 +281,38 @@ struct TestServer {
...
+    /// Headers to add to all client requests
+    client_headers: Vec<(HeaderName, HeaderValue)>,
+}
+
+impl TestConfig {
+    pub fn new() -> Self {
+        Default::default()
+    }
+    // add a name=value environment variable when starting the server
+    pub fn with_env(mut self, name: impl Into<String>, value: impl Into<String>) -> Self {
```
e-dard, line 303: 👌

tests/common/server_fixture.rs
```diff
@@ -289,20 +321,21 @@ struct Process {
 }
 
 impl TestServer {
-    fn new(env: Vec<(String, String)>) -> Self {
+    fn new(test_config: TestConfig) -> Self {
```
e-dard, line 324: nit `config`. `test_config` stutters a bit :)

tests/common/udp_listener.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// A UDP message received by this server
+#[derive(Clone)]
+pub struct Message {
+    data: Vec<u8>,
+}
+
+impl std::fmt::Debug for Message {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Message({} bytes: {}", self.data.len(), self.to_string())
```
e-dard, line 25:
```suggestion
        write!(f, "Message({} bytes: {})", self.data.len(), self.to_string())
```

pauldix:
Right, but I'm wondering why it would be feature flagged in the end to end test?

tests/common/udp_listener.rs
```diff
@@ -0,0 +1,134 @@
...
+                let mut data = vec![0; 1024];
+
+                select! {
+                    _ = captured_token.cancelled() => {
+                        println!("Received shutdown request");
+                        return;
+                    },
+                    res  = socket.recv_from(&mut data) => {
+                        let (sz, _origin) = res.expect("successful socket read");
+                        data.resize(sz, 0);
```
e-dard, line 77: Is the expectation here that you are truncating the buffer to a length equal to the number of bytes written into it? If so you could use `data.truncate(sz)` I think.

tests/end_to_end_cases/tracing.rs
```diff
@@ -0,0 +1,122 @@
...
+fn run_test() -> bool {
+    println!("Skipping test because jaeger feature not enabled");
+    false
+}
+
+async fn setup() -> (UdpCapture, ServerFixture) {
+    let udp_capture = UdpCapture::new().await;
+
+    let test_config = TestConfig::new()
+        .with_env("TRACES_EXPORTER", "jaeger")
```
e-dard, line 25: I wonder if any of these env names are used elsewhere and if they could be constants.

e-dard (approved):
LGTM 👍

alamb:
> Right, but I'm wondering why it would be feature flagged in the end to end test?

@pauldix  because `cargo test` doesn't build with support for jaeger (which is being tested...) If you try and set `TRACES_EXPORTER=jaeger` without building with `--feature=jaeger` you get an error about jaeger not being configured

tests/end_to_end_cases/tracing.rs
alamb: They are defined the `TracingConfig` here https://github.com/influxdata/influxdb_iox/blob/529854332126aad68ef79666e683aa388810ce8d/trace_exporters/src/lib.rs#L20-L65

I sort of think it is appropriate for the end to end tests to enumerate them again explicitly as the environment names form part of the "public API" for configuring IOx and ensuring it is clear if the config values are changed then users of IOx would have to change them too I think is reasonable

tests/common/udp_listener.rs
alamb:
```suggestion
                        data.truncate(sz, 0);
```

tests/common/server_fixture.rs
alamb: to be clear I did not change `test_config` to `config` because there are already way too many things named config and thus I felt this name was clearer if a bit more verbose and stuttery :)

alamb:
Related to https://github.com/influxdata/influxdb_iox/issues/2367

PR #2595, opened 2021-09-21
ci: add cargo audit

### LINKED ISSUE #2575: Run Cargo Audit (or similar) in CI
This will detect RUSTSEC vulnerabilities such as the one fixed by #2574

crepererum:
For now we ignore warnings (e.g. "crate is yanked" or "crate is
unmaintained") because:

- internal crates w/ names of crates.io crates (e.g. `query`) are
  treated like crates.io-crates even though they shouldn't, see
  https://github.com/rustsec/rustsec/issues/232
- many crates are currently unmaintained and require a bit of upstream
  work (e.g. `chrono` is currently not very active but uses an old
  version of `time` which uses the unmaintained `stdweb`)

Closes #2575.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,33 @@
+[advisories]
+ignore = [
+    # title: term is looking for a new maintainer
+    # why needed: used by `prettytable-rs` which is directly used by IOx but also by arrow
+    # upstream issue: https://github.com/phsym/prettytable-rs/issues/119
+    "RUSTSEC-2018-0015",
+
+    # title: memmap is unmaintained
+    # why needed: used by `symbolic` which is used by `pprof`
+    # upstream issue: https://github.com/getsentry/symbolic/issues/304
+    "RUSTSEC-2020-0077",
+
+    # title: difference is unmaintained
+    # why needed: used by `mockito`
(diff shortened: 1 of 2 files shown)


alamb (commented):
My one hesitation with this PR is the question "what is the purpose of running a test regularly whose output we ignore?" 

Is it a goal to get a clean audit run? If so should we track the work items to do so?

Is it a goal that our audit results don't get worse? If so perhaps we should figure out a way to "diff" the audit results or something to ensure nothing else gets added to the list?

I looked at the output of the run on circle CI and it looks reasonable.
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/10570/workflows/1645c821-6d45-405f-bf37-2c3160221663/jobs/66400

FWIW arrow 6 replaces `pretty-table` with the actively maintained comfytable: https://github.com/apache/arrow-rs/pull/656

So once we pick up arrow6 we can get rid of that dependency

tustvold:
An approach I've used previously is to explicitly whitelist things, along with the relevant upstream tracking issue(s). Disabling warnings as this PR does is still valuable as it will catch security vulnerabilities, but it'd be even better to catch all the things! 😀

crepererum:
OK, let me try a whitelist approach...

alamb (approved):

PR #2596, opened 2021-09-21
test: do not (ab)use the panic hook for replay tests

crepererum:
The old construct uses a single assert-statement for both:

- "bubble-up" scenario, were a panic should fire
- a check, were a panic should not fire

That makes it easy to add new tests. However we need two rather
questionable things to make that work:

- catch panic: to convert an assertion to a check
- a custom panic hook: to make tests not overly verbose (aka caught
  panics should not show up on stdout)

Esp. the custom panic hook doesn't work too well w/ multi-threaded tests
since it might swallow error messages from unrelated tests and makes
debugging of CI failures hard.

So instead of using assertions for checks, we now implement a proper
assertion and a check for each test. That's a bit more code per check
but easier probably more stable.

### DIFF
--- a/server/src/db/replay.rs
+++ b/server/src/db/replay.rs
@@ -816,44 +816,48 @@ mod tests {
                             .map(|(s1, s2)| (s1.as_ref(), s2.as_ref()))
                             .collect();
 
-                        Self::eval_assert(
-                            || {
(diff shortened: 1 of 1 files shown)


e-dard (approved):
LGTM 👍

PR #2597, opened 2021-09-21
chore: upgrade azure dependencies

crepererum:
Required for #2595 to reduce a potential ignore-list.


alamb (approved):
This also removes quite a few older packages (like older versions of `rand` 👍 )

PR #2600, opened 2021-09-21
feat: Add support to create new traces in IOx via JAEGER_DEBUG_HEADER

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/2367, related to https://github.com/influxdata/influxdb_iox/issues/2549

# Rationale
When making requests to IOx, it would be nice  if we could request tracing without having to make up unique trace ids / spans such as `4459495:30434:0:1` for each request, as described in https://github.com/influxdata/influxdb_iox/blob/main/CONTRIBUTING.md#step-3-send-a-request-with-trace-context

# Changes
1. Add `JAEGER_DEBUG_HEADER` consistent with https://pkg.go.dev/k8s.io/ingress-nginx/internal/ingress/controller/config
2. Remove redundant default values in trace_http (use the defaults from command line)
3. Update the docs
4. Update perf.py

# Behavior
1. If the `jaeger_trace_context_header_name` (defaults to `uber-trace-id`) is specified, then the trace/span from that is used
2. otherwise, if b3 headers are present, use that
3. otherwise, if  `JAEGER_DEBUG_HEADER` (default to `jaeger-debug-id`) is in the request, create a new trace and, and add a `trace-id` to the response, as described in https://github.com/jaegertracing/documentation/issues/349

# Results
I tried this out locally with jaeger via the following command

```shell
./target/debug/influxdb_iox database query   26f7e5a4b7be365b_917b97a92e883afc  'select count(*), avg(available) from mem group by host' --header jaeger-debug-id:my_id
```

### DIFF
--- a/perf/perf.py
+++ b/perf/perf.py
@@ -663,7 +663,7 @@ def run_test_battery(battery_name, router_id, writer_id, debug=False, do_trace=F
         headers = {}
         if do_trace:
             # TODO remove this after IOx can be configured to sample 100% of traces
-            headers['uber-trace-id'] = '%x:%x:0:1' % (random.randrange(0, 2 ** 64), random.randrange(0, 2 ** 64))
+            headers['jaeger-debug-id'] = 'from-perf'
         response = requests.get(url=query_url, params=params, headers=headers)
         time_delta = '%dms' % math.floor((time.time() - time_start) * 1000)
 
--- a/src/influxdb_ioxd.rs
+++ b/src/influxdb_ioxd.rs
@@ -238,7 +238,12 @@ async fn serve(
     let frontend_shutdown = tokio_util::sync::CancellationToken::new();
 
     let trace_header_parser = TraceHeaderParser::new()
-        .with_jaeger_header_name(config.tracing_config.jaeger_trace_context_header_name);
+        .with_jaeger_trace_context_header_name(
+            config
+                .tracing_config
+                .traces_jaeger_trace_context_header_name,
+        )
+        .with_jaeger_debug_name(config.tracing_config.traces_jaeger_debug_name);
 
     // Construct and start up gRPC server
 
--- a/trace_exporters/src/lib.rs
+++ b/trace_exporters/src/lib.rs
@@ -62,7 +62,7 @@ pub struct TracingConfig {
     )]
     pub traces_exporter_jaeger_service_name: String,
 
-    /// Tracing: http request header that contains the Jaeger service name
+    /// Tracing: specifies the header name used for passing trace context
     ///
     /// Only used if `--traces-exporter` is "jaeger".
     #[structopt(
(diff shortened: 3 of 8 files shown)


src/influxdb_ioxd.rs
```diff
@@ -238,7 +238,12 @@ async fn serve(
     let frontend_shutdown = tokio_util::sync::CancellationToken::new();
 
     let trace_header_parser = TraceHeaderParser::new()
-        .with_jaeger_header_name(config.tracing_config.jaeger_trace_context_header_name);
+        .with_jaeger_trace_context_header_name(
```
alamb, line 241: I renamed this field / parameter to follow the jaeger standard config names (even though it is a mouthful)

trace_exporters/src/lib.rs
```diff
@@ -62,15 +62,25 @@ pub struct TracingConfig {
...
     pub traces_exporter_jaeger_service_name: String,
 
-    /// Tracing: http request header that contains the Jaeger service name
+    /// Tracing: specifies the header name used for passing trace context
     ///
     /// Only used if `--traces-exporter` is "jaeger".
     #[structopt(
         long = "--traces-exporter-jaeger-trace-context-header-name",
         env = "JAEGER_TRACE_CONTEXT_HEADER_NAME",
         default_value = "uber-trace-id"
```
alamb, line 71: note the default value is specified here (only)

trace_http/src/ctx.rs
```diff
@@ -15,8 +15,6 @@ const B3_TRACE_ID_HEADER: &str = "X-B3-TraceId";
 const B3_PARENT_SPAN_ID_HEADER: &str = "X-B3-ParentSpanId";
 const B3_SPAN_ID_HEADER: &str = "X-B3-SpanId";
 
-const DEFAULT_JAEGER_TRACE_HEADER: &str = "uber-trace-id";
```
alamb: use defaults from `TracingConfig` now rather than something hard coded here

perf/perf.py
```diff
@@ -663,7 +663,7 @@ def run_test_battery(battery_name, router_id, writer_id, debug=False, do_trace=F
         headers = {}
         if do_trace:
             # TODO remove this after IOx can be configured to sample 100% of traces
```
crepererum, line 665:
```suggestion
```

This todo is somewhat fixed now?

trace_http/src/ctx.rs
```diff
@@ -411,4 +419,40 @@ mod tests {
...
+    #[test]
+    fn test_jaeger_debug_name() {
+        let parser = TraceHeaderParser::new().with_jaeger_debug_name("force-a-trace");
+
+        let collector: Arc<dyn TraceCollector> = Arc::new(trace::LogTraceCollector::new());
+
+        let mut headers = HeaderMap::new();
+
+        assert!(parser.parse(&collector, &headers).unwrap().is_none());
+        headers.insert("force-a-trace", HeaderValue::from_static("please do"));
```
crepererum, line 432: Should the `"please do"` string occur in the generated spans? Like if I force-trigger tracing, is the ID from this "force header" propagated through the system?

crepererum (approved):
Looks good, but I have to admit that I'm not too familiar w/ tracing so if you need someone checking the semantics, you probably needs someone else :D

perf/perf.py
alamb: I do think the todo is somewhat fixed, but I believe that @jacobmarble  had in mind some way to start up the IOx server so that it made traces / sent spans for *every* request that was made, regardless of any headers which were included. 

This PR improves things so that the contents of the header don't need to be calculated but I would say the TODO note still remains

trace_http/src/ctx.rs
alamb: That is an excellent idea -- I think that is how Jaeger is supposed to work (though it doesn't seem to work in our cloud for some reason, which @primeroz said was a known issue). Filed https://github.com/influxdata/influxdb_iox/issues/2614 to track

PR #2601, opened 2021-09-21
fix: All chunks now are applied delete predicates during scan

NGA-TRAN:
This is another task of #2396 that include:
1. A fix that have all chunks  of all scenarios apply the delete predicate during scan
2. A lot more test scenarios to have at least 2 deletes  at different chunk stages
3. A refactor


predicate/src/predicate.rs
```diff
@@ -187,26 +187,24 @@ impl Predicate {
         val
     }
 
-    /// Merge the given delete predicates into this select predicate
```
NGA-TRAN, line 190: all change sin this file is just a refactor to move all content of `add_negated_delete_exprs` to `merge_delete_predicates`

query/src/provider.rs
```diff
@@ -861,19 +861,7 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
...
-    ///                │   (optional)    │   <-- Only added if the input output_sort_key is not empty
-    ///                └─────────────────┘
-    ///                          ▲
-    ///                          │
-    ///                          │
-    ///                ┌─────────────────┐
-    ///                │IOxReadFilterNode│
-    ///                │    (Chunk)      │
-    ///                └─────────────────┘
-    ///```
```
NGA-TRAN, line 876: Since this function invokes the other function, I remove the duplicates function description that is no longer up to date

query/src/provider.rs
```diff
@@ -924,8 +910,9 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
     ) -> Result<Vec<Arc<dyn ExecutionPlan>>> {
         let mut plans: Vec<Arc<dyn ExecutionPlan>> = vec![];
 
-        // output is not required to be sorted or no chunks provided, only create a read filter for all chunks
-        if output_sort_key.is_empty() || chunks.is_empty() {
+        // Since now each chunk may include delete predicates, we need to create plan for each chunk but
+        // if there is no chunk, we still need to return a plan
+        if chunks.is_empty() {
```
NGA-TRAN, line 922: This is the bug fix

query_tests/src/scenarios.rs
```diff
@@ -1505,11 +1505,231 @@ impl DbSetup for DeleteMultiExprsFromOsOneMeasurementOneChunk {
...
+        )
+        .await;
+
+        // return scenarios to run queries
+        // NGA todo: bug scenario_rub_os_unload_rub, scenario_os
+        vec![scenario_rub_os]
+    }
+}
+
+// NGA todo: Add scenarios that have many different types of chunks at the same time (MUB, RUB, OS) when a or several deletes happen
```
NGA-TRAN, line 1729: I planned to add these tests into this PR but since there are clear scenarios that do not work yet, I will work on them before adding more complicated tests

predicate/src/predicate.rs
```diff
@@ -187,26 +187,24 @@ impl Predicate {
...
+    /// Merge the given delete predicates into this select predicate.
+    /// Since we wan to eliminate data filtered by the delete predicates,
```
alamb, line 191:
```suggestion
    /// Since we want to eliminate data filtered by the delete predicates,
```

predicate/src/predicate.rs
```diff
@@ -187,26 +187,24 @@ impl Predicate {
...
+    /// then added/merged into the selection one
     pub fn merge_delete_predicates<S>(&mut self, delete_predicates: &[S])
     where
         S: AsRef<Self>,
     {
-        self.add_negated_delete_exprs(delete_predicates);
-    }
+        // Create a list of disjunctive negated expressions.
+        // Example: there are two deletes as follows (note that time_range is stored separated in the Predicate
+        //  but we need to put it together with the exprs hee)
```
alamb, line 200:
```suggestion
        //  but we need to put it together with the exprs here)
```

query/src/provider.rs
alamb: Given I think we are assuming that deletes are not frequent, I wonder if it is worth keeping the optimization for what is likely to be the common case and have the check be something like

```rust
if (output_sort_key.is_empty() && no_delete_predicates(chunks)) || chunks.is_empty() {
...
}
```

query_tests/src/scenarios.rs
```diff
@@ -1505,11 +1505,231 @@ impl DbSetup for DeleteMultiExprsFromOsOneMeasurementOneChunk {
...
+        )
+        .await;
+
+        // delete happens when data in MUB then moved to RUB, then persisted, and then RUB is unloaded
+        let _scenario_os =
+            make_delete_mub_to_os_delete(lp_lines.clone(), pred1, pred2, table_name, partition_key)
+                .await;
+
+        // return scenarios to run queries
+        // NGA todo: BUG in scenario_os
```
alamb, line 1576: I thought this PR was supposed to fix the bug in the os scenario?

query_tests/src/scenarios.rs
```diff
@@ -1702,6 +2104,48 @@ async fn make_delete_rub_to_os(
     }
 }
 
+async fn make_delete_rub_to_os_delete(
```
alamb, line 2107: given the amount of code related to deletes, I wonder if it is worth moving the delete related code (eventually) into their own modules, like `query_tests/src/scenarios/delete.rs` or something

alamb (approved):

query/src/provider.rs
NGA-TRAN: Yeah, I thought about it, too. Let me create a ticket for this. So many delete tasks that without a ticket I will forget

query_tests/src/scenarios.rs
NGA-TRAN: Not yet. I am working on this fix

query_tests/src/scenarios.rs
NGA-TRAN: I have addressed this in this PR

query_tests/src/scenarios.rs
NGA-TRAN: @alamb It turn out the fix above did fix this and all the known failed tests. I will have an PR to turn all the tests on shortly

PR #2602, opened 2021-09-21
chore: Update datafusion

alamb:
# Rationale:
1. Incorporate change to make planning `async` added by @rdettai in https://github.com/apache/arrow-datafusion/pull/1013 :wave:

# Changes:
1. Update to use latest datafusion: https://github.com/apache/arrow-datafusion/commit/299ab7d1c37c707fcd500d3428abbdbe4dc5399b
2. run `cargo update` and check in results
3. Update code to use new `async` planning API


query/src/exec/context.rs
```diff
@@ -50,10 +51,11 @@ pub const DEFAULT_SCHEMA: &str = "iox";
 /// and is needed to create plans with the IOx extension nodes.
 struct IOxQueryPlanner {}
 
+#[async_trait]
 impl QueryPlanner for IOxQueryPlanner {
     /// Given a `LogicalPlan` created from above, create an
     /// `ExecutionPlan` suitable for execution
-    fn create_physical_plan(
+    async fn create_physical_plan(
```
alamb, line 58: We just have to sprinkle a little more `async`...

crepererum (approved):

PR #2605, opened 2021-09-21
feat: optimize scan for chunks without delete preidcates and without the need to sort output

### LINKED ISSUE #2603: Optimize scan for chunks with no delete predicates
# Rationale
PR #2601 fixed a bug in delete (for #2396) but remove the optimization to scan multiple chunks in one read operator. this ticket is to turn it back on by adding a function  `no_delete_predicates(chunks) -> bool`. So the current line:
  [ `if chunks.is_empty()`](https://github.com/influxdata/influxdb_iox/blob/4b0d6a8b084c162901324e5c7bdbf9086e21cb56/query/src/provider.rs#L922)
will turn to
`if (output_sort_key.is_empty() && no_delete_predicates(chunks)) || chunks.is_empty() {`

NGA-TRAN:
Closes #2603 

### DIFF
--- a/query/src/provider.rs
+++ b/query/src/provider.rs
@@ -950,12 +950,9 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
     }
 
     fn no_delete_predicates(chunks: &[Arc<C>]) -> bool {
-        for chunk in chunks {
-            if !chunk.delete_predicates().is_empty() {
-                return false;
-            }
-        }
-        true
+        chunks
+            .iter()
+            .all(|chunk| chunk.delete_predicates().is_empty())
     }
 
     /// Find the columns needed in the primary key across schemas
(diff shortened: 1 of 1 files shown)


query/src/provider.rs
```diff
@@ -947,6 +949,15 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
         sorted_chunk_plans
     }
 
+    fn no_delete_predicates(chunks: &[Arc<C>]) -> bool {
+        for chunk in chunks {
```
alamb, line 953: I think you might be able to write this more "idomatically" with something like

```rust
chunks.iter().all(|chunk| chunk.delete_predicates().is_empty())
```

But this way works too

alamb (approved):

query/src/provider.rs
NGA-TRAN: Let me try the rust way

PR #2606, opened 2021-09-22
feat: teach `Db` to preserve delete predicates

### LINKED ISSUE #2518: Persist delete predicates with chunk

crepererum:
Closes #2518.

### DIFF
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -1219,8 +1219,14 @@ pub(crate) fn checkpoint_data_from_catalog(catalog: &Catalog) -> CheckpointData 
             };
 
             files.insert(path, m);
+        }
 
-            // capture delete predicates
+        // capture delete predicates
+        // We should only report persisted chunks or chunks that are currently being persisted, because the
+        // preserved catalog does not care about purely in-mem chunks.
+        if matches!(guard.stage(), ChunkStage::Persisted { .. })
+            || guard.is_in_lifecycle(ChunkLifecycleAction::Persisting)
+        {
             for predicate in guard.delete_predicates() {
                 let predicate_ref: &Predicate = predicate.as_ref();
                 let addr = (predicate_ref as *const Predicate) as usize;
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -20,7 +20,7 @@ use snafu::{ensure, OptionExt, ResultExt, Snafu};
 
 use ::lifecycle::{LifecycleChunk, LockableChunk, LockablePartition};
 use data_types::{
-    chunk_metadata::{ChunkId, ChunkOrder, ChunkSummary},
+    chunk_metadata::{ChunkId, ChunkLifecycleAction, ChunkOrder, ChunkSummary},
     database_rules::DatabaseRules,
     partition_metadata::{PartitionSummary, TableSummary},
     server_id::ServerId,
@@ -166,6 +166,14 @@ pub enum Error {
 
     #[snafu(display("Cannot replay: {}", source))]
     ReplayError { source: crate::db::replay::Error },
+
+    #[snafu(display(
+        "Error while commiting delete predicate on preserved catalog: {}",
+        source
+    ))]
+    CommitDeletePredicateError {
+        source: parquet_file::catalog::core::Error,
+    },
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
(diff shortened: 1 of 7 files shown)


e-dard:
A PR description is always useful. Particularly for our future selves. Perhaps a short summary of the recent problem we noticed wiht deletes, what you found, and that this is the fix, would suffice.

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
         join_handle.await.unwrap();
     }
 
+    #[tokio::test]
+    async fn delete_predicate_preservation() {
+        // Test that delete predicates are stored within the preserved catalog
+        maybe_start_logging();
```
e-dard, line 3610: I don't think we're supposed to commit this line on tests are we?

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
         join_handle.await.unwrap();
     }
 
+    #[tokio::test]
+    async fn delete_predicate_preservation() {
```
e-dard, line 3608: Nice test 👌

e-dard (approved):
LGTM 👍

server/src/db.rs
```diff
@@ -521,24 +529,53 @@ impl Db {
...
+                .context(DeleteFromTable { table_name })?;
+            let partitions = table.partitions();
+            for partition in partitions {
+                let partition = partition.write();
+                let chunks = partition.chunks();
+                for chunk in chunks {
+                    // save the delete predicate in the chunk
+                    let mut chunk = chunk.write();
+                    chunk
+                        .add_delete_predicate(Arc::clone(&delete_predicate))
```
alamb, line 550: Will it be a problem here if this returns with an error? It may leave the chunks in a inconsistent state:

The predicate is applied to some chunks (in memory) but not all, and no predicates are persisted at all.

I see this type of partial failure was possible prior to this PR so I don't think we need to fix it here, but maybe it is worth a ticket to record the behavior / note we have more work to do

server/src/db.rs
```diff
@@ -521,24 +529,53 @@ impl Db {
...
+                    }
+                }
             }
         }
 
+        if !affected_persisted_chunks.is_empty() {
+            let mut transaction = self.preserved_catalog.open_transaction().await;
+            transaction
+                .delete_predicate(&delete_predicate, &affected_persisted_chunks)
+                .context(CommitDeletePredicateError)?;
```
alamb, line 572: likewise here -- if either of these calls fail (due to a S3 error, for example) I wonder if we should panic/ force the database through recovery (as the in memory and preserved state are now inconsistent)

server/src/db.rs
alamb: I think it is fine to commit. 

I have done so in many tests and I think it is quite helpful (so if you are debugging something you don't have to add this line and recompile).

Unless one has `RUST_LOG` set in your environment `maybe_start_logging` does not do anything.

server/src/db.rs
pauldix, line 550: Can this be updated to only persist to chunks that actually match the predicate?

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
...
+        )
+        .await
+        .unwrap();
+
+        // ==================== check: delete predicates ====================
+        let closure_check_delete_predicates = |db: &Db| {
+            for chunk in db.catalog.chunks() {
+                let chunk = chunk.read();
+                if chunk.addr().partition_key.as_ref() == "part_b" {
+                    // Strictly speaking not required because the chunk was persisted AFTER the delete predicate was
```
alamb, line 3737: I don't understand why `part_b` would not have the delete predicate

As I understand it, the sequence for that partition was
1. Write to `part_b`, move to RUB
2. Delete
3. Persist `part_b`

If we now restart the database, since `part_b` was persisted I would expect the delete predicates to still be present.

alamb (approved):

server/src/db.rs
NGA-TRAN: @alamb because the delete predicate is in the chunk (added when it was a RUB) so it will go in the chunk parquet file and no need to store it as extra in the chunk catalog

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
...
+        let expected = vec![
+            "+------+-----+----------+--------------------------------+",
+            "| part | row | selector | time                           |",
+            "+------+-----+----------+--------------------------------+",
+            "| a    | 10  | 0        | 1970-01-01T00:00:00.000000010Z |",
+            "| b    | 20  | 0        | 1970-01-01T00:00:00.000000020Z |",
+            "| c    | 30  | 0        | 1970-01-01T00:00:00.000000030Z |",
+            "| d    | 40  | 0        | 1970-01-01T00:00:00.000000040Z |",
+            "+------+-----+----------+--------------------------------+",
+        ];
```
NGA-TRAN, line 3758: Nice

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
...
+        // NOTE: partition "c" is gone here because it was not written to object store
+        let expected = vec![
+            "+------+-----+----------+--------------------------------+",
+            "| part | row | selector | time                           |",
+            "+------+-----+----------+--------------------------------+",
+            "| a    | 10  | 0        | 1970-01-01T00:00:00.000000010Z |",
+            "| b    | 20  | 0        | 1970-01-01T00:00:00.000000020Z |",
+            "| d    | 40  | 0        | 1970-01-01T00:00:00.000000040Z |",
+            "+------+-----+----------+--------------------------------+",
+        ];
```
NGA-TRAN, line 3786: Nice

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
...
+        let batches = run_query(Arc::clone(&db), "select * from cpu order by time").await;
+        assert_batches_sorted_eq!(&expected, &batches);
+
+        // ==================== do: remove checkpoint files ====================
```
NGA-TRAN, line 3790: Question: we have 2 kinds of file: checkpoint and catalog transaction, right? The below code is to delete the transaction files, not the checkpoint files?

server/src/db.rs
```diff
@@ -3559,6 +3604,239 @@ mod tests {
...
+        // NOTE: partition "c" is gone here because it was not written to object store
+        let expected = vec![
+            "+------+-----+----------+--------------------------------+",
+            "| part | row | selector | time                           |",
+            "+------+-----+----------+--------------------------------+",
+            "| a    | 10  | 0        | 1970-01-01T00:00:00.000000010Z |",
+            "| b    | 20  | 0        | 1970-01-01T00:00:00.000000020Z |",
+            "| d    | 40  | 0        | 1970-01-01T00:00:00.000000040Z |",
+            "+------+-----+----------+--------------------------------+",
+        ];
```
NGA-TRAN, line 3835: 👍

NGA-TRAN (approved):
Very nice.

server/src/db.rs
alamb: Ah, as in there are no delete predicates because it was applied when creating the parquet file and thus the rows have been permanently removed from storage 👍

server/src/db.rs
crepererum: Let me fix that in a follow-up PR. IMHO `add_delete_predicate` should not fail (because if it fails that's really an internal bug where it confuses some chunk state). If `add_delete_predicate` cannot fail, it could first collect the affected chunks, then create the transaction (which can fail) and then add the delete predicates to the in-memory state. That would be similar to how we handle other catalog transactions as well.

server/src/db.rs
crepererum: It deletes checkpoint files as the comment says. It's a bit confusing because both file types are produces by `iox_object_store.catalog_transaction_files` but there is a check within the loop (`file.is_checkpoint`) that ensures we're only removing checkpoint files here.

server/src/db.rs
NGA-TRAN: @alamb I think there are 2 different concepts here:
1. There is a delete predicate saved with the in-memory RUB chunk metadata and it will go to the Parquet file as is when it is persisted.
2. If the delete happens after the chunk is persisted, that delete predicate has to be kept somewhere.
 
Because `part_b` was persisted after delete,  it belongs to the first one.

We have not done hard delete yet hence the rows are still in the chunks

server/src/db.rs
NGA-TRAN: Thank @crepererum

PR #2607, opened 2021-09-22
feat: implement jaeger-agent protocol directly

tustvold:
_Most of this PR is auto-generated code, it is not nearly as intimidating as it may first seem._

OpenTelemetry has a fair few non-trivial dependencies, such as `crossbeam`, along with being a non-trivial dependency on their own. I initially looked into contributing some additional feature flags upstream to reduce the amount of stuff we would bring in, but this turned out to be highly non-trivial without either making significant changes to the crates layout or littering the codebase with conditional compilation attributes.

It is my expectation/understanding that IOx will only even need to talk to a jaeger-agent or an open telemetry collector, with the latter acting as the fan-out to whatever other trace collector is desired. Both of these have defined APIs using thrift and protobuf respectively and so can for the most part be auto-generated.

This PR therefore just implements the jaeger-agent protocol directly and removes the opentelemetry dependency completely.

Unfortunately as I discovered the thrift support within Rust is a little bit ropey, but I think I've documented all the particularly exciting quirks.


Cargo.toml
```diff
@@ -191,7 +191,6 @@ default = ["jemalloc_replacing_malloc"]
 azure = ["object_store/azure"] # Optional Azure Object store support
 gcp = ["object_store/gcp"] # Optional GCP object store support
 aws = ["object_store/aws"] # Optional AWS / S3 object store support
-jaeger = ["trace_exporters/jaeger"] # Enable optional jaeger tracing support
```
tustvold, line 194: This no longer has non-trivial additional dependencies so can just be enabled by default

src/influxdb_ioxd.rs
```diff
@@ -188,17 +188,27 @@ pub async fn main(config: Config) -> Result<()> {
...
         grpc_listener,
         http_listener,
         trace_collector,
         app_server,
     )
-    .await
+    .await;
+
+    if let Some(async_exporter) = async_exporter {
+        if let Err(e) = async_exporter.drain().await {
```
tustvold, line 207: This was actually "wrong" before, we never shutdown and drained the tracing exporter

trace_exporters/src/export.rs
```diff
@@ -0,0 +1,169 @@
...
+/// off a queue and writes them to the `AsyncExport`.
+///
+/// If this worker cannot keep up, and this queue fills up, spans will
+/// be dropped and warnings logged
+///
+/// Note: Currently this does not batch spans (#2392)
+///
+/// Note: This currently does not drain on shutdown
+#[derive(Debug)]
+pub struct AsyncExporter {
```
tustvold, line 35: This is the `OtelExporter` renamed and using a custom trait `AsyncExport` instead of the Otel `SpanExporter`. It also has logic to drain correctly on shutdown

trace_exporters/src/jaeger.rs
```diff
@@ -0,0 +1,297 @@
...
+            }
+            let to_read = buf.len().min(self.len - self.idx);
+            buf.copy_from_slice(&self.buffer[self.idx..(self.idx + to_read)]);
+            self.idx += to_read;
+            Ok(to_read)
+        }
+    }
+
+    #[tokio::test]
+    async fn test_jaeger() {
```
tustvold, line 208: We could easily use this to write complete end-to-end tracing tests. FYI @alamb

trace_exporters/src/lib.rs
```diff
@@ -7,13 +7,34 @@
...
 use structopt::StructOpt;
-use trace::TraceCollector;
 
-pub mod otel;
+pub mod export;
+
+mod jaeger;
+
+/// Auto-generated thrift code
+#[allow(
```
tustvold, line 22: The thrift code generation is not very good :laughing:

trace_exporters/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+```
+$ thrift --version
+Thrift version 0.13.0
+```
+
+Get the IDL definition
+
+```
+$ wget https://raw.githubusercontent.com/jaegertracing/jaeger-idl/master/thrift/jaeger.thrift https://raw.githubusercontent.com/jaegertracing/jaeger-idl/master/thrift/zipkincore.thrift https://raw.githubusercontent.com/jaegertracing/jaeger-idl/master/thrift/agent.thrift
```
tustvold, line 32: Unfortunately we have to bring in a load of only tangentially related definitions because the agent.thrift depends on them.

tustvold:
Marking as draft as there appear to be logical feature flag conflicts, but should be good for review

Cargo.lock
```diff
@@ -2592,46 +2592,6 @@ dependencies = [
...
 ]
 
-[[package]]
-name = "opentelemetry"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e1cf9b1c4e9a6c4de793c632496fa490bdc0e1eea73f0c91394f7b6990935d22"
-dependencies = [
- "async-trait",
- "crossbeam-channel",
```
tustvold, line 2602: This is still a dependency of rayon which is brought in by criterion - but this avoids it being bundled into our actual release builds

alamb (commented):
Starting to check this out...

trace_exporters/src/export.rs
```diff
@@ -0,0 +1,167 @@
...
+mod tests {
+    use super::*;
+    use trace::ctx::{SpanContext, SpanId, TraceId};
+
+    #[tokio::test]
+    async fn test_exporter() {
+        let (sender, mut receiver) = mpsc::channel(10);
+        let exporter = AsyncExporter::new(TestAsyncExporter::new(sender));
+
+        let root = SpanContext {
```
alamb, line 132: You could also use `SpanContext::new()` potentially here

trace_exporters/src/jaeger.rs
```diff
@@ -0,0 +1,297 @@
...
+        self.buf.extend_from_slice(buf);
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        let message_len = self.buf.len();
+        let r = self.socket.send(&self.buf);
+        self.buf.clear();
+        match r {
+            Ok(written) => {
```
alamb, line 130: I have no idea if this is ok or not (it seems like in this case it sends a truncated UDP message and drops the rest on the floor).

I suppose logging an error is probably the best thing to do now, though it might be worth logging `written` and `message_len` as well in the `error!` message so that we can tell if IOx is trying to do something crazy like send 100K messages, etc

trace_exporters/src/export.rs
```diff
@@ -0,0 +1,167 @@
...
+        self.join.clone().await
+    }
+}
+
+impl TraceCollector for AsyncExporter {
+    fn export(&self, span: Span) {
+        use mpsc::error::TrySendError;
+        match self.sender.try_send(Some(span)) {
+            Ok(_) => {
+                //TODO: Increment some metric
```
alamb, line 67: FYI there is a todo here. Perhaps the `AsyncExporter` could just have its own `AtomicUsize` recording for now and we can hook it up to the metrics crate if we find we want that level of detail

trace_exporters/src/jaeger.rs
```diff
@@ -0,0 +1,297 @@
...
+/// `JaegerAgentExporter` receives span data and writes it over UDP to a local jaeger agent
+///
+/// Note: will drop data if the UDP socket would block
+pub struct JaegerAgentExporter {
+    /// The name of the service
+    service_name: String,
+
+    /// The agent client that encodes messages
+    client:
+        AgentSyncClient<TCompactInputProtocol<NoopReader>, TCompactOutputProtocol<MessageWriter>>,
```
alamb, line 24: I don't understand the need to have an input that will never read from anything. Is this something about the autogenerated code from thrift?

trace_exporters/src/jaeger.rs
alamb: That is awesome -- I added basic end to end tests here: https://github.com/influxdata/influxdb_iox/blob/main/tests/end_to_end_cases/tracing.rs

But with this code we could replace the UDP listener nonsense and actually validate the contents of the messages 🎉

trace_exporters/src/thrift/agent.rs
```diff
@@ -0,0 +1,305 @@
+// Autogenerated by Thrift Compiler (0.13.0)
```
alamb, line 1: FWIW I am a big fan of checking this generated code in (rather than auto generating it as part of build.rs) as it works much better with my IDE.

alamb (approved):
This is quite cool -- I read the code and the tests, and it makes lots of sense to me. Thank you @tustvold . Very well done

I also tested it out locally and I saw spans making their way to my local jaeger instance 🏅 


```shell
TRACES_EXPORTER=jaeger TRACES_EXPORTER_JAEGER_AGENT_HOST=localhost TRACES_EXPORTER_JAEGER_AGENT_PORT=6831 cargo run --features=jaeger -- run -v --object-store=file --data-dir=$HOME/.influxdb_iox --server-id=42
```

Which caused an error as the `jaeger` feature has been removed.

Perhaps we can update the command in https://github.com/influxdata/influxdb_iox/blob/remove-opentelemetry/CONTRIBUTING.md#L277 to be something like

```shell
TRACES_EXPORTER=jaeger TRACES_EXPORTER_JAEGER_AGENT_HOST=localhost TRACES_EXPORTER_JAEGER_AGENT_PORT=6831 cargo run  -- run -v --object-store=file --data-dir=$HOME/.influxdb_iox --server-id=42
```

alamb:
The CI tests are failing as they refer to the `jaeger` feature which has been removed. I think the scripts just need some tweaking

trace_exporters/src/jaeger.rs
tustvold: The error log message a couple of lines down logs in the event that the message is truncated, and includes the message_len and written fields. 

In the event a message is truncated, there isn't an obvious way to recover. Thrift is really supposed to be sent over an ordered stream, e.g. TCP, it is a bit of a hack to send it over UDP. The protocol is basically exploiting that a UDP packet is either dropped entirely or arrives in tact, and so must ensure that a given message payload is never split across a UDP packet.

I will include this as a comment

trace_exporters/src/jaeger.rs
tustvold: Unfortunately...

trace_exporters/src/export.rs
tustvold: This is carried over from the old OtelExporter, created #2613 to track

PR #2608, opened 2021-09-22
chore: clean up dependencies a bit

crepererum:
This doesn't remove any crates from our dependency set but should make compilation a bit faster because more work can be done sooner.


e-dard (approved):
LGTM 👍

internal_types/Cargo.toml
```diff
@@ -11,7 +11,6 @@ arrow = { version = "5.0", features = ["prettyprint"] }
 hashbrown = "0.11"
 indexmap = "1.6"
 itertools = "0.10.1"
-observability_deps = { path = "../observability_deps" }
```
alamb, line 14: 💯

PR #2617, opened 2021-09-22
test: more delete tests

NGA-TRAN:
Another task of #2396 


query_tests/src/sql.rs
```diff
@@ -950,3 +950,46 @@ async fn sql_select_with_two_deleted_data_from_multi_exprs() {
...
+    ];
+
+    run_sql_test_case(
+        scenarios::delete::ThreeDeleteThreeChunks {},
+        "SELECT * from cpu",
+        &expected,
+    )
+    .await;
+
+    // Hmm: It seems a bug here. I am working on this next
```
alamb, line 977: 🤔

alamb (approved):
One thing I wonder while reading all these scenarios is "do we have delete tests that delete data that is in multiple partitions? All of the scenarios in this PR seem to put the data in a single partition.

Nice testing (and finding bugs!)

NGA-TRAN:
> One thing I wonder while reading all these scenarios is "do we have delete tests that delete data that is in multiple partitions? All of the scenarios in this PR seem to put the data in a single partition.
Hah, even with those many tests of different scenarios that I think will lead to bugs due to different chunk stages and in the life cycle, I still miss multi-partition scenarios. Will add them.

PR #2618, opened 2021-09-22
test: enable disabled predicate pushdown test

alamb:
This test now passes (likely due to some small variation in stack sizes)

Note the underling stack overflow reported in  https://github.com/apache/arrow-datafusion/issues/419 is fixed in https://github.com/influxdata/influxdb_iox/pull/2635


alamb:
(this test passed for me locally, which is strange)

crepererum:
> (this test passed for me locally, which is strange)

Different OS? Different ulimit? (aka is the stacksize for all observers the same?)

alamb:
> Different OS? Different ulimit? (aka is the stacksize for all observers the same?)

I am not sure what is different as the same test  failed for me locally at some point. There is a reproducer attached to  https://github.com/apache/arrow-datafusion/issues/419 that I will use to work this out, but it was strange that this test started working again in IOx

NGA-TRAN:
The tests work again maybe because of improvement in reducing  the number of nested level by reducing the number of expressions somewhere in IOx/DF. This happened before

alamb:
I think @NGA-TRAN 's assessment of what happened is likely correct, given my findings on https://github.com/apache/arrow-datafusion/issues/419

alamb:
FYI @NGA-TRAN  I have also removed the rust stack size workaround in this PR (and will merge the PR once I get the fix from DataFusion)

alamb:
Rust stack size workaround included in https://github.com/influxdata/influxdb_iox/pull/2635. Will merge this PR in with changes to query_tests

PR #2619, opened 2021-09-23
refactor: `CatalogChunk::add_delete_predicate` cannot fail

crepererum:
This is prep work for improving the error handling around predicate preservation.

### DIFF
--- a/server/src/db/catalog/chunk.rs
+++ b/server/src/db/catalog/chunk.rs
@@ -469,7 +469,7 @@ impl CatalogChunk {
         }
     }
 
-    pub fn add_delete_predicate(&mut self, delete_predicate: Arc<Predicate>) -> Result<()> {
+    pub fn add_delete_predicate(&mut self, delete_predicate: Arc<Predicate>) {
         debug!(
             ?delete_predicate,
             "Input delete predicate to CatalogChunk add_delete_predicate"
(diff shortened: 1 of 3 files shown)


server/src/db/catalog/chunk.rs
```diff
@@ -469,15 +469,17 @@ impl CatalogChunk {
...
+    pub fn add_delete_predicate(&mut self, delete_predicate: Arc<Predicate>) {
         debug!(
             ?delete_predicate,
             "Input delete predicate to CatalogChunk add_delete_predicate"
         );
         match &mut self.stage {
             ChunkStage::Open { mb_chunk: _ } => {
                 // Freeze/close this chunk and add delete_predicate to its frozen one
-                self.freeze_with_predicate(delete_predicate)?;
+                self.freeze_with_predicate(delete_predicate).expect(
```
alamb, line 480: 👍

alamb: I tried to make `freeze_with_predicate` infallable as well -- the only way it can fail is if `self.stage` is not `Open` -- so in other words, I think `expect` is correct.

It would be neat to be able to avoid the error check (maybe push the panic lower) but not as part of this PR

alamb (approved):

PR #2620, opened 2021-09-23
feat: correct pbjson serialization of Timestamp and Duration

tustvold:
Certain well-known-types have a non-standard [JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json). This PR implements the custom mapping for google::protobuf::Timestamp and google::protobuf::Duration


google_types/src/lib.rs
```diff
@@ -13,52 +13,13 @@
 mod pb {
     pub mod google {
         pub mod protobuf {
-            use chrono::{NaiveDateTime, Utc};
-            use std::convert::{TryFrom, TryInto};
-
             include!(concat!(env!("OUT_DIR"), "/google.protobuf.rs"));
             include!(concat!(env!("OUT_DIR"), "/google.protobuf.serde.rs"));
-
-            impl TryFrom<Duration> for std::time::Duration {
```
tustvold, line 22: These are moved into their respective modules to avoid this becoming a very large file as further mappings are implemented

google_types/Cargo.toml
```diff
@@ -12,6 +12,9 @@ pbjson = { path = "../pbjson" }
 prost = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 
+[dev-dependencies]
+serde_json = "1.0"
```
tustvold, line 16: This makes me a bit sad, but I could not work out an alternative way to test the serialization logic

pbjson_build/src/lib.rs
```diff
@@ -63,51 +74,58 @@ impl Builder {
...
+        let config = Config {
+            extern_types: Default::default(),
         };
 
-        match descriptor {
-            Descriptor::Enum(descriptor) => generate_enum(&config, type_path, descriptor, writer)?,
-            Descriptor::Message(descriptor) => {
-                if let Some(message) = resolve_message(descriptors, descriptor) {
-                    generate_message(&config, &message, writer)?
+        let iter = self.descriptors.iter().filter(move |(t, _)| {
```
tustvold, line 94: I'm not sure why Github has decided I've rewritten this entire method, the "line" was all that was changed to add the ability to exclude types from auto-pbjsonification

google_types/src/duration.rs
```diff
@@ -0,0 +1,159 @@
...
+                nanos: -(negated.subsec_nanos() as i32),
+            })
+        } else {
+            Ok(std::time::Duration::from_secs_f64(secs).into())
+        }
+    }
+}
+
+/// Splits nanoseconds into whole milliseconds, microseconds, and nanoseconds
+fn split_nanos(mut nanos: u32) -> (u32, u32, u32) {
```
alamb, line 83: is u32 enough? Or should it be u64?

pbjson_build/src/lib.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/2620/files?w=1 helps to see a reasonable diff

alamb (approved):
I looked through the test cases and they looked good to me. I don't really understand serde enough to give it a super thorough review, but the code looks clean and easy to understand

google_types/src/duration.rs
alamb: Specifically, my reading of https://developers.google.com/protocol-buffers/docs/proto3#json suggests that timestamps this could contain up to 9 fractional digits which could overflow a u32, perhaps

tustvold: The max value of the nanos field is 999,999,999 which is stored as an i32 in the protobuf, so I think this should be fine. It is nanoseconds less than a second, not nanoseconds since epoch

PR #2621, opened 2021-09-23
refactor: make `Partition::force_drop_chunk` similar to `Partition::drop_chunk`

crepererum:
- Bubbles up "not found" error, the caller should reason about it
- Returns deleted chunk

This will be useful if we wanna handle predicates correctly during the lifecycle (at the moment we might loose a few new predicates on the way).

### DIFF
--- a/server/src/db/catalog/partition.rs
+++ b/server/src/db/catalog/partition.rs
@@ -13,7 +13,7 @@ use persistence_windows::{
     min_max_sequence::OptionalMinMaxSequence, persistence_windows::PersistenceWindows,
 };
 use predicate::predicate::Predicate;
-use snafu::Snafu;
+use snafu::{OptionExt, Snafu};
 use std::{
     collections::{btree_map::Entry, BTreeMap},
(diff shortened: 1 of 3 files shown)


server/src/db/catalog/partition.rs
```diff
@@ -284,8 +284,13 @@ impl Partition {
     }
 
     /// Drop the specified chunk even if it has an in-progress lifecycle action
-    pub fn force_drop_chunk(&mut self, chunk_id: ChunkId) {
-        self.chunks.remove(&chunk_id);
+    pub fn force_drop_chunk(&mut self, chunk_id: ChunkId) -> Result<Arc<RwLock<CatalogChunk>>> {
```
alamb, line 287:
```suggestion
    /// returning the dropped chunk
    pub fn force_drop_chunk(&mut self, chunk_id: ChunkId) -> Result<Arc<RwLock<CatalogChunk>>> {
```

alamb (approved):

PR #2622, opened 2021-09-23
refactor: use dedicated in-memory `Expr` type during IO

crepererum:
Place a dedicated `Expr` type between datafusion and protobuf. While
this type is currently only used during serialization, it will be used
within `Predicate` in a follow-up change to enable de-duplication of
predicates and avoid the double parsing of datafusion expressions (we
are already somewhat parsing them when we check for valid delete
predicates).

This change looks larger than it is. In practice it just separates the
conversion "datafusion => protobuf" into "datafusion => IOx => protobuf"
and "protobuf => datafusion" into "protobuf => IOx => datafusion". So
(apart from the error types) this is functionally the same.

### DIFF
--- a/predicate/src/expr.rs
+++ b/predicate/src/expr.rs
@@ -1,0 +1,496 @@
+use std::{
+    convert::{TryFrom, TryInto},
+    ops::Deref,
+};
+
+use generated_types::influxdata::iox::catalog::v1 as proto;
+use ordered_float::OrderedFloat;
+use snafu::{OptionExt, ResultExt, Snafu};
+
+/// Single expression to be used as parts of a predicate.
+///
+/// Only very simple expression of the type `<column> <op> <scalar>` are supported.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Expr {
+    /// Column (w/o table name).
+    column: String,
+
+    /// Operator.
+    op: Op,
+
+    /// Scalar value.
+    scalar: Scalar,
+}
+
+impl Expr {
+    /// Column (w/o table name).
+    pub fn column(&self) -> &str {
+        &self.column
+    }
+
+    /// Operator.
+    pub fn op(&self) -> Op {
+        self.op
+    }
+
+    /// Scalar value.
+    pub fn scalar(&self) -> &Scalar {
+        &self.scalar
+    }
+}
+
+impl std::fmt::Display for Expr {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}{}{}", self.column(), self.op(), self.scalar())
+    }
+}
+
+impl From<Expr> for datafusion::logical_plan::Expr {
+    fn from(expr: Expr) -> Self {
+        let column = datafusion::logical_plan::Column {
+            relation: None,
+            name: expr.column,
+        };
+
+        datafusion::logical_plan::Expr::BinaryExpr {
+            left: Box::new(datafusion::logical_plan::Expr::Column(column)),
+            op: expr.op.into(),
+            right: Box::new(datafusion::logical_plan::Expr::Literal(expr.scalar.into())),
+        }
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum ProtoToExprError {
+    #[snafu(display("cannot deserialize operator: {}", source))]
+    CannotDeserializeOperator { source: crate::expr::ProtoToOpError },
+
+    #[snafu(display("illegal operator enum value: {}", value))]
+    IllegalOperatorEnumValue { value: i32 },
+
+    #[snafu(display("missing scalar"))]
+    MissingScalar,
+
+    #[snafu(display("cannot deserialize scalar: {}", source))]
+    CannotDeserializeScalar {
+        source: crate::expr::ProtoToScalarError,
+    },
+}
+
+impl TryFrom<proto::Expr> for Expr {
+    type Error = ProtoToExprError;
+
+    fn try_from(expr: proto::Expr) -> Result<Self, Self::Error> {
+        let op = proto::Op::from_i32(expr.op)
+            .context(IllegalOperatorEnumValue { value: expr.op })?
+            .try_into()
+            .context(CannotDeserializeOperator)?;
+
+        let scalar = expr
+            .clone()
+            .scalar
+            .context(MissingScalar)?
+            .try_into()
+            .context(CannotDeserializeScalar)?;
+
+        Ok(Expr {
+            column: expr.column,
+            op,
+            scalar,
+        })
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum DataFusionToExprError {
+    #[snafu(display("unsupported expression: {:?}", expr))]
+    UnsupportedExpression {
+        expr: datafusion::logical_plan::Expr,
+    },
+
+    #[snafu(display("unsupported operants: left {:?}; right {:?}", left, right))]
+    UnsupportedOperants {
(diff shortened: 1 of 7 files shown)


alamb:
I plan to review this in the morning

crepererum:
A word on error handling here:

**Before:**
- `datafusion =(can fail)=> protobuf`
- `protobuf =(can fail)=> datafusion`

**After:**
- `datafusion =(can fail)=> IOx =(cannot fail)=> protobuf`
- `protobuf =(can fail)=> IOx =(cannot fail)=> datafusion`

So now once you have the new in-memory format, you can convert it to protobuf or datafusion w/o failing.

For a future PR (so that this change doesn't get to large), I'll change the delete predicate parsing from 

```text
SQL clause =(can fail)=> datafusion =(check for supported predicates)=> datafusion
```

to 

```text
SQL clause =(can fail)=> datafusion =(can fail)=> IOx
```

aka replace the check w/ a type conversion (following the "parse, don't validate/typecheck" paradigm).

e-dard:
Before looking through the code I have a question - what are the expected uses for the `Predicate` type? Limited to delete predicates only, or for all predicates that are within gRPC InfluxDB read requests?

predicate/src/expr.rs
```diff
@@ -0,0 +1,496 @@
...
+    UnsupportedScalarValue {
+        value: datafusion::scalar::ScalarValue,
+    },
+}
+
+impl TryFrom<datafusion::scalar::ScalarValue> for Scalar {
+    type Error = DataFusionToScalarError;
+
+    fn try_from(scalar: datafusion::scalar::ScalarValue) -> Result<Self, Self::Error> {
+        // see https://github.com/apache/arrow-datafusion/blob/195b69995db8044ce283d72fb78eb6b74b8842f5/datafusion/src/sql/planner.rs#L274-L295
```
alamb, line 316: it is not clear to me why this code reference is important (as there is a pretty clear mapping from `ScalarValue` --> `Scalar`)

predicate/src/expr.rs
```diff
@@ -0,0 +1,496 @@
...
+
+use generated_types::influxdata::iox::catalog::v1 as proto;
+use ordered_float::OrderedFloat;
+use snafu::{OptionExt, ResultExt, Snafu};
+
+/// Single expression to be used as parts of a predicate.
+///
+/// Only very simple expression of the type `<column> <op> <scalar>` are supported.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Expr {
```
alamb, line 14: Is it fair to day that this `Expr`  is effectively a native rust Expr type that mirrors the protobuf definition but is not generated code?

I will probably end up confused at some point that this `Expr` is not the same as a datafusion`Expr` but maybe that is ok.

predicate/src/expr.rs
```diff
@@ -0,0 +1,496 @@
...
+}
+
+#[derive(Debug, Snafu)]
+pub enum DataFusionToExprError {
+    #[snafu(display("unsupported expression: {:?}", expr))]
+    UnsupportedExpression {
+        expr: datafusion::logical_plan::Expr,
+    },
+
+    #[snafu(display("unsupported operants: left {:?}; right {:?}", left, right))]
```
alamb, line 111: I think the correct English term here is `operand` rather than `operant`

predicate/src/expr.rs
```diff
@@ -0,0 +1,496 @@
...
+}
+
+impl TryFrom<datafusion::logical_plan::Expr> for Expr {
+    type Error = DataFusionToExprError;
+
+    fn try_from(expr: datafusion::logical_plan::Expr) -> Result<Self, Self::Error> {
+        match expr {
+            datafusion::logical_plan::Expr::BinaryExpr { left, op, right } => {
+                let (column, scalar) = match (left.deref(), right.deref()) {
+                    // The delete predicate parser currently only supports `<column><op><value>`, not `<value><op><column>`,
```
alamb, line 135: is this still true? I would have expected that we support `<value><op><column>` as well

cc @NGA-TRAN

alamb (approved):
I am with @e-dard  -- it would help to understand the usecase for this feature / functionality -- I am sure there is a good reason but I am struggling to understand how deduplication mentioned in the PR description is related to the change 

However, the code does what the description says and I don't see anything incorrect about it, so approving 🚀

predicate/src/expr.rs
crepererum: I've asked that on Slack a while ago and @e-dard said we only support `<column><op><value>`.

crepererum:
> Before looking through the code I have a question - what are the expected uses for the Predicate type? Limited to delete predicates only, or for all predicates that are within gRPC InfluxDB read requests?

Uh, I forgot about the non-delete predicates. What functionality do we support there? If the gRPC InfluxDB read predicates are more complicated, I would:

1. still merge this PR here since it doesn't affect the in-mem predicates yet (it's just used for the from/to protobuf route that we curently only use to store delete predicates within the preserved catalog)
2. Separate that in-mem `Predicate` type for the two use cases in a follow up PR.

e-dard:
> Uh, I forgot about the non-delete predicates. What functionality do we support there? If the gRPC InfluxDB read predicates are more complicated, I would:

Exactly my line of thinking :-) 

We support a similar, but richer set of functionality. Here is a good protobuf definition that got copied over from the Influxdata storage repo to IOx.

https://github.com/influxdata/influxdb_iox/blob/main/generated_types/protos/influxdata/platform/storage/predicate.proto#L9-L10

Since the contract is typically between the storage layer and the query layer, we don't have public docs around the supported expressions.

alamb:
> Since the contract is typically between the storage layer and the query layer, we don't have public docs around the supported expressions.



I thought all this code was related to what expressions are serialized in the catalog (at the moment, only delete predicates). Not what expressions are supported in the read buffer.

I would suggest we leave the "what is supported by the read buffer" as an implementation detail that is between the query layer and RUB rather than conflating with delete predicates

e-dard:
@alamb by storage layer I'm talking about IOx. By query layer I'm talking about a gRPC query client, e.g., a Flux engine.

predicate/src/expr.rs
NGA-TRAN: Yes, this is correct except time range which is not stored in expr so I think it is fine here. 
Not for this PR but related: There is one caution in the future for `expression deduplication` is that we can only deduplicate `full delete predicate` that includes both time range and expressions. We cannot dedupe an expression or just the time range of a `delete  predicate` because it will lead to wrong result. I will go over all related open tickets and add examples for this.

NGA-TRAN (approved):
LGTM.
Just for extra caution as I mentioned in a detailed comment, the time range needs to keep outside the expression in other fields of the `Predicate` if we do not want to mess up with current implementation

predicate/src/expr.rs
crepererum: I've renamed it to `DeleteExpr` since it will also ONLY be used for the delete use case

predicate/src/expr.rs
crepererum: removed the comment since it was more a personal reference and will likely be outdated quickly

predicate/src/expr.rs
alamb: I think that sounds much clearer 👍  Thank you

predicate/src/expr.rs
alamb: The underlying machinery in the read buffer / TSM may only support `<column><op><value>` but my question was more like can the user write `col = 1` as well as `1 = col`? Typically an optimizer or some pass will canonicalize the expression (so users can provide either order, but internally they get rewritten to a uniform form)

PR #2624, opened 2021-09-23
test: query with selection predicate on deleted data

NGA-TRAN:
Delete test for select query with selection predicates 

### DIFF
--- a/query_tests/src/sql.rs
+++ b/query_tests/src/sql.rs
@@ -975,62 +975,63 @@ async fn sql_select_with_three_deletes_from_three_chunks() {
     .await;
 }
 
-// #[tokio::test]
-// async fn sql_select_with_three_deletes_from_three_chunks_with_select_predicate() {
-// Hmm: It seems a bug here. I am working on this next
-// BUG:
-// let expected = vec![
-//     "+-----+-----+--------------------------------+",
-//     "| bar | foo | time                           |",
-//     "+-----+-----+--------------------------------+",
-//     "| 7   | me  | 1970-01-01T00:00:00.000000080Z |",
-//     "+-----+-----+--------------------------------+",
-// ];
+#[tokio::test]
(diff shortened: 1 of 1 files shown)


query_tests/src/sql.rs
```diff
@@ -975,62 +975,63 @@ async fn sql_select_with_three_deletes_from_three_chunks() {
...
+        "| 3   | you | 1970-01-01T00:00:00.000000070Z |",
+        "| 4   | me  | 1970-01-01T00:00:00.000000050Z |",
+        "| 5   | me  | 1970-01-01T00:00:00.000000060Z |",
+        "| 7   | me  | 1970-01-01T00:00:00.000000080Z |",
+        "+-----+-----+--------------------------------+",
+    ];
+
+    run_sql_test_case(
+        scenarios::delete::ThreeDeleteThreeChunks {},
+        "SELECT * from cpu where bar != 1.0",
```
NGA-TRAN, line 995: If we do not explicitly specify the float number `1.0`, we will get wrong results and can be flaky, too.

alamb: I bet it is "flaky" in the sense that the same data loaded into different types of chunks (eg. RUB vs non RUB) will get different answers

alamb (approved):
Gotta love the `.0`

query_tests/src/sql.rs
NGA-TRAN: Actually this caused me lot of headache today because if I changed the expected results which did not related to the query results at all, I got different query results. It is very very very weird. I think something that makes a pointer  points to a wrong place. @e-dard may have some idea

NGA-TRAN: @alamb I reviewed and you are correct that we get different results because of different chunk types. Got to love the test scenarios 😄

PR #2625, opened 2021-09-23
fix: Retry S3 requests that fail with 5xx and remove streaming put API that wasn't used

### LINKED ISSUE #2483: Database stuck in error state due to transient error while reading S3
after restarting a router a database was stuck in an error state. The cause of the error appears to be due to a transient error while reading from the object store 

```
      {
        "error": {
          "message": "error finding active generation directory in object storage: AWS S3-based Object Store error: Unable to list data. Bucket: toolsus1.a.influxcloud.net-iox, Error: Request ID: None Body: \u003c?xml version=\"1.0\" encoding=\"UTF-8\"?\u003e\n\u003cError\u003e\u003cCode\u003eInternalError\u003c/Code\u003e\u003cMessage\u003eWe encountered an internal error. Please try again.\u003c/Message\u003e\u003cRequestId\u003eDM2BQRWCZMQPZ2E4\u003c/RequestId\u003e\u003cHostId\u003eHrNj5drKDiJ0tK1tRmMxLKOAtK5HhxMaOxvjYn9by/CflmEGEYt56LroHz7lNUD0OeFmntOAduw=\u003c/HostId\u003e\u003c/Error\u003e"
        },
        "state": "DATABASE_STATE_KNOWN"
      },
```

carols10cents:
Closes #2483 

The future is here! And they're evenly distributed, all over!

This borrows heavily from [the s3-algo crate's retry of rusoto requests](https://github.com/openanalytics/s3-algo/blob/464f7481b644ea8b5f21ea2d943002a8ee90db06/src/lib.rs#L133) and wraps all S3 requests in the `s3_request` function, that uses the [futures-retry crate](https://crates.io/crates/futures_retry) to clone the future and run it again if it fails and certain conditions are met.

There's a lot of minor rearranging of the API to make futures and closures that make futures and closures... I hope it'll make sense if you look commit-by-commit, but please ask if anything is confusing. 

The `ObjectStore` trait needed to change, so there's a tiny bit of changes in all the object store implementations, but this only retries S3 requests. I haven't had a chance to investigate the other object store crates' retry implementations (or lack thereof) yet.

Some questions I have:

- Currently I've [hardcoded the number of times a request is retried to 10](https://github.com/influxdata/influxdb_iox/pull/2625/files#diff-44002b1abe67569142529e6752131b4327abb349c51dfe8ba3d5f3f30035ab3bR489). Should this be configurable via a command-line option/env var, or is this fine for now?
- What logging would be useful? Every time a request is retried? The error that prompted the retry? How many times a request was retried? Something else? Nothing?

### DIFF
--- a/object_store/src/aws.rs
+++ b/object_store/src/aws.rs
@@ -10,14 +10,14 @@ use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use futures::{
     stream::{self, BoxStream},
-    Stream, StreamExt, TryStreamExt,
+    Future, Stream, StreamExt, TryStreamExt,
 };
+use futures_retry::{FutureRetry, RetryPolicy};
 use rusoto_core::ByteStream;
 use rusoto_credential::{InstanceMetadataProvider, StaticProvider};
 use rusoto_s3::S3;
 use snafu::{OptionExt, ResultExt, Snafu};
-use std::convert::TryFrom;
-use std::{fmt, io};
+use std::{convert::TryFrom, fmt, io, time::Duration};
 
 /// A specialized `Result` for object store-related errors
 pub type Result<T, E = Error> = std::result::Result<T, E>;
@@ -379,39 +379,52 @@ impl AmazonS3 {
         use ListState::*;
 
         let raw_prefix = prefix.map(|p| p.to_raw());
+        let bucket = self.bucket_name.clone();
 
         let request_factory = move || rusoto_s3::ListObjectsV2Request {
-            bucket: self.bucket_name.clone(),
+            bucket,
             prefix: raw_prefix.clone(),
             delimiter: delimiter.clone(),
             ..Default::default()
         };
 
+        let bucket = self.bucket_name.clone();
         Ok(stream::unfold(ListState::Start, move |state| {
-            let request_factory = request_factory.clone();
+            let (bucket, request_factory) = (bucket.clone(), request_factory.clone());
+            let s3 = self.client.clone();
 
             async move {
-                let mut list_request = request_factory();
-
-                match state.clone() {
-                    HasMore(continuation_token) => {
-                        list_request.continuation_token = Some(continuation_token);
-                    }
+                let continuation_token = match state.clone() {
+                    HasMore(continuation_token) => Some(continuation_token),
                     Done => {
                         return None;
                     }
                     // If this is the first request we've made, we don't need to make any
                     // modifications to the request
-                    Start => {}
-                }
+                    Start => None,
+                };
 
-                let resp =
-                    self.client
-                        .list_objects_v2(list_request)
-                        .await
-                        .context(UnableToListData {
-                            bucket: &self.bucket_name,
-                        });
+                let resp = s3_request(move || {
+                    let (s3, bucket, request_factory, continuation_token) = (
+                        s3.clone(),
+                        bucket.clone(),
+                        request_factory.clone(),
+                        continuation_token.clone(),
+                    );
+
+                    async move {
+                        Ok(async move {
+                            s3.list_objects_v2(rusoto_s3::ListObjectsV2Request {
+                                continuation_token,
+                                ..request_factory()
+                            })
+                            .await
+                            .context(UnableToListData { bucket })
+                        })
+                    }
+                })
+                .await;
+
                 let resp = match resp {
                     Ok(resp) => resp,
                     Err(e) => return Some((Err(e), state)),
@@ -433,6 +446,45 @@ impl AmazonS3 {
         })
         .boxed())
     }
+}
+
+async fn s3_request<F, G, H>(future_factory: F) -> Result<rusoto_s3::ListObjectsV2Output>
+where
+    F: Fn() -> G + Unpin + Clone + Send + Sync + 'static,
+    G: Future<Output = Result<H, Error>> + Send,
+    H: Future<Output = Result<rusoto_s3::ListObjectsV2Output>> + Send,
+{
+    let mut attempts = 0;
+    // TODO: configurable
+    let n_retries = 10;
+    // TODO: let backoff =
+
+    FutureRetry::new(
+        move || {
+            let future_factory = future_factory.clone();
+
+            async move {
+                let request = future_factory().await?;
+
+                request.await
+            }
+        },
+        // retry
+        {
+            move |e| {
+                attempts += 1;
+                if attempts > n_retries {
+                    RetryPolicy::ForwardError(e)
+                } else {
+                    RetryPolicy::WaitRetry(Duration::from_millis(200))
+                }
+            }
+        },
+    )
+    .await
+    // TODO: log number of attempts?
+    .map(|(response, _attempts)| response)
+    .map_err(|(err, _attempts)| err)
 }
 
 impl Error {
--- a/object_store/src/aws.rs
+++ b/object_store/src/aws.rs
@@ -225,7 +225,7 @@ impl ObjectStoreApi for AmazonS3 {
         prefix: Option<&'a Self::Path>,
     ) -> Result<BoxStream<'a, Result<Vec<Self::Path>>>> {
         Ok(self
-            .list_objects_v2(prefix)
+            .list_objects_v2(prefix, None)
             .await?
             .map_ok(|list_objects_v2_result| {
                 let contents = list_objects_v2_result.contents.unwrap_or_default();
@@ -241,7 +241,61 @@ impl ObjectStoreApi for AmazonS3 {
     }
 
     async fn list_with_delimiter(&self, prefix: &Self::Path) -> Result<ListResult<Self::Path>> {
-        self.list_with_delimiter_and_token(prefix, &None).await
+        Ok(self
+            .list_objects_v2(Some(prefix), Some(DELIMITER.to_string()))
+            .await?
+            .try_fold(
+                ListResult {
+                    next_token: None,
+                    common_prefixes: vec![],
+                    objects: vec![],
+                },
+                |acc, list_objects_v2_result| async move {
+                    let mut res = acc;
+                    let contents = list_objects_v2_result.contents.unwrap_or_default();
+                    let mut objects = contents
+                        .into_iter()
+                        .map(|object| {
+                            let location = CloudPath::raw(
+                                object.key.expect("object doesn't exist without a key"),
+                            );
+                            let last_modified = match object.last_modified {
+                                Some(lm) => DateTime::parse_from_rfc3339(&lm)
+                                    .context(UnableToParseLastModified {
+                                        bucket: &self.bucket_name,
+                                    })?
+                                    .with_timezone(&Utc),
+                                None => Utc::now(),
+                            };
+                            let size = usize::try_from(object.size.unwrap_or(0))
+                                .expect("unsupported size on this platform");
+
+                            Ok(ObjectMeta {
+                                location,
+                                last_modified,
+                                size,
+                            })
+                        })
+                        .collect::<Result<Vec<_>>>()?;
+
+                    res.objects.append(&mut objects);
+
+                    res.common_prefixes.extend(
+                        list_objects_v2_result
+                            .common_prefixes
+                            .unwrap_or_default()
+                            .into_iter()
+                            .map(|p| {
+                                CloudPath::raw(
+                                    p.prefix.expect("can't have a prefix without a value"),
+                                )
+                            }),
+                    );
+
+                    Ok(res)
+                },
+            )
+            .await?)
     }
 }
 
@@ -314,6 +368,7 @@ impl AmazonS3 {
     async fn list_objects_v2(
         &self,
         prefix: Option<&CloudPath>,
+        delimiter: Option<String>,
     ) -> Result<BoxStream<'_, Result<rusoto_s3::ListObjectsV2Output>>> {
         #[derive(Clone)]
         enum ListState {
@@ -328,6 +383,7 @@ impl AmazonS3 {
         let request_factory = move || rusoto_s3::ListObjectsV2Request {
             bucket: self.bucket_name.clone(),
             prefix: raw_prefix.clone(),
+            delimiter: delimiter.clone(),
             ..Default::default()
         };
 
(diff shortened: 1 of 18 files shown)


tustvold (changes requested):
I think we can avoid bringing in a new dependency, which I think would be worthwhile doing.

Perhaps it is a something for a separate PR but I wonder if we might simply improve the utility of the errors returned by the object store API, so that it is easier for the callers to determine the class of error encountered and act on it. 

In many cases the plumbing for retries already exists, e.g. the Server and Database state machines, it just doesn't have a way to introspect the error in a meaningful way and so assumes the failure is fatal.

Currently it isn't even possible from the errors to determine if the error was that the object doesn't exist! Retries aside, I think having actionable errors returned would be good, as currently they're more like stack traces... This in turn would allow callers to retry based on the policy that makes sense for them. In almost all cases I can think of non-permanent errors should just be retried indefinitely, either directly or as part of a pre-existing retry loop.

I dunno, I missed the conversations that led to this PR so feel free to ignore me, but it seems a little strange to me that retries would be implemented at this level, where there is very little context as to the operations being performed

object_store/src/aws.rs
```diff
@@ -140,152 +140,191 @@ impl ObjectStoreApi for AmazonS3 {
...
-            .client
-            .get_object(get_request)
-            .await
-            .context(UnableToGetData {
-                bucket: self.bucket_name.to_owned(),
-                location: key.clone(),
-            })?
+        let s3 = self.client.clone();
+
+        let resp = s3_request(move || {
```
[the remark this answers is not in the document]
alamb: As in it will retry the `get` from the beginning, or will simply not retry the `get` at all?

FWIW I also agree that we can postpone retrying the `get` to some future PR if/when we have evidence that it is needed

object_store/src/aws.rs
```diff
@@ -355,78 +394,134 @@ pub(crate) fn new_failing_s3() -> Result<AmazonS3> {
...
+where
+    F: Fn() -> G + Unpin + Clone + Send + Sync + 'static,
+    G: Future<Output = Result<H, rusoto_core::RusotoError<E>>> + Send,
+    H: Future<Output = Result<R, rusoto_core::RusotoError<E>>> + Send,
+{
+    let mut attempts = 0;
+    // TODO: make the number of retries configurable
+    let n_retries = 10;
+
+    FutureRetry::new(
```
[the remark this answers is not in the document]
alamb: I am a big fan of reducing dependencies (we have over 500 already). Trading a little duplication for fewer dependencies seems like an excellent tradeoff to me

object_store/src/aws.rs
```diff
@@ -355,78 +394,134 @@ pub(crate) fn new_failing_s3() -> Result<AmazonS3> {
...
+                if attempts > n_retries || !should_retry {
+                    RetryPolicy::ForwardError(e)
+                } else {
+                    RetryPolicy::WaitRetry(Duration::from_millis(2u64.pow(attempts) * 50))
+                }
+            }
+        },
+    )
+    .await
+    // TODO: log number of attempts?
```
alamb, line 520: Recommend logging at `debug!` level the number of attempts and at `info!` level on an actual failure

object_store/src/aws.rs
```diff
@@ -355,78 +394,134 @@ pub(crate) fn new_failing_s3() -> Result<AmazonS3> {
...
 }
 
+async fn s3_request<E, F, G, H, R>(future_factory: F) -> Result<R, rusoto_core::RusotoError<E>>
+where
+    F: Fn() -> G + Unpin + Clone + Send + Sync + 'static,
+    G: Future<Output = Result<H, rusoto_core::RusotoError<E>>> + Send,
+    H: Future<Output = Result<R, rusoto_core::RusotoError<E>>> + Send,
+{
+    let mut attempts = 0;
+    // TODO: make the number of retries configurable
```
alamb, line 488: I recommend making it a crate level `pub const` so it is clear retries are occurring and the number can be changed, but no need to make it more configurable than that

alamb (approved):
When thinking about this PR it seems to me that combination of retries + streaming makes for the awkward API of a function that returns a stream.

I wonder if all the streaming complexity for PUT is really worth it (especially since we end up writing to tempfiles anyways). Could we make this whole thing simpler / easier to maintain by removing the API that implies we are doing streaming PUTs, make it clear we are buffering, and then just use the tempfile?



> Currently I've hardcoded the number of times a request is retried to 10. Should this be configurable via a command-line option/env var, or is this fine for now?

I think a hard coded constant is fine for now -- we can make it configurable if/when we have a need to change it. I am not sure it will need much tuning: either the request is likely to pass after the first or second retry or else it is going to fail pretty much non stop

> What logging would be useful? 

I would suggest: 
* `debug!` for: Every time a request is retried, including the error that prompted the retry
* `warn!` if the operation failed, including the number of times it was retried and the error that prompted the most recent retry

alamb:
> I dunno, I missed the conversations that led to this PR so feel free to ignore me, but it seems a little strange to me that retries would be implemented at this level, where there is very little context as to the operations being performed

@tustvold  what level would you recommend retries to be performed? If we did it at each callsite in IOx we would both duplicate a lot of code, as well as likely end up missing adding retries to some places that should have them.

The usecase for retries is "Amazon gives some sort of 500 internal error and retrying the request will succeed for some opaque and unknowable reason"

object_store/src/lib.rs
```diff
@@ -73,13 +73,14 @@ pub trait ObjectStoreApi: Send + Sync + 'static {
     fn new_path(&self) -> Self::Path;
 
     /// Save the provided bytes to the specified location.
-    async fn put<S>(
+    async fn put<F, S>(
```
alamb, line 76: it would be nice to make this docstring contain a bit more information about what `S` and `F` are, for people who are trying to use this crate and are not (yet) `futures` and rust trait experts yet

I think especially helpful would be the context that `bytes` is a function due to retries -- and that it will be called once for each attempt

tustvold:
> @tustvold what level would you recommend retries to be performed? If we did it at each callsite in IOx we would both duplicate a lot of code, as well as likely end up missing adding retries to some places that should have them.

I think we come at the same issue from two sides. The current object store API, with its asynchronous streams, request oriented design, etc... seems to me to be more of a low-level abstraction of the object store APIs as they actually are. When I call `put` request I expect it to do that - send a put request. In my mind this is a very good thing as the semantics are easily mappable onto the object store's API.

If I instead want to `WriteFile` or `WriteBytes` to object store, I would expect that to be a separate struct that would use the object store API to achieve this, with any valid retries, etc... This would avoid code duplication, whilst preserving the low level object store API for situations that need/want it. _It would also avoid the contextless retries that can be fertile ground for retry storms and similarly self-inflicted issues._

But perhaps we don't wish to keep this separation, in which case perhaps we should alter the API to make it more amenable to retries as you suggest. I don't feel particularly strongly about it, but I do feel that this PR as currently written feels a bit like there is an impedance mismatch between the API and the desired semantics

_TBC my "request changes" concerned the dependency, not necessarily the retries themselves_

carols10cents:
Ok I am seeing a number of suggestions that are smaller to do that I'm going to do as part of this PR:

- [ ] Avoid bringing in a new dependency by switching futures_retry for a `while` loop
- [ ] Revert the changes to retry `get` requests
- [ ] Change `put` requests to not do fake streaming since they're not doing real streaming anyway
- [ ] Make the `aws` method `list_objects_v2` `pub`
- [ ] Make the number of retries `pub const`
- [ ] Add logging
- [ ] Add documentation

Bigger changes that sound like they need more discussion:

- [ ] Expose more information about the semantics of the errors returned 
- [ ] Implement retry logic at a higher level

Does that sound right?

carols10cents:
Actually, I think `put` **is** doing streaming [if the length of the stream is specified](https://github.com/influxdata/influxdb_iox/blob/8e89dde85cab4d862bca0650b32b45e9e8c99ac2/object_store/src/aws.rs#L148), which it is in every non-test code scenario I can find:

- [catalog transaction files](https://github.com/influxdata/influxdb_iox/blob/8e89dde85cab4d862bca0650b32b45e9e8c99ac2/parquet_file/src/catalog/internals/proto_io.rs#L45)
- [parquet files](https://github.com/influxdata/influxdb_iox/blob/8e89dde85cab4d862bca0650b32b45e9e8c99ac2/parquet_file/src/storage.rs#L221)
- [database rules](https://github.com/influxdata/influxdb_iox/blob/8e89dde85cab4d862bca0650b32b45e9e8c99ac2/iox_object_store/src/lib.rs#L591)

alamb:
> The current object store API, with its asynchronous streams, request oriented design, etc... seems to me to be more of a low-level abstraction of the object store APIs as they actually are. When I call put request I expect it to do that - send a put request. In my mind this is a very good thing as the semantics are easily mappable onto the object store's API.

@tustvold  -- I agree with this assessment.

However, when I use an object store  and say `put` I expect it to do what i ask or else return an error of either
1.  I did something wrong
2. there is something fundamentally wrong with the service 

Transient 500 errors from AWS that will succeed if you rerun them do not fall into this category. I don't think handling such errors should *not* be done in IOx logic itself which is complicated enough without having to account for transient object store failures. It would be ideal if AWS did this itself but as nonsense such as https://github.com/influxdata/influxdb_iox/issues/2483 illustrates, they put it back to their clients to deal with

Are you primarily concerned with the extra `Fn()` stuff that is added to the APIs?

tustvold:
> Actually, I think put is doing streaming if the length of the stream is specified, which it is in every non-test code scenario I can find:

In all those examples though we have already hydrated the entire payload into a cheaply-cloneable Bytes array? I.e. the streaming is not necessary. TBC I like the streaming APIs, and they will be very useful for avoiding hydrating entire parquet files into memory, etc... and this was why I suggested what I did, but perhaps I've just muddied the water and we can add that later.

> Make the aws method list_objects_v2 pub

I meant more expose the semantics as the public API across all object stores, definitely something for a future PR though

> Bigger changes that sound like they need more discussion that I can make new issues for but won't be addressing for this PR:

I think this is sensible

> Transient 500 errors from AWS that will succeed if you rerun them do not fall into this category. I don't think handling such errors should not be done in IOx logic itself which is complicated enough without having to account for transient object store failures

I agree with this entirely, as written we handle these failures in both the AWS client and then also have to handle them in IOx as we can't adequately express the context of how long to retry for. If the get request corresponds to loading a parquet file for query, we want to try a couple of times and then bail out with an error to the client, if instead we are loading the database rules for a database we don't want to bail out ever, just keep retrying until we succeed

alamb:
@tustvold and I had a nice face to face  chat. I think the consensus is that the extra complexity required to support an streaming API and retries strikes us both as "not good" but the various ways we can think to improve the situation all require substantial effort that is hard to justify at this time

Given how important the retry feature is to IOx's resilience, our suggestion is to:
1. Go with the approach in this PR
2. Change the number of retries from 10 to 3 (to minimize request amplification)
3. Put a note in the backlog (maybe an issue, maybe not) that some "higher level" APIs ('put file' or 'put from memory') might make the crate easier to work with than the lower level 'put data from a stream that comes from a generator function'.

FWI @carols10cents

carols10cents:
> the various ways we can think to improve the situation all require substantial effort that is hard to justify at this time

> In all those examples though we have already hydrated the entire payload into a cheaply-cloneable Bytes array?

What if `put` takes a cheaply-cloneable `Bytes` array?

object_store/src/aws.rs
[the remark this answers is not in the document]
carols10cents: Filed https://github.com/influxdata/influxdb_iox/issues/2642 for GET requests.

carols10cents:
Ok, I think this is ready for rereview. I also ran the integration tests with AWS, Google, and Azure locally and all passed ✅

carols10cents:
And I've filed https://github.com/influxdata/influxdb_iox/issues/2643 for tracking enabling retrying at a higher abstraction level.

object_store/src/lib.rs
```diff
@@ -73,14 +74,7 @@ pub trait ObjectStoreApi: Send + Sync + 'static {
...
     /// Save the provided bytes to the specified location.
-    async fn put<S>(
-        &self,
-        location: &Self::Path,
-        bytes: S,
-        length: Option<usize>,
-    ) -> Result<(), Self::Error>
-    where
-        S: Stream<Item = io::Result<Bytes>> + Send + Sync + 'static;
+    async fn put(&self, location: &Self::Path, bytes: Bytes) -> Result<(), Self::Error>;
```
alamb, line 77: So the idea here is that since we currently have no actual streaming write usecase (aka all writes are either from a file or an in memory buffer we can simplify the API and just take `Bytes`)? If so I like it 👍

server/src/db.rs
```diff
@@ -4087,15 +4087,8 @@ mod tests {
...
-        let data = Bytes::default();
-        let len = data.len();
-
         iox_object_store
-            .put_parquet_file(
-                path,
-                futures::stream::once(async move { Ok(data) }),
-                Some(len),
-            )
+            .put_parquet_file(path, Bytes::new())
```
alamb, line 4091: it is so much nicer to read ❤️

object_store/src/gcp.rs
```diff
@@ -86,27 +86,7 @@ impl ObjectStoreApi for GoogleCloudStorage {
         CloudPath::default()
     }
 
-    async fn put<S>(&self, location: &Self::Path, bytes: S, length: Option<usize>) -> Result<()>
-    where
-        S: Stream<Item = io::Result<Bytes>> + Send + Sync + 'static,
-    {
-        let temporary_non_streaming = bytes
```
alamb, line 93: FYI @mkmik  -- the interface is simplified and now we don't need to buffer locally as well

object_store/src/aws.rs
```diff
@@ -10,18 +9,21 @@ use bytes::Bytes;
...
 use snafu::{OptionExt, ResultExt, Snafu};
-use std::convert::TryFrom;
-use std::{fmt, io};
+use std::{convert::TryFrom, fmt, time::Duration};
 
 /// A specialized `Result` for object store-related errors
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// The maximum number of times a request will be retried in the case of an AWS server error
+pub const MAX_NUM_RETRIES: u32 = 3;
```
alamb, line 25: 👍

object_store/src/aws.rs
```diff
@@ -355,75 +378,142 @@ pub(crate) fn new_failing_s3() -> Result<AmazonS3> {
...
+
+/// Handles retrying a request to S3 up to `MAX_NUM_RETRIES` times if S3 returns 5xx server errors.
+///
+/// The `future_factory` argument is a function `F` that takes no arguments and, when called, will
+/// return a `Future` (type `G`) that, when `await`ed, will perform a request to S3 through
+/// `rusoto` and return a `Result` that returns some type `R` on success and some
+/// `rusoto_core::RusotoError<E>` on error.
+///
+/// If the executed `Future` returns success, this function will return that success.
+/// If the executed `Future` returns a 5xx server error, this function will wait an amount of
```
alamb, line 471: 👍

object_store/src/aws.rs
```diff
@@ -355,75 +378,142 @@ pub(crate) fn new_failing_s3() -> Result<AmazonS3> {
...
+                    warn!(
+                        ?error,
+                        attempts, "maximum number of retries exceeded for AWS S3 request"
+                    );
+                    return Err(error);
+                } else if !should_retry {
+                    return Err(error);
+                } else {
+                    debug!(?error, attempts, "retrying AWS S3 request");
+                    let wait_time = Duration::from_millis(2u64.pow(attempts) * 50);
```
alamb, line 512: the retry times appear to be `50ms`, `100ms` and `200ms` which seems reasonable for the first try. Making it 100 (for `100ms`, `200ms` and `400ms`) might be ok too but that is just a hunch. 

I suggest making the `50` a module level constant as well

alamb (approved):
Looks great to me -- thank you @carols10cents  -- all of my original comments were addressed (very well!). It probably makes sense to give @tustvold  a chance to review tomorrow if he would like to

object_store/src/lib.rs
carols10cents: You got it dude!

alamb: full-house-michelle-tanner

🤣

alamb:
🎉

PR #2628, opened 2021-09-24
refactor: move delete predicate persistence into background job

crepererum:


NGA-TRAN (commented):
I defer the approval to someone else because I do not know the locking mechanism well. 
@crepererum : when you comeback from vacation, if you have time, can you go over this code with me? I would like to learn the rust locking mechanism and async in this PR.

NGA-TRAN (approved):
Thanks @crepererum for walking through the code with me. This prototype for you to test and prepare permanent work looks good. I guess when you turn it to permanent code, you will convert hard code values into constants or even configuration parameters but we will know when we reach there.

crepererum:
That needs a rebase, will take care of that later.

Issue #1603, opened 2021-06-01
Incorrect results with integer predicate on floating point column


alamb:
__Steps to reproduce:__
Note this test case is on the [alamb/read_with_predicate_failure](https://github.com/influxdata/influxdb_iox/tree/alamb/read_with_predicate_failure)  branch

Load Input line protocol (first two rows in 1 chunk, last row in a different RB chunk)
```
cpu bar=1 10
cpu bar=2 20
cpu bar=3 10
```

__Expected behaviour:__

```sql
select * from cpu where bar < 3
```
IOx returns the correct answer
```
"+-----+-------------------------------+",
"| bar | time                          |",
"+-----+-------------------------------+",
"| 1   | 1970-01-01 00:00:00.000000010 |",
"| 2   | 1970-01-01 00:00:00.000000020 |",
"+-----+-------------------------------+",
```

__Actual behaviour:__

IOx returns the wrong answer (no rows):
```
recom
"++",
"++",
```

__Note__

Note that when the constant is expressed as `3.0` the query results are correct:

```
select * from cpu where bar < 3.0
```

```
"+-----+-------------------------------+",
"| bar | time                          |",
"+-----+-------------------------------+",
"| 1   | 1970-01-01 00:00:00.000000010 |",
"| 2   | 1970-01-01 00:00:00.000000020 |",
"+-----+-------------------------------+",
```

alamb:
@e-dard  notes that this may be due to trying to compare floats / integers directly in the read buffer  --  no values pass the predicate because currently floats and integers are not comparable in the read buffer

some options are:
1. add support for that type of comparison in the read buffer 
2. validate expressions against a schema at a higher level and not push down expressions that aren’t going to work.

e-dard:
I have some thoughts about how I want to add support for comparing floats to integers in the Read Buffer but I don't think I will be able to get to that until next week. If you want to fix this in the mean time then one option would be to check predicates against Chunk schemas and only push down expressions that have exactly the same type as the column.

alamb:
Thanks @e-dard  -- I likely also won't get to this this week. Given https://github.com/influxdata/influxdb_iox/pull/1612 this may not as high a priority (generating a unsupported error is much less bad in my opinion than generating wrong answers)

alamb:
I am pretty sure #2631  doesn't fix this issue and github got over zealous reading the comment

PR #2631, opened 2021-09-24
feat: validate Read Buffer predicates

e-dard:
Related to #1603.

This PR adds some validation checks on the expressions within predicates supplied to the Read Buffer. This ensures that invalid predicates cannot cause the Read Buffer to panic, but instead result in an error that is propogated back to the query engine.

Currently, invalid predicates are those where making a comparison between the values and the column and the provided literal are not obvious, e.g., trying to apply the expression `region > 33` to a `region` column that has a string datatype.

This PR does not resolve https://github.com/influxdata/influxdb_iox/issues/1603 - in fact a predicate `temp = 22.2` for a `temp` column with a physical type `u64` is considered valid within the logic of this PR.

A follow on PR will teach the Read Buffer to coalesce compatible types into the column type automatically, which will ensure expressions like `temp > 100.3` will work when the `temp` column is `i64` or `u64`.

### DIFF
--- a/read_buffer/src/table.rs
+++ b/read_buffer/src/table.rs
@@ -1,11 +1,12 @@
 use crate::{
     column,
-    row_group::{self, ColumnName, Predicate, RowGroup},
+    row_group::{self, ColumnName, Literal, Predicate, RowGroup},
     schema::{AggregateType, ColumnType, LogicalDataType, ResultSchema},
     value::{OwnedValue, Scalar, Value},
 };
 use arrow::record_batch::RecordBatch;
 use data_types::{chunk_metadata::ChunkColumnSummary, partition_metadata::TableSummary};
+use datafusion::physical_plan::expressions::BinaryExpr;
 use internal_types::selection::Selection;
 use parking_lot::RwLock;
 use snafu::{ensure, Snafu};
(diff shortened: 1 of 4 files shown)


read_buffer/src/table.rs
```diff
@@ -684,6 +704,46 @@ impl MetaData {
         self.column_names.iter().map(|name| name.as_str()).collect()
     }
 
+    /// Determine, based on the table meta data, whether all of the provided
+    /// expressions can be applied, returning an error if any can't.
+    pub fn validate_exprs<'a>(&self, iter: impl IntoIterator<Item = &'a BinaryExpr>) -> Result<()> {
+        iter.into_iter()
+            .try_for_each(|expr| match self.columns.get(expr.column()) {
+                Some(col_meta) => match (col_meta.logical_data_type, expr.literal()) {
+                    (LogicalDataType::Integer, Literal::Integer(_))
```
e-dard, line 713: The set of types the Read Buffer considers comparable in terms of applying a `column op literal` expression.

NGA-TRAN (approved):
🎆 The error messages will save us time to figure out what is going 🍾

PR #2634, opened 2021-09-25
refactor: store chunks in iteration order

tustvold:
Follow on from https://github.com/influxdata/influxdb_iox/pull/2480#discussion_r708039190 that stores chunks within a partition in the order that they are iterated.

This is because iteration is significantly more common than lookups, and so it makes sense to optimise for this case. By storing in the iteration order we both avoid unnecessary sorts, and also eliminate the need for intermediate buffering


alamb (approved):
Another benefit is that this PR avoids the need to remember to sort the chunks correctly

Looks like a (very) nice clean up to me ❤️

PR #2640, opened 2021-09-27
fix: perf build cache

tustvold:
Caching and restoring "/target" will just cause it to grow indefinitely over time, this changes the perf CI to use the same caching mechanism as all the other jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,20 +201,11 @@ jobs:
           name: Install perf.py dependencies
           command: python3 -m pip install -r perf/requirements.txt
 
-      - restore_cache:
-          keys:
-            - cargo-lock-{{ checksum "Cargo.lock" }}
-      - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -201,20 +201,11 @@ jobs:
           name: Install perf.py dependencies
           command: python3 -m pip install -r perf/requirements.txt
 
-      - restore_cache:
-          keys:
-            - cargo-lock-{{ checksum "Cargo.lock" }}
-      - run:
-          name: Prime Rust build cache
-          command: cargo build --package influxdb_iox --bin influxdb_iox --package iox_data_generator --bin iox_data_generator
```
tustvold, line 209: This doesn't actually work because it builds with different options when invoked from perf

crepererum (approved):

PR #2644, opened 2021-09-27
refactor: delete tests

NGA-TRAN:
Refactoring delete tests to have it easier to add a more complicated ones when we finishing up tasks in #2396 

### DIFF
--- a/query_tests/src/scenarios/delete.rs
+++ b/query_tests/src/scenarios/delete.rs
@@ -5,6 +5,9 @@ use datafusion::logical_plan::{col, lit};
 use predicate::predicate::{Predicate, PredicateBuilder};
 
 use async_trait::async_trait;
+use query::QueryChunk;
+use server::Db;
+//use core::slice::SlicePattern;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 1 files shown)


query_tests/src/scenarios/delete.rs
```diff
@@ -618,1096 +930,291 @@ impl DbSetup for ThreeDeleteThreeChunks {
 }
 
 // -----------------------------------------------------------------------------
-// Helper functions
```
NGA-TRAN, line 621: Replace all helper functions with only 2 helper functions and from now on, when we add more delete tests, we do not need to add more helpers like we had to do before this refactoring.

query_tests/src/scenarios/delete.rs
```diff
@@ -38,28 +38,57 @@ impl DbSetup for DeleteFromMubOneMeasurementOneChunk {
             .add_expr(expr)
             .build();
 
-        // delete happens when data in MUB
-        let scenario_mub = make_delete_mub(lp_lines.clone(), pred.clone()).await;
```
NGA-TRAN, line 42: All the changes before the line "Helper functions" are just for using 2 new helper functions  which are a lot more general and easier to use, except there are a few changes in variable names and comments  to make more sense.

query_tests/src/scenarios/delete.rs
```diff
@@ -618,1096 +930,291 @@ impl DbSetup for ThreeDeleteThreeChunks {
...
-    assert_eq!(count_read_buffer_chunks(&db), 1);
-    assert_eq!(count_object_store_chunks(&db), 0);
-
-    DbScenario {
-        scenario_name: "Deleted data in RUB moved from MUB".into(),
-        db,
-    }
+#[derive(Debug, Clone)]
+pub enum ChunkType {
+    Mubo, // open MUB
```
alamb, line 943: Would it be possible to make these doc comments (so when I hover over them in my editor the definition of Mubo appears. Something like

```suggestion
    /// Open MUB
    Mubo, 
```

Also, I felt Mubo sounded like a perfect name for a startup and turns out there is already one: https://www.mubo-app.com/ 😆

query_tests/src/scenarios/delete.rs
```diff
@@ -618,1096 +930,291 @@ impl DbSetup for ThreeDeleteThreeChunks {
...
-    assert_eq!(count_mutable_buffer_chunks(&db), 0);
-    assert_eq!(count_read_buffer_chunks(&db), 1);
-    assert_eq!(count_object_store_chunks(&db), 0);
-
-    DbScenario {
-        scenario_name: "Deleted data in RUB moved from MUB".into(),
-        db,
-    }
+#[derive(Debug, Clone)]
+pub enum ChunkType {
```
alamb, line 942: Using this enum is likely useful to reduce repetition in other query_tests, but perhaps that refactoring can wait for another day

query_tests/src/scenarios/delete.rs
```diff
@@ -618,1096 +930,291 @@ impl DbSetup for ThreeDeleteThreeChunks {
...
-    }
+    /// delete predicate applies to chunks at their Os stage
+    Os,
 }
 
-async fn make_delete_rub_to_os_delete(
+/// Build a chunk that may move with life cycle before/after deletes
+///  Note the chunk in this function can be moved to different stages and delete predicates
+///  can be applied at different stages when the chunk is moved.
+async fn make_chunk_with_deletes_at_different_stages(
```
alamb, line 993: This is a very cool function

query_tests/src/scenarios/delete.rs
```diff
@@ -87,20 +116,41 @@ impl DbSetup for DeleteFromRubOneMeasurementOneChunk {
             .build();
 
         // delete happens to data in RUB
-        let scenario_rub =
-            make_delete_rub(lp_lines.clone(), pred.clone(), table_name, partition_key).await;
+        let preds = vec![Pred {
+            predicate: &pred,
+            delete_time: DeleteTime::Rub,
+        }];
+        let scenario_rub = make_chunk_with_deletes_at_different_stages(
```
alamb, line 123: You could almost reduce this even more if you implemented a `ChunkType::all()` function that returned an interator over all possible ChunkTypes.

Then you might be able to write something like (untested)

```rust
let mut scenarios = vec![];
for chunk_type in ChunkType::all() {
  scenarios.append(
    make_chunk_with_deletes_at_different_stages(
            lp_lines.clone(),
            ChunkType::RubOs,
            preds.clone(),
            table_name,
            partition_key,
        )
    )
}
scenarios
```

Also, given that the tests seem to be trying all combinations of `DeleteTime` and `ChunkType` you maybe could collapse the creation into two nested loops.

query_tests/src/scenarios/delete.rs
```diff
@@ -342,36 +484,62 @@ impl DbSetup for TwoDeleteMultiExprsFromMubOneMeasurementOneChunk {
             .build();
 
         // delete happens when data in MUB
```
alamb, line 486:
```suggestion
        // delete happens when data in MUB and at end
```

alamb (approved):
Screen Shot 2021-09-27 at 3 37 18 PM 👍 

This PR looks really great to me. Nice work @NGA-TRAN

query_tests/src/scenarios/delete.rs
NGA-TRAN: This is a good idea. I will explore this later (maybe after a few other higher priority PRs)

NGA-TRAN: But I did refactor it a bit in a different way

query_tests/src/scenarios/delete.rs
```diff
@@ -38,28 +46,13 @@ impl DbSetup for DeleteFromMubOneMeasurementOneChunk {
...
-        .await;
-
-        // delete happens when data in MUB then moved to RUB, then persisted, and then RUB is unloaded
-        let scenario_os =
-            make_delete_mub_to_os(lp_lines.clone(), pred, table_name, partition_key).await;
-
-        // return scenarios to run queries
-        vec![scenario_mub, scenario_rub, scenario_rub_os, scenario_os]
+        // delete happens when data in open MUB
+        delete_from_mub(lp_lines, preds, table_name, partition_key).await
```
NGA-TRAN, line 55: @alamb : `delete_from_mub`, `delete_from_rub`, `delete_from_rub_os`, and `delete_from_os` are 4 new functions I added in the second round refactor

query_tests/src/scenarios/delete.rs
```diff
@@ -282,24 +242,22 @@ impl DbSetup for DeleteMultiExprsFromOsOneMeasurementOneChunk {
...
-        // delete happens after data is persisted and RUB is unloaded
-        let scenario_os = make_delete_os(lp_lines.clone(), pred, table_name, partition_key).await;
-
-        // return scenarios to run queries
-        vec![scenario_rub_os, scenario_rub_os_unload_rub, scenario_os]
+        let preds = vec![Pred {
+            predicate: &pred,
+            delete_time: DeleteTime::RubOs,
+        }];
+        let mut scenarios =
```
alamb, line 249: this is much nicer 👍

query_tests/src/scenarios/delete.rs
NGA-TRAN: @alamb I have thought about this and my coming tests for 2  edge cases last night, I have realized that even though, after this PR, I no longer have to write 8 long functions for each test to cover all scenarios, I still need to write 3 short functions. With this further refactor, I will only need to write one function. I will spend today to work on this.
The more I think about delete and its work ahead, the sooner we need this to have better test coverage and save time to write them. And as you said, it will pay a path for us to refactor the whole query_tests crate

PR #2623, opened 2021-09-23
fix: Revert "feat: remove routerify (#2586)"

alamb:
This reverts commit 9fba7b0ba0c34ab4b2b5c865636fccf711b0b0b1 / #2586 as we have started seeing errors on our internal IOx deployment traced to this PR (see comments on linked ticket)

Created via `git revert 9fba7b0` and then fixing some conflicts

fixes https://github.com/influxdata/conductor/issues/546

PR #2649, opened 2021-09-28
revert: "fix: Revert "feat: remove routerify (#2586)" (#2623)"

### LINKED ISSUE #2630: refactor: Remove routerify from codebase
__Use case:__
We no longer user any of routerify features so we should remove it from the code base to save compile time, potential security issues,  dependencies, and our custom fork

__Proposal:__
Remove use of routerify

__Current behaviour:__
Routerify is compiled and hookedup but no features are used.

__Context:__
We tried to remove it previously in https://github.com/influxdata/influxdb_iox/pull/2586 but had to back out the change when we observed a high error rate on write -- https://github.com/influxdata/influxdb_iox/pull/2623

tustvold:
This reverts commit 18941fcbff83cc86aad89ea664febfa9f57355f3.

Originally attempted in #2586 this was reverted in #2623 as it was causing performance problems. This was narrowed down to debug output, which should have been resolved by #2648 

Resolves #2630


alamb (approved):
🚀

Issue #2633, opened 2021-09-25
Protobuf Encoding of ChunkId and ChunkOrder Permits Defaults


tustvold:
**Problem:**

Both ChunkId and ChunkOrder are represented as a plain `uint32` in the protobuf representations for gRPC and the catalog. This would be fine, however, they both can take on the value 0. As discussed at length in #2590 this has a couple of implications:

* All requests involving a chunk id or order, automatically default it to 0 if not specified
* When rendering the pbjson serialization of a ChunkSummary or similar, the id isn't rendered if it is 0

This is not necessarily problematic, but is definitely not ideal

**Proposal:**

Chunk IDs are no longer used for ordering within the catalog, and there has been a long-standing vision to migrate them eventually to just be UUIDs. As such the protobuf representation of chunk ids should be changed to be `string` from `uint32`. The internal representation within IOx of a `ChunkId` can remain unchanged for now. Not only does this solve the problems above, it also opens up the possibility of changing ChunkId to a UUID without this being a catalog + API breaking change.

Chunk Orders are used for ordering chunks within the catalog, and it therefore makes sense to keep their encoded representation as integers. However, much like other integer types appearing in the API, such as server ID, we should make them non-zero. @crepererum 's work to extract a type wrapper should greatly facilitate this.

**Alternatives Considered:**

Change one or both of the types to be `optional uint32`. This would be a perfectly valid approach, however, would require greater discipline to ensure it is consistently followed. Most `uint32` don't make sense to be optional, as the 0 values are not ambiguous, and so you end up in a situation where it isn't immediately obvious which encoding to use in what context, better imo to just not open up that can of worms.

crepererum:
Could we also serialize UUIDs as bytes or a fixed-size 128bit container (e.g. using two `fixed64` values) to make them cheaper to encode/decode? Note that the Rust UUID type is just a newtype wrapper for `[u8; 16]`.

tustvold:
I'd personally be fine with bytes or string, provided they render as a single "type" in pbjson idc :shrug:. Some people recommend against using string for UUIDs, stating endianness, but I'd expect clients to just treat them as an opaque sequence of bytes and so I think this doesn't really matter for our case

PR #2130, opened 2021-07-27
feat: DatabaseRules (de)serialisation support

domodwyer:
Allows Conductor to serialise these structures into JSON.

---

* feat: DatabaseRules (de)serialisation support (7d3b6bf8)

      Derives serde::Serialise & Deserialise for the DatabaseRules struct and 
      children.

PR #2652, opened 2021-09-28
feat: remove serde from data types

tustvold:
~~_This is currently blocked on #2633 but I have had it sitting in a branch for a while so thought I'd put it into a draft PR._~~

Now that pbjson can provide a stable serialization of our protobufs, we no longer need to serialize data_types. By removing this support not only do we save some compile time, but we avoid accidentally using a non-stable serialization format for things that might actually want or need stability, and/or the ability to evolve in API-compatible ways.

This was added in #2130 and I'm not entirely sure how conductor is making use of this, if it even is. However, following the extensive discussions on #2409 I'm fairly confident there is no correct way to use this serialization, as the data_types are defaulted and don't accurately reflect what was sent over the wire, nor can they accommodate unrecognized or missing fields.


tustvold:
I have confirmed that conductor makes no mention of `data_types` other than a transitive dependency from `generated_types` which can we can add an option to disable

tustvold:
Following #2739 conductor doesn't even have a transitive dependency on data_types and so I think this is good to go

mkmik:
I think conductor will only ever use json encoding of the protobuf layer (pbjson) and never the internal types

domodwyer (approved):
Serialising these directly was a semi-usable stop gap to unblock until we had #2409 and the time to decouple everything 👌 

I am unaware of anything needing serde support now.

PR #2655, opened 2021-09-28
chore: Make query logging a bit less verbose

alamb:
I have been messing with IOx locally, and I found that these particular lines show up in the logs by default, when I think they are supposed to be `debug`


NGA-TRAN (approved):
LGTM

Issue #2546, opened 2021-09-15
Not throw error when all of rows of a OS or RUB chunk are soft deleted


NGA-TRAN:
# Rationale
In the PR #2522 I have discovered an [error will be thrown here](https://github.com/influxdata/influxdb_iox/blob/74d3c2e6d23cae7c2bcde7cc497ea77eb6062139/server/src/db/lifecycle/persist.rs#L147) if all rows of an OS chunk are soft deleted. We will need to change this behavior.

Similarly, this to [RUB here](https://github.com/influxdata/influxdb_iox/blob/cbf166c035c5c28f7d359896ed07e806d3a2a48e/server/src/db/lifecycle/move_chunk.rs#L68)

NGA-TRAN:
FYI @crepererum

PR #2656, opened 2021-09-28
test: more tests that includes the repro of 2546

NGA-TRAN:
repro for #2546 

### DIFF
--- a/query_tests/src/sql.rs
+++ b/query_tests/src/sql.rs
@@ -844,6 +844,25 @@ async fn sql_select_all_different_tags_chunks() {
     .await;
 }
 
+// ----------------------------------------------
+// Delete tests
+
+// reproduce of https://github.com/influxdata/influxdb_iox/issues/2546
+// Not throw error when all of rows of a OS or RUB chunk are soft deleted
+// ignore until we have a fix that I am actively working on
+#[ignore]
+#[tokio::test]
+async fn sql_select_with_delete_from_one_expr_delete_all() {
+    let expected = vec!["++", "++"];
+
+    run_sql_test_case(
+        scenarios::delete::OneDeleteSimpleExprOneChunkDeleteAll {},
+        "SELECT * from cpu",
+        &expected,
+    )
+    .await;
+}
+
 #[tokio::test]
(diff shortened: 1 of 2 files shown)


query_tests/src/sql.rs
```diff
@@ -844,6 +844,25 @@ async fn sql_select_all_different_tags_chunks() {
     .await;
 }
 
+// ----------------------------------------------
+// Delete tests
+
+// reproduce of https://github.com/influxdata/influxdb_iox/issues/2546
+// Not throw error when all of rows of a OS or RUB chunk are soft deleted
```
alamb, line 851: 👍

query_tests/src/sql.rs
```diff
@@ -862,6 +881,35 @@ async fn sql_select_with_delete_from_one_expr() {
     .await;
 }
 
+#[tokio::test]
+async fn sql_select_with_delete_from_one_expr_with_select_predicate() {
```
alamb, line 885: I don't fully understand what extra coverage this test and the ones below offer. Are they just testing additional interactions of selection predicates with delete predicates?

alamb (approved):
Delete test that fails looks good 👍 

I don't understand the rationale for the other added tests in this PR; I don't think they harm things but I also don't understand how they help things either.

query_tests/src/sql.rs
NGA-TRAN: These are tests in my plan to add in and here are the reasons:
1. Since I have to merge negated delete predicates with normal select predicate and each delete predicate includes a set of expressions plus a time range, if we do not do it carefully (negate, and, or the right expressions), wrong results will return. It happened a few weeks ago when I just started merging them so I need more coverage tests of different kinds of expressions in the predicates and data set.
2. Since we have different kinds of chunk stages, we need to test the cases when delete predicates eliminate nothing, something, and everything from a chunk and many chunks with the similar combination of select predicates, on different set of data.
3. In those tests, I put one or many floats in the SQLs. My plan is when the integer bug in RUB is fixed, I will convert a part of them to integer to cover integer tests

Issue #2590, opened 2021-09-20
Mark all fields on management API fields as `optional` so generated code has a way to set `Option`


alamb:
__Use case:__
Since  the IOx Management API uses gRPC to serialize configuration, all fields are "optional" and thus and may not be present in a message a server receives. 

However, at least for the Rust code generator (tonic), if a field is not explicitly marked `optional` in the proto itself, the generated code for fields of primitive values (e.g. `u64`) is a `u64` value rather than `Option<u64>`, and thus there is no straightforward way for a client to NOT send a value for a primitive field. Even when the client does not want to change the field -- it HAS TO provide a value for that field.

This is problematic for when the client wants to use server side default values for some fields. Right now, for example, conductor needs to have logic / special magic numbers for such fields to specify if a default should be applied or not. 

Here is an example of such confusion [source](https://github.com/influxdata/conductor/blob/9b4a12619595398330307b08850ff7f4e313015c/libconductor/src/database/lifecycle_rules.rs#L76-L102):

```rust
            buffer_size_soft: v.buffer_size_soft.map(|v| v.get()).unwrap_or(0),
```
This code translates a `None` to a default value of `0` (via the call to `unwrap_or(0)`). 

Then on the IOx side `0` is interpreted as "default value". It was quite unobvious (to me) on the IOx side that the intent was for a field value of `0` is intended to mean "use the server default value". The [code](https://github.com/influxdata/influxdb_iox/blob/d9da7641539bb5f0b73ddb7eabdba5fca0595146/generated_types/src/database_rules/lifecycle.rs#L62)  relies on a conversion to `NonZeroU64` to fail .ok() is called:

```rust
            buffer_size_soft: (proto.buffer_size_soft as usize).try_into().ok(),
```

Further adding to the confusion:
1. @domodwyer  tells me there are some fields where `u64::MAX` must be used instead of `0`
2. The `tonic` code generator behavior is *different* for `struct` fields -- where all such fields are generated as `Option<..>` even if they are not marked as `optional` in the proto definition. 

__Proposal:__
Mark all configuration fields (e.g. DatabaseRules) as `optional` in the protos so that the generated code exposes the fact that all fields are actually optional and remove the special 

Doing so would likely require the   `.protoc_arg("--experimental_allow_proto3_optional")` option to be passed to tonic
https://github.com/influxdata/conductor/blob/9b4a12619595398330307b08850ff7f4e313015c/conductor_client/build.rs#L21

__Current behaviour:__
[...]

PR #2659, opened 2021-09-29
chore: expand protobuf documentation

tustvold:
This PR expands the documentation around protobuf to hopefully clear up confusion relating to #2590.

I also did a quick pass through the protobuf APIs adding some further documentation. I've not covered every field of every message, but I think I've covered most of the ones where the semantics are unclear.


generated_types/protos/influxdata/iox/management/v1/database_rules.proto
```diff
@@ -69,21 +73,30 @@ message LifecycleRules {
...
+  // it will be compacted and, if set, persisted. Writers will generally
+  // have this amount of time to send late arriving writes or this could
+  // be their clock skew.
+  //
+  // If 0, a server-side default is used
   uint32 late_arrive_window_seconds = 12;
 
   // Maximum number of rows before triggering persistence
+  //
+  // If 0, a server-side default is used
```
tustvold, line 86: I opted to avoid defining what the default is, so as to avoid the proto spec drifting away from the code. I don't feel particularly strongly about this though

docs/protobuf.md
```diff
@@ -3,21 +3,139 @@
...
-Tools always come to the rescue, but public protobuf tooling is embarrassingly behind.
+## Guidelines
 
-## prototxt
+Some general guidelines, in addition to those enforced by [buf lint](https://docs.buf.build/lint/overview):
 
-The `protoc` tool can (de)serialize to and from binary protobuf files,
-with the `--decode` and `--encode` flags. However, it needs to be passed the schema of the files *and*
-the fully qualified type name of the top level message in the file to be decoded.
+* Only use proto syntax 3
```
alamb, line 12: Perhaps you could add a link to proto3 syntax?

docs/protobuf.md
```diff
@@ -3,21 +3,139 @@
...
+* Non-nullable unsigned 32-bit integer -> `u32`
+* Nullable unsigned 32-bit integer -> `Option<u32>`
+* Non-nullable unsigned 32-bit integer array -> `Vec<u32>`
+* Nullable unsigned 32-bit integer array -> `Option<Vec<U32>>`
+* Non-nullable map -> `BTreeMap<_, _>`
+* Nullable map -> `Option<BTreeMap<_,_>>`
+* Non-nullable string -> `String`
+* Nullable string -> `Option<String>`
+* Non-nullable boolean -> `bool`
+* Nullable boolean -> `Option<bool>`
```
alamb, line 32:
```suggestion
```

I am not sure these extra type examples add much. They may be worth eliding to make this section more concise

jacobmarble:
There is a lot of helpful info in here, thanks for the update.

docs/protobuf.md
```diff
@@ -3,21 +3,139 @@
...
+
+Nullability allows a client to distinguish between a field specified to have the default value, and an unspecified
+field. Without nullability there is no way to represent this additional presence information.
+
+### Proto 3
+
```
alamb, line 38: I suggest you lead with the conclusion and then dive into the details. As it is, my eyes almost glazed over before I got to the conclusion (aka I think you "buried the lead" here)

Perhaps a lead sentence such as  "Nullability with proto3 can not be encoded for all types and thus care must be taken when converting to/from fields which such types in Rust"

docs/protobuf.md
```diff
@@ -3,21 +3,139 @@
...
+```protobuf
+message MyMessage {
+  oneof limit_opt {
+    uint64 limit = 1;
+  }
+}
+```
+
+In never versions of protoc there is also a short-hand for this, with the additional benefit that many code generators
```
alamb, line 85:
```suggestion
In newer versions of protoc there is also a short-hand for this, with the additional benefit that many code generators
```

docs/protobuf.md
```diff
@@ -3,21 +3,139 @@
...
-The `protoc` tool can (de)serialize to and from binary protobuf files,
-with the `--decode` and `--encode` flags. However, it needs to be passed the schema of the files *and*
-the fully qualified type name of the top level message in the file to be decoded.
+* Only use proto syntax 3
+* When rendering protobuf for CLIs, etc... prefer the [JSON](#json) representation over others (e.g. `std::fmt::Debug`)
+* Clearly document any additional semantics attached to a field's [default value](#nullability--presence--defaults)
 
-We have a script that hides the boring parts of finding all the `.proto` files and adding the right
-import paths etc. It also has a convenient command that lists all the fully qualified message types.
+## Nullability / Presence / Defaults
```
alamb, line 16: The content here is almost worthy of a blog post, in my mind :)

generated_types/protos/influxdata/iox/management/v1/database_rules.proto
```diff
@@ -35,10 +35,14 @@ message PartitionTemplate {
 message LifecycleRules {
   // Once the total amount of buffered data in memory reaches this size start
   // dropping data from memory
+  //
+  // If 0, no limit
```
alamb, line 39: 👍

alamb: Should it say "if 0 or unspecified"?  Or maybe that is adding to the confusion (I realize that the protobuf specification effectively creates the default value of 0)

generated_types/protos/influxdata/iox/management/v1/database_rules.proto
alamb: I like the approach in this PR

alamb (approved):
Looks like an improvement to me. 👍  thank you

I vaguely remember that @domodwyer  mentioned that for some fields, to get the default value something other than zero (`uint32::MAX`?) was required. However, I was not able to find any examples of that in current conductor,  so I may be mis-remembering

```shell
rg -n -H --no-heading -e 'MAX' $(git rev-parse --show-toplevel || pwd)
/Users/alamb/Software/conductor/lib/node/src/server_id.rs:125:            let lower = id.get_u32() & (u32::MAX >> 16);
/Users/alamb/Software/conductor/lib/node/src/server_id.rs:142:        let zeroed = server_id.get_u32() & (u32::MAX << 16);
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles.rs:24:const NODE_REQUEST_QUEUE_MAX: usize = 100;
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/txn_db_config/split_strategy_config.rs:192:                .upsert_database(Arc::clone(c), Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/txn_db_config/split_strategy_config.rs:197:                .delete_database(db.name().clone().into(), Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/query/query_cluster.rs:627:                .upsert_database(Arc::clone(config), Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/query/query_cluster.rs:702:                    .upsert_database(old, Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/query/query_cluster.rs:708:                .delete_database(Arc::new(db.clone()), Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/query/dispatcher.rs:20:            NODE_REQUEST_QUEUE_MAX,
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/query/dispatcher.rs:42:        let (node_queue_tx, _) = broadcast::channel(NODE_REQUEST_QUEUE_MAX);
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/dispatcher.rs:15:use crate::cluster::{roles::NODE_REQUEST_QUEUE_MAX, RequestDispatcher};
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/dispatcher.rs:37:        let (tx, _) = broadcast::channel(NODE_REQUEST_QUEUE_MAX);
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/unique_name_filter.rs:268:            $a.upsert_database($param, Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/unique_name_filter.rs:271:            $b.upsert_database($param, Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/unique_name_filter.rs:274:            $a.delete_database($param, Duration::MAX)
/Users/alamb/Software/conductor/lib/libconductor/src/cluster/roles/router/unique_name_filter.rs:277:            $b.delete_database($param, Duration::MAX)
/Users/alamb/Software/conductor/lib/util/src/backoff.rs:176:        b.count = i32::MAX;
/Users/alamb/Software/conductor/conductor_cli/src/commands/run.rs:102:        env = "DB_MAX_CONNECTIONS",
```

docs/protobuf.md
tustvold: I removed half of them, I wanted to make clear what nullability means for collection types, in particular how it differs from emptiness

docs/protobuf.md
tustvold: I've updated this section to hopefully be more clear, and to make it clearer that `prost` isn't the only code generator to use the lack of nullability to improve ergonomics. It was one of the reasons they removed nullability in the first place :laughing:  PTAL

docs/protobuf.md
tustvold: Pretty niche audience I imagine :laughing:

tustvold:
> I vaguely remember that @domodwyer mentioned that for some fields, to get the default value something other than zero (uint32::MAX?) was required.

Specifying something other than the protobuf default to get the server-side default would certainly be a relatively strange API, which if we are doing somewhere we should definitely change. I'm fairly certain, however, that we aren't.

Somewhat tangentially related to that point though, I don't see anything inherently wrong with an API endpoint assigning additional meaning to `u32::MAX` or any non-zero value tbh. Its no different in my mind from assigning constraints other than those inherent to the type - provided it is well documented it's fine. Sure you should avoid your APIs being wildly inconsistent, but other than the default value there's nothing "special" from protobuf's perspective

docs/protobuf.md
alamb: But for those who are interested, they *NEED* to know!

alamb:
> Somewhat tangentially related to that point though, I don't see anything inherently wrong with an API endpoint assigning additional meaning to u32::MAX ...

I agree there is nothing inherently wrong. 

However, it seems to me like if some fields use `0` to specify "use server side default" and others use `u32::MAX` to mean the same thing, the API would be hard to use correctly from a user's point of view

tustvold:
> However, it seems to me like if some fields use 0 to specify "use server side default" and others use u32::MAX to mean the same thing, the API would be hard to use correctly from a user's point of view

100% agree, the APIs should strive to be as consistent as possible. Further, using anything other than the 0 value to signal server side default would be likely be ill-advised, I even said a few lines earlier if we are doing this somewhere we should change this :sweat_smile:. 

My point was that the 0 value need not be the only "special value", e.g. using u32::MAX to indicate no limit or something, in addition to potentially using the 0 value to indicate server-side default, is perfectly valid.

tustvold:
I'm going to get this in, and if @domodwyer has any feedback when he gets back, I'll factor that into a follow up PR

Issue #2294, opened 2021-08-16
Incorrect query results while deduplicating


alamb:
@tustvold  noticed the following in an explain plan:

```
                      SortPreservingMergeExec: [state@1 ASC,city@0 ASC,time@2 ASC]          |
|               |               ExecutionPlan(PlaceHolder)                                          |
|               |                 RepartitionExec: partitioning=RoundRobinBatch(4)                  |
|               |                   IOxReadFilterNode: table_name=h2o, chunks=1 predicate=Predicate |
|               |                 RepartitionExec: partitioning=RoundRobinBatch(4)                  |
|               |                   IOxReadFilterNode: table_name=h2o, chunks=1 predicate=Predicate |
```

In pictures:

```
                          ▲                                        
                          │                                        
                          │                                        
                          │                                        
              ┌───────────────────────┐                            
              │SortPreservingMergeExec│                            
              └───────────────────────┘                            
                          ▲                                        
                          │                                        
            ┌─────────────┴─────────────┐                          
            │                           │                          
            │                           │                          
┌───────────────────────┐   ┌───────────────────────┐              
│    RepartitionExec    │   │    RepartitionExec    │◀─            
└───────────────────────┘   └───────────────────────┘  │           
            ▲                           ▲                          
            │                           │              │           
            │                           │                          
┌───────────────────────┐   ┌───────────────────────┐  │           
│   IOxReadFilterNode   │   │   IOxReadFilterNode   │              
└───────────────────────┘   └───────────────────────┘  │           
                                                  ─ ─ ─            
                                                 │                 
                                                                   
                                                 │                 
                                                                   
                                                 │                 
                                Adding these RepartitionExec nodes 
                               will result in incorrect results (as
                                 the stream is no longer sorted)   
                                                                   
```

He notes this may not actually produce incorrect results for the query at the moment because each chunk only yields a single record batch due to how `IOxReadFilter` is implemented; But if/when `IOxReadFilter` did yield multiple RecordBatches the sortedness would be not be maintained and thus the input to `SortPreservingMergeExec` would not necessairly be sorted, and rows could be lost (wrong results)

__Steps to reproduce:__
Not sure

__Expected behaviour:__
Do not repartition streams prior to `SortPreservingMergeExec`

__Actual behaviour:__
repartition streams prior to `SortPreservingMergeExec`

alamb:
FYI @NGA-TRAN

alamb:
[...]

PR #2660, opened 2021-09-29
HACK: return 1 row record batches from IOxReadFilterNode 

alamb:
Demonstrate that IOx still gets correct answers even when multiple record batches are returned from `IOxReadFilterNode` -- re https://github.com/influxdata/influxdb_iox/issues/2294

NOT FOR MERGING

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -148,7 +148,7 @@ jobs:
       - cache_restore
       - run:
           name: Cargo test
-          command: cargo test --workspace
+          command: cargo test --workspace  --no-fail-fast
       - cache_save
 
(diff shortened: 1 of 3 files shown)


alamb:
There is one failure: 
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/10817/workflows/9478565c-7865-41f1-87bd-91d0279cd6e9/jobs/69494?invite=true#step-104-5752

```
    end_to_end_cases::persistence::test_full_lifecycle
``` 

Which I believe is related to the sizes of the read buffer being slightly smaller with multiple record batches being fed

alamb:
I have no plans to merge this

PR #2661, opened 2021-09-29
fix: fallback to DataFusion predicates if unable to execute on RUB

### LINKED ISSUE #1603: Incorrect results with integer predicate on floating point column

e-dard:
Closes #1603 

This PR restricts the validation of predicates against the Read Buffer to only those that contain the same literal type as the column specified in the chunk schema.

For cases where the expression's literal differs in type (e.g., it's a float) from the column (e.g., it's an integer column) the new `validate_predicate` API will return an error. This API is used to validate predicates againts read buffer chunks before execution begins, which gives us the opportunity to fall-back to using the `Filter` step in the DF plan.

*Note* - I have a follow-on PR planned that will allow us to push expressions like `col_int > 3.2` down to the RUB, but I wanted to split it out of the "fix" step.

This PR fixes the issue of wrong query results, the follow on will allow us to push more types of predicate down (without breaking query results).

### DIFF
--- a/query_tests/src/sql.rs
+++ b/query_tests/src/sql.rs
@@ -779,6 +779,28 @@ async fn sql_predicate_pushdown_correctness_13() {
 }
 
 #[tokio::test]
+async fn sql_predicate_pushdown_correctness_14() {
+    // Test 14: on push-down expression with a literal type different from the
+    // column type.
+    //
+    // Check correctness
+    let expected = vec![
+        "+-------+--------+--------------------------------+---------+",
+        "| count | system | time                           | town    |",
+        "+-------+--------+--------------------------------+---------+",
+        "| 632   | 5      | 1970-01-01T00:00:00.000000120Z | reading |",
+        "| 632   | 6      | 1970-01-01T00:00:00.000000130Z | reading |",
+        "+-------+--------+--------------------------------+---------+",
+    ];
+    run_sql_test_case(
+        TwoMeasurementsPredicatePushDown {},
+        "SELECT * from restaurant where count > 500.76 and count < 640.0",
+        &expected,
+    )
+    .await;
+}
+
+#[tokio::test]
 async fn sql_deduplicate_1() {
     // This current expected is wrong because deduplicate is not available yet
     let sql =
--- a/read_buffer/src/table.rs
+++ b/read_buffer/src/table.rs
@@ -234,6 +234,14 @@ impl Table {
         Arc::clone(&self.table_data.read().meta)
     }
 
+    /// Validates if the predicate can be applied to the table based on the
+    /// schema and the predicate's expressions. Returns an error if the
+    /// predicate cannot be applied.
+    pub fn validate_predicate(&self, predicate: Predicate) -> Result<Predicate, Error> {
+        let table_data = self.table_data.read();
+        Ok(table_data.meta.validate_exprs(predicate)?.into())
+    }
+
     /// Determines if one of more row groups in the `Table` could possibly
     /// contain one or more rows that satisfy the provided predicate.
     pub fn could_pass_predicate(&self, predicate: &Predicate) -> bool {
--- a/read_buffer/src/table.rs
+++ b/read_buffer/src/table.rs
@@ -25,7 +25,7 @@ pub enum Error {
     #[snafu(display("table does not have InfluxDB timestamp column"))]
     NoTimestampColumn {},
 
-    #[snafu(display("unsupported column operation on {}: {}", column_name, msg))]
+    #[snafu(display("unsupported column operation on column \"{}\": {}", column_name, msg))]
     UnsupportedColumnOperation { msg: String, column_name: String },
 }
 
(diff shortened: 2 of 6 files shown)


query_tests/src/sql.rs
```diff
@@ -778,6 +778,28 @@ async fn sql_predicate_pushdown_correctness_13() {
     .await;
 }
 
+#[tokio::test]
```
e-dard, line 781: Test showing that the RUB now gives the correct results when using a predicate that contains a literal with a different type to the column (`count` is an unsigned integer column).

read_buffer/src/table.rs
```diff
@@ -289,13 +294,19 @@ impl Table {
...
-        // validate that the predicate can be applied to the row groups.
-        let (meta, row_groups) = self.filter_row_groups(predicate);
-        meta.validate_exprs(predicate.iter())?;
+        let (meta, row_groups) = {
+            let table_data = self.table_data.read();
+            (Arc::clone(&table_data.meta), table_data.data.clone())
+        };
+
+        // Determine if predicate can be applied to table.
+        let predicate: Predicate = meta.validate_exprs(predicate.clone())?.into();
```
e-dard, line 303: Physical operations on the Read Buffer will still validate predicates even if the caller already did it. I want the Read Buffer to return an error if someone forgot to validate a predicate, rather than incorrect query results.

read_buffer/src/table.rs
```diff
@@ -704,31 +733,32 @@ impl MetaData {
...
-            .try_for_each(|expr| match self.columns.get(expr.column()) {
+    /// Determine, based on the table meta data, whether each provided expression
+    /// can be applied as is, or successfully rewritten to a form that can be
+    /// applied. If an expression cannot be applied then an error is returned.
+    pub fn validate_exprs(
+        &self,
+        iter: impl IntoIterator<Item = BinaryExpr>,
+    ) -> Result<Vec<BinaryExpr>, Error> {
+        iter.into_iter().try_fold(vec![], |mut arr, expr| {
+            match self.columns.get(expr.column()) {
```
e-dard, line 744: I have removed a bunch of types that can be compared in this PR to get the read buffer producing correct results. In a follow on PR I will be putting these types back in and then where possibke _rewriting_ the expression so that the literal is of the correct type for the column. This will allow more predicates to be pushed down to the Read Buffer without having to make major changes to the type system in the depths of the RUB's execution engine.

server/src/db/chunk.rs
```diff
@@ -360,15 +365,21 @@ impl QueryChunk for DbChunk {
             }
             State::ReadBuffer { chunk, .. } => {
                 // Only apply pushdownable predicates
-                let rb_predicate =
-                    match to_read_buffer_predicate(predicate).context(PredicateConversion) {
-                        Ok(predicate) => predicate,
-                        Err(_) => read_buffer::Predicate::default(),
-                    };
+                let rb_predicate = chunk
```
e-dard, line 368: So now as well as converting the predicate to a Read Buffer predicate we also validate it against the RUB chunk. If it fails validation then we replace it with the default empty predicate ensuring that the Read Buffer will give the correct results and the filtering can be done higher up in the query engine.

e-dard: Of course, we probably could use the catalog of whatever to do this validation too. But since I wanted to add the logic to the Read Buffer anyway (trust but verify!) it seemed simple enough to just use the Read Buffer chunk's API to do the validation.

read_buffer/src/table.rs
```diff
@@ -289,13 +294,19 @@ impl Table {
...
+
+        // Determine if predicate can be applied to table.
+        let predicate: Predicate = meta.validate_exprs(predicate.clone())?.into();
+
+        // Determine if the negated predicates (deletes) can be applied to the
+        // table.
+        let mut n_predicates: Vec<Predicate> = vec![];
         for pred in negated_predicates {
-            meta.validate_exprs(pred.iter())?;
+            n_predicates.push(meta.validate_exprs(pred.clone())?.into());
```
NGA-TRAN, line 309: So the intension is either the `n_predicates` will look exactly the same as the `negated_predicates` or an error will be return?

server/src/db/chunk.rs
alamb: Given that `to_read_buffer_predicate` already returns a `Result` I was wondering if it would make sense to include a call to `validate_predicate` in the implementation of `to_read_buffer_predicate()` so it never got forgotten.

However, then I realized:
1. This is the only callsite: https://github.com/influxdata/influxdb_iox/search?q=to_read_buffer_predicate
2. The note below applies:  "the validity of a predicate also depends on the chunks's schema" which makes sense 👍

NGA-TRAN: I see, at this step you eliminate the predicates that read buffer cannot handle yet 👍

NGA-TRAN (approved):
The changes look good to me. After you merge this, I will modify some of the tests that failed before to test this further

server/src/db/chunk.rs
```diff
@@ -360,15 +365,21 @@ impl QueryChunk for DbChunk {
...
+                    // A predicate unsupported by the Read Buffer or against
+                    // this chunk's schema is replaced with a default empty
+                    // predicate.
+                    .validate_predicate(to_read_buffer_predicate(predicate).unwrap_or_default())
```
alamb, line 372: the same strategy of skipping parts of  list of predicate exprs but still applying those that can be, might be worth considering here;

As it is if the predicate looks like `time < 100000 AND time >= 2000000 AND float_col < 1` nothing will be pushed down into the read buffer, where as it seems like the `time < 100000 AND time >= 2000000 ` could and should

read_buffer/src/table.rs
```diff
@@ -330,8 +343,16 @@ impl Table {
         group_columns: &'input Selection<'_>,
         aggregates: &'input [(ColumnName<'input>, AggregateType)],
     ) -> Result<ReadAggregateResults> {
-        let (meta, row_groups) = self.filter_row_groups(&predicate);
-        meta.validate_exprs(predicate.iter())?;
+        //
+        // TODO(edd): add delete support if/when aggregates can be pushed down.
```
alamb, line 347: This type of thing is why I was suggesting not special casing the delete predicates (but instead somehow encode them in the normal `Predicate`) -- it would be extra work for the higher level query layer to transform delete predicates and identify special deleted ones, but then the RUB doesn't have to special case delete predicates in all of its APIs

alamb (approved):
> Note - I have a follow-on PR planned that will allow us to push expressions like col_int > 3.2 down to the RUB, but I wanted to split it out of the "fix" step.

I like this plan 👍 

I also verified that the reproducer on https://github.com/influxdata/influxdb_iox/compare/alamb/read_with_predicate_failure passes successfully with this change (by cherry-picking `e2feca14374e636d0e500ec6124d163132196df5` locally and fixing up some conflicts)

read_buffer/src/table.rs
e-dard: Right now yes. But as a follow on the predicate expressions may be rewritten, which is why the API consumes the predicate and returns a new one.

server/src/db/chunk.rs
e-dard: Yeah totally agree. I think it's worth considering though that not doing an "all or nothing" makes it harder to remove the extra Filter in the DF plan, mainly because it won't be as straightforward to know if the RUB could handle all the predicate. But worth thinking about.

alamb: In general, I think the easiest interface to the RUB to reason about would be:

`pred: Vec<PredicateExpr>`

Where each `PredicateExpr` could be whatever form the ReadBuffer has an optimization for. 

perhaps something like
```rust
enum PredicateExpr {
  BinaryExpr(BinaryExpr),
  NegatedExpr(BinaryExpr)
}
```

Issue #2665, opened 2021-09-29
Support Persisted Chunks Within compact_chunks Lifecycle Action


tustvold:
It should be possible to compact persisted chunks

alamb:
I would recommend writing up a design for this item -- specifically with a proposed API (do you say "compact all chunks in this partition", or "compact only these specific chunks", "compact all chunks in a table", etc)

Also, it would probably be good to explicitly spell out "can you compact chunks that are in different partitions" (answer is probably no, but it would be good to write down)

tustvold:
> I would recommend writing up a design for this item

Will do as part of this ticket. I was envisaging that it'd just be a case of updating the compact_chunks lifecycle action to verify all the provided chunks are either persisted or not, and if they are persisted to perform the necessary dance to upsert the resulting chunk into the persisted catalog and then the in-memory catalog.

Adding a gRPC API is tracked by #1934

alamb:
> Will do as part of this ticket. I was envisaging that it'd just be a case of updating the compact_chunks lifecycle action to verify all the provided chunks are either persisted or not, and if they are persisted to perform the necessary dance to upsert the resulting chunk into the persisted catalog and then the in-memory catalog.

Cool -- I don't think a detailed technical implementation design is needed (what you describe is probably good enough along with a well commented PR).

I think I was just getting at was clearly defining what was / was not allowed in terms of compacting persisted chunks together and how the user (e.g. conductor / an operator) will specify it

NGA-TRAN:
I close this because the `compact_object_chunks` and `compact_object_store_partition` are available. I have created #3355 to add those into our lifecycle

Issue #2047, opened 2021-07-19
Refactor: Create a catalog transaction type for acquiring and releasing catalog locks


alamb:
__Use case:__
IOx uses several different kinds of locks that may be held that relate to its handling of persisted (catalog) data:
1. CatalogLocks (on Partititions and Chunks)
2. Locks on the Persistet catalog  (e.g. [`TransactionHandle`](https://github.com/influxdata/influxdb_iox/blob/0c86d1dccfcde7fffff46705a3a3ec15cc6d6cf3/parquet_file/src/catalog.rs#L973-L979)
2. [`FlushHandle`](https://github.com/influxdata/influxdb_iox/blob/d71f38f27c41cfb0039d3d7d4ba305cbbc1f1506/persistence_windows/src/persistence_windows.rs#L64) (locks on persist windows)
3. the [`ChunkLifecycleAction`](https://github.com/influxdata/influxdb_iox/blob/00b89cd751fa43203855a095086ae255ff149dda/server/src/db/catalog/chunk.rs#L190) mark on Chunks
4. a [`cleanup_lock`](https://github.com/influxdata/influxdb_iox/blob/main/server/src/db.rs#L262) for controlling parquet file cleanup

Operations such as writing new data into IOx, or running lifecycle actions require some subset of these types of locks at various times. This leads to several challenges;

1. **Deadlocks**: The order in which to take the locks to prevent deadlock is (currently) implicit and unchecked, which produces intermittent deadlocks under load such as https://github.com/influxdata/influxdb_iox/issues/2045
2. **Manual Cleanup**: Due to some things which can't / shouldn't be held across `await`  points, there is manual cleanup lock like [`maybe_cleanup_failed`](https://github.com/influxdata/influxdb_iox/blob/00b89cd751fa43203855a095086ae255ff149dda/lifecycle/src/policy.rs#L428) which is both wasteful from a CPU perspective (see https://github.com/influxdata/influxdb_iox/issues/1962#issuecomment-879063198) but also I suspect a source of lurking bugs
1. **Cognitive Overhead** : Figuring out which locks are needed and in what order to acquire them for a desired operation is non trivial (especially as crates like `lifecycle` abstract them away behind traits)
2. **Visibility**: When some operation is waiting on locks / handles from another, there is currently no way to see what operation is holding the locks nor how long they have been waiting

__Proposal:__
[...]

PR #2674, opened 2021-09-30
feat: remove move lifecycle action

tustvold:
The move chunk lifecycle action is a relic of the old per-chunk lifecycle, and is not used by the lifecycle policy. It instead was kept around to avoid a large amount of test churn. This PR is that churn and migrates to use the compact_chunks lifecycle action instead.

Where possible I also took the opportunity to adjust tests to be agnostic to how chunk IDs are allocated, as this not only avoids coupling query tests unnecessarily to what the lifecycle is doing, but also opens the door to making chunk IDs UUIDs.


data_types/src/chunk_metadata.rs
```diff
@@ -127,6 +123,9 @@ pub struct ChunkSummary {
     /// The table of this chunk
     pub table_name: Arc<str>,
 
+    /// Order of this chunk relative to other overlapping chunks.
+    pub order: ChunkOrder,
```
tustvold, line 127: This is moved up in the order so that when sorting ChunkSummary they're sorted by order then id

query_tests/cases/in/duplicates.expected
```diff
@@ -1,4 +1,4 @@
--- Test Setup: OneMeasurementThreeChunksWithDuplicates
+-- Test Setup: OneMeasurementFourChunksWithDuplicates
```
tustvold, line 1: I renamed this as it was actually four chunks :laughing:

query_tests/src/pruning.rs
```diff
@@ -24,29 +24,15 @@ async fn setup() -> TestDb {
...
     let partition_key = "1970-01-01T00";
-    let mb_chunk = db
-        .rollover_partition("cpu", partition_key)
-        .await
-        .unwrap()
-        .unwrap();
-    db.move_chunk_to_read_buffer("cpu", partition_key, mb_chunk.id())
-        .await
-        .unwrap();
+    db.compact_open_chunk("cpu", partition_key).await.unwrap();
```
tustvold, line 27: There isn't a need to freeze a chunk before compacting or persisting, so the additional step can just be removed. Both lifecycle actions will freeze the chunk as part of proceeding

query_tests/src/sql.rs
```diff
@@ -381,28 +381,28 @@ async fn sql_select_from_system_chunk_columns() {
...
-        "| 1970-01-01T00 | 0        | o2         | time        | OpenMutableBuffer | 2         | 0          | 50        | 300       | 297          |",
-        "| 1970-01-01T00 | 1        | h2o        | city        | OpenMutableBuffer | 1         | 0          | Boston    | Boston    | 309          |",
-        "| 1970-01-01T00 | 1        | h2o        | other_temp  | OpenMutableBuffer | 1         | 0          | 72.4      | 72.4      | 297          |",
-        "| 1970-01-01T00 | 1        | h2o        | state       | OpenMutableBuffer | 1         | 0          | CA        | CA        | 309          |",
-        "| 1970-01-01T00 | 1        | h2o        | time        | OpenMutableBuffer | 1         | 0          | 350       | 350       | 297          |",
-        "+---------------+----------+------------+-------------+-------------------+-----------+------------+-----------+-----------+--------------+",
+        "+---------------+------------+-------------+-------------------+-----------+------------+-----------+-----------+--------------+",
```
tustvold, line 384: This output broke because the chunk IDs changed, I just took the opportunity to remove them

server/src/db.rs
```diff
@@ -482,12 +495,18 @@ impl Db {
...
         let fut = {
-            let partition = self.partition(table_name, partition_key)?;
-            let partition = LockableCatalogPartition::new(Arc::clone(self), partition);
+            let partition = self.lockable_partition(table_name, partition_key)?;
 
             // Do lock dance to get a write lock on the partition as well
             // as on the to-be-dropped chunk.
             let partition = partition.read();
+            LockablePartition::chunk(&partition, chunk_id).ok_or(
+                catalog::Error::ChunkNotFound {
```
tustvold, line 504: This error dance will hopefully be fixed by #2047

server/src/db.rs
```diff
@@ -608,6 +616,19 @@ impl Db {
...
         partition_key: &str,
+    ) -> Result<Arc<DbChunk>> {
+        self.compact_chunks(table_name, partition_key, |_| true)
+            .await
+    }
+
+    /// Compacts all chunks within a partition passing a predicate
+    ///
+    /// There is no lock gap between predicate evaluation and creation of the lifecycle action
+    pub async fn compact_chunks(
```
tustvold, line 627: This is a bit strange, but is likely temporary as #2047 will allow removing these "helper" methods from Db. They're only used by tests at the moment

server/src/db/catalog/chunk.rs
```diff
@@ -1062,46 +996,46 @@ mod tests {
 
         // set some action
         chunk
-            .set_lifecycle_action(ChunkLifecycleAction::Moving, &registration)
+            .set_lifecycle_action(ChunkLifecycleAction::Compacting, &registration)
```
tustvold, line 999: The actual lifecycle action is irrelevant for these tests, so just changed to using a different one

server/src/db/catalog/chunk.rs
```diff
@@ -1124,13 +1058,13 @@ mod tests {
...
             .unwrap();
 
         // clearing now fails because task is still in progress
         assert_eq!(
             chunk.clear_lifecycle_action().unwrap_err().to_string(),
-            "Internal Error: Cannot clear a lifecycle action 'Moving to the Read Buffer' for chunk Chunk('db':'table1':'part1':0) that is still running",
+            "Internal Error: Cannot clear a lifecycle action 'Compacting' for chunk Chunk('db':'table1':'part1':0) that is still running",
```
NGA-TRAN, line 1067: So now we use the word `compacting` for moving MUB(s) to a RUB and `persisting` for creating an OS for RUB(s)?

server/src/db/catalog/chunk.rs
```diff
@@ -1177,10 +1110,6 @@ mod tests {
         }
         assert_eq!(pred.exprs, expected_exprs1);
 
-        // Move the chunk
-        chunk.set_moving(&registration).unwrap();
-        // The chunk still should be frozen
-        assert_eq!(chunk.stage().name(), "Frozen");
```
NGA-TRAN, line 1183: 👍

NGA-TRAN (approved):
I am a big fan of this PR.

data_types/src/chunk_metadata.rs
```diff
@@ -87,9 +87,6 @@ impl ChunkStorage {
 /// Any lifecycle action currently in progress for this chunk
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 pub enum ChunkLifecycleAction {
-    /// Chunk is in the process of being moved to the read buffer
```
alamb, line 90: 👍

generated_types/protos/influxdata/iox/management/v1/chunk.proto
```diff
@@ -28,12 +28,11 @@ enum ChunkStorage {
 
 // Is there any lifecycle action currently outstanding for this chunk?
 enum ChunkLifecycleAction {
+  reserved 1;
```
alamb, line 31:
```suggestion
  // Previously actions that are no longer supported
  reserved 1;
```

generated_types/protos/influxdata/iox/management/v1/jobs.proto
```diff
@@ -27,12 +27,11 @@ message OperationMetadata {
   // The number of tasks that did not run to completion (e.g. panic)
   uint64 dropped_count = 16;
 
-  reserved 6;
+  reserved 6, 7;
```
alamb, line 30:
```suggestion
  // Previously operations that are no longer supported
  reserved 6, 7;
```

lifecycle/src/lib.rs
```diff
@@ -134,14 +134,6 @@ pub trait LockableChunk: Sized {
     /// Acquire an exclusive write lock on the chunk
     fn write(&self) -> LifecycleWriteGuard<'_, Self::Chunk, Self>;
 
-    /// Starts an operation to move a chunk to the read buffer
-    ///
-    /// TODO: Remove this function from the trait as it is
```
alamb, line 139: 🎉

query_tests/src/scenarios.rs
```diff
@@ -104,28 +104,25 @@ impl DbSetup for NoData {
...
+        db.rollover_partition(table_name, partition_key)
+            .await
+            .unwrap();
         assert_eq!(count_mutable_buffer_chunks(&db), 1); //
         assert_eq!(count_read_buffer_chunks(&db), 0); // nothing yet
         assert_eq!(count_object_store_chunks(&db), 0); // nothing yet
 
         // Now load the closed chunk into the RB
-        db.move_chunk_to_read_buffer(table_name, partition_key, ChunkId::new(0))
+        let chunk_id = db
```
alamb, line 115: the use of Ids rather than numbers is a nice improvement

query_tests/src/scenarios.rs
```diff
@@ -463,21 +457,19 @@ impl DbSetup for OneMeasurementTwoChunksDifferentTagSet {
...
         ];
         write_lp(&db, &lp_lines.join("\n")).await;
-        db.rollover_partition("h2o", partition_key).await.unwrap();
-        db.move_chunk_to_read_buffer("h2o", partition_key, ChunkId::new(1))
-            .await
-            .unwrap();
+        db.compact_open_chunk("h2o", partition_key).await.unwrap();
+
+        assert_eq!(count_mutable_buffer_chunks(&db), 0);
+        assert_eq!(count_read_buffer_chunks(&db), 2);
```
alamb, line 471: thank you for adding this check / proof there were still two distinct read buffer chunks. I thought we may have compacted the existing read buffer chunk as well

query_tests/src/scenarios.rs
```diff
@@ -885,21 +858,14 @@ pub async fn make_two_chunk_scenarios(
...
-            .await
-            .unwrap();
-
-        db.move_chunk_to_read_buffer(table_name, partition_key, ChunkId::new(0))
-            .await
-            .unwrap();
-
-        db.move_chunk_to_read_buffer(table_name, partition_key, ChunkId::new(1))
+        // Compact just the last chunk
+        db.compact_open_chunk(table_name, partition_key)
```
alamb, line 868: previously the code made 2 readbuffer chunks -- is that still th case? Or does this compact into a single chunk?

query_tests/src/sql.rs
alamb: To anyone else looking at this -- @tustvold  stopped selecting the `chunk_id` in this particular test query -- I at first thought he removed the column entirely from the table. 😆

server/src/db.rs
```diff
@@ -472,6 +476,15 @@ impl Db {
         })
     }
 
+    pub fn lockable_partition(
```
alamb, line 479: I do wonder about exposing this API as `pub` as it seems easy to misuse -- should we maybe make it `pub(crate)` or `#[cfg(test)]`?

server/src/db.rs
```diff
@@ -1732,17 +1754,19 @@ mod tests {
...
+                Instant::now() + Duration::from_secs(1),
+                || t6_write,
+            )
+            .await
+            .unwrap()
+            .id();
 
         // A chunk is now in the object store and still in read buffer
-        let expected_parquet_size = 1243;
+        let expected_parquet_size = 1244;
```
alamb, line 1769: do we know where the extra byte came from ? Is it due to ordering the fields differently in Chunk or something?

server/src/db/catalog/chunk.rs
alamb: that is my understanding too

server/src/db/lifecycle/move_chunk.rs
```diff
@@ -1,84 +0,0 @@
-use crate::db::catalog::chunk::CatalogChunk;
```
alamb, line 1: image

server/src/lib.rs
```diff
@@ -973,14 +985,25 @@ where
...
+                table: table_name.to_string(),
+            })?;
 
-        let guard = chunk.write();
+        let partition = partition.upgrade();
+        let chunk = chunk.write();
 
-        LockableChunk::move_to_read_buffer(guard).map_err(|e| Error::UnknownDatabaseError {
-            source: Box::new(e),
+        LockablePartition::compact_chunks(partition, vec![chunk]).map_err(|e| {
```
alamb, line 1003: Upon re-reading / seeing this code, the idea that `LockablePartition` / `LockableChunk` are parts of the "public" server API is a little confusing. This PR didn't make it worse, I am just observing it doesn't seem ideal.

tests/end_to_end_cases/management_cli.rs
```diff
@@ -855,19 +856,15 @@ async fn test_close_partition_chunk() {
...
-        chunk_id: 0,
-    });
-
-    assert_eq!(
-        Some(expected_job),
-        stdout.metadata.job,
-        "operation was {:#?}",
-        stdout
-    );
+    match iox_operation.metadata.job {
```
alamb, line 859: I wonder if we want to remove the `close-chunk` CLI command too as it is superceded by compact-chunk? Perhaps we could do so as a follow on ticket

alamb (approved):
looks great @tustvold  👍  Nice work

server/src/lib.rs
tustvold: I think once we have a CatalogTransaction type this will be a lot better, the intent is that not all lifecycle actions are member functions of Db, but rather you perform a transaction on Db. However, as you rightly say these low-level types make for a bit of a confusing API

server/src/db/catalog/chunk.rs
tustvold: We use the word `compacting` for compacting two chunks together. Currently they cannot be persisted, but after #2665 it will be any chunks

query_tests/src/sql.rs
tustvold: > Maybe add order by on the columns you want to stay the same for us to verify the changes do not break anything?

I believe the SQL test case sorts the output already?

NGA-TRAN: > I believe the SQL test case sorts the output already?

The test does sort output for comparing but what I meant was for us (human) to see nothing was changed except column chunk ID

tustvold: Aah - I can't do this as the thing it was sorting by has now changed, and even if the SQL contained an order by, it would be changed by the test

alamb: Perhaps we could add an order by (a follow on PR is fine) to the query so that future changes are easier to understand.

NGA-TRAN: Yeah, this is just a suggestion for easier to review. Not a must

tustvold: I'm really confused, this **IS** sorted it is just the sort order has changed as the chunk ID is no longer present. Even if chunk ID were present the order would change as the chunk IDs themselves have changed.

This test calls assert_batches_sorted_eq which already sorts the lines of output alphabetically, adding a sort expression to the SQL query would change nothing about how the output is rendered

NGA-TRAN: 😄 😄 😄 😄 
What we meant was if the `expected` output is sorted, it will be easier for us to review. Yes, the order by won't help the test but if you copy the `actual` output that is sorted, it will help our eyes. The actual output of sql like:
`SELECT * from system.chunk_columns order by table_name, column_name, min_value, max_value`

NGA-TRAN: And sorry for the confusion 😄 😄 😄 😄 😄 😄

tustvold: So the assertion compares the sorted expected and sorted actual, so yes I could change the order arbitrarily as written in the codebase. However, should the results change, the test output would print the sorted representation so we'd end up rehashing this discussion when they copied that output into the test :laughing: 

I'm going to leave it for now, it shouldn't change again and this way avoids surprises when the test output is sorted differently from what is actually in the code as "expected"

PR #2682, opened 2021-09-30
WIP feat: Add UUID to databases

### LINKED ISSUE #2675: Floating databases: Change from database name/generation in paths to UUIDs
- For this issue, still leave databases in a server's subdirectory and still scan that directory on 
    server startup
  - For this issue, still use tombstone files to distinguish active from inactive databases
  - When a database is created, create a UUID and store the database using UUID instead of name+generation id. Write the UUID in `rules.pb`.
  - Remove the concept of generation ID and switch to UUID throughout, especially in deleted/detailed database list output and delete database API
  - `influxdb_iox database restore` (and corresponding gRPC API) will take a database UUID rather than a database name + generation ID
  - Handle duplicate name detection during startup and when creating a database:
    - If we've added active database `a` to the database list, then we find another active database with a different UUID but also with the name `a` in its rules.pb, what should happen?
      - Put both `a` databases in an error state

carols10cents:
Closes #2675

Not ready yet


server/src/rules.rs
```diff
@@ -53,13 +57,13 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
...
-/// can use `internal_types::database_rules::DatabaseRules` in
+/// can use `data_types::database_rules::PersistedDatabaseRules` in
 /// blissful ignorance of such subtlties
 #[derive(Debug, Clone)]
 pub struct ProvidedDatabaseRules {
     /// Full database rules, with all fields set. Derived from
-    /// `encoded` by applying default values.
-    rules: Arc<DatabaseRules>,
+    /// `original` by applying default values.
+    full: Arc<DatabaseRules>,
```
alamb, line 66: 👍  this is a nice change of field name

alamb (approved):
> I was going to try changing the paths in this PR, but I think it's too much.

I agree -- doing it in two PRs sounds like a better idea to me

> If this gets deployed, the databases will go into an error state because the rules won't decode... then conductor will just recreate the databases, right? Or do I need to bump a _VERSION number? I forget.


I believe you need to increment the database transaction and metadata versions, as @crepererum  did in https://github.com/influxdata/influxdb_iox/pull/2716

https://github.com/influxdata/influxdb_iox/pull/2716/files#diff-6729dc92836fb7435cfea9e6d8fc274fb9c00f222486ab47d5ac7682163cd2ecR42

https://github.com/influxdata/influxdb_iox/pull/2716/files#diff-7fcb89263a9051c85a33281493fa8ca977e409abc5a74f2f03f28475b530a4caR124

generated_types/protos/influxdata/iox/management/v1/database_rules.proto
```diff
@@ -149,6 +149,15 @@ message DatabaseRules {
   WriteBufferConnection write_buffer_connection = 13;
 }
 
+// The UUID and database rules stored in object storage for a database. Operator-facing APIs should
+// use `DatabaseRules` instead.
+message PersistedDatabaseRules {
+  // The UUID uniquely identifying this database
+  string uuid = 1;
```
mkmik, line 156: perhaps we should use bytes for uniformity with other places where we will use uuid in a more space constrained circumstances (chunk ids)

```suggestion
  bytes uuid = 1;
```

CC: @crepererum 

WDYT?

crepererum: I agree, also see #2732.

crepererum:
> I believe you need to increment the database transaction and metadata versions

I think that's technically not required here since this change will just result in the catalog / DB not being found. The paths within the catalog are already relative to the DB root (aka if we change the prefix, the paths stay stable).

carols10cents:
I tested creating a database in a server with the code on `main` and then switching to this branch; the database would end up in an error state *and* updating the rules would fail, which I think would break everything.

So I added a commit so that instead, if parsing the existing rules file on startup as the wrapper type with a UUID fails, try to parse it as the rules type without a UUID. Then, the next time the rules get updated, the rules file will get written out with a UUID.

This code doesn't actually read the UUID from the rules file at all-- when the UUID is in the databases' object store paths, it will be read from there, and this transition code will go away.

Even though there are 2 approvals on this, I feel like the last commit is *just* big enough of a change that I'm going to leave this open until my tomorrow morning for anyone to re-review if you'd like.

server/src/lib.rs
```diff
@@ -2102,6 +2103,57 @@ mod tests {
         assert!(Arc::ptr_eq(&err, &foo_database.init_error().unwrap()));
     }
 
+    #[tokio::test]
+    async fn init_without_uuid() {
```
domodwyer, line 2107: This test perfectly captures the properties I was coming to look at - nice stuff 👍

domodwyer:
> I'm going to save this somewhere.... maybe in conductor's docs? Is it already there, and did I miss it? Then I can look this up next time I forget instead of bugging you again

I'll be writing up some end-user docs over the next few (working) days, and then eventually some more in-depth architecture docs to help onboard people into the project 👍  I'll be sure you add you as a reviewer to both!

> ..rest..

Thanks! Super helpful 🎉

PR #2683, opened 2021-10-01
feat: Implement optional SOCKS5 client proxy

mkmik:
Implement (optional) socks5 proxy in the iox client library.

This allows a local IOx client (e.g. a development version of conductor) to talk to IOx servers
running in a local k8s cluster even if the pod networking is not directly routable.

Among other things, this offers a simpler way to develop on a mac and test conductor locally
without having to cross-compile it for linux and run it in a container (or using complicated things
like teleport).

Demo:

```console
$ kubectl krew install socks5-proxy
...
$ kubectl socks5-proxy &
...
$ ALL_PROXY=socks5://127.0.0.1:1080 ./target/debug/influxdb_iox -h http://10.1.0.7:8082 server get
132072
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,3 +195,4 @@ aws = ["object_store/aws"] # Optional AWS / S3 object store support
 # to pick either heappy or jemalloc_replacing_malloc feature at least until we figure out something better.
 jemalloc_replacing_malloc = ["tikv-jemalloc-sys"]
 
+socks5 = [ "influxdb_iox_client/socks5" ]
--- a/client_util/src/connection.rs
+++ b/client_util/src/connection.rs
@@ -1,10 +1,12 @@
 use crate::tower::{SetRequestHeadersLayer, SetRequestHeadersService};
 use http::header::HeaderName;
 use http::{uri::InvalidUri, HeaderValue, Uri};
+#[cfg(feature = "socks5")]
+use hyper_socks2::SocksConnector;
 use std::convert::TryInto;
 use std::time::Duration;
 use thiserror::Error;
-use tonic::transport::Endpoint;
+use tonic::transport::{Channel, Endpoint};
 
 /// The connection type used for clients
 pub type Connection = SetRequestHeadersService<tonic::transport::Channel>;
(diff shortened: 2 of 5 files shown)


mkmik:
```
#!/bin/bash -eo pipefail
cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 366 security advisories (from /usr/local/cargo/advisory-db)
    Scanning Cargo.lock for vulnerabilities (462 crate dependencies)
Crate:         nix
Version:       0.20.1
Warning:       unsound
Title:         Out-of-bounds write in nix::unistd::getgrouplist
Date:          2021-09-27
ID:            RUSTSEC-2021-0119
URL:           https://rustsec.org/advisories/RUSTSEC-2021-0119
Dependency tree: 
nix 0.20.1

Crate:         nix
Version:       0.22.1
Warning:       unsound
Title:         Out-of-bounds write in nix::unistd::getgrouplist
Date:          2021-09-27
ID:            RUSTSEC-2021-0119
URL:           https://rustsec.org/advisories/RUSTSEC-2021-0119
Dependency tree: 
nix 0.22.1

error: 2 denied warnings found!

CircleCI received exit code 1
```

is this advisory because of something that the new dependencies use?

mkmik:
I created a dummy PR and it fails in the same way: https://github.com/influxdata/influxdb_iox/pull/2684

mkmik:
cargo-audit will pass once we rebase this on top of #2685

Cargo.toml
```diff
@@ -195,3 +195,4 @@ aws = ["object_store/aws"] # Optional AWS / S3 object store support
 # to pick either heappy or jemalloc_replacing_malloc feature at least until we figure out something better.
 jemalloc_replacing_malloc = ["tikv-jemalloc-sys"]
 
+socks5 = [ "influxdb_iox_client/socks5" ]
```
tustvold, line 198: Why does this need to be exposed as a feature flag here? It doesn't seem to be used by the IOx binaries?

client_util/src/connection.rs
```diff
@@ -86,6 +88,32 @@ impl std::default::Default for Builder {
...
+#[cfg(not(feature = "socks5"))]
+async fn connect(endpoint: Endpoint) -> Result<Channel> {
+    Ok(endpoint.connect().await?)
+}
+
+#[cfg(feature = "socks5")]
+async fn connect(endpoint: Endpoint) -> Result<Channel> {
+    use std::str::FromStr;
+
+    Ok(match std::env::var("ALL_PROXY").ok() {
```
tustvold, line 100: It feels off to me that client_util would have a hidden environment variable that alters its behaviour, rather I would expect the client (e.g. conductor) that is initializing the connection to specify a custom connector or something... e.g. a `build_with_connector` method or something

tustvold (commented):
If I understand correctly, this is functionality for Conductor not IOx. I wonder if we might rejig client_util slightly to allow injecting a custom connector, i.e. allow Conductor to add socks5 functionality without it needing to be part of the IOx codebase?

Cargo.toml
mkmik: yeah; I used it to demo it; I can remove it

client_util/src/connection.rs
mkmik: yeah sounds good; will refactor

mkmik:
ok, laziness doesn't pay, will move the code to conductor

mkmik:
Moved to https://github.com/influxdata/influxdb_iox/pull/2686

PR #2686, opened 2021-10-01
feat: Allow iox clients to specify connector

mkmik:
This allows to implement custom transports for iox clients (such as socks5 proxies, see #2683).

### DIFF
--- a/client_util/src/connection.rs
+++ b/client_util/src/connection.rs
@@ -4,7 +4,8 @@ use http::{uri::InvalidUri, HeaderValue, Uri};
 use std::convert::TryInto;
 use std::time::Duration;
 use thiserror::Error;
-use tonic::transport::Endpoint;
+use tonic::transport::{Channel, Endpoint};
+use tower::make::MakeConnection;
 
 /// The connection type used for clients
 pub type Connection = SetRequestHeadersService<tonic::transport::Channel>;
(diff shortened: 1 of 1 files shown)


client_util/src/connection.rs
```diff
@@ -89,22 +90,44 @@ impl std::default::Default for Builder {
...
+    {
+        let endpoint = self.create_endpoint(dst)?;
+        let channel = endpoint.connect().await?;
+        Ok(self.compose_middleware(channel))
+    }
+
+    /// Construct the [`Connection`] instance using the specified base URL and custom connector.
+    pub async fn build_with_connector<D, C>(self, dst: D, connector: C) -> Result<Connection>
+    where
+        D: TryInto<Uri, Error = InvalidUri> + Send,
```
tustvold, line 104: For those following along these are the constraints from `tonic::Endpoint::connect_with_connector()`

tustvold (approved):

client_util/src/connection.rs
mkmik: fwiw, I could have added a `.with_connector` method to the builder and have a single `.build` method but the amount of generics that would require (I once did the same for trogging) was too much for me today. (My battery ran out after 2h of compiling rust)

tustvold: I do not blame you, and imo separate build methods isn't really a problem if it makes the builder easier to understand. After all `Endpoint` is a "builder" with multiple "build" methods :laughing:

Issue #2687, opened 2021-10-01
Delete Predicate With Empty Timestamp Range Deletes Nothing


tustvold:
```
    #[tokio::test]
    async fn delete_all() {
        let db = make_db().await.db;
        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=1 10").await;
        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=3 23").await;
        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=2 26").await;
        db.delete("cpu", Default::default()).await.unwrap();

        let batches = run_query(db, "select count(*) from cpu").await;
        let expected = vec![
            "+-----------------+",
            "| COUNT(UInt8(1)) |",
            "+-----------------+",
            "| 0               |",
            "+-----------------+",
        ];
        assert_batches_sorted_eq!(&expected, &batches);
    }
```

Fails with 

```
expected:

[
    "+-----------------+",
    "| COUNT(UInt8(1)) |",
    "+-----------------+",
    "| 0               |",
    "+-----------------+",
]
actual:

[
    "+-----------------+",
    "| COUNT(UInt8(1)) |",
    "+-----------------+",
    "| 3               |",
    "+-----------------+",
]
```

Initial investigation shows that:

* The delete predicate is associated with the chunk
* Adding additional criteria, e.g. table name, to the predicate doesn't change the behaviour

e-dard:
I'm guessing this needs to be handled as a "delete table" type operation somewhere.

tustvold:
> I'm guessing this needs to be handled as a "delete table" type operation somewhere.

I don't think it needs to be handled any differently from any other predicates, it needs to be associated with the relevant chunks (all of them) and left to compaction to sort out. A separate but related task should then be to make the query/compaction detect when a delete predicate "covers" a chunk and optimise that case

alamb:
Given that all deletes are supposed to have a timerange, I think the expected behavior in this case should be to reject the delete (aka raise an error on the call to `Db::delete()`.

I think sorting out how to drop all data in a database is a reasonable thing to do but applying no predicate is likely not the ideal way

tustvold:
How about a chat at standup?

tustvold:
So this fails

```
#[tokio::test]
    async fn delete_all() {
        use datafusion::prelude::*;
        let db = make_db().await.db;
[...]

PR #2700, opened 2021-10-01
feat: allow compaction and persistence to retun no chunk (#2664)

### LINKED ISSUE #2546: Not throw error when all of rows of a OS or RUB chunk are soft deleted

### LINKED ISSUE #2664: Compaction Return No Chunk
Currently lifecycle actions such as compact_chunks must return a DbChunk, specifically `Result<Arc<DbChunk>>>`. In order to resolve #2546 we wish for compaction to be able to return no chunk in the scenario when all rows have been deleted. The return type should therefore be changed to `Result<Option<Arc<DbChunk>>>>`

tustvold:
Mostly plumbing, this PR allows the persistence and compaction lifecycle actions to return no chunk in the event that all rows are deleted.

The way the test for the persist lifecycle action has to be written makes me very sad, I started fixing the plumbing to write this in a slightly less hacky way, but it's taking a bit longer than I had anticipated and I understand there is some urgency to getting this in, so hacky it is :sweat_smile: 

Closes #2664 
Closes #2546


server/src/db/lifecycle/persist.rs
```diff
@@ -265,4 +276,88 @@ mod tests {
...
+            .unwrap();
+
+        assert!(maybe_chunk.is_none());
+
+        let chunks: Vec<_> = partition.read().chunk_summaries().collect();
+        assert_eq!(chunks.len(), 1);
+        assert_eq!(chunks[0].storage, ChunkStorage::ReadBuffer);
+        assert_eq!(chunks[0].row_count, 1);
+
+        // Add a third set of writes one of which overlaps the above chunk
```
tustvold, line 322: If it doesn't overlap the persist action just ignores it :laughing:

query_tests/src/scenarios.rs
```diff
@@ -1101,15 +1108,13 @@ impl DbSetup for ChunkOrder {
...
         // transform chunk 0 into chunk 2 by persisting
-        let new_chunk = db
-            .persist_partition(
-                "cpu",
-                partition_key,
-                Instant::now() + Duration::from_secs(1),
-            )
-            .await
-            .unwrap();
-        assert!(lockable_chunk.id < new_chunk.id());
```
alamb, line 1112: is there a reason to stop comparing the chunk ids? I am just wondering as it seems unrelated to this PR  (most other changes add a new `unwrap()`

server/src/db/lifecycle/compact.rs
```diff
@@ -103,45 +105,46 @@ pub(crate) fn compact_chunks(
...
-        let throughput = (input_rows as u128 * 1_000_000_000) / elapsed.as_nanos();
-
-        info!(input_chunks=query_chunks.len(), rub_row_groups=rb_row_groups,
-                input_rows=input_rows, output_rows=guard.table_summary().total_count(),
-                sort_key=%key_str, compaction_took = ?elapsed, fut_execution_duration= ?fut_now.elapsed(),
-                rows_per_sec=?throughput,  "chunk(s) compacted");
-
-        Ok(DbChunk::snapshot(&guard))
+            None => {
+                info!(%addr, ?chunk_ids, "no rows to persist");
```
alamb, line 144:
```suggestion
                info!(%addr, ?chunk_ids, "no rows to persist, no chunk created");
```

server/src/db/lifecycle/compact.rs
```diff
@@ -103,45 +105,46 @@ pub(crate) fn compact_chunks(
...
-        // input rows per second
-        let throughput = (input_rows as u128 * 1_000_000_000) / elapsed.as_nanos();
-
-        info!(input_chunks=query_chunks.len(), rub_row_groups=rb_row_groups,
-                input_rows=input_rows, output_rows=guard.table_summary().total_count(),
-                sort_key=%key_str, compaction_took = ?elapsed, fut_execution_duration= ?fut_now.elapsed(),
-                rows_per_sec=?throughput,  "chunk(s) compacted");
-
-        Ok(DbChunk::snapshot(&guard))
+            None => {
```
alamb, line 143: It doesn't really matter, but one thought i had while reading this code is that you might be able to an extra level of nesting (and make the diff smaller) by using a construct like

```rust
let rb_chunk = match maybe_rb_chunk {
  Some(rb_chunk) => rb_chunk,
  None => {
    info!(%addr, ?chunk_ids, "no rows to persist, no chunk created");
    return
  }
};
```

server/src/db/lifecycle/compact.rs
```diff
@@ -217,4 +221,31 @@ mod tests {
...
+        let db = make_db().await.db;
+
+        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=1 10").await;
+        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=3 23").await;
+        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=2 26").await;
+
+        let partition_keys = db.partition_keys().unwrap();
+        assert_eq!(partition_keys.len(), 1);
+
+        // Cannot simply use empty predicate (#2687)
```
alamb, line 236: and now I see where #2687  came from lol

server/src/db/lifecycle/compact.rs
```diff
@@ -217,4 +221,31 @@ mod tests {
...
+            .build();
+
+        // Delete everything
+        db.delete("cpu", Arc::new(predicate)).await.unwrap();
+        let chunk = db
+            .compact_partition("cpu", partition_keys[0].as_str())
+            .await
+            .unwrap();
+
+        assert!(chunk.is_none());
```
alamb, line 249: 👍

server/src/db/lifecycle/persist.rs
```diff
@@ -132,27 +132,33 @@ where
...
                 );
             }
 
-            // NGA todo: we hit this error if there are rows but they are deleted
-            // Need to think a way to handle this (https://github.com/influxdata/influxdb_iox/issues/2546)
-            let to_persist = to_persist.expect("should be rows to persist");
+            let to_persist = match to_persist {
+                Some(to_persist) => to_persist,
+                None => {
+                    info!(%addr, ?chunk_ids, "no rows to persist");
```
alamb, line 158:
```suggestion
                    info!(%addr, ?chunk_ids, "no rows to persist, no chunks created");
```

server/src/db/lifecycle/persist.rs
```diff
@@ -132,27 +132,33 @@ where
...
+                    return Ok(None);
+                }
+            };
```
alamb, line 161: There are several other things that `write_chunk_to_object_store` does such as updating the persistence windows and marking the chunk as "written to object store" (aka `guard.set_written_to_object_store(parquet_chunk)`)"

I wonder what are the implications of skipping these steps if there were no rows? 

The tests below show that somehow the source chunks get dropped, but I do still wonder about the persistence windows

server/src/db/lifecycle/persist.rs
```diff
@@ -265,4 +276,88 @@ mod tests {
...
+        // Try to persist third set of writes
+        let maybe_chunk = db
+            .persist_partition("cpu", partition_key.as_str(), t2 + late_arrival)
+            .await
+            .unwrap();
+
+        assert!(maybe_chunk.is_none());
+
+        // The already persisted chunk should remain
+        let chunks: Vec<_> = partition.read().chunk_summaries().collect();
```
alamb, line 358: I see that the source ReadBuffer chunk is removed as expected, which is good. I am not 100% how this happens, but it seems to be the correct behavior

alamb (approved):
I am not sure if the way the persist logic is short-circuited leaves the persist windows in a bad state. The tests in this PR demonstrate that a chunk that is persisted but the persistence has no rows is dropped, which is the expected behavior but I did not expect given the code (meaning I don't fully understand this). 


Otherwise this PR looks great to me 👍

server/src/db/lifecycle/persist.rs
tustvold: This is a good point, when the flush handle is dropped it isn't actually flushing out the persistable window. The tests pass because there is still an additional row to persist so what gets "orphaned" does eventually get flushed, but this may not always be the case.

Should be a simple fix to commit the flush handle, but very good spot! 👍

server/src/db/lifecycle/compact.rs
NGA-TRAN: Nice

server/src/db/lifecycle/persist.rs
NGA-TRAN: Very nice. Thanks @tustvold

NGA-TRAN (approved):
This looks good to me except the delete persistent test. I think you may want to ask @crepererum to have a look on Monday to make sure we do not change anything unexpectedly.

server/src/db.rs
```diff
@@ -3647,10 +3676,6 @@ mod tests {
         let partition_key = "part_d";
         write_lp(&db, "cpu,part=d row=40,selector=0i 40").await;
         write_lp(&db, "cpu,part=d row=41,selector=1i 41").await;
-        db.compact_partition(table_name, partition_key)
-            .await
-            .unwrap();
-
```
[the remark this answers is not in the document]
tustvold: The persist_partition on the line below will persist from all unpersisted chunks, i.e. it is an implicit "compaction". The compaction here makes no different to the resulting distribution of chunks, all the unpersisted data will end up in a single ReadBufferAndObjectStore chunk regardless of if there is a preceding compact_partition call

server/src/db.rs
```diff
@@ -679,7 +679,7 @@ impl Db {
         partition_key: &str,
         now: Instant,
         f_parquet_creation_timestamp: F,
-    ) -> Result<Arc<DbChunk>>
+    ) -> Result<Option<Arc<DbChunk>>>
     where
         F: Fn() -> DateTime<Utc> + Send,
     {
```
crepererum, line 685: GitHub doesn't allow me to comment outside the change-set, but [this](https://github.com/influxdata/influxdb_iox/pull/2700/files#diff-45906f4b20ff30f9970a08127270631ed8d53922585aefa8625a436827b39fddR729-R735) `ensure!` macro should probably be replaced with `return None;` (+ a test).

crepererum (approved):
Looks good except for the missing case in `persist_partition_with_timestamp`.

query_tests/src/scenarios.rs
tustvold: I removed it as it was testing a constraint that isn't supposed to be part of the contract, the only guarantee is chunk IDs are unique, there order is arbitrary

server/src/db/lifecycle/compact.rs
tustvold: Yeah I did that initially but put it in the wrong place, before the chunks were dropped, so I rejigged. But perhaps removing the return was overkill - will change

server/src/db.rs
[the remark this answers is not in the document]
tustvold: I'll back out this change and raise a separate PR to simplify these tests

query_tests/src/scenarios.rs
alamb: I see -- makes sense.  I suppose with the introduction of `ChunkOrder` that the Id can/should be treated as arbitrary

server/src/db.rs
tustvold: Created #2712 as this is actually a broader issue

PR #2715, opened 2021-10-04
refactor: introduce `DeletePredicate`

crepererum:
`DeletePredicate` is a simpler version of `Predicate` that is based on
IOx `DeleteExpr` instead of the full-blown DataFusion `Expr`. This will
allow us to do a couple of things (in follow-up changes):

- Order and de-duplicate delete predicates
- Normalize predicates
- Infallible serialization
- Smaller memory footprint

Note that this change only affects delete expressions. Query expressions
that are supported via the API are not changed. The query subsystem also
still uses the full-featured expressions/predicates (delete
expressions/predicates are converted to the more powerful DataFusion
version on-the-fly).


predicate/src/delete_predicate.rs
```diff
@@ -0,0 +1,736 @@
+use std::{collections::BTreeSet, convert::TryInto};
```
crepererum, line 1: I took the opportunity to move mostly all delete-predicate-related code out of `predicate.rs` so the code is a bit easier to read. Tests are moved alongside.

predicate/src/delete_predicate.rs
```diff
@@ -0,0 +1,736 @@
...
+/// Result type for Parser Cient
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Represents a parsed delete predicate for evaluation by the InfluxDB IOx
+/// query engine.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DeletePredicate {
+    /// Optional table restriction. If present, restricts the results
+    /// to only tables whose names are in `table_names`
+    pub table_names: Option<BTreeSet<String>>,
```
e-dard, line 75: If we are thinking about this from the perspective of the InfluxData delete API then either one table is specified, or no tables are. If no tables are specified then the delete applies to all tables.

I think the work to translate the lack of a table name into all the table names is going to happen higher up and closer to the specific InfluxData storage API. So I would expect by the time you are in here one would know exactly how many tables need to be touched.

Also, I think it would make things simpler if you had one predicate per table, rather than a predicate spanning many tables. It would likely make deduplication of predicates simpler to reason about.

predicate/src/delete_predicate.rs
```diff
@@ -0,0 +1,736 @@
...
+
+    /// Optional field restriction. If present, restricts the results to only
+    /// tables which have *at least one* of the fields in field_columns.
+    pub field_columns: Option<BTreeSet<String>>,
```
e-dard, line 79: FWIW this is hopefully going away soon (from the InfluxData API I mean)

predicate/src/delete_predicate.rs
```diff
@@ -0,0 +1,736 @@
...
+
+    /// Optional partition key filter
+    pub partition_key: Option<String>,
+
+    /// Optional timestamp range: only rows within this range are included in
+    /// results. Other rows are excluded
+    pub range: Option<TimestampRange>,
```
e-dard, line 86: I think this is either non-optional (if the delete is from the InfluxData delete API) or probably it doesn't need to sit explicitly on the predicate type. I'm wondering if it should just be baked into `exprs` to be honest.

e-dard (commented):
If these types are specifically for the InfluxData delete API then I wonder if the code should live a bit closer to those specific APIs.

query_tests/src/scenarios/delete.rs
```diff
@@ -122,22 +141,37 @@ impl DbSetup for TwoDeletesMultiExprsOneChunk {
...
         // pred1: delete from cpu where 0 <= time < 32 and bar = 1 and foo = 'me'
-        let expr1 = col("bar").eq(lit(1f64));
-        let expr2 = col("foo").eq(lit("me"));
-        let pred1 = PredicateBuilder::new()
-            .table("cpu")
-            .timestamp_range(0, 32)
-            .add_expr(expr1)
-            .add_expr(expr2)
-            .build();
+        let pred1 = DeletePredicate {
```
alamb, line 144: I'll be honest I prefer the `PredicateBuilder` implementation as I think it is clearer about what is going on -- perhaps if we were able to reuse `Predicate` in some way we could avoid having to rewrite this

alamb (approved):
I am not a huge fan of another structure that is mostly the same as `Predicate` though I can see the rationale for encapsulating infallible deletes and other delete specific behavior.

Would you be willing to entertain using the existing predicate internally like 

```rust
struct DeletePredicate {
  inner: Predicate
}
```

With an appropriate `TryInto` implementation to convert from a `Predicate`?

All that being said, I don't see any reason not to merge this PR as is either.

query_tests/src/scenarios/delete.rs
crepererum: Good point, I can certainly try this :+1:

predicate/src/delete_predicate.rs
alamb: These fields are copied from `Predicate` -- hopefully we can reuse the existing `Predicate` structure and then we will have only a single representation to optimize

crepererum: The thing is: I think some of these fields will always be `None` when the `Predicate` is created via a delete request. At least looking at gRPC interface I can see no way many of these fields are actually used (`field_columns`, `table_names`). For now I'll just convert the `range` to a non-optional type and leave the remaining fields. We might do some further simplification in a follow-up.

query_tests/src/scenarios/delete.rs
crepererum: I'll just use a `Default` impl for now (so the structure construction can be shortened) since I think the `DeletePredicate` type might get simplified quite a bit which makes the constructor less code-intensive.

predicate/src/delete_predicate.rs
NGA-TRAN: Right. I do not think we need this at all. Actually it even introduced confusions

NGA-TRAN (approved):
Since we want to use DeleteExpr in Delete Predicate, this is a necessary PR for it. Everything looks good. I suspect the DeletePredicate struct will have fewer fields. We will mostly keep the `time range` and `exprs`

predicate/src/delete_predicate.rs
```diff
@@ -0,0 +1,736 @@
...
+    /// Optional table restriction. If present, restricts the results
+    /// to only tables whose names are in `table_names`
+    pub table_names: Option<BTreeSet<String>>,
+
+    /// Optional field restriction. If present, restricts the results to only
+    /// tables which have *at least one* of the fields in field_columns.
+    pub field_columns: Option<BTreeSet<String>>,
+
+    /// Optional partition key filter
+    pub partition_key: Option<String>,
```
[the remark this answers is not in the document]
crepererum: Ok, will remove that in a follow-up PR.

PR #2716, opened 2021-10-04
refactor: make `ChunkOrder` non-zero

crepererum:
This will make it easier to handle missing values.

Helps with #2633.


alamb (approved):
seems like a good change to me

PR #2719, opened 2021-10-04
feat: don't auto-increment background worker now

tustvold:
I intend to eventually remove this, but at least in the mean time the auto-incrementing behaviour is rather confusing. I've spent a too much time tracking down off by one nanosecond errors originating from this, and I think it is no longer necessary :sweat_smile:


crepererum (approved):
IIRC this was once required to pass the tests (I think the persistence window once wanted that), but if the tests pass with this change it's probably alright :+1:

PR #2721, opened 2021-10-04
test: turn on tests of delete all

NGA-TRAN:
Now #2546 is fixed by #2700, let turn its tests on

### DIFF
--- a/query_tests/src/sql.rs
+++ b/query_tests/src/sql.rs
@@ -1155,6 +1155,3 @@ async fn sql_select_with_three_deletes_from_three_chunks_with_select_predicate()
     )
     .await;
 }
-
-#[tokio::test]
-async fn delete_empty_predicate() {}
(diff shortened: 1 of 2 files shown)


e-dard (approved):

query_tests/src/sql.rs
```diff
@@ -1160,3 +1155,6 @@ async fn sql_select_with_three_deletes_from_three_chunks_with_select_predicate()
     )
     .await;
 }
+
+#[tokio::test]
+async fn delete_empty_predicate() {}
```
tustvold, line 1160: ?

NGA-TRAN: Hah, will remove. It was the test I found no longer necessary :)

NGA-TRAN: it is removed @tustvold

tustvold (approved):

PR #2723, opened 2021-10-04
refactor: remove unused fields from `DeletePredicate` and remove intermediate `ParseDeletePredicate`

crepererum:
This "just" deletes code or makes code private, no functional change.

### DIFF
--- a/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
+++ b/generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
@@ -3,15 +3,14 @@ package influxdata.iox.catalog.v1;
 
 // Represents a parsed predicate for evaluation by the InfluxDB IOx query engine.
 message Predicate {
-  // Optional table restriction. If present, restricts the results to only tables these tables.
-  OptionalStringSet table_names = 1;
+  // Was `table_names`.
+  reserved 1;
 
-  // Optional field restriction. If present, restricts the results to only tables which have *at least one* of the
-  // fields in field_columns.
-  OptionalStringSet field_columns = 2;
+  // Was `field_columns`.
+  reserved 2;
 
-  // Optional partition key filter
-  OptionalString partition_key = 3;
+  // Was `partition_key`.
+  reserved 3;
 
   // Optional timestamp range: only rows within this range are included in results. Other rows are excluded.
   TimestampRange range = 4;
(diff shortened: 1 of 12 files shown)


predicate/src/delete_predicate.rs
```diff
@@ -93,252 +82,207 @@ pub struct DeletePredicate {
     pub exprs: Vec<DeleteExpr>,
 }
 
+impl DeletePredicate {
+    /// Parse and convert the delete grpc API into ParseDeletePredicate to send to server
+    pub fn try_new(start: &str, stop: &str, predicate: &str) -> Result<Self> {
+        // parse and check time range
+        let (start_time, stop_time) = parse_time_range(start, stop)?;
```
NGA-TRAN, line 89: 👍

predicate/src/delete_predicate.rs
```diff
@@ -93,252 +82,207 @@ pub struct DeletePredicate {
...
+
+        // Parse the predicate
+        let delete_exprs = parse_predicate(predicate)?;
```
NGA-TRAN, line 92: 👍

NGA-TRAN (approved):
LGTM

generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
```diff
@@ -3,15 +3,14 @@ package influxdata.iox.catalog.v1;
 
 // Represents a parsed predicate for evaluation by the InfluxDB IOx query engine.
 message Predicate {
-  // Optional table restriction. If present, restricts the results to only tables these tables.
```
alamb, line 6: Since nothing has one of these things serialized, perhaps we could simply delete the fields and not worry about backwards compatibility / reserved fields.

Not that it is a problem, I just wanted to point it out

alamb (approved):
looks a lot nicer to me

generated_types/protos/influxdata/iox/catalog/v1/predicate.proto
crepererum: As a general rule I just tend to block old protobuf field IDs so they're not reused in the future, w/o evaluating if these could unnecessary in a specific case. It just safes me a bit of worrying.

alamb: makes sense 👍

PR #2725, opened 2021-10-04
fix: make stop time inclusive

NGA-TRAN:
As a part of #2396 to make stop time included in the delete time range

### DIFF
--- a/query_tests/src/scenarios/delete.rs
+++ b/query_tests/src/scenarios/delete.rs
@@ -37,7 +37,7 @@ impl DbSetup for OneDeleteSimpleExprOneChunkDeleteAll {
             table_names: Some(IntoIterator::into_iter(["cpu".to_string()]).collect()),
             field_columns: None,
             partition_key: None,
-            range: TimestampRange { start: 0, end: 25 },
+            range: TimestampRange { start: 10, end: 20 },
             exprs: vec![],
         };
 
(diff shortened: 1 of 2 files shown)


query_tests/src/scenarios/delete.rs
```diff
@@ -37,7 +37,7 @@ impl DbSetup for OneDeleteSimpleExprOneChunkDeleteAll {
             table_names: Some(IntoIterator::into_iter(["cpu".to_string()]).collect()),
             field_columns: None,
             partition_key: None,
-            range: TimestampRange { start: 0, end: 25 },
+            range: TimestampRange { start: 10, end: 20 },
```
NGA-TRAN, line 40: The purpose of the test is to delete data in [10,20] range. This change makes sure the fix works as no tests on any scenarios fail

query_tests/src/scenarios/delete.rs
```diff
@@ -98,7 +98,7 @@ impl DbSetup for OneDeleteMultiExprsOneChunk {
             table_names: Some(IntoIterator::into_iter(["cpu".to_string()]).collect()),
             field_columns: None,
             partition_key: None,
-            range: TimestampRange { start: 0, end: 32 },
+            range: TimestampRange { start: 0, end: 30 },
```
NGA-TRAN, line 101: delete range of the test: [0, 30]

jacobmarble (approved):

Issue #2717, opened 2021-10-04
Check in generated_types protobuf code


alamb:
__Use case:__
My editor (rust-analyzer under emacs LSP mode) can not find the definitions of the classes in `generated_types` crate

__Proposal:__
The idea would be that we would check in generated types and then add a CI test that ensures that the checked in code and the generated code remains in sync by re-generating the protos and comparing them

Checking in the generated code from protobuf is how  IDPE does it in golang. It also reduces the number of dependencies for developers most days. Oh and it is consistent with what we do for flatbuffers.

__Current behaviour:__
`cargo build` invokes prost to generate code which is then `include!`ed in rust

__Desired behaviour:__
1. A new command that compiles the proto files into rs code to somewhere in the source tree
2. Remove said compiling code from build.rs
3. Add a new CI test that rebuilds the proto files and errors if the generated files are not consistent with what is checked in

__Alternatives considered:__
NA (status quo)

alamb:
cc @tustvold  @e-dard

e-dard:
This is a no brainer to me.

alamb:
I started work on an initial PR to check in generated types: https://github.com/influxdata/influxdb_iox/pull/2727

I hope to keep hacking on it (figure out how to make the final output generated code without the `include!` macro)

crepererum:
> Oh and it is consistent with what we do for flatbuffers.

I would argue that our flatbuffer build system is somewhat a mess and don't accept this as a point of reference.

IMHO generated code that can easily be derived from the existing data in the repository should never be checked into GIT. It makes larger diff, calls for messy merge conflicts, more upgrade pain (e.g. right now the generated code is implicitly updated by dependabot as well), and irrelevant load for developers. If I read this issue correctly that's all just to workaround a bug (or missing feature) in rust-analyzer. I would rather work on the upstream issue instead.

And as someone who touched the protobuf specs at least two times a week over the last months, I'm not really looking forward to handle even more manual labor.

alamb:
Another theoretical concern of re-generating code each time is that there is some possibility of different generated code being used in different build environments (e.g. using a different `protoc` version).

That being said, I had hoped that with an hour of messing around -- https://github.com/influxdata/influxdb_iox/pull/2727 -- I would be able to fix one of my pet peeves working with the iox code base. However, sadly that does not appear to be the case, and realistically I probably shouldn't be spending any more time on this issue as I have other higher priority things to work on at the moment. 

[...]

PR #2727, opened 2021-10-04
refactor: Check in generated protobuf (WIP)

alamb:
refactor: Check in generated protobuf

Closes https://github.com/influxdata/influxdb_iox/issues/2717

Still to do:
- [ ] Create a PR to pbjson : to pick up output dir change on https://github.com/alamb/pbjson/commit/41c918445202fa9dd5f4b9f35e2b755e20680df2
- [ ] Figure out how to expand `include!` macros (as rust analyzer still doesn't do the right thing there) -- maybe we need a new templating language -- handlebars? https://crates.io/crates/handlebars 
- [ ] Add a CI check that generated files did not change

# Rationale
1. Make it easier for normal (non CLion) IDEs to find definitions for generated code
2. See other rationale on #2717

# Changes
1. A new command that compiles the proto files into rs code to somewhere in the source tree
2. Remove said compiling code from build.rs
3. Add a new CI test that rebuilds the proto files and errors if the generated files are not consistent with what is checked in

# Flow for updating types:
```shell
cd generated_types/build/ && cargo run
```


alamb:
I plan to park this PR for a while (see comments on https://github.com/influxdata/influxdb_iox/issues/2717#issuecomment-934253850)

PR #2730, opened 2021-10-05
refactor: delete predicate clean-ups

crepererum:
See commits.

### DIFF
--- a/parquet_file/src/catalog/interface.rs
+++ b/parquet_file/src/catalog/interface.rs
@@ -1,5 +1,8 @@
 //! Abstract interfaces to make different users work with the perserved catalog.
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
 
 use data_types::chunk_metadata::{ChunkAddr, ChunkId};
 use iox_object_store::{IoxObjectStore, ParquetFilePath};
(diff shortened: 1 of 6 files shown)


parquet_file/src/catalog/interface.rs
```diff
@@ -130,17 +133,15 @@ pub trait CatalogState {
...
 
-    /// List of active delete predicates together with their chunks (by table name, partition key, and chunk ID).
+    /// Maps active delete predicates to their chunks (by table name, partition key, and chunk ID).
     ///
     /// This must only contains chunks that are still present in the catalog. Predicates that do not have any chunks
     /// attached should be left out.
-    ///
-    /// The vector itself must be sorted by [`DeletePredicate`]. The chunks list must also be sorted.
-    pub delete_predicates: Vec<(Arc<DeletePredicate>, Vec<ChunkAddrWithoutDatabase>)>,
+    pub delete_predicates: HashMap<Arc<DeletePredicate>, HashSet<ChunkAddrWithoutDatabase>>,
```
NGA-TRAN, line 146: I like converting this into hashMap and HashSet. I wonder if the Hash function on DeletePredicate struct is nothing or will cost us something?

NGA-TRAN (approved):
Makes sense and looks good to me

parquet_file/src/catalog/interface.rs
crepererum: It's not for free, but less hacky. Beforehand we were using the address of the content of `Arc<DeletePredicate>` which somewhat can be used for de-duplication but is very hacky. If it turns out that this is a bottleneck, we can still optimize it.

PR #2732, opened 2021-10-05
refactor: use protobuf bytes for transaction UUIDs

crepererum:
It's slightly more efficient and I plan to use the same format for UUID-based chunk IDs.

### DIFF
--- a/parquet_file/src/catalog/core.rs
+++ b/parquet_file/src/catalog/core.rs
@@ -39,7 +39,7 @@ pub use crate::catalog::internals::proto_parse::Error as ProtoParseError;
 /// Current version for serialized transactions.
 ///
 /// For breaking changes, this will change.
-pub const TRANSACTION_VERSION: u32 = 16;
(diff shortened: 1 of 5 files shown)


parquet_file/src/catalog/core.rs
```diff
@@ -492,15 +492,18 @@ impl OpenTransaction {
...
+            ),
+            None => (0, Bytes::new()),
         };
 
         Self {
             proto: proto::Transaction {
                 actions: vec![],
                 version: TRANSACTION_VERSION,
-                uuid: uuid.to_string(),
+                uuid: uuid.as_bytes().to_vec().into(),
```
tustvold, line 506: Was going to suggest https://docs.rs/bytes/1.1.0/bytes/struct.Bytes.html#method.copy_from_slice but then realised it literally just calls `to_vec` - https://docs.rs/bytes/1.1.0/src/bytes/bytes.rs.html#202-204 :laughing:

tustvold (approved):

PR #2739, opened 2021-10-05
refactor: make data_types optional dependency

tustvold:
`data_types` is not a particularly lightweight dependency, and brings in the line protocol parser among other things - see https://github.com/influxdata/influxdb_iox/pull/2737. The client crates do not need any of this and so by adding a feature flag on `generated_types` we can avoid them bringing this code in where it isn't needed


domodwyer (approved):

generated_types/src/lib.rs
```diff
@@ -140,11 +140,17 @@ pub fn protobuf_type_url_eq(url: &str, protobuf_type: &str) -> bool {
 pub use com::github::influxdata::idpe::storage::read::*;
 pub use influxdata::platform::storage::*;
 
+pub mod google;
+
+#[cfg(feature = "data_types_conversions")]
```
alamb, line 145: 👍  this is cool

Issue #2722, opened 2021-10-04
Add TimeProvider Abstraction


tustvold:
**Problem**

Currently within IOx we have three types of timestamp:

* Row timestamp - nanoseconds since epoch as `i64`
* Write timestamp - `chrono::DateTime<chrono::Utc>`
* Write instant - `std::time::Instant`

Additionally for the purposes of tests we do the following

* Plumb various forms of timestamp, or closures evaluating to timestamps, through method calls
* Have a somewhat strange `Mutex<Instant>` on `Db` called `background_worker_now_override`

This is rather confusing, and makes writing tests an exercise in frustration because some method might only plumb one or none of the timestamps through to the call site you are invoking and not the one you want, and so you end up spending inordinate amounts of time plumbing timestamps all over the place.

To add insult to injury:

* `std::time::Instant` cannot be directly formatted and appears to have differing resolution on aarch64-darwin
* `chrono::Duration` and `std::time::Duration` are not equivalent and so you end up juggling both in the same scope

**Proposal**

Introduce a custom `Time` struct within a new `time` crate. The internal representation is an implementation detail (probably u64 seconds + u32 nanos) but it should:

* Be convertible to nanoseconds since epoch
* Display as an rfc3339 date
* Use `std::time::Duration` for arithmetic

Secondly introduce a `TimeProvider` trait, with a single method

```
trait TimeProvider {
  fn now(&self) -> Time
}
```

A compliant `TimeProvider` should

* Guarantee returned `Time` are monotonically increasing
* Make best effort to keep ticking in the event of system clock change

Two implementations will be provided:

* `InstantProvider`: using `std::time::Instant` and performing an offset calculation like `to_approximate_datetime`
* `FixedProvider`: manually specified timestamps for tests

Systems that need timestamps will accept `Box<dyn TimeProvider>` to provide this.

**Alternatives Considered**

An earlier draft proposed relaxing the constraints on `PersistenceWindows` and using `chrono::Utc` this was discounted because:

* The guarantees around discontinuities are weaker, e.g. CLOCK_MONOTONIC will keep ticking if the system time changes, but if monotonic-ifying CLOCK_REALTIME by taking max of returned values, this would stop ticking if time traveled backwards
* Requires crates to have a dependency on chrono and juggle overloaded `Duration` types
* Doesn't eliminate the need for all the timestamp plumbing

NGA-TRAN:
[...]

Issue #2724, opened 2021-10-04
Use influxdata.pbdata.v1.TableBatch in Write Path


tustvold:
**Problem**

Currently the write path is predominantly in terms of the flatbuffer `influxdata.iox.write.v1.Entry` and this acts as a major complication within the write path.

The following are things that would be eliminated/fixed by using influxdata.pbdata.v1.TableBatch

* Translation from protobuf to Entry - [here](https://github.com/influxdata/influxdb_iox/blob/main/entry/src/entry.rs#L304)
* Translation from LP to Entry (although replaced by LP to PB)- [here](https://github.com/influxdata/influxdb_iox/blob/main/entry/src/entry.rs#L83)
* Custom Entry wrapper involving:
    * ouroborous wizardry - [here](https://github.com/influxdata/influxdb_iox/blob/main/entry/src/entry.rs#L737)
    * wild iterators - [here](https://github.com/influxdata/influxdb_iox/blob/main/entry/src/entry.rs#L768)
* Complexity due to encoded nature of Entry making reconstruction expensive
    * Write masks due to cost of reconstructing Entry - [here](https://github.com/influxdata/influxdb_iox/pull/2258/files)
    * Up-front sharding, partitioning, etc... - [here](https://github.com/influxdata/influxdb_iox/blob/main/entry/src/entry.rs#L83)
    * Splitting into separate partition writes - [here](https://github.com/influxdata/influxdb_iox/pull/2497/files#diff-340c5e5ddcbd504a868fad33d8b85bf2a67fe27c670fce22c6c2f607edd869e6R310)
* Lack of deep-copy support - https://github.com/google/flatbuffers/issues/264
* Lack of human-readable serialization for debugging - (only C++ flatbuffers supports JSON)
* https://github.com/influxdata/influxdb_iox/issues/2139
* https://github.com/influxdata/influxdb_iox/issues/2534
* https://github.com/influxdata/influxdb_iox/issues/2243

Additionally it is worth considering the following:

* Ingest is unlikely to be the performance bottleneck given it is `O(n)` and compaction/persistence are `O(nlog(n))`
* Flatbuffer verification is expensive, eliminating a number of its advantages, and necessary to avoid UB with foreign input
* The binary write format is:
    * Columnar and not highly nested, so decoding overheads should be limited by this
    * Supposed to be the performant ingest, converting from it to another representation somewhat undermines this?
    * An external facing API allowing for interesting use-cases interacting directly with the write buffer
[...]

PR #2743, opened 2021-10-05
refactor: use DateTime<Utc> in PersistenceWindows (#2722)

tustvold:
This is the first part of #2722 and migrates the PersistenceWindows to use `DateTime<Utc>`. 

To accommodate the lack of monotonicity inherent in `DateTime<Utc>` the PersistenceWindows take the maximum of the provided DateTime<Utc> and the maximum seen so far in calls to add_range. This does represent a regression in our ability to tolerate clock discontinuities, in particular if the system time goes backwards the persistence lifecycle will stall until time catches up. 

At least on Linux I think this can only occur if an administrator calls `clock_settime`, as NTP will use adjtime to smear the time correction. It is possible that leap seconds might cause the time to go back a second, but stalling the lifecycle for a second is unlikely to matter.

Originally I had hoped that it would be possible to construct a continuous clock that was printable using a combination of `std::time::Instant` and `Utc::now`. However, at least currently `std::time::Instant` doesn't keep ticking through suspend on all platforms and so this construction would drift - https://github.com/rust-lang/rust/issues/79462#issuecomment-734885062.

Using a single, printable clock at least imo outweighs the benefits of potential resilience to an unusual and untested edge case.


server/src/db.rs
```diff
@@ -663,31 +665,17 @@ impl Db {
 
     /// Persist given partition.
     ///
+    /// If `force` is `true` will persist all unpersisted data regardless of arrival time
```
tustvold, line 668: I opted to remove all timestamps from `persist_partition`, instead:

Callers wishing to persist everything should set force to true

Callers wishing to override the timestamp of the created chunks, or used to determine the writes to persist, should use the now_override (soon to be replaced with TimeProvider)

server/src/db.rs
```diff
@@ -700,7 +688,10 @@ impl Db {
             // get flush handle
             let flush_handle = partition
                 .persistence_windows_mut()
-                .map(|window| window.flush_handle(now))
+                .map(|window| match force {
+                    true => window.flush_all_handle(),
```
tustvold, line 692: I oscillated between special casing the persist all, and just having callers call window.flush_handle with chrono::MAX_DATETIME. I think a separate method is clearer and so I went with that

server/src/db.rs
```diff
@@ -838,7 +824,7 @@ impl Db {
                     .expect("lifecycle policy should be initialized");
 
                 policy
-                    .check_for_work(Utc::now(), self.background_worker_now())
+                    .check_for_work(self.utc_now(), Instant::now())
```
tustvold, line 827: This is an example of where mixed usage of `Instant` and `DateTime` can be very confusing. We currently only override one of these, so details about what type of time some implementation is using leak out. I spent a while chasing down a replay bug that boiled down to compaction triggering where it previously didn't because the mocked timestamps were larger...

query_tests/src/scenarios.rs
```diff
@@ -1094,11 +1053,8 @@ impl DbSetup for ChunkOrder {
             let partition = partition.read();
             let chunks = LockablePartition::chunks(&partition);
             let mut partition = partition.upgrade();
-            let flush_handle = LockablePartition::prepare_persist(
-                &mut partition,
-                Instant::now() + Duration::from_secs(1),
-            )
-            .unwrap();
+            let flush_handle =
+                LockablePartition::prepare_persist(&mut partition, chrono::MAX_DATETIME).unwrap();
```
tustvold, line 1057: I'm not a fan of the `chrono::MAX_DATETIME`, I hope to address this with further PRs which will remove timestamps from this method, like `Db::persist_partition`

persistence_windows/src/persistence_windows.rs
```diff
@@ -119,15 +119,17 @@ impl PersistenceWindows {
...
+            time_of_first_write: now,
+            time_of_last_write: now,
             max_sequence_numbers: Default::default(),
         }
     }
 
     /// Updates the late arrival period of this `PersistenceWindows` instance
-    pub fn set_late_arrival_period(&mut self, late_arrival_period: Duration) {
-        self.closed_window_period = late_arrival_period.min(DEFAULT_CLOSED_WINDOW_PERIOD);
+    pub fn set_late_arrival_period(&mut self, late_arrival_period: StdDuration) {
```
tustvold, line 129: #2722 should address the unfortunate mishmash of Duration types and just use `std::time::Duration`

lifecycle/src/policy.rs
```diff
@@ -1619,7 +1613,8 @@ mod tests {
             max_active_compactions: MaxActiveCompactions(NonZeroU32::new(10).unwrap()),
             ..Default::default()
         };
-        let now = Instant::now();
+        let now = from_secs(0);
```
alamb, line 1616: Is the goal eventually to remove the other uses of `Instant` here as well?

lifecycle/src/policy.rs
```diff
@@ -1112,7 +1106,7 @@ mod tests {
     }
 
     fn from_secs(secs: i64) -> DateTime<Utc> {
-        DateTime::from_utc(chrono::NaiveDateTime::from_timestamp(secs, 0), Utc)
+        Utc.timestamp(secs, 0)
```
alamb, line 1109: so much nicer -- it is too bad this form is not so obvious (to me) from the chrono documentation

persistence_windows/src/persistence_windows.rs
```diff
@@ -148,42 +150,35 @@ impl PersistenceWindows {
...
+        // DateTime<Utc> is not monotonic
+        let time_of_write = self.time_of_last_write.max(time_of_write);
         assert!(
             min_time <= max_time,
             "PersistenceWindows::add_range called with min_time ({}) > max_time ({})",
             min_time,
             max_time
         );
-        self.last_instant = received_at;
+        self.time_of_last_write = time_of_write;
```
alamb, line 181: 👍  that is a nice readability change I think

persistence_windows/src/persistence_windows.rs
```diff
@@ -555,38 +552,29 @@ mod tests {
...
                 table_name: Arc::from("table_name"),
                 partition_key: Arc::from("partition_key"),
             },
             late_arrival_period,
-            Instant::now(),
+            Utc::now(),
         )
     }
 
-    // *NOTE*: this test currently fails on (at least) aarch64 architectures
```
alamb, line 570: I wonder if we should ask @e-dard  to try this (newly rewritten) test on aarch64?

persistence_windows/src/persistence_windows.rs
```diff
@@ -601,62 +589,69 @@ mod tests {
...
     }
 
     #[test]
     fn starts_open_window() {
-        let mut w = make_windows(Duration::from_secs(60));
+        let mut w = make_windows(StdDuration::from_secs(60));
 
-        let i = Instant::now();
-        let start_time = Utc::now();
+        let row_t0 = Utc::now();
```
alamb, line 608: This is a nice change to avoid potential problems if successive calls to `Utc::now()` return the same value 👍

server/src/db.rs
```diff
@@ -663,31 +665,17 @@ impl Db {
...
     /// Errors if there is nothing to persist at the moment as per the lifecycle rules. If successful it returns the
     /// chunk that contains the persisted data.
     ///
-    /// The `now` timestamp should normally be `Instant::now()` but can be altered for testing.
     pub async fn persist_partition(
         self: &Arc<Self>,
         table_name: &str,
         partition_key: &str,
-        now: Instant,
+        force: bool,
```
alamb, line 677: 👍

query_tests/src/scenarios.rs
alamb: why did you need to change it to `MAX_DATETIME` rahter than using the existing logic like `now() + Duration::seconds(1)`?

server/src/db/lifecycle/persist.rs
```diff
@@ -19,18 +19,14 @@ use tracker::{TaskTracker, TrackedFuture, TrackedFutureExt};
 /// Split and then persist the provided chunks
 ///
 /// TODO: Replace low-level locks with transaction object
-pub fn persist_chunks<F>(
+pub fn persist_chunks(
```
alamb, line 22: is it worth explaining the use of `flush_handle` here?

alamb (approved):
I think it looks like a (very nice) improvement to me. 

I think @crepererum  may want to check out the change to `persist_partition` but I really like it over all 👍

crepererum (approved):

lifecycle/src/policy.rs
tustvold: Yes

tustvold:
Oops it would appear I accidentally linked a number of PRs to this instead of #2724

PR #2747, opened 2021-10-06
refactor: make `ChunkId` a UUID

crepererum:
**I originally planned to split this change into two (moving the type to UUIDs, introduce random UUIDs instead of linear counters). However this PR is the FULL change set, because along the way I've also removed the UUIDs from the parquet paths and already fixed most of the tests that assumed linear IDs.**

Reasons:

1. Helps w/ #2633 
2. Future-proof chunks IDs: we likely want to support catalog merges in the future (to re-configure sharding) and then chunks from multiple catalogs must have different IDs. UUIDs are an elegant way to archive that
3. Cleans up a lot of tests that are silently assuming chunk IDs start at 0 (which is an abstraction layer leak)
4. Removes the somewhat hacky UUID from our parquet files, because new chunk IDs can no longer collide w/ existing chunk IDs, hence we're never overwriting existing files
5. Manual catalog merges (copy + paste files from multiple catalogs) should "just work" using the catalog  rebuild, assuming you're not caring too much about the chunk ordering
6. As a small side-effect (that isn't used yet), chunk IDs are globally unique so in theory you could use the ID alone to address a chunk. However we don't have a global lookup table for that, but at least it prevents you from silly CLI typos (like you've meant to drop a chunk for `table1` but ended up dropping a chunk from `table2`).
7. IDs aren't really ordered anyways due to compaction. That's why we've introduced the chunk order a while ago (multiple chunks can have the same order, but as long as they originate from the same DB, they won't overlap in that case). So making assumption about ordering is somewhat fragile.

For low-level tests, `ChunkId::new_test` can be used to construct IDs from integers (this is sometimes helpful to make tests easier to read, see lifecycle tests for example). For high-level tests (e.g. tests that query the DB, use the CLI etc.), you must either work w/o chunk IDs (that's possible for most operations) or you need to retrieve the list of chunks before performing an action.

For IO, we use:

- **protobuf:** `bytes`
- **logging:** UUIDs emitted as strings (e.g. `3e5a1adf-3f17-4f93-8063-6fd6e09b2805`)
- **SQL:** same as logging

### DIFF
--- a/data_types/src/chunk_metadata.rs
+++ b/data_types/src/chunk_metadata.rs
@@ -206,14 +206,21 @@ impl ChunkSummary {
 pub struct ChunkId(Uuid);
 
 impl ChunkId {
-    pub fn new_random() -> Self {
+    /// Create new, random ID.
+    #[allow(clippy::new_without_default)] // `new` creates non-deterministic result
+    pub fn new() -> Self {
         Self(Uuid::new_v4())
     }
 
+    /// **TESTING ONLY:** Create new ID from integer.
+    ///
+    /// Since this can easily lead to ID collissions (which in turn can lead to panics), this must only be used for
+    /// testing purposes!
     pub fn new_test(id: u128) -> Self {
         Self(Uuid::from_u128(id))
     }
 
+    /// Get inner UUID.
     pub fn get(&self) -> Uuid {
         self.0
     }
--- a/data_types/src/chunk_metadata.rs
+++ b/data_types/src/chunk_metadata.rs
@@ -1,11 +1,16 @@
 //! Module contains a representation of chunk metadata
-use crate::partition_metadata::PartitionAddr;
+use std::{convert::TryFrom, num::NonZeroU32, sync::Arc};
+
+use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
-use std::{num::NonZeroU32, sync::Arc};
+use snafu::{ResultExt, Snafu};
+use uuid::Uuid;
+
+use crate::partition_metadata::PartitionAddr;
 
 /// Address of the chunk within the catalog
-#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Eq, PartialEq, PartialOrd, Ord, Serialize, Deserialize)]
 pub struct ChunkAddr {
     /// Database name
     pub db_name: Arc<str>,
@@ -183,10 +188,9 @@ pub struct DetailedChunkSummary {
 }
 
 impl ChunkSummary {
-    pub fn equal_without_timestamps(&self, other: &Self) -> bool {
+    pub fn equal_without_timestamps_and_ids(&self, other: &Self) -> bool {
         self.partition_key == other.partition_key
             && self.table_name == other.table_name
-            && self.id == other.id
             && self.storage == other.storage
             && self.lifecycle_action == other.lifecycle_action
             && self.memory_bytes == other.memory_bytes
@@ -199,31 +203,51 @@ impl ChunkSummary {
 ///
 /// This ID is unique within a single partition.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
-pub struct ChunkId(u32);
+pub struct ChunkId(Uuid);
 
 impl ChunkId {
-    pub const MAX: Self = Self(u32::MAX);
-
-    pub fn new(id: u32) -> Self {
-        Self(id)
+    pub fn new_random() -> Self {
+        Self(Uuid::new_v4())
     }
 
-    pub fn get(&self) -> u32 {
+    pub fn new_test(id: u128) -> Self {
+        Self(Uuid::from_u128(id))
+    }
+
+    pub fn get(&self) -> Uuid {
         self.0
-    }
-
-    /// Get next chunk ID.
-    ///
-    /// # Panic
-    /// Panics if `self` is already [max](Self::MAX).
-    pub fn next(&self) -> Self {
-        Self(self.0.checked_add(1).expect("chunk ID overflow"))
     }
 }
 
 impl std::fmt::Display for ChunkId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_tuple("ChunkId").field(&self.0).finish()
+    }
+}
+
+impl From<ChunkId> for Bytes {
+    fn from(id: ChunkId) -> Self {
+        id.get().as_bytes().to_vec().into()
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum BytesToChunkIdError {
+    #[snafu(display("Cannot convert bytes to chunk ID: {}", source))]
+    CannotConvertBytes { source: uuid::Error },
+}
+
+impl TryFrom<Bytes> for ChunkId {
+    type Error = BytesToChunkIdError;
+
+    fn try_from(value: Bytes) -> Result<Self, Self::Error> {
+        Ok(Self(Uuid::from_slice(&value).context(CannotConvertBytes)?))
+    }
+}
+
+impl From<Uuid> for ChunkId {
+    fn from(uuid: Uuid) -> Self {
+        Self(uuid)
     }
 }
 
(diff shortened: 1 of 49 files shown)


data_types/src/chunk_metadata.rs
```diff
@@ -227,6 +225,32 @@ impl std::fmt::Display for ChunkId {
...
 }
 
+impl From<ChunkId> for Bytes {
+    fn from(id: ChunkId) -> Self {
+        id.get().as_bytes().to_vec().into()
+    }
+}
+
+#[derive(Debug, Snafu)]
+pub enum BytesToChunkIdError {
```
e-dard, line 235: What do you think about moving this to the top of the file? I think it's good to keep errors together / in the same places.

generated_types/src/chunk.rs
```diff
@@ -199,7 +200,7 @@ mod test {
         let expected = ChunkSummary {
             partition_key: Arc::from("foo"),
             table_name: Arc::from("bar"),
-            id: ChunkId::new(42),
+            id: ChunkId::new_test(42),
```
e-dard, line 203: `ChunkId::from(42)` would sit better here with the other field initialisers.

query_tests/src/sql.rs
```diff
@@ -328,16 +328,16 @@ async fn sql_select_from_system_chunks() {
     //  test timestamps, etc)
 
     let expected = vec![
-        "+----+---------------+------------+-------------------+--------------+-----------+",
-        "| id | partition_key | table_name | storage           | memory_bytes | row_count |",
-        "+----+---------------+------------+-------------------+--------------+-----------+",
-        "| 0  | 1970-01-01T00 | h2o        | OpenMutableBuffer | 1639         | 3         |",
-        "| 0  | 1970-01-01T00 | o2         | OpenMutableBuffer | 1635         | 2         |",
-        "+----+---------------+------------+-------------------+--------------+-----------+",
+        "+---------------+------------+-------------------+--------------+-----------+",
```
e-dard, line 331: Why are we no longer exposing the Chunk IDs?

server/src/db/catalog.rs
```diff
@@ -552,4 +547,12 @@ mod tests {
     fn make_set(s: impl Into<String>) -> BTreeSet<String> {
         std::iter::once(s.into()).collect()
     }
+
+    fn sorted<T>(mut v: Vec<T>) -> Vec<T>
```
e-dard, line 551: nit: I would expect `sorted` to be a predicate function returning a `bool`. Could this be called `to_sorted` perhaps or `as_sorted`?

tests/end_to_end_cases/management_api.rs
```diff
@@ -525,7 +526,7 @@ async fn test_chunk_get() {
         Chunk {
             partition_key: "cpu".into(),
             table_name: "cpu".into(),
-            id: 0,
+            id: Uuid::from_u128(0).as_bytes().to_vec().into(),
```
e-dard, line 529: I think this might be a lot nicer if you implemented the suggestion of `From<u128> for ChunkId`.

tests/end_to_end_cases/system_tables.rs
```diff
@@ -48,11 +51,11 @@ async fn test_operations() {
     // parameterize on db_name1
 
     let expected_read_data = vec![
-        "+-----------+---------+---------------------------------+",
-        "| chunk_ids | status  | description                     |",
-        "+-----------+---------+---------------------------------+",
-        "| 0         | Success | Compacting chunks to ReadBuffer |",
-        "+-----------+---------+---------------------------------+",
+        "+---------+---------------------------------+",
```
e-dard, line 54: Again, why are we not exposing chunk IDs to the operator?

data_types/src/chunk_metadata.rs
crepererum: I'm more a fan of keeping the errors to where they are used TBH.

query_tests/src/sql.rs
crepererum: We are, but validating that they are "constant" (aka not random) is not a concern of this tests. This test is an integration-wire-up test. The exact semantics of this table are tested in `server::db::system_tables::chunks`.

tests/end_to_end_cases/system_tables.rs
crepererum: Same reason: wrong test layer. This is a wiring test and it's just easier to not query the IDs here. The actual semantics of this table is tested in `server::db::system_tables::operations`.

crepererum:
> The main design question I have is around the format of UUID. I would assume we will use `v4` UUIDs. If that is the case then I would like to see more validation that the UUIDs that are parsed from bytes are correct `v4` IDs.

1. for testing, we don't use UUID v4 but a somewhat hacky integer (I could pretend them to be a V4 as well, but that's even more of a lie). And testing also contains a bunch of IO and deserialization.
2. at the moment we use V4, however we might wanna change that in the future (e.g. to have IDs somewhat bound to the time and node or to use a totally different mechanism). Loading V4 IDs should still work in that case.

In general, I don't see a reason to validate the format here. What system came up w/ the UUID probably had a good reason to do so.

generated_types/src/chunk.rs
crepererum: As pointed out above: the conversion from integers should only be used for testing, hence I would like to avoid the official conversion.

server/src/db/catalog.rs
crepererum: done

tests/end_to_end_cases/management_api.rs
crepererum: done (w/ a different conversion path though)

crepererum:
> If we moved from `v4` UUIDs to another version are you suggesting we would still use the exact same bytes field/message to produce the IDs in the new format? If yes then how would we version the change between the ID formats?

I would use the same field, but I would assume that we stick to UUID as a container format but at most change the version of the UUID. UUIDs are actually versioned (using the first few bits, IIRC).

data_types/src/chunk_metadata.rs
```diff
@@ -199,25 +203,19 @@ impl ChunkSummary {
...
-
-    pub fn new(id: u32) -> Self {
-        Self(id)
+    pub fn new_random() -> Self {
+        Self(Uuid::new_v4())
     }
 
-    pub fn get(&self) -> u32 {
-        self.0
+    pub fn new_test(id: u128) -> Self {
```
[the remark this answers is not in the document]
alamb: it might help to add a rationale about *why* this shouldn't be used except for tests in the docstrings (so that future readers don't choose to ignore the advice thinking they know better)

generated_types/protos/influxdata/iox/catalog/v1/catalog.proto
```diff
@@ -56,8 +56,13 @@ message ChunkAddr {
   // Partition key.
   string partition_key = 2;
 
+  // Was uint32-based chunk ID.
```
alamb, line 59: 👍

iox_object_store/src/paths/parquet_file.rs
```diff
@@ -236,36 +196,19 @@ mod tests {
             result
         );
 
-        df.set_file_name("3");
```
alamb, line 239: why are these tests removed?

parquet_file/src/catalog/test_helpers.rs
```diff
@@ -452,16 +445,8 @@ where
             CatalogStateAddError::ParquetFileAlreadyExists { .. }
         ));
 
-        // does not exist - as different UUID
```
alamb, line 455: why is this thest removed?

query_tests/src/sql.rs
alamb: Or put another way "it is still possible to get chunk ids from system tables, it just isn't included in the test for system table as they are random numbers and would have to be normalized

The conversion is tested in https://github.com/influxdata/influxdb_iox/pull/2747/files#diff-976529f82a1a03f79eb75047b641057d14d306c8be74815402dcd7fec2ed9e93R203 so I think coverage is good enough for now

alamb (approved):

parquet_file/src/catalog/test_helpers.rs
crepererum: See me comment on `iox_object_store/src/paths/parquet_file.rs`: the additional UUID no longer exists, so same chunk ID now leads to the same file name, however chunk IDs can no longer collide.

crepererum:
Merging that now due to the high likelyhood of merge conflicts, if we really want to introduce UUID-type validation in the parser part, we can easily do that w/o breaking anything.

iox_object_store/src/paths/parquet_file.rs
alamb: Makes sense -- thank you for the clarification

Issue #2696, opened 2021-10-01
Error reading record batch while converting from SeriesSet: InvalidArgumentError("Sort requires at least one column")


alamb:
The `range_nsecs_group_count` test from https://github.com/influxdata/flux/blob/master/stdlib/universe/range_nsecs_test.flux is failing when run against IOx

The ttest
```
testcase range_nsecs_group_count {
    input = array.from(
        rows: [
            {_time: 2021-01-01T00:00:01.000000001Z, _field: "foo", _value: 1.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000002Z, _field: "foo", _value: 2.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value: 3.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000024Z, _field: "foo", _value: 11.24, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000009Z, _field: "bar", _value: 4.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000011Z, _field: "bar", _value: 5.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000015Z, _field: "bar", _value: 6.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000022Z, _field: "bar", _value: 1.2, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000031Z, _field: "bar", _value: 0.28, _measurement: "mm"},
        ],
    )
        |> group(columns: ["_field", "_measurement"])
        |> testing.load()
    want = array.from(
        rows: [
            {_value: 4},
        ],
    )
        |> group(columns: ["_field", "_measurement"])
    got = input
        |> range(start: 2021-01-01T00:00:01.000000001Z, stop: 2021-01-01T00:00:01.000000011Z)
        |> group()
        |> count()
        |> drop(columns: ["_start", "_stop"])

    testing.diff(got, want) |> yield()
}

```

The failure:
```
range_nsecs_group_count...fail: query error: rpc error: code = InvalidArgument desc = Error running grouping plans for database '54425547585a6420_25100627cc7529f5': Execution error: Error executing series set conversion: Error reading record batch while converting from SeriesSet: InvalidArgumentError("Sort requires at least one column")
```

Issue #2754, opened 2021-10-06
Flux test fails with drop: unsupported for aggregate count: Unsigned


alamb:
```
group_one_tag_count...fail: runtime error @53:12-53:46: drop: unsupported for aggregate count: Unsigned
```

Input:
```
// two fields, two tags keys, with three rows in each combo
inData = array.from(
    rows: [
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v0", _time: 2021-07-06T23:06:30Z, _value: 3},
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v0", _time: 2021-07-06T23:06:40Z, _value: 1},
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v0", _time: 2021-07-06T23:06:50Z, _value: 0},
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v1", _time: 2021-07-06T23:06:30Z, _value: 4},
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v1", _time: 2021-07-06T23:06:40Z, _value: 3},
        {_field: "f0", _measurement: "m0", t0: "t0v0", t1: "t1v1", _time: 2021-07-06T23:06:50Z, _value: 1},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:30Z, _value: 1},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:40Z, _value: 0},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v0", _time: 2021-07-06T23:06:50Z, _value: 4},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:30Z, _value: 4},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:40Z, _value: 0},
        {_field: "f0", _measurement: "m0", t0: "t0v1", t1: "t1v1", _time: 2021-07-06T23:06:50Z, _value: 4},
        {_field: "f1", _measurement: "m0", t0: "t0v0", t1: "t1v0", _time: 2021-07-06T23:06:30Z, _value: 0},
[...]

PR #2755, opened 2021-10-06
fix: do not sort the output of read_group with no group keys

alamb:
Resolves https://github.com/influxdata/influxdb_iox/issues/2696

This test from `fluxtest` used to fail:
```
range_nsecs_group_count...fail: query error: rpc error: code = InvalidArgument desc = Error running grouping plans for database '54425547585a6420_c40757b85db69ae4': Execution error: Error executing series set conversion: Error reading record batch while converting from SeriesSet: InvalidArgumentError("Sort requires at least one column")
```

With this fix, unfortunately the test still doesn't pass, but it hits another error (tracked by https://github.com/influxdata/influxdb_iox/issues/2754)

```
range_nsecs_group_count...fail: runtime error @95:12-95:46: drop: unsupported for aggregate count: Unsigned
```

The issue is that there are no tags in the input


query_tests/src/influxrpc/read_group.rs
```diff
@@ -78,6 +78,39 @@ async fn test_read_group_no_data_no_pred() {
...
+async fn test_read_group_data_no_tag_columns() {
+    let predicate = Predicate::default();
+    let agg = Aggregate::Count;
+    let group_columns = vec![];
+    let expected_results = vec![
+        "+-----+--------------------------------+",
+        "| foo | time                           |",
+        "+-----+--------------------------------+",
+        "| 2   | 1970-01-01T00:00:00.000000002Z |",
+        "+-----+--------------------------------+",
```
NGA-TRAN, line 101: Is this correct or wrong result?

PR #2756, opened 2021-10-06
fix: Cast count aggregates to Integer in read_group output

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/2754

# Rationale

Prior to this fix, fluxtest fails some tests like this:

```
range_nsecs_group_count...fail: runtime error @95:12-95:46: drop: unsupported for aggregate count: Unsigned
```

It is due to the fact that the flux table generator is not prepared to handle unsigned values for `count()` (it expects signed values).

# Changes
Cast the output of all `count()` aggregates to `Int64`

After this fix fluxtest is down to 26 failures (was over 30 yesterday) 🎉 
```
Ran 109 tests with 26 failure(s)
```


query/src/frontend/influxrpc.rs
```diff
@@ -1207,6 +1217,67 @@ impl InfluxRpcPlanner {
...
+        return Ok(plan_builder);
+    }
+
+    // it would be nice to have a way to get the current DF schema
+    // of the plan builder:
+    // https://github.com/apache/arrow-datafusion/issues/1074
+    let temp_plan = plan_builder.build().unwrap();
+    let schema = temp_plan.schema();
+
+    // in read_group and read_window_aggregate, aggregates are only
```
alamb, line 1247: this comment makes me realize the same treatment likely needs to be applied to read_window_aggregate -- will investigate that as a follow on PR

PR #2759, opened 2021-10-07
fix: interaction of lifecycle actions and delete predicates

### LINKED ISSUE #2627: Ensure that delete predicates don't get lost during lifecycle actions
Currently we capture delete predicates too here:

https://github.com/influxdata/influxdb_iox/blob/155d5a5cce267a538e9666f9f41c9d19223ef0fc/server/src/db/lifecycle/compact.rs#L69

https://github.com/influxdata/influxdb_iox/blob/155d5a5cce267a538e9666f9f41c9d19223ef0fc/server/src/db/lifecycle/persist.rs#L79

then we drop the chunk locks and later we use the captured predicates to create new chunks:

https://github.com/influxdata/influxdb_iox/blob/155d5a5cce267a538e9666f9f41c9d19223ef0fc/server/src/db/lifecycle/compact.rs#L118-L125

https://github.com/influxdata/influxdb_iox/blob/155d5a5cce267a538e9666f9f41c9d19223ef0fc/server/src/db/lifecycle/persist.rs#L143-L150

However in the meantime there might have been new predicates added to the chunks which we'll just forget. Note that the delete predicates that existed before dropping the lock might just be materialized during the compaction and split query, so we actually don't need to attach this set to the new chunks at all. To isolate the chunks that were added while the lifecycle action was running #2626 might be helpful.

crepererum:
Fixes #2627.

### DIFF
--- a/server/src/db/lifecycle/compact.rs
+++ b/server/src/db/lifecycle/compact.rs
@@ -12,7 +12,7 @@ use lifecycle::LifecycleWriteGuard;
 use observability_deps::tracing::info;
 use predicate::delete_predicate::DeletePredicate;
 use query::{compute_sort_key, exec::ExecutorType, frontend::reorg::ReorgPlanner, QueryChunkMeta};
-use std::{future::Future, sync::Arc};
+use std::{collections::HashSet, future::Future, sync::Arc};
 use tracker::{TaskTracker, TrackedFuture, TrackedFutureExt};
 
 /// Compact the provided chunks into a single chunk,
@@ -47,7 +47,7 @@ pub(crate) fn compact_chunks(
     let mut input_rows = 0;
     let mut time_of_first_write: Option<DateTime<Utc>> = None;
     let mut time_of_last_write: Option<DateTime<Utc>> = None;
-    let mut delete_predicates: Vec<Arc<DeletePredicate>> = vec![];
+    let mut delete_predicates_before: HashSet<Arc<DeletePredicate>> = HashSet::new();
     let mut min_order = ChunkOrder::MAX;
     let query_chunks = chunks
         .into_iter()
@@ -68,7 +68,7 @@ pub(crate) fn compact_chunks(
                 .map(|prev_last| prev_last.max(candidate_last))
                 .or(Some(candidate_last));
 
-            delete_predicates.extend(chunk.delete_predicates().iter().cloned());
+            delete_predicates_before.extend(chunk.delete_predicates().iter().cloned());
 
             min_order = min_order.min(chunk.order());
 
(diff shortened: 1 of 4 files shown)


server/src/db/lifecycle/compact.rs
```diff
@@ -108,13 +108,26 @@ pub(crate) fn compact_chunks(
...
-            // TODO: Reacquire chunk lock and check for "new" delete predicates (#2666)
-            partition
+            let chunk = partition
                 .force_drop_chunk(*id)
                 .expect("There was a lifecycle action attached to this chunk, who deleted it?!");
+
+            let chunk = chunk.read();
+            for pred in chunk.delete_predicates() {
+                if !delete_predicates_before.contains(pred) {
+                    delete_predicates_after.insert(Arc::clone(pred));
```
NGA-TRAN, line 120: Correct me if I am wrong. The `delete predicates` were added to the in-progress-compating chunks. So we want to keep them if those chunks are dropped here.

server/src/db/lifecycle/persist.rs
```diff
@@ -139,21 +139,33 @@ pub fn persist_chunks(
...
+                    }
+                }
             }
 
+            let delete_predicates = {
+                let mut tmp: Vec<_> = delete_predicates_after.into_iter().collect();
+                tmp.sort();
+                tmp
+            };
+
```
NGA-TRAN, line 161: This repeated code makes me wonder whether in the future we will be able to implement compacting and persisting in the same code. Maybe not

server/src/db/lifecycle/persist.rs
```diff
@@ -446,4 +468,113 @@ mod tests {
...
+        tokio::spawn(fut).await.unwrap().unwrap().unwrap();
+
+        // check in-mem delete predicates
+        let check_closure = |catalog: &Catalog| {
+            let chunks = catalog.chunks();
+            assert_eq!(chunks.len(), 1);
+            let chunk = &chunks[0];
+            let chunk = chunk.read();
+            let actual = chunk.delete_predicates();
+            let expected = vec![Arc::clone(&pred3)];
```
NGA-TRAN, line 562: Yeah, I am very curious where you eliminated the second call of `pred2`

server/src/db/lifecycle/write.rs
```diff
@@ -159,16 +160,27 @@ pub(super) fn write_chunk_to_object_store(
...
-            // because the chunk here was marked as "persisting".
+            // add delete predicates
+            //
+            // Delete predicates are handled in the following way
+            // 1. Predicates added before this chunk was created (aka before the DataFusion split plan was running):
+            //    They were materialized and are no longer part of the chunk.
+            // 2. Predicates added while this chunk was created (aka while the DataFusion split plan was running):
+            //    They were not materialized and must be added to this transaction.
+            // 3. Predicates added while we are persisting this chunk (aka while the "persisting" lifecycle action is active):
+            //    They are added to the outbound mailbox and will be handled by the background worker.
```
NGA-TRAN, line 180: Nice comments. Now I am no longer confused what `materialized` means 😄

server/src/db/lifecycle/compact.rs
```diff
@@ -253,4 +266,72 @@ mod tests {
...
+        let chunk = chunks[0].read();
+
+        let (_, fut) = compact_chunks(partition.upgrade(), vec![chunk.upgrade()]).unwrap();
+
+        // add more delete predicates
+        db.delete("cpu", Arc::clone(&pred2)).await.unwrap();
+        db.delete("cpu", Arc::clone(&pred3)).await.unwrap();
+
+        // finish future
+        tokio::spawn(fut).await.unwrap().unwrap().unwrap();
```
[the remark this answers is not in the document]
NGA-TRAN: Thanks for the examples @crepererum. Very clear

Issue #2688, opened 2021-10-01
Incorrect results fro mshow_tag_values with OR predicate 


alamb:
show_tag_values_measurement_or_predicate...fail: #  _diff=string,_value=string

https://github.com/influxdata/flux/blob/master/stdlib/influxdata/influxdb/schema/show_tag_values_measurement_or_predicate_test.flux is failing when run against IOX:

## Input
 ```
input = "
#group,false,false,false,false,true,true,true,true,true,true,true
#datatype,string,long,dateTime:RFC3339,long,string,string,string,string,string,string,string
#default,_result,,,,,,,,,,
,result,table,_time,_value,_field,_measurement,device,fstype,host,mode,path
,,0,2020-10-21T20:48:30Z,4881964326,inodes_free,disk,disk1s5,apfs,euterpe.local,ro,/
,,0,2020-10-21T20:48:40Z,4881964326,inodes_free,disk,disk1s5,apfs,euterpe.local,ro,/
,,0,2020-10-21T20:48:50Z,4881964326,inodes_free,disk,disk1s5,apfs,euterpe.local,ro,/
,,1,2020-10-21T20:48:30Z,4294963701,inodes_free,disk,disk2s1,hfs,euterpe.local,ro,/Volumes/IntelliJ IDEA CE
,,1,2020-10-21T20:48:40Z,4294963701,inodes_free,disk,disk2s1,hfs,euterpe.local,ro,/Volumes/IntelliJ IDEA CE
,,1,2020-10-21T20:48:50Z,4294963701,inodes_free,disk,disk2s1,hfs,euterpe.local,ro,/Volumes/IntelliJ IDEA CE
,,2,2020-10-21T20:48:30Z,488514,inodes_used,disk,disk1s5,apfs,euterpe.local,ro,/
,,2,2020-10-21T20:48:40Z,488514,inodes_used,disk,disk1s5,apfs,euterpe.local,ro,/
,,2,2020-10-21T20:48:50Z,488514,inodes_used,disk,disk1s5,apfs,euterpe.local,ro,/
,,3,2020-10-21T20:48:30Z,3578,inodes_used,disk,disk2s1,hfs,euterpe.local,ro,/Volumes/IntelliJ IDEA CE
,,3,2020-10-21T20:48:40Z,3578,inodes_used,disk,disk2s1,hfs,euterpe.local,ro,/Volumes/IntelliJ IDEA CE
[...]

PR #2763, opened 2021-10-07
fix: remove outdated "supported predicate" check in gRPC planner

alamb:
Related to https://github.com/influxdata/influxdb_iox/issues/2688

# Rationale

This old check for "support" is left over from a previous time before we had the `fluxtest` running; While the concerns it highlights are reasonable (and coverted by fluxtest

As we work through proper full support, the poorly implemented check for "supported" and "not supported" should be replaced with a proper implementation

# Change
1. Remove the check


query/src/frontend/influxrpc.rs
```diff
@@ -1274,77 +1266,6 @@ fn cast_aggregates(
...
-/// predicate in the grpc buffer
-struct SupportVisitor {}
-
-impl ExpressionVisitor for SupportVisitor {
-    fn pre_visit(self, expr: &Expr) -> DatafusionResult<Recursion<Self>> {
-        match expr {
-            Expr::Literal(..) => Ok(Recursion::Continue(self)),
-            Expr::Column(..) => Ok(Recursion::Continue(self)),
-            Expr::BinaryExpr { op, .. } => {
-                match op {
```
alamb, line 1302: note that this does not actually check for `IsNull` and `IsNotNull` 🤣

PR #2764, opened 2021-10-07
fix: Remove database name from structs where it's not really needed

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/2675

This doesn't switch from DB name + generation ID to UUID *yet*, but it was something easy to separate out.

When #2675 is fixed, the `IoxObjectStore` won't know what the database name is (unless it reads and parses the contents of the rules file). One place that was calling `iox_object_store.database_name()` was putting it in the `ChunkAddr`, `PartitionAddr`, and `Table`s to be used in `Job`s, logs, and `Metric`s.

But in every case, the database name is available from the `Db`, and it doesn't need to be passed around with each `ChunkAddr`, `PartitionAddr`, and `Table`.

For `Job`s, this PR moves the database name to the `Job` rather than being on the inner chunk or partition, and given that someday jobs might not need the database name at all (https://github.com/influxdata/influxdb_iox/issues/2758), it feels better to have all the database names on the same level (some jobs already had a database name field) to be in fewer spots when it can be removed.


crepererum:
This concerns the `ChunkAddr`:

1. I think this struct is used a lot for debugging, so we need to be aware that we are loosing some insights here (e.g. logs within the lifecycle policy).
2. Since chunk IDs are now UUID-based, we wouldn't actually need the `ChunkAddr` at all, since we COULD build a lookup table that resolves the chunk only based on its ID.
3. We're going to face the same issue / change that is proposed here when we wanna support table renames, because that's exactly the same shift (name -> UUID) just one level down.

Since I think we wanna have both proper logs and error messages as well as the ability to rename DBs, I would propose to leave the DB within the `ChunkAddr` but replace the actual DB name with something like this:

```rust
struct DatabaseId {
    /// Actual ID, used for `[Partial]{Eq,Ord}` and `Hash`
    id: Uuid,

    /// Current name, used for logging, `Debug` and `Display`
    name: ArcSwap<String>,
}

// all the impls go here
```

See [`arc_swap`](https://docs.rs/arc-swap/1.4.0/arc_swap/) for the special `Arc` that could be swapped out atomically.

PR #2765, opened 2021-10-07
refactor: Reduce string comparisons in gRPC expr tests

alamb:
I noticed while working on  https://github.com/influxdata/influxdb_iox/issues/2688 

# Rationale
DataFusion added support Expr::eq (`PartialEq`) a while ago, so we should use that instead of string comparisons

Also I didn't know about `unwrap_err()` when I originally wrote these tests

# Changes
Use `PartialEq` and `unwrap_err().to_string()`


src/influxdb_ioxd/rpc/storage/expr.rs
```diff
@@ -830,16 +830,10 @@ mod tests {
...
-
-        // compare the expression using their string representations
-        // as Expr can't be compared directly.
-        let converted_expr = format!("{:?}", converted_expr);
-        let expected_expr = format!("{:?}", expected_expr);
+        let converted_expr = &predicate.exprs;
 
         assert_eq!(
-            expected_expr, converted_expr,
+            &expected_expr, converted_expr,
```
alamb, line 836: I am basically writing a new test and wanted to do this refactor as a separate PR rather than part of the main one (which will already be substantial I think)

PR #2767, opened 2021-10-08
refactor: "min unpersisted ts" => "max persisted ts"

### LINKED ISSUE #2225: `PersistenceWindows` can panic / overflow with max timestamps
The `PersistenceWindows` hold row timestamps. To this we add 1ns to derive the next unpersisted timestamps (within the sequence number range of partially persisted data, so this TS is not monotonic). This addition can overflow or panic (in debug mode). Since the data is user-controlled (in contrast to the wallclock), this must not happen. Can easily tested with the following line protocol:

```text
table_1 foo=1 <i64::MAX>
```

which is

```text
table_1 foo=1 9223372036854775807
```

crepererum:
Store the "maximum persisted timestamp" instead of the "minimum
unpersisted timestamp". This avoids the need to calculate the next
timestamp from the current one (which was done via "max TS + 1ns").

The old calculation was prone to overflow panics. Since the
timestamps in this calculation originate from user-provided data (and
not the wall clock), this was an easy DoS vector that could be triggered
via the following line protocol:

```text
table_1 foo=1 <i64::MAX>
```

which is

```text
table_1 foo=1 9223372036854775807
```

Bonus points: the timestamp persisted in the partition
checkpoints is now the very same that was used by the split query during
persistence. Consistence FTW!

Fixes #2225.

### DIFF
--- a/generated_types/protos/influxdata/iox/catalog/v1/parquet_metadata.proto
+++ b/generated_types/protos/influxdata/iox/catalog/v1/parquet_metadata.proto
@@ -50,8 +50,11 @@ message PartitionCheckpoint {
   // Maps `sequencer_id` to the to-be-persisted minimum and seen maximum sequence numbers.
   map<uint32, OptionalMinMaxSequence> sequencer_numbers = 1;
 
-  // Minimum unpersisted timestamp.
-  google.protobuf.Timestamp min_unpersisted_timestamp = 2;
+  // Was Minimum unpersisted timestamp.
+  reserved 2;
+
+  // Maximum persisted timestamp.
+  google.protobuf.Timestamp max_persisted_timestamp = 3;
 }
 
 // Record of the playback state for the whole database.
--- a/persistence_windows/src/persistence_windows.rs
+++ b/persistence_windows/src/persistence_windows.rs
@@ -98,7 +98,7 @@ impl FlushHandle {
             Arc::clone(&self.addr.table_name),
             Arc::clone(&self.addr.partition_key),
             self.sequencer_numbers.clone(),
-            self.timestamp + chrono::Duration::nanoseconds(1),
+            self.timestamp,
         )
     }
 }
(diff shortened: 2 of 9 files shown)


generated_types/protos/influxdata/iox/catalog/v1/parquet_metadata.proto
```diff
@@ -50,8 +50,11 @@ message PartitionCheckpoint {
   // Maps `sequencer_id` to the to-be-persisted minimum and seen maximum sequence numbers.
   map<uint32, OptionalMinMaxSequence> sequencer_numbers = 1;
 
-  // Minimum unpersisted timestamp.
-  google.protobuf.Timestamp min_unpersisted_timestamp = 2;
+  // Was Minimum unpersisted timestamp.
+  reserved 2;
+
+  // Maximum persisted timestamp.
+  google.protobuf.Timestamp max_persisted_timestamp = 3;
```
tustvold, line 57: I think perhaps we should call this flush_timestamp, it isn't accurate to call it max_persisted_timestamp as it is possible for a later persist to use a smaller flush timestamp than a previous

server/src/db/replay.rs
```diff
@@ -331,10 +331,10 @@ fn filter_entry(
                 SequenceNumberSection::PartiallyPersisted => {
                     // TODO: implement row filtering, for now replay the entire batch
                     let maybe_mask = table_batch.timestamps().ok().map(|timestamps| {
-                        let min_unpersisted_ts = min_unpersisted_ts.timestamp_nanos();
+                        let max_persisted_ts = max_persisted_ts.timestamp_nanos();
```
tustvold, line 334: I think the above TODO can go?

persistence_windows/src/persistence_windows.rs
```diff
@@ -350,18 +350,26 @@ impl PersistenceWindows {
             persistable.max_time, timestamp,
             "persistable max time doesn't match handle"
         );
+
         // Everything up to and including persistable max time will have been persisted
-        let new_min = persistable.max_time + chrono::Duration::nanoseconds(1);
-        for w in self.closed.iter_mut().take(closed_count) {
-            if w.min_time < new_min {
-                w.min_time = new_min;
+        if let Some(new_min) = persistable
```
tustvold, line 355: I think this could be simplified with a saturating add (if chrono supports such a thing)

server/src/db/replay.rs
crepererum: done

generated_types/protos/influxdata/iox/catalog/v1/parquet_metadata.proto
crepererum: done

Issue #2750, opened 2021-10-06
Implement Delete Support for meta query operations in Read Buffer


e-dard:
TODO.

e-dard:
`column_names` completed in #2771. `column_values` to do.

PR #2771, opened 2021-10-08
feat: add negated predicate support to column_names in Read Buffer

e-dard:
Needs #2769 to merge first.

Towards #2750.

This PR adds dedicated support for handling soft deletes in the `column_names` Read Buffer operation. `column_names` is called when the InfluxDB RPC requires a `tag_keys` call.

The implementation follows the same pattern as `read_filter`. We evaluate all matching rows for the negated predicates, then we determine the relative complement between the filtered rows and the deleted rows.

With this final set of row IDs a column name is included in the results if we can find any non-null value at one of the final row IDs for the column in question.


read_buffer/src/chunk.rs
```diff
@@ -1276,13 +1283,86 @@ mod test {
...
+            )
+            .unwrap();
+        assert_eq!(result, to_set(&[]));
+
+        let result = chunk
+            .column_names(
+                Predicate::default(),
+                vec![Predicate::new(vec![BinaryExpr::from((
+                    "region", "!=", "west",
+                ))])], // all rows deleted
```
alamb, line 1341:
```suggestion
                ))])], // all rows except "east" deleted
```

PR #2774, opened 2021-10-08
fix: do not break chunk ordering during persistence

### LINKED ISSUE #2729: Persist Chunks Breaks Chunk Ordering
Currently the persist chunks lifecycle action performs a query on all chunks that match the flush predicate to produce a chunk to persist and a remainder. It skips any chunks that cannot match the predicate, this is incorrect as it will potentially bring forward later writes ahead of a skipped chunk. This should be fixed.

Currently this chunk selection logic is duplicated in `Db::persist_partition_with_timestamp` and the lifecycle policy, this duplication should ideally be removed and potentially better tested - https://github.com/influxdata/influxdb_iox/pull/2718#discussion_r721966100

crepererum:
Fixes #2729.

### DIFF
--- a/lifecycle/src/policy.rs
+++ b/lifecycle/src/policy.rs
@@ -387,7 +387,7 @@ where
             }
         };
 
-        let chunks = match select_persistable_chunks(&chunks, db_name, persist_handle.timestamp()) {
+        let chunks = match select_persistable_chunks(&chunks, persist_handle.timestamp()) {
             Ok(chunks) => chunks,
             Err(stall) => {
                 return stall;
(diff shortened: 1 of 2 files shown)


lifecycle/src/policy.rs
```diff
@@ -650,6 +611,73 @@ fn sort_free_candidates<P>(candidates: &mut Vec<FreeCandidate<'_, P>>) {
     })
 }
 
+pub fn select_persistable_chunks<'a, P, D>(
```
tustvold, line 614: I think this could do with documentation of the return type, as `bool` itself isn't very descriptive

lifecycle/src/policy.rs
```diff
@@ -650,6 +611,73 @@ fn sort_free_candidates<P>(candidates: &mut Vec<FreeCandidate<'_, P>>) {
...
+                        "chunk not eligible due to storage");
+                continue;
+            }
+        }
+
+        // Chunk's data is entirely after the time we are flushing
+        // up to, and thus there is reason to include it in the
+        // plan
+        if chunk.min_timestamp() > flush_ts {
+            // Ignore chunk for now, but we might need it later to closer chunk order gaps
```
tustvold, line 646:
```suggestion
            // Ignore chunk for now, but we might need it later to close chunk order gaps
```

lifecycle/src/policy.rs
```diff
@@ -650,6 +611,73 @@ fn sort_free_candidates<P>(candidates: &mut Vec<FreeCandidate<'_, P>>) {
...
+where
+    D: LockableChunk<Chunk = P>,
+    P: LifecycleChunk,
+{
+    let mut to_persist = Vec::new();
+    let mut to_persist_gap = Vec::new();
+
+    for chunk in chunks {
+        let chunk = chunk.read();
+        trace!(%db_name, chunk=%chunk.addr(), "considering chunk for persistence");
```
tustvold, line 628: As pointed out by @carols10cents on #2764 the `%db_name` is redundant and can be removed, if memory serves I added it to `ChunkAddr` for this reason, but clearly didn't update all the code paths :laughing:

tustvold (approved):
Nice :+1:

lifecycle/src/policy.rs
crepererum: done

PR #2778, opened 2021-10-08
fix: Remove all uses of `IoxObjectStore.database_name()`

carols10cents:
And pass the database name as a separate argument where needed, because `IoxObjectStore` soon won't know the database name.

Connects to #2675.


parquet_file/src/catalog/core.rs
```diff
@@ -305,7 +309,7 @@ impl PreservedCatalog {
         if Self::exists(&iox_object_store).await? {
             return Err(Error::AlreadyExists {});
         }
-        let state = S::new_empty(iox_object_store.database_name(), state_data);
+        let state = S::new_empty(db_name, state_data);
```
crepererum, line 312: Uh, that's indeed a bit of a mess. I think `S` doesn't really need the DB name, but I want to rework the `new_empty` logic anyways. Will do that after your PR.

PR #2790, opened 2021-10-11
refactor: cancel executor jobs on drop

### LINKED ISSUE #2027: Canceling a query doesn't seem to cancel queued / inflight fetches of parquet files from object store
__Steps to reproduce:__
I don't have a great reproducer now, but I wanted to file this as a placeholder until I do / this becomes an issue

Basically the reproducer is "run a query against a table with a large number of parquet chunks that can't complete within 30 seconds"

__Expected behaviour:__
When the connection is canceled / the query is shutdown any outstanding requests to object store are canclled and no new ones are started 

__Actual behaviour:__
I believe (as described on https://docs.google.com/document/d/1UmfoGuoVjtvmXOgp_AXWqShamlFVmydlxo06vv22iRU/edit#) that the parquet files continue to be fetched (and then discarded). Evidence that supports this behavior:
1. A subsequent query to a smaller table almost timed out (took 1000 more times)
2. For a significant amount of time after the query was canceled, the pod showed elevated CPU and network usage

__Environment info:__

Internal IOx cluster `query-3003` / `istio` database

crepererum:
Our executor is not meant as a fire-and-forget system. Instead the
submitter should always poll the result. Dropping the receiver side (aka
the job handle) should cancel the job.

Note that this is NOT sufficient to fix #2027 since datafusion also has a few places involving `spawn` where dropping the receiver doesn't cancel the spawned task.

### DIFF
--- a/query/src/exec/task.rs
+++ b/query/src/exec/task.rs
@@ -2,19 +2,67 @@
 //! intensive" workloads such as DataFusion plans
 
 use parking_lot::Mutex;
+use pin_project::{pin_project, pinned_drop};
 use std::{pin::Pin, sync::Arc};
 use tokio::sync::oneshot::Receiver;
+use tokio_util::sync::CancellationToken;
 
 use futures::Future;
 
 use observability_deps::tracing::warn;
 
-/// The type of thing that the dedicated executor runs
-type Task = Pin<Box<dyn Future<Output = ()> + Send>>;
+/// Task that can be added to the executor-internal queue.
+///
+/// Every task within the executor is represented by a [`Job`] that can be polled by the API user.
+struct Task {
+    fut: Pin<Box<dyn Future<Output = ()> + Send>>,
+    cancel: CancellationToken,
+}
+
+impl Task {
+    /// Run task.
+    ///
+    /// This runs the payload or cancels if the linked [`Job`] is dropped.
+    async fn run(self) {
+        tokio::select! {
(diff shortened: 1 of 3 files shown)


query/src/exec/task.rs
```diff
@@ -111,30 +159,35 @@ impl DedicatedExecutor {
     ///
     /// Currently all tasks are added to the tokio executor
     /// immediately and compete for the threadpool's resources.
-    pub fn spawn<T>(&self, task: T) -> Receiver<T::Output>
```
alamb, line 114: I guess my initial idea had been that dropping the receiver would effectively terminate the task but upon further reflection I realize dropping the receiver will let the task run to the point where it tries to send back the results (aka does almost all the work)

query/src/exec/task.rs
```diff
@@ -352,9 +406,33 @@ mod tests {
...
+        let barrier2 = Arc::new(AsyncBarrier::new(2));
+
+        let exec = DedicatedExecutor::new("Test DedicatedExecutor", 1);
+
+        let dedicated_task1 = exec.spawn(do_work_async(11, Arc::clone(&barrier1)));
+        let dedicated_task2 = exec.spawn(do_work_async(22, Arc::clone(&barrier2)));
+
+        drop(dedicated_task1);
+
+        barrier2.wait().await;
```
alamb, line 421: Am I correct that this test will wait forever without the code change in the PR?

PR #2792, opened 2021-10-11
feat: add time provider abstraction (#2722)

tustvold:
Adds the time provider abstraction, future PRs will plumb this in


time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.0.signed_duration_since(other.0).to_std().ok()
+    }
+
+    /// Adds given `Duration` to the current date and time.
+    ///
+    /// Returns `None` if it would result in overflow
+    ///
+    /// # Panic
+    ///
+    /// If `duration` cannot fit inside `chrono::Duration`
```
crepererum, line 135: Shouldn't this return `None` instead?

time/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "time"
+version = "0.1.0"
+authors = ["Raphael Taylor-Davies <r.taylordavies@googlemail.com>"]
```
crepererum, line 4:
```suggestion
```

`authors` is optional nowadays and is a bit discouraged for larger projects.

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        Some(Self(self.0.checked_add_signed(duration)?))
+    }
+
+    /// Subtracts the given `Duration` from the current date and time.
+    ///
+    /// Returns `None` if it would result in overflow
+    ///
+    /// # Panic
+    ///
+    /// If `duration` cannot fit inside `chrono::Duration`
```
crepererum, line 147: Same.

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+
+impl TimeProvider for SystemProvider {
+    fn now(&self) -> Time {
+        Time(Utc::now())
+    }
+}
+
+/// A [`TimeProvider`] that returns a fixed `Time` that can be set by [`MockProvider::set`]
+///
+/// By default returns UNIX_EPOCH
```
crepererum, line 182:
```suggestion
```

That `Default` impl doesn't exist

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.0
+    }
+}
+
+pub trait TimeProvider: std::fmt::Debug + Send + Sync + 'static {
+    /// Returns the current `Time`. No guarantees are made about monotonicity
+    fn now(&self) -> Time;
+}
+
+/// A [`TimeProvider`] that uses `Utc::now` as a clock source
```
crepererum, line 164: Maybe link the `chrono` docs.rs page here or explain which clock source is used.

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    clippy::clone_on_ref_ptr,
+    clippy::future_not_send
+)]
+
+use chrono::{DateTime, TimeZone, Timelike, Utc};
+use parking_lot::RwLock;
+use std::ops::{Add, Sub};
+use std::time::Duration;
+
+/// A UTC Timestamp returned by a `TimeProvider`
```
crepererum, line 15: Maybe mention that timestamps TRY to use the wall-clock.

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    fn sub(self, rhs: Duration) -> Self::Output {
+        let duration = chrono::Duration::from_std(rhs).unwrap();
+        Self(self.0 - duration)
+    }
+}
+
+impl Sub<Self> for Time {
+    type Output = Duration;
+
+    fn sub(self, rhs: Self) -> Self::Output {
```
crepererum, line 44:
```suggestion
    /// Calculates difference in wall-clock time.
    ///
    /// **Warning: Because monotonicity is not guaranteed, `t2 - t1` might be negative even when `t2` was generated after `t1!**
    ///
    /// This is unrelated that CPU time.
    fn sub(self, rhs: Self) -> Self::Output {
```

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        self.0.timestamp_nanos()
+    }
+
+    /// Returns the number of seconds since January 1, 1970 UTC
+    pub fn timestamp(&self) -> i64 {
+        self.0.timestamp()
+    }
+
+    /// Returns the minute number from 0 to 59.
+    pub fn minute(&self) -> u32 {
```
crepererum, line 110: Why is `minute` added here but `hour` and `second` are missing?

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+        Self {
+            now: RwLock::new(start),
+        }
+    }
+
+    pub fn set(&self, time: Time) {
+        *self.now.write() = time
+    }
+
+    pub fn inc(&self, duration: Duration) -> Time {
```
crepererum, line 199: This is untested.

time/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+impl std::fmt::Display for Time {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.to_rfc3339())
+    }
+}
+
+impl Time {
```
crepererum, line 61: Do we need all these methods? If so, I think every method deserves at least a single call during the tests.

crepererum (approved):
Smaller nits regarding testing and error handling.

PR #2793, opened 2021-10-11
refactor: create PreservedCatalogConfig struct

tustvold:
As discussed with @crepererum, currently plumbing through a new dependency of the preserved catalog results in a non-trivial amount of churn, as the new type must be plumbed through in a large number of method calls, test, etc...

This PR therefore adds a `PreservedCatalogConfig` struct to contain this information, which should help make plumbing in the new time provider in #2722  significantly less verbose. This will also help if we ever add further dependencies to the preserved catalog, e.g. metrics.


parquet_file/src/catalog/core.rs
```diff
@@ -260,59 +301,33 @@ impl PreservedCatalog {
         Ok(iox_object_store.wipe_catalog().await.context(Write)?)
     }
 
+    /// Deletes the catalog described by the provided config
+    pub async fn wipe_with_config(config: &PreservedCatalogConfig) -> Result<()> {
```
tustvold, line 305: This method is a bit off, but is needed because load_or_create_preserved_catalog does not live within this crate and I didn't want to make the members of `PreservedCatalogConfig` accessible outside the crate. This to avoid them being used for non-catalog related plumbing :sweat_smile: 

Longer term I think load_or_create_preserved_catalog should probably be a `PreservedCatalog::load_or_create` but wanted to leave that for another day

crepererum:
Please rebase this on top of #2778.

PR #2797, opened 2021-10-11
feat: allow write buffer producers to read sequencer IDs

crepererum:
Pulled out of #2497. I would like to do a bit of write buffer work (e.g. adding a format header for a potential transition to protobuf-based messages), and would like to merge this change beforehand. No matter how we implement message distribution, this feature here is going to be required anyways (since the producer side needs to detect the number of partitions).

### DIFF
--- a/write_buffer/src/core.rs
+++ b/write_buffer/src/core.rs
@@ -15,6 +15,11 @@ pub type WriteBufferError = Box<dyn std::error::Error + Sync + Send>;
 /// entries from the Write Buffer at a later time.
 #[async_trait]
 pub trait WriteBufferWriting: Sync + Send + Debug + 'static {
+    /// List all known sequencers.
+    ///
+    /// This list is sorted and not empty.
+    fn sequencer_ids(&self) -> Vec<u32>;
+
     /// Send an `Entry` to the write buffer using the specified sequencer ID.
     ///
     /// Returns information that can be used to restore entries at a later time.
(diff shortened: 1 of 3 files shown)


write_buffer/src/kafka.rs
```diff
@@ -121,22 +126,16 @@ impl KafkaBufferProducer {
...
-        if get_partitions(&database_name, &cfg).await?.is_empty() {
-            if let Some(cfg) = creation_config {
-                create_kafka_topic(&conn, &database_name, cfg.n_sequencers, &cfg.options).await?;
-            } else {
-                return Err("no partitions found and auto-creation not requested"
-                    .to_string()
-                    .into());
-            }
-        }
+        let partitions =
```
alamb, line 129: it took me a while to convince myself that `self.partitions` was sorted (as required per the doc comments of `sequencer_ids()` -- I wonder if some more doc comments are in order (perhaps on the docs of `get_partitions` and/or `maybe_auto_create_topics`

write_buffer/src/core.rs
```diff
@@ -15,6 +15,11 @@ pub type WriteBufferError = Box<dyn std::error::Error + Sync + Send>;
 /// entries from the Write Buffer at a later time.
 #[async_trait]
 pub trait WriteBufferWriting: Sync + Send + Debug + 'static {
+    /// List all known sequencers.
+    ///
+    /// This list is sorted and not empty.
+    fn sequencer_ids(&self) -> Vec<u32>;
```
alamb, line 21: Can there be duplicates in this list ? If not, maybe something like` BTreeSet<u32>` would communicate the sortedness / uniqueness of this lis (and avoid the need to document the expectations / implications lower down in the code about sortedness)

alamb (approved):
Are you planning to work on

write_buffer/src/kafka.rs
crepererum: Changed type from a `Vector` to a `BTreeSet`.

Issue #1473, opened 2021-05-11
Cross-Request Aggregation


tustvold:
Currently the write path for line-protocol is as follows:

* Entire string payload is parsed into ParsedLine<_>
* The list of ParsedLine<_> are aggregated into flatbuffer Entry's per-shard
* These Entry's are then sent where they need to go

This means that every Entry originates from a single write. This means any of the following leads to very small Entry payloads:

* IOx is receiving small write payloads
* IOx is configured with a large number of shards
* IOx is writing to something that mandates further sharding/partitioning (e.g. Kafka / Kinesis)

This has a severely detrimental impact on throughput should it occur.

An alternative approach would be to parse the data into lines of line protocol, and put these into per-destination queues. When a given queue reaches a certain size threshold, or a timeout expires, the contained data would be parceled up and written out. I have previously written an implementation of this concept [here](https://github.com/tustvold/rust-playground/blob/master/lib/stream/src/batch.rs).

In order to avoid interfering with the ongoing work on the write buffer, the proposed write-path would be as follows:

* Parse lines to some owning data structure
* Group the lines based on their destination shard
* For each shard group
    * If the destination shard is a remote IOx node - compute Entry and continue as before
    * If the destination shard is the local IOx node - compute Entry and continue as before (might be changed by #1472)
    * If the destination shard is not an IOx node - dump lines onto queue

Effectively this would allow non IOx shards to experiment with doing cross-request aggregation, and in future may allow us to do the same for IOx shards if we so wished

alamb:
I wonder how "when to acknowledge the write" would work in this situation? Would the write be held until the batch was written?

I also wonder about error handling. Say if we batched up three writes from different writers; When we decided to process the batch, the first succeeds, the middle one fails, what would we do with the third write? Would we try it anyways?

tustvold:
> I wonder how "when to acknowledge the write" would work in this situation? Would the write be held until the batch was written?

Yes. There are many ways to implement this, the most basic being a oneshot, a more sophisticated version using a Notified would also be possible.

> I also wonder about error handling. Say if we batched up three writes from different writers; When we decided to process the batch, the first succeeds, the middle one fails, what would we do with the third write? Would we try it anyways?

We have the same issue with sharding already, I understand that partial failures are permissible

> Until we observe this being a significant problem in our traces / metrics, I would personally put this into the "it might be premature optimization" category, but I may not have the whole story

[...]

PR #2803, opened 2021-10-12
feat: ability to link other spans in span context

crepererum:
This can be used when aggregating from multiple parent spans, e.g. when
we want to implement #1473.

### DIFF
--- a/trace_exporters/src/jaeger/span.rs
+++ b/trace_exporters/src/jaeger/span.rs
@@ -145,3 +145,18 @@ fn tag_from_meta(key: String, value: MetaValue) -> jaeger::Tag {
     };
     tag
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_split_trace_id_integer_conversion() {
+        // test case from
+        // https://github.com/open-telemetry/opentelemetry-specification/blob/639c7443e78800b085d2c9826d1b300f5e81fded/specification/trace/sdk_exporters/jaeger.md#ids
+        let trace_id = TraceId::new(0xFF00000000000000).unwrap();
+        let (high, low) = split_trace_id(trace_id);
+        assert_eq!(high, 0);
+        assert_eq!(low, -72057594037927936);
+    }
+}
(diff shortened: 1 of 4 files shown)


trace_exporters/src/jaeger/span.rs
```diff
@@ -51,13 +60,34 @@ impl From<Span> for jaeger::Span {
...
 
+        let references = if s.ctx.links.is_empty() {
+            None
+        } else {
+            Some(
+                s.ctx
+                    .links
+                    .into_iter()
+                    .map(|(trace_id, span_id)| {
+                        // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk_exporters/jaeger.md#links
```
tustvold, line 71: :ok_hand:

trace_exporters/src/jaeger/span.rs
```diff
@@ -1,12 +1,21 @@
...
-use crate::thrift::jaeger;
-use trace::span::{MetaValue, Span, SpanEvent, SpanStatus};
+use crate::thrift::jaeger::{self, SpanRef};
+use trace::{
+    ctx::TraceId,
+    span::{MetaValue, Span, SpanEvent, SpanStatus},
+};
+
+/// Split [`TraceId`] into high and low part.
+fn split_trace_id(trace_id: TraceId) -> (i64, i64) {
```
tustvold, line 9: The docs say that it should be big endian, but the Rust opentelemetry SDK does the same transformation so I don't know...

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-jaeger/src/exporter/mod.rs#L621

I think the golang SDK is not doing the same thing

https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/jaeger/jaeger.go#L195

It is possible that the Rust OpenTelemetry has a bug here?

trace_http/src/ctx.rs
```diff
@@ -211,10 +215,14 @@ fn decode_jaeger(
         return Ok(None);
     }
 
+    // Links cannot be specified via the HTTP header
+    let links = Vec::with_capacity(0);
```
tustvold, line 219: Why not just `vec![]`?

crepererum: Because I thought that would allocate memory, but apparently it doesn't: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=184113a9afa945cf3cf3b8f854f4c5ce
Will change that.

trace_exporters/src/jaeger/span.rs
crepererum: I think the high-low split makes sense. However I'm wondering about the `u64` to `i64` conversion. Is that supposed to be a wrapping conversion? I think our current code panics in debug mode due to overflow checks.

PR #2830, opened 2021-10-13
refactor: rework write buffer compaction as integration test

tustvold:
Much like #2820 this reworks write_buffer_reads_wait_for_compaction as an integration test, unblocking #2242


server/src/lib.rs
```diff
@@ -1225,9 +1225,43 @@ where
...
+
+    pub fn make_server(application: Arc<ApplicationState>) -> Arc<Server<TestConnectionManager>> {
+        Arc::new(Server::new(
+            TestConnectionManager::new(),
+            application,
+            Default::default(),
+        ))
+    }
+
+    pub async fn make_init_server(
```
crepererum, line 1248: Since this is now a public interface, some docstring would help. Or a rename, since `make_init_server` somewhat confused my (what is an Init Server?), maybe `make_initialized_server` or `make_server_and_wait_for_init`.

PR #2837, opened 2021-10-13
feat: Write a server config that lists database names and object storage locations

carols10cents:
Connects to #2676.

In the course of working on #2675, it became clear that not knowing the database name until reading the db rules wasn't quite the desired API. Instead of reading the db rules to find database names, and instead of storing multiple `link` files, each server will have one `config.pb` file in its `s3://<bucket..>/nodes/<server-id>/` directory that will contain the list of databases it owns with their names and object store locations.

This PR *writes* out the file after starting up the server and after databases are created to transition to using this file. The next PR will switch server startup to reading the file if it exists and falling back to scanning object storage until we've transition to the server config file always existing.

### DIFF
--- a/generated_types/protos/influxdata/iox/management/v1/server_config.proto
+++ b/generated_types/protos/influxdata/iox/management/v1/server_config.proto
@@ -4,6 +4,11 @@ option go_package = "github.com/influxdata/iox/management/v1";
 
 // Stores a server's map of the databases it owns. The keys are the database names and the values
 // are the database's location in object storage.
+//
+// Example (current): "foo" => "file:/1/foo" ("[object store root]/[server id]/[database name]")
+// Example (after completing the switch to floating databases):
+//         "foo" => "file:/dbs/3f25185a-0773-4ae8-abda-f9c3786f242b"
+//                  ("[object store root]/dbs/[database uuid]")
 message ServerConfig {
   map<string, string> databases = 1;
 }
--- a/generated_types/protos/influxdata/iox/management/v1/server_config.proto
+++ b/generated_types/protos/influxdata/iox/management/v1/server_config.proto
@@ -1,0 +1,22 @@
+syntax = "proto3";
+package influxdata.iox.management.v1;
+option go_package = "github.com/influxdata/iox/management/v1";
+
+// Stores a server's list of the databases it owns, including the database names and locations
+// in object storage.
+message ServerConfig {
+  repeated Database databases = 1;
+}
+
+message Database {
+  // The unencoded name of the database
+  //
+  // Must be a non-empty string containing no control characters
+  string name = 1;
+
+  // The location of the database's files in object storage relative to the application's
+  // object storage bucket configuration
+  //
+  // Must be a non-empty string
+  string location = 2;
+}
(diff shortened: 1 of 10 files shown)


server/src/lib.rs
```diff
@@ -1152,12 +1208,36 @@ async fn maybe_initialize_server(shared: &ServerShared) {
                     .expect("database unique");
             }
 
-            info!(server_id=%init_ready.server_id, "server initialized");
-            ServerState::Initialized(state)
+            let bytes = state.server_config();
```
alamb, line 1211: Do you envision that this is temporary / transition code to build a `config.pb` file for all existing databases (and will be removed eventually)?  It is a clever way to bootstrap `config.pb` files without bricking all existing databases 👍

server/src/lib.rs
```diff
@@ -1363,16 +1487,30 @@ mod tests {
...
+        // assert server config file exists and has 1 entry
+        let config = server_config(application.object_store(), server_id).await;
+        assert_config_contents(&config, &[(&name, iox_object_store.root_path())]);
+
         let db2 = DatabaseName::new("db_awesome").unwrap();
         let rules2 = DatabaseRules::new(db2.clone());
         let provided_rules2 = make_provided_rules(rules2);
 
-        server
+        let awesome = server
```
alamb, line 1498: oh, so awesome! 😆

server/src/lib.rs
```diff
@@ -1469,17 +1618,41 @@ mod tests {
...
+
+        let server = make_server(Arc::clone(&application));
         server.set_id(ServerId::try_from(1).unwrap()).unwrap();
         server.wait_for_init().await.unwrap();
 
         assert_eq!(server.db_names_sorted(), vec!["apples", "bananas"]);
 
-        let bananas = DatabaseName::new("bananas").unwrap();
-        let apples = DatabaseName::new("apples").unwrap();
+        // assert server config file has been recreated and contains 2 entries, even though
```
alamb, line 1634: nice

generated_types/protos/influxdata/iox/management/v1/server_config.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+package influxdata.iox.management.v1;
+option go_package = "github.com/influxdata/iox/management/v1";
+
+// Stores a server's list of the databases it owns, including the database names and locations
+// in object storage.
+message ServerConfig {
+  repeated Database databases = 1;
```
tustvold, line 8: I don't feel especially strongly about this, but I wonder if it might be clearer if instead of name being encoded within `Database`, instead this was a map keyed by database name?

server/src/lib.rs
```diff
@@ -1321,13 +1401,57 @@ mod tests {
         assert!(matches!(resp, Error::IdNotSet));
     }
 
+    async fn server_config_contents(object_store: &ObjectStore, server_id: ServerId) -> Bytes {
+        IoxObjectStore::get_server_config_file(object_store, server_id)
+            .await
+            .unwrap_or_else(|_| Bytes::new())
```
tustvold, line 1407: This is a neat way to workaround lack of prost Default :laughing:

server/src/lib.rs
carols10cents: Indeed!

Issue #2163, opened 2021-07-30
Re-enable row group pruning in parquet file reading


alamb:
__Use case:__
We "solved" https://github.com/influxdata/influxdb_iox/issues/2153 by disabling parquet predicate pushdown (due to https://github.com/apache/arrow-rs/issues/641). 

Disabling parquet predicate pushdown is probably ok given all parquet files we make have a single row group.

__Proposal:__
Re-enable pushdown of predicates into parquet scans. This will be important if we ever generate parquet files with multiple row groups or if datafusion gets smarter about using these pushed down predicates (i.e. does more than row group pruning)

__Desired behaviour:__
Back out the changes in XXXX

crepererum:
Outdated. I believe this was indirectly fixed by #3291.

PR #2838, opened 2021-10-13
chore: Update datafusion

alamb:
# Rationale:
Get the following changes into IOx
1. Integrate async planner / parquet improvements in https://github.com/apache/arrow-datafusion/pull/1010
1. Cleanup from @creperum on plan cancel https://github.com/apache/arrow-datafusion/pull/1105
2. Set the stage for reading parquet files without downloading the entire things locally
3. Get some other notable features from DataFusion such as`approx_distinct`, `cume_dist` and `percent_rank`

# Changes:
1. Update to use latest datafusion: https://github.com/apache/arrow-datafusion/commit/2454e468641d4d98af211c2800c0afec2732385b
2. Update to reflect new signatures

PR #2839, opened 2021-10-13
fix: re-enable parquet row group pruning

alamb:
Follow on to https://github.com/influxdata/influxdb_iox/pull/2838

Resolves https://github.com/influxdata/influxdb_iox/issues/2163 now that https://github.com/apache/arrow-rs/issues/641 has been fixed.

Also improves use of DataFusion's new file source API (Thanks @rdettai)

### DIFF
--- a/parquet_file/src/storage.rs
+++ b/parquet_file/src/storage.rs
@@ -8,11 +8,13 @@ use arrow::{
 use bytes::Bytes;
 use data_types::chunk_metadata::ChunkAddr;
 use datafusion::{
-    datasource::{object_store::local::LocalFileSystem, PartitionedFile},
-    logical_plan::Expr,
-    physical_plan::{
-        file_format::ParquetExec, ExecutionPlan, Partitioning, SendableRecordBatchStream,
+    datasource::{
+        file_format::{parquet::ParquetFormat, FileFormat, PhysicalPlanConfig},
+        object_store::{local::LocalFileSystem, ObjectStore, SizedFile},
+        PartitionedFile,
     },
+    logical_plan::Expr,
+    physical_plan::{Partitioning, SendableRecordBatchStream},
 };
 use datafusion_util::AdapterStream;
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


parquet_file/src/storage.rs
```diff
@@ -276,34 +289,43 @@ impl Storage {
...
-            debug!(?predicate, "Skipping predicate pushdown due to XXX");
-        }
-        let predicate = None;
-
         let object_store = Arc::new(LocalFileSystem {});
+        let filters: Vec<_> = predicate.into_iter().collect();
+
+        // Read statistics from the parquet file so we can use parquet
+        // row group pruning. It might be cool to use the statistics
+        // stored in the IOx catalog to do this (evetually)
```
alamb, line 297:
```suggestion
        // stored in the IOx catalog to do this (eventually)
```

parquet_file/src/storage.rs
```diff
@@ -276,34 +289,43 @@ impl Storage {
...
+        let format = ParquetFormat::default();
```
alamb, line 298: This uses the cool new `FileFormat` trait / API that was added to DataFusion to create the `ParquetExec` without having to do so directly. 👍

parquet_file/src/storage.rs
```diff
@@ -276,34 +289,43 @@ impl Storage {
...
-        // https://github.com/influxdata/influxdb_iox/issues/2163
-        if predicate.is_some() {
-            debug!(?predicate, "Skipping predicate pushdown due to XXX");
-        }
-        let predicate = None;
-
         let object_store = Arc::new(LocalFileSystem {});
+        let filters: Vec<_> = predicate.into_iter().collect();
+
+        // Read statistics from the parquet file so we can use parquet
```
alamb, line 295: Previously if filters were passed down to DataFusion several of the `query_tests` tests would fail

tustvold:
@alamb Is there a reason this is just sitting here orphaned?

alamb:
> @alamb Is there a reason this is just sitting here orphaned?

@tustvold  yes it is because I haven't gotten around to writing a test as suggested by @NGA-TRAN  -- given I don't really expect this to have a major performance difference I haven't made writing one a priority

alamb:
@tustvold  has an alternate plan

PR #2846, opened 2021-10-14
feat: wire span context frontent=>WriteBuffer

crepererum:
This doesn't do anything really useful yet because the consumer side cannot really use it (yet) due to missing wiring. Hence the end2end test is missing.

### DIFF
--- a/server/src/database.rs
+++ b/server/src/database.rs
@@ -27,6 +27,7 @@ use snafu::{ensure, OptionExt, ResultExt, Snafu};
 use std::{future::Future, sync::Arc, time::Duration};
 use tokio::{sync::Notify, task::JoinError};
 use tokio_util::sync::CancellationToken;
+use trace::ctx::SpanContext;
 use trace::{RingBufferTraceCollector, TraceCollector};
 use uuid::Uuid;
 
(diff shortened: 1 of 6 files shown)


server/src/database.rs
```diff
@@ -562,7 +563,11 @@ impl Database {
     /// - write it to a write buffer
     /// - write it to a local `Db`
     ///
-    pub async fn write_entry(&self, entry: entry::Entry) -> Result<(), WriteError> {
+    pub async fn write_entry(
+        &self,
+        entry: entry::Entry,
+        span_ctx: Option<SpanContext>,
```
alamb, line 569: Is the idea that we will (eventually) always have a span context? Or that we will only have a span context when the request is being traced?

If the latter, what do you think about making a new struct like `RequestContext` while you are doing the plumbing to save additional plumbing work if additional context needs to get passed

like 

```rust
#[Derive(Default)]
struct RequestContext {
  span_ctx: Option<SpanContext>
  // space for future request specific context here (like "source URI" or whatever)
}
```

Or s
```

File: src/influxdb_ioxd/http.rs [hunk 2]
```diff
@@ -493,6 +493,7 @@ async fn write<M>(
 where
     M: ConnectionManager + Send + Sync + Debug + 'static,
 {
+    let span_ctx = req.extensions().get().cloned();
```
alamb, line 496: this is cool

PR #2850, opened 2021-10-14
fix: do not panic when writing to an unknown sequencer

crepererum:
This was only broken for the mock implementation but let's fix it and add a generic test.


write_buffer/src/core.rs
```diff
@@ -696,6 +697,26 @@ pub mod test_utils {
...
+        T: TestAdapter,
+    {
+        let context = adapter.new_context(NonZeroU32::try_from(1).unwrap()).await;
+
+        let entry = lp_to_entry("upc user=1 100");
+
+        let writer = context.writing(true).await.unwrap();
+
+        // flip bits to get an unknown sequencer
+        let sequencer_id = !set_pop_first(&mut writer.sequencer_ids()).unwrap();
```
alamb, line 712: TIL that `!` on i32 performs bitwise complement!

```rust
fn main() {
    let v = 42;
    println!("to v {} or not to v {}", v, !v);
}
```

```shell
to v 42 or not to v -43
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ea9417edc30fb05982f19ae2b0f650a5

write_buffer/src/core.rs
```diff
@@ -696,6 +697,26 @@ pub mod test_utils {
...
+
+        writer
```
alamb, line 714: I am always a fan of testing the actual error message string to make sure that the actual error that comes back is expected .

This can avoid situations where a future refactor might change the configuration requirements for the sequencer resulting in an error before `store_entry` even tried to write to the sequencer_id; In such cases this test would still pass

alamb: something like 
```rust
let result = writer.store_entry(...).await.expect_err();
assert!(result.to_string().contains("Unknown sequencer"));

crepererum: Filed #2856 since this also affects other errors.

PR #2862, opened 2021-10-15
fix: Serialize relative db object store paths to the server config

carols10cents:
So that they can be deserialized, without parsing, to create a new
iox object store from the location listed in the server config.

Notably, the locations serialized don't start with the object storage's
prefix like "s3:" or "file:". The location is the same object storage as
the server configuration that was just read from object storage. Having
the server config on one type of object storage and the database files
on another type is not supported.

This felt a bit strange when, in reviewing the last PR, @alamb asked for examples of the database locations written to the server in the protobuf, and I noticed they were being written with the `mem:` or `file:` etc prefixes, which seemed a bit odd. Then I tried to use the paths in #2851 and it didn't work out so well 😬 

This should fix it!


server/src/lib.rs
```diff
@@ -2180,7 +2174,7 @@ mod tests {
 
         // server config contains foo
         let config = server_config(application.object_store(), server_id).await;
-        assert_config_contents(&config, &[(&foo_db_name, fake_db_path.to_string())]);
+        assert_config_contents(&config, &[(&foo_db_name, String::from("1/foo/"))]);
```
alamb, line 2177: 👍

object_store/src/path.rs
```diff
@@ -180,4 +194,30 @@ mod tests {
             "mem:foo/bar/baz.json"
         );
     }
+
+    #[test]
+    fn test_to_raw_string() {
+        assert_eq!(
```
alamb, line 200: Given that paths without filenames are treated differently, I wonder if it worth also testing paths of the form `/foo/bar` and/or `/foo/bar/`

PR #2873, opened 2021-10-18
feat: file-based write write_buffer

crepererum:
Closes #2849.


write_buffer/src/file.rs
```diff
@@ -0,0 +1,717 @@
...
+            0
+        };
+
+        // try to link scratchpad file to "current" dir
+        loop {
+            let cur_file = cur.join(sequence_number.to_string());
+            if tokio::fs::hard_link(&pad_file, &cur_file).await.is_ok() {
+                break;
+            }
+            sequence_number = sequence_number
```
alamb, line 250: I wonder if it is worth putting some sort of other safety breaker in here (like only retry up to 100 times) or something.

alamb: makes sense 👍

PR #2879, opened 2021-10-18
feat: arrow_util truncate and bitset append

tustvold:
As part of #2724 I found I needed some further functionality of the arrow_util data structures. This PR adds this.


arrow_util/src/bitset.rs
```diff
@@ -35,8 +35,68 @@ impl BitSet {
         self.buffer.resize(new_buf_len, 0);
     }
 
+    /// Appends `count` set bits
+    pub fn append_set(&mut self, count: usize) {
```
tustvold, line 39: This is literally the `append_bits` function below, with all instances of `to_set[...]` replaced with `0xFF`

arrow_util/src/bitset.rs
```diff
@@ -35,8 +35,44 @@ impl BitSet {
...
+
+        self.buffer.resize(new_buf_len, 0xFF);
+
+        let rem = new_len & 7;
+        if rem != 0 {
+            *self.buffer.last_mut().unwrap() &= (1 << rem) - 1;
+        }
+
+        self.len = new_len;
+        debug_assert_eq!(self.buffer.len(), new_buf_len);
```
alamb, line 56: given the call to `self.buffer.resize` above I wonder how much value this assert adds.

arrow_util/src/bitset.rs
```diff
@@ -35,8 +35,44 @@ impl BitSet {
...
+    }
+
+    /// Truncates the bitset to the provided length
+    pub fn truncate(&mut self, len: usize) {
+        let new_buf_len = (len + 7) >> 3;
+        self.buffer.truncate(new_buf_len);
+        let overrun = len & 7;
+        if overrun > 0 {
+            if let Some(last) = self.buffer.last_mut() {
```
alamb, line 65: Can `self.buffer.last_mut()` ever return `None` here?  Given `len` must have been greater than `0` I don't think it could be

arrow_util/src/bitset.rs
```diff
@@ -252,6 +288,53 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_append_fuzz() {
+        let mut mask = BitSet::new();
+        let mut all_bools = vec![];
+        let mut rng = rand::thread_rng();
+
+        for _ in 0..100 {
+            let len = (rng.next_u32() % 50) as usize;
```
alamb, line 298: Not that it matters, but 0,1 and 2 bytes worth of bits is probably the most valuable (aka `rng.next_u32 % 24`) -- I am not sure how much additional coverage lengths greater than 24 give.

arrow_util/src/bitset.rs
tustvold: A very good point :+1:

arrow_util/src/bitset.rs
tustvold: Yeah this dates from a prior more complex formulation, will remove

PR #2881, opened 2021-10-18
test: Add read_group test for missing fields

alamb:
As suggested by @NGA-TRAN  on https://github.com/influxdata/influxdb_iox/pull/2874 at https://github.com/influxdata/influxdb_iox/pull/2874#discussion_r731043321:

Add an additional read_group test


query_tests/src/influxrpc/read_group.rs
```diff
@@ -925,3 +926,33 @@ async fn test_grouped_series_set_plan_group_measurement_tag_count() {
     )
     .await;
 }
+
+#[tokio::test]
+async fn test_grouped_series_set_plan_group_field_pred_and_null_fields() {
+    // no predicate
+    let predicate = PredicateBuilder::default().table("h2o").build();
```
NGA-TRAN, line 933: Will there be more coverage to add exact the same test but on "o2" table?. Since "o2" has one row with `city` and one without `city`, I wonder what the output looks like

PR #2886, opened 2021-10-18
chore: misc updates

tustvold:
Updates indexmap and a number of patch version bumps


object_store/Cargo.toml
```diff
@@ -16,7 +16,7 @@ chrono = "0.4"
 cloud-storage = {version = "0.10.3", optional = true}
 futures = "0.3"
 # https://github.com/tkaitchuck/aHash/issues/95
```
alamb, line 18: shall we update the comment too? Like remove it?

Issue #2889, opened 2021-10-18
fluxtest: count_with_nulls...fail: panic: interface conversion: array.Interface is *array.Uint64, not *array.Int64


alamb:
the `count_with_nulls` fluxtest fails with:
```
...fail: panic: interface conversion: array.Interface is *array.Uint64, not *array.Int64
```

The flux / expected is:
```

do_test = (every, fn) => array.from(
    rows: [
        {_time: 2019-11-25T00:00:00Z, _measurement: "m0", _field: "f0", t0: "a-0", _value: 1.0},
        {_time: 2019-11-25T00:00:15Z, _measurement: "m0", _field: "f0", t0: "a-0", _value: 2.0},
        {_time: 2019-11-25T00:00:30Z, _measurement: "m0", _field: "f0", t0: "a-0", _value: 3.0},
        {_time: 2019-11-25T00:00:45Z, _measurement: "m0", _field: "f0", t0: "a-0", _value: 4.0},
        {_time: 2019-11-25T00:00:00Z, _measurement: "m0", _field: "f0", t0: "a-1", _value: 1.0},
        {_time: 2019-11-25T00:00:15Z, _measurement: "m0", _field: "f0", t0: "a-1", _value: 2.0},
        {_time: 2019-11-25T00:00:30Z, _measurement: "m0", _field: "f0", t0: "a-1", _value: 3.0},
        {_time: 2019-11-25T00:00:45Z, _measurement: "m0", _field: "f0", t0: "a-1", _value: 4.0},
        {_time: 2019-11-25T00:00:00Z, _measurement: "m0", _field: "f0", t0: "a-2", _value: 1.0},
        {_time: 2019-11-25T00:00:15Z, _measurement: "m0", _field: "f0", t0: "a-2", _value: 2.0},
        {_time: 2019-11-25T00:00:30Z, _measurement: "m0", _field: "f0", t0: "a-2", _value: 3.0},
        {_time: 2019-11-25T00:00:45Z, _measurement: "m0", _field: "f0", t0: "a-2", _value: 4.0},
    ],
)
    |> group(columns: ["_measurement", "_field", "t0"])
    |> testing.load()
    |> range(start: 2019-11-25T00:00:00Z, stop: 2019-11-25T00:01:00Z)
    |> aggregateWindow(every, fn)
    |> drop(columns: ["_start", "_stop"])

testcase count_with_nulls {
    want = array.from(
        rows: [
[...]

PR #2906, opened 2021-10-19
fix(read_window_aggregates): return aggregates as integer rather than unsigned

alamb:
# Rationale:

Resolves https://github.com/influxdata/influxdb_iox/issues/2889

flux expects `count` to return `Integer` rather than `Unsigned` (even though counts can never be negative ??)

# Changes
1. cast count appropriately (same fix as applied to `read_group` in https://github.com/influxdata/influxdb_iox/pull/2756)

# fluxtest status

TDB


query_tests/src/influxrpc/read_window_aggregate.rs
```diff
@@ -166,6 +166,34 @@ async fn test_read_window_aggregate_nanoseconds_measurement_pred() {
...
+    let predicate = PredicateBuilder::default()
+        .timestamp_range(100, 450)
+        .build();
+
+    let agg = Aggregate::Count;
+    let every = WindowDuration::from_nanoseconds(200);
+    let offset = WindowDuration::from_nanoseconds(0);
+
+    let expected_results = vec![
+        "Series tags={_field=temp, _measurement=h2o, city=Boston, state=MA}\n  IntegerPoints timestamps: [200, 400, 600], values: [1, 2, 1]",
```
alamb, line 181: without the fix, these return "UnsignedPoints" rather than "IntegerPoints"

query/src/frontend/influxrpc.rs
```diff
@@ -1171,6 +1171,8 @@ impl InfluxRpcPlanner {
             .sort(sort_exprs)
             .context(BuildingPlan)?;
 
+        let plan_builder = cast_aggregates(plan_builder, agg, &field_columns)?;
```
NGA-TRAN, line 1174: 👍

NGA-TRAN (approved):
I like this kind of easy-to-fix bug :)

Issue #2890, opened 2021-10-18
fluxtest: range_nsecs_bare_max fails / panics: Error reading record batch while converting from SeriesSet: ExternalError(Canceled)


alamb:
```
range_nsecs_bare_max...fail: query error: rpc error: code = InvalidArgument desc = Error running grouping plans for database '54425547585a6420_df3882e0e03da61a': Execution error: Error executing series set conversion: Error reading record batch while converting from SeriesSet: ExternalError(Canceled)
```

Which appears to have caused a panic:
```
Oct 18 17:02:32.446 ERROR panic_logging: thread 'IOx Query Executor Thread' panicked at 'attempt to add with overflow', query/src/func/window/internal.rs:261:9
thread 'IOx Query Executor Thread' panicked at 'attempt to add with overflow', query/src/func/window/internal.rs:261:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Relevant flux: 
```
testcase range_nsecs_bare_max {
    input = array.from(
        rows: [
            {_time: 2021-01-01T00:00:01.000000001Z, _field: "foo", _value: 1.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000002Z, _field: "foo", _value: 2.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value: 3.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000024Z, _field: "foo", _value: 11.24, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000009Z, _field: "bar", _value: 4.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000011Z, _field: "bar", _value: 5.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000015Z, _field: "bar", _value: 6.0, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000022Z, _field: "bar", _value: 1.2, _measurement: "mm"},
            {_time: 2021-01-01T00:00:01.000000031Z, _field: "bar", _value: 0.28, _measurement: "mm"},
        ],
    )
        |> group(columns: ["_field", "_measurement"])
        |> testing.load()
    want = array.from(
        rows: [
            {_time: 2021-01-01T00:00:01.000000005Z, _field: "foo", _value: 3.0, _measurement: "mm"},
[...]

PR #2908, opened 2021-10-19
fix: read_window_aggregate overflow

alamb:
# Rationale

Fix https://github.com/influxdata/influxdb_iox/issues/2890

Previously the `range_nsecs_bare_max` test caused a panic in IOx

```
Oct 18 17:02:32.446 ERROR panic_logging: thread 'IOx Query Executor Thread' panicked at 'attempt to add with overflow', query/src/func/window/internal.rs:261:9
thread 'IOx Query Executor Thread' panicked at 'attempt to add with overflow', query/src/func/window/internal.rs:261:9
```

# Changes:
1. Don't panic by modeling the golang overflow behavior. See an example here of golang  overflowing when adding two i64s: https://play.golang.org/p/uOR7Nc_pTE0

image

# fluxtest scorecard

```
---
Ran 109 tests with 7 failure(s)
```


query/src/func/window/internal.rs
```diff
@@ -258,7 +258,11 @@ impl Add<Duration> for i64 {
         if d.negative {
             nsecs = -nsecs;
         }
-        new_t + nsecs
+        // new_t + nsecs
+        // follow the golang behavior and ignore overflow
+        // see https://github.com/influxdata/influxdb_iox/issues/2890
+        let (v, _overflow) = new_t.overflowing_add(nsecs);
```
NGA-TRAN, line 264: 👍

NGA-TRAN (approved):

alamb:
@e-dard  makes sense -- thank you for the clarification

PR #2913, opened 2021-10-20
fix(logfmts): Use UTC for ./scripts/logfmts

mkmik:
Local timezones and sharing logs with geographically distributed colleagues don't mix together.

### DIFF
--- a/scripts/logfmts
+++ b/scripts/logfmts
@@ -23,4 +23,6 @@ if ! command -v gawk &> /dev/null; then
   echo "  ...."
 fi
 
+export TZ=C
+
 exec gawk -F"time=" '{print strftime("%Y-%m-%d %H:%M:%S", $2/1000000000), $1 }'


e-dard (approved):
😻

scripts/logfmts
```diff
@@ -23,4 +23,6 @@ if ! command -v gawk &> /dev/null; then
   echo "  ...."
 fi
 
+export TZ=C
```
alamb, line 26:
```suggestion
# Use UTC time
export TZ=C
```

Also I am not versed in the values of `TZ` -- would the value of `UTC` work (it is certainly clearer)

alamb (approved):

scripts/logfmts
mkmik: >  alamb removed the automerge label 6 minutes ago
to late :-(

`UTC` would work, unless you run this script on a host with a broken timezone database, which is not unlikely nowadays with all these funny ways of building container images; not that's it's likely to run this script in anything other than your laptop, just sharing why I have the habit of using `C` which always works since it's compiled in.

PR #2916, opened 2021-10-20
fix: perf CI cache

tustvold:
Blindly caching the /target directory will grow indefinitely over time. Cache retrieval is now starting to take a non-trivial amount of time in CI jobs.

This CI job is not running in docker like the others and so cannot use the same caching mechanism, but given this is re-installing all the components each time I'm inclined to think that caching the cargo registry is not going to net a meaningful return. If this CI job starts to take too long, we should probably look at reusing the build artifacts of other jobs and/or dockerising the dependencies.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,16 +200,9 @@ jobs:
           name: Install perf.py dependencies
           command: python3 -m pip install -r perf/requirements.txt
 
-      - restore_cache:
-          keys:
-            - cargo-lock-{{ checksum "Cargo.lock" }}
       - run:
           name: Prime Rust build cache
           command: cargo build --package influxdb_iox --bin influxdb_iox --package iox_data_generator --bin iox_data_generator
-      - save_cache:
-          key: cargo-lock-{{ checksum "Cargo.lock" }}
-          paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -200,17 +200,6 @@ jobs:
           name: Install perf.py dependencies
           command: python3 -m pip install -r perf/requirements.txt
 
-      - restore_cache:
-          keys:
-            - cargo-lock-{{ checksum "Cargo.lock" }}
-      - run:
```
tustvold, line 206: This doesn't actually work as it builds with different options, removing avoids possibility of future drift

tustvold:
> Skipping cache generation, cache already exists for key: cargo-lock-BYwS+ZFvtRliA1OpmrCSIz8QazT1SPY_OwI3UIx9ZjI=

Aah it appears I misunderstood how Circle CI build caching functions, I had understood it as upserting the cache key each time, which isn't what it does. I guess I just keep changing the Cargo.lock file and seeing it pop up in my PRs :sweat_smile:  It should not grow indefinitely, and so I'm going to close this PR

PR #2900, opened 2021-10-19
feat: add MutableBatch::extend_from

tustvold:
Draft as builds on #2882

This PR adds the ability to merge MutableBatch together

PR #2917, opened 2021-10-20
feat: add MutableBatch::extend_from_range

tustvold:
Follows on from #2904 and adds `MutableBatch::extend_from_range` this is the base primitive that will be used to implement partitioning, filtering, sharding, etc... for the write path.

Unfortunately unlike extend_from in #2900 we can't reuse the statistics and have to recompute them. This is a touch unfortunate and the code a bit cumbersome, but we can optimize if it ever shows up in traces. I toyed with the idea of pulling statistics out of MutableBatch all together, but this caused issues for `MutableBuffer::table_summary` so I left it in.


arrow_util/src/bitset.rs
```diff
@@ -181,12 +181,30 @@ impl BitSet {
...
+/// at the provided bit offset
+pub fn iter_set_positions_with_offset(
+    bytes: &[u8],
+    offset: usize,
+) -> impl Iterator<Item = usize> + '_ {
+    let mut byte_idx = offset >> 3;
+    let mut in_progress = bytes.get(byte_idx).cloned().unwrap_or(0);
+
+    let skew = offset & 7;
+    in_progress &= 0xFF << skew;
```
crepererum, line 206: Within a byte, the bits are arranged LSB (pos=0) to MSB (pos=8)?

tustvold: Correct

PR #2919, opened 2021-10-20
chore: Enable 'backwards compatibility' protobuf linting

alamb:
# Rationale
Conductor uses these protobuf definitions, so any change that is not backwards compatible will certainly require conductor changes.

# Changes
1. Add buf [breaking change detector](https://docs.buf.build/breaking/overview) lint to make sure any obviously breakages in protobuf are flagged
2. Note how to workaround it (after discussion with conductor devs)

# Note
This will not detect more subtle conductor breakages in IOx/conductor, but it will catch a class of changes that will definitely break conductor

Example failure when I explicitly broke backwards compatibility: https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/11840/workflows/c420826a-5a38-4f4c-b6ee-a9dde3a3274f/jobs/82390


.circleci/config.yml
```diff
@@ -289,6 +289,16 @@ jobs:
       - run:
           name: buf lint
           command: buf lint
+      - run:
+          name: buf breaking changes
+          command: |
+            echo "If you want to make changes forbidden by this lint, please"
```
tustvold, line 295: Is there some way we could inspect the PR labels or something, to allow suppressing it that way? I don't know if CircleCI supports this

.circleci/config.yml
```diff
@@ -289,6 +289,16 @@ jobs:
...
+            echo "coordinate with the conductor team and disable this lint"
+            echo "for your breaking change and re-enable it in a subsequent PR"
+            git fetch origin main
+            # compare against only changes in this branch (not against
+            # other stuff that may have been added to master since last merge)
+            MERGE_BASE=$(git merge-base origin/main $CIRCLE_BRANCH) sh -c 'buf breaking --against ".git#ref=$MERGE_BASE"'
```
tustvold, line 301: For others reviewing we have

```
breaking:
  use:
    - WIRE
    - WIRE_JSON
```

alamb:
> Is there some way we could inspect the PR labels or something, to allow suppressing it that way? I don't know if CircleCI supports this

Let me see what I can do

alamb:
Here is a proposal for how to skip the check if a `incompatible protobuf` label is added: 
 https://github.com/influxdata/influxdb_iox/pull/2922

Will wait to see if @crepererum  / @domodwyer  want to weigh in on either approach as well

crepererum (approved):
Alright, let's discuss the skipping logic in #2922 and lets merge the basic feature.

domodwyer (approved):

PR #2925, opened 2021-10-21
refactor: use Expr Display impl

e-dard:
The `Display` trait was implemented in https://github.com/apache/arrow-datafusion/pull/971, so this PR removes an outstanding `TODO`.


alamb (approved):
🎉

PR #2982, opened 2021-10-29
feat: add `RemoteService`

crepererum:
Ref #2980.

### DIFF
--- a/generated_types/protos/influxdata/iox/management/v1/service.proto
+++ b/generated_types/protos/influxdata/iox/management/v1/service.proto
@@ -44,13 +44,19 @@ service ManagementService {
   rpc ListChunks(ListChunksRequest) returns (ListChunksResponse);
 
   // List remote IOx servers we know about.
-  rpc ListRemotes(ListRemotesRequest) returns (ListRemotesResponse);
+  rpc ListRemotes(ListRemotesRequest) returns (ListRemotesResponse) {
+    option deprecated = true;
+  };
 
   // Update information about a remote IOx server (upsert).
-  rpc UpdateRemote(UpdateRemoteRequest) returns (UpdateRemoteResponse);
+  rpc UpdateRemote(UpdateRemoteRequest) returns (UpdateRemoteResponse) {
+    option deprecated = true;
+  };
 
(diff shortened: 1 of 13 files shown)


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -44,13 +44,19 @@ service ManagementService {
   rpc ListChunks(ListChunksRequest) returns (ListChunksResponse);
 
   // List remote IOx servers we know about.
-  rpc ListRemotes(ListRemotesRequest) returns (ListRemotesResponse);
+  rpc ListRemotes(ListRemotesRequest) returns (ListRemotesResponse) {
+    option deprecated = true;
```
alamb, line 48: Is the idea that once conductor is updated to use the new APIs, we can remove these deprecated ones?

If so, I suggest leaving a comment (with ticket reference) here with the backstory for the archaeological purposes of future readers

cc @domodwyer

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/mod.rs
```diff
@@ -51,6 +52,12 @@ where
         operations::make_server(Arc::clone(server_type.application.job_registry()))
     );
 
+    // remotes can be managed even if the server is not serving
```
alamb, line 55: is the eventual plan to only add the relevant services for a particular `server_type`? (Aka only make a management server for `database` and only make remove service for `router`?

alamb (approved):
Looks good to me -- as I understand it this is effectively a copy/paste of the existing part of the management API (and a subsequent PR will remove the portions from the management API)

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/mod.rs
crepererum: yes

generated_types/protos/influxdata/iox/management/v1/service.proto
crepererum: yes, done

PR #2983, opened 2021-10-29
feat: add `DeleteService`

crepererum:
Ref #2980.


generated_types/protos/influxdata/iox/delete/v1/service.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+package influxdata.iox.delete.v1;
+option go_package = "github.com/influxdata/iox/delete/v1";
+
+service DeleteService {
```
alamb, line 5: What would you think about calling this `DMLService`  (for Data Manipulation Langage, the standard term in databases for Insert/update/delete/ etc)? Or maybe `DataService`

That API could also include what is currently called the "flight" API (but is really the query apu) as well as the "write" API

However, I can see the value in having separate services for Delete and for Query and Write as well

What do you think?

alamb (approved):
👍

FWIW I think this is following the same pattern as https://github.com/influxdata/influxdb_iox/pull/2982

generated_types/protos/influxdata/iox/delete/v1/service.proto
crepererum: I think we shouldn't mix it w/ the flight API, but there's currently a `WriteService` which could be used for that.

Issue #570, opened 2020-12-17
Wrong results with first/last selector functions with null first/last selector functions


alamb:
I think there is a bug in the implementation of the first/last selector functions introduced here: https://github.com/influxdata/influxdb_iox/pull/565

Given this input which represents data where there is is no value  for `field2` at time=100
```
÷--------+--------+------+
| field1 | field2 | time |
+--------+--------+------+
| 70.6   |        | 100  |
| 70.4   | ss     | 200  |
| 70.5   | sa     | 300  |
+--------+--------+------+
```

Then the expected results I think should be as follows:
`first(field1, time)` -->  `(70.6, 100)`
`first(field2, time)` --> `(ss, 200)` 

However, the implementation in #556 is going to return 
`first(field2, time)` --> `(NULL, 300)` 

Which I don't think is correct

The tasks of this ticket is to:
1. Verify the actual expected behavior using InfluxDB
2. Fix the selector implementation in IOx to match

The implementation of the first / last selectors probably need to be updated to filter for nulls first before finding the min time

PR #2988, opened 2021-10-29
fix: Only use timestamps in first/last when there is a corresponding …

alamb:
# Rationale
Resolves https://github.com/influxdata/influxdb_iox/issues/570

# Changes
Only update the timestamp value for selectors `first` and `last` if there is an actual value. The logic for `min` / `max` selectors already take this into account as they operate row-wise.

# Example 
Here is an example of the issue. Prior to this PR, `first(value, time)` for this input 
```
value | time
--------------
NULL  | 100
A     | 200
B     | 300
```

would return `(NULL, 100)`

After this PR it returns `(A, 200)` correctly.

I am sure this code could be made faster (and reduce copy/pasta) but I think it is fine for now and went with the easy fix.


query/src/func/selectors/internal.rs
```diff
@@ -147,6 +147,26 @@ macro_rules! make_first_selector {
                     // the input type arguments should be ensured by datafusion
                     .expect("Second argument was time");
 
+                // Only look for times where the array array's NULL
+                // mask for the time array(time array should have no
+                // nulls)
```
carols10cents, line 152: This comment says "array array", that's not right, is it? And the sentence is missing the end of the "where" phrase -- only look for the times where the null mask... is what?

query/src/func/selectors/internal.rs
```diff
@@ -235,6 +255,26 @@ macro_rules! make_last_selector {
                     // the input type arguments should be ensured by datafusion
                     .expect("Second argument was time");
 
+                // Only look for times where the array array's NULL
```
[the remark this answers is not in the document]
carols10cents: Copy pasta of "array array" and the unfinished "where" phrase too 😢

query_tests/src/influxrpc/read_group.rs
```diff
@@ -540,6 +540,39 @@ async fn test_grouped_series_set_plan_first() {
...
+        // "h2o,state=MA,city=Boston temp=70.4,moisture=43.0 100000"
+        .table("h2o")
+        .add_expr(col("state").eq(lit("MA")))
+        .add_expr(col("city").eq(lit("Boston")))
+        .build();
+
+    let agg = Aggregate::First;
+    let group_columns = vec!["state"];
+
+    // expect timestamps to be present for
```
carols10cents, line 558: present for what?

carols10cents (approved):
Just some suggestions on comments that seemed unfinished. Code looks good!

query/src/func/selectors/internal.rs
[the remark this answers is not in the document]
alamb: 🥲

query/src/func/selectors/internal.rs
alamb: Fixed in fff3873d

query/src/func/selectors/internal.rs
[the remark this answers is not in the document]
alamb: fff3873d

query_tests/src/influxrpc/read_group.rs
alamb: Fixed in 3cab572a

PR #2990, opened 2021-10-30
fix: Handle errors and rule updates correctly with UUIDs

carols10cents:
Fixes #2975.

In testing and fixing some problems [in the UUID PR](https://github.com/influxdata/influxdb_iox/pull/2964) I found more problems to fix 😬 Andrew found some problems in that PR that I thought were just incorrect error handling, but I found out that updating database rules was updating the UUID 😰 because I was trying to reuse types too much. So I've split out a `PersistedDatabaseRules` type from `ProvidedDatabaseRules` (I'm not super pleased they're both `P...DatabaseRules`, better name suggestions much appreciated!). `PersistedDatabaseRules` contains `ProvidedDatabaseRules` plus the `UUID`, which better matches the proto type that's serialized to the rules file. A database's UUID is only assigned one place now, in server `create_database`.

This PR also makes the change Andrew suggested that we read the rules file on database restore to get the database name, and then read the rules file again during database initialization, so that you only have to specify the UUID and not the name too when restoring a database.


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -197,8 +197,9 @@ message RestoreDatabaseRequest {
   reserved 1;
   reserved "generation_id";
 
-  // the name of the database
-  string db_name = 2;
+  // Was the name of the database
+  reserved 2;
```
alamb, line 201: 🎉

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/management.rs
```diff
@@ -200,11 +199,10 @@ where
         request: Request<RestoreDatabaseRequest>,
     ) -> Result<Response<RestoreDatabaseResponse>, Status> {
         let request = request.into_inner();
-        let db_name = DatabaseName::new(request.db_name).field("db_name")?;
         let uuid = Uuid::from_str(&request.uuid).field("uuid")?;
 
         self.server
-            .restore_database(&db_name, uuid)
+            .restore_database(uuid)
```
alamb, line 205: I think now that the `name` is no longer needed to restore the database, we can also remove it from the CLI:

At the moment, the CLI still requires a `db_name` argument (which is ignored)
```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore ad3543bc-8bf6-4303-83b0-03c8a7fbb53f
error: The following required arguments were not provided:
    <uuid>

USAGE:
    influxdb_iox database restore <name> <uuid> --host <host>

For more information try --help
```

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -338,6 +338,22 @@ async fn delete_database() {
         .assert()
         .success()
         .stdout(predicate::str::contains(db));
+
+    // Restoring again is an error
+    Command::cargo_bin("influxdb_iox")
+        .unwrap()
+        .arg("database")
+        .arg("restore")
+        .arg(db)
```
alamb, line 347: As above, I don't think the `db` name is required here anymore

server/src/database.rs
```diff
@@ -1288,6 +1310,7 @@ struct DatabaseStateCatalogLoaded {
     lifecycle_worker: Arc<LifecycleWorker>,
     replay_plan: Arc<Option<ReplayPlan>>,
     provided_rules: Arc<ProvidedDatabaseRules>,
+    uuid: Uuid,
```
alamb, line 1313: I wonder if it would be cleaner / clearer to pass the `uuid` to `Db::new()` and keep it as a field -- rather than carrying it along in these db state machines?

alamb (commented):
👍  I tried out delete/ restore with a bunch of fat finger commands and it works much better now 👍 

I also tested that I could delete a db, restart the server and then restore the db

FWIW I think the naming of `PersistedDatabaseRules` and `ProvidedDatabaseRules` are quite clear (but I am biased as I introduced the `Provided` terminology orignially). I like awesome alliteration 😄 




```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore foo ad3543bc-8bf6-4303-83b0-03c8a7fbb53f
Error restoring database: Unexpected server error: Some entity that we attempted to create already exists: The database with UUID `ad3543bc-8bf6-4303-83b0-03c8a7fbb53f` named `my_awesome_db` is already active
```


# New Internal Error:
I also found if I provided a valid, but unknown UUID the CLI still reports an internal error:

```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore foo ad3543bc-8bf6-4303-83b0-03c8a7fbb533
Error restoring database: Unexpected server error: Internal error: Internal Error
```

In this case, the log says
```
2021-11-01T18:22:49.112161Z ERROR influxdb_iox::influxdb_ioxd::rpc::error: Unexpected error error=DatabaseRulesNotFound { uuid: ad3543bc-8bf6-4303-83b0-03c8a7fbb533, location: "42/ad3543bc-8bf6-4303-83b0-03c8a7fbb533/rules.pb" }
```

I don't think this is a critical defect (it is going to be an error, just percolating a nicer message to the CLI)

server/src/lib.rs
```diff
@@ -777,12 +789,30 @@ where
             let state = self.shared.state.read();
             let initialized = state.initialized()?;
 
-            if initialized.databases.contains_key(db_name) {
-                return DatabaseAlreadyExists { db_name }.fail();
-            }
             initialized.server_id
         };
 
+        let rules_bytes = IoxObjectStore::load_database_rules(
```
alamb, line 795: Is it worth a comment here saying this object store load is just for the db name and that the rules are reloaded later?

Or perhaps factoring this code into a function that does so to make it clearer?

alamb (approved):
sorry -- hit return too fast. I think this PR is good to merge in as is 👍  Nice work @carols10cents 

I did find another of the "internal error bugs" if you create a database, delete it, create a new db with the same name, delete that and then try and restore them both, the second restore generates an internal error:

```shell
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database delete db2
Deleted database db2
3347ad00-5998-434a-afb8-650b28248786
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database create db2
Created database db2 (ffa3b8e8-a20f-4510-8051-7f17c61dcc6d)
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database delete db2
Deleted database db2
ffa3b8e8-a20f-4510-8051-7f17c61dcc6d

(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore foo 3347ad00-5998-434a-afb8-650b28248786
Restored database 3347ad00-5998-434a-afb8-650b28248786
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore foo ffa3b8e8-a20f-4510-8051-7f17c61dcc6d
Error restoring database: Unexpected server error: Internal error: Internal Error
```

(this maybe the same thing as restoring the same uuid twice)

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/management.rs
carols10cents: ooooooops good catch!

server/src/lib.rs
carols10cents: Ooooh I like extracting into a function; there are like 4 errors that can happen in this little bit of code but nowhere else in this file; I can give the function its own error type :)

server/src/database.rs
carols10cents: The problem is when we load the databases on server startup, what we get in the server config file is the database name and the object store location for that database (which does contain the uuid, but I'm trying to avoid parsing those locations for meaning so that we can continue to change them in the future). We don't actually get the UUID [until the state machine reads the rules](https://github.com/influxdata/influxdb_iox/pull/2990/files#diff-805c83779fd625ae8d65cb912b1233551760a669c7d59c3f303a325c0da9fc9bR1232).

alamb: makes sense 👍

alamb (approved):
The changes look great to me

carols10cents:
Whew, finally got all these errors right, I think 🤞🏻

PR #2998, opened 2021-11-01
feat: add `DummyJobService`

crepererum:
Ref #2980.

### DIFF
--- a/generated_types/build.rs
+++ b/generated_types/build.rs
@@ -20,6 +20,7 @@ fn main() -> Result<()> {
 ///
 /// - `com.github.influxdata.idpe.storage.read.rs`
 /// - `influxdata.iox.catalog.v1.rs`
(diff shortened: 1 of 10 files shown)


tustvold:
Do we expect a non-database server to have long-running operations and by extension the need for this service. Just wondering why we need to split this service out of ManagementService?

crepererum:
> Do we expect a non-database server to have long-running operations and by extension the need for this service.
> Just wondering why we need to split this service out of ManagementService?

See #2998. The `ManagementService` contains functionality that doesn't really fall into the same category, hence I would like to kill it and move every method into a service that actually describes what's going on. 

There might be some compaction service or cleanup service in the future that could implement job/operations control.

crepererum:
@tustvold and I agreed that we're not going to split off this functionality and rather leave it closer to the database.

PR #3003, opened 2021-11-01
feat: migrate write buffer consumer to use DbWrite (#2724)

tustvold:
This migrates the WriteBufferConsumer to use `DbWrite` instead of `SequencedEntry`.

In practical terms this is just moving where the conversion takes place, but it is a step towards abstracting the write buffer away from flatbuffers


mutable_batch/src/payload.rs
```diff
@@ -204,12 +225,41 @@ pub struct DbWrite {
...
+    min_timestamp: i64,
+    max_timestamp: i64,
 }
 
 impl DbWrite {
     /// Create a new [`DbWrite`]
+    ///
+    /// # Panic
+    ///
+    /// Panics if
```
tustvold, line 237: These invariants are already enforced by the conversion logic in `sequenced_entry_to_writes`

write_buffer/src/core.rs
```diff
@@ -167,6 +169,22 @@ pub mod test_utils {
...
+        sequencer_id: u32,
+        span_context: Option<&SpanContext>,
+    ) -> DbWrite {
+        let (s, t) = writer
+            .store_entry(&entry, sequencer_id, span_context)
+            .await
+            .unwrap();
+
+        let sequenced = SequencedEntry::new_from_sequence(s, t, entry);
+        mutable_batch_entry::sequenced_entry_to_write(&sequenced).unwrap()
```
crepererum, line 185: Is migrating the write buffer producer some follow-up work?

crepererum (approved):

write_buffer/src/core.rs
tustvold: Yes

crepererum: No worries, I was assuming that and just wanted to check :+1:

mutable_batch/src/payload.rs
```diff
@@ -204,12 +225,41 @@ pub struct DbWrite {
...
+    ///
+    /// - `tables` is empty
+    /// - a MutableBatch is empty
+    /// - a MutableBatch lacks an i64 "time" column
     pub fn new(tables: HashMap<String, MutableBatch>, meta: WriteMeta) -> Self {
-        Self { tables, meta }
+        assert_ne!(tables.len(), 0);
```
alamb, line 243: I am a big fan of invariant checking, FWIW

PR #3004, opened 2021-11-01
feat: add `DeploymentService`

crepererum:
Ref #2980.


tustvold (approved):

alamb (commented):
For what it is worth, I really like the methodical "pull out one service at a time" approach this PR is taking to splitting up the APIs 👍

Issue #2738, opened 2021-10-05
Spike: Protobuf vs Flatbuffer Performance Evaluation


tustvold:
Prior to #2724 we want to quantify the performance differential between the two to quantify what we might lose by making the switch.

In particular the following should be considered and where possible measured:

* Flatbuffer verification speed
* Protobuf decode speed
* Compaction throughput
* Write buffer throughput

Previously the performance of JSON vs flatbuffers was evaluated - https://github.com/influxdata/influxdb_iox/pull/606

alamb:
I personally suggest measuring Compaction throughput and write buffer throughput (as a proxy for recovery speed) is all that matters -- the flatbuffer verification / protobuf decode speed is a secondary concern (as it is included in the other two measurements)

tustvold:
> the flatbuffer verification / protobuf decode speed is a secondary concern (as it is included in the other two measurements)

Indeed, I only split them up to allow confirming if there is an order of magnitude difference before spending the time to plumb the protobuf into the write path. Currently the write protocol is converted to a flatbuffer entry and so it wouldn't be an entirely fair comparison

alamb:
When you say "Protobuf Binary Write Protocol" is that referring to https://github.com/influxdata/influxdb-pb-data-protocol that @jacobmarble  worked on ? 

If so, I think the idea if 'switching to use protobuf after adding a dictionary encoding mechanism' makes all sorts of sense to me. It continues the high level idea from @pauldix  (and I think Jacob) to push some of the required ingest computation to the edge. 👍

tustvold:
> When you say "Protobuf Binary Write Protocol" is that referring to https://github.com/influxdata/influxdb-pb-data-protocol

Yes, although it is perhaps worth noting these benchmarks are not doing any of the optimisations listed there as I wasn't aware of them - I just read the proto spec :laughing: The performance difference of these is likely to be incredibly marginal if anything, so I'm confident these benchmarks still stand

tustvold:
@jacobmarble Correct

PR #3006, opened 2021-11-01
feat: add alternative string encodings to pbdata (#2724)

tustvold:
This PR adds two alternative string encodings to pbdata. Benchmarks in #2738 showed that these lead to a significant performance uplift vs the naive encoding as "repeated string". 

Note: the current pbdata gRPC write path will not make use of this code yet, as the pbdata -> entry conversion is separate code. As I intend to remove this code once the write path replumbing is complete, I've not updated this. If you attempt to write packed string or interned string data to IOx, it will interpret this as no rows for the column and error.

I will update https://github.com/influxdata/influxdb-pb-data-protocol once this PR is in


generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
```diff
@@ -31,6 +31,33 @@ message TableBatch {
...
 
+// A list of strings encoded as a contiguous string payload and a list of
+// byte offsets. The offset at index i records the byte offset of the start
+// of the i'th string, with index (i+1) recording the exclusive end offset
+//
+// For example ["hello", "influx", "", "data"] would be encoded as
+// - values: "helloinfluxdata"
+// - offsets: [0, 5, 11, 11, 15]
+//
+message PackedString {
```
tustvold, line 42: I'm not sure if this should be PackedString or PackedStrings?

mutable_batch_pb/src/lib.rs
```diff
@@ -97,16 +112,55 @@ pub fn write_table_batch(batch: &mut MutableBatch, table_batch: &TableBatch) ->
...
-                valid_mask,
-                values.string_values.iter().map(|x| x.as_str()),
-            ),
-            InfluxColumnType::Field(InfluxFieldType::String) => writer.write_string(
-                &column.column_name,
-                valid_mask,
-                values.string_values.iter().map(|x| x.as_str()),
-            ),
+            InfluxColumnType::Tag => {
+                if let Some(interned) = values.interned_string_values.as_ref() {
```
alamb, line 116: As written this seems like it will allow an input that has both `interned_string_values` and `packed_string_values` set. Perhaps we should check / disallow that case?

mutable_batch_pb/src/lib.rs
```diff
@@ -97,16 +112,55 @@ pub fn write_table_batch(batch: &mut MutableBatch, table_batch: &TableBatch) ->
...
+                            .map(|x| packed_string_idx(dictionary, *x as usize)),
+                    )
+                } else if let Some(packed) = values.packed_string_values.as_ref() {
+                    validate_packed_string(&column.column_name, packed)?;
+                    writer.write_string(&column.column_name, valid_mask, packed_string_iter(packed))
+                } else {
+                    writer.write_string(
+                        &column.column_name,
+                        valid_mask,
+                        values.string_values.iter().map(|x| x.as_str()),
```
alamb, line 160: it is unfortunate that we can't pass the strings directly in (aka that this function doesn't take ownership of the `table_batch` and pass these strings along)

mutable_batch_pb/src/lib.rs
```diff
@@ -127,6 +181,56 @@ pub fn write_table_batch(batch: &mut MutableBatch, table_batch: &TableBatch) ->
...
+    &strings.values[start_offset..end_offset]
+}
+
+/// Returns an iterator over the strings in a [`PackedString`]
+///
+/// # Panic
+///
+/// If the offsets array is not an increasing sequence of numbers less than
+/// the length of the strings array
+fn packed_string_iter(strings: &PackedString) -> impl Iterator<Item = &str> + '_ {
```
alamb, line 224: this appears to handle `PackedString`s with empty `offsets: vec![]` correctly, but it might be cool to add a test for that case explicity

alamb (approved):
While reading this PR I was wondering "do we really need both the `PackedString` and `InternedString` representation.

It seems like `PackedString` will be better for high cardinality values (e.g. pod ids or whatever) and `InternedString` would be better for low cardinality values so both make sense. 

However, I suspect `PackedString` and `String` are likely about the same (size and processing time) for high cardinality strings, so having both of them doesn't make sense to me.

alamb:
I think @jacobmarble  / @pauldix  might want to be heads up'ed (that isn't English I know) of this change to the pbdata structs...

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
```diff
@@ -31,6 +31,33 @@ message TableBatch {
...
+// indexes into this dictionary
+//
+// For example ["tag1", "tag1", "tag2", "tag2", "tag1"] might be encoded as
+// - dictionary: ["tag1, "tag2"]
+// - offsets: [0, 0, 1, 1, 0]
+//
+// There are no requirements that the dictionary is sorted, or its keys unique
+message InternedString {
+    PackedString dictionary = 1;
+    repeated uint32 values = 2;
```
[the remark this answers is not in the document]
tustvold: Protobuf already does varint encoding of integers, so it should do this automatically where possible - https://developers.google.com/protocol-buffers/docs/encoding#types

mutable_batch_pb/src/lib.rs
tustvold: MutableBatch stores string values as a PackedStringArray and so would have to copy the strings regardless

tustvold:
> However, I suspect PackedString and String are likely about the same (size and processing time) for high cardinality strings

I'm not sure I agree with this, the return of PackedString is avoiding allocating intermediate strings on decode combined with better data-locality - it's a single `(String, Vec<u32>)` vs `Vec<String>`. The cardinality wouldn't change this? This was borne out by the benchmarks on #2738 where PackedString was necessary to bring the protobufs inline with flatbuffer

> do we really need both the PackedString and InternedString representation

The `PackedString` type is necessary to define the `InternedString` type, but if by this you mean don't have the packed_string_values field - perhaps :shrug: I didn't think it hurt, and it avoids using a dictionary that isn't really a dictionary :grin: 

> Seems like we should remove string_values and always favor packed string (backwards compatibility aside)

I think its worth keeping for clients that don't need to squeeze out the last drop of performance, it will still be an order of magnitude faster than raw line protocol?

mutable_batch_pb/src/lib.rs
alamb: that is a good point

mutable_batch_pb/src/lib.rs
tustvold: Column type deduction will now fail if there are multiple values provided

alamb:
> I think its worth keeping for clients that don't need to squeeze out the last drop of performance, repeated string is slightly easier for clients to use and will still be an order of magnitude faster than raw line protocol?

Basically I think the tradeoff is "the client has more choices in how to encode string data" -- depending on your point of view that is either a good thing or a bad thing 😆

mutable_batch_pb/src/lib.rs
alamb: I must be missing something -- if the input has data in both the `interned_string_values` and `packed_string_vaues` field, won't this code insert data from `intered_string_values` and ignore the data in `packed_string_values`?

tustvold: Sorry I should have been more clear, the validation takes place earlier - https://github.com/influxdata/influxdb_iox/pull/3006/files#diff-ae16418a09c839b7fbb42373ec669ecba5581d946657e9bf3f5f6aa203b73866R290

This means it catches all variants of multiple types of data provided, and not just multiple different types of string

alamb: ah -- awesome

alamb: that is a clever mechanism, btw

PR #3008, opened 2021-11-02
refactor: decouple new router server configs from old management system

crepererum:
See commits. This would allow me to implement the router server on a more stable foundation.

Ref #2980.

### DIFF
--- a/generated_types/protos/influxdata/iox/router/v1/shard.proto
+++ b/generated_types/protos/influxdata/iox/router/v1/shard.proto
@@ -1,0 +1,48 @@
+syntax = "proto3";
+package influxdata.iox.router.v1;
+option go_package = "github.com/influxdata/iox/router/v1";
+
+// `ShardConfig` defines rules for assigning a line/row to an individual
+// host or a group of hosts. A shard
+// is a logical concept, but the usage is meant to split data into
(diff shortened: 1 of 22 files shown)


generated_types/protos/influxdata/iox/router/v1/shard.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
```
tustvold, line 1: Is it intentional that this appears to be a duplicate instead of a move?

tustvold (approved):
This seems to just move the write buffer stuff into a write_buffer package and the routing stuff into a routing package with no major code changes. Could you maybe highlight any areas where that isn't the case if it isn't :sweat_smile:

crepererum:
> This seems to just move the write buffer stuff into a write_buffer package and the routing stuff into a routing package with no major code changes. Could you maybe highlight any areas where that isn't the case if it isn't sweat_smile


For `shard.proto` it's not a move but a but a simplified copy.

crepererum:
Protobuf is actually compatible:

1. The package move is OK because we don't put this information into `Any`.
2. The router-related changes are OK because this stuff is not used (yet).

PR #3009, opened 2021-11-02
feat: pbdata encode (#2724) 

tustvold:
~~Draft as builds on #3006~~

This PR adds the ability to encode a MutableBatch as pbdata


mutable_batch_pb/src/lib.rs
```diff
@@ -11,779 +11,5 @@
     clippy::clone_on_ref_ptr
 )]
 
-use hashbrown::HashSet;
```
tustvold, line 14: This code is just moved to decode.rs

mutable_batch_pb/src/decode.rs
```diff
@@ -0,0 +1,778 @@
+//! Code to decode [`MutableBatch`] from pbdata protobuf
```
tustvold, line 1: This code is moved from lib.rs, unfortunately the Github diff isn't smart enough to pick up on this

tustvold:
Will split out changing the crate hierarchy

alamb (approved):

PR #3010, opened 2021-11-02
fix: dockerfiles should select database server mode by default

crepererum:


alamb (approved):

alamb:
I think @mkmik  / @domodwyer  should approve this PR prior to merging (mostly so they have a heads up / plan for how to deal with different modes)

mkmik (approved):

PR #3014, opened 2021-11-02
refactor: move pbdata decode into module

tustvold:
Split out from #3009 this just moves the pbdata decode logic into a separate module. This does not change any code


tustvold:
#3009  was approved so just going to move ahead with that

Issue #2436, opened 2021-08-29
Static Configuration


tustvold:
**Use Case:**

Currently IOx can only be provided with limited configuration at startup, specifically only server-global settings. This makes it tricky to manage declarative deployments of IOx, requiring some sort of sidecar or controller to push configuration into the IOx server once it has started.

It would be beneficial for integration tests, benchmarks, and possibly even some simple deployment scenarios, to be able to run IOx in a mode where the DatabaseRules are sourced from a set of configuration file(s) on startup. This would put IOx inline with most cloud-native tooling such as nginx, envoy, prometheus, telegraf, datadog-agent, etc...

**Proposal:**

Add a command line option `--conf-file` that instructs IOx to use the provided configuration file for database configuration.

Running with this option would disable the management API endpoints to mutate/delete database configuration, and would disable persistence of the rules files to object storage.

This JSON/YAML/TOML file would contain a list of DatabaseRules for this IOx instance. Additionally to avoid confusion with the database deletion tombstones, it should include the object store prefix for this database.

**Auto-reload:**

Not necessary for the first version, but the ability to auto-reload the configuration on update would make for a very compelling experience when combined with Kubernetes ConfigMaps. These can be mounted as regular files, and updated declaratively using standard Kubernetes tooling. When a ConfigMap is updated, the file mounted into any running containers is also updated. This is a very common pattern in kubernetes for services where restarts are not always desirable, e.g. ingress controllers.

Add a command line option `--autoreload` that instructs IOx to watch the provided configuration file, and apply any updates. If the config is invalid, or a database cannot be updated, the system should continue to operate with the old configuration and repeatedly log an error until the situation is resolved. 

Of course if this IOx replica is then restarted, it will lose this old configuration and be unable to startup with the new configuration, but this is a compromise - we avoid a typo bringing down the cluster at the cost of additional operator overhead. FWIW [nginx has the same behaviour](https://nginx.org/en/docs/control.html).

If a database is removed from the configuration, IOx should just forget about it - it should not write a delete tombstone as we cannot guarantee this will always take place (e.g. if the server is restarted before it has time to write it).

**Desired Behaviour:**

It should be possible to bring up an IOx instance, with statically configured databases

**Alternatives Considered:**

A bash pod that is provided with the configuration, and on startup waits for the server to start and then writes it out to the management API

alamb:
sounds like a useful feature to me, though one that would be non trivial to implement

tustvold:
[...]

PR #3016, opened 2021-11-03
feat: implement basic router server type

crepererum:
This only implements the deployment service and nothing else yet.


influxdb_iox/tests/common/server_fixture.rs
```diff
@@ -137,10 +163,26 @@ impl ServerFixture {
 
     /// Create a new server fixture and wait for it to be ready. This
     /// is called "create" rather than new because it is async and
-    /// waits. The database is left unconfigured (no writer id) and
+    /// waits. The database is left unconfigured (no server id) and
+    /// is not shared with any other tests.
+    pub async fn create_single_use_database() -> Self {
```
tustvold, line 168: This name is kind of confusing, it makes me think it is creating a `Database` not a `Server` configured in `Database` mode or whatever... I wonder if we could just have `create_single_use` with `ServerType` passed as argument?

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
```diff
@@ -112,8 +112,7 @@ where
     }
 
     fn shutdown_background_worker(&self) {
-        self.server.shutdown();
```
tustvold, line 115: I'm not a massive fan of this interface tbh, it seems to be copying the pattern of Db of having the caller own the background worker - which I realise in hindsight was a mistake that adds non-trivial complexity.

Not something for this PR, but I wonder if we could make the `ServerType` implementations own their own background workers much like `Server`, `Database`, `WriteBufferConsumer`, etc... ?

In fact I'm not entirely sure why `ServerType` needs its own background worker at all and can't just proxy through to `Server`...

influxdb_iox/src/influxdb_ioxd/server_type/router/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+        self::http::route_request(self, req).await
+    }
+
+    async fn server_grpc(self: Arc<Self>, builder_input: RpcBuilderInput) -> Result<(), RpcError> {
+        self::rpc::server_grpc(self, builder_input).await
+    }
+
+    async fn background_worker(self: Arc<Self>) {
+        self.shutdown.cancelled().await;
+        // no background worker
```
tustvold, line 63: Why have a shutdown token then? Perhaps see previous comment r.e. this interface

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/deployment.rs
```diff
@@ -0,0 +1,79 @@
...
+
+struct DeploymentService {
+    server: Arc<RouterServer>,
+    serving_readiness: ServingReadiness,
+}
+
+use crate::influxdb_ioxd::serving_readiness::ServingReadiness;
+
+#[tonic::async_trait]
+impl deployment_service_server::DeploymentService for DeploymentService {
```
tustvold, line 18: I'm really not a fan of having two implementation of DeploymentService one for router nodes and one for database nodes...

router/src/server.rs
```diff
@@ -0,0 +1,102 @@
...
+#[derive(Debug, Snafu)]
+#[allow(missing_copy_implementations)]
+pub enum SetServerIdError {
+    #[snafu(display("Server ID already set: {}", server_id))]
+    AlreadySet { server_id: ServerId },
+}
+
+/// Main entry point to manage a router node.
+#[derive(Debug)]
+pub struct RouterServer {
```
tustvold, line 18: Is this going to need to have its own state machine to parallel server::Server? I'm not sure what we gain by doing this?

tustvold (commented):
I have some reservations about completely splitting apart Routing from Database in this way, as I think it is going to result in a lot of code duplication? Additionally as we're likely to end up with further roles in the future, this feels like it won't scale very well?

I had perhaps envisaged router would just be an object on Server or possibly Database, much like we may have a WriteBufferConsumer or a Db or whatever....

crepererum:
> I had perhaps envisaged router would just be an object on Server or possibly Database, much like we may have a WriteBufferConsumer or a Db or whatever....

I don't think a router and a database should live in the same crate at all. They are totally different concepts. If they need code sharing, then that can be arranged in a different way.

influxdb_iox/src/influxdb_ioxd/server_type/router/mod.rs
crepererum: stale comment, will fix.

influxdb_iox/tests/common/server_fixture.rs
crepererum: > I wonder if we could just have `create_single_use` with `ServerType` passed as argument?

That's definitely possible, they way chosen here just resulted in less refactoring. I'll try the argument-driven approach.

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
crepererum: > In fact I'm not entirely sure why `ServerType` needs its own background worker at all and can't just proxy through to `Server`...

Because of this additional layer that somehow got introduced: `Application`.

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/deployment.rs
crepererum: So is your idea that we share the `Server` stuff? That can be arranged, but would require even more refactoring and at some point we wanna have query routing. So I would propose: let's take this duplication shortcut for now and then consolidate the interface afterwards.

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
tustvold: So `ApplicationState` was created as a container to reduce plumbing for "global-state". Some of this global state may not make sense in a "router" mode, but quite a lot of it probably does - everything bar job_registry and executor...

I can't help feeling trying to force a total separation is creating a lot of work for ourselves...

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/deployment.rs
tustvold: If this is the only service that is shared I guess I can deal :laughing:

tustvold:
> I don't think a router and a database should live in the same crate at all.

I totally agree with this, I want to split `Db` out of the server crate it categorically does not belong there. My suggestion was perhaps that we could share the same `Server` object and thereby re-use the same state machine, persistence logic, deployment service, etc...

tustvold (approved):
Had a good chat with @crepererum 

Some notes about this PR and things to follow in subsequent

* The duplication of DeploymentService is unfortunate but is relatively small in scope and therefore fine
* It would be good to revisit the background workers for ServerState
* It would be good if ApplicationState could somehow be shared with routers
* Routers at least initially won't have persisted configuration and may not ever (see #2436 )

The only change I would perhaps like to see in this PR is the `create_shared_database` as I think the name is too confusing

influxdb_iox/src/influxdb_ioxd/server_type/router/mod.rs
crepererum: done

influxdb_iox/tests/common/server_fixture.rs
crepererum: fixed

crepererum:
Follow-up work tracked under #3020.

PR #3018, opened 2021-11-03
feat: use protobuf encoding in write buffer (#2724) 

tustvold:
Draft as builds on #3017 

Adds the ability to decode protobuf payloads in the write buffer. There currently isn't a way to write such data into the write buffer, but one step at a time :laughing:


generated_types/protos/influxdata/iox/write_buffer/v1/write_buffer.proto
```diff
@@ -56,3 +58,10 @@ message WriteBufferCreationConfig {
   // Contains 0 or more key value pairs
   map<string, string> options = 2;
 }
+
+// A write payload for the write buffer
+message WriteBufferPayload {
+  oneof payload {
+    influxdata.pbdata.v1.DatabaseBatch write = 1;
```
crepererum, line 65: :+1: thanks for thinking about an easy extension point for deletes

crepererum (approved):

PR #3019, opened 2021-11-03
refactor: warn when we receive broken data from write buffer

crepererum:
We need to understand why this is the case and where broken data gets
stuck, because at the moment it seems that some Kafka hiccups can
poison the write buffer but it seems restarting IOx recovers the
situation.


mkmik:
CC: @primeroz

mkmik (approved):

primeroz:
:+1: 

> that some Kafka hiccups 

just to clarify that standard roll restarts are causing the issue, probably when the leader for a given partition change

mkmik:
@crepererum can we merge this?

PR #3021, opened 2021-11-03
feat: migrate write buffer producer to MutableBatch and pbdata (#2743)

tustvold:
Migrates the write buffer producer to be in terms of MutableBatch, and migrates the encoding to be using protobuf instead of flatbuffer


write_buffer/src/core.rs
```diff
@@ -25,17 +21,28 @@ pub trait WriteBufferWriting: Sync + Send + Debug + 'static {
...
         &self,
-        entry: &Entry,
         sequencer_id: u32,
-        span_context: Option<&SpanContext>,
-    ) -> Result<(Sequence, Time), WriteBufferError>;
+        write: &DbWrite,
+    ) -> Result<WriteMeta, WriteBufferError>;
+
+    /// Sends line protocol to the write buffer - primarily intended for testing
+    async fn store_lp(
```
tustvold, line 36: I opted to add a utility method to the trait to reduce the verbosity of tests

write_buffer/src/mock.rs
```diff
@@ -133,7 +135,19 @@ impl MockBufferSharedState {
             );
         }
 
-        entry_vec.push(Ok(entry));
+        writes_vec.push(Ok(write));
+    }
+
+    /// Push line protocol data with placeholder values used for write metadata
+    pub fn push_lp(&self, sequence: Sequence, lp: &str) {
```
tustvold, line 142: Much like WriteBufferWriting, this also has a utility line protocol method

write_buffer/Cargo.toml
```diff
@@ -14,6 +14,7 @@ http = "0.2"
 httparse = "1.5"
 mutable_batch = { path = "../mutable_batch" }
 mutable_batch_entry = { path = "../mutable_batch_entry" }
+mutable_batch_lp = { path = "../mutable_batch_lp" }
```
crepererum, line 17: It's a bit unfortunate that the write buffer now depends unconditionally on the LP parser. But since the LP-parser is used for basically every test anyways, I think it's somewhat OKish.

crepererum (commented):
On thing that I just realized: the consumer now produces a stream of `DbWrite`s which means for deletes we need to have a breaking change yet again. That's a bit of a waste since you've just refactored that particular type. Or was your assumption that the `DbWrite` can also carry deletes?

write_buffer/Cargo.toml
tustvold: I think it is pretty much a given that manipulates data ends up dependent on the line protocol parser, its a relatively lightweight dependency compared to some of the others, and FWIW is already a dependency of data_types...

tustvold:
> the consumer now produces a stream of DbWrites which means for deletes we need to have a breaking change yet again.

I think it should be a relatively straightforward change from DbWrite to something else, once I've worked out what that something else should be. I'm putting it in the bucket of, next weeks problems :laughing:

crepererum (approved):

alamb (commented):
I took a quick look through this PR and it looks good to me. 👍

PR #3027, opened 2021-11-04
refactor: remove Entry from Db interfaces (#2724)

tustvold:
Draft as builds on top of #3021 

This migrates the Db interfaces to no longer use Entry


server/src/db.rs
```diff
@@ -947,12 +938,10 @@ impl Db {
                 // buffer to return success before adding the entry to the mutable buffer.
 
                 // TODO: be smarter than always using sequencer 0
-                let meta = write_buffer
+                write_buffer
                     .store_write(0, &write)
                     .await
                     .context(WriteBufferWritingError)?;
-
-                write.set_meta(meta);
```
tustvold: The ability to sequence writes to a local database using a write buffer producer is something that I don't believe will be possible after @crepererum 's router mode rework, and I don't think ever really made a great deal of sense

server/src/db.rs
```diff
@@ -2382,19 +2364,8 @@ mod tests {
 
     #[tokio::test]
     async fn write_updates_persistence_windows() {
-        // Writes should update the persistence windows when there
```
tustvold, line 2389: See comment above about using a write buffer producer to sequence local writes

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
```diff
@@ -923,32 +923,10 @@ mod tests {
             .unwrap()
             .clone();
 
-        let entry_ingest = metric_registry
```
tustvold, line 926: I'm not sure how important these metrics are, we already have `ingest_lines`, `ingest_bytes`, etc... that measure the line protocol as it was ingested to the routers, and we also have `write_buffer_*` metrics that measure the data consumed from the write buffer so I'm not sure what these metrics are actually here for? I could report the in-memory size of the DbWrite, but I'm not entirely sure what significance this would have...

FYI @mkmik

crepererum (approved):

tustvold:
Fixed commit message

PR #3028, opened 2021-11-04
feat: add basic CRUD operations for router configs

crepererum:
It's not as big as it seems, it's mostly boilerplate to map the protobuf types to proper Rust types (similar to how it's done for database rules). Also: **no persistence (yet, if we ever want that)**


data_types/src/consistent_hasher.rs
```diff
@@ -53,6 +53,17 @@ where
     }
 }
 
+impl<T> Default for ConsistentHasher<T>
```
crepererum, line 56: That's because the derive macro results in overly strict bounds (i.e. you don't need `T: Default` for an empty ring).

data_types/src/router.rs
```diff
@@ -0,0 +1,135 @@
+use std::collections::BTreeMap;
```
crepererum, line 1: The shard configs here are partially copied from the old ones. The old ones will be removed once we remove router support from the `server` crate.

generated_types/protos/influxdata/iox/router/v1/router.proto
```diff
@@ -71,11 +68,14 @@ import "influxdata/iox/write_buffer/v1/write_buffer.proto";
...
   //
-  // The name is unique for this node.
+  // The name corresponds to the database name on the database node.
+  //
+  // The router name is unique for this router node.
   string name = 1;
 
-  // Sources of write requests.
-  WriteSources write_sources = 2;
+  // write source, current always HTTP and gRPC
```
crepererum, line 76: We only support routing from incoming HTTP and gRPC requests for now since consuming write buffers would require some kind of checkpointing and that requires design that is out of scope for now.

generated_types/src/lib.rs
```diff
@@ -194,19 +204,21 @@ pub use influxdata::platform::storage::*;
 
 pub mod google;
 
-#[cfg(feature = "data_types_conversions")]
+#[cfg(any(feature = "data_types_conversions", test))]
```
crepererum, line 207: w/o these changes the tests in these crate might just be skipped

influxdb_iox/tests/common/server_fixture.rs
```diff
@@ -518,52 +524,57 @@ impl TestServer {
...
-                    while !management_client
-                        .get_server_status()
-                        .await
-                        .unwrap()
-                        .initialized
-                    {
-                        interval.tick().await;
-                    }
-                };
+                if self.test_config.server_type == ServerType::Database {
```
crepererum, line 548: That was a bug in the generic test fixture that I had to address.

PR #3031, opened 2021-11-04
feat: simplify sharding

tustvold:
The current sharding configuration in addition to routing based on table name, can also route based on column values and custom row-level predicates.

This complicates using a columnar format internally (#2724) and (#2139), makes the implementation of deletes non-trivial, requires complex distributed query to route correctly, has eventual consistency implications for schema (#2938), and will break deduplication if using a non-tag column.

The new routing API added in https://github.com/influxdata/influxdb_iox/pull/3008/files#diff-6c1fc332cc7d90fd7b583cf760d5cfd38b637d817cab11501d2dddc395dfce29 no longer exposes this, but this PR removes the config from the existing protos in addition to the corresponding data_types structures


alamb (commented):
I think @pauldix  should probably weigh in on the possible implications of changing the possible hashing rules

generated_types/protos/influxdata/iox/management/v1/shard.proto
```diff
@@ -87,10 +88,12 @@ message DevNull {
 // HashRing is a rule for creating a hash key for a row and mapping that to
 // an individual node on a ring.
 message HashRing {
-  // If true the table name will be included in the hash key
-  bool table_name = 1;
-  // include the values of these columns in the hash key
-  repeated string columns = 2;
+  reserved 1;
```
alamb, line 91: Given that the protobufs are the user visible "API" the fact that these fields are no longer present seems like they could be brought back with fairly little fanfare.

In other words 👍

alamb (approved):

Issue #2680, opened 2021-09-30
Floating databases: Replace the tombstone file for database delete/restore with the release/claim API


carols10cents:
- ~~On server startup, instead of listing object storage contents and checking for tombstone file existence to know if a database is active or inactive, check the server's config.pb to see what databases it owns~~ this part is already done
  - Make delete/restore aliases for the release/claim database command/API
  - Remove all code having to do with tombstone files

carols10cents:
@mkmik moving this discussion here that was [started on the meta-issue](https://github.com/influxdata/influxdb_iox/issues/2672#issuecomment-931289790): 

> > Wait, so when the Disown API is used (instead of Delete), there _isn't_ a `DatabaseTombstone` context written?
> 
> When the Disown API is used, then the caller chooses what to put in the context. Usually the orchestrator tool (e.g. conductor) will have their own proto message that marks the operation as being done by conductor and gives some context (e.g. disowned as part of the migration from server 1 to server 2)

Ok, I think I understand, let me check:

If someone runs:

```
$ influxdb_iox database disown foo
```

and doesn't provide any context, then after the operation is complete, the `floating.pb` (or whatever we end up calling it) will contain:

```
FloatingDatabaseConfig {
    backlink = ''

    transactions = [
        OwnershipTransaction {
            backlink = 's3://<bucket..>/nodes/<previous-owner-server-id>'
            context = []
        }
    ]
}
```

and the characteristic of this `floating.pb` that identifies this database as "inactive" is the `backlink = ''`.

However, if someone runs:

```
$ influxdb_iox database delete foo
```

and doesn't provide any context, then after the operation is complete, the `floating.pb` will contain:

```
FloatingDatabaseConfig {
    backlink = ''

    transactions = [
        OwnershipTransaction {
            backlink = 's3://<bucket..>/nodes/<previous-owner-server-id>'
            context = [
                DeleteTombstone {}
            ]
        }
    ]
}
```

and we still only need to look at the `backlink` to see that this database is inactive.

At some point I was thinking we'd need to check the most recent `OwnershipTransaction` to see if it had a `DeleteTombstone` to know whether a database was owned/active or not, but I think that's captured in the current value of `backlink`, correct?

mkmik:
correct!

PR #3038, opened 2021-11-04
feat: Database disown API

### LINKED ISSUE #2678: Floating databases: Add a Disown API
- Add a gRPC API and CLI command that take:
    - database name
    - Optional UUID that must match the UUID of the database with the specified name for the disown operation to proceed
    - Optional context to be recorded in the transaction history of the database
  - When invoked, the server:
    - The server first checks and make absolutely sure it is the current owner of this database by verifying the database's `owner.pb` `id` points to the current server's ID
    - Writes into the database's `owner.pb` a new `OwnershipTransaction` that contains the current value of `id`, the provided context (if any), the context `DatabaseTombstone`, and the current time, and sets the toplevel `id` to empty string.
    - Deletes this database's entry in the `s3://<bucket..>/nodes/<server-id>/config.pb` file.
    - Removes the database from its in-memory list of databases.
    - Returns the UUID of the disowned database on success or an error message on failure.

carols10cents:
Closes #2678

Ok, I *think* I'm at least ready for feedback here... Notes: 

- I'm not at all sure that I've done the `google::proto::Any` type stuff or the `TimeProvider` stuff correctly.
- I have a TODO comment instead of an implementation for truncating the number of transaction history entries to 100, which was starting to feel just complicated enough that it should be tested, but unrelated enough from `server/src/database.rs` for me to want to write tests there, which leads to...
- I think the functions I currently have in `server/src/database.rs` that fetch, create, and update `OwnerInfo` protobuf need to be extracted to... `server/src/database/owner_info.rs` and made into methods on a pure-Rust (non-proto) `OwnerInfo` struct? Does that sound like the right place? Should I do this now or in a future PR?

--- 

Here's the help text for the new command:

```
$ influxdb_iox database disown --help
influxdb_iox-database-disown 0.1.0
Disown a database from its current server owner

USAGE:
    influxdb_iox database disown [OPTIONS] <name>

FLAGS:
        --help
            Prints help information

    -V, --version
            Prints version information


OPTIONS:
        --header <header>...
            Additional headers to add to CLI requests

            Values should be key value pairs separated by ':'
    -h, --host <host>
            gRPC address of IOx server to connect to [env: IOX_ADDR=]  [default: http://127.0.0.1:8082]

    -u, --uuid <uuid>
            Optionally, the UUID of the database to delete. This must match the UUID of the current database with the
            given name, or the disown operation will result in an error

ARGS:
    <name>
            The name of the database to disown
```


server/src/database.rs
```diff
@@ -294,24 +346,22 @@ impl Database {
...
-
-        iox_object_store
-            .put_owner_file(encoded)
-            .await
-            .context(SavingOwner)?;
+        let database_location = iox_object_store.root_path();
+        let server_location =
+            IoxObjectStore::server_config_path(application.object_store(), server_id).to_string();
+
+        let timestamp = Time::from_timestamp(295293, 3);
```
carols10cents, line 353: Errrrrr this is wrong, I thought this was in a test. Will fix!!!!

influxdb_iox/src/commands/database.rs
```diff
@@ -187,6 +190,23 @@ struct Delete {
...
+/// Disown a database from its current server owner
+#[derive(Debug, StructOpt)]
+struct Disown {
+    /// The name of the database to disown
+    name: String,
+
+    /// Optionally, the UUID of the database to delete. This must match the UUID of the current
+    /// database with the given name, or the disown operation will result in an error.
+    #[structopt(short, long)]
+    uuid: Option<String>,
```
carols10cents, line 202: And just noticed this can be 

```suggestion
    uuid: Option<Uuid>,
```

and structopt will take care of it... another fix incoming 😅

generated_types/protos/influxdata/iox/management/v1/server_config.proto
```diff
@@ -14,10 +16,32 @@ message ServerConfig {
...
+
+  // Copy of the path to this server's config file in object storage at the time of the
+  // transaction.
+  string location = 2;
+
+  // When the transaction took place
+  google.protobuf.Timestamp timestamp = 3;
+
+  // Context provided in the disown/adopt calls
+  repeated google.protobuf.Any context = 4;
```
alamb, line 46: I am interested to know the rationale for storing the extra metadata as protobuf is (and why not as a opaque string, for example)

The upsides of using protobuf is that perhaps IOx can look at this information and interpret it. The downside of using protobuf is that it is harder to use for some arbitrary payload I think (as you need to encode it using gRPC). 

If the idea is that the context is opaque to IOx (and not parsed/ interpreted by it) I think it would be a better user experience to simply store this context as a string (let users base64 encde stuff if they need binary).

In fact it seems that is exactly what the cli test cases do (store some string). Maybe the code would be simpler if this field's type is just `string`?

generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -48,8 +49,13 @@ service ManagementService {
   // Roughly follows the <https://google.aip.dev/134> pattern, except we wrap the response
   rpc UpdateDatabase(UpdateDatabaseRequest) returns (UpdateDatabaseResponse);
 
+  // Delete a database.
```
alamb, line 52: what is the difference between `delete` a database and `disown` a database? They both seem to (logically) do the same thing.

influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -447,6 +447,90 @@ async fn test_create_get_update_delete_restore_database() {
...
+    let db_name = rand_name();
+    let rules = DatabaseRules {
+        name: db_name.clone(),
+        ..Default::default()
+    };
+
+    // Create a database on one server
+    let created_uuid = client.create_database(rules.clone()).await.unwrap();
+
+    // Disown database returns the UUID
```
alamb, line 465:
```suggestion
    // Delete database returns the UUID
```
The command below this calls `delete_database`, not `disown_database` -- is that expected?

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -388,6 +388,250 @@ async fn delete_restore_database() {
...
+            .success()
+            .stdout(predicate::str::contains(format!(
+                "Disowned database {}",
+                db
+            )))
+            .get_output()
+            .stdout
+            .clone(),
+    )
+    .unwrap();
```
alamb, line 630: Is there any way (or plan) to be able to view the `context`s that were added to `disown` database calls?

server/src/lib.rs
```diff
@@ -745,6 +763,60 @@ where
...
+        if matches!(uuid, Some(specified) if specified != current) {
+            return UuidMismatch {
+                db_name: db_name.to_string(),
+                specified: uuid.unwrap(),
+                current,
+            }
+            .fail();
+        }
+
+        let timestamp = self.shared.application.time_provider().now();
```
alamb, line 794: 👍

server/src/lib.rs
```diff
@@ -745,6 +763,60 @@ where
...
+
+        {
+            let mut state = self.shared.state.write();
+
+            // Exchange FreezeHandle for mutable access via WriteGuard
+            let mut state = state.unfreeze(handle);
+
+            match &mut *state {
+                ServerState::Initialized(initialized) => {
+                    initialized.databases.remove(db_name);
```
alamb, line 809: I think we may need / want to call `join` here like

```
initialized.databases.remove(db_name).join().await;
```

To make sure the database has completed outstanding activity (and before this server reports the db's uuid). Otherwise something like conductor may try and have some other server adopt the database while there is still (an in memory `Database`) writing stuff to object store

Perhaps @tustvold  can check my understanding on this point

server/src/database.rs
```diff
@@ -294,24 +347,21 @@ impl Database {
...
-        let owner_info = management::v1::OwnerInfo {
-            id: server_id.get_u32(),
-            location: IoxObjectStore::server_config_path(application.object_store(), server_id)
-                .to_string(),
-        };
-        let mut encoded = bytes::BytesMut::new();
-        generated_types::server_config::encode_database_owner_info(&owner_info, &mut encoded)
-            .expect("owner info serialization should be valid");
-        let encoded = encoded.freeze();
+        let database_location = iox_object_store.root_path();
```
alamb, line 350: i think this code is looking much better now that the owner info stuff is refactored into `update_owner_info` 👍

server/src/database.rs
```diff
@@ -1178,6 +1221,124 @@ impl DatabaseStateDatabaseObjectStoreFound {
...
+
+    let new_transaction = management::v1::OwnershipTransaction {
+        id,
+        location,
+        timestamp: Some(timestamp.date_time().into()),
+        context,
+    };
+    transactions.push(new_transaction);
+
+    // TODO: only save latest 100 transactions
```
alamb, line 1320: I think this would be a good follow on PR

alamb (approved):
> I'm not at all sure that I've done the google::proto::Any type stuff or the TimeProvider stuff correctly.

The timeprovider stuff looks good to me. The `Any` stuff I am not sure about either (not being an expert in that area) but I have a suggestion (to use `string` that perhaps would avoid the use of `Any` at all)

> I think the functions I currently have in server/src/database.rs that fetch, create, and update OwnerInfo protobuf need to be extracted to... server/src/database/owner_info.rs and made into methods on a pure-Rust (non-proto) OwnerInfo struct? Does that sound like the right place? Should I do this now or in a future PR?

I think `server/src/database/owner_info.rs` is reasonable --  Another place for the pure Rust type would be in the `data_types` crates which we have used for similar things in the past

A future PR is probably the best

I think the code looks good to me.  I am curious about:
1. How disown is related to `delete` / `restore` (and what the plans are to unify this code. if any)

Is there any way to `adopt` a database (so that I could do some end to end manual testing)? I tried using `restore` but causes an error:

```shell
alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database create db1
Created database db1
99502bea-a058-4f0c-846f-0ec5032b548e

alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database disown db1
Disowned database db1
99502bea-a058-4f0c-846f-0ec5032b548e

alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database restore 99502bea-a058-4f0c-846f-0ec5032b548e
Error restoring database: Unexpected server error: Some entity that we attempted to create already exists: The database with UUID `99502bea-a058-4f0c-846f-0ec5032b548e` named `db1` is already active
```

Which seems strange

carols10cents:
>     1. How disown is related to `delete` / `restore` (and what the plans are to unify this code. if any)

That's tracked in https://github.com/influxdata/influxdb_iox/issues/2680. The TL;DR from the design doc are that delete/restore are going to become essentially aliases for disown/adopt. Delete will have a special context message it will put in the transaction history.

> Is there any way to adopt a database (so that I could do some end to end manual testing)? I tried using restore but causes an error:

Nope, not until https://github.com/influxdata/influxdb_iox/pull/3090 that builds on this one, sorry! And yes, until #2680 is done, delete and restore will be incompatible with disown/adopt (that is, temporarily, you can either delete+restore or disown+adopt, but when the floating databases project is done, you'll be able to delete+adopt or disown+restore to your heart's content).

> Which seems strange

Yes, you're getting that error because, at the moment, disown doesn't write the tombstone file, and restore currently refuses to restore unless there's a tombstone file to delete. Tombstone files will be going away as part of #2860.

generated_types/protos/influxdata/iox/management/v1/server_config.proto
mkmik: Imagine you don't want to use an `Any` and store an arbitrary payload. How would you do it? 

A `string` field can't contain any arbitrary encoding. You'd have to base64 the content if you have any non textual content. But that's fine, let's use a `bytes` then.

Ok, now you have a field that can contain whatever arbitrary payload you the client wants to put in there.

A few interesting questions arise:

1. But what is the type of this payload? The client who wrote it clearly knows what it wrote, but what about the new version of the client? Surely it can be up to the client to use some self describing file format that will tell which format/version it is.
2. What if you have multiple "clients" (or multiple layers of clients) that each wants to put some additional metadata into this; how would you tell them apart? A classic solution is to have some sort of key/name/... field associated with each blob (e.g. a string->blob (multi)map) so each client can quickly find it's stuff.

You could even combine (1) and (2) and add a "type" field that plays both roles: it tells clients (and humans!) what is the type of the payload and also lets clients (and humans!) associate the payloads with the clients and/or roles that such payloads play.

So basically we could just have a message like:

```proto
message TypedBlob {
  string type = 1;
  bytes blob = 2;
}
```

and then we can just have:

```proto
repeated TypedBlob context = 4;
```

Turns out this message type already exists and it's called [Any](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto)

You can put whatever you want into it; they payload is a byte array.

It just happens that languages with mature proto libraries also come with helpers to unpack a nested protobuf message encoded in the payload.

(As with everything in computer science there are always alternative ways of doing things; using Any in systems that use protobuf to define schemas flows the path of least surprise and most familiarity)

carols10cents: I'm not sure, that was what's in the design doc from @mkmik and @crepererum-- could one of you elaborate on the reason for having the transaction history `context`s be `Any` rather than `string`?

influxdb_iox/tests/end_to_end_cases/management_cli.rs
carols10cents: There wasn't anything about exposing that as a gRPC API in the design doc, so for now the only way is to look at the database's owner file in object storage.

generated_types/protos/influxdata/iox/management/v1/server_config.proto
mkmik: (I think I answered that in my previous comment that probably landed while you were still typing yours)

tustvold (approved):
Approving as the bulk of the PR looks good to me, I have a couple of comments that I think should be addressed before merging. In particular the shutdown behaviour and the Any serialization

influxdb_iox_client/src/client/management.rs
```diff
@@ -697,6 +725,45 @@ impl Client {
         Ok(uuid)
     }
 
+    /// Disown database
+    pub async fn disown_database(
+        &mut self,
+        db_name: impl Into<String> + Send,
+        uuid: Option<Uuid>,
+        context: Option<String>,
```
mkmik, line 733: the point of context is to allow arbitrary data to be passed as typed protobuf message. Here we're forcing that to be a string if the call comes through this internal API, which is suboptimal.

influxdb_iox_client/src/client/management.rs
```diff
@@ -697,6 +725,45 @@ impl Client {
...
+    ) -> Result<Uuid, DisownDatabaseError> {
+        let db_name = db_name.into();
+        let response = self
+            .inner
+            .disown_database(DisownDatabaseRequest {
+                db_name: db_name.clone(),
+                uuid: uuid.map(|u| u.as_bytes().to_vec()).unwrap_or_default(),
+                context: vec![google::protobuf::Any {
+                    type_url: protobuf_type_url("google.protobuf.StringValue"),
+                    value: context.unwrap_or_default().into(),
```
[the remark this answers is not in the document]
mkmik: technically `google.protobuf.StringValue` is not a string; it's a wrapper message that has one field, of type string.

If you want to produce a valid `Any` message whose payload is a `StringValue`, you have to marshal a `StringValue` protobuf message in a byte array and use that as the value.

Unfortunately the rust protobuf library doesn't yet come with useful helpers to streamline this operation.

That said, see my previous comment about whether or not we should be doing this string encoding in the first place.

FWIW, we can postpone the addition of this context parameter to another time and keep the initial API leaner.

generated_types/protos/influxdata/iox/management/v1/service.proto
mkmik: The floating db design doc subsumed deletion as a special kind of the disown operation.

We suggested that the IOx server may still offer a "delete" API operation which would be effectively a "sugar façade" in front of the disown operation, which would include a context message that recorded that the disown operation happened as a reaction to an express request by the API caller to delete the database. This context information may be useful for recovery tools to discriminate from orphaned databases left over from incorrectly aborted db migrations.

server/src/database.rs
```diff
@@ -269,13 +273,62 @@ impl Database {
...
+    }
+
     /// Create a restored database without any state. Returns its location in object storage
     /// for saving in the server config file.
     pub async fn restore(
         application: Arc<ApplicationState>,
         db_name: &DatabaseName<'static>,
         uuid: Uuid,
         server_id: ServerId,
+        timestamp: Time,
```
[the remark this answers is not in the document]
carols10cents: Oh, I was thinking it'd be easier to test if we passed time in, but I was also pretty sure I was missing something about the design 😂 Will fix!!

influxdb_iox_client/src/client/management.rs
alamb: Is the rationale for trying to attached typed protobuf payloads written down somewhere? I can't come up with any reason the payload should be non-opaque unless we have some usecase for iox itself to "peek" into its contents. 

In other words, I wonder "why not store all metadata as a string and then let the user of this API decide how they want to encode their messages"?

mkmik: I think it has been mentioned in the design doc but probably only very superficially and/or taking for granted some prorobuf API design patterns.

The key insight is that this is not about IOx itself having visibility into the content if this field; quite the opposite: it's about IOx just recording whatever its client passed during disown time. However while doing so it imposes a minimal structure that makes it easier for clients to deal with (possibly multiple layers of) custom data.

The only reason this is getting in the way here is that we're not embracing working with the grpc API directly and we're wrapping things back and forth between machine generated prorobuf rust wrappers and "native" rust API layers. This should just be a simple stupid field that gets copied from the client request without further fiddling with (as we'd do if it were a string)

carols10cents: I've copied this conversation over to the issue I made to track adding context back in to preserve it: https://github.com/influxdata/influxdb_iox/issues/3095#issuecomment-967339144

server/src/lib.rs
```diff
@@ -740,6 +740,11 @@ where
 
         let database = self.database(db_name)?;
         let uuid = database.delete().await.context(CannotMarkDatabaseDeleted)?;
+        database.shutdown();
+        let _ = database
+            .join()
+            .await
+            .log_if_error("database background worker");
```
alamb, line 747:
```suggestion
            .log_if_error("database background worker while deleting database");
```

server/src/lib.rs
```diff
@@ -788,6 +793,11 @@ where
         }
 
         let returned_uuid = database.disown().await.context(CannotDisownDatabase)?;
+        database.shutdown();
+        let _ = database
+            .join()
+            .await
+            .log_if_error("database background worker");
```
alamb, line 800:
```suggestion
            .log_if_error("database background worker while disowning database");
```

alamb (approved):
I took a look at https://github.com/influxdata/influxdb_iox/pull/3038/commits/815347b18a61d3998deaad3e798f139f39539b47 https://github.com/influxdata/influxdb_iox/pull/3038/commits/8bdfb253ce93b0cef5903a5b9db7f0023f096313 https://github.com/influxdata/influxdb_iox/pull/3038/commits/315cc3836b8d530777d71d1d3d9a819928e4d8be   https://github.com/influxdata/influxdb_iox/pull/3038/commits/56c16414a17750b0659107e8cf29e2f59e0a4f3e and https://github.com/influxdata/influxdb_iox/pull/3038/commits/63ee5b8317339b8eff00433cf1eb0ccac5b2829a

and I think they look great 👍

generated_types/protos/influxdata/iox/management/v1/server_config.proto
alamb: The downside in my mind of using `Any` is that it *requires* the clients to use protobuf to encode their metatadata.

For some client languages this is probably fine because protobuf support is excellent (`golang`) or good enough (`rust`). I don't know how well protobuf is supported in other languages. Though now that I type this I suppose that since the API itself is in terms of protobuf at least part of the client will have to deal with that anyways

🤷  I am not strongly opposed

PR #3039, opened 2021-11-05
feat: implement sharder for new router

crepererum:


tustvold (approved):
I'm not a fan that this is cloning the inner batches, but we can fix this up when we consolidate the routing logic

See - https://github.com/influxdata/influxdb_iox/pull/3035/files#diff-6d4e9e9a12c0366f056d3145e67db44f6695735adf895951630cccc50d29856bR302

Issue #2856, opened 2021-10-15
Rework write buffer errors


crepererum:
Currently write buffer errors are completely opaque and messages are not standardized. We should rework that to offer at least a few sane variants.

PR #3040, opened 2021-11-05
feat: add connection management for router

crepererum:
Implements connection pooling and mocking features. This is slightly similar to what the old `ConnectionManager` does but relies (like the write buffer) on dynamic dispatch so we don't have to use generics all over the place.

The `data_types` changes are required so we can hash write buffer configs.

### DIFF
--- a/router/src/connection_pool.rs
+++ b/router/src/connection_pool.rs
@@ -46,7 +46,10 @@ impl ConnectionPool {
     /// Create new connection pool.
     ///
     /// If `use_mock_grpc` is set only mock gRPC clients are created.
-    pub fn new(use_mock_grpc: bool, wb_factory: WriteBufferConfigFactory) -> Self {
+    pub async fn new(use_mock_grpc: bool, wb_factory: WriteBufferConfigFactory) -> Self {
+        // Note: this function is async even though it does not contain any `.await` calls because `LoadingCache::new`
+        // requires tokio to be running and even if documented people will forget about this.
+
         let grpc_clients = if use_mock_grpc {
             LoadingCache::new(|_connection_string: String| async move {
                 use crate::grpc_client::MockClient;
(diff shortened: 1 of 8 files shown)


router/src/connection_pool.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub async fn grpc_client(
+        &self,
+        connection_string: &str,
+    ) -> Result<Arc<dyn GrpcClient>, ConnectionError> {
+        let res = self
+            .grpc_clients
+            .get_with_meta(connection_string.to_string())
+            .await
+            .map_err(|e| Arc::new(e) as ConnectionError)?;
+        debug!(was_cached=%res.cached, %connection_string, "getting IOx write client");
```
alamb, line 108:
```suggestion
        debug!(was_cached=%res.cached, %connection_string, "getting IOx client");
```

I think the client will be used for more than just `write`, won't it?

router/src/grpc_client.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use mutable_batch_lp::lines_to_batches;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_mock() {
```
alamb, line 143: Your commitment to testing is admirable. 🏅 🏅 🏅  for these tests and structring the code for testability.

alamb (approved):
This code is very nice. Well done @crepererum 

> Implements connection pooling and mocking features. This is slightly similar to what the old ConnectionManager does but relies (like the write buffer) on dynamic dispatch so we don't have to use generics all over the place.

Is the intent that we will remove the old `ConnectionManager` at the completion of this work?

crepererum:
> Is the intent that we will remove the old `ConnectionManager` at the completion of this work?

yes.

router/src/connection_pool.rs
```diff
@@ -0,0 +1,152 @@
...
+    config::WriteBufferConfigFactory,
+    core::{WriteBufferError, WriteBufferWriting},
+};
+
+use crate::grpc_client::GrpcClient;
+
+type KeyWriteBufferProducer = (String, WriteBufferConnection);
+pub type ConnectionError = Arc<dyn std::error::Error + Send + Sync + 'static>;
+
+/// Stupid hack to fit the `Box<dyn ...>` in `WriteBufferError` into an `Arc`
```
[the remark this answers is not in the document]
crepererum: I have an even better proposal: #2856.

PR #3042, opened 2021-11-05
feat: router core logic

crepererum:
~Draft, on top of #3040.~

This implements the full core logic required for **write** routing. With #3035 I can then wire up the gRPC and HTTP APIs rather easily.

### DIFF
--- a/router/src/router.rs
+++ b/router/src/router.rs
@@ -167,14 +167,17 @@ mod tests {
     async fn test_write() {
         let server_id_1 = ServerId::try_from(1).unwrap();
         let server_id_2 = ServerId::try_from(2).unwrap();
+        let server_id_3 = ServerId::try_from(3).unwrap();
 
         let resolver = Arc::new(Resolver::new(Some(RemoteTemplate::new("{id}"))));
         let connection_pool = Arc::new(ConnectionPool::new_testing().await);
 
         let client_1 = connection_pool.grpc_client("1").await.unwrap();
         let client_2 = connection_pool.grpc_client("2").await.unwrap();
+        let client_3 = connection_pool.grpc_client("3").await.unwrap();
         let client_1 = client_1.as_any().downcast_ref::<MockClient>().unwrap();
         let client_2 = client_2.as_any().downcast_ref::<MockClient>().unwrap();
+        let client_3 = client_3.as_any().downcast_ref::<MockClient>().unwrap();
 
         let cfg = RouterConfig {
             name: String::from("my_router"),
@@ -188,6 +191,12 @@ mod tests {
                     },
                     MatcherToShard {
                         matcher: Matcher {
+                            table_name_regex: Some(Regex::new("foo_three").unwrap()),
+                        },
+                        shard: ShardId::new(30),
+                    },
+                    MatcherToShard {
+                        matcher: Matcher {
                             table_name_regex: Some(Regex::new("foo_.*").unwrap()),
                         },
                         shard: ShardId::new(20),
@@ -196,13 +205,13 @@ mod tests {
                         matcher: Matcher {
                             table_name_regex: Some(Regex::new("doom").unwrap()),
                         },
-                        shard: ShardId::new(30),
+                        shard: ShardId::new(40),
                     },
                     MatcherToShard {
                         matcher: Matcher {
                             table_name_regex: Some(Regex::new("nooo").unwrap()),
                         },
-                        shard: ShardId::new(40),
+                        shard: ShardId::new(50),
                     },
                     MatcherToShard {
                         matcher: Matcher {
@@ -228,6 +237,15 @@ mod tests {
                     WriteSinkSetConfig {
                         sinks: vec![WriteSinkConfig {
                             sink: WriteSinkVariantConfig::GrpcRemote(server_id_2),
+                            ignore_errors: false,
+                        }],
+                    },
+                ),
+                (
+                    ShardId::new(30),
+                    WriteSinkSetConfig {
+                        sinks: vec![WriteSinkConfig {
+                            sink: WriteSinkVariantConfig::GrpcRemote(server_id_3),
                             ignore_errors: false,
                         }],
                     },
@@ -275,11 +293,18 @@ mod tests {
                 "foo_bar x=9 9",
                 "nooo x=10 10",
                 "foo_bar x=11 11",
+                "foo_three x=12 12",
+                "doom x=13 13",
+                "foo_three x=14 14",
+                "www x=15 15",
+                "foo_three x=16 16",
+                "nooo x=17 17",
+                "foo_three x=18 18",
             ],
             &meta_2,
         );
         let err = router.write(write_2).await.unwrap_err();
-        assert_eq!(err.to_string(), "One or more writes failed: ShardId(20) => \"Write to sink set failed: Cannot write: poisened\", ShardId(30) => \"Did not find sink set for shard ID 30\"...");
+        assert_eq!(err.to_string(), "One or more writes failed: ShardId(20) => \"Write to sink set failed: Cannot write: poisened\", ShardId(40) => \"Did not find sink set for shard ID 40\"...");
         client_1.assert_writes(&[
             (
                 String::from("my_router"),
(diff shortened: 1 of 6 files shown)


router/src/router.rs
```diff
@@ -1,16 +1,91 @@
...
+
+#[derive(Debug, Snafu)]
+pub enum WriteError {
+    #[snafu(display("One or more writes failed: {}", fmt_write_errors(errors)))]
+    MultiWriteFailure {
+        errors: BTreeMap<ShardId, WriteErrorShard>,
+    },
+}
+
+fn fmt_write_errors(errors: &BTreeMap<ShardId, WriteErrorShard>) -> String {
```
alamb, line 33: 👍  limiting the number of errors printed is a nice idea

router/src/router.rs
```diff
@@ -1,16 +1,91 @@
...
+}
 
 /// Router for a single database.
 #[derive(Debug)]
 pub struct Router {
     /// Router config.
     config: RouterConfig,
+
+    /// Write sink sets, one per shard ID.
+    write_sink_sets: HashMap<ShardId, WriteSinkSet>,
```
alamb, line 59: Given that elsewhere we use `BTreeMap` to ensure a consistent iteration of `ShardId` what do you think about doing so here? If we use a `HashMap` I wonder if we would get non deterministic behavior (like non deterministic error messages or something)

I don't have any specific scenarios in mind, I am just thinking in general the more deterministic we can make things the easier it will be to understand its behavior

router/src/router.rs
```diff
@@ -24,22 +99,213 @@ impl Router {
...
+                    },
+                    MatcherToShard {
+                        matcher: Matcher {
+                            table_name_regex: Some(Regex::new("foo_.*").unwrap()),
+                        },
+                        shard: ShardId::new(20),
+                    },
+                    MatcherToShard {
+                        matcher: Matcher {
+                            table_name_regex: Some(Regex::new("doom").unwrap()),
```
alamb, line 195: Screen Shot 2021-11-08 at 6 34 24 AM

router/src/router.rs
```diff
@@ -24,22 +99,213 @@ impl Router {
...
+                "www x=8 8",
+                "foo_bar x=9 9",
+                "nooo x=10 10",
+                "foo_bar x=11 11",
+            ],
+            &meta_2,
+        );
+        let err = router.write(write_2).await.unwrap_err();
+        assert_eq!(err.to_string(), "One or more writes failed: ShardId(20) => \"Write to sink set failed: Cannot write: poisened\", ShardId(30) => \"Did not find sink set for shard ID 30\"...");
+        client_1.assert_writes(&[
```
alamb, line 281: 👍  this case ensures that client1 actually got the write. 👍 

I wonder if it is worth a test where client1 fails (but client 2 succeeds) to ensure that the write path doesn't stop at the first failure. This test may not be worth it at this time, but I wanted to mention it

router/src/server.rs
```diff
@@ -41,6 +46,9 @@ impl RouterServer {
             trace_collector,
             routers: Default::default(),
             resolver: Arc::new(Resolver::new(remote_template)),
+            connection_pool: Arc::new(
```
alamb, line 49: Will this  pool connections server-wide? (as in should the pool be created as part of `RouterServer::new()` or whatever calls `RouterServer::new()`)? As of this PR, there is only a single `RouterServer` created (in `influxdb_iox/src/commands/run/router.rs`) so I think all is good  but wanted to double check

router/src/server.rs
```diff
@@ -100,17 +112,26 @@ impl RouterServer {
         self.routers.write().remove(name).is_some()
     }
 
+    /// Get registered router, if any.
```
alamb, line 115: Could you please document what (logically) the `name` parameter represents? Is it the name of another server ? Its hostname? Its name in the config?

router/src/write_sink.rs
```diff
@@ -0,0 +1,361 @@
...
+            .await
+            .context(WriteFailure)?;
+
+        Ok(())
+    }
+}
+
+#[derive(Debug)]
+enum WriteSinkVariant {
+    GrpcRemote(VariantGrpcRemote),
```
alamb, line 114: I really like the naming convention of `Variant*` -- I find it much clearer than `GrpcRemote(GrpcRemote)` which seems to be a common Rust idiom 👍

alamb:
```suggestion
    /// Send write to a remote server via gRPC
    GrpcRemote(VariantGrpcRemote),
```

router/src/write_sink.rs
```diff
@@ -0,0 +1,361 @@
...
+    WriteBuffer(VariantWriteBuffer),
```
alamb, line 115:
```suggestion
    /// Send write to a "local" write buffer (which may be backed by kafka, local disk, etc)
    WriteBuffer(VariantWriteBuffer),
```

router/src/write_sink.rs
```diff
@@ -0,0 +1,361 @@
...
+                        sink_config,
+                        Arc::clone(&resolver),
+                        Arc::clone(&connection_pool),
+                    )
+                })
+                .collect(),
+        }
+    }
+
+    /// Write to sinks. Fails on first error.
```
alamb, line 191: To decrease the non determinism, what do you think about sorting the sinks on `db_name` or something? If writes are failed on the first error, then which system gets what write will be dependent on the order they are constructed.

However,  at the moment the order of construction seems to be the order they are defined in the config, which is deterministic and is easy to explain / reason about behavior.

So this is probably good the way it is written in this PR.

router/src/write_sink.rs
```diff
@@ -0,0 +1,361 @@
...
+            Default::default(),
+        );
+        sink_set.write(&write_1).await.unwrap();
+
+        let writes_1 = [(String::from("my_db"), write_1.clone())];
+        client_1.assert_writes(&writes_1);
+        client_2.assert_writes(&writes_1);
+        client_3.assert_writes(&writes_1);
+
+        client_2.poison();
```
alamb, line 345: 👍  for picking client 2 (so client 1 will succeed and client 3 will not get the write)

router/src/write_sink.rs
```diff
@@ -0,0 +1,361 @@
...
+            lines_to_batches("foo x=2 2", 0).unwrap(),
+            Default::default(),
+        );
+        sink_set.write(&write_2).await.unwrap_err();
+
+        let writes_2 = [
+            (String::from("my_db"), write_1.clone()),
+            (String::from("my_db"), write_2.clone()),
+        ];
+        client_1.assert_writes(&writes_2);
```
alamb, line 357: it might be worth a comment here explaining why  `client_2` and `client_3` are only expected to have `writes_1` (namely because the write to client2 fails and the write sink stops on first failure)

alamb (approved):
I really like how the error handling / writing logic is written. It is very easy to follow, where I found what is currently on main rather tricky (perhaps because it was embedded in Server / Db).

Very nice 👌

router/src/router.rs
crepererum: Not required, because the sharding steps uses deterministic iteration order and we want the fast `O(1)` lookup so sharding + sending isn't `O(n*log(n))`. I've added a few code comments to explain this.

router/src/router.rs
crepererum: done

router/src/server.rs
crepererum: Yes, there's a single pool for the whole server so that the number of connection IS NOT directly connected to the number of routers/databases handled by the server.

router/src/server.rs
crepererum: done

router/src/write_sink.rs
crepererum: done

router/src/write_sink.rs
crepererum: It's fully deterministic:

1. an incoming write (`DbWrite`) is only for a single DB
2. the sharding steps results in a `BTreeMap<ShardId, ...>`, so we write in the order of shard IDs
3. a sink set is per shard ID. The sinks within this set are a vector that has the same order as the in the config

router/src/write_sink.rs
crepererum: done

PR #3043, opened 2021-11-05
feat: remove legacy write service from influxdb_iox_client

tustvold:
This removes the usage of the legacy gRPC write API from influxdb_iox_client. The ability to write entries is removed, and line protocol is converted to protobuf on the client side

In line with how arrow-dependent functionality is hidden behind a feature flag, this moves the line protocol write API behind a feature flag - as it requires `mutable_batch` and by extension `arrow`


iox_data_generator/src/bin/create_database.rs
```diff
@@ -170,3 +164,86 @@ Examples:
 
     println!("Created database {}", db_name);
 }
+
+/// 3 rows of test data
```
tustvold, line 168: The data_generator proper only talks to the HTTP API, this create_database binary, however, writes some test data in using the gRPC write API. I'm not entirely sure why, but rather than introduce a dependency on mutable_batch, I just converted the lp to pbdata

influxdb_iox_client/src/client/write.rs
```diff
@@ -39,74 +45,60 @@ pub enum WriteError {
...
     /// [LineProtocol]: https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/#data-types-and-format
-    pub async fn write(
+    #[cfg(feature = "write_lp")]
+    pub async fn write_lp(
         &mut self,
-        db_name: impl Into<String> + Send,
-        lp_data: impl Into<String> + Send,
+        db_name: impl AsRef<str> + Send,
+        lp_data: impl AsRef<str> + Send,
+        default_time: i64,
```
alamb, line 77: As mentioned above, for the public facing API, I think it will be very confusing to require clients to provide a `default_time` here and think the parameter should be removed. 


If it is important for clients to provide a default time, a second `write_lp_with_default_time()` would probably be better in my opinion.

iox_data_generator/src/bin/create_database.rs
alamb: this is pretty yuck code, 🤮  the lp is so much nicer

alamb (commented):
I think the code does what the description says, but I have a question about the intent and think @pauldix  should weigh in. 

Specifically, should we be specifying default times on the client side?

Among other reasons to *NOT* fill out the times on the client:
1.  the existing influxdb [write API](https://docs.influxdata.com/influxdb/cloud/api/#tag/Write) does not  allow specifying the default time
3. Setting the timestamps on the client side may cause confusion / errors  if the client clocks have large amounts of skew
3. Users that want to specify the timestamps of their data can do so by setting it explicitly in the line protocol. Presumablyif they don't put timestamps in the line protocol, they want the server to take care of it

At the very least, I think the client should provide a default time of `now()` on the client. Perhaps better would be for the client to put `0` as the default time and then translate all such times on the server side to `now()`.


FWIW, I tested it out locally and worked great via `cli`:

```shell
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database write foo test_fixtures/lineproto/metrics.lp
1000 Lines OK
```


```shell
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database query foo "show tables"
+---------------+--------------------+---------------------+------------+
| table_catalog | table_schema       | table_name          | table_type |
+---------------+--------------------+---------------------+------------+
| public        | iox                | swap                | BASE TABLE |
| public        | iox                | diskio              | BASE TABLE |
| public        | iox                | net                 | BASE TABLE |
| public        | iox                | mem                 | BASE TABLE |
| public        | iox                | processes           | BASE TABLE |
| public        | iox                | disk                | BASE TABLE |
| public        | iox                | cpu                 | BASE TABLE |
| public        | iox                | system              | BASE TABLE |
| public        | system             | chunks              | BASE TABLE |
| public        | system             | columns             | BASE TABLE |
| public        | information_schema | tables              | VIEW       |
| public        | information_schema | columns             | VIEW       |
+---------------+--------------------+---------------------+------------+
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ 
```

iox_data_generator/src/bin/create_database.rs
tustvold: Not going to disagree with that one, I'm not sure how to avoid this... I welcome suggestions...

alamb: I wonder instead of 

```rust
                    Column {
                        column_name: "time".to_string(),
                        semantic_type: column::SemanticType::Time as _,
                        values: Some(column::Values {
                            i64_values: vec![100, 150, 200],
                            f64_values: vec![],
                            u64_values: vec![],
                            string_values: vec![],
                            bool_values: vec![],
                            bytes_values: vec![],
                            packed_string_values: None,
                            interned_string_values: None,
                        }),
                        null_mask: vec![],
                    },
```

It could use

```rust
                    Column {
                        column_name: "time".to_string(),
                        semantic_type: column::SemanticType::Time as _,
                        values: Some(column::Values {
                            i64_values: vec![100, 150, 200],
                            ..Default::default(),
                        }),
                        null_mask: vec![],
                    },
```
?

alamb (approved):
@pauldix  says requiring client to specify timestamps is desired.

iox_data_generator/src/bin/create_database.rs
tustvold: I didn't think this worked because Column doesn't implement Default, but it does so :tada:

influxdb_iox_client/src/client/write.rs
tustvold: Unless you object strongly I'm going to leave it as is for a few reasons:

* It makes explicit that this defaulting is taking place client side not server side
* It makes the call idempotent
* I still have an innate fear of `Utc::now` after the time replumbing nightmare

I think if we want defaulting of timestamps, we should loosen the constraints on pbdata, but otherwise I think we should be consistent in requiring them

alamb: > I still have an innate fear of Utc::now after the time replumbing nightmare

alamb: (I do not strongly object)

PR #3044, opened 2021-11-05
feat: remove influxdata.iox.write.v1.WriteService

tustvold:
~~Draft as builds on #3043~~

This removes the proto definition and accompanying server-side implementation


alamb (approved):
🧹

PR #3045, opened 2021-11-05
feat: remove flatbuffer entry

### LINKED ISSUE #2805: Remove "missing format header" from Kafka write buffer
Since #2806 we add a format header to our Kafka messages. Once this was deployed for a while we should remove the fallback that accepts messages w/o that header.

tustvold:
~Draft as builds on #3044~

Removes the entry flatbuffer from the codebase

Relates to #2724 

Closes #2805


write_buffer/src/kafka.rs
```diff
@@ -720,10 +719,17 @@ mod tests {
         let adapter = KafkaTestAdapter::new(conn);
         let ctx = adapter.new_context(NonZeroU32::new(1).unwrap()).await;
 
+        let headers = IoxHeaders::new(ContentType::Protobuf, None);
```
tustvold, line 722: This is necessary because the write buffer now errors if the header isn't present

alamb (commented):
❤️

alamb (approved):

alamb (approved):

alamb:
This may not be mergable as the `check-flattbuffers` check is removed but the repo has it marked as required. Perhaps @e-dard  could remove that requirement

alamb:
💥  what a great way to go into the weekend

PR #3048, opened 2021-11-05
fix: Store the restore database UUID as bytes instead of as a string

shepmaster:


carols10cents (approved):

carols10cents:
Merging because we discussed the incompatible protobuf in slack and determined it was fine because this proto isn't part of the transactions proto, nor is conductor currently using this API.

PR #3049, opened 2021-11-05
refactor: merge CI integration tests

tustvold:
This merges the IOx tests into a single job to avoid redundant compilation. It would be even better if we could do the compilation and then run the tests in parallel on multiple machines but one step at a time :laughing: . FWIW the machines we're running on are pretty beefy so we get quite good test run parallelism anyway.


tustvold:
The CI for #3050 shows this is actually running the Kafka tests - https://app.circleci.com/jobs/github/influxdata/influxdb_iox/90058?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

crepererum (changes requested):
Docs needs to be adjusted:
https://github.com/influxdata/influxdb_iox/blob/main/docs/testing.md#kafka-write-buffer

tustvold:
@crepererum in what sense? This doesn't change how the tests actually function, just how they're run in CI?

crepererum:
You delete `docker/ci-kafka-docker-compose.yml` which is used in the linked doc for local Kafka integration test. Would be nice if you could upgrade the doc to use some redpanda image instead (no docker-compose required).

crepererum (approved):

PR #3052, opened 2021-11-05
refactor: Extract a macro for matching errors in tests

carols10cents:
I'm about to add a bunch more assertions that follow this pattern and I was tired of writing it over and over, so I think it's time for this macro :)


test_helpers/src/lib.rs
```diff
@@ -146,3 +146,13 @@ macro_rules! assert_not_contains {
...
+
+#[macro_export]
+/// Assert that an operation fails with one particular error. Panics if the operation succeeds.
+/// Prints debug format of the error value if it doesn't match the specified pattern.
+macro_rules! assert_error {
+    ($OPERATION: expr, $(|)? $( $ERROR_PATTERN:pat_param )|+ $( if $GUARD: expr )? $(,)?) => {
+        let err = $OPERATION.unwrap_err();
+        assert!(matches!(err, $( $ERROR_PATTERN )|+ $( if $GUARD )?), "got: {:?}", err);
+    };
+}
```
shepmaster, line 158: This matches what I've added in a few other projects:

```rust
    macro_rules! assert_error {
        ($e:expr, $p:pat $(if $guard:expr)?) => {
            assert!(
                matches!($e, Err($p) $(if $guard)?),
                "Expected {}, but got {:?}",
                stringify!($p),
                $e,
            )
        };
    }
```

```rust
    macro_rules! assert_error {
        ($e:expr, $p:pat) => {
            assert!(
                matches!($e, Err($p)),
                "Expected {}, but got {:?}",
                stringify!($p),
                $e,
            )
        };
    }
```

test_helpers/src/lib.rs
```diff
@@ -146,3 +146,13 @@ macro_rules! assert_not_contains {
         );
     };
 }
+
+#[macro_export]
+/// Assert that an operation fails with one particular error. Panics if the operation succeeds.
+/// Prints debug format of the error value if it doesn't match the specified pattern.
+macro_rules! assert_error {
+    ($OPERATION: expr, $(|)? $( $ERROR_PATTERN:pat_param )|+ $( if $GUARD: expr )? $(,)?) => {
```
shepmaster, line 154: The handling of `|` [may not be needed in Rust 2021](https://doc.rust-lang.org/edition-guide/rust-2021/or-patterns-macro-rules.html)?

test_helpers/src/lib.rs
alamb: I agree that the inclusion of `Expected` and stringify! would be a good improvement to the usability of this macro (specifically when it failed)

alamb (approved):
❤️  this is *much* nicer @carols10cents 

Something perhaps to consider adding to the stdlib ;) (in all seriousness)

crepererum:
@alamb there's https://doc.rust-lang.org/nightly/std/assert_matches/macro.assert_matches.html but it's not stable yet.

carols10cents:
I added stringify! And I'm excited for when this becomes stable, I didn't know that was in the works, thank you Marco!

PR #3060, opened 2021-11-08
chore: ignore `RUSTSEC-2021-0122`

crepererum:


tustvold (approved):
#3045 came just in time :laughing:

PR #3061, opened 2021-11-08
ci: use new, public `quay.io/influxdb/iox` image

crepererum:


alamb (approved):
I wonder if there any way to test this prior to merging (specifically that whatever permissions are required are setup corectly)?

I looked at circle ci and I couldn't find evidence of a `perf_image` job being run on this branch:

https://app.circleci.com/pipelines/github/influxdata/influxdb_iox?branch=crepererum%2Fnew_quay_image

However, since we don't have auto image promotion, I also don't think it is the end of the world to merge this and then fix it in a subsequent PRs

Issue #2731, opened 2021-10-05
Deletes and Write Buffer


tustvold:
**Problem**

Currently deletes are unaware of the write buffer this means:

* A delete sent to a routing server will be a no-op as the server will have no local data
* A delete sent to a query server following a write buffer will be applied breaking eventual consistency

**Proposal**

Deletes should behave the same as other writes:

* Deletes sent to a routing server are dispatched to the write buffer
* Calling delete directly on a query server following a write buffer will return an error

The routing sever should validate that the delete predicate is valid as much as possible, but at least initially the write buffer format should just contain the string representation. In future we may be able to store the parsed representation instead.

NGA-TRAN:
A must to fully support #2396

tustvold:
So there is a slight complexity w.r.t the interaction between replay and deletes.

**Background**

For background, data is not persisted in sequence number order. This mandates a replay process to recover a restarted database, as it needs to correctly handle what has already been persisted. To do this replay identifies 3 regions within each sequencer during replay

* Fully Persisted - all this data has been persisted
* Partially Persisted - some of this data has been persisted
* Not Persisted - none of this data has been persisted

Replay is solely concerned with the "Partially Persisted" window, and it replays data until it reaches "Not Persisted" data for that sequencer.

**Proposal**

On encountering a delete in the "Partially Persisted" window we **must** only apply it to the in-memory chunks and not the persisted chunks - as the persisted chunks might contain data that was written by a write sequenced after the delete :exploding_head: 

This implicitly assumes that we are not persisting during replay - I have therefore prioritized #2811.

An additional complexity is that deletes themselves are lazily persisted during normal operation, we **must** therefore guarantee that any persist operation for a partition also persists any deletes in the sequence number range identified by its PartitionCheckpoint (the partially persisted window). Ticket TBC

My reasoning for why this works is:

* Any delete in the partially persisted window must have already been applied to any persisted chunks
* Deletes are applied to the in-memory data only, which during replay can only come from writes sequenced before the delete

I _think_ this works but would appreciate if someone else could check my logic. Perhaps @crepererum ?

alamb:
FWIW I think the logic above seems sound

> An additional complexity is that deletes themselves are lazily persisted during normal operation, we must therefore guarantee that any persist operation for a partition also persists any deletes in the sequence number range identified by its PartitionCheckpoint (the partially persisted window). Ticket TBC

[...]

PR #3065, opened 2021-11-08
refactor: move delete predicate proto to own package (#2731)

tustvold:
This moves the protobuf predicate representation into its own package so that it can be used by the write buffer encoding.

I may also move the delete gRPC service to use this representation instead of a string.

This is reported as a breaking change, but as we never serialize the paths in an Any this is fine


generated_types/protos/influxdata/iox/predicate/v1/predicate.proto
```diff
@@ -21,18 +21,6 @@ message Predicate {
   repeated Expr exprs = 5;
 }
 
-// A optional string set.
```
tustvold, line 24: These weren't being used anywhere so I just removed them

crepererum (approved):

e-dard (approved):
👍

PR #3068, opened 2021-11-08
feat: add write buffer delete encoding (#2731)

tustvold:
An RFC for how to represent deletes within the write buffer, and in particular how to handle delete distribution.

**Background**

Currently we have two closely related concepts, shards and sequencers. There are at least two possible models for how these interact:

* Single sequencer per shard, multiple shards per instance
* Single shard per instance, multiple sequencer per shard

Additionally IOx currently permits topologies with multiple layers of HTTP and gRPC routing. I'm not sure if this is intentional but I think it is...

When routing a delete, if it has a single destination the problem is straightforward - the delete is sent to the single destination. This would be the case where a delete specified a table name, and data is being distributed based on table. 

This gets more complex when the delete has multiple destinations. If the router simply cloned the delete to all possible destinations we could end up in a situation where the ordering of writes and deletes is ambiguous. 

Consider a situation where there is a single router server and a single query server, and writes are being distributed across two sequencers. If the delete is blindly sent to both sequencers, depending on the order the query server interleaves the reads from the two sequencers, the deletes will delete different data.

**Proposal**

The key invariant that simple cloning is violating is that a sequenced delete must only modify data that would have been sent to the same sequencer. The fix is therefore relatively straightforward, package information along with the delete that restricts them to exactly this data. This is what this PR does.

### DIFF
--- a/generated_types/protos/influxdata/iox/delete/v1/service.proto
+++ b/generated_types/protos/influxdata/iox/delete/v1/service.proto
@@ -2,12 +2,17 @@ syntax = "proto3";
 package influxdata.iox.delete.v1;
 option go_package = "github.com/influxdata/iox/delete/v1";
 
+import "influxdata/iox/predicate/v1/predicate.proto";
+import "influxdata/iox/predicate/v1/shard.proto";
+
 service DeleteService {
   // Delete data for a table on a specified predicate
   rpc Delete(DeleteRequest) returns (DeleteResponse);
 }
 
 // Request to delete data from a table on a specified predicate
+//
+// TODO: Use Delete
 message DeleteRequest {
   // name of the database
   string db_name = 1;
@@ -27,4 +32,16 @@ message DeleteRequest {
 }
 
 message DeleteResponse {
+}
+
+// A delete payload
+message DeletePayload {
+  // The name of the database
+  string db_name = 1;
+
+  // The predicate identifying data to delete
+  influxdata.iox.predicate.v1.Predicate predicate = 2;
+
+  // An optional conjunctive list of ShardPredicate restricting this delete to a specific shard
+  repeated influxdata.iox.predicate.v1.ShardPredicate shard_predicates = 3;
 }
--- a/generated_types/protos/influxdata/iox/predicate/v1/shard.proto
+++ b/generated_types/protos/influxdata/iox/predicate/v1/shard.proto
@@ -3,19 +3,11 @@ package influxdata.iox.predicate.v1;
 
 import "influxdata/iox/router/v1/shard.proto";
 
-// Represents a predicate that identifies the writes for a specific shard
+// A predicate that identifies the writes for a specific shard
 message ShardPredicate {
-  // An optional list of matchers for this shard
-  repeated influxdata.iox.router.v1.Matcher matchers = 1;
+  // The shard config
+  repeated influxdata.iox.router.v1.ShardConfig config = 1;
 
-  // An optional hash predicate
-  HashPredicate hash = 2;
-}
-
-// A HashPredicate matches if the SipHash 1-3 of the table name lies in
-// the inclusive range `start..=end`
-message HashPredicate {
-  uint64 start = 1;
-
-  uint64 end = 2;
+  // The shard ID
+  uint32 shard = 2;
 }
--- a/generated_types/protos/influxdata/iox/predicate/v1/shard.proto
+++ b/generated_types/protos/influxdata/iox/predicate/v1/shard.proto
@@ -1,0 +1,21 @@
+syntax = "proto3";
+package influxdata.iox.predicate.v1;
+
+import "influxdata/iox/router/v1/shard.proto";
+
+// Represents a predicate that identifies the writes for a specific shard
+message ShardPredicate {
+  // An optional list of matchers for this shard
+  repeated influxdata.iox.router.v1.Matcher matchers = 1;
+
+  // An optional hash predicate
+  HashPredicate hash = 2;
+}
+
+// A HashPredicate matches if the SipHash 1-3 of the table name lies in
+// the inclusive range `start..=end`
+message HashPredicate {
+  uint64 start = 1;
+
+  uint64 end = 2;
+}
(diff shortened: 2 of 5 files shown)


generated_types/protos/influxdata/iox/delete/v1/service.proto
```diff
@@ -28,3 +33,15 @@ message DeleteRequest {
...
+// A delete payload
+message DeletePayload {
+  // The name of the database
+  string db_name = 1;
+
+  // The predicate identifying data to delete
+  influxdata.iox.predicate.v1.Predicate predicate = 2;
+
+  // An optional conjunctive list of ShardPredicate restricting this delete to a specific shard
+  repeated influxdata.iox.predicate.v1.ShardPredicate shard_predicates = 3;
```
tustvold, line 46: I wasn't sure about this, but the repeated is to handle cases where there are many layers of data distribution. As stated in the PR description, I'm not sure how important supporting this actually is, but it isn't hard to support so :shrug:

generated_types/protos/influxdata/iox/predicate/v1/shard.proto
```diff
@@ -0,0 +1,21 @@
+syntax = "proto3";
+package influxdata.iox.predicate.v1;
+
+import "influxdata/iox/router/v1/shard.proto";
+
+// Represents a predicate that identifies the writes for a specific shard
+message ShardPredicate {
+  // An optional list of matchers for this shard
+  repeated influxdata.iox.router.v1.Matcher matchers = 1;
```
tustvold, line 9: I've realised a flaw with this, the Matchers are not guaranteed to be disjoint

crepererum:
Currently we still only use a single sequencer per shard. If we wanna change that (e.g. for high-throughput DBs), we probably need another "sharder level". Since this isn't implemented yet (it's not even designed) I think this whole matcher construct might be a bit premature.

tustvold:
> Currently we still only use a single sequencer per shard. If we wanna change that (e.g. for high-throughput DBs), we probably need another "sharder level". Since this isn't implemented yet (it's not even designed) I think this whole matcher construct might be a bit premature.

This is what I was alluding to with "There are at least two possible models" - I'm not entirely sure how we wish to have this system work and was trying to avoid having those discussions now and just implement something that will work regardless of what we eventually decide :laughing: . I agree it would be nicer to design that system first, but I was wary of opening that particular can of worms :sweat_smile:

tustvold:
Following discussion with @crepererum I have simplified the ShardPredicate to just bundle the ShardConfig. Whilst somewhat odoriferous this is somewhat counter-acted by the fact:

* It is only necessary for deletes with multiple destinations
* Deletes should be comparatively rare anyway
* Using ShardConfig ensures it evolves in parallel

The discussion also yielded a potentially nicer way to view what this PR is doing - it is effectively a form of lazy sharding, where the sharding decision is delegated to the query tier. This is necessary because the query tier is the only part that knows what the data is, what tables there are, etc... Currently we have a for loop for unqualified deletes, which won't work for router which don't know what the tables are - https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs#L491

alamb (commented):
> This gets more complex when the delete has multiple destinations. If the router simply cloned the delete to all possible destinations we could end up in a situation where the ordering of writes and deletes is ambiguous.

I vaguely remember @e-dard  or @pauldix  saying this exactly how the existing TSM system works (effectively sends a delete to all of the TSM partitions) which ensures within a partition there is a clear ordering of writes / deletes, but across partitions a write might (logically) get interleaved with a delete

I don't fully understand how recording a shard_id in a delete will solve the "a single delete applied to multiple shards may get interleaved with concurrent writes from other routers" -- but as mentioned above, I am not sure this is an important case to actually handle either

tustvold:
> sends a delete to all of the TSM partitions) which ensures within a partition there is a clear ordering of writes / deletes, but across partitions a write might (logically) get interleaved with a delete

This is correct, the distinction within TSM is that partitions (which are shards in IOx) exist as a concept within the storage nodes. To put it another way, the storage node knows which series belong to what TSM partition. This means when a delete arrives on partition 5, it can apply it to just the series that belong to partition 5. 

The problem in IOx is that a query server has no idea what the routing topology is, and this whole proposal is basically a mechanism to work around this fact. A perfectly valid alternative mechanism would be for the query servers to have a defined notion of sharding. I personally think this would be a better design, but was trying to avoid creating controversy by proposing it :laughing: Perhaps I'll just write up a design doc for this...

alamb:
> The problem in IOx is that a query server has no idea what the routing topology is, and this whole proposal is basically a mechanism to work around this fact. A perfectly valid alternative mechanism would be for the query servers to have a defined notion of sharding. I personally think this would be a better design, but was trying to avoid creating controversy by proposing it 😆 Perhaps I'll just write up a design doc for this...

I guess I was thinking "why not just send the delete to all shards?" (even if it could be proven, with some additional cleverness, that some shards didn't have any matching rows).

e-dard:
I talked through this with @tustvold. 

@alamb
> I vaguely remember @e-dard or @pauldix saying this exactly how the existing TSM system works (effectively sends a delete to all of the TSM partitions) which ensures within a partition there is a clear ordering of writes / deletes, but across partitions a write might (logically) get interleaved with a delete

Correct. All partitions apply all deletes. The key thing though is that TSM the sub-set of data held within a storage partition is physically isolated from other partitions. This means that when a partition applies the delete "to everything" it's actually only applying it to the sub-set of data it can see and is responsible for.

Just writing out what I understood from the conversation with @tustvold. The issue is that there is no such physical isolation inside of a query server. Therefore, assuming there were two partitions for the same database that were being consumed onto a query node. Yes the produce side could multiplex the same "all-tables" delete onto both partitions.

The issue will be that if one partition is ahead of the other it will process the delete first and the delete data for tables _that belongs to the other partition_. Now of course this data needs to be deleted anyway (and when the other partition catches up that delete will re-apply the same delete). But it is inconsistent and doesn't seem right. So I do think we need to prevent that from happening.

This PR fixes it by effectively taking a delete on the producer side of kafka and annotating the delete for each partition with something that restricts the delete to only data in that partition (e.g., a set of tables).

Another way to look at this is to teach the query server to only apply a delete to the chunks that belong to the partition (shard) the delete was consumed from.

tustvold:
We have instead decided to make the servers consuming from the write buffer shard aware

PR #3074, opened 2021-11-09
refactor: partially split up HTTP implementation and tests

crepererum:
This should make it easier to to wire up the HTTP write interface for
the router. No functional change intended, this just moves code around.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/http/write.rs
+++ b/influxdb_iox/src/influxdb_ioxd/http/write.rs
@@ -1,0 +1,236 @@
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use chrono::Utc;
+use data_types::{
+    names::{org_and_bucket_to_database, OrgBucketMappingError},
+    DatabaseName,
+};
+use hyper::{Body, Method, Request, Response, StatusCode};
+use mutable_batch::{DbWrite, WriteMeta};
+use observability_deps::tracing::debug;
+use serde::Deserialize;
+use snafu::{OptionExt, ResultExt, Snafu};
+
+use crate::influxdb_ioxd::{
+    http::utils::parse_body,
+    server_type::{ApiErrorCode, RouteError, ServerType},
+};
+
+use super::metrics::LineProtocolMetrics;
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Snafu)]
+pub enum HttpWriteError {
+    #[snafu(display("Internal error mapping org & bucket: {}", source))]
+    BucketMappingError { source: OrgBucketMappingError },
+
+    #[snafu(display(
+        "Internal error writing points into org {}, bucket {}:  {}",
+        org,
+        bucket_name,
+        source
+    ))]
+    WritingPoints {
+        org: String,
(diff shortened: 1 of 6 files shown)


influxdb_iox/src/influxdb_ioxd/http/write.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    /// Line protocol metrics.
+    fn lp_metrics(&self) -> Arc<LineProtocolMetrics>;
+
+    /// Perform write.
+    async fn write(
+        &self,
+        db_name: &DatabaseName<'_>,
+        write: DbWrite,
+    ) -> Result<(), InnerWriteError>;
```
crepererum, line 228: This is the interface a server type has to implement to be able to accept write requests. It's not super beautiful (esp. because we don't have a common "application" type for the server types yet), but it works. I expect this to be cleaned up a bit in the future.

influxdb_iox/src/influxdb_ioxd/http/write.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    /// Request was consumed and transformed into a response object. Routing was successfull.
+    Response(Response<Body>),
+}
+
+#[async_trait]
+pub trait HttpDrivenWrite: ServerType {
+    /// Routes HTTP write requests.
+    ///
+    /// Returns `RequestOrResponse::Response` if the request routed, otherwise the route does not match.
```
alamb, line 118:
```suggestion
    /// Returns `RequestOrResponse::Response` if the request was routed, 
    /// Returns `RequestOrResponse::Response` if the request did not match (and needs to be handled some other way)
```

influxdb_iox/src/influxdb_ioxd/http/write.rs
```diff
@@ -0,0 +1,236 @@
...
+        let db_name = org_and_bucket_to_database(&write_info.org, &write_info.bucket)
+            .context(BucketMappingError)?;
+
+        let body = parse_body(req, max_request_size).await.context(ParseBody)?;
+
+        let body = std::str::from_utf8(&body).context(ReadingBodyAsUtf8)?;
+
+        // The time, in nanoseconds since the epoch, to assign to any points that don't
+        // contain a timestamp
+        let default_time = Utc::now().timestamp_nanos();
```
alamb, line 148: at some point, this should probably take a `TimeProvider` instead of `Utc::now()` but that doesn't seem overly crucial at this point

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
```diff
@@ -267,174 +209,64 @@ impl From<server::Error> for ApplicationError {
...
-                stats.num_fields,
-                body.len(),
-                false,
-            );
-            Err(ApplicationError::WritingPoints {
-                org: write_info.org.clone(),
-                bucket_name: write_info.bucket.clone(),
-                source: Box::new(e),
-            })
+    match server_type
```
alamb, line 251: this is cool

alamb (approved):
Makes sense to me 👍

PR #3076, opened 2021-11-09
fix: HTTP server error handling clean up

### LINKED ISSUE #3075: Error returned when writing into a database that does not exist
I noticed when attempting to write into a database that does not exist an `error` is returned in the logs:

```
2021-11-09T11:50:45.641655Z ERROR influxdb_iox::influxdb_ioxd::http: Error while handling request error=Route error from run mode: Database 54425547585a6420_000000000153158e not found method=POST uri=/api/v2/write?bucket=000000000153158e&org=54425547585a6420 content_length=None
```

The `Error` log level should ideally be reserved for events/situations that require operator intervention, and where there is no sensible action for the server to take.

I added the following:

```rust
            let x = error.api_error_code();
            error!(%x, "code is");
```

to https://github.com/influxdata/influxdb_iox/blob/17ecec6d378578d94761a1e56cef8f769189a0ca/influxdb_iox/src/influxdb_ioxd/http/mod.rs#L146-L149

And noticed in the logs:

```
2021-11-09T11:50:45.641635Z ERROR influxdb_iox::influxdb_ioxd::http: code is x=100
2021-11-09T11:50:45.641655Z ERROR influxdb_iox::influxdb_ioxd::http: Error while handling request error=Route error from run mode: Database <redacted> not found method=POST uri=/api/v2/write?bucket=<redacted>&org=<redacted> content_length=None
```

This suggests that the application error associated with a database not existing:

https://github.com/influxdata/influxdb_iox/blob/17ecec6d378578d94761a1e56cef8f769189a0ca/influxdb_iox/src/influxdb_ioxd/server_type/mod.rs#L22-L33

Is getting lost in translation somehow...

crepererum:
1. Only log on error level for internal errors
2. Get rid of `api_error_code` since its interaction with nested errors
   is really confusing (e.g. not forwarding the correct error code
   doesn't have an effect because the way the actual response body is
   created but when debugging you'll see another error code).

Fixes #3075.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
@@ -41,17 +41,17 @@ impl From<ApiErrorCode> for u32 {
 pub trait RouteError: std::error::Error + snafu::AsErrorSource {
     fn response(&self) -> Response<Body>;
 
-    fn bad_request(&self) -> Response<Body> {
+    fn bad_request(&self, api_error_code: ApiErrorCode) -> Response<Body> {
         Response::builder()
             .status(StatusCode::BAD_REQUEST)
-            .body(self.body())
+            .body(self.body(api_error_code))
             .unwrap()
     }
 
-    fn internal_error(&self) -> Response<Body> {
+    fn internal_error(&self, api_error_code: ApiErrorCode) -> Response<Body> {
         Response::builder()
             .status(StatusCode::INTERNAL_SERVER_ERROR)
-            .body(self.body())
+            .body(self.body(api_error_code))
             .unwrap()
     }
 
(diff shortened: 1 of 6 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
```diff
@@ -62,23 +62,28 @@ pub trait RouteError: std::error::Error + snafu::AsErrorSource {
             .unwrap()
     }
 
-    fn no_content(&self) -> Response<Body> {
+    fn no_content(&self, api_error_code: ApiErrorCode) -> Response<Body> {
         Response::builder()
             .status(StatusCode::NO_CONTENT)
-            .body(self.body())
+            .body(self.body(api_error_code))
```
tustvold, line 68: I know this PR didn't introduce this, but why does NO_CONTENT have a body?

tustvold (commented):
Looking at the code we have the following error codes:

```
pub enum ApiErrorCode {
    /// An unknown/unhandled error
    UNKNOWN = 100,

    /// The database name in the request is invalid.
    DB_INVALID_NAME = 101,

    /// The database referenced already exists.
    DB_ALREADY_EXISTS = 102,

    /// The database referenced does not exist.
    DB_NOT_FOUND = 103,
}
```

Is there a particular reason we need to keep these error codes at all, if memory serves they date from before the gRPC API and were based around an expectation of conductor calling this API.

I wonder if we could just use the corresponding HTTP status codes and be done with it?

influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
crepererum: :shrug:

crepererum:
Let me clarify the open question (need to find someone who actually designed this).

crepererum:
Closing this PR, a proper clean up would conflict too much w/ #3077, so I'll postpone the clean up a bit.

PR #3077, opened 2021-11-09
feat: HTTP write interface for new router

crepererum:
~On top of #3076.~


influxdb_iox/src/influxdb_ioxd/http/write.rs
```diff
@@ -235,3 +235,300 @@ pub struct WriteInfo {
     pub org: String,
     pub bucket: String,
 }
+
+#[cfg(test)]
+pub mod test_utils {
```
alamb, line 240: I don't think it matters a lot, but I found the fact there `influxdb_iox/src/influxdb_ioxd/http/test_utils.rs` as well as a sub module `test_utils` in `influxdb_iox/src/influxdb_ioxd/http/write.rs` somewhat confusing (could we perhaps consolidate into `influxdb_iox/src/influxdb_ioxd/http/test_utils.rs` to improve the chance of finding relevant testing code? 

I don't feel strongly about this, but I wanted to point it out as i was reading this PR

influxdb_iox/src/influxdb_ioxd/http/write.rs
```diff
@@ -235,3 +235,300 @@ pub struct WriteInfo {
...
+                ("db_name", "MyOrg_MyBucket"),
+                ("status", "ok"),
+            ]))
+            .unwrap()
+            .fetch();
+        assert_eq!(observation.total, 98);
+        assert_eq!(observation.buckets[0].count, 1);
+        assert_eq!(observation.buckets[1].count, 0);
+
+        // Write to a non-existent database
```
alamb, line 483: stylistically I would find it easier to read this test (and quickly evaluate the current state of coverage) if there were a separate test for the positive and negative cases but combining both does save some repetition

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
```diff
@@ -453,75 +459,55 @@ mod tests {
...
-                test_server.url(),
-                bucket_name,
-                org_name
-            ))
-            .body(lp_data)
-            .send()
-            .await;
+    #[tokio::test]
+    async fn test_write_metrics() {
+        assert_write_metrics(setup_server().await, true).await;
```
alamb, line 497: Since it appears that `assert_write_metrics` is only called at this callsite , I don't understand the rationale for the indirection by factoring it into a function in another module. I am likely missing something

influxdb_iox/src/influxdb_ioxd/server_type/router/http.rs
```diff
@@ -1,49 +1,105 @@
...
-        (method, path) => Err(ApplicationError::RouteNotFound {
-            method,
-            path: path.to_string(),
+    match server_type
+        .route_write_http_request(req)
+        .await
+        .context(WriteError)?
+    {
+        RequestOrResponse::Response(resp) => Ok(resp),
+        RequestOrResponse::Request(req) => Err(ApplicationError::RouteNotFound {
```
alamb, line 79: Is there any test coverage for this case (where the router wasn't able to route http request?)

alamb (approved):
Looks good to me -- keep the changes flowing 👍 

I had some stylistic comments but I don't think any of them require any action

influxdb_iox/src/influxdb_ioxd/http/write.rs
crepererum: I tried to keep the `test_utils` close to the code that they help testing. So you don't end up w/ a 1000 lines test kitchen sink.

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
crepererum: It's also used in `influxdb_iox/src/influxdb_ioxd/server_type/router/http.rs`.

influxdb_iox/src/influxdb_ioxd/server_type/router/http.rs
crepererum: `assert_metrics` contains a 404-relalted check.

influxdb_iox/src/influxdb_ioxd/http/write.rs
crepererum: Yeah, now copy-pasting or abstracting all this "get the right observer" thing sounds like more effort than it is worth it.

influxdb_iox/src/influxdb_ioxd/http/write.rs
alamb: yeah, the issue is you may end up with 4 x 300 = 1200 line test kitchen sinks (as you may end up not being able to find something that is reusable). 🤷

PR #3079, opened 2021-11-10
ci: fix usage of old docker version

crepererum:
`--all-tags` as introduced for version 20.10 but appearently we're
running an older version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -360,10 +360,11 @@ jobs:
           # Disabling for now, and tracked further investigations
           # in https://github.com/influxdata/k8s-idpe/issues/3038
           docker_layer_caching: false
-          version: 20.10.7
+          version: 19.03.14
       - run: |
           sudo apt-get update
           sudo apt-get install -y docker.io
+          docker --version
       - run: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -380,7 +381,7 @@ jobs:
           sha256sum target/release/influxdb_iox
           COMMIT_SHA=$(git rev-parse --short HEAD)
           docker build -t quay.io/influxdb/iox:$COMMIT_SHA -t quay.io/influxdb/iox:main -f docker/Dockerfile.iox .
-          docker push quay.io/influxdb/iox --all-tags
+          docker push quay.io/influxdb/iox
```
tustvold, line 384: Does this not implicitly push `quay.io/influxdb/iox:latest` ?

tustvold (approved):
FWIW you might want to experiment with enabling the CI job for this specific branch if you want to iterate on what is going on

.circleci/config.yml
crepererum: Yes, it pushes all tags. That behavior was removed later on and `--all-tags` is required. But our debian builder is old so the docker client is old.

PR #1619, opened 2021-06-03
feat: Preserve full debug infos in release builds

mkmik:
The `Cargo.toml` file already says:

```toml
[profile.release]
debug = true
```

but we were overriding it in the circleci release script to include only line number info,
probably to shave off some time from the total build time.

The ability to debug a production binary outweighs the few seconds or even a minute overhead it takes to
add the debug infos. If building with full debug infos take much much longer, next time we disable them
it would be nice to include such a dramatic fact in the comments next to where we disable debug infos :-)

PR #3081, opened 2021-11-10
chore: faster and smaller IOx binaries

crepererum:
chore: faster and smaller IOx binaries

Use `codegen-units = 1`, thin-LTO and debug section compression to make our binary smaller (which is good for deploy and
test times) and faster.

# Summary

The binary size of `influxdb_iox` after building with:

```sh
cargo build --release --no-default-features --features="aws,gcp,azure,jemalloc_replacing_malloc"
```

The profile was:

```toml
[profile.release]
debug = true
```

The commit was:

```text
89ece8b493c924d05ebfa75d6661be5462eaeefa
```

The size results are:

| Method                                     | Size  |
| ------------------------------------------ | ----- |
| baseline                                   | 833MB |
| baseline + dbg compression                 | 222MB |
| baseline + strip                           |  49MB |
| codegen-units                              | 520MB |
| codegen-units + strip                      |  40MB |
| codegen-units + dbg compression            | 143MB |
| thin LTO                                   | 715MB |
| thin LTO + strip                           |  49MB |
| thin LTO + dbg compression                 | 199MB |
| codegen-units + thin LTO                   | 449MB |
| codegen-units + thin LTO + strip           |  40MB |
| codegen-units + thin LTO + dbg compression | 130MB |

For the methods that were successfully measured I couldn't really see any compile time differences on my laptop.

# Methods

## Strip
Remove debug symbols. We don't really want this, so this is just to get an idea of the size

```console
$ strip baseline
```

## Debug Sections compression
Debug sections make a large amount of our binary size (a stripped executable is 49MB instead of 833MB). Since we like to
have debug symbols we cannot just strip them. However these symbols are only used for:

- backtrace generation (something went wrong, not BAU)
- profiling
- debugging

So in normal operation and most test scenarios, we're just wasting memory. So we could compress them:

```console
$ objcopy --compress-debug-sections baseline baseline-dbg_compressed
```

There is also elfutils:

```console
$ eu-elfcompress test
```

Elfutils nearly ends up with the same size (220MB instead of 222MB that objcopy achieves), but takes more time and is
probably not worth it.

Note that compressed debug sections exist since many years. The Rust ecosystem supports reading them since over a year,
see:

- <https://github.com/gimli-rs/gimli/issues/195>
- <https://github.com/rust-lang/backtrace-rs/issues/342>

## Codegen Units
The rust compiler parallelizes codegen work. This split into units however means that optimizations are somewhat
limited. This can be change by:

```toml
[profile.release]
...
codegen-units = 1
```

As a nice side effect this should also make our code faster.

## Thin LTO
Get LLVM to run "thin" Link Time Optimization:

```toml
[profile.release]
...
lto = "thin"
```

As a nice side effect this should also make our code faster.

## Fat LTO
Get LLVM to run "fat" Link Time Optimization:

```toml
[profile.release]
...
lto = "fat"
```

There are no results for this because this took a massive amount of memory and CPU time and did not finish on my system.


shepmaster:
I suppose that split debug info wouldn't be useful?

crepererum:
> I suppose that split debug info wouldn't be useful?

IIRC this is not supported out of the box on Linux yet ( https://github.com/rust-lang/rust/issues/34651 ), we could however use objcopy and friends to do that manually I suppose.

**For debugging (needs full debug symbols):** We could lazy-load symbols (we don't have any infrastructure for that though, so we would need some S3 upload and download script or a symbol server).

**For profiling (I think that only needs line symbols):** It would be even more effort, because there we just use `pprof-rs` which uses `backtrace-rs` and I guess it would be tricky to hook into that.

**For panic backtraces (only needs line symbols):** I doubt that lazy loading is possible w/ reasonable effort.

So we need at least the line symbols within the image itself. We could lazy-load full symbols. However that's more effort than this PR here and I think the current solution is "good enough".

Another idea would be to compress the split symbols w/ something really strong (like zstd) and decompress on image startup, but that would delay the image startup and needs temp disc space, so that sounds a bit like a no go.

.circleci/config.yml
```diff
@@ -348,8 +348,11 @@ jobs:
       - run:
           name: Cargo release build with target arch set for CRoaring
           command: cargo build --release --no-default-features --features="aws,gcp,azure,jemalloc_replacing_malloc"
+      - run:
+          name: Compress debug symbols
+          command: objcopy --compress-debug-sections target/release/influxdb_iox
       - run: |
-          echo sha256sum after build is
+          echo "binary sha256sum after build is (not to be confused w/ the GIT SHA and the resulting image SHA!)"
```
alamb, line 355: ❤️  there are a lot of SHA's :)

alamb (commented):
FWIW I think debug sections (as opposed to symbols) are only needed if we actually want to attach `gdb` / `lldb` to a running `influxdb_iox` process. You don't need debug info for backtraces (you only need symbols).

Also, I have never had any luck in the past (admittedly this is like 10 years ago) getting anything other than a backtrace from a `release` binary (aka  what you get from `gcc -O3`). The code was always so optimized that any attempt to inspect locals was misleading at best. 

This, I wonder if **we should simply stop adding debug symbols** to release binaries, or maybe call `strip` to remove the debug sections in the binaries placed in the image? I also wonder if `lldb` / `gdb` knows how to read gzipped sections..

@mkmik  changed the build to add debug sections in https://github.com/influxdata/influxdb_iox/pull/1619 so perhaps he wants to weigh in

alamb:
Thank you @crepererum  -- this is very helpful

.circleci/config.yml
```diff
@@ -348,8 +348,11 @@ jobs:
       - run:
           name: Cargo release build with target arch set for CRoaring
           command: cargo build --release --no-default-features --features="aws,gcp,azure,jemalloc_replacing_malloc"
+      - run:
+          name: Compress debug symbols
+          command: objcopy --compress-debug-sections target/release/influxdb_iox
```
alamb, line 353: I have never seen this option before 👍

mkmik:
> FWIW I think debug sections (as opposed to symbols) are only needed if we actually want to attach `gdb` / `lldb` to a running `influxdb_iox` process. You don't need debug info for backtraces (you only need symbols).
> 
> Also, I have never had any luck in the past (admittedly this is like 10 years ago) getting anything other than a backtrace from a `release` binary (aka what you get from `gcc -O3`). The code was always so optimized that any attempt to inspect locals was misleading at best.
> 
> This, I wonder if **we should simply stop adding debug symbols** to release binaries, or maybe call `strip` to remove the debug sections in the binaries placed in the image? I also wonder if `lldb` / `gdb` knows how to read gzipped sections..
> 
> @mkmik changed the build to add debug sections in #1619 so perhaps he wants to weigh in

IIRC backtracks require DWARF debug info because that's the only way to walk the stack without frame pointers. (Frame pointers are no longer used by default on x86_64; they can be enabled with. RUSTFLAGS="-Cforce-frame-pointers=yes". Nowadays tools like dtrace and ebpf know how to walk the stack with dwarf info so there is less and less need to force generating code with frame pointers)

That said, that only requires a smaller amount of debug info; not sure if that's the "line symbols" that crepererum was mentioning.

mkmik:
I _think_ we already set debuginfo=1 which should be this minimal debug info needed for backtraces: Not sure if there is an even smaller set of debug info that has the bare minimum info for walking the stack without full line info

crepererum:
> I _think_ we already set debuginfo=1 which should be this minimal debug info needed for backtraces: Not sure if there is an even smaller set of debug info that has the bare minimum info for walking the stack without full line info

We don't for release builds:

https://github.com/influxdata/influxdb_iox/blob/f650962221bf9fbcce9aedba3c4f73e98e1cb7f6/.circleci/config.yml#L311

crepererum:
> not sure if that's the "line symbols" that crepererum was mentioning.

"line tables" would have been the correct naming, see <https://doc.rust-lang.org/cargo/reference/profiles.html#debug>

tustvold:
I have found that with release builds it is basically necessary to have the DWARF information in order to get sane profiles of a release build. The combination of symbol mangling, aggressive inlining, const-instantiation, monomorphisation and probably some other things, make the symbol table alone pretty limited.

Compare the output of `perf record` and `perf record --call-graph dwarf`

My memory of this is relatively fuzzy, but a long while ago I wrote a tool similar to https://godbolt.org/ for analysing Broadcom's embedded compilers - so I at one point understood this at least :laughing: Having a cursory look at the [spec](https://refspecs.linuxfoundation.org/elf/elfspec.pdf), my recollection seems not totally inaccurate

The ELF symbol table is only guaranteed to contain things necessary for linking, in practice this is potentially mangled non-inlined functions. Every symbol in this table may have information about its source location, although this may not be present. This is why you can sometimes set a breakpoint on a function name, but not the corresponding source lines. This is also why `gdb` and other source debuggers can behave strangely when setting line number breakpoints in release builds without DWARF information, it will set the breakpoint on the nearest line that has a symbol, which may not even be in the same function!

The DWARF information is what provides the necessary information to map lines of source code, to blocks of instructions. This includes information about inlined callsites, const instantiations, etc... I'm not sure if it contains information about monomorphisation, name mangling, etc... - I was focused on C code, but it is possible. In short without this information source debugging and profiling are likely to be severely impaired.

I'm not entirely sure what information is used to generate backtraces, but I'd be surprised if this is using the DWARF information and not just the symbol table

mkmik:
It's more than that. Without frame pointers you have no idea at which position of the stack you can find the address of the caller function. DWARF info allows you to reconstruct the original stack pointer as if it was at the functions entry point, because it tracks how the stack pointer gets manipulated at the various points throughout the function body.

This is stored in the .eh_frame section, which is technically not a debug section but it's still part of the DWARF spec, so it's often conflated with debug data (as it seems to be the case with the rust flags). That info also allows to track where each single register is spilled so that it can not only walk the stack in order to print a backtraces but also recover all state to a proper unwind, useful for catching/recovering exceptions/panics.

crepererum:
> > codegen-units + thin LTO + dbg compression
> 
> This seems like an unambiguous win to me. If I'm understanding correctly, we don't lose anything and get smaller binaries. **My only question would be the impact it has on compile times**, it's less common but whilst iterating on benchmarks release build performance can matter

I was wondering the same and locally it didn't make much of the difference (see comment right below the sizing table).

mkmik:
```console
$ objdump -h ./target/release/influxdb_iox | grep eh_frame | gawk --non-decimal-data '{printf("%d\n","0x" $3)}' | awk '{s+=$1} END {print s}' | numfmt --to=iec "$@"
4.8M
```

mkmik:
```console
$ objdump -h ./target/release/influxdb_iox | grep debug | gawk --non-decimal-data '{printf("%d\n","0x" $3)}' | awk '{s+=$1} END {print s}' | numfmt --to=iec "$@"
785M
```

mkmik:
```console
objdump -h ./target/release/influxdb_iox | grep debug | gawk --non-decimal-data '{n=sprintf("%d", "0x" $3); "numfmt --to=iec " n |& getline h; print h, $2}' | sort -h -r
211M .debug_info
182M .debug_loc
108M .debug_pubtypes
94M .debug_str
83M .debug_pubnames
67M .debug_ranges
39M .debug_line
3.0M .debug_aranges
1.9M .debug_abbrev
133K .debug_macro
304 .debug_frame
34 .debug_gdb_scripts
```

mkmik:
when compiling with `-C debuginfo=0` the compiler still generates the `.eh_frame` data which is enough to walk the stack and get the function addresses (which is a pre-requisite for function->symbol resolution).

```
stack backtrace:
   0: std::panicking::begin_panic
   1: rustack::bar
   2: rustack::foo
   3: rustack::main
```

The backtrace worked even if I manually stripped some debug sections (I tried `objcopy --remove-section .debug_str ./target/release/rustack`)

crepererum:
I would argue that we probably also want line numbers and not just function names though.

mkmik:
I agree. Let's not over optimize the image size.

tustvold (approved):

PR #3082, opened 2021-11-11
ci: run clippy for all features

### LINKED ISSUE #537: ci: run cargo check under all cfg variations
__Use case:__
Currently code that is behind a cfg attribute and is not compiled in by default is not checked as part of the CI process, specifically the `test_aws` and `test_gcs` cfgs (others?) in the `object_store`.

__Proposal:__
Until we can run e2e against these stores, we should at least run `cargo check` to ensure the code is valid.

__Current behaviour:__
Code behind the cfg is ignored and not checked for correctness.

__Desired behaviour:__
A CI failure when code in these areas does not compile.

__Alternatives considered:__
Ignoring the problem at the expense of future sanity.

crepererum:
This indeed fixes some issues in our object store implementation.

Closes #537.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ jobs:
       - cache_restore
       - run:
           name: Clippy
-          command: cargo clippy --all-targets --workspace -- -D warnings
+          command: cargo clippy --all-targets --all-features --workspace -- -D warnings
       - cache_save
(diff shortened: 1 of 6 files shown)


e-dard (approved):
👌

alamb (commented):
👌

PR #3084, opened 2021-11-11
refactor: extract DML types (#2731)

tustvold:
As part of routing deletes through the write buffer, I need to define an in-memory type for the data that is being sequenced. Following internal discussion we decided to call this DML to be consistent with other database's name for this.

It didn't seem write to define these types in `mutable_batch` as it has no concept of a delete, but it also seemed unfortunate to define them in `write_buffer` and therefore force unrelated code to depend on things like rdkafka. I therefore lifted and renamed WriteMeta and DbWrite into a new crate called dml


dml/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+use data_types::database_rules::{ShardConfig, ShardId, Sharder};
+use data_types::partition_metadata::{StatValues, Statistics};
+use data_types::sequence::Sequence;
+use mutable_batch::MutableBatch;
+use time::Time;
+use trace::ctx::SpanContext;
+
+/// Metadata information about a DML operation
+#[derive(Debug, Default, Clone)]
+pub struct DmlMeta {
```
tustvold, line 25: This was previously called WriteMeta

dml/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+/// A DML operation
+#[derive(Debug, Clone)]
+pub enum DmlOperation {
+    /// A write operation
+    Write(DmlWrite),
+}
+
+/// A collection of writes to potentially multiple tables within the same database
+#[derive(Debug, Clone)]
+pub struct DmlWrite {
```
tustvold, line 95: This was previously called DbWrite

dml/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    /// Returns the number of bytes read from the wire if relevant
+    pub fn bytes_read(&self) -> Option<usize> {
+        self.bytes_read
+    }
+}
+
+/// A DML operation
+#[derive(Debug, Clone)]
+pub enum DmlOperation {
```
tustvold, line 88: This is what the write buffer will "sequence". To be added in a separate PR

alamb: specifically `Delete` will be added in a separate PR 👍

alamb (approved):
Looks like a nice refactor to me

PR #3087, opened 2021-11-11
feat: remove legacy per-line partitioning logic

tustvold:
This was replaced by PartitionWrite added by #2974 and is no longer used


data_types/Cargo.toml
```diff
@@ -8,8 +8,6 @@ readme = "README.md"
 
 [dependencies] # In alphabetical order
 bytes = "1.0"
-chrono = "0.4"
```
tustvold, line 11: :tada:

PR #3088, opened 2021-11-11
chore: allow image server mode to be chosen via env var

crepererum:


mkmik (commented):
hmm, not sure I like this. Usually it's nice to be able to run other commands with a docker image, e.g. `docker run $IOX_IMAGE database list" without having to force the entrypoint

crepererum:
@mkmik fixed, PTAL

mkmik (approved):
Thanks!

PR #3094, opened 2021-11-12
refactor: move delete predicate into data_types (#2731)

tustvold:
As part of #2731 I need to be able to express a delete to `DmlOperation`. Unfortunately simply making `dml` depend on `predicate` would drag `datafusion`, `serde_json`, `tonic`, etc... along for the ride. This is not ideal.

I therefore pulled DeletePredicate and friends into `data_types` and rewrote the conversion logic in `predicate` to use this. Unfortunately the trait orphan rules mean that From/TryFrom, etc... cannot be defined here. I'm not sure this is a big loss as they weren't actually being used anywhere outside the crate.

Related to this, currently `predicate` is a dependency of query and other parts of the system, however, it also bundles all the conversion logic. Not only does this make for a confusing API leakage, where details of the Influx delete compatibility serialization leak into other parts of the system, but also drags along the entire tonic dependency chain. I've added TODOs to move these out of `predicate`, possibly into `predicate_conversions` or closer to where they are actually used. FYI @crepererum


dml/src/lib.rs
```diff
@@ -195,6 +197,119 @@ impl DmlWrite {
     }
 }
 
+/// Represents a parsed delete predicate for evaluation by the InfluxDB IOx
```
tustvold, line 200: Just moved from `predicate`

predicate/src/delete_predicate.rs
```diff
@@ -297,14 +271,17 @@ fn parse_time_range(start: &str, stop: &str) -> Result<(i64, i64)> {
...
+/// https://docs.influxdata.com/influxdb/v2.0/write-data/delete-data/, which happens before
+/// the parsing of timestamps and sql predicate. The examples below will show FLUX DELETE's syntax which is
+/// different from SQL syntax so we need this extra parsing step before invoking sqlparser to parse the
+/// sql-format predicates and timestamps
+///
+/// TODO: Move alongside HTTP code
 #[derive(Debug, Default, PartialEq, Clone)]
-/// data of a parsed delete
-pub struct ParsedDelete {
+pub struct HttpDeleteRequest {
```
tustvold, line 284: I opted to rename this, as it took me an embarrassingly long time to work out what it was

parquet_catalog/Cargo.toml
```diff
@@ -10,6 +10,7 @@ bytes = "1.0"
 data_types = { path = "../data_types" }
 datafusion = { path = "../datafusion" }
 datafusion_util = { path = "../datafusion_util" }
+dml = { path = "../dml" }
```
tustvold, line 13: Perhaps I'll create a `delete_predicate` crate to avoid this being such a chonky dependency addition :thinking:

data_types/src/delete_predicate.rs
```diff
@@ -0,0 +1,100 @@
+use crate::timestamp::TimestampRange;
```
tustvold, line 1: These structs are just moved out of predicate

tustvold:
I originally opted to move these into `dml` but doing so introduced a dependency on `mutable_batch` for crates such as `parquet_catalog`, etc... I therefore decided to just move them into data_types. I do have a slight concern that data_types is getting rather broad, but perhaps this is unavoidable.

data_types/src/delete_predicate.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub range: TimestampRange,
+
+    /// Optional arbitrary predicates, represented as list of
+    /// expressions applied a logical conjunction (aka they
+    /// are 'AND'ed together). Only rows that evaluate to TRUE for all
+    /// these expressions should be returned. Other rows are excluded
+    /// from the results.
+    pub exprs: Vec<DeleteExpr>,
+}
+
```
NGA-TRAN, line 18: I do not see `impl DeletePredicate`. Any reason for that?

NGA-TRAN (approved):
All the changes seem reasonable to the move of the struct. Nice job, @tustvold

data_types/src/delete_predicate.rs
tustvold: There were only two member functions before, try_new and `all_column_names_but_time`. `try_new` requires the parsing logic and so was left behind in `predicate` and renamed to `parse_delete_predicate`. `all_column_names_but_time` was a bit of a strange member function imo, and also needs schema::TIME_COLUMN_NAME. Rather than introducing a schema dependency into data_types or moving the constant around, I just modified `QueryChunkMeta::delete_predicate_columns` to "inline" this logic.

tustvold:
Will leave this open for @crepererum to review on Monday morning if he wishes to

data_types/src/delete_predicate.rs
NGA-TRAN: Yeah, saw those. Sounds good

crepererum (approved):

PR #3096, opened 2021-11-12
feat: WriteBufferReader use DmlOperation (#2731)

tustvold:
In order to support deletes within the write buffer, we need the reader stream to support payloads of more than one type. This alters the write buffer reader to be a stream of DmlOperation instead of DmlWrite. Currently as DmlOperation has only one variant this is a little bit redundant, but will make more sense once deletes are in there as well.


crepererum (approved):
Looks good. Just checking: is `write_buffer::codec::CONTENT_TYPE_PROTOBUF` still correct? (esp. the exact message that is named by it)

PR #3107, opened 2021-11-15
feat: improve `RouterClient` errors

crepererum:
Found while working on conductor.


tustvold (approved):
I'm not a massive fan of wrapping each of these methods with its own error type that just exposes a subset of the variants of tonic::Status, but I guess this is the convention at the moment :shrug:

alamb:
> I'm not a massive fan of wrapping each of these methods with its own error type that just exposes a subset of the variants of tonic::Status, but I guess this is the convention at the moment 🤷

What is the alternative @tustvold ?  Would you suggest a single `Error` type shared by all the function calls as is done elsewhere in the code?

tustvold:
> What is the alternative @tustvold ? Would you suggest a single Error type shared by all the function calls as is done elsewhere in the code?

Possibly, or just returning the `tonic::Status`... The gRPC designers already went to the effort of defining useful error variants, and mapping them to error codes, it seems somewhat unnecessary to be building our own... It's also possible I'm just grumpy about the amount of error propagation code we have, and this is an easy target :laughing:

alamb:
🤔  indeed

PR #3108, opened 2021-11-15
refactor: move delete predicate proto serialization to generated_types

tustvold:
This achieves a couple of ends:

* It allows write_buffer to serialize deletes without depending on `predicate` and by extension `datafusion`
* It allows query to depend on `predicate` without depending on `generated_types` and by extension `tonic`
* Is consistent with how other protobuf conversion is handled
* Can use TryFrom, etc...
* Is slightly less verbose


generated_types/src/write_buffer.rs
```diff
@@ -47,14 +44,8 @@ impl TryFrom<write_buffer::WriteBufferConnection> for WriteBufferConnection {
...
 
-        let direction: Direction =
-            Direction::from_i32(proto.direction).ok_or_else(|| FieldViolation {
-                field: "direction".to_string(),
-                description: "Cannot decode enum variant from i32".to_string(),
-            })?;
-
         Ok(Self {
-            direction: direction.try_into()?,
+            direction: Direction::from_i32(proto.direction).required("direction")?,
```
tustvold, line 48: Drive by fix to reduce the verbosity

predicate/Cargo.toml
```diff
@@ -9,7 +9,6 @@ chrono = "0.4"
 data_types = { path = "../data_types" }
 datafusion = { path = "../datafusion" }
 datafusion_util = { path = "../datafusion_util" }
-generated_types = { path = "../generated_types" }
```
tustvold, line 12: :tada:

generated_types/src/delete_predicate.rs
```diff
@@ -0,0 +1,161 @@
...
+            exprs: predicate.exprs.into_iter().map(Into::into).collect(),
+        }
+    }
+}
+
+impl TryFrom<proto::Predicate> for DeletePredicate {
+    type Error = FieldViolation;
+
+    fn try_from(value: Predicate) -> Result<Self, Self::Error> {
+        let range = value
```
tustvold, line 35: This pattern shows up a fair bit in this conversion code, I'll look to extract another helper extension trait in a follow up PR.

generated_types/src/delete_predicate.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    #[test]
+    fn test_roundtrip() {
+        let round_trip = |expr: DeleteExpr| {
+            let serialized: proto::Expr = expr.clone().into();
+            let deserialized: DeleteExpr = serialized.try_into().unwrap();
+            assert_eq!(expr, deserialized);
+        };
+
+        round_trip(DeleteExpr {
```
alamb, line 139: Why did you change this test to use the structs directly rather than `parse_delete_predicate` and a string? Is it because the parsing code lives in the `predicate` crate and this module can only use things from `data_types`?

alamb (approved):

generated_types/src/delete_predicate.rs
tustvold: I left that part of the test behind in `predicate`. The display test being in `predicate` is actually a little bit odd now that the type itself lives in data_types, I'll move it in a follow up.

PR #3109, opened 2021-11-15
refactor: improve usability of proto conversion traits

tustvold:
These extension traits grew somewhat organically as I implemented the gRPC API, and consequently are a little bit confused. This PR renames some things to make them clearer, and adds a load of documentation about how and why to use them, and also a bit about error handling in gRPC in general.


generated_types/src/google.rs
```diff
@@ -310,10 +359,15 @@ impl From<PreconditionViolation> for tonic::Status {
...
+///
+/// This is useful where a field has fallible `TryFrom` conversion logic, which doesn't
+/// return [`FieldViolation`]. This will format the returned error and add the field
+/// name as context
+///
 pub trait FieldViolationExt {
     type Output;
 
-    fn field(self, field: &'static str) -> Result<Self::Output, FieldViolation>;
+    fn scope(self, field: &'static str) -> Result<Self::Output, FieldViolation>;
```
tustvold, line 370: This makes it consistent with the name of the member function on `FieldViolation` to further scope an error

generated_types/src/google.rs
```diff
@@ -345,106 +399,125 @@ impl From<QuotaFailure> for tonic::Status {
...
+///
+/// * Within conversion code for a oneof enumeration
+/// * Converting from a scalar field to a custom Rust type
+///
+/// In a lot of cases, the type will be `Option<proto::Type>` or `Vec<proto::Type>`
+/// in which case `FromOptionalField` or `FromRepeatedField` should be used instead
+///
 pub(crate) trait FromField<T> {
-    fn scope(self, field: impl Into<String>) -> Result<T, FieldViolation>;
+    fn from_field(self, field: impl Into<String>) -> Result<T, FieldViolation>;
```
tustvold, line 414: I felt the `from_` prefix helped to identify that a type conversion is taking place

generated_types/src/google.rs
```diff
@@ -345,106 +399,127 @@ impl From<QuotaFailure> for tonic::Status {
...
         res.map_err(|e| e.scope(field))
     }
 }
+
+/// An extension trait that adds the method `non_empty` to any `String`
+///
+/// This is useful where code wishes to require a non-empty string is specified
+///
+/// TODO: Replace with NonEmptyString type implementing TryFrom?
+pub(crate) trait NonEmptyString {
```
tustvold, line 495: This is FromFieldString renamed and moved lower in the file

generated_types/src/google.rs
tustvold: Unfortunately it would appear clippy does not like it...

generated_types/src/google.rs
```diff
@@ -345,106 +399,127 @@ impl From<QuotaFailure> for tonic::Status {
...
+///
+/// This is primarily used to define other extension traits but may be useful for:
+///
+/// * Conversion code for a oneof enumeration
+/// * Converting from a scalar field to a custom Rust type
+///
+/// In a lot of cases, the type will be `Option<proto::Type>` or `Vec<proto::Type>`
+/// in which case `FromOptionalField` or `FromRepeatedField` should be used instead
+///
+pub trait FromField<T> {
```
tustvold, line 413: These traits are intentionally now made public, they were only crate local whilst being iterated on

generated_types/src/google.rs
```diff
@@ -1,5 +1,52 @@
...
+//! to abstract away concerns such as API evolution, default values, etc...
+//!
+//! Unfortunately, writing this translation code is somewhat cumbersome and so this module
+//! contains a collection of extension traits to improve the ergonomics of writing the potentially
+//! fallible conversion from the protobuf representation to the corresponding `data_types` type
+//!
+//! Each type should implement the following:
+//!
+//! * From<data_types::Type> for proto::Type
+//! * TryFrom<proto::Type, Error=FieldViolation> for data_types::Type
```
crepererum, line 38:
```suggestion
//! * `From<data_types::Type> for proto::Type`
//! * `TryFrom<proto::Type, Error=FieldViolation> for data_types::Type`
```

generated_types/src/google.rs
```diff
@@ -345,106 +399,127 @@ impl From<QuotaFailure> for tonic::Status {
...
 
         res.map_err(|e| e.scope(field))
     }
 }
+
+/// An extension trait that adds the method `non_empty` to any `String`
+///
+/// This is useful where code wishes to require a non-empty string is specified
+///
+/// TODO: Replace with NonEmptyString type implementing TryFrom?
```
crepererum, line 494: I think that's desirable, but can wait for another PR.

crepererum (approved):
Thanks, I personally also struggled a bit with the extension traits in the past. The new names are definitely better.

alamb (commented):
looks like a nice change to me

PR #3111, opened 2021-11-15
feat: Rename database disown/adopt to release/claim

### LINKED ISSUE #3110: Rename disown/adopt
> Disown and adopt feel like unnecessarily provocative terms. Can we use “release” and “claim”?

I agree and I'm happy to make this change if "release" and "claim" are the terms we'd rather use!

carols10cents:
Closes #3110.

Yay for better names!!


alamb (approved):
👍  looks good to me

PR #3113, opened 2021-11-15
fix: ensure `ConsistenHasher` is consistent

### LINKED ISSUE #3063: ConsistentHasher Hash Function Stability
**Problem**

The ConsistentHasher implementation makes use of [`std::collections::hash_map::DefaultHasher`](https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html) to compute hashes within the ring. Currently this is using SipHash 1-3, seeded with keys 0, 0, however, it carries this warning 

> The internal algorithm is not specified, and so it and its hashes should not be relied upon over releases. 

Changing the algorithm would effectively break sharding for a cluster, and so this is not ideal

**Proposal**

We should pin a specific hash function for use within `ConsistentHasher`. Unfortunately the SipHash implementation used by the stdlib is marked deprecated and cannot be used directly, but it has been re-exported as a [crate](https://crates.io/crates/siphasher).

Ahash, used by `hashbrown::HashMap` and in some other places, does not have a [stable output](https://github.com/tkaitchuck/ahash#goals-and-non-goals) and would not be suitable.

crepererum:
The std `DefaultHasher` is NOT guaranteed to stay the same, so let's
directly use the `SipHasher13` which at the moment (2021-11-15) is used
by the standard lib.

Fixes #3063.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -792,6 +792,7 @@ dependencies = [
  "ordered-float 2.8.0",
  "percent-encoding",
  "regex",
+ "siphasher",
  "snafu",
  "test_helpers",
  "time 0.1.0",
@@ -3973,6 +3974,12 @@ dependencies = [
  "num-bigint 0.2.6",
  "num-traits",
 ]
+
+[[package]]
+name = "siphasher"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "533494a8f9b724d33625ab53c6c4800f7cc445895924a8ef649222dcb76e938b"
 
 [[package]]
 name = "slab"
--- a/data_types/Cargo.toml
+++ b/data_types/Cargo.toml
@@ -13,6 +13,7 @@ observability_deps = { path = "../observability_deps" }
 ordered-float = "2"
 percent-encoding = "2.1.0"
 regex = "1.4"
+siphasher = "0.3"
 snafu = "0.6"
 time = { path = "../time" }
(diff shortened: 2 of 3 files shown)


Cargo.lock
```diff
@@ -3974,6 +3975,12 @@ dependencies = [
  "num-traits",
 ]
 
+[[package]]
+name = "siphasher"
```
alamb, line 3979: If we truly want to guarantee a stable hash algorithm, I think we should copy/paste the relevant code into the `influxdb_iox` repo so we have full control. Doing this effectively means we are relying on siphasher not t change that they do in some future release (e.g. 0.3.8 for example). 

I also always hope we can avoid more dependencies in our build / codebase which would be another reason to avoid this new dependency

alamb (approved):
I suggest inlining the `SipHasher13` implementation rather than using a crate, but I think this PR as written is also an improvement over main 👍

Cargo.lock
crepererum: This crate doesn't implement some hash, but a well defined one:

https://www.aumasson.jp/siphash/siphash.pdf

So breaking changes within this crate shouldn't affect the output (also, we have somewhat a regression test in this module) but are more likely "bump minimum Rust version for performance / feature things".

alamb: I was more thinking like "oh we fixed some corner case where our implementation was not consistent with the standard / other implementations in XYZ way"

Which could reasonably be treated as a bugfix / minor change but could still break IOx b/c `hash(input)` may not be the same anymore

Issue #3020, opened 2021-11-03
clean up IOxd interface


crepererum:
Follow-up tasks after #3016 is merged:

* It would be good to revisit the background workers for ServerState
* It would be good if ApplicationState could somehow be shared with routers

_Originally posted by @tustvold in https://github.com/influxdata/influxdb_iox/pull/3016#pullrequestreview-796580395_

PR #3115, opened 2021-11-15
refactor: clarify `ServerType` background worker handling

crepererum:
Ref #3020.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
@@ -39,8 +39,10 @@ impl From<tonic::transport::Error> for RpcError {
 pub trait ServerType: std::fmt::Debug + Send + Sync + 'static {
     type RouteError: HttpApiErrorSource;
 
+    /// Metric registry associated with the server.
     fn metric_registry(&self) -> Arc<Registry>;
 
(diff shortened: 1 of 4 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/mod.rs
```diff
@@ -51,9 +53,14 @@ pub trait ServerType: std::fmt::Debug + Send + Sync + 'static {
         req: Request<Body>,
     ) -> Result<Response<Body>, Self::RouteError>;
 
+    /// Construct and serve gRPC subsystem.
     async fn server_grpc(self: Arc<Self>, builder_input: RpcBuilderInput) -> Result<(), RpcError>;
 
-    async fn background_worker(self: Arc<Self>);
+    /// Join shutdown worker.
+    ///
+    /// This MUST NOT before `shutdown` is called, otherwise the server is deemed to be dead and the process will exit.
```
tustvold, line 61:
```suggestion
    /// This MUST NOT exit before `shutdown` is called, otherwise the server is deemed to be dead and the process will exit.
```

tustvold (approved):
I still find it funky that `join` is driving the shutdown behaviour, but I see that this is a consequence of the way `IOxExecutionContext` is written... I will look into at some point :grin:

PR #3116, opened 2021-11-15
Ntran/compact os chunks

### LINKED ISSUE #3089: Compact specified persisted chunks
__Use case:__
Currently, our persisted chunks may be small, overlapped with other chunks (in terms of timestamp and/or duplicates), and include soft deleted data which highly lead to suboptimal query performance, less compression, and more complicated data migration. We need a way to compact contiguous chunks, overlapped chunks and chunks with deletes.

__Proposal:__
After considering many things in this [Object Store Design](https://docs.google.com/document/d/1Rl5IYJA6e1hhH49xsNFMJPkDiqo9Buqk1wxgZS035B8/edit#), we decide to start with the simplest thing first: implement a function to compact given chunks.

This function will error if either
  1. The given chunks are not all persisted yet
  2. The given chunks are not contiguous

Steps: 
  1 . The chunks will be scan to deduplicate and hard delete data if needed.
  2 . The result will be written into a new persisted chunk.
  3. If the given persisted chunks have RUBs, unload those RUBs.
  4. Mark given chunks no longer needed to get dropped in the background later
  
__Current behaviour:__
We cannot compact persisted chunks

__Desired behaviour:__
Users can manually compact specified chunks

NGA-TRAN:
Closes #3089


server/src/db/lifecycle/persist.rs
```diff
@@ -571,4 +780,80 @@ mod tests {
         .unwrap();
         check_closure(&catalog);
     }
+
+    #[tokio::test]
+    async fn test_compact_os_negative() {
```
NGA-TRAN, line 785: @crepererum : this PR is not ready for review but I would like to go over this test to ask you about the use of flush handle and the time.inc function

server/src/db/lifecycle/persist.rs
```diff
@@ -223,6 +230,213 @@ pub fn persist_chunks(
...
+///    . provided chunk(s) not belong to the provided partittion
+///    . not all provided chunks are persisted
+///    . the provided chunks are not contiguous
+///
+// Steps:
+// 1 . The chunks will be scan to deduplicate and hard delete data if needed.
+// 2 . The result will be written into a new persisted chunk.
+// 3. If the given persisted chunks have RUBs, unload those RUBs.
+// 4. Mark given chunks no longer needed to get dropped in the background later
+pub(crate) fn compact_object_store_chunks(
```
NGA-TRAN, line 247: At first I added this function in compact.rs file but I soon discovered that this function is closer to persist_chunks so I decided to move it here

NGA-TRAN: One more comment here is this function is very long and very similar to `compact_chunks` and `persist_chunks` but they have a bit of different here and there. We might want to see if it is worth to refactor them.

server/src/db/lifecycle/persist.rs
```diff
@@ -571,4 +785,312 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_compact_os_negative() {
+        // Tests that nothing will get compacted
+
+        test_helpers::maybe_start_logging();
+
+        let (db, time) = test_db().await;
+        let late_arrival = Duration::from_secs(1);
+        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=1 10").await;
+        // todo: ask Marco why we need this time.inc function? it closes what?
```
NGA-TRAN, line 798: @crepererum : I do not understand well the `time_provider` and `flush_handle`. A short chat will help.

server/src/db/lifecycle/persist.rs
```diff
@@ -571,4 +785,312 @@ mod tests {
...
+        write_lp(db.as_ref(), "cpu,tag1=chunk3,tag2=a bar=2 30").await;
+        db.persist_partition("cpu", partition_keys[0].as_str(), true)
+            .await
+            .unwrap()
+            .unwrap()
+            .id();
+        //
+        // Add a MUB
+        write_lp(db.as_ref(), "cpu,tag1=chunk4,tag2=a bar=2 40").await;
+        // todo: Need to ask Marco why there is no handle created here
```
NGA-TRAN, line 854: @crepererum : similar question here but a chat is better

server/src/db/lifecycle/persist.rs
```diff
@@ -571,4 +785,312 @@ mod tests {
         .unwrap();
         check_closure(&catalog);
     }
+
+    #[tokio::test]
```
NGA-TRAN, line 789: The 3 tests below are quite long. I hope reviewers can point out a way to shorten them

generated_types/protos/influxdata/iox/management/v1/jobs.proto
```diff
@@ -91,6 +92,22 @@ message CompactChunks {
   repeated bytes chunks = 5;
 }
 
+// Compact chunks into a single chunk
+message CompactObjectStoreChunks {
```
NGA-TRAN, line 96: I will create a ticket to finish the gRPC API to invoke compact_object_store_chunks.

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+        let persisted_rows = persisting_chunk.as_ref().map(|p| p.rows()).unwrap_or(0);
+
+        let persist_fut = {
+            let partition = LockableCatalogPartition::new(Arc::clone(&db), partition);
+
+            // drop compacted chunks
+            let mut partition_write = partition.write();
+            let mut delete_predicates_after = HashSet::new();
+            for id in &chunk_ids {
+                let chunk = partition_write.force_drop_chunk(*id).expect(
```
tustvold, line 364: I think this just removes the chunk from the in-memory catalog, I think the preserved catalog will still contain this chunk

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+///
+// Steps:
+// 1 . The chunks will be scan to deduplicate and hard delete data if needed.
+// 2 . The result will be written into a new persisted chunk.
+// 3. If the given persisted chunks have RUBs, unload those RUBs.
+// 4. Mark given chunks no longer needed to get dropped in the background later
+pub(crate) fn compact_object_store_chunks(
+    partition: LifecycleWriteGuard<'_, Partition, LockableCatalogPartition>,
+    chunks: Vec<LifecycleWriteGuard<'_, CatalogChunk, LockableCatalogChunk>>,
+    flush_handle: FlushHandle,
```
tustvold, line 249: A FlushHandle encodes the notion of persisting unpersisted data, which this isn't doing. I suspect this will therefore be generating invalid PartitionCheckpoint, etc... for data that this is not actually persisting

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+            let persisting_chunk = LockableCatalogChunk {
+                db,
+                chunk: Arc::clone(new_chunk),
+                id: new_chunk_id,
+                order: min_order,
+            };
+
+            // Now write the chunk to object store
+            let persisting_chunk = persisting_chunk.write();
+            write_chunk_to_object_store(partition_write, persisting_chunk, flush_handle)?.1
```
tustvold, line 417: This lifecycle action is very tightly coupled with what it means to persist previously unpersisted data. I'm fairly certain using this is incorrect.

I would instead expect this lifecycle action to create a new catalog transaction that drops the old chunks from the preserved catalog, and adds the new chunk (if any).

server/src/db/lifecycle/persist.rs
```diff
@@ -572,4 +786,312 @@ mod tests {
...
+        assert_eq!(summary_chunks[0].storage, ChunkStorage::OpenMutableBuffer);
+        assert_eq!(summary_chunks[0].row_count, 1);
+        // the result of compacting 2 persisted chunks
+        assert_eq!(
+            summary_chunks[1].storage,
+            ChunkStorage::ReadBufferAndObjectStore
+        );
+        // 2 rows left because 4 duplicated & deleted rows have been removed
+        assert_eq!(summary_chunks[1].row_count, 2);
+    }
```
tustvold, line 1096: I think we should have a test that reloads the catalog from object storage afterwards, to ensure that things have been correctly updated there also. Perhaps the ReplayTest machinery could be extended to support this?

tustvold (changes requested):
I think this cannot simply reuse the persist machinery that exists for persisting unpersisted data, as it stands I think as written this will flush data from the PersistenceWindows and generate PartitionCheckpoints as if it were a normal persist operation, but without actually persisting that data.

I think some tests of recovering the resulting catalog would show the issue, we likely need to write some ReplayTest cases for this operation

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+            // This function will also error out if the chunk is not yet persisted
+            chunk.set_compacting_object_store(&registration)?;
+            Ok(DbChunk::snapshot(&*chunk))
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    // Verify if all the provided chunks are contiguous
+    // TODO: ask Raphael if this should be done after dropping partition lock. It seems this check is better here
+    //       to avoid a situation a newly persisted chunk added after dropping lock
+    if !partition.contiguous_object_store_chunks(&chunk_orders) {
```
alamb, line 318: 👍

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+
+        // Compacting query plan
+        let (schema, plan) =
+            ReorgPlanner::new().compact_plan(schema, query_chunks.iter().map(Arc::clone), key)?;
+        let physical_plan = ctx.prepare_plan(&plan).await?;
+        // run the plan
+        let stream = ctx.execute_stream(physical_plan).await?;
+
+        // create a read buffer chunk for persisting
+        // todo: after the code works well, this will be improved to pass this RUB pass
```
alamb, line 351: I would expect this to write directly to object store (not to RUB)

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+        let persisting_chunk = collect_rub(stream, &addr, metric_registry.as_ref()).await?;
```
alamb, line 352: I think renaming this to something like `compacting_chunk` would make the code easier to read / reason about

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,213 @@ pub fn persist_chunks(
...
+                        .persistence_windows_mut()
+                        .unwrap()
+                        .flush(flush_handle);
+                    return Ok(None);
+                }
+            };
+
+            // Create a RUB chunk before persisting
+            // This RUB will be unloaded as needed during the lifecycle's maybe_free_memory
+            // todo: this step will be improved to avoid creating RUB if needed
```
alamb, line 398: The RUB in my mind is all about fast in memory query performance; I don't understand why we would write to the RUB at all in a compact plan

I think we should just write the compacted chunk directly into Object Store and bypass the RUB entirely -- otherwise we risk both:
1. Blowing out memory with a huge chunk
2. evicting some more recent data that is being queried to hold data that won't be queried.

alamb (commented):
I defer to @tustvold  on the coding details of the lifecycle action, but the interface for the `CompactObjectStoreChunks` action looks good 👍

lifecycle/src/policy.rs
```diff
@@ -908,6 +909,20 @@ mod tests {
             Ok(db.registry.lock().complete(()))
         }
 
+        fn compact_object_store_chunks(
```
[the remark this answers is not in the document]
NGA-TRAN: This is needed for Testing. The Test implements LockablePartition that includes this function

server/src/db/lifecycle/persist.rs
NGA-TRAN: Yeah, I can move this to its own module, if after our chat, we think it is better

server/src/db/lifecycle/persist.rs
NGA-TRAN: I am looking for answer of what FlushHandle does here, too. Sounds like we should not use it

server/src/db/lifecycle/persist.rs
NGA-TRAN: This is a good idea. Let us chat and see how to do this.

server/src/db/lifecycle/persist.rs
NGA-TRAN: Yes, we definitely needs that test. 
And I think we also need an integration test at a  higher level that uses gRPC to compact os chunks and then use SQL (or gRPC API) to query data again. This will be done after I add the gRPC API

server/src/db/lifecycle/persist.rs
```diff
@@ -222,6 +229,228 @@ pub fn persist_chunks(
...
+        for id in &chunk_ids {
+            // todo: return context for error
+            let _result = db
+                .drop_chunk(
+                    &*partition_addr.table_name,
+                    &*partition_addr.partition_key,
+                    *id,
+                )
+                .await;
+        }
```
NGA-TRAN, line 436: @tustvold : from your comment that this code does not drop chunk from `preserved catalog`, I have explored I add this `drop_chunk` from preserved catalog here. I know:

- @crepererum may let us know this is not necessary if it is done in the background later (not sure)
- And that this function needs to rewrite to use new transaction and avoid flush handle but I just spend sometime mostly to learn the code

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,681 @@
...
+        // Step 3: persist the newly created chunk (data)
+        // todo
+
+        // =============================================================
+        // Step 4: switch to newly compacted-and-persisted chunk
+        // and drop old chunks in one transaction
+        // todo
+
+        // the below code will be replaced with 2 functions, one for step 3 and one for step4
+
```
NGA-TRAN, line 117: @tustvold and @crepererum : 
I have moved compact os work to its own module and split the long function into smaller ones. The Steps 3 and 4 will be done after our discussion. Here are my questions (you will probably provide more insights besides these questions):
1. Flush Handle & time_provider
2. We do not want to create a RUB for this but, so far, have we tried to create an OS  chunk without a RUB one? If we have not, should we go through RUB first and then improve it in the following PRs?
3. Should we drop chunks from both non-preserved and the preserved catalogs? Or the chunks will be automatically dropped from preserved catalog in the background in a future lifecycle?
4. How to define a new transaction?

crepererum: Some loose thoughts (we have a meeting later, but I already wanted to write down a bit of context):

> 1. Flush Handle & time_provider

No flush handle required because you don't need to flush any unpersisted data. You'll need a time provider though for the transaction (the parquet file contains a creation time stamp). The provider is used to allow you to mock this time stamp for tests.

> 2. We do not want to create a RUB for this but, so far, have we tried to create an OS  chunk without a RUB one? If we have not, should we go through RUB first and then improve it in the following PRs?

It should be possible to create a OS chunk directly. We do that when reading from the persisted catalog during startup:

https://github.com/influxdata/influxdb_iox/blob/76790cadd8e2a13f6676da4bdc4ca4a03d01effd/server/src/db/load.rs#L249-L256

> 3. Should we drop chunks from both non-preserved and the preserved catalogs? Or the chunks will be automatically dropped from preserved catalog in the background in a future lifecycle?

Yes, you need to drop it from both and the preserved drop MUST be done BEFORE the in-mem drop because its IO can fail (in which case you MUST NOT add any new chunks or delete the old ones from the in-mem catalog). Remember to also add the new chunk to the persisted catalog (within the same transaction).

> 4. How to define a new transaction?

https://github.com/influxdata/influxdb_iox/blob/76790cadd8e2a13f6676da4bdc4ca4a03d01effd/server/src/db/lifecycle/write.rs#L162-L186

There's also `remove_parquet` to drop the existing chunks.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,681 @@
...
+        // the result of compacting 2 persisted chunks
+        assert_eq!(
+            summary_chunks[1].storage,
+            ChunkStorage::ReadBufferAndObjectStore
+        );
+        // 2 rows left because 4 duplicated & deleted rows have been removed
+        assert_eq!(summary_chunks[1].row_count, 2);
+    }
+
+    // todo: add more tests
```
crepererum, line 680: We also need a test (or extend an existing one) that the compaction "survives" a DB restart (aka that the add and remove operations are correctly persisted). You might wanna have a look at

https://github.com/influxdata/influxdb_iox/blob/76790cadd8e2a13f6676da4bdc4ca4a03d01effd/server/src/db/lifecycle/persist.rs#L468

crepererum: I also would like to see a test that delete predicates that are added while the compaction is running are correctly handled. You might also wanna have a look at the same test (`test_delete_predicate_propagation`) on how to do that (it requires some future-trickery, but it's not too hard).

NGA-TRAN:
After talking with @crepererum and @tustvold and identifying the tasks of this function, I will keep this PR for my reference only and start working on other PRs. This functions needs many PRs until we mark it done

e-dard:
Hey @NGA-TRAN, I wonder if we should close this PR (the branch will still be available).

NGA-TRAN:
Sure @e-dard . I will remove this branch too

Issue #3095, opened 2021-11-12
Accept and store context for disown/adopt transaction records


carols10cents:
This was part of the [original floating databases design](https://github.com/influxdata/influxdb_iox/issues/2672) that ended up being somewhat complex, so we've decided to defer its implementation. 

The idea was that the CLI and gRPC APIs for disown and adopt should accept optional context to be recorded in the transaction history of the database. The type of context was to be:

```
repeated google.protobuf.Any context
```

Preserving relevant discussion from https://github.com/influxdata/influxdb_iox/pull/3038, particularly [this thread](https://github.com/influxdata/influxdb_iox/pull/3038#discussion_r747825338):

Andrew:

> I am interested to know the rationale for storing the extra metadata as protobuf is (and why not as a opaque string, for example)
> 
> The upsides of using protobuf is that perhaps IOx can look at this information and interpret it. The downside of using protobuf is that it is harder to use for some arbitrary payload I think (as you need to encode it using gRPC). 
> 
> If the idea is that the context is opaque to IOx (and not parsed/ interpreted by it) I think it would be a better user experience to simply store this context as a string (let users base64 encde stuff if they need binary).
> 
> In fact it seems that is exactly what the cli test cases do (store some string). Maybe the code would be simpler if this field's type is just `string`?

MKM: 

> 
> Imagine you don't want to use an `Any` and store an arbitrary payload. How would you do it? 
> 
> A `string` field can't contain any arbitrary encoding. You'd have to base64 the content if you have any non textual content. But that's fine, let's use a `bytes` then.
> 
> Ok, now you have a field that can contain whatever arbitrary payload you the client wants to put in there.
> 
> A few interesting questions arise:
> 
> 1. But what is the type of this payload? The client who wrote it clearly knows what it wrote, but what about the new version of the client? Surely it can be up to the client to use some self describing file format that will tell which format/version it is.
> 2. What if you have multiple "clients" (or multiple layers of clients) that each wants to put some additional metadata into this; how would you tell them apart? A classic solution is to have some sort of key/name/... field associated with each blob (e.g. a string->blob (multi)map) so each client can quickly find it's stuff.
> 
> You could even combine (1) and (2) and add a "type" field that plays both roles: it tells clients (and humans!) what is the type of the payload and also lets clients (and humans!) associate the payloads with the clients and/or roles that such payloads play.
> 
> So basically we could just have a message like:
> 
> ```proto
> message TypedBlob {
>   string type = 1;
[...]

PR #3118, opened 2021-11-15
fix: Make delete/restore aliases for release/claim; remove tombstone

### LINKED ISSUE #2680: Floating databases: Replace the tombstone file for database delete/restore with the release/claim API

carols10cents:
Fixes #2680

This change makes the "delete" CLI/gRPC API an alias for "release" and the "restore" CLI/gRPC API an alias for "claim".

The original design that involved storing context had delete/restore creating and inserting special context that release/claim would not, but [we've deferred context entirely](https://github.com/influxdata/influxdb_iox/issues/3095), so now there's even less difference.

The only teeny tiny difference is that the "release" API optionally takes a UUID to check before releasing to make absolute sure we have the database we think we do, while currently "delete" doesn't have the ability to take an optional UUID at the moment (but it would be easy to add if that's desired).

You should now be able to mix and match these sets of commands, that is, in addition to:

- `delete database` then `restore database`
- `release database` then `claim database`

you should also be able to do:

- `delete database` then `claim database`
- `release database` then `restore database`


alamb (approved):
I feel like we might even want to remove the "delete" and "restore" commands entirely (so that we don't have two sets of commands that are interchangeable).

I realize this is more of an API design question for @pauldix  / @jacobmarble / @mkmik  but I think it is worth reviewing

alamb (approved):
🏅  -- very nice

jacobmarble:
(I haven't been following along closely.)

Seems like "delete" and "release" would have different semantics? If I issue a "delete database" I expect to never see that database again. If I "release database" then I expect to go to the docs before I touch that with a ten-foot pole, then understand it, then issue it, knowing my data isn't going to be deleted.

Do I have this wrong?

PR #3121, opened 2021-11-16
refactor: DMLOperation write path (#2731)

tustvold:
I initially planned to have separate member functions for writes and deletes, but whilst implementing this I realised I was factoring out lots of common logic. Additionally this was made more complex because you can't construct `&DmlOperation` from `&DmlWrite` or `&DmlDelete`.

This PR therefore changes the write path to use `DmlOperation`, it's a little bit confused because we currently have two parrallel routing implementations but hopefully we can keep iterating and eventually converge on something good :sweat_smile:


write_buffer/src/core.rs
```diff
@@ -21,15 +21,15 @@ pub trait WriteBufferWriting: Sync + Send + Debug + 'static {
...
     fn sequencer_ids(&self) -> BTreeSet<u32>;
 
-    /// Send a [`DmlWrite`] to the write buffer using the specified sequencer ID.
+    /// Send a [`DmlOperation`] to the write buffer using the specified sequencer ID.
     ///
     /// The [`dml::DmlMeta`] will be propagated where applicable
     ///
     /// Returns the metadata that was written
-    async fn store_write(
+    async fn store_operation(
```
tustvold, line 29: I wasn't sure whether to call this `store_operation` or just `store`

router/src/grpc_client.rs
```diff
@@ -17,7 +17,7 @@ pub type WriteError = Box<dyn std::error::Error + Send + Sync>;
 #[async_trait]
 pub trait GrpcClient: Sync + Send + std::fmt::Debug + 'static {
     /// Write data to the given database.
-    async fn write(&self, db_name: &str, write: &DmlWrite) -> Result<(), WriteError>;
+    async fn write(&self, db_name: &str, write: &DmlOperation) -> Result<(), WriteError>;
```
tustvold, line 20: I couldn't think of a less confusing name for this function

crepererum: I think "write" isn't to bad since it's similar to a file handle. "Send" would be another option.

The docstring should probably be updated, it's more a "send DML operation to the given database".

write_buffer/src/core.rs
crepererum: `store_operation` seems fine

crepererum (approved):
Thanks.

Issue #2752, opened 2021-10-06
Add histogram metrics for ingestion


mkmik:
Now that we can use custom histogram bucket sizes (thanks to #2365!), it's finally possible to emit a metric that measures the distribution of sizes of LP batches and/or eentries we write to / read from the write buffer (e.g. kafka)

- [x] `ingest_batch_size_bytes`
- [ ] `write_buffer_ingest_entry_size`
- [ ] (bonus points) compressed sizes for write buffer entries (may be hard since the compression is done transparently by the Kafka client library)

CC: @primeroz

alamb:
I am going to take a shot at adding the histogram for kafka ingestion sizes

PR #3120, opened 2021-11-16
feat: emit Kafka stats as metrics instead of logs

crepererum:
This maps a subset of Kafka stats as metrics. The set can -- of course
-- be changed in the future depending on our needs.

Fixes #3100.

PR #3129, opened 2021-11-16
feat: Add kafka write buffer consumer metrics

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/2752

Takes advantage of the great work in https://github.com/influxdata/influxdb_iox/pull/3120 from @crepererum

# Rationale
1. To tune the Kafka configuration, we need to understand the distribution of sizes of entries we read from the kafka

# Changes
1. Add `write_buffer_ingest_entry_size` metric
2. Tests for same


write_buffer/src/kafka.rs
```diff
@@ -205,18 +214,30 @@ impl WriteBufferReading for KafkaBufferConsumer {
...
 
-                    let kafka_headers = message.headers().into_iter().flat_map(|headers| (0..headers.count()).map(|idx| headers.get(idx).unwrap()));
+                    let kafka_headers = header_iter(&message);
                     let headers = IoxHeaders::from_headers(kafka_headers, trace_collector.as_ref())?;
-
                     let payload = message.payload().ok_or_else::<WriteBufferError, _>(|| {
                         "Payload missing".to_string().into()
                     })?;
 
+                    // Estimate size of data read from kafka as
```
alamb, line 233: I could not find any way in  `rdkafka` to get the actual message size as it was received (e.g if it was compressed) nor the actual in memory message size.

@mkmik  I am not sure if you would like this metric to include headers and the key size, or if you would like it to just include the payload

write_buffer/src/kafka.rs
```diff
@@ -385,11 +406,31 @@ impl KafkaBufferConsumer {
             .collect::<Result<BTreeMap<u32, Arc<StreamConsumer<ClientContextImpl>>>, KafkaError>>(
             )?;
 
+        let write_buffer_ingest_entry_size: Metric<U64Histogram> = metric_registry
+            .register_metric_with_options(
+                "write_buffer_ingest_entry_size",
+                "distribution of ingested Kafka message sizes",
+                || {
+                    U64HistogramOptions::new([
+                        1024,
```
alamb, line 415: @mkmik / @primeroz  Are these ok histogram boundaries for kafka message ingest size (aka the size of messages read from kafka)?

write_buffer/src/kafka.rs
mkmik: the payload sizes are usually big enough for this to usually not matter much

write_buffer/src/kafka.rs
mkmik: in #2969 I also made sure to include messages larger than the megabyte (because we observed such batch sizes in tools)

alamb: I am happy to use the same sizes as in https://github.com/influxdata/influxdb_iox/pull/2969; I reduced the bucket sizes in this PR as I recalled @primeroz said we had a 1MB limit on kafka message size. Please let me know what you would like and I will make this PR do it

primeroz: just to be fair i _suggest_ we work on the [default expected max message size of 1 MebiByte ](https://kafka.apache.org/documentation/#brokerconfigs_message.max.bytes) if possible but with extra work , and especially with a distribution of sizes that limit the number of large messages , we could send bigger messages. 

If we can stay in the limit of the default size it will make everything easier especially as we scale up number of topics

mkmik: I agree we should limit to 1MB.

However if for any reason we end up tweaking that parameter, we wouldn't have observability that help us figure out which databases are payload . Changing the buckets here requires changing the code.

Perhaps we could have less fine grained by buckets above the 1MB mark?

primeroz: yeah i tihnk we should definetely expose some bigger buckets like `3MB, 5MB, 10MB, Infinity`... and maybe reduce the buckets in the <1MB ... especially if those are per `iox database` to limit crdinality ?

mkmik: yeah, honestly I don't know if we can keep putting the `iox database` as a label there.
At least not until we have iox itself processing the telemetry.

Our /metrics is already exploding in size right now. We're going to have 50k databases in a single region

primeroz: :+1: i do think tht metrics like this don't need the precision of per-database cardinality ... i would love a single per iox-router metric with the histogram of write sizes

mkmik: yeah, the per-database labels would be useful in order to spot the outliers. But the are other techniques (sadly not supported by our monitoring system) which work much much better at doing that: `exemplars`

write_buffer/src/kafka.rs
```diff
@@ -205,18 +214,30 @@ impl WriteBufferReading for KafkaBufferConsumer {
...
                     let headers = IoxHeaders::from_headers(kafka_headers, trace_collector.as_ref())?;
-
                     let payload = message.payload().ok_or_else::<WriteBufferError, _>(|| {
                         "Payload missing".to_string().into()
                     })?;
 
+                    // Estimate size of data read from kafka as
+                    // payload len + key len + headers
+                    let kafka_read_size: usize = payload.len() +
+                        message.key().map(|key| key.len()).unwrap_or(0) +
```
e-dard, line 236: Haven't tested this but I think you can do. YOLO!!

```suggestion
                        message.key().map(|key| key.len()).unwrap_or_default() +
```

write_buffer/src/kafka.rs
e-dard: Prometheus will be broken for more than tens of tenants on IOx. I don't think it's critical to know the distribution of writes into kafka on a per-tenant basis. I suspect we could infer that from other places like the front door.

e-dard: I _think_ in TSM Kafka we tend to have message sizes around `1.8MB`.

e-dard (approved):
Nice 👍

write_buffer/src/kafka.rs
alamb: Ok, so now I am pretty confused 🤔 

What boundaries do you want on these histograms? How about something like

```
[1K, 16K, 256K, 1MB, 3MB, 10MB, Infinity]?
```

primeroz: yeah , for me that is good set of buckets . In case i would just add one in between 256 and 1MB , maybe `768`

mkmik: LGTM

alamb: Updated boundaries in 9abdf540

write_buffer/src/kafka.rs
alamb: In 18f5947a

Issue #1953, opened 2021-07-12
Panics captured by logging layer lose message


mkmik:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behaviour.

1. Cause a panic, e.g. noticed when sending very large http requests to `/api/v2/write` 

__Expected behaviour:__

```
level=error msg="thread 'tokio-runtime-worker' panicked at 'Should have been able to read the next chunk: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody })', src/influxdb_ioxd/http.rs:427:27" 
```

__Actual behaviour:__

```
level=error msg="thread 'tokio-runtime-worker' panicked at 'UNKNOWN', src/influxdb_ioxd/http.rs:427:27" span=1657324937750249551 target="panic_logging" location="panic_logging/src/lib.rs:110" time=1626093262739824399
```

__Logs__

The stderr stream shows the original message:

```
thread 'tokio-runtime-worker' panicked at 'Should have been able to read the next chunk: hyper::Error(Body, Custom { kind: UnexpectedEof, error: IncompleteBody })', src/influxdb_ioxd/http.rs:427:27
```

alamb:
Interestingly, there is a test for catching the context here https://github.com/influxdata/influxdb_iox/blob/99093b18fb7f60b0229705c5fce79c42f86657a5/src/influxdb_ioxd/rpc/storage/service.rs#L1892-L1896

I tried locally (on a mac) and I got messages both in debug and release mode

mkmik:
To make the thing even more mysterious, the logs contain some examples where the message is correctly preserved in the logfmt body; so whatever is going on, it happens only in some cases (the best kind of bugs)

carols10cents:
Preserving what I put in chat: one possible cause of this is calling `panic!` with something other than a `String` or `&str`. It doesn't look like we're doing this in our code.

`cargo build -vvv` shows warnings from your dependencies, so I ran `RUSTFLAGS="-W non_fmt_panic" cargo build -vvv`.

This spot in the `proc-macro-error` crate is the only candidate I found: https://gitlab.com/CreepySkeleton/proc-macro-error/-/blob/master/src/lib.rs#L472

The `proc-macro-error::AbortNow` type is private, so we can't just try downcasting to that, but we could fork `proc-macro-error` and change that call to `panic` and see if that's it?

alamb:
[...]

PR #3130, opened 2021-11-16
fix: capture all panic messages in logs

alamb:
Resolves https://github.com/influxdata/influxdb_iox/issues/1953

# Rationale
1. `panic` messages that are not `&str` or `String` we printed as `UNKNOWN`

# Changes
1. Use std::fmt::Display for `panic_info` rather than trying to replicate it ourself


panic_logging/src/lib.rs
```diff
@@ -91,35 +91,7 @@ impl Drop for SendPanicsToTracing {
...
-    };
-
-    if let Some(s) = panic_info.payload().downcast_ref::<&str>() {
-        log_panic(s);
-    } else if let Some(s) = panic_info.payload().downcast_ref::<&String>() {
-        log_panic(s);
-    } else {
-        log_panic("UNKNOWN");
-    }
+    error!(panic_info=%panic_info, "Thread panic");
```
alamb, line 94: I am pretty sure I didn't use the `std::fmt::Display` implementation of `PanicInfo` was because I didn't know any better initially

panic_logging/src/lib.rs
```diff
@@ -91,35 +91,7 @@ impl Drop for SendPanicsToTracing {
...
-            Some(name) => name.into(),
-            None => "unnamed".into(),
-        };
-        error!(
-            "thread '{}' panicked at '{}', {}",
-            thread_name, s, location_string
-        );
-    };
-
-    if let Some(s) = panic_info.payload().downcast_ref::<&str>() {
```
alamb, line 116: The idea is to let `PanicInfo` handle this correctly -- see https://doc.rust-lang.org/src/core/panic/panic_info.rs.html#130-145

domodwyer (approved):
Looks good! I'll do the same in Conductor 👍

panic_logging/src/lib.rs
NGA-TRAN: Hah, simple

NGA-TRAN (approved):

alamb:
> Hah, simple

Yeah   -- seems so simple now 🤦

mkmik:
Thanks!

Issue #3029, opened 2021-11-04
IOx management `CreateDatabase`  timesout prior to database creation if write buffer connection can not be established in time


alamb:
As pointed out by @domodwyer  

__Steps to reproduce:__

Try and create a database with an invalid write buffer connection (for example, with nothing listening on port 9092):

```json
{
  "rules": {
    "name": "my_db_connection_string",
    "writeBufferConnection" : {
      "direction":  1,
      "type": "kafka",
      "connection": "localhost:9092"
    }
  }
}
```

Put this content into the file `/tmp/new_db.json` and then try and create a datatabase

```shell
grpcurl --plaintext -d "`cat /tmp/new_db.json`" localhost:8082 influxdata.iox.management.v1.ManagementService.CreateDatabase
```

Locally on my machine, this api will timeout after 1 minute (probably the `grpcurl` timeout). Conductor times out after 30 seconds:

```shell

alamb@MacBook-Pro k8s-idpe % time grpcurl --plaintext -d "`cat /tmp/new_db.json`" localhost:8082 influxdata.iox.management.v1.ManagementService.CreateDatabase
<" localhost:8082 influxdata.iox.management.v1.ManagementService.CreateDatabase
ERROR:
  Code: InvalidArgument
  Message: Cannot initialize database: error creating write buffer: Meta data fetch error: BrokerTransportFailure (Local: Broker transport failure)
grpcurl --plaintext -d "`cat /tmp/new_db.json`" localhost:8082   0.02s user 0.01s system 0% cpu 1:00.05 total
alamb@MacBook-Pro k8s-idpe %
```

Note that the database *was* created (but can't start due to the non existent write buffer):

```
grpcurl --plaintext -d '{}' localhost:8082 influxdata.iox.management.v1.ManagementService.GetServerStatus
{
  "serverStatus": {
    "initialized": true,
    "databaseStatuses": [
      ...
      {
        "dbName": "my_db_connection_string",
        "error": {
          "message": "error creating write buffer: Meta data fetch error: BrokerTransportFailure (Local: Broker transport failure)"
        },
        "state": "DATABASE_STATE_CATALOG_LOAD_ERROR"
      },
      ...
    ]
  }
}
```

__Expected behaviour:__
I expect the `CreateDatabase` API to return a message within 30 seconds (hopefully much sooner) with either an error or a message that the database was created but is still initializing

__Actual behaviour:__
gRPC client times out waiting for a response

__Environment info:__

[...]

PR #3131, opened 2021-11-16
fix: Default kafka timeout to be shorter than gRPC timeout (60 sec --> 10 sec)

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3029

# Rationale
  
The hard coded timeouts for some kafka operations were longer than gRPC timeouts, resulting in the gRPC request to IOx timing out when the kafka configuration wasn't working.

# Changes
1. Make timeout a symbolic constant
2. Change value from 60 seconds to 10 seconds


write_buffer/src/kafka.rs
```diff
@@ -404,7 +414,10 @@ async fn get_partitions(
     let metadata = tokio::task::spawn_blocking(move || {
         let probe_consumer: BaseConsumer = cfg.create()?;
 
-        probe_consumer.fetch_metadata(Some(&database_name), Duration::from_secs(60))
+        probe_consumer.fetch_metadata(
+            Some(&database_name),
+            Duration::from_millis(KAFKA_OPERATION_TIMEOUT_MS),
```
alamb, line 419: this happens to be the call that fails in `create_database` but if talking to kafka is taking more than 10 seconds I think we are likely going to have problems so I changed all the hard coded constants

domodwyer (approved):
Thanks - this will make things much nicer 👍

crepererum (approved):

Issue #3122, opened 2021-11-16
Linux-based ARM64 CI


crepererum:
__Use case:__
1. Test that IOx runs under ARM64
2. Build docker image (#3119)

__Proposal:__
Setup a Linux-based ARM64 CircleCI runner that runs `test` and `build`. The remaining jobs should be covered well enough by x64 (`test_perf`, `test_heappy`, `perf_image`) or only require a single arch (`lint`, `fmt`, `cargo_audit`, `doc`, `protobuf-lint`).

The `ci_image` job also needs to be changed to build images for both architectures.

__Current behaviour:__
We only test under x64.

__Desired behaviour:__
We also test and build ARM64.

__Alternatives considered:__
Not doing it.

crepererum:
@jacobmarble IMHO this is still an open issue, currently this is blocked by circleci that lacks sufficient ARM64 support (they don't support all the features we need).

jacobmarble:
> circleci that lacks sufficient ARM64 support (they don't support all the features we need).

Would you care to name a couple of those features? [I found a few that might on your list](https://circleci.com/docs/using-arm#limitations):
- doesn't support Arm in Docker
- doesn't support Arm in GCP (Google's fault)

I tried simply changing the resource class for our build_dev job and it failed with this error:
> Resource class "arm.2xlarge" is not a valid resource class.

cannonpalms:
> [I tried simply changing the resource class for our build_dev job](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/22946/workflows/cd5cfdac-5228-4cbf-bd50-91c73ae1652a/jobs/199638) and it failed with this error:

> > Resource class "arm.2xlarge" is not a valid resource class.

I believe that this _is_ a valid resource class, but not for the `docker` executor--only for the `LinuxVM` executor, which I'm guessing is a `machine` executor. https://circleci.com/docs/configuration-reference#resourceclass
![image](https://user-images.githubusercontent.com/2457559/187292574-64a6c829-2465-4cf3-a773-5a31a73a4f83.png)

The docs link above is quite difficult to read, but this note could be interpreted to apply to ARM LinuxVM resource classes:
> This resource requires review by our support team. [Open a support ticket](https://support.circleci.com/hc/en-us/requests/new) if you would like to request access.

If so, I believe @gaughen holds the keys to the CircleCI-support-request castle.

gaughen:
[...]

Issue #3119, opened 2021-11-16
docker run quay.io/influxdb/iox on Apple Silicon fails to run


jacobmarble:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behaviour.

1. Get a MacBook Pro (13-inch, M1, 2020) running MacOS 12.0.1
2. `docker run quay.io/influxdb/iox:bc112448`

__Expected behaviour:__
Describe what you expected to happen.

Image runs in container.

__Actual behaviour:__
Describe What actually happened.

```console
% docker run quay.io/influxdb/iox:bc112448
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

__Environment info:__

`docker run quay.io/influxdb/iox:bc112448`

Darwin 21.1.0 arm64

Docker Desktop 4.2.0 (Kubernetes is not running)

__Config:__
none

__Logs:__
n/a

crepererum:
Context:

We currently only build images for AMD64 and Linux. I think we could also build images for ARM64 (and Linux) given that we have a CI runner that works under ARM64. A bit more context: <https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/> . quay.io should support multi-arch as per <https://thenewstack.io/red-hats-quay-3-container-supports-multiple-architectures/>.

crepererum:
Is this feature no longer wanted? 🤔

jacobmarble:
I consider this issue to be unimportant, as I can build IOx outside of Docker, so I closed it. However, I can see how this could become important as we try to expand our ability to run small IOx instances for development purposes.

crepererum:
Having an ARM64 CI docker image is also a requirement for #3122.

PR #3133, opened 2021-11-17
chore: update one-shot Dockerfile to not depend on rust:ci

jacobmarble:
Related to #3119 and #3122

`docker build .` should just work, but it does not. The existing Dockerfile depends on image `quay.io/influxdb/rust:ci`, which is not published for ARM64, and doesn't seem to speed up the build process much anyhow.

This change simplifies that Dockerfile, though build times may increase.


crepererum:
This is NOT the dockerfile that we use for our prod images. We use https://github.com/influxdata/influxdb_iox/blob/main/docker/Dockerfile.iox . I'm wondering if we really need to dockerfiles here.

Dockerfile
```diff
@@ -1,9 +1,13 @@
 #syntax=docker/dockerfile:1.2
-FROM quay.io/influxdb/rust:ci as build
+FROM rust:1.56-slim-buster as build
```
e-dard, line 2: Would `rust:slim-buster` use the latest stable release?

crepererum: Please use `bullseye` instead of `buster` to be in line with our CI.

jacobmarble:
> This is NOT the dockerfile that we use for our prod images. We use https://github.com/influxdata/influxdb_iox/blob/main/docker/Dockerfile.iox . I'm wondering if we really need to dockerfiles here.

I added this Dockerfile because I just want to `docker build .` Yes, I'm aware of docker/Dockerfile.iox, but it doesn't work on MacOS because it copies the binary from the local workstation (non-Linux) into a container image (Linux).

crepererum (approved):

PR #3138, opened 2021-11-17
ci: update CI image builder to use newer docker

crepererum:
This is a precondition to build ARM64 CI images.


.circleci/config.yml
```diff
@@ -371,7 +371,8 @@ jobs:
   #
   # To modify the contents of the CI image, update docker/Dockerfile.ci
   ci_image:
-    machine: true
+    machine:
+      image: ubuntu-2004:202107-02
```
tustvold, line 375: Could you provide a link to where these are defined - it took me a lot of digging to not find a list :sweat_smile:

.circleci/config.yml
```diff
@@ -371,7 +371,8 @@ jobs:
   #
   # To modify the contents of the CI image, update docker/Dockerfile.ci
   ci_image:
-    machine: true
+    machine:
```
tustvold, line 374: I'm not sure if there is a benefit to using a machine vs [setup_remote_docker](https://circleci.com/docs/2.0/building-docker-images/) as we do above. Perhaps there might be an advantage to being consistent?

tustvold (approved):
Left some comments you can take or leave

.circleci/config.yml
crepererum: done

.circleci/config.yml
crepererum: added a doc section explaining the reasoning

Issue #3126, opened 2021-11-16
Flush Deletes on Persist


tustvold:
As discussed on https://github.com/influxdata/influxdb_iox/issues/2731#issuecomment-969122076 a persist operation must guarantee that any outstanding deletes are also persisted from the mailbox

PR #3142, opened 2021-11-17
refactor: extract Mailbox type (#3126)

tustvold:
This extracts the delete mailbox functionality into its own type, with the addition of synchronisation logic to allow coordination between persist lifecycle actions and the background worker.

An argument could be made that this additional synchronisation is redundant and could just make use of the single catalog transaction, but this would require duplicating logic in multiple places or pushing the mailbox into the PreservedCatalog. Having discussed with @crepererum I think this is simpler


alamb (approved):
I think the code does what it says , though I think it is worth considering using a tokio `Channel`

crepererum (approved):

internal_types/src/mailbox.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+}
+
+impl<T> Mailbox<T> {
+    /// Add an item to the inbox
+    pub fn push(&self, item: T) {
+        self.inbox.lock().push(item)
+    }
+
+    /// Get a handle to consumer data from this [`Mailbox`]
```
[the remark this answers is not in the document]
tustvold: I too like to consumer data :laughing:

tustvold:
> I think it is worth considering using a tokio Channel

I don't think this is a comparable abstraction to a channel:

* The contents can be inspected prior to flushing (important as catalog transactions are fallible)
* Potentially multiple consumers acquire exclusive access for the duration of a catalog transaction
* We need to be sure to take everything in the mailbox and not just some stuff

If there is some way to get this some other way, I'd be interested to know :smile:

server/src/db.rs
```diff
@@ -839,27 +840,15 @@ impl Db {
...
-                            // https://github.com/influxdata/influxdb_iox/issues/2626 is implemented
-                            guard.drain(0..todo.len());
-                        }
-                        Err(e) => {
-                            error!(%e, "cannot preserve delete predicates");
-                        }
-                    }
+                let handle = self.delete_predicates_mailbox.consume().await;
+                match self.preserve_delete_predicates(handle.outbox()).await {
+                    Ok(()) => handle.flush(),
```
alamb, line 845: should we possibly "unflush" the handle here or `panic!`? It seems like as written the deletes are dropped (not successfully applied)

alamb:
https://github.com/influxdata/influxdb_iox/pull/3148 <-- added the difference between channels 👍

server/src/db.rs
tustvold: I'm not sure I follow, if we get an error the handle is dropped and nothing is removed from the collection?

alamb: Ah -- I see -- because the handle holds the lock on the outbox. Sorry my bad.

PR #3144, opened 2021-11-17
feat: Move server config paths beneath 'nodes'

### LINKED ISSUE #3037: Move server files in object storage from /[server-id]/ to /nodes/[server-id]
I just reread the [floating databases doc](https://docs.google.com/document/d/1CXBZK-otK6-svjRca68bnk7VWiL8MGlUsFX2aQYcVbI/edit#heading=h.ro8gybfpbwe5) and noticed the server paths are different than they are currently:

```diff
- s3://<bucket..>/<server-id>/config.pb
+ s3://<bucket..>/nodes/<server-id>/config.pb
```

carols10cents:
Closes #3037.

This migrates server configuration files from `/[server id]/config.pb` to `/nodes/[server id]/config.pb`. 

On server startup:
- This code looks in `/nodes/[server id]/config.pb` for a server config. 
- If it doesn't find any content there, it falls back to looking in `/[server id]/config.pb` (this is temporary to ensure smooth migration). 
- If it doesn't find anything in either location, it assumes this is a new server that doesn't have any databases to initialize. 
- After initializing any databases found, it writes out the server configuration file at the new location (this code already existed before this PR as part of the migration from listing object storage to using the server config file initially). 
- Future server restarts will only use the config found in `/nodes/[server id]/config.pb` and won't look in `/[server id]/config.pb`.

It shouldn't be *required* to clear object storage, but we might want to to get rid of the old config files that won't be used after the config is written out at the new path.


iox_object_store/src/lib.rs
```diff
@@ -76,9 +76,23 @@ pub struct IoxObjectStore {
 impl IoxObjectStore {
     /// Get the data for the server config to determine the names and locations of the databases
     /// that this server owns.
+    ///
+    /// TEMPORARY: Server config used to be at the top level instead of beneath `/nodes/`. Until
```
alamb, line 80: Do we have a plan to transition deployments? In other words, is there any code that will move the server config files or will we have to do that manually?

Unless we move the config files via code perhaps this backwards compatibility code is just postponing the day of reckoning?

alamb (approved):
I just doubled checked https://docs.google.com/document/d/1CXBZK-otK6-svjRca68bnk7VWiL8MGlUsFX2aQYcVbI/edit# and confirmed this is the intended behavior. 

My only question is if we should bother with backwards compatibility code (or conversely if we should instead add some sort of object store migration code from old location to new location) 🤔 

FYI @mkmik

iox_object_store/src/lib.rs
carols10cents: I tried to explain how the transition will work in the PR description? Did I miss anything you're concerned about migrating?

alamb: No -- sorry @carols10cents  -- I just didn't read it closely enough (aka that code to write to the new config location already exists rather than being added as part of this PR).

alamb:
Merging this one in so it doesn't bitrot -- given that we have a migration path, I don't think anything else special needs to be done.

Thanks @carols10cents !

PR #3147, opened 2021-11-18
feat: flush delete mailbox on persist (#3126)

### LINKED ISSUE #3126: Flush Deletes on Persist

tustvold:
Closes #3126


server/src/db/replay.rs
```diff
@@ -2568,6 +2598,72 @@ mod tests {
...
+                        "+-----+------------------+--------------------------------+",
+                        "| 15  | b                | 1970-01-01T00:00:00.000000015Z |",
+                        "+-----+------------------+--------------------------------+",
+                    ],
+                )]),
+                Step::MakeWritesPersistable,
+                Step::Persist(vec![("table_1", "tag_partition_by_b")]),
+                Step::Restart,
+                Step::Replay,
+                Step::Assert(vec![Check::Query(
```
tustvold, line 2650: Without the changes to the lifecycle this assertion fails. This is because when persisting partition b, the delete to partition a is still in the mailbox. The persist, however, generates a DatabaseCheckpoint that indicates we are fully persisted up to sequencer number 2. The result is that we only replay from after sequencer number 2, and lose the delete :scream:

server/src/db/lifecycle/write.rs
```diff
@@ -155,6 +157,13 @@ pub(super) fn write_chunk_to_object_store(
                 .context(ParquetChunkError)?,
             );
 
+            // Collect any pending delete predicate from any partitions and include them in
```
alamb, line 160: 💯  for the comments

server/src/db/replay.rs
```diff
@@ -2568,6 +2598,72 @@ mod tests {
...
+                    ],
+                )]),
+                Step::MakeWritesPersistable,
+                Step::Persist(vec![("table_1", "tag_partition_by_a")]),
+                Step::Delete(vec![TestDelete {
+                    sequencer_id: 0,
+                    sequence_number: 1,
+                    table_name: None,
+                    predicate: DeletePredicate {
+                        range: TimestampRange { start: 0, end: 11 },
```
alamb, line 2627: What do you think about increasing this range to be to 20 (so it also would apply to the data that was ingested *after* it) to demonstrate the delete is applied correctly to only the data that was ingested prior ot it?

```suggestion
                        range: TimestampRange { start: 0, end: 20 },
```

alamb (approved):
👌  nice tests

And total kudos to @crepererum for setting up such a clear testing framework 👨‍🍳 😗 👍

crepererum (approved):
Sorry, I've somewhat reviewed this PR but didn't write a comment :D Looks good, thanks!

PR #3148, opened 2021-11-18
docs: document differences between Mailbox and channel

alamb:
~Pad by github stats by~ Perminently document @tustvold 's description of how `Mailbox` differs from `channel` in https://github.com/influxdata/influxdb_iox/pull/3142#issuecomment-972687495 to help future readers


internal_types/src/mailbox.rs
```diff
@@ -10,6 +10,17 @@ use std::sync::Arc;
...
+///
+/// 1. The contents can be inspected prior to remval from the mailbox
+/// (important as catalog transactions are fallible)
+///
+/// 2. Potentially multiple consumers can acquire exclusive access for
+/// the duration of a catalog transaction
+///
+/// 3. Users can ensure that everything in the mailbox is
+/// consumed
+///
```
tustvold, line 23:
```suggestion
/// This is different from a tokio channel in the following ways:
///
/// 1. The contents can be inspected prior to removal from the mailbox
/// (important as catalog transactions are fallible)
///
/// 2. Potentially multiple consumers can acquire exclusive access for
/// the duration of a catalog transaction
///
/// 3. Users can ensure that everything in the mailbox is consumed
///
```

tustvold (approved):

internal_types/src/mailbox.rs
tustvold: Darn kodiak was too fast

alamb: https://github.com/influxdata/influxdb_iox/pull/3152 lol

Issue #3150, opened 2021-11-18
IOx does not accept all valid line protocol


e-dard:
Whilst exploring IOx's compatibility with InfluxQL I noticed that many tests continued to fail after I fixed something in IOx that I expected to enable them to pass.

On closer investigation I determined that many of them were failing because data was not being written into IOx during the test setup.

Here is an example of some valid line protocol that one of these tests needs:

```
dupnames,region=us-east,day=1 value=10,day=3i 123456
```

Here is what happens if you write this into IOx:

```
$ influxdb_iox database create test_db
Created database test_db
d8c69b9f-9399-4b63-a418-1d39e82bb718

$ influxdb_iox database write test_db /tmp/lp

Error writing: Error converting lines: error writing line 1: Unable to insert iox::column_type::field::integer type into a column of iox::column_type::tag
```

It's pretty obvious what's going on here - IOx does not know how to handle two columns in the same table requiring the same column name. This is a pretty significant issue that we will need to resolve.

I have verified that this line protocol is valid and is accepted by our Cloud platform. It is queryable via Flux or InfluxQL. Here is an example of similar Flux output (notice the tag and field `day`).

![Screenshot 2021-11-18 at 13 41 54](https://user-images.githubusercontent.com/501993/142426323-aac679f3-2ccd-4c11-9133-1a2922fda5a2.png)

rbetts:
Fwiw, InfluxQL allows users to disambiguate tag and field identifiers using this syntax:

```
SELECT "<field_key>"::field,"<tag_key>"::tag           Returns a specific field and tag. The ::[field | tag] syntax specifies the identifier’s type. Use this syntax to differentiate between field keys and tag keys that have the same name.
```

See https://docs.influxdata.com/influxdb/v1.8/query_language/explore-data/#the-basic-select-statement

... and I'm realizing I don't really understand how flux does this.

pauldix:
Ugh, this was actually a bug in the original line protocol, or rather an unintentional omission that became a feature. We didn't have any enforcement that tag names and field names be unique, which is what we should have been doing all along. Within InfluxQL (the only query language we supported back then) there was no way to query data if you had done this because we didn't expect people to do this.

[...]

PR #3151, opened 2021-11-18
test: show valid line protocol does not get written into IOx

e-dard:
This PR shows a demonstration of IOx not correctly handling valid line protocol, as described in more detail in #3150

### DIFF
--- a/influxdb_line_protocol/src/lib.rs
+++ b/influxdb_line_protocol/src/lib.rs
@@ -1229,6 +1229,20 @@ mod test {
     }
 
     #[test]
+    fn parse_identical_tag_field_name() {
+        let input = "dupnames,day=monday day=1.2";
+        let vals = parse(input).unwrap();
+
+        assert_eq!(vals[0].series.measurement, "dupnames");
+        assert_eq!(vals[0].series.tag_set.as_ref().unwrap()[0].0, "day");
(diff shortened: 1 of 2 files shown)


influxdb_line_protocol/src/lib.rs
```diff
@@ -1228,6 +1228,20 @@ mod test {
         assert_eq!(vals[0].field_set[0].1.unwrap_i64(), 23);
     }
 
+    #[test]
+    fn parse_identical_tag_field_name() {
```
e-dard, line 1232: This test passes - the parser can handle this valid LP

mutable_batch_lp/src/lib.rs
```diff
@@ -180,4 +180,17 @@ mod tests {
             &[batch["mem"].to_arrow(Selection::All).unwrap()]
         );
     }
+
+    #[test]
```
e-dard, line 184: Because we need two columns called `space` in the same table, the MUB doesn't know what to do and barfs trying to convert the LP into record batches.

crepererum: What's the expected outcome here? Esp: how would this map to the SQL data model?

e-dard:
Closing as we're going to continue to reject these sorts of writes in IOx.

PR #3153, opened 2021-11-18
feat: skip applying deletes to persisted chunks during replay (#3125) 

### LINKED ISSUE #3125: Only Replay Deletes To In-Memory Chunks
As discussed on https://github.com/influxdata/influxdb_iox/issues/2731#issuecomment-969122076 when replaying, deletes should only be applied to unpersisted chunks.

tustvold:
~~_Draft as builds on #3147_~~

Closes #3125


server/src/db/replay.rs
```diff
@@ -370,6 +370,19 @@ impl<'a> WriteFilter for ReplayFilter<'a> {
     }
 }
 
+impl<'a> DeleteFilter for ReplayFilter<'a> {
```
alamb, line 373: This is cool (to reuse the `ReplayFilter` idea) 👍

alamb (approved):
I reviewed 91bf55786688ed25cc145d09ded76a306c411f1f and it makes sense to me 👍

server/src/db/replay.rs
tustvold: It should provide the ability to filter partitions and tables in future if this proves to be a relevant optimisation

PR #3156, opened 2021-11-18
feat: compact_object_store_chunks : initial implementation that also lists steps to complete

NGA-TRAN:
This is the first part of  #3089. The function is available but only step 1 is implemented and tested. Step 2-4 will be in the following PRs

### DIFF
--- a/lifecycle/src/policy.rs
+++ b/lifecycle/src/policy.rs
@@ -706,7 +706,6 @@ mod tests {
         Drop(ChunkId),
         Unload(ChunkId),
         Compact(Vec<ChunkId>),
-        CompactOS(Vec<ChunkId>),
         Persist(Vec<ChunkId>),
     }
 
@@ -913,8 +912,7 @@ mod tests {
             _partition: LifecycleWriteGuard<'_, TestPartition, Self>,
             _chunks: Vec<LifecycleWriteGuard<'_, TestChunk, Self::Chunk>>,
         ) -> Result<TaskTracker<()>, Self::Error> {
-            // This is just a trait function that this test may not need
-            unimplemented!();
+            unimplemented!("The test does not need compact os chunks");
         }
 
         fn prepare_persist(
--- a/lifecycle/src/policy.rs
+++ b/lifecycle/src/policy.rs
@@ -910,17 +910,11 @@ mod tests {
         }
 
         fn compact_object_store_chunks(
-            partition: LifecycleWriteGuard<'_, TestPartition, Self>,
-            chunks: Vec<LifecycleWriteGuard<'_, TestChunk, Self::Chunk>>,
+            _partition: LifecycleWriteGuard<'_, TestPartition, Self>,
+            _chunks: Vec<LifecycleWriteGuard<'_, TestChunk, Self::Chunk>>,
         ) -> Result<TaskTracker<()>, Self::Error> {
             // This is just a trait function that this test may not need
-            // Nothing is compacted in this function yet
-            // If you want this functionality, import and invoke compact_object_store_chunks
-            let event = MoverEvents::CompactOS(chunks.iter().map(|x| x.addr.chunk_id).collect());
-            let db = partition.data().db;
-            db.events.write().push(event);
-
-            Ok(db.registry.lock().complete(()))
+            unimplemented!();
         }
 
         fn prepare_persist(
--- a/lifecycle/src/policy.rs
+++ b/lifecycle/src/policy.rs
@@ -706,6 +706,7 @@ mod tests {
         Drop(ChunkId),
         Unload(ChunkId),
         Compact(Vec<ChunkId>),
+        CompactOS(Vec<ChunkId>),
         Persist(Vec<ChunkId>),
     }
 
@@ -902,6 +903,20 @@ mod tests {
 
             let event = MoverEvents::Compact(chunks.iter().map(|x| x.addr.chunk_id).collect());
 
+            let db = partition.data().db;
+            db.events.write().push(event);
+
+            Ok(db.registry.lock().complete(()))
+        }
+
+        fn compact_object_store_chunks(
+            partition: LifecycleWriteGuard<'_, TestPartition, Self>,
+            chunks: Vec<LifecycleWriteGuard<'_, TestChunk, Self::Chunk>>,
+        ) -> Result<TaskTracker<()>, Self::Error> {
+            // This is just a trait function that this test may not need
+            // Nothing is compacted in this function yet
+            // If you want this functionality, import and invoke compact_object_store_chunks
+            let event = MoverEvents::CompactOS(chunks.iter().map(|x| x.addr.chunk_id).collect());
             let db = partition.data().db;
             db.events.write().push(event);
 
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -676,7 +676,9 @@ impl Db {
             let partition = partition.upgrade();
 
             // invoke compact
-            let (_, fut) = lifecycle::compact_object_store::compact_object_store_chunks(partition, chunks).context(LifecycleError)?;
+            let (_, fut) =
+                lifecycle::compact_object_store::compact_object_store_chunks(partition, chunks)
+                    .context(LifecycleError)?;
             fut
         };
 
--- a/server/src/db.rs
+++ b/server/src/db.rs
@@ -651,6 +651,38 @@ impl Db {
         fut.await.context(TaskCancelled)?.context(LifecycleError)
     }
 
+    /// Compact all provided persisted chunks
+    pub async fn compact_object_store_chunks(
+        self: &Arc<Self>,
+        table_name: &str,
+        partition_key: &str,
+        chunk_ids: Vec<ChunkId>,
+    ) -> Result<Option<Arc<DbChunk>>> {
+        if chunk_ids.is_empty() {
+            return Ok(None);
+        }
+
+        // Use explicit scope to ensure the async generator doesn't
+        // assume the locks have to possibly live across the `await`
+        let fut = {
+            let partition = self.partition(table_name, partition_key)?;
+            let partition = LockableCatalogPartition::new(Arc::clone(self), partition);
+            let partition = partition.read();
+
+            // todo: set these chunks
+            let chunks = vec![];
+
+            // Lock partition for write
+            let partition = partition.upgrade();
+
+            // invoke compact
+            let (_, fut) = lifecycle::compact_object_store::compact_object_store_chunks(partition, chunks).context(LifecycleError)?;
+            fut
+        };
+
+        fut.await.context(TaskCancelled)?.context(LifecycleError)
+    }
+
     /// Persist given partition.
     ///
     /// If `force` is `true` will persist all unpersisted data regardless of arrival time
--- a/server/src/db/catalog/partition.rs
+++ b/server/src/db/catalog/partition.rs
@@ -374,14 +374,26 @@ impl Partition {
 
     /// Return true if there are no other persisted chunks that are in the middle of
     /// the provided chunk orders
+    // NGA todo: There is test_compact_os_non_contiguous_chunks in
+    // compact_object_store.rs to test this but I will add more unit tests right here
+    // when PR #3167 ChunkGenerator is merged
     pub fn contiguous_object_store_chunks(&self, chunk_orders: &BTreeSet<ChunkOrder>) -> bool {
+        // Last order in the chunk_orders for comparison
+        let last_order_element = chunk_orders.iter().rev().next();
+        let last_order = match last_order_element {
+            Some(last_order) => last_order,
+            None => {
+                return true;
+            } // provided chunk_orders is empty
+        };
+
         let chunks = self.chunks();
         for chunk in chunks {
             let chunk = chunk.read();
             if chunk.is_persisted() {
                 let order = chunk.order();
                 // this chunk does not belong to chunk_orders but in the middle of them
-                if !chunk_orders.contains(&order) && chunk_orders.range(order..).count() > 0 {
+                if !chunk_orders.contains(&order) && order < *last_order {
                     return false;
                 }
             }
--- a/server/src/db/catalog/partition.rs
+++ b/server/src/db/catalog/partition.rs
@@ -14,7 +14,11 @@ use persistence_windows::{
 };
 use schema::Schema;
 use snafu::{OptionExt, Snafu};
-use std::{collections::BTreeMap, fmt::Display, sync::Arc};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt::Display,
+    sync::Arc,
+};
 use time::{Time, TimeProvider};
 use tracker::RwLock;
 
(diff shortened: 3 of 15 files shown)


server/src/db.rs
```diff
@@ -651,6 +651,40 @@ impl Db {
...
+
+        // Use explicit scope to ensure the async generator doesn't
+        // assume the locks have to possibly live across the `await`
+        let fut = {
+            let partition = self.partition(table_name, partition_key)?;
+            let partition = LockableCatalogPartition::new(Arc::clone(self), partition);
+            let partition = partition.read();
+
+            // todo: set these chunks
+            let chunks = vec![];
```
NGA-TRAN, line 673: This will be done after all todos inside `compact_object_store_chunks(partition, chunks)` completed

lifecycle/src/policy.rs
```diff
@@ -908,6 +909,20 @@ mod tests {
             Ok(db.registry.lock().complete(()))
         }
 
+        fn compact_object_store_chunks(
+            partition: LifecycleWriteGuard<'_, TestPartition, Self>,
+            chunks: Vec<LifecycleWriteGuard<'_, TestChunk, Self::Chunk>>,
+        ) -> Result<TaskTracker<()>, Self::Error> {
+            // This is just a trait function that this test may not need
```
crepererum, line 916: if you don't need it, just put `unimplemented!("some reason")` in there.

server/src/db/catalog/partition.rs
```diff
@@ -368,6 +372,23 @@ impl Partition {
         self.chunks.iter()
     }
 
+    /// Return true if there are no other persisted chunks that are in the middle of
+    /// the provided chunk orders
+    pub fn contiguous_object_store_chunks(&self, chunk_orders: &BTreeSet<ChunkOrder>) -> bool {
```
crepererum, line 377: This function is complex enough to demand a unit test.

server/src/db/catalog/partition.rs
```diff
@@ -368,6 +372,23 @@ impl Partition {
...
+        let chunks = self.chunks();
+        for chunk in chunks {
+            let chunk = chunk.read();
+            if chunk.is_persisted() {
+                let order = chunk.order();
+                // this chunk does not belong to chunk_orders but in the middle of them
+                if !chunk_orders.contains(&order) && chunk_orders.range(order..).count() > 0 {
```
crepererum, line 384: `chunk_orders.range(order..).count() > 0` seems like a really costly check just to determine if there's anything at/after `order` within `chunk_order`.

You could probably get the last element once (before iterating over `chunks`):

```rust
let last_element: Option<ChunkOrder> = chunk_orders.iter().rev().next();
```

and then just compare that:

```rust
last_element.map(|o| order <= o).unwrap_or(false)
```

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+) -> Result<(
+    Time,
+    Time,
+    u64,
+    HashSet<Arc<DeletePredicate>>,
+    Vec<Arc<DbChunk>>,
+    ChunkOrder,
+    DatabaseCheckpoint,
+    PartitionCheckpoint,
+)> {
```
crepererum, line 168: Please use a struct here, esp. since the tuple elements are not strongly-typed (e.g. `u64` could be anything). Always prefer a struct over a tuple + docs. Performance-wise they are the same btw.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+            delete_predicates.extend(chunk.delete_predicates().iter().cloned());
+
+            min_order = min_order.min(chunk.order());
+            chunk_orders.insert(chunk.order());
+
+            // Todo:get chunk's datatbase_checkpoint and partition_checkpoint of the chunk and keep max
+
+            // Set chunk in the right action which is compacting object store
+            // This function will also error out if the chunk is not yet persisted
+            chunk.set_compacting_object_store(registration)?;
```
crepererum, line 230: I'm pretty sure we have this bug in other places as well, but whenever this state transition fails (e.g. at the third chunk) you might have some chunks already in the new lifecycle (e.g. the first two chunks). This lifecycle action will never be cleaned and the chunks are in this state forever. So ideally we would abort the lifecycle action here as well.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+    // Todo: ask Marco if this cleanup_lock is needed
+    // fetch shared (= read) guard preventing the cleanup job from deleting our files
+    let _guard = db.cleanup_lock.read().await;
+
+    // Create a new chunk for this stream data
+    //let table_name = Arc::from("cpu");
+    let table_name = Arc::from(partition_addr.table_name.to_string());
+    let partition_key = Arc::from(partition_addr.partition_key.to_string());
+
+    let chunk_id = ChunkId::new();
```
crepererum, line 347: @tustvold Since we've talked about it: it's perfectly fine to create freestanding chunk IDs w/o involving the in-mem catalog.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+        let partition = LockableCatalogPartition::new(Arc::clone(&db), Arc::clone(&db_partition));
+        let partition = partition.read();
+        let chunks = LockablePartition::chunks(&partition);
+        assert_eq!(chunks.len(), 4);
+        let partition = partition.upgrade();
+        let chunk1 = chunks[0].read();
+        let chunk3 = chunks[2].read();
+        let compact_non_persisted_chunks =
+            compact_object_store_chunks(partition, vec![chunk1.upgrade(), chunk3.upgrade()]);
+        assert!(compact_non_persisted_chunks.is_err());
```
crepererum, line 491: please check the error variant to string repr.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+        assert_eq!(partition_keys.len(), 1);
+        let db_partition = db.partition("cpu", &partition_keys[0]).unwrap();
+
+        // Test 1: no chunks provided
+        let partition = LockableCatalogPartition::new(Arc::clone(&db), Arc::clone(&db_partition));
+        let partition = partition.read().upgrade();
+        let compact_no_chunks = compact_object_store_chunks(partition, vec![]);
+        assert!(compact_no_chunks.is_err());
+
+        // test 2: persisted non persisted chunks
```
crepererum, line 441: Could you please split this into multiple test functions? I find "series-style" tests quite hard to reason about.

crepererum (commented):
Needs style and test changes, but I think overall that's a good start from where we can piece-by-piece add the remaining functionality + tests (like partition/db checkpoints and delete predicates, ...).

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: This is a good point. I think this action is needed to let other conflicting processes not change these chunks. However, we have not tried to change the action back after we are done. In this case, if it succeeds, all chunks will be dropped and we do not need to do anything. However, if it fails, then we have to set the action to the chunks' original ones.

What do you think if I create a ticket to handle setting the action to its stage before compacting/persisting for not only this case but for all available code that we believe hitting this bug? @crepererum and @tustvold

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: Any action for me here or this message is for @tustvold only?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,500 @@
...
+    db: &'a Db,
+    partition_addr: &'a PartitionAddr,
+    stream: SendableRecordBatchStream,
+    partition_checkpoint: PartitionCheckpoint,
+    database_checkpoint: DatabaseCheckpoint,
+    time_of_first_write: Time,
+    time_of_last_write: Time,
+    chunk_order: ChunkOrder,
+) -> Result<Arc<ParquetChunk>> {
+    // Todo: ask Marco if this cleanup_lock is needed
```
NGA-TRAN, line 338: @crepererum See question above

server/src/db/lifecycle/compact_object_store.rs
tustvold: There is already code to handle this - https://github.com/influxdata/influxdb_iox/blob/main/lifecycle/src/policy.rs#L412

There has been discussion around removing the need for this loop as a performance optimisation, but it hasn't been a priority.

NGA-TRAN: > removing the need for this loop as a performance optimisation

@tustvold : Can you be clearer what `this loop` means? Do I need any action here?

tustvold: > `this loop`

The code loops through every chunk looking for failed lifecycle actions, it would be nice to detect the failure as a consequence of some handle being dropped and enqueue recovery at "source". This would eliminate the need to go looking for failed jobs. The main reason it was implemented this way is to allow for backoff, but backoff could be achieved in other ways.

>  Do I need any action here

No action needed, if it shows up in a profile we can look into then :smile:

NGA-TRAN: Sounds good. In this specific case we need this loop to check for OS chunks and get their chunk order anyway.

NGA-TRAN:
@crepererum I have addressed all you comments.
@tustvold : you have to review this first thing Monday morning or I will bug you 😄 
I need this PR in to have the following pieces go easier

server/src/db/lifecycle/compact_object_store.rs
crepererum: This was only meant to be an FYI, nothing actionable :)

server/src/db/lifecycle/compact_object_store.rs
crepererum: Yes, and I think you're not holding it long enough here. The lock MUST be acquired BEFORE you store a parquet file and MUST be released AFTER you completed the preserved catalog transaction. Otherwise the "cleanup parquet files"-background-worker might just delete the parquet file that you are about to add to the catalog.

crepererum (changes requested):
The DB clean-up lock needs to be moved a bit.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,572 @@
...
+
+    use data_types::database_rules::LifecycleRules;
+    use lifecycle::{LockableChunk, LockablePartition};
+    use query::{QueryChunk, QueryDatabase};
+    use std::{
+        num::{NonZeroU32, NonZeroU64},
+        time::Duration,
+    };
+
+    // Todo: this as copied from persist.rs and should be revived to match the needs here
```
tustvold, line 426: I think you can just use `make_db` here which calls `TestDb::builder().build().await`.

These tests don't need to be overriding the late_arrive_window_seconds as compacting object store is not concerned with such concepts. By extension I think it doesn't need the time provider, as Wall time has no bearing on its behaviour. 

`worker_backoff_millis` is just a hack to stop the lifecycle policy interfering and starting compactions, etc... in parallel - but I think for these tests there is unlikely to be anything to trigger such behaviour and so you might get away without it

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,572 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_compact_os_non_contiguous_chunks() {
+        test_helpers::maybe_start_logging();
+
+        let (db, time) = test_db().await;
+        let late_arrival = Duration::from_secs(1);
+        write_lp(db.as_ref(), "cpu,tag1=cupcakes bar=1 10").await;
+        time.inc(late_arrival);
```
tustvold, line 509: All this stuff about time providers and late arrival is not needed here, as we aren't concerned with persisting unpersisted data (which must have been in-memory for late_arrival_seconds)

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,572 @@
...
+        write_lp(db.as_ref(), "cpu,tag1=chunk3,tag2=a bar=2 30").await;
+        db.persist_partition("cpu", partition_keys[0].as_str(), true)
+            .await
+            .unwrap()
+            .unwrap()
+            .id();
+        //
+        // Add a MUB
+        write_lp(db.as_ref(), "cpu,tag1=chunk4,tag2=a bar=2 40").await;
+        // todo: Need to ask Marco why there is no handle created here
```
tustvold, line 540: Legacy comment?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,572 @@
...
+            partition: partition.addr().clone(),
+            chunks: vec![chunk1.id(), chunk3.id()],
+        });
+
+        let compact_non_contiguous_persisted_chunks =
+            mark_chunks_to_compact(partition, vec![chunk1, chunk3], &registration);
+        let err = compact_non_contiguous_persisted_chunks.unwrap_err();
+        assert!(err
+            .to_string()
+            .contains("Cannot compact the provided persisted chunks. They are not contiguous"));
```
tustvold, line 562: Would be good to print the error string on assertion failure

tustvold (commented):
Looks good to me, some minor nits.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -0,0 +1,572 @@
...
+    // Verify if all the provided chunks are contiguous
+    if !partition.contiguous_object_store_chunks(&chunk_orders) {
+        return ChunksNotContiguous {}.fail();
+    }
+
+    let time_of_first_write = time_of_first_write.expect("Should have had a first write somewhere");
+    let time_of_last_write = time_of_last_write.expect("Should have had a last write somewhere");
+
+    // drop partition lock
+    let _partition = partition.into_data().partition;
```
tustvold, line 230: Could also use `std::mem::drop(partition)` or simply leave it to exiting the scope of the function to drop the lock

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -82,26 +82,39 @@ pub(crate) fn compact_object_store_chunks(
...
+        // Step 2: Compact the os chunks into a stream
+        let compacted_stream = compact_chunks(&db, &compacting_os_chunks.os_chunks).await?;
+        let compacted_rows;
+        let _schema = compacted_stream.schema;
+        let sort_key = compacted_stream.sort_key;
+
+        // Step 3: Start to persist files and update the preserved catalog accordingly
+        // This process needs to hold cleanup lock to avoid the persisted file was deleted right after
+        // it is created and before it is updated in the preserved catalog
+        {
```
crepererum, line 94: :+1:

crepererum (approved):

NGA-TRAN:
@tustvold 
Hah: I forgot to remove "ataumerge" and kodiak has merged it after it got an approval. I am addressing your comments and they will go with my next PR which might include new pieces

NGA-TRAN (commented):
ALl @tustvold 's comments have been addressed in #3183

Issue #2477, opened 2021-09-07
Predicate and limit pushdown for system tables


tustvold:
Currently predicates and limits are not pushed down into [scans for system tables](https://github.com/influxdata/influxdb_iox/blob/main/server/src/db/system_tables.rs#L139), this combined with the fact the current design aims to produce a [single RecordBatch](https://github.com/influxdata/influxdb_iox/blob/main/server/src/db/system_tables.rs#L116) results in query timeouts on systems with large catalogs.

This should be fixed.

For reference the query I was running is

```
select table_name,
       partition_key,
       row_count,
       (cast(min_value as bigint) - cast(previous_max_value as bigint)) / 1000000000 as delta_seconds
from (select *,
             LAG(max_value)
             OVER (partition by table_name, partition_key order by table_name, partition_key, chunk_id) as previous_max_value
      from system.chunk_columns
      where column_name = 'time'
        and storage = 'ReadBufferAndObjectStore' and table_name like 'http_api%');
```

alamb:
This is a valuable feature / idea for sure. 👍

alamb:
TLDR: as implied by this ticket's description, the vast majority of time spent in system table queries is spent creating the `RecordBatch`  for the queries. Adding support for `limit`, `projection` and `predicates` will help.

## Analysis
I inserted a "bunch" of data into my local instance using load generator with commands like

```shell
../target/release/iox_data_generator -h localhost:8080 --token foo --org org --bucket bucket -s schemas/cap-write.toml  --start '356 days ago' --batch_size=1000
```

This resulted in  `1122347` rows in the `system.column_chunks` table and I ran the queries (with a release `influxdb_iox` build)  as shown under "Query Timings".

I observed a few things:
1. The complexity of the query did not appreciably affect query performance (`count(*)` took about the same as something with several subqueries) ;
1. Creating a `String` for each UUID (which are then copied and freed) was a hot spot in the profile (see screen shot below)
3. There is no `limit` in theses queries but I often write queries like `select * from system.chunks limit 10` as a way to see the schema of the table.


## Unscientific Instruments profile
![Screen Shot 2021-11-17 at 3 00 00 PM](https://user-images.githubusercontent.com/490673/142288180-b6575875-923d-4941-ba67-0c7541a959c9.png)

## Thoughts and Analysis

There are many different ways to optimize system table creation so one
key consideration is how to avoid spending too much effort here.

[...]

PR #3157, opened 2021-11-18
(WIP) feat: Only create columns for system tables that are requested

alamb:
Editor's note: throwing this up as I switch to something else -- I sketched out an interface I would like but it needs some touch up and then to refactor system table creation to use it

Builds on https://github.com/influxdata/influxdb_iox/pull/3155

Related to  https://github.com/influxdata/influxdb_iox/issues/2477

# Rationale
1. Creating the raw data for the `RecordBatch` for system tables takes significant time for large databases
2. DataFusion already does "projection pushdown" (aka specifies what columns are needed)

Changes:
1. Interpret and use projection from DataFusion
2. XXX

### DIFF
--- a/server/src/db/system_tables.rs
+++ b/server/src/db/system_tables.rs
@@ -9,11 +9,7 @@
 
 use super::catalog::Catalog;
 use crate::JobRegistry;
-use arrow::{
-    datatypes::{Field, Schema, SchemaRef},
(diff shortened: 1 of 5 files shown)


alamb:
haven't had any time to work on this, and it has bitrotted 😢  will revisit when I have more time

PR #3160, opened 2021-11-18
fix: Small cleanups in Cargo.tomls

carols10cents:
Some things I noticed while investigating why crates are rebuilding so often. These fixes won't help that problem much, but they do make the Cargo.tomls a bit cleaner and fix a few crates that couldn't be compiled with `-p`.

I *think* we're using the new feature resolver because all the crates are using `edition = "2021"`, but [the docs say to specify the resolver in the `[workspace]` table of a virtual workspace.](https://doc.rust-lang.org/nightly/cargo/reference/resolver.html#resolver-versions)


influxdb_iox/Cargo.toml
```diff
@@ -103,7 +103,6 @@ write_buffer = { path = "../write_buffer" }
 
 # Crates.io dependencies, in alphabetical order
 assert_cmd = "2.0.2"
-flate2 = "1.0"
```
alamb, line 106: 🎉

alamb (approved):
Thank you @carols10cents

influxdb_iox/Cargo.toml
crepererum: It's not really gone, it's part of the normal deps ;)

PR #3161, opened 2021-11-19
fix: don't retry "forever" sending Kafka messages

### LINKED ISSUE #3030: Router fails writing to write buffer when kafka leader changes
__Steps to reproduce:__

1. have a routing tier that writes to kafka
2. perform a rolling restart of kafka (e.g. an upgrade or a change in parameters)

__Expected behaviour:__

The routing tier should continue to write to kafka despite the leader changing under its feet.
A small increase latency is expected

__Actual behaviour:__

Routing tier returns errors for a while and then recovers.

__Environment info:__

infludxdata's iox deployment on stag03's

__Logs:__

```
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9737\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026100004984263
level=error msg="librdkafka: Global error: BrokerTransportFailure (Local: Broker transport failure): iox-kafka-1.iox-kafka-brokers.iox.svc:9092/1: Disconnected (after 48985ms in state UP, 1 identical error(s) suppressed)" log.target=rdkafka::client log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=92 target="log" time=1636026105434016353
level=error msg="librdkafka: Global error: AllBrokersDown (Local: All broker connections are down): 4/4 brokers are down" log.target=rdkafka::client log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=92 target="log" time=1636026105434035113
level=error msg="librdkafka: FAIL [thrd:iox-kafka-1.iox-kafka-brokers.iox.svc:9092/1]: iox-kafka-1.iox-kafka-brokers.iox.svc:9092/1: Connect to ipv4#10.136.108.142:9092 failed: Connection refused (after 1ms in state CONNECT)" log.target=librdkafka log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=63 target="log" time=1636026105435571770
level=error msg="librdkafka: Global error: BrokerTransportFailure (Local: Broker transport failure): iox-kafka-1.iox-kafka-brokers.iox.svc:9092/1: Connect to ipv4#10.136.108.142:9092 failed: Connection refused (after 1ms in state CONNECT)" log.target=rdkafka::client log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=92 target="log" time=1636026105435602555
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9752\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026110014719372
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9749\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026150054380625
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9762\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026160064964580
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9748\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026170074046188
level=error msg="Error while handling request" error="Internal error writing points into org 3bd2399eac571807, bucket 8dc736eef15fc59d:  Cannot write to write buffer: Message production error: MessageTimedOut (Local: Message timed out)" method=POST uri=/api/v2/write?bucket=8dc736eef15fc59d&org=3bd2399eac571807 content_length="Some(\"9756\")" target="influxdb_iox::influxdb_ioxd::http" location="influxdb_iox/src/influxdb_ioxd/http.rs:423" time=1636026210114558357
level=error msg="librdkafka: FAIL [thrd:iox-kafka-2.iox-kafka-brokers.iox.svc:9092/2]: iox-kafka-2.iox-kafka-brokers.iox.svc:9092/2: Connect to ipv4#10.136.23.205:9092 failed: Connection timed out (after 128736ms in state CONNECT)" log.target=librdkafka log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=63 target="log" time=1636026213366958414
level=error msg="librdkafka: Global error: BrokerTransportFailure (Local: Broker transport failure): iox-kafka-2.iox-kafka-brokers.iox.svc:9092/2: Connect to ipv4#10.136.23.205:9092 failed: Connection timed out (after 128736ms in state CONNECT)" log.target=rdkafka::client log.module_path=rdkafka::client log.file=/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/rdkafka-0.27.0/src/client.rs log.line=92 target="log" time=1636026213367020019
```

CC: @crepererum @primeroz

crepererum:
When a Kafka broker pod is recreated (for whatever reason) and gets a
new IP while doing so, the following happened:

1. Old broker pod gets terminated, but is still reachable via DNS and
   TCP.
2. rdkafka looses its connection, re-creates it using the old IP. The
   TCP connection can be established (this heavily depends on the K8s
   network setup), but won't be able to send any messages because the
   old broker is already shutting down / dead.
3. New broker gets created w/ new IP (but same DNS name).
4. Somewhat in parallel to step 3: rdkafka gets informed by other
   brokers that the topic lost its leader and then that the topic has
   the new leader (which has the same identity as the old one). Since
   leader changes in Kafka can also happen when brokers are totally
   healthy, it doesn't conclude that its TCP connection might be broken
   and tries to send messages to the new broker via the old TCP
   connection.
5. It takes very long (~130s on my test setup) for the old
   rdkafka->broker TCP connection to break. Since
   `message.send.max.retries` has a default of `2147483647` rdkafka will
   not give up on the application level.
5. rdkafka re-connects, while doing so resolves via DNS the new broker
   IP and is happy.

An alternative fix that was tried: Use the `connect` rdkafka callback to
hook into the place where it would issue the UNIX `connect` call. There
we can manipulate the socket. Setting `TCP_USER_TIMEOUT` to 5000ms also
solves the issue somewhat, but might have different implications (also
it then takes around 5s to kill the connection). Since this is a more
hackish implementation and somewhat an unofficial way to configure
rdkafka, I decided against it.

Test Setup
==========

```rust
#[tokio::test]
async fn write_forever() {
    maybe_start_logging();
    let conn = maybe_skip_kafka_integration!();
    let adapter = KafkaTestAdapter::new(conn);
    let ctx = adapter.new_context(NonZeroU32::new(1).unwrap()).await;

    let writer = ctx.writing(true).await.unwrap();
    let lp = "upc user=1 100";
    let sequencer_id = set_pop_first(&mut writer.sequencer_ids()).unwrap();

    for i in 1.. {
        println!("{}", i);

        let tables = mutable_batch_lp::lines_to_batches(lp, 0).unwrap();
        let write = DmlWrite::new(tables, DmlMeta::unsequenced(None));
        let operation = DmlOperation::Write(write);
        let res = writer.store_operation(sequencer_id, &operation).await;
        dbg!(res);

        tokio::time::sleep(Duration::from_secs(1)).await;
    }
}
```

Make sure to set the the rdkafka `log` config to `all`. Then use KinD,
setup a 3-node Strimzi cluster and start the test binary within the K8s
cluster. You need to start a debug container that is close enough to
your developer system (e.g. an old Debian DOES NOT work if you run
bleeding edge Arch):

```console
$(host) kubectl run -i --tty --rm debug --image=archlinux --restart=Never -n kafka -- bash
````

Then you copy over the test binary the container using [cargo-with](https://github.com/cbourjau/cargo-with):

```console
$(host) cargo with 'kubectl cp {bin} kafka/debug:/foo' -- test -p write_buffe
````

Within the container shell that you've just created, start the
forever-running test (make sure to set `KAFKA_CONNECT` according to your
Strimzi setup!):

```console
$(container) TEST_INTEGRATION=1 KAFKA_CONNECT=my-cluster-kafka-bootstrap:9092 RUST_BACKTRACE=1 RUST_LOG=debug ./foo write_forever --nocapture
````

The test should run and tell you that it is delivering messages. It also
tells you within the debug logs which broker it sends the messages to.
Now you need to kill the broker (in my example it was `my-cluster-kafka-1`):

```console
$(host) kubectl -n kafka delete pod my-cluster-kafka-1
````

The test should now stop to deliver messages and should error. Without
this patch it might take over 100s for it to recover even after the
deleted pod was re-created. With this patch it quickly is able to
deliver data again after the broker comes back online.

Fixes #3030.


mkmik:
Does the iox client see one (or more errors) once the 10 retries are exhausted?

tustvold (approved):

mkmik (approved):

tustvold:
> rdkafka will not give up on the application level

With this change are we relying on rdkafka giving up and returning an error to the clients, this seems unfortunate? Will the router just return errors for a period until the broker comes back? I'm confused...

crepererum:
> > rdkafka will not give up on the application level
> 
> With this change are we relying on rdkafka giving up and returning an error to the clients, this seems unfortunate? Will the router just return errors for a period until the broker comes back? I'm confused...

We're currently bubbling all sorts of errors back to the client, w/ and w/o this patch (e.g. I've seen a phase where rdkafka thinks the topic is gone entirely). I agree that this should be fixed, but that's not the point of this PR. The point here is that rdkafka (which I meant w/ "application level", sorry for the confusion) should check if its TCP connection is actually working instead of retrying forever.

Issue #2209, opened 2021-08-06
PartitionCheckpoint Without Parquet File


tustvold:
Currently it is only possible to generate a PartitionCheckpoint associated with a parquet file. With the ability to drop unpersisted data, however, there comes a need to be able to generate a PartitionCheckpoint without any associated data. I suspect it would be valid to only include this in the catalog transactions, and not persist a file, provided the catalog-rebuild from parquet files is aware that a later DatabaseCheckpoint may be inconsistent with an earlier PartitionCheckpoint.

This is likely a requirement to properly fix #2208 and #2207

jacobmarble:
@tustvold can this be closed? Looks like #2207 and #2208 are resolved.

NGA-TRAN:
@tustvold : this sounds important to avoid bugs. Is this still valid and need our attention soon?

tustvold:
@NGA-TRAN I think #3145 and #3146 which motivate this issue end up being just optimisations, as opposed to necessary for correctness. I started implementing this, but in so doing convinced myself that you couldn't get invalid behaviour just sub-optimal. I could be wrong though?

NGA-TRAN:
@tustvold I do not know this area well so I let you decide. If you think this is not important and won't help much, close it is a good option 😄

crepererum:
Could we be a bit more precise on the "suboptimal but not invalid" behavior here? IIRC the issue is when you delete the entire partition, we don't store any checkpoint for it. I think we still have some subtle bug here that we lead forever-open persistence windows if you have a partition that gets write which is immediately deleted w/o any persistence action in the meantime. The in-mem persistence windows don't track the delete, so databases checkpoints created during the persistence of other partitions will just have a minimum of the never-persisted write. In that case we won't be able to perform replay properly after a day or so.

tustvold:
Aah yes, I remember. You need this to correctly handle a sequencer with deletes but not writes. As nothing will checkpoint its position in the stream, stuff will get rotated out of Kafka and eventually the database will error in replay complaining about skipped sequence numbers

> a partition that gets write which is immediately deleted w/o any persistence action in the meantime

I think this is a separate bug if this is the case, it shouldn't be possible to delete unpersisted data. (TBH I question the value of unsequenced delete APIs at all).

crepererum:
I wasn't talking about unsequenced data here. I think we're not updating the persistence windows when we receive deletes via a write buffer.

tustvold:
[...]

PR #3163, opened 2021-11-19
refactor: add parquet chunk generator (#2209)

tustvold:
In order to write tests for #2209 I need to be able to construct parquet chunks with meaningful checkpoints, currently they are all the same.

Additionally, the current code to generate parquet files for testing purposes I at least found incredibly confusing, there are lots of free functions that generate slightly different variants of the components that comprise a ParquetChunk. This is all not helped by the fact there are 5 different types of metadata nested within each other which was just :exploding_head: 

Finally, generating checkpoints is inherently stateful, they can't go backwards, and so I need somewhere for this state to be stored.

I therefore extracted a `ChunkGenerator` that can generate `ParquetChunk`. It also returns the `IoxMetadata` that was encoded into that `ParquetChunk` - this avoids tests other than those that wish to test the round-trip from having to do the decode dance.

I've updated the tests within parquet_file to use this, and will update the tests in parquet_catalog in a follow up PR.


parquet_file/src/test_utils.rs
```diff
@@ -895,7 +890,7 @@ pub fn read_data_from_parquet_data(schema: SchemaRef, parquet_data: Vec<u8>) ->
 
 /// Create test metadata by creating a parquet file and reading it back into memory.
 ///
-/// See [`make_chunk`] for details.
+/// TODO(raphael): Replace with ChunkGenerator
```
tustvold, line 893: This is the function used by `parquet_catalog` - I intend to replace it with usage of ChunkGenerator in a follow up PR

parquet_file/src/storage.rs
```diff
@@ -660,8 +634,7 @@ mod tests {
...
             "+----------------+---------------+-------------------+------------------+-------------------------+------------------------+----------------------------+---------------------------+----------------------+----------------------+-------------------------+------------------------+----------------------+----------------------+-------------------------+------------------------+----------------------+-------------------+--------------------+------------------------+-----------------------+-------------------------+------------------------+-----------------------+--------------------------+-------------------------+-----------------------------+",
         ];
-        assert_eq!(num_rows, actual_num_rows);
```
crepererum, line 663: why is the n-rows check gone?

crepererum (approved):
I agree that the testing setup is somewhat organically grown and should be cleaned up a bit, so thanks for tackling it :+1:

parquet_file/src/storage.rs
tustvold: The n rows check is still there, the check of the RecordBatches that were written to Parquet has been removed. I can add it back if you feel strongly, but as it stands the ChunkGenerator isn't exposing this and I felt it somewhat redundant?

crepererum: Since we read the record batches as-is and we don't have any higher-level dedup logic in place, I think this is fine :+1:

PR #3167, opened 2021-11-19
refactor: use ChunkGenerator in parquet_catalog (#2209)

tustvold:
#3163 added a ChunkGenerator for generating persisted parquet files for tests. This PR migrates parquet_catalog to use this abstraction, and removes the old code.

This is necessary functionality for #2209 as I need to write tests with real partition checkpoints stored in the parquet files


parquet_catalog/src/core.rs
```diff
@@ -1889,102 +1902,56 @@ mod tests {
...
-            };
-            state.insert(info.clone()).unwrap();
-            t.add_parquet(&info);
+            for _ in 0..4 {
+                let (chunk, _) = generator.generate().await;
+                let info = CatalogParquetInfo::from_chunk(&chunk);
+                expected.push(chunk);
+                state.insert(info.clone()).unwrap();
+                t.add_parquet(&info);
+            }
```
NGA-TRAN, line 1911: It is nice to see you refactor and use generator to simplify the test code everywhere 👍

parquet_catalog/src/dump.rs
```diff
@@ -304,11 +298,11 @@ File {
...
                                             "part1",
                                         ],
-                                        file_name: "00000000-0000-0000-0000-000000000000.parquet",
+                                        file_name: "00000000-0000-0000-0000-000000000001.parquet",
                                     },
                                 ),
-                                file_size_bytes: 33,
-                                metadata: b"metadata omitted (937 bytes)",
+                                file_size_bytes: 3052,
+                                metadata: b"metadata omitted (935 bytes)",
```
NGA-TRAN, line 305: Have you verified all the changes here? they are quite different including the file name

parquet_catalog/src/dump.rs
```diff
@@ -500,7 +488,7 @@ File {
                                 },
                             },
                             chunk_order: ChunkOrder(
-                                5,
+                                1,
```
NGA-TRAN, line 491: Some explanation about the difference in chunk order will help

NGA-TRAN (approved):
I can see I will use this parquet chunk generator for my tests. My only comment is verification of the difference test output. Nice work, @tustvold

parquet_catalog/src/dump.rs
tustvold: It's just bundling slightly different data within the parquet files, in terms of testing the catalog's mechanics it is immaterial what the parquet data actually is. But will let @crepererum confirm

parquet_catalog/src/dump.rs
tustvold: Sorry, should have commented. Previously all chunks would have the same ChunkOrder. ChunkGenerator uses the same as the chunk id. This shouldn't have any impact on these tests, as the order only impacts query

parquet_file/src/test_utils.rs
```diff
@@ -100,102 +96,6 @@ pub async fn load_parquet_from_store_for_path(
...
-        column_summaries,
-    )
-    .await
-}
-
-/// Create a test chunk by writing data to object store.
-///
-/// TODO: This code creates a chunk that isn't hooked up with metrics
-///
-/// TODO(raphael): Replace with ChunkGenerator
```
alamb, line 143: Mission Accomplished !

parquet_catalog/src/dump.rs
crepererum: That's OK, this test isn't really meant to be a size check anyways

crepererum (approved):
Thank you!

Issue #3117, opened 2021-11-15
Sync features of dependencies + dev-dependencies to avoid rebuilds between build/test


carols10cents:
I think we're running into this:

> `cargo test` would trigger a rebuild compared to `cargo build`. It's a limitation of how dependencies are constructed, but also an intentional design to avoid duplicate builds during a normal `cargo test` for most projects. If you want to avoid rebuilds in all situations, you'll have to add features to your normal dependencies to sync them with the dev-dependencies.

_Originally posted by @ehuss in https://github.com/rust-lang/cargo/issues/2904#issuecomment-632292650_

carols10cents:
Now that we have a workspace hack crate, I'm going to go through each of the entries in `workspace-hack/Cargo.toml` to analyze:

- Why they are built with different feature sets
- Why they are dependencies at all
- Whether there's anything we can/should change about the way we depend on them

I'll post my analysis here and link to changes.


## ahash

```
[dependencies]
ahash = { version = "0.7", features = ["std"] }
[build-dependencies]
ahash = { version = "0.7", features = ["std"] }
```

- https://crates.io/crates/ahash
- https://docs.rs/ahash/0.7.6
- Depended on by arrow-util, datafusion, hashbrown, pprof

> std: This enables features which require the standard library. (On by default) This includes providing the utility classes AHashMap and AHashSet.

It looks like the only part of ahash used in iox is [`RandomState`](https://docs.rs/ahash/0.7.6/ahash/struct.RandomState.html), which doesn't require the standard library. Therefore, we can safely use `default-features = false` to our deps on ahash in arrow-util.

It doesn't look like datafusion uses any of the ahash structures that need std either, so I'm sending in a PR to turn default features off there too.

hashbrown doesn't enable ahash's std feature.

pprof depends on inferno which depends on ahash, and inferno does use the data structures enabled by ahash std, but pprof is optional to iox, so I don't think this matters. (This led me to exclude pprof from cargo-hakari's config)

carols10cents:
## bytes

```
[dependencies]
bytes = { version = "1", features = ["std"] }
[build-dependencies]
bytes = { version = "1", features = ["std"] }
```

- https://crates.io/crates/bytes
- https://docs.rs/bytes/1.1.0/bytes/
- Depended on by:
[...]

PR #3168, opened 2021-11-19
feat: Add a workspace hack crate by using `cargo hakari`

carols10cents:
Connects to #3117.

## What is a workspace hack crate?

It's a crate that explicitly turns on all features of dependencies that any crate in the workspace uses.

## Why would we want such a thing?

When building the whole workspace, like with `cargo test --workspace`, Cargo unifies all the features of all the crates used. When building just one crate in the workspace, like with `cargo test -p server`, Cargo only builds dependencies with features that the `server` crate or its dependencies use, which are probably a different set than the whole workspace. This causes lots of crates to be rebuilt. The workspace hack crate makes it so that more crates can be shared by always unifying all the features.

## What is `cargo hakari`?

[cargo-hakari](https://crates.io/crates/cargo-hakari) is a crate that manages workspace hack crates. It has some excellent documentation!

## Well? Does it help?

Yes!

### Before the workspace hack crate

- Delete `/target`
- Run `cargo test --workspace`
- Run `cargo test -p server`, 79 crates get compiled
- Run `cargo test -p query_tests`, 49 crates get compiled
- `ls -l target/debug/deps | wc -l` => 10986

### After the workspace hack crate

- Delete `/target`
- Run `cargo test --workspace`
- Run `cargo test -p server`, 2 crates get compiled
- Run `cargo test -p query_tests`, 1 crate gets compiled
- `ls -l target/debug/deps | wc -l` => 10525, difference of 461 files that no longer need to be compiled

## Does this solve all our problems?

Well, no. As you'll see in the workspace hack crate, we've got LOTS of features turned on for LOTS of crates, and we might not need all of them. The next step I'm going to work on is researching if we need them all and if we can trim down this list of features.

Also, this will make initial builds (and builds after updating Rust or cleaning out `/target`) take longer, but at the tradeoff of needing to *re*build less while developing, so I think it's worth it.

This also adds yet another tool to use and yet another CI job to keep the workspace hack crate up-to-date. Please read the documentation in `workspace-hack/README.md` and let me know if it explains enough. Is there anywhere else it should be documented or linked to so that it's clear what to do if this CI job fails?


tustvold (commented):
Is cargo smart enough to know that these aren't real dependencies - i.e. the code isn't actually used? 

Basically I'm curious because I've put a fair bit of effort into reducing transitive dependencies, perhaps I'm wrong and this doesn't have a discernible impact on the compile times of crates - but I'd be interested to know either way. It has certainly been my experience that crates with fewer dependencies compile faster, but wonder if this is something you've explored.

Effectively this is a fairly massive transitive dependency blob brought in by every crate, so instinctively I'd expect this to hurt link times at the very least...

Edit: Perhaps there are some heavy hitters like tokio, that could benefit from this treatment, but not others? Perhaps we could just use this for ones where the dependency is "beyond our control" and not for ones like cache_loader_async where the Cargo.toml within the IOx repo itself are just inconsistent?

workspace-hack/Cargo.toml
```diff
@@ -0,0 +1,84 @@
...
+# are managed by hakari.
+
+### BEGIN HAKARI SECTION
+[dependencies]
+ahash = { version = "0.7", features = ["std"] }
+bytes = { version = "1", features = ["std"] }
+cache_loader_async = { version = "0.1", features = ["ttl-cache"] }
+chrono = { version = "0.4", features = ["clock", "libc", "oldtime", "serde", "std", "time", "winapi"] }
+clap = { version = "2", features = ["ansi_term", "atty", "color", "strsim", "suggestions", "vec_map"] }
+criterion = { version = "0.3", features = ["async", "async_tokio", "cargo_bench_support", "futures", "tokio"] }
```
tustvold, line 21: Forcing everything to depend on criterion (which is solely used by benchmarks), and brings in things like crossbeam, etc... seems indicative of my concern r.e. massive transitive dependencies.

workspace-hack/Cargo.toml
```diff
@@ -0,0 +1,84 @@
...
+regex-automata = { version = "0.1", features = ["regex-syntax", "std"] }
+regex-syntax = { version = "0.6", features = ["unicode", "unicode-age", "unicode-bool", "unicode-case", "unicode-gencat", "unicode-perl", "unicode-script", "unicode-segment"] }
+reqwest = { version = "0.11", features = ["__rustls", "__tls", "blocking", "default-tls", "hyper-rustls", "hyper-tls", "json", "native-tls-crate", "rustls", "rustls-tls", "rustls-tls-webpki-roots", "serde_json", "stream", "tokio-native-tls", "tokio-rustls", "webpki-roots"] }
+serde = { version = "1", features = ["derive", "rc", "serde_derive", "std"] }
+serde_json = { version = "1", features = ["indexmap", "preserve_order", "std"] }
+smallvec = { version = "1", default-features = false, features = ["union"] }
+tikv-jemalloc-sys = { version = "0.4", features = ["background_threads_runtime_support", "stats", "unprefixed_malloc_on_supported_platforms"] }
+tokio = { version = "1", features = ["bytes", "fs", "full", "io-std", "io-util", "libc", "macros", "memchr", "mio", "net", "num_cpus", "once_cell", "parking_lot", "process", "rt", "rt-multi-thread", "signal", "signal-hook-registry", "sync", "time", "tokio-macros", "winapi"] }
```
tustvold, line 53: "winapi" :thinking:

carols10cents:
Yep, I can absolutely work on reducing this *before* we merge it in! And looks like I broke heappy because that's mutually exclusive from some of the features that are now all on... 😅

workspace-hack/Cargo.toml
alamb: Yeah, `criterion` is not a small dependency -- it would be great actually to remove it as a development dependency of all our crates except for some specific benchmark one

carols10cents: Ah, yep-- I just pushed a commit to exclude those workspace crates that are only benchmarks or tests (aside from query_test that we do want to share dependencies), and that removed criterion.

shepmaster:
> Is cargo smart enough to know that these aren't real dependencies - i.e. the code isn't actually used?

One concrete equivalent may be the playground. That has [250+ crates that are always specified in the Cargo.toml](https://github.com/integer32llc/rust-playground/blob/master/compiler/base/Cargo.toml). Compiling / linking an empty `fn main() {}` takes one second on an EC2 t2.nano:

```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 1.06s
     Running `target/debug/playground`
```

> crates with fewer dependencies compile faster,

I'd expect that to be true, but mostly because of the total amount of code crunched, not the raw number of dependencies.

> Perhaps there are some heavy hitters like tokio, that could benefit from this treatment, but not others?

I expect that taking the workspace hack to the logical extreme would ultimately end up with "shim crates", (e.g. `workspace-hack-tokio`, `workspace-hack-rand`, etc.). Any crate that depends on `tokio`, directly or indirectly, would be modified to depend on that shim crate. That would allow a crate lower down in the dependency tree that doesn't need tokio to completely skip compiling tokio, while still only compiling tokio once at the higher levels of the tree.

I don't know if hakari has support for that fine-grained of support or not. There's also a question of how valuable that path would be for The Average Iox Developer™. If we expect that most Iox developers are going to *usually* build something high up in the dependency tree, then it doesn't matter _as much_ if they build tokio "too early".

crepererum:
If I understand this correctly the tradeoff here is:

_We build more stuff upfront for the entire workspace (aka more crates and features) but in return we get fast task switching within our workspace._

Important here to note is that we don't update dependencies that often. Mostly on Mondays (dependabot PRs) and then once or twice during the week (e.g. datafusion, arrow or some major dependencies). Taken this into account I think it's absolutely worth it.

.circleci/config.yml
```diff
@@ -137,6 +138,19 @@ jobs:
...
       - store_artifacts:
           path: rustdoc.tar.gz
+  workspace_hack_checks:
+    docker:
+      - image: quay.io/influxdb/rust:ci
+    steps:
+      - checkout
+      - rust_components
+      - cache_restore
+      - run:
```
crepererum, line 148: Maybe add some `echo "In case this fails run ... locally"` so people know how to recover from a failing check?

crepererum (approved):

.guppy/hakari.toml
```diff
@@ -0,0 +1,38 @@
...
+# Write out exact versions rather than a semver range. (Defaults to false.)
+# exact-versions = true
+
+# Don't search in these crates for dependencies, and don't have these crates depend on the
+# workspace-hack crate.
+#
+# Includes most bench- or test-only crates except for query_tests, as that crate is built often
+# and should share as many dependencies as possible.
+[traversal-excludes]
+workspace-members = [
```
tustvold, line 29: I think we should possibly also exclude the client crates, otherwise I think conductor will end up depending on IOx's workspace hack which seems off?

tustvold (approved):
> I expect that taking the workspace hack to the logical extreme would ultimately end up with "shim crates"

Oh god no, that would be horrible, and complete overkill :laughing: I more meant that there are some crates in there, such as clap, cache_loader_async, tracing-log, tracing-subscriber, etc... that surprise me if they're not just a case of fixing the feature flags across the workspace to be consistent. But I guess maybe that is for a followup.

> Compiling / linking an empty fn main() {} takes one second on an EC2 t2.nano:

Neat! I'll keep trying to reduce transitive coupling within the workspace, as that will still help iteration speed, but good to know the number of dependencies doesn't directly impact iterative compilation :+1: Still want to get the number of dependencies down though :laughing:

.guppy/hakari.toml
carols10cents: Looks like conductor is depending on [influxdb_iox_client](https://github.com/influxdata/conductor/blob/28f300bb00208037a4fc8feee0478868e1c2ec86/lib/node/Cargo.toml#L16) and [trogging](https://github.com/influxdata/conductor/blob/28f300bb00208037a4fc8feee0478868e1c2ec86/conductor_cli/Cargo.toml#L41), so I've excluded those.

.circleci/config.yml
carols10cents: Great idea, done!

workspace-hack/README.md
```diff
@@ -0,0 +1,21 @@
...
+documentation][hakari-docs].
+
+[hakari]: https://crates.io/crates/cargo-hakari
+[hakari-docs]: https://docs.rs/cargo-hakari/0.9.6/cargo_hakari/about/index.html
+
+## CI failures
+
+If the `workspace_hack_checks` CI job is failing, there are two possible reasons and solutions:
+
+- If `cargo hakari generate --diff` fails, that means a crate has started or stopped using a
```
alamb, line 16: 👨‍🍳 💋

sunshowers:
Hey, I'm the author of `cargo hakari` and this bubbled up -- glad y'all are using this! Always happy to answer questions on the [cargo-guppy repo](https://github.com/facebookincubator/cargo-guppy/issues/new) :)

PR #3169, opened 2021-11-19
feat: reduce cardinality of job metrics

tustvold:
The size of the metrics payloads produced by IOx are causing issues for our operational tooling.

As pointed out by @mkmik the job metrics are orders of magnitude higher cardinality than anything else

```
; awk -F'{' </tmp/foo  '{print $1}' | sort | uniq -c | sort -n| tail
 648 write_buffer_ingest_request_duration_seconds_bucket
4680 catalog_loaded_chunks
4680 catalog_loaded_rows
6039 influxdb_iox_job_completed_cpu_seconds_count
6039 influxdb_iox_job_completed_cpu_seconds_sum
6039 influxdb_iox_job_completed_wall_seconds_count
6039 influxdb_iox_job_completed_wall_seconds_sum
9488 influxdb_iox_job_count
144936 influxdb_iox_job_completed_cpu_seconds_bucket
144936 influxdb_iox_job_completed_wall_seconds_bucket
```

Some buckets are also duplicated

This PR fixes the duplicate buckets and removes table name from the attributes. This should significantly reduce the size of the metrics payloads. The job metadata is accessible as a system table, and so there is still a mechanism to view the full cardinality data (with limited historical recall) should an operator so wish

Related to #3013


server/src/job.rs
```diff
@@ -134,10 +134,6 @@ impl JobRegistryMetrics {
             Duration::from_millis(100),
```
alamb, line 134: I recommend even fewer buckets (until we have a specific usecase for this level of granularity)
```
10ms
100ms
1s
10s
100s
INF
```

alamb (approved):

mkmik (approved):

server/src/job.rs
mkmik: I'm fine with having a couple of buckets for very long running jobs. But we don't need a very high precision there

tustvold: Will update

PR #3170, opened 2021-11-19
fix: Clean up dependencies based on what's in the workspace-hack crate

### LINKED ISSUE #3117: Sync features of dependencies + dev-dependencies to avoid rebuilds between build/test

carols10cents:
The workspace hack crate lists all dependencies that use different feature sets in different parts of the dependency tree (in order to unify them). I looked at all the crates and features listed to make sure all the crates, features, and feature set variations really needed to be there.

I indeed found a number of inconsistencies, changes we could make, updates, crates that should be excluded from the workspace crate analysis because their dependency tree is more separate, and lots more!

I recommend looking at this PR commit-by-commit as I tried to document the reason for each change in the commit message.

Fixes #3117.

Regarding the commit `fix: In end-to-end tests, use influxdb_iox_client instead of influxdb2_client`, Paul has approved in chat:

>  I guess I wouldn't expect to see IOx depend on influxdb2_client. As long as IOx has a legit integration test to ensure it supports the InfluxDB v2 write and delete APIs, that's enough. All other API that IOx implements is outside the scope of what the influxdb2_client would ever support (i.e. it's all API that is either IOx native API or the storage gRPC API).

There are tests that use `/api/v2/write` and `/api/v2/delete` in https://github.com/influxdata/influxdb_iox/blob/0517adf7528d8d96a6ffa63ec998526dba3a0c0b/influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs#L418-L545


Cargo.lock
```diff
@@ -4972,15 +4899,6 @@ dependencies = [
  "webpki",
```
alamb, line 4899: The number of dependencies removed in `Cargo.lock` is truly impressive 👍

alamb (approved):
I went through this PR commit by commit -- very impressive @carols10cents ❤️ thank you

crepererum (approved):

PR #3185, opened 2021-11-23
feat: compute max checkpoints from chunks' IOx Metadata and tests for contiguous chunks

NGA-TRAN:
Another part of #3089. Two different pieces:
  1. Read IoxMetadata from Parquet file to get the chunk checkpoints
  2. Modify contiguous_chunks to have it work with duplicated orders and add tests for it


server/src/db/catalog/partition.rs
```diff
@@ -374,31 +386,49 @@ impl Partition {
 
     /// Return true if there are no other persisted chunks that are in the middle of
     /// the provided chunk orders
-    // NGA todo: There is test_compact_os_non_contiguous_chunks in
-    // compact_object_store.rs to test this but I will add more unit tests right here
-    // when PR #3167 ChunkGenerator is merged
-    pub fn contiguous_object_store_chunks(&self, chunk_orders: &BTreeSet<ChunkOrder>) -> bool {
-        // Last order in the chunk_orders for comparison
+    pub fn contiguous_chunks(
```
NGA-TRAN, line 389: Thanks @tustvold to point out:
  . We have to consider all chunks, not only the persisted one.
  . Chunk order can be duplicate

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -268,6 +281,26 @@ struct CompactingOsChunks {
     partition_checkpoint: PartitionCheckpoint,
 }
 
+/// Check points of this creating chunk will have min range None and max range is the max of all max
+//  Todo: ask Marco and Raphael: How about sequencer_number of there are may of them?
```
NGA-TRAN, line 285: @crepererum and @tustvold : See if my statement for min_max range is correct and I also have a question above

tustvold: I'm not sure I follow what this comment is saying. It should just take the maximum PartitionCheckpoint and DatabaseCheckpoint from the chunks.

PartitionCheckpoint already implements partial_cmp (and chunks within the same partition should always be comparable), and DatabaseCheckpoint provides a fold method that can be used

server/src/db/catalog/chunk.rs
```diff
@@ -732,6 +732,16 @@ impl CatalogChunk {
         }
     }
 
+    /// Return parquet chunk if it is persisted
+    pub fn parquet_chunk(&self) -> Option<Arc<ParquetChunk>> {
```
tustvold, line 736:
```suggestion
    pub fn parquet_chunk(&self) -> Option<&Arc<ParquetChunk>> {
```

Pet peeve of mine, but where there is no reason to do so, it is better to return & vs a cloned reference - this allows borrow chaining in e.g. map to work correctly, otherwise it complains about a dropped temporary.

(there is also the minor point of avoiding unnecessary atomic operations)

server/src/db/catalog/partition.rs
```diff
@@ -374,31 +386,44 @@ impl Partition {
...
+        &self,
+        chunk_ids: &BTreeSet<ChunkId>,
+        chunk_orders: &BTreeSet<ChunkOrder>,
+    ) -> Result<bool> {
+        if chunk_orders.is_empty() {
+            return Ok(true);
+        }
+
+        // First and last order in the chunk_orders for comparison
+        let first_order_element = chunk_orders.iter().next();
```
tustvold, line 399: The line above has verified that `chunk_orders` is not empty and so I think this could just call unwrap and not bother with the additional error plumbing

server/src/db/catalog/partition.rs
```diff
@@ -374,31 +386,44 @@ impl Partition {
...
     /// the provided chunk orders
-    // NGA todo: There is test_compact_os_non_contiguous_chunks in
-    // compact_object_store.rs to test this but I will add more unit tests right here
-    // when PR #3167 ChunkGenerator is merged
-    pub fn contiguous_object_store_chunks(&self, chunk_orders: &BTreeSet<ChunkOrder>) -> bool {
-        // Last order in the chunk_orders for comparison
+    pub fn contiguous_chunks(
+        &self,
+        chunk_ids: &BTreeSet<ChunkId>,
+        chunk_orders: &BTreeSet<ChunkOrder>,
```
tustvold, line 392:
```suggestion
        orders: RangeInclusive<ChunkOrder>,
```

The code path calling this is already computing the min order, it seems unnecessary to build up a BTreeSet only to generate a min and max

server/src/db/catalog/partition.rs
```diff
@@ -374,31 +386,44 @@ impl Partition {
...
         let chunks = self.chunks();
         for chunk in chunks {
             let chunk = chunk.read();
-            if chunk.is_persisted() {
-                let order = chunk.order();
-                // this chunk does not belong to chunk_orders but in the middle of them
-                if !chunk_orders.contains(&order) && order < *last_order {
-                    return false;
-                }
+            let order = chunk.order();
```
tustvold, line 417: Using `self.keyed_chunks()` instead of `self.chunks()` would avoid the need for additional locking

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -224,21 +219,39 @@ fn mark_chunks_to_compact(
...
+            // read IoxMetadata from the parquet chunk's  metadata
+            if let Some(parquet_chunk) = chunk.parquet_chunk() {
+                let iox_parquet_metadata = parquet_chunk.parquet_metadata();
+                let iox_metadata = iox_parquet_metadata
+                    .decode()
+                    .context(ParquetMetaRead)?
+                    .read_iox_metadata()
+                    .context(ParquetMetaRead)?;
+
+                db_ckpts.push(iox_metadata.database_checkpoint);
```
tustvold, line 233: I'm not sure I understand the need to buffer up all the partition and database checkpoints to just compute a maximum for both?

server/src/db/catalog/partition.rs
```diff
@@ -211,11 +214,20 @@ impl Partition {
...
     ) -> &Arc<RwLock<CatalogChunk>> {
-        assert_eq!(chunk.table_name().as_ref(), self.table_name());
-
         let chunk_id = ChunkId::new();
         let chunk_order = self.next_chunk_order();
 
+        self.create_open_chunk_with_specified_id_order(chunk, chunk_id, chunk_order)
+    }
+
+    pub fn create_open_chunk_with_specified_id_order(
```
tustvold, line 223:
```suggestion
    fn create_open_chunk_with_specified_id_order(
```

server/src/db/catalog/partition.rs
```diff
@@ -511,6 +536,150 @@ mod tests {
...
+        ids.insert(ChunkId::new_test(3));
+        ids.insert(ChunkId::new_test(1));
+        let mut orders = BTreeSet::new();
+        orders.insert(ChunkOrder::new(3).unwrap());
+        orders.insert(ChunkOrder::new(1).unwrap());
+        assert!(!partition.contiguous_chunks(&ids, &orders).unwrap());
+    }
+
+    #[test]
+    fn test_contiguous_chunks_three_chunk_with_duplicated_orders() {
```
tustvold, line 615: :ok_hand:

NGA-TRAN:
@tustvold Thanks for your comments and hints, I have made this with few lines and also computed max checkpoints easily. Can you have a quick look? If nothing critical, I would like to merge this today before my long weekend break.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -223,19 +221,60 @@ fn mark_chunks_to_compact(
             delete_predicates.extend(chunk.delete_predicates().iter().cloned());
 
             min_order = min_order.min(chunk.order());
+            max_order = max_order.max(chunk.order());
```
tustvold, line 224: Is chunk_orders still used?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -184,18 +186,14 @@ fn mark_chunks_to_compact(
...
 
-    // initialize checkpoints
-    let database_checkpoint = DatabaseCheckpoint::new(Default::default());
-    let partition_checkpoint = PartitionCheckpoint::new(
-        Arc::clone(&partition_addr.table_name),
-        Arc::clone(&partition_addr.partition_key),
-        Default::default(),
-        Time::from_timestamp_nanos(0),
-    );
+    let mut database_checkpoint: Option<DatabaseCheckpoint> = None;
```
tustvold, line 195: This could just be DatabaseCheckpoint::default FWIW

server/src/db/lifecycle/error.rs
```diff
@@ -71,8 +71,27 @@ pub enum Error {
...
+    #[snafu(display(
+        "Error reading IOx Metadata from Parquet IoxParquetMetadata: {}",
+        source
+    ))]
+    ParquetMetaRead {
+        source: parquet_file::metadata::Error,
+    },
+
+    #[snafu(display(
+        "Cannot compact chunks because of error during computing max partition checkpoint"
```
tustvold, line 89:
```suggestion
        "Cannot compact chunks because of error computing max partition checkpoint"
```

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -223,19 +221,60 @@ fn mark_chunks_to_compact(
...
 
             // Set chunk in the right action which is compacting object store
             // This function will also error out if the chunk is not yet persisted
             chunk.set_compacting_object_store(registration)?;
-            Ok(DbChunk::snapshot(&*chunk))
+            Ok(DbChunk::parquet_file_snapshot(&*chunk))
         })
         .collect::<Result<Vec<_>>>()?;
 
+    if database_checkpoint.is_none() || partition_checkpoint.is_none() {
```
tustvold, line 269: I don't think this error case can occur - i.e. you can just use unwrap. You error if any single chunk lacks a checkpoint, and you've already checked that there are a non-zero number of chunks

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -223,19 +221,60 @@ fn mark_chunks_to_compact(
...
+                        Some(Ordering::Less) => {
+                            partition_checkpoint = Some(iox_metadata.partition_checkpoint);
+                        }
+                        _ => {}
+                    }
+                } else {
+                    partition_checkpoint = Some(iox_metadata.partition_checkpoint);
+                }
+            } else {
+                return ChunksNotPersisted {}.fail();
```
tustvold, line 259: You could make `set_compacting_object_store` return `Result<&ParquetChunk>` and use that to avoid this additional error variant but :shrug:

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -223,19 +221,60 @@ fn mark_chunks_to_compact(
...
+                // fold all database_checkpoints into one for the compacting chunk
+                if let Some(db_ckpt) = &mut database_checkpoint {
+                    db_ckpt.fold(&iox_metadata.database_checkpoint);
+                } else {
+                    database_checkpoint = Some(iox_metadata.database_checkpoint);
+                }
+
+                // keep max partition_checkpoint for the compacting chunk
+                if let Some(part_ckpt) = &partition_checkpoint {
+                    match part_ckpt.partial_cmp(&iox_metadata.partition_checkpoint) {
```
tustvold, line 246: You could use `snafu::OptionExt` here

tustvold (approved):
Just some minor nits left

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: I though the same and did actually use unwrap but after a second thought I added this check here to capture some silly bugs we humans inevitably introduce especially if some code added in the future

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: Brain is not working to digest this :). Skip to then

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: Good catch

Issue #2286, opened 2021-08-13
Reduce lifecycle log verbosity


tustvold:
`LifecyclePolicy` is producing a substantial amount of logs that can make it very hard to see what is going on in a running IOx instance. Part of this might be due to the fact we are overriding the worker backoff time (I think to 100ms) and so the loop runs more frequently than it might otherwise.

A lot of the time, however, it is logging the exact same message, e.g. "persisting partition as exceeds row threshold" or "Chunk to persist has outstanding action" ideally it would only log these sort of messages once until the condition clears.

For example, this code section is logging the fact that the threshold is exceeded before checking if it has already scheduled work to address this

```
if persistable_row_count >= rules.persist_row_threshold.get() {
    info!(%db_name, %partition, persistable_row_count, "persisting partition as exceeds row threshold");
} else if persistable_age_seconds >= rules.persist_age_threshold_seconds.get() {
    info!(%db_name, %partition, persistable_age_seconds, "persisting partition as exceeds age threshold");
} else {
    debug!(%db_name, %partition, persistable_row_count, "partition not eligible for persist");
    return false;
}

let mut chunks = LockablePartition::chunks(&partition);

// Upgrade partition to be able to rotate persistence windows
let mut partition = partition.upgrade();

let persist_handle = match LockablePartition::prepare_persist(&mut partition, now) {
    Some(x) => x,
    None => {
        debug!(%db_name, %partition, "no persistable windows or previous outstanding persist");
        return false;
    }
};
```

PR #3189, opened 2021-11-23
refactor: lifecycle observability cleanup (#2286)

tustvold:
Some minor cleanup in advance of a PR to reduce the verbosity of the lifecycle policy's logging (#2286)


lifecycle/src/policy.rs
```diff
@@ -459,7 +459,7 @@ where
             let stall_compaction_persisting = if rules.persist && !self.suppress_persistence {
                 let persisting = self.maybe_persist_chunks(&db_name, partition, &rules, now);
                 if persisting {
-                    debug!(%db_name, %partition, reason="persisting", "stalling compaction");
+                    debug!(%db_name, partition=%partition.read(), reason="persisting", "stalling compaction");
```
tustvold, line 462: This will get reworked as part of #3189, but is also unlikely to matter in practice

lifecycle/src/guard.rs
```diff
@@ -129,15 +130,9 @@ impl<'a, P, D> LifecycleReadGuard<'a, P, D> {
     }
 }
 
-impl<'a, P, D> Debug for LifecycleReadGuard<'a, P, D> {
+impl<'a, P: Display, D> Display for LifecycleReadGuard<'a, P, D> {
```
tustvold, line 133: `LifecycleReadGuard` and `LifecycleWriteGuard` are changed to use the display implementation of the locked payload, e.g. `Partition` or `Chunk`, and not the wrapper type - e.g. `LockableCatalogChunk`

lifecycle/src/guard.rs
```diff
@@ -179,14 +175,9 @@ impl<'a, P, D> LifecycleWriteGuard<'a, P, D> {
...
-impl<'a, P, D> Debug for LifecycleWriteGuard<'a, P, D> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "LifecycleWriteGuard{{..}}")
-    }
-}
-impl<'a, P, D: Display> Display for LifecycleWriteGuard<'a, P, D> {
+impl<'a, P: Display, D> Display for LifecycleWriteGuard<'a, P, D> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{} locked for write", self.data)
+        self.guard.fmt(f)
```
tustvold, line 180: The "locked for write" caused confusion for people when interpreting the log lines, so best to just remove - it should be clear from the line number what type of lock it is

server/src/db/lifecycle.rs
```diff
@@ -123,30 +121,11 @@ impl lifecycle::PersistHandle for CatalogPersistHandle {
 pub struct LockableCatalogPartition {
     pub db: Arc<Db>,
     pub partition: Arc<RwLock<Partition>>,
-    /// Human readable description of what this CatalogPartiton is
-    pub display_string: String,
 }
 
 impl LockableCatalogPartition {
     pub fn new(db: Arc<Db>, partition: Arc<RwLock<Partition>>) -> Self {
-        let display_string = {
```
tustvold, line 132: This is best described as a hack, as part of reducing the lifecycle log verbosity I need to obtain the PartitionAddr instead of a display string. I experimented with putting a PartitionAddr on LockableCatalogPartition but it turned into a monster PR, and so decided against it. The two log lines that don't already have a locked partition, will just have to re-lock it

server/src/db/lifecycle/persist.rs
```diff
@@ -271,36 +271,35 @@ mod tests {
...
-        persist_chunks(partition, chunks, handle)
+        persist_chunks(partition_guard, chunks, handle)
             .unwrap()
             .1
             .await
             .unwrap()
             .unwrap();
 
-        assert!(db_partition
+        assert!(partition
```
tustvold, line 302: Technically this is now acquiring a different type of read lock (upgradeable vs normal) but this is a test so this doesn't matter

server/src/db/lifecycle/compact.rs
```diff
@@ -191,22 +191,20 @@ mod tests {
         let partition_keys = db.partition_keys().unwrap();
         assert_eq!(partition_keys.len(), 1);
 
-        let db_partition = db.partition("cpu", &partition_keys[0]).unwrap();
+        let partition = db.lockable_partition("cpu", &partition_keys[0]).unwrap();
```
tustvold, line 194: The long term "vision" if you will is that these Lockable* become the API for performing operations on Db, and the current member functions are just extracted as test helpers. This is not a fully formed vision, however

crepererum:
Could you actually document the outcome of this PR somewhere (PR description or commit message)? Aka: how does this affect the logging output?

tustvold:
It's primarily a refactor PR, but the following lines will be changed

```
level=info msg="persisting partition as exceeds age threshold" db_name=dd1a42c069c58fc4_a0f19879ab88c086 partition="Partition('dd1a42c069c58fc4_a0f19879ab88c086':'total_count_metric':'2021-11-23 11:00:00') locked for read" persistable_age_seconds=1800 target="lifecycle::policy" location="lifecycle/src/policy.rs:368" time=1637673912103230439
```

To

```
level=info msg="persisting partition as exceeds age threshold" db_name=dd1a42c069c58fc4_a0f19879ab88c086 partition="Partition('dd1a42c069c58fc4_a0f19879ab88c086':'total_count_metric':'2021-11-23 11:00:00') locked for read" persistable_age_seconds=1800 target="lifecycle::policy" location="lifecycle/src/policy.rs:368" time=1637673912103230439
```

i.e. the "locked for read" and "locked for write" output is removed. Everything else is unchanged

lifecycle/src/guard.rs
alamb: I am not sure I will remember to look at the line number, but if the ambiguity causes me confusion I'll propose something different

lifecycle/src/policy.rs
alamb: the concern is that this change now acquires another read lock?

The only thing I would double check is that this doesn't inadvertently cause a deadlock when `DEBUG` logging is enabled (by trying to acquire a read lock when a write lock is already held, for example). 

I may have done that in the past 😢  (aka turning on debug logging ended up causing deadlocks)

server/src/db/lifecycle.rs
alamb: As long as the change to relock the chunk as part of logging doesn't cause a deadlock or some other untoward behavior, the change in this PR seems like an improvement to me

alamb (approved):
Please test that doing `RUST_LOG=debug cargo test all` doesn't deadlock. 

Otherwise this looks like a nice set of improvements to me

lifecycle/src/policy.rs
tustvold: If it could, it would deadlock the lifecycle loop as the very first thing `maybe_compact_chunks` does (which follows this) is to acquire the same lock :sweat_smile:

alamb: > it would deadlock the lifecycle loop as the very first thing

with the assumption that `debug!` logging was enabled -- if logging wasn't enabled it might be a lurking timebomb

tustvold: You misunderstand. If there were a lock held in this context that would deadlock calling partition.read(), it would deadlock in the below call regardless of whether debug logging is enabled. i.e. enabling debug would just change the line of the deadlock :laughing:

PR #3191, opened 2021-11-23
feat: HTTP delete support for router

### LINKED ISSUE #3154: Delete Routing
**Problem**

Once we have support for deletes within the write buffer fully fleshed out #2731, the next obvious question is how do writes get there.

Currently deletes from both the HTTP and gRPC endpoints are written directly to `Db::delete`, this means they skip any routing or sharding logic and are never written to the write buffer. Currently if a delete request is made to a router, it is effectively a no-op.

**Proposal**

Hook up deletes with the new routing abstractions being added in #2980

crepererum:
~On top of #3187.~

Closes #3154.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/http/dml.rs
+++ b/influxdb_iox/src/influxdb_ioxd/http/dml.rs
@@ -27,27 +27,15 @@ pub enum HttpDmlError {
     #[snafu(display("Internal error mapping org & bucket: {}", source))]
     BucketMappingError { source: OrgBucketMappingError },
 
-    #[snafu(display(
-        "User error writing points into org {}, bucket {}:  {}",
-        org,
-        bucket_name,
-        source
-    ))]
+    #[snafu(display("User error writing points into {}:  {}", db_name, source))]
     WritingPointsUser {
-        org: String,
-        bucket_name: String,
+        db_name: String,
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
-    #[snafu(display(
-        "Internal error writing points into org {}, bucket {}:  {}",
-        org,
-        bucket_name,
-        source
-    ))]
+    #[snafu(display("Internal error writing points into {}:  {}", db_name, source))]
     WritingPointsInternal {
-        org: String,
-        bucket_name: String,
+        db_name: String,
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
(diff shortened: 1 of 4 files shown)


influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -55,37 +93,66 @@ pub enum HttpWriteError {
...
+            e @ Self::ParsingDelete { .. } => e.invalid(),
+            e @ Self::BuildingDeletePredicate { .. } => e.invalid(),
         }
     }
 }
 
 /// Write error when calling the underlying server type.
 #[derive(Debug, Snafu)]
-pub enum InnerWriteError {
+pub enum InnerDmlError {
```
crepererum, line 142: I've merged writes and deletes into a single DML HTTP trait.

influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -180,14 +247,23 @@ pub trait HttpDrivenWrite: ServerType {
                         .unwrap(),
                 ))
             }
-            Err(InnerWriteError::NotFound { .. }) => {
+            Err(InnerDmlError::DatabaseNotFound { db_name }) => {
```
tustvold, line 250: Perhaps we could extract this InnerDMLError -> HttpDmlError into a function, or even better a `From` implementation?

influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -21,17 +23,53 @@ use super::{
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Snafu)]
-pub enum HttpWriteError {
+pub enum HttpDmlError {
```
tustvold, line 26: I'm not sure I follow the reasoning for splitting out separate Writing and Deleting variants for the same class of error? If there were ambiguity I'd have expected the caller to add this context?

influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -513,6 +712,90 @@ pub mod test_utils {
         }
     }
 
+    pub async fn assert_delete_unknown_database<T>(test_server: TestServer<T>)
```
tustvold, line 715: Could we have some doc comments?

tustvold (approved):
I think there are ways the amount of error plumbing could be reduced, but otherwise this does what it says on the tin

influxdb_iox/src/influxdb_ioxd/http/dml.rs
crepererum: Mostly because `snafu` is driving me mad. Some variants are shared and otherwise I would have to come up w/ really creative variant names.

influxdb_iox/src/influxdb_ioxd/http/dml.rs
crepererum: done

influxdb_iox/src/influxdb_ioxd/http/dml.rs
crepererum: done

influxdb_iox/src/influxdb_ioxd/http/dml.rs
tustvold: :laughing: I know the feeling, this is fine - at some point I expect we will go on a great error cleanup, but its hardly a high priority

Issue #3190, opened 2021-11-23
IOx not returning tag sets keys in canonical order


e-dard:
When returning results for the gRPC API IOx returns frames containing series keys. These series keys contains tags including special ones for the measurement and the field.

In the TSM data model measurements and field keys are represented by special tag keys `\x00` and `\xff` respectively. These byte values were chosen so that they would always be the first and last value in a series key. This has the nice property that they can be prepended and appended to an arbitrary (sorted) series key.

IOx has been correctly returning series keys ordered lexicographically by their tag sets. However, it has been using `_measurement` and `_field` as the tag keys for measurement and field. The Flux engine can interpret these tag keys, but the InfluxQL engine generally wants the binary format. Since the ordering of `_measurement`/`_field` and `\x00`/`\xff` differs, converting from one representation to the other will involve an extra sort.

We should just emit series keys ordered by the byte representation of the special tag key. This would mean we would return keys like:

```
\x00=cpu,host=foo,server=a,\xff=temp
```

or 

```
_measurement=cpu,host=foo,server=a,_field=temp
```

Issue #3165, opened 2021-11-19
Teach IOx to sometimes return binary tag keys 


e-dard:
The internal InfluxQL test suite is currently failing against IOx because in some cases it expects the `_measurement` and `_field` keys to be returned in a different format.

See https://github.com/influxdata/influxdb_iox/issues/3164

Once we know how to interpret this new field we need to teach the IOx gRPC server to rewrite the `_measurement` and `_field` keys where necessary.

For example, instead of returning frames like:

```
Response gRPC frames frames=
SeriesFrame, 
    tags: _field=data,_measurement=test1, type: 1

IntegerPointsFrame, timestamps: [1586520000000000000, 1587384000000000000, 1589112000000000000, 1589976000000000000], values: "1,2,3,4"
```

We instead need to return frames like

```
Response gRPC frames frames=
SeriesFrame, 
    tags: \x00=test1,\xff=data, type: 1

IntegerPointsFrame, timestamps: [1586520000000000000, 1587384000000000000, 1589112000000000000, 1589976000000000000], values: "1,2,3,4"
```

### Definition of Done

- IOx gRPC API returns the data in the format expected by the caller. 
- Multiple internal InfluxQL tests start to pass.
- No regressions to the Flux test suite.

e-dard:
Blocked on https://github.com/influxdata/influxdb_iox/issues/3165

PR #3193, opened 2021-11-23
refactor: update tag set ordering for gRPC API

e-dard:
Closes https://github.com/influxdata/influxdb_iox/issues/3190

Towards: https://github.com/influxdata/influxdb_iox/issues/3165

## Rationale

When returning results for the gRPC API IOx returns frames containing series keys. These series keys contains tags including special ones for the measurement and the field.

In the TSM data model measurements and field keys are represented by special tag keys `\x00` and `\xff` respectively. These byte values were chosen so that they would always be the first and last value in a series key. This has the nice property that they can be prepended and appended to an arbitrary (sorted) series key.

IOx has been correctly returning series keys ordered lexicographically by their tag sets. However, it has been using `_measurement` and `_field` as the tag keys for measurement and field. The Flux engine can interpret these tag keys, but the InfluxQL engine generally wants the binary format. Since the ordering of `_measurement`/`_field`and `\x00`/`\xff` differs, converting from one representation to the other will involve an extra sort.

This PR changes how the tag sets of series keys are ordered.

Instead of returning series keys ordered like this: `_field=temp,_measurement=cpu,host=foo,server=a` we will now return them ordered like this: `_measurement=cpu,host=foo,server=a,_field=temp`.

This has the benefit that when we need to return them using their binary format we can *rewrite the series keys* rather than having to re-sort the tags.

## Testing

I have re-ran the internal Flux test suite and there are no regressions.


crepererum (approved):

query/src/exec/seriesset/series.rs
```diff
@@ -250,26 +253,35 @@ impl SeriesSet {
...
-            Tag {
-                key: "_field".into(),
-                value: field_name.into(),
-            },
-            Tag {
-                key: "_measurement".into(),
-                value: Arc::clone(&self.table_name),
-            },
-        ];
+        // N.B., in order to emit series sets in the same "tag order" as they
```
e-dard, line 256: Here is the behaviour change. The rest of the PR is test case updates.

query/src/exec/seriesset/series.rs
```diff
@@ -250,26 +253,35 @@ impl SeriesSet {
...
+        // correct lexicographical order, in a TSM data-model these tags are
+        // actually stored as `\x00` and `\xff` respectively. Therefore the
+        // expectation is that "_measurement" is emitted first and "_field"
+        // last.
+        //
+        // This also ensures that the output will be sorted first by
+        // "_measurement" and then "_field" even when there are no groups
+        // requested.
+
+        // Prepend key with "_measurement"
```
alamb, line 268: Very similar to how the code for GroupFrames works: 👍 

https://github.com/influxdata/influxdb_iox/blob/94368c339c4d033622042d48a14a9e5735c50bc8/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/data.rs#L43-L52

alamb (approved):

PR #3194, opened 2021-11-23
refactor: use serialized predicates instead of strings for gRPC deletes

### LINKED ISSUE #3192: gRPC `DeleteRequest` should take a parsed predicate, not a string
This

https://github.com/influxdata/influxdb_iox/blob/50e9e02ff7242a0b3069e777bc9217fa11620c32/generated_types/protos/influxdata/iox/delete/v1/service.proto#L28-L30

Should use this type here:

https://github.com/influxdata/influxdb_iox/blob/50e9e02ff7242a0b3069e777bc9217fa11620c32/generated_types/protos/influxdata/iox/predicate/v1/predicate.proto#L4-L5

crepererum:
IOx is the only consumer of this API so we might just use the serialized
form. Cloud2 uses the HTTP API which supports SQL-like predicates.

Fixes #3192.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/delete.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/delete.rs
@@ -1,11 +1,12 @@
 use std::fmt::Debug;
 use std::sync::Arc;
 
+use data_types::non_empty::NonEmptyString;
 use data_types::DatabaseName;
-use generated_types::google::FieldViolationExt;
+use dml::{DmlDelete, DmlMeta};
+use generated_types::google::{FieldViolationExt, FromOptionalField, OptionalField};
 use generated_types::influxdata::iox::delete::v1::*;
-use predicate::delete_predicate::parse_delete_predicate;
-use server::{connection::ConnectionManager, Error, Server};
+use server::{connection::ConnectionManager, Server};
 use tonic::Response;
 
 struct DeleteService<M: ConnectionManager> {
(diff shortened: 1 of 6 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/delete.rs
```diff
@@ -23,13 +22,13 @@ where
...
-        let DeleteRequest {
+        let DeleteRequest { payload } = request.into_inner();
+        let DeletePayload {
             db_name,
             table_name,
-            start_time,
-            stop_time,
             predicate,
-        } = request.into_inner();
+        } = payload.unwrap_field("payload")?;
```
tustvold, line 30: :heart:

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/delete.rs
```diff
@@ -38,20 +37,17 @@ where
...
-                    stop_time,
-                    predicate,
-                }))
-            }
-            Ok(del_predicate) => {
-                // execute delete
-                db.delete(&table_name, Arc::new(del_predicate))
-                    .map_err(default_db_error_handler)?;
-            }
+        let tables = if table_name.is_empty() {
```
tustvold, line 40: You could consider using `db.store_delete` which would avoid duplicating this logic

router/src/grpc_client.rs
```diff
@@ -74,9 +74,7 @@ impl GrpcClient for RealClient {
                             .table_name()
                             .map(|s| s.to_owned())
                             .unwrap_or_default(),
-                        delete.predicate().range.start.to_string(),
```
tustvold, line 77: :tada:

tustvold (approved):

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/delete.rs
crepererum: Didn't know that exists, will give it a try...

crepererum: done

PR #3196, opened 2021-11-23
feat: reduce lifecycle log verbosity (#2286)

### LINKED ISSUE #2286: Reduce lifecycle log verbosity

tustvold:
So I got a fair way through implementing a reporting setup that would aggregate the data from multiple iterations of the policy loop and periodically report summaries - but in the end I decided to abandon this. We already have a fair amount of observability into what the lifecycle decided is doing from the operations system (which logs, produces metrics and has a system table), and from the chunks system table and corresponding metrics. 

I think this is sufficient for most practical purposes, I have therefore lowered the level of the logging so it will be skipped by default. It would be good to have observability into chunks being dropped, how long the policy loop is taking, etc... but these are best exposed as metrics and not logs (#2819)

Closes #2286


e-dard (approved):

alamb (commented):
I think this is a good pragmatic choice. 👍

PR #3204, opened 2021-11-24
refactor: prepare to remove router functionality from `server` crate

crepererum:
See commits.


influxdb_iox/tests/end_to_end_cases/deployment_cli.rs
```diff
@@ -0,0 +1,49 @@
...
+use assert_cmd::Command;
+use predicates::prelude::*;
+
+#[tokio::test]
+async fn test_server_id_database() {
+    assert_server_id(ServerFixture::create_single_use(ServerType::Database).await).await;
+}
+
+#[tokio::test]
+async fn test_server_id_router() {
```
tustvold, line 12: What does a router server ID actually do?

tustvold (approved):

influxdb_iox/tests/end_to_end_cases/deployment_cli.rs
crepererum: Nothing at the moment, we might use it for logging or to persist router configs (which we currently don't do).

PR #3206, opened 2021-11-24
refactor: clean up write buffer end2end tests

crepererum:
Remove the ones that are covered by `write_pb.rs` or that are obsolete
(i.e. rely on behavior that only exist in the legacy router code).
Migrate the remaining ones to use an actual router.

### DIFF
--- a/influxdb_iox/tests/end_to_end_cases/write_buffer.rs
+++ b/influxdb_iox/tests/end_to_end_cases/write_buffer.rs
@@ -3,11 +3,9 @@ use crate::{
         server_fixture::{ServerFixture, ServerType, TestConfig},
         udp_listener::UdpCapture,
     },
-    end_to_end_cases::scenario::{rand_name, DatabaseBuilder},
+    end_to_end_cases::scenario::{rand_name, wildcard_router_config, DatabaseBuilder},
 };
 use arrow_util::assert_batches_sorted_eq;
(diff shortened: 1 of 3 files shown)


influxdb_iox/tests/end_to_end_cases/write_buffer.rs
```diff
@@ -228,46 +119,6 @@ async fn reads_come_from_write_buffer() {
     check.await.unwrap();
 }
 
-#[tokio::test]
-async fn cant_write_to_db_reading_from_write_buffer() {
```
tustvold, line 232: Doing a search for "it's configured to only read from the " suggests this is the only test for this error handling logic?

crepererum: yes, and this logic will no longer be required for the new router code since you cannot configure it to be reading and writing at the same time.

tustvold: You will still be able to configure a database that is reading from a write buffer, and want to prevent it receiving direct writes? via gRPC?

crepererum: Uh, good point. I should leave this test then

tustvold (approved):

PR #3207, opened 2021-11-24
refactor: remove write routing logic from `Server`

crepererum:
~On top of #3206 and requires #3210.~

 API users should use `Database` instead, there's no need any longer to mirror
this API in `Server`. Note that `Database` is better than `Db` in this
case, because the former can also check if we're writing from a write
buffer and can easily reject unsequenced writes. The `pub` modifiers
were adjusted to make it impossible to write through `Db` directly.

### DIFF
--- a/server/src/resolver.rs
+++ b/server/src/resolver.rs
@@ -14,10 +14,6 @@ impl RemoteTemplate {
         let template = template.into();
         Self { template }
     }
-
-    fn get(&self, id: &ServerId) -> GrpcConnectionString {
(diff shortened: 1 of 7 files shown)


server/src/resolver.rs
```diff
@@ -14,10 +14,6 @@ impl RemoteTemplate {
         let template = template.into();
         Self { template }
     }
-
```
crepererum, line 17: This file gets removed in a follow-up PR. This PR just removes obviously unused code to silence clippy.

tustvold (approved):
:heart:

PR #3211, opened 2021-11-24
feat: test interaction of write buffer and lifecycle (#2811)

### LINKED ISSUE #2811: Persistence Enabled During Replay
Currently persistence is enabled before replay completes, this is problematic as it may lead to replay checkpoints going backwards. The fix is simply to move [this line](https://github.com/influxdata/influxdb_iox/blob/main/server/src/database.rs#L1136) after db.perform_replay. 

However, writing a test is likely to prove somewhat complex. This ticket tracks adding such a test.

tustvold:
This reworks the existing write_buffer_reads_wait_for_compaction into a more general test that also validates that replay waits for compaction, and persistence is disabled during replay

Closes #2811

This test fails if you move `self.lifecycle_worker.unsuppress_persistence();` to before `db.perform_replay` in `DatabaseStateCatalogLoaded::advance`


server/tests/write_buffer_lifecycle.rs
```diff
@@ -0,0 +1,265 @@
...
+    write_group2(&mock_shared_state);
+
+    database.wait_for_init().await.unwrap();
+    let db = database.initialized_db().unwrap();
+
+    let start = Instant::now();
+    loop {
+        if count_persisted_chunks(&db) > 1 {
+            // As soon as replay finishes the lifecycle should have persisted everything in
+            // table_1 into a single chunk
```
crepererum, line 207: > As soon as replay finishes the lifecycle should have persisted everything in `table_1` into a single chunk

Then why are you checking that there are two chunk summaries here?

crepererum (commented):
I don't really understand how phase 3 ensures that we're not persisting during replay.

tustvold:
@crepererum I can't say I blame you, it took me the best part of day to devise this somewhat arcane test :sweat_smile: I've added some more docs that may help?

crepererum (approved):
That makes more sense, thank you!

PR #3212, opened 2021-11-25
refactor: remove delete predicates from read filter interface

tustvold:
There is more cleanup that can be done here, but wanted to get sign off on the change to the public API.

QueryChunk is the interface for the query engine, it is implemented by DbChunk which is a snapshot of a catalog chunk, along with a test implementation. Given the catalog chunk is the owner of the delete predicates, it therefore seems strange that you would need to provide them to the QueryChunk? In my opinion they are an implementation detail of how the specific implementation of QueryChunk decides to handle deletes.

This PR therefore removes this from the interface.


query/src/provider/physical.rs
```diff
@@ -118,22 +118,14 @@ impl<C: QueryChunk + 'static> ExecutionPlan for IOxReadFilterNode<C> {
...
-            .read_filter(&self.predicate, selection, &del_preds)
-            .map_err(|e| {
-                DataFusionError::Execution(format!(
-                    "Error creating scan for table {} chunk {}: {}",
-                    self.table_name,
-                    chunk.id(),
-                    e
-                ))
-            })?;
+        let stream = chunk.read_filter(&self.predicate, selection).map_err(|e| {
```
e-dard, line 121: I think this code highlights the deletes leaking unnecessarily out of the chunk abstraction

e-dard (approved):
This change seems pretty sane to me. Would be good to get @NGA-TRAN's sign off because she has been working on this front and may have reasons for the way the code is organised.

query/src/provider/physical.rs
alamb: As in this PR reduces the leakage or that the PR increases the leakage?

alamb (approved):

query/src/provider/physical.rs
e-dard: Reduces. Before, the delete predicates were being fetched out of the chunk only to be put back into the chunk's `read_filter` implementation.

alamb: Yes I agree this PR encapsulates the deletes more

server/src/db/chunk.rs
```diff
@@ -346,19 +346,23 @@ impl QueryChunk for DbChunk {
...
         debug!(?predicate, "Input Predicate to read_filter");
         self.access_recorder.record_access();
 
-        debug!(?delete_predicates, "Input Delete Predicates to read_filter");
+        let delete_predicates: Vec<_> = self
+            .meta
+            .delete_predicates
+            .iter()
+            .map(|pred| Arc::new(pred.as_ref().clone().into()))
+            .collect();
```
NGA-TRAN, line 361: 👍

NGA-TRAN (commented):
Great catch and thanks for he cleanup, @tustvold

PR #3213, opened 2021-11-25
fix: don't project on schema metadata mismatch

tustvold:
Whilst working on adding the ability to reload a RUB chunk from a parquet file I ran into an issue where a projection node would be added to a plan, because of a mismatch in schema metadata between input and output. This in turn was due to what I think is a separate bug in how the provider determines what the input schema is.

This then combined with https://github.com/apache/arrow-datafusion/issues/1361 to result in the RecordBatches produced by the scan not containing field metadata, which would cause the RUB to panic complaining about missing schema metadata.

Ultimately a projection node is not going to help with a schema mismatch, this must be handled at a higher-level. The only thing this node handles is if there is a column selection that doesn't match the source. This PR therefore adjusts the logic to only enable in such a case


query/src/provider.rs
```diff
@@ -708,8 +708,15 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
         let input_schema = input.schema();
         let output_schema = output_schema.as_arrow();
 
-        // If the schemas are the same, nothing to do
-        if input_schema == output_schema {
+        // If columns are the same, nothing to do
+        let same_columns = input_schema.fields().len() == output_schema.fields().len()
```
e-dard, line 712: Would be good to add a test case that now passes thanks to this change. Perhaps something to do with preserving meta data?

query/src/provider.rs
```diff
@@ -708,8 +708,15 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
...
+            && input_schema
+                .fields()
+                .iter()
+                .zip(output_schema.fields())
+                .all(|(a, b)| a.name() == b.name());
```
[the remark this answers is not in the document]
tustvold: It's an invariant that must be handled higher up (and is) there is no mechanism for querying across chunks with incompatible schemas

query/src/provider.rs
tustvold: Test added, and created ticket for the underlying bug that causes the schema metadata mismatch in the first place (#3214)

Issue #3143, opened 2021-11-17
Perf Remote Target Environment


tustvold:
**Problem**

As part of developing IOx we wish to be able to reproduce or simulate specific workloads in a controlled environment, and potentially incorporate this into a performance regression suite run by CI.

We have the beginnings of this in the form of the perf scripts written by @jacobmarble. However, these are currently limited to running tests on a local machine, which has a couple of implications:

* Performance characteristics may vary across different hardware / OSes
* Cannot easily test the impact of network/disk IO
* Limited to the resources of a single machine
* May be impractical to leave "soaking" whilst getting on with other work
* Cannot easily test different deployment topologies

**Proposal**

Extend `perf` with the concept of a target environment. This would replace the current command line flags that configure aspects such as what object store implementation to use. Instead `perf` would be provided a `toml` file that identifies a deployment target, and provides configuration options specific to that target. 

One deployment target would be `docker-local` which would use the same combination of docker and `./target` currently and with the same options that are currently exposed via CLI.

Another deployment target would be `kustomize` which would take a set of k8s manifests, and use [kustomize](https://kustomize.io/) to apply them to a target cluster. _I don't have a particular affinity for kustomize, I chose it simply because it is built-in to kubectl, it's the flexibility of using a templating tool that matters to me, if people would prefer using jsonnet or helm or whatever_ :shrug:

The expectation would be to end up with a collection of target environment configs checked into git, allowing test batteries to be run against a number of different toplogies and environments, depending on what the developer is trying to simulate/reproduce.

This would also provide an obvious extension point for additional target environments, e.g. `local`, or `podman` or `k8s-jsonnet` or `helm` or `conductor` if we so wished.

e-dard:
I think that we should probably try and align effort around Remocal. Jumping into those details might be best done out of band of the public tracker.

I certainly like the idea of having multiple targets 👍

alamb:
I think it may also be worth considering splitting up the two concerns into two sets of tools:
1. Setting up a IOx cluster/clusters
2. Running a battery of tests (perhaps driving the `data_generator` that @pauldix  has been working on)

Today `perf.py` does both of these pieces -- splitting it up into two (or more) tools might make it easier to customize rather than have one tool that does it all

tustvold:
> I think that we should probably try and align effort around Remocal.

I was hoping to keep the scope down and avoid that particular can of worms by just focusing on orchestrating IOx and not the whole shebang, but happy to discuss.

[...]

PR #3215, opened 2021-11-25
feat: test bench

crepererum:
The functionality is not massive yet, but it's a start.


test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+        echo "cannot set server ID yet, waiting..."
+        sleep 1
+    fi
+done
+
+# wait for rule/config updates
+FINGERPRINT_DONE=/fingerprint.done
+FINGERPRINT_STAGING=/fingerprint.staging
+F_CURRENT=/current
+while true; do
```
crepererum, line 34: I had to move the config sync into the K8s cluster because broadcasting configs to all nodes 1 by 1 -- including the port forwarding -- was just super annoying. This is not the most beautiful or resilient code, but it works for now. Also it expects you to specify the whole gRPC update message the the congigs must be wrapped into a `{"rules": ...}` or `{"router": ...}` before placing them into the config maps.

tustvold: I'm guessing the proper fix for this is #2436 ?

test_bench/README.md
```diff
@@ -0,0 +1,57 @@
...
+```console
+$ docker push some.registry.io/foo/iox-glue
+```
+
+Next we'll "kustomize" the test bench. You can use `k8s/overlays/demo` as a starting point. There you should at least
+have a look at `images-patch.yml` to set the image with the glue code that you've just built. Then you can deploy the
+whole thing:
+
+```console
+$ kubectl kustomize ./k8s/overlays/demo| kubectl apply -f -
```
tustvold, line 35: Btw `kubectl apply -k` is shorthand for this

test_bench/k8s/overlays/demo/images-patch.yml
```diff
@@ -0,0 +1,15 @@
...
+  name: not-important
+spec:
+  template:
+    spec:
+      containers:
+        - name: iox
+          image: quay.io/influxdb/iox:main
+          imagePullPolicy: Always
+        - name: glue
+          image: europe-west3-docker.pkg.dev/influxdata-fusion/mneumann-repository-docker-1/iox-glue
```
tustvold, line 14: Is this intentional?

test_bench/glue/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM alpine:latest
```
mkmik, line 1:
```
FROM nixery.dev/shell/curl/grpcurl
```

and you can avoid manually installing grpcurl

test_bench/k8s/overlays/demo/images-patch.yml
```diff
@@ -0,0 +1,15 @@
+---
```
tustvold, line 1: Kustomize has some sugar for overriding images without writing patch files - https://github.com/kubernetes-sigs/kustomize/blob/master/examples/image.md

test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+    status=$?
+    set -e
+
+    if [[ $status == 0 ]]; then
+        echo "server ID set"
+        break
+    else
+        echo "cannot set server ID yet, waiting..."
+        sleep 1
+    fi
```
mkmik, line 27:
```suggestion
    if grpcurl -d "{\"id\": $server_id}" -allow-unknown-fields -plaintext "$INFLUXDB_IOX_GRPC_BIND_ADDR" influxdata.iox.deployment.v1.DeploymentService.UpdateServerId; then
        echo "server ID set"
        break
    else
        echo "cannot set server ID yet, waiting..."
        sleep 1
    fi
```

test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+# extract ordinal index from server ID
+[[ $HOSTNAME =~ -([0-9]+)$ ]] || (echo "invalid hostname" && exit 1)
+ordinal=${BASH_REMATCH[1]}
```
tustvold, line 7: A nicer way to do this would be to use the downward API to mount the ordinal as an environment variable - https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+            # select create/update routing depending on the config type
+            if [[ $type == "database" ]]; then
+                echo "Create database..."
+
+                set +e
+                grpcurl -d @ < "$F_CURRENT" -allow-unknown-fields -plaintext "$INFLUXDB_IOX_GRPC_BIND_ADDR" influxdata.iox.management.v1.ManagementService.CreateDatabase
+                status=$?
+                set -e
+
+                if [[ $status == 0 ]]; then
```
mkmik, line 68:
```suggestion
                if grpcurl -d @ < "$F_CURRENT" -allow-unknown-fields -plaintext "$INFLUXDB_IOX_GRPC_BIND_ADDR" influxdata.iox.management.v1.ManagementService.CreateDatabase; then
```

test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+                    in_sync=1
+                else
+                    echo "cannot create database, try updating it..."
+
+                    set +e
+                    grpcurl -d @ < "$F_CURRENT" -allow-unknown-fields -plaintext "$INFLUXDB_IOX_GRPC_BIND_ADDR" influxdata.iox.management.v1.ManagementService.UpdateDatabase
+                    status=$?
+                    set -e
+
+                    if [[ $status == 0 ]]; then
```
mkmik, line 79:
```suggestion
                    if grpcurl -d @ < "$F_CURRENT" -allow-unknown-fields -plaintext "$INFLUXDB_IOX_GRPC_BIND_ADDR" influxdata.iox.management.v1.ManagementService.UpdateDatabase; then
```

test_bench/glue/Dockerfile
crepererum: That image didn't work with my local podman/buildah builder, probably due to some bug in buildah and I stopped investigating.

test_bench/glue/entrypoint.sh
mkmik: etc

tustvold (commented):
This is awesome :heart: 

Edit: will approve once I've tried it out

test_bench/glue/entrypoint.sh
```diff
@@ -0,0 +1,119 @@
...
+    done
+
+    # TODO: delete resources that are no longer present, should be done by comparing the list of resources on the server VS the ones in config files by name
+
+    # store fingerprints for next round
+    rm -rf "$FINGERPRINT_DONE"
+    mkdir "$FINGERPRINT_DONE"
+    if [ -n "$(ls -A "$FINGERPRINT_STAGING")" ]; then
+        cp "$FINGERPRINT_STAGING"/* "$FINGERPRINT_DONE"/
+    fi
```
mkmik, line 116: unix pro tip to handle the case where a source dir is empty

```suggestion
    cp "$FINGERPRINT_STAGING"/. "$FINGERPRINT_DONE"/
```

test_bench/k8s/overlays/demo/images-patch.yml
crepererum: Pff, somewhat. If you're using our gcloud account as well you can use this image.

test_bench/glue/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV CURL_FLAGS="--proto =https --tlsv1.2 -sSf" \
+    GRPCURL_VERSION="1.8.5"
+
+RUN apk add --no-cache bash curl
+
+RUN curl ${CURL_FLAGS} -Lo grpcurl.tar.gz https://github.com/fullstorydev/grpcurl/releases/download/v${GRPCURL_VERSION}/grpcurl_${GRPCURL_VERSION}_linux_x86_64.tar.gz \
+    && tar xvf grpcurl.tar.gz \
+    && mv grpcurl /bin
+
+COPY entrypoint.sh /entrypoint.sh
```
tustvold, line 12: I wonder if this script could instead be a configmap that gets mounted in? This would allow updating the script without needing to rebuild the image?

test_bench/README.md
```diff
@@ -0,0 +1,57 @@
...
+## Architecture
+The test bench is meant to run in [Kubernetes]. To create different test setups or to customize a setup to a specific
+cluster we use [Kustomize]. The [Kustomize] configs can be found in `k8s/`.
+
+To glue IOx to [Kubernetes] we need some helper code which can be found under `glue/`.
+
+## Getting Started
+First you need to build an image with the glue code:
+
+```console
```
mkmik, line 18: perhaps we could simplify this by using a nixery image with the required tools and then just injecting the script via kustomize, instead of having to create an image with the script, pushing it somewhere etc

I see you already have to deal with kustomize in order to inject the right glue image.

test_bench/glue/entrypoint.sh
crepererum: I think that's currently not possible, see https://github.com/kubernetes/kubernetes/issues/40651 Or I didn't find any way to only get the ordinal w/o the hostname. Otherwise I could also set the server ID env variable in the first place and can completely remove this gRPC call here.

test_bench/glue/Dockerfile
crepererum: I personally don't really like this way to abuse config maps. Ideally we keep at least the script file freestanding so that syntax highlighting and tools like `shellcheck` just work.

test_bench/README.md
mkmik: is this using the same version of kustomize?

I remember when `-k` was using a builtin version that was way behind the latest version and `kubectl kustomize` was picking up the external process if you had it installed (perhaps I'm misremembering)

test_bench/glue/Dockerfile
tustvold: You could use kustomize ConfigMapGenerator to keep it as a free-floating script that gets bundled up into a configmap as a deployment detail?

mkmik: I also like the script being in a real file (so that you can run `shellcheck` on it, your IDE etc)

whatever extra step we have to do to get the script from a plain file into a configmap are likely less problematic
that having to build and push a custom docker image.

(relevant comment https://github.com/influxdata/influxdb_iox/pull/3215/files#r757065978)

https://github.com/influxdata/influxdb_iox/pull/3215/files#r757065978

test_bench/glue/entrypoint.sh
tustvold: Ahaha I just went digging through some old code to work out how I did it... Yeah I used the downward API to get the hostname as an environment variable, and then overrode the cmd with 

```
/bin/bash -c "exec process --ordinal ${HOSTNAME##*-}"
```

test_bench/README.md
tustvold: Yeah, that does ring a bell...

test_bench/glue/Dockerfile
mkmik: are you sure you included `/shell/` as the first path component?

crepererum: Yeah, it was a different weird bug like "I was expecting more/less image layers, but found that many on disk".

test_bench/README.md
crepererum: done

test_bench/glue/Dockerfile
crepererum: done

test_bench/k8s/overlays/demo/images-patch.yml
crepererum: done

test_bench/k8s/overlays/demo/images-patch.yml
crepererum: not relevant any longer

test_bench/glue/Dockerfile
crepererum: not relevant any longer

tustvold (approved):

mkmik (approved):

tustvold:
Tagging #3143 as I believe this relates to it

Issue #3209, opened 2021-11-24
Reproducible Query Performance Tests 


tustvold:
**Problem**

In order to quantitatively evaluate various design trade-offs it is desirable to be able to construct an IOx server with a given state and run queries against it. This is currently quite difficult to achieve and requires manually ingesting data, and then triggering lifecycle actions to move the data into the desired storage locations.

**Proposal**

* Add the ability to reload a parquet file back into the RUB and expose this in the management API and CLI
* Expose the ability to rebuild the catalog for a set of parquet files written by IOx in the CLI

The combination of these two pieces of functionality would allow you to construct an IOx server with a subset of data sourced from some system, potentially written into IOx by the data generator, and then test the query performance with chunks in the RUB or not. If/when we add parquet file caching, the procedure would be much the same where you would need an API to prefetch or evict certain files.

**Alternatives Considered**

A "snapshot" mode, that would take a list of parquet files and a config file and construct an IOx server with a corresponding in-memory state. This would be significantly more intrusive to implement, as it would need to still behave largely like a standard `database` mode process, but with a seeded initial state. It is also unclear how you would run this on a remote instance.

A CLI tool to automate loading data into a server and driving the lifecycle manually to a desired end state. This potentially combined with the ability to upsert entire parquet files. This would likely work, but it is better if you can avoid re-ingesting data.

e-dard:
This is what my dreams are made of ❤️

alamb:
The high level proposal looks good to me 👍 

Here is a proposed API.

```bash
# Create a new, empty database
./influxdb_iox database create my_db

# tell the database to adopt all parquet files located in <url>, recursively
# URL must be accessable to the server (TODO other restrictions on url)
./influxdb_iox database import s3://iox-bucket/servers/21/data/foo

```

The idea would be that all metadata from all parquet files with prefix `/data/foo` would be read, and added to the catalog of `my_db`. No other catalogs would be modified, and no data would be moved in object store.

Open questions:
1. Should we move data into the proper location in object storage for `my_db` to  take control of them? Pros would be that my_db now acts like any other catalog. Cons is that data movement is required

tustvold:
@alamb I was actually proposing something far more basic. You would need to take the files and copy them to an S3 bucket with the topology expected by IOx. All the CLI would provide is the ability to reconstruct a catalog from those files. Writing a generalized data migration tool seems like a load of extra work that doesn't help with being able to run query tests?

alamb:
[...]

PR #3216, opened 2021-11-25
feat: reload chunk back into read buffer (#3209)

tustvold:
Adds lifecycle machinery necessary to load a chunk from object storage back into the read buffer. A subsequent PR will add the gRPC and CLI plumbing to expose this

Part of #3224


server/src/db/lifecycle/load.rs
```diff
@@ -0,0 +1,131 @@
...
+    // Drop locks
+    let chunk = chunk.into_data().chunk;
+
+    let ctx = db.exec.new_context(ExecutorType::Reorg);
+
+    let fut = async move {
+        let key = compute_sort_key(std::iter::once(db_chunk.summary()));
+
+        // Cannot move query_chunks as the sort key borrows the column names
+        let (_, plan) = ReorgPlanner::new().compact_plan(
```
crepererum, line 49: So the intention here is to do some optimization work during loading, i.e. sort the data and drop duplicates? Is that really required for files that were written by IOx itself?

tustvold: > sort the data and drop duplicates

The dedup plan is smart enough to know it doesn't need to do these for a parquet chunk (or if it isn't it should be). The data is already sorted and cannot contain duplicates. This is basically just a mechanism to reuse a lot of code to perform a standard scan on a parquet chunk on a thread owned by the reorg execution context

crepererum (approved):

PR #3217, opened 2021-11-26
feat: disallow unsequenced deletes when reading from write buffer

crepererum:
Similar to how writes to a databases that reads from a write buffer are
forbidden deletes should be rejected as well.

### DIFF
--- a/influxdb_iox/tests/end_to_end_cases/write_buffer.rs
+++ b/influxdb_iox/tests/end_to_end_cases/write_buffer.rs
@@ -10,6 +10,10 @@ use generated_types::influxdata::iox::write_buffer::v1::{
     write_buffer_connection::Direction as WriteBufferDirection, WriteBufferConnection,
 };
 use influxdb_iox_client::{
+    delete::{
+        generated_types::{Predicate, TimestampRange},
+        DeleteError,
+    },
     management::{generated_types::WriteBufferCreationConfig, CreateDatabaseError},
     write::WriteError,
 };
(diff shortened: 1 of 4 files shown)


server/src/db.rs
```diff
@@ -521,12 +521,12 @@ impl Db {
...
 
     /// Store a delete
-    pub fn store_delete(&self, delete: &DmlDelete) -> Result<()> {
+    pub(crate) fn store_delete(&self, delete: &DmlDelete) -> Result<()> {
         self.store_filtered_delete(delete, DeleteFilterNone::default())
     }
 
     /// Store a delete with the provided [`DeleteFilter`]
-    pub fn store_filtered_delete(
+    pub(crate) fn store_filtered_delete(
```
tustvold, line 529: I think these should be pub still? If the eventual goal is to split Db into its own crate?

influxdb_iox/tests/end_to_end_cases/write_buffer.rs
```diff
@@ -153,6 +157,23 @@ async fn cant_write_to_db_reading_from_write_buffer() {
 
     assert_contains!(err.to_string(), "only allowed through write buffer");
     assert!(matches!(dbg!(err), WriteError::ServerError(_)));
+
+    // Deleteing from this database is an error; all data comes from write buffer
```
tustvold, line 161:
```suggestion
    // Deleting from this database is an error; all data comes from write buffer
```

tustvold (approved):

server/src/db.rs
crepererum: Yeah, for that refactoring we would need to make it public. I'll revert this specific change.

PR #3218, opened 2021-11-26
refactor: remove connection manager and resolver from `Server`

crepererum:


tustvold (commented):
:heart: It makes me irrationally happy to finally start seeing all this expunged

tustvold (approved):

PR #3219, opened 2021-11-26
refactor: remove write buffer producer logic from `Db`

### LINKED ISSUE #2243: Pull WriteBuffer out of Db onto Database

crepererum:
As a side effect, writing to database no longer requires `async`.

Closes #2243.


tustvold (approved):
Yay! :tada:

PR #3221, opened 2021-11-26
refactor: remove write buffer direction

crepererum:
The direction was required when a database could read or write from/to a
write buffer. Now it is clear from the usage context of a write buffer
context which of the two applications is meant (databases read, routers
write) so the direction flag is no longer required.


tustvold (approved):

alamb:
It is very cool to see the final phases of this work and all the deleted code ❤️  -- nice work @crepererum

crepererum:
@alamb there's more to come (routing config in database rules)

PR #3222, opened 2021-11-26
refactor: consolidate sharding implementation

crepererum:
Move sharding implementation from router to `dml` (where the types can
be consumed w/o cloning the data) and only support the new sharding
configs.

The old sharding configs will be removed soon.


dml/src/lib.rs
```diff
@@ -198,23 +216,23 @@ impl DmlWrite {
...
-    pub fn shard(
-        self,
-        config: &ShardConfig,
-    ) -> Result<HashMap<ShardId, Self>, data_types::database_rules::Error> {
-        let mut sharded_tables = HashMap::new();
+    pub fn shard(self, config: &ShardConfig) -> BTreeMap<ShardId, Self> {
+        let mut batches: HashMap<ShardId, HashMap<String, MutableBatch>> = HashMap::new();
+
         for (table, batch) in self.tables {
-            let shard = config.shard(&table)?;
```
tustvold, line 207: Should probably remove the now orphaned logic in data_types

dml/src/lib.rs
```diff
@@ -198,23 +216,23 @@ impl DmlWrite {
     }
 
     /// Shards this [`DmlWrite`]
-    pub fn shard(
-        self,
-        config: &ShardConfig,
-    ) -> Result<HashMap<ShardId, Self>, data_types::database_rules::Error> {
```
tustvold, line 204: Less error handling is always a win imo

tustvold (approved):

dml/src/lib.rs
crepererum: next PR

PR #3223, opened 2021-11-26
fix: Implement a dummy Storage.Offsets method

mkmik:
Closes https://github.com/influxdata/conductor/issues/766

(@brettbuddin this is the reason why your https://github.com/influxdata/idpe/pull/12426 failed to deploy on staging)

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
@@ -11,12 +11,12 @@ use tonic::Status;
 
 use data_types::{error::ErrorLogger, names::org_and_bucket_to_database, DatabaseName};
 use generated_types::{
-    google::protobuf::Empty, storage_server::Storage, CapabilitiesResponse, Capability,
-    Int64ValuesResponse, MeasurementFieldsRequest, MeasurementFieldsResponse,
-    MeasurementNamesRequest, MeasurementTagKeysRequest, MeasurementTagValuesRequest, Predicate,
-    ReadFilterRequest, ReadGroupRequest, ReadResponse, ReadSeriesCardinalityRequest,
-    ReadWindowAggregateRequest, StringValuesResponse, TagKeyMetaNames, TagKeysRequest,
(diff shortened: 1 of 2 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -11,12 +11,12 @@ use tonic::Status;
 
 use data_types::{error::ErrorLogger, names::org_and_bucket_to_database, DatabaseName};
 use generated_types::{
-    google::protobuf::Empty, storage_server::Storage, CapabilitiesResponse, Capability,
-    Int64ValuesResponse, MeasurementFieldsRequest, MeasurementFieldsResponse,
-    MeasurementNamesRequest, MeasurementTagKeysRequest, MeasurementTagValuesRequest, Predicate,
-    ReadFilterRequest, ReadGroupRequest, ReadResponse, ReadSeriesCardinalityRequest,
-    ReadWindowAggregateRequest, StringValuesResponse, TagKeyMetaNames, TagKeysRequest,
-    TagValuesRequest, TimestampRange,
+    google::protobuf::Empty, offsets_response::PartitionOffsetResponse, storage_server::Storage,
```
rogpeppe, line 14: ISTM that this might be better formatted one entry per line to avoid awkward to parse diffs like this. I dunno if that's ever idiomatic in Rust or not.

rogpeppe (approved):

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
mkmik: We use rustfmt and for some reason its authors decided that it the aesthetics of the use section are important.
This is not the only place where I noticed a small change causes big reformat

mkmik:
I also got an informal approval from @e-dard on slack before I set off writing this PR.
It looks harmless so I'll merge optimistically

alamb:
Thanks @mkmik !

I have proposedan end to end test in https://github.com/influxdata/influxdb_iox/pull/3232

PR #3243, opened 2021-11-29
refactor: improve Kafka client usage

crepererum:
With the new rust-rdkafka release (merged in #3234) managing multiple
consumer streams becomes a bit easier. Also we can just reuse consumer
clients for multiple metadata requests. In total that provides:

- use only a single client connection for consumers (we had multiple
  connection attempts during startup and one client per stream)
- use only two clients for producers (sadly we need a consumer client to
  probe the partitions during startup)
- consumers no longer need to poll the stream to receive statistics


tustvold (approved):
I think this looks good, if it works SGTM :grin:

Issue #3225, opened 2021-11-26
Rebuild Catalog


tustvold:
We wish to be able to rebuild the catalog of a database in the event the catalog is lost or corrupted. This functionality is currently implemented `parquet_catalog::rebuild::rebuild_catalog` but is not user-accessible. It should be possible to request that IOx attempts to rebuild the catalog for a Database, if it has failed to load i.e. it is in `DatabaseState::CatalogLoadError`. 

This would be similar to the similar to the existing `WipePreservedCatalog` but would instead run the rebuild logic.

This will also enable building a catalog from a set of parquet files produced by IOx.

* Create new database
* Stop IOx so it doesn't delete "orphaned parquet files"
* Delete transactions directory so catalog load on startup fails
* Add parquet files to `data` directory
* Startup IOx and run a catalog rebuild e.g. `influxdb_iox database recover rebuild`

Split out from #3209

alamb:
@tustvold  would this feature be able to rebuild a catalog from a subset of parquet files (e.g. if only the parquet files for a particular table were copied to a new location)??

tustvold:
@alamb Yup, depending on what you copied you might even be able to resume where you left off from the same write buffer.

Longer-term I think we will want more sophisticated catalog splitting machinery, that's able to preserve deletes, etc... but this is a quick and dirty mechanism that doesn't require anything more than a bit of gRPC plumbing, and that we want for disaster recovery anyway.

alamb:
I think this is a fine interim step (rebuild from what is actually there). We can work on more sophisticated strategies, as you say, as part of a larger story

PR #3245, opened 2021-11-29
refactor: cleanup database recovery (#3225)

tustvold:
In preparation for #3225 this PR does some minor cleanup to the existing WipePreservedCatalog and SkipReplay code paths to hopefully reduce the duplication.

This also finishes up the work started by @carols10cents in #2570 so that the API for Database is slightly more consistent


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/management.rs
```diff
@@ -421,15 +421,8 @@ impl management_service_server::ManagementService for ManagementService {
         let tracker = self
             .server
             .wipe_preserved_catalog(&db_name)
-            .map_err(|e| match e {
```
tustvold, line 424: This error mapping confused me, as I can't see  a way for server::wipe_preserved_catalog to return anything other than DatabaseNotFound or WipePreservedCatalog

server/src/lib.rs
```diff
@@ -874,24 +874,14 @@ impl Server {
     ///
     /// The DB must exist in the server and have failed to load the catalog for this to work
     /// This is done to prevent race conditions between DB jobs and this command
-    pub fn wipe_preserved_catalog(
```
tustvold, line 877: I'm tempted to just remove this method from server in a future PR

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/error.rs
```diff
@@ -1,5 +1,5 @@
 use generated_types::google::{
-    FieldViolation, InternalError, NotFound, PreconditionViolation, QuotaFailure,
+    AlreadyExists, FieldViolation, InternalError, NotFound, PreconditionViolation, QuotaFailure,
```
tustvold, line 2: I wonder about defining these conversions as From implementations, to remove the need for all the map_err everywhere :thinking:

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/error.rs
```diff
@@ -17,6 +17,12 @@ pub fn default_server_error_handler(error: server::Error) -> tonic::Status {
         Error::DatabaseNotInitialized { db_name } => {
             tonic::Status::unavailable(format!("Database ({}) is not yet initialized", db_name))
         }
+        Error::DatabaseAlreadyExists { db_name } => AlreadyExists {
```
tustvold, line 20: Aside from more consistent error messages, using the `AlreadyExists` payload will embed the corresponding grpc details payload for clients to potentially decode

influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -338,7 +338,7 @@ async fn test_create_get_update_release_claim_database() {
     let err = client.claim_database(released_uuid).await.unwrap_err();
     assert_contains!(
         err.to_string(),
-        format!("A database with the name `{}` already exists", db_name)
+        format!("Resource database/{} already exists", db_name)
```
tustvold, line 341: This is a consequence of changing the default error mapping, this makes it consistent with the other `AlreadyExists` style errors, in addition to providing the grpc details payload should a client wish to decode it

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/error.rs
crepererum: I think the issue is:

1. we don't map all errors
2. For the ones we don't map, we convert them to an opaque "internal error" and create a log statement for debugging. IMHO `From` should be free from side effects, so logging isn't really an option.
3. Sometimes we convert to google errors, sometimes we directly create a `tonic::Status`.

Overall I feel that  the error handling handling could also be improved by not having these overlarge error types w/ ton of variants :/

server/src/database.rs
```diff
@@ -28,22 +29,30 @@ use std::{future::Future, sync::Arc, time::Duration};
 use time::Time;
 use tokio::{sync::Notify, task::JoinError};
 use tokio_util::sync::CancellationToken;
+use tracker::{TaskTracker, TrackedFutureExt};
 use uuid::Uuid;
 
+/// Matches an error [`DatabaseState`] and clones the contained state
+macro_rules! error_state {
+    ($s:expr, $transition: literal, $variant:ident) => {
+        match &**$s.shared.state.read() {
```
crepererum, line 38: `&**$`

Matrix

server/src/lib.rs
crepererum: I agree. Most of the method on `Server` should be removed. API users should either user `Database` or `Db` directly.

server/src/database.rs
```diff
@@ -487,47 +496,40 @@ impl Database {
...
-                _ => {
-                    return InvalidState {
-                        db_name,
-                        state: state.state_code(),
-                        transition: "WipePreservedCatalog",
-                    }
-                    .fail()
-                }
-            };
+    pub async fn wipe_preserved_catalog(&self) -> Result<TaskTracker<Job>, Error> {
```
crepererum, line 499: why is this `async` now?

server/src/database.rs
```diff
@@ -487,47 +496,40 @@ impl Database {
...
+        let registry = self.shared.application.job_registry();
+        let (tracker, registration) = registry.register(Job::WipePreservedCatalog {
+            db_name: Arc::from(db_name.as_str()),
+        });
 
         let shared = Arc::clone(&self.shared);
 
-        Ok(async move {
-            let db_name = &shared.config.name;
+        tokio::spawn(
```
crepererum, line 513: Who tracks this subtask and can stop it? I'm somewhat allergic to untracked `tokio::spawn` calls.

server/src/database.rs
tustvold: It allows removing the `TransitionInProgress` error variant. I'm not sure how I feel about this, I was just going for consistency with the other methods - I personally am not a massive fan of yet more async :laughing:

server/src/database.rs
tustvold: The tracker registration can stop it, this is just the tokio::spawn that used to be in server moved into Database

server/src/database.rs
crepererum: I don't understand how adding an `async` to a function that never `await`s allows you to remove an enum variant :thinking:

tustvold: There is a `handle.await` two lines down?

server/src/database.rs
crepererum: Uuh, I will probably ban `tokio::spawn` at one point and allow this call only at places that opt-in, e.g. this `spawn` call could be done somewhere in the task registration. But anyway, that's not your concern now.

crepererum (approved):

server/src/database.rs
tustvold: I would not object to such a change

Issue #3247, opened 2021-11-29
Inconsistent Management API Responses


tustvold:
**Problem**

Some operations within the management API return a longrunning operations handle as soon as the task has been "accepted":

* ClosePartitionChunk
* WipePreservedCatalog

Others don't return a longrunning operation at all:

* SkipReplay
* PersistPartition
* DropPartition
* UnloadPartitionChunk

This makes for an inconsistent API which is not ideal.

The major reason to favor eagerly returning a longrunning operation handle is it allows setting tight request [deadlines](https://grpc.io/blog/deadlines/), the major downside is it complicates client implementations. 

**Proposal**

There are broadly 3 categories of operation:

1. Many Minutes - WipePreservedCatalog
2. Seconds to minutes - PersistPartition, ClosePartitionChunk
3. Milliseconds to Seconds - SkipReplay, UnloadPartitionChunk, DropPartition

I think creating and returning longrunning task handles for category 3 is unnecessary, similarly I think not eagerly returning a handle for category 1 is not a good idea. I'm not sure what the correct choice for category 2 is, perhaps create longrunning operation internally, but use synchronous RPC with the client?

crepererum:
A major difference between category 2 and 3 is that the former involves object store IO. I think if it's avoidable we should never block a request on external IO (query request being a notable exception to this rule).

FWIW I would also favor a bit more implementation complexity if it results in a more consistent API design. This design not only affects that gRPC interface but also the Rust client as well as the CLI. So I would also like to see category 3 to return job handles eagerly, but that's not a too strong opinion.

alamb:
I agree with @crepererum. So specifically, a counter proposal is  that all APIs return long running all categories you refer to and adding code to the Rust client that waits for the operation to complete successfully that is convenient to use.

> major downside is it complicates client implementations.

FWIW I am not sure it complicates client implementations (as the client already has to handle long running operations for certain APIs)

tustvold:
No objections from me to always eagerly returning an operations handle :+1:

Issue #3224, opened 2021-11-26
Reload Persisted Chunk into RUB


tustvold:
**Problem**

Once the lifecycle decided to unload a chunk from the read buffer, there is no mechanism for loading it back. In the abscence of a parquet cache (#1491) this forces IOx to fetch the parquet files for every query. We wish to be able to quantify how queries perform against the read buffer, compared to parquet, and so being able to construct the desired in-memory state is important

**Proposal**

Add a lifecycle action, and accompanying gRPC and CLI plumbing to allow moving a chunk in `ObjectStoreOnly` back into `ReadBufferAndObjectStore`.

Split out from #3209

PR #3249, opened 2021-11-29
feat: support forcing persistence of partition irrespective of arrival time

tustvold:
I originally tried to roll addressing #3247 into this, but it got really complicated. This just adds the ability to force persistence, which I need in order to properly test #3224 

Edit: I found an alternative way to test #3224 but still think this change makes sense


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -402,6 +402,9 @@ message PersistPartitionRequest {
 
   // the table name
   string table_name = 3;
+
+  // persist all data for partition irrespective of arrival time
```
alamb, line 406: I agree this makes sense

influxdb_iox/src/commands/database/partition.rs
```diff
@@ -229,7 +229,7 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
             } = persist;
 
             client
-                .persist_partition(db_name, table_name, partition_key)
```
alamb, line 232: Is there any value exposing a `--force` cli flag too?

alamb (approved):
makes sense to me

influxdb_iox/src/commands/database/partition.rs
tustvold: Done

PR #3250, opened 2021-11-29
feat: load RUB from object store (#3224)

### LINKED ISSUE #3224: Reload Persisted Chunk into RUB

tustvold:
Closes #3224


NGA-TRAN:
Will review this first thing tomorrow

crepererum (approved):

NGA-TRAN (commented):
I guess you will (or already have) tests for this. 

One thing I am been thinking for all the tests (not only this one) is we may want a way to test all/most errors especially the one happens during catalog update. Just something for us to think.

PR #3252, opened 2021-11-29
feat: update os compacted chunk in both in-memory and preserved catalogs

### LINKED ISSUE #3255: Delete predicate handling during parquet compaction
# Abstract
We need to be a bit careful w/ delete predicates during compaction of preserved chunks (i.e. parquet files).

# Assumptions
The schema that will be used makes the following assumptions:

## A.1 Chunk UUIDs
Chunks are addressed by UUIDs (version 4). This makes chunk addresses unique, even over multiple server restarts.

## A.2 Excessive Preserved Delete Predicates
The preserved catalog might contain delete predicates for unknown chunks. The unknown chunks are just ignored during catalog loading.

## A.3 Lazy Predicate Preservation
Incoming new predicates are never preserved directly but are attached to affected chunks and are also added to a mailbox. The mailbox content is stored regularly either by a background worker loop or during every other chunk preservation.

## A.4 Failable IO
Transaction to the preserved catalog might fail. To avoid out-of-sync situations of the in-mem catalog and the preserved catalog, we use the following schema:

1. Mark affected in-mem chunks
2. Transaction IO
3. If successful: commit in-mem change

## A.5 Short-lived locks
Chunks and partitions MUST NOT be locked during IO or longer calculations (e.g. compaction).

# Schema
The compaction job does the following:

1. Generate UUID for the result chunk
2. With partition write lock:
   1. Mark all to-be-compacted chunks using a lifecycle marker `CompactingOS`. This lifecycle marker shall also contain the UUID from step 1.
3. Run compaction (no locks at this point!).
4. With parquet clean-up lock.
   1. Write parquet file.
   2. Create preserved catalog transaction with new persisted chunk (no delete predicates required here). Exit if it fails.
5. With partition lock:
   1. Add new parquet-file-based chunk using the UUID from step 1
   2. Delete compacted chunks.

The mailbox insertion mechanism here

https://github.com/influxdata/influxdb_iox/blob/f165faf5bce152bef6676f4af7fef699c5cfd085/server/src/db.rs#L573-L583

must be modified to also check if a chunk is in the newly introduced `CompactingOS` state. If so, it also marks the to-be-created chunk as affected. This works due to assumption A.2.

NGA-TRAN:
Close #3255 
Another piece of #3089 


server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -114,27 +117,33 @@ pub(crate) fn compact_object_store_chunks(
...
-        } // End of cleanup locking
-
-        // Step 4: Update the in-memory catalogs to use the newly created os_chunk
-        //   . Drop all os_chunks from the in-memory catalog
-        //   . Add the new created os_chunk in the in-memory catalog
-        //  This step can be done outside a transaction because the in-memory catalog
-        //    was design to false tolerant
+            // Step 3.2: Update the in-memory catalog to use the newly created os_chunk
+            //   . Drop all os_chunks from the in-memory catalog
+            //   . Add the new created os_chunk in the in-memory catalog
```
NGA-TRAN, line 127: @crepererum : Since we need the delete predicates for updating the preserve catalog, I think we need to update the in-memory catalog before the preserved catalog to avoid loosing delete predicates due to lock gaps. Let me know if this is feasible solution

NGA-TRAN (commented):
@tustvold and @crepererum : these are the last 2 pieces of the compact_os_chunks puzzle. My next step is to add the rest of the tests

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -375,12 +394,107 @@ async fn persist_stream_to_chunk<'a>(
...
+        transaction.delete_predicate(&predicate, &[chunk_addr.clone().into()]);
+    }
+
+    // Add pending delete predicates
+    for (predicate, chunks) in delete_handle.outbox() {
+        transaction.delete_predicate(predicate, chunks);
+    }
+
+    // Close/commit the transaction
+    transaction.commit().await.context(CommitError)?;
```
tustvold, line 484: You need to call `handle.flush` if the transaction commits - otherwise nothing actually gets flushed from the mailbox :laughing: 

FWIW I don't think it is necessary to flush pending deletes from the mailbox, as this operation doesn't update any of the checkpointing information, but I guess it can't hurt

tustvold (changes requested):
This drops chunks from the in-memory catalog before updating the preserved catalog, as this latter operation is fallible we either need some way to rollback the changes to the in-memory catalog or to perform the operations in the other order.

I've tried to explain how this might work, but it might be easiest to talk through how to do this synchronously

server/src/db/lifecycle/compact_object_store.rs
crepererum: I was about to answer, but it took me a while to write down what I had in mind. @tustvold and I had a conservation about this issue a while back and IIRC we came up w/ this plan: #3255.

In general there's no way you can do the in-mem transaction before the persisted transaction because the latter one can always fail due to IO.

NGA-TRAN: Thanks both @tustvold and @crepererum. Very helpful

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -375,12 +394,107 @@ async fn persist_stream_to_chunk<'a>(
...
+///
+async fn update_in_memory_catalog<'a>(
+    db: &'a Arc<Db>,
+    chunk_ids: &'a [ChunkId],
+    delete_predicates_before: &'a HashSet<Arc<DeletePredicate>>,
+    iox_metadata: IoxMetadata,
+    parquet_chunk: Arc<ParquetChunk>,
+) -> Result<(Arc<DbChunk>, Vec<Arc<DeletePredicate>>)> {
+    // Drop the old chunks and capture leftover delete predicates
+    let partition = db
```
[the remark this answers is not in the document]
NGA-TRAN: Let me keep the pointer to partition instead. I was a bit confused about the way to get partition and acquire its read/write lock but I think I got it now.

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: I will integrate this with #3255

NGA-TRAN (commented):
Thanks @tustvold and @crepererum . I think I got the idea. Will put them in the right order

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: @tustvold With the new changes, I am not sure if we need the `handle.flush` anymore. Can you check?

generated_types/protos/influxdata/iox/management/v1/chunk.proto
```diff
@@ -44,6 +44,7 @@ enum ChunkLifecycleAction {
   CHUNK_LIFECYCLE_ACTION_DROPPING = 4;
 
   /// Chunk is in the process of being compacted
+  // todo: how to make this enum(ChunkId)?
```
NGA-TRAN, line 47: @tustvold : can we do this or we need another field for ChunkId?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -498,11 +613,279 @@ mod tests {
...
+        assert_eq!(summary_chunks.len(), 2);
+        // MUB
+        assert_eq!(summary_chunks[0].storage, ChunkStorage::OpenMutableBuffer);
+        assert_eq!(summary_chunks[0].row_count, 1);
+        // OS: the result of compacting all 3 persisted chunks
+        assert_eq!(summary_chunks[1].storage, ChunkStorage::ObjectStoreOnly);
+        assert_eq!(summary_chunks[1].row_count, 2);
+    }
+
+    #[ignore]
```
NGA-TRAN, line 816: This test does not pass yet. I know where the bug happens but need more time to investigate. It does not affect others though. I will fix it after this PR is merged

NGA-TRAN (commented):
@tustvold and @crepererum : These PR does 2 things:
1. Handle  delete predicates per your suggestions. I feel quite confident about the changes.
2. Handle empty result.

You may want to review them all

server/src/db.rs
```diff
@@ -573,13 +573,21 @@ impl Db {
 
                     // We should only report persisted chunks or chunks that are currently being persisted, because the
                     // preserved catalog does not care about purely in-mem chunks.
+
+                    let (chunk_id, compacting_os) =
+                        match chunk.in_lifecycle_compacting_object_store() {
+                            Some(id) => (id, true),
```
crepererum, line 579: In this case you need to use BOTH chunk IDs (the ID of the chunk itself AND the ID of the to-be-created chunk). The reason is that the compaction that is currently in progress might fail (either the compaction itself fails or the commit to the preserved catalog fails) and in the failure case we don't wanna loose the delete predicate.

This is somewhat a tee construct, i.e. it sends the delete predicates to multiple chunks (the existing one and the to-be-created one).

server/src/db/lifecycle/compact_object_store.rs
crepererum: I think it's still the same. There's no need to touch the mailbox here, since the background worker (or another persistence lifecycle action) will lazily take care of that.

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -372,7 +396,90 @@ async fn persist_stream_to_chunk<'a>(
...
+    let delete_handle = db.delete_predicates_mailbox.consume().await;
+    let mut delete_predicates = vec![];
+    for (predicate, chunks) in delete_handle.outbox() {
+        for chunk in chunks {
+            if chunk.chunk_id == iox_metadata.chunk_id {
+                delete_predicates.push(Arc::clone(predicate));
+            }
+        }
+    }
+    delete_predicates.sort();
```
crepererum, line 451: (This is my 2nd try writing this because the GitHub UI scoping of "cancel" is buggy :angry: ).

This is a bit tricky because now you have a lock gap between "we determine the delete predicates" and "we modify the partition". During this gap new predicates might be added that are just lost (at least from the in-mem view).

# Solution 1
Move the `let mut partition = partition.write();` to the start of the function. That SHOULD work but I find it a somewhat weak lock interaction.

# Solution 2
This is similar to what we do in other places.

1. In `mark_chunks_to_compact` collect for each chunk that you're about to include the delete predicates (e.g. in a `BTreeMap`). Let's call this variable `delete_predicates_before`. For these predicates we know that they are materialized by the compaction.
2. Remove the marked code block.
3. Some lines down replace the `for id in chunk_ids {` block with:

```rust
let mut delete_predicates = HashSet::new();
for id in chunk_ids {
    let chunk = partition
        .force_drop_chunk(*id)
        .expect("There was a lifecycle action attached to this chunk, who deleted it?!");
    let before = delete_predicates_before.get(id).expect("chunk existed before compaction");
    for pred in chunk.delete_predicates() {
        if !before.contains(&pred) {
            delete_predicates.insert(pred);
        }
    }
}
```

CC @tustvold could you please double-check my lock reasoning here?

server/src/db.rs
NGA-TRAN: Good point

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: @crepererum I like solution 2 better. In that case, we do not need to read delete predicates from mailbox at all, right?

crepererum: > I like solution 2 better. In that case, we do not need to read delete predicates from mailbox at all, right?

Right.

NGA-TRAN: I think we still need both the `delete_predicate_after` from compacted chunks and the ones in the mailbox. Let me add them in

NGA-TRAN: Hah, did not see your response. Let three of us catch up on this after standup so we are on the same page

NGA-TRAN: Actually I agree with @crepererum that we do not need to get the ones in the mailbox because they will be done in the background later

NGA-TRAN:
@crepererum and @tustvold : I think all the known locking gaps have been solved. Can you have a look to get this merged? I am focusing on more tests now. This needs a lot of tests

server/src/db.rs
```diff
@@ -573,14 +573,31 @@ impl Db {
 
                     // We should only report persisted chunks or chunks that are currently being persisted, because the
                     // preserved catalog does not care about purely in-mem chunks.
+
+                    // Capture ID of to be created chunk for the in-progress compacting OS chunks
+                    let (to_be_created_chunk_id, compacting_os) =
```
crepererum, line 578: I think you can make this an `Option<ChunkId>` because the `to_be_created_chunk_id` is not used if `compacting_os = false`.

server/src/db.rs
```diff
@@ -573,14 +573,31 @@ impl Db {
...
                         || chunk.is_in_lifecycle(ChunkLifecycleAction::Persisting)
+                        || compacting_os
                     {
                         affected_persisted_chunks.push(ChunkAddrWithoutDatabase {
                             table_name: Arc::clone(&chunk.addr().table_name),
                             partition_key: Arc::clone(&chunk.addr().partition_key),
                             chunk_id: chunk.addr().chunk_id,
                         });
+
+                        if compacting_os {
```
tustvold, line 594:
```suggestion
                        if let Some(to_be_created) = to_be_created_chunk_id {
```

Should remove the need for the boolean and the `unwrap`

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -149,7 +168,7 @@ pub(crate) fn compact_object_store_chunks(
             rows_per_sec=?throughput,
             "object store chunk(s) compacted");
 
-        Ok(None) // todo: will be a real chunk when all todos done
+        Ok(dbchunk) // todo: consider to return Ok(None) when appropriate
```
tustvold, line 171: Is this todo outdated? This is an Option?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -372,7 +408,90 @@ async fn persist_stream_to_chunk<'a>(
...
+
+        let chunk = chunk.read();
+        for pred in chunk.delete_predicates() {
+            if !delete_predicates_before.contains(pred) {
+                delete_predicates_after.insert(Arc::clone(pred));
+            }
+        }
+    }
+
+    let delete_predicates = {
```
tustvold, line 468: Why do we sort the delete predicates?

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -372,7 +408,90 @@ async fn persist_stream_to_chunk<'a>(
...
+    Ok(())
+}
+
+async fn update_in_memory_catalog(
+    chunk_ids: &[ChunkId],
+    iox_metadata: IoxMetadata,
+    parquet_chunk: Option<Arc<ParquetChunk>>,
+    partition: Arc<RwLock<Partition>>,
+    delete_predicates_before: HashSet<Arc<DeletePredicate>>,
+) -> Result<Option<Arc<DbChunk>>> {
```
tustvold, line 450:
```suggestion
) -> Option<Arc<DbChunk>> {
```

This should be infallible

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -498,11 +625,279 @@ mod tests {
...
     // todo: add tests
-    //   . compact 2 contiguous OS chunks
-    //   . compact 3 chunks with duplicated data
-    //  . compact with deletes before compacting
     //  . compact with deletes happening during compaction
+    //  . compact with deletes/duplicates that lead to empty result
+    //    --- this needs more coding before testing
     //  . verify checkpoints
     //   . replay
+    //  . en-to-end tests to not only verify row num but also data
```
tustvold, line 902:
```suggestion
    //  . end-to-end tests to not only verify row num but also data
```

generated_types/protos/influxdata/iox/management/v1/chunk.proto
crepererum: I think protobuf enums are solely marker/integer-based and cannot contain any payload. So I think we need a new field on `Chunk` or even better a new struct:

```proto
message ChunkLifecycleAction {
  enum Type {
    // No lifecycle
    CHUNK_LIFECYCLE_ACTION_UNSPECIFIED = 0;

    /// Chunk is in the process of being written to object storage
    CHUNK_LIFECYCLE_ACTION_PERSISTING = 2;

    /// Chunk is in the process of being compacted
    CHUNK_LIFECYCLE_ACTION_COMPACTING = 3;

    ...
    }

    Type type = 1;
    bytes target_chunk_id = 2;
}
```

Since the protobuf is only used between our CLI client and the server and we don't preserve the data anywhere, I think this somewhat bigger change is OK.

tustvold (approved):
I _think_ this is good, but its a bit late in the day so I might have missed something. Will look through again tomorrow morning

server/src/db/lifecycle/compact_object_store.rs
NGA-TRAN: I asked myself the same question. It was done in both `compact` and `persist` chunk. @crepererum has some reason for it?

generated_types/protos/influxdata/iox/management/v1/chunk.proto
NGA-TRAN: I will address this in a different PR

alamb: Different PR is https://github.com/influxdata/influxdb_iox/pull/3275

Issue #3251, opened 2021-11-29
Use gRPC Error Model in gRPC Client


tustvold:
**Problem**

Currently within influxdb_iox_client we create an error enumeration for every method. This results in a huge amount of largely copy-pasted error boilerplate. With exception to the flight API, the gRPC client is pretty much entirely error plumbing. 

Aside from all the standard reasons DRY is good, the sheer wall of error handling hides the fact some of it is wrong.

For example, `unload_partition_chunk` converts all responses with a `FailedPrecondition` status code as:

```
#[error("Cannot perform operation due to wrong chunk lifecycle state: {}", .0.message())]
    LifecycleError(tonic::Status),
```

This is wrong, as `FailedPrecondition` is also used to signify that the writer ID is not set.

**Proposal**

The designers of gRPC already spent invested a significant amount of effort into making meaningful error variants, the clients should be switched over to a single error enumeration with the following variants using the detail payloads found in `generated_types::google`

```
pub enum Error {
    FieldViolation(FieldViolation),
    InternalError(InternalError),
    AlreadyExists(AlreadyExists),
    NotFound(NotFound),
    PreconditionViolation(PreconditionViolation),
    QuotaFailure(QuotaFailure),
    MissingDetails(tonic::Code),
}
```

Aside from eliminating a large amount of boilerplate, this accurately reflects the error model sent over the wire, and therefore avoids the problem of drifting error mappings

alamb:
From my perspective, as long as the details of the original message are not lost (aka what actually happened if an `InternalError` is returned) this sounds like an improvement to me

domodwyer:
Totally up for using the google errors for the on-the-wire serialisation - standards! 🎉 

I think it'd be a big step backwards in usability for the end user of the client if the API client no longer mapped these errors to something semantically meaningful though. As an example, at the moment deleting a database presents an enumeration of possible outcomes:

```rust
        match client.release_database(name.into(), uuid).await{
            Ok(_) => ...
            Err(ApiError::DatabaseNotFound { .. }) => ...
            Err(ApiError::NoServerId) => ...
            Err(ApiError::InvalidUuid { .. }) => // Name matches, UUID does not ...
            Err(ApiError::ServerError { .. }) => ...
        }
```

Which is descriptive, concise and an absolute treat to use and programatically match on! If these error types were removed, we'd have to match against the entire set of possible google errors, and pick apart each one and match on the types & strings inside them:

```rust
[...]

PR #3253, opened 2021-11-29
feat: rebuild catalog (#3225)

### LINKED ISSUE #3225: Rebuild Catalog

tustvold:
Closes #3225

Adds the ability to rebuild a catalog.

I tested this locally by running a server with

```
cargo run -- run -vvv database --object-store file --data-dir ~/.iox/data/ --server-id 1000 --wipe-catalog-on-error no
```

I then wrote some data into it, and persisted it

```
cargo run database partition persist DB_NAME TABLE_NAME PARTITION_KEY --force
```

I then stopped the server running and nvaigated to the database's directory within `~/.iox/data`. I then ran

```
rm -rf transactions
mkdir -p transactions/00000000000000000000/
echo "INVALID" > transactions/00000000000000000000/78451672-db8c-4f7d-bcca-a4e1d8b5a74d.txn
```

I then started the server up again with the same command and it failed to startup (as expected)

```
2021-11-29T22:21:42.223538Z ERROR server::database: database in error state - operator intervention required db_name=my_database e=error loading catalog: Cannot load preserved catalog: Error during protobuf IO: Error during protobuf deserialization: failed to decode Protobuf message: Transaction.previous_uuid: invalid wire type: SixtyFourBit (expected LengthDelimited) state=CatalogLoadError
```

From a separate terminal I then ran

```
cargo run -- database recover rebuild my_database
```

Then

```
cargo run -- operation wait 0
```

And once that finished I was able to query the database, and all the chunks had been discovered :tada:


influxdb_iox/src/commands/database/recover.rs
```diff
@@ -72,7 +87,7 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
             }
 
             let operation = client
-                .wipe_persisted_catalog(db_name)
+                .wipe_preserved_catalog(db_name)
```
tustvold, line 90: Some parts of the system referred to it as the "persisted catalog" and others the "preserved catalog". I went with the latter as this reflects the structures in parquet_catalog and the gRPC API.

influxdb_iox_client/src/client/management.rs
```diff
@@ -298,7 +298,31 @@ pub enum GetServerStatusError {
 
 /// Errors returned by [`Client::wipe_persisted_catalog`]
 #[derive(Debug, Error)]
-pub enum WipePersistedCatalogError {
+pub enum WipePreservedCatalogError {
```
tustvold, line 301: #3251 will avoid this silliness

parquet_catalog/src/rebuild.rs
```diff
@@ -65,14 +65,10 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 ///
 /// - **Metadata Read Failure:** There is a parquet file with metadata that cannot be read. Set
 ///   `ignore_metadata_read_failure` to `true` to ignore these cases.
-pub async fn rebuild_catalog<S>(
+pub async fn rebuild_catalog(
```
tustvold, line 68: FYI @crepererum I couldn't see a strong reason that this needed to take a CatalogState and so I removed this as it made it easier to use. Please let me know if I'm missing something here

crepererum: The reason is pretty simple: you probably wanna use the rebuilt catalog straight away. Now you're basically loading it twice.

influxdb_iox/tests/common/server_fixture.rs
```diff
@@ -304,6 +305,20 @@ impl ServerFixture {
     pub fn dir(&self) -> &Path {
         self.server.dir.path()
     }
+
+    /// Poison persisted catalog
+    pub fn poison_catalog(&self, database_uuid: Uuid) {
```
crepererum, line 310: We already have code for that:

https://github.com/influxdata/influxdb_iox/blob/f165faf5bce152bef6676f4af7fef699c5cfd085/parquet_catalog/src/test_helpers.rs#L234

server/src/database.rs
```diff
@@ -532,6 +542,54 @@ impl Database {
...
+                    Arc::clone(shared.application.time_provider()),
+                );
+                parquet_catalog::rebuild::rebuild_catalog(config, false)
+                    .await
+                    .map_err(Box::new)
+                    .context(RebuildPreservedCatalog { db_name })?;
+
+                {
+                    let mut state = shared.state.write();
+                    *state.unfreeze(handle) = DatabaseState::RulesLoaded(current_state);
```
crepererum, line 582: If you would return the catalog state from `rebuild_catalog` (as it was done before) you could advance the state even further.

parquet_catalog/src/rebuild.rs
tustvold: Is that really a problem? I can change back, but it forces the catalog rebuild to be coordinated by `Db` as `Loader` and friends are private implementation details of it? This seems more complicated?

influxdb_iox/tests/common/server_fixture.rs
tustvold: This server fixture does not have access to a PreservedCatalog... Its in a separate process

parquet_catalog/src/rebuild.rs
crepererum: Let's be pragmatic then :+1:

crepererum (approved):

alamb:
FWIW I am going to work on testing this feature by documenting my usecase of "how to rebuild a part of a production cluster"

PR #3258, opened 2021-11-30
refactor: cleanup Db write path error handling

tustvold:
Following @crepererum work to split the write buffer producer out of `Db` and the routing logic out of `Database` we can start to hack away at the great wall of error plumbing. In particular there are now only a small number of possible error variants for a write, and we don't need to route writes via `Database` at all :tada:


influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -661,14 +661,12 @@ pub mod test_utils {
                 .await
                 .unwrap();
 
-            // This currently results in an InternalServerError which is correctly recorded
-            // as a server error, but this should probably be a BadRequest client error (#2538)
```
tustvold, line 665: :tada: Accidental bug fixes are the best bug fixes

server/src/db.rs
```diff
@@ -958,25 +957,29 @@ impl Db {
...
-            rules.lifecycle_rules.immutable
-        };
-        debug!(%immutable, "storing entry");
-        if immutable {
-            return DatabaseNotWriteable {}.fail();
+    /// Returns an error if this [`Db`] cannot accept a given DML operation
+    fn can_store(&self, meta: &DmlMeta) -> Result<(), DmlError> {
+        let rules = self.rules.read();
+        ensure!(!rules.lifecycle_rules.immutable, DatabaseNotWriteable);
+        if rules.write_buffer_connection.is_some() {
```
tustvold, line 964: The meta contains a sequence number iff it came from the write buffer, it's a bit of a cheat, but it codifies the property we actually care about - that we aren't mixing sequenced and unsequenced data

server/src/db.rs
```diff
@@ -160,6 +125,34 @@ pub enum Error {
...
+    TableNotFound {
+        table_name: String,
+        source: CatalogError,
+    },
+
+    #[snafu(display(
+        "Storing database write failed with the following error(s), and possibly more: {}",
+        errors.iter().map(ToString::to_string).collect::<Vec<_>>().join(", ")
+    ))]
+    SchemaErrors { errors: Vec<schema::merge::Error> },
```
tustvold, line 153: The only errors that can occur on write are schema errors :tada:

server/src/db.rs
```diff
@@ -537,7 +532,11 @@ impl Db {
     /// Returns an error if the table cannot be found in the catalog
     ///
     /// **WARNING: Only use that when no write buffer is used.**
-    pub fn delete(&self, table_name: &str, delete_predicate: Arc<DeletePredicate>) -> Result<()> {
+    pub fn delete(
```
tustvold, line 535: I want to remove this method in a future PR

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
```diff
@@ -176,52 +176,26 @@ impl HttpDrivenDml for DatabaseServerType {
...
+        let db = self
+            .server
+            .db(db_name)
+            .map_err(|_| InnerDmlError::DatabaseNotFound {
+                db_name: db_name.to_string(),
+            })?;
+
+        db.store_operation(&op).map_err(|e| match e {
+            // TODO: missing tables should not be special-cased (#3259)
+            server::db::DmlError::TableNotFound { table_name, .. } => {
```
alamb, line 188: well that is certainly a lot nicer

alamb (approved):
I reviewed the code and the tests carefully and it looks like a significant improvement to me

crepererum (approved):

PR #3260, opened 2021-11-30
refactor: add resource type enumeration (#3251)

tustvold:
Use an enumeration for ResourceType, this helps to ensure consistency and may form part of a story to simplify the client errors (#3251).


influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/deployment.rs
```diff
@@ -22,7 +23,7 @@ impl deployment_service_server::DeploymentService for DeploymentService {
     ) -> Result<Response<GetServerIdResponse>, Status> {
         match self.server.server_id() {
             Some(id) => Ok(Response::new(GetServerIdResponse { id: id.get_u32() })),
-            None => return Err(NotFound::default().into()),
+            None => return Err(NotFound::new(ResourceType::ServerId, Default::default()).into()),
```
tustvold, line 26: Returning NotFound here is a little bit strange, but I guess harmless :smile:

crepererum (approved):

generated_types/src/google.rs
```diff
@@ -238,15 +238,67 @@ impl From<InternalError> for tonic::Status {
     }
 }
 
+/// A resource type within the gRPC error model
+#[derive(Debug, Clone)]
+pub enum ResourceType {
```
alamb, line 243: Is it worth pointing out somewhere (docstrings?) that this is IOx specific extension (I found it a little confusing that `Chunk` appeared in a module named `google`

alamb (commented):
Looks like an improvement to me 👍

PR #3262, opened 2021-11-30
fix: GIT version in `--version`

crepererum:
The GIT version was always "UNKNOWN", a regression introduced by #2965.
Also adds a regression test.


influxdb_iox/tests/end_to_end_cases/debug_cli.rs
```diff
@@ -28,6 +28,20 @@ async fn test_dump_catalog() {
...
 
+#[tokio::test]
+async fn test_git_version() {
+    Command::cargo_bin("influxdb_iox")
+        .unwrap()
+        .arg("--version")
+        .assert()
+        .success()
+        .stdout(
+            predicate::str::contains("UNKNOWN")
```
alamb, line 39: ❤️

alamb (approved):
Awesome test

PR #3263, opened 2021-11-30
fix: don't error on delete to non-existent table (#3259)

### LINKED ISSUE #3259: Silently Ignore Delete Predicates with Non-Existent Tables
**Problem**

Currently if a delete specifies a table name predicate, and this table is not found, IOx returns an error. For the HTTP endpoint this is translated to a 404, for gRPC a precondition violation?!

From @e-dard 

> Right now, if we are restricting ourselves to the delete operation API (which we use to implement delete support in Cloud) then I would expect to get a 204 with no data deleted.

Given the table name is just a restriction on the predicate, this also makes sense to me

**Proposal**

Silently drop delete predicates with tables that don't exist, in much the same way as we would silently drop delete predicates with timestamp ranges that don't match any chunks.

tustvold:
Closes #3259

Table predicates should not be special cased w.r.t error reporting


e-dard (approved):
LGTM 👍

PR #3265, opened 2021-12-01
feat: cleanup gRPC server errors (#3251) 

tustvold:
This PR reworks the PreconditionViolation error to be an enumeration of possible violations returned by IOx, this should make it easier to maintain consistency.


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/error.rs
```diff
@@ -5,29 +5,28 @@ use observability_deps::tracing::error;
...
+    use server::Error;
 
     match error {
-        Error::IdNotSet => PreconditionViolation {
-            category: "Writer ID".to_string(),
-            subject: "influxdata.com/iox".to_string(),
-            description: "Writer ID must be set".to_string(),
-        }
+        Error::IdNotSet => PreconditionViolation::ServerIdNotSet.into(),
+        Error::DatabaseNotInitialized { db_name } => PreconditionViolation::DatabaseInvalidState(
```
tustvold, line 12: Database and Server have been changed to return PreconditionViolation in the event they aren't initialized, this reflects the fact we don't actually have sufficient context to know if they are either still initializing, in which case Unavailable is correct, or stuck in an error state, in which case it is not. Unavailable implies a simple retry of the request will eventually work, returning PreconditionViolation errs on the side of caution. I've created #3266 to track this

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/write_pb.rs
```diff
@@ -34,14 +34,10 @@ impl write_service_server::WriteService for PBWriteService {
...
-                category: String::from("router"),
-                subject: String::from("influxdata.com/iox"),
-                description: e.to_string(),
-            }
-            .into()
-        })?;
+        router
+            .write(write)
+            .await
+            .map_err::<tonic::Status, _>(|e| tonic::Status::aborted(e.to_string()))?;
```
tustvold, line 40: A routing error isn't a precondition failure. I _think_ aborted is the closest to the semantics that we wish to convey, namely that the operation was potentially partially successful and needs to be retried at a higher level than a simple blind retry which would be signified by Unavailable

generated_types/src/google.rs
```diff
@@ -371,27 +371,93 @@ impl From<NotFound> for tonic::Status {
     }
 }
 
-#[derive(Debug, Default, Clone)]
-pub struct PreconditionViolation {
-    pub category: String,
-    pub subject: String,
-    pub description: String,
+#[derive(Debug, Clone)]
+pub enum PreconditionViolation {
```
alamb, line 375: Is it fair to say that this is effectively making a Rust `Error` struct with IOx specific semantics which is serialized using the generic Protobuf messages?

alamb:
```suggestion
/// This is an IOx specific Error which is serialized using the generic google Protobuf encoding
pub enum PreconditionViolation {
```

generated_types/src/google.rs
```diff
@@ -371,27 +371,93 @@ impl From<NotFound> for tonic::Status {
...
+        subject: String,
+        description: String,
+    },
+}
+
+impl PreconditionViolation {
+    fn description(&self) -> String {
+        match self {
+            Self::ServerIdNotSet => "Server ID must be set".to_string(),
+            Self::DatabaseImmutable => "database must be mutable".to_string(),
```
alamb, line 400: Nit: should it be "Database" (with a capital) to maintain consistency?

generated_types/src/google.rs
```diff
@@ -371,27 +371,93 @@ impl From<NotFound> for tonic::Status {
...
+            Self::ChunkInvalidState(description) => description.clone(),
+            Self::Unknown { description, .. } => description.clone(),
+        }
+    }
 }
 
-fn encode_precondition_failure(violations: Vec<PreconditionViolation>) -> Result<Any, EncodeError> {
-    use rpc::precondition_failure::Violation;
+impl From<PreconditionViolation> for rpc::precondition_failure::Violation {
+    fn from(v: PreconditionViolation) -> Self {
```
alamb, line 411: Should there be the alternate code too (aka `impl TryFrom<rpc::precodition_failure::Violation for PreconditionViolation`) so that users of `influxdb_iox_client` can get structured error types as well?

Or perhaps there is one that I have missed

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/error.rs
```diff
@@ -90,9 +94,9 @@ pub fn default_database_error_handler(error: server::database::Error) -> tonic::
             error!(%source, "Unexpected error while rebuilding catalog");
             InternalError {}.into()
         }
-        Error::RulesNotUpdateable { .. } => tonic::Status::failed_precondition(error.to_string()),
-        Error::CannotPersistUpdatedRules { .. } => {
-            tonic::Status::failed_precondition(error.to_string())
+        Error::CannotPersistUpdatedRules { source } => {
```
alamb, line 97: This new pattern certainly looks much clearer to me

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/write_pb.rs
alamb: I think most of this error handling plumbing was a matter of "follow existing patterns"

Aborted seems strange to me for what amounts to a configuration failure, right? At least the test that is changed lower in this PR has a misconfigured router. 

Also, do you envision a similar `Aborted` struct for enumerating possible types of abort messages?

alamb (approved):

influxdb_iox/src/influxdb_ioxd/server_type/router/rpc/write_pb.rs
tustvold: > Aborted seems strange to me for what amounts to a configuration failure, right

I had actually assumed the errors here were write errors, i.e. can't connect to upstream errors. I'm not sure why configuration errors would manifest in the write path and not at configuration time...

generated_types/src/google.rs
tustvold: Client changes will follow in a separate PR :smile:

PR #3269, opened 2021-12-01
feat: `jaeger-debug-id` from data generator

crepererum:


influxdb2_client/src/lib.rs
```diff
@@ -134,6 +144,12 @@ impl Client {
         if let Some(auth) = &self.auth_header {
             req = req.header("Authorization", auth);
         }
+        if self.jaeger_debug {
+            req = req.header(
+                "jaeger-debug-id",
```
primeroz, line 149: if what we are trying to do is generate traces to be looked up in `https://tracing.aws.influxdata.io` this needs to be `influx-debug-id`

crepererum: fixed

iox_data_generator/src/bin/iox_data_generator.rs
```diff
@@ -139,6 +139,12 @@ Logging:
                 .help("Generate this many samplings to batch into a single API call. Good for sending a bunch of historical data in quickly if paired with a start time from long ago.")
                 .takes_value(true)
         )
+        .arg(
+            Arg::with_name("jaeger_debug")
+                .long("jaeger_debug")
+                .help("Generate jaeger debug header with given key during write")
```
alamb, line 145: I wonder if we care about sampling or something so that we don't end up generating traces for all writes

iox_data_generator/src/bin/iox_data_generator.rs
```diff
@@ -139,6 +139,12 @@ Logging:
                 .help("Generate this many samplings to batch into a single API call. Good for sending a bunch of historical data in quickly if paired with a start time from long ago.")
                 .takes_value(true)
         )
+        .arg(
+            Arg::with_name("jaeger_debug")
```
alamb, line 143: I recommend keeping the name consistent with the parameter, so `"jaeger_debug_header"`

alamb (approved):

iox_data_generator/src/bin/iox_data_generator.rs
primeroz: yes please, at least make it configurable so we can turn it down if it sarts to overwhelm the jaeger environment. 

At the moment the service is designed to be small and mostly be used with `forced header` so it will be easy to overwhelm.

crepererum: It's basically a all-or-nothing switch which I think serves the purpose at the moment since it's opt-in. If you wanna generate data and track their flow through the system (we also track it cross-Kafka btw.), just set this header. But for most load tests, you probably don't wanna do that.

iox_data_generator/src/bin/iox_data_generator.rs
crepererum: done

iox_data_generator/src/bin/iox_data_generator.rs
primeroz: i know this is a weird request but can you try and let me know ( just mentione me in the PRs ) when tracing is enabled for iox ? 

I just want to see the effect of the Jaeger Server setup

PR #3271, opened 2021-12-01
feat: decode error details payloads (#3251)

tustvold:
Adds the logic necessary to decode the gRPC error details payloads, a follow on PR will then plumb these into the influxdb_iox_client.

As explained in the module doc, these error payloads are meant to be extra icing on the cake. For the most part the gRPC API should be such that status codes alone are sufficient, with clients not forced to decode these payloads in order to implement a functional client. However, we are producing these errors so we might as well expose them.


generated_types/src/google.rs
```diff
@@ -647,3 +766,64 @@ impl<T> OptionalField<T> for Option<T> {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bytes::Bytes;
+
+    #[test]
+    fn test_error_roundtrip() {
```
domodwyer, line 776: This is the stuff! ❤️

domodwyer (approved):

PR #3272, opened 2021-12-01
fix: accept PB batches w/ trimmed repeated tail values

### LINKED ISSUE #3186: `DmlMeta::bytes_read` for write buffer read and write doesn't match
Actually checking that `bytes_read` as reported by the writer is the same as reported by the reader leads to failures. This is mostly because different sizes are measured (w/ or w/o headers).

crepererum:
Fixes #3186.


crepererum:
Ups, tracking the wrong issue here

PR #3273, opened 2021-12-01
fix: accept PB batches w/ trimmed repeated tail values

### LINKED ISSUE #3128: IOx PbData Repeated Values
IOx does not currently support repeated trailing values in pbdata payloads as described [here](https://github.com/influxdata/influxdb-pb-data-protocol#optimization-2-trim-repeated-tail-values). It will instead return an error if it is sent such a payload

crepererum:
Fixes #3128.


mutable_batch_pb/src/decode.rs
```diff
@@ -192,6 +201,84 @@ pub fn write_table_batch(batch: &mut MutableBatch, table_batch: &TableBatch) ->
     Ok(())
 }
 
+/// Inner state of [`RepeatLastElement`].
```
tustvold, line 204: I wonder if this could be written more concisely using a combination of std::iter::from_fn and Peekable?

Something like

```
fn repeat_last<I>(i: impl Iterator<Item=I>) -> impl Iterator<Item=I>{
    let mut iter = i.peekable();
    let mut last = iter.peek();
    std::iter::from_fn(move || {
        match iter.peek() {
            Some(n) => {
                last = Some(n.clone);
                n
            }
            None => last,
        }
    })
}
```

mutable_batch_pb/src/decode.rs
```diff
@@ -790,4 +873,200 @@ mod tests {
             "column \"tag3\" contains invalid offset 3 at index 2",
         );
     }
+
+    #[test]
```
tustvold, line 877: :ok_hand:

mutable_batch_pb/src/decode.rs
```diff
@@ -790,4 +873,200 @@ mod tests {
...
+            "| 1970-01-01T00:00:00.000000008Z |",
+            "| 1970-01-01T00:00:00.000000009Z |",
+            "+--------------------------------+",
+        ];
+
+        assert_batches_eq!(expected, &[batch.to_arrow(Selection::All).unwrap()]);
+    }
+
+    #[test]
+    fn test_optimization_trim_repeated_tail_values() {
```
tustvold, line 958: This test made my head hurt, but it's good :1st_place_medal: I think...

tustvold (approved):
Looks good to me, I think you could write the iterator more concisely but up to you

mutable_batch_pb/src/decode.rs
crepererum: It just ensures that I've added the iterator wrapper to all places, hence the large number of columns.

tustvold: It was the interplay of the truncated null masks, with the truncated iterators that was rather head hurty :laughing:

mutable_batch_pb/src/decode.rs
crepererum: I think it's not that easy, because you still need more state because `peek()` doesn't clone but only references the next element. And you don't wanna just blindly clone on every `peek()` because for strings that's quite expensive. To avoid that you again need some state enum and while you might get away w/ slightly less state (because `next` can be moved into `Peekable`), complexity-wise it's about the same.

I honestly wished that this wrapper would be offered by upstream (std or itertools).

tustvold: > because for strings that's quite expensive

Are they not `&str` here? The writer shouldn't need owned strings...

PR #3274, opened 2021-12-01
refactor: not persist empty chunk resulting from deleting & deduplicating

NGA-TRAN:
This was found during PR #3252 


parquet_file/src/storage.rs
```diff
@@ -145,31 +145,6 @@ impl Storage {
...
-    /// a specified partitioned chunk to a parquet file of this storage
-    ///
-    /// returns the path to which the chunk was written, the size of
-    /// the bytes, and the parquet metadata
-    pub async fn write_to_object_store_if_having_data(
-        &self,
-        chunk_addr: ChunkAddr,
-        stream: SendableRecordBatchStream,
-        metadata: IoxMetadata,
     ) -> Result<Option<(ParquetFilePath, usize, IoxParquetMetaData)>> {
```
alamb, line 148: It might help to update the docstring of this function to explain that `Ok(None)` is returned if there was no parquet file created because there was no data to persist after applying deletes

alamb (approved):
Looks like a nice clean up to me

parquet_file/src/storage.rs
NGA-TRAN: Will do

PR #3275, opened 2021-12-01
feat: add chunk Id into lifecycle action proto to have it work with lifecycle action for compacting os chunks

NGA-TRAN:
This is a followup of #3252

I found this PR a bit wired but if we go with different solution, we will deal with similar weirdness elsewhere I guess.
This is the context for this PR: https://github.com/influxdata/influxdb_iox/pull/3252/files#r759733797

What do you think @alamb , @crepererum , @tustvold ?

### DIFF
--- a/data_types/src/chunk_metadata.rs
+++ b/data_types/src/chunk_metadata.rs
@@ -225,10 +225,6 @@ impl ChunkId {
         Self(Uuid::from_u128(id))
     }
 
-    pub fn new_uuid(id: Uuid) -> Self {
-        Self(id)
-    }
-
     /// Get inner UUID.
     pub fn get(&self) -> Uuid {
         self.0
--- a/data_types/src/chunk_metadata.rs
+++ b/data_types/src/chunk_metadata.rs
@@ -225,6 +225,10 @@ impl ChunkId {
         Self(Uuid::from_u128(id))
     }
 
+    pub fn new_uuid(id: Uuid) -> Self {
+        Self(id)
+    }
+
     /// Get inner UUID.
     pub fn get(&self) -> Uuid {
         self.0
--- a/generated_types/Cargo.toml
+++ b/generated_types/Cargo.toml
@@ -15,7 +15,6 @@ regex = "1.4"
 serde = { version = "1.0", features = ["derive"] }
 tonic = "0.5"
 time = { path = "../time" }
-uuid = { version = "0.8", features = ["v4"] }
 workspace-hack = { path = "../workspace-hack"}
 
 [dev-dependencies]
--- a/generated_types/Cargo.toml
+++ b/generated_types/Cargo.toml
@@ -15,6 +15,7 @@ regex = "1.4"
 serde = { version = "1.0", features = ["derive"] }
 tonic = "0.5"
 time = { path = "../time" }
+uuid = { version = "0.8", features = ["v4"] }
 workspace-hack = { path = "../workspace-hack"}
 
 [dev-dependencies]
--- a/generated_types/protos/influxdata/iox/management/v1/chunk.proto
+++ b/generated_types/protos/influxdata/iox/management/v1/chunk.proto
@@ -27,22 +27,22 @@ enum ChunkStorage {
 
 enum Action {
   // No lifecycle
-  UNSPECIFIED = 0;
+  ACTION_UNSPECIFIED = 0;
 
   /// Chunk is in the process of being written to object storage
-  PERSISTING = 2;
+  ACTION_PERSISTING = 2;
 
   /// Chunk is in the process of being compacted
-  COMPACTING = 3;
+  ACTION_COMPACTING = 3;
 
   /// Chunk is about to be dropped from memory and (if persisted) from object store.
-  DROPPING = 4;
+  ACTION_DROPPING = 4;
 
   /// Chunk is in the process of being compacted
-  COMPACTING_OBJECT_STORE = 5;
+  ACTION_COMPACTING_OBJECT_STORE = 5;
 
   /// Chunk is being loaded into the read buffer
-  LOADING_READ_BUFFER = 6;
+  ACTION_LOADING_READ_BUFFER = 6;
 }
 
 // Is there any lifecycle action currently outstanding for this chunk?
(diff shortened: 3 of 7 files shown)


NGA-TRAN:
I will fix the failure in protobuf-lint if we agree to go with this solution

data_types/src/chunk_metadata.rs
```diff
@@ -225,6 +225,10 @@ impl ChunkId {
         Self(Uuid::from_u128(id))
     }
 
+    pub fn new_uuid(id: Uuid) -> Self {
```
crepererum, line 228: I think you can just use the `TryFrom<Bytes>` implementation from this file.

generated_types/protos/influxdata/iox/management/v1/chunk.proto
```diff
@@ -25,30 +25,33 @@ enum ChunkStorage {
...
+  LOADING_READ_BUFFER = 6;
+}
+
+// Is there any lifecycle action currently outstanding for this chunk?
+message ChunkLifecycleAction {
+  Action action = 1;
+  
+  // The id of to-be-created chunk.
+  // UUID is stored as 16 bytes in big-endian order.
+  bytes target_chunk_id = 2;
```
crepererum, line 54: You probably want to make sure this is represented as `Bytes` instead of `Vec<u8>` in Rust. Just add a new entry to this list here:

https://github.com/influxdata/influxdb_iox/blob/f085af034ee411e6e8357955ab7048607a3bdb55/generated_types/build.rs#L83-L100

generated_types/Cargo.toml
```diff
@@ -15,6 +15,7 @@ regex = "1.4"
 serde = { version = "1.0", features = ["derive"] }
 tonic = "0.5"
 time = { path = "../time" }
+uuid = { version = "0.8", features = ["v4"] }
```
crepererum, line 18: Assuming you apply all suggestions, I think you don't even need this dependency here.

generated_types/src/chunk.rs
```diff
@@ -61,15 +62,32 @@ impl From<ChunkStorage> for management::ChunkStorage {
 
 impl From<Option<ChunkLifecycleAction>> for management::ChunkLifecycleAction {
     fn from(lifecycle_action: Option<ChunkLifecycleAction>) -> Self {
+        let random_uuid = ChunkId::new().get().as_bytes().to_vec();
```
crepererum, line 65: No need to use a random UUID here. Since the UUID is unused in this case, just use an empty byte string (`Bytes::new()`), which is the protobuf-version of "missing / not specified".

generated_types/src/chunk.rs
```diff
@@ -150,22 +167,30 @@ impl TryFrom<management::ChunkLifecycleAction> for Option<ChunkLifecycleAction>
...
-            management::ChunkLifecycleAction::Unspecified => Ok(None),
+        let management::ChunkLifecycleAction {
+            action,
+            target_chunk_id,
+        } = proto;
+
+        let chunk_id: [u8; 16] = target_chunk_id.try_into().unwrap_or_else(|v: Vec<u8>| {
+            panic!("Expected a Vec of length {} but it was {}", 16, v.len())
+        });
+        let chunk_id = Uuid::from_bytes(chunk_id);
```
crepererum, line 178: I would only evaluate the `target_chunk_id` when you actually expect it, i.e. only for `action == management::Action::CompactingObjectStore.into()`.

tustvold:
I can't help feeling exposing what is an implementation detail of the interaction between deletes and compacting persisted chunks in the protobuf API is a pretty major smell. Was there a reason we can't just add an `Option<ChunkId>` to `CatalogChunk` and leave it like that? This seems like quite a lot of churn for something I'm not even sure we want to expose?

crepererum:
> I can't help feeling exposing what is an implementation detail of the interaction between deletes and compacting persisted chunks in the protobuf API is a pretty major smell. Was there a reason we can't just add an `Option<ChunkId>` to `CatalogChunk` and leave it like that? This seems like quite a lot of churn for something I'm not even sure we want to expose?

We could also add it to the chunk itself, I don't mind TBH.

alamb:
I am sorry I am late to the party, but I don't fully understand the rationale for this change so I don't feel like I can offer an opinion on the code

Is the rationale
1. So we can write end to end tests that interact with IOx via the management API?
2. so that in some future a client could run its own "what chunks to compact" algorithm?
3. Something else?

NGA-TRAN:
@alamb 

> Is the rationale
>1. So we can write end to end tests that interact with IOx via the management API?
>2. so that in some future a client could run its own "what chunks to compact" algorithm?
>3. Something else?

This is something else. It is for ChunkLifeCycleAction that @crepererum and @tustvold know better why we need to put it in Management API for our own internal communication (not users). As in this solution, the future to-be-compacted chunkID is kept in there, too, which I think is not that bad. However, we may want to keep that ID hidden from this API by adding it into the catalog chunk.

Regardless, I am glad I worked on this PR because I learned a lot of in the proto and rust. @crepererum's hints to make it better is super helpful. I think I will address Marco comments for me to learn more to handle with photo and rust (will commit in branch won't  merge). Then I will go with the second solution because I think the second solution will help with future optimistic concurrency control that I will create a discussion after this

crepererum:
@alamb I think #3255 might answer your question. Tl;Dr;: While compacting OS chunks, we need to know a priori what the UUID of the target chunk is so we don't loose in-flight delete predicates. This PR is about the question where we keep this UUID (as part of the lifecycle action VS as part of the chunk).

NGA-TRAN:
I have addressed review comments but let us not merge this and explore adding chunk ID in the CatalogChunk instead

alamb:
> @alamb I think #3255 might answer your question. Tl;Dr;: While compacting OS chunks, we need to know a priori what the UUID of the target chunk is so we don't loose in-flight delete predicates. This PR is about the question where we keep this UUID (as part of the lifecycle action VS as part of the chunk).

Thanks @crepererum  and @NGA-TRAN  -- I think I am still missing "why do we want to expose this inprogress chunk id in the management API?"

NGA-TRAN:
@alamb We do not want to expose it. It just happens to be a part of a structure that is mapped to a proto message. The other solution #3285 does not expose it but eventually we may want to expose it in `ChunkSummary` which also has a counterpart proto message. LOL

tustvold:
Closing as went with #3285

Issue #3266, opened 2021-12-01
Return gRPC Unavailable Whilst Initializing


tustvold:
#3265 made it so that a PreconditionViolation is returned if the server/database isn't initialized. This errs on the side of caution, as it doesn't claim that a simple retry will fix an initialization error. However, it might improve user experience if we returned Unavailable whilst the initializing and not encountered an error.

PR #3277, opened 2021-12-02
refactor: cleanup gRPC client error handling (#3251)

### LINKED ISSUE #3251: Use gRPC Error Model in gRPC Client

tustvold:
Closes #3251

This PR updates the client to use a new custom `Error` type in place of `tonic::Status` that uses the logic added in #3271 to decode the details payloads. 

An earlier cut of this then wrapped these in a semantic error wrapper that "projected" the variants of this that clients would be expected to handle for that endpoint. In practice as I wired this up I discovered there were actually only ever three variants of interest:

* NotFound - for resource based operations
* AlreadyExists - for create operations
* InvalidArgument - for incorrect data specified

The rest all fell in the bucket of either IOx will never return them, or if it did there isn't much that the recipient can be expected to do in response. I felt on balance this did not warrant the additional complexity of the error classification system - but I can add it back in a follow on PR if people feel strongly.

This PR therefore achieves a couple of goals:

* It ensures that the endpoints all expose a minimal baseline of error variants - a number simply treated everything as a ServerError :sweat_smile:
* It ensures the error semantics are uniform across the client, and by extension our APIs
* Related to the above, it forces us to design the APIs in such a way that the gRPC error model is sufficient
* Significantly reduces the amount of plumbing code necessary to add a gRPC endpoint and expose it properly in the client

Once this is merged I will update conductor with these changes

Note: this does not yet update the flight client, I will update that at a later date - it actually has some logic that isn't just error plumbing :laughing:


influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -965,7 +974,10 @@ async fn test_close_partition_chunk_error() {
         .await
         .expect_err("expected error");
 
-    assert_contains!(err.to_string(), "Database not found");
+    assert_contains!(
+        err.to_string(),
+        "Resource database/this database does not exist not found"
```
tustvold, line 979: Created #3278 to track making this error message less confusing

influxdb_iox/tests/end_to_end_cases/deployment_api.rs
```diff
@@ -80,12 +74,8 @@ async fn test_serving_readiness_router() {
         .set_serving_readiness(false)
         .await
         .unwrap();
-    let err = write_client.write_lp(name, lp_data, 0).await.unwrap_err();
-    assert!(
-        matches!(&err, WriteError::ServerError(status) if status.code() == Code::Unavailable),
```
tustvold, line 85: This is one of the many examples of client error variants that practically treat everything as an unexpected server error

influxdb_iox/src/commands/database.rs
```diff
@@ -66,6 +45,9 @@ pub enum Error {
 
     #[error("Error in partition subcommand: {0}")]
     Catalog(#[from] recover::Error),
+
+    #[error("Client error: {0}")]
+    ClientError(#[from] influxdb_iox_client::error::Error),
```
tustvold, line 50: I opted to collapse this into a single variant with no additional context, as it is sort of implicit from the command being run. This is one of the nice things from using a single error, the users can add context if they wish to, but aren't forced to plumb through every variant separately

influxdb_iox_client/src/client/error.rs
```diff
@@ -0,0 +1,139 @@
...
+impl ServerErrorDetails for PreconditionViolation {
+    fn try_decode(status: &tonic::Status) -> Option<Self> {
+        generated_types::google::decode_precondition_violation(status).next()
+    }
+}
+
+/// The errors returned by this client
+#[derive(Error, Debug)]
+#[allow(missing_docs)]
+pub enum Error {
```
tustvold, line 61: I debated splitting off some of these variants into an `UnexpectedError` variant, as a number are not things IOx will return. I decided against it because it would then be a breaking change if we did decide we wanted to use one of those variants.

I think it is quite a nice property that this is a stable error enumeration that will not change

crepererum:
Regarding the error types: I'm pretty sure that conductor uses "not available", which is issued while the server is booting up. It is important to differentiate that error from "internal" because "not available" can always safely be retried.

tustvold:
Yeah about that... - https://github.com/influxdata/influxdb_iox/issues/3266 :sweat_smile:

influxdb_iox_client/src/client/error.rs
crepererum: > I decided against it because it would then be a breaking change if we did decide we wanted to use one of those variants.

That depends on your definition of "breaking" (e.g. Rust type level VS client-server interaction). There's [`non_exhaustive`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) which is usually used for this case. But I agree that it's nice to map all the variants so error changes on the server side don't end up with too much work on the client side.

influxdb_iox_client/src/client/error.rs
```diff
@@ -0,0 +1,139 @@
...
+/// A generic opaque error
+pub type StdError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+/// A gRPC error payload with optional [details](https://cloud.google.com/apis/design/errors#error_details)
+#[derive(Error, Debug, Clone)]
+#[error("{message}")]
+pub struct ServerError<D> {
+    message: String,
+    details: Option<D>,
+}
```
crepererum, line 15: I think these two members should be `pub`, otherwise it's hard to read them out on the conductor side.

Is the `message` meant to be a full half-human-readable message and `details` are machine-readable / parsed?

I think (and that also goes back to the `pub` comment) it would be nice to have one end2end test that demonstrates the whole chain, e.g. issuing some client command (e.g. get database), match the error (e.g. `NotFound`), get the details and use the boilerplate from #3271 to decode the type enum (e.g. it's a database and has indeed the name that we've used during the GET call).

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/management.rs
```diff
@@ -106,10 +103,7 @@ impl management_service_server::ManagementService for ManagementService {
             .ok_or_else(|| FieldViolation::required("rules"))?;
 
         let provided_rules =
-            ProvidedDatabaseRules::new_rules(rules).map_err(|e| FieldViolation {
-                field: "rules".to_string(),
-                description: e.to_string(),
-            })?;
+            ProvidedDatabaseRules::new_rules(rules).map_err(|e| e.scope("rules"))?;
```
tustvold, line 106: Drive by fix, previously this would report all FieldViolation as being for `rules`. Whilst technically not incorrect, it's nicer if we can report the actual field that caused the issue

influxdb_iox_client/src/client/error.rs
tustvold: Done PTAL

influxdb_iox/tests/common/server_fixture.rs
```diff
@@ -638,13 +638,13 @@ impl TestServer {
 
                     let mut health = influxdb_iox_client::health::Client::new(channel);
 
-                    match health.check_deployment().await {
-                        Ok(_) => {
+                    match health.check_deployment().await.unwrap() {
```
alamb, line 641: By changing this API to a `bool`, how would a client (e.g. conductor) report why the server is not ready? 

When I have been using conductor to make clusters, when something goes wrong it is very helpful to have conductor / conductor's logs be able to report what is wrong rather than having go hunt down the specfic pod / iox server and try and get its logs

influxdb_iox/tests/end_to_end_cases/deployment_api.rs
alamb: Is the underlying server error message about what went wrong available in `r`? I am not saying it has to be some different `enum` variant, I am just very interested in ensuring that a user can see what went wrong.

(I also realize "what went wrong" may also not be available in the current error enums, so perhaps this is future work)

influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -373,7 +372,7 @@ async fn release_database() {
     let err = client.release_database(&db_name, None).await.unwrap_err();
     assert_contains!(
         err.to_string(),
-        format!("Could not find database {}", db_name)
+        format!("Resource database/{} not found", db_name)
```
alamb, line 405: I have seen the convention of `database/db_name` and now `database_uuid/uuid`, etc be used in error messages. That seems to imply to me that the client exposes some list of "entities" (aka types of things) such as `database`, `uuid`, etc

I wonder 🤔  should this list of things be documented anywhere / represented as actual types in `influxdb_iox_client`? (not for this PR, just musing)

influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -699,7 +701,10 @@ async fn test_partition_get_error() {
         .await
         .expect_err("expected error");
 
-    assert_contains!(err.to_string(), "Database not found");
+    assert_contains!(
+        err.to_string(),
+        "Resource database/this database does not exist not found"
```
alamb, line 745: 👍 for the message

influxdb_iox/tests/end_to_end_cases/management_api.rs
```diff
@@ -717,7 +722,11 @@ async fn test_partition_get_error() {
         .await
         .expect_err("exepcted error getting partition");
```
alamb, line 762:
```suggestion
        .expect_err("expected error getting partition");
```

alamb: nothing to do with this PR, just something I noticed

influxdb_iox_client/src/client/health.rs
```diff
@@ -39,32 +21,35 @@ impl Client {
...
             .await?
             .into_inner();
 
-        match status.status() {
-            ServingStatus::Serving => Ok(()),
-            ServingStatus::NotServing => Err(Error::NotServing),
-            _ => Err(Error::InvalidResponse(status.status)),
+        match response.status() {
+            ServingStatus::Serving => Ok(true),
+            ServingStatus::NotServing => Ok(false),
```
alamb, line 38: it would be great to include details here about *why* the server is not serving

influxdb_iox_client/src/client/management.rs
```diff
@@ -599,48 +153,32 @@ impl Client {
...
                 uuid: uuid.map(|u| u.as_bytes().to_vec()).unwrap_or_default(),
             })
-            .await
-            .map_err(|status| match status.code() {
-                tonic::Code::NotFound => ReleaseDatabaseError::DatabaseNotFound { name: db_name },
-                tonic::Code::FailedPrecondition => ReleaseDatabaseError::NoServerId,
-                tonic::Code::InvalidArgument => ReleaseDatabaseError::InvalidArgument(status),
-                _ => ReleaseDatabaseError::ServerError(status),
-            })?;
+            .await?;
```
alamb, line 181: that certainly looks nicer

alamb (approved):
I think this looks like a good change to me. Thank you @tustvold  for leading the charge

> In practice as I wired this up I discovered there were actually only ever three variants of interest:

> Once this is merged I will update conductor with these changes

👍  thank you. I wonder if it might be a good idea to try updating conductor on a branch to point at these changes prior to merging. I am not familiar enough with the conductor codebase to know for sure, but I thought there were a few places where it was trying to react to the specific error on the server side.

influxdb_iox/tests/common/server_fixture.rs
tustvold: This is all the context that the gRPC status endpoint exposes, it's a third-party gRPC [spec](https://github.com/influxdata/influxdb_iox/blob/main/generated_types/protos/grpc/health/v1/service.proto#L9). This change just makes the client more faithful to it

influxdb_iox/tests/end_to_end_cases/deployment_api.rs
tustvold: In this specific case the message payload will be "service not ready to serve" as this is what the ServingReadiness interceptor injects. In short the error message given to `tonic::Status` is visible in it entirety in the client

influxdb_iox/tests/end_to_end_cases/management_api.rs
tustvold: They already are :tada: This was what https://github.com/influxdata/influxdb_iox/pull/3260 did

influxdb_iox_client/src/client/health.rs
tustvold: As mentioned above, this is a third-party proto spec that doesn't carry this context. The GetServerStatus endpoint provides more context and is the IOx specific endpoint

influxdb_iox/tests/end_to_end_cases/management_api.rs
alamb: 🤔  I don't see an enumeration for `database_uuid`

tustvold: https://github.com/influxdata/influxdb_iox/pull/3260/files#diff-96057df473448d082ec5896e8c1feceb04ba6262b02e94922c9ca62fd358831cR248 ?

alamb:
Gah -- I swear I checked for that. Sorry I missed it

tustvold:
https://github.com/influxdata/conductor/pull/803 is updated to this commit and is passing CI, so going to go ahead and get this in before the inevitable merge conflicts arise

alamb:
> influxdata/conductor#803 is updated to this commit and is passing CI, so going to go ahead and get this in before the inevitable merge conflicts arise



Nice work

PR #3280, opened 2021-12-02
refactor: more trace logging in RUB

e-dard:
PR adds some tracing log lines to help me with timings and understand where there are points for query optimisation in the RUB.

Log lines are added as `trace` level, meaning they're compiled out of release builds.


read_buffer/src/chunk.rs
```diff
@@ -185,9 +185,19 @@ impl Chunk {
...
-        self.table
+        debug!(%predicate, ?select_columns, ?negated_predicates, "read_filter called");
+        let now = std::time::Instant::now();
+        let result = self
+            .table
             .read_filter(&select_columns, &predicate, negated_predicates.as_slice())
-            .context(TableError)
+            .context(TableError);
+
+        let row_groups = match &result {
```
alamb, line 195: I wonder if you could write this like

```rust
let row_groups = result.as_ref().and_then(|result| result.row_groups()).unwrap_or(0);
```

Though now that I write that I am not sure it is any more concise or readable 🤔

read_buffer/src/row_group.rs
```diff
@@ -400,24 +417,37 @@ impl RowGroup {
...
+            let (col_name, col) = self.column_name_and_column(expr.column());
 
             // Explanation of how this buffer pattern works. The idea is that
             // the buffer should be returned to the caller so it can be re-used
             // on other columns. Each call to `row_ids_filter` returns the
             // buffer back enabling it to be re-used.
-            match col.row_ids_filter(&expr.op, &expr.literal_as_value(), dst) {
+            let now = std::time::Instant::now();
+            let row_ids = col.row_ids_filter(&expr.op, &expr.literal_as_value(), dst);
+            trace!(elapsed=?now.elapsed(), rows=?match &row_ids{
```
alamb, line 428: maybe this is a candidate to make a function like `RowIDsOption::len()` to avoid repetition

alamb (approved):

read_buffer/src/chunk.rs
e-dard: Ended up at:

```rust
let row_groups = result
            .as_ref()
            .map(|result| result.row_groups())
            .unwrap_or(0);
```

Good shout.

alamb: 👨‍🍳 💋

read_buffer/src/row_group.rs
e-dard: Reason I didn't was because I want the context of the total number of rows in the column for one of the variants.

PR #3283, opened 2021-12-02
chore: Update Rust and fix or allow new Rust+clippy warnings

carols10cents:
The biggest change is probably that Rust is now able to detect that fields are unread. I marked all those with `#[allow(dead_code)]` because I'm not sure if they're valid or if they're signs of work to come or if they're old code that can be cleaned up.


crepererum:
I think some of these dead code warnings might actually be valid, but I think we can easily iterate over these `allow` statements later.

crepererum (approved):

alamb:
Yay -- thank you @carols10cents

PR #3284, opened 2021-12-02
feat: Add force flag to ClaimDatabase

alamb:

# Rationale

I want to copy a database from our production cluster and start it in my local development environment. IOx is complaining the database is owned by another server. While this is technically true, I want to use it anyways lol.

Here is what I am doing:
Run
```shell
Run IOx like this:
./target/release/influxdb_iox run -v --object-store=file --data-dir=$HOME/.influxdb_iox --server-id=42 --wipe-catalog-on-error=false
# Copy files for an entire database I downloaded from S3
cp -R ~/Downloads/3f4ff3f8-a448-431d-bf3d-0de350f08b2b ~/.influxdb_iox/dbs/
```
Then I try to claim this database:
```shell
./target/debug/influxdb_iox database claim 3f4ff3f8-a448-431d-bf3d-0de350f08b2b
Error claiming database: Unexpected server error: Client specified an invalid argument: The database with UUID `3f4ff3f8-a448-431d-bf3d-0de350f08b2b` is already owned by the server with ID 263145
```

(this is due to the fact I yoinked if from production where another server did indeed own it)

Note I can't even delete the owner file as a workaround 😭 :

```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database claim 3f4ff3f8-a448-431d-bf3d-0de350f08b2b
Error claiming database: Unexpected server error: Client specified an invalid argument: error getting database owner info: error loading owner info: Object at location dbs/3f4ff3f8-a448-431d-bf3d-0de350f08b2b/owner.pb not found: No such file or directory (os error 2)
```

# Changes:
1. Add a `force` flag to the management API, and CLI that allows clamining a database even if it is supposedly owned by another server or the owner.pb is not found
2. Add tests for the feature (much larger than the code change)

Note it still doesn't work perfectly for my database downloaded database downloaded (that will be the next PR):

```
cd /Users/alamb/Software/influxdb_iox && RUST_BACKTRACE=1 CARGO_TARGET_DIR=/Users/alamb/Software/iox-target cargo run -- database claim 3f4ff3f8-a448-431d-bf3d-0de350f08b2b  --force
    Finished dev [unoptimized + debuginfo] target(s) in 0.47s
     Running `/Users/alamb/Software/iox-target/debug/influxdb_iox database claim 3f4ff3f8-a448-431d-bf3d-0de350f08b2b --force`
Error claiming database: Unexpected server error: The operation was cancelled: Timeout expired
```

# CLI Usage Info

```
USAGE:
    influxdb_iox database claim [FLAGS] [OPTIONS] <uuid>

FLAGS:
        --force
            Force this server to claim this database, even if it is ostensibly owned by another server.

            WARNING: ONLY do this if you are sure that no other servers are writing to this database (for example, the
            data files have been copied somewhere). If another server is currently writing to this database, corruption
            will very likely occur
        --help
            Prints help information

    -V, --version
            Prints version information

```


server/src/database.rs
```diff
@@ -252,31 +252,63 @@ impl Database {
...
             .await
             .context(IoxObjectStoreError)?;
 
         let owner_info = fetch_owner_info(&iox_object_store)
             .await
-            .context(FetchingOwnerInfo)?;
+            .context(FetchingOwnerInfo);
+
+        // try to recreate owner_info if force is specified
+        let owner_info = match owner_info {
```
alamb, line 279: here is the actual logic change -- the rest of this PR is either plumbing the argument through or tests

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -669,6 +669,208 @@ async fn claim_database() {
...
+        .assert()
+        .success()
+        .stdout(predicate::str::contains(format!(
+            "Claimed database {}",
+            db_name
+        )));
+}
+
+#[tokio::test]
+async fn migrate_database_from_one_server_to_another() {
```
carols10cents, line 755: Hm, this test name is giving me a bit of pause... I expected it to be testing the normal "release on one server, claim on another" migration path that *doesn't* involve moving files in object storage, but this is testing your use case where you *do* want to move files in object storage for debugging purposes. Maybe...

```suggestion
async fn migrate_database_object_store() {
```

or 

```suggestion
async fn migrate_database_files_from_one_server_to_another() {
```

or 

```suggestion
async fn migrate_database_files_and_force_claim() {
```

?

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -669,6 +669,208 @@ async fn claim_database() {
...
+            .stdout(predicate::str::contains("Created"))
+            .get_output()
+            .stdout
+            .clone(),
+    )
+    .unwrap();
+
+    let db_uuid = stdout.lines().last().unwrap().trim();
+
+    // figure out where the database lives and copy its data to a temporary directory.
```
carols10cents, line 803: Along the same lines, I think in the future it would help to have documented *why* you might do this and that it's not really a production use case, something like:

```suggestion
    // figure out where the database lives and copy its data to a temporary directory,
    // as you might copy data from remote object storage to local disk for debugging.
```

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -669,6 +669,208 @@ async fn claim_database() {
...
+
+    // however with --force the owner.pb file is updated forcibly
+    Command::cargo_bin("influxdb_iox")
+        .unwrap()
+        .arg("database")
+        .arg("claim")
+        .arg(db_uuid)
+        .arg("--host")
+        .arg(addr)
+        .arg("--force") // sudo make me a sandwich
```
carols10cents, line 865: 😂

carols10cents (approved):
Looking pretty close!! Just a few tiny suggestions.

Also, do you think it's worth testing that the ownership transaction history in the owner.pb file is updated as expected in both the "deleted owner.pb" case and the "existing owner.pb" case? Similarly to these tests: https://github.com/influxdata/influxdb_iox/blob/2c04215c1fa082bbe0c812f764931fc3a32cf18a/server/src/database.rs#L1571-L1633

influxdb_iox/tests/end_to_end_cases/management_cli.rs
alamb: 👍  I like `migrate_database_files_from_one_server_to_another` (this is meant to be an end to end test of the workflow I am actually doing -- namely copying an entire database from object store somewhere locally to my machine (aka a different object store) and reimporting it

influxdb_iox/tests/end_to_end_cases/management_cli.rs
alamb: https://xkcd.com/149/ is one of my favorite comics of all time.

alamb:
Thanks for the review  @carols10cents 

> Also, do you think it's worth testing that the ownership transaction history in the owner.pb file is updated as expected in both the "deleted owner.pb" case and the "existing owner.pb" case? Similarly to these tests:

I thought about this and decided not to do it because:
1. The codepath isn't different (so I don't think there is code left to be covered)
2. I don't have a usecase for doing anything specific with the transaction history of a database that was forcibly claimed so I any test scenario would have to be artificial

I am going to merge this one in and then move on to my next `--force` command -- rebuilding catalogs!

PR #3289, opened 2021-12-03
fix: Capture the right 'no data' while parquet has no data

NGA-TRAN:
Another piece of #3089 

When compaction result has no data, its stream still includes some bytes. We need to check number of row groups for the actual no-chunk-data


parquet_file/src/metadata.rs
```diff
@@ -923,52 +927,14 @@ mod tests {
         // setup: preserve chunk to object store
         let mut generator = ChunkGenerator::new().await;
         generator.set_config(GeneratorConfig::NoData);
-        let (chunk, _) = generator.generate().await;
```
NGA-TRAN, line 926: Since we no longer need to generate/create parquet file without row groups, we only need test no parquet file and stop. There are no other details of the file to test. Thus, I only keep one test to verify no parquet file created and remove the other necessary tests.

parquet_file/src/storage.rs
```diff
@@ -470,10 +473,10 @@ mod tests {
         };
 
         // create parquet file
-        let (_record_batches, schema, _column_summaries, _num_rows) =
-            make_record_batch("foo", TestSize::Full);
+        let (record_batches, schema, _column_summaries, _num_rows) =
+            make_record_batch("foo", TestSize::Minimal);
```
NGA-TRAN, line 477: Need data for the test below to work. otherwise the below line will return None.

parquet_file/src/test_utils/generator.rs
```diff
@@ -66,13 +66,13 @@ impl ChunkGenerator {
         &self.partition
     }
 
-    pub async fn generate(&mut self) -> (ParquetChunk, IoxMetadata) {
+    pub async fn generate(&mut self) -> Option<(ParquetChunk, IoxMetadata)> {
```
NGA-TRAN, line 69: Will generate nothing/None if no data provided

NGA-TRAN (commented):
Since I changed the behavior to not generate parquet file without row group, I found myself  modifying many related tests. 
@tustvold and @crepererum : you may want to verify if I delete important tests

parquet_file/src/storage.rs
```diff
@@ -155,16 +155,19 @@ impl Storage {
         let schema = stream.schema();
         let data = Self::parquet_stream_to_bytes(stream, schema, metadata).await?;
```
alamb, line 156: Having to ensure all the other codepaths are prepared to handle parquet files with no row groups seems like it may be more complicated than preventing such parquet files from being created in the first place

Rather than checking if there are now rows by checking the returned `Vec` 's size, I wonder if it would be possible to push the check into `parquet_stream_to_bytes`. Since that function appears to see all the `RecordBatches` maybe it could track how many rows were actually written and return `None` if nothing actually was?

https://github.com/influxdata/influxdb_iox/blob/11067bfe3f9151b586856d6f4990be4fc3ce595b/parquet_file/src/storage.rs#L130-L151

parquet_file/src/storage.rs
```diff
@@ -149,10 +149,15 @@ impl Storage {
         {
             let mut writer = ArrowWriter::try_new(mem_writer.clone(), schema, Some(props))
                 .context(OpeningParquetWriter)?;
+            let mut no_stream_data = true;
             while let Some(batch) = stream.next().await {
+                no_stream_data = false;
```
alamb, line 156: I wonder if we have to check `batch.num_rows() > 0` -- I don't know if we ever produce zero length record batch rows 🤔

parquet_file/src/storage.rs
```diff
@@ -149,10 +149,15 @@ impl Storage {
...
                 let batch = batch.context(ReadingStream)?;
                 writer.write(&batch).context(WritingParquetToMemory)?;
             }
+            if no_stream_data {
```
alamb, line 160: 👍

alamb (approved):
looks good to me

parquet_file/src/storage.rs
NGA-TRAN: I think it is safe with this check that no stream at all. We do not need to check num_rows

alamb: I was thinking of the case where a delete predicate filtered out all rows , for example, in which case I think the result would be a single `RecordBatch` that had zero rows

NGA-TRAN: This is a good point that has lead to a complicated locking mechanism we have implemented :)
With the current approach, the deleted/duplicated rows are eliminated before this step which means whatever comes out of this stream will be put into a new chunk. If a delete happens concurrently with persistent step of the compaction (aka right at this moment of the code), its delete predicate will be kept temporarily in a mailbox and will be attached to this chunk's in-memory catalog  after the persistence is done.

Issue #3282, opened 2021-12-02
Parquet Scan Performance


tustvold:
**Problem**

Currently the scan path for a parquet file in object storage is as follows:

* DbChunk::snapshot - obtain snapshot of ParquetChunk by cloning Arc
* DbChunk::read_filter
* ParquetChunk::read_filter
* Storage::read_filter
* Storage::download_and_scan_parquet

This final step downloads the file from object storage to a temporary file, and then runs Datafusion's ParquetExec on the file returning the yielded SendableRecordBatchStream.

This will download the same parquet file repeatedly, even if the object store implementation is disk-based. This is clearly unfortunate.

**Proposal**

Introduce a new trait `ParquetReader` with a single method (or more depending on how implementation shakes out)

```
trait ParquetReader {
    fn read_filter(
        predicate: &Predicate,
        selection: Selection<'_>,
        schema: SchemaRef,
        path: ParquetFilePath,
    ) -> Result<SendableRecordBatchStream>
}
```

Each `ParquetChunk` will contain an `Arc<dyn ParquetReader>` instead of `Arc<IoxObjectStore>` that will be used by `ParquetChunk::read_filter`.

Multiple implementations of `ParquetChunk` will then be written:

* ParquetReaderStore - streams data from object storage
* ParquetReaderFileStore - streams data from a file-system based object store
* ParquetReaderCached - streams data from a local parquet cache

**Alternatives Considered**

There is some existing plumbing for an `ObjectStoreFileCache` within the `object_store` crate. Unfortunately this is limited by the fact it has to still expose the ObjectStore API, limiting the ability to push down projections, predicates etc..., avoid additional memory copies, stream data, etc... This proposal instead abstracts away the object store, allowing great implementation flexibility

Another question is whether the cache should support writes, I decided against this for two reasons:

* It requires a non-trivial amount of additional functionality from implementations
* Longer-term the servers performing writes may not be the same ones servicing queries
* Query should drive caching not ingest

e-dard:
This proposal seems like a good one to me 👍🏻

crepererum:
Why is the `ParquetReader` dynamic per chunk instead of per object store?

tustvold:
> Why is the ParquetReader dynamic per chunk instead of per object store?

[...]

PR #3291, opened 2021-12-03
feat: simplify parquet reader (#3282)

tustvold:
This PR migrates away from using ParquetExec to read files, in preparation for #3282.

Aside from drastically simplifying the code, this also fixes an issue where we were performing blocking IO, namely writing to a temporary file, in a regular tokio task. It also avoids having to use named temporary files, which avoids them accidentally getting orphaned in `/tmp` (see [here](https://docs.rs/tempfile/3.2.0/tempfile/#resource-leaking)).

Some very unscientific local testing showed no performance differential from this change (which isn't surprising)


parquet_file/src/storage.rs
```diff
@@ -246,106 +192,44 @@ impl Storage {
...
+                break;
             }
         }
+
         Ok(())
     }
 
     pub fn read_filter(
-        predicate: &Predicate,
+        _predicate: &Predicate,
```
tustvold, line 227: This is temporary, but doesn't represent a regression as they weren't enabled before (#2163) and I don't think we are persisting parquet files with multiple row groups anyway. It should be relatively straightforward to re-enable in future should we decide to do so

parquet_file/src/storage.rs
```diff
@@ -246,106 +192,46 @@ impl Storage {
...
-            .context(OpenTempFile)?;
-
-        debug!(?path, ?temp_file, "Beginning to read parquet to temp file");
-        let mut read_stream = store
-            .get_parquet_file(&path)
-            .await
+        let mut file = tempfile::tempfile().context(OpenTempFile)?;
+
+        debug!(?path, ?file, "Beginning to read parquet to temp file");
+        let read_stream = futures::executor::block_on(store.get_parquet_file(&path))
```
tustvold, line 199: This is doubly stupid in the case of a local filesystem, as not only are we copying the files to a temporary location from a location already on disk, but all the tokio filesystem IO is blocking and so we're needlessly proxying it via a threadpool only to use it in a blocking manner :laughing:

parquet_file/src/storage.rs
alamb: See also https://github.com/influxdata/influxdb_iox/pull/2839 which did reenable it (but I haven't yet merged due to lack of test)

I am pretty sure I have seen iox make parquet files with multiple row groups, but I can't seem to find any at this moment. 

In any event, I think now that we are getting to the point where we can actually run queries on parquet files, we can figure out how, if at all, we want to re-enable pruning

alamb (approved):
This PR got me thinking that `ParquetExec` is the code that does Row Group Pruning, so that is something we would have to add to IOx in the future; If we could find some way to reuse the code in DataFusion (and the tests), that would be ideal. But we can deal with that as needed

tustvold:
> If we could find some way to reuse the code in DataFusion (and the tests)

I _think_ this should be possible, after all we are already using `PruningPredicate` to prune chunks in IOx, and most of the mechanics of implementing row group pruning are actually implemented in `parquet` but I suppose we shall see.

PR #3292, opened 2021-12-03
feat: Add force flag to RebuildCatalog

alamb:
# Rationale

The usability of copying parquet files from one server to another (local) is not great.

There are at least two things that make it so
1. Users must currently must "break the catalog" as described in [the docs](https://github.com/influxdata/influxdb_iox/blob/main/docs/server_startup.md#break-the-catalog ) in order to get a database that has started and ended up in an error state
2. The server must be started with the ` --wipe-catalog-on-error=false` flag (or else the db will just clear its catalog when there is an error rather than staying in an error state)

What I want is a way to force an IOx database to effectively "rescan" all the parquet files in its data directory and rebuild the catalog from those parquet files even if it already has a perfectly good data file

Prior to this PR, for example, if I wanted to rebuild the `foo` database with the following command, it doesn't work if foo has successfully started

```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox database recover rebuild foo

Error in partition subcommand: Error rebuilding preserved catalog: The system is not in a state required for the operation's execution: database (foo) in invalid state (Initialized) for transition (RebuildPreservedCatalog)
```

# Changes:
1. Add a `--force` flag to `database recover rebuild` that will force IOx to rebuild the catalog, regardless of the state of the database when the rebuild command is issued
2. Update docs
3. tests


generated_types/src/job.rs
```diff
@@ -1,6 +1,29 @@
 use crate::influxdata::iox::management::v1 as management;
 use data_types::job::Job;
 
+impl management::operation_metadata::Job {
+    /// Return the db_name for this job
+    pub fn db_name(&self) -> &str {
```
alamb, line 6: used in tests

influxdb_iox/tests/end_to_end_cases/database_migration.rs
```diff
@@ -128,4 +163,101 @@ async fn migrate_database_files_from_one_server_to_another() {
             "Claimed database {}",
             db_name
         )));
+
+    //
```
alamb, line 167:
```suggestion
```

influxdb_iox/tests/end_to_end_cases/database_migration.rs
```diff
@@ -128,4 +163,101 @@ async fn migrate_database_files_from_one_server_to_another() {
...
+
+    // Now start another server and create a database to receive the files
+    let fixture = ServerFixture::create_shared(ServerType::Database).await;
+    let addr = fixture.grpc_base();
+
+    let db_name = rand_name();
+    let new_db_uuid = create_readable_database(&db_name, fixture.grpc_channel()).await;
+    let sql_query = "select count(*) from the_table";
+
+    // No data for cpu_load_short yet
```
alamb, line 182:
```suggestion
    // No data for the_table yet
```

influxdb_iox/tests/end_to_end_cases/database_migration.rs
```diff
@@ -128,4 +163,101 @@ async fn migrate_database_files_from_one_server_to_another() {
...
+
+    println!(
+        "Find tmp_path: \n\n{}\n\n",
+        Command::new("find")
+            .arg(tmp_dir.path().to_string_lossy().to_string())
+            .assert()
+            .success()
+            .to_string()
+    );
+
```
alamb, line 205:
```suggestion
```

server/src/database.rs
```diff
@@ -547,33 +569,87 @@ impl Database {
...
+    /// Rebuilding the catalog from parquet files. This can be used to
+    /// recover from a CatalogLoadError, or if new parquet files are
+    /// added to the data directory
+    pub async fn rebuild_preserved_catalog(&self, force: bool) -> Result<TaskTracker<Job>, Error> {
         let handle = self.shared.state.read().freeze();
         let handle = handle.await;
 
         let db_name = &self.shared.config.name;
-        let current_state = error_state!(self, "RebuildPreservedCatalog", CatalogLoadError);
+
```
alamb, line 580: Here is the core logic change. It would be great if @tustvold  could review to ensure it doesn't violate any of the startup statemachine invariants

server/src/database.rs
```diff
@@ -547,33 +569,87 @@ impl Database {
...
+            expected: "Object store initialized",
+        })?;
 
         tokio::spawn(
             async move {
                 let db_name = &shared.config.name;
 
-                // First attempt to wipe the catalog
-                PreservedCatalog::wipe(&current_state.iox_object_store)
+                // shutdown / stop the DB if it is running so it can't
```
alamb, line 611: Here is the core logic change. It would be great if @tustvold  could review to ensure it doesn't violate any of the startup statemachine invariants

server/src/database.rs
```diff
@@ -582,10 +658,12 @@ impl Database {
...
 
-                {
-                    let mut state = shared.state.write();
-                    *state.unfreeze(handle) = DatabaseState::RulesLoaded(current_state);
-                }
+                // Double check the state hasn't changed (we hold the
+                // freeze handle to make sure it does not)
+                assert!(matches!(&**shared.state.read(), DatabaseState::Known(_)));
+
+                // TODO: is this necessary?
```
alamb, line 665:
```suggestion
```

NGA-TRAN (approved):
I approve this in case you want to merge for a quick test. I do not see any red flags but since I am not familiar with this code, my review is suboptimal

alamb:
> I approve this in case you want to merge for a quick test. I do not see any red flags but since I am not familiar with this code, my review is suboptimal

Thanks @NGA-TRAN  

I think it can wait another day or so to merge to give @crepererum  and / or @tustvold  a chance if they want to review too. I just find myself wanting to use this flag for testing on another PR and I can't :)

crepererum (approved):

server/src/database.rs
```diff
@@ -547,33 +569,87 @@ impl Database {
...
         let handle = handle.await;
 
         let db_name = &self.shared.config.name;
-        let current_state = error_state!(self, "RebuildPreservedCatalog", CatalogLoadError);
+
+        {
+            // If the force flag is not specified, can only rebuild
+            // the catalog if it is in ended up in an error loading
+            let state = self.shared.state.read();
+            if !force && !matches!(&**state, DatabaseState::CatalogLoadError { .. }) {
```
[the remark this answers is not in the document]
alamb: Ah, that is a good point 👍  I will add that warning

server/src/database.rs
alamb: Will give it a try

server/src/database.rs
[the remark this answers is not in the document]
alamb: Added in 01ae5325

alamb:
> I already messed with the semi-linear state machine in #3318 😅. I think it's OK, but at one point we should probably draw a nice diagram of the FSM.

It isn't exactly a diagram but maybe an improvement: https://github.com/influxdata/influxdb_iox/pull/3349

PR #3293, opened 2021-12-03
 feat: don't copy parquet files out of disk object store (#3282) 

tustvold:
Draft as builds on #3291 and a little bit rough and ready still

This changes the ObjectStore to return `GetResult` instead of a `BoxedStream`. This allows for two things:

* Simplifies the API for clients that just want the Bytes
* Allows the disk implementation to return the raw File object

This in turn allows the ParquetFileReader to not need to create a temporary file. In local testing this had no performance impact, but this may make a difference for some workloads - especially where temporary files are not `tmpfs` (i.e. RAM-backed).

I'm not sure I especially like this, in #3282 I proposed doing something slightly different, but this was certainly less painful to implement :laughing:


object_store/src/lib.rs
```diff
@@ -582,6 +576,65 @@ impl<P: ObjectStorePath> ObjectMeta<P> {
     }
 }
 
+/// Result for a get request
+pub enum GetResult<E> {
```
alamb, line 580: 👍  this is a clever solution.

object_store/src/memory.rs
```diff
@@ -188,8 +189,7 @@ mod tests {
             .get(&location)
             .await
             .unwrap()
-            .map_ok(|b| bytes::BytesMut::from(&b[..]))
-            .try_concat()
+            .bytes()
```
alamb, line 192: Implementing `bytes()` and `stream()`  is much nicer, in my opinion, than dealing with `Box`ed streams directly 👍

alamb (approved):
This looks fine to me -- but to be clear this is only going to have some effect for environments using the file object store, right?

Or do you envision also implementing caching at the object store level as well (aka the S3 object store implementation would download data to files and pass back handles to those files)?

tustvold:
> This looks fine to me -- but to be clear this is only going to have some effect for environments using the file object store, right?

Correct, this is just to get a sense of the fastest the cache could possibly be and also to assist with local development

> do you envision also implementing caching at the object store level as well

No, I think it makes sense to treat parquet files as special

PR #3295, opened 2021-12-03
refactor: remove create database from data generator

pauldix:
This removes the create database command line flag and associated code from the data generator runner. Creation of databases should live outside the generator in other tools.


alamb (approved):
Makes sense to me

PR #3296, opened 2021-12-03
feat: grpc call for compact object store chunks

### LINKED ISSUE #3124: gRPC API to call compact object store chunks
__Use case:__
Now we have the function #3089 (PR #3116), let us implement the gRPC call to invoke that

__Proposal:__
Build a simple gRPC call for users to manually compact object store chunks through a newly built function in #3089. User will provide `db_name`, `table_name`, `partition_key` and a set of `chunk_id` they want to compact. Those chunks have to be already persisted and contiguous in that partition to get compacted

NGA-TRAN:
Closes #3124
Another piece of #3089 


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -90,6 +90,11 @@ service ManagementService {
 
   // Drop partition from memory and (if persisted) from object store.
   rpc DropPartition(DropPartitionRequest) returns (DropPartitionResponse);
+
+  // Compact given object store chunks 
+  //
+  // Errors if the chunks are not compacted yet and not contiguous
+  rpc CompactObjectStoreChunks(CompactObjectStoreChunksRequest) returns (CompactObjectStoreChunksResponse);
```
NGA-TRAN, line 97: Now I have this ready, after the fix in https://github.com/influxdata/influxdb_iox/pull/3289 is merged, I will add test for this client as well as end-to-end tests in next PR.

generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -472,3 +477,22 @@ message DropPartitionRequest {
...
+
+  // the table name
+  string table_name = 3;
+
+  // the chunk ids
+  repeated bytes chunk_ids = 4;
+}
+
+message CompactObjectStoreChunksResponse {
+}
```
alamb, line 497: I wonder if the request should return a JobId (so that the compaction can happen as an async task rather than a synchronous one). I can imagine some compaction operations might take longer than the 30 second gRPC timeout

Something like this (following the model of `RebuildCatalog`):

```suggestion
  // The operation that tracks the work for compacting the chunks
  google.longrunning.Operation operation = 1;
}
```

https://github.com/influxdata/influxdb_iox/blob/main/generated_types/protos/influxdata/iox/management/v1/service.proto#L429-L432

influxdb_iox/src/commands/database/partition.rs
```diff
@@ -67,6 +67,25 @@ struct Persist {
     force: bool,
 }
 
+/// Compact Object Store Chunks
+///
+/// Errors if the chunks are not yet compacted and not contiguous. If successful it returns the
+/// compacted chunk or None if no data after compacted due to rows eliminated from deletes and deduplication
```
alamb, line 73: i don't think this comment is 100% accurate anymore -- the command doesn't seem to return the compacted chunk (it seems to print `"Ok"` on success).

influxdb_iox/src/commands/database/partition.rs
```diff
@@ -152,6 +171,12 @@ enum Command {
     /// chunk that contains the persisted data.
     Persist(Persist),
 
+    /// Compact Object Store Chunks
+    ///
+    /// Errors if the chunks are not yet compacted and not contiguous. If successful it returns the
+    /// compacted chunk or None if no data after compacted due to rows eliminated from deletes and deduplication
```
alamb, line 177: likewise here the reference to returning the compacted chunk I think is erronious

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/management.rs
```diff
@@ -534,6 +534,38 @@ impl management_service_server::ManagementService for ManagementService {
...
+            .db(&db_name)
+            .map_err(default_server_error_handler)?;
+
+        let mut chunk_id_ids = vec![];
+        for chunk_id in chunk_ids {
+            let chunk_id = ChunkId::try_from(chunk_id).scope("chunk_id")?;
+            chunk_id_ids.push(chunk_id);
+        }
+
+        db.compact_object_store_chunks(&table_name, &partition_key, chunk_id_ids)
```
alamb, line 563: As I mentioned above, since I think this operation can take quite a while, doing it as an asyncrhonous Job and returning a job ID might make more sense here rather than waiting for the job to finish.

influxdb_iox_client/src/client/management.rs
```diff
@@ -503,4 +503,34 @@ impl Client {
...
+        &mut self,
+        db_name: impl Into<String> + Send,
+        table_name: impl Into<String> + Send,
+        partition_key: impl Into<String> + Send,
+        chunk_ids: Vec<Bytes>,
+    ) -> Result<(), Error> {
+        let db_name = db_name.into();
+        let partition_key = partition_key.into();
+        let table_name = table_name.into();
+        // let chunk_ids = chunk_ids
```
alamb, line 520: Left over?

alamb (approved):
I do think making `compact_chunks` an async operation (by returning a JobId) is / will be important. However, I also think it could be done as a follow on PR as well.

influxdb_iox/src/commands/database/partition.rs
NGA-TRAN: Good catch

generated_types/protos/influxdata/iox/management/v1/service.proto
NGA-TRAN: Great suggestion and link. Working on this

NGA-TRAN:
@alamb O have addressed all your comments. Can you have a quick look to see if they work the way you suggested? I  adding tests to verify the job returned messages before merging

generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -472,3 +477,24 @@ message DropPartitionRequest {
...
+  // the table name
+  string table_name = 3;
+
+  // the chunk ids
+  repeated bytes chunk_ids = 4;
+}
+
+message CompactObjectStoreChunksResponse {
+  // The operation that tracks the work for compacting object store chunks
+  google.longrunning.Operation operation = 1;
```
alamb, line 498: 👍

alamb (approved):
Looks good. Thanks @NGA-TRAN

PR #3297, opened 2021-12-04
refactor: Add agent name to data generator

pauldix:
This is work leading up to giving the data generator the ability to write to many databases. The plan is to specify which agents databases will use to write data.


alamb (approved):
Looks good to me -- there appear to be a few other places that need to be updated to include the new `name` field that are causing the CI tests to fail

pauldix:
Ah, CI saves me again. I was only running the package tests locally 😄

PR #3313, opened 2021-12-06
chore: provide a single way to build prod IOx images

### LINKED ISSUE #3310: Provide a single way to build IOx images
We currently have two ways to build IOx images:

1. Via CircleCI and "copy binary to docker image" (see [here](https://github.com/influxdata/influxdb_iox/blob/1099f5c4173dc312491f23448e3bbd9ecb9921dd/.circleci/config.yml#L325-L388))
2. Via an ordinary `Dockerfile`

This is mostly a historical artifact and has the following issues:

- it's easy to get these two things out of sync (e.g. variant two already uses different features and doesn't provide a proper entrypoint to multiplex between server modes).
- Variant 1 doesn't work for Mac users or for Linux users w/ a reasonable new distro (because glibc isn't compatible)

So I propose to just use variant 2 and port over all functionality that we need from variant 1.

crepererum:
Closes #3310.


tustvold (approved):

crepererum:
I've manually triggered the `perf_image` workflow:

- CircleCI job: <https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/13569/workflows/3f76a6a2-7795-4c71-a864-970ff36fa1d2/jobs/100835>
- Quay artifact: `quay.io/influxdb/iox:c77637e5` / `quay.io/influxdb/iox:crepererum_issue3310`

Issue #3311, opened 2021-12-06
`migrate_database_files_from_one_server_to_another` is flaky


e-dard:
From our M1 builder:

```
failures:

---- end_to_end_cases::database_migration::migrate_database_files_from_one_server_to_another stdout ----
****************
Server 127.0.0.1:8091 Logging to "/var/folders/yb/j9nqj1kj3gl_yltwrsqk9nd00000gn/T/.tmp2nsSwy"
****************
Waiting for HTTP server to be up: error sending request for url (http://127.0.0.1:8091/health): error trying to connect: tcp connect error: Connection refused (os error 61)
Waiting for gRPC API to be up: Connection error: transport error (error trying to connect: tcp connect error: Connection refused (os error 61))
Waiting for HTTP server to be up: error sending request for url (http://127.0.0.1:8091/health): error trying to connect: tcp connect error: Connection refused (os error 61)
Waiting for gRPC API to be up: Connection error: transport error (error trying to connect: tcp connect error: Connection refused (os error 61))
Successfully connected to server
Successfully got a response from HTTP: Response { url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(8091), path: "/health", query: None, fragment: None }, status: 200, headers: {"content-length": "2", "date": "Mon, 06 Dec 2021 13:42:53 GMT"} }
Deployment service is running
Successfully started TestServer (grpc http://127.0.0.1:8093, http http://127.0.0.1:8091)
Copying data from "/var/folders/yb/j9nqj1kj3gl_yltwrsqk9nd00000gn/T/influxdb_ioxAB8PfF/dbs/55929411-92a0-4efd-bde7-c204c136547f" to "/var/folders/yb/j9nqj1kj3gl_yltwrsqk9nd00000gn/T/.tmppJWe9p"
****************
Start TestServer Output
****************
[2m2021-12-06T13:42:52.943471Z[0m [32m INFO[0m [2minfluxdb_iox::structopt_blocks::object_store[0m[2m:[0m Using File for object storage
[2m2021-12-06T13:42:52.946670Z[0m [32m INFO[0m [2mserver::application[0m[2m:[0m using specified number of threads per thread pool [3mnum_threads[0m[2m=[0m8
[...]

PR #3314, opened 2021-12-06
fix: fix flay migrate_database_files_from_one_server_to_another and prepare for reuse + fix

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3311

This fixes a flaky test and prepares much of it for reuse in https://github.com/influxdata/influxdb_iox/pull/3292

I will comment inline about the actual fix


influxdb_iox/tests/end_to_end_cases/database_migration.rs
```diff
@@ -1,18 +1,47 @@
...
+    let target_dir = target_dir.as_ref();
+
+    // if the target database has had no data written yet, it will
+    // have no `data` directory
+    println!("Ensuring {:?} directory exists", target_dir);
+    Command::new("mkdir")
+        .arg("-p")
+        .arg(target_dir.to_string_lossy().to_string())
+        .assert()
+        .success();
```
alamb, line 28: This is the fix for https://github.com/influxdata/influxdb_iox/pull/3292 (if the target directory doesn't exist the files end up in the wrong spot)

influxdb_iox/tests/end_to_end_cases/database_migration.rs
```diff
@@ -1,18 +1,47 @@
...
-async fn migrate_database_files_from_one_server_to_another() {
+/// Copy the `source_dir` directory into the `target_dir` directory using the `cp` command
+fn cp_dir(source_dir: impl AsRef<Path>, target_dir: impl AsRef<Path>) {
+    let source_dir = source_dir.as_ref();
+    let target_dir = target_dir.as_ref();
+
+    // needed so that if the target server has had no databases
+    // created yet, it will have no `data` directory. See #3292
+    println!("Ensuring {:?} directory exists", target_dir);
+    Command::new("mkdir")
```
alamb, line 24: This is the fix for https://github.com/influxdata/influxdb_iox/pull/3292 (if the target directory doesn't exist the files end up in the wrong spot)

The rest of this PR is just refactoring into reusable pieces I am using for a new test

e-dard (approved):

influxdb_iox/tests/end_to_end_cases/database_migration.rs
crepererum: Do we really need to shell out to `mkdir` and `cp` to create an copy a bit of FS content? Should the [stdlib](https://doc.rust-lang.org/std/fs/index.html#functions) be sufficient for that?

alamb: I couldn't find anything in the stdlib that will do `cp -R` (aka recursive directory copy) and I thought shelling out to `cp` was less icky than importing a new crate.

PR #3318, opened 2021-12-06
fix: retry write buffer creation during database init

### LINKED ISSUE #3287: Kafka topic auto-creation is racy
__Steps to reproduce:__

1. Create many routers and databases w/ Kafka-based write buffers in parallel
2. Look for `error creating write buffer: Cannot create non-empty topic` errors in gRPC responses and logs

__Expected behaviour:__
No errors.

__Actual behaviour:__
`error creating write buffer: Cannot create non-empty topic`

__Environment info:__
* test bench

__Background:__
It's very likely that this check-do-check construct doesn't work as expected, maybe because there's a small delay between "admin command executed" and "partitions visible":

https://github.com/influxdata/influxdb_iox/blob/f17b670fe917d204c2d2a05c77df811141f44481/write_buffer/src/kafka.rs#L593-L609

crepererum:
Closes #3287.


write_buffer/src/kafka.rs
```diff
@@ -526,13 +531,30 @@ where
...
+                Err("Topic exists but has no partitions".to_string().into())
+            } else {
+                Ok(Some(partitions))
+            }
+        }
+        Some(error_code) => {
+            let error_code: RDKafkaErrorCode = error_code.into();
+            match error_code {
+                RDKafkaErrorCode::UnknownTopic | RDKafkaErrorCode::UnknownTopicOrPartition => {
+                    Ok(None)
```
alamb, line 552: maybe worth a comment here explaining that the caller is responsible for creating the topic (the docstring implies it above so more comments aren't required, I just feel it would be clearer if they were present)

write_buffer/src/kafka.rs
```diff
@@ -590,8 +612,11 @@ where
     C: Consumer<D> + Send + Sync + 'static,
     D: ConsumerContext,
 {
-    let mut partitions = get_partitions(database_name, consumer).await?;
-    if partitions.is_empty() {
+    loop {
+        if let Some(partitions) = get_partitions(database_name, consumer).await? {
```
alamb, line 616: What exactly happens under load? Is it that there is some delay between when the call to `create_kafka_topic` and when a subsequent call to `get_partitions()` will reflect those newly created topics?

I ask because as written, it seems like this code will continually retry to recreate the topics, rather than create them once and then loop trying to get them.  I am not sure if this is a problem or not but wanted to point it out

alamb (approved):

server/src/database.rs
```diff
@@ -1806,6 +1811,69 @@ mod tests {
...
+            server_id,
+            wipe_catalog_on_error: false,
+            skip_replay: false,
+        };
+        let database = Arc::new(Database::new(Arc::clone(&application), db_config.clone()));
+
+        // wait for a bit so the database fails because the mock is missing
+        let database_captured = Arc::clone(&database);
+        tokio::time::timeout(Duration::from_millis(100), async move {
+            database_captured.wait_for_init().await.unwrap();
```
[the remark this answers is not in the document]
crepererum: improved

write_buffer/src/kafka.rs
crepererum: done

write_buffer/src/kafka.rs
crepererum: fixed

PR #3328, opened 2021-12-07
feat: add `system.queries `system table

alamb:
# Rationale

I want to get a snapshot of actual customer queries that were run against an IOx system so that with a snapshot of the data I can reproduce, profile, and optimize IOx's query performance

To do so I need to know what actual queries were sent to IOx, in a form I can replay them. 

# Changes:
1. Record the last 100 (picked arbitrarily) queries (both SQL and gRPC) queries that are run in memory
2. Expose that information via `system.queries` system table
3. Tests

# Design notes
I put the log on `Db` (as it is a db-wide query log), which needed a little code rearranging in the storage service. It seems a little less than ideal but I don't think it is terrible either and I can't come up with anything better


tustvold:
FYI - https://github.com/influxdata/influxdb_iox/pull/3334 should fix the pbjson

influxdb_iox/src/influxdb_ioxd/server_type/database/http.rs
```diff
@@ -252,6 +252,8 @@ async fn query(
 
     let db = server.db(&db_name)?;
 
+    db.record_query("sql", &q);
```
alamb, line 255: here is where SQL queries (that come from HTTP) are recorded

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/flight.rs
```diff
@@ -170,6 +171,8 @@ impl Flight for FlightService {
             .db(&database)
             .map_err(default_server_error_handler)?;
 
+        db.record_query("sql", &read_info.sql_query);
```
alamb, line 174: SQL queries from flight are recorded here

query_tests/cases/in/all_chunks_dropped.expected
```diff
@@ -9,6 +9,7 @@
 | public        | system             | chunk_columns       | BASE TABLE |
 | public        | system             | operations          | BASE TABLE |
 | public        | system             | persistence_windows | BASE TABLE |
```
alamb, line 11: this is a strange test -- I will update it to not be a replica of system tables -- will change it so it only selects iox tables (not system tables)

e-dard (approved):
I looks through this and it all makes sense to me!

arrow_util/src/test_util.rs
```diff
@@ -89,3 +92,39 @@ pub fn sort_record_batch(batch: RecordBatch) -> RecordBatch {
 
     RecordBatch::try_new(batch.schema(), sort_output).unwrap()
 }
+
+/// Apply a
```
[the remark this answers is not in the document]
alamb: hahah -- will fix that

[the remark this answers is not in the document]
alamb: In https://github.com/influxdata/influxdb_iox/pull/3328/commits/91139468eaf0a915a47867eb41f1a32ee4c5073a

PR #3330, opened 2021-12-08
feat: add ability for data generator to write to many buckets

pauldix:
This adds the ability for the data generator to write to many databases. A new command line argument, `bucket_list`, is added which should be a file name. The file should contain a list of databsaes, one per line, with the structure of `<org>_<bucket>`. This is a little odd given the data generator expects org and bucket separately, but I expect the file that we'll be using will be database names, which have this format.

The configuration can specify what percentage of the list should get written to by which agents at what sampling interval. This should allow configurations where databases get different levels of ingest and different types (as specified via different agent specs). The structure is a little wonky, but I think it'll get the job done. The next step is to run some perf tests to see how the data generator performs if writing to 10k databases.

### DIFF
--- a/iox_data_generator/schemas/full_example.toml
+++ b/iox_data_generator/schemas/full_example.toml
@@ -177,12 +177,12 @@ bool = true
 # amounts of write load as well as different schemas through specifying different agents.
 [[bucket_writers]]
 # 20% of the buckets specified in the --bucket_list file will have these agents writing to them
-percent = 0.2
+ratio = 0.2
 # for each of those buckets, have 3 of the first_agent writing every 10s, and 1 of the another_example writing every minute.
 agents = [{name = "first_agent", count = 3, sampling_interval = "10s"}, {name = "another_example", sampling_interval = "1m"}]
 
 [[bucket_writers]]
 # the remaining 80% of the buckeets specified will write using these agents
-percent = 0.8
+ratio = 0.8
 # we'll only have a single agent of another_example for each database
 agents = [{name = "another_example", sampling_interval = "1s"}]
--- a/iox_data_generator/schemas/full_example.toml
+++ b/iox_data_generator/schemas/full_example.toml
@@ -103,14 +103,10 @@ for_each = [
     "foo_bar"
 ]
 
+# Agent specs can be referenced later on by bucket writers, which specify how frequently
+# data should be written and by how many different agents.
 [[agents]]
 name = "first_agent"
-# Create this many agents. Agents are single threaded so the way to get more paralellism is to
-# increase the number of agents.
-count = 2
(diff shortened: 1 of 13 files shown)


iox_data_generator/schemas/full_example.toml
```diff
@@ -171,4 +167,22 @@ count = 2
...
+
+# bucket_writers specify how to split up the list of supplied buckets to write to. If
+# only a single one is specified via the CLI flags, then you'd want only a single bucket_writer
+# with a percent of 1.0.
+#
+# These make it possible to split up a large list of buckets to write to and send different
+# amounts of write load as well as different schemas through specifying different agents.
+[[bucket_writers]]
+# 20% of the buckets specified in the --bucket_list file will have these agents writing to them
+percent = 0.2
```
crepererum, line 180: I find the naming a bit confusing here. I would expect that a field called "percent" is a float withing the range `[0,100]`. Also see:

> In mathematics, a percentage (from Latin per centum "by a hundred") is a number or ratio expressed as a fraction of 100. ([Wikipedia](https://en.wikipedia.org/wiki/Percentage))

A float in `[0,1]` I would rather call "ratio".

iox_data_generator/src/specification.rs
```diff
@@ -110,6 +229,33 @@ pub struct TagSetsSpec {
     pub for_each: Vec<String>,
 }
 
+/// The specification for what should be written to the list of provided org buckets.
+/// Buckets will be written to by one or more agents with the given sampling interval and
+/// agent count.
+#[derive(Deserialize, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct BucketWriterSpec {
```
crepererum, line 237: Is there no way to assign a "bucker writer spec" only to a subset of buckets, e.g. via regex?

crepererum (changes requested):

crepererum:
> The next step is to run some perf tests to see how the data generator performs if writing to 10k databases.

I got 1k databases running, but I doubt 10k would work w/ our current Kafka implementation :D

iox_data_generator/schemas/full_example.toml
pauldix: Ah, I'll change it to that. I wanted the ability to have `0.001` for example if I know I'm feeding in 10k databases and I only want 10 of them to use those agents.

iox_data_generator/src/specification.rs
pauldix: I can add that as a feature in a follow on PR. This will split them up based on the `percent` (soon to be `ratio`) and it does them in order. So you could organize the input file so that the databases are in the order you want them to be assigned. Definitely not very user friendly though, so I'll add the regex feature too.

PR #3337, opened 2021-12-08
chore: clean up `all_chunks_dropped.sql`

alamb:
# Rationale:
This test failed when I added a new system table but it is unrelated to system tables. See https://github.com/influxdata/influxdb_iox/pull/3328#discussion_r764920939

# Changes
Make the test only check non-system tables


NGA-TRAN (approved):
Nicer to select what we need only

Issue #3338, opened 2021-12-08
Flaky test    end_to_end_cases::database_migration::migrate_table_files_from_one_server_to_another


alamb:
Runs:
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/13659/workflows/e8fb865d-2d00-4fe9-8ea3-f671c58e1009/jobs/101734

Also https://circleci.com/gh/influxdata/influxdb_iox/101739?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Example output:
```
thread 'end_to_end_cases::database_migration::migrate_table_files_from_one_server_to_another' panicked at 'assertion failed: t_0.elapsed() < Duration::from_secs(10)', influxdb_iox/tests/common/server_fixture.rs:289:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:100:14
   2: core::panicking::panic
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:50:5
   3: end_to_end::common::server_fixture::ServerFixture::wait_server_initialized::{{closure}}
             at ./tests/common/server_fixture.rs:289:13
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/future/mod.rs:80:19
   5: end_to_end::end_to_end_cases::database_migration::migrate_table_files_from_one_server_to_another::{{closure}}
             at ./tests/end_to_end_cases/database_migration.rs:233:5
   6: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/future/mod.rs:80:19
   7: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/future/future.rs:119:9
   8: tokio::runtime::basic_scheduler::Inner<P>::block_on::{{closure}}::{{closure}}
[...]

PR #3339, opened 2021-12-08
fix: flaky `migrate_table_files_from_one_server_to_another ` test

alamb:
fixes https://github.com/influxdata/influxdb_iox/issues/3338

# Analysis
The test wants to wait for the rebuild database to be ready, but it was actually verifying that *all*  databases on the shared fixture had initialized, which is not always true (as some tests leave their databases in a bad state).  Here is a failure with more diagnostic information that shows what is happening: [circleci](https://circleci.com/api/v1.1/project/github/influxdata/influxdb_iox/101769/output/105/0?file=true&allocation-id=61b0dba46785d124eab10d1a-0-build%2F69B73C28)

# Changes
Only check the initialization status for the database that is being rebuilt


pauldix (approved):
lgtm

PR #3340, opened 2021-12-08
fix: allow setting the same server ID twice

crepererum:
This is important for idempotence and simplifies clients and helper
scripts a lot.


pauldix (approved):
Nice enhancement

alamb (approved):

PR #3346, opened 2021-12-08
feat: add regex to bucket writers assignment in data generator

### LINKED ISSUE #3344: Allow data generator to assign agents to databases through regex match
Currently, the data generator allows users to specify what ratio of the supplied list of databases should use what configs. A user should alternatively be able to assign agents to databases based on a regex match against their name. These would be mutually exclusive in a data generator spec. Either the user can use ratios or they can use regexes.

pauldix:
This adds the ability to specify a regex to match against database names when specifying what agents should write to which buckets in the data generator.

A default has also been added for ratio so that it doesn't need to be specified if only a single database writer is defined.

Closes #3344

### DIFF
--- a/iox_data_generator/src/specification.rs
+++ b/iox_data_generator/src/specification.rs
@@ -29,20 +29,20 @@ pub enum Error {
     InvalidSamplingInterval { source: humantime::DurationError },
 
     #[snafu(display(
-        "Agent {} referenced in bucket_writers, but not present in spec",
+        "Agent {} referenced in database_writers, but not present in spec",
         agent
     ))]
     AgentNotFound { agent: String },
 
-    #[snafu(display("bucket_writers can only use ratio or regex, not both"))]
-    BucketWritersConfig,
+    #[snafu(display("database_writers can only use database_ratio or database_regex, not both"))]
+    DatabaseWritersConfig,
 
     #[snafu(display(
-        "bucket_writer missing regex. If one uses a regex, all others must also use it"
+        "database_writer missing database_regex. If one uses a regex, all others must also use it"
     ))]
     RegexMissing,
 
-    #[snafu(display("bucket_writers regex {} failed with error: {}", regex, source))]
+    #[snafu(display("database_writers regex {} failed with error: {}", regex, source))]
     RegexCompile { regex: String, source: regex::Error },
 }
 
@@ -69,8 +69,8 @@ pub struct DataSpec {
     pub tag_sets: Vec<TagSetsSpec>,
     /// The specification for the agents that can be used to write data to databases.
     pub agents: Vec<AgentSpec>,
-    /// The specification for writing to the provided list of buckets.
-    pub bucket_writers: Vec<BucketWriterSpec>,
+    /// The specification for writing to the provided list of databases.
+    pub database_writers: Vec<DatabaseWriterSpec>,
 }
 
 impl DataSpec {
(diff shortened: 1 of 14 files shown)


iox_data_generator/src/specification.rs
```diff
@@ -76,25 +88,10 @@ impl DataSpec {
...
-            // validate the agents actually exist in the spec
-            for a in &w.agents {
-                let _ = self.agent_by_name(&a.name).unwrap_or_else(|| {
-                    panic!(
-                        "agent {} referenced in bucket writers, but isn't in the spec",
-                        a.name
-                    )
-                });
-            }
+        let use_ratio =
```
alamb, line 91: Would it make sense to do `use_ratio = <if any bucker writer has ratio set>` rather than just checking the first one?

When written this way it seems like if some writer other than the first has a `ratio` specified, it will be silently ignored in favor of `regex`

iox_data_generator/src/specification.rs
```diff
@@ -732,4 +763,183 @@ agents = [{name = "bar", sampling_interval = "1m", count = 3}]
...
+[[agents.measurements]]
+name = "cpu"
+[[agents.measurements.fields]]
+name = "host"
+template = "server"
+
+[[agents]]
+name = "bar"
+[[agents.measurements]]
+name = "whatevs"
```
alamb, line 859: whatevs whatevs!

alamb (approved):

iox_data_generator/src/specification.rs
```diff
@@ -115,19 +112,43 @@ impl DataSpec {
...
-
-            for org_bucket in &buckets[start..end] {
+            let selected_buckets = if use_ratio {
+                if w.regex.is_some() {
+                    return BucketWritersConfig.fail();
+                }
+
+                if start >= buckets.len() {
+                    warn!(
+                        "Bucket_writers percentages > 1.0. Writer {:?} and later skipped.",
```
crepererum, line 122:
```suggestion
                        "Bucket_writers ratio > 1.0. Writer {:?} and later skipped.",
```

iox_data_generator/src/specification.rs
```diff
@@ -236,8 +263,12 @@ pub struct TagSetsSpec {
 #[serde(deny_unknown_fields)]
 pub struct BucketWriterSpec {
```
crepererum, line 264: As a future user, it took me quite a while to understand the data flow graph here and why "ratio" and "regex" don't work together (or at least why it doesn't make sense). I think what would help:

1. rename `ratio` to `database_ratio`
2. rename `regex` to `database_regex`
3. replace the term "bucket" with "database" everywhere except at the one place where it specifies a bucket within an org (i.e. `database_name = <org>_<bucket>`
4. clearly state that the ratio here is an interval "cum sum prev. ratios" to "cum sum prev. ratios + ratio" of the database name list. The database name list is NOT shuffled or anything.

crepererum (changes requested):

iox_data_generator/src/specification.rs
pauldix: Does it make sense why you wouldn't use them together?

My thinking was that if you're using ratios, you're describing how to split up the entire list of provided databases. If you're using regex, then you have some naming scheme in the provided databases so you want to split them up explicitly based on that naming scheme.

pauldix: The bucket/database thing is also where I thought this whole thing is a little wonky. The data generator writes to the InfluxDB 2.0 API so it writes to org & bucket. However, the list this takes is a database list because I assumed that would be easiest to generate since that's what would be getting used to configure the IOx servers. I can change it to `database`.

crepererum: > Does it make sense why you wouldn't use them together?

I'm wondering how the dataflow graph would look like if they are used together. What is "ratio" going to measure them, pre regex split or post regex split?

crepererum (approved):

PR #3347, opened 2021-12-08
feat: Optionally build Kafka

### LINKED ISSUE #3315: Put write_buffer, rdkafka, etc behind a feature flag
OSS users shouldn't have to build write_buffer, rdkafka, etc by default if they're not going to have a write buffer. Add a feature flag that's off by default.

carols10cents:
Closes #3315.

This saves me ~15 seconds on a release build. According to the Cargo timings analysis, on this branch:

```
$ cargo build --release 
```

builds 441 total units (crates?) in 7m 59.2s.

```
$ cargo build --release --features kafka
```

builds 455 total units in 8m 13.6s.

### DIFF
--- a/write_buffer/src/config.rs
+++ b/write_buffer/src/config.rs
@@ -12,7 +12,6 @@ use trace::TraceCollector;
 use crate::{
     core::{WriteBufferError, WriteBufferReading, WriteBufferWriting},
     file::{FileBufferConsumer, FileBufferProducer},
-    kafka::{KafkaBufferConsumer, KafkaBufferProducer},
     mock::{
         MockBufferForReading, MockBufferForReadingThatAlwaysErrors, MockBufferForWriting,
         MockBufferForWritingThatAlwaysErrors, MockBufferSharedState,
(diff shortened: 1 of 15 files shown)


write_buffer/src/config.rs
```diff
@@ -142,6 +129,37 @@ impl WriteBufferConfigFactory {
...
+        Ok(Arc::new(kafka_buffer) as _)
+    }
+
+    #[cfg(not(feature = "kafka"))]
+    async fn kafka_buffer_producer(
+        &self,
+        _db_name: &str,
+        _cfg: &WriteBufferConnection,
+    ) -> Result<Arc<dyn WriteBufferWriting>, WriteBufferError> {
+        panic!(
```
crepererum, line 157: Can we please not panic if a binary is not compiled using the right features? For object store, we normally just return an error. That would be quite easy here as well since `WriteBufferError` is basically just a `Box<dyn Error>`:

```rust
String::from(
    "`WriteBufferWriting` of type `kafka` requested, but Kafka support was not included \
    in this build by enabling the `kafka` feature"
).into()
```

crepererum (changes requested):
Nitpick regarding the error handling. I also think you have to adjust the `perf.py`:

https://github.com/influxdata/influxdb_iox/blob/8f81120d54ea10e4a3c6c316c17b98eb716145c2/perf/perf.py#L400-L403

write_buffer/src/config.rs
carols10cents: Absolutely!

carols10cents:
@crepererum 2 new commits on the end here ready for re-review! Thank you!!

crepererum (approved):
:)

PR #3348, opened 2021-12-09
refactor: Rename `QueryResults::to_batches` and reduce test duplication

alamb:
# Rationale

I found myself with some time and had a few software itches to scratch

# Changes:
1. Rename `QueryResults::to_batches` to `QueryResults::collect` to be consistent with DataFusion
1. Use QueryResults::collect instead of `collect_query` in end to end test code


pauldix (approved):
Nice. As they say, if you've got time to lean, you've got time to clean.

Issue #3343, opened 2021-12-08
Support compact_object_store_partition


NGA-TRAN:
__Use case:__
Following up the work in #3089 that compacts all specified object store chunks that belong to the same partition, let us support compact all OS chunks of a given partition. Currently, most interactions between users and IOx through specifying partition. This support will go well with that perspective

__Proposal:__
This `compact_object_store_partition` will be done similar to `compact_object_store_chunks` but instead of getting the OS chunk ids as a parameter, we will read them from the given partition 

__Current behaviour:__
Not available yet

__Desired behaviour:__
The  `compact_object_store_partition` will compact all of the given partition's OS chunks into an OS chunk

PR #3350, opened 2021-12-09
test: propogate delete tests for compact OS chunks

NGA-TRAN:
One more important test for #3089 and #3343

Test deletes that happen before and during OS compaction

### DIFF
--- a/server/src/db/lifecycle/compact_object_store.rs
+++ b/server/src/db/lifecycle/compact_object_store.rs
@@ -495,7 +495,10 @@ async fn update_in_memory_catalog(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{db::{test_helpers::write_lp, load::load_or_create_preserved_catalog, catalog::Catalog}, utils::make_db};
+    use crate::{
+        db::{load::load_or_create_preserved_catalog, test_helpers::write_lp},
+        utils::make_db,
+    };
     use data_types::{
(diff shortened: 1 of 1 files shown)


server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -732,10 +738,10 @@ mod tests {
             .unwrap()
             .unwrap()
             .id();
-        //
+
         // persist chunk 2
-        write_lp(&db, "cpu,tag1=cookies bar=2 20"); // delete
-        write_lp(&db, "cpu,tag1=cookies bar=3 30"); // duplicate & delete
```
NGA-TRAN, line 738: the duplicate comment was wrong

server/src/db/lifecycle/compact_object_store.rs
```diff
@@ -717,13 +720,16 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_compact_os_duplicates_and_hard_deletes() {
+    async fn test_compact_os_soft_and_hard_deletes() {
```
NGA-TRAN, line 723: I have modified the current hard delete test to make it cover both hard and soft delete. This is a concurrency test. This test is quite long but I neither want to split it into many small tests nor put the setup into a different function. The reason is they are tightly together and we want to see the data there to verify the output.

e-dard (approved):
Love these tests @NGA-TRAN !

PR #3356, opened 2021-12-10
WIP: less async trait => faster compiles

carols10cents:
not ready yet, this is looking like a good direction though


alamb (commented):
This PR is amazingly epic @carols10cents  -- perhaps as part of it you can add a note in the https://github.com/influxdata/influxdb_iox/blob/main/docs/style_guide.md that we don't wan to be using `async_trait` along with the rationale.

Maybe even worth a blog post (and then perhaps linking to the the `async_trait` crate docs) saying that `async_trait` comes at the expense of non trivial additional compilation time

docs/style_guide.md
```diff
@@ -296,3 +296,61 @@ fn google_cloud() -> Result<()> {
...
+impl SomeTrait for SomeType {
+    fn some_method(&self, arg: Type) -> BoxFuture<'_, ReturnType> {
+        async move {
+            // body
+        }.boxed()
+    }
+}
+```
+
+In words:
```
alamb, line 333: ❤️

carols10cents:
I think there's progress being made fixing the underlying rustc issue, so I'm putting this on hold for the moment.

carols10cents:
😬 I'm going to close this PR as it's just accumulating merge conflicts, but I'll keep the branch around in case anyone wants to cherry-pick anything for particular crates while we're waiting. The PR that might fix the issue is still in limbo :(

PR #3361, opened 2021-12-13
refactor: allow database init to recover from errors

### LINKED ISSUE #3335: Database Init Recovery
# Current Situation
At the moment we require "operator intervention" for basically every database init error:

https://github.com/influxdata/influxdb_iox/blob/76befe94ad14cd121d6fc5c58aa112997d9e211a/server/src/database.rs#L837-L850

Note that we we already retry many IO operations, but this might not be sufficient of larger issues (e.g. unhealthy K8s cluster, connectivity issues, AWS downtimes like yesterday). So at the moment we basically terminate the init process and wait for some operator to do something. This is quite some burden for servers w/ 1k+ databases. Also we don't actually provide gRPC APIs and CLI commands to recover a database from all the error states ("force rebuild catalog" from #3292 being a notable exception / escape hatch).

The init process is a bit special because in contrast to admin actions, catalog inspect calls, queries, writes and metrics (basically all HTTP and gRPC) as well as lifecycle-driven ops and write buffer-driven writes -- all of which are retried by the external or internal driver/actor--, IOx itself is the actual driver and currently doesn't retry.

# Proposal
Add a retry logic to most/all init states. Most error states can be reset to a non-error state that they have originated from.

We might need to spend some thought if this also applies to "replay" though, because this would probably double-replay some messages. So we might need to reload the preserved catalog.

The error states should be kept however, since they convey important information to the users and conductor. I think [`wait_for_init`](https://github.com/influxdata/influxdb_iox/blob/76befe94ad14cd121d6fc5c58aa112997d9e211a/server/src/database.rs#L484-L512) should also stop and fail on the first error, so we don't block forever.

# Prior Art
In #3318 the error handling was changed for the write buffer creation, so we don't terminate database initialization when a Kafka cluster is unhealthy. However the wiring could probably be improved to add an improved, potentially increasing backoff mechanism.

# "Real World" Cases
- #3287
- test bench errors if minio gets restarted, e.g.:
  `2021-12-08T10:55:30.848240Z ERROR server::database: database in error state - operator intervention required db_name=myorg_mytemplate3343 e=error getting database owner info: error loading owner info: AWS S3-based Object Store error: Unable to GET data. Bucket: iox, Location: dbs/de87efa5-c2c1-405d-a227-3437240ed319/owner.pb, Error: Error during dispatch: connection error: Connection reset by peer (os error 104) (HttpDispatch(HttpDispatchError { message: "Error during dispatch: connection error: Connection reset by peer (os error 104)" })) state=OwnerInfoLoadError`

crepererum:
Closes #3335.


server/src/database.rs
```diff
@@ -715,7 +729,7 @@ impl Drop for Database {
 #[derive(Debug)]
 struct DatabaseShared {
     /// Configuration provided to the database at startup
-    config: DatabaseConfig,
+    config: RwLock<DatabaseConfig>,
```
alamb, line 732: I don't fully understand the need for a RwLock -- it seemed to me that one of the nice properties of the various database states was using the type system to encode the state transitions / when things could change

Allowing `config` to be changed seems to break that nice property

If you need to allow multiple readers, perhaps you could just use `config: Arc<DatabaseConfig>` rather than a lock?

server/src/database.rs
```diff
@@ -869,20 +879,29 @@ async fn background_worker(shared: Arc<DatabaseShared>) {
...
 }
 
 /// Try to drive the database to `DatabaseState::Initialized` returns when
 /// this is achieved or the shutdown signal is triggered
 async fn initialize_database(shared: &DatabaseShared) {
-    let db_name = &shared.config.name;
+    let db_name = shared.config.read().name.clone();
     info!(%db_name, "database initialization started");
 
+    // error throttle controshared.config.name.l
```
alamb, line 901: I don't understand this comment (what is a `controshared.config.name.l`?)

server/src/database.rs
```diff
@@ -891,50 +910,61 @@ async fn initialize_database(shared: &DatabaseShared) {
...
                     );
-                    None
+                    throttled_error = true;
+
+                    // exponential backup w/ jitter, decorrelated
+                    // see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+                    let mut rng = thread_rng();
+                    sleep = Duration::from_secs_f64(MAX_BACKOFF.as_secs_f64().min(
+                        rng.gen_range(INIT_BACKOFF.as_secs_f64()..(sleep.as_secs_f64() * 3.0)),
+                    ));
```
alamb, line 956: Given this has some magic numbers (and I am a big confused about the calculation for the backoff), what would you think about using a constructor on `TransactionOrWait` and encapsulating `TransationAndSleep` , and throttled. 

The logic might be easier to discover.

Maybe something like 
```
struct Backoff {
  sleep: Duration
}

impl Backoff {
  fn new() -> Self {
    // default backoff
   }
   fn backoff(self) -> Self {
      // put random backoff here
    }
}
```

Then instead of 
```rust
    let mut throttled_error = false;
    let mut sleep = INIT_BACKOFF;
``` 

In your main loop you could have something like

```rust
    let mut backoff: Option<Backoff> = None;
```

And update it in the loop

server/src/database.rs
```diff
@@ -891,50 +910,61 @@ async fn initialize_database(shared: &DatabaseShared) {
...
                         %db_name,
                         %e,
                         %state,
-                        "database in error state - operator intervention required"
+                        "database in error state - wait until retry"
                     );
-                    None
+                    throttled_error = true;
+
+                    // exponential backup w/ jitter, decorrelated
```
alamb, line 951:
```suggestion
                    // exponential backoff w/ jitter, decorrelated
```

server/src/database.rs
```diff
@@ -891,50 +910,61 @@ async fn initialize_database(shared: &DatabaseShared) {
...
+                    // see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+                    let mut rng = thread_rng();
+                    sleep = Duration::from_secs_f64(MAX_BACKOFF.as_secs_f64().min(
+                        rng.gen_range(INIT_BACKOFF.as_secs_f64()..(sleep.as_secs_f64() * 3.0)),
```
alamb, line 955: By using `INIT_BACKOFF .. sleep*3` doesn't that mean the backoff can actually decrease (rather than increase) each iteration?

I would have expected it to be something like `sleep = min(MAX_BACKOFF, sleep.exp(1 + rand()))` 

My brain may be all mushy today from reviewing so much disparate code though

server/src/database.rs
```diff
@@ -943,40 +973,61 @@ async fn initialize_database(shared: &DatabaseShared) {
...
-                        info!(%db_name, %e, "cannot create write buffer, wait a bit and try again");
-                        tokio::time::sleep(INIT_BACKOFF).await;
                         DatabaseState::WriteBufferCreationError(state, Arc::new(e))
                     }
                     Err(e) => DatabaseState::ReplayError(state, Arc::new(e)),
                 }
             }
-            state => unreachable!("{:?}", state),
+            DatabaseState::ReplayError(state, _) => {
+                warn!(%db_name, "throwing away loaded catalog to recover from replay error");
```
alamb, line 1015: 👍

server/src/database.rs
```diff
@@ -1060,19 +1111,45 @@ pub enum InitError {
...
-///
-/// 1. It reaches `Initialized`
+/// ```text
+///                      (end)
+///                        ^
+///                        |
+///                 [NoActiveDatabase]
+///                        ^
+///                        |
+/// (start)------------>[Known]<------------>[DatabaseObjectStoreLookupError]
```
alamb, line 1121: I think the state machine actually looks more like this (the error state transitions go directly to the next stage rather than back

```
...
[Known]------------>[DatabaseObjectStoreLookupError]
   |                              |
   |                              |
   v                              |
[DatabaseObjectStoreFound] <------+
...
```

server/src/database.rs
```diff
@@ -1081,25 +1158,19 @@ enum DatabaseState {
     OwnerInfoLoaded(DatabaseStateOwnerInfoLoaded),
     RulesLoaded(DatabaseStateRulesLoaded),
     CatalogLoaded(DatabaseStateCatalogLoaded),
+
+    // Terminal state (positive)
```
alamb, line 1162:
```suggestion
    // Terminal state (success)
```

server/src/database.rs
```diff
@@ -1081,25 +1158,19 @@ enum DatabaseState {
...
     Initialized(DatabaseStateInitialized),
 
-    // Error states
-    /// Terminal State
-    DatabaseObjectStoreLookupError(DatabaseStateKnown, Arc<InitError>),
-    /// Terminal State
+    // Terminal state (negative)
```
alamb, line 1165:
```suggestion
    // Terminal state (failure)
```

server/src/database.rs
alamb: Maybe could put this `warn!` into the actual `rollback()` function to ensure it was called.

alamb (commented):
This PR is quite cool @crepererum  -- it is neat to see that the databases can get out of  error states. 👍 

I think this PR should really have some sort of testing as well

Perhaps you could write an end to end test 
1. creates a database
2. releases it
3. move away owner.pb
4. restarts it
5. waits for the db to get into `OwnerInfoLoadError`
6. puts owner.pb back
7. waits for the db to initialize successfully

I also had some questions about the use of a `RwLock` and the calculation of exponential backoff that I would like answered before I feel comfortable approving this PR.

server/src/database.rs
crepererum: The lock is here to allow the change of `skip_replay`. The old solution directly changed the state which for the new state machine is kinda tricky, because you need to hit the precise moment when the DB is in the `ReplayError` state. So instead we're now using the boolean flag to control this and allow the normal startup system to just use the flag to converge to the desired end state (in this case by skipping replay).

server/src/database.rs
crepererum: Probably a "paste random code here" error :sweat_smile:

server/src/database.rs
crepererum: I think you're observation is correct. However this should only happen w/ a low likelyhood and is done to decorrelate the retry events. (see linked AWS post)

server/src/database.rs
crepererum: fixed

server/src/database.rs
crepererum: done

server/src/database.rs
crepererum: done

server/src/database.rs
crepererum: I've tried to improve the docs a bit and explain why not all wait operations are error backoffs.

crepererum:
@alamb I've added a test. I've also changed the state machine so it can also recover from `NoActiveDatabase` (there was no reason why it shouldn't be able to do so) and extended the `claim_database` test accordingly.

server/src/database.rs
```diff
@@ -2101,6 +2101,41 @@ mod tests {
...
+
+        // create second database
+        let database = Database::new(Arc::clone(&application), config);
+        database.wait_for_init().await.unwrap_err();
+
+        // recover database by fixing owner file
+        iox_object_store.put_owner_file(owner_backup).await.unwrap();
+        tokio::time::timeout(Duration::from_secs(10), async move {
+            loop {
+                if database.wait_for_init().await.is_ok() {
```
alamb, line 2154: 👍

alamb (approved):
Thank you @crepererum 


> The lock is here to allow the change of skip_replay. The old solution directly changed the state which for the new state machine is kinda tricky, because you need to hit the precise moment when the DB is in the ReplayError state. So instead we're now using the boolean flag to control this and allow the normal startup system to just use the flag to converge to the desired end state (in this case by skipping replay).

I still don't really like this but I admit I am now just whining as I don't have a better alternate suggestion.

alamb:
cc @tustvold you may be interested in this

server/src/database.rs
tustvold: > because you need to hit the precise moment when the DB is in the ReplayError state

I don't follow why this is a problem, if you acquire a freeze handle you effectively prevent any concurrent transitions?

server/src/database.rs
```diff
@@ -891,50 +920,53 @@ async fn initialize_database(shared: &DatabaseShared) {
...
-                | DatabaseState::DatabaseObjectStoreFound(_)
-                | DatabaseState::OwnerInfoLoaded(_)
-                | DatabaseState::RulesLoaded(_)
-                | DatabaseState::CatalogLoaded(_)
-                | DatabaseState::WriteBufferCreationError(_, _) => {
+                _ if state.error().is_none() || (state.error().is_some() && throttled_error) => {
                     match state.try_freeze() {
-                        Some(handle) => Some((DatabaseState::clone(&state), handle)),
+                        Some(handle) => {
+                            TransactionOrWait::Transaction(DatabaseState::clone(&state), handle)
```
tustvold, line 928: It occurs to me that this dates from before you could wait on a freeze handle, we could probably remove this logic now

server/src/database.rs
```diff
@@ -869,20 +879,39 @@ async fn background_worker(shared: Arc<DatabaseShared>) {
...
 /// Try to drive the database to `DatabaseState::Initialized` returns when
 /// this is achieved or the shutdown signal is triggered
 async fn initialize_database(shared: &DatabaseShared) {
-    let db_name = &shared.config.name;
+    let db_name = shared.config.read().name.clone();
     info!(%db_name, "database initialization started");
 
+    // error throttling
+    // - checks if the current error was already throttled
+    // - keeps a backoff duration that will change over the curse of multiple errors
```
tustvold, line 911:
```suggestion
    // - keeps a backoff duration that will change over the course of multiple errors
```

server/src/database.rs
```diff
@@ -1060,19 +1108,55 @@ pub enum InitError {
 
 /// The Database startup state machine
 ///
-/// A Database starts in DatabaseState::Known and advances through the
-/// non error states in sequential order until either:
-///
-/// 1. It reaches `Initialized`
-///
-/// 2. It is reset to `Known` and starts initialization again
+/// ```text
```
tustvold, line 1111: :ok_hand:

tustvold (commented):
I think it is possible to remove the need for the new RWLock, along with the throttled_error complexity, but otherwise looks good

crepererum:
@tustvold simplified the locking and removed the user-visible state rollback (as we've agreed on during a 1:1 chat).

server/src/database.rs
crepererum: Done, it doesn't fully eliminate the variables though.

server/src/database.rs
```diff
@@ -676,30 +680,30 @@ impl Database {
 
     /// Recover from a ReplayError by skipping replay
     pub async fn skip_replay(&self) -> Result<(), Error> {
-        let db_name = &self.shared.config.name;
```
tustvold, line 679: I don't think this method needs to be changed anymore - and therefore the RWLock can be removed?

crepererum: You need it to trigger "skip replay" via gRPC. It's the only way to toggle the flag for a single database when an admin see it's broken and accepts that replay can be skipped for that particular database.

alamb:
THanks @crepererum

Issue #3354, opened 2021-12-10
Add "database initialization state" to CLI: `database list --detailed`


alamb:
__Use case:__
I am reviewing databases in production and at least some of them ave having problems (due to other tickets). However, it is hard to tell which databases have problems and which are ok

When I list the status of the databases on this server I can get the Name and UUId. I don't see anything about their status (are they initialized, etc) so I can't easily figure out what is going on with what database from this output

```
(arrow_dev) alamb@MacBook-Pro:~/Software/influxdb_iox$ ./target/debug/influxdb_iox  database list --detailed
+-----------------------------------+--------------------------------------+
| Name                              | UUID                                 |
+-----------------------------------+--------------------------------------+
| 797a0180da80497c_26141866d829c862 | a9f4cb7a-aaa7-4a0e-84b6-f5fde5f85315 |
| 797a0180da80497c_418d700797265a54 | ecaea0ee-a230-4665-b8e0-0e0d42b62c7c |
| 797a0180da80497c_4d2743db995d4bce | 18c168ae-c589-4b81-a959-b7c1f81bb84b |
| 797a0180da80497c_5f0b4940239581e0 | e090335f-f032-4455-b4b0-c25c9e1cc60d |
| 797a0180da80497c_628819f03a1a47d5 | 4174cac7-7696-4061-9e89-dd054e79e98d |
| 797a0180da80497c_6adb5627945fe0d4 | 0b1e21cc-c9a9-479d-9982-d72e883c3eb3 |
| 797a0180da80497c_73515dc225b48ddc | 0d7b8c1d-2991-4c3d-a13c-ffd83521b9a9 |
| 797a0180da80497c_96d6733751adec41 | 7db47e09-d1a2-4c83-ac87-661486f6cdba |
| 797a0180da80497c_972935bc24f310a6 | f11890ec-0053-41d1-806e-b2452d02263c |
| 797a0180da80497c_a909f07d9af6d54a | 8a4ca746-5b77-4378-9832-cd79e765c686 |
| 797a0180da80497c_a9434e557b465991 | edcc1760-db46-492e-983c-227f514be6a8 |
| 797a0180da80497c_aca5b85fb8c2ef6e | fdefab1f-a58f-4331-9c08-763a24b70ce7 |
| 797a0180da80497c_b6805dec6a42dea4 | b922ebe4-f253-4d63-8635-6612662930a7 |
[...]

PR #3377, opened 2021-12-14
feat: Add database initialization state and errors to CLI and remove list_databases_detailed gRPC

alamb:
# Rationale

Fixes: https://github.com/influxdata/influxdb_iox/issues/3354

When we were having database level problems, it was hard to find out which specific database(s) on a server was having problems and what the problem was

It also turns out there was a non trivial amount of overlap between `GetServerStatus` and the `ListDatabaseDetailed` management gRPC calls

# Changes:
1. Remove the ListDetailedDatabase call from gRPC (I verfied that conductor doesn't use the gRPC)
2. Add UUID field to ServerStatus
1. Add "state" and "error" columns to the output of list-databases

So even with a new end to end test, this PR still removes code overall 

# Examples

```shell
cargo run -- database list --detailed
```

When things are all good:
```
+-----------------------------------+--------------------------------------+-------------+--------+
| Name                              | UUID                                 | State       | Error  |
+-----------------------------------+--------------------------------------+-------------+--------+
| 26f7e5a4b7be365b_917b97a92e883afc | 2c552958-dd89-438b-9bd7-8c8b23fb3dfa | Initialized | <none> |
| my_db                             | dd627e7d-a7a7-4fa4-9ae1-0854c8abd55a | Initialized | <none> |
| tracing_spans                     | b9fe3669-7641-46b6-a691-68b5f91b17e5 | Initialized | <none> |
+-----------------------------------+--------------------------------------+-------------+--------+
```

When I broke a local database on purpose
```sql
+-----------------------------------+--------------------------------------+--------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Name                              | UUID                                 | State              | Error                                                                                                                                                                               |
+-----------------------------------+--------------------------------------+--------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 26f7e5a4b7be365b_917b97a92e883afc | 2c552958-dd89-438b-9bd7-8c8b23fb3dfa | Initialized        | <none>                                                                                                                                                                              |
| my_db                             | <UUID not yet known>                 | OwnerInfoLoadError | error getting database owner info: error loading owner info: Object at location dbs/dd627e7d-a7a7-4fa4-9ae1-0854c8abd55a/owner.pb not found: No such file or directory (os error 2) |
| tracing_spans                     | b9fe3669-7641-46b6-a691-68b5f91b17e5 | Initialized        | <none>                                                                                                                                                                              |
+-----------------------------------+--------------------------------------+--------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```


generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -34,9 +34,6 @@ service ManagementService {
   // Claim a released database.
   rpc ClaimDatabase(ClaimDatabaseRequest) returns (ClaimDatabaseResponse);
 
-  // List databases with their metadata.
```
alamb, line 37: I am surprised that the `buf` breaking change lint doesn't flag removing a gRPC call as a breaking change 🤔

influxdb_iox/src/commands/database.rs
```diff
@@ -254,19 +254,54 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
...
+                if !initialized {
+                    eprintln!("Can not list databases. Server is not yet initialized");
+                    if let Some(err) = error {
+                        println!("WARNING: Server is in error state: {}", err.message);
+                    }
+                    return Ok(());
+                }
 
-                if !databases.is_empty() {
+                if !database_statuses.is_empty() {
```
alamb, line 270: This is the new CLI formatting code (that now uses `get_server_status`)

generated_types/protos/influxdata/iox/management/v1/service.proto
alamb: kudos to @tustvold  for pointing out the `GetServerStatus` call

alamb:
FYI @carols10cents

generated_types/protos/influxdata/iox/management/v1/service.proto
```diff
@@ -414,6 +388,9 @@ message DatabaseStatus {
 
   // Current initialization state of the database.
   DatabaseState state = 3;
+
+  // The UUID of the database, if known
+  optional bytes uuid = 4;
```
tustvold, line 393:
```suggestion
  bytes uuid = 4;
```
I don't think there is a need for explicit nullability for this field, i.e. being able to distinguish between no UUID and an empty bytes payload. Not using optional makes this consistent with the rest of the [API](https://github.com/influxdata/influxdb_iox/blob/main/docs/protobuf.md#explicit-presence)

tustvold (approved):
I'd rather not use explicit nullability in the proto, otherwise LGTM :+1:

generated_types/protos/influxdata/iox/management/v1/service.proto
tustvold: > I am surprised that the buf breaking change lint doesn't flag removing a gRPC call as a breaking change

I have a sneaky suspicion `buf` breaking change only audits protobuf changes and not gRPC

generated_types/protos/influxdata/iox/management/v1/service.proto
alamb: The reason I think it needs to be optional is that the UUID isn't available to `database` known until later in the initialization sequence. I actually don't understand how the server would know about a database's name but not its UUID but that appears to be how the code is structured today:

https://github.com/influxdata/influxdb_iox/blob/main/server/src/database.rs#L1174-L1191

https://github.com/influxdata/influxdb_iox/blob/main/server/src/database.rs#L411-L414

tustvold: This is why I really dislike this "feature" of proto 3. **All** fields in protobuf are optional, specifying the `optional` label just means adds explicit nullability. That is the ability to distinguish tell if the user specified the default value, or if the field was defaulted by the protocol.

Without the `optional` label, `uuid` is still optional - it will just be implicitly defaulted by the protocol to an empty bytes array. I tried to expand upon this in the linked doc, but perhaps this can be further clarified?

alamb: @tustvold  and I had a chat and I will remove the `optional` thing in the protobuf and instead encode no UUID as empty bytes

alamb: fixed in https://github.com/influxdata/influxdb_iox/pull/3377/commits/6609fbbbff7bf11f2d94947b190ed8e22b7513de

alamb:
> Without the optional label, uuid is still optional - it will just be implicitly defaulted by the protocol to an empty bytes array. I tried to expand upon this in the linked doc, but perhaps this can be further clarified?

I think what I was missing was "what to do instead" (which in this case was encode "no UUID" as "empty array of bytes"). All good now 👍

alamb:
rebased to satisfy `Semantic Pull Request` (it didn't like the message of the merge commit 🙄 , which will become irrelevant anyways as this PR is destined for `squash`ing)

PR #3378, opened 2021-12-14
docs: IOx Catalogs

NGA-TRAN:


docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
... represents the structure of its actual `Data Organization` and hence looks similar to Figure 1. Note that the bottom layer of Figure 3 shows physical chunks of data (`O-MUB 1`, `RUB 2`, `O-MUB 3`, `F-MUB 4`, `RUB 5`, and `OS 5`) that the catalog does not store but points to. In short, the `In-Memory Catalog` contains a set of [`Tables`](https://github.com/influxdata/influxdb_iox/blob/8a2410e161996603a4147e319c71e5bb38ca9cb7/server/src/db/catalog/table.rs#L20), each of which consists of a set of [`Partitions`](https://github.com/influxdata/influxdb_iox/blob/8a2410e161996603a4147e319c71e5bb38ca9cb7/server/src/db/catalog/partition.rs#L116). A `Partition` includes a set of [`CatalogChunks`](https://github.com/influxdata/influxdb_iox/blob/76befe94ad14cd121d6fc5c58aa112997d9e211a/server/src/db/catalog/chunk.rs#L195), each of which is represented by a [`ChunkStage`](https://github.com/influxdata/influxdb_iox/blob/76befe94ad14cd121d6fc5c58aa112997d9e211a/server/src/db/catalog/chunk.rs#L130) that points to the corresponding Data Chunk(s). Please refer to the previous document, [IOx Data Organization and LifeCycle](link), for the description of chunk stages and chunk types supported in IOx.
```
tustvold, line 70: My personal feeling is that code-specifics are best documented alongside the code they document - helps avoid drift. If it were up to me, I would keep this document higher level

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+
+Each object of the catalog contains information for us to operate the chunk lifecycle, run queries, as well as measure the health of the system[^prop]. For example, the `In-Memory Catalog` object itself includes [`CatalogMetrics`](https://github.com/influxdata/influxdb_iox/blob/873ce27b0c0c5e9da33e6f4fae8c5be7c163c16c/server/src/db/catalog/metrics.rs#L21) that measure how often its tables, partitions, and chunks are locked, what kinds of locks (e.g. shared or exclusive), and their lock wait time[^lock]. This information helps the IOx team to measure lock contention to adjust the database with a right setup. The `(L)` next to each object or property indicates they are modifiable but must be locked while doing so. Locking ensures the catalog consistency and integrity but will bring down the throughput and performance of the system, hence should be always measured for appropriate actions.
```
tustvold, line 72: Minor nit, this sort of seems to imply locks are necessarily bad. I know you know, but it could perhaps be more clear that the issue is lock contention that prevents possible parallelism, not the existence of the locks themselves. If you have a serial workload, eliminating the locks is not going to make it any faster :sweat_smile:

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+
+The `Preserved Catalog` keeps [`DatabaseCheckpoints`](https://github.com/influxdata/influxdb_iox/blob/b39e01f7ba4f5d19f92862c5e87b90a40879a6c9/persistence_windows/src/checkpoint.rs#L554) which represent a sequence number of writes before which all the writes are persisted. To restart a database, the `In-Memory Catalog` will be rebuilt to the state of the latest `DatabaseCheckpoint` found in the `PreservedCatalog`, then all the writes after that will be (re)ingested [^write].
```
tustvold, line 140: This is not correct. The `DatabaseCheckpoint` and `PartitionCheckpoint` are not used to rebuild the `Preserved Catalog`. The preserved catalog's transaction stream is replayed to reconstruct the `Preserved Catalog`. The catalog also has a notion of a checkpoint, to truncate how far back transactions need to replayed from.

The `Preserved Catalog` is then used to populate the `In-memory catalog` with persisted chunks, and to identify the `DatabaseCheckpoint` and PartitionCheckpoint`. 

These are then used to reconstruct the state of the `In-memory Catalog`, in particular the unpersisted data at the time of the last persist operation. This procedure is replay. 

There are some subtleties to it to avoid duplicating data, documented [here](https://github.com/influxdata/influxdb_iox/blob/main/persistence_windows/src/checkpoint.rs), that aim to avoid duplicating already persisted data but these are technically optimisations.

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+
+If a database is still running but in an inconsistent state, it will be rebuilt by momentarily stopping accepting writes and discarding all in-memory data of `O-MUBs`, `F-MUBs`, `RUBs`. Then the `In-Memory Catalog` will be brought back to the latest `DatabaseCheckpoint` found in the `PreservedCatalog` as above and then all the writes after that will be (re)ingested.
```
tustvold, line 142: I'm not sure what process this is referring to

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+
+Since Preserved Catalog only saves data to a consistent state and includes the well-known slow IO reads and writes, their transactions and related locks on persisting objects will impact the data lifecycle workload, especially if a lot of data continues getting persisted. This is another reason for us to monitor transaction and locking contention to adjust our `Data LifeCycle Policy` to ensure ingesting data flow smoothly until persisted.
```
tustvold, line 144: This seems to imply locks are held whilst writing to object store, this is sort of true but also rather misleading.

Whilst performing IO the catalog locks are **not** held. _The rust type system physically won't let you hold them across an await_. Instead the catalog locks are only held whilst evaluating the policy, and are then "downgraded" to lifecycle actions recorded on the chunks that are being manipulated. 

These are sort of "locks", which is why this is sort of true, however, they only prevent other **lifecycle actions** from manipulating those **same** chunks. They do not prevent queries, lifecycle actions on other chunks, etc... from proceeding

tustvold (commented):
Thank you for this, I think this will be a very useful resource going forward :+1:

However, I think there is some confusion between the interaction between the in-memory and preserved catalogs. In particular with regards to how recovery proceeds and what locks are held when. I've tried to provide some clarifying comments, but let me know if anything isn't clear

docs/catalogs.md
crepererum, line 70: `(link)` placeholder

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+│---- ---- ----│              │ ---- ---- │                                                                       
+└──────────────┘              │ ---- ---- │                                                                       
+                              └───────────┘                                                                       
+Figure 1: Data Organization in an IOx Server
+
+```
+
+
+Each database of IOx owns three catalogs demonstrated in Figure 2. The `In-Memory Catalog` (or simply the `Catalog`), contains metadata to access all chunk types in both memory and Object Store described in [IOx Data Organization and LifeCycle](link). The `Preserved Catalog` is used to rebuild the database in case it is shutdown or unexpectedly destroyed by accidents or disasters. The `Query Catalog` includes extra information to access the right chunks and their statistics quickly for running queries, recording necessary logs, and reporting the results back to users. The `Other Properties` are a set of properties created/rebuilt with the database based on some default configuration parameters of the IOx Server or specified by the user who creates it.
```
crepererum, line 46: `(link)` placeholder

docs/catalogs.md
alamb: Maybe it is is referring to the (yet to be implemented) https://github.com/influxdata/influxdb_iox/issues/3381 lol

docs/catalogs.md
```diff
@@ -0,0 +1,167 @@
...
+
+[^write]: InfluxDB ingests data to IOx from Kafka so we know which writes to start (re)ingesting.
+
+## Query Catalog
+[`Query Catalog`](https://github.com/influxdata/influxdb_iox/blob/218042784fdb3bd0aa2c13dcaaf2f39190d42329/server/src/db/access.rs#L123) shown in Figure 4 is an `In-Memory Catalog` plus extra information to access the right chunks and their statistics quickly for running queries, logging necessary information or metrics, and reporting the results back to users. For instance, since IOx uses [DataFusion](https://github.com/apache/arrow-datafusion) to plan and run queries, `UserTables` provide the required interface to DataFusion for doing so. Similarly, since we want to query our own catalog data such as number of tables, partitions, chunks types and so on, we construct `SystemTables` interface to send to DataFusion for that purpose. `ChunkAccess` helps find and prune chunks of the query table that do not include data needed by the query[^query]. `QueryLog` logs necessary information such as times and types the queries are run.
```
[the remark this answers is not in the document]
alamb: yes -- I also think calling this concept a `Catalog` is confusing -- a view on to the `In-Memory` catalog I think is a much more precise description

alamb (commented):
Thanks for writing this down @NGA-TRAN  👍 

I think this document, as @tustvold  has pointed out, might be stronger if we left out the arguments (which I agree with, BTW) for why careful locking is critical

jacobmarble (approved):

NGA-TRAN:
@alamb @crepererum and @tustvold : I will look into detailed comments but from skimming the comments , I will make this document shorter and focus on:

1. Emphasize we have one `catalog` which is the memory one. I would **not** draw a new figure but use the data figure (Figure 1) to add high level description for it.
2. Then I will describe n a very high level that we save this catalog and rebuild it as needed. Will neither try to dig into the-how nor mention lock contention. I do not think it is necessary at this time to mention them. We will add more if we have measurements that are worth documenting.
3. I will also describe very briefly how we use that catalog to query

NGA-TRAN:
@alamb , @tustvold I have rewritten this doc to make it very high level. Can you help review it? 

I know Marco is on vacation and do not want to bother him :)

docs/catalogs.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Catalog
+Figure 1 below shows the organization of both IOx Catalog and Data. The bottom layer `Data Chunks` are physical data described in previous [document](data_organization_lifecycle.md). All the layers from `Catalogs` to `Chunk Stages` are metadata managed by IOx `Catalog`. Each database has its own `Catalog`[^cat] that contains a set of `Tables`, each consists of a set of `Partitions`. A partition includes many `Chunks`, each is represented by a `ChunkStage` that points to their corresponding physical chunk data.
+
+[^cat]: This design might be changed in the future to meet other use cases.
+
+Since the catalog is the central information shared to all users of the database, it needs to keep consistent and integral. Each object of the catalog must be locked in either shared or exclusive mod when accessed or modified. Each catalog object also contains information for us to operate the chunk lifecycle, run queries, as well as measure the health of the system.
```
alamb, line 12:
```suggestion
Since the catalog is the central information shared to all users of the database, it must be kept consistent. Each object of the catalog must be locked in either shared or exclusive mode when accessed or modified. Each catalog object also contains information to operate the chunk lifecycle, run queries, as well as measure the health of the system. 
```

docs/catalogs.md
```diff
@@ -0,0 +1,90 @@
...
+[^cat]: This design might be changed in the future to meet other use cases.
+
+Since the catalog is the central information shared to all users of the database, it needs to keep consistent and integral. Each object of the catalog must be locked in either shared or exclusive mod when accessed or modified. Each catalog object also contains information for us to operate the chunk lifecycle, run queries, as well as measure the health of the system. 
+
+**Examples of information for operating the chunk lifecycle**
+* A chunk includes `LifecycleAction` that keeps track of on-going action of a chunk (e.g. `compacting`, `persisting`) to avoid running the same job on the chunk. 
+* A partition contains `PersistenceWindows` that  keep track of ingested data within a partition to determine when it can be persisted. This allows IOx to receive out of order writes in their timestamps while persisting mostly in non-time overlapping Object Store files.
+
+**Examples of information for measuring the health of the system**
+* A catalog consists of metrics that measure how often its tables, partitions, and chunks are locked, what kinds of locks (e.g. shared or exclusive), and their lock wait time.
```
alamb, line 19:
```suggestion
* A catalog contains metrics that measure how often its tables, partitions, and chunks are locked, what kinds of locks (e.g. shared or exclusive), and their lock wait time.
```

docs/catalogs.md
```diff
@@ -0,0 +1,90 @@
...
+          └────────────┘                          └────────────┘                                               
+                 │                                       │                                                     
+       ┌─────────▼───────┐               ┌───────────────┼─────────────────────┐                               
+       ▼                 ▼               ▼               ▼                     ▼                               
+   ┌───────┐         ┌───────┐       ┌───────┐       ┌───────┐             ┌───────┐                           
+   │Chunk 1│         │Chunk 2│       │Chunk 3│       │Chunk 4│             │Chunk 5│               Chunks      
+   └───────┘         └───────┘       └───────┘       └───────┘             └───────┘                           
+       │                 │               │               │                     │                               
+       ▼                 ▼               ▼               ▼                     ▼                               
+┌──────────────┐ ┌──────────────┐ ┌──────────────┐┌──────────────┐     ┌──────────────┐
```
alamb, line 61: I would personally not include `ChunkStage` as a box in this picture -- but rather would put it as an annotation on the `Chunks` (in other words collapse `Chunk 1` and `ChunkStage 1 (Open)` into the same box

docs/catalogs.md
```diff
@@ -0,0 +1,90 @@
...
+│---- ---- ----│ │  ---- ----   │  │ ---- ---- │ │ ---- ---- ---- ││ ---- ---- ││ ---- ---- │                  
+└──────────────┘ │  ---- ----   │  └───────────┘ │ ---- ---- ---- ││ ---- ---- ││ ---- ---- │                  
+                 └──────────────┘                │ ---- ---- ---- │└───────────┘└───────────┘                  
+                                                 └────────────────┘                                            
+Figure 1: Catalog and Data Organization
+```
+
+
+## Rebuild a Database from its Catalog
+Since the `Catalog` is the core of the database, losing the catalog means losing the database; or rebuilding a database means rebuilding its catalog. Thus to rebuild a catalog,  its information needs to save in a durable storage.
```
alamb, line 81:
```suggestion
Since the `Catalog` is the core of the database, losing the catalog means losing the database; or rebuilding a database means rebuilding its catalog. Thus to rebuild a catalog,  its information needs to saved in a durable storage.
```

docs/catalogs.md
```diff
@@ -0,0 +1,90 @@
...
...des two major steps: first rebuild the catalog that links to the already persisted `OS` chunks, then rerun ingesting data of `O-MUBs`, `F-MUBs` and `RUBs`. To perform those 2 steps, IOx saves `catalog transactions` that include minimal information of the latest stage of all valid `OS` chunks (e.g, chunks are not deleted), and `checkpoints` of when in the past to re-ingest data. At the beginning of the `Catalog rebuild`, those transactions will be read from the Object Store, and their information is used to rebuild the Catalog and rerun the necessary ingestion. Due to IOx `DataLifeCyclePolicy` that is responsible for when to trigger compacting chunk data, the rebuilt catalog may look different from the one before but as long as all the data in previous `O-MUBs`, `F-MUBs`, and `RUBs` are reloaded/recovered and tracked in the catalog, it does not matter which chunk types the data belong to.  Refer to [Catalog Persistence](catalog_persistence.md) for the original design of Catalog Persistence and [CheckPoint](https://github.com/influxdata/influxdb_iox/blob/b39e01f7ba4f5d19f92862c5e87b90a40879a6c9/persistence_windows/src/checkpoint.rs) for the detailed implementation of IOx Checkpoints.
```
alamb, line 83:
```suggestion
Basically, if an IOx server goes down unexpectedly, we will lose the in-memory catalog shown in Figure 1 and all of its in-memory data chunks `O-MUBs`, `F-MUBs`, and `RUBs`. Only data of `OS` chunks are not lost. As pointed out in [IOx Data Organization and LifeCycle](data_organization_lifecycle.md) that chunk data is persisted contiguously with their loading time, `OS` chunks only include data ingested before the one in `MUBs` and `RUBs`. Thus, in principal, rebuilding the catalog includes two major steps: first rebuild the catalog that links to the already persisted `OS` chunks, then rerun ingesting data of `O-MUBs`, `F-MUBs` and `RUBs`. To perform those 2 steps, IOx saves `catalog transactions` that include minimal information of the latest stage of all valid `OS` chunks (e.g, chunks are not deleted), and `checkpoints` of when in the past to re-ingest data. At the beginning of the `Catalog rebuild`, those transactions will be read from the Object Store, and their information is used to rebuild the Catalog and rerun the necessary ingestion. Due to IOx `DataLifeCyclePolicy` that is responsible for when to trigger compacting chunk data, the rebuilt catalog may look different from the one before but as long as all the data in previous `O-MUBs`, `F-MUBs`, and `RUBs` are reloaded/recovered and tracked in the catalog, it does not matter which chunk types the data belong to.  Refer to [Catalog Persistence](catalog_persistence.md) for the original design of Catalog Persistence and [CheckPoint](https://github.com/influxdata/influxdb_iox/blob/b39e01f7ba4f5d19f92862c5e87b90a40879a6c9/persistence_windows/src/checkpoint.rs) for the detailed implementation of IOx Checkpoints.
```

alamb (approved):
I left some small comments, but this looks very nice to me. Thanks for updating it @NGA-TRAN

alamb:
🎉

PR #3379, opened 2021-12-15
fix: treat early server worker exit as proper error

### LINKED ISSUE #3375: IOx Exit Code
**Problem**

Currently if IOx encounters an error, such as a background worker panicking, it will trigger server shutdown. However, the server will then exit with a zero exit-code.

**Proposal**

When shutting down because of an error, IOx should return a non-zero exit code

crepererum:
Instead of just logging the issue, also make sure the error gets
propagated all the way up to the exit code.

Fixes #3375.

### DIFF
--- a/influxdb_iox/src/commands/run/test.rs
+++ b/influxdb_iox/src/commands/run/test.rs
@@ -1,0 +1,67 @@
+//! Implementation of command line option for running server
+
+use std::sync::Arc;
+
+use crate::{
+    influxdb_ioxd::{
+        self,
+        server_type::{
+            common_state::{CommonServerState, CommonServerStateError},
+            test::{TestAction, TestServerType},
+        },
+    },
+    structopt_blocks::run_config::RunConfig,
+};
+use metric::Registry;
+use structopt::StructOpt;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Run: {0}")]
+    Run(#[from] influxdb_ioxd::Error),
+
+    #[error("Invalid config: {0}")]
+    InvalidConfig(#[from] CommonServerStateError),
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "run",
+    about = "Runs in test mode",
+    long_about = "Run the IOx test server.\n\nThe configuration options below can be \
+    set either with the command line flags or with the specified environment \
+    variable. If there is a file named '.env' in the current working directory, \
+    it is sourced before loading the configuration.
+
+Configuration is loaded from the following sources (highest precedence first):
+        - command line arguments
+        - user set environment variables
+        - .env file contents
+        - pre-configured default values"
+)]
+pub struct Config {
+    #[structopt(flatten)]
+    pub(crate) run_config: RunConfig,
+
+    /// Test action
+    #[structopt(
+        long = "--test-action",
+        env = "IOX_TEST_ACTION",
+        default_value = "none"
+    )]
+    test_action: TestAction,
+}
+
+pub async fn command(config: Config) -> Result<()> {
+    let common_state = CommonServerState::from_config(config.run_config.clone())?;
+    let server_type = Arc::new(TestServerType::new(
+        Arc::new(Registry::new()),
+        common_state.trace_collector(),
+        config.test_action,
+    ));
+
+    Ok(influxdb_ioxd::main(common_state, server_type).await?)
+}
--- a/influxdb_iox/src/commands/run/test.rs
+++ b/influxdb_iox/src/commands/run/test.rs
@@ -50,7 +50,9 @@ pub struct Config {
     #[structopt(
         long = "--test-action",
         env = "IOX_TEST_ACTION",
-        default_value = "none"
+        default_value = "None",
+        possible_values = &TestAction::variants(),
+        case_insensitive = true,
     )]
     test_action: TestAction,
 }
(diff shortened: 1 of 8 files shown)


influxdb_iox/src/commands/run/test.rs
```diff
@@ -0,0 +1,67 @@
...
+        - command line arguments
+        - user set environment variables
+        - .env file contents
+        - pre-configured default values"
+)]
+pub struct Config {
+    #[structopt(flatten)]
+    pub(crate) run_config: RunConfig,
+
+    /// Test action
```
carols10cents, line 49: Do we want the valid values for `--test-action` documented here? Or is this command something we only expect to use in internal tests that we deliberately don't want to explain to end users?

influxdb_iox/src/influxdb_ioxd/server_type/test.rs
```diff
@@ -0,0 +1,127 @@
...
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_ref() {
+            "none" => Ok(Self::None),
+            "early-return-from-grpc-worker" => Ok(Self::EarlyReturnFromGrpcWorker),
+            "early-return-from-server-worker" => Ok(Self::EarlyReturnFromServerWorker),
+            "panic-in-grpc-worker" => Ok(Self::PanicInGrpcWorker),
+            "panic-in-server-worker" => Ok(Self::PanicInServerWorker),
+            _ => Err(format!("Invalid test action: {}", s)),
```
carols10cents, line 49: If we do want to document the possible values to show up in the `--help` output, I think it would also be nice to have the possible values in the error message here, so that if you type a value wrong it'll say something like `Invalid test action: arly-return-from-grpc-worker. Valid values are: none, early-return-from-grpc-worker...`

influxdb_iox/src/influxdb_ioxd/server_type/test.rs
```diff
@@ -0,0 +1,127 @@
...
+            panic!("Test panic in gRPC worker");
+        }
+        if self.test_action == TestAction::EarlyReturnFromGrpcWorker {
+            return Ok(());
+        }
+
+        let builder = setup_builder!(builder_input, self);
+        serve_builder!(builder);
+
+        Ok(())
```
carols10cents, line 110: Personally, I think these functions with multiple `if` blocks would be better expressed with `match`:

```suggestion
        match self.test_action {
            TestAction::PanicInGrpcWorker => panic!("Test panic in gRPC worker"),
            TestAction::EarlyReturnFromGrpcWorker => Ok(()),
            _ => {
                let builder = setup_builder!(builder_input, self);
                serve_builder!(builder);

                Ok(())
            }
        }
```

but I don't feel super strongly about it, especially because this is test code :)

carols10cents (approved):
I added a panic in the database background worker and observed on `main` the server exit code was 0, and on this branch it's 1 🎉 !

Just a few small thoughts to take or leave :)

influxdb_iox/src/influxdb_ioxd.rs
```diff
@@ -276,19 +285,28 @@ where
...
+                }
                 Err(error) => {
                     error!(%error, "gRPC server error");
-                    res = res.and(Err(Error::ServingRpc{source: error}))
+                    res = res.and(Err(Error::ServingRpc{source: error}));
                 }
             },
             result = http_server => match result {
-                Ok(_) => info!("HTTP server shutdown"),
+                Ok(_) if frontend_shutdown.is_cancelled() => info!("HTTP server shutdown"),
```
alamb, line 302: 👍

influxdb_iox/tests/end_to_end_cases/run_cli.rs
```diff
@@ -93,3 +93,54 @@ fn test_deprecated_cli_without_server_type() {
     process.kill().unwrap();
     process.wait().unwrap();
 }
+
+#[test]
+fn test_early_return_from_grpc_worker() {
```
alamb, line 98: Nice

alamb (approved):
🥇 for some great testing

I liked @carols10cents  code style suggestions as well

crepererum:
Fixed the test action CLI. Now we have:

```console
$ cargo run -- run test --help
...
        --test-action <test-action>
            Test action [env: IOX_TEST_ACTION=]  [default: None]  [possible values: None, EarlyReturnFromGrpcWorker,
            EarlyReturnFromServerWorker, PanicInGrpcWorker, PanicInServerWorker]
...

$ cargo run -- run test --test-action=Foo
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/influxdb_iox run test --test-action=Foo`
error: 'Foo' isn't a valid value for '--test-action <test-action>'
        [possible values: EarlyReturnFromGrpcWorker, EarlyReturnFromServerWorker, None, PanicInGrpcWorker, PanicInServerWorker]

File: influxdb_iox/src/influxdb_ioxd/server_type/test.rs [hunk 6]
```diff
@@ -27,27 +28,14 @@ impl HttpApiErrorSource for ApplicationError {
...
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.to_lowercase().as_ref() {
-            "none" => Ok(Self::None),
-            "early-return-from-grpc-worker" => Ok(Self::EarlyReturnFromGrpcWorker),
-            "early-return-from-server-worker" => Ok(Self::EarlyReturnFromServerWorker),
-            "panic-in-grpc-worker" => Ok(Self::PanicInGrpcWorker),
-            "panic-in-server-worker" => Ok(Self::PanicInServerWorker),
-            _ => Err(format!("Invalid test action: {}", s)),
-        }
+arg_enum! {
```
alamb, line 31: TIL https://docs.rs/clap/2.34.0/clap/macro.arg_enum.html 👍

PR #3384, opened 2021-12-15
chore: add data generator config for large database

pauldix:
Writes a total of 225k measurements with varying amounts of throughput.


alamb (approved):
I ran this using 

```
cargo run   -p iox_data_generator  -- -s iox_data_generator/schemas/big_db.toml  --output /tmp/data.lp --bucket foo --org bar
```

And checked the output in `/tmp/data.lp` 

Sweet sweet synthetic data. Can't get enough

crepererum (approved):

PR #3388, opened 2021-12-16
test: `object_store` aws test using localstack

crepererum:
This removes 3 "nonexisting region" tests that where testing very
specific error behavior that no local emulator (minio and localstack)
replicate and that don't add much value. It's better to test our AWS
code at all than being to picky.

#3389 demonstrates that we're indeed testing something.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,6 +168,7 @@ jobs:
             }
 
   test:
+    # setup multiple docker images (see https://circleci.com/docs/2.0/configuration-reference/#docker)
     docker:
       - image: quay.io/influxdb/rust:ci
       - image: localstack/localstack
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,6 +170,7 @@ jobs:
   test:
     docker:
       - image: quay.io/influxdb/rust:ci
+      - image: localstack/localstack
     resource_class: xlarge  # use of a smaller executor tends crashes on link
     environment:
       # Disable incremental compilation to avoid overhead. We are not preserving these files anyway.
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -182,18 +183,31 @@ jobs:
...
+      AWS_SECRET_ACCESS_KEY: test
+      AWS_ENDPOINT: http://127.0.0.1:4566
+      INFLUXDB_IOX_BUCKET: iox-test
     steps:
       - run:
           name: Run Kafka
           # Sudo needed because data directory owned by root but container runs as unprivileged user
           command: sudo /opt/redpanda/bin/redpanda --redpanda-cfg /etc/redpanda/redpanda.yaml --overprovisioned --smp 1 --memory 1G --reserve-memory 0M
           background: true
+      - run:
```
tustvold, line 197: I wonder if this might just be moved into an entrypoint script for `localstack/localstack`. The added benefit is you get `awslocal` that  handles the `--endpoint-url` for you

.circleci/config.yml
```diff
@@ -170,6 +170,7 @@ jobs:
   test:
     docker:
       - image: quay.io/influxdb/rust:ci
+      - image: localstack/localstack
```
tustvold, line 173: Maybe provide a link to https://circleci.com/docs/2.0/configuration-reference/#docker it took me a long time to find out what this was doing.

It occurs to me that this might be a nicer way to run redpanda :thinking:

tustvold (approved):

.circleci/config.yml
crepererum: Then we need our own localstack image which I wanted to avoid. The entrypoint is actually not that trivial as well, see <https://github.com/localstack/localstack/blob/master/bin/docker-entrypoint.sh>.

.circleci/config.yml
crepererum: added a comment

.circleci/config.yml
tustvold: I guess I was thinking using `bash -c` and some hackery :laughing:

PR #3390, opened 2021-12-16
refactor: use redpanda docker image

tustvold:
Something I saw @crepererum do in #3388 and thought was a nicer way to do this


.circleci/config.yml
```diff
@@ -170,6 +170,8 @@ jobs:
   test:
     docker:
       - image: quay.io/influxdb/rust:ci
+      - image: vectorized/redpanda:v21.9.2
```
tustvold, line 173: This is the more recent of the two images used in idpe

crepererum (approved):

.circleci/config.yml
alamb: 🎉  this is very cool

alamb (approved):

PR #3393, opened 2021-12-16
refactor: make object store prefixes segment based

### LINKED ISSUE #3226: Catalog Lists Transactions Outside Transactions Directory
**Problem**

Run IOx

```
cargo run -- -vvv run database --object-store file --data-dir ~/.iox/data/ --server-id 1000
```

Create a database

```
cargo run -- database create my_database
```

Stop IOx and navigate to the object store directory and rename the corresponding `transactions` directory to `transactions-old`

Start IOx, as IOx currently runs with wipe_catalog_on_error (#1522), this attempts to wipe the catalog and fails

```
2021-11-26T16:22:33.169850Z ERROR server::db::load: cannot load catalog, so wipe it: Error during protobuf IO: Error during store read operation: Object at location dbs/9d40a878-3257-4fb6-ad32-7e40c250035d/transactions/00000000000000003400/3c0ca095-4a83-4b2e-9eef-71d4dbb59d8b.ckpt not found: No such file or directory (os error 2)
2021-11-26T16:22:33.257229Z ERROR server::database: database in error state - operator intervention required db_name=my_database e=error loading catalog: Cannot wipe preserved catalog: Error during store write operation: File-based Object Store error: Unable to delete file /home/raphael/.iox/data/dbs/9d40a878-3257-4fb6-ad32-7e40c250035d/transactions/00000000000000000000/70a93309-0412-4a97-8dba-652a5b29a3bf.txn: No such file or directory (os error 2) state=CatalogLoadError
```

If you move transactions-old up a directory the server starts up without issue. If you now move transactions-old back you get a fork detected

```
2021-11-26T16:44:54.297788Z ERROR server::db::load: cannot load catalog, so wipe it: Fork detected. Revision 0 has two UUIDs 70a93309-0412-4a97-8dba-652a5b29a3bf and a17c9c5a-4127-41da-8278-f79cbf00d327. Maybe two writer instances with the same server ID were running in parallel?
2021-11-26T16:44:54.386319Z ERROR server::database: database in error state - operator intervention required db_name=my_database e=error loading catalog: Cannot wipe preserved catalog: Error during store write operation: File-based Object Store error: Unable to delete file /home/raphael/.iox/data/dbs/9d40a878-3257-4fb6-ad32-7e40c250035d/transactions/00000000000000000000/70a93309-0412-4a97-8dba-652a5b29a3bf.txn: No such file or directory (os error 2) state=CatalogLoadError
```

crepererum:
Don't use string prefixes, e.g. `foo/bar/` is a prefix of `foo/bar/x`
but NOT of `foo/bar_baz/y`.

This also removes some heuristics during the cloud storage parsing that
assumed that file names always contain a dot but directories don't.

Technically we should now always be able to know whether a path points
to a file or a directory:

- Rust (manually constructed): we use `DirsAndFileName` which knows the
  difference (i.e. if `file_name` is set)
- in-mem store: we also use `DirsAndFileName`
- file system: this was fixed by #1523
  (ccd094dfcfca6e27fec97835f233a86dadd54272 and 464667d8b8d571718392c0e1f970e8f9c4d44376)
- cloud: cloud doesn't know about directories. So all paths that these
  APIs return and that end with a `/` are directories (can only occur in
  `list_with_delimiter`); everyting else is a file

Path string representations are now acting occurdingly (i.e. always end
with an `/` if they point to a directory).

Fixes #3226.

### DIFF
--- a/object_store/src/aws.rs
+++ b/object_store/src/aws.rs
@@ -284,21 +284,28 @@ impl ObjectStoreApi for AmazonS3 {
         &'a self,
         prefix: Option<&'a Self::Path>,
     ) -> Result<BoxStream<'a, Result<Vec<Self::Path>>>> {
+        let prefix_is_dir = prefix.map(|path| path.is_dir()).unwrap_or(true);
+
         Ok(self
             .list_objects_v2(prefix, None)
             .await?
-            .map_ok(|list_objects_v2_result| {
+            .map_ok(move |list_objects_v2_result| {
                 let contents = list_objects_v2_result.contents.unwrap_or_default();
 
                 contents
                     .into_iter()
                     .flat_map(|object| object.key.map(CloudPath::raw))
+                    .filter(move |path| {
+                        prefix_is_dir || prefix.map(|prefix| prefix == path).unwrap_or_default()
+                    })
                     .collect()
             })
             .boxed())
     }
 
     async fn list_with_delimiter(&self, prefix: &Self::Path) -> Result<ListResult<Self::Path>> {
+        let prefix_is_dir = prefix.is_dir();
+
         Ok(self
             .list_objects_v2(Some(prefix), Some(DELIMITER.to_string()))
             .await?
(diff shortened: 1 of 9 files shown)


carols10cents (approved):
This looks good! I added a commit to fix a few small typos and comments that needed to be updated to say "not match" instead of "matches" to reflect the behavior change.

Locally, I ran the integration tests for s3, google, and azure that actually contact those services, and they all pass 🎉 

Just to document for future reference: Marco and I chatted about this and this makes the meaning of a "prefix match" in `object_store` different than what the cloud storage services mean, but because IOx doesn't use the partial directory matches ever, this change seems fine.

PR #3401, opened 2021-12-17
feat: use `TCP_USER_TIMEOUT` for rdkafka

crepererum:
Nothing to see here, just a tiny experiment.


alamb:
Given we have removed rdkafka, I think this PR is unlikely to ever be merged, so closing it down; Please reopen if that was a mistake

Issue #3372, opened 2021-12-14
Add support for TagValuesGroupedByMeasurementAndTagKeyRequest gRPC method


e-dard:
Internally we use the `TagValuesGroupedByMeasurementAndTagKeyRequest` RPC method to push down computation associated with `SHOW TAGS KEYS` meta queries. This functionality was added to TSM storage after we first forked over the RPC API to IOx. Therefore, IOx does not have the ability to handle this RPC  call.

This issue tracks adding support for `TagValuesGroupedByMeasurementAndTagKeyRequest`.

- [x] Add initial support for the RPC, returning an unimplemented error. #3373
- [x] Add support to handle `SHOW TAG VALUES FROM measurement WITH KEY = "column" WHERE some_predicate` https://github.com/influxdata/influxdb_iox/pull/3406
- [x] Add support to handle querying across multiple measurements (tables) using regex matches https://github.com/influxdata/influxdb_iox/pull/3406
- [x] Add  support to handle `WITH KEY = some_predicate`. https://github.com/influxdata/influxdb_iox/pull/3408

PR #3406, opened 2021-12-17
feat: increase support for TagValuesGroupedByMeasurementAndTagKey

e-dard:
Towards #3372

This PR implements the functionality in IOx to support many instantiations of a `TagValuesGroupedByMeasurementAndTagKey` RPC call.

Specifically, this PR adds support for 

* specifying measurements as literals or regex matches;
* providing arbitrary predicates that at least one returned tag value must satisfy;
* specifying a tag key as a literal match.

The following types of InfluxQL meta-queries are covered:

```
SHOW TAG VALUES FROM measurement WITH KEY = 'column'
SHOW TAG VALUES FROM measurement WITH KEY = 'column' WHERE 'other_column' = 'foo'
SHOW TAG VALUES FROM measurement WITH KEY = 'column' WHERE 'other_column' ~= /^.*a$/'
SHOW TAG VALUES FROM measurement1,measurement2 WITH KEY = 'column' WHERE 'other_column' ~= /^.*a$/'
SHOW TAG VALUES FROM /^Foo.+/ WITH KEY = 'column' WHERE 'other_column' != 'bar'
```

### Testing

As well as the included unit tests I also ran the internal InfluxQL compatibility tests and determined that this PR reduces the number of failures from **106** to **90**.

### TODO

There is still no support for predicates on the tag key. For example, in InfluxQL you can do the following:

```
SHOW TAG VALUES FROM measurement WITH KEY ~= /Ab.+c$/ WHERE 'other_column' = 'foo'
```

I will likely tackle that in the same way I tackled the measurements with regexes - by pre-materialising matching tag keys. This seems reasonable given everything currently gets fully materialised anyway.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
@@ -2,7 +2,10 @@
 //! implemented in terms of the [`QueryDatabase`](query::QueryDatabase) and
 //! [`DatabaseStore`]
 
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{BTreeSet, HashMap},
+    sync::Arc,
+};
 
 use snafu::{OptionExt, ResultExt, Snafu};
 use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -877,6 +898,84 @@ where
...
+    req: TagValuesGroupedByMeasurementAndTagKeyRequest,
+    span_ctx: Option<SpanContext>,
+) -> Result<Vec<TagValuesResponse>, Error>
+where
+    D: QueryDatabase + ExecutionContextProvider + 'static,
+{
+    use generated_types::tag_key_predicate::Value;
+
+    // Extract the tag key string literal.
+    // TODO - currently only eq operation supported, as in `WITH KEY = 'foo'`.
```
e-dard, line 915: Extended support for patterns on the `WITH KEY` is the final part of https://github.com/influxdata/influxdb_iox/issues/3372

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -877,6 +898,84 @@ where
...
+        return NotYetImplemented {
+            operation: format!(
+                "tag key predicate operation not supported: {:?}",
+                tag_key_op
+            ),
+        }
+        .fail();
+    };
+
+    // Because we need to return tag values grouped by measurements and tag
```
e-dard, line 935: I think materialising up front is a reasonable decision here because everything gets materialised for all these meta query type plans anyway...

e-dard: (I do think we will have to do something about it in the future though)

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -877,6 +898,84 @@ where
...
+        db_name.clone(),
+        req.measurement_patterns,
+        span_ctx.clone(),
+    )
+    .await?;
+
+    let mut responses = vec![];
+    for name in measurements.into_iter() {
+        // get all the tag values associated with this measurement name and tag key
+        let values = tag_values_impl(
```
e-dard, line 949: For each measurement and tag key get the tag values. When I add support for multiple tag keys matching a pattern there will be an extra loop in here or I'll push the pattern down to the tag values implementation.

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1028,6 +1127,89 @@ where
     Ok(field_list)
 }
 
+/// Materialises a collection of measurement names. Typically used as part of
+/// a plan to scope and group multiple plans by measurement name.
+async fn materialise_measurement_names<D>(
```
e-dard, line 1132: This could probably live somewhere else but it's only called by one other function.

alamb: I suggest it goes into `src/query/frontend/influxrpc.rs` but if you did so you'll need to figure out how to share / convert `LiteralOrRegex`

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1028,6 +1126,89 @@ where
...
+                            },
+                            Node {
+                                node_type: Type::Literal as i32,
+                                children: vec![],
+                                value: Some(Value::RegexValue(pattern)),
+                            },
+                        ],
+                        value: Some(Value::Comparison(Comparison::Regex as i32)),
+                    };
+                    let resp = measurement_name_impl(
```
alamb, line 1188: nice

alamb (approved):
Looks very slick to me

It is neat to see you being able to re-use the existing calls to form the results 👍

e-dard:
@alamb will refactor `materialise_measurement_names` into the right place as a follow on as I have another PR building on this one.

PR #3408, opened 2021-12-20
feat: add support for predicates on SHOW TAG VALUES WITH KEY = pred

### LINKED ISSUE #3372: Add support for TagValuesGroupedByMeasurementAndTagKeyRequest gRPC method

e-dard:
Closes #3372

**Need to merge this first** https://github.com/influxdata/influxdb_iox/pull/3406 

This PR closes out the known outstanding `SHOW TAG VALUES` InfluxQL support using the newer `TagValuesGroupedByMeasurementAndTagKeyRequest` RPC request.

This PR adds support for applying predicates to the `WITH key ` clause in `SHOW TAG VALUES WITH KEY` statements.

The following operators are supported:

  - literal equality `WITH KEY = "some_column"`
  - literal inequality `WITH KEY != "some_column"`
  - regex match `WITH KEY =~ /foo.+/`
  - regex not match `WITH KEY !~ /foo.+/`
  - list match `WITH KEY IN ("col1", "col2", "col3")

I implemented several of the above by materialising the column names for a table and then applying the appropriate logic to filter away columns (tag keys) that don't match. Because everything has to be materialised currently to support this RPC I think that's OK. In the future we could improve on this by pushing down a plan to the system tables or similar, and let DF handle the application of these tag key predicates.

### Testing

As well as the included tests, this reduces the failures on our internal suite from `90` to `80`.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
@@ -913,8 +913,7 @@ async fn tag_values_grouped_by_measurement_and_tag_key_impl<D>(
 where
     D: QueryDatabase + ExecutionContextProvider + 'static,
 {
-    // Extract the tag key string literal.
-    // TODO - currently only eq operation supported, as in `WITH KEY = 'foo'`.
+    // Extract the tag key predicate.
     // See https://docs.influxdata.com/influxdb/v1.8/query_language/explore-schema/#show-tag-values
     // for more details.
     let tag_key_pred = req
@@ -1131,9 +1130,6 @@ where
 
 /// Materialises a collection of measurement names. Typically used as part of
 /// a plan to scope and group multiple plans by measurement name.
-///
-/// TODO(edd): this might be better represented as a plan against the `tables`
-/// system table.
 async fn materialise_measurement_names<D>(
     db: Arc<D>,
     db_name: DatabaseName<'static>,
--- a/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
+++ b/influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
@@ -15,13 +15,14 @@ use tonic::Status;
 use data_types::{error::ErrorLogger, names::org_and_bucket_to_database, DatabaseName};
 use generated_types::{
     google::protobuf::Empty, literal_or_regex::Value as RegexOrLiteralValue,
-    offsets_response::PartitionOffsetResponse, storage_server::Storage, CapabilitiesResponse,
-    Capability, Int64ValuesResponse, LiteralOrRegex, MeasurementFieldsRequest,
-    MeasurementFieldsResponse, MeasurementNamesRequest, MeasurementTagKeysRequest,
-    MeasurementTagValuesRequest, OffsetsResponse, Predicate, ReadFilterRequest, ReadGroupRequest,
-    ReadResponse, ReadSeriesCardinalityRequest, ReadWindowAggregateRequest, StringValuesResponse,
-    TagKeyMetaNames, TagKeysRequest, TagValuesGroupedByMeasurementAndTagKeyRequest,
-    TagValuesRequest, TagValuesResponse, TimestampRange,
+    offsets_response::PartitionOffsetResponse, storage_server::Storage, tag_key_predicate,
+    CapabilitiesResponse, Capability, Int64ValuesResponse, LiteralOrRegex,
+    MeasurementFieldsRequest, MeasurementFieldsResponse, MeasurementNamesRequest,
+    MeasurementTagKeysRequest, MeasurementTagValuesRequest, OffsetsResponse, Predicate,
+    ReadFilterRequest, ReadGroupRequest, ReadResponse, ReadSeriesCardinalityRequest,
+    ReadWindowAggregateRequest, StringValuesResponse, TagKeyMetaNames, TagKeysRequest,
+    TagValuesGroupedByMeasurementAndTagKeyRequest, TagValuesRequest, TagValuesResponse,
+    TimestampRange,
 };
 use observability_deps::tracing::{error, info, trace};
 use predicate::predicate::PredicateBuilder;
@@ -48,7 +49,7 @@ use crate::influxdb_ioxd::{
 };
 use trace::ctx::SpanContext;
 
-use super::TAG_KEY_MEASUREMENT;
+use super::{TAG_KEY_FIELD, TAG_KEY_MEASUREMENT};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
@@ -156,6 +157,9 @@ pub enum Error {
     #[snafu(display("Missing tag key predicate"))]
     MissingTagKeyPredicate {},
 
+    #[snafu(display("Tag Key regex error: {}", source))]
+    InvalidTagKeyRegex { source: regex::Error },
+
     #[snafu(display("Error sending results via channel:  {}", source))]
     SendingResults {
         source: Box<dyn std::error::Error + Send + Sync>,
@@ -214,6 +218,7 @@ impl Error {
             Self::NotYetImplemented { .. } => Status::internal(self.to_string()),
             Self::MeasurementLiteralOrRegex { .. } => Status::invalid_argument(self.to_string()),
             Self::MissingTagKeyPredicate {} => Status::invalid_argument(self.to_string()),
+            Self::InvalidTagKeyRegex { .. } => Status::invalid_argument(self.to_string()),
         }
     }
 }
@@ -908,28 +913,15 @@ async fn tag_values_grouped_by_measurement_and_tag_key_impl<D>(
 where
     D: QueryDatabase + ExecutionContextProvider + 'static,
 {
-    use generated_types::tag_key_predicate::Value;
-
     // Extract the tag key string literal.
     // TODO - currently only eq operation supported, as in `WITH KEY = 'foo'`.
     // See https://docs.influxdata.com/influxdb/v1.8/query_language/explore-schema/#show-tag-values
     // for more details.
-    let tag_key_op = req
+    let tag_key_pred = req
         .tag_key_predicate
         .context(MissingTagKeyPredicate {})?
         .value
         .context(MissingTagKeyPredicate {})?;
-    let tag_key_lit = if let Value::Eq(lit) = tag_key_op {
-        lit
-    } else {
-        return NotYetImplemented {
-            operation: format!(
-                "tag key predicate operation not supported: {:?}",
-                tag_key_op
-            ),
-        }
-        .fail();
-    };
 
     // Because we need to return tag values grouped by measurements and tag
     // keys we will materialise the measurements up front, so we can build up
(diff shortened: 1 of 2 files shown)


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -908,28 +913,15 @@ async fn tag_values_grouped_by_measurement_and_tag_key_impl<D>(
 where
     D: QueryDatabase + ExecutionContextProvider + 'static,
 {
-    use generated_types::tag_key_predicate::Value;
-
     // Extract the tag key string literal.
     // TODO - currently only eq operation supported, as in `WITH KEY = 'foo'`.
```
alamb, line 917: this comment seems like it needs an update

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -944,32 +936,43 @@ where
...
+        .await?;
 
-        responses.push(TagValuesResponse {
-            measurement: name,
-            key: tag_key_lit.clone(),
-            values,
-        });
+        for key in tag_keys {
+            // get all the tag values associated with this measurement name and tag key
+            let values = tag_values_impl(
```
alamb, line 950: Like `materialize_measurement_names` this code logically might better belong in `query/src/frontend/influxrpc.rs`, though can be moved there as a follow on PR

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1128,6 +1131,9 @@ where
 
 /// Materialises a collection of measurement names. Typically used as part of
 /// a plan to scope and group multiple plans by measurement name.
+///
+/// TODO(edd): this might be better represented as a plan against the `tables`
```
alamb, line 1135: I think https://github.com/influxdata/influxdb_iox/blob/main/query/src/frontend/influxrpc.rs#L239 (`table_names_plan` is likely as efficient as you are going to get) 

Depending on the predicate `system.tables` may not have the correct information

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1209,6 +1215,70 @@ where
...
+    span_ctx: Option<SpanContext>,
+) -> Result<BTreeSet<String>, Error>
+where
+    D: QueryDatabase + ExecutionContextProvider + 'static,
+{
+    use generated_types::tag_key_predicate::Value;
+
+    if let Value::Eq(elem) = tag_key_predicate {
+        // If the predicate is a simple literal match then return that value
+        // regardless of whether or not the tag key exists.
```
alamb, line 1236: that is a fascinating semantic -- I assume it is what InfluxQL expects -- it is not "intuitive" to me

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1209,6 +1215,70 @@ where
...
+        return Ok(vec![elem].into_iter().collect::<BTreeSet<_>>());
+    } else if let Value::In(elem) = tag_key_predicate {
+        // If the predicate is a list of literal matches then return those
+        // regardless of whether or not those tag keys exist.
+        return Ok(elem.vals.into_iter().collect::<BTreeSet<_>>());
+    }
+
+    // Otherwise materialise the tag keys for this measurement and filter out
+    // any that don't pass the provided tag key predicate.
+    let mut tag_keys = tag_keys_impl(
```
alamb, line 1246: Given you can pass predicates to `tag_keys_impl` now, I wonder if it would be possible to simple convert `tag_key_preciate` into a `Predicate` and let `tag_keys_impl` apply the values

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1209,6 +1215,70 @@ where
...
+        Value::Neq(value) => tag_keys.retain(|elem| elem != &value),
+        Value::EqRegex(pattern) => {
+            let re = regex::Regex::new(&pattern).context(InvalidTagKeyRegex)?;
+            tag_keys.retain(|elem| re.is_match(elem));
+        }
+        Value::NeqRegex(pattern) => {
+            let re = regex::Regex::new(&pattern).context(InvalidTagKeyRegex)?;
+            tag_keys.retain(|elem| !re.is_match(elem));
+        }
+        x => unreachable!("predicate should have been handled already {:?}", x),
```
alamb, line 1276: I do think this kind of predicate might all already be handled in `tag_keys_impl`

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1967,6 +2037,71 @@ mod tests {
...
+                }),
+                Some(make_state_geq_ma_predicate()),
+                vec![TagValuesResponse {
+                    measurement: "table_b".into(),
+                    key: "state".into(),
+                    values: vec!["MA".into()],
+                }],
+            ),
+            (
+                "SHOW TAG VALUES FROM 'table_a' WITH KEY in (\"state\", \"foo\")",
```
alamb, line 2091: maybe should we have a test where the predicate is the literal match?

aka is there coverage of

```rust
    if let Value::Eq(elem) = tag_key_predicate {
        // If the predicate is a simple literal match then return that value
        // regardless of whether or not the tag key exists.
```?

**Review by alamb — APPROVED** [review]
I reviewed https://github.com/influxdata/influxdb_iox/pull/3408/commits/c4a3640b8be62a90be020bcbb5ed38c90d7e1a5b as I think the others are included as part of https://github.com/influxdata/influxdb_iox/pull/3406

I saw a few "TODO" or "make a query in terms of system table" type comments.

In terms of follow on work / optimizing this implementation, if it turns out to be needed, would be to implement a custom plan in `InfluxRPCPlanner` following the model of the existing requests. I don't think there is any particular need to make these changes now, but it might be worth considering

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 4 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:1236]: Ah I wonder if the comment has confused you. This comment means: if the user has issues the query

```
SHOW TAG VALUES WITH key = "my_key"
```

Then attempt to find all the tag values for the column `my_key` across all tables, even if the `my_key` column does not exist on those tables. It's an optimisation to avoid materialising all column names (tag keys) in this function only to throw them all away.

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 5 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:1246]: I didn't think that `tag_keys_impl` knew how to apply predicates on projections? I would be surprised if this had been implemented in IOx because we only taught IOx how to handle `_measurement` and `_field`, but not something like `_tag_key`. To be clear - the `tag_keys_predicate` allows the user to basically say "find me all distinct values for any column that matches this pattern".

Like:

```
SHOW TAG VALUES WITH KEY =~ /Foo.*/
```

would return tag values for columns: `{Foobar, Foozoo, Foofoo}` etc.

Does `_field` work against none field columns? 🤔 Maybe that could do it?

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 6 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:1276]: This line exists because the predicate was handled right at the top of the function (`Value::Eq` and `Value::In`). No need to inefficiently materialise all tag keys if those two predicate types were used.

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 7 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:2091]: Already covered in previous PR, e.g., line 1997

https://github.com/influxdata/influxdb_iox/pull/3408/files/5b277c875427e451e57ad10945b7e8ee327e3e61..c4a3640b8be62a90be020bcbb5ed38c90d7e1a5b#diff-420c0b3ff21cbbddbc0941f124e784caaf4c5aac7a81d15da08a19544114858eR1997

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 2 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:950]: will be moving both in a follow on

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 6 above)
Comment (alamb) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:1272]: I guess I was thinking that if `tag_keys_impl` didn't efficiently handle those kinds of predicates, it could be extended to do so. `tag_keys_impl` has a bunch of code to try and evaluate predicates using metadata only rather than running a full plan. 

I am thinking in particular of https://github.com/influxdata/influxdb_iox/blob/a9db30cdd08bb1efd45a8c32b0b30cac2819b672/query/src/frontend/influxrpc.rs#L363-L370

File: influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs (hunk 5 above)
Comment (e-dard) [reply influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs:1242]: > What happens if there are other predicates in the SHOW TAG VALUES query (is that even possible)?
>
> Like 
```
SHOW TAG VALUES WITH KEY =~ /Foo.*/ AND time > 2w
```

Yes that's possible and the previous PR #3406 took care of that support. The syntax is:

```
SHOW TAG VALUES WITH KEY =~ /Foo.*/ WHERE time > 2w
```

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
e-dard: That predicate (as far as I understand it) is a predicate for determining which column names to return based on applying a predicate to all values within said columns. If any rows match then return the column name.

This predicates above though are basically just filters on columns names, that is - they don't care what values are in the column they just care whether the column name matches the filter. I guess I'm not sure what value there is in putting those into the planner so to speak, because they're only used in this single situation (`SHOW TAG VALUES WITH KEY _SOME_FILTER_ WHERE....`). And also only used by InfluxQL.

alamb: >  I guess I'm not sure what value there is in putting those into the planner so to speak

The value I was hypostulating was to make flux queries with predicates on `_field` faster

I think this is enough of a niche usecase it isn't worth worrying about until we get additional evidence that it is

e-dard:
@alamb merging this so it's not hanging around on the pile into the new year. I suspect we will probably come back to some of the logic in the RPC service and maybe need to extend or move things around a bit as we get further through the outstanding InfluxQL defects.

PR #3411, opened 2021-12-21
docs: IOx Transactions & Locks

NGA-TRAN:

### DIFF
--- a/docs/transactions_locks.md
+++ b/docs/transactions_locks.md
@@ -1,35 +1,43 @@
 # IOx Transactions and Locks
 
-To keep data consistent, IOx needs to lock objects that are being modified and group actions that need to either all succeed or all fail into a transaction. This document describes types of locks and transactions that IOx supports. As illustrated in two previous documents, [Data Organization and LifeCycle](data_organization_lifecycle.md) and [Catalog](catalogs.md), IOx separates handling `Physical Data Chunks` from its metadata known as `Catalog`. Since `Catalog` is the core information used to operate the database and points to physical `Data Chunks`, most of the locking  and transaction activities will happen on `Catalog Objects`. However, before digging into Catalog's transactions and locks, let us see how `Physical Data Chunks` are read and written.
+In order to keep data consistent, IOx needs to **lock** objects that are being read or modified, and group actions that need to either all succeed or all fail into a **transaction**. This document describes types of locks and transactions that IOx supports. As illustrated in two previous documents, [Data Organization and LifeCycle](data_organization_lifecycle.md) and [Catalog](catalogs.md), IOx separates handling `Physical Data Chunks` from its metadata known as `Catalog`. Since `Catalog` is the core information used to operate the database and points to physical `Data Chunks`, most of the locking  and transaction activities will happen on `Catalog Objects`. However, before digging into Catalog's transactions and locks, let us see how `Physical Data Chunks` are read and written.
 
 ## Read and Write Physical Data Chunks
-As described in [IOx Data Organization and LifeCycle](data_organization_lifecycle.md), there are four types of `Data Chunks` in IOx: `O-MUB`, `F-MUB`, `RUB`, and `OS` in which only `O-MUB` is mutable, the others are immutable. So when a write is issued, it will go to its corresponding `O-MUB`, but when a read occurs, it may need to read all types of chunks that may contain needed data[^prune].
+As described in [IOx Data Organization and LifeCycle](data_organization_lifecycle.md), there are four types of `Data Chunks` in IOx: `O-MUB`, `F-MUB`, `RUB`, and `OS` in which only `O-MUB` is mutable, the others are immutable. When a write is issued, it will go to its corresponding `O-MUB`, but when a read occurs, it may need to read all types of chunks that may contain needed data[^prune].
 
 [^prune]: IOx supports chunk pruning to eliminate chunks that do not contains needed data but it is beyond the scope of this document.
 
-* **Read a `F-MUB`, a `RUB` or a `OS`:** When an immutable data chunk is identified to be read, its reference will be returned without locking because it never gets modified. If IOx wants to remove this chunk, it will wait until no references to the chunk before doing so. Note that the `Catalog Objects` described in Figure 1 of [IOx Catalog](catalogs.md)  such as `Table`, `Partition`, and `Chunk` that lead IOx to this data chunk  will be locked for `read` temporarily while searching for data chunks to ensure the data chunk won't be removed by other concurrent actions. See next section for Catalog locking.
+* **Read a `F-MUB`, a `RUB` or a `OS`:** When an immutable data chunk is identified to be read, its reference will be returned without locking because it never gets modified. If IOx wants to remove this chunk, it will wait until no references to the chunk before doing so. Next section will describe how the catalog objects that lead IOx to this data chunk is locked.
 
-* **Read and Write `O-MUB`:** Unlike immutable data chunks, when a mutable data chunk is identified to be read, a snapshot of its data at that moment will be returned and the query will be run on that snapshot which is isolated from the chunk. This enables IOx to continue running queries if the chunk is then modified to add ingesting data. However, if a read comes while a write is happening, it has to wait to get the next snapshot. IOx plans to improve this limitation in the future by returning the previous snapshot for running the query without waiting for a write to finish.
+* **Read and Write `O-MUB`:** Unlike immutable data chunks, when a mutable data chunk is identified to be read, a snapshot of its data at that moment will be returned and the query will be run on that snapshot which is isolated from the chunk. This enables IOx to continue running queries if the chunk is then modified to add ingesting data. However, if a read comes while a write is happening, it has to wait to get the next snapshot. See next section for the detail of catalog locking and transaction on writes.
+
 
 ## Catalog Transactions and Locks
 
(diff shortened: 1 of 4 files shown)


docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+In order to query a table, we need through these steps:
+1. Step 1: Identifying Data Chunks of the table and take their snapshots.
+1. Step 2: Pruning unnecessary Chunks
+1. Step 3: Building a query plan for those Chunks
+1. Step 4: Executing the plan
+
+Since querying only involves reading activities, no transactions are needed. As pointed out above, locks no longer needed when snapshots or references of the Data Chunk are identified, only the first step acquires locks and the subject of discussion of this document.
+
+**Step 1: Identify Data Chunks of the tables and take their snapshots** are implemented in [filtered_chunks](https://github.com/influxdata/influxdb_iox/blob/8a2410e161996603a4147e319c71e5bb38ca9cb7/server/src/db/catalog.rs#L288)[^tables] and as follows:
+  1. <span style="color:red">Acquire</span> `S` lock on all Catalog Tables.
```
NGA-TRAN, line 88: I am not sure why this color does not show up when I am on `display in rich diff`

NGA-TRAN (commented):
I have to say that there is so much information to make the document useful. This document is so long but I am not sure how to shorten it but still convey how locks and transactions in IOX work.

I hope I will get a lot feedback to make it better and easier to read.

server/src/db/catalog.rs
```diff
@@ -310,6 +310,7 @@ impl Catalog {
             }
             None => itertools::Either::Right(table.partitions()),
         });
+        // todo: table lock should be released here
```
alamb, line 313: I think we would also have to review callers to make sure they don't rely on the table being locked / not changing. But I can't imagine that they would

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Read and Write Physical Data Chunks
+As described in [IOx Data Organization and LifeCycle](data_organization_lifecycle.md), there are four types of `Data Chunks` in IOx: `O-MUB`, `F-MUB`, `RUB`, and `OS` in which only `O-MUB` is mutable, the others are immutable. When a write is issued, it will go to its corresponding `O-MUB`, but when a read occurs, it may need to read all types of chunks that may contain needed data[^prune].
+
+[^prune]: IOx supports chunk pruning to eliminate chunks that do not contains needed data but it is beyond the scope of this document.
```
alamb, line 8:
```suggestion
[^prune]: IOx supports chunk pruning during query planning to eliminate chunks that do not contains needed data but it is beyond the scope of this document.
```

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+
+* **Read a `F-MUB`, a `RUB` or a `OS`:** When an immutable data chunk is identified to be read, its reference will be returned without locking because it never gets modified. If IOx wants to remove this chunk, it will wait until no references to the chunk before doing so. Next section will describe how the catalog objects that lead IOx to this data chunk is locked.
```
alamb, line 10:
```suggestion
* **Read a `F-MUB`, a `RUB` or a `OS`:** When an immutable data chunk is identified to be read, its reference will be returned without locking because it never gets modified. If IOx wants to remove this chunk, it will wait until no references to the chunk exist before doing so. Next section will describe how the catalog objects that lead IOx to this data chunk is locked.
```

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+    ───────────────────────────────────────|────────|────────
+      Possible to acquire read lock (S)?   |  yes   |  no
+    ───────────────────────────────────────|────────|────────
+      Possible to acquire write lock (X)?  |  no    |  no
+
+    
+Table1 1: Common `Shared` and `Exclusive` Locks and Their Compatibility
+```
+
+IOx supports one more lock in the middle, `SX`, as shown in Table 2 and implemented in [Freezable](https://github.com/influxdata/influxdb_iox/blob/fa47fb5582cb7527817a8c2834b82b5eb604ad46/internal_types/src/freezable.rs). `SX` is a lock that is first acquired for `read` but will be upgraded to `write` at some point. This allows other concurrent threads to read the `SX`-acquired object before it is upgraded to `write` but won't be able to acquire `SX` or `X` on that object. This means while `Thread 1` is holding `SX` on object `T`, other threads cannot modify T; and no matter how many threads are holding `S` lock on `T` while `Thread 1` is holding `SX` on it, `Thread 1` is still able to upgrade its lock to `X` for modifying `T`.
```
alamb, line 50: ❤️

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+
+```text
+
+                     \           Thread 1 ->           |     S               | SX (read-then-write) | X (write)
+         Thread 2 |   \                                |  (read)             | (read-then-write)    |
+                  V    \                               |                     |                      |
+    ───────────────────────────────────────────────────|─────────────────────|──────────────────────|───────────
+      Possible to acquire read lock (S)?               |     yes             |        yes           |   no
+    ───────────────────────────────────────────────────|─────────────────────|──────────────────────|───────────
+      Possible to acquire read-then-write lock (SX) ?  |     yes?            |        no            |   no
```
alamb, line 66: I think it is possible to get a `SX` lock (Freeze handle) when there are other outstanding read locks so 👍  @tustvold  should confirm.

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+Let us go over a few common actions in IOx Data Lifecycle to explore when locks and transactions are used.
+
+### Example 1: Query Data of a Table, T1
+In order to query a table, we need through these steps:
+1. Step 1: Identifying Data Chunks of the table and take their snapshots.
+1. Step 2: Pruning unnecessary Chunks
+1. Step 3: Building a query plan for those Chunks
+1. Step 4: Executing the plan
+
+Since querying only involves reading activities, no transactions are needed. As pointed out above, locks no longer needed when snapshots or references of the Data Chunk are identified, only the first step acquires locks and the subject of discussion of this document.
```
alamb, line 85:
```suggestion
Since querying only involves reading activities, no transactions are needed. As pointed out above, locks are no longer needed once snapshots or references of the Data Chunk are identified, so only the first step acquires locks and the subject of discussion of this document.
```

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+  1. Identify Catalog Chunks of `P1` which are, let say, `C1` and `C2`.
+  1. <span style="color:red">Acquire</span> `S` lock on `C1` and `C2`.
+  1. Take snapshots for `C1` and `C2` (snapshot can either be real snapshot for `O-MUB` or reference for `F-MUB`, `RUB` and `OS`)
+  1. <span style="color:green">Release all</span> `S` locks on Chunks (`C1` and `C2`), Partition (`P1`), and Tables.
+  1. Return the snapshots.
+
+[^tables]: The filtered_chunks function can return chunks of many tables but we only need chunks of one table in this example.
+
+### Example 2: Write to an O-MUB Chunk of Partition, P1, of Table, T1
+A write in IOx is implemented in [store_filtered_write](https://github.com/influxdata/influxdb_iox/blob/ccba68fe3ed3b41f06992cd1c11eefe720ed3ad4/server/src/db.rs#L1053) that can trigger to create a table or partitions as needed but in this example, we only focus on writing data to an available `O-MUB` chunk of Partition, `P1`, of Table, `T1`. Similar to reading, the lock acquisition and release are in the same order but on `X` lock instead.
```
alamb, line 100:
```suggestion
A write in IOx is implemented in [store_filtered_write](https://github.com/influxdata/influxdb_iox/blob/ccba68fe3ed3b41f06992cd1c11eefe720ed3ad4/server/src/db.rs#L1053) that can trigger the creation of a table or partitions as needed but in this example, we only focus on writing data to an available `O-MUB` chunk of Partition, `P1`, of Table, `T1`. Similar to reading, the lock acquisition and release are in the same order but on `X` lock instead.
```

alamb (approved):

docs/transactions_locks.md
```diff
@@ -0,0 +1,147 @@
...
+
+1. Step a: <span style="color:red">Acquire</span> `S` lock on **cleanup lock** to ensure no cleanup jobs can remove any parquet files.
+1. Step b: Write the compacted data to a parquet file, `PF`, in the object store.
+1. Step c: <span style="color:red">Open</span>  a Catalog Transaction, `CT`
```
tustvold, line 128: This acquires another lock, only a single catalog transaction can be being performed at once. 

As mentioned above, I would suggest splitting out details of preserved catalog transactions into their own document, and keep them separate from those on the in-memory catalog. Mixing the concepts together I think is a source of a lot of confusion

tustvold (commented):
I would suggest reducing the scope of this document to just be transactions on the in-memory catalog. Separate documents can then cover preserved-catalog transactions, the mechanics of persistence, the startup state machines, etc... and any locking necessary to achieve the desired semantics. (I'm happy to take a stab at writing some of these)

Trying to cover all uses of locks and transactions in IOx in a single document I think is going to be very long and confusing. For better or worse the design of IOx has independent subsystems glued together by lifecycle actions, with many of these subsystems providing their own transaction abstractions. Therefore trying to describe a single "transaction", or even a single "catalog", is either going to be inaccurate or very confusing.

To give a concrete example of this, `Freezable` is not actually used by the in-memory catalog at all - it is used as an implementation detail of the startup state machines and `PersistenceWindows`. The in-memory catalog only uses a combination of [RWLock](https://docs.rs/parking_lot/0.11.2/parking_lot/type.RwLock.html), and lifecycle actions recorded on chunks. Therefore the fact this document even mentions it suggests to me a scope issue...

NGA-TRAN:
Thanks @alamb and @tustvold: since we will have new catalog, I put this on hold for now.

alamb:
closing old PR to keep list clean -- please reopen if we still plan to work on this

I think this documentation change is likely not relevant as we switch to NG

PR #1293, opened 2021-04-23
fix: Make a ServerId type that's consistent everywhere

carols10cents:
Servers have an ID, which can only be set once per server instance and cannot be zero.

These IDs also get passed around with writes and replication and used in object storage.

The "set only once" constraint only needs to be enforced on the `Server` ID getter/setter methods, but the nonzero part should be enforced everywhere. There were a few places (in tests) that were using 0 for the concept of a server ID, which shouldn't be possible anywhere.

`ServerId` should be a separate type from `NonZeroU32` (rather than an alias) so that they're not interchangeable (and if we want to change to, say, a `NonZeroU64` inner type, we'll have one place to change it).

This does change the management service gRPC definitions, I think for the better, because some servers might never be writers but they still need an ID. So it makes more sense to set a server ID rather than a writer ID. We'll send a PR to conductor soon.

PR #3415, opened 2021-12-22
fix: Remove some straggling references to writer ID that should be server ID

carols10cents:
Was reminded of this today, just a small cleanup of places I missed renaming in https://github.com/influxdata/influxdb_iox/pull/1293.


alamb (approved):
LGTM!

PR #3417, opened 2021-12-23
refactor: Extract server::db into its own crate

### LINKED ISSUE #2821: Split Db out of Server
Once #2243 and #2242 are complete it should be possible to create a separate Db crate. 

Aside from reducing the size of the server crate, some crates such as `query_tests` only need `Db` and will not need to pull in the entirety of `Server`, the write buffer, etc... This will likely help with its compilation times

carols10cents:
Closes #2821.

`JobRegistry` also needed to be extracted to a crate so that both server and db could use it.


alamb (approved):
❤️  I love it @carols10cents  -- as you mentioned this is a good one to merge in while there are no other outstanding code changes to minimize conflicts

Thus since I think this is fairly non controversial I will merge it in ✅

tustvold:
Omg, what a great Christmas present! 🎉

PR #3418, opened 2022-01-01
docs: document ingest flow

tustvold:
Whilst reading #3411 I realised we don't actually have any high-level documentation of the ingest flow. I therefore wrote up something quick.

I personally think the flow is easier to understand as the composition of smaller parts, and so that is what I've tried to do here. As it happens this reflects the code, which also tries to separate the concerns across various subsystems. In my opinion, ingest is inherently sufficiently complex that trying to be both broad and deep is liable to just confuse.


docs/ingest.md
```diff
@@ -0,0 +1,117 @@
+# Ingest
+
+A user expects to be able to write/delete data from an IOx database with an API call to an IOx server. Critically the user expects:
+
+* An API response in a matter of seconds or less
+* A successful API call to eventually take effect
```
carols10cents, line 6: I might clarify what "take effect" means here-- queryable, in the WAL, and persisted? What else?

docs/ingest.md
```diff
@@ -0,0 +1,117 @@
...
+
+This document aims to provide a high-level overview of how a number of related, but independent, subsystems come together to achieve this functionality within IOx:
+
+* _In-Memory Catalog_ - metadata about the data contained within this server 
+* _Preserved Catalog_ - preserved metadata used to rebuild the in-memory catalog and start ingestion on restart
+* _Write Buffer_ - a write-ahead log for modifications that haven't reached the Preserved Catalog
```
carols10cents, line 12: Do you want to address that the write buffer is optional if you aren't concerned with durability? Or because IOx as a service will have a write buffer, is this document only focused on that case? I think either way you go is fine, but it might be a good caveat to make somewhere?

docs/ingest.md
```diff
@@ -0,0 +1,117 @@
...
+
+These two new chunks are then atomically swapped into the _In-Memory Catalog_ replacing the chunks from which they were sourced. 
+
+Any delete predicates that landed on the source chunks after the snapshots were taken, will be carried across to the two new chunks.
+
+_Much like compaction, this must process a contiguous sequence of chunks within the partition to ensure ordering is preserved_ 
+
+**Write data to parquet**
+
+The data to persist is now located in a single chunk in the _In-Memory Catalog_, and is written to a single parquet file in object storage.
```
carols10cents, line 78: Object storage is another aspect that is potentially optional if you're only interested in in-memory persistence, but IOx as a service will be using object storage

docs/ingest.md
```diff
@@ -0,0 +1,117 @@
...
+
+**Preserved Catalog**
+
+The parquet file is now written, but it needs to be recorded in the _Preserved Catalog_, along with the position reached in the _Write Buffer_. On restart this will enable IOx to skip over data that has already been persisted.
+
+It would be perfectly correct to simply record the min unpersisted positions in the _Write Buffer_ and on restart resume playback from there. However, as we persist data for different partitions separately, and out-of-order within a partition, this would likely result in re-ingesting lots of already persisted data. 
+
+As such a more [sophisticated approach](https://github.com/influxdata/influxdb_iox/blob/b39e01f7ba4f5d19f92862c5e87b90a40879a6c9/persistence_windows/src/checkpoint.rs) is used that more accurately describes what remains unpersisted.
```
carols10cents, line 86: Just thinking about maintaining these docs, maybe it would be a good idea to link to `main` rather than a commit SHA, and leave a note in `persistence_windows/src/checkpoint.rs` that if the file gets moved, these docs should be updated?

carols10cents (approved):
Looks good to me,  a few small suggestions you can take or leave!

docs/ingest.md
tustvold: Good point - added that caveat.

It does make me think of a somewhat strange topology where you have a write buffer and no object storage. Technically we allow this, the semantics are perhaps a little bit confused though :sweat_smile:

PR #3420, opened 2022-01-03
chore: lint markdown docs and use relative code references in docs

tustvold:
Currently our markdown docs reference code sections with links to explicit git shas, this is stable but potentially stale. As suggested by @carols10cents [here](https://github.com/influxdata/influxdb_iox/pull/3418#discussion_r777576481) it would be nicer to reference the files directly.

This PR therefore changes to do this, but also adds a simple script that checks for broken links in CI. I originally tried to write this in bash but it was painful, and this way it will be easy enough to add future lints if we so wish.


docs/catalogs.md
```diff
@@ -23,64 +23,64 @@ Since the catalog is the central information shared to all users of the database
 
 
 ```text
-                                                             ┌───────────┐
```
tustvold, line 26: My editor seems to like to remove trailing whitespace, I can disable but I'm inclined to think this is a good thing...

.circleci/config.yml
```diff
@@ -458,6 +468,7 @@ workflows:
       - lint
       - cargo_audit
       - protobuf-lint
+      - docs-lint
```
crepererum, line 471: this should also be added to the `requires` list of the `perf_image` job below

crepererum (approved):

Issue #3412, opened 2021-12-22
Query Performance characterization of parquet, read buffer, and compaction


alamb:
In order to characterize where IOx is now and help us prioritize where additional query performance work is worthwhile I would like to do the following analysis:

I also want to hone the workflow of such experiments rather than obsess about which particular query I picked

So the goal is to pick one query from the production logs, and run it against:
1. Parquet files as they were downloaded from production (default)
2. After loading all parquet into the read buffer (so the queries go against that rather than parquet)
2. Compact all parquet files into a single file per partition (using the API from @Nga Tran)
4. Compact all parquet files into a single file per partition and load them to the read buffer

alamb:
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-0bb25ec3-7fff-61b1-f9b9-67249fec00ff"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">Here are average query execution times for 6 customer queries that were mirrored to IOx in production</span><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">.</span></p><br /><div dir="ltr" style="margin-left:0pt;" align="left">

Experiment | avg_duration_ms | RSS (GB) | Object Store Size MB | Memory Size MB
-- | -- | -- | -- | --
Parquet | 874.73 | 0.98 | 423 | 0.4
ReadBuffer | 5.35 | 4.5 | 423 | 2592
Compacted Parquet | 51.42 | 0.6 | 146 | 0.14
Compacted ReadBuffer | 2.72 | 1.8 | 146 | 896

[...]

PR #3421, opened 2022-01-04
feat: Add `table_name` to `Partition` response in management client

alamb:
# Rationale:

I want to be able to call `management::Client::compact_object_store_chunks` as part of https://github.com/influxdata/influxdb_iox/issues/3412:

```rust
pub async fn compact_object_store_chunks(
    &mut self,
    db_name: impl Into<String> + Send,
    table_name: impl Into<String> + Send,
    partition_key: impl Into<String> + Send,
    chunk_ids: Vec<Bytes>
) -> Result<IoxOperation, Error>
...
```

However, to do so you I must supply the `table_name` which is not returned in the call to `list_partitions` or `get_partition`

# Changes:
1. Add `table_name` to `Partition` in management protobuf
2. Thread `PartitionAddr` around enough to get the info

# Aside
I can't help thinking the access of Partiton info / Chunk info could be more uniform but for now, this seems ok


data_types/src/partition_metadata.rs
```diff
@@ -59,13 +59,6 @@ impl PartitionSummary {
     }
 }
 
-/// Metadata and statistics for a Chunk *within* a partition
```
alamb, line 62: drive by cleanup -- this structure is not used

generated_types/protos/influxdata/iox/management/v1/partition.proto
```diff
@@ -9,4 +9,7 @@ option go_package = "github.com/influxdata/iox/management/v1";
 message Partition {
   // The partition key of this partition
   string key = 1;
+
```
alamb, line 12: The point of this PR is to add this field and fill it correctly in the management response

query_tests/src/scenarios.rs
```diff
@@ -676,12 +676,13 @@ impl DbSetup for TwoMeasurementsManyFieldsTwoChunks {
         ];
         write_lp(&db, &lp_lines.join("\n"));
 
-        assert_eq!(count_mutable_buffer_chunks(&db), 2);
-        assert_eq!(count_read_buffer_chunks(&db), 1);
+        assert_eq!(count_mutable_buffer_chunks(&db), 4);
```
alamb, line 679: This db is partitioned on `table_name` and there are two tables (each with 2 MUbs and a rub). Previously the code grouped on partition_key (which doesn't have the table name) so it was off by a factor of 2

carols10cents (approved):
LGTM!

PR #3422, opened 2022-01-04
refactor: Remove `Result` in QueryDatabase trait (none of the functions can fail)

alamb:
~built on https://github.com/influxdata/influxdb_iox/pull/3421, so draft until that is done~

This is a cleanup I noticed while working on https://github.com/influxdata/influxdb_iox/pull/3421

# Rationale
None of these methods can fail (any more) so no need to return `Result`

# Changes:
1. Remove fallible methods and associated `Error type


carols10cents (approved):
Love it 👍🏻

Issue #3403, opened 2021-12-17
Allow wipe catalog (`database recover wipe` in `ReplayError` and other states it makes sense


alamb:
__Use case:__
We have a database that is in `ReplayError` state on startup (either due to some bug in replay, kafka problem, or something else) and I want to be able to get the database back into a state where it is ingesting new data

I don't care about any data that is lost (because perhaps I will replay it from OOB backup)

__Proposal:__
Allow `database recover wipe --force` command in states where it makes sense

As @crepererum  notes, it doesn't make sense to wipe catalog in all the states (e.g. in `Known` because you don't have sufficient information), but states like `ReplayError` certainly seems ok

__Current behaviour:__
I was trying what I thought was the ☢️   option of wiping the catalog, but it turned out I can't do so with the CLI

```
$ influxdb_iox database recover wipe ab6ac6aa1fb6b962_0468f4ff1130467c --force
Error in partition subcommand: Error wiping preserved catalog: The system is not in a state required for the operation's execution: database (ab6ac6aa1fb6b962_0468f4ff1130467c) in invalid state (ReplayError) for transition (WipePreservedCatalog)
command terminated with exit code 1
```

__Desired behaviour:__
The command above to succeed and the database to ingest new data (and all old history is lost)

__Alternatives considered:__
Describe other solutions or features you considered.

PR #3424, opened 2022-01-06
refactor: Split database.rs into smaller modules

alamb:
re: https://github.com/influxdata/influxdb_iox/issues/3403

# Rationale:

Navigating a 2000+ line module was making my head hurt (and made it harder to navigate the code). Now is a good time to split this monster up as no one else is making concurrent modifications

# Changes:
This is not meant to have any functional changes, just code reorganziation

Split out major chunks of code into their own modules:
1.  State into --> `state.rs`
2. OwnerInfo code --> `owner.rs`
3. Database initialization and `DatabaseState*` enums init.rs that has DatabaseState


carols10cents (approved):
IT'S SO BEAUTIFUL!!!

PR #3425, opened 2022-01-06
feat: Allow wipe catalog  in `ReplayError` and `WriteBufferCreationError` states

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3403

# Rationale

When databases in our production environment are broken due to `ReplayError` and I am trying to apply a "hammer" to fix them, I can't apply that hammer in a few other error states (which can only be applied when the database is in the `CatalogLoadError` state)

# Changes:
Allow wiping the catalog in the various states

Follow the model of rebuild preserved catalog


carols10cents (approved):
LGTM!

alamb:
cc @tustvold  / @crepererum  -- if you have a chance to double check the logic here that would be great. I will address any comments as a follow on PR.

Thank you @carols10cents ❤️

PR #3427, opened 2022-01-07
refactor: Clean up code by using `lit` and `lit_timestamp_nano` 

alamb:
# Rationale
I believe it is easier to understand the intent of the code when the actual value is not buried in 3 levels of wrapper creation.

I saw this screen spew while working on gRPC code (probably from when I didn't know better)

# Changes
1. Replaces some uses of `Expr::Literal(ScalarValue(Some(...)))` with `lit` and `lit_timestamp_nano`


e-dard (approved):
LGTM

Issue #3428, opened 2022-01-07
InfluxQL: Predicates with `OR` and `_field` and `_measurement` get wrong answers


alamb:
Found while working on https://github.com/influxdata/influxdb_iox/pull/3427. Specifically, while reviewing code and comments on #2910 for handling `_field` predicates (see https://github.com/influxdata/influxdb_iox/pull/2910/files#r732650703).

I don't yet know if this bug will be detected in the InfluxQL test suite -- but I wanted to get it filed now sooner rather than later.

__Steps to reproduce:__

With this data:
```
h2o,state=MA,city=Boston temp=70.4 50
h2o,state=MA,city=Boston other_temp=70.4 250
h2o,state=CA,city=Boston other_temp=72.4 350
o2,state=MA,city=Boston temp=53.4,reading=51 50
o2,state=CA temp=79.0 300
```

Run a query with the following predicate, to should pick `other_temp` from `h20` and `temp` from `o2`:
```
(_field = 'other_temp' AND _measurement = 'h2o') OR (_field = 'temp' AND _measurement = 'o2')
```

I will add this tests case in a PR

__Expected behaviour:__

Expect the following 4 series:
```
Series tags={_measurement=h2o, city=Boston, state=CA, _field=other_temp}\n  FloatPoints timestamps: [350], values: [72.4]",
Series tags={_measurement=h2o, city=Boston, state=MA, _field=other_temp}\n  FloatPoints timestamps: [250], values: [70.4]",
Series tags={_measurement=o2, state=CA, _field=temp}\n  FloatPoints timestamps: [300], values: [79.0]",
Series tags={_measurement=o2, city=Boston, state=MA, _field=temp}\n  FloatPoints timestamps: [50], values: [53.4]",
```
__Actual behaviour:__

Returns 5 series -- it *also* (incorrectly returns) `temp` for h20

```
Series tags={_measurement=h2o, city=Boston, state=CA, _field=other_temp}\n  FloatPoints timestamps: [350], values: [72.4]",
Series tags={_measurement=h2o, city=Boston, state=MA, _field=other_temp}\n  FloatPoints timestamps: [250], values: [70.4]",
        // This series should not be here (_field = temp)
Series tags={_measurement=h2o, city=Boston, state=MA, _field=temp}\n  FloatPoints timestamps: [50, 100000], values: [70.4, 70.4]",
[...]

PR #3429, opened 2022-01-07
test: Reproducer for #3428

alamb:
Add some test for influxrpc queries demonstrating incorrect results due to #3428


e-dard:
These tests seem to be passing according to CI. Do you still want this reviewed and merged?

alamb:
> These tests seem to be passing according to CI. Do you still want this reviewed and merged?


Yes, please @e-dard  -- the tests are passing bc the wrong answers are checked in

I just figured I would write the test

e-dard:
Ah ok. I'll look through it.

An alternative way to do it would be to check-in the expected correct answers and then add the `#[ignore]` attribute, which should ensure CI passes.

alamb:
> An alternative way to do it would be to check-in the expected correct answers and then add the #[ignore] attribute, which should ensure CI passes.

that is fine with me too -- whichever you would prefer

e-dard (approved):

PR #3441, opened 2022-01-11
feat: Add initial iox_catalog skeleton

pauldix:
This isn't ready to merge because the tests won't pass without going through the setup and I haven't fixed warnings and any clippy errors. But it's here to have a look. I also need to update table and constraint names with whatever @NGA-TRAN puts together in her writeup.

Any ideas on how we should set up the tests for this? I want to keep it as simple as possible for now and actually use Postgres. I think that's fine to have everyone have it on their laptop. I'm not sure how to make that go in Circle CI. Maybe @domodwyer or @mkmik know from setting up Conductor CI?

### DIFF
--- a/iox_catalog/Cargo.toml
+++ b/iox_catalog/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "iox_catalog"
+version = "0.1.0"
+authors = ["Paul Dix <paul@pauldix.net>"]
+edition = "2021"
+
+[dependencies] # In alphabetical order
+async-trait = "0.1.42"
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
+futures = "0.3"
+observability_deps = { path = "../observability_deps" }
+snafu = "0.6.10"
+sqlx = { version = "0.5", features = [ "runtime-tokio-native-tls" , "postgres" ] }
+tokio = { version = "1.13", features = ["full", "io-util", "macros", "parking_lot", "rt-multi-thread", "time"] }
+influxdb_line_protocol = { path = "../influxdb_line_protocol" }
--- a/iox_catalog/Cargo.toml
+++ b/iox_catalog/Cargo.toml
@@ -9,7 +9,7 @@ async-trait = "0.1.42"
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 futures = "0.3"
 observability_deps = { path = "../observability_deps" }
-snafu = "0.6.10"
+snafu = "0.7"
 sqlx = { version = "0.5", features = [ "runtime-tokio-native-tls" , "postgres" ] }
 tokio = { version = "1.13", features = ["full", "io-util", "macros", "parking_lot", "rt-multi-thread", "time"] }
 influxdb_line_protocol = { path = "../influxdb_line_protocol" }
--- a/iox_catalog/Cargo.toml
+++ b/iox_catalog/Cargo.toml
@@ -13,3 +13,6 @@ snafu = "0.7"
 sqlx = { version = "0.5", features = [ "runtime-tokio-native-tls" , "postgres" ] }
 tokio = { version = "1.13", features = ["full", "io-util", "macros", "parking_lot", "rt-multi-thread", "time"] }
 influxdb_line_protocol = { path = "../influxdb_line_protocol" }
+
+[dev-dependencies] # In alphabetical order
+dotenv = "0.15.0"
--- a/iox_catalog/Cargo.toml
+++ b/iox_catalog/Cargo.toml
@@ -13,6 +13,7 @@ snafu = "0.7"
 sqlx = { version = "0.5", features = [ "runtime-tokio-native-tls" , "postgres" ] }
 tokio = { version = "1.13", features = ["full", "io-util", "macros", "parking_lot", "rt-multi-thread", "time"] }
 influxdb_line_protocol = { path = "../influxdb_line_protocol" }
+workspace-hack = { path = "../workspace-hack"}
 
 [dev-dependencies] # In alphabetical order
 dotenv = "0.15.0"
--- a/iox_catalog/README.md
+++ b/iox_catalog/README.md
@@ -1,0 +1,23 @@
+# IOx Catalog
+This crate contains the code for the IOx Catalog. This includes the definitions of namespaces, their tables, 
+the columns of those tables and their types, what Parquet files are in object storage and delete tombstones. 
+There's also some configuration information that the overal distributed system uses for operation.
+
+To run this crate's tests you'll need Postgres installed and running locally. You'll also need to set the 
+`DATABASE_URL` environment variable so that sqlx will be able to connect to your local DB. For example with 
+user and password filled in:
+
+```
+DATABASE_URL=postgres://<postgres user>:<postgres password>@localhost/iox_shared
+```
+
+You'll then need to create the database and run the migrations. You can do this via the sqlx command line.
+
+```
+cargo install sqlx-cli
+sqlx database create
+sqlx migrate run
+```
+
+This will set up the database based on the files in `./migrations` in this crate. SQLx also creates a table 
+to keep track of which migrations have been run.
--- a/iox_catalog/README.md
+++ b/iox_catalog/README.md
@@ -1,10 +1,10 @@
 # IOx Catalog
-This crate contains the code for the IOx Catalog. This includes the definitions of namespaces, their tables, 
-the columns of those tables and their types, what Parquet files are in object storage and delete tombstones. 
+This crate contains the code for the IOx Catalog. This includes the definitions of namespaces, their tables,
+the columns of those tables and their types, what Parquet files are in object storage and delete tombstones.
 There's also some configuration information that the overal distributed system uses for operation.
 
-To run this crate's tests you'll need Postgres installed and running locally. You'll also need to set the 
-`DATABASE_URL` environment variable so that sqlx will be able to connect to your local DB. For example with 
+To run this crate's tests you'll need Postgres installed and running locally. You'll also need to set the
+`DATABASE_URL` environment variable so that sqlx will be able to connect to your local DB. For example with
 user and password filled in:
 
 ```
@@ -19,5 +19,15 @@ sqlx database create
 sqlx migrate run
 ```
 
-This will set up the database based on the files in `./migrations` in this crate. SQLx also creates a table 
+This will set up the database based on the files in `./migrations` in this crate. SQLx also creates a table
 to keep track of which migrations have been run.
+
+## Tests
+
+To run the Postgres integration tests, ensure the above setup is complete first.
+
+* Set `DATABASE_URL=<dsn>` env (see above)
+* Set `TEST_INTEGRATION=1`
+* Run `cargo test`
+
+**CAUTION:** existing data in the database is dropped when tests are run
--- a/iox_catalog/migrations/20211229171744_initial_schema.sql
+++ b/iox_catalog/migrations/20211229171744_initial_schema.sql
@@ -56,7 +56,8 @@ CREATE TABLE IF NOT EXISTS iox_catalog.sequencer
     kafka_topic_id INT NOT NULL,
     kafka_partition INT NOT NULL,
     min_unpersisted_sequence_number BIGINT,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    CONSTRAINT sequencer_unique UNIQUE (kafka_topic_id, kafka_partition)
     );
 
 CREATE TABLE IF NOT EXISTS iox_catalog.sharding_rule_override
--- a/iox_catalog/migrations/20211229171744_initial_schema.sql
+++ b/iox_catalog/migrations/20211229171744_initial_schema.sql
@@ -2,7 +2,9 @@
 -- iox_shared schema
 BEGIN;
 
-CREATE TABLE IF NOT EXISTS public.kafka_topics
+CREATE SCHEMA IF NOT EXISTS iox_catalog;
+
+CREATE TABLE IF NOT EXISTS iox_catalog.kafka_topic
 (
     id INT GENERATED ALWAYS AS IDENTITY,
     name VARCHAR NOT NULL,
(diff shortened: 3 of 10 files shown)


iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+            .unwrap();
+        let query_pool = QueryPool::create_or_get(SHARED_QUERY_POOL, "foo", &pool)
+            .await
+            .unwrap();
+
+        (pool, kafka_topic, query_pool)
+    }
+
+    #[tokio::test]
+    async fn test_catalog() {
```
pauldix, line 736: I wasn't sure what the best way to have a set of tests that each require a fresh db setup to run would look like. They'd have to run serially or something? So I just put everything in this single large test. Just making sure everything works for now, it's not exhaustive.

mkmik: look at this: https://github.com/influxdata/conductor/blob/873ece9b33c6a0f9bfab72a76c8dbc0bbe2d1d8e/lib/util/src/postgres/test_utils.rs#L30

it creates a random schema for every test and sets the search path in the after_connect hook; this way each test can run in parallel

iox_catalog/README.md
```diff
@@ -0,0 +1,23 @@
...
+You'll then need to create the database and run the migrations. You can do this via the sqlx command line.
+
+```
+cargo install sqlx-cli
+sqlx database create
+sqlx migrate run
+```
+
+This will set up the database based on the files in `./migrations` in this crate. SQLx also creates a table 
+to keep track of which migrations have been run.
```
NGA-TRAN, line 23: I will install Postgres and follow this instructions

iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -0,0 +1,241 @@
...
+    parquet_file_id BIGINT NOT NULL,
+    PRIMARY KEY (delete_tombstone_id, parquet_file_id)
+    );
+
+ALTER TABLE IF EXISTS public.namespaces
+    ADD FOREIGN KEY (kafka_topic_id)
+    REFERENCES public.kafka_topics (id) MATCH SIMPLE
+    ON UPDATE NO ACTION
+       ON DELETE NO ACTION
+	NOT VALID;
```
NGA-TRAN, line 121: I think I will be able to replace each `ALTER TABLE` here with a line or two inside the `CREATE TABLE` to make the DML SQL simpler. I will give it a try when this PR is merged

NGA-TRAN (commented):
I have skimmed this and it looks good to me. I have some idea what it does and how to work with Postgres. I let @domodwyer, @mkmik or @crepererum to help you with the setup. I think I have enough information to start working on a checklist and opening tickets.

iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+        Ok(rec)
+    }
+}
+
+#[derive(Debug, sqlx::FromRow, Eq, PartialEq)]
+pub struct Column {
+    id: i32,
+    table_id: i32,
+    name: String,
+    data_type: i16,
```
crepererum, line 253: Shouldn't this be `ColumnType`?

iox_catalog/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+    clippy::future_not_send,
+    clippy::use_self,
+    clippy::clone_on_ref_ptr
+)]
+
+const SHARED_KAFKA_TOPIC: &str = "iox_shared";
+const SHARED_QUERY_POOL: &str = SHARED_KAFKA_TOPIC;
+const TIME_COLUMN: &str = "time";
+
+pub mod postgres;
```
crepererum, line 18: Could we split this up differently? (could also be in a follow-up PR)

I think the API structs that catalog users will use via Rust (e.g. `KafkaTopic`) should live in one place, like `mod interface`, and because there are quite a few structs it would be nice if they are a bit grouped instead of all in one giant module. Luckily these objects don't hold any IO state (e.g. a query pool or transaction like it is common for ORMs)

The actual IO interface (e.g. Postgres) should probably be abstracted away. A pattern that works for this is what's normally called a "repository", e.g.:

```rust
// one for "postgres", "mock", "file", ...
trait KafkaTopicRepo {
    async fn create_or_get(&self, name: &str) -> Result<KafkaTopic>;
}

// one for "postgres", "mock", "file", ...
trait NamespaceRepo {
    async fn create(
        &self,
        name: &str,
        retention_duration: &str,
        kafka_topic_id: i32,
        query_pool_id: i16,
    ) -> Result<Namespace>;

    async fn get_by_name(&self, name: &str) -> Result<Option<Namespace>>;
}

// one for "postgres", "mock", "file", ...
trait RepoCollection {
    async fn kafka_topic(&self) -> Result<Arc<dyn KafkaTopicRepo>>;
    async fn namespace(&self) -> Result<Arc<dyn NamespaceRepo>>;
}
```

This also allows you to create very simple mock repos, e.g. for `NamespaceRepo` this can just contain a `Mutex<Vec<Namespace>>`.

crepererum: On top of this repo concept, you can also have ONE generic test suite that only uses this interface (you might wanna add a `list` method to all repo traits). This would make the tests reusable and also easier to understand.

crepererum (commented):
I think we need to work on the interface quite a bit.

iox_catalog/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+version = "0.1.0"
+authors = ["Paul Dix <paul@pauldix.net>"]
+edition = "2021"
+
+[dependencies] # In alphabetical order
+async-trait = "0.1.42"
+chrono = { version = "0.4", default-features = false, features = ["clock"] }
+futures = "0.3"
+observability_deps = { path = "../observability_deps" }
+snafu = "0.6.10"
```
domodwyer, line 12: We will need to upgrade to snafu 0.7 to avoid consumers compiling two snafu versions and any potential weirdness that comes from it - there's an automated tool: https://github.com/shepmaster/snafu-upgrade-assistant

Definitely doesn't need doing now though!

iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -0,0 +1,241 @@
...
+    PRIMARY KEY (id),
+    CONSTRAINT table_name_unique UNIQUE (namespace_id, name)
+    );
+
+CREATE TABLE IF NOT EXISTS public.column_names
+(
+    id INT GENERATED ALWAYS AS IDENTITY,
+    table_id INT NOT NULL,
+    name VARCHAR NOT NULL,
+    data_type SMALLINT NOT NULL,
```
domodwyer, line 47: It's personal preference, but there are [enums](https://www.postgresql.org/docs/9.1/datatype-enum.html) in Postgres that'll avoid us having to do the "number to meaning" conversions in code and sqlx can handle the conversion. It'll also enforce only one of the valid data types is specified, rather than some random number - polish for after the spike perhaps?

iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+impl std::fmt::Display for ColumnType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        let s = self.as_str();
+
+        write!(f, "{}", s)
+    }
+}
+
+impl TryFrom<i16> for ColumnType {
+    type Error = ();
```
domodwyer, line 508:
```suggestion
    type Error = Box<dyn std::error::Error>;
```

and then swap the `Err(())` tbelow o be:

```rust
_ => Err("invalid column value".into()),
```

This way callers will still have a useful error message to display/log and you don't need to make another type.

iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+/// against the schema, or if new schema is defined, attempt to insert it into the Postgres
+/// catalog. If any new schema is created or found, this function will return a new
+/// `NamespaceSchema` struct which can replace the passed in one in cache.
+///
+/// If another writer attempts to create a column of the same name with a different
+/// type at the same time and beats this caller to it, an error will be returned. If another
+/// writer adds the same schema before this one, then this will load that schema here.
+pub async fn validate_or_insert_schema(
+    lines: Vec<ParsedLine<'_>>,
+    schema: Arc<NamespaceSchema>,
```
domodwyer, line 534: This fn doesn't need _ownership_ of `schema` - it just needs to borrow it while the fn runs in order to validate it.

If we change this line to borrow the schema, instead of passing in an Arc:

```suggestion
    schema: &NamespaceSchema,
```

We fix two things:
* No pointless `.clone()` of the schema in the caller to pass an `Arc` to this fn
* A caller that has a bare `NamespaceSchema` doesn't have to wrap it in an `Arc` purely to call this fn

The cool bit is, this still [accepts `Arc<NamespaceSchema>`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e25e86789ba920e1902837d6153d1406) too.

iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+
+/// Given the lines of a write request and an in memory schema, this will validate the write
+/// against the schema, or if new schema is defined, attempt to insert it into the Postgres
+/// catalog. If any new schema is created or found, this function will return a new
+/// `NamespaceSchema` struct which can replace the passed in one in cache.
+///
+/// If another writer attempts to create a column of the same name with a different
+/// type at the same time and beats this caller to it, an error will be returned. If another
+/// writer adds the same schema before this one, then this will load that schema here.
+pub async fn validate_or_insert_schema(
```
domodwyer, line 532: It might be worth decoupling the "create" and "validate" parts to allow callers to check if the schema is valid without necessarily updating the DB - this will be nice because validating doesn't have to be async, and will have predictable latency (no network I/O involved) which is helpful if we have to validate under a lock or something.

Should we do the schema update within a transaction that we roll back if we error out? This would prevent permanently setting column schemas in a request we return an error to the user for, or half-setting the columns in a request before hitting a conflict and bailing - I'm happy to help with this 👍

As an example of using a txn:

```rust
// Create the txn
let mut tx = pool.begin().await?;

sqlx::query("SQL HERE")
    .execute(&mut tx)
    .await?;

sqlx::query("MORE SQL")
    .execute(&mut tx)
    .await?;

tx.commit().await?;
``` 

If you drop `tx` without calling commit, the txn is rolled back. To support this, we need to pass the txn into the `create_or_get()` methods.

If you want to avoid iterating over all the fields twice after splitting this fn into two steps (iterating once to validate, and then once again to update the schema) we can return a `ChangeSet` from `validate()` that we feed to the update step: `update_schema(changes: ChangeSet)`?

iox_catalog/src/postgres.rs
```diff
@@ -0,0 +1,874 @@
...
+                                        },
+                                    );
+                                }
+                            }
+                        }
+                    }
+                }
+
+                // validate existing fields or insert
+                for (key, value) in &line.field_set {
```
domodwyer, line 584: This validation stuff is complex - I trust it's legit but we should definitely add some test cases!

domodwyer:
Sorry that's a bit of a chunky review - most of it is trivial and deferrable!

iox_catalog/migrations/20211229171744_initial_schema.sql
pauldix: Is that something that'll work in Sqlite too?

iox_catalog/src/postgres.rs
pauldix: I was deliberately avoiding using a transaction against the DB. Most requests will have no new schema defined. For the few that do, it'll likely be a small part of the overall request.

The validation shouldn't be holding a lock, which is why I had it accepting an Arc. The schema is held in cache in the router, when the write comes in it gets a clone of that namespace schema from the cache, sends it to this method, which will insert any new schema as it goes. If there's anything new, it'll return from this method, which the caller will then use to update the cache with the returned struct.

I actually want it injecting the schema as it goes and doing so without holding a transaction open against the db.

iox_catalog/migrations/20211229171744_initial_schema.sql
domodwyer: Kind of - there appears to be [various ways](https://stackoverflow.com/questions/5299267/how-to-create-enum-type-in-sqlite) of doing something similar in sqlite. The PG enums wouldn't leak out of the PG implementation though. 

One benefit of using enums that came to mind is that when quering PG directly (using psql or whatever), you'll see the data types displayed as `String` or `Tag` instead of `5` and `7` and having to look up their meaning. Hardly critical stuff though!

iox_catalog/src/lib.rs
crepererum: I would opt for an abstraction quite early. Otherwise postgres leaks through the whole code base and it will be hard to migrate to a proper interface. Also w/o any mocking testing of the remaining components will be a bit of a pain (I really don't want to interact w/ postgres just to test the ingestor core logic).

iox_catalog/src/postgres.rs
domodwyer: I don't know if this is a provlem, but with incremental schema setting you can get some weird behaviour!

Imagine racing two parallel requests for the same table, each specifying new columns that contradict with each other:

```
a="string", b=42
b="string", a=42
```

Incremental processing of these requests in parallel might lead to `a` being set to `string` AND `b` being set to `string` if both requests have their first columns processed before either proceses the second colum - both requests will fail, and neither request will ever succeed even if retried. The resulting schema is:

```
a: string
b: string
```

but the user never sent any requests that had a schema of two strings and their intended requests will fail. 

Super edge-case stuff though I think, but worth considering - I'm sure we can come back to it after the spike if the behaviour is undesirable.

crepererum (approved):
I think that's good enough for now. For follow-up PRs, we can (in parallel) improve the following things:

- **transaction handling:** we probably don't wanna commit every single `get_or_create` call, esp. for high-cardinality data structures like partitions and columns. We could use a "unit of work" (UoW) pattern here that basically couples the repo collection on a transaction.
- **mocking:** implement mocked repos and port the current postgres-specific tests over to generic tests that can be used by all backends
- **interface cleanups:** see my comment at `NamespaceSchema::add_tables_and_columns`.

domodwyer (approved):
Lets get using it and iterate as needed 🎉

iox_catalog/src/interface.rs
```diff
@@ -0,0 +1,392 @@
...
+            tables: BTreeMap::new(),
+            kafka_topic_id,
+            query_pool_id,
+        }
+    }
+
+    /// Adds tables and columns to the `NamespaceSchema`. These are created
+    /// incrementally while validating the schema for a write and this helper
+    /// method takes them in to add them to the schema.
+    pub fn add_tables_and_columns(
```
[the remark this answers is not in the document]
pauldix: The `Schema` objects are just for in memory caches. by the time this function gets called, all the tables and columns have already been added to the actual repos. This function is only used by a single caller so I could eliminate it and just put the code in that spot. It's not really meant to be part of a public facing API.

Issue #2289, opened 2021-08-14
Inconsistent statistics (try 3) in IOx


alamb:
Next phase of https://github.com/influxdata/influxdb_iox/issues/2124

I added what has turned out to be a very good bug finding aid (as well as a great way to brick IOx in dev) that asserts that the row count statistics are consistent for all columns. We had temporarily turned in into a warning, and I thought I had fixed https://github.com/influxdata/influxdb_iox/issues/2124 so re-enabled the assert.

Turns out that it was not fixed and asserted again. @NGA-TRAN  has turned off the assert again in https://github.com/influxdata/influxdb_iox/pull/2250 and this ticket tracks the work to fix it

In the long running saga

__Expected behaviour:__
Counts for all columns in a chunk should be consistent

__Actual behaviour:__
Sometimes the counts are mismatched

__Logs:__

```
thread 'IOx Query Executor Thread' panicked at 'assertion failed: `(left == right)`
  left: `3161247`,
 right: `3170727`: Mismatch counts in table http_query_response_bytes column org_id, expected 3170727 got 3161247', data_types/src/partition_metadata.rs:109:13
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:117:5
   4: data_types::partition_metadata::TableSummary::total_count
             at ./home/rust/project/data_types/src/partition_metadata.rs:109:13
   5: <server::db::access::ChunkAccess as query::pruning::PruningObserver>::was_pruned
             at ./home/rust/project/server/src/db/access.rs:164:24
   6: query::pruning::must_keep
             at ./home/rust/project/query/src/pruning.rs:98:17
   7: query::pruning::prune_chunks::{{closure}}
             at ./home/rust/project/query/src/pruning.rs:56:21
   8: core::iter::adapters::filter::filter_try_fold::{{closure}}
             at ./rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/iter/adapters/filter.rs:44:25
[...]

PR #3444, opened 2022-01-11
fix: parquet column count statistics (#2124)

tustvold:
Initially StatValues tracked null_count and non-null count. At some point this was changed to track total_count and null_count. 

Unfortunately the parquet implementation doesn't appear to have been properly updated. This didn't cause any tests to fail as the parquet crate was internally consistent on treating `total_count` as the non-null count, including in its tests. This only caused issues when interfacing with other code in IOx.

In particular it causes log output such as

```
2022-01-11T19:18:55.861248Z  WARN data_types::partition_metadata: Mismatch in statistics count, see #2124 table_name=user_asset column_name=old_amount column_count=234152 previous_count=236852
```

When running IOx with data that would previously print this error, the warning is no longer printed.

Also related to https://github.com/influxdata/conductor/issues/763


parquet_file/src/metadata.rs
```diff
@@ -715,8 +715,7 @@ fn read_statistics_from_parquet_row_group(
                 .fail()?;
             }
 
-            let count =
-                (row_group.num_rows().max(0) as u64).saturating_sub(parquet_stats.null_count());
+            let count = row_group.num_rows().max(0) as u64;
```
tustvold, line 718: This is the change, the rest is just tests plumbing

parquet_file/src/metadata.rs
```diff
@@ -952,7 +953,8 @@ mod tests {
         let n_rows = decoded.md.file_metadata().num_rows() as u64;
         assert!(n_rows >= decoded.md.num_row_groups() as u64);
         for summary in &chunk.table_summary().columns {
-            assert!(summary.total_count() <= n_rows);
+            assert_eq!(summary.total_count(), n_rows);
+            assert!(summary.null_count() <= summary.total_count());
```
alamb, line 957: Could we also assert that `decoded.md.null_count() == summary.null_count()`?

alamb:
Also possibly related https://github.com/influxdata/influxdb_iox/issues/2289 -- part of the same story

alamb (approved):
Nice catch

Issue #3450, opened 2022-01-12
InfluxQL failure with inclusive timestamp predicates (TestServer_WhereTimeInclusive)


alamb:
```
    --- FAIL: TestServer_WhereTimeInclusive/last_GTE/LTE (0.01s)
```

This seems related to  https://github.com/influxdata/influxdb_iox/issues/3445 which @e-dard  is investigating, but seems clearly to show IOx returning the wrong answer

__Input___

```
		fmt.Sprintf(`cpu value=1 %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:00:01Z").UnixNano()),
		fmt.Sprintf(`cpu value=2 %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:00:02Z").UnixNano()),

		fmt.Sprintf(`cpu value=3 %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:00:03Z").UnixNano()),
```

Which corresponds to this line protocol data:
```
cpu value=1 946684801000000000
cpu value=2 946684802000000000

cpu value=3 946684803000000000

```

This query:
__Query__
```sql
SELECT * from cpu where time >= '2000-01-01T00:00:03Z' and time <= '2000-01-01T00:00:03Z'
```

__Expected__
exp:    {"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","value"],"values":[["2000-01-01T00:00:03Z",3]]}]}]}

__Actual__
actual: {"results":[{"statement_id":0}]}

Looking at the log, IOx gets this predicate with `TimestampRange { start: 946684803000000000, end: 946684803000000001 }`

```
2022-01-12T17:11:17.788082Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: read filter db_name=54425547585a6420_000000000153158f req.range=Some(TimestampRange { start: 946684803000000000, end: 946684803000000001 }) predicate=((TagRef:_m[0x00] == "cpu") AND node: NONE)
```

alamb:
I can reproduce this error locally

```diff
diff --git a/query_tests/src/influxrpc/read_filter.rs b/query_tests/src/influxrpc/read_filter.rs
index fc1c4936..9cc2892d 100644
--- a/query_tests/src/influxrpc/read_filter.rs
+++ b/query_tests/src/influxrpc/read_filter.rs
[...]

PR #3453, opened 2022-01-12
fix: Fix max timestamp value comparison in chunk metadata

alamb:
Fixes https://github.com/influxdata/influxdb_iox/issues/3450

# Background
A `TimestampRange` structure was being used to represent *both* a predicate (where the bound is exclusive) and also set of a min/max values of data in mutable buffer and parquet chunks, where the bound is *inclusive*)

So in this specific case, we have a data min and max values of

```
946684801000000000 - 946684803000000000
```

But a predicate range of

```
946684801000000000 - 946684803000000001
```

The desired test is "are there any values in the data range that fall within the predicate range?". Clearly the answer to this question should be "yes"

However, the way the code is written it is computing "not is range disjoint with other" which was returning false.

# Fix
Create a separate `TimestampMinMax` struct to try and make it clear the semantics are (very) subtlety different.

# InfluxQL test update:
The `TestServer_WhereTimeInclusive` suite now passes (previously `TestServer_WhereTimeInclusive/last_GTE/LTE` was failing)
 
```
--- PASS: TestServer_WhereTimeInclusive (1.42s)
    --- PASS: TestServer_WhereTimeInclusive/all_GTE/LTE (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/all_GTE (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/all_LTE (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/first_GTE/LTE (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/last_GTE/LTE (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/before_GTE/LTE (0.01s)
    --- PASS: TestServer_WhereTimeInclusive/all_GT/LT (0.01s)
    --- PASS: TestServer_WhereTimeInclusive/first_GT/LT (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/last_GT/LT (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/all_GT (0.02s)
    --- PASS: TestServer_WhereTimeInclusive/all_LT (0.02s)
```

There are now 11 failing tests (down from 14) 🎉 
```
(arrow_dev) alamb@MacBook-Pro-2:~/Software/idpe$ cat /tmp/influxql.log | grep -v AAL | grep -v 'logger\.go' | grep '^--- FAIL'  | wc -l
      11
```


mutable_buffer/src/snapshot.rs
```diff
@@ -144,10 +145,10 @@ impl ChunkSnapshot {
         self.schema
             .find_index_of(TIME_COLUMN_NAME)
             .and_then(|idx| match &self.summary.columns[idx].stats {
-                Statistics::I64(stats) => Some(
-                    !TimestampRange::new(stats.min? as _, stats.max? as _)
-                        .disjoint(timestamp_range),
```
alamb, line 149: this `!disjoint` check was returning the wrong value when `stat.max == timestamp_range.max+1`

parquet_file/src/chunk.rs
```diff
@@ -83,9 +83,9 @@ pub struct ParquetChunk {
     /// Schema that goes with this table's parquet file
     schema: Arc<Schema>,
 
-    /// Timestamp range of this table's parquet file
-    /// (extracted from TableSummary)
-    timestamp_range: Option<TimestampRange>,
+    /// min/max time range of this table's parquet file
+    /// (extracted from TableSummary), if known
+    timestamp_min_max: Option<TimestampMinMax>,
```
alamb, line 88: Changed to use `TimestampMinMax`

alamb:
Marking this one as a draft as the `TestServer_Query_ExactTimeRange` InfluxQL test started newly failing after this change.

data_types/src/timestamp.rs
```diff
@@ -27,11 +34,34 @@ impl TimestampRange {
...
+    }
 
     #[inline]
-    /// Returns if this range is disjoint w.r.t the provided range
-    pub fn disjoint(&self, other: &Self) -> bool {
-        self.end <= other.start || self.start >= other.end
+    /// Returns true if any of the values between min / max
+    /// (inclusive) are within the specified timestamp range if
+    /// self.contains(v_min) || self.contains(v_max)
+    pub fn contains(&self, range: TimestampRange) -> bool {
```
e-dard, line 61: I don't have strong opinions on this but in other projects we have tended to call this "overlaps".

data_types/src/timestamp.rs
```diff
@@ -64,16 +94,33 @@ mod tests {
...
-
-        assert!(r1.disjoint(&r2));
-        assert!(r2.disjoint(&r1));
-        assert!(!r1.disjoint(&r3));
-        assert!(!r3.disjoint(&r1));
-        assert!(!r2.disjoint(&r3));
-        assert!(!r3.disjoint(&r2));
+    fn test_timestamp_range_contains_min_max() {
+        let range = TimestampRange::new(100, 200);
+        assert!(!TimestampMinMax::new(0, 99).contains(range));
```
e-dard, line 99: These look good 👍

e-dard (approved):
LGTM 👍

data_types/src/timestamp.rs
alamb: renamed to `overlaps` in b4b7855a

PR #3456, opened 2022-01-13
fix: undetected gzip HTTP body truncation

domodwyer:
If all the stars aligned in just the right way, it was possible to silently drop line protocol fields / timestamp due to truncation of the HTTP body when reading a gzipped request.

---

* fix: undected gzip HTTP body truncation (3d329018)

      When reading the gzip-encoded body of a HTTP request, the stream is read up
      until the configured maximum number of allowable bytes, at which point the
      body was silently trucated. This could allow fields in submitted line protocol
      to be silently lost (amongst other bad things).

      This change ensures that truncation results in a RequestSizeExceeded error.

### DIFF
--- a/influxdb_iox/src/influxdb_ioxd/http/utils.rs
+++ b/influxdb_iox/src/influxdb_ioxd/http/utils.rs
@@ -86,11 +86,24 @@ pub async fn parse_body(
 
         // Read at most max_size bytes to prevent a decompression bomb based
         // DoS.
-        let mut decoder = decoder.take(max_size as u64);
+        //
+        // In order to detect if the entire stream ahs been read, or truncated,
+        // read an extra byte beyond the limit and check the resulting data
+        // length - see test_read_gzipped_body_truncation.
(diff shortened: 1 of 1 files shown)


crepererum (approved):

mkmik (approved):
good catch!

influxdb_iox/src/influxdb_ioxd/http/utils.rs
```diff
@@ -133,4 +151,52 @@ mod tests {
...
+        //
+        // Effectively trimming off the " B=42" suffix.
+        let body = "bananas,A=";
+        let body = iter::once(body)
+            .chain(iter::repeat("A").take(MAX_BYTES - body.len()))
+            .chain(iter::once(" B=42\n"))
+            .flat_map(|s| s.bytes())
+            .collect::<Vec<u8>>();
+
+        // Apply gzip compression to the body
```
alamb, line 178: 🪙 👍  very nice test

PR #3460, opened 2022-01-13
feat: add query completion duration to queries system table

### LINKED ISSUE #3457: Add query complete time to Queries System Table
Currently the `queries` system table has query issue time, which records the wall-clock time when the query was issued.

Whilst investigation problems with Query mirroring internally I would find it helpful to easily see which queries are still running and how long queries took to run.

I would like to add two new columns to this table:

 1. completed: a boolean column where the value indicated whether the query has completed or not.
 2. completed_duration: a duration column where the value is `NULL` until the query completes, and then is it is the non-null duration that the query took to complete.

### Proposed Approach

I have taken a quick look around and would probably do something like this:

First, add an arc-wrapped atomic to the `QueryLogEntry` that can be used to track the duration a query took to complete:

https://github.com/influxdata/influxdb_iox/blob/a750da43f1ae580948f1ee470b6a782c9aab09cc/db/src/query_log.rs#L9-L19

Second, I would change the `record_query` method on the `QueryDatabase` trait to return a closure. This closure will be used to signal a query was completed:

https://github.com/influxdata/influxdb_iox/blob/a750da43f1ae580948f1ee470b6a782c9aab09cc/query/src/lib.rs#L80-L104

When the closure is called it would update the atomic on the `QueryLogEntry`. This would be handled in the `record_query` method  of the `QueryCatalogAccess` implementation of the `QueryDatabase` trait.

https://github.com/influxdata/influxdb_iox/blob/a750da43f1ae580948f1ee470b6a782c9aab09cc/db/src/access.rs#L213-L240

Thirdly, I would update the queries system table to add the new columns. We could use a sentinel duration, e.g., `-1` to indicate that a query was still running (setting the completed column for the row to `false` and the `query_duration` column to `NULL`). 

Fourthly, I would update all the points that call `record_query` to hold onto the closure and then call it once the query was completed.

Something like:

```rust
let finish =  db.record_query("sql", &q);
...
...
...
...

finish();
```

e-dard:
Closes #3457 

This PR adds a new `completed_duration` column to the `queries` system table, which populates rows with the wall-clock time the query took to execute.

### TLDR 

It looks like this :-)

```
edd@m1:~|⇒  influxdb_iox database query query_log "select * from system.queries"
+-----------------------------+------------+-------------------------------------+--------------------+
| issue_time                  | query_type | query_text                          | completed_duration |
+-----------------------------+------------+-------------------------------------+--------------------+
| 2022-01-13T19:32:01.281203Z | sql        | show tables                         | 25.382ms           |
| 2022-01-13T19:32:10.657779Z | sql        | select count(1) from query_log      | 60.239ms           |
| 2022-01-13T19:32:33.337761Z | sql        | select * from system.queries        | 2.951ms            |
| 2022-01-13T19:34:31.852304Z | sql        | select count(1) from query_log      | 10.927ms           |
| 2022-01-13T19:34:34.649583Z | sql        | select * from system.queries        | 1.974ms            |
| 2022-01-13T19:34:48.961946Z | sql        | select * from query_log limit 1     | 10.135ms           |
| 2022-01-13T19:34:51.180407Z | sql        | select * from system.queries        | 1.949ms            |
| 2022-01-13T19:35:09.921408Z | sql        | select min(hostname) from query_log | 9.839ms            |
| 2022-01-13T19:35:14.093952Z | sql        | select * from system.queries        |                    |
+-----------------------------+------------+-------------------------------------+--------------------+
```

### Approach

I extended the `QueryLogEntry` type to include an atomic `i64` that is populated once a query completes. The `QueryLogEntry` is returned to the caller on creation, who is responsible with moving it into a RAII token (`QueryCompletedToken`). 

When the token (which is returned from `record_query`) is dropped at the end of query execution the wall-clock time between now and the issue time of the `QueryLogEntry` is updated via the atomic field.

The `queries` system table always populates results with the most recent value for the completed duration - `NULL` indicates that the query is still running.

Incidentally: during this work I discovered that it's not possible to pretty print Duration arrays in Arrow yet. I overrode the behaviour in IOx and filed https://github.com/apache/arrow-rs/issues/1168

### DIFF
--- a/arrow_util/src/display.rs
+++ b/arrow_util/src/display.rs
@@ -1,4 +1,4 @@
-use arrow::array::{ArrayRef, DurationMillisecondArray, TimestampNanosecondArray};
+use arrow::array::{ArrayRef, DurationNanosecondArray, TimestampNanosecondArray};
 use arrow::datatypes::{DataType, TimeUnit};
 use arrow::error::{ArrowError, Result};
 use arrow::record_batch::RecordBatch;
@@ -42,13 +42,13 @@ fn array_value_to_string(column: &ArrayRef, row: usize) -> Result<String> {
             Ok(ts.to_rfc3339_opts(SecondsFormat::AutoSi, use_z))
         }
         // TODO(edd): see https://github.com/apache/arrow-rs/issues/1168
-        DataType::Duration(TimeUnit::Millisecond) if column.is_valid(row) => {
+        DataType::Duration(TimeUnit::Nanosecond) if column.is_valid(row) => {
             let dur_column = column
                 .as_any()
-                .downcast_ref::<DurationMillisecondArray>()
+                .downcast_ref::<DurationNanosecondArray>()
                 .unwrap();
 
-            let duration = std::time::Duration::from_millis(
+            let duration = std::time::Duration::from_nanos(
                 dur_column
                     .value(row)
                     .try_into()
--- a/arrow_util/src/display.rs
+++ b/arrow_util/src/display.rs
@@ -1,6 +1,6 @@
-use arrow::array::{ArrayRef, TimestampNanosecondArray};
+use arrow::array::{ArrayRef, DurationMillisecondArray, TimestampNanosecondArray};
 use arrow::datatypes::{DataType, TimeUnit};
-use arrow::error::Result;
+use arrow::error::{ArrowError, Result};
 use arrow::record_batch::RecordBatch;
 
 use comfy_table::{Cell, Table};
@@ -40,6 +40,21 @@ fn array_value_to_string(column: &ArrayRef, row: usize) -> Result<String> {
             // convert to string in preferred influx format
             let use_z = true;
             Ok(ts.to_rfc3339_opts(SecondsFormat::AutoSi, use_z))
+        }
+        // TODO(edd): see https://github.com/apache/arrow-rs/issues/1168
+        DataType::Duration(TimeUnit::Millisecond) if column.is_valid(row) => {
+            let dur_column = column
+                .as_any()
+                .downcast_ref::<DurationMillisecondArray>()
+                .unwrap();
+
+            let duration = std::time::Duration::from_millis(
+                dur_column
+                    .value(row)
+                    .try_into()
+                    .map_err(|e| ArrowError::InvalidArgumentError(format!("{:?}", e)))?,
+            );
+            Ok(format!("{:?}", duration))
         }
         _ => {
             // fallback to arrow's default printing for other types
--- a/db/src/access.rs
+++ b/db/src/access.rs
@@ -242,7 +242,7 @@ impl QueryDatabase for QueryCatalogAccess {
         // When the query token is dropped the query entry's completion time
         // will be set.
         let entry = self.query_log.push(query_type, query_text);
-        QueryCompletedToken::new(move || self.query_log.set_completed(Arc::clone(&entry)))
+        QueryCompletedToken::new(move || self.query_log.set_completed(entry))
     }
 }
 
--- a/db/src/access.rs
+++ b/db/src/access.rs
@@ -22,7 +22,7 @@ use predicate::predicate::{Predicate, PredicateBuilder};
 use query::{
     provider::{ChunkPruner, ProviderBuilder},
     pruning::{prune_chunks, PruningObserver},
-    QueryChunk, QueryChunkMeta, QueryDatabase, DEFAULT_SCHEMA,
+    QueryChunk, QueryChunkMeta, QueryCompletedToken, QueryDatabase, DEFAULT_SCHEMA,
 };
 use schema::Schema;
 use std::{any::Any, sync::Arc};
@@ -234,8 +234,15 @@ impl QueryDatabase for QueryCatalogAccess {
             .map(|table| Arc::clone(&table.schema().read()))
     }
 
-    fn record_query(&self, query_type: impl Into<String>, query_text: impl Into<String>) {
-        self.query_log.push(query_type, query_text)
+    fn record_query(
+        &self,
+        query_type: impl Into<String>,
+        query_text: impl Into<String>,
+    ) -> QueryCompletedToken<'_> {
+        // When the query token is dropped the query entry's completion time
+        // will be set.
+        let entry = self.query_log.push(query_type, query_text);
+        QueryCompletedToken::new(move || self.query_log.set_completed(Arc::clone(&entry)))
     }
 }
 
--- a/db/src/query_log.rs
+++ b/db/src/query_log.rs
@@ -9,6 +9,9 @@ use std::{
 use parking_lot::Mutex;
 use time::{Time, TimeProvider};
 
+// The query duration used for queries still running.
+const UNCOMPLETED_DURATION: i64 = -1;
+
 /// Information about a single query that was executed
 #[derive(Debug)]
 pub struct QueryLogEntry {
@@ -33,7 +36,7 @@ impl QueryLogEntry {
             query_type,
             query_text,
             issue_time,
-            query_completed_duration: (-1_i64).into(),
+            query_completed_duration: UNCOMPLETED_DURATION.into(),
         }
     }
 
@@ -42,7 +45,7 @@ impl QueryLogEntry {
             .query_completed_duration
             .load(atomic::Ordering::Relaxed)
         {
-            -1 => None,
+            UNCOMPLETED_DURATION => None,
             d => Some(Duration::from_nanos(d as u64)),
         }
     }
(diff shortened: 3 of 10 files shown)


arrow_util/src/display.rs
```diff
@@ -41,6 +41,21 @@ fn array_value_to_string(column: &ArrayRef, row: usize) -> Result<String> {
             let use_z = true;
             Ok(ts.to_rfc3339_opts(SecondsFormat::AutoSi, use_z))
         }
+        // TODO(edd): see https://github.com/apache/arrow-rs/issues/1168
```
e-dard, line 44: Filed this. I might try and get to it later in the week.

db/src/system_tables/queries.rs
```diff
@@ -94,19 +105,36 @@ mod tests {
...
+        let now = Time::from_rfc3339("1996-12-20T16:40:01+00:00").unwrap();
+        read_filter_entry.set_completed(now);
+
+        let expected = vec![
+            "+----------------------+-------------+-------------------+--------------------+",
+            "| issue_time           | query_type  | query_text        | completed_duration |",
+            "+----------------------+-------------+-------------------+--------------------+",
+            "| 1996-12-19T16:39:57Z | sql         | select * from foo |                    |",
+            "| 1996-12-20T16:39:57Z | sql         | select * from bar |                    |",
+            "| 1996-12-20T16:39:57Z | read_filter | json goop         | 4s                 |",
```
e-dard, line 134: Using a duration column type gives us a nice representation of the elapsed time.

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -245,9 +245,9 @@ where
             .db_store
             .db(&db_name)
             .context(DatabaseNotFoundSnafu { db_name: &db_name })?;
-        db.record_query("read_filter", defer_json(&req));
+        let _query_completed_token = db.record_query("read_filter", defer_json(&req));
```
e-dard, line 248: when the token is dropped the query will be timed.

query/src/lib.rs
```diff
@@ -77,6 +77,31 @@ pub trait QueryChunkMeta: Sized {
...
+
+impl<'a> QueryCompletedToken<'a> {
+    pub fn new(f: impl Fn() + Send + 'a) -> Self {
+        Self { f: Box::new(f) }
+    }
+}
+
+impl<'a> Drop for QueryCompletedToken<'a> {
+    fn drop(&mut self) {
+        (self.f)()
```
e-dard, line 101: time the query...

query/src/lib.rs
```diff
@@ -77,6 +77,31 @@ pub trait QueryChunkMeta: Sized {
...
+}
+
+impl<'a> Debug for QueryCompletedToken<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("QueryCompletedToken").finish()
+    }
+}
+
+impl<'a> QueryCompletedToken<'a> {
+    pub fn new(f: impl Fn() + Send + 'a) -> Self {
```
e-dard, line 94: We need a lifetime so that we can show that the function lives as long as the `QueryLog` the entry came from

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/flight.rs
```diff
@@ -172,7 +172,7 @@ impl Flight for FlightService {
             .db(&database)
             .map_err(default_server_error_handler)?;
 
-        db.record_query("sql", &read_info.sql_query);
+        let _query_completed_token = db.record_query("sql", &read_info.sql_query);
```
e-dard, line 175: FYI - because the data is streamed out of here, I'm pretty sure that this isn't really capturing the entire query time (particularly streaming results back to the client)...

db/src/system_tables/queries.rs
alamb: very nice -- it is also a nice nudge to adding better / full support for durations

db/src/query_log.rs
```diff
@@ -25,8 +33,25 @@ impl QueryLogEntry {
             query_type,
             query_text,
             issue_time,
+            query_completed_duration: (-1_i64).into(),
```
alamb, line 36: It is probably fine, but maybe worth a named constant like

```rust
const UNCOMPLETED_DURATION: i64 = -1;
``` 

Rather than `-1`? 

Given there are only two uses I don't think this is a big deal at all

db/src/access.rs
```diff
@@ -234,8 +234,15 @@ impl QueryDatabase for QueryCatalogAccess {
...
-        self.query_log.push(query_type, query_text)
+    fn record_query(
+        &self,
+        query_type: impl Into<String>,
+        query_text: impl Into<String>,
+    ) -> QueryCompletedToken<'_> {
+        // When the query token is dropped the query entry's completion time
+        // will be set.
+        let entry = self.query_log.push(query_type, query_text);
+        QueryCompletedToken::new(move || self.query_log.set_completed(Arc::clone(&entry)))
```
alamb, line 245:
```suggestion
        QueryCompletedToken::new(move || self.query_log.set_completed(entry))
```

since `entry` is already an `Arc` I don't think you need to clone it again

You'll also likely have to make `QueryCompletedToken::new() take a `FnOnce` rather than a `Fn` like this:

```diff
diff --git a/query/src/lib.rs b/query/src/lib.rs
index 143faabe..188540ed 100644
--- a/query/src/lib.rs
+++ b/query/src/lib.rs
@@ -81,7 +81,8 @@ pub trait QueryChunkMeta: Sized {
 /// a `QueryDatabase`. It is used to trigger side-effects (such as query timing)
 /// on query completion.
 pub struct QueryCompletedToken<'a> {
-    f: Box<dyn Fn() + Send + 'a>,
+    // Is an option so it can be taken during drop()
+    f: Option<Box<dyn FnOnce() + Send + 'a>>
 }
 
 impl<'a> Debug for QueryCompletedToken<'a> {
@@ -91,14 +92,16 @@ impl<'a> Debug for QueryCompletedToken<'a> {
 }
 
 impl<'a> QueryCompletedToken<'a> {
-    pub fn new(f: impl Fn() + Send + 'a) -> Self {
-        Self { f: Box::new(f) }
+    pub fn new(f: impl FnOnce() + Send + 'a) -> Self {
+        Self { f: Some(Box::new(f)) }
     }
 }
 
 impl<'a> Drop for QueryCompletedToken<'a> {
     fn drop(&mut self) {
-        (self.f)()
+        if let Some(f) = self.f.take() {
+            (f)()
+        }
     }
 }
```

I don't think it really matters, but figured I would point it out

db/src/query_log.rs
```diff
@@ -49,29 +74,74 @@ impl QueryLog {
...
+            entry.query_completed_duration(),
+            Some(Duration::from_millis(0))
+        );
+
+        // when the query completes some time in the future.
+        time_provider.set(Time::from_timestamp_millis(300));
+        entry.set_completed(time_provider.now());
+        assert_eq!(
+            entry.query_completed_duration(),
+            Some(Duration::from_millis(200))
```
alamb, line 144: 👍

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/flight.rs
alamb: well one could argue that the time the client takes to get results back should be included in query duration (as the resources are tied up for the entire time)

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -245,9 +245,9 @@ where
...
 
-        let results = read_filter_impl(db, db_name, req, span_ctx)
+        let results = read_filter_impl(Arc::clone(&db), db_name, req, span_ctx)
```
alamb, line 250: I didn't understand the need for`Arc::clone()` here, but then I tried to remove it and see it is because `QueryCOmpletedToken` has a lieftime associated with it.

I tried to get rid of that and I found that it was because a `self` gets captured ... 

Another approach would be to `Arc::clone()` the `time_provider` so that the completion token doesn't have a lifetime associated with it. 

But this works too 🤷 and i think structure is a matter of preference

query/src/lib.rs
alamb: specifically, it is b/c the function that is bound has a `self` reference I believe

alamb (approved):
LGTM -- I have some suggestions but nothing that would prevent this PR from being merged

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/flight.rs
e-dard: I suspect that for most queries of interest (like, slow ones...), processing will be the dominant factor (and this the query time here will be accurate). So I think this will do for now, and if we wanted a more accurate one we would be better off measuring it at the connection layer or on the client side.

db/src/access.rs
e-dard: Good suggestion. Will add this in.

PR #3464, opened 2022-01-14
fix: reject empty org & bucket

domodwyer:
Not an earth shattering issue, but probably not the behaviour we want - if a client specifies the `org` and `bucket` as empty, there's probably a problem upstream.

---

* fix: reject empty org & bucket (5cce69b4)

      Previously a request that specified an empty org & bucket value would be 
      mapped to a database named "_".

      This commit changes the org/bucket mapping fn to return an error if either org
      or bucket is empty.


NGA-TRAN (approved):
LGTM

PR #3465, opened 2022-01-14
feat: iox_catalog sequencers

domodwyer:
Support for the `sequencer` table + misc doc improvements.

---

* refactor: ensure sequencers are unique (093251fb)

      Adds a unique constraint to ensure only one sequencer record exists for each
      Kafka (topic, partition).

* test: use DSN from env for integration tests (c628d326)

      Removes the hard-coded DSN, instead sourcing it from the DATABASE_URL 
      environment variable.

* docs: integration testing for iox_catalog (5cf93f8b)

      Documents the required steps in order to run the Postgres integration tests
      for the iox_catalog crate.

* feat(iox_catalog): create & list sequencers (c87796de)

      Adds support for interacting with the "sequencer" table.

* chore: update lockfile (595f7d33)

      Running cargo in iox_catalog generates a lockfile diff.

### DIFF
--- a/iox_catalog/migrations/20211229171744_initial_schema.sql
+++ b/iox_catalog/migrations/20211229171744_initial_schema.sql
@@ -56,7 +56,8 @@ CREATE TABLE IF NOT EXISTS iox_catalog.sequencer
     kafka_topic_id INT NOT NULL,
     kafka_partition INT NOT NULL,
     min_unpersisted_sequence_number BIGINT,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    CONSTRAINT sequencer_unique UNIQUE (kafka_topic_id, kafka_partition)
     );
 
 CREATE TABLE IF NOT EXISTS iox_catalog.sharding_rule_override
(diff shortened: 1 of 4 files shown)


iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -56,7 +56,8 @@ CREATE TABLE IF NOT EXISTS iox_catalog.sequencer
     kafka_topic_id INT NOT NULL,
     kafka_partition INT NOT NULL,
     min_unpersisted_sequence_number BIGINT,
-    PRIMARY KEY (id)
+    PRIMARY KEY (id),
+    CONSTRAINT sequencer_unique UNIQUE (kafka_topic_id, kafka_partition)
```
domodwyer, line 60: @pauldix I think this is needed right? I believe we would want one `min_unpersisted_sequence_number` per `(topic, partition)` pair.

pauldix: It is, good catch!

pauldix (approved):

PR #3468, opened 2022-01-14
feat(router2): http write handler

domodwyer:
Implements the write HTTP endpoint and plumbs in the skeleton for DML request handling. 

Includes tests for #3456 and #3464 to avoid any regressions.

I shall follow this up with a delete endpoint impl.

---

* feat: abstract DML handler (80b12d41)

      Defines the DmlHandler trait responsible for processing a request in some
      abstract way, decoupling the HTTP/gRPC request handlers from the underlying
      routing logic.

* feat: router2 HTTP handlers (40a290f6)

      Implements the HTTP v2 write API endpoint for router2.


router2/src/dml_handler/mod.rs
```diff
@@ -0,0 +1,14 @@
+//! DML operation handling.
+//!
+//! An IOx node operating as a router exposes multiple API interfaces (HTTP,
+//! gRPC) which funnel requests into a common [`DmlHandler`] implementation,
+//! responsible for processing the request before pushing the results into the
+//! appropriate write buffer sink.
+
+mod r#trait;
```
mkmik, line 8: I never seen this syntax; is this a special syntax or just a naming convention?

domodwyer: This sacrificial hash pleases the AST gods - `trait` is a reserved keyword but this lets you use it as not-a-keyword, it's effectively `mod trait;`

mkmik (approved):
LGTM!

domodwyer:
Yeah that's a good point! Currently you have to create a op to call `op.shard()` but the sharding could just as easily operate on the set of parsed lines themselves 👍

tustvold:
> but the sharding could just as easily operate on the set of parsed lines themselves

Not fully caught up on this, but FWIW the reason sharding is done on DMLOperation was to handle non-line protocol write payloads. The data would get written into a uniform in-memory representation so that the sharding logic could be shared. Not sure if this is still a concern

domodwyer:
A `DmlWrite` is constructed from a `HashMap<TableName, MutableBatch>` and all write paths I looked at had this hashmap as an intermediate in order to create the DML write, so I think we can shard that rather than the `DmlWrite` itself.

I plan on pulling the sharding logic out of the DML types, so a sharder operates on a set of writes/deletes rather than the ops sharding themselves - this way we can have different sharder impls for different sharding logic, such as the column overrides in the future. I'll probably wind up with a trait impl for both the in-memory write & delete structures to support sharding, so even if we have some non-hashmap-thing we should still be good so long as we can return a table name, etc 👍

Issue #3458, opened 2022-01-13
Can't deserialize canceled `IoxOperation` with `serde_json` after pbjson update to 2.0: "data did not match any variant of untagged enum Content"


alamb:
List the minimal actions needed to reproduce the behaviour.

Try and encode/decode an `IOxOperation` that is canceled using `serde_json` (and thus implicitly `pbjson`)

This (non canceled) operation successfully parses as an `:`

```json
{
  "operation": {
    "name": "51",
    "metadata": {
      "typeUrl": "type.googleapis.com/influxdata.iox.management.v1.OperationMetadata",
      "value": "CPGJARCYr4beAxgBKgcKBYCU69wDaAE="
    },
    "done": true,
    "response": {
      "typeUrl": "type.googleapis.com/google.protobuf.Empty"
    }
  },
  "metadata": {
    "cpuNanos": "17649",
    "wallNanos": "1002543000",
    "totalCount": "1",
    "successCount": "1",
    "dummy": {
      "nanos": [
        "1000000000"
      ]
    }
  }
}
```

This FAILS to parse as an `IoxOperation`(note that it has an `error` rather than `response`)
```json
{
  "operation": {
    "name": "50",
    "metadata": {
      "typeUrl": "type.googleapis.com/influxdata.iox.management.v1.OperationMetadata",
      "value": "CNlDELjejwEYASoHCgWAlOvcA3gB"
    },
    "done": true,
    "error": {
      "code": 1,
      "message": "Job cancelled"
    }
  },
  "metadata": {
    "cpuNanos": "8665",
    "wallNanos": "2355000",
    "totalCount": "1",
    "cancelledCount": "1",
    "dummy": {
      "nanos": [
        "1000000000"
      ]
    }
  }
}
```

Full repro: 
[repro.zip](https://github.com/influxdata/influxdb_iox/files/7864314/repro.zip)
1. Unzip
2. Point `Cargo.toml` at your local checkout of `influxdb_iox` with the changes from https://github.com/influxdata/influxdb_iox/pull/3455
3. `cargo run`

__Expected behaviour:__
Test should pass and no panic produced)

__Actual behaviour:__
Test errors with
```
thread 'main' panicked at 'Parsing: Error("data did not match any variant of untagged enum Content", line: 10, column: 15)', src/main.rs:41:60
```

alamb:
@tustvold  FYI

tustvold:
I plan to look into this today

[...]

PR #3469, opened 2022-01-16
fix: serde_json `arbitrary_precision` (#3458)

tustvold:
https://github.com/serde-rs/json/pull/848 fixed the problem pbjson was running into. There are still some unexpected behaviours of `arbitrary_precision` so I still standby the change to disable it in arrow-rs pending https://github.com/serde-rs/json/issues/846, but pbjson should be back to working now :smile: 

Closes https://github.com/influxdata/influxdb_iox/issues/3458


alamb (approved):
Nice 👍

tustvold:
Trying to bump CI to see if Semantic Pull Request comes back to life

PR #3472, opened 2022-01-17
feat(router2): implement delete API handler

domodwyer:
Follows on from #3468 adding a HTTP handler for the `/api/v2/delete` endpoint.

---

* feat(router2): implement delete API handler (6f2e10ca)

      Adds support for the HTTP v2 delete API endpoint.

### DIFF
--- a/router2/src/server/http.rs
+++ b/router2/src/server/http.rs
@@ -8,6 +8,7 @@ use data_types::names::{org_and_bucket_to_database, OrgBucketMappingError};
 use futures::StreamExt;
 use hyper::{header::CONTENT_ENCODING, Body, Method, Request, Response, StatusCode};
 use observability_deps::tracing::*;
+use predicate::delete_predicate::parse_http_delete_request;
 use serde::Deserialize;
 use thiserror::Error;
 use time::{SystemProvider, TimeProvider};
@@ -54,6 +55,10 @@ pub enum Error {
     #[error("failed to parse line protocol: {0}")]
     ParseLineProtocol(mutable_batch_lp::Error),
 
+    /// Failure to parse the request delete predicate.
+    #[error("failed to parse delete predicate: {0}")]
+    ParseDelete(#[from] predicate::delete_predicate::Error),
+
     /// An error returned from the [`DmlHandler`].
     #[error("dml handler error: {0}")]
     DmlHandler(#[from] DmlError),
@@ -79,6 +84,7 @@ impl Error {
             Error::RequestSizeExceeded(_) => StatusCode::PAYLOAD_TOO_LARGE,
             Error::ParseLineProtocol(_) => StatusCode::BAD_REQUEST,
             Error::DmlHandler(DmlError::Internal(_)) => StatusCode::INTERNAL_SERVER_ERROR,
+            Error::ParseDelete(_) => StatusCode::BAD_REQUEST,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


pauldix (approved):

router2/src/server/http.rs
```diff
@@ -476,25 +554,135 @@ mod tests {
...
+        ok,
+        query_string = "?org=bananas&bucket=test",
+        body = r#"{"start":"2021-04-01T14:00:00Z","stop":"2021-04-02T14:00:00Z", "predicate":"_measurement=its_a_table and location=Boston"}"#.as_bytes(),
+        dml_handler = [Ok(())],
+        want_result = Ok(_),
+        want_dml_calls = [MockDmlHandlerCall::Delete(d)] => {
+            assert_eq!(d.table_name, "its_a_table");
+            assert_eq!(d.start_time, "2021-04-01T14:00:00Z");
+            assert_eq!(d.stop_time, "2021-04-02T14:00:00Z");
+            assert_eq!(d.predicate, "location=Boston");
```
NGA-TRAN, line 593: A few detailed for the following-up PR:
1. Do we need to support empty table which means we will delete all tables of the specified namespace (a customer DB) . With this new design, I do not think we should do it but this is a business question and @e-dard has talked with business side for this and knows the answer better.
2. `start_time` and `stop_time` must always have values. We may want to test for that.
3. `predicate` can be empty. we want to test that , too.
4. `we may need to add for `predicates with more complicated expression and include many columns such as `location=Boston and (temp>45 or town!=Andover)`

NGA-TRAN (approved):
LGTM. I just add a few details for follow-up PRs

router2/src/server/http.rs
domodwyer: I admire your testing mentatily - TEST ALL THE THINGS 🔍 

These tests ensure that the HTTP handler behaviour is correct - specifically:

* The HTTP request body is correctly read, and passed into the delete predicate parser, producing the correct output.
* An error generated by the predicate parser is returned and correctly mapped to an error response to the user.

The behaviour of the predicate parser and edge cases are tested separately, and pretty extensively [here](https://github.com/influxdata/influxdb_iox/blob/6f2e10cab6c59e7684bedb6ae70605d6f604cf7d/predicate/src/delete_predicate.rs#L385) (by @crepererum I think 👍 )

It looks like it accepts both [empty tables and empty predicates](https://github.com/influxdata/influxdb_iox/blob/6f2e10cab6c59e7684bedb6ae70605d6f604cf7d/predicate/src/delete_predicate.rs#L636-L637) - if we want to disallow this for the new architecture I shall add validation to this in the (upcoming) DML handler so that the validation is applied to all APIs (http & grpc) as they will all funnel into a common DML handler - I'm envisioning something like this:

```
    ┌──────────────┐     ┌──────────────┐
    │   HTTP API   │     │   gRPC API   │
    └──────────────┘     └──────────────┘
            │                    │
            │  ┌──────────────┐  │
            └─▶│ DML Handler  │◀─┘
               └───────┬──────┘
                       │
            ┌─Deletes──┴──Writes─┐
            │                    │
            │                    │
            ▼                    ▼
                         ┌──────────────┐
    Not Implemented      │    Schema    │
         (yet)           │  Validation  │
                         └───────┬──────┘
                                 ▼
                         ┌──────────────┐
                         │   Sharder    │
                         └───────┬──────┘
                                 ▼
                         ┌──────────────┐
                         │  Write Sink  │
                         └───────┬──────┘
                                 │
                        ┌────────▼─────┐
                        │    Kafka     ├┐
                        └┬─────────────┘├┐
                         └┬─────────────┘│
                          └──────────────┘
```

NGA-TRAN: @domodwyer If you use the existing code that include those tests, we are good then

NGA-TRAN: And very nice diagram to describe the process 👍

PR #3475, opened 2022-01-17
feat: implement memory backed iox_catalog

pauldix:
* Adds a memory based catalog, useful for testing purposes.
* Separates getting the namespace schema from the namespace and moves the schema code out interface out of postgres.
* Make postgres and mem catalog implementations public
* Make iox_catalog tests generic for any backend implementation

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -2,7 +2,7 @@
 
 use async_trait::async_trait;
 use influxdb_line_protocol::FieldValue;
-use snafu::Snafu;
+use snafu::{OptionExt, Snafu};
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::fmt::Formatter;
@@ -33,6 +33,9 @@ pub enum Error {
         name
     ))]
     UnknownColumnType { data_type: i16, name: String },
+
+    #[snafu(display("namespace {} not found", name))]
+    NamespaceNotFound { name: String },
 }
 
 /// A specialized `Error` for Catalog errors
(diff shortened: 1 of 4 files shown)


iox_catalog/src/interface.rs
```diff
@@ -181,6 +187,63 @@ impl NamespaceSchema {
         }
     }
 
+    /// Gets the namespace schema including all tables and columns.
+    pub async fn get_by_name<T: RepoCollection + Send + Sync>(
```
crepererum, line 191: I would prefer if the data structs (namespace, table, column) would NOT interact w/ repos but would be rather simple and that the CRUD operations are performed by the repos / repo collection or by a freestanding function (for this specific example here I don't really see a reason to attach this method to a type/trait).

iox_catalog/src/lib.rs
```diff
@@ -195,3 +196,99 @@ pub async fn create_or_get_default_records<T: RepoCollection + Sync + Send>(
 
     Ok((kafka_topic, query_pool, sequencers))
 }
+
+#[cfg(test)]
+mod tests {
```
crepererum, line 201: :tada:

crepererum (approved):
Minor nitpick, I really the new abstract test suite :+1:

domodwyer (approved):
Lovely stuff

iox_catalog/src/mem.rs
```diff
@@ -0,0 +1,294 @@
...
+use crate::interface::{
+    Column, ColumnRepo, ColumnType, Error, KafkaTopic, KafkaTopicRepo, Namespace, NamespaceRepo,
+    QueryPool, QueryPoolRepo, RepoCollection, Result, Sequencer, SequencerRepo, Table, TableRepo,
+};
+use async_trait::async_trait;
+use std::convert::TryFrom;
+use std::fmt::Formatter;
+use std::sync::{Arc, Mutex};
+
+/// In-memory catalog that implements the `RepoCollection` and individual repo traits fromt
```
NGA-TRAN, line 13:
```suggestion
/// In-memory catalog that implements the `RepoCollection` and individual repo traits from
```

NGA-TRAN (approved):
LGTM

iox_catalog/src/interface.rs
pauldix: I had it under `NamespaceSchema` because that's what the function returns. I could either do this as a free standing function like `validate_or_insert_schema` or I could move it under the `RepoCollection` trait. If I did the latter, I'd probably move `validate_or_insert_schema` and `create_or_get_default_records` to be under the trait as well. You have a preference on which?

pauldix: moved to a freestanding function: 5e464727d141dfcfbaa3ecd4b821bdd4d3e0c6cd

PR #3476, opened 2022-01-18
feat: add `size`  methods to DML-related types

crepererum:
This will be helpful when we want to batch DML operations in memory
(e.g. when using RSKafka).

This also ensures that `MBChunk` accounts for the column names that
are stored within `MutableBatch`.


tustvold (approved):

alamb:
Semantic PR seemed to have been stuck, so I change the title of this PR to unstick it (so kodiak could begin working through the merge queue)

PR #3480, opened 2022-01-18
feat: add Partition scaffolding to iox_catalog crate

pauldix:


NGA-TRAN (approved):

iox_catalog/src/interface.rs
```diff
@@ -131,6 +131,21 @@ impl SequencerId {
     }
 }
 
+/// Unique ID for a `Sequencer`
```
NGA-TRAN, line 134: @pauldix : this is why I got confused. Here "Unique ID for a `Sequencer`". Then below it was said for IOx Partition

pauldix: Fixed in #3482

Issue #3439, opened 2022-01-11
InfluxQL: Timestamp range predicates with `MaxNanoTime` are incorrect (TestServer_Query_ShowQueries_Future)


alamb:
TestServer_Query_ShowQueries_Future/show_tag_keys

__Steps to reproduce:__
Use data like:

```
cpu,host=server01 value=100 %d`, models.MaxNanoTime
```

__query___

```sql
SHOW TAG KEYS FROM cpu

```

__Expected behaviour:__

exp:    {"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["tagKey"],"values":[["host"]]}]}]}

__Actual behaviour:__

actual: {"results":[{"statement_id":0}]}

The actual request that is made to IOx looks like the following

```
2022-01-11T16:24:46.699337Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: measurement_tag_keys db_name=54425547585a6420_000000000153158f range=Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }) measurement=cpu predicate=<NONE>
```

Note the end of the timerange is specified as `9223372036854775806`

The data is present:
```sql
$ influxdb_iox database query 54425547585a6420_000000000153158f  'select * from cpu'
+----------+--------------------------------+-------+
| host     | time                           | value |
+----------+--------------------------------+-------+
| server01 | 2262-04-11T23:47:16.854775806Z | 100   |
+----------+--------------------------------+-------+
```

# Other example

The same basic problem happens with the other test in `TestServer_Query_ShowQueries_Future`, 

```
TestServer_Query_ShowQueries_Future/show_field_keys
```

__query___

```sql
SHOW FIELD KEYS
```

__Expected behaviour:__

exp:    {"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["fieldKey","fieldType"],"values":[["value","float"]]}]}]}

__Actual behaviour:__

actual: {"results":[{"statement_id":0}]}

The actual request that is made to IOx looks like the following

```
2022-01-11T16:28:00.176321Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: measurement_names db_name=54425547585a6420_000000000153158f range=Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }) predicate=<NONE>
```

And somehow or other it appears to have pruned out all the chunks
```
[...]

PR #3485, opened 2022-01-19
fix: Special case tag_keys with max timestamp range

alamb:
Fixes https://github.com/influxdata/influxdb_iox/issues/3439

# Rationale
There is a special case handling for the `tag_key` schema query when the predicate spans the entire time range ( include min/max nano time:

https://github.com/influxdata/idpe/blob/e6246019a22c40a264415819d2445f26cbe7fff8/storage/tsdb/tsm1/engine_schema.go#L539-L541

# Changes
1. Implement the same special case in IOx
2. tests


predicate/src/predicate.rs
```diff
@@ -207,6 +207,23 @@ impl Predicate {
...
         }
     }
+
+    /// Removes the timestamp range from this predicate, if the range
+    /// is for the entire min/max valid range.
+    ///
+    /// This is used in certain cases to retain compatibility with the
+    /// existing storage engine
+    pub fn clear_timestamp_if_max_range(mut self) -> Self {
+        self.range = self.range.take().and_then(|range| {
```
alamb, line 217: here is the check

query/src/frontend/influxrpc.rs
```diff
@@ -339,6 +339,9 @@ impl InfluxRpcPlanner {
     {
         debug!(predicate=?predicate, "planning tag_keys");
 
+        // Special case
+        let predicate = predicate.clear_timestamp_if_max_range();
```
alamb, line 343: here is where it is called

e-dard (approved):
Seems like a reasonable solution to me 👍

Issue #3486, opened 2022-01-19
InfluxQL: TimestampRange is not limited to min/max nano timestamps


alamb:
It turns out that the valid timestamp range for the Influx StorageRPC is not `i64::MIN` to `i64::MAX` but instead is `[i64::MIN +2, i64::MAX - 1]`

However, IOX will allow through predicates that are outside of this range

Iox should also clamp the timestamp range of the predicates to min/max as well

Here is an example of such code 
https://github.com/influxdata/idpe/blob/83d1efa4afc7afc9cbb892c8a6ab064523afc52b/storage/store/store.go#L960-L965

PR #3487, opened 2022-01-19
fix(InfluxQL): Limit `TimestampRange` to Min/Max times

alamb:
~built on https://github.com/influxdata/influxdb_iox/pull/3485 so draft until that is done~

# Rationale
Fixes https://github.com/influxdata/influxdb_iox/issues/3486

# Changes
1. Clamp the min/max values of `TimestampRange` to `MIN_NANO_TIME` and `MAX_NANO_TIME`
2. Make `TimestampRange::start and TimestampRange::end` non public so the invariants can be enforced


data_types/src/timestamp.rs
```diff
@@ -36,15 +36,17 @@ pub const MAX_NANO_TIME: i64 = i64::MAX - 1;
...
+    /// Start defines the inclusive lower bound. Minimum value is [MIN_NANO_TIME]
+    start: i64,
+    /// End defines the exclusive upper bound. Maximum value is [MAX_NANO_TIME]
+    end: i64,
 }
 
 impl TimestampRange {
     pub fn new(start: i64, end: i64) -> Self {
         debug_assert!(end >= start);
+        let start = start.max(MIN_NANO_TIME);
```
alamb, line 48: Here is the key / only substantial change -- the rest of this PR is tests and updates now that `start` and `end` are not public

e-dard (approved):

PR #3488, opened 2022-01-19
feat(router2): sharding abstraction & logic

domodwyer:
This PR adds a sharder abstraction that should be flexible enough for us to provide different implementations of sharding logic for different situations, such as:

* Sharding by the contents of a LP write
* Sharding by table & namespace (implemented in this PR)
* Sharding by column values
* Sharding one of the above to multiple shards (i.e. sharding a delete to multiple sequencers when using column overrides)
* ...etc...

I will build on this to construct a sharded set of sequencers next.

---

* refactor: API agnostic DML delete handler (d62e1c67)

      Changes the DmlHandler::delete() trait method to accept required params, and
      accept a DeletePredicate instead of a HttpDeleteRequest so that it can be
      re-used in the gRPC handler.

* refactor: use static DatabaseName instances (aa308043)

* feat: sharder trait & impl (118fdfd0)

      This commit defines the Sharder trait that should allow us to implement 
      multiple sharding strategies over a defined set of input types (such as a
      MutableBatch for writes, DeletePredicate for deletes, etc).

      This commit also includes a jump hash implementation that consistently shards
      (table name, namespace) tuples to a given shard for all input types.


router2/src/dml_handler/nop.rs
```diff
@@ -29,12 +29,15 @@ impl DmlHandler for NopDmlHandler {
         Ok(())
     }
 
-    async fn delete(
+    async fn delete<'a>(
         &self,
-        delete: predicate::delete_predicate::HttpDeleteRequest,
+        namespace: DatabaseName<'static>,
+        table: impl Into<String> + Send + Sync + 'a,
+        predicate: DeletePredicate,
```
NGA-TRAN, line 36: 👍

router2/src/sharder/trait.rs
```diff
@@ -0,0 +1,24 @@
...
+/// sharding.
+///
+/// NOTE: It is a system invariant that deletes are routed to (all of) the same
+/// sequencers as a write for the same table.
+pub trait Sharder<P>: Debug + Send + Sync {
+    /// The type returned by a sharder.
+    ///
+    /// This could be a shard ID, a sequencer, an array of multiple sequencers,
+    /// etc.
+    type Item: Debug + Send + Sync;
```
NGA-TRAN, line 20: I like this abstract

NGA-TRAN (approved):
I do not look deeply into the details of seed and consistent hash method but I do like the `Sharder` abstract trait and that you reuse `delete_predicate` type

pauldix (approved):
LGTM

router2/src/sharder/table_namespace_sharder.rs
```diff
@@ -0,0 +1,208 @@
...
+        key.hash(&mut state);
+        let mut key = state.finish();
+
+        let mut b = -1;
+        let mut j = 0;
+        while j < self.shards.len() as i64 {
+            b = j;
+            key = key.wrapping_mul(2862933555777941757).wrapping_add(1);
+            j = ((b.wrapping_add(1) as f64) * (((1u64 << 31) as f64) / (((key >> 33) + 1) as f64)))
+                as i64
```
crepererum, line 93: What's this and where does it come from? Do we really roll our own hashing algorithm here?

router2/src/sharder/table_namespace_sharder.rs
```diff
@@ -0,0 +1,208 @@
...
+    ///
+    /// # Panics
+    ///
+    /// This constructor panics if the number of elements in `shards` is 0.
+    pub fn new(shards: impl IntoIterator<Item = T>) -> Self {
+        // A randomly generated static siphash key to ensure all router
+        // instances hash the same input to the same u64 sharding key.
+        //
+        // Generated with: xxd -i -l 16 /dev/urandom
+        let key = [
```
crepererum, line 50: We should have some fixture tests ensuring that hashes are stable, even when we update our codebase / dependencies.

crepererum (commented):
Why can we not reuse `DmlOperation::shard` for this?

router2/src/sharder/table_namespace_sharder.rs
domodwyer: This is the hashing implementation specified as part of the jumphash algorithm - paper linked in the type comment 👍

router2/src/sharder/table_namespace_sharder.rs
domodwyer: Good idea 👍

domodwyer:
I always intended to separate out sharding from the DML type - it seems like a conflation of responsiblities to me - the sharding code should have the singular responsiblity of just sharding some data, and the DML should define data which other code can operate on (separating logic/data). This lets you abstract sharding away and easily substitute different sharding strategies, reuse the sharding logic, stub out sharding for tests, etc.

If you were asking why we shard the `MutableBatch` map directly, it avoids creating the DML only to break it apart and build more DML instances immediately after - check out the comments [on this PR](https://github.com/influxdata/influxdb_iox/pull/3468).

router2/src/sharder/table_namespace_sharder.rs
```diff
@@ -0,0 +1,230 @@
...
+    // the cluster are producing the same mapping of K->B. However, this would
+    // not be the case during a rolling deployment where some nodes are using
+    // one mapping, and new nodes using another.
+    //
+    // This test being updated necessitates a stop-the-world deployment (stop
+    // all routers, deploy new hashing code on all routers, resume serving
+    // traffic) to inconsistently routing of ops. Also prepare a roll-back
+    // strategy would that accounts for this mapping change.
+    #[test]
+    fn test_key_bucket_fixture() {
```
domodwyer, line 222: I hope this comment adiquetly explains the "here be dragons" nature of this test failing and what it means to us in a broader sense.

PR #3489, opened 2022-01-19
refactor: consolidate all query_test Scenarios

alamb:
Builds on https://github.com/influxdata/influxdb_iox/pull/3485 so draft until that is merged

# Rationale
1. If all scenarios are defined in the same place, it is easier to find an existing scenario than write a new one
2. I am preparing to remove the use of `async_trait` in favor of a straight up future (to improve compile speeds) and I want them all in the same place

Possible downsides:
1. having the data and expected results in different files may not be as convenient, but I think the ongoing development improvement will be worth it.

This consolidation also makes clear the scale of the challenge the rust compiler is facing.

# Changes:
Move code around (no logical change intended)


alamb:
Following @carols10cents 's lead and closing this one as it is accumulating merge conflicts

PR #3490, opened 2022-01-19
feat: Add parquet file records to iox_catalog

pauldix:
* Adds ParquetFile and scaffolding to IOx catalog
* Changed the file_location in parquet_file to object_store_id which is a uuid

### DIFF
--- a/iox_catalog/migrations/20211229171744_initial_schema.sql
+++ b/iox_catalog/migrations/20211229171744_initial_schema.sql
@@ -84,15 +84,15 @@ CREATE TABLE IF NOT EXISTS iox_catalog.parquet_file
     id BIGINT GENERATED ALWAYS AS IDENTITY,
     sequencer_id SMALLINT NOT NULL,
     table_id INT NOT NULL,
-    partition_id INT NOT NULL,
-    file_location VARCHAR NOT NULL,
+    partition_id BIGINT NOT NULL,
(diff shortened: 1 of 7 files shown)


iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -84,15 +84,15 @@ CREATE TABLE IF NOT EXISTS iox_catalog.parquet_file
     id BIGINT GENERATED ALWAYS AS IDENTITY,
     sequencer_id SMALLINT NOT NULL,
     table_id INT NOT NULL,
-    partition_id INT NOT NULL,
-    file_location VARCHAR NOT NULL,
+    partition_id BIGINT NOT NULL,
+    object_store_id uuid NOT NULL,
```
NGA-TRAN, line 88: What is this `object_store_id` for? Is it a unique id that identifies a parquet file in the object store and also used for its file location?

NGA-TRAN (approved):
LGTM. Nice set of tests

iox_catalog/migrations/20211229171744_initial_schema.sql
pauldix: yeah, it's used in the file location in object store. It will get generated by the ingester before saving any record to the catalog

PR #3493, opened 2022-01-19
feat: add sequencer get_by_kafka_topic_id_and_partition to catalog

pauldix:
Also update ingester to get sequencer record and not attempt to create


NGA-TRAN (approved):
LGTM

PR #3499, opened 2022-01-20
feat(router2): shard writes/deletes into write buffer

domodwyer:
This PR adds the logic to glue together a `Sharder` (#3488) and a `WriteBuffer` implementation.

---

* feat: Sequencer wrapper (28b340f4)

      This type wraps an underlying WriteBufferWriter implementation, tagging it
      with a sequencer ID it should use when enqueuing operations to the buffer.

* feat: mock sharder (ed3a2eba)

      Implements a mock Sharder impl that returns pre-configured responses to 
      shard(), and captures the input to the call.

* feat: sharded write buffer (e953f649)

      Implements sharding of ops into an underlying WriteBuffer.

      Writes are sharded by some abstract Sharder impl, collated per shard to 
      maximise the size of each op (and therefore compression efficiency), converted
      into a DML operation and then enqueued in parallel to the underlying
      WriteBuffer implementation.

      Deletes are modelled as being mapped to a single write buffer shard, which is
      the case while we support sharding based on the table & namespace only.
      Deletes will be extended to support (potentially) multiple shards when column
      overrides are implemented.


crepererum (commented):
Since the actual write buffer is a runtime, not a compile-time configuration, IOx doesn't use a type parameter for `WriteBufferWriting` but rather goes for `Arc<dyn WriteBufferWriting>`, which is also what the config system provides you:

https://github.com/influxdata/influxdb_iox/blob/56abfba0f0e5f45274154d5b2938cde85b4c85d8/write_buffer/src/config.rs#L85-L91

domodwyer:
Ah yeah I forgot it's not just kafka! Good spot 👍

lukebond (approved):
looks nice to me :)

pauldix (approved):

router2/src/sharded_write_buffer.rs
```diff
@@ -0,0 +1,435 @@
...
+            .any(|v| v.table_name == "bananas" && v.payload.rows() == 1));
+        assert!(calls
+            .iter()
+            .any(|v| v.table_name == "platanos" && v.payload.rows() == 1));
+        assert!(calls
+            .iter()
+            .any(|v| v.table_name == "another" && v.payload.rows() == 1));
+        assert!(calls
+            .iter()
+            .any(|v| v.table_name == "table" && v.payload.rows() == 1));
```
NGA-TRAN, line 323: 👍

router2/src/sharded_write_buffer.rs
```diff
@@ -0,0 +1,435 @@
...
+
+        // Configure the first shard to write to one write buffer
+        let write_buffer1 = init_write_buffer(1);
+        let write_buffer1_state = write_buffer1.state();
+        let shard1 = Arc::new(Sequencer::new(0, Arc::new(write_buffer1)));
+
+        // Configure the second shard to write to a write buffer that always fails
+        let write_buffer2 = init_write_buffer(1);
+        // Non-existant sequencer ID to trigger an error.
+        let shard2 = Arc::new(Sequencer::new(13, Arc::new(write_buffer2)));
```
NGA-TRAN, line 365: 👍

router2/src/sharded_write_buffer.rs
```diff
@@ -0,0 +1,435 @@
...
+        // Call the ShardedWriteBuffer and drive the test
+        let ns = DatabaseName::new("namespace").unwrap();
+        w.delete(ns, TABLE, predicate.clone(), None)
+            .await
+            .expect("delete failed");
+
+        // Assert the sharder saw all the tables
+        let calls = sharder.calls();
+        assert_matches!(calls.as_slice(), [MockSharderCall{table_name, ..}] => {
+            assert_eq!(table_name, TABLE);
```
NGA-TRAN, line 418: 👍

NGA-TRAN (approved):
👍

Issue #3462, opened 2022-01-13
InfluxQL: unsupported regex (TestServer_Query_Tags): Internal(\"error compiling regex pattern:


alamb:
Causing the InfluxQL `TestServer_Query_Tags` test to fail

```
    --- FAIL: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping) (0.03s)
    --- FAIL: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping)#01 (0.02s)
```

__Steps to reproduce:__
Use data like:

```
		fmt.Sprintf("cpu,host=server01 value=100,core=4 %d", now.UnixNano()),
		fmt.Sprintf("cpu,host=server02 value=50,core=2 %d", now.Add(1).UnixNano()),

		fmt.Sprintf("cpu1,host=server01,region=us-west value=100 %d", mustParseTime(time.RFC3339Nano, "2015-02-28T01:03:36.703820946Z").UnixNano()),
		fmt.Sprintf("cpu1,host=server02 value=200 %d", mustParseTime(time.RFC3339Nano, "2010-02-28T01:03:37.703820946Z").UnixNano()),
		fmt.Sprintf("cpu1,host=server03 value=300 %d", mustParseTime(time.RFC3339Nano, "2012-02-28T01:03:38.703820946Z").UnixNano()),

		fmt.Sprintf("cpu2,host=server01 value=100 %d", mustParseTime(time.RFC3339Nano, "2015-02-28T01:03:36.703820946Z").UnixNano()),
		fmt.Sprintf("cpu2 value=200 %d", mustParseTime(time.RFC3339Nano, "2012-02-28T01:03:38.703820946Z").UnixNano()),

		fmt.Sprintf("cpu3,company=acme01 value=100 %d", mustParseTime(time.RFC3339Nano, "2015-02-28T01:03:36.703820946Z").UnixNano()),
		fmt.Sprintf("cpu3 value=200 %d", mustParseTime(time.RFC3339Nano, "2012-02-28T01:03:38.703820946Z").UnixNano()),

		fmt.Sprintf("status_code,url=http://www.example.com value=404 %d", mustParseTime(time.RFC3339Nano, "2015-07-22T08:13:54.929026672Z").UnixNano()),

		fmt.Sprintf("status_code,url=https://influxdb.com value=418 %d", mustParseTime(time.RFC3339Nano, "2015-07-22T09:52:24.914395083Z").UnixNano()),

```

[...]

PR #3502, opened 2022-01-20
fix(InfluxQL): Support RegEx with escape sequences not supported by Rust regex

alamb:
# Rationale
Fixes https://github.com/influxdata/influxdb_iox/issues/3462

The issue is that `\:` is not a valid escape sequence for regular expressions in Rust

When I change the expression from `https\://influxdb\.com` to `https://influxdb\.com` it works as expected

The specific error is:

```
error: unrecognized escape sequence
```

Which comes from regex-syntax at:
https://docs.rs/regex-syntax/0.6.25/src/regex_syntax/ast/parse.rs.html#1445-1538

Specifically, ':' is not a metacharacter as defined in: https://docs.rs/regex-syntax/latest/regex_syntax/fn.is_meta_character.html (and [source](https://docs.rs/regex-syntax/latest/src/regex_syntax/lib.rs.html#207-213))

```rust
pub fn is_meta_character(c: char) -> bool {
    match c {
        '\\' | '.' | '+' | '*' | '?' | '(' | ')' | '|' | '[' | ']' | '{'
        | '}' | '^' | '$' | '#' | '&' | '-' | '~' => true,
        _ => false,
    }
}
```

# Changes:
1. Translate all `\<char>` sequences that are not valid meta characters to their unescaped form

# InfluxQL test suite improvement:

Before this PR:
```shell
go test -v ./influxql/v1tests -count 1  -run 'TestServer_Query_Tags/single_field_\(regex_tag_match_with_escaping\)'
...
--- FAIL: TestServer_Query_Tags (1.01s)
    --- FAIL: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping) (0.10s)
    --- FAIL: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping)#01 (0.05s)
FAIL
```
After this PR:

```shell
--- PASS: TestServer_Query_Tags (0.47s)
    --- PASS: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping) (0.04s)
    --- PASS: TestServer_Query_Tags/single_field_(regex_tag_match_with_escaping)#01 (0.03s)

```

# Note

I am not super happy with this code: it seems overly tricky and I worry about edge cases, and it took me far longer to get this than I would like to admit. 

That being said, I don't really have any other ideas for alternate implementations 🤷  So any suggestions welcome


query_tests/src/influxrpc/read_filter.rs
```diff
@@ -477,7 +477,7 @@ async fn test_read_filter_data_pred_using_regex_match() {
         .build();
 
     let expected_results = vec![
-    "Series tags={_measurement=h2o, city=LA, state=CA, _field=temp}\n  FloatPoints timestamps: [200], values: [90.0]",
+        "Series tags={_measurement=h2o, city=LA, state=CA, _field=temp}\n  FloatPoints timestamps: [200], values: [90.0]",
```
alamb, line 480: driveby whitespace change

crepererum (approved):
Mmh, I'm not really sure why `regex-syntax` is so strict here. If they introduce new special characters, it's a breaking change anyways because if will change how the unescaped variant is interpreted. So I agree that this is probably the quick fix.

I wasn't sure at which parsing stage escaping is applied, esp. when considering other special characters like `[abc]+` (what about chars within the character group?). Playing around a bit w/ https://regex101.com/ it seems that escaping always goes first, which is good. So I'll approve this solution here.

However, maybe we (you, I, someone else from the team) can contribute this back to upstream, even if it's only an opt-in flag of [`RegexBuilder`](https://docs.rs/regex/1.5.4/regex/struct.RegexBuilder.html).

alamb:
> Mmh, I'm not really sure why regex-syntax is so strict here. If they introduce new special characters, it's a breaking change anyways because if will change how the unescaped variant is interpreted. So I agree that this is probably the quick fix.

> However, maybe we (you, I, someone else from the team) can contribute this back to upstream, even if it's only an opt-in flag of RegexBuilder.

I poked around some more and I found that this is a known issue / request in regex from 2008:  https://github.com/rust-lang/regex/issues/501 (actually there is another issue exactly for the specific `\:` issue https://github.com/rust-lang/regex/issues/727 😆 ) 

Given the discussion on that ticket it doesn't look like the rust implementation is likely to be changed anytime soon. Thus I think the best I can do here is to add a reference to that ticket in IOx's code and move on

alamb:
Added comments / rationale in 713d4be0

PR #3505, opened 2022-01-21
refactor: Clean up the Catalog API

pauldix:
This updates the catalog API to make it easier to work with for consumers. I also found a bug in the MemCatalog implementation while refactoring the tests to work with the new API definition. Consumers will now be able to Arc wrap the catalog and use it across awaits.

### DIFF
--- a/ingester/src/server.rs
+++ b/ingester/src/server.rs
@@ -3,13 +3,13 @@
 
 use std::sync::Arc;
 
-use iox_catalog::interface::{KafkaPartition, KafkaTopic, KafkaTopicId, RepoCollection};
+use iox_catalog::interface::{Catalog, KafkaPartition, KafkaTopic, KafkaTopicId};
 
 /// The [`IngesterServer`] manages the lifecycle and contains all state for
 /// an `ingester` server instance.
 pub struct IngesterServer<'a, T>
 where
-    T: RepoCollection + Send + Sync,
+    T: Catalog,
 {
     /// Kafka Topic assigned to this ingester
     kafka_topic: KafkaTopic,
@@ -21,7 +21,7 @@ where
 
 impl<'a, T> IngesterServer<'a, T>
 where
-    T: RepoCollection + Send + Sync,
+    T: Catalog,
 {
     /// Initialize the Ingester
     pub fn new(topic: KafkaTopic, shard_ids: Vec<KafkaPartition>, catalog: &'a Arc<T>) -> Self {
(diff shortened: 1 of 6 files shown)


iox_catalog/src/interface.rs
```diff
@@ -247,32 +246,32 @@ impl std::fmt::Display for ParquetFileId {
...
 }
 
-/// Container that can return repos for each of the catalog data types.
+/// Trait that contains methods for working with the catalog
 #[async_trait]
-pub trait RepoCollection {
+pub trait Catalog: Send + Sync {
     /// repo for kafka topics
-    fn kafka_topic(&self) -> Arc<dyn KafkaTopicRepo + Sync + Send>;
+    fn kafka_topics(&self) -> &dyn KafkaTopicRepo;
```
crepererum, line 253: I like the usage of references here. If we ever want to add commits / transactions (to change multiple objects at the same time), this will be way easier because we could couple the transaction state to the lifetime of `Catalog` (probably renaming the type) and the repos could not outlive a transaction any longer.

crepererum (approved):

ingester/src/server.rs
```diff
@@ -3,13 +3,13 @@
...
 
-use iox_catalog::interface::{KafkaPartition, KafkaTopic, KafkaTopicId, RepoCollection};
+use iox_catalog::interface::{Catalog, KafkaPartition, KafkaTopic, KafkaTopicId};
 
 /// The [`IngesterServer`] manages the lifecycle and contains all state for
 /// an `ingester` server instance.
 pub struct IngesterServer<'a, T>
 where
-    T: RepoCollection + Send + Sync,
+    T: Catalog,
```
NGA-TRAN, line 12: I like the name `Catalog` interface. A lot better in this context

iox_catalog/src/interface.rs
```diff
@@ -881,61 +874,84 @@ pub(crate) mod test_helpers {
...
 
-        let tables = table_repo.list_by_namespace_id(namespace.id).await.unwrap();
+        let tables = catalog
+            .tables()
+            .list_by_namespace_id(namespace.id)
+            .await
+            .unwrap();
         assert_eq!(vec![t], tables);
+
+        // test we can create a table of the same name in a different namespace
```
NGA-TRAN, line 907: 👍

NGA-TRAN (approved):
I like this cleanup

PR #3506, opened 2022-01-21
feat: Add scaffolding for ingester server

pauldix:
This also includes stuff from #3505 so once that gets merged in this will be a little easier to review.

* Adds a new ingester command to start an ingester server
* Moves previous ingester server over to handler
* Skeleton for gRPC and HTTP handlers

### DIFF
--- a/influxdb_iox/src/commands/run/ingester.rs
+++ b/influxdb_iox/src/commands/run/ingester.rs
@@ -1,0 +1,121 @@
+//! Implementation of command line option for running ingester
+
+use std::sync::Arc;
+
+use crate::{
+    clap_blocks::run_config::RunConfig,
+    influxdb_ioxd::{
+        self,
+        server_type::{
+            common_state::{CommonServerState, CommonServerStateError},
+            ingester::IngesterServerType,
+        },
+    },
+};
+use ingester::handler::IngestHandlerImpl;
+use ingester::server::grpc::GrpcDelegate;
+use ingester::server::{http::HttpDelegate, IngesterServer};
+use iox_catalog::interface::{Catalog, KafkaPartition};
+use iox_catalog::postgres::PostgresCatalog;
+use observability_deps::tracing::*;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Run: {0}")]
+    Run(#[from] influxdb_ioxd::Error),
+
+    #[error("Cannot setup server: {0}")]
+    Setup(#[from] crate::influxdb_ioxd::server_type::database::setup::Error),
+
+    #[error("Invalid config: {0}")]
+    InvalidConfig(#[from] CommonServerStateError),
+
+    #[error("Catalog error: {0}")]
+    Catalog(#[from] iox_catalog::interface::Error),
+
+    #[error("Kafka topic {0} not found in the catalog")]
+    KafkaTopicNotFound(String),
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug, clap::Parser)]
+#[clap(
+    name = "run",
+    about = "Runs in ingester mode",
+    long_about = "Run the IOx ingester server.\n\nThe configuration options below can be \
+    set either with the command line flags or with the specified environment \
+    variable. If there is a file named '.env' in the current working directory, \
+    it is sourced before loading the configuration.
+Configuration is loaded from the following sources (highest precedence first):
+        - command line arguments
+        - user set environment variables
+        - .env file contents
+        - pre-configured default values"
+)]
+pub struct Config {
+    #[clap(flatten)]
+    pub(crate) run_config: RunConfig,
+
+    /// Postgres connection string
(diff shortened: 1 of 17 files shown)


influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -0,0 +1,121 @@
...
+use observability_deps::tracing::*;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("Run: {0}")]
+    Run(#[from] influxdb_ioxd::Error),
+
+    #[error("Cannot setup server: {0}")]
+    Setup(#[from] crate::influxdb_ioxd::server_type::database::setup::Error),
```
domodwyer, line 29: This looks unused - it references the "database" mode of running - I just realised I left it in the `router2` command file too 🤦

influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -0,0 +1,121 @@
...
+        .await?
+    {
+        Some(k) => k,
+        None => return Err(Error::KafkaTopicNotFound(config.kafka_topic)),
+    };
+    let kafka_partitions: Vec<_> = (config.kafka_partition_range_start
+        ..config.kafka_partition_range_end)
+        .map(KafkaPartition::new)
+        .collect();
+
```
domodwyer, line 106: It might be worth checking `kafka_partitions.len() > 0` as a sanity check - could go inside the `IngestHandlerImpl` when it comes to fleshing it out too.

If the caller put `kafka_partition_range_start: 10, kafka_partition_range_end: 0` you'd get [0 partitions](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6e3f2da749984639dcb99077efc3e00) and it's not super obvious at a glance the range limits are the wrong way round.

iox_catalog/src/interface.rs
```diff
@@ -247,32 +246,32 @@ impl std::fmt::Display for ParquetFileId {
     }
 }
 
-/// Container that can return repos for each of the catalog data types.
+/// Trait that contains methods for working with the catalog
 #[async_trait]
-pub trait RepoCollection {
+pub trait Catalog: Send + Sync {
```
domodwyer, line 251: nice 🎉

influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    let kafka_topic = match catalog
+        .kafka_topics()
+        .get_by_name(&config.kafka_topic)
+        .await?
+    {
+        Some(k) => k,
+        None => return Err(Error::KafkaTopicNotFound(config.kafka_topic)),
+    };
```
domodwyer, line 101: A nice, concise way of expressing this is to use the [`.ok_or(MyError)`] combinator to map the `Option<KafkaTopic>` we have here into a `Result` we can use `?` on:

```
Some(KafkaTopic) -> becomes -> Ok(KafkaTopic)
None             -> becomes -> Err(MyError)
```

Which means you can use `?` to return the error early:

```suggestion
    let kafka_topic = catalog
        .kafka_topics()
        .get_by_name(&config.kafka_topic)
        .await?
        .ok_or(Error::KafkaTopicNotFound(config.kafka_topic))?;
```

This pattern can be applied whenever you find yourself writing `Some(v) => v / None => Err(...)` 👍 

[`.ok_or(MyError)`]: https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

ingester/src/handler.rs
```diff
@@ -0,0 +1,62 @@
...
+    kafka_partitions: Vec<KafkaPartition>,
+    /// Catalog of this ingester
+    pub iox_catalog: Arc<dyn Catalog>,
+}
+
+impl std::fmt::Debug for IngestHandlerImpl {
+    fn fmt(&self, _f: &mut Formatter<'_>) -> std::fmt::Result {
+        todo!()
+    }
+}
```
domodwyer, line 25: If you were to change the Catalog trait from:

```rust
pub trait Catalog: Send + Sync { 
```

to require "all catalog implementations must be debug printable" by adding a `Debug` bound:

```rust
pub trait Catalog: Send + Sync + Debug { 
```

You would be able to remove this manual debug impl for `IngestHandlerImpl` and it would let you `#[derive(Debug)]` for an easier life 🏖️

domodwyer (approved):
ITS ALIVE :shipit:

influxdb_iox/src/commands/run/ingester.rs
NGA-TRAN: Thanks for the explanation, @domodwyer . Easy to understand for my future use, too

ingester/src/handler.rs
```diff
@@ -0,0 +1,62 @@
+//! Ingest handler
```
NGA-TRAN, line 1: I like you put ingester's handler into this separate file

NGA-TRAN (approved):
Nice

pauldix:
Will fix up the commented things in a follow up PR as I have one in the works that builds on this. Thanks!

PR #3516, opened 2022-01-24
feat: chunk pruning metrics

tustvold:
We've observed what appear to be chunk pruning performance issues when there are a large number of chunks, in particular it appears to be taking ~50ms for each additional predicate in a catalog with around 3000 chunks.

This PR therefore adds two set of metrics that will help prove this hypothesis

* query_access_catalog_snapshot_total and query_access_catalog_snapshot_duration_seconds_total - these will identify the time spent in the catalog snapshotting chunks and performing the first level of filtering
* query_access_prune_total and query_access_prune_duration_seconds_total - these will identify the time spent in the Datafusion predicate pruning logic

I suspect there may be better names, for these - suggestions welcome


alamb (approved):

db/src/access.rs
```diff
@@ -37,16 +38,57 @@ const QUERY_LOG_SIZE: usize = 100;
...
     /// Keyed by table name
     tables: Mutex<HashMap<Arc<str>, TableAccessMetrics>>,
 }
 
 impl AccessMetrics {
     fn new(registry: &metric::Registry, db_name: Arc<str>) -> Self {
+        let attributes = Attributes::from([("db_name", db_name.to_string().into())]);
+        let catalog_snapshot_duration = registry
+            .register_metric::<DurationCounter>(
+                "query_access_catalog_snapshot_duration",
```
e-dard, line 66: Is this the full metric name? I don't see how this maps to `query_access_catalog_snapshot_duration_seconds_total`.

e-dard (approved):
LGTM, just a question about naming in the code.

db/src/access.rs
tustvold: The unit and metric type suffix are added automatically to force consistency

Issue #3507, opened 2022-01-24
IOx makes temporary files for each query


alamb:
https://github.com/apache/arrow-datafusion/issues/1637 in DataFusion adds a memory manager / disk manager (🎉 ) but as it is currently implemented / integrated into IOx this means each query creates a temp file, which in heavily loaded databases could be quite expensive. 

__Steps to reproduce:__
Run queries in IOx

__Expected behaviour:__
Directory is only created when a query actually needs it (which these days is practically speaking, never)

__Actual behaviour:__
Directory is created on each query

Upstream related issue: https://github.com/apache/arrow-datafusion/issues/1636

alamb:
🤔  this seems to still be happening. I see logs like this locally (and remoted):

```
2022-01-27T18:45:40.201495Z  INFO datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
2022-01-27T18:45:40.202059Z DEBUG datafusion::execution::disk_manager: Created directory TempDir { path: "/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmpobQs5k" } as DataFusion working directory    
2022-01-27T18:45:40.202146Z  INFO datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
2022-01-27T18:45:40.202841Z DEBUG datafusion::execution::disk_manager: Created directory TempDir { path: "/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmpT4r291" } as DataFusion working directory    
2022-01-27T18:45:40.202904Z  INFO datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
2022-01-27T18:45:40.203419Z DEBUG datafusion::execution::disk_manager: Created directory TempDir { path: "/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmphImBiK" } as DataFusion working directory    
2022-01-27T18:45:40.203485Z  INFO datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
2022-01-27T18:45:40.203942Z DEBUG datafusion::execution::disk_manager: Created directory TempDir { path: "/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmpRdqi4J" } as DataFusion working directory    
```

PR #3517, opened 2022-01-24
chore: Update datafusion, upgrade to arrow/parqet/arrow-flight 8.0.0

alamb:
~draft, builds on https://github.com/influxdata/influxdb_iox/pull/3525~

~NOTE: The tests fail on this branch -- looks related to https://github.com/apache/arrow-datafusion/pull/1596~

# Rationale:
1. Prepare the way for getting https://github.com/apache/arrow-datafusion/pull/1668 merged asap so we can fix https://github.com/influxdata/influxdb_iox/issues/3507
2. Upgrade to arrow 8.0

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/63d24bf5bd0d34e84ccc10c9bc51b5bb754017b1
2. Update to arrow 8.0.0
2. Updates for API changes in DataFusion

# Blocking
- [x] https://github.com/apache/arrow-datafusion/issues/1676
- [x] https://github.com/apache/arrow-datafusion/issues/1686


alamb:
There is another problem: `SortPreservingMergeStream` appears not to be stable after https://github.com/apache/arrow-datafusion/pull/1596 was merged

Specifically, the previous implementation of  `SortPreservingMergeSteam` would always pick inputs with a lower 'stream index' when the sort keys were tied. 

Thus,  if stream 1 and stream 2 had the same record values, it was guaranteed that the row from stream1 came out before stream 2 which IOx relies on to correctly deduplicate data. The new `SortPreservingMergeStream` implementation does not seem to have this property.

alamb:
Filed https://github.com/apache/arrow-datafusion/pull/1687

Cargo.lock
```diff
@@ -1717,7 +1726,7 @@ dependencies = [
  "flate2",
  "futures",
  "generated_types",
- "hashbrown",
+ "hashbrown 0.11.2",
```
alamb, line 1729: I will make a follow on PR to upgrade to hashbrown 0.12 in IOx

query/src/provider.rs
```diff
@@ -2280,7 +2280,9 @@ mod test {
             "| 10        | MT   | 1970-01-01T00:00:00.000007Z    |",
             "+-----------+------+--------------------------------+",
         ];
-        assert_batches_eq!(&expected, &batch);
+        // Since output is partially sorted, allow order to vary and
+        // test to still pass
+        assert_batches_sorted_eq!(&expected, &batch);
```
alamb, line 2285: this change is due to the slightly different merge implementation in DataFusion

query/src/pruning.rs
```diff
@@ -142,6 +142,12 @@ impl<'a> PruningStatistics for ChunkMetaStats<'a> {
         // on a single chunk at a time
         1
     }
+
+    fn null_counts(&self, column: &Column) -> Option<ArrayRef> {
```
alamb, line 146: DataFusion pruning has started to learn about `IS NULL` predicates

workspace-hack/Cargo.toml
```diff
@@ -37,13 +38,14 @@ num-bigint = { version = "0.4", features = ["std"] }
...
 once_cell = { version = "1", features = ["alloc", "parking_lot", "race", "std"] }
+parquet = { version = "8", features = ["arrow", "base64", "brotli", "experimental", "flate2", "lz4", "snap", "zstd"] }
 rand = { version = "0.8", features = ["alloc", "getrandom", "libc", "rand_chacha", "rand_hc", "small_rng", "std", "std_rng"] }
 regex = { version = "1", features = ["aho-corasick", "memchr", "perf", "perf-cache", "perf-dfa", "perf-inline", "perf-literal", "std", "unicode", "unicode-age", "unicode-bool", "unicode-case", "unicode-gencat", "unicode-perl", "unicode-script", "unicode-segment"] }
 regex-automata = { version = "0.1", features = ["regex-syntax", "std"] }
 regex-syntax = { version = "0.6", features = ["unicode", "unicode-age", "unicode-bool", "unicode-case", "unicode-gencat", "unicode-perl", "unicode-script", "unicode-segment"] }
 reqwest = { version = "0.11", features = ["__tls", "default-tls", "hyper-tls", "json", "native-tls-crate", "serde_json", "tokio-native-tls"] }
 serde = { version = "1", features = ["derive", "rc", "serde_derive", "std"] }
-serde_json = { version = "1", features = ["arbitrary_precision", "indexmap", "preserve_order", "raw_value", "std"] }
```
alamb, line 46: no more `arbitrary precision` 🎉  -- cc @tustvold

crepererum (approved):

Cargo.lock
alamb: https://github.com/influxdata/influxdb_iox/pull/3551

PR #3519, opened 2022-01-25
refactor: have the deduplicate work without chunk statistics

NGA-TRAN:


NGA-TRAN:
This PR does:
  . Add functions to let providers know this chunks do not have statistics and should not rely on it
  . Add tests to compact non-duplcate  and duplicate data. The test with different columns in different record batches does not pass yet. Will work on this next.

Not include in this PR and will be done after the bug above: Tests that apply delete predicates

tustvold:
I'm a bit confused by this change, why do we have chunks without statistics? `MutableBatch` already has statistics tracking built-in?

query/src/lib.rs
```diff
@@ -236,6 +240,44 @@ where
...
+
+    true
+}
+
+pub fn compute_sort_key_for_chunks<'a, C>(schema: &'a Schema, chunks: &'a [C]) -> SortKey<'a>
+where
+    C: QueryChunkMeta,
+{
+    if !chunks_have_stats(chunks) {
+        // chunks have not enough stats, return its  pk
```
tustvold, line 271: Won't this sort order therefore be completely arbitrary? We really want to be sorting lower to higher cardinality...

query/src/lib.rs
```diff
@@ -236,6 +240,44 @@ where
...
+        self.as_ref().has_stats()
+    }
+}
+
+/// return true if all the chunks inlcude statistics
+pub fn chunks_have_stats<C>(chunks: &[C]) -> bool
+where
+    C: QueryChunkMeta,
+{
+    // If at least one of the provided chunk cannot provide stats,
```
tustvold, line 254: This is kind of sad, given the stats are already computed it seems a shame to not just expose them

NGA-TRAN: Agree

query/src/lib.rs
NGA-TRAN: Arbitrary now. We can make it deterministic by sorting by name. 
My goal is to keep this sorted by their cardinality as we did before. We measured performance before and it  was a lot faster using lowest cardinality first. But  let us measure performance on this new no-stats  on customer flow of data and decide

NGA-TRAN:
BTW, I just discovered I put `query` crate in `dependencies`, I will move it and appropriate ones to `dev dependencies`

query/src/lib.rs
```diff
@@ -52,6 +52,10 @@ pub trait QueryChunkMeta: Sized {
         !self.delete_predicates().is_empty()
     }
 
+    /// return true of this chunk has summary which
```
alamb, line 55: what would you think about changing the API of `summary` to return an `Option` rather than a separate function `has_stats`?

My rationale is that then the compiler can help ensure that all places that call `summary()` can properly handle when a chunk doesn't have any.

So instead of 
```rust
    /// Return a reference to the summary of the data
    fn summary(&self) -> &TableSummary;
```

Do something like
```rust
    /// Return a reference to the summary of the data, if available
    fn summary(&self) -> Option<&TableSummary>;
```

query/src/lib.rs
```diff
@@ -236,6 +240,44 @@ where
...
+
+/// return true if all the chunks inlcude statistics
+pub fn chunks_have_stats<C>(chunks: &[C]) -> bool
+where
+    C: QueryChunkMeta,
+{
+    // If at least one of the provided chunk cannot provide stats,
+    // do not need to compute potential duplicates. We will treat
+    // as all of them have duplicates
+    for chunk in chunks {
```
alamb, line 257: I think you could express this in idiomatic rust like

```rust
chunks.iter().all(|c| c.has_stats()))
```

But I don't think there is any practical difference

alamb (approved):

query/src/lib.rs
NGA-TRAN: I like this idea. Let me change it

NGA-TRAN:
@alamb You have approved this PR but since your suggestion of changing the summary function to return an Option, many more files have been changed. Can you skim quickly to see if these are what you expected?

query/src/lib.rs
NGA-TRAN: The PK is already sorted lexicographically but time column always last

query/src/lib.rs
NGA-TRAN: Done

alamb (approved):
I think it looks good -- obviously the various places that will now panic if there is no summary is not ideal but I suspect the lack of statistics is a temporary state

PR #3520, opened 2022-01-25
feat: buffer writes when writing to RSKafka

crepererum:
This will scale better than our current "create a Kafka request per incoming HTTP request" behavior.


write_buffer/src/rskafka/aggregator.rs
```diff
@@ -0,0 +1,310 @@
...
+    record::Record,
+};
+use schema::selection::Selection;
+use time::{Time, TimeProvider};
+use trace::{ctx::SpanContext, span::SpanRecorder, TraceCollector};
+
+use crate::codec::{ContentType, IoxHeaders};
+
+#[derive(Debug)]
+struct WriteAggregator {
```
NGA-TRAN, line 19: I am a bit confused about the name `Aggregator` because the work is mostly on a Span rather than actual aggregation but it seems like this word is used somewhere and you just reuse it. Adding comments describing this struct would be helpful then.

NGA-TRAN (approved):
LGTM

write_buffer/src/rskafka/aggregator.rs
crepererum: done

PR #3526, opened 2022-01-25
feat: support line protocol precision parameter (#3522)

### LINKED ISSUE #3522: Write Path Precision Parameter
The IOx write path should understand and interpret the [precision](https://docs.influxdata.com/influxdb/v2.0/api/#tag/Write) parameter in line protocol writes

tustvold:
This closes #3522 for the current write path. I suspect the new write path will also need similar work. FYI @domodwyer


domodwyer:
Thanks for the heads up 👍

domodwyer:
@crepererum regarding the duplication of HTTP handlers, the router2 handlers use different types, so duplicating them seemed cleaner than changing all the existing code until we've decided this is the archtecture path we want - they're realtively simple anwyay 👍 

It also has the nice side effect of keeping the router's HTTP handlers in the router crate!

PR #3528, opened 2022-01-25
feat: initialise ShardedWriteBuffer

domodwyer:
Adds arguments to the `router2` CLI mode to accept write buffer config and initialise the write buffer layers.

The source of truth for "how many sequencers should the router shard into" is the number of kafka partitions reported by the write buffer/kafka itself.

---

* feat: initialise ShardedWriteBuffer (633cb357)

      Initialises a ShardedWriteBuffer for the hard-coded "iox_shared" topic.

      Adds the following CLI flags:

          * --write-buffer: type of buffer [kafka, rskafka, file]
          * --write-buffer-addr: write buffer endpoint address

      The server uses these config options to initialise the appropriate write 
      buffer backend, and configure the TableNamespaceSharder to shard operations
      over the set of sequencers exposed by the write buffer.


lukebond (approved):

influxdb_iox/src/commands/run/router2.rs
```diff
@@ -51,19 +57,73 @@ Configuration is loaded from the following sources (highest precedence first):
...
+async fn init_write_buffer(
+    config: &Config,
+    topic: &str,
+    metrics: Arc<metric::Registry>,
+) -> Result<ShardedWriteBuffer<TableNamespaceSharder<Sequencer>>> {
+    let write_buffer_config = WriteBufferConnection {
+        type_: config.write_buffer_type.clone(),
+        connection: config.write_buffer_connection_string.clone(),
+        connection_config: Default::default(),
+        creation_config: None,
```
crepererum, line 103: If you set this to `None` (which is valid), the write buffer will NOT auto-create the Kafka topic. You can set it to an empty map though if you want and get auto-creation for free (this might be simpler for the MVP phase).

crepererum (approved):

PR #3531, opened 2022-01-25
feat: Add db_name/namespace to DmlWrite and DmlDelete

pauldix:
This is required for the new ingester to be able to work with the write buffer. The protobuf that gets serialized over Kafka already includes the database name, it just wasn't getting carried through to the marshaled Dml operation.


crepererum (changes requested):
Please adjust

https://github.com/influxdata/influxdb_iox/blob/52866fe6a90ea372957a29ec4c604552824b9ee6/dml/src/lib.rs#L426

To check the names as well. We should probably implement a proper `PartialEq` at some point.

dml/src/lib.rs
```diff
@@ -177,6 +177,8 @@ impl From<DmlDelete> for DmlOperation {
 /// A collection of writes to potentially multiple tables within the same database
 #[derive(Debug, Clone)]
 pub struct DmlWrite {
+    /// The namespace being written to
+    namespace: String,
```
domodwyer, line 181: You'll need a getter to be able to read this back out the struct (same with deletes) 👍

influxdb_iox/src/influxdb_ioxd/http/dml.rs
```diff
@@ -239,7 +239,7 @@ pub trait HttpDrivenDml: ServerType {
             "inserting lines into database",
         );
 
-        let write = DmlWrite::new(tables, DmlMeta::unsequenced(span_ctx));
+        let write = DmlWrite::new(db_name.as_str(), tables, DmlMeta::unsequenced(span_ctx));
```
domodwyer, line 242: We can avoid converting the `DatabaseName` to a `str` and then [to a `String`](https://github.com/influxdata/influxdb_iox/blob/8180b007aec9cee33276a33a67de002f73d30032/dml/src/lib.rs#L224) straight after by using `&db_name`:

```suggestion
        let write = DmlWrite::new(&db_name, tables, DmlMeta::unsequenced(span_ctx));
```

This works because there is a `From<&DatabaseName> for String` [impl allowing the direct conversion](https://github.com/influxdata/influxdb_iox/blob/8180b007aec9cee33276a33a67de002f73d30032/data_types/src/database_name.rs#L88-L92).

Same thing for the delete path too 👍

router2/src/sharded_write_buffer.rs
```diff
@@ -99,7 +99,11 @@ where
         }
 
         let iter = collated.into_iter().map(|(sequencer, batch)| {
-            let dml = DmlWrite::new(batch, DmlMeta::unsequenced(span_ctx.clone()));
+            let dml = DmlWrite::new(
+                namespace.as_str(),
```
domodwyer, line 103:
```suggestion
                &namespace,
```

router2/src/sharded_write_buffer.rs
```diff
@@ -134,6 +138,7 @@ where
         trace!(sequencer_id=%sequencer.id(), %table_name, %namespace, "routing delete to shard");
 
         let dml = DmlDelete::new(
+            namespace.as_str(),
```
domodwyer, line 141:
```suggestion
                &namespace,
```

domodwyer (approved):
Makes sense to me!

pauldix:
Ok, fixed up and pushed

pauldix:
Ok wow, was a bit more of a pain than I thought to fix all the testing machinery. Hopefully working now

write_buffer/src/core.rs
```diff
@@ -388,9 +437,30 @@ pub mod test_utils {
         let sequencer_id_1 = set_pop_first(&mut sequencer_ids_1).unwrap();
         let sequencer_id_2 = set_pop_first(&mut sequencer_ids_1).unwrap();
 
-        let w_east_1 = write(&writer_1, entry_east_1, sequencer_id_1, None).await;
-        let w_west_1 = write(&writer_1, entry_west_1, sequencer_id_2, None).await;
-        let w_east_2 = write(&writer_2, entry_east_2, sequencer_id_1, None).await;
+        let w_east_1 = write(
+            context.database_name(),
```
crepererum, line 441: I think for these tests it would be helpful if the name that is used to create the `DmlWrite` and the one that is used during write buffer creation are different because this is exactly the use case you're aiming for (e.g. create an `iox_shared` Kafka topic into which we serialize a write for the namespace `customer_1`).

I know this might sound a bit pedantic, but it's too easy to confuse these two names somewhere in the code.

With that I think `context.database_name()` is not actually required.

crepererum (changes requested):

crepererum:
I'm stealing this PR now.

PR #3532, opened 2022-01-25
feat: add per-partition tracing

tustvold:
With this change IOx now produces traces with per-partition granularity. This will provide better insight into what Datafusion is actually doing, what tasks are being spawned, what chunks scanned, etc...

I intend to build upon this, in particular adding more information about what the scan is actually doing, what storage it is fetching from, etc...

image


influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
```diff
@@ -448,7 +448,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_query_tracing() {
-        let collector = Arc::new(RingBufferTraceCollector::new(100));
+        let collector = Arc::new(RingBufferTraceCollector::new(1000));
```
tustvold, line 451: On my machine, Datafusion creates a LOT of partitions... We will likely need to keep an eye on this in production, but imo it is a useful piece of information if we have plans creating 50 partitions to fetch 10 rows each :laughing:

alamb: DataFusion creates a number of partitions equal to the number of cores... by default... This is likely to be "non ideal"

Perhaps we should consider limiting the target concurrency (maybe cap it at 8?) 


https://github.com/apache/arrow-datafusion/blob/master/datafusion/src/execution/context.rs#L872-L874

query/src/exec/query_tracing.rs
```diff
@@ -89,66 +94,124 @@ fn send_metrics_to_tracing(
...
+    // Get the timings of the parent operator
+    let parent_start_time = parent_span.start.unwrap_or(default_end_time);
+    let parent_end_time = parent_span.end.unwrap_or(default_end_time);
 
-    span.start = parent_span.start;
+    let mut operator_span = parent_span.child(plan_name.clone());
+    let mut operator_metrics = SpanMetrics {
+        output_rows: None,
+        elapsed_compute_nanos: None,
+    };
```
alamb, line 108: I think the code would be clearer if this was declared closer to where it was used (aka only in the `Some(Metrics)` case clause below

alamb (approved):

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
e-dard: We are typically running on 8–10 cores in production, so I think it's OK for now.

e-dard (approved):

query/src/exec/query_tracing.rs
tustvold: It is used outside that in order to append the aggregate metrics to the operator span that is generated

PR #3535, opened 2022-01-25
fix: merge record batches into one batch

NGA-TRAN:
Because our query provider only process one record batch per chunk, we need to merge all record batches in the QueryableChunk into one to have it work


ingester/src/query.rs
```diff
@@ -16,26 +16,31 @@ use predicate::{
...
+    // PaddNulls { source: arrow::error::ArrowError },
+    #[snafu(display("Internal error while concat record batches {}", source))]
+    ConcatBatches { source: arrow::error::ArrowError },
 }
 
 /// A specialized `Error` for Ingester's Query errors
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-// todo: move this function to a more appropriate crate
+// todo: Ask alamb whether these 2 functions should be moved to a more appropriate crate
```
NGA-TRAN, line 43: @alamb : a question for you here

ingester/src/query.rs
```diff
@@ -50,6 +55,57 @@ pub fn merge_record_batch_schemas(batches: &[Arc<RecordBatch>]) -> Arc<Schema> {
     Arc::new(merger.build())
 }
 
+/// Merge the record bacthes into one record batch
+/// and padd null values to columns that are not available in certain bacthes
+pub fn merge_record_batches(batches: Vec<Arc<RecordBatch>>) -> Result<Option<RecordBatch>> {
```
NGA-TRAN, line 60: We need this because currently our provider only works on Chunk with one RecordBatch

ingester/src/query.rs
alamb: https://github.com/influxdata/influxdb_iox/tree/main/arrow_util might be a possible one

It is interesting to note there is similar logic that (just) got merged into DataFusion in https://github.com/apache/arrow-datafusion/pull/1622

ingester/src/query.rs
alamb: Yeah this schema combination is done after the provider of the plan (so that we don't bother with record batches / rows that are going to be most/all filtered.

ingester/src/query.rs
```diff
@@ -50,6 +55,57 @@ pub fn merge_record_batch_schemas(batches: &[Arc<RecordBatch>]) -> Arc<Schema> {
...
+                .expect("A record batch should have been created")
+        })
+        .collect::<Vec<_>>();
+
+    // Combine batches
+    if batches.is_empty() {
+        return Ok(None);
+    }
+
+    Ok(Some(
```
alamb, line 104: It might be useful to write some unit tests here. 

Cases that I would suggest are (ripped off from https://github.com/apache/arrow-datafusion/pull/1622#discussion_r789853739):

1. columns in different orders in different batches (e.g. one batch has (`a`, `b`, `c`) columns and one has (`b`, `a`)
3.  Columns with the same name and different types (expect an error)
4. batches with different subsets of the data  (`a`, `b`) and (`b`, `c`) for example

ingester/src/compact.rs
```diff
@@ -471,31 +474,6 @@ mod tests {
         assert_batches_eq!(&expected, &[batch2.clone()]);
         batches.push(Arc::new(batch2));
 
-        // verify data from both batches
```
alamb, line 474: why was this test removed?

alamb (approved):
makes sense to me 👍

ingester/src/query.rs
NGA-TRAN: Yeah, there so similar work :)

NGA-TRAN: And I am moving these function to arrow_util

ingester/src/query.rs
NGA-TRAN: Adding these

ingester/src/query.rs
NGA-TRAN: @alamb : there is an issue here. I use `SchemaMerger` which belongs to `schema` crate to merge the schema. I do not think we should import `schema` in `arrow_util` crate. We may want move these to somewhere else or I have to rewrite the function merge schema to not use the schema merger. The former may be more reasonable

ingester/src/compact.rs
NGA-TRAN: It was removed because the verification was wrong. The output was very funny (batches with different columns are combined in a funny way) and not useful at all. The single batch verification above is good enough

ingester/src/query.rs
alamb: You might be able to keep the schema merging outside `arrow_util` by passing the merged schema in. Like

```rust
pub fn merge_record_batches(schema: SchemaRef, batches: Vec<Arc<RecordBatch>>) -> Result<Option<RecordBatch>> {
...
}
```
🤔

NGA-TRAN: Yeah, I think it will work and I will move the `merge_record_batch_schemas` function into the `schema` crate

PR #3538, opened 2022-01-26
feat: decouple read buffer row group size from Datafusion batch size

e-dard:
This PR introduces a Read Buffer `ChunkBuilder` that can be used to create a single Read Buffer chunk using one or more record batches.

The builder provides an API wherein the caller can provide a chunk single schema and then submit record batches as input. These record batches are buffered up until a desired row group size (in rows) is met, then they're concatenated and compressed into a Read Buffer row group. 

The `ChunkBuilder` aims to minimise in-memory footprint by compressing record batches into Read Buffer row groups as soon as enough data for a row group is submitted. That is, the builder will create row groups before the final `build()` method is called and a `Chunk` is yielded.

Effectively, then the memory foot print can be as large as the case where this builder didn't exist, and the Datafusion execution `BATCH_SIZE` was bumped to the Read Buffer's desired row group size.

When `build()` is called on the `ChunkBuilder` a single `Chunk` is yielded. If there were any pending record batches in the builder, these are compressed into a final row group. This means a row group may be smaller than the desired row group size but I'm not overly concerned about that at the moment.

To improve test coverage of this builder I have adapted all the Read Buffer tests using the existing API to use the new `ChunkBuilder`.

**Note** The obvious side effect of introducing this change (aside from the hopefully wonderful compression gains) are that setting `ROW_GROUP_MIN_ROWS` to something significantly larger than the typical Datafusion batch size (currently `1000`), is that we increase the upper bound on the size of data emitted from the Read Buffer as record batches. I am not convinced this is an immediate concern, but when it becomes an issue I plan to tackle it by teaching the Read Buffer to emit smaller batches of results, which I have tracked here https://github.com/influxdata/influxdb_iox/issues/3539


db/src/lifecycle.rs
```diff
@@ -357,29 +357,34 @@ fn collect_rub(
...
 
         adapted_stream
-            .try_for_each(|batch| {
-                chunk.upsert_table(batch);
-                future::ready(Ok(()))
-            })
+            .try_for_each(|batch| future::ready(chunk_builder.push_record_batch(batch)))
             .await?;
 
+        // The adapted stream may not hav yielded any record batches.
```
e-dard, line 377:
```suggestion
        // The adapted stream may not have yielded any record batches.
```

read_buffer/src/chunk.rs
```diff
@@ -16,6 +16,8 @@ use std::{
     convert::TryFrom,
 };
 
+const ROW_GROUP_MIN_ROWS: usize = 100_000;
```
e-dard, line 19:
```suggestion
// The desired minimum row group size, used as the default for the `ChunkBuilder`.
const ROW_GROUP_MIN_ROWS: usize = 100_000;
```

e-dard, line 19: `100_000` is typically a good row group size for the Read Buffer. I expect that we may want to revisit this depending on what we learn with current testing efforts. Further, we may wish to expose this as something configurable in the future.

read_buffer/src/chunk.rs
```diff
@@ -53,25 +64,23 @@ impl Chunk {
...
         let table_name = table_name.into();
         let row_group = record_batch_to_row_group(&table_name, table_data);
-        let storage_statistics = row_group.column_storage_statistics();
-
-        let table = Table::with_row_group(table_name, row_group);
 
-        metrics.update_column_storage_statistics(&storage_statistics);
-
-        Self { metrics, table }
+        Self::new_from_row_group(table_name, row_group, metrics)
```
e-dard, line 72: This is a drive-by refactor that unifies the logic of `new` and `new_from_row_group`

read_buffer/src/chunk.rs
```diff
@@ -669,45 +831,15 @@ mod test {
         }
     }
 
-    #[derive(Debug, Default)]
-    struct ChunkBuilder {
```
e-dard, line 673: This was previously a test-only helper. Now we use the public `ChunkBuilder` in the tests for more test coverage of the public API.

db/src/lifecycle.rs
```diff
@@ -357,29 +357,34 @@ fn collect_rub(
...
-        let mut chunk = read_buffer::RBChunk::new(table_name, first_batch, chunk_metrics);
+        let mut chunk_builder = read_buffer::RBChunkBuilder::new(table_name.as_ref(), schema)
+            .with_metrics(chunk_metrics);
 
         adapted_stream
-            .try_for_each(|batch| {
-                chunk.upsert_table(batch);
-                future::ready(Ok(()))
-            })
+            .try_for_each(|batch| future::ready(chunk_builder.push_record_batch(batch)))
```
alamb, line 374: 👍  this is a nice change

read_buffer/src/chunk.rs
alamb:
```suggestion
const DEFAULT_ROW_GROUP_MIN_ROWS: usize = 100_000;
```

read_buffer/src/chunk.rs
alamb: Looks like it also means that `new_from_row_group` will now update metrics as well? This seems like an improvement to me but I wanted to confirm the intended change in behavior

read_buffer/src/chunk.rs
```diff
@@ -352,6 +362,159 @@ impl std::fmt::Debug for Chunk {
...
+            current_rows: 0,
+            record_batches: vec![],
+            row_group_min_rows: ROW_GROUP_MIN_ROWS,
+            chunk_metrics: None,
+            row_groups: vec![],
+        }
+    }
+
+    /// Provide metrics for the built `Chunk`. If no metrics are provided then
+    /// unregistered metrics will be
```
alamb, line 395: If no metrics are provided, then no metrics are reported, right? I don't understand what the `unregistered metrics` bit is trying to say here

read_buffer/src/chunk.rs
```diff
@@ -352,6 +362,159 @@ impl std::fmt::Debug for Chunk {
...
+        if self.current_rows >= self.row_group_min_rows {
+            return self.snapshot_rowgroup();
+        }
+
+        Ok(())
+    }
+
+    // Force a rowgroup snapshot.
+    fn snapshot_rowgroup(&mut self) -> Result<(), ArrowError> {
+        let concat_batch =
```
alamb, line 430: FWIW this will effectively copy all the data to contiguous memory -- if this consumes too much maybe eventually we could allow a `RowGroup` to be created from a Vec<RecordBatch> directly and avoid the copy

read_buffer/src/chunk.rs
```diff
@@ -352,6 +362,159 @@ impl std::fmt::Debug for Chunk {
...
+        // chunk with.
+        if self.is_empty() {
+            return ChunkBuilderNoInputSnafu {
+                table_name: self.table_name.clone(),
+            }
+            .fail();
+            // Snapshot remaining batches to a row group (accepting that it may
+            // be smaller than desired)
+        } else if !self.record_batches.is_empty() {
+            self.row_groups.push(record_batch_to_row_group(
```
alamb, line 462: Perhaps this could just call `self.snaphot_row_group()` (perhaps after lowering the minimum row group threshold to 0) to save some repetition?

alamb (approved):
LGTM -- my only potential concern is the extra copy created by `concat_recordbatch` 👍 

This is a much nicer API

read_buffer/src/chunk.rs
e-dard: Yes that was the unification 😂

read_buffer/src/chunk.rs
e-dard: Ah that is not something I realised would happen. I will file a ticket. The only reason I used `concat` was that it was quicker than updating the internal logic of `RowGroup`. I can do in the future as a perf improvement.

e-dard: Added a `PERF` comment in code and field this https://github.com/influxdata/influxdb_iox/issues/3540

read_buffer/src/chunk.rs
e-dard: fixed.

PR #3541, opened 2022-01-26
DEMO (NOT FOR MERGING): DataFusion MergeStream Fix

alamb:
Built from https://github.com/influxdata/influxdb_iox/pull/3517

Demonstrates then when the following fixes from datafusion are applied to #3157  IOx passes its tests
1. https://github.com/apache/arrow-datafusion/pull/1687
2.  https://github.com/apache/arrow-datafusion/pull/1678


alamb:
served its purpose -- all done now

Issue #3436, opened 2022-01-11
Should empty Tags be modeled as Null or the empty string?


tustvold:
**Problem**

As reported by @alamb [here](https://github.com/influxdata/influxdb_iox/issues/3430#issuecomment-1009986314) the storagerpc layer does not appear to distinguish between an empty tag value and null tag value.

It would appear this is more general than just the storagerpc layer as `cpu,host= value=3` is considered invalid line protocol.

IOx, however, permits empty tag values and views these as distinct from null tag values. This applies to the binary write protocol, the mutable buffer, SQL, etc...

**Proposal**

We should clarify if tag columns should be explicitly nullable or not. If they should be explicitly nullable, i.e. support an empty string as distinct from NULL, we will likely need to rewrite storage and delete predicates. If they are not nullable we will need to alter the way null padding works in the ingest and schema adapter components.

I'm personally not sure which would be more implementation effort, part of me feels no NULLs will be simpler, but I'm honestly not sure.

FYI @pauldix @e-dard

alamb:
I think this issue can be handled at the influxql query layer, which is what I plan to do for https://github.com/influxdata/influxdb_iox/issues/3430 https://github.com/influxdata/influxdb_iox/issues/3434.

As @tustvold  notes above:
1. lineprotocol doesn't have the concept of nulls. 
2.  IOx maps lineprotocol to relational model by treating any column that wasn't specified as NULL
3. The Influxdb storage RPC queries expect that the value of such columns evaluate to the empty string `''` in terms of predicates

My plan is to handle the mapping of NULL --> `''` at the predicate evaluation layer.

alamb:
I have been thinking about this some more and I am somewhat worried about making missing tags look like the empty string at query time.

I will have a think but it may be we need to go with @tustvold 's suggestion of model missing tags as `''` rather than `NULL` if that is what the query model expects 🤔

PR #3543, opened 2022-01-26
Spike showing how to use '' on ingest rather than null for tags

alamb:
Spike for solution to https://github.com/influxdata/influxdb_iox/issues/3436 as described in https://docs.google.com/document/d/1Od7J5iu8eWm9CoUufwKUYhw0tpQpSMMWJ5n03SHGP54/edit#

High plan:
- [x] find basic test cases from other PRs
- [x] Change ingest path to use "" for missing tags
- [x] Change schema adapter stream
- [ ] Get basic test cases passing
- [ ] Verify InfluxQL tests pass
- [ ] Review damage elsewhere


alamb:
Planning on trying a different approach now... See doc

PR #3544, opened 2022-01-26
feat: Ingester persistence

### LINKED ISSUE #3512: Ingester persist to parquet
The `ingester` crate needs a new file called `persist.rs` that will contain copy-pasted code from existing parts of iox that do persisting of data to parquet files.

carols10cents:
Closes #3512.

@NGA-TRAN I'm happy to rebase and resolve conflicts after your PR; I put in a stub `IoxMetadata` type without any fields that should be the type you're adding instead. And I wasn't sure if this `persist` function would create that metadata or it would be passed in... so the arguments to these functions are probably wrong.

And this isn't actually called anywhere yet :)


ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+use snafu::{OptionExt, ResultExt, Snafu};
+use std::{
+    io::{Cursor, Seek, SeekFrom, Write},
+    sync::Arc,
+};
+use uuid::Uuid;
+
+/// IOx-specific metadata.
+#[allow(missing_copy_implementations)]
+pub struct IoxMetadata {}
```
NGA-TRAN, line 30: Since you need it, I think you may want to add the following info into it. I will rebase yours instead

```
pub struct IoxMetadata {
    /// The uuid used as the location of the parquet file in the OS.
    /// This uuid will later be used as the catalog's ParquetFileId 
    pub object_store_id: Uuid,

    /// Timestamp when this file was created.
    pub creation_timestamp: Time,

    /// sequencer id of the data
    pub sequencer_id: SequencerId,

    /// table id of the data
    pub table_id: TableId,

    /// table name of the data
    pub table_name: Arc<str>,

    /// partition id of the data
    pub partition_id: PartitionId,
    
    /// parittion key of the data
    pub partition_key: Arc<str>,

    /// Time of the first write of the data
    /// This is also the min value of the column `time`
    pub time_of_first_write: Time, 

    /// Time of the last write of the data
    /// This is also the max value of the column `time`
    pub time_of_last_write: Time,

    /// sequence number of the first write
    pub min_sequence_number: SequenceNumber,

    /// sequence number of the last write 
    pub max_sequence_number: SequenceNumber,
}
```

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+    sequencer_id: SequencerId,
+    partition_id: PartitionId,
+    metadata: IoxMetadata,
+    min_sequence_number: SequenceNumber,
+    max_sequence_number: SequenceNumber,
+    min_time: Timestamp,
+    max_time: Timestamp,
+    stream: SendableRecordBatchStream,
+    object_store: &ObjectStore,
+    parquet_file_repo: &dyn ParquetFileRepo,
```
NGA-TRAN, line 76: Because the IoxMetadata already includes many of these, they can be removed from this argument list

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+            table_id,
+            partition_id,
+            object_store_id,
+            min_sequence_number,
+            max_sequence_number,
+            min_time,
+            max_time,
+        )
+        .await
+        .context(CreatingParquetFileInCatalogSnafu)?;
```
NGA-TRAN, line 119: The catalog update will be done after this step in a transaction. This should not be in here. And because of that, you do not need the argument `parquet_file_repro` either

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+
+    let bytes = Bytes::from(data);
+
+    let path = parquet_file_object_store_path(
+        object_store,
+        namespace_id,
+        table_id,
+        sequencer_id,
+        partition_id,
+        object_store_id,
```
NGA-TRAN, line 99: These lines are part of the `metadata: IoxMetadata` which you already added in the `data`. I am not sure what is the best way to store this `metadata` in parquet. Maybe @crepererum are able to provide us some advice

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+        source: parquet::errors::ParquetError,
+    },
+
+    #[snafu(display("Error converting to vec[u8]: Nothing else should have a reference here"))]
+    WritingToMemWriter {},
+}
+
+/// Convert the given stream of RecordBatches to bytes
+async fn parquet_stream_to_bytes(
+    mut stream: SendableRecordBatchStream,
```
NGA-TRAN, line 176: Let us discuss in person tomorrow about this:
I am debating whether to keep this input as a `SendableRecordBatchStream` (my preference) or a `Vec<RecordBatch>`. The reason is we need to compute `min_time` and `max_time` for the `time` column of this data before writing it to parquet. Currently, this `min_time` and `max_time` are a part of the `IoxMetadata` and need to get computed before this step. Hence, I already get all batches of the stream and put them into a vector. If we keep doing so, this argument will be Vec<RecordBatch>. However, if we compute `min_time` and `max_time` inside this function, then this argument can be kept as a stream. See related `Point 1` and `Point 2` below.

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+    schema: SchemaRef,
+    metadata: IoxMetadata,
+) -> Result<Vec<u8>, ParquetStreamToBytesError> {
+    let metadata_bytes = metadata.to_protobuf().context(MetadataEncodeFailureSnafu)?;
+
+    let props = writer_props(&metadata_bytes);
+
+    let mem_writer = MemWriter::default();
+    {
+        let mut writer = ArrowWriter::try_new(mem_writer.clone(), schema, Some(props))
```
NGA-TRAN, line 186: Point 1: Since `min_time` and `max_time` are part of metada, and as mention above, they are not computed at this time yet, we need to store the `metadata` after  this.

ingester/src/persist.rs
```diff
@@ -0,0 +1,411 @@
...
+            .context(OpeningParquetWriterSnafu)?;
+        let mut no_stream_data = true;
+        while let Some(batch) = stream.next().await {
+            no_stream_data = false;
+            let batch = batch.context(ReadingStreamSnafu)?;
```
NGA-TRAN, line 191: The `min_time` and `max_time` can be computed here

NGA-TRAN (commented):
I have added @crepererum as a reviewer.
@carols10cents , @pauldix and @crepererum : let us have a quick discussion tomorrow (Thursday) because I think my comments do not convey everything.
I will try to commit things I am working right now and point you there to explain more what I mean

ingester/src/persist.rs
crepererum: Can you use the existing storage method because this one already takes care of that?

https://github.com/influxdata/influxdb_iox/blob/bc44d33108d9bfed54349935d1377cfe6005d404/parquet_file/src/storage.rs#L104-L109

or

https://github.com/influxdata/influxdb_iox/blob/bc44d33108d9bfed54349935d1377cfe6005d404/parquet_file/src/storage.rs#L140-L145

carols10cents: I copied this code from there, but that code uses `ChunkAddr` and the new ingester does not.

generated_types/protos/influxdata/iox/preserved_catalog/v1/parquet_metadata.proto
```diff
@@ -8,37 +8,59 @@ message IoxMetadata {
   // Metadata format version.
   uint32 version = 1;
 
+  // Object store ID. Used in the parquet filename. 16 bytes in big-endian order.
```
carols10cents, line 11: Because there's already a protobuf message named `IoxMetadata`, and because protobuf is made to be evolvable, I changed this proto definition rather than making a new one. Let me know if I should make a new one instead.

ingester/src/persist.rs
```diff
@@ -0,0 +1,403 @@
...
+        proto_msg.encode(&mut buf)?;
+
+        Ok(buf)
+    }
+}
+
+/// Current version for serialized metadata.
+///
+/// For breaking changes, this will change.
+pub const METADATA_VERSION: u32 = 100;
```
carols10cents, line 109: I kept the concept of a metadata version number from `parquet_file/src/metadata.rs` but started the number here as 100 rather than bumping the [current 10](https://github.com/influxdata/influxdb_iox/blob/4a96e522900850795946707a07ffa896885074bd/parquet_file/src/metadata.rs#L124) to 11.

Do we want to have the whole version-checking, catalog-wiping behavior we've had in IOx in the past? (If so, I need to port over the version-checking logic in the deserialization code, which also isn't currently in this PR) 

Or are we committing to backwards compatibility and not wiping the catalog in the new design, in which case this field is irrelevant and should be deprecated instead of having a value set here?

Or is this field irrelevant because I actually should have created a different proto definition that won't have this field?

ingester/src/persist.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// sequence number of the first write
+    pub min_sequence_number: SequenceNumber,
+
+    /// sequence number of the last write
+    pub max_sequence_number: SequenceNumber,
+}
+
+impl IoxMetadata {
+    /// Convert to protobuf v3 message.
+    fn to_protobuf(&self) -> std::result::Result<Vec<u8>, prost::EncodeError> {
```
carols10cents, line 81: You'll notice this struct only has a method for serialization, no deserialization yet. I sort of wanted to have deserialization here too so I could do some round-trip tests, but the ingester won't actually need to deserialize parquet files, correct? The queriers will need to? So should the deserialize logic only live in the queriers?

ingester/src/persist.rs
```diff
@@ -0,0 +1,403 @@
...
+///
+/// For breaking changes, this will change.
+pub const METADATA_VERSION: u32 = 100;
+
+/// File-level metadata key to store the IOx-specific data.
+///
+/// This will contain [`IoxMetadata`] serialized as base64-encoded [Protocol Buffers 3].
+///
+/// [Protocol Buffers 3]: https://developers.google.com/protocol-buffers/docs/proto3
+pub const METADATA_KEY: &str = "IOX:metadata";
```
carols10cents, line 116: This key's name is the same as [in `parquet_file/src/metadata.rs`](https://github.com/influxdata/influxdb_iox/blob/4a96e522900850795946707a07ffa896885074bd/parquet_file/src/metadata.rs#L131), is that ok or should it be something different?

generated_types/protos/influxdata/iox/preserved_catalog/v1/parquet_metadata.proto
carols10cents: Ugh, nevermind, I don't think this is going to work with what `prost` does, since it doesn't generate Rust fields for deprecated/reserved fields. I'm going to make a new proto definition.

ingester/src/persist.rs
NGA-TRAN: Yes, I think so

ingester/src/persist.rs
carols10cents: Ugh, nevermind, because I ended up making a different proto definition, I dropped this field. Let me know if this should still be present, and we can start it back at 1.

ingester/src/persist.rs
NGA-TRAN: The Querier and Compactor will read Parquet files

ingester/src/persist.rs
NGA-TRAN: I think it is fine to keep it the same but @crepererum know the answer better

NGA-TRAN (approved):
The code LGTM. 
Regarding the tests, have we had tests to verify the written content? if not, I think we may want to add tests to read back the data and verify them. But this can be done in the following PR.

ingester/src/persist.rs
```diff
@@ -0,0 +1,394 @@
...
+    ClosingParquetWriter {
+        source: parquet::errors::ParquetError,
+    },
+
+    #[snafu(display("Error converting to vec[u8]: Nothing else should have a reference here"))]
+    WritingToMemWriter {},
+}
+
+/// Convert the given metadata and RecordBatches to parquet file bytes
+async fn parquet_bytes(
```
crepererum, line 198: Why are we duplicating functionality of <https://github.com/influxdata/influxdb_iox/blob/main/parquet_file/src/storage.rs> here?

carols10cents: My understanding is that this code is going to replace the `parquet_file` crate eventually. The `parquet_file` crate works with `IoxObjectStore`, `ChunkAddr`, and a bunch of other things that `ingester` doesn't use. Am I incorrect? Should the `parquet_file` crate be evolved instead?

pauldix: Ideally I think the `parquet_file` crate would be evolved. The ingester should definitely be reusing object store code where we can.

pauldix: The question is what is possible while we have the requirement of operating the old structure and the new at the same time all of of `main`

carols10cents: Ok, should I be breaking existing code that uses `parquet_file` like https://github.com/influxdata/influxdb_iox/blob/442d63e65b43452b46468bf999bb77d3b22761d5/db/src/lifecycle/write.rs#L20-L24 or should I refactor so that both old and new work?

crepererum: You're free to touch existing code if you wanna evolve/split/improve the `parquet_file` interface. Ideally you shouldn't break existing persisted data though.

carols10cents:
@pauldix @crepererum Merge conflicts resolved, ready for your reviews!

generated_types/protos/influxdata/iox/ingest/v1/parquet_metadata.proto
```diff
@@ -0,0 +1,47 @@
...
+  bytes object_store_id = 1;
+
+  // Timestamp when this file was created.
+  google.protobuf.Timestamp creation_timestamp = 2;
+
+  // Unique namespace ID.
+  int32 namespace_id = 3;
+
+  // Unique name of the namespace.
+  string namespace_name = 4;
```
crepererum, line 19: Do we have ID and name here so we can use IDs everywhere but still can rebuild the catalog from files?

parquet_file/src/metadata.rs
```diff
@@ -454,6 +457,87 @@ impl IoxMetadata {
     }
 }
 
+/// IOx-specific metadata.
+///
+/// # Serialization
+/// This will serialized as base64-encoded [Protocol Buffers 3] into the file-level key-value
+/// Parquet metadata (under [`METADATA_KEY`]).
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct IoxMetadata {
```
crepererum, line 466: TBH I thought we can shoehorn everything that we need during the MVP into the existing (now old) `IoxMetadata`, but I'm OK w/ creating an entirely new object since it's probably easier and requires weird conversion hacks all over the place.

generated_types/protos/influxdata/iox/ingest/v1/parquet_metadata.proto
carols10cents: I'm not sure why namespace_name is here; I didn't add it in my version of this type's definition because I didn't need to use it anywhere. Nga added it in hers, I assume just for consistency?

parquet_file/src/metadata.rs
carols10cents: Yeaaahhhh I think it's much cleaner to have a new type, there are too many differences.

carols10cents:
@pauldix I'm going to merge this since Marco has approved; I'm happy to make any modifications in a new PR if you notice anything that isn't quite right :)

PR #3545, opened 2022-01-26
refactor: remove table_names from Predicate

tustvold:
This alters `Predicate` to no longer contain a list of table names, and instead makes this an implementation detail of the InfluxRPC storage layer. This greatly simplifies pruning, along with a number of other codepaths, as it now takes place over an array of chunks from the same table with a defined table-wide schema.

Unfortunately this turned into a rather substantial PR, due to the verbosity of the InfluxRPC implementation and its corresponding tests, but it is mostly mechanical. 

This change was inspired by the observation that nothing actually issues multi-table predicates, as `TableProvider` never creates them, and the influx RPC query path has to "normalize" them to per-table predicates to handle special tags such as `__measurement`.


db/src/lib.rs
```diff
@@ -1358,82 +1355,46 @@ pub mod test_helpers {
         ctx.collect(physical_plan).await.unwrap()
     }
 
-    pub fn mutable_chunk_ids(db: &Db, partition_key: &str) -> Vec<ChunkId> {
-        mutable_tables_chunk_ids(db, TableNameFilter::AllTables, partition_key)
-    }
-
-    pub fn mutable_tables_chunk_ids(
+    /// Returns the [`ChunkId`] of every chunk containing data in the mutable buffer
+    pub fn chunk_ids_mub(
```
tustvold, line 1359: These were renamed to make them shorter :sweat_smile:

db/src/lib.rs
```diff
@@ -832,22 +832,17 @@ impl Db {
...
-    }
-
-    /// Return chunk summary information for all chunks in the specified
-    /// tables and partition across all storage systems
-    pub fn partition_tables_chunk_summaries(
+    /// Return chunk summary information for all chunks
+    ///
+    /// If `table_name` is not `None` restricts to chunks in that table
+    /// If `partition_key` is not `None` restricts to chunks in that partition
+    pub fn filtered_chunk_summaries(
```
tustvold, line 839: The `filtered` prefix is to distinguish it from `QueryDatabase::chunk_summaries`

db/src/utils.rs
```diff
@@ -190,100 +185,53 @@ pub async fn make_db_time() -> (Arc<Db>, Arc<time::MockProvider>) {
     (db, provider)
 }
 
-fn chunk_summary_iter(db: &Db) -> impl Iterator<Item = ChunkSummary> + '_ {
```
tustvold, line 193: This was actually wrong before, it would count a chunk multiple times in the event of multiple tables containing chunks with the same partition key

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/expr.rs
```diff
@@ -192,6 +201,40 @@ impl AddRpcNode for PredicateBuilder {
             }
         }
     }
+
+    /// Adds an optional table name restriction to the existing list
+    pub fn table_option(self, table: Option<String>) -> Self {
```
tustvold, line 206: These methods are moved from `PredicateBuilder`

influxdb_iox/src/influxdb_ioxd/planner.rs
```diff
@@ -12,6 +11,7 @@ use query::{
 };
 
 pub use datafusion::error::{DataFusionError as Error, Result};
+use predicate::rpc_predicate::InfluxRpcPredicate;
```
tustvold, line 14: This is the new `InfluxRpcPredicate` that combines a `Predicate` with the none or more table names provided in the storage query

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/expr.rs
```diff
@@ -866,29 +924,39 @@ mod tests {
...
         let rpc_predicate = make_tagref_not_equal_predicate(TAG_KEY_MEASUREMENT, "foo");
 
-        let predicate = PredicateBuilder::default()
+        let predicate = InfluxRpcPredicateBuilder::default()
             .rpc_predicate(Some(rpc_predicate))
             .expect("successfully converting predicate")
             .build();
 
-        let expected_exprs = vec![col("_measurement").not_eq(lit("foo"))];
+        let tables = ["foo", "bar"];
```
tustvold, line 932: Here we test the conversion to per-table `Predicate`

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -774,20 +773,6 @@ where
     }
 }
 
-trait SetRange {
```
tustvold, line 777: These extension traits are no longer necessary, the methods are implemented directly on `InfluxRpcPredicateBuilder`

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1728,7 +1712,7 @@ mod tests {
         let db_info = org_and_bucket();
 
         // predicate specifies m4, so this is filtered out
-        let chunk = TestChunk::new("my_table").with_error("This is an error");
+        let chunk = TestChunk::new("m5").with_error("This is an error");
```
tustvold, line 1715: These tests were previously exploiting the fact that TestDatabase wasn't pruning chunks correctly - the query is restricted to a different chunk, and so theoretically should have been ignoring the chunk. With the changes in this PR the pruning logic is inherent to how predicates are calculated so this needs to be updated

predicate/src/lib.rs
```diff
@@ -12,3 +12,4 @@ pub mod delete_expr;
 pub mod delete_predicate;
 pub mod predicate;
 pub mod regex;
+pub mod rpc_predicate;
```
tustvold, line 15: The `predicate` crate seemed like the most sensible place to put this, in order to keep it with the other variants

predicate/src/rpc_predicate.rs
```diff
@@ -0,0 +1,391 @@
...
+            .map(|table| {
+                let predicate = normalize_predicate(&table, &self.inner);
+                (table, predicate)
+            })
+            .collect()
+    }
+
+    /// Same as [`Self::table_predicates`] but taking a function that lazily computes
+    /// the table_names
+    pub fn table_predicates_lazy<F, I>(&self, table_names: F) -> Vec<(String, Predicate)>
```
tustvold, line 100: This is used so that the catalog is only asked for the list of table names if the query doesn't have table name restrictions

predicate/src/rpc_predicate.rs
```diff
@@ -0,0 +1,391 @@
...
+        self.table_predicates(std::iter::once_with(table_names).flatten())
+    }
+
+    /// Returns the table names this predicate is restricted to if any
+    pub fn table_names(&self) -> Option<&BTreeSet<String>> {
+        self.table_names.as_ref()
+    }
+}
+
+/// Predicate that has been "specialized" / normalized for a
```
tustvold, line 114: This logic and corresponding tests are moved from `query::frontend::influxrpc`

predicate/src/predicate.rs
```diff
@@ -353,41 +335,6 @@ impl PredicateBuilder {
         self
     }
 
-    /// Adds an optional table name restriction to the existing list
-    pub fn table_option(self, table: Option<String>) -> Self {
```
tustvold, line 357: These were moved to `InfluxRpcPredicateBuilder`

query/src/frontend/influxrpc.rs
```diff
@@ -9,17 +9,18 @@ use data_types::chunk_metadata::ChunkId;
 use datafusion::{
     error::{DataFusionError, Result as DatafusionResult},
     logical_plan::{
-        binary_expr, lit, when, Column, DFSchema, DFSchemaRef, Expr, ExprRewriter, LogicalPlan,
```
tustvold, line 12: I recommend viewing this file with whitespace ignored https://github.com/influxdata/influxdb_iox/pull/3545/files?w=1

query/src/frontend/influxrpc.rs
```diff
@@ -236,68 +210,68 @@ impl InfluxRpcPlanner {
...
-                let predicate = normalizer.normalized(table_name);
-
-                // Try and apply the predicate using only metadata
-                let pred_result = chunk
-                    .apply_predicate_to_metadata(&predicate)
-                    .map_err(|e| Box::new(e) as _)
-                    .context(CheckingChunkPredicateSnafu {
-                        chunk_id: chunk.id(),
-                    })?;
+        let table_predicates = rpc_predicate.table_predicates_lazy(|| database.table_names());
```
tustvold, line 228: Rather than getting a list of chunks and then determining the tables from those chunks, we instead determine the tables and then iterate through the chunks for each table

query/src/frontend/influxrpc.rs
```diff
@@ -353,75 +330,74 @@ impl InfluxRpcPlanner {
...
+                // deleted data before getting tag keys
+                let mut do_full_plan = chunk.has_delete_predicates();
 
-            let column_names: Vec<&str> = schema
-                .tags_iter()
-                .map(|f| f.name().as_str())
-                .collect::<Vec<&str>>();
+                // Try and apply the predicate using only metadata
+                let pred_result = chunk
+                    .apply_predicate_to_metadata(predicate)
```
tustvold, line 344: I suspect this could be updated to use `prune_chunks_metadata` but I went for the smaller diff to start with

query/src/frontend/influxrpc.rs
```diff
@@ -631,33 +608,38 @@ impl InfluxRpcPlanner {
...
 
-        // map table -> Vec<Arc<Chunk>>
-        let chunks = database.chunks(normalizer.unnormalized());
-        let table_chunks = self.group_chunks_by_table(&mut normalizer, chunks)?;
+        let table_predicates = rpc_predicate.table_predicates_lazy(|| database.table_names());
+        let mut field_list_plan = FieldListPlan::with_capacity(table_predicates.len());
+
+        for (table_name, predicate) in &table_predicates {
+            let chunks = database.chunks(table_name, predicate);
+            let chunks = prune_chunks_metadata(chunks, predicate)?;
```
tustvold, line 632: This is necessary because `group_chunks_by_table` used to call this

query/src/frontend/influxrpc.rs
```diff
@@ -838,40 +820,6 @@ impl InfluxRpcPlanner {
         Ok(SeriesSetPlans::new(ss_plans))
     }
 
-    /// Creates a map of table_name --> Chunks that have that table that *may* pass the predicate
-    fn group_chunks_by_table<C>(
```
tustvold, line 842: This logic is moved into `prune_chunks_metadata`

query/src/frontend/influxrpc.rs
```diff
@@ -1429,6 +1375,33 @@ impl InfluxRpcPlanner {
     }
 }
 
+/// Prunes the provided list of chunks using [`QueryChunk::apply_predicate_to_metadata`]
+///
+/// TODO: Should this logic live with the rest of the chunk pruning logic?
```
tustvold, line 1380: I don't see a particular reason for not doing this in `QueryDatabase::chunks()` but I figured this PR was large enough

query/src/frontend/influxrpc.rs
```diff
@@ -1766,227 +1739,6 @@ fn make_selector_expr<'a>(
         .alias(col_name))
 }
 
-/// Creates specialized / normalized predicates that are tailored to a specific
-/// table.
-#[derive(Debug)]
-struct PredicateNormalizer {
```
tustvold, line 1772: `Predicate` is what was normalized, and `InfluxRpcPredicate` un-normalized

query/src/frontend/influxrpc.rs
```diff
@@ -1766,227 +1739,6 @@ fn make_selector_expr<'a>(
...
-/// _value > 34.2
-/// ```
-///
-/// When evaluated on table "cpu" then the expression is rewritten as a
-/// collection of disjunctive expressions against all field columns
-/// ```text
-/// ("field1" > 34.2 OR "field2" > 34.2 OR "fieldn" > 34.2)
-/// ```
-#[derive(Debug)]
-struct TableNormalizedPredicate {
```
tustvold, line 1838: Logic moved into `predicate::rpc_predicate`

query/src/test.rs
```diff
@@ -900,14 +914,7 @@ impl QueryChunk for TestChunk {
             return Ok(predicate_match);
         }
 
-        // otherwise fall back to basic filtering based on table name predicate.
-        let predicate_match = if !predicate.should_include_table(&self.table_name) {
```
tustvold, line 904: This pruning now happens higher up

query_tests/src/influxrpc/field_columns.rs
```diff
@@ -1,6 +1,7 @@
 use arrow::datatypes::DataType;
```
tustvold, line 1: Updating these tests showed me an additional value-add for porting these test cases to not be defined in code, it was painful :sweat_smile:

query_tests/src/scenarios.rs
```diff
@@ -664,8 +664,8 @@ impl DbSetup for TwoMeasurementsManyFieldsTwoChunks {
         ];
         write_lp(&db, &lp_lines.join("\n"));
 
-        assert_eq!(count_mutable_buffer_chunks(&db), 4);
```
tustvold, line 667: These were previously wrong, as the method was counting chunks twice

data_types/src/chunk_metadata.rs
```diff
@@ -86,6 +86,21 @@ impl ChunkStorage {
             Self::ObjectStoreOnly => "ObjectStoreOnly",
         }
     }
+
+    /// Returns true if has data in the mutable buffer
+    pub fn has_mutable_buffer(&self) -> bool {
```
e-dard, line 91: super nit, but I wonder if these should be `is_mutable_buffer` etc because only one of the methods can be true. To me `has` implies the possibility of a multiple possible truths.

db/src/catalog.rs
```diff
@@ -294,27 +261,21 @@ impl Catalog {
 
     /// Calls `map` with every chunk and returns a collection of the results
     ///
-    /// If `partition_key` is Some(partition_key) only returns chunks
-    /// from the specified partition.
-    ///
-    /// `table_names` specifies which tables to include
+    /// If `table_name` is not `None` restricts to chunks in that table
+    /// If `partition_key` is not `None` restricts to chunks in that partition
```
e-dard, line 265:
```suggestion
    /// If `table_name` is `Some` restricts to chunks in that table.
    /// If `partition_key` is `Some` restricts to chunks in that partition.
```

db/src/lib.rs
```diff
@@ -832,22 +832,17 @@ impl Db {
...
-        self.partition_tables_chunk_summaries(TableNameFilter::AllTables, partition_key)
-    }
-
-    /// Return chunk summary information for all chunks in the specified
-    /// tables and partition across all storage systems
-    pub fn partition_tables_chunk_summaries(
+    /// Return chunk summary information for all chunks
+    ///
+    /// If `table_name` is not `None` restricts to chunks in that table
+    /// If `partition_key` is not `None` restricts to chunks in that partition
```
e-dard, line 838:
```suggestion
    /// If `table_name` is `Some` restricts to chunks in that table.
    /// If `partition_key` is `Some` restricts to chunks in that partition.
```

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/expr.rs
```diff
@@ -146,15 +149,21 @@ pub enum GroupByAndAggregate {
...
+#[derive(Debug, Default)]
+pub struct InfluxRpcPredicateBuilder {
+    table_names: Option<BTreeSet<String>>,
+    inner: PredicateBuilder,
 }
 
-impl AddRpcNode for PredicateBuilder {
+impl InfluxRpcPredicateBuilder {
+    /// Sets the timestamp range
+    pub fn set_range(mut self, range: Option<RPCTimestampRange>) -> Self {
```
e-dard, line 160: This seems redundant to me because providing `None` has no side effects on anything. What about the following as an alternative?

```rust
pub fn set_range(mut self, range: RPCTimestampRange) -> Self
```

e-dard (approved):
These changes make sense to me.  👍 

I tested this branch against our internal Flux and InfluxQL integration suites and got equivalent output to `main`. 

I would ask @alamb to do the same as he was working on InfluxQL tests more recently than I and may have a more up to date branch to test against.

data_types/src/chunk_metadata.rs
tustvold: Whilst true for the mutable buffer, the read_buffer and object_store methods are not exclusive.

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/expr.rs
tustvold: This was kept from the old code, and reduces code in the service handlers, otherwise they have to have an if block which is especially clunky as this is a consuming builder

db/src/access.rs
```diff
@@ -215,16 +211,15 @@ impl ChunkAccess {
     /// Returns all chunks that may have data that passes the
```
alamb, line 211:
```suggestion
    /// Returns all chunks from `table_name` that may have data that passes the
```

db/src/access.rs
```diff
@@ -295,6 +285,15 @@ impl QueryDatabase for QueryCatalogAccess {
             .map(|table| Arc::clone(&table.schema().read()))
     }
 
+    /// Return a covering set of chunks for a particular partition
```
alamb, line 288:
```suggestion
    /// Return a covering set of chunks for a particular table and predicate
```

data_types/src/chunk_metadata.rs
e-dard: Doh. Of course :-)

predicate/src/lib.rs
alamb: I think it makes sense

alamb (approved):
Epic 🤘

query_tests/src/influxrpc/read_group.rs
```diff
@@ -755,9 +765,6 @@ impl DbSetup for MeasurementForGroupKeys {
 
 #[tokio::test]
 async fn test_grouped_series_set_plan_group_by_state_city() {
-    // no predicate
```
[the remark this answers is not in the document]
tustvold: Aah yeah, my IDE expands the parameter names for me, I can see why this would be advantageous when reviewing PRs as Github doesn't...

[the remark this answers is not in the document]
tustvold: I went for a quick middle ground and replaced `Default::default()` with `InfluxRpcPredicate::default()`

predicate/src/rpc_predicate.rs
tustvold: Done

PR #3546, opened 2022-01-27
ci: Run clippy lint for undocumented unsafe code

pierwill:
The [`clippy::undocumented_unsafe_blocks`](https://rust-lang.github.io/rust-clippy/master/#undocumented_unsafe_blocks) lint is allowed by default. As a "restriction lint" it must be explicitly specified.

Turning this on will cause the following code to fail in CI

```rust
unsafe {
    [...snip...]
}
```

while allowing this:

```rust
// SAFETY: The following code is safe because [good reasons].
unsafe {
    [...snip...]
}
```


pierwill:
[This](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/14477/workflows/a396217b-1057-4476-a53f-fd6bb67e2b1f/jobs/110010?invite=true#step-104-737) may actually be a bug in clippy...

crepererum:
I think you might need to indent the line after the `SAFETY` word, but not sure. Also can you please instead of adding the lint via CLI add it to the individual crates, like:

https://github.com/influxdata/influxdb_iox/blob/7261571abfe1b6088810732d5811240ec585c7cf/ingester/src/lib.rs#L6-L13

pierwill:
Asking about this now on the Zulip channel for clippy.

pierwill:
This does *not* reproduce [on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f1f31ccdb2091ca583c025b5a6dfb655).

pierwill:
Ah okay, I think the problem is with the version of clippy.

data_types/src/lib.rs
```diff
@@ -2,7 +2,12 @@
 //! servers including replicated data, rules for how data is split up and
 //! queried, and what gets stored in the write buffer database.
 
-#![deny(rustdoc::broken_intra_doc_links, rustdoc::bare_urls, rust_2018_idioms)]
```
crepererum, line 5: Could you please add this to ALL crates? I know it's a bit cumbersome but sadly there isn't any workspace-wide way to declare `allow`/`warn`/`deny` lints.

pierwill: Sure thing 👍 . 

However, I'm still investigating the the mysterious failure of this lint to work on this repo only....

pierwill:
This is an incredibly frustrating bug. 😠

alamb:

alamb:

PR #3547, opened 2022-01-27
feat: IoxMetadata for parquet file

### LINKED ISSUE #3513: Parquet metadata change/repurpose/rename?
The parquet metadata that gets persisted needs to be slightly different in the new design, but the changes also need to be backwards compatible with the existing design.

- What can be repurposed?
- What needs to be renamed?
- What needs to be added?
- What needs to be deprecated?

NGA-TRAN:
Closes #3513 
@carols10cents and @pauldix : this is ready for review.


ingester/src/compact.rs
```diff
@@ -25,18 +30,161 @@ pub enum Error {
...
+    let max = time_col.iter().max().context(MaxTimeSnafu{})??;
+    
+    Ok((min, max))
+}
+
+/// Compact a giving persisting batch
+/// Return compacted data with its metadata
+pub async fn compact_persting_batch(time_provider: Arc<dyn TimeProvider>, executor: &Executor, batch: PersistingBatch) 
+//-> Result<(SendableRecordBatchStream, IoxMetadata)> {
+-> Result<(Vec<RecordBatch>, Option<IoxMetadata>)> {
```
NGA-TRAN, line 141: I do want to keep this return `SendableRecordBatchStream`. To do so, the `min_time` and `max_time` won't be computed and put in IoxMetadata yet. They will be computed at persistence step when we read the record batches form the stream

ingester/src/compact.rs
```diff
@@ -25,18 +30,161 @@ pub enum Error {
...
+        .await
+        .context(CollectStreamSnafu{})?;
+
+    // no data after compaction
+    if output_batches.iter().all(|b| b.num_rows() == 0) {
+        return Ok((vec![], None));
+    }
+
+    // Compute min and max of the `time` column 
+    let (min_time, max_time) = compute_timenanosecond_min_max(&output_batches)?;
```
NGA-TRAN, line 164: Lines 154 to 164 should be removed if we compute `time_min` and `time_max` after and outside this function

NGA-TRAN (commented):
@carols10cents , @pauldix and @crepererum
This is the work I am doing and mentioned in [Carol's PR](https://github.com/influxdata/influxdb_iox/pull/3544).  They are related and I want us to discuss to find a good way to implement them

ingester/src/compact.rs
```diff
@@ -25,18 +30,161 @@ pub enum Error {
...
+//-> Result<(SendableRecordBatchStream, IoxMetadata)> {
+-> Result<(Vec<RecordBatch>, Option<IoxMetadata>)> {
+
+        // Nothing to compact
+        if batch.data.data.is_empty() {
+            return Ok((vec![], None));
+        }
+
+    // Get table name and partittion key
+    let table_name = batch.table_id.get().to_string(); // todo: implement get_table_name(table_id) in the catalog interface or store it with the persisting batch
```
pauldix, line 149: This mapping is going to be in the in-memory cache that I'm wiring up. Would be good to get the interface defined and get these two together.

ingester/src/compact.rs
```diff
@@ -25,18 +30,161 @@ pub enum Error {
...
+    let partition_key = batch.partition_id.get().to_string(); // todo: implement get_partition_key(partition_id) in the catalog interface
```
pauldix, line 150: Same as above

ingester/src/compact.rs
```diff
@@ -70,35 +241,63 @@ mod tests {
...
-    use arrow::record_batch::RecordBatch;
+    use arrow::{array::TimestampNanosecondArray, record_batch::RecordBatch};
     use arrow_util::assert_batches_eq;
     use iox_catalog::interface::SequenceNumber;
     use query::test::{raw_data, TestChunk};
+    use time::SystemProvider;
 
     #[tokio::test]
-    async fn test_compact_one_batch_no_dupilcates() {
+    async fn test_compact_persisting_batch_on_one_record_batch_no_dupilcates() {
```
NGA-TRAN, line 251: I modify this test to invoke the `compact_persisting_batch` to test IoxMetadata, too

NGA-TRAN:
@carols10cents : I will rebase after your merge to use your IoxMetadata

ingester/src/compact.rs
```diff
@@ -25,20 +30,186 @@ pub enum Error {
...
+        .column(index)
+        .as_any()
+        .downcast_ref::<TimestampNanosecondArray>()
+        .context(TimeCastingSnafu {})?;
+
+    let min = time_col
+        .iter()
+        .min()
+        .context(MinTimeSnafu {})?
+        .context(MinTimeSnafu {})?;
```
carols10cents, line 149: So these look like they're needed because the time column optionally contains values, and then the whole column could be empty, which would mean there's no min or max. I wonder if those would be the two errors that would be useful to report, rather than have one error for min and one for max... 

what I mean is, define these two error variants instead of `MinTime` and `MaxTime`:

```
#[snafu(display("Timestamp column had 0 values"))]
NoTimestampValues,

#[snafu(display("Timestamp column had all null values"))]
AllNullTimestampValues,
```

and then both here and when getting the max, it would be:

```
let min = time_col
    .iter()
    .min()
    .context(NoTimestampValuesSnafu)?
    .context(AllNullTimestampValuesSnafu)?;   
```

I could also see panicking here with `expect` instead because all of our record batches are supposed to have time values, so if the time column doesn't have any values or has all null values, something has gone very wrong. Not sure if we want to crash for that, though.

ingester/src/compact.rs
```diff
@@ -25,20 +30,186 @@ pub enum Error {
...
     #[snafu(display("Error while building delete predicate from start time, {}, stop time, {}, and serialized predicate, {}", min, max, predicate))]
     DeletePredicate {
         source: predicate::delete_predicate::Error,
         min: String,
         max: String,
         predicate: String,
     },
+
+    #[snafu(display("The Record batch is empty"))]
+    EmptyBatch {},
```
carols10cents, line 45: Because these error variants don't have any fields, you can leave out the curly brackets here and where they're used, if you want

carols10cents (approved):
Just some small thoughts about error handling, overall LGTM!

ingester/src/compact.rs
NGA-TRAN: This is a very good point. I like both solutions but I think the expect to panic is more appropriate because the time column must have values. What do you think @pauldix?

ingester/src/compact.rs
NGA-TRAN: Will remove the curly brackets

ingester/src/compact.rs
NGA-TRAN: I have made them panic but can also convert them to snafu if needed

ingester/src/compact.rs
NGA-TRAN: Done

PR #3549, opened 2022-01-27
refactor(router2): DML handler stack

domodwyer:
Refactors the DML handler into a composable trait in preparation for the schema validator.

This will allow us to construct a request handling pipeline with isolated/decoupled/easily testable steps through which DML requests progress.

---

* refactor: composable DmlHandler stack (e28c0967)

      Changes the DmlHandler trait to allow composition of handler logic in order to
      construct the complete request processing pipeline.

* feat: debug log write/delete requests (908b8fca)

      Log requests hitting the HTTP endpoint at DEBUG.

* refactor: dml_handler -> dml_handlers (30744206)


router2/src/dml_handlers/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+//!                       ┌──────────────┐
+//!                       │ Write Buffer │
+//!                       └──────────────┘
+//!                               │
+//!                               │
+//!                      ┌────────▼─────┐
+//!                      │    Kafka     ├┐
+//!                      └┬─────────────┘├┐
+//!                       └┬─────────────┘│
+//!                        └──────────────┘
```
domodwyer, line 39: This diagram shows the internal parts of router2 and how they fit together.

I am working on the schema validation handler now, and in the future this pipeline can include decorators adding a partitioning step, multi-kafka-topic selection, etc.

router2/src/dml_handlers/trait.rs
```diff
@@ -17,23 +18,30 @@ pub enum DmlError {
...
     /// Write `batches` to `namespace`.
     async fn write(
         &self,
         namespace: DatabaseName<'static>,
         batches: HashMap<String, MutableBatch>,
-        payload_stats: PayloadStatistics,
-        body_len: usize,
         span_ctx: Option<SpanContext>,
-    ) -> Result<(), DmlError>;
+    ) -> Result<(), Self::Error>;
```
domodwyer, line 44: This is the main change of the PR - removing call-specific fields that were only used for logging (instead, logging in the caller) in order to have a cleanly composibile trait.

The rest of this PR is minor refactoring to accomidate this change.

router2/src/sharder/mock.rs
```diff
@@ -50,14 +64,40 @@ where
...
+        namespace: &data_types::DatabaseName<'_>,
+        payload: &MutableBatch,
+    ) -> &Self::Item {
+        let mut guard = self.0.lock();
+        guard.record_call(MockSharderCall {
+            table_name: table.to_string(),
+            namespace: namespace.to_string(),
+            payload: MockSharderPayload::MutableBatch(payload.clone()),
+        });
+        Box::leak(Box::new(
```
mkmik, line 90: I'm not familiar with Box::leak. Does this literally just return a reference to a heap allocated object that will never be freed, but this is ok because this is just for tests?

mkmik (approved):
LGTM. Nice.

(with a small non-blocking question inline)

router2/src/dml_handlers/trait.rs
```diff
@@ -17,23 +18,30 @@ pub enum DmlError {
...
+    #[error(transparent)]
+    WriteBuffer(#[from] ShardError),
+
     /// An unknown error occured while processing the DML request.
     #[error("internal dml handler error: {0}")]
     Internal(Box<dyn std::error::Error + Send + Sync>),
 }
 
-/// An abstract handler of DML requests.
+/// A composable, abstract handler of DML requests.
```
alamb, line 30: I am not sure what is meant by "composable" for whatever it is worth (not that it is wrong)

router2/src/server/http.rs
```diff
@@ -397,12 +422,6 @@ mod tests {
 
                     let calls = dml_handler.calls();
                     assert_matches!(calls.as_slice(), $($want_dml_calls)+);
-
-                    // If this was a write op, ensure the body length matches
```
alamb, line 401: why was this removed?

router2/src/sharder/mock.rs
alamb, line 90: It seems strange to have to return a reference to something that is never freed; Maybe the trait should be change to return `Self::Item` to avoid this?

alamb (approved):

router2/src/sharder/mock.rs
domodwyer: Correct! It allows you to return a ref to an item that the mock holds - otherwise you'd be returning item owned by the current function.

Obviously a bad idea to do in prod code, but as you note this is a test 👍

router2/src/dml_handlers/trait.rs
domodwyer: So much terminology! https://en.wikipedia.org/wiki/Composability

router2/src/server/http.rs
domodwyer: Because the mock no longer captures this.

It was kinda pointless anyway - nothing really uses it - I just asserted it before because it was there.

router2/src/sharder/mock.rs
domodwyer: This is for tests - the actual code does indeed return refs (which lets you return `&ReallyBigNonClonableType` - similar to how iterators lend out their contents).

domodwyer:
Force pushed to rebase onto main - no code changes 👍

router2/src/dml_handlers/trait.rs
domodwyer: I guess this is a good analogy: 😂 

russian nesting dolls

alamb: LOL I was not asking for a wikipedia link to a definition -- I was more asking how this term applied to the trait (like will there be DML handlers within DML handlers)?

router2/src/sharder/mock.rs
alamb: I was thinking the actual traits could define `Item = &ReallyBigNonCloneableType`

router2/src/dml_handlers/trait.rs
domodwyer: Ah I see - I was surprised!

Yes, a DML handler will call a DML handler and so on until the chain is complete - standard decorator stuff to add behaviour in layers, much like middleware composes together to add tracing/auth logic/etc.

router2/src/sharder/mock.rs
domodwyer: Do you mean the implementation specifies `Item = &ReallyBigNonCloneableType`?

You would need generic associated types (an unstable/nightly only rust feature) in order to express that `Item` is generic over any lifetime - try fiddling [with this example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=570b124bed3b576313d8debb1080dea5).

router2/src/dml_handlers/trait.rs
alamb: Got it -- 👍

router2/src/sharder/mock.rs
alamb: It can be accomplished with `Arc` https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9ff23926a11993372ab7ebe897d4c509 

though that is not ideal of course

domodwyer: Yeah but then if you were returning array indexes or something `Copy` you'd have to wrap them in Arc's to give them back - I think returning refs is preferable at the expense of the mocks for testing only.

alamb: I agree it seems like a reasonable tradeoff

router2/src/dml_handlers/mod.rs
NGA-TRAN: Thanks for the diagram, @domodwyer. Easy to understand what you have been building 👍

PR #3553, opened 2022-01-27
feat: columnar predicate pruning

tustvold:
Builds on the change in #3545 to remove table names from predicates, to rework chunk pruning to be more "columnar".

In particular all the statistics for the chunks are collected into a single RecordBatch and the pruning predicate evaluated against this in one pass. This should hopefully significantly improve its performance and reduce the malloc thrashing we were seeing in production.


db/src/catalog/table.rs
```diff
@@ -110,6 +111,31 @@ impl Table {
         self.partitions.values().filter_map(|x| x.read().summary())
     }
 
+    /// Calls `map` with every chunk and returns a collection of the results
+    ///
+    /// If `partition_key` is `Some` restricts to chunks in that partition.
+    pub fn filtered_chunks<F, C>(&self, partition_key: Option<&str>, map: F) -> Vec<C>
```
tustvold, line 117: This is moved from catalog.rs, to allow query to get a consistent snapshot of both schema and chunks.

db/src/access.rs
```diff
@@ -247,22 +257,13 @@ impl ChunkPruner<DbChunk> for ChunkAccess {
...
     type Observed = DbChunk;
 
     fn was_pruned(&self, chunk: &Self::Observed) {
-        let metrics = self.access_metrics.table_metrics(chunk.table_name());
-        metrics.pruned_chunks.inc(1);
         let chunk_summary = chunk.summary().expect("Chunk should have summary");
-        metrics.pruned_rows.inc(chunk_summary.total_count())
-    }
-
-    fn could_not_prune_chunk(&self, chunk: &Self::Observed, reason: &str) {
```
tustvold, line 260: We no longer have this granularity of reporting, given it was only ever a debug log I'm inclined to think this is no great a loss

query/src/pruning.rs
```diff
@@ -23,140 +27,212 @@ pub trait PruningObserver {
...
-
-    /// Called when the specified chunk could not be pruned, for some reason
-    fn could_not_prune_chunk(&self, _chunk: &Self::Observed, _reason: &str) {}
 }
 
 /// Given a Vec of prunable items, returns a possibly smaller set
 /// filtering those that can not pass the predicate.
-pub fn prune_chunks<C, P, O>(observer: &O, summaries: Vec<C>, predicate: &Predicate) -> Vec<C>
+///
+/// TODO(raphael): Perhaps this should return `Result<Vec<bool>>` instead of
```
tustvold, line 35: As this is now effectively a single-shot, the PruningObserver plumbing seems a tad unnecessary, but definitely something for a follow up PR.

query/src/pruning.rs
```diff
@@ -600,104 +690,6 @@ mod test {
         assert_eq!(names(&pruned), vec!["chunk1", "chunk4", "chunk6"]);
     }
 
-    #[test]
-    fn test_pruned_incompatible_types() {
```
tustvold, line 604: This is testing an eventuality that can no longer occur, chunks must have compatible schema within a table

db/src/access.rs
```diff
@@ -107,23 +107,22 @@ impl AccessMetrics {
...
-                let metrics = TableAccessMetrics {
-                    pruned_chunks: self.pruned_chunks.recorder(attributes.clone()),
-                    pruned_rows: self.pruned_rows.recorder(attributes),
-                };
-
-                (Arc::clone(table), metrics)
-            });
-            metrics
-        })
+    fn table_metrics(&self, table: &str) -> Arc<TableAccessMetrics> {
```
tustvold, line 110: `PruningObserver` is now directly on `TableAccessMetrics` so we need an `Arc` to avoid holding a lock whilst pruning is evaluated

db/src/access.rs
```diff
@@ -51,7 +51,7 @@ struct AccessMetrics {
     pruned_rows: Metric<U64Counter>,
 
     /// Keyed by table name
-    tables: Mutex<HashMap<Arc<str>, TableAccessMetrics>>,
+    tables: Mutex<HashMap<String, Arc<TableAccessMetrics>>>,
```
alamb, line 54: any particular reason to change to owned String?

db/src/access.rs
alamb: This one is fine -- I think it is important to have in the debug logs how many chunks were pruned so that it is easier to determine for wrong answer queries if the pruning is the problem or if the issue lies elsewhere

db/src/catalog/table.rs
alamb: Maybe it would be more precisely called `filter_map_chunks`

query/src/provider.rs
```diff
@@ -92,7 +92,13 @@ enum ColumnType {
 /// Something that can prune chunks based on their metadata
 pub trait ChunkPruner<C: QueryChunk>: Sync + Send + std::fmt::Debug {
     /// prune `chunks`, if possible, based on predicate.
-    fn prune_chunks(&self, chunks: Vec<Arc<C>>, predicate: &Predicate) -> Vec<Arc<C>>;
+    fn prune_chunks(
+        &self,
+        table_name: &str,
+        schema: Arc<Schema>,
```
alamb, line 98: recommend naming this `table_schema` to make it is the overall table's schema (rather than the schema each chunk is expected to have)

query/src/pruning.rs
alamb: `Result<Vec<bool>>` makes sense to me

query/src/pruning.rs
alamb: I think we should add (or perhaps modify an existing test) to show pruning working when the chunks to be pruned have different (but compatible) schemas .

alamb (approved):
I think this code looks 👍  -- very nice @tustvold 

The only thing I would like to see is a test demonstrating chunk pruning working with chunks that have different schemas but that could be added as a follow on test

db/src/access.rs
tustvold: There wasn't a particular reason to keep the `Arc` as it was never getting passed it

db/src/catalog/table.rs
tustvold: I'd prefer to keep this consistent with the name on the catalog proper

query/src/pruning.rs
tustvold: Is this not what `test_pruned_different_schema` does a couple of lines up?

tustvold:
I'm going to get this in and then follow up, I'd like to get the production build chugging along :sweat_smile:

PR #1842, opened 2021-06-29
feat: add gRPC listener for new write protocol

jacobmarble:
Creates a new write service using gRPC/protobuf.

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #3554, opened 2022-01-27
feat: schema validation for MutableBuffer instances

domodwyer:
Changes the `validate_or_insert_schema()` function to operate on a (set of) `MutableBatch` rather than `ParsedLine`.

This lets us validate writes where the original LP is not avialable (for example, downstream of the router, or via the non-LP gRPC write interface) as well as allowing us to validate the parsed `MutableBuffer` columns, rather than the columns in all the parsed LP lines (a smaller N) only to then parse them into a `MutableBuffer` anyway.

---

* refactor: Debug bounds on Catalog trait (3edf459a)

* feat: validate MutableBatch schema (e088bca3)

      Changes the schema validation code to validate MutableBatch instances
      (coming from a pre-parsed LP write, and non-LP-based writes) instead of parsed
      LP lines.


iox_catalog/src/interface.rs
```diff
@@ -720,6 +720,36 @@ impl TryFrom<i16> for ColumnType {
...
+        match value {
+            InfluxColumnType::Tag => Self::Tag,
+            InfluxColumnType::Field(InfluxFieldType::Float) => Self::F64,
+            InfluxColumnType::Field(InfluxFieldType::Integer) => Self::I64,
+            InfluxColumnType::Field(InfluxFieldType::UInteger) => Self::U64,
+            InfluxColumnType::Field(InfluxFieldType::String) => Self::String,
+            InfluxColumnType::Field(InfluxFieldType::Boolean) => Self::Bool,
+            InfluxColumnType::Timestamp => Self::Time,
+            // The "IOx" data model is unused.
+            InfluxColumnType::IOx(_) => unimplemented!(),
```
domodwyer, line 734: I could not find any use of this `InfluxColumnType::IOx(_)` variant - is there a reason to keep it around?

iox_catalog/src/lib.rs
```diff
@@ -28,150 +29,129 @@ pub mod interface;
...
+async fn validate_mutable_batch(
+    mb: &MutableBatch,
+    table_name: &str,
+    schema: &mut Cow<'_, NamespaceSchema>,
+    catalog: &dyn Catalog,
+) -> Result<()> {
+    // Check if the table exists in the schema.
+    //
+    // Because the entry API requires &mut it is not used to avoid a premature
+    // clone of the Cow.
```
domodwyer, line 73: Easily the strangest sentence I have wrote all week 🐄

iox_catalog/src/lib.rs
```diff
@@ -28,150 +29,129 @@ pub mod interface;
...
+        validate_mutable_batch(batch, table_name, &mut schema, catalog).await?;
+    }
+
+    match schema {
+        Cow::Owned(v) => Ok(Some(v)),
+        Cow::Borrowed(_) => Ok(None),
+    }
+}
+
+async fn validate_mutable_batch(
```
domodwyer, line 64: This fn is nearly as long as the old one (in terms of actual lines) but most of it is comments explaining the process - actual code is much reduced 👍

iox_catalog/src/lib.rs
```diff
@@ -211,94 +191,265 @@ mod tests {
...
+                "f2" => ColumnType::F64,
+                "time" => ColumnType::Time,
+            ],
+            "m2": [
+                "f1" => ColumnType::Bool,
+                "t3" => ColumnType::Tag,
+                "time" => ColumnType::Time,
+            ],
+        }
+    );
```
domodwyer, line 336: The test macro above makes expressing "given a series of these writes, assert the schema in both the cache and the database matches this" trivial.

Easy testing -> more tests 🎉

iox_catalog/Cargo.toml
```diff
@@ -14,6 +14,13 @@ tokio = { version = "1.13", features = ["io-util", "macros", "parking_lot", "rt-
...
 workspace-hack = { path = "../workspace-hack"}
 uuid = { version = "0.8", features = ["v4"] }
+mutable_batch = { path = "../mutable_batch" }
+schema = { path = "../schema" }
 
 [dev-dependencies] # In alphabetical order
 dotenv = "0.15.0"
+mutable_batch_lp = { path = "../mutable_batch_lp" }
+paste = "1.0.6"
+pretty_assertions = "1.0.0"
```
mkmik, line 24: I wonder if we really needs this;
last time I looked at some assertion errors in my IDE I _think_ it has been highlighted sufficiently well

iox_catalog/src/interface.rs
```diff
@@ -667,6 +650,23 @@ impl ColumnSchema {
     }
 }
 
+impl PartialEq<mutable_batch::column::Column> for ColumnSchema {
+    fn eq(&self, other: &mutable_batch::column::Column) -> bool {
```
NGA-TRAN, line 654: So this function compare the type of `ColumnSchema` with the type of `MutableBatch's column`? With this trait, will we use `x.eq(y)` or `x == y`?   I feel a bit confused with either. Should we make the function `type_equal` or something like that to make it clearer?

iox_catalog/src/interface.rs
NGA-TRAN: Agree we do not need it but maybe someone created it had a reason?

iox_catalog/src/lib.rs
NGA-TRAN: 😆

iox_catalog/Cargo.toml
domodwyer: I don't get assertion output highlighting?!

This gives a diff-like colourised output that makes it obvious to spot an accidentally transposed single character, or extra space, etc in nested structures. It's a pure dev life improvement for a light dependency, but only really matters when a test fails which hopefully doesn't happen - I don't mind either way really 👍

iox_catalog/src/lib.rs
```diff
@@ -28,150 +29,129 @@ pub mod interface;
...
+    // becomes a Cow::owned() copy and the modified copy should be inserted into
+    // the schema before returning.
+
+    for (name, col) in mb.columns() {
+        // Check if the column exists in the cached schema.
+        //
+        // If it does, validate it. If it does not exist, create it and insert
+        // it into the cached schema.
+        match table.columns.get(name.as_str()) {
+            Some(existing) if existing == col => {
```
NGA-TRAN, line 118: The `==` here is for comparing data type that I asked above, right? It is good that you have comments describing the code  but I have to say the code will be clearer if you replace `==` with a function like `same_type`

iox_catalog/src/lib.rs
```diff
@@ -28,150 +29,129 @@ pub mod interface;
...
-        new_schema.add_tables_and_columns(new_tables, new_columns);
-        return Ok(Some(new_schema));
+    if let Cow::Owned(table) = table {
+        // The table schema was mutated and needs inserting into the namespace
+        // schema to make the changes visible to the caller.
+        assert!(schema
+            .to_mut()
+            .tables
+            .insert(table_name.to_string(), table)
+            .is_some());
```
NGA-TRAN, line 151: Sorry but I really do not understand the purpose of the needs to insert this mutated table into the namespace

iox_catalog/src/lib.rs
```diff
@@ -211,94 +191,265 @@ mod tests {
...
+            ],
+        }
+    );
+
+    test_validate_schema!(
+        two_writes_conflicting_column_types,
+        lp = [
+            "m1,t1=a f1=2i",
+            // Second write has conflicting type for f1.
+            "m1,t1=a f1=2.0",
```
NGA-TRAN, line 427: 👍

iox_catalog/src/lib.rs
```diff
@@ -211,94 +191,265 @@ mod tests {
...
+        }
+    );
+
+    test_validate_schema!(
+        two_writes_tag_field_transposition,
+        lp = [
+            // x is a tag
+            "m1,t1=a,x=t f1=2i",
+            // x is a field
+            "m1,t1=a x=t,f1=2i",
```
NGA-TRAN, line 445: 👍

NGA-TRAN (approved):
Looks good. The test framework and  test coverage look great. I just have a suggestion and a question.

iox_catalog/Cargo.toml
domodwyer: What editor do you use? I want diffs!

iox_catalog/src/interface.rs
domodwyer: Yeah makes sense, bit of a overstretched overloading 😂 

I'll push a commit in a sec 👍

iox_catalog/Cargo.toml
alamb: In general I am a fan of reducing the time taken to compile iox, so if a crate is used just to add some simply functionalty that could be copy/pasted or reimplemented easily I prefer to do the latter

However, fancy diffs doesn't seem to fall into that category so if it helps @domodwyer  be more productive I think the new dependency is reasonable

iox_catalog/src/lib.rs
domodwyer: That's OK! I'll try to explain:

This function [takes](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L67) a [`Cow<NamespaceSchema>`](https://doc.rust-lang.org/std/borrow/enum.Cow.html) type that provides _copy-on-write_ semantics. This matches perfectly with the behaviour of the existing function (which I retained in the refactor) where a new, updated `NamespaceSchema` is returned to you if the one you provided needed to be updated (this has some caching advantages such as not needing an exclusive `&mut` reference to the data which would preclude sharing it across many requests in parallel). We use the same trick to lazily clone the table schema only when needed to update it.

When we enter this fn, we lookup the table in the `NamespaceSchema` given in the fn arguments - and if it exists we [borrow it](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L74-L75) for the rest of the validation process - this means it gets wrapped in a `Cow::Borrowed(table_schema)`.

We then run the validation process and can take one of two paths:

* We either observe no new columns so we do not need to update the table schema
* We have to add a new column and need to update the table schema

For the first case, we don't have to change anything, so we don't have to return a new schema to the user - what they had is fine - we should avoid cloning the table schema in this case as we will simply throw it away unchanged.

For the second case, we need to copy the existing table schema and insert the new column into it. To obtain a new, mutable copy of the table schema, we call [`to_mut()`](https://doc.rust-lang.org/std/borrow/enum.Cow.html#method.to_mut) which turns it into an owned, mutable copy (copy-on-write!) - we do that [here](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L148).

Assuming we updated the table schema, at the end of the validation fn we now have TWO table schemas - the original one we [borrowed at the start of the fn](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L75) and the [**copy we created**](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L96) in order to add new columns to it. [If we have](https://github.com/influxdata/influxdb_iox/blob/3fce437b163a4d109d5a05407f1a874d58712898/iox_catalog/src/lib.rs#L144) a `Cow::Owned(table_schema)` then the `table_schema` in it is the copy we created and we need to put it back into the `NamespaceSchema` the caller gave us, replacing the old unmodified version with our new, modified copy so they can see it!

iox_catalog/src/interface.rs
domodwyer: I have pushed the change in 02b83e0eefee608cd1916fedbe4b8a6da88d4170 👍

iox_catalog/src/interface.rs
alamb: I believe this is something @jacobmarble  added to handle a different "IOx native" ingest format (other than line protocol). I don't think it is currently used

alamb: Added in https://github.com/influxdata/influxdb_iox/pull/1842 I think

domodwyer: Thanks! I'll chat to @jacobmarble when he's back 👍

pauldix (approved):

iox_catalog/src/interface.rs
domodwyer: Awesome - thanks for clarifying.

I created https://github.com/influxdata/influxdb_iox/issues/3562 and will PR it tomorrow 👍

iox_catalog/Cargo.toml
mkmik: > What editor do you use? I want diffs!
> 
> edit: I guess it'd also be helpful in CI failures in that case if it's IDE specific

CLion

image

image

image


Unfortunately I had to switch to vscode because I can't use CLion remotely on my wife's M1 mac mini and compiling rust on my laptop is just too slow for my nerves. And vscode apparently doesn't know how to do that trick :-(

mkmik: I tried https://marketplace.visualstudio.com/items?itemName=ryu1kn.partial-diff but it's not very practical because a) you have to select the two snippets manually b) it doesn't work in the terminal pane where the cargo test results end up

Issue #3434, opened 2022-01-10
InfluxQL predicate with negated regexp and empty tag (TestServer_Query_With_EmptyTags)


alamb:
Similar to #3430 -- read https://github.com/influxdata/influxdb_iox/issues/3430#issuecomment-1009986314 before working on this one
```
    --- FAIL: TestServer_Query_With_EmptyTags/where_regex_exact_(not) (0.02s)
```

__Steps to reproduce:__
Use data like:

```
cpu value=1
cpu,host=server01 value=2
```

Run an influxQL query

```sql
-- "where regex exact (not)"
select value from cpu where host !~ /^server01$/
-- "where regex at least one char",
select value from cpu where host =~ /.+/
```

__Expected behaviour:__
The series for the `cpu value=1` should be returned

			exp:     `{"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","value"],"values":[["2009-11-10T23:00:02Z",1]]}]}]}`,

__Actual behaviour:__
Nothing is returned

__Expected behaviour:__
The series for the `cpu value=1` should be returned
```
"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","value"],"values":[["2009-11-10T23:00:02Z",1]]}]}]
```

Issue #3430, opened 2022-01-07
InfluxQL support for tag predicate against empty string (TestServer_Query_With_EmptyTags)


alamb:
Similar in issue to #3434  -- maybe need the same fix

__Steps to reproduce:__
Use data like:

```
cpu value=1 
cpu,host=server01 value=2
```

Run an influxQL query

```sql
select value from cpu where host = ''
```

__Expected behaviour:__
The series for the `cpu value=1` should be returned

			exp:     `{"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","value"],"values":[["2009-11-10T23:00:02Z",1]]}]}]}`,

__Actual behaviour:__
Nothing is returned

__Environment info:__
This is from the `TestServer_Query_With_EmptyTags/where_empty_tag` test

alamb:
IOx sees the predicate `exprs: [#host = Utf8("")]` 

```
2022-01-07T20:00:10.697631Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: read filter db_name=54425547585a6420_000000000153158f req.range=Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }) predicate=((TagRef:_m[0x00] == "cpu") AND node: NONE)
2022-01-07T20:00:10.698190Z DEBUG query::frontend::influxrpc: planning read_filter predicate=Predicate { table_names: Some({"cpu"}), field_columns: Some({"value"}), partition_key: None, range: Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }), exprs: [#host = Utf8("")], value_expr: [] }
```

However, since IOx's model treats a missing column as null (rather than as an empty string) the predicate ends up being `where host = ""` rather than `where host IS NULL`

To fix this we need to teach the influxrpc layer to rewrite tag = "" predicates into `tag IS NULL` type predicates  I think

alamb:
The core issue here appears to be a mismatch in semantics when a column is missing.

IOx treats a missing column (`tag`, `field`) as `NULL`, which is inline with 

However, the storagerpc query language appears to expect that the value of a missing tag will be represented as the empty string `''`. This leads to the discrepancy in the test:

For this series (which does not have a `host` tag):
```
cpu value=1 
```

A sql predicate of `host = ''` would be treated like `NULL = ''`, (which is NULL / not true), which *WILL NOT* return the series.

[...]

PR #3557, opened 2022-01-27
fix(InfluxQL): treat null tags as `''` rather than `null` in storagerpc queries

alamb:
# Rationle

See doc: https://docs.google.com/document/d/1Od7J5iu8eWm9CoUufwKUYhw0tpQpSMMWJ5n03SHGP54/edit#

closes https://github.com/influxdata/influxdb_iox/issues/3436
closes https://github.com/influxdata/influxdb_iox/issues/3430
closes https://github.com/influxdata/influxdb_iox/issues/3434

# Changes
## Predicate gRPC translation

Rewrites references to tags into case statements

```
TagRef(tag_col) = "cpu"
```

```
CASE
  WHEN tag_col IS NULL THEN ''
  ELSE tag_col
END = "cpu"
```

## Shenanigans
And adds shenanigans to ensure that by the time the ReadBuffer, pruning or parquet sees a predicate (for a metadata query or a scan), it will almost never have a `CASE` statement in it.

# Implementation checklists
- [x] Add end to end tests: https://github.com/influxdata/influxdb_iox/pull/3563
- [x] Add gRPC frontend translation (this PR)
- [x] ~Support predicate pruning with CASE ... expressions: https://github.com/apache/arrow-datafusion/issues/1692~ -- decided we didn't need this if we use [this strategy](https://docs.google.com/document/d/1Od7J5iu8eWm9CoUufwKUYhw0tpQpSMMWJ5n03SHGP54/edit#)
- [x] Simplify `CASE ...` expressions: https://github.com/apache/arrow-datafusion/issues/1693
- [x] Add a public Expr simplification API https://github.com/apache/arrow-datafusion/issues/1694
- [x] Fold past CASE blocks / translate to boolean CASE: https://github.com/influxdata/influxdb_iox/issues/3585
- [x] Apply simplifications to expressions with chunk schema prior to passing to read buffer / parquet  https://github.com/influxdata/influxdb_iox/pull/3588
- [x] https://github.com/influxdata/influxdb_iox/pull/3602


alamb:
The story so far:
1. All tests pass (including newly added ones for matching on empty tags) 🎉 
2. However, I am very worried about the performance implications of committing this PR because of the performance implications.

Specifically there are several places where `column <op> constant` are highly optimized that will not recognize tags any longer when the column reference is wrapped in this CASE .. nonsense.

The two places I am worried about are
1. chunk pruning
2. ReadBuffer metadata query + scan

The good news is that in almost all cases, a chunk will likely have all values for the tag column (or no values for the tag column) and we can simplify the expression in that case.


So that leaves two blockers to merging this PR in my mind:
1. Make chunk pruning work with these expressions
2. Figure out how to simplify the exprssion prior to handing them to the read buffer

Both of these will likely require changes in DataFusion upstream.

alamb:
Update: We have a new plan (thanks to @tustvold) for removing `CASE` statements: https://docs.google.com/document/d/1Od7J5iu8eWm9CoUufwKUYhw0tpQpSMMWJ5n03SHGP54/edit#

It seems like it should be good!

predicate/src/rewrite.rs
```diff
@@ -144,6 +164,110 @@ fn inline_case(case_on_left: bool, left: Expr, right: Expr, op: Operator) -> Exp
...
+    match expr {
+        Expr::BinaryExpr { left, op: _, right } if is_lit(right) => is_col(left),
+        Expr::BinaryExpr { left, op: _, right } if is_lit(left) => is_col(right),
+        _ => None,
+    }
+}
+
+impl ExprRewriter for IOxPredicateRewriter {
+    fn mutate(&mut self, expr: Expr) -> Result<Expr> {
+        // look for this structure:
```
alamb, line 218: This is they very definition of a special case 🤔  But I think in this case the potential performance gain is worth it. It would be great to make this general purpose somehow

alamb:
I am happy to say this PR is now ready to go once all the upstream DataFusion PRs are merged and the PRs that depend on it are merged to main

query/src/pruning.rs
```diff
@@ -626,6 +627,55 @@ mod test {
         assert_eq!(names(&pruned), vec!["chunk2", "chunk3"]);
     }
 
+    #[test]
+    fn test_pruned_is_null() {
+        test_helpers::maybe_start_logging();
+        // Verify that type of predicate is pruned if column1 is null
+        // (this is a common predicate type created by the INfluxRPC planner)
```
alamb, line 634: it actually turns out that by the time chunk pruning happens this has been simplified away, but I had already written the test case and figured it would be good to leave in

influxdb_iox/tests/end_to_end_cases/influxrpc.rs
```diff
@@ -340,8 +340,6 @@ pub async fn read_filter_regex_operator() {
 }
 
 #[tokio::test]
-// fixed in https://github.com/influxdata/influxdb_iox/pull/3557
```
alamb, line 343: tests now pass 🎉

predicate/src/rpc_predicate.rs
```diff
@@ -172,14 +173,23 @@ fn normalize_predicate(
             // column projection set.
             rewrite_field_column_references(&mut field_projections, e)
         })
+        .map(|e| {
```
alamb, line 176: I don't really like this chain.... I will think of how to make it less grotty

query/src/frontend/influxrpc.rs
```diff
@@ -2055,11 +2064,10 @@ mod tests {
 
         let actual_predicate = test_db.get_chunks_predicate();
 
-        // verify that the predicate was rewritten to foo = 'bar'
-        let expected_predicate = PredicateBuilder::new()
-            // (foo = 'bar') OR false
-            .add_expr(col("foo").eq(lit("bar")))
-            .build();
+        // verify that the predicate was rewritten to (NOT col IS NULL) AND (foo = 'bar')
```
tustvold, line 2067: I think this comment is out of date?

predicate/src/rpc_predicate.rs
```diff
@@ -172,14 +173,23 @@ fn normalize_predicate(
...
+            // apply IOx specific rewrites (that unlock other simplifications)
+            rewrite::rewrite(e).expect("rewrite failed")
+        })
         .map(|e| {
             if let Some(schema) = &schema {
-                e.simplify(&SimplifyAdapter::new(schema.as_ref()))
-                    .expect("Expression simplificiation failed")
+                let adapter = SimplifyAdapter::new(schema.as_ref());
+                // simplify twice to ensure "full" cleanup
+                e.simplify(&adapter)
```
tustvold, line 184: Just noting that we will need to be aware that predicate simplification takes place as part of the RPC predicate conversion, and therefore will not apply to SQL. This isn't a bad thing, just noting it :smile:

query/src/pruning.rs
tustvold: I'm right in thinking in this test case it won't be simplified away, but would be if the predicate followed the normal RPC predicate path?

tustvold (approved):

query/src/frontend/influxrpc.rs
alamb:
```suggestion
        // verify that the predicate was rewritten to `foo = 'bar'`
```

query/src/pruning.rs
alamb: yes

PR #3558, opened 2022-01-27
feat: router2 sharder benchmarking

lukebond:
First pass on some basic benchmarking for the router2 sharder.


domodwyer:
I did not notice the `automerge` but I don't think anyone will object 👍

lukebond:
can add that as an additional set of benchmarks in another PR :+1:

PR #3560, opened 2022-01-27
fix: chunk ordering

tustvold:
When compacting or persisting chunks we take a consecutive sequence of chunks as determined by chunk order. For example, if the catalog contained chunks (order: 1, id: 1), (order: 1, id: 2), (order: 2, id: 3), you could take ids [1, 2], [1, 2, 3] or [3] but not [1], [2], [2, 3], [1, 3] etc... 

As such the reasoning went that you could select an arbitrary order from this slice to use for the resulting chunk(s). 

Unfortunately this missed a detail of persist, it takes a consecutive slice of **unpersisted** chunks and completely ignores the orders of persisted chunks. This meant you could end up with ambiguous ordering of overlapping persisted data. 

In particular consider the following sequence:

* Write into chunk (order: 1, id: 0)
* Persist part of (order: 1, id:0) to yield (order: 1, id:1) and (order: 1, id: 2) where the former is persisted
* Write into a new chunk (order: 2, id: 3)
* Persist data from (order: 1, id: 2) and (order: 2, id: 3) to (order: ?, id: 4)

If you take the minimum order (1) here you've introduced ambiguity, as the data written by the second persist may overlap with that of the first persist.

Fortunately, the fix is incredibly simple - take the maximum chunk order of the input instead of the minimum. This ensures that new data always increases the chunk order, avoiding ambiguity. 

It is still possible to persist two chunks with the same order, but new data can't be included in the second persist that wasn't present in memory at the time of the first, and we never persist data that overlaps with what we currently have in memory.

_This should also reduce situations where chunks are forced to participate in compaction on account of having an overlapping chunk order._


db/src/lifecycle/compact.rs
```diff
@@ -47,7 +47,7 @@ pub(crate) fn compact_chunks(
     let mut time_of_first_write: Option<Time> = None;
     let mut time_of_last_write: Option<Time> = None;
     let mut delete_predicates_before: HashSet<Arc<DeletePredicate>> = HashSet::new();
-    let mut min_order = ChunkOrder::MAX;
+    let mut max_order = ChunkOrder::MIN;
```
alamb, line 50: Given the subtleties involved, I think it would very much help to add the description (and the scenarios / examples specifically) from the PR as comments in the code

db/src/lifecycle/persist.rs
```diff
@@ -50,7 +50,7 @@ pub fn persist_chunks(
     let mut time_of_last_write: Option<Time> = None;
     let mut query_chunks = vec![];
     let mut delete_predicates_before: HashSet<Arc<DeletePredicate>> = HashSet::new();
-    let mut min_order = ChunkOrder::MAX;
+    let mut max_order = ChunkOrder::MIN;
```
alamb, line 53: not sure how, but it would be great to get the comments here too (maybe pull the code that computes the output order into its own `ChunkOrderer` struct and beaucoup de comments?

alamb (approved):
I read the description of this PR and it makes sense to me

I reviewed that the code implements what is described in the description 👍

tustvold:
I'm going to lump the documentation portion into https://github.com/influxdata/influxdb_iox/issues/2047 and potentially pick that up next week time permitting

crepererum (approved):

PR #3563, opened 2022-01-27
test: add end to end tests that query missing tags

alamb:
Add the end to end tests (marked as `#ignore`) from https://github.com/influxdata/influxdb_iox/pull/3557 to keep the size of that PR down.

re #3436


influxdb_iox/tests/end_to_end_cases/influxrpc.rs
```diff
@@ -317,45 +317,85 @@ async fn measurement_fields_endpoint(
...
-    let mut storage_client = StorageClient::new(fixture.grpc_channel());
-    let mut write = fixture.write_client();
-
-    let scenario = Scenario::new();
-    scenario.create_database(&mut management).await;
-
-    load_read_group_data(&mut write, &scenario).await;
-
-    let read_source = scenario.read_source();
+pub async fn read_filter_regex_operator() {
```
alamb, line 320: this was refactored so I didn't end up repeating the same thing over and over again

influxdb_iox/tests/end_to_end_cases/influxrpc.rs
```diff
@@ -317,45 +317,88 @@ async fn measurement_fields_endpoint(
...
+        TimestampRange {
             start: 0,
             end: 2001, // include all data
-        }),
-        predicate: Some(make_regex_match_predicate("host", "^b.+")),
-        ..Default::default()
-    };
+        },
+        // https://github.com/influxdata/influxdb_iox/issues/3430
+        // host = '' means where host is not present
```
NGA-TRAN, line 353: I think the delete bug I just mentioned in slack is due to this. I had 2 batches and one does not have the tag the delete predicate is applied to. Maybe after this fix that test will pass. Will create a ticket for it

NGA-TRAN (approved):
LGTM

influxdb_iox/tests/end_to_end_cases/influxrpc.rs
alamb: I think it is deeper than just deletes -- influxql expects that `tag = ''` matches for a series without a tag (aka when the data was never present not just when it was deleted)

PR #3567, opened 2022-01-28
feat: two pass chunk prune

tustvold:
Currently when pruning chunks the system will collect the statistics for all columns mentioned by any of the predicates across all chunks in a table. It will then evaluate all the predicates against this `RecordBatch` of aggregated statistics. This is fantastic because it gives the most accurate prune possible, handling complex predicate expressions, etc... However, it is relatively expensive to do for every chunk in the table.

As it happens most queries will have some sort of predicate on time, and this will typically prune out large numbers of chunks. We can therefore do an initial pass with just the time predicates, before performing the more expensive prune on what is left.

This is partly an experiment, I'm not entirely sure what difference this will make, but thought it was simple enough to be worth a shot :grin:

### DIFF
--- a/query/src/pruning.rs
+++ b/query/src/pruning.rs
@@ -8,13 +8,14 @@ use arrow::array::{
 use arrow::datatypes::{DataType, Int32Type, TimeUnit};
 
 use data_types::partition_metadata::{StatValues, Statistics};
+use datafusion::logical_plan::{Expr, Operator};
 use datafusion::{
     logical_plan::Column,
     physical_optimizer::pruning::{PruningPredicate, PruningStatistics},
 };
 use observability_deps::tracing::{debug, trace};
 use predicate::predicate::Predicate;
-use schema::Schema;
+use schema::{Schema, TIME_COLUMN_NAME};
 
 use crate::{group_by::Aggregate, QueryChunkMeta};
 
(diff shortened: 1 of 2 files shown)


query/src/pruning.rs
```diff
@@ -54,19 +55,67 @@ where
...
         }
     };
+
+    // A large number of chunks can likely be ruled out based on time alone
+    //
+    // Additionally, the timestamp statistics are cheaper to aggregate and
+    // evaluate predicates against than the strings found in tag columns
+    //
+    // We therefore do an initial pass with just the time predicates
+    if let Some(time_predicate) = extract_time_predicate(&filter_expr) {
```
alamb, line 65: I am of two minds here -- what you have will work in more cases than what influxQL creates, but it is also more expensive (as it goes off and invokes the general pruning predicate logic).

I wonder what you think of really special casing the `low <= time < high` type predicate and skip datafusion pruning all together:

```rust
if let Some(timestamp_range) = extract_time_predicate(&filter_expr) {
  chunks = chunks.iter()
    .filter(|chunk| chunk_in_range(chunk, timestamp_range)
    .collect()
}
```

That will lower the overhead of the pruning to about the minimum I can think of...

alamb (approved):
If we go with this approach, I think we should have at least some basic test it is working -- or if you want to see if it makes a difference first,  perhaps we can leave a ticket to polish or remove it after that measurement.

tustvold:
Going to implement this differently as suggested

PR #3568, opened 2022-01-28
refactor: decouple error types for DmlHandler

domodwyer:
This commit was supposed to be part of https://github.com/influxdata/influxdb_iox/issues/3549 but was lost during a rebase and is part of the same end goal - refactoring the `DmlHandler` to be an easily-composable trait.

---

* refactor: decouple error types for DmlHandler (b38deaa7)

      Allows the DmlHandler to return different types for each method.

      This enables a DmlHandler implementation decorating an inner handler to return
      the inner handler's error directly, avoiding any "wrapper" errors.


alamb (approved):
Small PR 👍

Issue #3570, opened 2022-01-28
SQL Chunk Pruning


tustvold:
Creating a ticket to track some limitations of the current approach, as fixing these isn't an immediate priority but they should be fixed at some point.

`DbSchemaProvider` snapshots the catalog in `SchemaProvider::table` limiting the predicates that are available at this point

`ChunkTableProvider::scan` adds the pushed down predicates to `Predicate::exprs` without extracting timestamp ranges. Perhaps as mentioned by @alamb on #3545  `Predicate` should just be a collection of `Expr` and have logic to extract the time range on demand, either way the mismatch between influxrpc and SQL is unfortunate

lesam:
This should really have been two issues I think. This one:

> ChunkTableProvider::scan adds the pushed down predicates to Predicate::exprs without extracting timestamp ranges. Perhaps as mentioned by @alamb on https://github.com/influxdata/influxdb_iox/pull/3545 Predicate should just be a collection of Expr and have logic to extract the time range on demand, either way the mismatch between influxrpc and SQL is unfortunate

is being addressed in https://github.com/influxdata/influxdb_iox/issues/5453 . I'm not sure if this one:

> DbSchemaProvider snapshots the catalog in SchemaProvider::table limiting the predicates that are available at this point

is still true or not.

PR #3571, opened 2022-01-28
feat: catalog timestamp pruning

tustvold:
This is an alternative to #3567 that instead pushes timestamp predicates down to the in-memory catalog.

This should be both faster, and simpler than #3567, but will currently only work for InfluxRPC. I've created #3570 to track fixing up the SQL logic to also take advantage of this.


data_types/src/partition_metadata.rs
```diff
@@ -377,6 +378,14 @@ impl Statistics {
             Self::String(v) => v.update_for_nulls(num_nulls),
         }
     }
+
+    /// Convert this to [`TimestampMinMax`] if possible
+    pub fn timestamp_min_max(&self) -> Option<TimestampMinMax> {
```
alamb, line 383: 👍

alamb (approved):
🎖️

PR #3576, opened 2022-01-28
feat(router2): write schema validation!

domodwyer:
Builds on the work in #3554 to perform schema validation on incoming writes in `router2`.

Failed schema validation logs at `WARN` - below is an example of two requests, with the second attempting to write a float value to an existing i64 column named `val`:

Screenshot 2022-01-28 at 16 34 46

The resulting HTTP 400 error body:

```json
{
  "code": "invalid",
  "message": "dml handler error: Column val is type i64 but write has type iox::column_type::field::float"
}
```

Because nothing is currently creatring the namespace records in the catalog, I used the following SQL to bootstrap the DB before sending the request:

```sql
INSERT INTO "iox_catalog"."query_pool" ("name") VALUES ('query');
INSERT INTO "iox_catalog"."kafka_topic" ("name") VALUES ('iox-shared');
INSERT INTO "iox_catalog"."namespace" ("name", "retention_duration", "kafka_topic_id", "query_pool_id") VALUES ('bananas_platanos', 'inf', 1, 1);
```

I shall automate this record creation in `router2` next.

---

* refactor: get_schema_by_name no Option (f9f9beac)

      Fetching a NamespaceSchema always succeeds and never returns None.

* feat: schema validation (c81f2072)

      Implements a write schema validation DML handler, denying requests that 
      conflict with the schema within the global catalog. Additive schema changes
      are accepted, incrementally updating the global catalog schema.

      Deletes are passed through unchanged and unvalidated.

* feat: cache NamespaceSchema in validator (65980237)

      Adds an in-memory cache of table schemas to the SchemaValidator DML handler.

      The cache pulls from the global catalog when observing a column for the first
      time, and pushes the column type to set it for subsequent requests if it does
      not exist (this pull & push is done by atomically by the catalog in an
      "upsert" call).

      The in-memory cache is sharded by namespace, with each shard guarded by an
      individual lock to minimise contention between readers (the expected average
      case) and writers (only when adding new columns/tables).

      Relies on the catalog to serialise new column creation and validate parallel
      creation requests.

* refactor: enable schema validation (39d489d9)

      Adds the SchemaValidator to the DML handler stack - this adds it into the
      request path in router2.


router2/src/dml_handlers/schema_validation.rs
```diff
@@ -0,0 +1,451 @@
...
+        let schema = self.cache.read().get(&namespace).map(Arc::clone);
+        let schema = match schema {
+            Some(v) => v,
+            None => {
+                // Pull the schema from the global catalog or error if it does
+                // not exist.
+                let schema = get_schema_by_name(&namespace, &*self.catalog)
+                    .await
+                    .map_err(|e| {
+                        warn!(error=%e, %namespace, "failed to retrieve namespace schema");
```
e-dard, line 145: Is this a common pattern? I guess I would expect either some logging or an error propagation. Could this logging be done where ever this error is handled (I assume it must be handled rather than catastrophic).

e-dard (approved):
Can't speak to the logic but I had a run through the code.

router2/src/dml_handlers/schema_validation.rs
domodwyer: It's not - I'm not in love with it either.

All these errors bubble up through the stack and [hit this fn](https://github.com/influxdata/influxdb_iox/blob/4744c5804e0d180b18e988ec8a7691d53fdb53f8/router2/src/server/http.rs#L71-L91) that turns them into HTTP errors which doesn't seem conducive to nice error handling either. Logging it here can capture the context as log fields at least...

Option to suggestions 👍

router2/src/dml_handlers/schema_validation.rs
```diff
@@ -0,0 +1,451 @@
...
+/// # Correctness
+///
+/// The correct functioning of this schema validator relies on the catalog
+/// implementation correctly validating new column requests against any existing
+/// (but uncached) columns, returning an error if the requested type does not
+/// match.
+///
+/// The catalog must serialise column creation to avoid `set(a=tag)` overwriting
+/// a prior `set(a=int)` write to the catalog (returning an error that `a` is an
+/// `int` for the second request).
```
NGA-TRAN, line 76: 👍

router2/src/dml_handlers/schema_validation.rs
NGA-TRAN: I think this kind of log is good to see unexpected things happening in the system

NGA-TRAN (approved):
Nice comments on implementation and the tests look good too

alamb:
Changed title to try and kick `Semantic Pull Request`, did not work sadly 😭

domodwyer:
I also tried, but I like your title addition 😂

alamb:
third-time was the charm!

PR #3579, opened 2022-01-28
fix: reduce overhead of creating physical expressions

alamb:
Draft until https://github.com/apache/arrow-datafusion/pull/1700 is merged

I am putting this up now while it is fresh in my mind.

# Rationale
Creating `ExecutionContextState`s is expensive and getting more so (see https://github.com/apache/arrow-datafusion/issues/1690)

# Changes
Only create what is needed and use a simpler API


query/src/util.rs
```diff
@@ -68,11 +69,7 @@ pub fn df_physical_expr(
     input: &dyn ExecutionPlan,
     expr: Expr,
 ) -> std::result::Result<Arc<dyn PhysicalExpr>, DataFusionError> {
-    // To create a physical expression for a logical expression we need appropriate
```
alamb, line 71: such overhead! at the moment on master this actually creates a tempfile!

alamb:
Superceded by https://github.com/influxdata/influxdb_iox/pull/3586/files

PR #3580, opened 2022-01-28
refactor: Start consolidating test scenarios in query_tests

alamb:
(This us a nice background task while waiting for some tests to compile...)

# Rationale
1. If all scenarios are defined in the same place, it is easier to find an existing scenario than write a new one
2. I would like to remove the use of `async_trait` in favor of a straight up Future (to improve compile speeds) and I want them all in the same place
3. https://github.com/influxdata/influxdb_iox/pull/3489 got too big  / attracted merge conflicts, so I am going to do this incrementally instead

Possible downsides:
1. having the data and expected results in different files may not be as convenient, but I think the ongoing development improvement will be worth it.

This consolidation also will make it clear the scale of the challenge the rust compiler is facing.

# Changes:
Move code around (no logical change intended)


NGA-TRAN (approved):
The moves to library and utils make sense to me

PR #3584, opened 2022-01-31
feat: allow IOx catalog to setup itself (no SQLx CLI required)

crepererum:
- DB creation must still be done externally. IMHO it's part of the infrastructure and it's unlikely that the IOx user has sufficient permissions to perform this operation.
- the new functionality is used by the test suite, but will later be exposed as a CLI command


iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -1,7 +1,5 @@
 -- Add migration script here
 -- iox_shared schema
-BEGIN;
```
crepererum, line 3: SQLx already sets up a transaction for us. Starting yet another one leads to warnings on the Postgres side.

iox_catalog/src/postgres.rs
```diff
@@ -43,7 +46,7 @@ impl PostgresCatalog {
                     // Tag the connection with the provided application name.
                     c.execute(sqlx::query("SET application_name = '$1';").bind(app_name))
                         .await?;
-                    let search_path_query = format!("SET search_path TO {}", schema_name);
+                    let search_path_query = format!("SET search_path TO public,{}", schema_name);
```
crepererum, line 49: This is required because the migration tracking table (`_sqlx_migrations`) can currently only live within the `public` schema. The IOx tables will be inserted into `iox_catalog`.

domodwyer:
I think we can use the built-in [sqlx migrator](https://docs.rs/sqlx/0.4.0-beta.1/sqlx/macro.migrate.html) macro to handle the leg work - we can avoid having to read the SQL files, build a rs file from it, etc.

In conductor, we just do this:

```rust
static MIGRATOR: Migrator = sqlx::migrate!("./migrations");
```

and add `println!("cargo:rerun-if-changed=migrations");` to the `build.rs` so it picks up the changes as needed - no need to manually build up the files ourselves.

crepererum:
@domodwyer done :)

domodwyer (approved):
🚀

NGA-TRAN (approved):
I was about to clarify the work to find .sql files, gather their content into, and then resolve. But then they were all disappear (I guess it was addressed @domodwyer's comment). Since this is just a setup, I think it looks good. If there does not exists migration tests, ask them in the follow-up PR to verify if this works?

crepererum:
> If there does not exists migration tests, ask them in the follow-up PR to verify if this works?

Migrations are tested, at least to the point that they create a sane DB state ("empty DB => schemas are present"). I deem this sufficient for now. If add more migrations in the future, I  think we need a more sophisticated migration testing setup including data fixtures, but that's something we can solve later (it's also pointless to build this now when we don't even have a real migration yet).

PR #3588, opened 2022-01-31
feat: Simplify predicates in the `InfluxRpcFrontend` before using them

alamb:
~Draft until upstream DataFusion PR https://github.com/apache/arrow-datafusion/pull/1717 merges~

Part of https://github.com/influxdata/influxdb_iox/pull/3557 / https://github.com/influxdata/influxdb_iox/issues/3436

# Rationale
The plan for handling the last influxql simplification involves using a `CASE` statement and then rewriting it away to nothing most of the time. 

# Changes
1. Split `QueryDatabase` trait into the part that is needed to rewrite predicates (list of table names and table schema)
2. Apply `Expr::simplify` to each table normalized predicate
3. Tests for same


db/src/access.rs
```diff
@@ -310,6 +299,19 @@ impl QueryDatabase for QueryCatalogAccess {
     }
 }
 
+impl PredicateTableInfo for QueryCatalogAccess {
```
alamb, line 302: Moved the functions into a subset trait

predicate/src/rpc_predicate.rs
```diff
@@ -166,10 +190,52 @@ fn normalize_predicate(table_name: &str, predicate: &Predicate) -> Predicate {
             None => predicate.field_columns = Some(field_projections),
         };
     }
-
     predicate
 }
 
+struct SimplifyAdapter {
+    // TODO avoid re-creating this each time....
```
alamb, line 197: The way to do so would be to cache the structure in the catalog. I will do so if we detect this taking too long during planning

predicate/src/rpc_predicate.rs
```diff
@@ -156,6 +172,14 @@ fn normalize_predicate(table_name: &str, predicate: &Predicate) -> Predicate {
             // column projection set.
             rewrite_field_column_references(&mut field_projections, e)
         })
+        .map(|e| {
+            if let Some(schema) = &schema {
+                e.simplify(&SimplifyAdapter::new(schema.as_ref()))
```
alamb, line 177: here is the call to simplify!

predicate/src/rpc_predicate.rs
alamb: Or make a better way in DataFusion: https://github.com/apache/arrow-datafusion/issues/1725

```suggestion
    // TODO avoid re-creating this each time....
    // https://github.com/apache/arrow-datafusion/issues/1725
```

alamb: https://github.com/apache/arrow-datafusion/pull/1726

alamb:
I am happy to improve / change this API -- if anyone has thoughts or suggestions please let me know

query/src/frontend/influxrpc.rs
```diff
@@ -1945,4 +1951,120 @@ mod tests {
...
+        F: FnOnce(&TestDatabase, InfluxRpcPredicate) + Send,
+    {
+        let chunk0 = Arc::new(
+            TestChunk::new("h2o")
+                .with_id(0)
+                .with_tag_column("foo")
+                .with_time_column(),
+        );
+
+        let silly_predicate = PredicateBuilder::new()
```
alamb, line 2042: I am particularly proud of this test that shows the predicates being simplified in the rpc planner prior to being passed to the chunks

predicate/src/rpc_predicate.rs
```diff
@@ -166,10 +190,53 @@ fn normalize_predicate(table_name: &str, predicate: &Predicate) -> Predicate {
...
         };
     }
-
     predicate
 }
 
+struct SimplifyAdapter {
+    // TODO avoid re-creating this each time....
+    // https://github.com/apache/arrow-datafusion/issues/1725
+    df_schema: DFSchema,
```
tustvold, line 199: TIL, why does DataFusion have its own schema type? Could we make the arrow-rs schema more extensible?

tustvold (approved):
Very nice :+1: 

I think it would be even cooler to have a test of the case goop gubbins that we now generate, but not sure what remains to go in to enable this

predicate/src/rpc_predicate.rs
alamb: > TIL, why does DataFusion have its own schema type? Could we make the arrow-rs schema more extensible?

It was because the schema needs a relation (aka "table") name

alamb:
> I think it would be even cooler to have a test of the case predicate gubbins that we now generate for tags, but not sure what remains to go in to enable this

I believe there is a test like this in https://github.com/influxdata/influxdb_iox/pull/3557 which I will highlight

alamb:
>> I think it would be even cooler to have a test of the case predicate gubbins that we now generate for tags, but not sure what remains to go in to enable this

> I believe there is a test like this in #3557 which I will highlight

While there is a test that covers this case in #3557  it is implicit -- I have also written an explicit test here:

https://github.com/influxdata/influxdb_iox/pull/3619

PR #3589, opened 2022-01-31
feat: Implement deserializing IoxMetadata from protobuf

### LINKED ISSUE #3587: Implement deserialization of new `IoxMetadata` type
Like `IoxMetadataOld::from_protobuf`; this will be useful for round-trip tests and the query path side will need this eventually.

carols10cents:
Fixes #3587.


parquet_file/src/metadata.rs
```diff
@@ -532,12 +535,72 @@ impl IoxMetadata {
...
         uuid == self.object_store_id
     }
 }
 
+/// Parse big-endian UUID from protobuf.
+pub fn parse_uuid(bytes: &[u8]) -> Result<Option<Uuid>> {
+    if bytes.is_empty() {
+        Ok(None)
+    } else {
+        let uuid = Uuid::from_slice(bytes).context(UuidParseSnafu {})?;
```
crepererum, line 599: I've ran into this issue as well and find it a bit frustrating: every 128bit integer (or equivalent bytes vector) is in theory a valid UUID, even though the type might be unknown and the UUID crate actually only fails when the length of the slice is wrong:

https://github.com/uuid-rs/uuid/blob/2d6c147bdfca9612263dd7e82e26155f7ef8bf32/src/builder.rs#L309-L317

However protobuf neither provides us w/ 128bit integers nor with fixed-size vectors and also doesn't have some "well known" UUID type (see https://github.com/protocolbuffers/protobuf/issues/2224). And hand-rolling structs (e.g. by splitting the int into fixed-size high/low 64bit integers, see protobuf issue) seems a bit excessive as well.

So I guess reading UUIDs can fail :/

crepererum (approved):

NGA-TRAN (approved):
LGTM

PR #3590, opened 2022-02-01
refactor: reusable ingester/router2 CLI pieces

crepererum:
Sharing is caring.

Write buffer configs and catalog DSN handling are shared between ingester and router2. The DSN is also going to be used by admin/debug CLI commands, so make them reusable (which also ensures that args/envs are homogeneous).


influxdb_iox/src/clap_blocks/catalog_dsn.rs
```diff
@@ -0,0 +1,25 @@
...
+use iox_catalog::{
+    interface::{Catalog, Error},
+    postgres::PostgresCatalog,
+};
+
+/// CLI config for catalog DSN.
+#[derive(Debug, Clone, clap::Parser)]
+pub struct CatalogDsnConfig {
+    /// Postgres connection string
+    #[clap(long = "--catalog-dsn", env = "INFLUXDB_IOX_CATALOG_DNS")]
```
lukebond, line 12: `s/DNS/DSN/` here ( i appreciate that this error existed in the code you copied!)

lukebond (approved):
seems sensible to me!

influxdb_iox/src/clap_blocks/catalog_dsn.rs
crepererum: fixed

PR #3594, opened 2022-02-01
feat: add support for parsing sql expressions into RPCNode

e-dard:
This PR adds support for converting a string expression into an InfluxRPC Predicate node.

## Rationale

IOx implements support for the InfluxDB TSM based gRPC read API. Therefore it can support being a backend for Flux and InfluxQL queries. However, IOx itself does not contain a Flux or InfluxQL query engine. As a product developer and an operator this makes it hard to issue arbitrary gRPC requests to IOx.

Whilst we do now have some ability to replay entire gRPC requests that have been logged in the `queries` system table, it would be of considerable use if we could issue stand-alone requests from the CLI. It would also open the door up to having more data-driven gRPC tests in IOx, much like we do for SQL.

## Changes

This PR introduces an API that lets one specify an expression as a string, which very closely matches the syntax format of an [InfluxQL WHERE clause](https://docs.influxdata.com/influxdb/v1.8/query_language/explore-data/#the-where-clause).

The API then returns an InfluxRPC Predicate node. 

For example, you can provide the following types of clauses:

```
// tag matches
server = 'foo'

// field expressions (using the `::field` syntax)
server = 'foo' and "temp"::field > 100.2

// spaces in tag keys
"my host with spaces" = 'bar'

// logical expressions
"expression_1" = 'foo' AND "expression_2" = 'bar

// expressions with more than two children
"tk1" = 'a' OR "tk2" = 'a' OR "tk3" = 'a'

// precedence via parentheses
"server" = 'a' OR ("temp"::field > 12.3 AND "host" = 'foo')

// special tag keys
"_measurement" = 'cpu' OR "_field" = 'temp'
```

## Next Steps

The follow on to this work is to start working on the CLI command that will use this API and build the gRPC requests themselves.

The API doesn't yet support regex operators or specifying unsigned integers.

### DIFF
--- a/predicate/Cargo.toml
+++ b/predicate/Cargo.toml
@@ -9,6 +9,7 @@ chrono = { version = "0.4", default-features = false }
 data_types = { path = "../data_types" }
 datafusion = { path = "../datafusion" }
 datafusion_util = { path = "../datafusion_util" }
+generated_types = { path = "../generated_types" }
 itertools = "0.10"
 observability_deps = { path = "../observability_deps" }
 ordered-float = "2"
--- a/predicate/src/sql.rs
+++ b/predicate/src/sql.rs
@@ -1,0 +1,370 @@
+//! This module has logic to translate a sub-set of DataFusion expressions into
+//! RPC Nodes (predicates).
+use generated_types::{
+    node::Comparison as RPCComparison, node::Logical as RPCLogical, node::Type as RPCType,
+    node::Value as RPCValue, Node as RPCNode,
+};
+
+use snafu::Snafu;
+use sqlparser::ast::{BinaryOperator as Operator, Expr, Ident, Value};
+
+const TAG_KEY_FIELD: [u8; 1] = [255];
+const TAG_KEY_MEASUREMENT: [u8; 1] = [0];
+
+/// Parse Error
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("unable to parse '{}' into numerical value", value))]
+    ParseError { value: String, msg: String },
+
+    #[snafu(display("unexpected value '{:?}'", value))]
+    UnexpectedValue { value: Value },
+
+    #[snafu(display("unexpected expression type: '{:?}'", expr))]
+    UnexpectedExprType { expr: Expr },
+
+    #[snafu(display("unexpected operator: '{:?}'", op))]
+    UnexpectedBinaryOperator { op: Operator },
+
+    #[snafu(display(
+        "unsupported identifier type: '{:?}'. Supported Types: field, tag",
+        ident
+    ))]
+    UnsupportedIdentType { ident: String },
+}
+
+/// Result type for Parser Cient
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Converts a sub-set of Datafusion expressions into RPCNodes.
+pub fn expr_to_rpc_node(expr: Expr) -> Result<RPCNode> {
+    build_node(&expr)
+}
+
+// Builds an RPCNode given the value Expr and the converted children
+fn build_node(expr: &Expr) -> Result<RPCNode> {
+    match expr {
+        Expr::Cast { expr, data_type } => match data_type {
+            sqlparser::ast::DataType::Custom(ident) => {
+                if !ident.0.is_empty() {
+                    if let Some(Ident { value, .. }) = ident.0.get(0) {
+                        match value.as_str() {
+                            "field" => {
+                                // extract field key identifier
+                                if let Expr::Identifier(field) = &**expr {
+                                    return make_leaf(
+                                        RPCType::FieldRef,
+                                        RPCValue::FieldRefValue(field.value.clone()),
+                                    );
+                                }
+                                return UnexpectedExprTypeSnafu {
+                                    expr: *expr.clone(),
+                                }
+                                .fail();
+                            }
+                            "tag" => {
+                                // extract tag key identifier
+                                if let Expr::Identifier(tag) = &**expr {
+                                    return make_leaf(
+                                        RPCType::TagRef,
+                                        RPCValue::TagRefValue(make_tag_name(tag.value.clone())),
+                                    );
+                                }
+                                return UnexpectedExprTypeSnafu {
+                                    expr: *expr.clone(),
+                                }
+                                .fail();
+                            }
+                            _ => todo!(),
+                        }
+                    }
+                }
+
+                UnsupportedIdentTypeSnafu {
+                    ident: ident.to_string(),
+                }
+                .fail()
+            }
+            _ => UnsupportedIdentTypeSnafu {
+                ident: data_type.to_string(),
+            }
+            .fail(),
+        },
+        Expr::Identifier(c) => {
+            // See https://docs.influxdata.com/influxdb/v1.8/query_language/explore-data/#syntax
+            // This allows one to specify whether a column name should map to
+            // a field or tag.
+            if let Some(stripped) = c.value.strip_suffix("::field") {
+                return make_leaf(
+                    RPCType::FieldRef,
+                    RPCValue::FieldRefValue(stripped.to_owned()),
+                );
+            }
+            make_leaf(
+                RPCType::TagRef,
+                RPCValue::TagRefValue(make_tag_name(c.value.clone())),
+            )
+        }
+        Expr::Value(v) => match v {
+            Value::Boolean(b) => make_lit(RPCValue::BoolValue(*b)),
+            Value::Number(n, _) => make_lit(parse_number(n)?),
+            Value::DoubleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::SingleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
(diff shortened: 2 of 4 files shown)


predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+///   * Numbers are parsed into integers where possible, but fall back to floats
+///   * Use parentheses to denote precedence.
+///   * _measurement and _field will be correctly converted into the binary format.
+///  
+/// Unsupported:
+///   * Regex operators are not yet supported.
+///   * Unsigned integers cannot yet be supported because there is no current
+///     way to denote them (we need to add a `u` suffix support).
+///
+pub fn expr_to_rpc_predicate(expr: &str) -> Result<RPCPredicate> {
```
e-dard, line 72: This is the main API method that the CLI command will make use of.

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+    Ok(RPCPredicate {
+        root: Some(build_node(&parser.parse_expr().context(ExprParseSnafu)?)?),
+    })
+}
+
+// Builds an RPCNode given the value Expr and the converted children
+fn build_node(expr: &Expr) -> Result<RPCNode> {
+    match expr {
+        Expr::Nested(expr) => make_node(RPCType::ParenExpression, vec![build_node(expr)?], None),
+        Expr::Cast { expr, data_type } => match data_type {
```
e-dard, line 87: When you issue `something::field` it's treated as a `CAST`. We can utilise this variant to build the correct `TagRef` expression (tag filters) or `FieldRef` expression (field filters).

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+                    ident: ident.to_string(),
+                }
+                .fail()
+            }
+            _ => UnsupportedIdentTypeSnafu {
+                ident: data_type.to_string(),
+            }
+            .fail(),
+        },
+        Expr::Identifier(c) => {
```
e-dard, line 131: tag filters without an explicit `::tag`

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+            Value::SingleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::HexStringLiteral(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::NationalStringLiteral(v) => make_lit(RPCValue::StringValue(v.clone())),
+            _ => UnexpectedValueSnafu {
+                value: v.to_owned(),
+            }
+            .fail(),
+        },
+        Expr::BinaryOp { left, op, right } => {
+            build_binary_node(build_node(left)?, op.clone(), build_node(right)?)
```
e-dard, line 152: Because this API is planned to mainly be used for tests and CLI interaction I'm not overly concerned about the style of recursion I used and the stack. The implementation is simple this way and we can revisit if we run into problems.

predicate/Cargo.toml
```diff
@@ -9,6 +9,7 @@ chrono = { version = "0.4", default-features = false }
 data_types = { path = "../data_types" }
 datafusion = { path = "../datafusion" }
 datafusion_util = { path = "../datafusion_util" }
+generated_types = { path = "../generated_types" }
```
tustvold, line 12: Generally I've been trying to avoid the query engine, etc... from depending on the tonic, gRPC stack. It's not a massive deal, but I don't see a particular reason this logic needs to live in `predicate`

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+/// Expects expressions like the following:
+///
+///   * server = 'host' AND "temp"::field > 22
+///   * "env"::tag = 'eu' OR "env"::tag = 'us' OR "env"::tag = 'asia'
+///   * "host" = 'a' AND ("temp"::field > 100.3 OR "cpu"::field = 'cpu-1')
+///   * "_measurement" = 'cpu'
+///
+/// Notes:
+///
+///   * Tag keys can be optionally surrounded in double quotes.
```
tustvold, line 59: Is there a particular motivation to mix single and double quotes?

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+                RPCValue::TagRefValue(make_tag_name(c.value.clone())),
+            )
+        }
+        Expr::Value(v) => match v {
+            Value::Boolean(b) => make_lit(RPCValue::BoolValue(*b)),
+            Value::Number(n, _) => make_lit(parse_number(n)?),
+            Value::DoubleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::SingleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::HexStringLiteral(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::NationalStringLiteral(v) => make_lit(RPCValue::StringValue(v.clone())),
```
tustvold, line 145: TIL, SQL has special syntax for unicode strings. Guess this dates from before UTF-8

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+                RPCType::TagRef,
+                RPCValue::TagRefValue(make_tag_name(c.value.clone())),
+            )
+        }
+        Expr::Value(v) => match v {
+            Value::Boolean(b) => make_lit(RPCValue::BoolValue(*b)),
+            Value::Number(n, _) => make_lit(parse_number(n)?),
+            Value::DoubleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::SingleQuotedString(v) => make_lit(RPCValue::StringValue(v.clone())),
+            Value::HexStringLiteral(v) => make_lit(RPCValue::StringValue(v.clone())),
```
tustvold, line 144: Is a hex string literal a string?

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+            build_binary_node(build_node(left)?, op.clone(), build_node(right)?)
+        }
+        _ => UnexpectedExprTypeSnafu {
+            expr: expr.to_owned(),
+        }
+        .fail(),
+    }
+}
+
+fn parse_number(number: &str) -> Result<RPCValue, Error> {
```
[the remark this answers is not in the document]
tustvold: I'm not sure this won't parse "1.0" as an integer?

predicate/src/sql.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        for expr_str in exprs {
+            let expr = make_sql_expr(&expr_str);
+            let exp_rpc_node = make_tag_expr(&expr_str);
+            assert_eq!(expr, exp_rpc_node)
+        }
+
+        // Using the double quoted syntax for a tag key.
+        let expr = make_sql_expr(r#""my,stuttering,tag,key" != 'foo'"#);
+        let exp_rpc_node = make_tag_expr("my,stuttering,tag,key != 'foo'");
```
tustvold, line 350: I think this test is a tad circular? `make_tag_expr` ends up using much of the same logic being tested?

tustvold (approved):
I think this is good, I've left some comments about location and testing

predicate/src/sql.rs
e-dard: Tag keys can either be absent of quotes, e.g., `server = 'host'` or cam optionally include quotes `"server" = 'host'`. I'm attempting to be as faithful to the InfluxQL semantics as possible, e.g., to avoid having to edit things we might copy in from the field.

predicate/src/sql.rs
e-dard: honestly I'm not sure - I probably don't even need to include this variant but I figured I'd just try and do something with any strings.

predicate/src/sql.rs
e-dard: yeah I realise that, but it's preferable to having hundreds of lines of structs in my opinion. I'll change `make_tag_expr` to only initialise the structs and not use any of the sql module API under test.

e-dard: Addressed this by only using structs in `make_tag_expr` and `make_field_expr`  test helpers.

predicate/Cargo.toml
alamb: I think it would be very cool to avoid pulling in generated-types into `predicate`

How about making a new crate `influxrpc` or something and adding the to/from logic in there (we can also extract code from `influxdb_iox` crate that translates the `RPCPredicate` to `Expr` there as well

predicate/src/sql.rs
```diff
@@ -0,0 +1,508 @@
...
+            RPCLogical::Or,
+            make_tag_expr("env = asia"),
+        )
+        .unwrap();
+
+        assert_eq!(expr, exp_rpc_node);
+    }
+
+    #[test]
+    fn test_from_sql_expr_nested() {
```
alamb, line 490: this is quite cool

alamb (commented):
Love it

predicate/src/sql.rs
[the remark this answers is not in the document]
e-dard: It will, but that's ok. The read API is supposed to be flexible about numerical data types. For example if you have an integer field `temp` you should be able to query it from InfluxQL/Flux with expressions like `temp > 22.4` or `temp > 22`. And the same for float fields.

PR #3595, opened 2022-02-01
feat: flush interface for write buffer producers

### LINKED ISSUE #3504: feat: allow callers to flush kafka buffers
I spoke to @crepererum to understand the buffering behaviour of the kafka write buffer(s) we use.

My summary is:

* We buffer in the kafka client(s)
* Calling [`store_operation()`](https://github.com/influxdata/influxdb_iox/blob/e953f6493f1a98ee426cd742d23659144870597c/write_buffer/src/core.rs#L29) places the write in a buffer and returns when the buffer is flushed to kafka
* We flush when the buffer has accumulated ~10MB of operations (uncompressed)
* We then compress this buffer to the on-wire size, which also includes a little metadata and push it to kafka

__Use case:__
When a router shuts down, we currently drop the writes in the buffer and the client gets disconnected/an error response.

__Proposal:__
Expose a `flush()` on the [`WriteBufferWriting`](https://github.com/influxdata/influxdb_iox/blob/e953f6493f1a98ee426cd742d23659144870597c/write_buffer/src/core.rs#L18) interface to allow the caller to flush the underlying buffer during shutdown

__Current behaviour:__
Writes are dropped and errors are returned - this will cause a spike of errors for clients during rolling deployments.

__Desired behaviour:__
Stop accepting new writes and flush the underlying buffer.

crepererum:
Closes #3504.


domodwyer (approved):
❤️

Issue #553, opened 2020-12-11
style: decide on project style for coercing &str into String


e-dard:
(This issue seems somewhat limited and pedantic but I think it's an interesting test issue for the wider point of "how do we programatically enforce any coding-style we might want in the project).

There are a few ways you can take a borrowed `&str` or string literal and create an allocated `String`.

```rust
"foo".into()

"foo".to_string()

"foo".to_owned()

String::from("foo")
```

Whilst it was not always the case (please correct me if I'm wrong @carols10cents / @shepmaster) as of today these all basically perform the same in Rust. I think in fact several of them end up calling one of the other alternatives. E.g., `to_string` calls `String::from` calls `to_owned`.

Implementation-wise it doesn't matter too much which we call, but I think from a consistency point of view I think it could be confusing for people starting on Rust or the project. 

Therefore I propose that where possible (which I _think_ is always?) we use `.to_owned` when converting from `str` to `String`. I like `to_owned` because it clearly describes what's happening. 

Assuming we agree to care about this the next steps are:

- [ ] decide on choice of method.
- [ ] update style guide
- [ ] replace instances in project with the blessed method.

My only concern with this is that it's not something that's worth investing cognitive load in (via reviewing) because fundamentally there is no difference in correctness or performance between the approaches, so ideally this is something we would have a clippy/style lint for. I have no idea if you can have custom clippy lints for example.

carols10cents:
There is this clippy lint! https://rust-lang.github.io/rust-clippy/master/index.html#str_to_string

alamb:
I think the consensus is 

“If clippy can do it, then do whatever clippy can enforce” -- so like @carols10cents said

domodwyer:
So I actually started running this lint and got:

![image](https://user-images.githubusercontent.com/9275968/102387588-77bfce00-3fc8-11eb-9eb7-5a627553a86b.png)

🤷

carols10cents:
WHO LINTS THE LINTERS

carols10cents:
It looks like `str_to_string` [was readded in Rust 1.50](https://github.com/rust-lang/rust-clippy/pull/6333) (relevant discussion [in this issue](https://github.com/rust-lang/rust-clippy/issues/5610) if you're interested) so we could turn that on and fix the problems in all the crates (someday, https://github.com/rust-lang/cargo/issues/5034, someday I will get to you)

e-dard:
I started to introduce these lints in #3597

PR #3597, opened 2022-02-01
refactor: introduce to_string clippy lints in read buffer

e-dard:
Following on a little bit from the discussion in https://github.com/influxdata/influxdb_iox/issues/553 I added the following lints to the `read_buffer` crate as a little experiment:

* [str_to_string](https://rust-lang.github.io/rust-clippy/master/#str_to_string)
* [string_string](https://rust-lang.github.io/rust-clippy/master/)

These lints encourage the use of `to_owned` rather than `to_string` for `&str` and `String` types. And I'm all about the consistency...

Unfortunately `clippy --fix` did not pick these up as being automatically fixable, which would make handling the whole code-base a bit arduous. I also tried the imaginatively called ` __CARGO_FIX_YOLO=1  cargo clippy --fix` incantation but that didn't help either.

I'm tempted to just try a massive rename on the code base and deal with incorrect `to_owned` calls manually, but I'm not sure if it will pan out.

### DIFF
--- a/read_buffer/src/chunk.rs
+++ b/read_buffer/src/chunk.rs
@@ -264,7 +264,7 @@ impl Chunk {
                 if !table_meta.has_column(column_name) {
                     return ColumnDoesNotExistSnafu {
                         column_name: column_name.to_string(),
-                        table_name: self.table.name().to_string(),
+                        table_name: self.table.name().to_owned(),
                     }
                     .fail();
                 }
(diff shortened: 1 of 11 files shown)


domodwyer (approved):
`__CARGO_FIX_YOLO=1` ha no way..... 

I admit I did not review every line, but I am also all about that consistency 👌

PR #3599, opened 2022-02-01
feat: expose all relevant configs for rskafka write buffers

crepererum:
So whoever creates the write buffer can now configure rskafka, which makes experiments / performance tuning way easier.


write_buffer/src/rskafka/config.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// Maximum message size in bytes.
+    ///
+    /// extracted from `max_message_size`. Defaults to `None` (rskafka default).
+    pub max_message_size: Option<usize>,
+}
+
+impl TryFrom<&BTreeMap<String, String>> for ClientConfig {
+    type Error = WriteBufferError;
+
+    fn try_from(cfg: &BTreeMap<String, String>) -> Result<Self, Self::Error> {
```
alamb, line 19: 👍

write_buffer/src/rskafka/config.rs
```diff
@@ -0,0 +1,256 @@
...
+        Ok(Self {
+            max_message_size: cfg.get("max_message_size").map(|s| s.parse()).transpose()?,
+        })
+    }
+}
+
+/// Config for topic creation.
+#[derive(Debug, PartialEq, Eq)]
+pub struct CreationConfig {
```
alamb, line 28:
```suggestion
pub struct TopicCreationConfig {
```
?

write_buffer/src/rskafka/config.rs
```diff
@@ -0,0 +1,256 @@
...
+    pub max_batch_size: Option<i32>,
+}
+
+impl TryFrom<&BTreeMap<String, String>> for ConsumerConfig {
+    type Error = WriteBufferError;
+
+    fn try_from(cfg: &BTreeMap<String, String>) -> Result<Self, Self::Error> {
+        Ok(Self {
+            max_wait_ms: cfg
+                .get("consumer_max_wait_ms")
```
alamb, line 90: Do you think it would be easier to work with the code if these fields are consistently called `consumer_max_wait_ms` ? The way this is written the same logical thing seems to also be named `max_wait_ms` (presumably because the `consumer` part of the name is redundant as part of the `ConsumerConfig`) name

alamb (approved):
Looks like a nice improvement to me 👍

write_buffer/src/rskafka/config.rs
crepererum: I think the namespacing here is OK:

On the Rust side the `ConsumerConfig` makes clear that this is something concerning the "consumer".

The generic/opaque `connection_config: BTreeMap<String, String>` (which is where we get this data from) is shared by the `ConnectionConfig`, `ProducerConfig` and `ConsumerConfig` and there I wanted to have some more clarity which of them are affected (no prefix => all connections, producer/consumer prefix => producer/consumer). Not having clear prefixes within the string dict is (IMHO) kinda annoying when working w/ rdkafka for example.

write_buffer/src/rskafka/config.rs
crepererum: done

write_buffer/src/rskafka/config.rs
alamb: > Not having clear prefixes within the string dict is (IMHO) kinda annoying when working w/ rdkafka for example.

Yes I agree the UX is much better with prefixed names

PR #3604, opened 2022-02-02
refactor: move sql parsing -> RPC predicate into own crate

e-dard:
As mentioned by @tustvold on #3594 putting the parsing logic into the `predicate` crate introduced a new dependency on `generated_types`.

This PR moves that code out into a new `parser` crate. Alternative name suggestions welcome...


tustvold (approved):
`influxrpc_parser` might be more descriptive... `parser` is a tad, well... amiguous :laughing:

e-dard:
@tustvold updated in 83b562e

PR #3605, opened 2022-02-02
feat: reduce partition fanout

tustvold:
We are seeing partition fanouts into the thousands in our production workloads, this is in large part due to the `Repartition` physical optimizer being a little bit partition happy. I've filed https://github.com/apache/arrow-datafusion/issues/1731 to track this, but in the meantime I think we should just disable this optimization.

I also took the opportunity to set the target_partitions based on the number of threads in the `DedicatedExecutor` as it otherwise defaults to num_cpus, which in docker-ised context may be larger than what has actually been allocated to the workload.

### DIFF
--- a/query/src/exec/context.rs
+++ b/query/src/exec/context.rs
@@ -11,6 +11,10 @@ use datafusion::{
     execution::context::{ExecutionContextState, QueryPlanner},
     execution::{DiskManager, MemoryManager},
     logical_plan::{LogicalPlan, UserDefinedLogicalNode},
+    physical_optimizer::{
+        aggregate_statistics::AggregateStatistics, coalesce_batches::CoalesceBatches,
+        hash_build_probe_order::HashBuildProbeOrder, merge_exec::AddCoalescePartitionsExec,
+    },
     physical_plan::{
         coalesce_partitions::CoalescePartitionsExec,
         displayable,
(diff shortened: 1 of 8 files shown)


crepererum:
> I also took the opportunity to set the target_partitions based on the number of threads in the DedicatedExecutor as it otherwise defaults to num_cpus, which in docker-ised context may be larger than what has actually been allocated to the workload.

I think `num_cpus` [takes cgroups into accounts](https://github.com/seanmonstar/num_cpus/blob/master/src/linux.rs), so this shouldn't be an issue. The std `available_parallelism` currently only handles logical CPUs, but there is [a PR](https://github.com/rust-lang/rust/pull/92697) to fix that as well.

tustvold:
Good to know, was not aware :+1:. Regardless though, we expose num_threads as a CLI parameter and so we should plumb it through to the planner

crepererum (approved):

e-dard (approved):
LGTM 👍

query/src/exec/context.rs
```diff
@@ -164,12 +168,22 @@ const BATCH_SIZE: usize = 1000;
 
 impl IOxExecutionConfig {
     pub(super) fn new(exec: DedicatedExecutor) -> Self {
+        let physical_optimizers = vec![
```
alamb, line 171: I worry this list will get out of sync with what the rest of DataFusion is doing. Perhaps we can add a pub function in DataFusion to get the current default list of optimizers or something

alamb: I am somewhat worried that if we disable this optimization we'll slow down aggregate queries dramatically. But I suppose we can do it and see

query/src/exec/context.rs
```diff
@@ -164,12 +168,22 @@ const BATCH_SIZE: usize = 1000;
...
+            Arc::new(AggregateStatistics::new()),
+            Arc::new(HashBuildProbeOrder::new()),
+            Arc::new(CoalesceBatches::new()),
+            // Arc::new(Repartition::new()), - https://github.com/apache/arrow-datafusion/issues/1731
+            Arc::new(AddCoalescePartitionsExec::new()),
+        ];
+
         let execution_config = ExecutionConfig::new()
             .with_batch_size(BATCH_SIZE)
+            .with_target_partitions(exec.num_threads())
```
alamb, line 181: What if we just did this?

query/src/exec/context.rs
tustvold: At least currently the chunk count is sufficient that we don't really need additional partitioning... If anything we need the opposite :sweat_smile:

query/src/exec/context.rs
tustvold: Following @crepererum 's comment I don't think this will change anything in our production environments. We have 24 CPUs, and each chunk is getting repartioned into 24

tustvold:
Following some discussion I'm going to fix this upstream

Issue #3598, opened 2022-02-01
Create an InfluxRPC CLI client tool


e-dard:
IOx implements InfluxDB's read API. This API allows IOx to act as a storage back-end for either a Flux or InfluxQL query engine. 

The RPC message format for a read request is [quite complex](https://github.com/influxdata/influxdb_iox/blob/er/refactor/read_buffer_clippy/generated_types/protos/influxdata/platform/storage/storage_common.proto#L0-L1). The predicate messages [even more so](https://github.com/influxdata/influxdb_iox/blob/er/refactor/read_buffer_clippy/generated_types/protos/influxdata/platform/storage/predicate.proto).

It would be useful to be able to issue these requests against IOx, without having to rely on a Flux/InfluxQL query engine to craft the request. Whilst we do have some functionality to replay previously submitted requests using the `system.queries` table that relies on us having the data for those queries to hand (e.g., via downloading a live dataset).

In a previous life we had a `storectl` command that would allow one to express to TSM storage some facets of a read request you wanted to issue and the RPC would then be crafted for you. I proposed to create a similar tool for IOx.

### Expected Usage

Something like:

```
// SELECT * FROM "cpu" WHERE "server" = 'abc' AND time >= x AND time < y
$ influxdb_iox storage read_filter \
  --start <optional_ns_or_time_string> \ 
  --stop <optional_ns_or_time_string> \
  --predicate _measurement=cpu AND server = 'abc'
```

```
// SHOW TAG VALUES FROM "cpu" WITH KEY = 'host'
$ influxdb_iox storage tag_values \
  --start <optional_ns_or_time_string> \ 
  --stop <optional_ns_or_time_string> \
  --tag_key_eq "host"
  --predicate _measurement=cpu
```

We will have sub-commands for the various operations.

### Result Visualisation

TBD - ideally a tabular format.

## To Complete

- [ ] Teach IOx to parse SQL expressions into RPC predicates. #3594 
- [ ] Add scaffolding for a new influxdb_iox sub-command.
- [ ] Implement logic to craft and emit an RPC request to an IOx server using `read_filter`.
- [ ] Add support for read_group.
- [ ] Add support for aggregate_window.
- [ ] Add support for tag keys.
- [ ] Add support for tag values.
- [ ] Add support for measurements.

PR #3606, opened 2022-02-02
feat: add skeleton `storage` CLI command

e-dard:
Towards: #3598

This PR adds a new `influxdb_iox` sub-command: `storage`. This sub-command will be responsible for interpreting provided arguments, crafting InfluxRPC read request messages, executing read requests against IOx, and displaying the results to the user of the CLI.

The contribution in this PR is a skeleton CLI sub-command that can deal with necessary arguments for a `ReadFilter` RPC request. It does not issue an RPC request (future PRs).

## Example Usage

Here are some examples of how the CLI can be used. All storage sub-commands will accepts:

 * `start`: an optional starting timestamp. The CLI can parse either integer values representing nanoseconds since (or before) the epoch, or alternatively timestamps in `RFC3339` format.
 * `stop`: see `start` above. Note `stop` is exclusive.
 * `predicate`: an optional predicate that can be used to filter results. The predicate is parsed into a `Predicate` `RPCNode` with the work done in #3594.

These arguments are accepted as options on the `storage` sub-command because all storage RPC requests accept those arguments.

Not providing anything will query all measurements and all time. **Note**: this is an operator tool and some understanding of the consequences of that should be assumed. Having said that I might consider restricting what happens in this case. For example we could default the time range to the last minute, or require a measurement.

```
$ influxdb_iox storage read_filter
```

Query all tables with a time-range:

```
$ influxdb_iox storage --start 2022-02-02T11:30:00Z --stop 2022-02-02T11:35:00Z read-filter
```

You can also use integers if you prefer:

```
$ influxdb_iox storage --start 1643801400000000000 --stop 1643801700000000000 read-filter
```

Query the `temp` field from the `cpu` table:

```
$ influxdb_iox storage --predicate "_measurement=cpu AND _field=temp" read-filter
```

Currently the CLI command just spits the config back out:

```
Unimplemented: config is Config { command: ReadFilter(ReadFilter), start: -9223372036854775806, stop: 9223372036854775806, predicate: Predicate { root: Some(Node { node_type: LogicalExpression, children: [Node { node_type: ComparisonExpression, children: [Node { node_type: TagRef, children: [], value: Some(TagRefValue([0])) }, Node { node_type: TagRef, children: [], value: Some(TagRefValue([99, 112, 117])) }], value: Some(Comparison(Equal)) }, Node { node_type: ComparisonExpression, children: [Node { node_type: TagRef, children: [], value: Some(TagRefValue([255])) }, Node { node_type: TagRef, children: [], value: Some(TagRefValue([116, 101, 109, 112])) }], value: Some(Comparison(Equal)) }], value: Some(Logical(And)) }) } }`
```

You can see that the built up `RPCNode` includes the special tag key for `_measurement` (`[0]`) and `_field` (`[255]`):

```
[
  Node { node_type: TagRef, children: [],  value: Some(TagRefValue([0])) }, 
  Node { node_type: TagRef, children: [], value: Some(TagRefValue([99, 112, 117])) }
]

...

[
  Node { node_type: TagRef, children: [],  value: Some(TagRefValue([255])) }, 
  Node { node_type: TagRef, children: [], value: Some(TagRefValue([116, 101, 109, 112])) }
]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1781,6 +1781,7 @@ dependencies = [
  "influxdb_iox_client",
  "influxdb_line_protocol",
  "influxdb_storage_client",
+ "influxrpc_parser",
  "ingester",
  "internal_types",
  "iox_catalog",
--- a/influxdb_iox/Cargo.toml
+++ b/influxdb_iox/Cargo.toml
@@ -16,6 +16,7 @@ influxdb_iox_client = { path = "../influxdb_iox_client", features = ["flight", "
 influxdb_line_protocol = { path = "../influxdb_line_protocol" }
 ingester = { path = "../ingester" }
 internal_types = { path = "../internal_types" }
+influxrpc_parser = { path = "../influxrpc_parser"}
 iox_catalog = { path = "../iox_catalog" }
 iox_object_store = { path = "../iox_object_store" }
 job_registry = { path = "../job_registry" }
--- a/influxdb_iox/src/commands/storage.rs
+++ b/influxdb_iox/src/commands/storage.rs
@@ -1,11 +1,16 @@
+use generated_types::Predicate;
+use influxrpc_parser::predicate;
 use time;
 
-use snafu::Snafu;
+use snafu::{ResultExt, Snafu};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
     #[snafu(display("Unable to parse timestamp '{:?}'", t))]
     TimestampParseError { t: String },
+
+    #[snafu(display("Unable to parse predicate: {:?}", source))]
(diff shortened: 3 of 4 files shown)


tustvold (approved):

alamb:
I wonder what we think about calling this command `influxrpc` (for consistency with the rest of the codebase) rather than `storage` and putting it in the `database` subcommand

So then one would invoke it like:

```shell
$ influxdb_iox database influxrpc my_db read-filter  --start 2022-02-02T11:30:00Z --stop 2022-02-02T11:35:00Z 
```

It would also then be more consistent with sql 

```shell
$ influxdb_iox database query my_db "select * from foo"
```

PR #3609, opened 2022-02-02
chore: Use ubuntu mirror in test_perf

mkmik:
The `test_perf` CI target fails because archive.ubuntu.com is having a bad day.
ubuntu has an infrastructure for mirrors, let's use it.

https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/14764/workflows/3b97f8f7-5fca-4b93-9efb-839233ceea27/jobs/113164


tustvold (approved):
I still find it uncomfortable that apt uses http :laughing:

.circleci/config.yml
```diff
@@ -279,6 +279,11 @@ jobs:
       CARGO_NET_GIT_FETCH_WITH_CLI: "true"
       RUST_BACKTRACE: "1"
     steps:
+      - run:
+          name: Setup ubuntu mirror
+          command: |
+            mirror=$(curl -s http://mirrors.ubuntu.com/mirrors.txt | head -n 1)
+            sudo sed -i "s;http://.*ubuntu.com/ubuntu/;$mirror;g" /etc/apt/sources.list
```
jdstrand, line 286: Thanks for the PR. While I understand why you're doing this, I don't care for this since it is simply pulling the first mirror that happens to be in the list. ATM, that is http://repos.forethought.net/ubuntu/. Now, this is mostly ok from a security perspective because apt uses a signed archive and the repos don't have the signing key, so they can't inject malicious files into the mirror. They can be out of date though (though assuming Check-Valid-Until is not changed, it can't be too out of date (eg, and suffer from serious replay attacks). They can also be slow. They can also go down.

I suggest instead of choosing a random mirror, choose a closer, official one. Ie, http://us.archive.ubuntu.com/ubuntu which is up now.

jdstrand (changes requested):

jdstrand:
> I still find it uncomfortable that apt uses http laughing

Apt uses a gpg-signed archive with various checks for to ensure security.

.circleci/config.yml
mkmik: Ok doing in https://github.com/influxdata/influxdb_iox/pull/3612

> They can also be slow [..] also go down

Yeah; my reasoning was that while they could be down at least by retrying one stands some chance of making some progress.

> They can also go down with a higher likelihood than the official mirror.

but if that's true, then probably it's better to manually pick a mirror.

Ideally we should have a centralized place where we can choose the mirror we want to use and have all the CI jobs use that.

jdstrand: I like the centralized place to configure idea since it would allow us to document the fallback strategies in one place, we only have to change in this one place since everything just looks in this one place, which would make us able to fix problems faster. Another option for ultimate resilience is to setup our own mirror/proxy and have everything point to it, but this isn't free.

mkmik: > Another option for ultimate resilience is to setup our own mirror/proxy and have everything point to it, but this isn't free.

setting up our own mirror/proxy is not only not sweat-free and also not risk-free; chances are that it would be less reliable than the official mirror

> I like the centralized place to configure idea since it would allow us to document the fallback strategies in one place, we only have to change in this one place since everything just looks in this one place, which would make us able to fix problems faster

I think the easiest solution would be to serve a static page at some super simple place and curl that page from the CI scripts (and gracefully handle when curl fails to fetch that page)

jdstrand: > setting up our own mirror/proxy is not only not sweat-free and also not risk-free; chances are that it would be less reliable than the official mirror

It depends on how it is setup. Some of the proxy software would not shield the clients for errors from the official archive and could be buggy themselves, but if it is a full mirror (eg I've had success with `debmirror`), then the script to update the mirror wouldn't push the changes to the mirror unless is sync'd correctly in the first place, so it would see the errors in where it was mirroring from (eg, archive.ubuntu.com) but all the CI clients would only see the last successful sync. It could be made reliable, but would not be free of cost or maintenance, for sure.

The deployments team said they plan to ask CircleCI if they could maintain said mirror. That could be nice.

jdstrand: > I think the easiest solution would be to serve a static page at some super simple place and curl that page from the CI scripts (and gracefully handle when curl fails to fetch that page)

Yeah, this is not unlike your original PR but we control the contents of the static page.

PR #3612, opened 2022-02-02
fix: Use recommended ubuntu mirror

mkmik:
Addressing https://github.com/influxdata/influxdb_iox/pull/3609#discussion_r797570030


tustvold (approved):

jdstrand (approved):
Thanks for the update! LGTM. It's probably over-engineering since the official archive are usually up, but I wonder if it makes sense to have a comment:
```
# Use us.archive.ubuntu.com since it is close. If it is down, try archive.ubuntu.com. If it is
# down too, temporarily choose something from http://mirrors.ubuntu.com/mirrors.txt that
# is up to date.
mirror="http://us.archive.ubuntu.com/ubuntu"
```

Not a blocker. Feel free to commit as is.

PR #3614, opened 2022-02-02
feat: increase trace buffer size

tustvold:
We're seeing some dropped spans from some particularly chonky traces, it doesn't hurt to just bump the queue here. I expect this may lead to packets getting dropped elsewhere, but who knows :smile: 

I briefly toyed with #2392 - however, estimating the size of the UDP packets ahead of time is non-trivial and in practice extremely unlikely to matter.


alamb (approved):
:shipit:

PR #3617, opened 2022-02-02
feat: combine non-overlapping chunks without deletes

tustvold:
There was already code to do this, it was just not being used in the case where  `self.no_duplicates` evaluates to `false`.


query_tests/cases/in/duplicates.expected
```diff
@@ -19,9 +19,7 @@
 |               |                 RepartitionExec: partitioning=RoundRobinBatch(4)                                                                            |
 |               |                   IOxReadFilterNode: table_name=h2o, chunks=1 predicate=Predicate                                                           |
 |               |         RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                    |
-|               |           IOxReadFilterNode: table_name=h2o, chunks=1 predicate=Predicate                                                                   |
-|               |         RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                    |
-|               |           IOxReadFilterNode: table_name=h2o, chunks=1 predicate=Predicate                                                                   |
+|               |           IOxReadFilterNode: table_name=h2o, chunks=2 predicate=Predicate                                                                   |
```
alamb, line 22: FWIW there is a test that verifies the correct output of this query too in https://github.com/influxdata/influxdb_iox/blob/combine-non-overlapping-chunks/query_tests/src/sql.rs#L690

alamb (approved):
I reviewed the logic and the test changes and I think they look good to me, 

FYI @NGA-TRAN 

Backstory on the rationale for this PR for anyone else following along is that @tustvold has been observing executions on our mirrored production system that have many more DataFusion partitions than we expected.

NGA-TRAN:
Sounds like this is an optimization to remove unnecessary work 👍

PR #3618, opened 2022-02-02
feat: Allow sql test runner to compare sorted output

alamb:
# Rationale:
I desperately want to move more of `query_tests` to be data driven rather than code driven (to reduce compile times as well as debugging cycle times)

Last time I tried to move code from `sql.rs` into its own file, I found i really needed to call `assert_batches_sorted_eq!` rather than `assert_batches_eq` so that the results didn't diff based on sort order

# Changes (in separate commits):
1. Introduce the `--IOX_COMPARE: sorted` annotation for sql tests
2. Port over some system table.sql to demonstrate how it works


query_tests/src/sql.rs
```diff
@@ -123,177 +123,6 @@ async fn sql_select_from_school() {
     .await;
 }
 
-#[tokio::test]
```
alamb, line 126: moved to data files!

query_tests/src/runner/parse.rs
```diff
@@ -1,20 +1,84 @@
...
+        self.query.sql.is_empty()
+    }
+
+    /// Creates a Query and resets this builder to default
+    fn build_and_reset(&mut self) -> Option<Query> {
+        if !self.is_empty() {
+            Some(std::mem::take(&mut self.query))
+        } else {
+            None
+        }
```
e-dard, line 70: alternatively:

```suggestion
        (!self.is_empty()).then(|| std::mem::take(&mut self.query))
```

tustvold (approved):

query_tests/src/runner.rs
```diff
@@ -363,70 +379,189 @@ SELECT * from disk;
...
+        assert_eq!(results.output_contents, expected);
+        results.runner_result.unwrap();
+    }
+
+    /// Result of running the test_input with an expected output
+    struct RunResult {
+        /// Result of running the test case
+        runner_result: Result<()>,
+
+        /// the path that expected file was located in
```
e-dard, line 529:
```suggestion
        /// The path that expected file was located in
```

query_tests/src/runner.rs
```diff
@@ -363,70 +379,189 @@ SELECT * from disk;
...
+        expected_path: PathBuf,
+
+        /// The output file that the runner actually produced
+        output_contents: String,
+
+        /// the path that the output file was written to
```
e-dard, line 535:
```suggestion
        /// The path that the output file was written to
```

e-dard (approved):
LGTM 👍  Like the approach.

PR #3619, opened 2022-02-02
test: add end to end for read_filter and empty string predicates

alamb:
~Draft until https://github.com/influxdata/influxdb_iox/pull/3557 is merged~

@tustvold mentioned https://github.com/influxdata/influxdb_iox/pull/3588#pullrequestreview-870835156 that it would be nice to have an explicit tests for the RPC expr rewrite.

This PR adds such a test as well as reducing some duplication (thus breaking it into its own PR)


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1480,25 +1480,18 @@ mod tests {
 
         // also ensure the plumbing is hooked correctly and that the predicate made it
         // down to the chunk
-        let actual_predicates = fixture
```
alamb, line 1483: I refactored the predicate check into a helper method to avoid some repetition (as I was about to copy/paste this stuff yet again)

e-dard (approved):
LGTM

PR #3621, opened 2022-02-02
fix: Correct a 'long' argument name so the ingester command can run

carols10cents:
I can't currently do `cargo run -- run ingester --help` because it errors with:

```
thread 'main' panicked at 'App ingester: Long option names must be unique for each argument, but '--write-buffer-partition-range-start' is in use by both 'write-buffer-partition-range-start' and 'write-buffer-partition-range-end'', /Users/carolnichols/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.0.13/src/build/app/debug_asserts.rs:86:17
```

This was introduced in https://github.com/influxdata/influxdb_iox/pull/3569 and wasn't caught because there aren't currently any tests that call the ingester command. I'm working on adding some now! :)


alamb (approved):

crepererum (commented):
:facepalm:

PR #3622, opened 2022-02-02
fix: remove incorrect comment

alamb:
I found an out of date comment while reviewing  https://github.com/influxdata/influxdb_iox/pull/3617


query_tests/src/sql.rs
```diff
@@ -517,7 +517,6 @@ async fn sql_predicate_pushdown_correctness_14() {
 
 #[tokio::test]
 async fn sql_deduplicate_1() {
-    // This current expected is wrong because deduplicate is not available yet
```
alamb, line 520: I verified the output below, and there are no duplicated rows (aka every combination of (`time`, `state`, `city`) is unique that I can see)

NGA-TRAN: Agree. The comment was obsolete and I think we updated the expect but  forgot to remove the comment after the deduplicate is available

NGA-TRAN (approved):

query_tests/src/sql.rs
alamb: That is what I thought (hoped) too but wanted to triple check 😅

PR #3624, opened 2022-02-03
test: ensure that rskafka and rdkakfa work together

crepererum:
Before removing rdkafka ensure that:

- rskafka can consume existing messages produced by rdkafka so we do not
  need to drain existing topics
- rdkafka can consume new messages produced by rskafka so we can roll
  back

I ran the whole `write_buffer` test suite (including the newly added
tests) using Apache Kafka as well as Redpanda.


write_buffer/src/lib.rs
```diff
@@ -19,3 +19,6 @@ pub mod kafka;
 pub mod mock;
 
 pub mod rskafka;
+
+#[cfg(all(test, feature = "kafka"))]
```
alamb, line 23: is the idea that this code will get removed when rdkafka is removed as well, and that you are putting it in to show the migration will go smoothly?

write_buffer/src/rskafka/mod.rs
```diff
@@ -198,6 +198,9 @@ impl WriteBufferReading for RSKafkaConsumer {
             let stream = stream.map(move |res| {
                 let (record, _watermark) = res?;
 
+                // store new offset already so we don't get stuck on invalid records
```
alamb, line 201: 👍  good find. Testing FTW!

write_buffer/src/rskafka_kafka_test.rs
```diff
@@ -0,0 +1,229 @@
...
+    rskafka::{RSKafkaConsumer, RSKafkaProducer},
+};
+
+use async_trait::async_trait;
+use data_types::{server_id::ServerId, write_buffer::WriteBufferCreationConfig};
+use test_helpers::maybe_start_logging;
+use time::TimeProvider;
+use trace::RingBufferTraceCollector;
+
+struct RdRsTestAdapter {
```
alamb, line 26:
```suggestion
/// A struct that uses `rdkafka` to write messages to kafka
/// and `rskafka` to read messages back to ensure
/// messages written during the transition period from 
/// `rdkafka` to `rskafka` can be read correctly
struct RdRsTestAdapter {
```

write_buffer/src/rskafka_kafka_test.rs
```diff
@@ -0,0 +1,229 @@
...
+            database_name: random_topic_name(),
+            n_sequencers,
+            time_provider,
+            metric_registry: metric::Registry::new(),
+            trace_collector: Arc::new(RingBufferTraceCollector::new(100)),
+        }
+    }
+}
+
+struct RsRdTestAdapter {
```
alamb, line 56:
```suggestion
/// A struct that uses `rskafka` to write messages to kafka
/// and `rdkafka` to read messages back, for the reason
/// explained in [RdRsTestAdapter]
struct RsRdTestAdapter {
```

alamb (approved):
👍  very cool

write_buffer/src/lib.rs
crepererum: Exactly :+1:

write_buffer/src/rskafka/mod.rs
```diff
@@ -468,6 +469,62 @@ mod tests {
         while jobs.try_next().await.unwrap().is_some() {}
     }
 
+    #[tokio::test]
+    async fn test_offset_after_broken_message() {
```
alamb, line 473: 👍

PR #3625, opened 2022-02-03
chore: remove rdkafka dependency

crepererum:
All features are now covered by rskafka. This also removes the need to
specify a server ID for write buffer consumers. This was only used for
rdkafka since there we needed to specify a consumer group, even though
we did not use any transactions.


mkmik (approved):

domodwyer (approved):
Congratulations - a cool acheivement! 🎉

PR #3626, opened 2022-02-03
refactor: extract NamespaceSchema cache

domodwyer:
This PR moves the namespace schema cache out of the `SchemaValidator`, allowing the cached state to be shared across multiple handlers.

As part of the "namespace auto-creation" behaviour we'll be adding for testing, this cache will allow a quick "do we know this namespace already exists" check in the write path, and if not, we can send a request to the catalog to ensure a record exists for it - this will avoid having to hit the catalog for every write request.

---

* refactor: extract NamespaceSchema cache (3cc44816)

      Breaks the in-memory cache of NamespaceSchema out into a decoupled type that
      can be shared across multiple DML handlers.


router2/src/namespace_cache.rs
```diff
@@ -0,0 +1,23 @@
...
+use data_types::DatabaseName;
+use iox_catalog::interface::NamespaceSchema;
+
+/// An abstract cache of [`NamespaceSchema`].
+pub trait NamespaceCache: Debug + Send + Sync {
+    /// Return the [`NamespaceSchema`] for `namespace`.
+    fn get_schema(&self, namespace: &DatabaseName<'_>) -> Option<Arc<NamespaceSchema>>;
+
+    /// Place `schema` in the cache, unconditionally overwriting any existing
+    /// [`NamespaceSchema`] mapped to `namespace`.
```
alamb, line 17:
```suggestion
    /// [`NamespaceSchema`] mapped to `namespace`, returning
    /// the previous value, if any
```

alamb (approved):
LGTM

router2/src/namespace_cache.rs
domodwyer: Thanks! 

For some reason when my editor wraps the line it sometimes chunks off words making for a thoroughly weird sentance 🤷

PR #3628, opened 2022-02-03
feat(router2): namespace auto-creation

domodwyer:
This PR adds a new handler into the write path which automatically populates the catalog with the requested namespace (if it does not already exist).

To avoid putting catalog I/O in the request path of every write, this PR utilises the namespace cache split out in #3626 to avoid sending a "create namespace" request for namespaces the router already knows exist. In the case of a cache miss, a create request is pessimistically sent to the catalog and response handled appropriately. We can expect this behaviour to cause a just-booted router to exhibit higher latency at the start of a test run until the namespace cache has warmed.

This is a bit of a shim to enable testing the new architecture and will most likely not be in the final prod build - nonetheless it's tested, etc like any other code in case it does wind up living on.

---

* feat(router2): namespace auto-creation (3351c84e)

      Decorate the existing request handler pipeline with a layer that implicitly
      creates the namespace when a write request is received.

### DIFF
--- a/influxdb_iox/src/commands/run/router2.rs
+++ b/influxdb_iox/src/commands/run/router2.rs
@@ -136,7 +136,7 @@ pub async fn command(config: Config) -> Result<()> {
         ns_cache,
         topic_id,
         query_id,
-        "inf".to_owned(),
+        iox_catalog::INFINITE_RETENTION_POLICY.to_owned(),
         handler_stack,
     );
     //
--- a/influxdb_iox/src/commands/run/router2.rs
+++ b/influxdb_iox/src/commands/run/router2.rs
@@ -14,7 +14,6 @@ use crate::{
         },
     },
 };
-use iox_catalog::interface::QueryPoolId;
 use observability_deps::tracing::*;
 use router2::{
     dml_handlers::{NamespaceAutocreation, SchemaValidator, ShardedWriteBuffer},
@@ -68,6 +67,14 @@ pub struct Config {
 
     #[clap(flatten)]
     pub(crate) write_buffer_config: WriteBufferConfig,
+
+    /// Query pool name to dispatch writes to.
+    #[clap(
+        long = "--query-pool",
+        env = "INFLUXDB_IOX_QUERY_POOL_NAME",
+        default_value = "iox-shared"
+    )]
+    pub(crate) query_pool_name: String,
 }
 
 pub async fn command(config: Config) -> Result<()> {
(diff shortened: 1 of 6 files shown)


influxdb_iox/src/commands/run/router2.rs
```diff
@@ -90,7 +93,35 @@ pub async fn command(config: Config) -> Result<()> {
...
+        });
+
+    let handler_stack = NamespaceAutocreation::new(
+        catalog,
+        ns_cache,
+        topic_id,
+        QueryPoolId::new(1),
+        "inf".to_owned(),
+        handler_stack,
+    );
```
domodwyer, line 124: @pauldix this hard codes the namespace that is created to use:

* The correct kafka topic ID from the catalog
* Always sets the query pool to 1
* Always sets the retention policy to "inf"

Will these hard-coded values be sufficient for testing?

router2/src/dml_handlers/mod.rs
```diff
@@ -4,44 +4,65 @@
...
+//!                          ┌──────────────┐
+//!                          │ Write Buffer │
+//!                          └──────────────┘
+//!                                  │
+//!                                  │
+//!                         ┌────────▼─────┐
+//!                         │    Kafka     ├┐
+//!                         └┬─────────────┘├┐
+//!                          └┬─────────────┘│
+//!                           └──────────────┘
```
domodwyer, line 44: This updated router2 request flow diagram shows how this `Namespace Autocreation` addition works with the rest of the router.

influxdb_iox/src/commands/run/router2.rs
pauldix: Hmmm, the `router` shouldn't be getting a kafka topic from the startup config. It can potentially write to many kafka topics, depending on what's in the DB. It would be better to use `catalog.kafka_topics().get_by_name(SHARED_KAFKA_TOPIC)` and for the query pool `catalog.query_pools().get_by_name(SHARED_QUERY_POOL)`. Assuming that we're still using the database to keep this configuration.

That config will ultimately be created in k8s-idpe, but the running servers need to get it from somewhere.

"inf" is fine for the retention policy, but might be good to have that as a constant in the catalog interface.

domodwyer: Added code to resolve a query pool at startup and made it clear this is for testing purposes and will be removed in the future 👍

influxdb_iox/src/commands/run/router2.rs
```diff
@@ -90,7 +98,56 @@ pub async fn command(config: Config) -> Result<()> {
...
+        .map(|v| v.id)
+        .unwrap_or_else(|| {
+            panic!(
+                "no kafka topic named {} in catalog",
+                &config.write_buffer_config.topic
+            )
+        });
+    let query_id = catalog
+        .query_pools()
+        .create_or_get(&config.query_pool_name)
```
NGA-TRAN, line 132: So the router creates a query pool in the catalog if not available?

NGA-TRAN (approved):
LGTM. Just wonder if the Query pool is created when not available in all cases or just for the test case of creating namespace

influxdb_iox/src/commands/run/router2.rs
domodwyer: That's correct - it means we don't have to externally create the record in postgres before we can start the router (for testing purposes - we'll get rid of this when we're done)

PR #3632, opened 2022-02-03
fix: use standard DSN argument for router2 CLI

crepererum:
- support long-form (instead of relying on positional arguments)
- use same code as everying else


domodwyer (approved):
Ah thanks, I did look but didn't find that 👍

lukebond (approved):

Issue #3564, opened 2022-01-27
Update persistence info to the catalog


NGA-TRAN:
# Rationale
This is the third step of #3491 to update the catalog after a parquet file is created by the Ingester. We have to insert data into these table and remove the corresponding PersistingBatch in one transaction:
1. One row in `Parquet_File` table representing the newly created parquet file. Information in the IoxMetata will be used to insert this row.
2. N rows to `Processed_Tombstone` table in which N is the number of tombstones applied during persistence that creates the parquet file.

This means:
  - Insert into Parquet_File
  - Insert n-rows into Process_Tombstone
  - Delete PerstingBatch from the memory
  - Commit the Postgres inserts

pauldix:
There's no need to hold the Postgres transaction open while removing the stuff from memory. I'd have a single function on the `Catalog` trait that is something like:

```
  asnc fn add_parquet_file_with_tombstones(&self, meta: &ParquetMeta, tombstones: &[Tombstone]);
```

Call that function which does the inserts within a Postgres transaction, but commits it all before returning, then remove the persisting batch from memory. 

The query API lets the query server specify a sequence_number to have all data greater than it. If the query server happens to pick up the new parquet file from the catalog and sends the request before the persisting batch has been removed from memory, it won't matter because the persisting batch will be excluded because of the sequence number.

NGA-TRAN:
Thanks @pauldix . That makes it even simpler. Mainly 2 steps:
1. Update Postgres's parquet_file and processed_tombstone in one transaction using available function.
2. Remove the PersistingBatch

PR #3636, opened 2022-02-03
feat: insert the persisted info into the catalog in one transaction

NGA-TRAN:
a task of #3564 


NGA-TRAN:
I have turn this PR to draft and will modify it a bit. Do not review it yet

iox_catalog/src/interface.rs
```diff
@@ -443,6 +470,8 @@ pub trait ParquetFileRepo: Send + Sync {
     #[allow(clippy::too_many_arguments)]
     async fn create(
         &self,
+        // this transaction is only provided when this record is inserted in a transaction
+        txt: Option<&mut Transaction<'_, Postgres>>,
```
crepererum, line 474: Yeah cannot stick a postgres-specific transaction to the generic interface. A file-based or in-mem catalog doesn't have postgres. And even if that would be a generic transaction I think the interface here is the wrong way around. The repositories should be bound to a transaction or UoW (Unit of Work) so it's clear which scope all getters and setters have. I can draft a design for this on Monday if you wish me to do that.

domodwyer: I think sinking a little bit of time into getting the API right for txns up front is a great idea 👍

iox_catalog/src/mem.rs
```diff
@@ -90,6 +102,50 @@ impl Catalog for MemCatalog {
...
+        if let Err(error) = processed_tombstones {
+            // failed to insert processed tombstone, remove the above
+            // inserted parquet file from the catalog
+            self.remove_parquet_file(parquet.object_store_id);
+            return Err(error);
+        }
+        let processed_tombstones = processed_tombstones.unwrap();
+
+        Ok((parquet, processed_tombstones))
+    }
```
domodwyer, line 148: I think it would be nice to move this logic _outside_ of the catalog as it only calls public catalog methods and doesn't need any internal postgres state - it uses the catalog, it is not implementing the catalog (aside from txns).

I guess @crepererum already has similar plans with his txn proposal - it would be nice to to have to duplicate this logic for different catalog implementations (the in-mem catalog implementation is almost identical) when they will do the same thing via the public catalog API.

iox_catalog/src/postgres.rs
```diff
@@ -558,6 +627,99 @@ RETURNING *
...
+    async fn exist(&self, id: ParquetFileId) -> Result<bool> {
+        let read_result =
+            sqlx::query_as::<_, ParquetFile>(r#"SELECT count(*) FROM parquet_file WHERE id = $1;"#)
+                .bind(&id) // $1
+                .fetch_all(&self.pool)
+                .await
+                .map_err(|e| Error::SqlxError { source: e })?;
+
+        Ok(!read_result.is_empty())
+    }
```
domodwyer, line 640: I think this logic is wrong - the result will always have 1 row with `count = 0` so it is not empty:

```sql
iox=# SELECT count(*) FROM iox_catalog.parquet_file WHERE id =42;
 count
-------
     0
(1 row)
```

I would imagine we need to access the count value in the result and compare it to 0 instead of checking if there are no rows.

We will have to change the query from a `ParquetFile` we're specifying in the `query_as::<_, ParquetFile>` generics as it is not returning the columns that go into a `ParquetFile` - I can help with this if you want 👍

iox_catalog/src/postgres.rs
```diff
@@ -558,6 +627,99 @@ RETURNING *
...
+
+    async fn count(&self) -> Result<usize> {
+        let read_result =
+            sqlx::query_as::<_, ProcessedTombstone>(r#"SELECT count(*) FROM parquet_file;"#)
+                .fetch_all(&self.pool)
+                .await
+                .map_err(|e| Error::SqlxError { source: e })?;
+
+        Ok(read_result.len())
+    }
```
domodwyer, line 650: Same as `exist()` I think this logic is probably wrong.

domodwyer (changes requested):
I think we're missing test coverage for the postgres catalog functions `exist` and `count` and I think they need refactoring to give the correct result we're after - I'm happy to help with this if you want 👍

iox_catalog/src/postgres.rs
NGA-TRAN: Good catch about count(*) always return value @domodwyer.
Let me add tests and see if I figure out how to change this. Otherwise, I will ping you for help

NGA-TRAN:
@domodwyer 
> I think we're missing test coverage for the postgres catalog functions exist and count and I think they need refactoring to give the correct result we're after - I'm happy to help with this if you want 

Yeah, I put this as draft to add tests for Postgres functions. Let me do that and I think I might need your help to get column and read value if I cannot figure out from sql documents

NGA-TRAN:
@crepererum and @domodwyer : this is ready for review. If there are no major issues, I think we may want to have this merged to have all behaviors available for testing persistence. When Marco has the new abstract available, we will refactor it.

iox_catalog/src/interface.rs
```diff
@@ -1339,4 +1434,198 @@ pub(crate) mod test_helpers {
...
+        catalog
+            .add_parquet_file_with_tombstones(&parquet, &[t3.clone(), t1.clone()])
+            .await
+            .unwrap_err();
+        // Since the transaction is rollback, t3 is not yet added
+        assert!(!catalog
+            .processed_tombstones()
+            .exist(parquet_file.id, t3.id)
+            .await
+            .unwrap());
```
domodwyer, line 1591: ❤️

domodwyer (approved):
Looking good! 👍

PR #3641, opened 2022-02-04
fix: partition range in ingester CLI has INCLUSIVE end

crepererum:


carols10cents (approved):
Oooh yep, I just ran into this too.

Issue #3445, opened 2022-01-12
Determine Correct time-range semantics in IOx


e-dard:
As we get further into the InfluxQL compatibility test suite there appears to be some subtleties coming to light that indicate that the internal semantics of what a time range domain is might be incorrect.

### What We Know

 - We know that Flux has a [range](https://docs.influxdata.com/flux/v0.x/stdlib/universe/range/) function, that uses the domain `[start, stop)`.
 - We know that InfluxQL lets the user specify different operators for time, effectively making the domain up to the user. They could specify `[start, stop]`, `(start, stop)` etc
 - We know that in IOx we currently use the semantics `[start, stop)`.
 - We know that we are seeing failing issues around timestamp handling https://github.com/influxdata/influxdb_iox/issues/3439
 - Apparently internally TSM storage uses the domain `[start, stop]`, which differs from IOx.

### What We Don't Know

 - How InfluxQL/Flux manipulate timestamps in the `TimeRange` protobuf message, when sending RPC read requests.

### Plan

 - Figure out exactly what values get sent in the `TimeRange` for InfluxQL/Flux requests, which will inform IOx's semantics.
 - Fix IOx's semantics.

e-dard:
## InfluxQL RPC

### No time-range (all time)

When I execute the following InfluxQL query:

```
SELECT * FROM cpu
```

the semantics of the query are to return all data (nothing should be bounded by time). The `Timerange` RPC message sent to storage/IOx is as follows:

```
read filter 
  db_name=54425547585a6420_000000000153158f 
  req.range=Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }) 
  predicate=((TagRef:_m[0x00] == "cpu") AND node: NONE)
```

So that clearly shows that InfluxQL sends the minimum and maximum timestamps supported in the InfluxDB data model. Note that these values are not the minimum and maximum values for the `i64` data type. This is because we save a few nanoseconds to optionally use as sentinel values.

### Explicit time-range

Let's try adding a time range using the minimum and maximum values, like:

```
SELECT * FROM cpu WHERE time OP -9223372036854775806 AND time OP 9223372036854775806
```

Where `OP` is one of the possible operators (`<`, `<=`, `>`, `>=`). We can make a table to show how changing the operator changes the RPC value sent to storage.

| Operator  | RPC start time  | RPC stop time  | 
|---|---|---|
| `<`  | n/a  | 9223372036854775806  |
[...]

PR #3642, opened 2022-02-04
feat: special case max timestamp range for table_names and field_columns

tustvold:
Similar in principle to https://github.com/influxdata/influxdb_iox/pull/3485 this special cases max timestamp ranges for table_names and field_columns.

This "fixes" the behaviour of `SHOW MEASUREMENTS` and `SHOW FIELD KEYS` with respect to a row with the maximum timestamp of 9223372036854775806.

This is because despite neither of these queries specifying a time predicate, the RPC request has one that exludes this row.

```
2022-02-04T13:58:03.867065Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: measurement_names db_name=54425547585a6420_000000000153158f range=Some(TimestampRange { start: -9223372036854775806, end: 9223372036854775806 }) predicate=<NONE>
```

Interestingly the actual measurement_fields request has the "correct" time range, but this seems to follow a measurement_names that has the problematic predicate

```
2022-02-04T13:58:03.875796Z  INFO influxdb_iox::influxdb_ioxd::server_type::database::rpc::storage::service: measurement_fields db_name=54425547585a6420_000000000153158f range=Some(TimestampRange { start: -9223372036854775808, end: 9223372036854775807 }) predicate=<NONE>
```

It is unclear if this is actually an upstream bug, so creating as a draft and we can merge this depending on what is decided.

Relates to #3445


query/src/frontend/influxrpc.rs
```diff
@@ -219,6 +219,9 @@ impl InfluxRpcPlanner {
     {
         debug!(?rpc_predicate, "planning table_names");
 
+        // Special case predicates that span the entire valid timestamp range
+        let rpc_predicate = rpc_predicate.clear_timestamp_if_max_range();
```
alamb, line 223: I wonder if we should just always apply this special case logic? But perhaps then we would get discrepancies the other way if `storage` didn't implement it for all entry points? 🤷

alamb (approved):

query/src/frontend/influxrpc.rs
tustvold: I think the inconsistency arises based on whether storage consults the index or the data itself. Tbh I personally would prefer IOx was consistent even if that introduced discrepancies w.r.t storage. In such a case I'd vote for removing this special case altogether...

alamb: 🔜  (after IOx is running the workload). I think we need to hold our noses for the time being in service of TSM compatiblity

alamb:
Shall we merge this one @tustvold ?

tustvold:
I've re-run this branch against the InfluxRPC test suite and it looks good - so I think let's go for it

PR #3643, opened 2022-02-04
test: Add logging to make `db` tests more debuggable

alamb:
While debugging https://github.com/influxdata/influxdb_iox/pull/3633 I found some things that would have made my life easier


db/src/replay.rs
```diff
@@ -981,6 +981,7 @@ mod tests {
 
     #[tokio::test]
     async fn replay_ok_two_partitions_persist_second() {
+        test_helpers::maybe_start_logging();
```
alamb, line 984: this means the tests now respect `RUST_LOG=debug`

carols10cents (approved):

PR #3646, opened 2022-02-04
feat: catalog update topic command can now take multiple topics

lukebond:
This PR changes the `catalog topic update` subcommand to allow you to specify multiple topics. This will be used for the tool that syncs from the CRD list of topics, so that it only has to make one call.

### DIFF
--- a/influxdb_iox/src/commands/catalog/topic.rs
+++ b/influxdb_iox/src/commands/catalog/topic.rs
@@ -1,5 +1,6 @@
 //! This module implements the `catalog topic` CLI subcommand
 
+use futures::{stream::FuturesUnordered, StreamExt};
 use thiserror::Error;
 
 use crate::clap_blocks::catalog_dsn::CatalogDsnConfig;
@@ -30,8 +31,8 @@ struct Update {
     #[clap(flatten)]
     catalog_dsn: CatalogDsnConfig,
 
-    /// The name of the topic
-    db_name: String,
+    /// The name(s) of the topic(s) to update
+    topic_names: Vec<String>,
 }
 
 /// All possible subcommands for topic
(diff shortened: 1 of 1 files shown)


crepererum:
How would the CLI look like when we need additional per-topic parameters? I think we should rather call the CLI once per topic (we don't have that many topics anyways, so the overhead is OK).

influxdb_iox/src/commands/catalog/topic.rs
```diff
@@ -43,10 +44,23 @@ enum Command {
...
+                .map(|n| async move { topics_repo.create_or_get(n).await })
+                .collect::<FuturesUnordered<_>>()
+                .collect::<Vec<_>>()
+                .await
+                .into_iter()
+                .collect::<Result<Vec<_>, _>>()?
+                .iter()
+                .for_each(|t| {
+                    println!("{}", t.id);
+                });
```
domodwyer, line 63: If you import [`TryStream`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#) it handles `Result` nicely - in this example, [`try_for_each`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each) might be helpful!

```suggestion
            update
                .topic_names
                .iter()
                .map(|n| async move { topics_repo.create_or_get(n).await })
                .collect::<FuturesUnordered<_>>()
                .try_for_each(|t| std::future::ready(Ok(println!("{}", t.id))))
                .await?;
```

domodwyer (commented):
Thought this would be a good moment to introduce `TryStream` 🎉

influxdb_iox/src/commands/catalog/topic.rs
lukebond: much nicer :) thanks

lukebond:
> How would the CLI look like when we need additional per-topic parameters? I think we should rather call the CLI once per topic (we don't have that many topics anyways, so the overhead is OK).

i hadn't thought of that, but i figured since the postgres schema has only name in the topic table it'd be okay? but yeah, expanding that table would break the tool that uses this CLI command i'm adding.

crepererum:
> > How would the CLI look like when we need additional per-topic parameters? I think we should rather call the CLI once per topic (we don't have that many topics anyways, so the overhead is OK).
> 
> i hadn't thought of that, but i figured since the postgres schema has only name in the topic table it'd be okay? but yeah, expanding that table would break the tool that uses this CLI command i'm adding.

IMHO we should always be open to adding more columns to the table, esp. during this very early MVP phase. Looking at some of the tables I personally doubt that the design is final.

That said I'm not in favor of this interface, but I also don't veto it since YOU are both implementer and primary consumer. :slightly_smiling_face:

lukebond:
I think you're right, I'll just loop at the other end! :+1:

PR #3647, opened 2022-02-04
feat: emit tabular results for storage read_filter command

e-dard:
Towards #3598

This PR adds a "pretty printer" output format for gRPC frame results emitted by IOx on its gRPC Storage API. You will now see these results using this pretty format by default when using the `influxb_iox storage` commands.

### Rationale

I know that the Arrow crate has some pretty printing functionality for record batches, so the obvious thing in my mind was to convert from gRPC response frames to Arrow record batches and then leverage that printing functionality. Whilst we naturally have the ability to go from record batch -> gRPC frames in IOx (in the influxRPC API for example), we don't have a way to go in the other direction. The bulk of this PR is just the plumbing to do that.

### Output

It looks like this.

Query all measurements all time (probably not a good idea to do this on a big dataset - will need some guard rails probably).

```
⇒  influxdb_iox storage 54425547585a6420_000000000153158f read-filter
_measurement: air_temperature
rows: 12
+--------------+-------+-------------------+------------------------+-------------------------------+
| location     | state | sea_level_degrees | tenk_feet_feet_degrees | time                          |
+--------------+-------+-------------------+------------------------+-------------------------------+
| coyote_creek | CA    | 77.2              |                        | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 77.1              |                        | 1970-01-01 00:00:01.600756160 |
| coyote_creek | CA    |                   | 40.8                   | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    |                   | 41                     | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    | 77.5              |                        | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    | 78                |                        | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    |                   | 41.1                   | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    |                   | 40.9                   | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    | 77.3              |                        | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    | 77.6              |                        | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    |                   | 40                     | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    |                   | 40.9                   | 1970-01-01 00:00:01.600756160 |
+--------------+-------+-------------------+------------------------+-------------------------------+



_measurement: h2o_temperature
rows: 12
+--------------+-------+----------------+-----------------+-------------------------------+
| location     | state | bottom_degrees | surface_degrees | time                          |
+--------------+-------+----------------+-----------------+-------------------------------+
| coyote_creek | CA    | 51.3           |                 | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 50.9           |                 | 1970-01-01 00:00:01.600756160 |
| coyote_creek | CA    |                | 55.1            | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    |                | 50.2            | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    | 40.2           |                 | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    | 40.1           |                 | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    |                | 55.8            | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    |                | 54.7            | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    | 50.4           |                 | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    | 49.2           |                 | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    |                | 65.2            | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    |                | 63.6            | 1970-01-01 00:00:01.600756160 |
+--------------+-------+----------------+-----------------+-------------------------------+
```

Filter on one measurement and field:

```
⇒  influxdb_iox storage --predicate "_measurement = 'h2o_temperature' AND _field = 'bottom_degrees'" 54425547585a6420_000000000153158f read-filter
_measurement: h2o_temperature
rows: 6
+--------------+-------+----------------+-------------------------------+
| location     | state | bottom_degrees | time                          |
+--------------+-------+----------------+-------------------------------+
| coyote_creek | CA    | 51.3           | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 50.9           | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    | 40.2           | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    | 40.1           | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    | 50.4           | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    | 49.2           | 1970-01-01 00:00:01.600756160 |
+--------------+-------+----------------+-------------------------------+
```

Filter on a field value:

```
⇒  influxdb_iox storage --predicate "sea_level_degrees < 78.0" 54425547585a6420_000000000153158f read-filter
_measurement: air_temperature
rows: 10
+--------------+-------+-------------------+------------------------+-------------------------------+
| location     | state | sea_level_degrees | tenk_feet_feet_degrees | time                          |
+--------------+-------+-------------------+------------------------+-------------------------------+
| coyote_creek | CA    | 77.2              |                        | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 77.1              |                        | 1970-01-01 00:00:01.600756160 |
| coyote_creek | CA    |                   | 40.8                   | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    |                   | 41                     | 1970-01-01 00:00:01.600756160 |
| puget_sound  | WA    | 77.5              |                        | 1970-01-01 00:00:01.568756160 |
| puget_sound  | WA    |                   | 41.1                   | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    | 77.3              |                        | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    | 77.6              |                        | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    |                   | 40                     | 1970-01-01 00:00:01.568756160 |
| santa_monica | CA    |                   | 40.9                   | 1970-01-01 00:00:01.600756160 |
+--------------+-------+-------------------+------------------------+-------------------------------+

```

**Note!** - couple of unusual things with the above output... The InfluxRPC API does not support projections (because in the original TSM storage model everything is a series key rather than columnar). So, even though I'm filtering on one field value I'm getting other fields return too (`tenk_feet_feet_degrees`). This is because the two fields share the same row in IOx (because they have the same tagset in the inputted line protocol).

The reason that they're not displayed in the same row in this output is because I'm dealing with denormalised frames not rows, and the effort to organise them into normalised rows doesn't seem worth it right now (the value of this tool is for debugging/performance analysis, not as something "user facing").

If you wanted to just see the field you could filter on it!

```
⇒  influxdb_iox storage --predicate "_field = 'sea_level_degrees' AND sea_level_degrees < 77.5" 54425547585a6420_000000000153158f read-filter
_measurement: air_temperature
rows: 3
+--------------+-------+-------------------+-------------------------------+
| location     | state | sea_level_degrees | time                          |
+--------------+-------+-------------------+-------------------------------+
| coyote_creek | CA    | 77.2              | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 77.1              | 1970-01-01 00:00:01.600756160 |
| santa_monica | CA    | 77.3              | 1970-01-01 00:00:01.568756160 |
+--------------+-------+-------------------+-------------------------------+
```


influxdb_iox/src/commands/storage/request.rs
```diff
@@ -0,0 +1,33 @@
...
+    generated_types::ReadFilterRequest {
+        predicate,
+        read_source: Some(org_bucket),
+        range: Some(TimestampRange { start, end: stop }),
+        key_sort: read_filter_request::KeySort::Unspecified as i32, // IOx doesn't support any other sort
+        tag_key_meta_names: TagKeyMetaNames::Text as i32,
+    }
+}
+
+// TODO Add the following helpers for building requests:
```
alamb, line 23: I like where this is headed 👍

influxdb_iox/src/commands/storage/response.rs
```diff
@@ -0,0 +1,780 @@
...
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("arrow error: {:?}", source))]
+    Arrow { source: arrow::error::ArrowError },
+
+    #[snafu(display("frame type currently unsupported: {:?}", frame))]
+    UnsupportedFrameType { frame: String },
+
+    #[snafu(display("tag keys must be valid UTF-8: {:?}", source))]
```
alamb, line 25: Don't at least some responses have the special binary `0xff` or `0x00` tag keys?

influxdb_iox/src/commands/storage/response.rs
```diff
@@ -0,0 +1,780 @@
...
+    let mut current_table_frame: Option<(IntermediateTable, SeriesFrame)> = None;
+
+    if frames.is_empty() {
+        return Ok(all_tables);
+    }
+
+    for frame in frames {
+        match frame {
+            generated_types::read_response::frame::Data::Group(_) => {
+                return UnsupportedFrameTypeSnafu {
```
alamb, line 70: it will be interesting to think through how to pretty-print groups 🤔 Not for this PR of course


```
⇒  influxdb_iox storage --predicate "_field = 'sea_level_degrees' AND sea_level_degrees < 77.5" 54425547585a6420_000000000153158f read-filter
_measurement: air_temperature
rows: 3
+--------------+-------+-------------------+-------------------------------+
| location     | state | sea_level_degrees | time                          |
+--------------+-------+-------------------+-------------------------------+
>>======================= Group keys = coyote_creek, CA ==================<<
| coyote_creek | CA    | 77.2              | 1970-01-01 00:00:01.568756160 |
| coyote_creek | CA    | 77.1              | 1970-01-01 00:00:01.600756160 |
>>======================= Group keys = santa_monica, CA ==================<<
| santa_monica | CA    | 77.3              | 1970-01-01 00:00:01.568756160 |
+--------------+-------+-------------------+-------------------------------+
```

influxdb_iox/src/commands/storage/response.rs
```diff
@@ -0,0 +1,780 @@
...
+        print_batches(&[rb]).context(ArrowSnafu)?;
+        println!("\n\n");
+    }
+    Ok(())
+}
+
+// This function takes a set of InfluxRPC data frames and converts them into an
+// Arrow record batches, which are suitable for pretty printing.
+fn into_record_batches(frames: &[Data]) -> Result<BTreeMap<String, RecordBatch>> {
+    // Run through all the frames once to build the schema of each table we need
```
alamb, line 56: 👍

influxdb_iox/src/commands/storage/response.rs
```diff
@@ -0,0 +1,780 @@
...
+                    .map(|x| Some(x.to_owned()))
+                    .collect::<Vec<_>>();
+                column.extend_string(&values);
+
+                let time_column = &mut current_table.time_column;
+                time_column.extend_from_slice(&f.timestamps);
+            }
+        };
+
+        // If the current frame contained field values/timestamps then we need
```
alamb, line 183: I don't think there is any requirement that the frames all need to have the same timestamps (like one frame could have timestamps `[1, 2, 10]` and the other could have `[3, 7, 10, 11]` 

I may have missed it, but there doesn't seem to be code to handle this case

influxdb_iox/src/commands/storage/response.rs
```diff
@@ -0,0 +1,780 @@
...
+                "+------+------------+--------+------+---------+-------------------------------+",
+                "| foo  |            | a      | 1.1  |         | 1970-01-01 00:00:00.000000001 |",
+                "| foo  |            | a      | 2.2  |         | 1970-01-01 00:00:00.000000002 |",
+                "| foo  |            | a      | 3.3  |         | 1970-01-01 00:00:00.000000003 |",
+                "| foo  |            | a      | 4.4  |         | 1970-01-01 00:00:00.000000004 |",
+                "| foo  |            | a      | 5.1  |         | 1970-01-01 00:00:00.000000005 |",
+                "| foo  |            | a      | 5.2  |         | 1970-01-01 00:00:00.000000006 |",
+                "| foo  |            | a      | 5.3  |         | 1970-01-01 00:00:00.000000007 |",
+                "| foo  |            | a      | 10.4 |         | 1970-01-01 00:00:00.000000010 |",
+                "| foo  |            | a      |      | 22      | 1970-01-01 00:00:00.000000001 |",
```
alamb, line 758: I think a valuable test case would be two fields with the same tagset / timestamps so that there is a single row with values for more than one field

alamb (approved):
😍 

I internally waffled about if this "tabularized" format would  obscure too much of the response (about what frames are returned and their types, for example), but I have convinced myself that if out we need that level of detail we can always add some other alternate pretty print implementation.

influxdb_iox/src/commands/storage/response.rs
e-dard: Only if you request them - and this client won't request them.

influxdb_iox/src/commands/storage/response.rs
e-dard: oh that's a nice idea.

influxdb_iox/src/commands/storage/response.rs
e-dard: If there are subsequent frames for the same series frame "parent" then the timestamps will be ordered. The only reason we would get more than one timestamp frame would be if the prior ones had more than 1000 values in them (a TSM block size). 

Subsequent timestamp blocks for different series frames can just be emitted because the tag columns would have updated.

influxdb_iox/src/commands/storage/response.rs
e-dard: Ah so now I remember I haven't added helpful comments on the PR yet :-). This test case is supposed to show you that the output is not aiming to move the values "onto the same row" like the way we store them. Recall that these fields don't get emitted in the same frames by the InfluxRPC API (because each different field means a different series frame). So I figured there wasn't a huge value in trying to rebuild the internal IOx data model just to show the operator what the data is. 

If that doesn't make sense I can explain it better on Monday :-)

alamb: If the behavior is intentional that makes sense to me. A short disclaimer on the output that it isn't the same as the way IOx stores the rows  might help avoid confusion in the future.

PR #3650, opened 2022-02-05
feat: add method to catalog to get its partition info by id

pauldix:
This is a little bit specific for how things are structured in IngesterData right now. Easy enough to take back out later if/when we restructure.

This is in advance of wiring up persistence in the ingester.


crepererum (approved):

domodwyer (approved):

NGA-TRAN (approved):
LGTM

PR #3651, opened 2022-02-07
feat: expose write buffer connection config via CLI

crepererum:
This allows us to configure aspects like message sizes and latency.


domodwyer (approved):
Nice idea 👍

PR #3652, opened 2022-02-07
fix: use a single metric registry for ingester

crepererum:
With this change write buffer ingestion metrics are showing up under
`/metrics`


domodwyer (approved):
🚢

PR #3654, opened 2022-02-07
feat: lazy query log

tustvold:
This makes formatting within the query log lazy, allowing us to buffer up more values.

I suspect this may just push the memory issues to query time, as we don't support lazy materialization of system tables, but hey ho


alamb (approved):

query/src/lib.rs
```diff
@@ -106,6 +106,8 @@ impl<'a> Drop for QueryCompletedToken<'a> {
     }
 }
 
+pub type QueryText = Box<dyn std::fmt::Display + Send + Sync>;
```
alamb, line 109: Adding a note here about the rationale for this (so that we can postpone the stringification until we need it to keep memory overhead lower) I think would help future readers understand the construction

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -1303,31 +1303,29 @@ where
...
+            // This buffering is unfortunate but `Formatter` doesn't implement `std::io::Write`
+            match serde_json::to_string_pretty(&self.s) {
+                Ok(s) => f.write_str(&s),
+                Err(e) => write!(f, "error formatting: {}", e),
             }
         }
     }
 
-    DeferredToJson { s }
+    Box::new(DeferredToJson { s: s.clone() })
```
alamb, line 1328: 🤔  I do wonder how much overhead the raw RPC request (rather than a jsonified version) has compared to a json stringified version. 8 byte pointers add up pretty fast.

Maybe we could save it as "non prettified json" and just reformat it prior to system table construction if the overhead is too much

tustvold: I suppose we could also store the encoded protobuf representation :thinking:

alamb:
Changed title to try and get semantic PR to become unstuck

PR #3656, opened 2022-02-07
feat: gitops adapter

lukebond:
This is a tool I'm tentatively calling the "gitops adapter", as it is the link between our k8s-idpe manifests and the iox catalog config. At present it only deals with kafka topics but its remit may expand.

A new CRD was created in k8s-idpe here: https://github.com/influxdata/k8s-idpe/pull/5969 - it holds a list of kafka topic names.

This PR is a tool that will run in k8s, watching KafkaTopicList resources and calls out to the iox catalog via the CLI to update the topic names.

This tool started out as a copy-paste of the conductor controller and works in the same way. So while it looks like a lot of code it's an existing pattern that works.

The bulk of the unique content is the behaviour of the `reconcile` function in `src/main.rs` and the `reconcile_topics` function it calls out to.


gitops_adapter/src/main.rs
```diff
@@ -0,0 +1,538 @@
...
+                        trace!(
+                            "Updated catalog with kafka topic {}. stdout: {}",
+                            topic,
+                            String::from_utf8_lossy(&output.stdout).to_string().trim()
+                        );
+                        // The CLI returns an ID on success; try to parse it here to ensure it
+                        // worked; not sure that return zero is enough? e.g. --help will return 0.
+                        // also, we'd like to print the IDs out later
+                        String::from_utf8_lossy(&output.stdout)
+                            .to_string()
```
mkmik, line 193: `.to_string()` doesn't seem to be necessary

mkmik (approved):
minor comment

gitops_adapter/src/main.rs
lukebond: thanks

PR #3660, opened 2022-02-07
refactor: catalog transaction

crepererum:
Setup an inteface to handle Units of Work within our catalog. Previously
both the Postgres and the in-mem backend used "mini-transactions on
demand". Now the caller has a clear way to establish boundaries and
gets read and write isolation. A single `Arc<dyn Catalog>` can create as
many `Box<dyn UnitOfWork>` as you like, but note that depending on the
backend you may not scale infinitely (postgres will likely impose
certain limits and the in-mem backend limits concurrency to 1 to keep
things simple).

**Note that while Postgres poisons transactions on the first failure, the in-mem backend is currently very generous. I'll adjust the in-mem backend to be more aggressive in a follow-up PR to catch bugs earlier.**


iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,102 @@ pub trait Catalog: Send + Sync + Debug {
     /// Setup catalog for usage and apply possible migrations.
     async fn setup(&self) -> Result<(), Error>;
 
+    /// Create a new unit of work.
```
alamb, line 311: I know it sounds all stodgy / old skool but I would like to request calling this object using the term `Transaction` somewhere so that it is easier for someone familiar with Databases to find / understand quickly what it is.

Names I think are preferable  include `CatalogTransaction` (my personal preference), `IoxTransaction`, `IoxCatalogTransaction`, `TransactionUnit`, `Transaction`

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,102 @@ pub trait Catalog: Send + Sync + Debug {
...
+        async fn abort_inplace(&mut self) -> Result<(), Error>;
+    }
+}
+
+/// Unit of work of a [`Catalog`].
+///
+/// A unit of work provides a consistent view on data and stages writes (this normally maps to a database transaction).
+/// Repositories can cheaply be borrowed from it. To finalize a unit of work, call [commit](Self::commit) or [abort](Self::abort).
+///
+/// Note that after any method in any repository in this unit returned an error, the unit MIGHT be poisened and will
```
alamb, line 343:
```suggestion
/// Note that after any method in this unit returned an error, the unit MIGHT be poisoned and will
```

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,102 @@ pub trait Catalog: Send + Sync + Debug {
...
+/// return errors for all operations, depending on the backend.
+#[async_trait]
+pub trait UnitOfWork: Send + Sync + Debug + sealed::UnitOfWorkFinalize {
+    /// Commit unit of work.
+    ///
+    /// If successfull, all changes will be visible to other Unit of Works. Note that a failed commit does NOT mean that
+    /// the unit of not committed (e.g. the failure might be due to an IO error after a database transaction was already
+    /// finished).
```
alamb, line 351: This doesn't sound quite right to me -- a commit should either entirely succeed or entirely fail in a way that the observer can determine. Otherwise it seems to miss a large amount of the potential value of this abstraction

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,102 @@ pub trait Catalog: Send + Sync + Debug {
...
+    async fn commit(mut self: Box<Self>) -> Result<(), Error> {
+        self.commit_inplace().await
+    }
+
+    /// Abort unit of work, throwing away all changes.
+    async fn abort(mut self: Box<Self>) -> Result<(), Error> {
+        self.abort_inplace().await
+    }
+
     /// repo for kafka topics
```
alamb, line 361: I don't quite follow the "repo" terminology here

iox_catalog/src/interface.rs
```diff
@@ -1542,90 +1617,129 @@ pub(crate) mod test_helpers {
...
+
+        let k3 = uow
+            .kafka_topics()
+            .get_by_name("test_uow_isolation")
+            .await
+            .unwrap();
+        assert!(k3.is_none());
+        uow.abort().await.unwrap();
+
+        insertion_task.await.unwrap();
```
alamb, line 1743: is it worth testing that the `"test_uow_isolation"` topic succeeds after `insertion_task` is complete?

alamb (commented):
I feel quite strongly about the naming (specifically that it be called something with `Transaction`)

Otherwise I think this interface looks great and the PR overall is 🏅 . 

Nice work @crepererum

iox_catalog/src/interface.rs
crepererum: I think we mean the same thing, I'll improve the wording.

iox_catalog/src/interface.rs
crepererum: done

iox_catalog/src/interface.rs
crepererum: done

iox_catalog/src/interface.rs
crepererum: Added a doc section to the `Transaction` docs.

iox_catalog/src/interface.rs
crepererum: done

crepererum:
Renamed Unit of Work to Transaction

alamb (approved):
its-so-beautiful-cry

Thank you

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,115 @@ pub trait Catalog: Send + Sync + Debug {
...
+/// are implemented.
+#[async_trait]
+pub trait Transaction: Send + Sync + Debug + sealed::TransactionFinalize {
+    /// Commit transaction.
+    ///
+    /// # Error Handling
+    /// If successfull, all changes will be visible to other transactions.
+    ///
+    /// If an error is returned, the transaction may or or not be committed. This might be due to IO errors after the
+    /// transaction was finished. However in either case, the transaction is atomic and can only succeed or fail
```
alamb, line 363: 👍

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,115 @@ pub trait Catalog: Send + Sync + Debug {
...
+    /// Creating transactions is potentially expensive. Holding one consumes resources. The number of parallel active
+    /// transactions might be limited per catalog, so you MUST NOT rely on the ability to create multiple transactions in
+    /// parallel for correctness but only for scaling.
+    async fn start_transaction(&self) -> Result<Box<dyn Transaction>, Error>;
+}
+
+/// Secret module for [sealed traits].
+///
+/// [sealed traits]: https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
+pub(crate) mod sealed {
```
domodwyer, line 322: Nice 👍

iox_catalog/src/interface.rs
```diff
@@ -309,59 +308,115 @@ pub trait Catalog: Send + Sync + Debug {
...
+///
+///
+/// # Repositories
+/// The methods (e.g. "get or create") for handling entities (e.g. namespaces, tombstones, ...) are grouped into
+/// *repositories* with one *repository* per entity. A repository can be thought of a collection of a single entity.
+/// Getting repositories from the transaction is cheap.
+///
+/// Note that a repository might internally map to a wide range of different storage abstractions, ranging from one or
+/// more SQL tables over key-value key spaces to simple in-memory vectors. The user should and must not care how these
+/// are implemented.
```
domodwyer, line 354: Might be worth defining the drop semantics of a txn - I assume it is something like "dropping a transaction instance terminates the underlying transaction, releasing any state from the backend" that implementors are expected to uphold?

domodwyer (approved):
Love it 👍

iox_catalog/src/interface.rs
crepererum: I'll add a comment

crepererum: done, also added a test.

iox_catalog/src/lib.rs
```diff
@@ -43,7 +43,7 @@ pub mod postgres;
 pub async fn validate_or_insert_schema<'a, T, U>(
     tables: T,
     schema: &NamespaceSchema,
-    catalog: &dyn Catalog,
+    txn: &mut dyn Transaction,
```
pauldix, line 46: Question on this modification. Previously, this function was designed so that in most cases, the catalog would never be used (and thus Postgres wouldn't be touched). I see in the tests this is sending over a transaction, which in the Postgres catalog implementation is actually doing something on the sqlx side. If this transaction isn't used (i.e. because the cached schema contains the info or no new schema is being defined) does this actually touch Postgres? I want to make sure that this isn't introducing a bunch of load because of the API change.

crepererum: Good point. For this specific method might wanna open the transaction on-demand then instead of potentially prematurely.

domodwyer: Yeah it's a great point - I just looked through and I think it always opens a txn - there's no lazy opening in the SQLx impl nor our code.

Great spot 🎉

PR #3661, opened 2022-02-07
feat: lazy system tables

tustvold:
This reworks the SystemTable plumbing to allow them to yield multiple batches, this will:

* Allow more parallelism in queries
* Apply limits more effectively
* Avoid building enormous record batches that exceed size limits of arrays

Currently only QueryLog is hooked up to take advantage of this, which feeds into #3654


db/src/system_tables.rs
```diff
@@ -133,21 +138,20 @@ impl SchemaProvider for SystemSchemaProvider {
...
+
 /// The minimal thing that a system table needs to implement
 trait IoxSystemTable: Send + Sync {
     /// Produce the schema from this system table
     fn schema(&self) -> SchemaRef;
 
-    /// Get the contents of the system table as a single RecordBatch
-    fn batch(&self) -> Result<RecordBatch>;
+    /// Get the contents of the system table
+    fn scan(&self, batch_size: usize) -> Result<BatchIterator>;
```
tustvold, line 149: This is the major change, rather than yielding a single `RecordBatch` implementations now yield a boxed iterator

db/src/system_tables.rs
```diff
@@ -170,56 +174,152 @@ where
...
+                    batch,
+                    Arc::clone(&self.projected_schema),
+                    self.projection.as_ref(),
+                )
+            })
+        }))
+    }
+}
+
+fn project_schema(
```
tustvold, line 271: This is just split apart as we need to project the schema ahead of creating the ExecutionPlan

db/src/system_tables/chunks.rs
```diff
@@ -30,9 +31,14 @@ impl IoxSystemTable for ChunksTable {
         Arc::clone(&self.schema)
     }
 
-    fn batch(&self) -> Result<RecordBatch> {
-        from_chunk_summaries(self.schema(), self.catalog.chunk_summaries())
-            .log_if_error("system.chunks table")
+    fn scan(&self, _batch_size: usize) -> Result<BatchIterator> {
```
tustvold, line 34: Here is an example of a system table that is now lazy but still doesn't support streaming batches

e-dard (approved):
LGTM 👍

db/src/system_tables.rs
```diff
@@ -170,56 +174,152 @@ where
         _filters: &[datafusion::logical_plan::Expr],
         _limit: Option<usize>,
```
alamb, line 175: there is a limit passed in here that could potentially be applied as well

db/src/system_tables.rs
```diff
@@ -170,56 +174,152 @@ where
...
+            projection: self.projection.clone(),
+        }))
+    }
+
+    fn statistics(&self) -> Statistics {
+        Statistics::default()
+    }
+}
+
+struct SystemTableStream {
```
alamb, line 243: There is a lot of ceremony to create one of these, isn't there 🤔

db/src/system_tables.rs
alamb: I think you might be able to use `Schema::project` here https://docs.rs/arrow/8.0.0/arrow/datatypes/struct.Schema.html#method.project

db/src/system_tables.rs
```diff
@@ -170,56 +174,152 @@ where
...
+        })
+        .collect::<DataFusionResult<Vec<_>>>()?;
+
+    Ok(Arc::new(Schema::new(projected_columns)))
+}
+
 /// Creates a DataFusion ExecutionPlan node that scans a single batch
 /// of records.
-fn scan_batch(
+fn project_batch(
```
alamb, line 299: See also `RecordBatch::project`: https://docs.rs/arrow/8.0.0/arrow/record_batch/struct.RecordBatch.html#method.project

alamb (approved):
👌  Nice work

db/src/system_tables.rs
```diff
@@ -170,134 +173,97 @@ where
...
+        self
+    }
 
-    fn seq_array(start: u64, end: u64) -> ArrayRef {
-        Arc::new(UInt64Array::from_iter_values(start..end))
+    fn schema(&self) -> SchemaRef {
+        Arc::clone(&self.projected_schema)
     }
 
-    #[tokio::test]
```
tustvold, line 228: We now use upstream logic to do batch and schema projection

alamb (approved):
😎  nice

PR #3662, opened 2022-02-07
feat: add timeouts to write chunk

tustvold:
We are seeing lifecycle actions getting stuck in the write chunk to object store lifecycle action.

```
+------+---------+--------------------------------+--------------------+----------------+------------+---------------------+--------------------------------------+-------------------------------------+
| id   | status  | start_time                     | cpu_time_used      | wall_time_used | table_name | partition_key       | chunk_ids                            | description                         |
+------+---------+--------------------------------+--------------------+----------------+------------+---------------------+--------------------------------------+-------------------------------------+
| 4871 | Running | 2022-02-03T12:29:25.643600598Z | 00:00:00.009393752 |                | buckets    | 2022-01-31 21:00:00 | 12da0fc7-fcb6-4fdf-aa2a-5e452664b778 | Persisting chunks to object storage |
+------+---------+--------------------------------+--------------------+----------------+------------+---------------------+--------------------------------------+-------------------------------------+
```

We know from this that:

* The operation is not taking up CPU time
* The operation has not completed
* The future associated with the task has not been dropped

This means in all likelihood this is a case of `await`ing a future that never completes. 

This PR therefore adds timeouts to each of these operations which will:

* Allow the system to recover
* Log to identify where it is getting stuck


db/src/lifecycle/write.rs
```diff
@@ -111,7 +116,9 @@ pub(super) fn write_chunk_to_object_store(
         // catalog-level transaction for preservation layer
         {
             // fetch shared (= read) guard preventing the cleanup job from deleting our files
-            let _guard = db.cleanup_lock.read().await;
+            let _guard = timeout(TIMEOUT, db.cleanup_lock.read())
+                .await
+                .log_if_error("write chunk cleanup lock")?;
```
alamb, line 121: 👍  for error discrimination

alamb (approved):

PR #3663, opened 2022-02-07
feat: add tag_values to storage cli

e-dard:
Towards #3598.

This PR adds support for `tag_values` to the storage CLI. The tag values RPC requires that a column (tag key) is specified. Thus, the command us used as follows:

```
⇒  influxdb_iox storage db_name tag-values host

tag values: 3
+-----------------------+
| values                |
+-----------------------+
| -                     |
| foo                   |
| bar                   |
+-----------------------+
```

If you specify `_measurement` or `_field` the value will be correctly transformed into the byte-specific format used by the storage read API (`\x00` or `\xff` respectively).

```
⇒  influxdb_iox storage db_name tag-values _measurement

tag values: 3
+-----------------------+
| values                |
+-----------------------+
| m1                    |
| m2                    |
| m3                    |
+-----------------------+
```

### DIFF
--- a/influxdb_iox/src/commands/storage.rs
+++ b/influxdb_iox/src/commands/storage.rs
@@ -110,12 +110,11 @@ fn parse_db_name(db_name: &str) -> Result<OrgAndBucket, ParseError> {
 #[derive(Debug, clap::Parser)]
 enum Command {
     /// Issue a read_filter request
-    ReadFilter(ReadFilter),
+    ReadFilter,
 }
 
-/// Create a new database
-#[derive(Debug, clap::Parser)]
-struct ReadFilter {}
+// #[derive(Debug, clap::Parser)]
+// struct ReadFilter {}
 
 /// Create and issue read request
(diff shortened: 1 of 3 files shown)


influxdb_iox/src/commands/storage.rs
```diff
@@ -109,13 +109,14 @@ fn parse_db_name(db_name: &str) -> Result<OrgAndBucket, ParseError> {
...
-    ReadFilter(ReadFilter),
+    ReadFilter,
+    TagValues(TagValues),
 }
 
-/// Create a new database
 #[derive(Debug, clap::Parser)]
-struct ReadFilter {}
+struct TagValues {
+    tag_key: String,
```
alamb, line 118:
```suggestion
    /// The name of the tag for which to retrieve values
    tag_key: String,
```

alamb (approved):
The only thing that is probably good to add at some point are some tests. But we can do that perhaps as a follow on PR

PR #3664, opened 2022-02-08
feat: add WriteBufferErrorKind

tustvold:
We recently observed some clients getting `OffsetOutOfRange` responses from Kafka, whilst there is likely an underlying issue here, the error handling for this caused a veritable log explosion.

This is because rsfkaka correctly bubbles up this error up to the consumer, as it doesn't make sense to retry it, however, the write buffer logic in `db` immediately retries any write buffer error. This is to avoid invalid data payloads from stalling out ingest, but has the unfortunate side-effect of going ballistic in the event of a systematic error.

This PR adds a `WriteBufferErrorKind` to allow distinguishing between errors due to invalid payloads, and errors due to something else


write_buffer/src/core.rs
```diff
@@ -10,7 +12,106 @@ use futures::stream::BoxStream;
...
+impl From<rskafka::client::producer::Error> for WriteBufferError {
+    fn from(e: rskafka::client::producer::Error) -> Self {
+        Self {
+            inner: Box::new(e),
+            kind: WriteBufferErrorKind::IO,
+        }
+    }
+}
+
+impl From<String> for WriteBufferError {
```
tustvold, line 86: I don't like this, but this PR turned into an absolute monster otherwise...

db/src/write_buffer.rs
```diff
@@ -142,14 +142,27 @@ async fn stream_in_sequenced_entries<'a>(
...
-                warn!(
+                error!(
                     %e,
                     %db_name,
                     sequencer_id,
-                    "Error converting write buffer data to SequencedEntry",
+                    "Error reading record from write buffer",
                 );
+
+                match e.kind() {
```
tustvold, line 153: This is what this PR is all about

write_buffer/src/core.rs
```diff
@@ -10,7 +12,106 @@ use futures::stream::BoxStream;
...
+            inner: e.into(),
+            kind: WriteBufferErrorKind::Unknown,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub enum WriteBufferErrorKind {
+    /// This operation failed for an unknown reason
+    Unknown,
```
crepererum, line 107: Eventually we probably wanna get rid of the "unknown" variant.

crepererum (commented):
I like the introduction of structured errors. However what I don't like is where the errors are handled. IMHO they should either be handled in rskafka or in the write buffer crate, not all over the place in the consumer code. The stream the consumer uses should be consumable w/o any backoffs.

crepererum (approved):
Since this is blocking prod, let's merge it. I'll clean it up in a follow-up.

PR #3676, opened 2022-02-08
fix: write buffer traces

crepererum:
Fix a few edge cases in our traces that came up when using this functionality in our test bench. Jaeger is now a bit more happy. Sadly it doesn't really like our links due to https://github.com/jaegertracing/jaeger-ui/issues/662 and it's a bit cumbersome to navigate them.


domodwyer (approved):
I will have to do something similar in the router shortly - there's not a lot of observability going on yet!

PR #3681, opened 2022-02-08
feat: make stop/stop/predicate global

e-dard:
This propagates through options to the child sub-commands. Put another way, before you had to do:

```
$ influxdb_iox storage my_db --start 1641340800000000000 --stop 1643932800000000000 --predicate "_measurement = 'abc'" read-filter 
```

and now you can do:

```
$  influxdb_iox storage my_db read-filter --start 1641340800000000000 --stop 1643932800000000000 --predicate "_measurement = 'abc'"
```


tustvold (approved):
:heart:

PR #3607, opened 2022-02-02
refactor: Consolidate more query_test DbSetups

alamb:
Follow on from https://github.com/influxdata/influxdb_iox/pull/3580

My ultimate plan is to switch from `async_trait` to explicit futures,
but first I want to see the scale of the problem we are dealing with

PR #3682, opened 2022-02-08
refactor: Complete move of test scenarios into library.rs

alamb:
refactor: Complete move of test scenarios into library.rs

Follow on to https://github.com/influxdata/influxdb_iox/pull/3607

My ultimate plan is to switch from `async_trait` to explicit futures,
but first I want to see the scale of the problem we are dealing with


query_tests/src/influxrpc/read_window_aggregate.rs
```diff
@@ -78,36 +72,6 @@ async fn test_read_window_aggregate_no_data_no_pred() {
     .await;
 }
 
-struct MeasurementForWindowAggregate {}
```
alamb, line 81: Moves into a different file, that is all

e-dard (approved):

PR #3686, opened 2022-02-09
fix: increase default `producer_max_batch_size`

crepererum:
Let's try to get prod back online.


tustvold (approved):
Isn't it currently set to something even larger?

crepererum:
> Isn't it currently set to something even larger?

Let's see if this works, otherwise I'll wire it through to conductor because this requires us to also bump the "max message size" for consumers.

Issue #3368, opened 2021-12-13
Abort Database Init State Machine


tustvold:
Currently database shutdown waits for replay to finish or error, this by extension means that server shutdown waits for replay to finish.

The scenario where I observed this, had one database with an incredibly large replay window with replay taking a significant length of time, and another database that panicked during replay. The latter database would trigger server shutdown, but the server would wait until replay of the other database had finished before actually exiting.

It should be possible to trigger shutdown of database's in the process of initializing and have them terminate in a timely manner

crepererum:
It's probably sufficient to call `abort` on this handle here:

https://github.com/influxdata/influxdb_iox/blob/d0f588d3b800894fe0ebd06b6f9a184ca6a603d7/server/src/database.rs#L188

tustvold:
This problem is more general, and can also occur if initialization gets stuck in a given stage. Going to pick this one up as it is preventing recovering production, where initialization is stalling because Kafka no longer contains the offset to resume consumption from.

PR #3689, opened 2022-02-09
refactor: simplify Database retry backoff logic (#3368)

tustvold:
Whilst working on #3368 I realised this retry loop could be simplified


server/src/database/init.rs
```diff
@@ -616,62 +607,15 @@ pub(crate) async fn initialize_database(shared: &DatabaseShared) {
...
 
-    // error throttling
-    // - checks if the current error was already throttled
-    // - keeps a backoff duration that will change over the course of multiple errors
-    let mut throttled_error = false;
+    // A backoff duration for retrying errors that will change over the course of multiple errors
     let mut backoff = INIT_BACKOFF;
 
     while !shared.shutdown.is_cancelled() {
-        // Acquire locks and determine if work to be done
```
tustvold, line 626: I found this block quite hard to follow as it is potentially waiting for the error retry of the previous iteration, by moving it to the end of the while loop I think it is easier to follow

server/src/database/init.rs
```diff
@@ -744,6 +683,37 @@ pub(crate) async fn initialize_database(shared: &DatabaseShared) {
             *state.unfreeze(handle) = next_state;
             shared.state_notify.notify_waiters();
         }
+
```
crepererum, line 686: Just checking: in this case did we already drop all the locks from the DB (including read locks) so gRPC requests are free to modify the state?

crepererum: If the answer is "yes, all locks gone", feel free to merge it

Issue #3629, opened 2022-02-03
Shard delete DML ops with no specified table across all sequencers


domodwyer:
Currently `router2` does not special-case the logic for delete requests that do not contain a table name.

When a delete is received by the router that contains no table name, it should be sent to all sequencers for the topic so that the (potentially unconstrained) delete is applied to all shards.

PR #3690, opened 2022-02-09
feat: split & assign partition key in router

domodwyer:
This PR adds partitioning to the `router2` implementation by introducing a new `Partitioner` into the request pipeline that operates over row-wise line-protocol writes suitable for use with the HTTP API. We're not currently using the gRPC API, which is column-wise and therefore will need a different partitioning implementation when we wish to enable gRPC writes.

This change keeps the namespace auto-creation (#3628) as the first handler that is called, ensuring we don't bother partitioning a request if the namespace cannot be created.

The partitioned writes are tagged with the generated partition key, and passed through the handler stack until reaching the write buffer where the key is logged and dropped - further work will allow us to pass this partition key across the write-buffer API boundary for use in the ingester in the future.

---

* refactor: parametrise DML handler input type (fc409874)

      Allow a DML handler to specify the write input type on which it operates.

      This allows us to construct a write handler pipeline that transforms the 
      request as it passes through the various handlers. We'll use this to implement
      a handler that annotates a normal set of table writes with the partition key,
      modifying downstream handlers to expect this annotated input.

* test: MockDmlHandler generic over write input (575d8f26)

      Allow the MockDmlHandler to capture any input type given to the write() 
      method. This lets us reuse the mock across all handler implementations, 
      regardless of their expected write input type.

* feat: row-wise line protocol partitioner (748a83bf)

      Implements write partitioning by date for line-protocol strings received by
      the HTTP API.

      This change moves the LP parsing logic out of the HTTP handler (nicer!) and
      into the Partitioner where it is partitioned as each line is parsed to
      minimise overhead - it's cheaper to partition row-wise data than trying to
      split out column-wise data later. The existing MutableBatch builder logic
      (LineConverter) is slightly modified for this use to avoid duplication.

* refactor: namespace auto-creator generic input (2eb2ddb0)

      Changes the NamespaceAutocreation handler to be generic over any WriteInput.

      This allows the NamespaceAutocreation layer to be placed anywhere in the 
      handler stack, without needing a prior transformation or specific write type.

* refactor: enable Partitioner in request pipeline (0415fa13)

      Adds the Partitioner DML handler into the handler stack, modifying the input
      types of down-stream handlers to accept the partitioned data.

### DIFF
--- a/router2/src/dml_handlers/mock.rs
+++ b/router2/src/dml_handlers/mock.rs
@@ -71,11 +71,12 @@ macro_rules! record_and_return {
 impl DmlHandler for Arc<MockDmlHandler> {
     type WriteError = DmlError;
     type DeleteError = DmlError;
+    type WriteInput = HashMap<String, MutableBatch>;
 
     async fn write(
         &self,
         namespace: DatabaseName<'static>,
-        batches: HashMap<String, MutableBatch>,
+        batches: Self::WriteInput,
         _span_ctx: Option<SpanContext>,
     ) -> Result<(), Self::WriteError> {
         record_and_return!(
--- a/router2/src/dml_handlers/mock.rs
+++ b/router2/src/dml_handlers/mock.rs
@@ -1,20 +1,18 @@
-use std::{collections::VecDeque, sync::Arc};
+use std::{collections::VecDeque, fmt::Debug, sync::Arc};
 
 use async_trait::async_trait;
 use data_types::{delete_predicate::DeletePredicate, DatabaseName};
 
-use hashbrown::HashMap;
-use mutable_batch::MutableBatch;
 use parking_lot::Mutex;
 use trace::ctx::SpanContext;
 
 use super::{DmlError, DmlHandler};
 
 #[derive(Debug, Clone)]
-pub enum MockDmlHandlerCall {
+pub enum MockDmlHandlerCall<W> {
     Write {
         namespace: String,
-        batches: HashMap<String, MutableBatch>,
+        batches: W,
     },
     Delete {
         namespace: String,
@@ -23,23 +21,42 @@ pub enum MockDmlHandlerCall {
     },
 }
 
-#[derive(Debug, Default)]
-struct Inner {
-    calls: Vec<MockDmlHandlerCall>,
+#[derive(Debug)]
+struct Inner<W> {
+    calls: Vec<MockDmlHandlerCall<W>>,
     write_return: VecDeque<Result<(), DmlError>>,
     delete_return: VecDeque<Result<(), DmlError>>,
 }
 
-impl Inner {
-    fn record_call(&mut self, call: MockDmlHandlerCall) {
+impl<W> Default for Inner<W> {
+    fn default() -> Self {
+        Self {
+            calls: Default::default(),
+            write_return: Default::default(),
+            delete_return: Default::default(),
+        }
+    }
+}
+
+impl<W> Inner<W> {
+    fn record_call(&mut self, call: MockDmlHandlerCall<W>) {
         self.calls.push(call);
     }
 }
 
-#[derive(Debug, Default)]
-pub struct MockDmlHandler(Mutex<Inner>);
+#[derive(Debug)]
+pub struct MockDmlHandler<W>(Mutex<Inner<W>>);
 
-impl MockDmlHandler {
+impl<W> Default for MockDmlHandler<W> {
+    fn default() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<W> MockDmlHandler<W>
+where
+    W: Clone,
+{
     pub fn with_write_return(self, ret: impl Into<VecDeque<Result<(), DmlError>>>) -> Self {
         self.0.lock().write_return = ret.into();
         self
@@ -50,7 +67,7 @@ impl MockDmlHandler {
         self
     }
 
-    pub fn calls(&self) -> Vec<MockDmlHandlerCall> {
+    pub fn calls(&self) -> Vec<MockDmlHandlerCall<W>> {
         self.0.lock().calls.clone()
     }
 }
@@ -68,10 +85,13 @@ macro_rules! record_and_return {
 }
 
 #[async_trait]
-impl DmlHandler for Arc<MockDmlHandler> {
+impl<W> DmlHandler for Arc<MockDmlHandler<W>>
+where
+    W: Debug + Send + Sync,
+{
     type WriteError = DmlError;
     type DeleteError = DmlError;
-    type WriteInput = HashMap<String, MutableBatch>;
+    type WriteInput = W;
 
     async fn write(
         &self,
--- a/router2/src/dml_handlers/mod.rs
+++ b/router2/src/dml_handlers/mod.rs
@@ -4,44 +4,49 @@
 //! processing handler chain:
 //!
 //! ```text
-//!                ┌──────────────┐    ┌──────────────┐
-//!                │   HTTP API   │    │   gRPC API   │
-//!                └──────────────┘    └──────────────┘
-//!                        │                   │
-//!                        └─────────┬─────────┘
-//!                                  │
-//!                                  ▼
-//!                     ╔═ DmlHandler Stack ═════╗
-//!                     ║                        ║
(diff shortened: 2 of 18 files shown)


router2/src/dml_handlers/mod.rs
```diff
@@ -4,44 +4,49 @@
...
+//!                           ┌──────────────┐
+//!                           │ Write Buffer │
 //!                           └──────────────┘
+//!                                   │
+//!                                   │
+//!                          ┌────────▼─────┐
+//!                          │    Kafka     ├┐
+//!                          └┬─────────────┘├┐
+//!                           └┬─────────────┘│
+//!                            └──────────────┘
```
domodwyer, line 49: This diagram shpos the new updated request pipeline, including the position of the new `Partitioner`.

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -81,15 +83,20 @@ where
...
         &self,
         namespace: DatabaseName<'static>,
-        writes: HashMap<String, MutableBatch>,
+        writes: Self::WriteInput,
         span_ctx: Option<SpanContext>,
     ) -> Result<(), ShardError> {
         let mut collated: HashMap<_, HashMap<String, MutableBatch>> = HashMap::new();
 
+        // Extract the partition key & DML writes.
+        let (partition_key, writes) = writes.into_parts();
```
domodwyer, line 98: Here (in the write buffer layer) the partition key is extracted from the partitioned batch, and later logged.

in the future we would push this through the write buffer too.

domodwyer (commented):
It may be easier to go commit-by-commit for this PR when reviewing 👍

router2/src/dml_handlers/partitioner.rs
```diff
@@ -0,0 +1,449 @@
...
+            table,tag1=A,tag2=B val=42i 1644347270670952000\n\
+        ",
+        inner_write_returns = [Ok(()), Ok(()), Ok(())],
+        want_writes = [
+            "1970-01-01" => ["bananas"],
+            "2016-06-13" => ["platanos", "another"],
+            "2022-02-08" => ["table"],
+        ],
+        want_handler_ret = Ok(())
+    );
```
domodwyer, line 338: @pauldix these are the partition keys generated for the input LP (the date portion of the LP timestamp) 👍

router2/src/dml_handlers/mod.rs
NGA-TRAN: Does it mean the Partitioning is done before Sharding? I think it is fine to do like this but will it be easier to understand and have better performance if we partition after sharding because a shard will have many partitions

router2/src/dml_handlers/partitioner.rs
```diff
@@ -0,0 +1,449 @@
...
+                    fields = stats.num_fields,
+                    "partitioned write complete"
+                );
+                Ok(())
+            })
+            .await
+            .map_err(|e| PartitionError::Inner(Box::new(e.into())))
+    }
+
+    /// Pass the delete request through unmodified to the next handler.
```
NGA-TRAN, line 175: I am not sure if this means a delete also belong to a partition. @Paul has changed the design to have deletes not belong to any partition because the delete time range can be large than a partition

router2/src/dml_handlers/partitioner.rs
```diff
@@ -0,0 +1,449 @@
...
+        /// A static default time to use in tests (1971-05-02 UTC).
+        static ref DEFAULT_TIME: Time = Time::from_timestamp_nanos(42000000000000000);
+    }
+
+    // Generate a test case that partitions "lp" and calls a mock inner DML
+    // handler, which returns the values specified in "inner_write_returns".
+    //
+    // Assert the partition-to-table mapping in "want_writes" and assert the
+    // handler write() return value in "want_handler_ret".
+    macro_rules! test_write {
```
NGA-TRAN, line 211: I like you create macro and use it for all your tests. I need to learn this trick to do mine

router2/src/dml_handlers/sharded_write_buffer.rs
NGA-TRAN: I think I have got your idea to partition data first and then put them in the right shard. I wonder if partitioning on after sharding will be more optimized or not really

NGA-TRAN (approved):
The code look good. I just have questions regarding what should be done first partitioning or sharding and whether you partition delete.

router2/src/dml_handlers/partitioner.rs
```diff
@@ -0,0 +1,449 @@
...
+where
+    D: DmlHandler<WriteInput = Partitioned<HashMap<String, MutableBatch>>>,
+    T: TimeProvider,
+{
+    type WriteError = PartitionError;
+    type DeleteError = D::DeleteError;
+
+    type WriteInput = String;
+
+    /// Parse the input line-protocol string and emit partitioned batches to the
```
crepererum, line 102: Why is "parsing" and partitioning a single step? We could just partition the `MutableBatch` as the current stack does. It's easier to reason about, easier to test, and I very much doubt that touching the data twice is really an performance issue at the moment. We're about to copy the data at least 20 times afterwards anyways.

router2/src/dml_handlers/mock.rs
```diff
@@ -1,20 +1,18 @@
...
-use hashbrown::HashMap;
-use mutable_batch::MutableBatch;
 use parking_lot::Mutex;
 use trace::ctx::SpanContext;
 
 use super::{DmlError, DmlHandler};
 
 #[derive(Debug, Clone)]
-pub enum MockDmlHandlerCall {
+pub enum MockDmlHandlerCall<W> {
```
crepererum, line 12: I see you're a fan of very "generic" code. That's OK, but could you please document the generic parameters a bit more? What is `W`, esp. since `batches: W` doesn't make any sense to me.

tustvold:
Just an observation we already have the ability to partition `MutableBatch` into a collection of `PartitionWrite`. These can then be written to new (or existing) `MutableBatch`. We used to partition line protocol, but it was explicitly removed to avoid having to implement a separate notion of partitioning for the gRPC API.

_As an aside the IOx CLI doesn't write to the HTTP API, it converts to line protocol locally and then writes via the gRPC API_

router2/src/dml_handlers/partitioner.rs
domodwyer: I thought I was going to have to implement splitting the MB myself and figured it would be faster to get row-wise splitting working first to get us going (as I think we're using in the data generator?). I too dislike mixing the "parse" and "partition" responsiblities!

Fortunately as @tustvold [points out](https://github.com/influxdata/influxdb_iox/pull/3690#issuecomment-1034755903) this MB splitting is all ready to go so I'll refactor to use that.

domodwyer:
> Just an observation ...

A useful one! I hadn't found the splitting stuff as I was looking in the router code (forgetting it is all done consumer side currently). Thanks 👍 

As for HTTP/gRPC I think we're primarily targeting the HTTP API for the time being to get the test bench running which I believe uses the HTTP API - we'll be supporting gRPC at some point I'm sure though.

router2/src/dml_handlers/mock.rs
domodwyer: Yep that's fair! This was one of those `cargo check` -> edit -> `cargo check` -> edit induced namings 😂

router2/src/dml_handlers/partitioner.rs
domodwyer: This impl does not partition deletes, they are passed through without partitioning - we need to implement https://github.com/influxdata/influxdb_iox/issues/3629 to fully support deletes yet though.

router2/src/dml_handlers/mod.rs
domodwyer: I'm not sure it makes any difference - if you have data that is split into 3 partitions, and each partition into 2 shards, you have 3x2 "writes" and the same if you shard then partition (2x3)!

I might be missing a subtlety though - with the changes to partition the MB directly we'll be able to move this handler anywhere in the stack to empirically test what's best in the test bench 🎉

PR #3695, opened 2022-02-09
feat: abort database initialization (#3368)

tustvold:
The freeze handle within the Database state machine prevents concurrent in-flight mutations to the database state. This is generally a good thing, however, if an initialization process stalls out without returning an error - it will continue to hold this lock, preventing the operator from intervening to recover the system.

We need a mechanism to not only interrupt the current initialization action, but also prevent new ones from being created.

As such this adds a new `Aborted` state to the initialization state machine, along with a `Notify` that can be used to signal the initialization logic to bail out.

The astute may ask, "what about initialized databases", this case is actually already handled and is how the restart process works. Whilst running the database does not hold a freeze handle, so the restart process can acquire one, change the Database state and signal the background loop with `state_notify`. The background worker will exit out of its main loop, after waiting for everything to shutdown, and return to the init loop.

### DIFF
--- a/server/src/database.rs
+++ b/server/src/database.rs
@@ -183,6 +183,7 @@ impl Database {
             shutdown: Default::default(),
             state: RwLock::new(Freezable::new(DatabaseState::new_known())),
             state_notify: Default::default(),
+            abort_initialization: Default::default(),
         });
 
         let handle = tokio::spawn(background_worker(Arc::clone(&shared)));
@@ -230,17 +231,13 @@ impl Database {
         Ok(uuid)
     }
 
-    /// Triggers shutdown of this `Database`
-    pub fn shutdown(&self) {
-        let db_name = self.name();
-        info!(%db_name, "database shutting down");
-        self.shared.shutdown.cancel()
-    }
-
     /// Triggers a restart of this `Database` and wait for it to re-initialize
     pub async fn restart(&self) -> Result<(), Arc<InitError>> {
         let db_name = self.name();
         info!(%db_name, "restarting database");
+
+        // Abort any in-progress initialization
+        self.abort_initialization();
 
         let handle = self.shared.state.read().freeze();
         let handle = handle.await;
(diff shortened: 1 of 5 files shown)


server/src/database/init.rs
```diff
@@ -619,57 +694,34 @@ pub(crate) async fn initialize_database(shared: &DatabaseShared) {
...
-                    },
-                    Err(e) => DatabaseState::ReplayError(state, Arc::new(e)),
-                }
+            _ => {}
+        }
+
+        // Try to advance to the next state
+        let next_state = tokio::select! {
+            next_state = state.advance(shared) => next_state,
+            _ = shared.abort_initialization.notified() => {
```
tustvold, line 718: Here is where we now respond to shutdown and abort signals.

server/src/database/init.rs
```diff
@@ -171,6 +175,9 @@ pub(crate) enum DatabaseState {
     CatalogLoadError(DatabaseStateRulesLoaded, Arc<InitError>),
     WriteBufferCreationError(DatabaseStateCatalogLoaded, Arc<InitError>),
     ReplayError(DatabaseStateCatalogLoaded, Arc<InitError>),
+
+    // Database initialization was aborted
+    Aborted(DatabaseStateKnown),
```
tustvold, line 180: The aborted state, discards any potentially invalid intermediate state. This is intentional as it avoids shenanigans due to `Db` with a partially replayed catalog or something similar

server/src/database/init.rs
```diff
@@ -619,57 +697,34 @@ pub(crate) async fn initialize_database(shared: &DatabaseShared) {
...
-                    Ok(state2) => match state2.advance(shared).await {
-                        Ok(state2) => DatabaseState::Initialized(state2),
-                        Err(e) => DatabaseState::ReplayError(state, Arc::new(e)),
-                    },
-                    Err(e) => DatabaseState::ReplayError(state, Arc::new(e)),
-                }
+            _ => {}
+        }
+
+        // Try to advance to the next state
```
tustvold, line 718: This is where we now respond to signals

server/src/database.rs
```diff
@@ -256,6 +253,27 @@ impl Database {
...
+    /// to the initialized state, and will require intervention.
+    ///
+    /// If the database is already initialized, this will have no effect
+    ///
+    /// This is useful for interrupting the initialization of a database, so that the
+    /// process can be restarted with different parameters
+    pub fn abort_initialization(&self) {
+        let db_name = self.name();
+        info!(%db_name, "aborting database init");
+        self.shared.abort_initialization.notify_waiters();
```
alamb, line 267: I guess I am worried about the complexity of having two similar (but not the same) notions of `shutdown` and `aborted`, especially in logic that is already complicated enough. 

I don't understand why we need both `abort_initialization` and `shutdown` -- in all the places that `abort` is checked for, can't we simply check the `shutdown` token?  Or perhaps remove the shutdown notifier and run all shutdowns through the `Aborted` (maybe rename to `Shutdown`) state?

server/src/database/init.rs
alamb: I also don't understand why this needs  `aborted` -- aka why isn't checking `shared.shutdown.cancelled()` here enough?

tustvold: Because cancelled will cause the background worker to exit it out, it cannot be reset - it's terminal

alamb (approved):
Update on this PR is that @tustvold  and I had a good chat (at least I think it was good 😆 ) and will re-asses tomorrow after some time to reflect.

There may be a way to unify this type of logic more coherently, but if not we'll just hold our nose and merge this as we don't want to spend too much time making this code beautiful if its ultimate fate is to be replaced by the NG architecture anyways

PR #3707, opened 2022-02-09
feat: Turn protobuf predicates into predicate::Predicate 

carols10cents:
This (mostly) implements the deserialization of a protobuf query containing a predicate received by the ingester in a flight doGet request.

Most of the code is inspired by the protobuf and conversion code in ballista, with some improvements. And I think I might have found a few bugs that I'm going to send in.

There are a few parts of the ballista conversion code that are commented out/unimplemented. Some of those I was able to figure out, I think. I still have a few `unimplemented`s in here; this PR's complexity feels like a fractal though, and every time I think I'm almost done, I go another level in.

I'm not sure how complex of queries we're expecting to send from the querier to the ingester.

There's also an `unimplemented!` for the serialization part that the querier will need to turn `predicate::Predicate`s into protobuf to send to the `ingester`; I figured that could come in another PR.


alamb (commented):
Thank you @carols10cents 

At a high level I would really prefer this code NOT be in IOx as it will just be one more thing we have to continually keep up to date (as new Exprs are added to datafusion, etc)

There is some recent work to break DataFusion into smaller crates -- see https://github.com/apache/arrow-datafusion/issues/1750

What would you think about making moving most of this code into a PR in the DataFusion repo to add a new crate (perhaps `datafusion-serialization`, or `ballista/rust/serialization`) and moving all the protobuf definition / serialization code there?

Benefits:
1. No code duplication across repos
2. IOx would naturally pick up changes to serialization that happened in the datafusion repo
3. Other people could potentially use that serialization library for their own project

?

carols10cents:
Yeah, I do think it would be better if this was in the datafusion repo, but it is going to be a bit of work to detangle the datafusion serialization part from the ballista serialization part since we only want the datafusion part. I'll file a datafusion bug with my plan to start discussion there on Monday!

alamb:
> Yeah, I do think it would be better if this was in the datafusion repo, but it is going to be a bit of work to detangle the datafusion serialization part from the ballista serialization part since we only want the datafusion part. I'll file a datafusion bug with my plan to start discussion there on Monday!


If this is blocking something in Iox / NG I think it would be fine to merge in for the short term, but for the long term I think it would be much better to have in DataFusion/Ballista

alamb (approved):
Proper extraction is tracked in https://github.com/apache/arrow-datafusion/issues/1832. However I think it is fine to have it in iox in the interim

PR #3711, opened 2022-02-10
feat: restartable `Database` (#3368)

tustvold:
As pointed out on #3695 having two notions of shutdown is confusing, in particular there is a separate notion of shutting down a database (abort in #3695) and shutting down the database worker.

This PR adds a new `Shutdown` state to `DatabaseState` that implies no background worker is running. The background worker will attempt to initialize the database, and on shutdown will simply exit, previously it would loop back to trying to initialize the database.

As a `Database` background worker can now change, the watchdog logic had to change. Previously a database worker loop exiting would trigger server shutdown, now it just triggers the state transition for the database to shutdown. I think this is probably actually better. We will likely want to add a gRPC API to restart a shutdown database.


server/src/database.rs
```diff
@@ -627,126 +603,176 @@ impl Database {
...
 
-        // Future that completes if the WriteBufferConsumer exits
-        let consumer_join = match &write_buffer_consumer {
-            Some(consumer) => futures::future::Either::Left(consumer.join()),
-            None => futures::future::Either::Right(futures::future::pending()),
-        }
-        .fuse();
-        futures::pin_mut!(consumer_join);
+/// The background worker for `Database` - there should only ever be one
+async fn background_worker(shared: Arc<DatabaseShared>, shutdown: CancellationToken) {
```
tustvold, line 676: I recommend looking at https://github.com/influxdata/influxdb_iox/pull/3711/files?w=1

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
```diff
@@ -257,67 +258,57 @@ mod tests {
...
             .create_database(make_rules(&other_db_name))
             .await
             .unwrap();
 
-        let other_db = server.database(&other_db_name).unwrap();
+        // Configure a panic in the worker of the database we're about to create
+        let panic_key = "database background worker: panic_test";
+        db::utils::register_panic_key(panic_key);
 
-        let serve_fut = test_serve(
```
tustvold, line 269: A panic in a database worker no longer triggers server termination

server/src/database.rs
```diff
@@ -222,43 +216,52 @@ impl Database {
         .await
         .context(CannotReleaseSnafu { db_name })?;
 
-        let mut state = self.shared.state.write();
-        let mut state = state.unfreeze(handle);
-        *state = DatabaseState::new_no_active_database();
```
tustvold, line 227: I hope to expunge this notion from the Database state machine in a future PR, as it very confusing. An inactive Database should just not exist :smile:

server/src/lib.rs
```diff
@@ -395,32 +395,6 @@ impl ServerStateInitialized {
             }
         };
 
-        // Spawn a task to monitor the Database and trigger server shutdown if it fails
```
tustvold, line 398: This is the server watchdog that has been removed

server/src/lib.rs
```diff
@@ -1413,6 +1375,9 @@ mod tests {
 
     #[tokio::test]
     async fn load_databases() {
+        let apples_name = DatabaseName::new("apples").unwrap();
```
tustvold, line 1378: This test was previously racy. It would keep `database` around despite dropping the `server` they came from. These zombie database would technically be shutdown, but as this wasn't an explicit state in database this was invisible

tustvold:
If we like this I intend to follow it up with PRs to

* Pre-empt init on shutdown
* Pass IOxObjectStore to Database on construction, hopefully eliminating a lot of complexity
* Add gRPC APIs to shutdown and restart databases (with options)
* Ability to wipe catalogs of shutdown databases

server/src/database/init.rs
```diff
@@ -138,6 +143,10 @@ pub enum InitError {
 ///                         |
 ///                         V
 ///                   [Initialized]
+///
+///                         |
+///                         V
+///                     [Shutdown]
 ///                         |
 ///                         V
 ///                       (end)
```
alamb, line 152:
```suggestion
```

server/src/database/init.rs
```diff
@@ -153,6 +162,8 @@ pub enum InitError {
 ///    to dump the potentially half-modified in-memory catalog before retrying.
 #[derive(Debug, Clone)]
 pub(crate) enum DatabaseState {
+    // Database not running, with an optional shutdown error
```
alamb, line 165: I like it 👍

server/src/database/init.rs
```diff
@@ -603,14 +614,14 @@ const MAX_BACKOFF: Duration = Duration::from_secs(500);
 
 /// Try to drive the database to `DatabaseState::Initialized` returns when
 /// this is achieved or the shutdown signal is triggered
-pub(crate) async fn initialize_database(shared: &DatabaseShared) {
+pub(crate) async fn initialize_database(shared: &DatabaseShared, shutdown: CancellationToken) {
```
alamb, line 617: I like passing in this token here

server/src/lib.rs
```diff
@@ -1424,24 +1389,41 @@ mod tests {
             .expect("failed to create database");
         let bananas_uuid = bananas.uuid().unwrap();
 
-        std::mem::drop(server);
+        assert!(bananas.is_initialized());
+
+        // Shutdown server
```
alamb, line 1394: shutting down the server explicitly and waiting for it to stop 👨‍🍳 😗

server/src/lib.rs
```diff
@@ -1424,24 +1389,41 @@ mod tests {
...
 
         assert_eq!(server.db_names_sorted(), vec!["apples", "bananas"]);
 
-        std::mem::drop(server);
+        let bananas_object_store = bananas.iox_object_store().unwrap();
 
-        bananas
-            .iox_object_store()
-            .unwrap()
+        // Shutdown server
```
alamb, line 1418: I don't know if it would add any value but you could also add:

```suggestion
        assert!(!apples.is_shutdown());
        assert!(!bananas.is_shutdown());
        // Shutdown server
```

To demonstrate that the server shutdown causes the databases to shutdown

server/src/database.rs
```diff
@@ -627,126 +599,178 @@ impl Database {
...
             warn!(%db_name, "database dropped without waiting for worker termination");
         }
     }
 }
 
-/// The background worker for `Database` - there should only ever be one
-async fn background_worker(shared: Arc<DatabaseShared>) {
-    let db_name = shared.config.read().name.clone();
-    info!(%db_name, "started database background worker");
+/// Spawn a new background worker for a database in the shutdown state
```
alamb, line 614:
```suggestion
/// Spawn a new background worker for a database in the `shutdown` state
/// The state is reset to `Known` then the background worker attempts to drive the 
/// Database through initialization
```

server/src/database.rs
```diff
@@ -180,38 +173,39 @@ impl Database {
...
+            let uuid = state
+                .uuid()
+                .context(CannotReleaseUnownedSnafu { db_name })?;
 
-            (uuid, iox_object_store)
+            (store, uuid)
         };
 
-        info!(%db_name, %uuid, "releasing database");
+        self.shutdown();
```
alamb, line 207: this is such a nicer pattern 👍

server/src/database.rs
```diff
@@ -180,38 +173,39 @@ impl Database {
...
+        let _ = self.join().await;
```
alamb, line 208: I think it might be worth logging an error if the `join()` returns `Err`. It also might just make more sense to return an error if the background worker got an error before we go mess with the object store state? 🤔

server/src/database.rs
```diff
@@ -222,43 +216,52 @@ impl Database {
...
     }
 
     /// Triggers a restart of this `Database` and wait for it to re-initialize
     pub async fn restart(&self) -> Result<(), Arc<InitError>> {
         let db_name = self.name();
         info!(%db_name, "restarting database");
 
-        let handle = self.shared.state.read().freeze();
-        let handle = handle.await;
+        // Ensure database is shut down
```
alamb, line 234: 😍

alamb (approved):
I like it a lot 🏅 

https://www.youtube.com/watch?v=28GLa9T2CtI

server/src/database.rs
tustvold: If the database worker panics I would still like to be able to restart the database :)

PR #3713, opened 2022-02-10
feat: Have the hakari CI job automatically add a commit if needed

### LINKED ISSUE #3680: Dependabot doesn't run `cargo hakari generate`
Dependabot updates like #3675 are flagged by our CI by default because dependabot doesn't run `cargo hakari generate` to update the `workspace-hack` crate. A developer can obviously jump in an provide a fixup, but it's additional effort. It would be nice if we can somehow hook `cargo hakari generate` into dependabot.

Ref https://github.com/dependabot/dependabot-core/issues/1524 which I think is a similar issue/cause.

CC @carols10cents

carols10cents:
Fixes #3680.

When someone (either dependabot or a person) opens a PR that needs changes to the workspace hack crate to update the features listed, or if other crates need to be changed to make them depend on the workspace hack crate, instead of CI, this will add a commit to the PR.

[Here's a successful CI run where there were no changes to commit](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/15381/workflows/914a7c73-6033-47c1-9aed-bede1d1656a3/jobs/119962).

When there are changes to commit, Circle CI needs permission to push to the repo according to these docs - https://support.circleci.com/hc/en-us/articles/360018860473-How-to-push-a-commit-back-to-the-same-repository-as-part-of-the-CircleCI-job so I'll need an admin to do that.

I also added this commit to [this dependabot PR](https://github.com/influxdata/influxdb_iox/pull/3709). Interestingly, cargo-hakari is trying to undo the dependabot changes 😂 but [this run](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/15383/workflows/6256b04c-5368-4f5a-ab7e-1cede6a024cd/jobs/119991) shows everything working except the push back to github because Circle CI doesn't have permissions.


.circleci/config.yml
```diff
@@ -170,19 +170,30 @@ jobs:
...
       - run:
-          name: Check that the workspace hack crate contains all features in use
+          name: Configure git
           command: |
-            cargo hakari generate --diff || {
-              echo "If this fails, fix it by running \`cargo hakari generate\` locally and committing the changes"
-              exit 1
-            }
+            git config user.email "49699333+dependabot[bot]@users.noreply.github.com"
+            git config user.name "dependabot[bot]"
```
crepererum, line 176: Since this will push to ALL PRs, not just the dependabot ones, I think we should use a neutral config here. How about:

```suggestion
            git config user.email "????"
            git config user.name "CircleCI[bot]"
```

(I'm not sure about the email address though)

crepererum (commented):
Thinking about it: could we filter the GIT push to only run on dependabot PRs (e.g. by matching on the `user.name` of the last commit since IIRC CircleCI doesn't get a PR title). I personally find it quite annoying if bots mess w/ my PRs (but other devs might have a different opinion)

domodwyer:
I would like this - I hate that lint 😂 

Can I help at all?

carols10cents:
@domodwyer I definitely need help with giving Circle CI permissions to push to the repo, I don't have the power to do that! Putting in chat a todiscuss about whether to do this for everyone or just dependabot...

.circleci/config.yml
carols10cents: Fun fact: GitHub doesn't care at all about whether this email address is valid or actually belongs to the user making the commit ;) So i've set it to circleci@influxdata.com, which doesn't need to actually exist? Does that seem ok?

carols10cents:
Ok I'm going to leave this as a demonstration/test but close it and open a different PR for actually merging

carols10cents:
reopening to retry something

carols10cents:
ok done testing for now

carols10cents:
Don't mind me

PR #3715, opened 2022-02-10
refactor: require UUID to create Database

tustvold:
Draft as builds on #3711 

`IOxObjectStore::load` performs an object store list operation in order to determine if a rules file exists at the requisite path. As this is both async, and potentially fallible this is done as part of the init state machine of `Database`. Unfortunately this means that the UUID of a Database along with its `IOxObjectStore` is not known until it has reached a certain part of the init process.

This complicates a fair bit of logic, and most importantly makes for somewhat questionable logic in the routines that need to shutdown the database. For example, wipe catalog takes a copy of the IOxObjectStore from the initialized database, then shuts down the Database, and uses the now orphaned IOxObjectStore to perform the operation.

This PR therefore makes the UUID and by extension the IOxObjectStore a immutable property of Database provided at creation time. If the rules file doesn't exist, initialization will stall out in `RulesLoadError`, much as it would have done before if the rules file were deleted between `IOxObjectStore::load` and the `OwnerInfoLoaded -> RulesLoaded` transition.

This simplifies the logic, however, means **we will no longer be able to load ServerConfig with non-UUID databases**.


server/src/database/init.rs
```diff
@@ -106,15 +103,9 @@ pub enum InitError {
...
-///  o-o           o--------|-------------------o                         o-o
-///  | V           V        V                   V                         V |
-/// [NoActiveDatabase]<--[Known]------------->[DatabaseObjectStoreLookupError]
-///        |                |                           |
-///        o----------------+---------------------------o
-///                         |
-///                         |                                     o-o
-///                         V                                     V |
-///              [DatabaseObjectStoreFound]------>[OwnerInfoLoadError]
+///                         |----------------------------o     o-o
```
alamb, line 106: 😍

server/src/lib.rs
```diff
@@ -286,6 +286,12 @@ pub enum InitError {
 
     #[snafu(display("error persisting initial server config to object storage: {}", source))]
     PersistInitialServerConfig { source: object_store::Error },
+
+    #[snafu(display("invalid database name in server config: {}", source))]
+    InvalidDatabase { source: DatabaseNameError },
+
+    #[snafu(display("invalid database location in server config: {}", source))]
```
alamb, line 293:
```suggestion
    #[snafu(display("invalid database uuid in server config while finding location: {}", source))]
```

server/src/lib.rs
```diff
@@ -366,36 +441,19 @@ impl ServerStateInitialized {
...
-                        .fail();
-                    }
-                } else {
-                    return DatabaseAlreadyExistsSnafu {
-                        db_name: config.name,
-                    }
-                    .fail();
-                }
+            )))),
+            Entry::Occupied(_) => DatabaseAlreadyExistsSnafu {
```
alamb, line 452: well that is a lot nicer

server/src/lib.rs
```diff
@@ -1453,134 +1432,10 @@ mod tests {
         assert!(apples.init_error().is_none());
 
         let err = bananas.wait_for_init().await.unwrap_err();
-        assert_contains!(err.to_string(), "No rules found to load");
+        assert_contains!(err.to_string(), "rules.pb not found");
         assert!(Arc::ptr_eq(&err, &bananas.init_error().unwrap()));
     }
 
-    #[tokio::test]
```
alamb, line 1460: to confirm my understanding, the reason this test is removed is:

"this is testing that the "very old" style paths for databases (that did not include uuid) will no longer load. After this PR those "very old" style paths are not supported anymore (which is fine) so this test is no longer needed either"

alamb (approved):
Looks like a great simplification to me

PR #3717, opened 2022-02-10
feat: preempt init shutdown (#3368) 

### LINKED ISSUE #3368: Abort Database Init State Machine

tustvold:
Closes #3368 

This is an alternative implementation of #3695 that instead of introducing a new abort signal, uses the shutdown signal. This builds on #3711 that made Database restartable


server/src/database/init.rs
```diff
@@ -260,6 +260,55 @@ impl DatabaseState {
...
         }
     }
+
+    /// Try to advance to the next state
+    ///
+    /// # Panic
+    ///
+    /// Panics if the database cannot be advanced (already initialized or shutdown)
+    async fn advance(self, shared: &DatabaseShared) -> Self {
+        match self {
```
tustvold, line 270: This is just moved to make the select easier to write

server/src/database/init.rs
```diff
@@ -516,51 +565,23 @@ pub(crate) async fn initialize_database(shared: &DatabaseShared, shutdown: Cance
...
-                // Already initialized
-                break;
+            _ => {}
+        }
+
+        // Try to advance to the next state
+        let next_state = tokio::select! {
+            next_state = state.advance(shared) => next_state,
+            _ = shutdown.cancelled() => {
+                info!(%db_name, "initialization aborted by shutdown");
```
alamb, line 582: 👍

alamb (approved):
very nice

PR #3722, opened 2022-02-11
feat: server and database state gauge metrics

tustvold:
This adds two gauge metrics `server_state` and `database_state` that report the init status of the server and database respectively. In the case of the `server_state` it also reports the server id.

This should make it easier to view the state of all databases across one or more IOx servers.


mkmik (approved):
LGTM

server/src/lib.rs
```diff
@@ -2288,6 +2382,61 @@ mod tests {
...
+
+        database.shutdown();
+        database.join().await.unwrap();
+
+        let reporter = report();
+        let db_state = reporter.metric("database_state").unwrap();
+        assert_eq!(db_state.observations.len(), 1);
+        assert_eq!(
+            db_state.observations[0].0,
+            Attributes::from(&[("name", "test_db"), ("state", "Shutdown")])
```
e-dard, line 2436: Just to be clear then, when a database is no longer initialised the gauge for that metric doesn't get decremented to `0`, instead the metric just doesn't get emitted (and a different metric is emitted in it's place with `state=Shutdown`).

It's important to understand/document the semantics because it affects how an operator uses the metrics to understand what's going on (e.g., how they build a monitoring query).

e-dard (approved):
Looks good. 👍

Issue #3723, opened 2022-02-11
Revert hardcoded Kafka options for prod


crepererum:
placeholder

PR #3724, opened 2022-02-11
fix: hard-code prod kafka config

crepererum:
Prod has a larger max msg. size for Kafka (10MB instead of 1MB), but
currently we're unable to wire all the write buffer configs through. As
a quick fix lets hard code the config. This however breaks the write
buffer when running under default Kafka (1MB), so we should reverse this
(tracked under #3723).


mkmik (approved):

alamb:
🚀

PR #3725, opened 2022-02-11
feat: static router configuration (#2436)

tustvold:
Part of #2436 

Provides the ability to give the router command a path to a file to load configuration from. This file is expected to contain a pbjson encoded protobuf message, that in turn contains a list of the `Router` protobuf messages used by the gRPC APIs. Currently this file must be JSON, but it would be relatively trivial to extend to yaml, toml, etc... if we so wished.

If provided a config file, the mutation APIs will return an error, but the immutable APIs still function as normal.


mkmik (approved):
simple is good

PR #3730, opened 2022-02-11
refactor: traitify database configuration (#2436)

tustvold:
Relates to #2436 

This consolidates the various server config logic into a single `ConfigProviderObjectStorage` with a corresponding `ConfigProvider` trait. Aside from unifying code that was previously distributed across a number of different files and functions, this will also allow a subsequent PR to provide a `ConfigProvider` backing onto a local filesystem.


server/src/config.rs
```diff
@@ -0,0 +1,225 @@
...
+use generated_types::google::FieldViolation;
+use generated_types::influxdata::iox::management;
+use iox_object_store::IoxObjectStore;
+use object_store::ObjectStore;
+use snafu::{ensure, ResultExt, Snafu};
+use std::sync::Arc;
+use uuid::Uuid;
+
+/// A generic opaque error
+pub type StdError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
alamb, line 14:
```suggestion
pub type BoxedError = Box<dyn std::error::Error + Send + Sync + 'static>;
```

I don't feel strongly, just I felt slightly mislead with the `std` prefix

server/src/config.rs
```diff
@@ -0,0 +1,225 @@
...
+/// of a database server
+#[async_trait]
+pub trait ConfigProvider: std::fmt::Debug + Send + Sync {
+    /// Returns a list of database name and UUID pairs
+    async fn fetch_server_config(
+        &self,
+        server_id: ServerId,
+    ) -> Result<Vec<(String, Uuid)>, StdError>;
+
+    /// Returns a list of database names and UUID pairs
```
alamb, line 26:
```suggestion
    /// Persists a list of database names and UUID pairs
```

It also might help to make it clear if this list overwrites the existing list, or if it is additive, or if it only writes deltas, etc

server/src/config.rs
```diff
@@ -0,0 +1,225 @@
...
+    ) -> Result<(), StdError>;
+
+    /// Returns the configuration for the database with the given `uuid`
+    async fn fetch_rules(&self, uuid: Uuid) -> Result<ProvidedDatabaseRules, StdError>;
+
+    /// Persists the configuration for the database with the given `uuid`
+    async fn store_rules(&self, uuid: Uuid, rules: &ProvidedDatabaseRules) -> Result<(), StdError>;
+}
+
+/// Error enumeration for [`ConfigProviderObjectStorage`]
```
alamb, line 40: As a follow on PR this would be really nice to move into `server/src/config/object_store.rs` to keep the module smaller

alamb (approved):
LGTM

Something else that might be nice to consider while you are messing with this code would be to more cleanly separate the "environmental" config from the database config

Specifically, I mean pulling the kafka connection string out of the `DatabaseRules` and putting them somewhere else

PR #3731, opened 2022-02-13
doc: add doc about installation prerequisites

jackwener:
Closes #No related issue

When I build `influxdb_iox`, an error occurred.

I found that it is due to missing `openssl`, so I open a PR to polish the README.


README.md
```diff
@@ -136,6 +137,26 @@ PROTOC_INCLUDE=/opt/homebrew/include
...
+
+Install the `openssl` library.
+
+```shell
+# ubuntu
+sudo apt-get install libssl-dev
+
+# opensuse
+sudo zypper install libopenssl-devel
+```
```
e-dard, line 150:
```suggestion
Check if you have `openssl` installed with `$ openssl version`. If it's missing then openssl can typically be installed with your system package manager.
```

I suggest that we don't get into the habit of enumerating all the ways to install a build dependency for the various platforms IOx targets.

e-dard (changes requested):
Hey @jackwener thanks so much for the contribution. I propose a small change, which I'll add to the branch if you're happy with that.

e-dard:
Thank you for the contribution @jackwener 🙏. I have confirmed you have signed the CLA.

@jackwener the commit style we use here is [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), you will need to rewrite the [eef2a07](https://github.com/influxdata/influxdb_iox/pull/3731/commits/eef2a07825e870a91b024dce4e91cac8ab1136de) commit to have the prefix `docs:` rather than `[doc]: `.

jackwener:
Thanks for you review, I've fix those problems.

domodwyer (approved):
Thanks @jackwener!  ❤️

PR #3732, opened 2022-02-13
feat: static database configuration (#2436)

### LINKED ISSUE #2436: Static Configuration

### LINKED ISSUE #2463: Delete Predicate API

tustvold:
Closes #2436

Follows on from #3730 and allows specifying a server's configuration as a single config file, much like #3725 added for routers.

There are some aspects of this design worth highlighting:

* The config file contains UUIDs. Whilst somewhat unfortunate from a UX perspective, especially since they must be base64 encoded, this gives greater flexibility and preserves the ability to migrate databases between nodes
* The config file separates the list of databases from the rules, this is to reduce duplication where you wish to declare a number of databases with the same configuration
* Does not create ownership files in object storage, as I couldn't devise a satisfactory mechanism to "release" a database. FWIW the static configuration should make it easier to audit the sorts of collisions this mechanism was designed to protect against anyway
* Databases fetch the configuration file on startup, and so restarting a database will load any configuration changes. We may wish to provide a similar mechanism at the server level to reconcile added/removed databases - e.g. `SIGHUP`


generated_types/protos/influxdata/iox/management/v1/server_config.proto
```diff
@@ -42,3 +45,23 @@ message OwnershipTransaction {
...
 }
+
+// A user-provided configuration file listing the databases for this server
+message ServerConfigFile {
+  message DatabaseConfig {
+    /// The name of this database
+    string name = 1;
+
+    // The UUID for this database
+    bytes uuid = 2;
```
e-dard, line 56: I know it's not part of this API, but this whole DB name / UUID coupling is starting to bother me. It's not a good UX at all. I don't want to have to tie these pieces of information together all the time to do something as an operator.

tustvold: No disagreements from me, FWIW I fully intend to not actually use UUIDs when generating these configs in k8s-idpe. IOx just cares that they are a unique sequence of 16 bytes, not that they are a valid UUID :smile:

influxdb_iox/tests/end_to_end_cases/management_cli.rs
```diff
@@ -1806,3 +1807,166 @@ pub async fn setup_load_and_persist_two_partition_chunks(
...
+            "bufferSizeSoft": 1048576,
+            "bufferSizeHard": 10485760,
+            "mubRowThreshold": 1337
+          }
+        }
+      ],
+      "databases": [
+        {
+          "name": "test_db1",
+          "uuid": "tRFBjozQEey+58uRDU+9uA==",
```
e-dard, line 1834: Is it possible to base64 encode these inside of IOx when parsing the configuration? Worse than having to specify a UUID in the config is having to base64 encode it. Could we (inside IOx) do this so that the uuid config value is more readable to an operator in a file/repo/wherever?

e-dard (approved):
Code LGTM 👍 

I think it would be better to represent the uuids (given we are stuck with them for now) as plain text and then get IOx to encode them when it parses the config. I don't know if that's possible or not though?

influxdb_iox/tests/end_to_end_cases/management_cli.rs
tustvold: If by a string you mean like "c007048b-273a-4317-baa5-133086885d90" we definitely could

generated_types/protos/influxdata/iox/management/v1/server_config.proto
alamb: The usecase for uuid I think is so that we can handle duplicated database names (aka delete a database, make a new one with the same name, and now want to undelete the initial one)

influxdb_iox/tests/end_to_end_cases/management_cli.rs
tustvold: Going to do that, the base64 is actually an artifact of the protobuf <-> JSON conversion for bytes

generated_types/src/google.rs
```diff
@@ -491,6 +493,7 @@ impl PreconditionViolation {
             Self::PartitionInvalidState(description) => description.clone(),
             Self::ChunkInvalidState(description) => description.clone(),
             Self::RouterConfigImmutable => "router configuration is not mutable".to_string(),
+            Self::DatabaseConfigImmutable => "database configuration is not mutable".to_string(),
```
alamb, line 496: Maybe we could hint what the user can do to fix this. 

`"database configuration is not mutable (specified from config file)"`

?

influxdb_iox/src/commands/run/database.rs
```diff
@@ -88,17 +88,25 @@ pub struct Config {
         default_value = "no"
     )]
     pub skip_replay_and_seek_instead: BooleanFlag,
+
+    /// Path to a configuration file to use for routing configuration, this will
+    /// disable dynamic configuration via `influxdata.iox.management.v1.ManagementService`
+    ///
+    /// The config file should contain a JSON encoded `influxdata.iox.management.v1.ServerConfigFile`
```
alamb, line 95: 👍

influxdb_iox/src/influxdb_ioxd/server_type/database/mod.rs
```diff
@@ -28,6 +29,7 @@ pub struct DatabaseServerType {
     pub lp_metrics: Arc<LineProtocolMetrics>,
     pub max_request_size: usize,
     pub serving_readiness: ServingReadiness,
+    config_immutable: bool,
```
alamb, line 32: I am surprised this is needed as well -- I wonder if we could have instead have added `ConfigProvider::can_update(&self) -> bool` as the config provider was already threaded through

alamb (commented):
Looks like a good set of changes overall 👍

PR #3733, opened 2022-02-14
chore: Update datafusion, update `arrow`/`parquet`/`arrow-flight` to 9.0

alamb:
# Rationale:
1. Update to arrow 9.0.0 to get perf improvements from @tustvold 

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/81e76edc330067332ac9a3972525ecbe92953267
2. Update arrow / flight / parquet to 9.0.0


alamb:
Looking into failures

parquet_catalog/src/dump.rs
```diff
@@ -301,8 +301,8 @@ File {
                                         file_name: "00000000-0000-0000-0000-000000000001.parquet",
                                     },
                                 ),
-                                file_size_bytes: 3052,
-                                metadata: b"metadata omitted (935 bytes)",
+                                file_size_bytes: 1809,
```
alamb, line 304: @tustvold  (and @crepererum ) -- I updated these sizes because the test data now has only a single row group (due to https://github.com/apache/arrow-rs/pull/1214)

The changes look reasonable to me but I was hoping you could offer a second check. In particular I am not sure how important it is to have multiple row groups in the test file to maintain coverage.

crepererum: I think the test code of "parquet statistics -> IOx statistics" kinda relies on that, let me check...

alamb: Sounds good -- I will do so 👍

parquet_file/src/storage.rs
```diff
@@ -87,11 +87,22 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(Debug, Clone)]
 pub struct Storage {
     iox_object_store: Arc<IoxObjectStore>,
+
+    // If `Some`, restricts the size of the row groups created in the parquet file
+    max_row_group_size: Option<usize>,
```
alamb, line 92: added this field to `Storage` so that the tests could restrict the size of row group written and ensure more than one row group is present

parquet_catalog/src/dump.rs
alamb: I have made this change (mostly in https://github.com/influxdata/influxdb_iox/pull/3733/commits/04ae4c9424c03b64adcbcf12561a6657f0d21b22)

ingester/src/persist.rs
```diff
@@ -38,7 +40,17 @@ pub async fn persist(
         .expect("record_batches.is_empty was just checked")
         .schema();
 
-    let data = parquet_file::storage::Storage::parquet_bytes(record_batches, schema, metadata)
+    // Make a fake IOx object store to conform to the parquet file
```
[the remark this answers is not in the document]
tustvold: `IOxObjectStore::existing` might be better, I think this can spuriously fail for network reasons

tustvold (approved):
Re-reviewed and looks good, I'd recommend using IOxObjectStore::existing as IOxObjectStore::create makes actual requests IIRC

alamb:
Sorry for the rebase, but the semantic PR check doesn't like when I do manual merges from master 😢 

Screen Shot 2022-02-14 at 5 12 15 PM

crepererum (approved):
Thanks

ingester/src/persist.rs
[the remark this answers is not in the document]
alamb: In 9d68a6dd

PR #3734, opened 2022-02-14
fix: use encoded data size estimation instead of mutable buffer

crepererum:
For sparse data the PB-encoded data (our Kafka wire format) is way
smaller than the MutableBatch (up to a factor 20). So lets use this one
to estimate the size during batching.


tustvold (approved):
If I understand this correctly we're trading potentially encoding a MutableBatch to a protobuf multiple times, in exchange for a more accurate size estimation. I think this makes sense, but I'm not sure if the additional encoding costs will outweigh the benefits of aggregation

crepererum:
> If I understand this correctly we're trading potentially encoding a MutableBatch to a protobuf multiple times, in exchange for a more accurate size estimation. I think this makes sense, but I'm not sure if the additional encoding costs will outweigh the benefits of aggregation

If that becomes an issue we can optimize the encoding. Also I it's part of the routing tier which can be scaled up rather easily.

PR #3742, opened 2022-02-14
feat: build Query Plan that queries QueryableBatch with filters

### LINKED ISSUE #3729: Build a Query plan for QueryableBatch that applies Projection & Predicate pushdown
# Rationale
This is a task of #3637. Now we have Projection & Predicate pushdown available (#3693, #3719) for QueryableBatch, which ticket is to build and run a Query Plan that applies those pushdown.

The implementation is similar to [compact](https://github.com/influxdata/influxdb_iox/blob/bf6691e56fd4e910d609c0315eca8bb9a0ceec80/ingester/src/compact.rs#L159) but accepts 2 parameters `Selection` and `Predicate` for the pushdown work

NGA-TRAN:
Closes #3729


ingester/src/query.rs
```diff
@@ -259,8 +261,7 @@ impl QueryChunk for QueryableBatch {
         // Return stream of data
         let dummy_metrics = ExecutionPlanMetricsSet::new();
         let mem_metrics = MemTrackingMetrics::new(&dummy_metrics, 0);
-        let stream =
-            SizedRecordBatchStream::new(self.schema().as_arrow(), stream_batches, mem_metrics);
+        let stream = SizedRecordBatchStream::new(schema.as_arrow(), stream_batches, mem_metrics);
```
NGA-TRAN, line 264: Fixed a bug here. The provided schema must be the one of the projection pushed down

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+        "Error while executing the query for getting Ingester data to send to Querier"
+    ))]
+    ExecutePlan { source: DataFusionError },
+}
+
+/// A specialized `Error` for Ingester's Query errors
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Query a given Queryable Batch
+pub async fn query(
```
NGA-TRAN, line 45: This is the main function of this PR

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
+//! Hanlde all requests from Querier
```
alamb, line 1:
```suggestion
//! Handle all requests from Querier
```

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+        Selection::All => None,
+        Selection::Some(columns) => Some(
+            data.schema()
+                .compute_select_indicies(columns)
+                .context(SelectColumnsSnafu)?,
+        ),
+    };
+
+    let mut expr = vec![];
+    if let Some(filter_expr) = predicate.filter_expr() {
```
alamb, line 64: 👍  https://github.com/influxdata/influxdb_iox/blob/85aa019f5004a0c6c536a75384d7dd2428547beb/predicate/src/lib.rs#L80-L93

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+        expr.push(filter_expr);
+    }
+
+    let ctx = executor.new_context(ExecutorType::Reorg);
```
alamb, line 68: I am not sure about this threadpool  (Maybe `ExecutorType::Query` is more appropriate?)

However, perhaps the specific thread pools are less important as the query / ingest / compact roles are broken down into separate processes

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[snafu(display(
+        "Error while executing the query for getting Ingester data to send to Querier"
+    ))]
+    ExecutePlan { source: DataFusionError },
+}
+
+/// A specialized `Error` for Ingester's Query errors
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// Query a given Queryable Batch
```
alamb, line 44:
```suggestion
/// Query a given Queryable Batch, applying selection and filters as appropriate
```

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+        assert_batches_eq!(&expected, &output_batches);
+    }
+
+    #[tokio::test]
+    async fn test_query_with_filter() {
+        test_helpers::maybe_start_logging();
+
+        // create input data
+        let batches = create_one_record_batch_with_influxtype_no_duplicates().await;
+        //let tombstones = vec![create_tombstone(1, 1, 1, 1, 0, 200000, "tag1=UT")];
```
alamb, line 137: left over?

ingester/src/querier_handler.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        // build queryable batch from the input batches
+        let batch = make_queryable_batch_with_deletes("test_table", 1, batches, tombstones);
+
+        // make filters
+        // Only read 2 columns: "tag1" and "time"
+        let selection = Selection::Some(&["tag1", "time"]);
+
+        // tag1=VT
+        let expr = col("tag1").eq(lit("UT"));
```
alamb, line 183: I think this filter says "UT" rather than "VT"

```suggestion
        let expr = col("tag1").eq(lit("VT"));
```

query/src/frontend/reorg.rs
```diff
@@ -68,11 +82,16 @@ impl ReorgPlanner {
             .build()
             .context(CreatingProviderSnafu { table_name })?;
 
-        // Logical plan to scan all columns with no predicates
-        let plan = LogicalPlanBuilder::scan(table_name, Arc::new(provider) as _, None)
-            .context(BuildingPlanSnafu)?
-            .build()
-            .context(BuildingPlanSnafu)?;
+        // Logical plan to scan given columns and apply predicates
+        let plan = LogicalPlanBuilder::scan_with_filters(
```
alamb, line 86: it might be worth double checking that `LogicalPlanBuilder::scan_with_filters` with an empty `Vec<Expr>` doesn't end up adding a `LogicalPlan::Filter` in the plan (at least by the time the physical plan is executed)

alamb (approved):

ingester/src/querier_handler.rs
NGA-TRAN: This is a good point. I will add a todo and create a ticket for us to come back to this resource management for router, ingester, querier, compacter

ingester/src/querier_handler.rs
NGA-TRAN: I will modify comment to make it `tag`=UT` to have the result as expected

query/src/frontend/reorg.rs
NGA-TRAN: I have verified:
1. run test in debug mod and the plan does not include FilterExec if the filters is an empty vector
2. In DF code, the `scan` always invokes `scan_with_filters` with empty filters

ingester/src/querier_handler.rs
NGA-TRAN: #3752

PR #3745, opened 2022-02-15
fix: defined behaviour when seeking to an unknown sequence number

crepererum:
The new, defined behavior is: "return an error once and then end the stream".


write_buffer/src/core.rs
```diff
@@ -605,11 +609,28 @@ pub mod test_utils {
...
-        assert!(handler_1_1_a.stream().poll_next_unpin(&mut cx).is_pending());
-        assert!(handler_1_2_a.stream().poll_next_unpin(&mut cx).is_pending());
+        let err = handler_1_1_a
+            .stream()
+            .next()
+            .await
+            .expect("stream not ended")
+            .unwrap_err();
+        assert_eq!(err.kind(), WriteBufferErrorKind::UnknownSequenceNumber);
+        assert!(handler_1_1_a.stream().next().await.is_none());
```
crepererum, line 623: This is the essential behavior: return an error and end the stream.

write_buffer/src/core.rs
```diff
@@ -168,15 +178,19 @@ pub trait WriteBufferWriting: Sync + Send + Debug + 'static {
...
     ///
     /// Note that due to the mutable borrow, it is not possible to have multiple streams from the same
     /// [`WriteBufferStreamHandler`] instance at the same time. If all streams are dropped and requested again, the last
-    /// offsets of the old streams will be the start offsets for the new streams. If you want to prevent that either
-    /// create a new [`WriteBufferStreamHandler`] or use [`seek`](Self::seek).
+    /// sequence number of the old streams will be the start sequence number for the new streams. If you want to
+    /// prevent that either create a new [`WriteBufferStreamHandler`] or use [`seek`](Self::seek).
+    ///
+    /// If the sequence number that the stream wants to read is unknown (either because it is in the future or because
+    /// some retention policy removed it already), the stream will return an error w/
```
e-dard, line 192:
```suggestion
    /// some retention policy removed it already), the stream will return an error with
```

I would avoid shorthand in public API docs.

write_buffer/src/mock.rs
```diff
@@ -393,12 +393,19 @@ pub struct MockBufferStreamHandler {
 
     /// Offset within the sequencer IDs.
     offset: u64,
+
+    /// Flags is the stream is terminated, e.g. due to "offset out of range"
```
e-dard, line 397:
```suggestion
    /// Flags if the stream is terminated, e.g. due to "offset out of range"
```

write_buffer/src/mock.rs
```diff
@@ -429,6 +436,27 @@ impl WriteBufferStreamHandler for MockBufferStreamHandler {
...
+                        None
+                    }
+                })
+                .max()
+                .map(|x| x + 1)
+                .unwrap_or_default();
+            if self.offset > next_offset {
+                self.terminated = true;
+                return Poll::Ready(Some(Err(WriteBufferError::unknown_sequence_number(
+                    String::from("unknown sequence number"),
```
e-dard, line 456: Is it worth logging the offsets here to see how far away they were?

e-dard (approved):
Couple of typos but LGTM 👍

PR #3746, opened 2022-02-15
feat(router2): MutableBatch partitioning

domodwyer:
This PR adds partitioning to the `router2` implementation by introducing a new `Partitioner` into the request pipeline that operates splits `MutableBatch` instances. 

This change keeps the namespace auto-creation (#3628) as the first handler that is called, ensuring we don't bother partitioning a request if the namespace cannot be created.

The partitioned writes are tagged with the generated partition key, and passed through the handler stack until reaching the write buffer where the key is logged and dropped - further work will allow us to pass this partition key across the write-buffer API boundary for use in the ingester in the future.

This PR replaces the row-wise implementation in #3690 and is pretty much identical other than the partitioning impl in commit 92218ce8.

---

* refactor: parametrise DML handler input type (e99922d5)

      Allow a DML handler to specify the write input type on which it operates.

      This allows us to construct a write handler pipeline that transforms the 
      request as it passes through the various handlers. We'll use this to implement
      a handler that annotates a normal set of table writes with the partition key,
      modifying downstream handlers to expect this annotated input.

* test: MockDmlHandler generic over write input (5c254339)

      Allow the MockDmlHandler to capture any input type given to the write() 
      method. This lets us reuse the mock across all handler implementations, 
      regardless of their expected write input type.

* feat: write partitioner (92218ce8)

      Implements a write partitioning DML handler that splits per-table MutableBatch
      instances into per-partition, per-table MutableBatch and concurrently calls
      the inner DML handler with each.

* refactor: namespace auto-creator generic input (c64e9f0d)

      Changes the NamespaceAutocreation handler to be generic over any WriteInput.

      This allows the NamespaceAutocreation layer to be placed anywhere in the 
      handler stack, without needing a prior transformation or specific write type.

* refactor: enable Partitioner in request pipeline (e0558000)

      Adds the Partitioner DML handler into the handler stack, modifying the input
      types of down-stream handlers to accept the partitioned data.


lukebond (approved):

domodwyer:
Good questions!

> the partition template can change over time

For the time being we opted to test with date-based partitioning to get us going - it'll definitely be configurable in the future though. It's pretty easy to support router side 👍 Might be worth discussing the priority of this in a chat (@pauldix).

> what is the rationale for using raw_entry_mut()

This is a good one! The entry API lets us look up an existing key and modify it, or insert a value if it does not exist (an upsert API, amungst other functionalities).

The `raw_entry_mut()` use here is to defer an allocation of the table name - the normal [entry API](https://docs.rs/hashbrown/0.12.0/hashbrown/hash_map/struct.HashMap.html#method.entry) has a signature of:

```rust
fn entry(&mut self, key: K)
```

Where `K` is the [generic type of the key](https://docs.rs/hashbrown/0.12.0/hashbrown/hash_map/struct.HashMap.html#impl-7) for the `HashMap` - in our case, this is an owned `String` because we have a `HashMap<String, value>`, so the entry API for our use case effectively becomes:

```rust
fn entry(&mut self, key: String)
```

Which means when we use the entry API, we need to clone the borrowed (`&str`) table name into an owned `String` to give it to the `entry()` lookup method. At this point, there are two possible outcomes:

* There was no entry for the table name - great, no problem, the owned `String` we cloned earlier is placed into the hashmap as the key and we give it the value - the hashmap now owns the `String` key.
* There was already an entry for the table name, meaning the hashmap already had a `String` key - we needlessly cloned the `&str` to a `String` to look up the key, only to throw it away unused because the key already exists! We don't need to put the `String` into the hashmap - a complete waste :(

However, `raw_entry_mut()` has a [`from_key()`](https://docs.rs/hashbrown/0.12.0/hashbrown/hash_map/struct.RawEntryBuilderMut.html#method.from_key) builder method that accepts the borrowed form of the key (simplified signature below):

```rust
fn from_key(self, k: &Q) -> RawEntryMut<...>
    where
        K: Borrow<Q>,
        Q: Hash + Eq,
```

By using `raw_entry_mut()` we can use the borrowed form of the key (`&str`) for the lookup (`Q` in the [signature](https://docs.rs/hashbrown/0.12.0/hashbrown/hash_map/struct.RawEntryBuilderMut.html#method.from_key) is the borrowed form of `K` as defined by the bound `K: Borrow<Q>` which can be read as "K is borrowed as type Q") to avoid cloning the table name to a `String` up front, instead cloning it only when we know we need to give an owned `String` key to the hashmap because the key doesn't exist which you can [see here](https://github.com/influxdata/influxdb_iox/pull/3746/files#diff-73ef3490e65635110094b2edac98b720496b10b6b47bab3055aab749dffa3fbeR108):

```rust
.or_insert_with(|| (table_name.to_owned(), MutableBatch::default()));
```

This is a handy optimisation for us as we can expect the same table name to appear many times within the same write due to batching, so if we had 100 LP lines for `cpu` measurements, we'd only have to allocate 1 `cpu` string instead of 100 🎉

PR #3748, opened 2022-02-15
fix: start Kafka write buffer stream at "earliest" offset, not at "0"

crepererum:


write_buffer/src/kafka/mod.rs
```diff
@@ -139,10 +140,23 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
...
+        let start_offset: Option<i64> = {
+            // need to trick a bit to make this async function `Send`
+            *next_offset.lock()
+        };
+        let start_offset = match start_offset {
+            Some(x) => x,
+            None => {
+                // try to guess next offset from upstream
+                self.partition_client
+                    .get_offset(OffsetAt::Earliest)
```
tustvold, line 152: Won't this race against retention, is there not some magic offset of `-1` or something that fetches the oldest data?

crepererum: Both race against upstream. But you're right, we should probably do this within rskafka where the offset of the underlying stream is tracked.

tustvold (approved):
Let's get this in and perhaps remove it once we fix it upstream

PR #3760, opened 2022-02-16
docs: Improve iox_catalog testing docs

mkmik:
1. the `iox_catalog` README.md mentions that by following the docs the migration script is run, but that only happens with `sqlx database setup`, not with `sqlx database create`.

2. added an example on how to craft the DSN string for talking to a db over a unix domain socket (useful because by default the homebrew postgres on macos doesn't open a tcp port)

3. added a section about the schema search path.


iox_catalog/README.md
```diff
@@ -11,11 +11,18 @@ user and password filled in:
 DATABASE_URL=postgres://<postgres user>:<postgres password>@localhost/iox_shared
 ```
 
+You can omit the host part if your postgres is running on the default unix domain socket (useful on macos because, by default,
+the config installed by `homebrew install postgres` doesn't listen to a TCP port):
```
domodwyer, line 15: I thought it was always `brew`?

```suggestion
the config installed by `brew install postgres` doesn't listen to a TCP port): 
```

domodwyer (approved):
👍

Issue #3577, opened 2022-01-28
Pre-warm router table schema cache


domodwyer:
When a router starts up, it will have an empty in-memory cache (#3576) and the inital writes will experience noticably higher latency while the cache is population is driven by requests - this will also likely cause a significant spike in load on the catalog DB.

This ticket covers pre-warming the cache by pulling the database start at startup (in a single DB call) or some other means.

PR #3764, opened 2022-02-16
feat: router2 metrics

domodwyer:
This PR adds instrumentation to each stage of the request handling pipeline to allow us to observe the latencies of the various operations within the router. It adds latency histograms the following stages:

* Submitting ops to the write buffer (inc. waiting for internal buffer flush)
* Schema validation (inc. catalog interactions)
* Write partitioning
* Overall e2e request handling latency

Each metric is broken down by success/error response and operation type (write/delete).

Because the request pipeline is constructed as a chain of nested handlers, the timing for each stage is inclusive of the child handlers it calls:

```
              ┌ ─                                          
                                                           
              │     ┌───────────────┐                      
                    │  NS Creation  │                      
              │     └───────────────┘                      
                            │  ┌───────────────┐           
              │             │  │  Partitioner  │           
                            │  └───────────────┘           
              │             │          │                   
                            │          │                   
 Cumulative   │             │          │  ┌───────────────┐
   Timings                1.5s        1s  │    etc...     │
              │             │          │  └───────────────┘
                            │          │                   
              │             │          │                   
                            │  ┌───────────────┐           
              │             │  │  Partitioner  │           
                            │  └───────────────┘           
              │     ┌───────────────┐                      
                    │  NS Creation  │                      
              │     └───────────────┘                      
                                                           
              └ ─                                          
```

This is temporary, and I'll switch to external chaining (rather than internally nesting) handlers in the next PR to give us better timing data - splitting them up to keep it small.

Additionally the namespace cache is instrumented to measure:

* Get hit/miss count
* Put insert/update count

This will help us measure the incidence of https://github.com/influxdata/influxdb_iox/issues/3578 and https://github.com/influxdata/influxdb_iox/issues/3577 and weight remediation of them appropriately.

---

* feat: instrumented namespace cache (92fe507e)

      Decorates the NamespaceCache with a set of cache get hit/miss counters, and
      put insert/update counters to expose cache behaviour.

* feat: metric instrumentation for DML handlers (40e5b193)

      Adds a decorator type over a DmlHandler implementation that records call 
      latency for writes & deletes, broken down by result (success/error).

* feat: add instrumentation to request pipeline (d6d0ae8d)

      Wraps the sharded write buffer, schema validator, partitioner and overall
      request handler in instrumentation to record call latencies and export them
      via the /metrics endpoint.

### DIFF
--- a/router2/src/dml_handlers/instrumentation.rs
+++ b/router2/src/dml_handlers/instrumentation.rs
@@ -34,33 +34,28 @@ pub struct InstrumentationDecorator<T, P = SystemProvider> {
 
 impl<T> InstrumentationDecorator<T> {
     /// Wrap a new [`InstrumentationDecorator`] over `T` exposing metrics
-    /// prefixed by `name`.
-    ///
-    /// # Memory leak
-    ///
-    /// Calling this method constructs a set of metric names derived from `name`
-    /// and leaks them once to acquire a static str for each.
+    /// labelled with `handler=name`.
     pub fn new(name: &'static str, registry: Arc<metric::Registry>, inner: T) -> Self {
         let buckets = || {
             U64HistogramOptions::new([5, 10, 20, 40, 80, 160, 320, 640, 1280, 2560, 5120, u64::MAX])
         };
 
         let write: Metric<U64Histogram> = registry.register_metric_with_options(
-            leak_concat_name(name, "_write_duration_ms"),
+            "dml_handler_write_duration_ms",
             "write handler call duration in milliseconds",
             buckets,
         );
         let delete: Metric<U64Histogram> = registry.register_metric_with_options(
-            leak_concat_name(name, "_delete_duration_ms"),
-            "write handler call duration in milliseconds",
+            "dml_handler_delete_duration_ms",
+            "delete handler call duration in milliseconds",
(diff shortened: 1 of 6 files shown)


router2/src/dml_handlers/instrumentation.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        res
+    }
+}
+
+fn leak_concat_name(prefix: &str, suffix: &str) -> &'static str {
+    let s = format!("{}{}", prefix, suffix);
+    Box::leak(s.into_boxed_str())
+}
```
crepererum, line 137: :see_no_evil:

crepererum (approved):
I like the layering.

A bit of testing would be nice though, at least some simple smoke tests to check that we indeed are producing some metrics. See https://github.com/influxdata/influxdb_iox/blob/a3e15ade62e63f88412b052ac1f5826bb7e0d06a/db/src/lib.rs#L1607-L1761 for examples.

domodwyer:
Pushed a commit to use labels rather than distinct metric names, but the tests asserting metrics are collected will be in the next PR as I already had an e2e test for the chained-handler change - hope you don't mind 👍

Issue #3635, opened 2022-02-03
Panic Simplifying Invalid Predicate


tustvold:
Calling IOx with an invalid predicate, for example one that compares incompatible types, causes a panic.

```
2022-02-03T18:34:45.499339Z ERROR panic_logging: Thread panic panic_info=panicked at 'Expression simplificiation round 1 failed: Plan("'Utf8 = Int64' can't be evaluated because there isn't a common type to coerce the types to")', predicate/src/rpc_predicate.rs:185:22
thread 'IOx Query Executor Thread' panicked at 'Expression simplificiation round 1 failed: Plan("'Utf8 = Int64' can't be evaluated because there isn't a common type to coerce the types to")', predicate/src/rpc_predicate.rs:185:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This should instead return an error to the user

alamb:
@tustvold  do you happen to remember how you found this panic? I am struggling to find a reproducer

tustvold:
I believe one of the InfluxQL test cases ran into it

alamb:
Thanks @tustvold  I'll try rerunning those

Note I created https://github.com/influxdata/influxdb_iox/pull/3763 to show my (failing) progress so far to create a reproducer

alamb:
This will reproduce the issue: 

```
go test -v ./influxql/v1tests -run TestServer_Query_ShowTagValues
```

```
2022-02-16T14:25:53.854939Z DEBUG query::frontend::influxrpc: planning tag_values rpc_predicate=InfluxRpcPredicate { table_names: Some({"_"}), inner: Predicate { field_columns: None, partition_key: None, range: None, exprs: [CASE WHEN #value IS NULL THEN Utf8("") ELSE #value END = Int64(5000)], value_expr: [] } } tag_name="host"
2022-02-16T14:25:53.855226Z ERROR panic_logging: Thread panic panic_info=panicked at 'Expression simplificiation round 1 failed: Plan("'Utf8 = Int64' can't be evaluated because there is
n't a common type to coerce the types to")', predicate/src/rpc_predicate.rs:185:22
thread 'IOx Query Executor Thread' panicked at 'Expression simplificiation round 1 failed: Plan("'Utf8 = Int64' can't be evaluated because there isn't a common type to coerce the types
to")', predicate/src/rpc_predicate.rs:185:22
```

input data looks like:
```
	writes := []string{
[...]

PR #3767, opened 2022-02-16
fix: Make errors in rewriting return `Error` rather than a `panic`

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3635
Closes https://github.com/influxdata/conductor/issues/889

# Rationale:
We (ahem, as the royal We) were being lax during predicate rewriting and `panic'ing` rather than properly handling errors for unsanitary predicates.

# Changes:
1. Return an error rather than `panic` while rewriting predicates
2. Tests for same


alamb:
cc @onelson

onelson:
I'm currently building it to re-run the flux tests, but I don't expect my test results to differ from yours. This looks like a rather mundane diff overall :sweat_smile:.

PR #3769, opened 2022-02-16
fix: supervise and shutdown ingester background tasks

### LINKED ISSUE #3761: Ingester: handler threads can die / finish unnoticed
For every sequencer we create a tokio task that essentially contains [`stream_in_sequenced_entries`](https://github.com/influxdata/influxdb_iox/blob/a3e15ade62e63f88412b052ac1f5826bb7e0d06a/ingester/src/handler.rs#L169-L274). There are two ways this function can exit:

1. The stream can just end. This happens when we encounter a Kafka error that we cannot deal with. This is likely always a bug in rskafka because most errors should be retried. However silently exiting is not a good option either.
2. Panic: if the function panics, it will just exit and nobody will notice (except that we log that)

My proposal would be to handle this similar to how we handle this in the current architecture: add a `join` function to `IngestHandlerImpl` that panics if one of the tasks exits early (either clean or via panic) and hook that up into the CLI here:

https://github.com/influxdata/influxdb_iox/blob/a3e15ade62e63f88412b052ac1f5826bb7e0d06a/influxdb_iox/src/influxdb_ioxd/server_type/ingester.rs#L68-L70

This way the CLI would exit if this is happening and K8s would restart the container. This way we have monitoring, retries and backoffs all solved.

### LINKED ISSUE #3762: Ingester: handler threads don't offer "clean shutdown"
To stop ingester handler tasks the only way is to drop `IngestHandlerImpl` which will abort the handler tasks. It would be nice if for a "normal" server shutdown we can have a cleaner solution that will at least finish it's current tasks (e.g. finish in-process persist operation or finish to the handling of the current DML operation).

crepererum:
Closes #3761.
Closes #3762.

Props for @tustvold who came up with the join handle management! :slightly_smiling_face:


ingester/src/handler.rs
```diff
@@ -50,20 +56,43 @@ const INGEST_PAUSE_DELAY: Duration = Duration::from_millis(100);
...
+    /// Wait until the handler finished  to shutdown.
+    ///
+    /// Use [`shutdown`](Self::shutdown) to trigger a shutdown.
+    async fn join(&self);
+
+    /// Shut down background workers.
+    fn shutdown(&self);
+}
+
+/// A [`JoinHandle`] that can be pulled from multiple sites.
```
tustvold, line 73:
```suggestion
/// A [`JoinHandle`] that can be cloned
```

ingester/src/handler.rs
```diff
@@ -110,16 +139,18 @@ impl IngestHandlerImpl {
             Arc::new(SystemProvider::new()),
         ));
         let manager = Arc::clone(&lifecycle_manager);
+        let shutdown = CancellationToken::new();
+        let shutdown_captured = shutdown.clone();
         let handle = tokio::task::spawn(async move {
-            run_lifecycle_manager(manager, persister).await;
+            run_lifecycle_manager(manager, persister, shutdown_captured).await;
```
tustvold, line 145: FWIW if you remove the `async move` and just call `run_lifecycle_manager()` and pass the future to `tokio::task::spawn` you might be able to move the captured dance

ingester/src/handler.rs
```diff
@@ -140,28 +171,54 @@ impl IngestHandlerImpl {
...
 }
 
+#[async_trait]
 impl IngestHandler for IngestHandlerImpl {
     fn query(&self, _request: IngesterQueryRequest) -> Result<QueryData> {
         todo!()
     }
+
+    async fn join(&self) {
+        let mut unordered: FuturesUnordered<_> = self.join_handles.iter().cloned().collect();
```
tustvold, line 196: Since these `JoinHandle` don't actually "do" anything, i.e. we don't really need to poll them in parallel, I'd be tempted to just have a simple loop and avoid the `FuturesUnordered` shenanigans

ingester/src/handler.rs
```diff
@@ -181,12 +238,19 @@ async fn stream_in_sequenced_entries(
...
     mut metrics: SequencerMetrics,
+    shutdown: CancellationToken,
 ) {
     let mut watermark_last_updated: Option<Instant> = None;
     let mut watermark = 0_u64;
     let mut stream = write_buffer_stream.stream().await;
 
-    while let Some(db_write_result) = stream.next().await {
+    while let Some(db_write_result) =
+        tokio::select!(next = stream.next() => next, _ = shutdown.cancelled() => {return})
```
tustvold, line 248:
```suggestion
        tokio::select!(next = stream.next() => next, _ = shutdown.cancelled() => return)
```

ingester/src/lifecycle.rs
```diff
@@ -238,10 +239,19 @@ const CHECK_INTERVAL: Duration = Duration::from_secs(1);
...
+        if shutdown.is_cancelled() {
+            return;
+        }
+
         manager.maybe_persist(&persister).await;
-        tokio::time::sleep(CHECK_INTERVAL).await;
+
+        tokio::select!(
+            _ = tokio::time::sleep(CHECK_INTERVAL) => {},
+            _ = shutdown.cancelled() => {return;},
```
tustvold, line 253:
```suggestion
            _ = shutdown.cancelled() => return,
```

tustvold: Perhaps also some info logs?

tustvold (approved):
I'm not very familiar with the code this plugs into, but it seems plausible to me.

It would perhaps be cool to have an end-to-end test of this, as such logic is often very fragile. Perhaps could copy the panic_key concept used by the old architecture.

Also some strategic info logging can go a long way to helping with when the logic inevitably does go awry :smile:

domodwyer (approved):
Looks good to me! I'll plumb something into the router shorlty 🎉

ingester/src/handler.rs
crepererum: This doesn't detect early exists, because you might still poll the first handler while the 2nd one in the list is early gone. Will add a comment and tests.

ingester/src/handler.rs
crepererum: done

crepererum:
Added some solid tests.

PR #3771, opened 2022-02-16
feat: generate a trace ID when using gRPC client

e-dard:
This PR basically lets you easily trace commands issued from the IOx gRPC client. To do this you need to:

- Have IOx running with an appropriate tracing configuration set (e.g., jaeger running locally).
- Have not changed the `--traces-exporter-jaeger-trace-context-header-name` argument from the default of `uber-trace-id`.

Assuming that, then when you want to trace your query (or other operation) you can now do:

```
$ influxdb_iox database query mydb "select * from cpu" --gen-trace-id
Trace ID set to fcde5072a178bf81:1112223334445:0:1

...
...
```

The trace ID will be automatically generated (based on a hash of the current time) and then emitted to you. IOx will then trace the command and ship that trace to jaeger.

You can then take the first part of the emitted ID, in this case `fcde5072a178bf81` and use that to lookup your trace, e.g., visiting `http://localhost:16686/trace/fcde5072a178bf81`.

### DIFF
--- a/influxdb_iox/src/main.rs
+++ b/influxdb_iox/src/main.rs
@@ -209,8 +209,12 @@ fn main() -> Result<(), std::io::Error> {
                     trace_exporters::DEFAULT_JAEGER_TRACE_CONTEXT_HEADER_NAME,
                 )
                 .unwrap();
-                let value = http::header::HeaderValue::from_str(gen_trace_id().as_str()).unwrap();
+                let trace_id = gen_trace_id();
+                let value = http::header::HeaderValue::from_str(trace_id.as_str()).unwrap();
                 builder = builder.header(key, value);
+
+                // Emit trace id information
+                println!("Trace ID set to {}", trace_id);
             }
 
             match builder.build(&host).await {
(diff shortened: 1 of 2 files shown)


influxdb_iox/src/main.rs
```diff
@@ -135,6 +140,11 @@ struct Config {
     #[clap(long, global = true)]
     header: Vec<KeyValue<http::header::HeaderName, http::HeaderValue>>,
 
+    #[clap(long, global = true)]
```
alamb, line 143: 👍

influxdb_iox/src/main.rs
```diff
@@ -190,10 +200,23 @@ fn main() -> Result<(), std::io::Error> {
...
+            let mut builder = headers.into_iter().fold(Builder::default(), |builder, kv| {
                 builder.header(kv.key, kv.value)
             });
 
+            if config.gen_trace_id {
+                let key = http::header::HeaderName::from_str(
+                    trace_exporters::DEFAULT_JAEGER_TRACE_CONTEXT_HEADER_NAME,
+                )
+                .unwrap();
+                let trace_id = gen_trace_id();
```
alamb, line 212: I like how it makes a trace id rather than having to specify it

alamb (approved):
👨‍🍳 👌

PR #3776, opened 2022-02-17
test: isolated catalog tests

domodwyer:
This PR enables catalog tests to run in parallel without interfering with each other.

While working on the catalog I added a couple tests and (eventually) discovered they were trashing each other's global catalog state as it was shared across tests. This is currently avoided by running all the test code runs in one test (`test_catalog`).

The test code paths now generate a random schema name for each catalog test (or more specifically, for each call to [`setup_db`](https://github.com/influxdata/influxdb_iox/blob/44e9eaf92b0740af5e2087fc03cf0220cea7323c/iox_catalog/src/postgres.rs#L928-L968)). The prod code paths [continue to use `iox_catalog`](https://github.com/influxdata/influxdb_iox/blob/44e9eaf92b0740af5e2087fc03cf0220cea7323c/influxdb_iox/src/clap_blocks/catalog_dsn.rs#L47) as the schema name.

A list of schmeas after running `TEST_INTEGRATION=1 cargo test` 3 times against a clean PG:

```
       List of schemas
         Name         | Owner
----------------------+-------
 bxnrwtwjigvjapwowyda | dom
 iox_catalog          | dom
 lrzbxfrakfnkkmnahyoc | dom
 pbxwasxwfjbgqmavanju | dom
 public               | dom
(5 rows)
```

---

* refactor: do not specify schema in migrations (3b378418)

      Allow the caller to set the Postgres schema a migration should be applied to,
      rather than restricting the migration to a specific, hard-coded schema.

      BREAKING CHANGE: manually adds a new migration that precedes the existing
      migration to ensure the iox_catalog schema exists before applying the
      migration. You'll probably have to drop any existing databases and migrate
      from scratch:

          sqlx database drop; sqlx database create;

* test(iox_catalog): isolated catalog tests (44e9eaf9)

      This commit changes the iox_catalog test harness so that each test is run in
      an independent, randomly generated schema to avoid concurrent tests
      interfering with each other.

      Each test creates a randomly-named schema, grants permissions to the new 
      schema, runs the full migration stack against the new schema, and then 
      executes the code under test.

### DIFF
--- a/iox_catalog/migrations/20211229171744_initial_schema.sql
+++ b/iox_catalog/migrations/20211229171744_initial_schema.sql
@@ -2,7 +2,7 @@
 -- iox_shared schema
 CREATE SCHEMA IF NOT EXISTS iox_catalog;
 
-CREATE TABLE IF NOT EXISTS iox_catalog.kafka_topic
+CREATE TABLE IF NOT EXISTS kafka_topic
 (
     id INT GENERATED ALWAYS AS IDENTITY,
     name VARCHAR NOT NULL,
@@ -10,7 +10,7 @@ CREATE TABLE IF NOT EXISTS iox_catalog.kafka_topic
     CONSTRAINT kafka_topic_name_unique UNIQUE (name)
     );
 
-CREATE TABLE IF NOT EXISTS iox_catalog.query_pool
+CREATE TABLE IF NOT EXISTS query_pool
 (
     id SMALLINT GENERATED ALWAYS AS IDENTITY,
     name VARCHAR NOT NULL,
@@ -18,7 +18,7 @@ CREATE TABLE IF NOT EXISTS iox_catalog.query_pool
     CONSTRAINT query_pool_name_unique UNIQUE (name)
     );
 
-CREATE TABLE IF NOT EXISTS iox_catalog.namespace
+CREATE TABLE IF NOT EXISTS namespace
 (
     id INT GENERATED ALWAYS AS IDENTITY,
     name VARCHAR NOT NULL,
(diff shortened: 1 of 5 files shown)


iox_catalog/src/postgres.rs
```diff
@@ -926,10 +926,45 @@ mod tests {
     }
 
     async fn setup_db() -> PostgresCatalog {
+        // create a random schema for this particular pool
```
mkmik, line 929: I loved this aspect of the conductor test suite; thanks!

iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -2,23 +2,23 @@
 -- iox_shared schema
 CREATE SCHEMA IF NOT EXISTS iox_catalog;
```
mkmik, line 3: isn't this now in `20210217134322_create_schema.sql` ?

iox_catalog/migrations/20211229171744_initial_schema.sql
```diff
@@ -2,23 +2,23 @@
...
 
-CREATE TABLE IF NOT EXISTS iox_catalog.kafka_topic
+CREATE TABLE IF NOT EXISTS kafka_topic
```
mkmik, line 5: modifying the existing migrations means that all developers will have to drop their databases when they re-run the tests.

are you worried that just adding a migration that drops all the tables and then another migration that adds the stuff in the "current" schema would slow down test setup on CI too much?

mkmik: > modifying the existing migrations means that all developers will have to drop their databases when they re-run the tests.

also I think testbench instances may be affected (they are meant to be disposable though)

domodwyer: The migration will run in the new, random schema so no one should have to drop anything - I only mention potentially needing to drop the DB if someone is running it "in prod" mode locally (using the `iox_catalog` schema).

Everything so far is a throw-away env - we've been modifying the original migration as needed until we deploy somewhere persistent.

iox_catalog/migrations/20211229171744_initial_schema.sql
domodwyer: Ah thanks - picking stuff across got messy.

mkmik (approved):

lukebond (approved):
thank you! i copied over part of it but didn't get the random schema working it seems :)

PR #3787, opened 2022-02-18
feat: catalog now supports dsn-file://

### LINKED ISSUE #3780: support dsn-file:// in iox catalog

lukebond:
Closes #3780

Ports this feature over from conductor.

I struggled a bit with static &str lifetimes, setting the app name after connect, and using the right app name (from dsn or from connect() fn), compounded by some conflicts in the postgres catalog code. So would appreciate some eyes on it!


mkmik (approved):
LGTM

iox_catalog/src/postgres.rs
```diff
@@ -240,6 +217,143 @@ impl Catalog for PostgresCatalog {
     }
 }
 
+/// Creates a new [`sqlx::Pool`] from a database config and an explicit DSN.
+///
+/// This function doesn't support the IDPE specific `dsn-file://` uri scheme.
+async fn new_raw_pool(
+    app_name: &str,
+    schema_name: &str,
```
domodwyer, line 225: As you're already creating `Arc<str>` in the caller, you can use arcs here and avoid some clones:

```suggestion
    app_name: Arc<str>,
    schema_name: Arc<str>,
```

It's not a big deal though - it's 1 clone of each str per connection per `new_raw_pool()` call.

iox_catalog/src/postgres.rs
```diff
@@ -240,6 +217,143 @@ impl Catalog for PostgresCatalog {
...
+                        let new_pool = new_raw_pool(app_name, schema_name, &new_dsn).await?;
+                        pool.replace(new_pool);
+                        Ok(Some(new_dsn))
+                    }
+                }
+
+                match try_update(&app_name, &schema_name, &current_dsn, &dsn_file, &pool).await {
+                    Ok(None) => {}
+                    Ok(Some(new_dsn)) => {
+                        info!("replacing hotswap pool");
```
domodwyer, line 332:
```suggestion
                        info!("replaced hotswap pool");
```

This is logged _after_ the call to `pool.replace()` - I'm sure this will confuse someone on day! Might be worth a log before too in case we're picking through logs?

PR #3788, opened 2022-02-18
refactor(router2): non-intrusive handler chain

domodwyer:
This refactor changes the way DML handlers are chained together in the router in order to produce more useful metrics (see #3764).

While this change caused a fair bit of code churn, the changes are pretty trivial compiler-driven fixups - the result is much nicer to test / work with though! An example of the benefit of this is it moved responsibility for parallelising certain operations (like writes across shards) out from the request handler and into an external type, separating responsiblities and making the parallisation pluggable / configurable / optional.

---

* refactor: chain DML handlers (bb132b61)

      The router is composed of several DML handlers called in sequence in order to
      construct the full request handling pipeline. Prior to this commit, each
      handler nested the next handler it calls internally, producing a nested call
      chain that resulted metrics (added in #3764) recording cumulative latency like
      this:

                    ┌ ─

                    │     ┌───────────────┐
                         │  NS Creation  │
                   │     └───────────────┘
                                 │  ┌───────────────┐
                   │             │  │  Partitioner  │
                                 │  └───────────────┘
                   │             │          │
                                 │          │
      Cumulative   │             │          │  ┌───────────────┐
        Timings                1.5s        1s  │    etc...     │
                   │             │          │  └───────────────┘
                                 │          │
                   │             │          │
                                 │  ┌───────────────┐
                   │             │  │  Partitioner  │
                                 │  └───────────────┘
                   │     ┌───────────────┐
                         │  NS Creation  │
                   │     └───────────────┘

                    └ ─

      This meant it was hard to determine the latency of a single handler without
      knowing (and subtracting the latency of) all the child handlers it calls.

      This commit replaces the intrusive nested handler call chain with an external
      Chain combinator type to compose together individual handlers, resulting in
      correct per-handler timings and simpler code/tests:

                ┌───────────────┐
               │  NS Creation  │
               └───────────────┘
                       │
                      .5s       ┌───────────────┐
                       └───────▶│  Partitioner  │
                                └───────────────┘
                                        │
                                       1s    ┌───────────────┐
                                        └───▶│    etc...     │
                                             └───────────────┘

* test: router handler stack integration test (497615d7)

      Adds an integration test covering the router's HTTP handler stack.

      Given a well-formed HTTP write, the test asserts:

          * Write passes through the stack without error
         * Response code sent to client
         * Write buffer message is enqueued
         * Catalog namespace record is created
         * Metric handler is invoked and the hit is recorded

### DIFF
--- a/influxdb_iox/src/commands/run/router2.rs
+++ b/influxdb_iox/src/commands/run/router2.rs
@@ -18,8 +18,8 @@ use data_types::database_rules::{PartitionTemplate, TemplatePart};
 use observability_deps::tracing::*;
 use router2::{
     dml_handlers::{
-        InstrumentationDecorator, NamespaceAutocreation, Partitioner, SchemaValidator,
-        ShardedWriteBuffer,
+        DmlHandlerChainExt, FanOutAdaptor, InstrumentationDecorator, NamespaceAutocreation,
+        Partitioner, SchemaValidator, ShardedWriteBuffer,
     },
     namespace_cache::{metrics::InstrumentedCache, MemoryNamespaceCache, ShardedCache},
     sequencer::Sequencer,
@@ -90,15 +90,15 @@ pub async fn command(config: Config) -> Result<()> {
 
     let catalog = config.catalog_dsn.get_catalog("router2").await?;
 
+    // Initialise the sharded write buffer and instrument it with DML handler
+    // metrics.
     let write_buffer = init_write_buffer(
         &config,
         Arc::clone(&metrics),
         common_state.trace_collector(),
     )
     .await?;
-
-    // Wrap the write buffer with metrics
-    let handler_stack =
+    let write_buffer =
         InstrumentationDecorator::new("sharded_write_buffer", Arc::clone(&metrics), write_buffer);
 
     // Initialise an instrumented namespace cache to be shared with the schema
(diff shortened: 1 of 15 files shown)


influxdb_iox/src/commands/run/router2.rs
```diff
@@ -167,17 +164,33 @@ pub async fn command(config: Config) -> Result<()> {
...
+        // This block initialises a fan-out adaptor that parallelises partitioned
+        // writes into the handler chain it decorates (schema validation, and then
+        // into the sharded write buffer), and instruments the parallelised
+        // operation.
+        .and_then(InstrumentationDecorator::new(
+            "parallel_write",
+            Arc::clone(&metrics),
+            FanOutAdaptor::new(schema_validator.and_then(write_buffer)),
+        ));
+
```
domodwyer, line 193: I like how this small block of code shows the modularity of the chained handlers - it now defines the entire request pipeline for the router, specifying the order in which handlers are invoked, where parallisation occurs, and what latency metrics are recorded.

router2/src/dml_handlers/chain.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl<T, U> Chain<T, U> {
+    fn new(first: T, second: U) -> Self {
+        Self { first, second }
+    }
+}
+
+#[async_trait]
+impl<T, U> DmlHandler for Chain<T, U>
```
domodwyer, line 43: This defines the combinator type used to chain together the handlers (the impl behind the `and_then()` call).

lukebond (approved):
love the elegance of the end result! and thanks as always for the great explanation and diagrams in the PR

PR #3790, opened 2022-02-18
feat: add wildcard support to persist partition

tustvold:
I found myself wanting to trigger persistence across all tables for an old partition, just to get it persisted. I originally did this in a bash script, but figured it was generally useful enough to be part of the CLI


lukebond (approved):
nice!

PR #3791, opened 2022-02-18
feat: trigger persistence if over soft limit and no evictable chunks

tustvold:
This tweaks the lifecycle to trigger persistence of the largest partition, if the soft limit has been exceed and no chunks can be evicted from memory. 

This avoids stalling ingest, at the expense of potentially persisting smaller parquet files. However, this will still respect the late arrival window, i.e. it will avoid persisting data that only just arrived, and therefore should still not create parquet files with overlapping time ranges.

The primary motivator for this change is not actually handling misconfigured databases with too little memory allocated to them, although it does help with this, but rather when catching up on ingest from Kafka. In such a scenario ingest may run many multiples faster than normal, potentially ingesting data from multiple hours in minutes. The lifecycle, however, is unaware of this and will wait the normal persist_age_threshold_seconds to persist this data. This has a tendency to result in a database hitting its hard limit and the catchup stalling out, with this change it should instead persist the largest partition and carry on.


influxdb_iox/tests/end_to_end_cases/persistence.rs
```diff
@@ -114,7 +116,7 @@ async fn test_full_lifecycle() {
     wait_for_exact_chunk_states(
         &fixture,
         &db_name,
-        vec![ChunkStorage::ObjectStoreOnly],
+        vec![ChunkStorage::ReadBufferAndObjectStore],
```
tustvold, line 119: This test becomes non-deterministic if the soft limit is set too low, as the lifecycle may now trigger persistence. Raising the soft limit avoids this, but means this no longer tests the unloading of the read buffer. We have other tests of this, so I don't think this is a great loss

e-dard (approved):
This makes sense to me 👍

PR #3802, opened 2022-02-21
feat: preserve catalog on error (#1522)

### LINKED ISSUE #1522: Do not wipe preserved catalog on error
Placeholder.

tustvold:
Closes #1522 

As outlined in the issue, this dates from before we had proper tooling to restore/recover various classes of catalog error. We now do, and so this default can be changed


lukebond (approved):
seems reasonable to me although i don't know the full implications. it was indeed a long time ago

Issue #3777, opened 2022-02-17
new command and crate for compactor


lukebond:
The runner will just take an initialised catalog and initialised object store and then do nothing

PR #3803, opened 2022-02-21
feat: adding compactor CLI command and crate

lukebond:
Skeleton of CLI command and crate for compactor. Mostly copied ingester and router2 here and there. Some TODO's left to indicate where to build out the bits to make it work.

Closes: #3777

### DIFF
--- a/compactor/src/handler.rs
+++ b/compactor/src/handler.rs
@@ -1,0 +1,82 @@
+//! Compactor handler
+
+use async_trait::async_trait;
+use iox_catalog::interface::Catalog;
+use object_store::ObjectStore;
+use observability_deps::tracing::warn;
+use std::{fmt::Formatter, sync::Arc};
+use thiserror::Error;
+use tokio_util::sync::CancellationToken;
+
+#[derive(Debug, Error)]
+#[allow(missing_copy_implementations, missing_docs)]
+pub enum Error {}
+
+/// The [`CompactorHandler`] does nothing at this point
+#[async_trait]
+pub trait CompactorHandler: Send + Sync {
+    /// Wait until the handler finished  to shutdown.
+    ///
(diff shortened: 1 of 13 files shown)


compactor/src/handler.rs
```diff
@@ -0,0 +1,82 @@
...
+use async_trait::async_trait;
+use iox_catalog::interface::Catalog;
+use object_store::ObjectStore;
+use observability_deps::tracing::warn;
+use std::{fmt::Formatter, sync::Arc};
+use thiserror::Error;
+use tokio_util::sync::CancellationToken;
+
+#[derive(Debug, Error)]
+#[allow(missing_copy_implementations, missing_docs)]
```
mkmik, line 12: why `missing_copy_implementations`?

compactor/src/handler.rs
```diff
@@ -0,0 +1,82 @@
...
+pub enum Error {}
+
+/// The [`CompactorHandler`] does nothing at this point
+#[async_trait]
+pub trait CompactorHandler: Send + Sync {
+    /// Wait until the handler finished  to shutdown.
+    ///
+    /// Use [`shutdown`](Self::shutdown) to trigger a shutdown.
+    async fn join(&self);
```
mkmik, line 21: is it necessary that the CompactorHandler is Send+Sync?

what happens when multiple threads call `join` or shutdown?

crepererum: > what happens when multiple threads call join or shutdown?

The semantics are: `shutdown` is sync but doesn't block, it just flags the handler for shutdown. You may call this as often as you want. `join` blocks until the server finished shutdown (you can call this at any point and from as many threads as you wish).

crepererum (approved):

lukebond:
thanks @crepererum for elucidating the stuff i copied over from ingester :laughing: :+1:

PR #3804, opened 2022-02-21
feat: add tracing down to the chunk level

e-dard:
This PR plumbs tracing down to the Chunk layers by plumbing the `IOxExecutionContext` into the `InfluxRPC` planner.

Currently it only provides tracing for `read_filter`. More to be added in subsequent PRs.

Spans in the emitted trace will tell you the following about chunks:

 - predicates applied
 - projected columns
 - storage format (mub, rub, parquet).

The motivation for plumbing `IOxExecutionContext` through everything is that it seems likely it will make a useful way of passing other context through the query engine in the future.


influxdb_iox/src/influxdb_ioxd/planner.rs
```diff
@@ -119,7 +119,7 @@ impl Planner {
     where
         D: QueryDatabase + 'static,
     {
-        let planner = InfluxRpcPlanner::new();
+        let planner = InfluxRpcPlanner::default();
```
tustvold, line 122: Is there a particular reason you opted to not plumb in `self.ctx.child_ctx("influxrpc_planner")` here?

query/src/exec/task.rs
```diff
@@ -69,13 +69,14 @@ impl<T> PinnedDrop for Job<T> {
...
 /// them) on a separate tokio Executor
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct DedicatedExecutor {
     state: Arc<Mutex<State>>,
 }
 
 /// Runs futures (and any `tasks` that are `tokio::task::spawned` by
 /// them) on a separate tokio Executor
+#[derive(Default)]
```
tustvold, line 79: I'm not sure this is correct, I think this will create a sort of zombie `DedicatedExecutor`...

query/src/provider.rs
```diff
@@ -304,17 +313,27 @@ pub(crate) struct Deduplicater<C: QueryChunk + 'static> {
...
             overlapped_chunks_set: vec![],
             in_chunk_duplicates_chunks: vec![],
             no_duplicates_chunks: vec![],
+            ctx,
         }
     }
 
+    #[cfg(test)]
+    // creates a deduplicator with a default executor
+    pub(crate) fn default_context() -> Self {
```
tustvold, line 333: Typically we've opted to put these sorts of methods in a test_util, rather than as static members or member functions.

query/src/provider.rs
```diff
@@ -405,6 +424,7 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
         if self.no_duplicates() {
             // Neither overlaps nor duplicates, no deduplicating needed
             let mut non_duplicate_plans = Self::build_plans_for_non_duplicates_chunks(
+                self.ctx.child_ctx("build_plans_for_non_duplicates_chunks"),
```
tustvold, line 427: :ok_hand:

trace/src/span.rs
```diff
@@ -145,7 +145,7 @@ impl From<i64> for MetaValue {
 ///
 /// If not created with a `Span`, e.g. this request is not being sampled, all operations
 /// called on this `SpanRecorder` will be a no-op
-#[derive(Debug, Default)]
+#[derive(Clone, Debug, Default)]
```
tustvold, line 148: I'm not sure about `Clone` here - you will end up with duplicate spans with the same context?

query/src/exec/context.rs
```diff
@@ -246,6 +249,7 @@ impl IOxExecutionConfig {
 ///
 /// An IOxExecutionContext is created directly from an Executor, or from
 /// an IOxExecutionConfig created by an Executor
+#[derive(Clone, Default)]
```
tustvold, line 252: As mentioned below, I think the `Default` implementation is currently a tad broken - I don't think you can actually run anything on it.

I think it would be better to either, make `DedicatedExecutor` optional, otherwise it runs on the current tokio thread pool. I'm not entirely sure how much churn this would create though, I think it should be manageable.

tustvold (commented):
This mostly looks good to me, its definitely a pain having to plumb it in at this stage, but hopefully its a once and done kind of thing :sweat_smile: 

I do think there is a slight issue with the way the default is implemented, but nothing hugely major here. Nice work :+1:

influxdb_iox/src/influxdb_ioxd/planner.rs
e-dard: I will do these, I was just limiting the spike to `read_filter` initially.

query/src/exec/context.rs
```diff
@@ -549,10 +553,13 @@ impl IOxExecutionContext {
...
-        self.exec
-            .spawn(fut)
-            .await
-            .unwrap_or_else(|e| Err(Error::Execution(format!("Join Error: {}", e))))
+        match &self.exec {
+            Some(exec) => exec
+                .spawn(fut)
+                .await
+                .unwrap_or_else(|e| Err(Error::Execution(format!("Join Error: {}", e)))),
+            None => todo!(), // spawn onto current thread pool
```
tustvold, line 561: TBH this could probably be left unimplemented!() for a while

tustvold (commented):
LGTM :+1:

db/src/lifecycle/write.rs
```diff
@@ -97,8 +97,13 @@ pub(super) fn write_chunk_to_object_store(
             collect_checkpoints(flush_handle.checkpoint(), &db.catalog);
 
         // Get RecordBatchStream of data from the read buffer chunk
+        // TODO(edd): wire up execution context
```
tustvold, line 100: I presume this isn't a problem

db/src/streams.rs
```diff
@@ -37,7 +44,13 @@ impl futures::Stream for ReadFilterResultsStream {
         mut self: std::pin::Pin<&mut Self>,
         _: &mut Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        Poll::Ready(Ok(self.read_results.next()).transpose())
+        let mut ctx = self.ctx.child_ctx("next_row_group");
+        let rb = self.read_results.next();
```
tustvold, line 48: Was going to suggest https://doc.rust-lang.org/std/option/enum.Option.html#method.inspect but it appears to be nightly only

tustvold (approved):

db/src/access.rs
```diff
@@ -368,7 +369,8 @@ impl SchemaProvider for DbSchemaProvider {
             schema
         };
 
-        let mut builder = ProviderBuilder::new(table_name, schema);
+        // TODO(edd): wire up execution provider
```
alamb, line 372: I don't understand this TODO

influxdb_iox/src/influxdb_ioxd/planner.rs
alamb: time now? Or in a follow on PR?

ingester/src/query.rs
```diff
@@ -230,9 +230,12 @@ impl QueryChunk for QueryableBatch {
     /// streams from several different `QueryChunk`s.
     fn read_filter(
         &self,
+        mut ctx: IOxExecutionContext,
         predicate: &Predicate,
         selection: Selection<'_>,
     ) -> Result<SendableRecordBatchStream, Self::Error> {
+        ctx.set_metadata("storage", "ingester");
+        ctx.set_metadata("projection", format!("{}", selection));
```
alamb, line 238: I am surprised clippy didn't get all annoyed and suggest `selection.to_string()` here

query/src/exec/context.rs
```diff
@@ -246,16 +249,17 @@ impl IOxExecutionConfig {
 ///
 /// An IOxExecutionContext is created directly from an Executor, or from
 /// an IOxExecutionConfig created by an Executor
+#[derive(Default)]
 pub struct IOxExecutionContext {
     inner: ExecutionContext,
 
-    /// Dedicated executor for query execution.
+    /// Optional dedicated executor for query execution.
```
alamb, line 256: Under what circumstances don't we have a dedicated executor? Making this optional means there is now some possibility it isn't set when needed

alamb (approved):

influxdb_iox/src/influxdb_ioxd/planner.rs
e-dard: follow on

query/src/exec/context.rs
e-dard: I chatted this through with @tustvold and the idea was that if it's `None` the current thread pool will be used for execution, rather than there being no threadpool available.

PR #3806, opened 2022-02-21
feat: sequencer metrics during replay

tustvold:
Currently there are no sequencer metrics during replay, which is unfortunate as this is often the time when you are most interested in them :laughing:. This PR therefore adds these metrics in.

_The replay code really needs some TLC, it living in Db is a bit of a historical artifact, but hey ho. If it lives for much longer I may spend some time cleaning it up, but not today :smile:_


db/src/write_buffer.rs
```diff
@@ -94,6 +94,53 @@ impl Drop for WriteBufferConsumer {
     }
 }
 
+/// [`WatermarkFetcher`] contains the logic to periodically fetch a high watermark from
+/// a [`WriteBufferReading`]. As this may be an expensive operation, and is solely used
+/// for observability, we only wish to do this periodically
+pub struct WatermarkFetcher {
```
tustvold, line 100: This is the logic in stream_in_sequenced_entries` extracted into a type so that it can be reused.

db/src/write_buffer.rs
```diff
@@ -94,6 +94,53 @@ impl Drop for WriteBufferConsumer {
...
+            // skip over invalid data in the write buffer so recovery can succeed
+            Err(e) => {
+                debug!(
+                    %e,
+                    sequencer_id=self.sequencer_id,
+                    "Error while reading sequencer watermark",
+                )
+            }
+        }
+        self.last_updated = now;
```
tustvold, line 139: The fact an error counts as an update is a little odd, but is what the code previously did and I guess gives retry backoff of a form

db/src/replay.rs
```diff
@@ -223,14 +226,21 @@ pub async fn perform_replay(
...
 
+            let mut watermark_fetcher =
+                WatermarkFetcher::new(Duration::from_secs(10), *sequencer_id);
+            let mut metrics = ingest_metrics.new_sequencer_metrics(*sequencer_id);
             let mut stream = handler.stream().await;
             while let Some(dml_operation) = stream.try_next().await.context(EntrySnafu {
                 sequencer_id: *sequencer_id,
             })? {
+                let watermark = watermark_fetcher.fetch(write_buffer).await;
+                let ingest_recorder = metrics.recorder(watermark).operation(&dml_operation);
```
tustvold, line 237: There is an interesting disparity here, an error in the write buffer is fatal to replay, the `try_next()` above, but not the normal ingestion loop which simply records an error.

tustvold:
Working on adding a test of this

db/src/replay.rs
```diff
@@ -923,6 +936,30 @@ mod tests {
...
+                                }
+                                false => {
+                                    let maybe_observation = maybe_instrument
+                                        .and_then(|x| Some(x.get_observer(attributes)?.fetch()));
+
+                                    matches!(maybe_observation, Some(x) if x == *expected)
+                                }
+                            }
+                        }
+                        _ => unimplemented!(),
```
tustvold, line 961: We could easily add support for different observation types in future, should we so wish

db/src/replay.rs
```diff
@@ -1041,6 +1103,28 @@ mod tests {
...
+                    ),
+                    Check::Metric(
+                        "write_buffer_last_max_ts",
+                        attributes.clone(),
+                        metric::Observation::U64Gauge(0),
+                    ),
+                    Check::Metric(
+                        "write_buffer_sequence_number_lag",
+                        attributes.clone(),
+                        metric::Observation::U64Gauge(2),
```
tustvold, line 1125: Here we can se that replay has updated the metrics

db/src/write_buffer.rs
```diff
@@ -94,6 +94,53 @@ impl Drop for WriteBufferConsumer {
...
+            delay,
+            last_updated: Instant::now() - delay,
+            watermark: 0,
+            sequencer_id,
+        }
+    }
+
+    /// Fetch the current high watermark, or return the last watermark if
+    /// sufficient time has not passed
+    pub async fn fetch(&mut self, write_buffer: &dyn WriteBufferReading) -> u64 {
```
crepererum, line 120: You might also use a tokio task instead of manual updates, but I don't have a too strong opinion on that.

db/src/lib.rs
```diff
@@ -897,13 +898,20 @@ impl Db {
         consumer: Arc<dyn WriteBufferReading>,
     ) -> Result<BTreeMap<u32, Box<dyn WriteBufferStreamHandler>>> {
         use crate::replay::{perform_replay, seek_to_end};
-
+        let ingest_metrics =
+            WriteBufferIngestMetrics::new(&self.metric_registry, self.name.as_ref());
```
crepererum, line 902: Is it OK that to just drop these metrics on the floor after replay is finished and rebuild the metrics update later?

db/src/replay.rs
```diff
@@ -502,6 +513,9 @@ mod tests {
 
         /// Check query results.
         Query(&'static str, Vec<&'static str>),
+
+        /// Check metric
+        Metric(&'static str, metric::Attributes, metric::Observation),
```
crepererum, line 518: Nice framework integration :)

crepererum (approved):

db/src/lib.rs
tustvold: Yeah - the metric registry is global, so it's fine. None of these counters have local state

db/src/write_buffer.rs
tustvold: The real fix is probably to somehow expose the high watermark returned by Fetch, I'll create an rskafka ticket

PR #3811, opened 2022-02-22
refactor: generic+reusable task registry

crepererum:
This adds an interface to the existing task registry (`AbstractTaskRegistry`) and moves the job counter metrics into a reusable interface, so it can easily be integrated into NG as well.


tracker/src/task/metrics.rs
```diff
@@ -0,0 +1,281 @@
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    sync::Arc,
+    time::Duration,
+};
+
+use crate::{AbstractTaskRegistry, TaskId, TaskRegistration, TaskTracker};
+
+/// Function that extracts metric attributes from job metadata.
+pub type FAttributes<T> = Box<dyn Fn(&mut metric::Attributes, &T) + Send>;
```
tustvold, line 10: I wonder if this should instead return `metric::Attributes` rather than modifying an existing one?

I think this can also be `FnMut`

tustvold (approved):
LGTM :+1:

tracker/src/task/metrics.rs
crepererum: done

PR #3819, opened 2022-02-22
refactor: move `DedicatedExecutor` from `query` to `executor`

crepererum:
The threaded executor is not query specific (at least not this
particular layer) and can easily be used by other subsystems. In many
cases, it's likely better than a random `tokio::task::spawn` calls w/o
awaiting results.


tustvold (approved):
Makes sense to me, although at least to me a crate called `tracker` seems like an odd place to find a `DedicatedExecutor`

crepererum:
> Makes sense to me, although at least to me a crate called `tracker` seems like an odd place to find a `DedicatedExecutor`

Created a new crate `executor`.

alamb:
> Created a new crate executor.

👍

PR #3830, opened 2022-02-23
feat: emit trace spans for router stages

domodwyer:
Emit spans for each instrumented router stage, including the error message of any handler failures.

---

* feat: emit trace spans for router stages (b20dce80)

      Configures the instrumentation decorator to emit a trace span covering the
      duration of the decorated handler's execution, recording the success/error
      result and and error message, if any.


tustvold (approved):
LGTM :+1:

FWIW it might be nice to have some tests of this, there should be a fair few examples of using the `RingBufferTraceCollector` for this purpose

Issue #2035, opened 2021-07-16
Verify and Improve if Chunk Resorting happen during scan 


NGA-TRAN:
<!--

Thank you for suggesting an idea to improve InfluxDB IOx. 

Have you read the contributing section of the README? Please do if you haven't.
https://github.com/influxdata/influxdb_iox/blob/main/README.md

* Please ask usage questions in the Influx Slack (there is an #influxdb-iox channel).
    * https://influxdata.com/slack
* If the feature you're interested in already has an open *or* closed ticket associated with it (please search) please don't add "+1" or "me too" comments. The Github reaction emojis are a way to indicate your support for something and will contribute to prioritisation.

-->

# Rationale
Chunk sorting is needed during scan to deduplicate data. However, if the chunk is already sorted on the order we want, we do not need to resort them. Based on our design, we do not expect the sort order get changed so if the chunk needs to get resort, there is something wrong in our design. This ticket is to verify if the resort ever happens and on which data sets for us to improve our design.

Fist, let us dig into our current designs of `compaction` and `scan` where sort and resort (if any) happen.

## Compactions

Compaction happens when we want to compact many chunks (MUBs and RUBs) into one RUB chunk, or when we move a MUB into and RUB. The diagram below shows those 2 processes. They do similar work: scan data of the chunk(s), sort them, and write them into a new RUB chunk. Thus when a RUB is created, it is always sorted. OS chunks are created from corresponding RUB chunks so they are also sorted.

The sort key includes all primary columns (tags and time) of the chunk and the column sort order is computed based on the tag column cardinality or distinct value  (the smaller the better) to have best RLE encoding. Time column is always last in the sort order.

For example, if the primary keys are `tag1, tag2, tag3, time` whose corresponding cardinalities are 100, 30, 80, 50, their sort key will be `tag2, tag3, tag1, time`

                                                                         
```
             CHUNKS COMPACTION                         MUB -> RUB        
                                                                         
                                                                         
                                                                         
              .───────────.                          .───────────.       
             (  RUB chunk  )                        (  RUB chunk  )      
              `───────────'                          `───────────'       
                    ▲                                      ▲             
                    │                                      │             
                    │                                      │             
           ┌─────────────────┐                    ┌─────────────────┐    
           │      Sort       │                    │      Sort       │    
           └─────────────────┘                    └─────────────────┘    
                    ▲                                      ▲             
                    │                                      │             
                    │                                      │             
[...]

PR #3831, opened 2022-02-23
feat: add sort ordinal to chunk_columns system table (#2035)

tustvold:
Adds the sort ordinal to the chunk_columns system table, which will help diagnose sort related inconsistencies.

I originally experimented with adding this to ChunkColumnSummary or similar, however, the sort key is stored as part of the in-memory CatalogChunk's arrow schema, which not all storage's are aware of. The quickest solution was just to expose this.


e-dard (approved):
Awesome 👍

PR #3834, opened 2022-02-23
test: catalog `maybe_skip_integration!` and `should_panic`

crepererum:


domodwyer (approved):
I added this, so sorry if it caused you a headache! Could you explain what's going on / what the change accomplishes?

PR #3835, opened 2022-02-23
feat: `DedicatedExecutor` async `join` and job `detach`.

crepererum:
This brings `DedicatedExecutor` in line with our remaining stack and allows it (w/ some additional code) to act as a foundation for the ingester/compactor task pool.


executor/src/lib.rs
```diff
@@ -55,10 +58,21 @@ pub type Error = tokio::sync::oneshot::error::RecvError;
 #[derive(Debug)]
 pub struct Job<T> {
     cancel: CancellationToken,
+    detached: bool,
     #[pin]
     rx: Receiver<T>,
 }
 
+impl<T> Job<T> {
+    /// Detached job so dropping it does not cancel it.
```
alamb, line 67:
```suggestion
    /// Detach job so dropping it does not cancel it.
```

executor/src/lib.rs
```diff
@@ -463,7 +571,36 @@ mod tests {
...
+        // wait a bit just to make sure that our tasks doesn't get dropped
+        tokio::time::sleep(Duration::from_millis(10)).await;
+        assert_eq!(exec.tasks(), 2);
+
+        // unblock tasks
+        barrier1.wait().await;
+        barrier2.wait().await;
+        wait_for_tasks(&exec, 0).await;
+
+        exec.join().await;
```
alamb, line 603: I may not fully understand the usecase of calling `detach` on the tasks -- but it seems like the goal is to allow the tasks to continue to run even when the `DedicatedExecutor` has been `join()`d.

Thus, I was thinking this test should call `exec.join().await` the exec before unblocking the tasks, rather than after? This would likely require something different than `wait_for_tasks` of course

executor/src/lib.rs
```diff
@@ -158,13 +199,17 @@ impl DedicatedExecutor {
...
+                tx_shutdown.send(()).ok();
             })
         });
 
         let state = State {
-            requests: Some(tx),
-            thread: Some(thread),
+            requests: Some(tx_tasks),
             task_refs: Arc::new(()),
+            completed_shutdown: rx_shutdown.map_err(Arc::new).boxed().shared(),
```
alamb, line 211: TIL: https://docs.rs/futures/0.3.21/futures/future/trait.FutureExt.html#method.shared 👍

alamb (approved):
The changes make sense to me. I had only one substantial question on the `detach` test

executor/src/lib.rs
crepererum: It's not about `join`, but rather what happens when the task handle (`Job`) gets dropped. Normally we would abort the task, but there are good cases where you want a "fire and forget"-like behavior and don't want to come up with some artificial container for the handles.

crepererum: I see how this is confusing. The point of this test was rather about the first task (for which we call `discard` and that function deliberately consumes `self`) and to kinda show that another task can still run. I'll remove the 2nd task since it doesn't really add anything to the test.

alamb: Ah, got it -- sorry I was confused. Maybe a comment that says the task should be running even though it is dropped would also help.

👍

crepererum: I rewrote the test and I thinks it's a bit clearer now what is happening.

alamb:
Updated the branch to use the larger executor from #3885

executor/src/lib.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/3835/commits/f4ed4ae18fa31ac312d7f645fb817c1a4bd18e1f looks much better (thank you for the comment!)

alamb (approved):

PR #3837, opened 2022-02-23
feat: catalog instrumentation

domodwyer:
This PR adds metrics for each method in the catalog so we can see how much we're smashing postgres with the big stick, and how long it takes to do stuff when being hit with said stick 🏏 

---

* feat: catalog instrumentation decorator (a0b43323)

      Adds a MetricDecorator type that wraps a Catalog impl, delegating calls to the
      underlying impl and recording per-method call latency broken down by
      success/error state.

* feat: instrument postgres catalog impl (aaf89519)

      Wraps the postgres implementation of the catalog with a MetricDecorator.

      This is slightly intrusive, with the metrics registry being pushed into the
      PostgresCatalog type in order to decorate the impls returned when calling the
      Catalog::repositories() and Catalog::start_transaction() methods (rather than
      being a pure decorator) in order to use static dispatch and let the compiler
      optimise away as much overhead as possible.


iox_catalog/src/metrics.rs
```diff
@@ -0,0 +1,249 @@
...
+    };
+}
+
+decorate!(
+    impl_trait = KafkaTopicRepo,
+    methods = [
+        "kafka_create_or_get" = create_or_get(&mut self, name: &str) -> Result<KafkaTopic>;
+        "kafka_get_by_name" = get_by_name(&mut self, name: &str) -> Result<Option<KafkaTopic>>;
+    ]
+);
```
domodwyer, line 168: I didn't want the decorator to be a pain in the ass to keep in-sync with the actual catalog impl (and I also didn't want to write out all the decorators by hand) so this macro builds them. If we add more methods or change the signature, we should be able to modify this easily to add the instrumentation.

Realistically this seems like it might be useful to automatically derive instrumented decorators for other traits - I might make this a `#[derive(metric_decorator)]` procedural macro in the future if it comes up again.

crepererum (approved):

PR #3839, opened 2022-02-23
fix: Remove min_time and max_time from IngesterQueryRequest

carols10cents:
As discussed in chat because they're already in the `Predicate`.


NGA-TRAN (approved):
Looks good. Thanks Carol for a quick PR

PR #3843, opened 2022-02-23
feat: skeleton of querier CLI

### LINKED ISSUE #3840: add querier CLI
- as run subcommand
- long-running server with API

can be blank and no nothing for now, ready to flesh out as the crate grows

lukebond:
Closes #3840

Basically a copy-pasta of the compactor CLI and server/handler. Does nothing but has the skeleton in place to build it out.


influxdb_iox/src/commands/run/querier.rs
```diff
@@ -0,0 +1,75 @@
...
+    #[clap(flatten)]
+    pub(crate) catalog_dsn: CatalogDsnConfig,
+}
+
+pub async fn command(config: Config) -> Result<(), Error> {
+    let common_state = CommonServerState::from_config(config.run_config.clone())?;
+
+    let catalog = config.catalog_dsn.get_catalog("querier").await?;
+
+    let metric_registry: Arc<metric::Registry> = Default::default();
```
domodwyer, line 63: I just changed this in #3837 - sorry!

```suggestion
    let metric_registry: Arc<metric::Registry> = Default::default();

    let catalog = config.catalog_dsn.get_catalog("querier", Arc::clone(&metric_registry)).await?;
```

querier/src/handler.rs
```diff
@@ -0,0 +1,69 @@
...
+/// Implementation of the `QuerierHandler` trait (that currently does nothing)
+pub struct QuerierHandlerImpl {
+    /// The global catalog for schema, parquet files and tombstones
+    catalog: Arc<dyn Catalog>,
+    /// A token that is used to trigger shutdown of the background worker
+    shutdown: CancellationToken,
+}
+
+// LB: copied this over from the ingester handler; doesn't seem to be needed there either
+impl std::fmt::Debug for QuerierHandlerImpl {
```
domodwyer, line 35: How come we can't `#[derive(Debug)]` this impl?

domodwyer (approved):
Skeletons! ☠️

querier/src/handler.rs
crepererum: Probably because the handler contained types that don't implement `Debug` themselves. If that's not the case here you can just use `derive`.

querier/src/handler.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        Self { catalog, shutdown }
+    }
+}
+
+#[async_trait]
+impl QuerierHandler for QuerierHandlerImpl {
+    async fn join(&self) {
+        // join querier threads here
+        todo!();
```
crepererum, line 54: Did you test the CLI? Because I'm pretty sure it more or less will crash immediately due to this `todo!()`.

querier/src/server/http.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Initialise a new [`HttpDelegate`] passing valid requests to the
+    /// specified `querier_handler`.
+    pub fn new(querier_handler: Arc<I>) -> Self {
+        Self { querier_handler }
+    }
+
+    /// Routes `req` to the appropriate handler, if any, returning the handler
+    /// response.
+    pub fn route(&self, _req: Request<Body>) -> Result<Response<Body>, Error> {
+        unimplemented!()
```
crepererum, line 48: Maybe return `NotFound` instead of panicking.

influxdb_iox/src/influxdb_ioxd/server_type/querier.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// Provide a placeholder gRPC service.
+    async fn server_grpc(self: Arc<Self>, builder_input: RpcBuilderInput) -> Result<(), RpcError> {
+        let builder = setup_builder!(builder_input, self);
+        // TODO add a service here
+        serve_builder!(builder);
+
+        Ok(())
+    }
```
crepererum, line 62: So either you keep the delegates and use them here or your remove them, but having them and not wiring them up sounds like some hours of debugging later. You're call.

crepererum (approved):
Small nits.

querier/src/handler.rs
lukebond: on starting the server has no service added so immediately stops and then calls this fn and panics. once it has a service it should stay up, and then panic on stop unless this is implemented. do i have it wrong?

crepererum: Just `join` on the shutdown token, this should work for now. It doesn't DO anything for real, but at least it behaves like a normal service (e.g. answers `/health` requests). That would allow me to already integrate it into the test bench. Implementing a real service will take a few more days.

lukebond: :+1:  thanks

lukebond: okay, so i thought it was stopping immediately just because there was no grpc service registered, but in fact it was something i'd done wrong around the join and shutdown functions. i've fixed it now and i will also go back and fix it in the compactor server which does the same thing.

influxdb_iox/src/influxdb_ioxd/server_type/querier.rs
lukebond: removed the delegates that weren't wired up

PR #3844, opened 2022-02-23
fix: chunk overlap missing columns

tustvold:
The overlap check logic doesn't work correctly in the event of missing columns, leading the deduplicator to produce larger groups of chunks to deduplicate than necessary. This fixes the logic.

Tagging https://github.com/influxdata/conductor/issues/904


query/src/provider/overlap.rs
```diff
@@ -189,52 +190,44 @@ where
...
-                    // name didn't match, so try and find the next
-                    // place it does.  Since there may be missing keys
-                    // in each side, need to check each in turn
-                    //
-                    // Note this will result in O(num_tags) stack
-                    // frames in the worst case, but we expect the
-                    // number of tags to be relatively small (~20 at
-                    // the time of this writing)
-                    return Ok(self.potential_overlap_impl(self_idx + 1, other, other_idx)?
-                        || self.potential_overlap_impl(self_idx, other, other_idx + 1)?);
```
tustvold, line 231: This `||` is the source of the problem, it will only say that the chunks don't overlap if both paths return `false`. 

Consider the scenario of `s1.name == "time"`, in such a case the left branch will always evaluate to `Ok(true)` as there are no more columns on that side. The result is that the time column is ignored for determining overlap :scream:

e-dard: oh boy!

e-dard (approved):

PR #3847, opened 2022-02-23
feat: Add row count to the parquet_file record in the catalog

### LINKED ISSUE #3842: Add row_count to parquet_file
The ingester should include `row_count` as part of the summary information in the `parquet_file` record in the catalog. This will be useful for quick debugging and likely for driving compaction behavior.

carols10cents:
Fixes #3842.


iox_catalog/migrations/20220223192306_add_row_count_to_parquet_file.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE
+  IF EXISTS parquet_file
```
mkmik, line 2: Why do we need "if exists" here?

mkmik (approved):
LGTM. Minor question

iox_catalog/migrations/20220223192306_add_row_count_to_parquet_file.sql
carols10cents: So that the migration won't fail if the `parquet_file` table doesn't exist. Not sure how likely this is, but [this existing migration](https://github.com/influxdata/influxdb_iox/blob/main/iox_catalog/migrations/20220221145409_routing_limits.sql) had it and I don't think it would hurt anything. 🤷🏻‍♀️

pauldix (approved):

iox_catalog/migrations/20220223192306_add_row_count_to_parquet_file.sql
mkmik: The table must exist otherwise something is wrong with the order with which migrations are applied.
If that happens I prefer an error message to silently skipping a file.

(not a big deal though)

crepererum (approved):

iox_catalog/migrations/20220223192306_add_row_count_to_parquet_file.sql
NGA-TRAN: Since we are in early stage, this will mostly affect our current test data. I think for production migration, we have to use different way such as bulk load/ingest Paul mentioned yesterday

NGA-TRAN (approved):
LGTM

Issue #3783, opened 2022-02-17
Ingester: ingester should ignore replayed data that has already been persisted


pauldix:
Once #3702 is fixed, the ingester should be updated so that it ignores any replayed data that has already been persisted. This will require IngesteData to keep track of the last persisted sequence number for each table, which will also be needed for the querier API on the ingester.

cc @NGA-TRAN

PR #3848, opened 2022-02-23
feat: add method to get table persistence information from catalog

pauldix:
This is required to finish up with #3783


iox_catalog/src/interface.rs
```diff
@@ -464,6 +464,27 @@ pub trait TableRepo: Send + Sync {
 
     /// Lists all tables in the catalog for the given namespace id.
     async fn list_by_namespace_id(&mut self, namespace_id: NamespaceId) -> Result<Vec<Table>>;
+
+    /// Gets the table persistence info for the given sequencer
+    async fn get_table_persist_info(
+        &mut self,
+        sequencer_id: SequencerId,
+        namespace_id: NamespaceId,
+        table_name: &str,
```
crepererum, line 473: Why are we not using the table ID here?

domodwyer (approved):

iox_catalog/src/interface.rs
pauldix: Because I don't have it at the point where it comes up. This is the first lookup of the table in the ingester when the data comes through in a mutable batch. If we decide to refactor the mutable batches to use IDs rather than strings for namespace, table name and column name, there's a ton of cleanup we can do on the ingester side.

PR #3852, opened 2022-02-24
feat: extend catalog so we can recover `ParquetChunk`s from it

crepererum:
There are a few bits missing to be able to successfully create a `ParquetChunk` based on the current catalog design. This PR adds the catalog-specific functionality.


iox_catalog/src/postgres.rs
```diff
@@ -930,11 +960,13 @@ impl ParquetFileRepo for PostgresTxn {
         max_sequence_number: SequenceNumber,
         min_time: Timestamp,
         max_time: Timestamp,
+        file_size_bytes: i64,
+        parquet_metadata: Vec<u8>,
```
domodwyer, line 964: `i64` I assume because PG dosn't support `usize`? Perhaps we can accept & convert from a `usize` within the impl and hide this unfrotunate impl detail.

It might be overkill, but some sort of `EncodedMetadata` kind of wrapper instead of bare `Vec<u8>` to get the compiler to check the right things are going to the right places would be cool - it could use [`#[sqlx(transparent)]`](https://github.com/influxdata/influxdb_iox/blob/c8b964ebefc35f8a54f89985a9a0074540884bb8/iox_catalog/src/interface.rs#L84-L850) so the wrapper type is responsible for the DB glue too!

domodwyer (approved):

iox_catalog/src/postgres.rs
crepererum: will do that in a follow-up PR.

crepererum: will do that in a follow-up PR.

NGA-TRAN (commented):
LGTM

PR #3853, opened 2022-02-24
feat: wire up `ParquetFilePath` for NG

crepererum:
It's a bit of a duck-type hack, but if we wanna just `ParquetFileChunk`
in the new architecture, we somehow need it to accept new-gen paths.
Also path handling should be somewhat centralized since
ingester/compactor/querier all need to construct them. So having a
`ParquetFilePath` that supports both path styles seems to be a
not-to-bad solution. This should obviously be cleaned up in some
not-to-distant future.


NGA-TRAN (approved):
Makes sense to me. I think the way to use `variant`'s `Old` and `New`, and `new_old_gen` and `new_new_gen` to distinguish the 2 cases and let us to reuse the same crate is good. 

My minor comment is we do plan to remove the old ones after the new ones are in use, so the `new` word does not mean anything at that time anymore. Maybe use something like `ng` instead. But this is very minor and can be replaced anytime in the future.

crepererum:
> My minor comment is we do plan to remove the old ones after the new ones are in use, so the `new` word does not mean anything at that time anymore. Maybe use something like `ng` instead. But this is very minor and can be replaced anytime in the future.

Same goes for NG. Once we're done, we need to remove old/new/ng/2 from the code base entirely.

PR #3854, opened 2022-02-24
feat: namespace cache table/column count metrics

domodwyer:
Records the table and column counts across all namespaces in the namespace cache.

---

* feat: inc/dec gauge metrics (77f64921)

      Adds methods to increment, or decrement gauge metrics.

* refactor: borrow metric registry (4024e95c)

      There's no need for the namespace metrics to take (shared) ownership of the
      metric registry, so lend it at the call site instead of cloning the arc.

* feat: instrument namespace cache contents (0ddc35ce)

      Adds two new metrics:

         * namespace_cache_table_count: total number of tables in cache
         * namespace_cache_column_count: total number of columns in cache

      The metric decorator keeps a running total of each of the table and column
      counts as namespaces are inserted into the cache, and adjusts the value
      accordingly when an existing namespace is overwrote.


router2/src/namespace_cache/metrics.rs
```diff
@@ -70,15 +89,210 @@ where
...
+        assert_eq!(
+            cache.put_insert_counter.observe(),
+            Observation::U64Counter(1)
+        ); // Unchanged
+        assert_eq!(
+            cache.put_update_counter.observe(),
+            Observation::U64Counter(4)
+        );
+        assert_eq!(cache.table_count.observe(), Observation::U64Gauge(2));
+        assert_eq!(cache.column_count.observe(), Observation::U64Gauge(7));
```
crepererum, line 268: would be nice to have a test that also uses a different namespace. Otherwise it seems that `table_count`/`column_count` are reset on each `put_schema` call.

crepererum (approved):
Looks good, small test nitpick.

router2/src/namespace_cache/metrics.rs
domodwyer: Good shout 👍

PR #3857, opened 2022-02-24
feat: create `ParquetChunk` in querier

crepererum:
Adds a small adapter that is able to produce `ParquetChunk`s for NG.

The interface is kinda unfinished and the test is mostly a smoke test, but after I set up some more wiring (e.g. create a proper catalog chunk and insert it into the catalog), we can have some query tests.


NGA-TRAN (approved):
The ParquetChunkAdapter looks good. You have created the test chunk in this PR. I can see you will query it in your next PR :) 👍

Issue #3845, opened 2022-02-23
Chunk Sort Order Lost on Restart


tustvold:
**Problem**

Create a database and persist some data

```
$ cargo run -- database create org_bucket
$ cargo run -- database write org_bucket ~/Downloads/data.lp
$ cargo run -- database partition persist org_bucket "2021-04-26 13:00:00" --all-tables --force
```

Data has a sort ordinal

```
org_bucket> select * from system.chunk_columns;
+---------------------+--------------------------------------+------------+-------------+-----------------+-----------+------------+---------------------+---------------------+--------------+-------------+--------------+
| partition_key       | chunk_id                             | table_name | column_name | storage         | row_count | null_count | min_value           | max_value           | memory_bytes | chunk_order | sort_ordinal |
+---------------------+--------------------------------------+------------+-------------+-----------------+-----------+------------+---------------------+---------------------+--------------+-------------+--------------+
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | tag0        | ObjectStoreOnly | 10000     | 0          | value0              | value1              |              | 1           | 1            |
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | tag1        | ObjectStoreOnly | 10000     | 0          | value0              | value9              |              | 1           | 2            |
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | tag2        | ObjectStoreOnly | 10000     | 0          | value0              | value9              |              | 1           | 3            |
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | tag3        | ObjectStoreOnly | 10000     | 0          | value0              | value9              |              | 1           | 4            |
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | tag4        | ObjectStoreOnly | 10000     | 0          | value0              | value0              |              | 1           | 0            |
| 2021-04-26 13:00:00 | b257e6ec-dfa3-4690-a243-4cf0b852a41a | m0         | time        | ObjectStoreOnly | 10000     | 0          | 1619444823731174000 | 1619444859727574000 |              | 1           | 5            |
+---------------------+--------------------------------------+------------+-------------+-----------------+-----------+------------+---------------------+---------------------+--------------+-------------+--------------+
```

[...]

PR #3858, opened 2022-02-24
refactor: use ChunkMetadata in catalog interface (#3845)

tustvold:
As part of #3845 I want to add more catalog metadata. Currently these are plumbed in as a wall of arguments to the various catalog functions. This PR refactors this code to instead take the `ChunkMetadata` payload, which I subjectively think is cleaner.


db/src/catalog/chunk.rs
```diff
@@ -558,13 +560,25 @@ impl CatalogChunk {
...
+    /// # Panic
+    ///
+    /// Panics if this chunk is not mutable
     pub fn record_write(&mut self) {
-        self.access_recorder.record_access();
-
-        let now = self.time_provider.now();
-        self.time_of_first_write = self.time_of_first_write.min(now);
-        self.time_of_last_write = self.time_of_last_write.max(now);
+        match &mut self.stage {
```
tustvold, line 568: As an added bonus you can't record a write on an immutable chunk :sweat_smile:

db/src/catalog/chunk.rs
```diff
@@ -558,13 +560,25 @@ impl CatalogChunk {
...
+            ChunkStage::Open {
+                time_of_first_write,
+                time_of_last_write,
+                ..
+            } => {
+                let now = self.time_provider.now();
+                *time_of_first_write = now.min(*time_of_first_write);
+                *time_of_last_write = now.max(*time_of_last_write);
+            }
+            _ => unreachable!("chunk not mutable"),
```
e-dard, line 578: nit: sometimes it's good to enumerate all the states and panic for each one, in case we decide to add a new mutable stage in the future.

e-dard (approved):
Code looks good. Have fun trying to get it merged today 😂

Issue #3859, opened 2022-02-24
iox_catalog(mem): deadlock acquiring multiple RepoCollection 


domodwyer:
__Steps to reproduce:__
Attempt to call `repositories()` twice to acquire two handles to a `RepoCollection` impl:

```diff
diff --git a/iox_catalog/src/interface.rs b/iox_catalog/src/interface.rs
index 26fa67d0..7effcd63 100644
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -1062,6 +1062,9 @@ pub(crate) mod test_helpers {
     use std::{ops::DerefMut, sync::Arc, time::Duration};

     pub(crate) async fn test_catalog(catalog: Arc<dyn Catalog>) {
+        let mut _c1 = catalog.repositories().await;
+        let mut _c2 = catalog.repositories().await;
+
         test_setup(Arc::clone(&catalog)).await;
         test_kafka_topic(Arc::clone(&catalog)).await;
         test_query_pool(Arc::clone(&catalog)).await;
```

This is a behavioural difference with the postgres impl, preventing clean substitution - code may work fine with PG but deadlock using the in-mem catalog.

crepererum:
I thought I documented this, but apparently that's not the case. You shouldn't keep the repo collection around. It should rather be used as `catalog.repositories().await.foo.add_or_create(...).await`. It's somewhat expected behavior to simplify the in-mem implementation.

PR #3860, opened 2022-02-24
refactor: parallel column resolution

domodwyer:
Make things faster the easy way - by hiding latency with parallelism!

While adding tests to this I noticed the in-mem catalog behaves differently than the postgres impl, and deadlocks when attempting to get two handles to the `RepoCollection` (https://github.com/influxdata/influxdb_iox/issues/3859).

---

* refactor: parallel column resolution (b07f15be)

      A quick change to perform the ColumnRepo::create_or_get() calls in parallel
      (up to a maximum of 3 in-flight at any one time) in order to mitigate the
      latency of the call and reduce the overall schema validation call duration.

      The in-flight limit is enforced to avoid starving the DB connection pool of
      connections.

### DIFF
--- a/ingester/src/data.rs
+++ b/ingester/src/data.rs
@@ -1141,7 +1141,6 @@ mod tests {
     use mutable_batch_lp::lines_to_batches;
     use mutable_batch_lp::test_helpers::lp_to_mutable_batch;
     use object_store::ObjectStoreApi;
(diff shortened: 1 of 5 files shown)


NGA-TRAN (approved):
I think this is a nice optimization and hope you will get the performance numbers you estimate to have

PR #3861, opened 2022-02-24
fix: while executing deduplication, do not return empty record batches

### LINKED ISSUE #3782: Ingester: compact creates extra record batch
The `compact` function in the ingester produces an extra record batch for some reason. If you run this with an input record batch that has only a single row, it will output two record batches, one with the row and one empty. If you run it with an input record batch of two rows, it will split it into two record batches (one row each). If you run it with input of 3 rows, it will split into 2 record batches (one with 2 and one with 1). It should just be returning a single record batch if it's under whatever the batch size is.

NGA-TRAN:
Closes #3782 


ingester/src/compact.rs
```diff
@@ -354,6 +355,43 @@ mod tests {
...
+
+        // compact
+        let exc = Executor::new(1);
+        let stream = compact(&exc, compact_batch).await.unwrap();
+        let output_batches = datafusion::physical_plan::common::collect(stream)
+            .await
+            .unwrap();
+
+        // verify no empty record batches - bug #3782
+        assert_eq!(output_batches.len(), 1);
```
NGA-TRAN, line 382: This test has only one row so verify here only one batch is returned

ingester/src/compact.rs
```diff
@@ -377,6 +415,10 @@ mod tests {
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
+        // verify no empty record bacthes - bug #3782
+        assert_eq!(output_batches.len(), 2);
+        assert_eq!(output_batches[0].num_rows(), 1);
+        assert_eq!(output_batches[1].num_rows(), 1);
```
NGA-TRAN, line 421: This test has 2 returned rows and the deduplicate always return 2 batches. See #3782 for the reason

ingester/src/compact.rs
```diff
@@ -411,6 +453,10 @@ mod tests {
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
+        // verify no empty record bacthes - bug #3782
+        assert_eq!(output_batches.len(), 2);
+        assert_eq!(output_batches[0].num_rows(), 6);
+        assert_eq!(output_batches[1].num_rows(), 1);
```
NGA-TRAN, line 459: Same here: 2 batches

tustvold (approved):

PR #3862, opened 2022-02-25
feat: create CatalogChunk in querier

crepererum:
Now we have catalog chunks (w/o delete predicates), next thing will be a whole catalog.


querier/src/chunk.rs
```diff
@@ -25,11 +58,20 @@ pub struct ParquetChunkAdapter {
...
-    partition_keys: RwLock<HashMap<PartitionId, Arc<str>>>,
+    partition_cache: RwLock<HashMap<PartitionId, Arc<str>>>,
 
-    /// Table name cache.
-    table_names: RwLock<HashMap<TableId, Arc<str>>>,
+    /// Table name and namespace cache.
+    table_cache: RwLock<HashMap<TableId, (Arc<str>, NamespaceId)>>,
+
+    /// Namespace name cache.
+    namespace_cache: RwLock<HashMap<NamespaceId, Arc<str>>>,
```
NGA-TRAN, line 74: Similar to #3865, I think we need Sequencer information here to make sure to get all or appropriate data this Querier is responsible to.

NGA-TRAN (commented):
We need to map Sequencers to Queriers to make sure we read the appropriate data. It is fine if you do it in the following PR but we may want to think about the design first

PR #3863, opened 2022-02-25
refactor: configurable max catalog connections

domodwyer:
Quick addition to the shared catalog config block.

---

* refactor: configurable max catalog connections (35792526)

      Allow the maximum number of catalog (postgres) connections to be specified as
      part of the catalog configuration.


e-dard (approved):
✅

Issue #3214, opened 2021-11-25
ChunkTableProvider Set Sort Key


tustvold:
The Deduplicator and by extension ChunkTableProvider have functionality to produce data that is sorted. Unfortunately, they do not produce RecordBatch with a schema with this set. This is because the schema is provided by the caller, but the sort key is determined by the Deduplicator. This should be fixed so that the caller can request a sort order if desired, and have this respected by the Deduplicator and included in the output record batches

PR #3864, opened 2022-02-25
feat: correctly sort data and store in catalog metadata

tustvold:
The current code has a couple of problems:

* Sort order is stored as part of a chunk's schema, and this differs from the underlying data
* Sort order is lost on restart (#3845)
* ChunkTableProvider ignores the sort order specified by ReorgPlanner (#3214)

This PR addresses all of these


query/src/frontend/reorg.rs
```diff
@@ -453,14 +424,14 @@ mod test {
             "+-----------+------------+------+--------------------------------+",
             "| field_int | field_int2 | tag1 | time                           |",
             "+-----------+------------+------+--------------------------------+",
-            "| 100       |            | AL   | 1970-01-01T00:00:00.000000050Z |",
-            "| 70        |            | CT   | 1970-01-01T00:00:00.000000100Z |",
+            "| 1000      | 1000       | WA   | 1970-01-01T00:00:00.000028Z    |",
```
tustvold, line 427: This is actually a pretty serious bug, that is a consequence of #3214 - `ReorgPlanner` requests a sort order, but the data wasn't actually sorted with respect to this :scream:

query_tests/src/table_schema.rs
```diff
@@ -200,197 +209,7 @@ async fn list_schema_location_all() {
         Selection::All,
         "restaurant",
         expected_schema,
+        None,
     )
     .await;
 }
-
```
tustvold, line 206: These tests no longer made sense as the sort key is not part of the schema

schema/src/sort.rs
```diff
@@ -68,12 +69,33 @@ impl std::fmt::Display for ColumnSort {
     }
 }
 
-#[derive(Debug, Default, Eq, PartialEq, Clone)]
-pub struct SortKey<'a> {
-    columns: IndexMap<&'a str, SortOptions>,
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct SortKey {
```
tustvold, line 73: `SortKey` is changed to an owning type, as it needs to be stored as part of a chunk's metadata. As an added bonus it also makes is easier to work with

db/src/catalog/chunk.rs
```diff
@@ -88,6 +88,9 @@ pub struct ChunkMetadata {
     /// into IOx. Note due to the compaction, etc... this may not be the chunk
     /// that data was originally written into
     pub time_of_last_write: Time,
+
+    /// Sort key for this chunk
+    pub sort_key: Option<SortKey>,
```
tustvold, line 93: I experimented with using `SortKey::default()` in place of `None`, but this just let to confusing bugs. `Option` forces code paths to explicitly handle the case of no sort key.

query/src/provider.rs
```diff
@@ -554,7 +561,7 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
         output_schema: Arc<Schema>,
         chunks: Vec<Arc<C>>, // These chunks are identified overlapped
         predicate: Predicate,
-        output_sort_key: &SortKey<'_>,
```
tustvold, line 557: The deduplicate associated functions are altered to take the sort key that they should use, this ensures that a consistent sort key is used across the groups

schema/src/sort.rs
```diff
@@ -68,22 +69,79 @@ impl std::fmt::Display for ColumnSort {
     }
 }
 
-#[derive(Debug, Default, Eq, PartialEq, Clone)]
-pub struct SortKey<'a> {
-    columns: IndexMap<&'a str, SortOptions>,
+#[derive(Debug, Default)]
+pub struct SortKeyBuilder {
```
tustvold, line 73: I extracted a builder so that SortKey can be immutable and therefore use `Arc` for cheaper clones

query/src/provider.rs
```diff
@@ -1341,26 +1295,9 @@ mod test {
         let chunk1 = Arc::new(
             TestChunk::new("t")
                 .with_id(1)
-                .with_time_column_with_full_stats(
```
tustvold, line 1344: These tests don't actually care about statistics

query/src/provider.rs
```diff
@@ -381,22 +373,20 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
...
         chunks: Vec<Arc<C>>,
         predicate: Predicate,
-        sort_output: bool,
+        output_sort_key: Option<SortKey>,
     ) -> Result<Arc<dyn ExecutionPlan>> {
-        // Initialize an empty sort key
-        let mut output_sort_key = SortKey::with_capacity(0);
-        if sort_output {
-            // Compute the output sort key which is the super key of chunks' keys base on their data cardinality
-            output_sort_key = compute_sort_key_for_chunks(&output_schema, chunks.as_ref());
```
tustvold, line 397: This will produce an incorrect sort key if output_schema has a selection, and the chunks have no statistics. This is now fixed, as it passes in the primary key schema, not the output_schema

query/src/provider.rs
```diff
@@ -649,42 +650,26 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
...
         let plan = Self::build_sort_plan_for_read_filter(
             table_name,
             Arc::clone(&input_schema),
             Arc::clone(&chunks[0]),
             predicate,
-            &sort_key,
+            Some(sort_key),
         )?;
 
-        // The sort key of this chunk might only the subset of the super sort key
```
tustvold, line 675: We no longer need to select the sort key, as arrow_sort_key_exprs contains the logic to skip columns that aren't found

db/src/lib.rs
```diff
@@ -2128,7 +2128,7 @@ mod tests {
         // Read buffer + Parquet chunk size
         catalog_chunk_size_bytes_metric_eq(registry, "mutable_buffer", 0);
         catalog_chunk_size_bytes_metric_eq(registry, "read_buffer", 1700);
-        catalog_chunk_size_bytes_metric_eq(registry, "object_store", 1248);
+        catalog_chunk_size_bytes_metric_eq(registry, "object_store", 1259);
```
e-dard, line 2131: I used to get annoyed having to update these numbers all the time, but now I'm sort of fond of being able to see the cost each change has on the size.

query/src/provider.rs
```diff
@@ -785,32 +770,42 @@ impl<C: QueryChunk + 'static> Deduplicater<C> {
...
-        if !output_sort_key.is_empty() {
-            // build schema of PK
-            let pk_schema = Self::compute_pk_schema(&[Arc::clone(&chunk)]);
-            // Merge it in the output_schema
-            input_schema = Self::compute_input_schema(&input_schema, &pk_schema);
+        let mut schema_merger = SchemaMerger::new().merge(&output_schema).unwrap();
+
+        let chunk_schema = chunk.schema();
+
+        // Cols of sort key
```
e-dard, line 781: I think that what's happening here is we are injecting into the schema given to the read filter node any columns present in both the chunk and the sort key. I think it might be worth explaining a bit why this has to happen. I'm also a little bit unsure of how this always guarantees the resulting sort key has the same sort as the data.

e-dard (approved):
I went through this pretty carefully and it LGTM. Nice reduction in code too!

query/src/provider.rs
tustvold: I've added some more docs, the schema handling in this is definitely a little suspect - I've preserved what the existing logic does, but will likely follow up with something further in this space

alamb (commented):
I can't say I reviewed every line of this PR carefully but what I read makes sense to me.  Nice work

FYI @NGA-TRAN  in case you are interested -- this PR will store the sort key directly in the catalog  (as fields on `Chunk`) rather than in the schema.

NGA-TRAN:
It makes sense to store sort key in the chunk and its corresponding parquet file to retrieve back. 
We may need to do the same for the NG parquet file. CC @pauldix @crepererum and @carols10cents 
FYI: The current parquet file is sorted but on lexicographical tag names and then time. Whether it is optimal for our new design is a different topic. 
Let us discuss this "storing sort key for NG batch & parquet file" incur discussion today and we will open a ticket for NG work accordingly

alamb:
I think storing the sort key in the chunk / parquet file too -- I don't have a particularly strong opinion if the sort key needs to be attached to the arrow `Schema` as metadata or as our own IOx specific metadata added to the file

PR #3865, opened 2022-02-25
feat: sync namespaces in querier

crepererum:
With this PR and #3862 we have somewhat both ends of the sync pipeline, being able to create namespaces and chunks, but the catalog still needs to be filled with tables and the created chunks.


querier/src/database.rs
```diff
@@ -0,0 +1,211 @@
...
+    backoff_config: BackoffConfig,
+
+    /// Catalog.
+    catalog: Arc<dyn Catalog>,
+
+    /// Metric registry
+    metric_registry: Arc<metric::Registry>,
+
+    /// Namespaces.
+    namespaces: RwLock<HashMap<Arc<str>, Arc<QuerierNamespace>>>,
```
NGA-TRAN, line 34: Since data of a Namespace may be split into many Sequencers/Shards, we need a way to map a Querier (or a QuerierDatabase) to corresponding Sequencers. We may need to think a bit about the design here. 

In the design of the `Ingester`, [IngesterData](https://github.com/influxdata/influxdb_iox/blob/c965221df14fc9a9157d135239e30057148f159c/ingester/src/data.rs#L93) includes a list of SequencerData; then each [SequencerData](https://github.com/influxdata/influxdb_iox/blob/c965221df14fc9a9157d135239e30057148f159c/ingester/src/data.rs#L307) contains a list of [NamespaceData](https://github.com/influxdata/influxdb_iox/blob/c965221df14fc9a9157d135239e30057148f159c/ingester/src/data.rs#L93).

We do not need to do the same in the Querier but we may need to think to keep them the same/in sync if possible.

@pauldix may have thought about this already. We may keep a list of namespaces separately to know quickly which db/namespace this Querier supports but, in parallel, we many need to have the list of Sequencers to get data for a specific Namesapce

NGA-TRAN (commented):
The PR does it intends to do which is syncing namespaces in querier 🎉 . I think we may need to see if we need to sync Sequencer and Namespace at the same time because the Querier won't work correctly without syncing both of them.

querier/src/database.rs
crepererum: I we use the old query infrastructure for the MVP, then we have to have a single in-mem catalog per namespace. To see how can still split per-sequencer data have a look at #3862: The old-gen partitions do NOT map directly to NG partitions but rather to `sequencer X partition`.

NGA-TRAN (approved):
Approve this for @crepererum to move forward with our discussion on slack

PR #3874, opened 2022-02-28
feat: allow persisting all partitions and/or all tables from CLI

tustvold:
A follow on from #3790 that also allows persisting all partitions within a table, or even all partitions and all tables.

It also allows continuing on error, which can be useful when running without `--force` as a partition might contain unpersisted data that hasn't been present in memory for long enough to be persisted.


influxdb_iox/src/commands/database/partition.rs
```diff
@@ -255,38 +268,70 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
...
+            };
+
+            let table_filter = match (table_name, all_tables) {
+                (Some(table_name), false) => Some(table_name),
+                (None, true) => None,
+                _ => return Err(Error::MissingTableName),
+            };
+
+            let partition_tables: HashSet<_> = client
+                .list_chunks(&db_name)
```
tustvold, line 290: It is theoretically possible that this might be a rather large list - although I suspect we may have other problems before then :sweat_smile:

influxdb_iox/src/commands/database/partition.rs
```diff
@@ -61,9 +64,11 @@ struct Persist {
     db_name: String,
 
     /// The partition key
-    partition_key: String,
+    #[clap(long)]
+    partition_key: Option<String>,
```
tustvold, line 68: These need to be changed to be arguments as we have two optional arguments now

influxdb_iox/src/commands/database/partition.rs
alamb: What would the problem be with the large list? That it will take a long time to fetch? You could always special case and use `client.list_partition_chunks()` if `partition_filter.is_some()` if worried about it

influxdb_iox/src/commands/database/partition.rs
```diff
@@ -255,38 +271,70 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
...
+                .into_iter()
+                .filter(|chunk| {
+                    let partition_mismatch =
+                        matches!(&partition_filter, Some(x) if &chunk.partition_key != x);
+
+                    let table_mismatch = matches!(&table_filter, Some(x) if &chunk.table_name != x);
+
+                    let already_persisted = ChunkStorage::try_from(chunk.storage())
+                        .map(|s| s.has_object_store())
+                        .unwrap_or(false);
```
alamb, line 304: Should this raise an error if the returned storage is not known? It would evidence a mismatch between server and client I think

alamb (approved):
I also tried it out locally, and it worked well 👍 

```shell
alamb@MacBook-Pro-2 influxdb_iox % ./target/debug/influxdb_iox database partition list 26f7e5a4b7be365b_917b97a92e883afc
./target/debug/influxdb_iox database partition list 26f7e5a4b7be365b_917b97a92e883afc
[
  "2022-03-01 21:00:00",
  "2022-03-01 20:00:00",
  "2022-03-02 15:00:00",
  "2022-03-01 21:00:00",
  "2022-03-01 20:00:00",
  "2022-03-02 15:00:00",
  "2022-03-02 15:00:00",
  "2022-03-01 20:00:00",
  "2022-03-01 21:00:00",
  "2022-03-02 15:00:00",
  "2022-03-01 20:00:00",
  "2022-03-01 21:00:00",
  "2022-03-01 20:00:00",
  "2022-03-02 15:00:00",
  "2022-03-01 21:00:00",
  "2022-03-02 15:00:00",
  "2022-03-01 20:00:00",
  "2022-03-01 21:00:00"
alamb@MacBook-Pro-2 influxdb_iox % ./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc
< database partition persist 26f7e5a4b7be365b_917b97a92e883afc
Error in partition subcommand: Must either specify a --partition-key or --all-partitions
alamb@MacBook-Pro-2 influxdb_iox % ./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions
./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions
Error in partition subcommand: Must either specify --table-name or --all-tables
alamb@MacBook-Pro-2 influxdb_iox % ./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions --all-tables
./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions --all-tables
Persisting partition: "2022-03-02 15:00:00", table: "cpu"
Error in partition subcommand: Client error: The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: cpu:2022-03-02 15:00:00
...
alamb@MacBook-Pro-2 influxdb_iox % ./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions --all-tables --continue-on-error
./target/debug/influxdb_iox database partition persist 26f7e5a4b7be365b_917b97a92e883afc --all-partitions --all-tables --continue-on-error
Persisting partition: "2022-03-02 15:00:00", table: "cpu"
Error persisting partition: "2022-03-02 15:00:00", table: "cpu": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: cpu:2022-03-02 15:00:00
Persisting partition: "2022-03-02 15:00:00", table: "disk"
Error persisting partition: "2022-03-02 15:00:00", table: "disk": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: disk:2022-03-02 15:00:00
Persisting partition: "2022-03-02 15:00:00", table: "diskio"
Error persisting partition: "2022-03-02 15:00:00", table: "diskio": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: diskio:2022-03-02 15:00:00
Persisting partition: "2022-03-02 15:00:00", table: "mem"
Error persisting partition: "2022-03-02 15:00:00", table: "mem": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: mem:2022-03-02 15:00:00
Persisting partition: "2022-03-02 15:00:00", table: "swap"
Error persisting partition: "2022-03-02 15:00:00", table: "swap": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: swap:2022-03-02 15:00:00
Persisting partition: "2022-03-02 15:00:00", table: "system"
Error persisting partition: "2022-03-02 15:00:00", table: "system": The system is not in a state required for the operation's execution: Cannot persist partition because it cannot be flushed at the moment: system:2022-03-02 15:00:00
Error in partition subcommand: Some operations returned an error, but --continue-on-error passed
```

influxdb_iox/src/commands/database/partition.rs
tustvold: Yeah - as it is for all time it could conceivably reach a point where pagination would be relevant - but not a now problem

PR #3875, opened 2022-02-28
chore: upgrade Rust to 1.59

crepererum:
Mostly a few new clippy crates around `flat_map`, `and_then`, and
"underscore locks" (!!!):
https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock


db/src/lib.rs
```diff
@@ -3023,7 +3022,7 @@ mod tests {
 
         let task = tokio::spawn(async move {
             sender.send(()).unwrap();
-            let _ = chunk_a.read();
+            let _read = chunk_a.read();
```
alamb, line 3025: this one seems unnecessary 😆

alamb (approved):
All in all I like the improvements to the code from the new clippy lints (which I haven't found to be the case all that often) 👍

db/src/lib.rs
crepererum: Yeah, you could also just drop get the lock and instantly drop it. I kept the underscore version for the sake of visibility (i.e. "a guard was created") and consistency.

alamb: makes sense 👍

Issue #3805, opened 2022-02-21
Write Buffer Aggregator Index Out Of Bounds


tustvold:
The write buffer frequently panics with output along the lines of

```
2022-02-15T14:52:26.013800841Z level=error msg="Thread panic" panic_info="panicked at 'index out of bounds: the len is 0 but the index is 0', write_buffer/src/kafka/aggregator.rs:467:22" target="panic_logging" location="panic_logging/src/lib.rs:94" time=1644936746013732368
```

This concerns the logic for matching the list of produced offsets, to the tags that produced them. In particular `Deaggregator::tag_to_record` appears to be empty, when it is expected it is expected to not be. I have never seen the panic caused by a non-empty `tag_to_record`. This suggests to me some sort of bug in the flush logic, either in IOx or rskafka, that causes `Aggregator::flush` to be called more than once, but I've yet to find a smoking gun

tustvold:
Will close once I've confirmed issue is fixed, and have cleaned up additional logging

crepererum:
This is fixed (at least we no longer see this issue).

PR #3877, opened 2022-02-28
feat: harden write buffer aggregator (#3805)

tustvold:
Manual inspection doesn't seem to be yielding insight into what is going on, so lets add some more assertions and logs and hope they shed some more light on the situation.


write_buffer/src/kafka/aggregator.rs
```diff
@@ -464,6 +500,15 @@ impl StatusDeaggregator for Deaggregator {
...
     ) -> Result<Self::Status, aggregator::Error> {
+        assert_eq!(input.len(), self.tag_to_record.len(), "invalid offsets");
+        assert!(
+            self.tag_to_record.len() > tag.0,
+            "tag {} out of range (tag_to_record: {:?}, offsets: {:?})",
+            tag.0,
+            self.tag_to_record,
+            input
+        );
+
```
tustvold, line 511: The line below is the one that is panic-ing. At least now we should at the very least gain the context of what the `Deaggregator` state is

write_buffer/src/kafka/aggregator.rs
```diff
@@ -370,9 +375,40 @@ impl Aggregator for DmlAggregator {
...
         state.flush_writes();
 
+        assert_eq!(self.state.current_writes.len(), 0, "incomplete flush");
+        assert_eq!(
+            self.state.completed_ops.len(),
+            self.state.tag_to_record.len(),
+            "missing records following flush"
+        );
+
+        info!(
```
tustvold, line 405: This log line may be rather spammy, but at this point the more context the better - we can always lower its level later

write_buffer/src/kafka/aggregator.rs
```diff
@@ -322,20 +329,18 @@ impl Aggregator for DmlAggregator {
...
-                            let mut md2 = op_md;
-                            std::mem::swap(agg, &mut agg2);
-                            std::mem::swap(record, &mut record2);
-                            std::mem::swap(md, &mut md2);
 
-                            let flushed_tag = agg2.tag;
-                            agg2.finalize_span();
-                            self.state.push_op(record2, md2, flushed_tag);
+                            // Replace current aggregator
+                            let (agg, record, md) = o.insert((new_agg, op_record, op_md));
```
tustvold, line 340: Just a drive by refactor to make it clearer

crepererum (approved):
Looks good. You might wanna put a bit of logging into `reserve_tag` as well. This is (ensured by the `Tag` type) the only place where tags are created, which is exactly the tag that later on in `deaggregate` `self.tag_to_record[tag.0]` panics.

tustvold:
> You might wanna put a bit of logging into reserve_tag as well.

I think this should already be covered by the existing length assertions, as the tags are just based on the length of `tag_to_record` but I could be mistaken?? I'll run with this for now, and see what happens

PR #3879, opened 2022-03-01
feat: split release build and deploy

tustvold:
The release builds take a while, sometimes in excess of 20 minutes. This PR therefore:

* Start the release build immediately instead of waiting for the tests to pass
* Only deploy the release if the tests pass
* Bump the no output timeout to 30 minutes :sob: 

If approved, I will test with the build_perf flag prior to merge


.circleci/config.yml
```diff
@@ -564,6 +586,7 @@ workflows:
     when: << pipeline.parameters.build_perf >>
     jobs:
       - build_release
+      - deploy_release
```
tustvold, line 589: This is potentially not desired, but preserves the existing behaviour

.circleci/config.yml
```diff
@@ -450,24 +447,47 @@ jobs:
...
+      - setup_remote_docker:
+          version: 19.03.13
+          docker_layer_caching: true
+      - run:
+          name: Update docker branch tags
+          command: |
+            COMMIT_SHA="$(git rev-parse --short HEAD)"
+            BRANCH="$(echo "$CIRCLE_BRANCH" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9]/_/g')"
+            
+            docker pull quay.io/influxdb/iox:"$COMMIT_SHA"
```
tustvold, line 474: You can avoid pulling the image, by talking to the registry directly with some hacky curl, I figured it was less fragile to just pull the images

.circleci/config.yml
crepererum: I think that's OK. You can build branches and get branch tags (i.e. they are not called `main`).

crepererum (approved):

crepererum (changes requested):

crepererum (approved):

PR #3880, opened 2022-03-01
refactor: clean up `querier` a bit

crepererum:
Before adding more and more features, here is a bit of a clean up and
prep work:

- Pull out caching into its own module and add proper tests for it.
- Start to build a test infrastructure so tests are shorter and easier
  to read. This doesn't fully pay off just yet but gets more and more
  important when we actually sync tables and chunks.


alamb (commented):
FWIW the structure of this PR looks good to me -- if you need me to give it a deeper review let me know @NGA-TRAN

NGA-TRAN (approved):
I am still reviewing but I send a couple of comments out because

PR #3881, opened 2022-03-01
test: querier shutdown and background task handling

crepererum:
This is similar to what we already have for the ingester.


querier/src/poison.rs
```diff
@@ -0,0 +1,93 @@
+use std::{
```
alamb, line 1: This is quite a cool (and unexpected, to me) structure.

Can you possibly add some comments about what it does and what it is for and how to use the code in this module?

querier/src/database.rs
```diff
@@ -134,12 +137,20 @@ impl QuerierDatabase {
...
     database: Arc<QuerierDatabase>,
     shutdown: CancellationToken,
+    poison_cabinet: Arc<PoisonCabinet>,
 ) {
     loop {
         if shutdown.is_cancelled() {
             info!("Database sync shutdown");
             return;
         }
+        if poison_cabinet.contains(&PoisonPill::DatabaseSyncPanic) {
```
alamb, line 147: It bothers me stylistically that `DatabaseSyncPanic` and `DatabaseSyncExit` are defined in different modules than their handling code. 

I wonder would it be possible for the code to `panic!()` and handle sync exit to be defined in the same file as the enums 🤔  but maybe that is overly complicated. 

I don't think any changes are needed, I just wanted to mention this

querier/src/handler.rs
```diff
@@ -118,3 +127,69 @@ impl Drop for QuerierHandlerImpl {
         }
     }
 }
+
+#[cfg(test)]
```
alamb, line 131: these are very clever tests

alamb (approved):
Screen Shot 2022-03-01 at 3 13 41 PM

From what I can see this PR doesn't actually change the shutdown / background handling, it just adds tests for them (which is 💯  but it might be worth updating the PR title)

 cc @NGA-TRAN

querier/src/database.rs
crepererum: I agree that it is not ideal, the issue is that `return` doesn't propagate very nicely w/o too much macro magic. In general I found the actual wiring of the "is pill present" check very specific to the actual code, since it somewhat depends where/who you await tasks or streams. A more complicated example exists within the `ingester`:

https://github.com/influxdata/influxdb_iox/blob/5f1095ef7d1fa294dc28fa767c33b4e6a524a66a/ingester/src/handler.rs#L296-L316

I think after the compactor also gets a similar system, I will generalize it and see what would make a somewhat nicer API, also coming up with a generic `PoisonCabinet` and a proper trait for the pill.

NGA-TRAN (commented):
This PR is for me to learn something new 👍

PR #3889, opened 2022-03-01
feat: wire up flight into the querier

crepererum:
No proper tests yet. Also note that `show databases;` doesn't work because that command doesn't use flight but the management API. Also there are no tables, partitions or chunks! What works however is:

```text
> use myorg_mybucket1;
You are now in remote mode, querying database myorg_mybucket1
myorg_mybucket1> show tables;
+---------------+--------------------+---------------------+------------+
| table_catalog | table_schema       | table_name          | table_type |
+---------------+--------------------+---------------------+------------+
| public        | system             | chunks              | BASE TABLE |
| public        | system             | columns             | BASE TABLE |
| public        | system             | queries             | BASE TABLE |
| public        | information_schema | tables              | VIEW       |
| public        | information_schema | columns             | VIEW       |
+---------------+--------------------+---------------------+------------+
Returned 8 rows in 51.966891ms
myorg_mybucket1>
```

(some system tables are a bit pointless for NG, but we can clean that up later).


influxdb_iox/src/influxdb_ioxd/server_type/querier/rpc/flight.rs
```diff
@@ -0,0 +1,22 @@
...
+};
+use data_types::DatabaseName;
+use querier::{database::QuerierDatabase, namespace::QuerierNamespace};
+
+use crate::influxdb_ioxd::rpc::flight::{make_server as make_server_inner, QueryDatabaseProvider};
+
+impl QueryDatabaseProvider for QuerierDatabase {
+    type Db = QuerierNamespace;
+
+    fn db(&self, db_name: &DatabaseName<'_>) -> Result<Arc<Self::Db>, tonic::Status> {
```
alamb, line 14: A soapbox moment: why did we rename the concept of "database" to "namespace"? (it seems inverted compared to most other systems where the hierarchy is:

`database -> schema (namespace) -> tables`

I realize this is really a question for @pauldix

querier/src/namespace.rs
```diff
@@ -1,17 +1,29 @@
...
 use crate::chunk::ParquetChunkAdapter;
 
 /// Maps a catalog namespace to all the in-memory resources and sync-state that the querier needs.
 #[derive(Debug)]
 pub struct QuerierNamespace {
     /// Old-gen DB catalog.
-    catalog: DbCatalog,
+    db_catalog: Arc<DbCatalog>,
 
     /// Adapter to create old-gen chunks.
```
alamb, line 28: I would like to understand your plan, at some point, of what the end state of the querier looks like

querier/src/namespace.rs
```diff
@@ -64,3 +100,62 @@ impl QuerierNamespace {
...
+
+    fn chunk_summaries(&self) -> Vec<ChunkSummary> {
+        self.catalog_access.chunk_summaries()
+    }
+
+    fn record_query(
+        &self,
+        query_type: impl Into<String>,
+        query_text: QueryText,
+    ) -> QueryCompletedToken<'_> {
```
alamb, line 134: this will conflict with changes I made in https://github.com/influxdata/influxdb_iox/pull/3891  (which removes the lifetime from `QueryCompletedToken` but I am happy to resolve them

querier/src/namespace.rs
```diff
@@ -34,13 +52,29 @@ impl QuerierNamespace {
         metric_registry: Arc<metric::Registry>,
         object_store: Arc<ObjectStore>,
         time_provider: Arc<dyn TimeProvider>,
+        exec: Arc<Executor>,
     ) -> Self {
+        let db_catalog = Arc::new(DbCatalog::new(
```
alamb, line 57: 👍

alamb (approved):
Looks cool to me -- I am not sure how the query catalog access works to expose chunks to the new system, but a nice step in the right direction

alamb:
cc @NGA-TRAN  you should probably also follow along with this one too

querier/src/database.rs
```diff
@@ -75,6 +82,11 @@ impl QuerierDatabase {
         self.namespaces.read().values().cloned().collect()
     }
 
+    /// Get namespace if it exists
+    pub fn namespace(&self, name: &str) -> Option<Arc<QuerierNamespace>> {
+        self.namespaces.read().get(name).cloned()
```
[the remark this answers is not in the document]
crepererum: > So if someone queries a just-ingested namespace and the Querier is not synced up to have the new namespace yet, we will say we cannot query this?

At the moment: yes. We're not triggering ANY iox_catalog request when you issue a query.

> ...we may need to go and see if it exists in the catalog and run the query

I doubt that we wanna do that unconditionally anyways due to DoS prevention. So we'll likely end up w/ a somewhat in-between solution: sync if a new query comes in and the last sync was more than X milliseconds ago.

Issue #3823, opened 2022-02-22
QueryLog Success/Failure


tustvold:
QueryLog records the completed_duration, but provides no indication if the query ran actually ran to completion or was aborted for some reason - panic, timeout, upstream hung up, etc... This makes it hard to reason about query execution time, as by definition the aborted queries will have truncated execution times.

Issue #3821, opened 2022-02-22
Query Log Incorrect Completed Duration for Flight Queries


tustvold:
The `QueryCompletedToken` is only held for the length of time it takes to construct the `GetStream` before it is dropped and the query is recorded as "completed". The token needs to be carried into the stream, so that it is dropped when the stream is exhausted

tustvold:
Reopening as was not resolved by #3891

PR #3891, opened 2022-03-01
feat: add `success` column to system.queries

alamb:
Resolves https://github.com/influxdata/influxdb_iox/issues/3823

# Rationale
In order to understand query execution time, we also need to understand if the query completed successfully or not

# Changes
Add `success` flag to `system.query` table, tests for same


influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -244,14 +244,19 @@ where
...
+
+        let mut query_completed_token = db.record_query("read_filter", defer_json(&req));
 
         let results = read_filter_impl(Arc::clone(&db), db_name, req, span_ctx)
             .await?
             .into_iter()
             .map(Ok)
             .collect::<Vec<_>>();
 
+        // Code above wraps everything in Ok; Assert here to catch if that changes
```
alamb, line 256: the amount of repetition in this file is not great. However, I am not totally sure how to get around it

influxdb_iox/src/influxdb_ioxd/rpc/flight.rs
```diff
@@ -182,7 +182,7 @@ where
 
         let db = self.server.db(&database)?;
 
-        let _query_completed_token = db.record_query("sql", Box::new(read_info.sql_query.clone()));
+        let query_completed_token = db.record_query("sql", Box::new(read_info.sql_query.clone()));
```
alamb, line 185: the changes in this file are what required removing the lifetime from `QueryCompletionToken` as now the token is passed into a future

tustvold: This will also likely help with #3821

influxdb_iox/tests/end_to_end_cases/system_tables.rs
```diff
@@ -168,12 +207,14 @@ async fn test_queries() {
...
-        "| sql        | select query_type, query_text from system.queries where query_type = 'sql' and query_text like '%read_filter%' |",
-        "+------------+----------------------------------------------------------------------------------------------------------------+",
+        "+------------+-------------------------------------------------------------------------------------------------------------------------+---------+",
+        "| query_type | query_text                                                                                                              | success |",
+        "+------------+-------------------------------------------------------------------------------------------------------------------------+---------+",
+        "| sql        | select query_type, query_text, success from system.queries where query_type = 'read_filter'                             | true    |",
+        "| sql        | select foo from blarg read_filter                                                                                       | false   |",
+        // note that the currently outstanding query is not (yet) successful
```
tustvold, line 215: Could add `completed_duration is not NULL as is_complete` as a column to make this clearer??

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
```diff
@@ -526,6 +548,10 @@ where
         .map(Ok)
         .collect::<Vec<_>>();
 
+        // Code above wraps everything in Ok; Assert here to catch if that changes
+        assert!(results.iter().all(|r| r.is_ok()));
+        query_completed_token.set_success();
+
```
tustvold, line 554:
```suggestion
        if results.iter().all(|r| r.is_ok()) {
            query_completed_token.set_success();
        }
```

?

tustvold: This appears in other locations as well

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
tustvold: I think when we adjust this code to stream responses we will likely need to create types for these streams, which might allow sharing some more of this logic

tustvold (approved):
LGTM - one thing that occurs to me as I write this is it might be useful to distinguish between errors and cancellation, but potentially something for a future PR.

influxdb_iox/src/influxdb_ioxd/server_type/database/rpc/storage/service.rs
alamb: in https://github.com/influxdata/influxdb_iox/pull/3891/commits/b2de42ad30b1a28081f6db268641b60208c6c125

influxdb_iox/tests/end_to_end_cases/system_tables.rs
alamb: in https://github.com/influxdata/influxdb_iox/pull/3891/commits/611fa912cdbd926305a455c113309a3f18c51147

PR #3894, opened 2022-03-01
fix: make catalog db migration idempotent again

### LINKED ISSUE #3893: Database setup is not idempotent
__Steps to reproduce:__

```bash
#!/bin/bash

psql -c 'DROP DATABASE IF EXISTS iox_shared'
psql -c 'CREATE DATABASE iox_shared'
cargo run -q -- catalog setup --catalog-dsn postgres:///iox_shared
cargo run -q -- catalog setup --catalog-dsn postgres:///iox_shared
```

__Expected behaviour:__

```
DROP DATABASE
CREATE DATABASE
OK
OK
```

__Actual behaviour:__

```
DROP DATABASE
CREATE DATABASE
OK
Catalog error: Unhandled sqlx error: while executing migrations: error returned from database: column "max_tables" of relation "namespace" already exists
```

__Environment info:__

Reproduced at commit 5f1095ef7d1fa294dc28fa767c33b4e6a524a66a

mkmik:
Closes #3893

This PR moves the schema creation from the migration script (//iox_catalog/migrations/20210217134322_create_schema.sql) to the `.setup()` call.

This ensures that the `iox_catalog` schema exists before the sqlx migration machinery is executed, otherwise the namespace where sqlx creates the `_sqlx_migrations` in, will be different the first and the subsequent runs (this violating schema migration idempotence).

(more detailed description in a comment in the code).

This PR also updates the README.md, instructing to run `influxdb_iox catalog setup` instead of `sqlx database setup`. This is not strictly necessary right now, since for development purposes it may work to keep the migration table in the `public` namespace, but it's a good idea to use the same database setup codepath for development and CI/prod.


iox_catalog/src/postgres.rs
```diff
@@ -274,6 +274,16 @@ async fn new_raw_pool(
...
+                // We need to create the schema if we're going to set it as the first item of the search_path
+                // otherwise when we run the sqlx migration scripts for the first time, sqlx will create the
+                // `_sqlx_migrations` table in the public namespace (the only namespace that exists), but the second
+                // time it will create it in the `<schema_name>` namespace and re-run all the migrations without
+                // skipping the ones already applied (see #3893).
+                //
+                // This makes the migrations/20210217134322_create_schema.sql step unnecessary; we need to keep that
+                // file because migration files are immutable.
+                let create_schema_query = format!("CREATE SCHEMA IF NOT EXISTS {}", schema_name);
+                c.execute(sqlx::query(&create_schema_query)).await?;
```
crepererum, line 286: I think a DB modification (aka `create schema`) shouldn't be a side effect of a `connect` routine, but rather the same routine that also runs the migrations:

https://github.com/influxdata/influxdb_iox/blob/5f1095ef7d1fa294dc28fa767c33b4e6a524a66a/iox_catalog/src/postgres.rs#L202-L209

crepererum (changes requested):

iox_catalog/src/postgres.rs
mkmik: yeah I was too lazy to weave through the `schema_name`

mkmik: @crepererum PTAL

crepererum (approved):
Thank you

PR #3897, opened 2022-03-02
feat: batch column upsert

domodwyer:
Adds a new batch upsert method to the column repo in order to upsert N columns to the catalog in one round trip.

As part of 210f6b40 the scope of the catalog trait taken by the schema validation was reduced, causing some minor code churn. The only functional change is within the body of `validate_mutable_batch()` where the new bulk method is called.

---

* docs: metric decorate macro (89dba1b7)

      Some helpful tips for using the metric decorator's decorate! macro when 
      modifying the catalog traits.

* feat: batch upsert of columns to catalog (da145ffb)

      Adds ColumnRepo::create_or_get_many() to upsert multiple columns in one round
      trip to Postgres.

* feat: batch column upsert for schema validation (8de453ed)

      Uses the new ColumnRepo::create_or_get_many() catalog method to perform a bulk
      upsert of (potentially) new columns to the catalog during schema validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2173,6 +2173,7 @@ checksum = "ec58677acfea8a15352d42fc87d11d63596ade9239e0a7c9352914417515dbe6"
 name = "iox_catalog"
 version = "0.1.0"
 dependencies = [
+ "assert_matches",
  "async-trait",
  "data_types2",
  "dotenv",
--- a/ingester/src/data.rs
+++ b/ingester/src/data.rs
@@ -1217,7 +1217,7 @@ mod tests {
     use iox_catalog::{mem::MemCatalog, validate_or_insert_schema};
     use mutable_batch_lp::{lines_to_batches, test_helpers::lp_to_mutable_batch};
     use object_store::ObjectStoreApi;
-    use std::time::Duration;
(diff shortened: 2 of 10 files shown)


alamb (commented):
Sorry for being late to the party, so please excuse my ignorance

Can you confirm that the purpose / rationale of this PR is to improve catalog performance?

If so, are you working on a particular scenario and do you have numbers about how much faster this approach is? It would be cool to know how much better this makes things

domodwyer:
I had to rebase to resolve a bunch of conflicts - I made no functional changes.

> Can you confirm that the purpose / rationale of this PR is to improve catalog performance?

The scope of this PR is to do one round trip for N columns instead of N, which should reduce the load on PG, reduce contention in the connection pool, etc. The _hypothesis_ is that this improves catalog performance purely by doing less things.

> If so, are you working on a particular scenario and do you have numbers about how much faster this approach is? It would be cool to know how much better this makes things

I'm trying to merge this so I can get it in our testbench setup to acuire these numbers :) I imagine it'll be helpful for all scenarios where there's > 1 new column in a write.

crepererum (approved):

PR #3898, opened 2022-03-02
feat: http ingest metrics

domodwyer:
Quick one to plumb in some handy ingest metrics for the HTTP write endpoint.

---

* feat: http ingest metrics (6d9709be)

      Records LP line count, field count & request body size (decompressed, byte
      size) for writes, and request body byte size for deletes.

### DIFF
--- a/router2/src/server/http.rs
+++ b/router2/src/server/http.rs
@@ -8,6 +8,7 @@ use data_types::names::{org_and_bucket_to_database, OrgBucketMappingError};
 use futures::StreamExt;
 use hashbrown::HashMap;
 use hyper::{header::CONTENT_ENCODING, Body, Method, Request, Response, StatusCode};
+use metric::U64Counter;
 use mutable_batch::MutableBatch;
 use observability_deps::tracing::*;
 use predicate::delete_predicate::{parse_delete_predicate, parse_http_delete_request};
@@ -154,6 +155,11 @@ pub struct HttpDelegate<D, T = SystemProvider> {
     max_request_bytes: usize,
     time_provider: T,
     dml_handler: D,
+
+    write_metric_lines: U64Counter,
+    write_metric_fields: U64Counter,
+    write_metric_body_size: U64Counter,
+    delete_metric_body_size: U64Counter,
 }
 
 impl<D> HttpDelegate<D, SystemProvider> {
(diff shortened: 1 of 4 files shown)


router2/src/server/http.rs
```diff
@@ -162,11 +168,40 @@ impl<D> HttpDelegate<D, SystemProvider> {
     ///
     /// HTTP request bodies are limited to `max_request_bytes` in size,
     /// returning an error if exceeded.
-    pub fn new(max_request_bytes: usize, dml_handler: D) -> Self {
+    pub fn new(max_request_bytes: usize, dml_handler: D, metrics: &metric::Registry) -> Self {
+        let write_metric_lines = metrics
+            .register_metric::<U64Counter>(
+                "http_write_lines_count",
```
e-dard, line 174: Can you make all of these suffixes `_total` rather than `_count`? It is more in line with the preferred naming for Prometheus metrics, which is how we are going to continue exposing these right?

router2/src/server/http.rs
```diff
@@ -162,11 +168,40 @@ impl<D> HttpDelegate<D, SystemProvider> {
...
+            .recorder(&[]);
+        let write_metric_fields = metrics
+            .register_metric::<U64Counter>(
+                "http_write_fields_count",
+                "cumulative number of line protocol fields successfully routed",
+            )
+            .recorder(&[]);
+        let write_metric_body_size = metrics
+            .register_metric::<U64Counter>(
+                "http_write_body_size",
```
e-dard, line 186: I think this should be named `http_write_body_bytes_total`

router2/src/server/http.rs
```diff
@@ -162,11 +168,40 @@ impl<D> HttpDelegate<D, SystemProvider> {
...
+                "cumulative byte size of successfully routed (decompressed) line protocol write requests",
+            )
+            .recorder(&[]);
+        let delete_metric_body_size = metrics
+            .register_metric::<U64Counter>(
+                "http_delete_body_size",
```
e-dard, line 192: I think it should be `http_delete_body_bytes_total`

e-dard (commented):
I think the metric names should be renamed to follow the Prometheus convention https://prometheus.io/docs/practices/naming/

But the code LGTM 👍

domodwyer:
Yep good shout - I pushed a rewrite that changes the names only 👍

e-dard (approved):

PR #3900, opened 2022-03-02
feat: sync partitions in querier

crepererum:
Now there we have partitions, the last missing bit are chunks (probably 1/2 day of work).


db/src/catalog.rs
```diff
@@ -102,13 +102,23 @@ impl Catalog {
             .collect()
     }
 
-    /// Get a specific table by name, returning `None` if there is no such table
+    /// Get a specific table by name, returning `TableNotFound` if there is no such table
     pub fn table(&self, table_name: impl AsRef<str>) -> Result<MappedRwLockReadGuard<'_, Table>> {
         let table_name = table_name.as_ref();
         RwLockReadGuard::try_map(self.tables.read(), |tables| tables.get(table_name))
             .map_err(|_| TableNotFoundSnafu { table: table_name }.build())
```
alamb, line 109: Can this use `.context(TableNotFound { table: table_name})?`? 

I realize you did not make this code in this PR I just noticed it while reviewing

db/src/catalog.rs
```diff
@@ -102,13 +102,23 @@ impl Catalog {
...
     }
 
+    /// Get a specific table by name, returning `TableNotFound` if there is no such table
+    pub fn table_mut(
+        &self,
+        table_name: impl AsRef<str>,
+    ) -> Result<MappedRwLockWriteGuard<'_, Table>> {
+        let table_name = table_name.as_ref();
+        RwLockWriteGuard::try_map(self.tables.write(), |tables| tables.get_mut(table_name))
+            .map_err(|_| TableNotFoundSnafu { table: table_name }.build())
```
alamb, line 119:
```suggestion
            .context(TableNotFound { table: table_name })
```

?

querier/src/namespace.rs
```diff
@@ -199,8 +220,97 @@ impl QuerierNamespace {
...
+        info!(
+            add = to_add.len(),
+            delete = to_delete.len(),
+            actual = actual_partitions.len(),
+            desired = desired_partitions.len(),
+            namespace = self.name.as_ref(),
+            "Syncing partitions",
+        );
+
+        let mut per_table_add_delete: HashMap<_, (Vec<_>, Vec<_>)> = HashMap::new();
```
alamb, line 278: It would help me to add a comment here explaining what the keys and values represented

Looks like a mapping from "table_name" to (partitions_to_add, and partitions_to_delete)?

alamb (approved):

crepererum:
> @crepererum -- how are you testing / running the querier locally?

test bench in GKE (google cloud)

db/src/catalog.rs
crepererum: The problem is that `try_map` returns an error but `TableNotFound` doesn't have an inner (= `source`) error because the inner error is irrelevant for the API user here. So `context` gets confused because it has no place to put the `try_map` error.

db/src/catalog.rs
crepererum: see above

querier/src/namespace.rs
crepererum: done

PR #3901, opened 2022-03-02
feat: rustls as TLS backend

### LINKED ISSUE #1354: Use Rustls
[Rustls](https://github.com/ctz/rustls) is a Rust implementation of SSL. This has a number of advantages:

* It's faster and uses less memory than OpenSSL
* Doesn't require libssl as a runtime dependency <sup>1</sup>
* Doesn't rely on OpenSSL <sup>2</sup>

_1. Some people might argue that cryptography libraries should be dynamically linked, but I think that ship sailed with docker (pun somewhat intended)_
_2. This is sort of misleading, rustls depends on ring, which has implementations borrowed from Google's BoringSSL, which itself derives code from OpenSSL :turtle:_

Most crates already have options to switch to using rustls, but it is possible that some of the more esoteric dependencies don't. Last time I looked the azure cloud SDK didn't have support, but in almost all cases it is simply a case of plumbing feature flags through correctly - and so should be a fairly straightforward PR to upstream

domodwyer:
This PR removes our dependency on OpenSSL, instead using [rustls](https://github.com/rustls/rustls) as the TLS implementation. 

Some benefits of this include:

* OpenSSL is a pain as a dependency on various platforms/envs/cross-compiling
* rustls was reported as being [faster](https://jbp.io/2019/07/01/rustls-vs-openssl-performance.html) (in 2019, requires a little 🧂)
* May well improve compile times - OpenSSL is a beast!
* Perhaps most importantly, is designed ground up to be a lower-risk TLS implementation, helped by the memory safety of rust and by implementing only features/protocol versions/ciphers that are commonly used in modern deployments

However a notable behavioural difference is the lack of IP SAN support - certificates that specify an IP address as the subject name will not work with rustls, though this is being [actively funded and worked on](https://www.abetterinternet.org/post/preparing-rustls-for-wider-adoption/). This may cause us an issue, and **we should be prepared to revert this PR** if it breaks functinoality in our env (which will manifest as connection errors with IP addresses as the endpoint). The driving reason I made this change was to support cross-compilation (macOS M1 -> musl linux) to avoid having to compile IOx in docker (sloooooow) in order to push new containers for testing - I can certainly live without it.

As part of this change I had to update the azure SDK pin to pick up rustls support, which had moved some things around.

Closes #1354.

---
* refactor: bump Azure SDK (4785a7d0)

      Update to the latest Azure SDK to pick up rustls support.

* refactor: use rustls (46bb107b)

      Removes openssl as a dependency, switching to rustls[1] as the TLS 
      implementation throughout.

      It is important to note that this change brings with it a significant 
      behavioural difference - rustls does not currently support IP SANs in 
      certificates (instead only supporting fully-qualified names / DNS) and this
      will manifest as a failure to connect to IP endpoints over TLS. This might be
      a blocker that prevents us using rustls exclusively, but there's noe asy way
      to know without trying it. Fortunately the rustls project has received funding
      to work on IP SAN support[2].

      [1]: https://github.com/rustls/rustls
      [2]: https://www.abetterinternet.org/post/preparing-rustls-for-wider-adoption/

* docs: remove openssl references (b391cdde)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -646,9 +646,10 @@ dependencies = [
  "hex",
  "jsonwebtoken",
  "lazy_static",
- "openssl",
+ "pem 0.8.3",
  "percent-encoding",
  "reqwest",
+ "ring",
  "serde",
  "serde_json",
  "tokio",
@@ -923,6 +924,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b2466559f260f48ad25fe6317b3c8dac77b5bdb5763ac7d9d6103530663bc90"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "ct-logs"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c1a816186fa68d9e426e3cb4ae4dff1fcd8e4a2c34b781bf7a822574a0d0aac8"
+dependencies = [
+ "sct 0.6.1",
 ]
 
 [[package]]
@@ -1419,21 +1429,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
-dependencies = [
- "foreign-types-shared",
-]
-
-[[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
-
-[[package]]
 name = "form_urlencoded"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


Cargo.lock
```diff
@@ -5339,13 +5413,14 @@ dependencies = [
...
-name = "tokio-native-tls"
-version = "0.3.0"
+name = "tokio-rustls"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f7d995660bd2b7f8c1568414c1126076c13fbb725c40112dc0120b78eb9b717b"
+checksum = "bc6844de72e57df1980054b38be3a9f4702aba4858be64dd700181a8a6d0e1b6"
 dependencies = [
- "native-tls",
+ "rustls 0.19.1",
```
alamb, line 5421: looks like there are several versions of rusttls now used -- but we can sort that out later perhaps

alamb (approved):
Looks reasonable to me - it might be worth doing two PRs -- one with the upgrade of the azure libraries and one with switch over to rust tls.

Cargo.lock
crepererum: Likely an upstream issue, e.g. https://github.com/launchbadge/sqlx/pull/1505

domodwyer:
I rebased to pick up the latest changes (more native tls!) and split the azure change into a separate commit as I merge PRs so the history sticks around - we can revert the rustls commit separately if needed 👍

However, I have no idea what's up with the workspace hack thing - I might have to summon the wizard @carols10cents 🧙‍♀️

crepererum:
@dom just run `cargo hakari generate` locally and commit the changes.

crepererum (approved):

domodwyer:
I tried that, but unfortunately _no pasa nada_:

```console
~iox % cargo hakari generate
info: no changes detected
```

domodwyer:
@crepererum I bestow the title of wizard upon you 🧙‍♂️

PR #3902, opened 2022-03-02
fix: jemalloc feature flag

domodwyer:
Adds missing feature flags to make the `jemalloc_replacing_malloc` feature truely optional - in the current state, compiling iox with `--no-default-features` errors out.

---

* fix: jemalloc feature flag (6204a904)

      In #2435 we moved the jemalloc metrics around, and missed a few feature gates
      to avoid pulling in jemalloc unless it is requested.


alamb (approved):
Seems fine with me. 

I wonder if it worth adding a command like `cargo check --no-default-features` in the circle ci config to make sure this doesn't get broken in the future

PR #3903, opened 2022-03-02
fix: apply `_field` restrictions that do not match any fields

alamb:
Fixes https://github.com/influxdata/conductor/issues/891

# Rationale:

A predicate like `_field = 'foo'` means restrict to only data that has the field named `foo`. This was being ignored by IOx when ruling out chunks, resulting in incorrect results.

# Changes:
Properly evaluate the `_field` predicates

# TODO
- [x] verify that the reported test case passes with this fix
- [x] Update internal CI test that was skipping this test (PR is up)


db/src/access.rs
```diff
@@ -232,6 +232,25 @@ impl ChunkAccess {
             .catalog_snapshot_duration
             .inc(start.elapsed());
 
+        // if there is a field restriction on the predicate, only
+        // chunks with that field should be returned. If the chunk has
+        // none of the fields specified, then it doesn't match
+        let chunks = match &predicate.field_columns {
```
alamb, line 238: this is the fix

query_tests/src/influxrpc/tag_values.rs
```diff
@@ -332,6 +332,46 @@ async fn list_tag_values_field_col() {
...
+#[tokio::test]
+async fn list_tag_values_field_col_does_not_exist() {
+    let tag_name = "state";
+    let predicate = PredicateBuilder::default()
+        .timestamp_range(0, 1000) // get all rows
+        // since this field doesn't exist this predicate should match no values
+        .add_expr(col("_field").eq(lit("not_a_column")))
+        .build();
+    let predicate = InfluxRpcPredicate::new(None, predicate);
+    let expected_tag_keys = vec![];
```
alamb, line 344: this test fails (incorrectly returns `MA`, `CA`, and `NY`)

onelson: This test is a nice explanation of the problem. Huh.

onelson (approved):
I don't really know enough of the context to vet the change, but if the tests are green this LGTM :shipit:

db/src/access.rs
```diff
@@ -233,6 +233,25 @@ impl ChunkAccess {
...
+            None => chunks,
+            Some(field_columns) => {
+                chunks
```
tustvold, line 242: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain might be a nicer way to write this

tustvold (approved):
I'm not entirely sure how a bug in chunk pruning would cause more data to be returned that expected, if anything this feels like it is masking an underlying bug, but this code looks correct to me

alamb:
So I thought more about this. Part of the confusion is that the InfluxRPC uses predicates for a concept that is very different in SQL: selection

`_field = 'foo'` means "select only the `foo` column. However, if there is no `foo` column there should be no columns (or rows) selected. So I guess we could also add a `false` Expr to the `Predicate` if there were no columns. But I am having a hard time coming up with a reproducer for that case.

PR #3905, opened 2022-03-02
test: Add some end-to-end ish NG tests

carols10cents:
Blocked by https://github.com/influxdata/influxdb_iox/pull/3944

Technically this test is only going router -> ingester -> queried as if the querier would through the flight API, it doesn't actually start a querier or compactor yet.

But I think there's a good framework here... certainly more that could be done as far as making more reusable/readable components rather than setting a bunch of environment variables for the server processes... let me know if there are any improvements that you'd like to see before merging!


crepererum:
I think it's OK if the end2end test require postgres in the beginning, however I would like to have purely file-based (e.g. via sqlite) `iox_catalog` backend at some point, similar to what we have for the object store and the write buffer.h

alamb:
> I think it's OK if the end2end test require postgres in the beginning, however I would like to have purely file-based (e.g. via sqlite) iox_catalog backend at some point, similar to what we have for the object store and the write buffer.


I agree with both of these items. I am happy to work on the 'pure file based' version. I am gathering my thoughts into a plan over the next few days.

alamb (commented):
What would you think about making a new end to end test framework (copying the relevant parts over) -- `influxdb_iox/tests/end_to_end_ng.rs`?

I think trying to maintain the existing harness and make it work for both ng and the og will be quite challenging

carols10cents:
> What would you think about making a new end to end test framework (copying the relevant parts over) -- influxdb_iox/tests/end_to_end_ng.rs?

I am very into that.

influxdb_iox/src/commands/run/router2.rs
```diff
@@ -143,15 +143,9 @@ pub async fn command(config: Config) -> Result<()> {
     let mut txn = catalog.start_transaction().await?;
     let topic_id = txn
         .kafka_topics()
-        .get_by_name(config.write_buffer_config.topic())
+        .create_or_get(&config.write_buffer_config.topic())
```
carols10cents, line 146: @domodwyer Is this ok to do? It's in the section marked "testing only"... without this, the catalog doesn't have the kafka topic `iox_shared` from the file-based write buffer in it, and I'm not sure if this is an acceptable way to go about putting it in the catalog or if there's a different setup step I've missed.

influxdb_iox/tests/common_ng/server_fixture.rs
```diff
@@ -0,0 +1,568 @@
...
+    /// This is currently implemented as a singleton so all tests *must*
+    /// use a new router and not interfere with the existing routers.
+    pub async fn create_shared(server_type: ServerType) -> Self {
+        // Try and reuse the same shared server, if there is already
+        // one present
+        static SHARED_SERVERS: OnceCell<parking_lot::Mutex<HashMap<ServerType, Weak<TestServer>>>> =
+            OnceCell::new();
+        let shared_servers = SHARED_SERVERS.get_or_init(|| parking_lot::Mutex::new(HashMap::new()));
+
+        let mut shared_servers = shared_servers.lock();
```
alamb, line 99: I suspect the shared stuff is copy/paste from the old end to end harness (which is fine) but I wonder if we will be using shared servers going forward or if we should delete this code just keep `create_single_server` until sharing is needed?

alamb (approved):
All in all I like it @carols10cents , thank you! I see no reason we can't merge this and iterate afterwards. 

Some thoughts on follow on work:
1. Is there any chance you can document how to run these tests somewhere (including the postgres requirement and what environment variables need to be configured)?
2. As a follow on, shall we hook this into circle CI?


I took it for a spin and I coudn't get the test to pass for me locally. Maybe because some sort of database setup / migration is needed? If so, perhaps we can have `TestServer` itself run the required database migration step?

Here is what happened:

```shell
DATABASE_URL="postgresql://localhost:5432/alamb" TEST_INTEGRATION=1 cargo test --test end_to_end_ng

...
INSERT INTO
  kafka_topic (name)
VALUES
  ($1) ON CONFLICT ON CONSTRAINT kafka_topic_name_unique DO
UPDATE
SET
  name = kafka_topic.name RETURNING *;
  
2022-03-03T21:46:18.114025Z  WARN iox_catalog::postgres: Dropping PostgresTxn w/o finalizing (commit or abort)
2022-03-03T21:46:18.114334Z  INFO sqlx::query: /* SQLx ping */; rows: 0, elapsed: 186.941µs  
Server command failed: Error in router2 subcommand: Catalog error: Unhandled sqlx error: error returned from database: relation "kafka_topic" does not exist
****************
End Router2 TestServer Output
****************


failures:
    end_to_end_ng_cases::ng::router2_through_ingester

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 30.01s


```


I tested various ways to run the tests incorrectly and nice error reporting 👍 
```shell
alamb@MacBook-Pro-2:~/Software/influxdb_iox$ TEST_INTEGRATION=1 cargo test --test end_to_end_ng
    Finished test [unoptimized + debuginfo] target(s) in 0.50s
     Running tests/end_to_end_ng.rs (target/debug/deps/end_to_end_ng-22af823990daa9f2)

running 1 test
test end_to_end_ng_cases::ng::router2_through_ingester ... FAILED

failures:

---- end_to_end_ng_cases::ng::router2_through_ingester stdout ----
thread 'end_to_end_ng_cases::ng::router2_through_ingester' panicked at 'TEST_INTEGRATION is set, but variable(s) DATABASE_URL need to be set', influxdb_iox/tests/end_to_end_ng_cases/ng.rs:15:24

```

influxdb_iox/src/commands/run/router2.rs
```diff
@@ -143,15 +143,9 @@ pub async fn command(config: Config) -> Result<()> {
...
+        .create_or_get(config.write_buffer_config.topic())
         .await?
-        .map(|v| v.id)
-        .unwrap_or_else(|| {
-            panic!(
-                "no kafka topic named {} in catalog",
-                config.write_buffer_config.topic()
-            )
-        });
+        .id;
```
crepererum, line 148: This was somehow already covered by `influxdb_iox catalog topic update <my_topic_name>`. I think we somewhat concluded that we don't wanna do auto-creation for topics because we likely wanna add more parameters to `topic` in the future (also see #3646).

influxdb_iox/tests/common_ng/server_fixture.rs
```diff
@@ -0,0 +1,568 @@
...
+    grpc_channel: Connection,
+}
+
+impl ServerFixture {
+    /// Create a new server fixture and wait for it to be ready. This
+    /// is called "create" rather than new because it is async and
+    /// waits. The shared router is configured with a server id and
+    /// can be used immediately
+    ///
+    /// This is currently implemented as a singleton so all tests *must*
```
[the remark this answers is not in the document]
crepererum: I thought being a singelton was the whole point of `create_shared`? If you wanna have a non-singelton process call `create_single_use`. So I think this should rather read as:

```suggestion
    /// This is implemented as a singleton so all tests *must*
```

or even

```suggestion
    /// All tests *must*
```

influxdb_iox/tests/common_ng/server_fixture.rs
crepererum: Let's keep the single-use code, because I guess we're going to need this rather soon for the management API tests (e.g. list all X).

crepererum (approved):
Smaller nitpicks/questions.

influxdb_iox/src/commands/run/router2.rs
domodwyer: I don't think we want this I'm afraid - we had spoke about who is responsible for creating the record, and determined we'd use a bootstrap CLI method or something similar (basically some tool that populates the DB outside of the normal processes) to ensure there's no weird starting-up-a-new-router-instance-and-suddenly-you-have-200-more-topics kinds of bugs. The router probably won't have all the information to populate a topic record in the future either :(

influxdb_iox/tests/common_ng/server_fixture.rs
```diff
@@ -0,0 +1,568 @@
...
+    time::Duration,
+};
+use tempfile::{NamedTempFile, TempDir};
+use tokio::sync::Mutex;
+
+// These port numbers are chosen to not collide with a development ioxd server
+// running locally.
+static NEXT_PORT: AtomicU16 = AtomicU16::new(8090);
+
+pub const DEFAULT_SERVER_ID: u32 = 32;
```
domodwyer, line 25: We shouldn't need this - we're ID-less in NG land 🎉

influxdb_iox/tests/common_ng/server_fixture.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    pub async fn wait_for_grpc(&self) {
+        let mut interval = tokio::time::interval(Duration::from_millis(1000));
+
+        loop {
+            match (self.grpc_channel().await, self.test_config.server_type) {
+                (Ok(channel), ServerType::Router2) => {
+                    let mut health = influxdb_iox_client::health::Client::new(channel);
+
+                    match health.check("influxdata.pbdata.v1.WriteService").await {
```
domodwyer, line 446: Does this pass currently? The gRPC impl is a [stub](https://github.com/influxdata/influxdb_iox/blob/845cfaedd1d27d6f5e586178c7ccbb23672784e0/router2/src/server/grpc.rs#L32) in `router2` at the moment...

I can quickly knock an impl up though which is probably easier long term - I think this is the correct approach 👍

influxdb_iox/tests/end_to_end_ng_cases/ng.rs
```diff
@@ -0,0 +1,98 @@
...
+    },
+    maybe_skip_integration,
+};
+use arrow_util::assert_batches_sorted_eq;
+use data_types2::{IngesterQueryRequest, SequencerId};
+use hyper::{Body, Client, Request, StatusCode};
+use tempfile::TempDir;
+
+#[tokio::test]
+async fn router2_through_ingester() {
```
domodwyer, line 14: Well isn't this lovely! Really cool 🌮

domodwyer (commented):
This will be a great addition - thanks for taking the time

influxdb_iox/src/commands/run/router2.rs
alamb: For testing it is likely very helpful to auto create topics -- perhaps we can add an explicit `auto_create_topics` configuration flag to control if the topics are auto created or not?

domodwyer: I can see why that'd be nice - this code will be removed entirely at some point and I imagine we'll still want the behaviour - what about moving the bootstrap code into the setup of the fixture?

domodwyer:
@carols10cents I think you might want to take a look at the [`influxdb_iox catalog topic`](https://github.com/influxdata/influxdb_iox/blob/845cfaedd1d27d6f5e586178c7ccbb23672784e0/influxdb_iox/src/commands/catalog.rs#L61-L63) CLI command that bootstraps the kafka records in the catalog (instead of implicitly creating them in the router) 👍

influxdb_iox/tests/common_ng/server_fixture.rs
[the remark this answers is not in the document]
alamb: I guess I was saying i hadn't seen a `create_shared` in the tests so far

influxdb_iox/src/commands/run/router2.rs
alamb: Moving the bootstrap code into the fixture sounds good to me 👍

influxdb_iox/tests/common_ng/server_fixture.rs
domodwyer: I managed to do this while waiting for benchmarks to complete 🎉 

https://github.com/influxdata/influxdb_iox/pull/3936

influxdb_iox/tests/common_ng/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+// Helper macro to skip tests if TEST_INTEGRATION and DATABASE_URL environment variables are
+// not set.
+#[macro_export]
+macro_rules! maybe_skip_integration {
+    () => {{
+        use std::env;
+        dotenv::dotenv().ok();
+
+        let required_vars = ["DATABASE_URL"];
```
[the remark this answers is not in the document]
carols10cents: Sure, I was making this consistent with [what `iox_catalog` does in its tests](https://github.com/influxdata/influxdb_iox/blob/c0ea6d1cfc4c38335fddc04277c6e563b550f239/iox_catalog/src/postgres.rs#L1377), but I can change that too :)

influxdb_iox/tests/common_ng/server_fixture.rs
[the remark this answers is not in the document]
carols10cents: Yeah there's only one test so far, and I'm not sure if we want shared or not yet, I haven't really thought through the postgres implications. So I just kept both methods around.

influxdb_iox/tests/common_ng/server_fixture.rs
carols10cents: Whoops, great catch I missed!

influxdb_iox/tests/common_ng/server_fixture.rs
carols10cents: Yeah this test just hits the `/health` gRPC thingie, which the `add_service!` macro sets up. I think that just tests if you can connect at all, not whether the response is success or failure. Good to see you've implemented gRPC writing anyway though!!

carols10cents:
> @carols10cents I think you might want to take a look at the [`influxdb_iox catalog topic`](https://github.com/influxdata/influxdb_iox/blob/845cfaedd1d27d6f5e586178c7ccbb23672784e0/influxdb_iox/src/commands/catalog.rs#L61-L63) (idempotent) CLI command that bootstraps the kafka records in the catalog (instead of implicitly creating them in the router) 👍

Will do-- I didn't know about that command, I've been using `influxdb_iox catalog setup`, which apparently doesn't set up the topic? Should it?

carols10cents:
> 2\. As a follow on, shall we hook this into circle CI?

It's already in there! [The test job sets TEST_INTEGRATION and DATABASE_URL](https://github.com/influxdata/influxdb_iox/blob/c0ea6d1cfc4c38335fddc04277c6e563b550f239/.circleci/config.yml#L220-L230), and [`cargo test` runs this test](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16339/workflows/da24b306-eeac-4344-9e58-2ca782a98d67/jobs/130530?invite=true#step-109-882).

carols10cents:
> I took it for a spin and I coudn't get the test to pass for me locally. Maybe because some sort of database setup / migration is needed? If so, perhaps we can have `TestServer` itself run the required database migration step?

Ah yes, currently this requires the same test setup as the `iox_catalog` tests [as documented in that README](https://github.com/influxdata/influxdb_iox/blame/c0ea6d1cfc4c38335fddc04277c6e563b550f239/iox_catalog/README.md#L6-L43). The problem with running that as part of the test is that it blows away the database, which other tests might want to use.... I'm not sure how best to handle that.

alamb:
> The Rust test harness doesn't really have a way to run something once before all tests... do you have any thoughts?

What about as part of `ServerFixture::create_single_use_with_config` and `ServerFixture::create_shared`?

carols10cents:
> What about as part of ServerFixture::create_single_use_with_config and ServerFixture::create_shared?

Each single use and each shared would still need to use their own database though, right? Or else they could still trample each other?

alamb:
> Each single use and each shared would still need to use their own database though, right? Or else they could still trample each other?

Yes, I am hoping we eventually come to the conclusion it is ok to use a postgres `schema` to isolate multiple concurrent tests from each other. @domodwyer  and @pauldix  is that ok with you?

carols10cents:
> Yes, I am hoping we eventually come to the conclusion it is ok to use a postgres `schema` to isolate multiple concurrent tests from each other. @domodwyer and @pauldix is that ok with you?

Ahh I just noticed that [the iox_catalog tests are creating a schema with a random name](https://github.com/influxdata/influxdb_iox/blob/e09f39d6a069d9859f7fa38f8819263bd005fb71/iox_catalog/src/postgres.rs#L1431-L1442).

carols10cents:
I decided to separate out the env var change into https://github.com/influxdata/influxdb_iox/pull/3944 to get comments on that change specifically.

iox_catalog/README.md
```diff
@@ -52,21 +52,6 @@ DIFFERENT database name for your test database than your `INFLUXDB_IOX_CATALOG_D
...
-```
-
-* Set up the test database as above but specify the test database URL as follows (`catalog setup`
-  will NOT pick up the `TEST_` version of the environment variable from your `.env` file so it
-  needs to be specified explicitly WITHOUT the `TEST_` prefix):
-
-```
-INFLUXDB_IOX_CATALOG_DSN=<testdsn> cargo run -q -- catalog setup
-INFLUXDB_IOX_CATALOG_DSN=<testdsn> cargo run -q -- catalog topic update iox-shared
-```
```
crepererum, line 68: Who creates the database now? I think the DB creation is only wired up when you run the `influxdb_iox` test suite. So if you start w/ a fresh postgres and just run `<all the env vars...> cargo test -p iox_catalog`, it will fail because the DB is missing, no?

influxdb_iox/tests/end_to_end_ng_cases/ng.rs
```diff
@@ -0,0 +1,98 @@
...
+    // Set up router2 ====================================
+
+    let test_config = TestConfig::new(ServerType::Router2)
+        .with_postgres_catalog(&database_url)
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_TYPE", "file")
+        .with_env(
+            "INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS",
+            n_sequencers.to_string(),
+        )
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_ADDR", &write_buffer_string);
```
crepererum, line 35: I think the integration tests ALWAYS wanna use a file-based write buffer, so this entire setup (including the tempdir stuff above) can be part of the server fixture. However this could also be done in a follow-up PR if you like. I just wanna make sure that writing new integration tests is as easy as possible and has the least "copy+paste" overhead.

iox_catalog/README.md
carols10cents: Ah, great catch! I will add the db creation to the `iox_catalog` tests too.

influxdb_iox/tests/end_to_end_ng_cases/ng.rs
```diff
@@ -0,0 +1,98 @@
...
+    let response = client.request(request).await.unwrap();
+    assert_eq!(response.status(), StatusCode::NO_CONTENT);
+
+    // Set up ingester ===================================
+
+    let test_config = TestConfig::new(ServerType::Ingester)
+        .with_postgres_catalog(&database_url)
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_TYPE", "file")
+        .with_env("INFLUXDB_IOX_PAUSE_INGEST_SIZE_BYTES", "20")
+        .with_env("INFLUXDB_IOX_PERSIST_MEMORY_THRESHOLD_BYTES", "10")
```
crepererum, line 64: Same as for the write buffer config: I think this can be set by default. If we wanna make this configurable, I would rather offer typed methods liek `TestConfig::with_pause_ingest_size_bytes(...)`. Same goes for the write buffer partition range.

Similar to the comment above: this can be a follow-up PR.

influxdb_iox/tests/common_ng/server_fixture.rs
```diff
@@ -0,0 +1,577 @@
+use assert_cmd::prelude::*;
```
crepererum, line 1: I kinda rely on the fact that this is copy+paste+modify of the existing code. One thing to test manually would be to place a `panic` at the end of the integration test and see if the server stdout is indeed printed (technically how this works is that we always print the captured server stdout when the test fixture is dropped and Rust's `libtest` captures that printout and emits it for every failed test).

crepererum (approved):
Looks good, I think that's a solid start. Two points (also see inline comments:)

1. the `iox_catalog` tests should be able to run w/o relying on the integration test be executed beforehand (this should be fixed before merging)
2. the test fixture could use more typed options and defaults w/o relying on the test setting the correct env variables (this can be a follow-up PR)

Issue #3884, opened 2022-03-01
Split API logic out of influxdb_iox crate into separate crates


tustvold:
The influxdb_iox crate is extremely slow to compile, and this is adding a frustrating amount of latency to getting code into production.

I therefore collected a self-profile to try to understand what it was spending time doing

```
$ RUSTFLAGS="-Zself-profile=$PWD/profiles -Zself-profile-events=default,args" cargo build --release
$ crox --dir profiles --minimum-duration 1000
```

[chrome_profiler.zip](https://github.com/influxdata/influxdb_iox/files/8161890/chrome_profiler.zip)

If my interpretation of this is correct, it is primarily bottlenecked on LLVM codegen, with the majority of time spent in `LLVM_module_optimize` and `LLVM_passes`.

I therefore collected data about what functions were creating the most IR.

[lines.zip](https://github.com/influxdata/influxdb_iox/files/8161941/lines.zip)

This suggests to me that the tonic, and hyper code are responsible for a significant amount of generated LLVM IR. This is not hugely surprising as they rely extensively on generics. I therefore think that there might be value in splitting the various gRPC and HTTP server and client code into separate crates so that they can be compiled in parallel.

Perhaps something like

* `influxdb_iox_common`
* `influxdb_iox_database`
* `influxdb_iox_router`
* `influxdb_iox_router2`
* `influxdb_iox_compactor`
* `influxdb_iox_querier`
* `influxdb_iox_cli`

alamb:
I like the sound of this plan. Please let m eknow if I can help

alamb:
I was thinking sequencing wise we would do a single PR that broke out one or two of the crates to define the pattern and then just apply that pattern to everything else

This might also help with with querier side of things (which will likely be a plumbing exercise I think) so I can justify it that way

NGA-TRAN:
@alamb Does it mean 'influx_iox_compactor` will be used for the new NG compactor? We do not have any code for it yet but I will add one soon/tomorrow.

alamb:
> @alamb Does it mean 'influx_iox_compactor` will be used for the new NG compactor? We do not have any code for it yet but I will add one soon/tomorrow.


Yes, I think so @NGA-TRAN  -- I hope to have the first PR done shortly and then you can just follow the model of the other services (database, router, router2) and hopefully not be affected much

NGA-TRAN:
Thanks @alamb . I plan to reuse/invoke available compaction functions

[...]

PR #3908, opened 2022-03-02
refactor: split `influxdb_ioxd`, `clap_blocks` and `serving_readiness` from influxdb_iox crate

alamb:
~Draft as I ran out of time today, but hope to finish up tomorrow~

# Rationale
Speed up compile times by breaking the `influxdb_iox` crate into smaller subcrates: https://github.com/influxdata/influxdb_iox/issues/3884

# Changes:

This is phase 1 of splitting the API logic out into separate crates to improve IOx compile times. It is mostly mechanical and I have commented in places where something more complicated is going on. 

1. Splits the code that implements the various run modes into its own `influxdb_ioxd` top level crate
2. Splits out `clap_blocks` and `serving_readiness` into their own top level crates (as they are needed by both influxdb_iox and influxdb_ioxd)

In subsequent PRs I plan to split `influxdb_ioxd` further into `influxdb_ioxd_common`,  `influxdb_ioxd_database`, etc.


clap_blocks/src/run_config.rs
```diff
@@ -66,3 +63,30 @@ pub struct RunConfig {
     #[clap(flatten)]
     pub(crate) object_store_config: ObjectStoreConfig,
 }
+
+impl RunConfig {
```
alamb, line 67: These functions are to help encapsulate the struct a bit more (rather than making all fields `pub`)

influxdb_iox/src/commands/run/database.rs
```diff
@@ -96,11 +94,34 @@ pub struct Config {
...
+        self.config_file.as_ref()
+    }
+}
+
 pub async fn command(config: Config) -> Result<()> {
     let common_state = CommonServerState::from_config(config.run_config.clone())?;
 
-    let application = make_application(&config, common_state.trace_collector()).await?;
-    let app_server = make_server(Arc::clone(&application), &config)?;
+    let application = make_application(
```
alamb, line 112: `Config` is defined in `influxdb_iox` so I had to update the `make_application` and `make_server` to take only the fields they needed rather than the whole struct

influxdb_ioxd/src/server_type/database/setup.rs
```diff
@@ -1,71 +1,75 @@
...
         object_storage,
-        config.num_worker_threads,
+        num_worker_threads,
         trace_collector,
         config_provider,
     )))
 }
 
-pub fn make_server(application: Arc<ApplicationState>, config: &Config) -> Result<Arc<Server>> {
+pub fn make_server(
```
alamb, line 57: Here is one non trivial change to break a dependency on `Config` (defined in `influxdb_iox`)

influxdb_ioxd/Cargo.toml
```diff
@@ -0,0 +1,139 @@
+[package]
```
alamb, line 1: This likely has a bunch of unused dependencies (it is mostly copy/pasted from influxdb_iox). I plan to clean it up in a follow on PR

crepererum:
> Speed up compile times by breaking the influxdb_iox crate into smaller subcrates

Does this make a difference? I.e. is it faster?

influxdb_ioxd/src/server_type/database/rpc/storage/expr.rs
```diff
@@ -251,7 +251,7 @@ impl InfluxRpcPredicateBuilder {
 /// 1. Flatten `None` value nodes with `children` of length 1 (semantically the
 /// same as the child itself). Specifically, if the input is:
 ///
-/// ```
+/// ```text
```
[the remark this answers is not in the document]
crepererum: I think it might be a public/private thing? :thinking:

crepererum (approved):

alamb:
> Does this make a difference? I.e. is it faster?

This PR is very unlikely to make things faster as `influxdb_iox` depends on `influxdb_ioxd`

It will likely be faster if things are changed in `influxdb_iox`, but I didn't test that

alamb:
Merging this one in sooner to avoid merge conflicts

PR #3924, opened 2022-03-03
feat: Auto commit cargo hakari generate changes

### LINKED ISSUE #3680: Dependabot doesn't run `cargo hakari generate`

carols10cents:
Fixes #3680.

When someone (either dependabot or a person) opens a PR that needs changes to the workspace hack crate to update the features listed, or if other crates need to be changed to make them depend on the workspace hack crate, instead of CI, this will add a commit to the PR.

Take a look at the previous PR for this: https://github.com/influxdata/influxdb_iox/pull/3713

It's this branch plus a commit where I deleted part of the workspace-hack's Cargo.toml, and the hakari CircleCI job was able to push a new commit to the PR that fixed it 🎉

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,19 +170,30 @@ jobs:
       - rust_components
       - cache_restore
       - run:
-          name: Check that the workspace hack crate contains all features in use
+          name: Configure git
           command: |
-            cargo hakari generate --diff || {
-              echo "If this fails, fix it by running \`cargo hakari generate\` locally and committing the changes"
-              exit 1
-            }
+            git config user.email "49699333+dependabot[bot]@users.noreply.github.com"
+            git config user.name "dependabot[bot]"
       - run:
-          name: Check that all crates in the workspace depend on the workspace hack crate
+          name: Commit any changes to the features in the workspace hack crate
           command: |
-            cargo hakari manage-deps --dry-run || {
-              echo "If this fails, fix it by running \`cargo hakari manage-deps\` locally and committing the changes"
-              exit 1
-            }
+            cargo hakari generate
+            if [[ $(git status --porcelain) ]] ; then
+              git commit -am "[skip ci] Run cargo hakari generate"
+              git push origin "$CIRCLE_BRANCH"
+            else
+              echo "No changes to commit"
+            fi
+      - run:
+          name: Commit any changes to crates so they all depend on the workspace hack crate
+          command: |
+            cargo hakari manage-deps
+            if [[ $(git status --porcelain) ]] ; then
+              git commit -am "[skip ci] Run cargo hakari manage-deps"
+              git push origin "$CIRCLE_BRANCH"
+            else
+              echo "No changes to commit"
+            fi
 
   test:
     # setup multiple docker images (see https://circleci.com/docs/2.0/configuration-reference/#docker)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,8 +172,8 @@ jobs:
       - run:
           name: Configure git
           command: |
-            git config user.email "49699333+dependabot[bot]@users.noreply.github.com"
-            git config user.name "dependabot[bot]"
+            git config user.email "circleci@influxdata.com"
+            git config user.name "CircleCI[bot]"
       - run:
           name: Commit any changes to the features in the workspace hack crate
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -173,26 +173,26 @@ jobs:
       - rust_components
       - cache_restore
       - run:
+          name: Check git log
+          command: |
+            if [[ $(git rev-list --count --author=CircleCI origin/main..) -ne 0 ]]; then
+              echo "There's already a cargo hakari commit on this branch; stopping'"
+              circleci-agent step halt
+            else
+              echo "No cargo hakari commits here; checking"
+            fi
+      - run:
           name: Configure git
           command: |
             git config user.email "circleci@influxdata.com"
             git config user.name "CircleCI[bot]"
       - run:
-          name: Commit any changes to the features in the workspace hack crate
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -166,23 +166,37 @@ jobs:
...
+              git push origin "$CIRCLE_BRANCH"
+            else
+              echo "No changes to commit"
+            fi
+      - run:
+          name: Commit any changes to crates so they all depend on the workspace hack crate
+          command: |
+            cargo hakari manage-deps
+            if [[ $(git status --porcelain) ]] ; then
+              git commit -am "[skip ci] Run cargo hakari manage-deps"
```
alamb, line 195: I am not familiar with `[skip ci]`  -- but it seems to me like we should actually re-run CI after changing cargo-hakari (otherwise it could potentially break the build without being flagged)

carols10cents: Hmmmmmmmmmm. I'm imagining a hakari bug somehow where it thrashes on the workspace-hack crate and commits infinitely? Could that happen?

alamb: probably... I wonder if we could skip auto committing if there is already another auto commit?

carols10cents: Whew I think I've got it... on this branch, if you look at [the workspace hack CI job](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16349/workflows/a082fd9f-7551-4c56-acea-22c67bff1269/jobs/130633), you'll see that the job didn't see any existing cargo hakari commits made by CircleCI on this branch, ran the cargo hakari commands, and didn't find any changes to commit.

If you look at the other PR https://github.com/influxdata/influxdb_iox/pull/3713, you'll see:

1. [A commit that causes workspace hack changes](https://github.com/influxdata/influxdb_iox/pull/3713/commits/ee1d4706c68d3db60bbc07b3383f2e4345748c17)-- [the job sees there is no hakari commit yet, runs the commands, commits the changes, and pushes them](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16350/workflows/123a1bb4-546a-4c2f-b114-e761b0e8c3ff/jobs/130645)
2. [On the commit the job added](https://github.com/influxdata/influxdb_iox/pull/3713/commits/d35132422a2b9fefd91f11322b1ea81e84a480b7), [the "Check git log" step sees there's already a hakari commit then stops running the job](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16351/workflows/97a0a9b7-dedb-453d-bf97-659ae08a4fa2/jobs/130661) (it's still a success)

What do you think?

alamb: The only thing that is slightly strange to me is that the job that pushes the new commit is marked as "success": https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16350/workflows/123a1bb4-546a-4c2f-b114-e761b0e8c3ff/jobs/130645

I wonder if we should mark it as a failure (to signify that that commit would not be suitable to merge) ?

.circleci/config.yml
```diff
@@ -166,23 +166,37 @@ jobs:
...
       - run:
-          name: Check that the workspace hack crate contains all features in use
+          name: Check git log
           command: |
-            cargo hakari generate --diff || {
-              echo "If this fails, fix it by running \`cargo hakari generate\` locally and committing the changes"
-              exit 1
-            }
+            if [[ $(git rev-list --count --author=CircleCI origin/main..) -ne 0 ]]; then
+              echo "There's already a cargo hakari commit on this branch; stopping'"
```
alamb, line 179: 👍

alamb (approved):
I think we should give it a try!

alamb:
Thanks @carols10cents  this is going to be great

crepererum (approved):

.circleci/config.yml
domodwyer, line 179: I think it might be worth doing a `cargo hakari generate --diff` check (like what we had before) at this point (after determining there was a commit already, before exiting sucessfully) to prevent the check getting confused and green-lighting the merge with a bad hakari config? Is that even possible?

domodwyer (approved):
image

.circleci/config.yml
carols10cents: Good call! I could see a race condition with kodiak happening otherwise.

.circleci/config.yml
carols10cents: 🤷🏻‍♀️ I have no idea if it's possible, but I can certainly add this check in so we can see!

carols10cents:
Ok, this PR passes as expected.

Over on [my test PR](https://github.com/influxdata/influxdb_iox/pull/3713),

- A commit changing Cargo.toml files [failed the workspace hack checks job](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16381/workflows/4097fdf5-b3fa-4be4-b6b1-a67a52fa13f4/jobs/130989) as expected
- That job did [push an additional commit](https://github.com/influxdata/influxdb_iox/pull/3713/commits/8eb343748b3e7a41ae71ee6f322b362f606411de) as desired
- All CI jobs were run on the additional commit, including checking hakari again, which found no changes so [the job passed](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16382/workflows/1e02fde7-45d9-4901-8bdd-39ee0b777c76/jobs/131000).

I'm not sure how to cause the theoretical problem where there's a hakari commit but still hakari changes, but at least I haven't seen that case *falsely* invoked.

So I think this is ready to merge :)

PR #3925, opened 2022-03-03
feat: Send parquet and tombstone seq nums with ingester query response

### LINKED ISSUE #3867: Add more fields into IngesterQueryResponse
# Rationale

Now our [TableData](https://github.com/influxdata/influxdb_iox/blob/c965221df14fc9a9157d135239e30057148f159c/ingester/src/data.rs#L518) includes 2 fields `parquet_max_sequence_number` and `tombstone_max_sequence_number`, we should return those with [IngesterQueryResponse](https://github.com/influxdata/influxdb_iox/blob/c965221df14fc9a9157d135239e30057148f159c/ingester/src/data.rs#L1123).

Rename `max_sequencer_number` to `parquet_max_sequence_number`

carols10cents:
Fixes #3867.


NGA-TRAN (approved):
LGTM

PR #3927, opened 2022-03-03
feat: more router metrics

domodwyer:
These metrics will help us answer a few questions that have popped up over the last couple days.

More metrics == more insights 👀 

---

* refactor: write table count metric (e00986c56)

      Record the number of tables in each write - this will let us observe the total
      number of tables a router instance has observed, which when combined with the
      existing metrics helps us understand the shape
      (distribution of tables/lines/fields) of the workload hitting the routers.

* refactor: sequencer metrics (bb9b140f4)

      Records per-sequencer (kafka partition) enqueue latency / counts broken down
      by operation success/error.

* refactor: NamespaceCache latency histograms (6b5283bf3)

      Switches the get/put counters to latency histograms to record the duration of
      each call - this might be interesting!

* feat: schema validation benchmarks (14d90d101)

      Useful for confirming the scalability of the schema check algorithm.

### DIFF
--- a/influxdb_iox/src/commands/run/router2.rs
+++ b/influxdb_iox/src/commands/run/router2.rs
@@ -227,7 +227,7 @@ async fn init_write_buffer(
     let write_buffer = Arc::new(
         config
             .write_buffer_config
-            .writing(metrics, trace_collector)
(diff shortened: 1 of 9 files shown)


mkmik (approved):
love this; thanks!

PR #3928, opened 2022-03-04
feat: add a flag to enable per-partition tracing

tustvold:
#3532 added per-partition tracing. Unfortunately certain tracing setups struggle to handle the granularity of these traces, resulting in dropped spans. This PR therefore gates this feature behind an environment variable so that setups can opt-in to this functionality. This is much the same approach taken #2135 where the metric granularity was causing problems


e-dard (approved):
👍

query/src/exec/query_tracing.rs
```diff
@@ -14,6 +14,14 @@ use hashbrown::HashMap;
 use observability_deps::tracing::debug;
 use trace::span::{Span, SpanRecorder};
 
+const TIMESTAMP_METRICS_ENABLE_ENV: &str = "INFLUXDB_IOX_PER_PARTITION_TRACING";
```
alamb, line 17: the name of this environment variable seems not to match its rust code, which may be confusing. Maybe something like `const INFLUXDB_IOX_PER_PARTITION_TRACING` instead?

query/src/exec/query_tracing.rs
```diff
@@ -59,7 +67,13 @@ impl Drop for TracedStream {
     fn drop(&mut self) {
         if let Some(span) = self.span_recorder.span() {
             let default_end_time = Utc::now();
-            send_metrics_to_tracing(default_end_time, span, self.physical_plan.as_ref());
+            let per_partition_tracing = per_partition_tracing();
```
alamb, line 70: I wonder if parsing the environment variable all the time is wasteful... but it makes sense to try it this way and then optimize to something fancier if it shows up in profiles

alamb (approved):

alamb:
Removed `automerge` flag in case you want to rename the variable

query/src/exec/query_tracing.rs
tustvold: Oh lol - copy pasta :laughing:

query/src/exec/query_tracing.rs
tustvold: Not sure how expensive `std::env::var` is, but the parse part is a simple match block and so should be pretty fast - https://doc.rust-lang.org/src/core/str/traits.rs.html#585

tustvold: I actually updated this to use the BooleanFlag construct to be consistent with the rest of the CLI, and added caching. PTAL :smile:

docs/tracing.md
```diff
@@ -85,6 +85,14 @@ For example, a command such as this should do the trick:
 TRACES_EXPORTER=jaeger TRACES_EXPORTER_JAEGER_AGENT_HOST=localhost TRACES_EXPORTER_JAEGER_AGENT_PORT=6831 cargo run -- run -v --server-id=42
 ```
 
+Additional trace granularity, in particular traces with spans for each DataFusion partition, can be enabled with
```
alamb, line 88: 👍

query/src/exec/query_tracing.rs
```diff
@@ -14,6 +15,26 @@ use hashbrown::HashMap;
 use observability_deps::tracing::debug;
 use trace::span::{Span, SpanRecorder};
 
+const PER_PARTITION_TRACING_ENABLE_ENV: &str = "INFLUXDB_IOX_PER_PARTITION_TRACING";
+fn per_partition_tracing() -> bool {
+    use std::sync::atomic::{AtomicU8, Ordering};
+    static TRACING_ENABLED: AtomicU8 = AtomicU8::new(u8::MAX);
+
+    match TRACING_ENABLED.load(Ordering::Relaxed) {
```
alamb, line 23: I wonder if `AtomicBoolean` would be sufficient

alamb (approved):

query/src/exec/query_tracing.rs
tustvold: I need a ternary, unfortunately

PR #3931, opened 2022-03-04
feat: improve trace naming

tustvold:
The current logic simply truncates the description to the 20 characters. This is unfortunate as:

* The resulting span names vary based on the arguments in the plan
* The span names are arbitrarily truncated, resulting in `SortPreservingMergeE` among other things
* The actual full description is lost

This PR changes to use the leading operator name from the description as the span name, and include the full description as a metadata field. This provides the full context, whilst also allowing use-cases like searching for all traces containing `SortExec`, `DeduplicateExec`, etc... or finding time spent in `SortExec` spans, etc...

An example trace with the changes can be seen below

image


query/src/exec/query_tracing.rs
```diff
@@ -297,16 +295,17 @@ mod tests {
...
         let traces = TraceBuilder::new();
         send_metrics_to_tracing(Utc::now(), &traces.make_span(), &exec);
 
         let spans = traces.spans();
         assert_eq!(spans.len(), 1);
-        // name is truncated to 20 cahracters
-        assert_eq!(spans[0].name, "TestExec: This is a ", "span: {:#?}", spans);
+
+        // name is truncated to the operator name
+        assert_eq!(spans[0].name, "TestExec - Foo", "span: {:#?}", spans);
```
alamb, line 308: I think it would be worth a test that `desc` is properly populated with the full description too

alamb (approved):
👍

query/src/exec/query_tracing.rs
```diff
@@ -349,28 +357,59 @@ mod tests {
         println!("Spans: \n\n{:#?}", spans);
         assert_eq!(spans.len(), 12);
 
-        let check_span = |span: &Span, expected_start, expected_end| {
+        let check_span = |span: &Span, expected_start, expected_end, desc: Option<&str>| {
             assert_eq!(span.start, expected_start, "expected start; {:?}", span);
             assert_eq!(span.end, expected_end, "expected end; {:?}", span);
+            assert_eq!(span.metadata.get("desc").map(|x| x.string().unwrap()), desc);
```
alamb, line 363: 👍

alamb (approved):

Issue #3935, opened 2022-03-04
Subsequently inserted string field values are lost


alamb:
__Steps to reproduce:__

1. Download the `input.lp` from: [input.zip](https://github.com/influxdata/influxdb_iox/files/8186350/input.zip)

Then load the data and query it

```
influxdb_iox database create test_db
influxdb_iox database write test_db /tmp/input.lp
influxdb_iox database query test_db 'select * from m'
```

__Expected behaviour:__
The loaded data should be

```
+-----+----+----+----+-------+----------------------+
| f0  | f1 | f2 | f3 | f4    | time                 |
+-----+----+----+----+-------+----------------------+
| cat | 10 | 10 | 10 | true  | 2021-12-16T00:00:00Z |
| dog | 11 | 11 | 11 | false | 2021-12-16T00:01:00Z |
| cat | 12 | 12 | 12 | true  | 2021-12-16T00:02:00Z |
| emu | 13 | 13 | 13 | false | 2021-12-16T00:03:00Z |
| bird| 14 | 14 | 14 | true  | 2021-12-16T00:04:00Z |
+-----+----+----+----+-------+----------------------+
```

__Actual behaviour:__

lost all values other than `cat`:

```shell
(arrow_dev) alamb@MacBook-Pro-2:~/Software/idpe$ cat /tmp/input.lp
m f0="cat" 1639612800000000000
m f0="dog" 1639612860000000000
m f0="emu" 1639612920000000000
m f0="rat" 1639612980000000000
m f0="bird" 1639613040000000000
m f1=10i 1639612800000000000
m f1=11i 1639612860000000000
m f1=12i 1639612920000000000
m f1=13i 1639612980000000000
m f1=14i 1639613040000000000
m f2=10u 1639612800000000000
m f2=11u 1639612860000000000
m f2=12u 1639612920000000000
m f2=13u 1639612980000000000
m f2=14u 1639613040000000000
m f3=10 1639612800000000000
m f3=11 1639612860000000000
m f3=12 1639612920000000000
m f3=13 1639612980000000000
m f3=14 1639613040000000000
m f4=true 1639612800000000000
m f4=false 1639612860000000000
m f4=true 1639612920000000000
m f4=false 1639612980000000000
m f4=true 1639613040000000000
(arrow_dev) alamb@MacBook-Pro-2:~/Software/idpe$ ../influxdb_iox/target/debug/influxdb_iox database create test_db
Created database test_db
[...]

PR #3939, opened 2022-03-04
fix: correctly set nulls in string columns

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3935

# Rationale

Re https://github.com/influxdata/conductor/issues/892

`NULL` strikes again!

The validity bit in String fields was not being correctly set. Thus string columns that were supposed to be `NULL` were actually seen as `""`, a subtle but critical difference for correctness!

This bug was likely masked for string columns specifically because most of our tests compare batches by printing them to strings where `""` and `null` are rendered the same.

Change:
1. Properly use validity mask
2. explicit tests for null when inserting null line protocol

I also verified this now passes the fluxtest suite:

```shell
isTypeTableString...success
```


mutable_batch/src/column.rs
```diff
@@ -298,7 +298,7 @@ impl Column {
                     .context(CreatingArrowArraySnafu)?;
                 Arc::new(UInt64Array::from(data))
             }
-            ColumnData::String(data, _) => Arc::new(data.to_arrow()),
+            ColumnData::String(data, _) => Arc::new(data.to_arrow(Some(nulls))),
```
alamb, line 301: this was the problem: the null mask that was computed didn't make it to the created `StringArray`

query_tests/src/influxrpc/field_columns.rs
```diff
@@ -165,7 +165,7 @@ async fn test_field_name_plan() {
             Field {
                 name: "field2".into(),
                 data_type: DataType::Utf8,
-                last_timestamp: 1000, // Need to verify with alamb if 1000 is the right one. It looks to me it should be 100
```
alamb, line 168: LOL @NGA-TRAN  had spotted this problem many 🌔 's ago

mutable_batch_lp/src/lib.rs
```diff
@@ -225,7 +225,83 @@ mod tests {
...
+
+        // Verify the nullness of the string column ("" not the same as null)
+        let f0 = &batch.columns()[0];
+        assert!(f0.is_valid(0));
+        assert!(!f0.is_valid(1));
+
+        // Verify the nullness of the f1 column ("" not the same as null)
+        let f1 = &batch.columns()[1];
+        assert!(!f1.is_valid(0));
+        assert!(f1.is_valid(1));
```
NGA-TRAN, line 262: 👍

query_tests/src/influxrpc/field_columns.rs
NGA-TRAN: 🎉

NGA-TRAN (approved):
This NULL never-ended issues remind me of easy-to-us vs more-work-for-developer

tustvold (approved):
Nice find :+1: 

Perhaps we should have an option to format nulls differently for use by tests...

alamb:
> Perhaps we should have an option to format nulls differently for use by tests...

Might definitely be nice

PR #3940, opened 2022-03-04
feat: ingester seeks kafka partition on initialization

### LINKED ISSUE #3851: Ingester: seek kafka partitions based on sequencer state in catalog
The ingester initialization should seek to the kafka partitions based on the `min_unpersisted_sequence_number` in the catalog

pauldix:
Fixes #3851


ingester/src/handler.rs
```diff
@@ -661,6 +666,114 @@ mod tests {
...
+                        // verify there's data in the buffer
+                        if let Some((b, _)) = data.snapshot("cpu", "1970-01-01").await {
+                            if let Some(b) = b.first() {
+                                match b.data.num_rows() {
+                                    0 => continue,
+                                    1 => has_measurement = true,
+                                    _ => panic!("initialization did a seek to the beginning rather than the min_unpersisted"),
+                                }
+                            }
+                        }
```
crepererum, line 762: How does this show that `w2` was ingested but `w1` was not? If I understand this correctly, this only shows that there's a single row ingested, but this could also be because it only ingested `w1` and did not yet pull `w2`.

ingester/src/handler.rs
```diff
@@ -174,11 +174,16 @@ impl IngestHandlerImpl {
...
-            let stream_handler = write_buffer
+            let mut stream_handler = write_buffer
                 .stream_handler(kafka_partition.get() as u32)
                 .await
                 .context(WriteBufferSnafu)?;
 
+            stream_handler
+                .seek(sequencer.min_unpersisted_sequence_number as u64)
+                .await
+                .context(WriteBufferSnafu)?;
```
crepererum, line 185: So this concludes the "seeking" part. However for some partitions we already persisted more than this and IIRC we wanted to filter out these writes as well. Is this already covered / implemented / tracked somewhere?

ingester/src/handler.rs
crepererum: To me this panic sounds a bit racy TBH.

ingester/src/handler.rs
pauldix: Yes, that was in #3849

ingester/src/handler.rs
pauldix: Ok, I think I found a cleaner way to verify this that isn't racey. I look at the min sequence number instead and panic if it's the same as that first write.

crepererum (approved):

PR #3942, opened 2022-03-04
feat: ingester persists on cold partitions

### LINKED ISSUE #3869: Ingester: add partition cold for writes duration persistence trigger in lifecycle
The ingester should persist partitions that have been cold for writes for some period of time to clear them out of the buffer to free up memory. Name the flag `-persist-partition-cold-threshold-seconds` and have the default be 300 seconds.

pauldix:
Add configuration and lifecycle to trigger partition persistence if it hasn't received a write in a given number of secods.

Fixes #3869


influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -134,6 +134,15 @@ pub struct Config {
     )]
     pub persist_partition_age_threshold_seconds: u64,
 
+    /// If a partition has had data buffered and hasn't received a write for this
+    /// period of time, it will be persisted. The default value is 5 minutes (in seconds).
```
alamb, line 138:
```suggestion
    /// period of time, it will be persisted. The default value is 300 seconds (5 minutes).
```

ingester/src/lifecycle.rs
```diff
@@ -142,17 +151,24 @@ impl LifecycleManager {
         sequence_number: SequenceNumber,
         bytes_written: usize,
     ) -> bool {
+        let now = self.time_provider.now();
```
alamb, line 154: Maybe worth calling this after the lock is acquired (so that `now` isn't too far in the past if you have to wait for the lock)? I don't think it will really matter though

ingester/src/lifecycle.rs
```diff
@@ -189,9 +205,13 @@ impl LifecycleManager {
                 .checked_duration_since(s.first_write)
                 .map(|age| age > self.config.partition_age_threshold)
                 .unwrap_or(false);
+            let is_cold = now
```
alamb, line 208: I wonder if we still need `aged_out` with the `is_cold` check? I can't think of a reason to force persistence of a partition that still is getting active writes if it isn't too large 🤔

alamb (approved):
Looks like the code very nicely does what it claims to do 👍

ingester/src/lifecycle.rs
pauldix: Aged out is so that we have control over how far back we'd have to read into Kafka on startup. It gives us the ability to put an upper bound on ingester recovery time.

PR #3944, opened 2022-03-04
fix: Use more consistent env vars for database connections

carols10cents:
I decided to pull this out of https://github.com/influxdata/influxdb_iox/pull/3905. [Over there, @alamb asked](https://github.com/influxdata/influxdb_iox/pull/3905#discussion_r819082786) why the environment variable was named `DATABASE_URL` and had this suggestion:

> What would you think about using the same environment variable as the command line, for consistency?
>
> I think the CLI uses `INFLUXDB_IOX_CATALOG_DSN`
>
> https://github.com/influxdata/influxdb_iox/blob/d6afda0227fcb11efac3c65996e3da481481b100/clap_blocks/src/catalog_dsn.rs#L30-L33
>
> Or perhaps something derived from the same name like `TEST_INFLUXDB_IOX_CATALOG_DSN` if we want to try and avoid the classic "I accidentally ran the tests suite against the production database" type scenario 😆 

I chose `DATABASE_URL` because that's what `iox_catalog` was using, and `iox_catalog` was using it because that's what `sqlx` uses.

This PR proposes to change that by:

- Creating the database, if it doesn't already exist, in `cargo run -- catalog setup` so that `sqlx database create` isn't needed at all 
  -  I believe now the only reason people will need to install `sqlx-cli` will be to manage migrations? Does that sound correct? I have adjusted the documentation in `iox_catalog/README.md` to reflect my understanding
- Creating the `SHARED_KAFKA_TOPIC` in `cargo run -- catalog setup` so that the router can use it without you also needing to run `cargo run -- catalog topic` (not really required for the env var change, but sort of related, and I wanted to check that assumption for setup in this context)
- Use `INFLUXDB_IOX_CATALOG_DSN` instead of `DATABASE_URL` in code
- Use `TEST_INFLUXDB_IOX_CATALOG_DSN` in integration tests

The only place this is awkward is that, currently, to run the `iox_catalog` integration tests, you'll need to first run once:

```
INFLUXDB_IOX_CATALOG_DSN=[manually insert your TEST_INFLUXDB_IOX_CATALOG_DSN here] cargo run -q -- catalog setup
```

because we wouldn't want `catalog setup` to look at the `TEST_INFLUXDB_IOX_CATALOG_DSN` in your .env file normally (or would we?????)

We could also change the tests so that they call what `catalog setup` does in a future PR to remove this need.

Please check my assumptions and understanding!!


mkmik:
the migrations are actually handled by `influxdb_iox catalog setup`. The only thing we use the sqlx CLI binary for is to create the database. We could show the equivalent psql call instead.

carols10cents:
> the migrations are actually handled by `influxdb_iox catalog setup`. The only thing we use the sqlx CLI binary for is to create the database. We could show the equivalent psql call instead.

I meant more if people are creating migrations; to get the filenames in the right format, it's best to do `sqlx migrate create`, right?

mkmik:
> > the migrations are actually handled by `influxdb_iox catalog setup`. The only thing we use the sqlx CLI binary for is to create the database. We could show the equivalent psql call instead.
> 
> I meant more if people are creating migrations; to get the filenames in the right format, it's best to do `sqlx migrate create`, right?

Right. But that's a relatively infrequent operation (as opposed to prepping the environment for running integration tests)

> Also, this PR changes `influxdb_iox catalog setup` so that it creates the database if it doesn't exist, is that a problem?

No, that's quite useful

iox_catalog/src/postgres.rs
```diff
@@ -275,6 +286,11 @@ async fn new_raw_pool(
     dsn: &str,
     max_conns: u32,
 ) -> Result<sqlx::Pool<Postgres>, sqlx::Error> {
+    // Create the database if it doesn't exist
+    if !Postgres::database_exists(dsn).await? {
+        Postgres::create_database(dsn).await?;
+    }
```
crepererum, line 292: Couple of comments:

- Creating databases normally requires special permissions that our normal prod user doesn't have. I think it's still OK to at least try to create the DB if it isn't present, however...
- This setup seems very racy if we have let's say 20 nodes booting up, each of them checking if the DB exists and then trying to create it. I guess `create_database` will fail if the DB already exists, so that should be handled here.

So I'm not sure if it's really worth the hassle or if we should create the DB explicitly in our test fixture.

crepererum (changes requested):

crepererum:
> The only place this is awkward is that, currently, to run the `iox_catalog` integration tests, you'll need to first run once:
> 
> ```
> INFLUXDB_IOX_CATALOG_DSN=[manually insert your TEST_INFLUXDB_IOX_CATALOG_DSN here] cargo run -q -- catalog setup
> ```
> 
> because we wouldn't want `catalog setup` to look at the `TEST_INFLUXDB_IOX_CATALOG_DSN` in your .env file normally (or would we?????)

We can the integration test suite / fixture not call `catalog setup` with the right parameters automatically? To me it kinda seems that we're leaking quite a bunch of test assumptions into our prod code here, even though in my head this adaptation should normally be done by the test framework / suite / fixture.

carols10cents:
@crepererum @mkmik Ok, now this PR only changes the environment variable names and the documentation -- in the next PR, I will make the server fixture take care of running the commands that are in the docs in this PR for the tests, and then I will remove those docs. The setup and commands for non-tests then won't change.

What do you think?

crepererum (approved):

crepererum:
> @crepererum @mkmik Ok, now this PR only changes the environment variable names and the documentation -- in the next PR, I will make the server fixture take care of running the commands that are in the docs in this PR for the tests, and then I will remove those test setup docs. The setup and commands for non-tests then won't change.
> 
> What do you think?

Sounds like a good plan to me :)

domodwyer (approved):
You're literally maintaining my sanity as well as code at this point.

mkmik (approved):

Issue #3958, opened 2022-03-07
NG Querier: end2end tests for flight and InfluxRPC


crepererum:
Blocked on #3905.

alamb:
We have coverage of the flight SQL interface in https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_ng_cases/querier.rs

We need to port the grpc tests now

PR #3959, opened 2022-03-07
feat: generic InfluxRPC impl, used by querier

### LINKED ISSUE #3933: NG Querier: connect influxRPC
The idea here is to allow flux and influxQL queriers to run against the NG querier

crepererum:
Luckily the "storage" API was already using a trait that was very
similar to the one that was introduced in #3889 to abstract the flight
API. Here we move a single, shared trait; move the storage API
implementation to a common module and use it to implemenet InfluxRPC for
both the OG database and the NG querier.

Closes #3933.

End2end tests are blocked on #3905 and are tracked by #3958.


influxdb_ioxd/src/rpc/flight.rs
```diff
@@ -120,12 +122,6 @@ struct ReadInfo {
     sql_query: String,
 }
 
-pub trait QueryDatabaseProvider: std::fmt::Debug + Send + Sync + 'static {
```
alamb, line 123: 👍

influxdb_ioxd/src/rpc/storage/mod.rs
```diff
@@ -15,15 +15,18 @@ pub mod input;
 pub mod service;
 
 use generated_types::storage_server::{Storage, StorageServer};
-use server::DatabaseStore;
```
alamb, line 18: I may have missed it, but we can probably remove the `DatabaseStore` trait entirely after this PR

influxdb_ioxd/src/rpc.rs
```diff
@@ -9,7 +9,9 @@ use trace_http::ctx::TraceHeaderParser;
 use crate::server_type::{RpcError, ServerType};
 use serving_readiness::ServingReadiness;
 
+pub(crate) mod common;
```
alamb, line 12: ❤️ 

I like how this structure is coming together

influxdb_ioxd/src/rpc/common.rs
```diff
@@ -0,0 +1,9 @@
+use std::sync::Arc;
+
+use query::{exec::ExecutionContextProvider, QueryDatabase};
+
+pub trait QueryDatabaseProvider: std::fmt::Debug + Send + Sync + 'static {
```
alamb, line 5: Can we perhaps add some docstrings to this trait (especially for the rationale -- "To keep the RPC implementations from directly relying on the query engine" or something?)

alamb (approved):
👨‍🍳 ❤️ 

I am a big fan -- nice work @crepererum  

I do wonder if we should be tracking somewhere what end to end tests we should be adding. Maybe it is obvious.

influxdb_ioxd/src/rpc/storage/mod.rs
NGA-TRAN: If I understand correctly, the `DatabaseStore` was only implemented for `Server` and `TestDatabaseStore`.  `DatabaseStore` is now replaced by `QueryDatabaseProvider` and the `db_or_create` trait function is implemented directly as a function. Hence it can be removed entirely

NGA-TRAN (approved):
LGTM

influxdb_ioxd/src/rpc/storage/mod.rs
crepererum: done (deleted)

crepererum:
> I do wonder if we should be tracking somewhere what end to end tests we should be adding. Maybe it is obvious.

TBH I don't know yet, so whoever implements #3958 has to think about it :sweat_smile:

PR #3960, opened 2022-03-07
chore: reduce router log verbosity (#3805)

tustvold:
Drops some very chatty logging that was added to help debugging #3805 down to debug level. Also removes a TODO that I no longer intend to do


write_buffer/src/kafka/aggregator.rs
```diff
@@ -514,7 +514,6 @@ impl StatusDeaggregator for Deaggregator {
                 "tag {} out of range (database_name: {}, tag_to_record: {:?}, offsets: {:?})",
                 tag.0, self.database_name, self.tag_to_record, input
             );
-            // TODO: Temporary non-fatal assertion to reduce log spam (#3805)
```
tustvold, line 517: I don't think there is anything harmful about leaving this as not a panic

domodwyer (approved):

PR #3961, opened 2022-03-07
build: remove openssl dependency

domodwyer:
Looks like #3908 accidentally reintroduced OpenSSL as a dependency.

We should blacklist openssl with something like [`cargo-deny`](https://github.com/EmbarkStudios/cargo-deny) to avoid this dance.

---

* build: remove openssl dependency (baae6910f)

      Switches a stray reqwest dependency to use rustls.


tustvold (approved):

domodwyer:
> Sorry about that

No problem! I've opened a PR (#3963) that uses cargo-deny to warn us if we depend on OpenSSL by accident in the future 👍

Issue #3953, opened 2022-03-04
NG Compactor: remove fully processed tombstones


NGA-TRAN:
This ticket implements step 7 of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). The input is a list of [parquet files](https://github.com/influxdata/influxdb_iox/blob/6ba5e518976a1feadb09b27b74967ac57b0e9971/data_types2/src/lib.rs#L624) and the output is a list of list of overlapped parquet files. Something like

```rust
fn remove_fully_processed_tombstones(&[TombstoneId]) -> Result<()>
``` 

Note that this function need to find fully processed tombstones in the given list first

Update: can the given tomstonesIds a HashSet instead?

PR #3962, opened 2022-03-07
feat: wire up tombstones into querier 

### LINKED ISSUE #3932: NG Querier: Tombstone Handling
Idea is to have the querier handle tombstones correctly

crepererum:
Closes #3932


db/src/catalog/chunk.rs
```diff
@@ -541,6 +541,25 @@ impl CatalogChunk {
         }
     }
 
+    pub fn override_delete_predicates(&mut self, delete_predicates: Vec<Arc<DeletePredicate>>) {
```
alamb, line 544: Can we please add a docstring that explains what this does so future readers don't have to look at the code?

I also wonder if calling it `set_delete_predicates` might be a more standard name?

querier/src/namespace.rs
```diff
@@ -438,6 +450,160 @@ impl QuerierNamespace {
...
+        // parse predicates and lookup table names in advance
+        // table name -> (sequencer_id -> list of predicates)
+        let mut predicate_cache = self.predicate_cache.lock().await;
+        let mut predicates_by_table_and_sequencer: HashMap<_, HashMap<_, Vec<_>>> =
+            HashMap::with_capacity(tombstones_by_table_and_sequencer.len());
+        for (table_id, tombstones_by_sequencer) in tombstones_by_table_and_sequencer {
+            let table_name = self.catalog_cache.table_name(table_id).await;
+            let mut predicates_by_sequencer = HashMap::with_capacity(tombstones_by_sequencer.len());
+            for (sequencer_id, mut tombstones) in tombstones_by_sequencer {
+                // sort tombstones by ID so that predicate lists are stable
```
alamb, line 488: I wonder if the the tombstones should be sorted by creation time instead, so that deletes that were applied earlier are applied before deletes that came after?

querier/src/namespace.rs
```diff
@@ -438,6 +450,160 @@ impl QuerierNamespace {
...
+                            .expect("malformed partition key")
+                            .0
+                            .parse()
+                            .expect("malformed partition key"),
+                    );
+
+                    let predicates = match predicates_by_sequencer.get(&sequencer_id) {
+                        Some(predicates) => predicates,
+                        None => {
+                            // don't skip modification since we might need to remove delete predicates from chunks
```
alamb, line 563: the sooner we get end to end tests of this logic the better I will feel. These types of comments make my internal "race detector" spider sense go all crazy

alamb (approved):

querier/src/namespace.rs
crepererum: Technically this cannot even happen at the moment since the catalog doesn't offer any way to delete tombstones (they can only be marked as processed for certain files) but I thought we might wanna clean up tombstones at some point or use a smarter but more complex query that doesn't pull fully-processed tombstones.

querier/src/namespace.rs
crepererum: Does this ever make a difference? I mean deletes are purely subtractive.

db/src/catalog/chunk.rs
crepererum: done

crepererum:
Merging this now, if there are open issues/question we can solve this in a follow-up PR.

querier/src/namespace.rs
alamb: Yeah I don't have any specific suggestion or issue here, just a vague feeling of unease. I should probably keep it to myself :)

querier/src/namespace.rs
alamb: I can't come up with a problem scenario off the top of my head

querier/src/namespace.rs
crepererum: It's OK. It's technically untested code, so we could also just stick `unimplemented!` in there.

NGA-TRAN (commented):
Sorry I reviewed this yesterday but forgot to click the final send button

querier/src/namespace.rs
NGA-TRAN: In the compaction, there is a step to remove tombstones that are fully processed. 
https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#
"Remove fully processed tombstones: [TICKET #3953](https://github.com/influxdata/influxdb_iox/issues/3953) for each tombstone in the tombstone_list created in step 4, check if they are fully processed (see the original architecture document for how). If so., remove  it and its corresponding processed tombstones in one transaction."

PR #3964, opened 2022-03-07
build: propagate feature flags to influxdb_ioxd

domodwyer:
This change fixes the optionality of the jemalloc allocator and makes it possible to compile IOx _without_ using `jemalloc`.

---

* build: propagate feature flags to influxdb_ioxd (aaec1c782)

      When you did a "cargo build --no-default-features" before this commit, the
      influxdb_iox crate was compiled without the default features
      (namely jemalloc stuff) but the influxdb_ioxd crate was compiled with jemalloc
      (it was not possible to compile without jemalloc support).

      This commit fixes propagation of the jemalloc flag
      ("jemalloc_replacing_malloc") into the influxdb_ioxd crate.


lukebond (approved):

alamb:
Maybe it is worth a change to the CI to ensure this doesn't get broken on you (again)

PR #3971, opened 2022-03-07
refactor: Remove NG dependencies on iox_object_store

carols10cents:
And only use the object_store crate directly.

80% of the iox_object_store crate is logic that isn't relevant to NG, so rather than special-casing NG ParquetFilePaths in iox_object_store and having to create "dummy" `IoxObjectStore` instances, make a new crate, parquet_file2, that shares some code with the parquet_file crate but doesn't use `iox_object_store` at all.

This also separates some logic in `parquet_file` that was only relevant to NG. This sets us up for an easy deletion when we're ready, by moving the few re-exported types from `parquet_file` to `parquet_file2` and then deleting all remaining code in the `parquet_file` crate (and deleting the `iox_object_store` crate entirely as-is).

The coupling of OG <-> NG is now moved a bit into the `db` crate, which has a new chunk type and a bit of copy-pasted code to treat parquet files from OG or NG the same-- I'm eyeing that separation for extraction soon.


db/src/catalog/chunk.rs
```diff
@@ -386,6 +404,38 @@ impl CatalogChunk {
         chunk
     }
 
+    /// Creates a new chunk that is only registered via an object store reference (= only exists in
+    /// parquet) as created by the ingester.
+    pub fn new_ingested(
```
alamb, line 409: FYI @crepererum  -- this may affect your work to untangle querier and `db`

db/src/catalog/chunk.rs
```diff
@@ -178,6 +178,20 @@ pub enum ChunkStage {
...
+    /// Chunk cannot receive new data. It is persisted.
+    Ingested {
+        /// Metadata (statistics, schema) about this chunk
+        meta: Arc<ChunkMetadata>,
+
+        /// Parquet chunk that lives immutable within the object store.
+        parquet: Arc<parquet_file2::chunk::ParquetChunk>,
+
+        /// In-memory version of the parquet data.
+        read_buffer: Option<Arc<RBChunk>>,
```
alamb, line 193: I am not sure an ingested chunk can ever be in the ReadBuffer -- I know we will eventually want to put it there but for now having the code imply the `read_buffer` is even possible (when I think it is always `None`) may be more confusing than it is worth

db/src/catalog/chunk.rs
```diff
@@ -670,6 +760,14 @@ impl CatalogChunk {
                     vec![]
                 }
             }
+            ChunkStage::Ingested { read_buffer, .. } => {
+                if let Some(read_buffer) = &read_buffer {
+                    read_buffer.column_sizes()
+                } else {
+                    // TODO parquet statistics
```
alamb, line 767: this is probably important to hook up, right?

alamb (commented):
The basic idea makes (a lot) of sense to me. Thank you @carols10cents 

I think @crepererum  should also review this tomorrow if possible as I think he is thinking about how the ingester / querier use the existing code in `db`

db/src/catalog/chunk.rs
```diff
@@ -178,6 +178,20 @@ pub enum ChunkStage {
         /// In-memory version of the parquet data.
         read_buffer: Option<Arc<RBChunk>>,
     },
+
+    /// Chunk in a _persisted_ stage as created by the ingester.
+    ///
+    /// Chunk cannot receive new data. It is persisted.
+    Ingested {
```
crepererum, line 185: "Ingested" doesn't really capture the whole story though. The file might also have been compacted. On the other hand it's kinda name bikeshedding because using `db` is somewhat an intermediate solution anyways.

crepererum (commented):
TBH I'm not sure this is any cleaner than before. We now have even more NG-OG interaction (all the changes in `db` which BTW are going to be thrown away soon anyways) and a non/half-tested new crate that replicates a lot of functionality of an existing one, only because the OG and NG parquet chunks have 1 difference: the path. I think if we don't wanna depend on `iox_object_store` we could also loosen the type signatures in `parquet_file` to accept something like `Either<ParquetFilePath, ParquetFilePath2>` or something.

PR #3972, opened 2022-03-08
feat: add compaction_level and created_at to parquet_file

### LINKED ISSUE #3945: Ingester: Add 2 fields level and creation_time into PARQUET_FILE catalog table
# Rationale
Per new requirements in [NG Compactor](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#), we need to add 2 fields, `level` and `creation_timestamp` into catalog table PARQUET_FILE.

When a parquet file is created in the Ingester, if its size > MAX_COMPACT_SIZE (configurable value during deployment and can be changed later), its `level` will be 1 because we do not need to compact it anymore. Otherwise, its `level` will be 0. The file’s `creation_time` is the current time and should be matched with its corresponding [IoxMetadata's creation_timestamp](https://github.com/influxdata/influxdb_iox/blob/ea4a245697c690b1a3a94c6d98e36b0766cfc898/parquet_file/src/metadata.rs#L511)

pauldix:
Adds compaction_level and created_at to parquet_file. The created_at time is set on create automatically by the catalog (either Postgres or the Mem mock).

This touched a bunch of files because of the addition of a time provider to the mem catalog, but should otherwise be pretty straight forward.

Closes #3945


data_types2/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 description = "Shared data types in the Iox NG architecture"
 
 [dependencies]
+chrono = { version = "0.4"}
```
crepererum, line 8: Wasn't the consensus a while a ago that we use our own `time` everywhere? It has a `Time` type.

data_types2/src/lib.rs
```diff
@@ -648,8 +649,15 @@ pub struct ParquetFile {
     pub parquet_metadata: Vec<u8>,
     /// the number of rows of data in this file
     pub row_count: i64,
+    /// the compaction level of the file
+    pub compaction_level: i16,
```
crepererum, line 653: What's a compaction level? Does this count how often the data passed throw compaction? Can data from different levels be compacted together?

data_types2/Cargo.toml
NGA-TRAN: Question for @crepererum : We implemented `TimeProvider` for `SystemProvider` so it is fine to use SystemProvider to provide time, right?

crepererum: You shouldn't use `SystemProvider` directly but always `Arc<dyn TimeProvider>`, because this will later on safe us soo much time if we wanna mock this side effect (by replacing `SystemProvider` w/ `MockProvider`). So the goal is that all calls to `now()` (or similar functions asking for the current time) go throw the `TimeProvider`.

data_types2/src/lib.rs
crepererum: It would be nice if (in this PR or at a later point) this would be clear from just reading the source code or "nearby" docs. Nothing against a design doc, but the moment the code is created (i.e. design is set in stone) it would be helpful to have the docs as well.

NGA-TRAN (commented):
In general it looks good. Only tow things to consider: which time we use and whether we want to keep the `created_at` consistent with the actual `creation_timestamp` of the parquet file in OS

data_types2/Cargo.toml
pauldix: sqlx uses the chrono type to map timestamp in Postgres.

pauldix: I can see if there's a way to use another type. Or I could just use a bigint in Postgres

iox_catalog/migrations/20220307201050_add_parquet_level_and_created_at.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE
+    IF EXISTS parquet_file
+    ADD
+    COLUMN compaction_level smallint NOT NULL DEFAULT 0,
+    ADD
+    COLUMN created_at timestamp WITH TIME ZONE NOT NULL DEFAULT NOW();
```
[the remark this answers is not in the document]
crepererum: Another point: can we please use integers for timestamps in postgres, NOT the PG-native type? I think converting timestamps (including min/max values) is quite a pain.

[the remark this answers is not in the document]
pauldix: I could convert this to just be a bigint like the others and not auto assign it and instead accept it as a parameter on create. Thoughts @crepererum?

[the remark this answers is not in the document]
crepererum: > I could convert this to just be a bigint like the others and not auto assign it and instead accept it as a parameter on create. Thoughts @crepererum?

That would be perfect, also passing it in would be helpful for testing. If you like this also enables us to set the timestamp to the time we've persisted the file to OS (instead of when we announced the file to the catalog)

data_types2/Cargo.toml
crepererum: Let's go for bigint then (also see discussion below regarding timestamps in general).

data_types2/src/lib.rs
carols10cents: When I was reading the design doc, it seemed to me that "compaction level" was more like "should be considered for potential compaction (or not)". I'm not sure of a shorter way to convey that (or if I'm understanding correctly).

> There are only 2 values for level: 0 for recent and 1 for non-recent.

Is there any possibility that there will be more levels in the future? If not, could this be a boolean value instead of an integer? Then could it be called `consider` or `ignore` something along those lines?

NGA-TRAN: With the current design, we only need 2 values. However, we will consider both of them (level 0 first and then level 1 if they are overlapped with level 0) so the `ignore` word may be not that appropriate.

pauldix: I'd like to leave open the possibility for other levels in the future, but for now it's either compacted or not. Although level 1 files will frequently have to be compacted because of new data or tombstones.

pauldix:
Ok, I updated it to just use a timestamp passed in during create which ends up removing a ton of stuff since the mem catalog no longer needs a time provider. I think it should be good to go now.

NGA-TRAN (approved):
LGTM

Issue #3974, opened 2022-03-08
NG Querier: Move aways from `db`


crepererum:
Using `db` to execute queriers on the new architecture was nice to proof that data in NG can be queried, but has several draw backs:

1. it pulls in the whole OG `db` stack including all the data structures that are no longer relevant
2. we cannot inspect queries "life" so we rely on a somewhat naive sync-loop to keep `db` and `iox_catalog` in sync

This issue is about replacing `db` with `querier`-specific data structures with our own, but potentially keeping the sync loop for now (if that's easier).

alamb:
In my view, the rationale for this work is that it is more or less "required" technically to add features to the querier that will be hard in the "populate Db" type mode -- such as talking to the Ingester https://github.com/influxdata/influxdb_iox/issues/3988 and adding ReadBuffer caching https://github.com/influxdata/influxdb_iox/issues/3989

For parquet files, there is a 1:1 mapping between chunks in OG and NG. For data in the ingester, the mapping is much less clear to me. Without this work, we might have to do something like hack in an "Ingester" chunk into the OG as there is no direct equivalent

PR #3975, opened 2022-03-08
refactor: split `querier::namespace` into submodules

crepererum:
This makes it easier to see what's required to support the query
interface.

Ref #3974


alamb (approved):

NGA-TRAN (commented):
👍

PR #3976, opened 2022-03-08
refactor: remove `QueryDatabase::partition_addr`

crepererum:
- This was not actually used by the query engine.
- The query engine doesn't have a concept of a "partition", it only
  cares about chunks.
- Unbound access to all partitions in the database is quite expensive
  (esp. on NG).

Ref #3974.


alamb (approved):
nice cleanup

NGA-TRAN (commented):
I really like the fact that you start seeing what actually need to build a plan and execute it for a query

PR #3977, opened 2022-03-08
refactor: remove `QueryDatabase::chunk_summaries`

crepererum:
- This is not used by the query engine at all.
- The query engine should not care about ALL chunks but only about the
  chunks it gets via `QueryDatabase::chunks` (which includes a table
  name and a predicate).
- All other users of that API are NOT really query-related.

Ref #3974.


tustvold (approved):
I think this is an orphan and can go

PR #3979, opened 2022-03-08
(WIP) feat: Add Routes non templated service router

alamb:
This ports the code from https://github.com/hyperium/tonic/pull/830 and gets it to compile outside the tonic tree, in case that is helpful for dynamically creating gRPC router


influxdb_ioxd/src/rpc/routes.rs
```diff
@@ -0,0 +1,120 @@
...
+///
+/// The usecase for this is to dynamically add handlers to a tonic Server.
+///
+/// The Router in tonic 0.6 is generic so that the type of the handler
+/// is part of the type of the router, meaning you can't pass around a
+/// generic router without knowing exactly the type of the last
+/// handler (vomit!)
+///
+/// TODO remove axum dependency, all the code overlap
+#[derive(Debug, Default, Clone)]
```
alamb, line 25:
```suggestion
/// Derived from https://github.com/hyperium/tonic/pull/830
#[derive(Debug, Default, Clone)]
```

alamb:
Update here is that the way the existing tonic router is instantiated I can't dynamically route within a service -- as the tonic router gets chained together with Ors based on the service name

But I now have another idea (I need to figure out how to wrap the service)

alamb:
🤮

Issue #3926, opened 2022-03-03
Proposal: Local all-in-one mode NG


alamb:
__Use case:__
* As a developer working on a feature such as query performance or a correctness bug
* I want to run / poke at a local instance of IOx NG in my development environment
* so that I can use the tools I feel most comfortable and productive with (e.g. debugger, modify code / rerun, etc) to be maximally productive

__Proposal:__
Allow users to create a fully functioning, single node, minimal IOx NG set of services. The cluster will require only the most minimal configuration options. Anything more sophisticated (for example, multiple routers, or queriers, etc) is out of scope and will be orchestrated by some other system (like our cloud deployment system for InfluxDB Cloud)

Specifically propose adding a command called `run2` after the similarly named `run` command in the current system This would be invoked like this (with reasonable defaults):

```shell
 influxdb_iox run2 --pg_dsn postgresql://localhost:5432/alamb  --object-store=file --data-dir=$HOME/.influxdb_iox
```

This will setup the following configuration (not at all configurable):
```text
┌─────────────┐  ┌─────────────┐   ┌─────────────┐   ┌─────────────┐
│   Router    │  │  Ingester   │   │   Querier   │   │  Compactor  │
└─────────────┘  └──────┬──────┘   └──────┬──────┘   └──────┬──────┘
       │                │                 │                 │       
       │                │                 │                 │       
       │                │                 │                 │       
       └────┬───────────┴─────┬───────────┴─────────┬───────┘       
            │                 │                     │               
            │                 │                     │               
         .──▼──.           .──▼──.               .──▼──.            
        (       )         (       )             (       )           
        │`─────'│         │`─────'│             │`─────'│           
        │       │         │       │             │       │           
        │.─────.│         │.─────.│             │.─────.│           
        (       )         (       )             (       )           
         `─────'           `─────'               `─────'            
                                                                    
        Existing         Object Store        kafka / redpanda       
        Postgres        (file, mem, or                   
       (started by      pre-existing)         (file, mem, or        
          user)        configured with        pre-existing)         
                        --object-store       configured with        
```

At some point in the future, (but not part of this project) we may provide a way to run without a pre-existing postgres installation.

__Alternatives considered:__

It may seem tempting to use some other technology like `gopm` or `docker compose` to manage the multiple processes required for IOx NG. However, this is non ideal , as @lukebond  noted, as then we have another tool to learn and another way to maintain configuration

[...]

PR #3981, opened 2022-03-08
(WIP) refactor: Sketch out dynamic tonic service builder interface

alamb:
# Backstory

I am trying to create an IOx "All In One Mode" -- see #3926 . Ideally I am hoping to run the gRPC services from the router, ingester, compactor and querier on the same listener.

The code structure I am hoping for is like the following (so the Compactor, Ingester, etc encapsulate the services they need and can add them on demand)

Router Mode:
```rust
let iox_router2 = create_router2_service(...)

let builder = RpcBuilder::new();
let builder = iox_router2.add_services(builder)
builder.serve()
```

And in All in One Mode:
```rust
let iox_router2 = create_router2_service(...)
let ingester = create_ingester(...)
let querier = create_querier(...)

let builder = RpcBuilder::new();
let builder = iox_router2.add_services(builder)
let builder = ingester.add_services(builder)
let builder = querier.add_services(builder)
builder.serve()
```

Simple, right?  Maybe.... but I can't figure it out as the `tonic::Server` or tonic::Router` are both generic so there is no way I have found to implement the builder as shown.

This PR adds the interface for the builder and shows it can compile, but I can't figure out how to actually implement it. 

I also put up a PR with code from the unreleased version of tonic here: https://github.com/influxdata/influxdb_iox/pull/3979 in case that is helpful


influxdb_ioxd/src/rpc.rs
```diff
@@ -19,158 +25,82 @@ pub fn service_name<S: NamedService>(_: &S) -> &'static str {
     S::NAME
 }
 
-#[derive(Debug)]
-pub struct RpcBuilderInput {
```
alamb, line 23: This removed code shows how things currently work -- basically each `ServerType` impl uses macros to instantiate the common services so that the actual types never leave the instantiation sites...

influxdb_ioxd/src/server_type/test.rs
```diff
@@ -81,16 +81,14 @@ impl ServerType for TestServerType {
         })
     }
 
-    async fn server_grpc(self: Arc<Self>, builder_input: RpcBuilderInput) -> Result<(), RpcError> {
+    async fn add_grpc_services(
+        self: Arc<Self>,
+        builder: RpcBuilder2,
+    ) -> Result<RpcBuilder2, RpcError> {
         match self.test_action {
             TestAction::PanicInGrpcWorker => panic!("Test panic in gRPC worker"),
```
alamb, line 89: this particular service is probably not quite right anymore, but it is good enough to show the problem

influxdb_ioxd/src/rpc.rs
```diff
@@ -19,158 +25,82 @@ pub fn service_name<S: NamedService>(_: &S) -> &'static str {
...
-            socket,
-            ..
-        } = $builder;
-
-        let stream = TcpListenerStream::new(socket);
-        inner
-            .serve_with_incoming_shutdown(stream, shutdown.cancelled())
-            .await?;
-    }};
+impl RpcBuilder2 {
```
alamb, line 39: This is all I want 🤪  a simple "builder" whose type doesn't change when services are added to it!

alamb:
This would be so much nicer, but I couldn't make it work

PR #3987, opened 2022-03-08
refactor: move parquet chunk's new and decode to parquet_file crate

NGA-TRAN:
refactor so I can use them in Compactor too


NGA-TRAN:
FYI @crepererum

alamb (approved):
Seems fine to me -- while this is mixing more code from OG and NG together in the parquet_file crate I think it is inevitable as we now scurry to switch from OG to NG as fast as possible

PR #3990, opened 2022-03-09
refactor: drop trace log verbosity

domodwyer:
This is very verbose - pages and pages of scrollback! Maybe it was `INFO` for a reason though?

---

* refactor: drop trace log verbosity (8af9330cd)

      This is very shouty, logging one INFO per HTTP request - the router logs are
      almost all this one line repeated.


crepererum:
I think the reason was that this doesn't normally happen in prod because there we don't force traces to be rooted in IOx itself but instead "the cloud" creates it for us. Within the test bench we are kinda lazy and the data generator doesn't emit trace roots. I think we should rather modify the data generator to produce proper trace roots and forward them to jaeger.

domodwyer:
Is there any value in surfacing that as a log message at INFO though? From what I can see it's emitted for `/health` and `/metrics` hits too.

crepererum:
> Is there any value in surfacing that as a log message at INFO though? From what I can see it's emitted for `/health` and `/metrics` hits too.

I think the intention was "we're emitting traces that are not really part of environment". However I think we could at least lower the verbosity to `debug`.

CC @tustvold who (I think) originally wrote this piece of code.

tustvold:
It was actually @alamb 😅 I think lowering to debug makes sense, as it could be useful for debugging a production issue. Trace is stripped out of release builds and so we would lose this ability

Issue #3985, opened 2022-03-08
NG Querier: Proper Cache Layer


crepererum:
Add a proper cache layer that shields the postgres catalog from "each flight/InfluxRPC query results in multiple Postgres queries". Requirements are:

- [x] only cache data that is needed
- [x] work w/ async codebase #3991
- [ ] #4616
- [x] be generic (mostly just take a function that is executed on cache misses) #3991
- [x] TTL #4027
- [ ] #4618
- [ ] #4617
- [x] one cache per topic/object type (no giant cache data structure) #3991
- [x] one cached object may answer multiple requests (e.g. a cached table can answer both ID->name as well as name->ID mappings) #3991
- [x] replaces the current `querier::cache` #3991

~Might use the `cache_loader_async` crate (or not if it doesn't fit).~ Didn't fit, because:

- tests esp. regarding cancellation and panics were basically missing
- complex interface
- we would need to replace/wrap many features (e.g. for metrics, integration with `TimeProvider`, size-based LRU) so there wasn't much left (the semi-tested core logic at most)

alamb:
I am not sure being generic is a requirement for what is worth

Also it is worth thinking about:
1. How will locally caching data in ReadBuffer and Parquet be connected to this system

crepererum:
> I am not sure being generic is a requirement for what is worth

I just don't wanna write the infrastructure stuff (metrics, management controls, LRU/TTL handling) multiple times. I just want something that is consistent and well tested, so I don't need to bother with it all too much when reasoning about the querier.

> How will locally caching data in ReadBuffer and Parquet be connected to this system

IMHO that's just another KV cache. "Cache" per se doesn't have to be "cached in RAM", it can also mean "this is a cached parquet file on disk" (I don't wanna write a caching framework that deals with storage backends or serialization though).

crepererum:
Closing this since the basic work is done and we can use the open tickets to track the concrete open items.

PR #3991, opened 2022-03-09
feat: querier test system, ground work

crepererum:
See #3985 for the motivation.

This introduces a cache system for the querier which can later be
extended to support the remaining features listed in #3985 (e.g.
metrics, LRU/TTL).

All current caches are wired up to go throw the new cache system. Once
we move away from (ab)using `db`, the set of caches will be different
but the system will remain.


querier/src/cache_system/broadcast.rs
```diff
@@ -0,0 +1,144 @@
+use observability_deps::tracing::warn;
```
crepererum, line 1: This file is copied from `rskafka` where we had a similar application.

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,430 @@
+use std::{collections::HashMap, future::Future, hash::Hash, sync::Arc};
```
crepererum, line 1: This is the actual cache system.

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+}
+
+/// Inner cache state that is usually guarded by a lock.
+///
+/// The state parts must be updated in a consistent manner, i.e. while using the same lock guard.
+#[derive(Debug)]
+struct CacheState<K, V> {
+    cached_entries: HashMap<K, V>,
```
crepererum, line 140: Currently we store everything in a "keep all" hash map. That's obviously not the desired end state. I'll introduce an interface for that bit (that also takes care of pruning entries) later, but I didn't want to blow up this PR any further.

querier/src/cache/mod.rs
```diff
@@ -0,0 +1,66 @@
+use backoff::BackoffConfig;
```
crepererum, line 1: Everything in `querier::cache` is actually just moved around code. The caches existed beforehand but in a giant struct and hand-managed hashmaps. They are not split apart. The tests are the same (just moved to the corresponding modules).

querier/src/cache_system/driver.rs
crepererum: @tustvold I could use your review here.

crepererum:
FYI: @tustvold proposed to just use a shared oneshot channel instead of `BroadcastOnce`. This would also avoid the usage of the weird panic guard. I'll implement that tomorrow. This shouldn't affect the interface or the tests though.

tustvold:
Looks good to me, nice work. As you've articulated, whilst I like the idea of reusing BroadcastOnce, I think the fact we don't need peek() functionality and already have external state storage means we can do something simpler

querier/src/cache_system/broadcast.rs
crepererum: this file is gone now

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,430 @@
...
+            state,
+            k,
+            did_not_panic: false,
+        }
+    }
+
+    /// Flag that the loader did not panic.
+    ///
+    /// Dropping this guard after calling this function will NOT broadcast anything.
+    fn did_not_panic(&mut self) {
```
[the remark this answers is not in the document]
crepererum: this code is gone now

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,430 @@
...
+
+impl<K, V> Cache<K, V>
+where
+    K: Clone + Eq + Hash + std::fmt::Debug + Send + 'static,
+    V: Clone + std::fmt::Debug + Send + 'static,
+{
+    /// Create new, empty cache with given loader function.
+    pub fn new<T, F>(loader: T) -> Self
+    where
+        T: Fn(K) -> F + Send + Sync + 'static,
```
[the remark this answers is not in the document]
crepererum: done

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,430 @@
...
+
+        tokio::time::sleep(Duration::from_millis(10)).await;
+
+        assert_eq!(Arc::strong_count(&loader), 1);
+    }
+
+    fn setup() -> (Arc<Cache<u8, String>>, Arc<Loader>) {
+        let loader = Arc::new(Loader::default());
+        let loader_captured = Arc::clone(&loader);
+        let cache = Arc::new(Cache::new(move |key| {
```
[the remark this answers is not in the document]
crepererum: done

crepererum:
@tustvold I've removed `BroadcastOnce`, which resulted in way simpler and far less code. I've also introduced a `Loader` trait. So PTAL :slightly_smiling_face:

tustvold (approved):

querier/src/cache_system/driver.rs
```diff
@@ -0,0 +1,391 @@
...
+                    // If we panic here then `tx` will be dropped and the receivers will be notified.
+                    let v = loader.load(k_for_loader).await;
+
+                    // broadcast result
+                    // It's OK if the receiver side is gone. This might happen during shutdown
+                    tx.send(v.clone()).ok();
+
+                    // remove "running" state and store result
+                    let mut state = state_captured.lock();
+                    drop(
```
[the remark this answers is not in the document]
crepererum: I'll add a comment. In short: to convince `rustc` that we don't need the shared future that is returned by `remove`.

PR #3992, opened 2022-03-09
refactor(ingester): eliminate unnecessary lock

domodwyer:
The `LifecycleManager` is designed such that only a single call to `maybe_persist()` can be running at any one time, and a `persist_running` lock was added with the intent to enforce this. This PR removes the lock, and instead uses the type system to enforce this serialisation property.

---

* refactor: eliminate unnecessary lock (874be0097)

      This commit splits the API of the LifecycleManager into two:

          * LifecycleManager: singleton responsible for evaluating partitions
             and running persist tasks.

          * LifecycleHandle: a handle for each sequencer ingester(s) to update
             the global LifecycleManager state when applying ops.

      This keeps the accessible API & responsibilities of each caller distinct and
      allows us to leverage the type system to enforce linearisation of calls to
      LifecycleManager::maybe_persist() without resorting to an
      (unnecessary) mutex guard for serialisation.


ingester/src/lifecycle.rs
```diff
@@ -134,62 +207,27 @@ impl LifecycleManager {
...
+        }
     }
 
     /// This will persist any partitions that are over their size or age thresholds and
     /// persist as many partitions as necessary (largest first) to get below the memory threshold.
     /// The persist operations are spawned in new tasks and run at the same time, but the
     /// function waits for all to return before completing.
-    pub async fn maybe_persist<P: Persister>(&self, persister: &Arc<P>) {
-        // ensure that this is only running one at a time
-        self.persist_running.lock().await;
```
domodwyer, line 191: This lock is what sent me down this PR path - the guard is not held, and therefore does not properly serialise calls.

Fortunately the caller hierarchy for `LifecycleManager::maybe_persist()` is pretty simple and would never lead to two concurrent `maybe_persist()` calls anyway, so this lock was redundant and was easily modified to have the compiler enforce linearisation by changing `maybe_persist(&self)` to take `&mut self` instead.

ingester/src/lifecycle.rs
```diff
@@ -18,16 +18,89 @@ use time::{Time, TimeProvider};
...
-/// The lifecycle manager keeps track of the size and age of partitions across all sequencers.
-/// It triggers persistence based on keeping total memory usage around a set amount while
-/// ensuring that partitions don't get too old or large before being persisted.
+/// A handle for sequencer consumers to interact with the global
+/// [`LifecycleManager`] instance.
+///
+/// This handle presents an API suitable for ingester tasks to query and update
+/// the [`LifecycleManager`] state.
+#[derive(Debug, Clone)]
+pub struct LifecycleHandle {
```
domodwyer, line 27: This is the real change - the tasks that are reading from the sequencers and applying ops now hold a `LifecycleHandle` rather than a reference to the `LifecycleManager`, through which they push state updates with `log_write()` as they did before the change.

crepererum:
I'll have a look after lunch, I'm definitely interested to see how you moved the lock into the type system.

alamb:
I think we may want to ask @tustvold  to take a peek at this

ingester/src/lifecycle.rs
```diff
@@ -18,16 +18,89 @@ use time::{Time, TimeProvider};
...
+    time_provider: Arc<dyn TimeProvider>,
+
+    config: Arc<LifecycleConfig>,
+
+    /// The state shared with the [`LifecycleManager`].
+    state: Arc<Mutex<LifecycleState>>,
```
crepererum, line 33: Ah, so both the manager and the handle actually use a lock around the state, it's just that the manager offers all the things that must not be run concurrently (i.e. `maybe_persist`) and ensures that by using `&mut self` for these operations, while the handle offers operations that can be triggered concurrently (and uses `&self` for them). So it's somewhat a "internal unsafe, but external safe API" approach. I like it (used something along these lines for the `iox_catalog` transaction scaffolding as well).

crepererum (approved):

NGA-TRAN (approved):

PR #3993, opened 2022-03-09
fix: Remove data_types as a dependency of ng compactor

carols10cents:
All NG crates should only be depending on `data_types2`. `compactor` isn't actually using anything from `data_types`.


alamb (commented):
🎉

PR #3997, opened 2022-03-09
WIP: POC of using datafusion-proto

carols10cents:
This is the start of using the new datafusion-proto crate to do the protobuf conversions for datafusion types... but I don't think pbjson-build likes what I'm doing, or I haven't told pbjson-build the right things yet... Can you help @tustvold ?


carols10cents:
oh dear and `buf lint` doesn't like datafusion's proto... and this doesn't include a CI check to make sure the datafusion proto file is in sync with the one in the datafusion-proto crate

generated_types/build.rs
```diff
@@ -118,6 +119,7 @@ fn generate_grpc_types(root: &Path) -> Result<()> {
 
     pbjson_build::Builder::new()
         .register_descriptors(&descriptor_set)?
+        .extern_path(".datafusion", "::datafusion::protobuf")
```
tustvold, line 122: I can look into more tomorrow morning, but unless datafusion is using pbjson to generate serialize impls for the proto, this would cause issues if you then tried to use the DF types in a message with pbjson enabled. It effectively is telling pbjson to ignore the protos for DF which will then error if it needs them for some reason

carols10cents: This was my attempt at getting pbjson to work... it doesn't work without this line either, both ways I get orphan rule problems:

```
error[E0277]: the trait bound `datafusion::protobuf::Column: Serialize` is not satisfied
    --> /home/rust/project/target/debug/build/generated_types-83c4971067478a96/out/influxdata.iox.ingester.v1.serde.rs:20:48
     |
20   |             struct_ser.serialize_field("left", v)?;
     |                        ---------------         ^ the trait `Serialize` is not implemented for `datafusion::protobuf::Column`
     |                        |
     |                        required by a bound introduced by this call
     |
note: required by a bound in `longrunning::_::_serde::ser::SerializeStruct::serialize_field`
    --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/serde-1.0.136/src/ser/mod.rs:1899:12
     |
1899 |         T: Serialize;
     |            ^^^^^^^^^ required by this bound in `longrunning::_::_serde::ser::SerializeStruct::serialize_field`

```

tustvold: The line is correct, you can't define these protons outside of DF, you will also need to disable pbjson for any messages in IOx that use the DF messages (or add pbjson to DF)

carols10cents:
@alamb We need to decide whether we want to either: 

- disable pbjson for any proto messages containing datafusion exprs
- Add pbjson to datafusion-proto

Do you have thoughts on that?

alamb:
> Do you have thoughts on that?

@carols10cents  
My personal thoughts are that we should treat datafusion `Exprs` as opaque to IOx's native serialization (aka store them as `bytes` in our proto files). 

Rationale:
1. While it is the case that today both IOx and DataFusion use protobuf to encode the expressions, I don't think anything in IOx needs to assume / expect that.
2.  Treating DataFusion exprs as opaque would allow the implementation to change (e.g. to use substrait.io or something) without IOx having to change

If we went with

alamb:
> And also decide whether we want to ignore the datafusion .proto file from buf lint analysis or get it to conform, 

Making it conform seems like a good idea

> and implement a check in CI that the .proto file we've copied into iox matches the .proto file shipped with the version of datafusion-proto we're using

👍

carols10cents:
> My personal thoughts are that we should treat datafusion `Exprs` as opaque to IOx's native serialization (aka store them as `bytes` in our proto files).
> 
> Rationale:
> 
>     1. While it is the case that today both IOx and DataFusion use protobuf to encode the expressions, I don't think anything in IOx needs to assume / expect that.
> 
>     2. Treating DataFusion exprs as opaque would allow the implementation to change (e.g. to use substrait.io or something) without IOx having to change

Oooohhh well if that's the case then we might not need the .proto file over here!

alamb:
@tustvold and @crepererum  and @mkmik  -- do you have 😱  reactions to this proposal: https://github.com/influxdata/influxdb_iox/pull/3997#issuecomment-1077949298 (aka encode Expr's opaquely?)

tustvold:
My initial thought is it will make calling or introspecting the API using any non-custom tooling painful and so I'm not a massive fan tbh...

If the fear is allowing future evolution of the payloads, a oneof seems like it would be a better fit.

Unlike an opaque token, the content of the predicate is part of the API contract and critically something the user must construct. This seems at odds with using an opaque serialisation...

alamb:
I don't think users construct datafusion `Exprs` - in this case it is part of serializing a filter predicate to send to the `ingester` (aka it is a message not stored on disk).

I don't have a strong desire for opaque things in the rpc format, but I also don't think we need to spend a huge amount of plumbing to provide transparent Expr encoding if it is a pain. 

As for the choice between these options:

> * disable pbjson for any proto messages containing datafusion exprs
> * Add pbjson to datafusion-proto


I would say "whatever is easier"

alamb:
> If the fear is allowing future evolution of the payloads, a oneof seems like it would be a better fit. You're going to need some way to distinguish payload types, you might as well use the built in functionality.


My thinking here is that all evolution / handling backwards compatibility would happen in datafusion and IOx can remain oblivious to it... Though of course that assumes DataFusion will implement backwards compatible serialization, which may not be the case 🤔

tustvold:
Definitely, as this is an internal API we should avoid spending unnecessary amounts of time making it "perfect". My only comment was that we will at some point need to debug these calls, and an opaque type will complicate both making ad-hoc requests with tools like grpcurl and inspecting protocol traffic using tools like wireshark. Definitely not a deal breaker, I'm happy with whatever the group decides 😀

PR #3998, opened 2022-03-09
chore: Run dependabot more frequently

alamb:
# Rationale
I would like dependabot to do the annoying work of updating all the `Cargo.toml` files that have arrow versions for me, but since I like to get arrow updates into IOx a day or two after they are released, a weekly schedule for dependabot dosn't help me

# Changes
Run Dependabot daily to pick up dependency changes


tustvold (approved):
What is the downside of this? Why wouldn't you set it to run every hour or minute...

alamb:
> What is the downside of this? Why wouldn't you set it to run every hour or minute...

I think `daily` is the most frequent that is supported: https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleinterval

😆

Issue #3999, opened 2022-03-09
NG Compactor: Split compacted data into 2 files


NGA-TRAN:
# Rationale

This ticket implements step 6.b.ii of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). The input is the compacted data and the output is 2 files split based on the given split_rules. Something like

```rust
fn split compact(
    CompactedData, 
    split_rules,  // needs to decide what the rules are
) -> Result<out1, out2)>

```

alamb:
First of all, when the partition is "cold" (aka not getting writes anymore) I think we should compact the entire thing into a single file (subject to the max file size limit)

For "hot" partitions that are still getting writes, I agree with @NGA-TRAN  that approach 1 sounds good 👍  because:
1. It seems like it is less work (we can reuse the existing query plans)
2. It wouldn't require much rework if we eventually decide we need something like approach 2

I think the whole point of splitting when compacting a "hot" partition (that is still getting data) is to make sure the larger part doesn't overlap with new data that will come in. A static 10/90 split based on timestamp seems reasonable at first, though of course we will be able to find pathological counter examples for any policy that is chosen,

NGA-TRAN:
@alamb Good point regarding cold and hot partition. Let assume we do this for hot partition. I have added 2 tasks in the `Improvement` section of parent ticket #3904 to handle cold partition and hierarchical partition if needed

Issue #3949, opened 2022-03-04
NG Compactor: Group overlapped files for a given set of files


NGA-TRAN:
# Rationale
This ticket implements step 4 of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). The input is a list of [parquet files](https://github.com/influxdata/influxdb_iox/blob/6ba5e518976a1feadb09b27b74967ac57b0e9971/data_types2/src/lib.rs#L624) and the output is a list of list of overlapped parquet files. Something like

```rust
fn group_overlapped_files(&[ParquetFile]) -> Result(Vec<Vec<ParquetFile>>)
```

Issue #3946, opened 2022-03-04
NG Compactor: Read catalog for level-0 and level-1 parquet files


NGA-TRAN:
# Rationale
The ticket is to implement step 1 and step 3 of the [Loop Find & Compact Files in the NG Compactor design](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx).

For step 1: build a function like this 
```rust
     get_level_0(sequencer_id) ->  Result<Vec<ParquetFile>>
```

Step 3:
```rust
   get_level_1(sequencer_id, table_id, partition_id) -> Result<Vec<ParquetFile>>
```

carols10cents:
Going to start this, building atop https://github.com/influxdata/influxdb_iox/pull/3972

carols10cents:
@NGA-TRAN For step 3, for this aspect of level 1 files:

```
( ( min_time <= level-0-file.min_time AND 
    max_time >= level-0-file.min_time) )  OR
( ( min_time > level-0-file.min_time AND 
     min_time <= level-0-file.max_time) ) 
```

and looking at this code in compact.rs: https://github.com/influxdata/influxdb_iox/blob/95f461b90769c48cc920cba9e13c70bc8e7ed87d/compactor/src/compact.rs#L82-L86

there are multiple `level-0-file`s to consider here when trying to find level 1 files. So is the `get_level_1` function supposed to:

* Option 1: Use the min of all level 0 file min times and the max of all level 0 file max times, to find all level 1 files that overlap with any time of the outer bounds of all level 0 files, but some level 1 files might not actually overlap with any
* Option 2: Iterate through each level 0 file and request the level 1 files that overlap with each one, which means calling `get_level_1` N times where N=number of level 0 files, and this might put duplicates into the list of files to consider if a level 1 file overlaps with more than one level 0 file
* Something else?

Thanks!

NGA-TRAN:
@carols10cents I think we will go with Option 1 with  extra notes:
1. Since only files in the same partition (= same sequence_id, table_id, partition_id) may overlap, the option 1 should be repeated for each same-partition group of level-0 files.
2. After you get level-1 files in option 1, you may want to go over them and remove the ones that are not really overlapped with any level 0. Since the compaction is not cheap, we do not want to compact data that are not really needed.

NGA-TRAN:
[...]

PR #4001, opened 2022-03-09
feat: Implement the compact function for NG Compactor

### LINKED ISSUE #3907: NG Compactor: Compact a given list of parquet files of a partition with a given of overlapped tombstones
# Rationale
This is a feature of #3904 that compact one or more parquet files of a given partition with a given list of tombstones

UPDATE: The design below is a bit out of date, the actual implementation will stick to the [Compactor design](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#) 

# Design

### Function signature:

```rust
fn compact_parquet_files(
     parquet_files: &[iox_catalog::interface::ParquetFile], 
     tombstones: &[iox_catalog::interface::Tomestone]
) -> Result<iox_catalog::interface::ParquetFile>
```

### Idea

- Object store data of the given `parquet_files` will be read into `ParquetChunks` (existing code). 
- [compact_plan](https://github.com/influxdata/influxdb_iox/blob/3d047073b9ac89dc69b39d7dc424d5dfa7c3d18a/query/src/frontend/reorg.rs#L112) will be used to compact those `ParquetChunks`. In order to do so, we need a wrapper as follows that implements `QueryChunk` and `QueryChunkMeta`
   ```rust
        struct QueryableParquetChunk {
             data: Arc<ParquetChunk>.
             delete_predicates: Vec<Arc<DeletePredicate>>,  // converted from the given tombstones and QueryChunk requires this
             table_name: String,  // QueryChunk requires this
        }
   ```

### Other needed work

1. Verify if all given parquet files belong to 1 partitions
2.  Sort the parquet files in their ascending `min_sequence_number` so the deduplication will work correctly. Note:
      i. Throw error if their sequence number ranges overlap (if this happens, there must be a bug in Ingester)
      ii. It is fine to have a gap between 2 ranges because those may be used for the sequence number of tombstones
3. Merge schema of all parquet files
4. Compute sort key of the schema. For NG architecture, since we do not have column statistics, the sort key will be lexicographical order  of tag names + time column
5. Then invoke `compact_plan` to build a plan and then executer it to return a stream of record batches
6. Persist the stream (reuse existing code done for NG Ingester)

NGA-TRAN:
Closes #3907

This PR compacts a given a list of overlapped files. Note that the compacted result is a list of record batches. The work to split them into 90/10 will be done in #3999


NGA-TRAN:
I am working on adding tests for this

compactor/src/compact.rs
```diff
@@ -167,47 +240,399 @@ impl Compactor {
         }
         tombstones
     }
-}
 
-#[derive(Debug, Clone, Copy, PartialEq, PartialOrd, Eq, Ord)]
-struct TablePartition {
```
NGA-TRAN, line 173: Many helper structs and functions in this file are moved to the same crate `utils.rs`

compactor/src/compact.rs
```diff
@@ -167,47 +240,399 @@ impl Compactor {
...
-    sequencer_id: SequencerId,
-    table_id: TableId,
-    partition_id: PartitionId,
-}
+    // Compact given files. Assumption tat the given files are overlaped in time
+    // If the assumption does not meet, we will spend time not to compact anything but put data together
+    async fn compact(
```
NGA-TRAN, line 246: This is the focus of this PR

compactor/src/query.rs
```diff
@@ -0,0 +1,216 @@
...
+        source: parquet_file::metadata::Error,
+    },
+}
+
+/// A specialized `Error` for Compactor's query errors
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// QueryableParquetChunk that implements QueryChunk and QueryMetaChunk for building query plan
+#[derive(Debug, Clone)]
+pub struct QueryableParquetChunk {
```
NGA-TRAN, line 40: In order to build a query plan on parquet chunks, we have to make each chunk a queryable one by implementing 2 traits, QueryChunk and QueryChunkMeta. This file is for that purpose.

The reason I do not want to use equivalent ones in the Querier because the Querier abstracts all chunks (MUB, RUB, OS) into a DbChunk who is then turned into a queryable one. It is many things about them the Compactor does not need to know.

ingester/src/compact.rs
```diff
@@ -40,71 +40,16 @@ pub enum Error {
...
+
+    #[snafu(display("Error computing min and max for record batches: {}", source))]
+    MinMax { source: query::util::Error },
 }
 
 /// A specialized `Error` for Ingester's Compact errors
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// Return min and max for column `time` of the given set of record batches
-pub fn compute_timenanosecond_min_max(batches: &[RecordBatch]) -> Result<(i64, i64)> {
```
NGA-TRAN, line 60: I have moved these functions to query crate to reuse them in the compactor

compactor/src/compact.rs
```diff
@@ -167,47 +240,399 @@ impl Compactor {
...
+                        sequencer_id_1: overlapped_files[0].data.sequencer_id,
+                        table_id_1: overlapped_files[0].data.table_id,
+                        partition_id_1: overlapped_files[0].data.partition_id,
+                        sequencer_id_2: file.data.sequencer_id,
+                        table_id_2: file.data.table_id,
+                        partition_id_2: file.data.partition_id,
+                    });
+                }
+            }
+        }
```
shepmaster, line 281: The current code checks the first element against itself and can be simplified

```suggestion
        if let Some((head, tail)) = overlapped_files.split_first() {
            for file in tail {
                let is_same = file.data.sequencer_id == head.data.sequencer_id
                    && file.data.table_id == head.data.table_id
                    && file.data.partition_id == head.data.partition_id;

                ensure!(is_same, NotSamePartitionSnafu {
                    sequencer_id_1: head.data.sequencer_id,
                    table_id_1: head.data.table_id,
                    partition_id_1: head.data.partition_id,
                    sequencer_id_2: file.data.sequencer_id,
                    table_id_2: file.data.table_id,
                    partition_id_2: file.data.partition_id,
                })
            }
        }
```

compactor/src/compact.rs
```diff
@@ -167,47 +240,399 @@ impl Compactor {
...
+        let mut min_sequence_number = query_chunks[0].min_sequence_number();
+        let mut max_sequence_number = query_chunks[0].max_sequence_number();
+        for c in &query_chunks {
+            if min_sequence_number > c.min_sequence_number() {
+                min_sequence_number = c.min_sequence_number();
+            }
+            if max_sequence_number < c.max_sequence_number() {
+                max_sequence_number = c.max_sequence_number();
+            }
+        }
```
shepmaster, line 305: Same idea as above, plus using the existing `min` and `max` functions

```suggestion
        let (head, tail) = query_chunks.split_first().unwrap();
        let mut min_sequence_number = head.min_sequence_number();
        let mut max_sequence_number = head.max_sequence_number();
        for c in tail {
            // use std::cmp::{min, max};
            min_sequence_number = min(min_sequence_number, c.min_sequence_number());
            max_sequence_number = max(max_sequence_number, c.max_sequence_number());
        }
```

compactor/src/utils.rs
```diff
@@ -0,0 +1,117 @@
...
+#[derive(Debug, Clone)]
+pub struct ParquetFileWithTombstone {
+    pub(crate) data: Arc<ParquetFile>,
+    pub(crate) tombstones: Vec<Tombstone>,
+}
+
+impl ParquetFileWithTombstone {
+    /// Return all tombstone ids
+    pub fn tombstone_ids(&self) -> HashSet<TombstoneId> {
+        self.tombstones.iter().map(|t| t.id).collect::<HashSet<_>>()
```
shepmaster, line 48: Type can be inferred from the return type of the function

```suggestion
        self.tombstones.iter().map(|t| t.id).collect()
```

compactor/src/query.rs
```diff
@@ -0,0 +1,216 @@
...
+    data: Arc<ParquetChunk>,                      // data of the parquet file
+    iox_metadata: Arc<IoxMetadata>,               // metadata of the parquet file
+    delete_predicates: Vec<Arc<DeletePredicate>>, // converted from tombstones
+    table_name: String,                           // needed to build query plan
+}
+
+impl QueryableParquetChunk {
+    /// Initialize a QueryableParquetChunk
+    pub fn new(
+        table_name: &str,
```
shepmaster, line 50: We accept a `&str` but then convert it into a `String`. Below, we accept a `String` for the table name, convert it into a reference, then throw away the original `String`. If this were `table_name: impl Into<String>` then you could transfer ownership of the `String` and avoid the extra allocation.

query/src/util.rs
```diff
@@ -198,6 +216,55 @@ impl<'a> ExprRewriter for MissingColumnsToNull<'a> {
...
+        }
+        if max_time < max {
+            max_time = max;
+        }
+    }
+    Ok((min_time, max_time))
+}
+
+/// Return min and max for column `time` in the given record batch
+pub fn compute_timenanosecond_min_max_for_one_record_bacth(
```
shepmaster, line 236:
```suggestion
pub fn compute_timenanosecond_min_max_for_one_record_batch(
```

query/src/util.rs
```diff
@@ -198,6 +216,55 @@ impl<'a> ExprRewriter for MissingColumnsToNull<'a> {
     }
 }
 
+/// Return min and max for column `time` of the given set of record batches
+pub fn compute_timenanosecond_min_max(batches: &[RecordBatch]) -> Result<(i64, i64)> {
+    let mut min_time = i64::MAX;
+    let mut max_time = i64::MIN;
+    for batch in batches {
+        let (min, max) = compute_timenanosecond_min_max_for_one_record_bacth(batch)?;
+        if min_time > min {
```
shepmaster, line 225: Would suggest same usage of `min` and `max` methods as above

query/src/util.rs
```diff
@@ -198,6 +216,55 @@ impl<'a> ExprRewriter for MissingColumnsToNull<'a> {
...
+    Ok((min_time, max_time))
+}
+
+/// Return min and max for column `time` in the given record batch
+pub fn compute_timenanosecond_min_max_for_one_record_bacth(
+    batch: &RecordBatch,
+) -> Result<(i64, i64)> {
+    if batch.num_columns() == 0 {
+        return Err(Error::EmptyBatch);
+    }
```
shepmaster, line 241:
```suggestion
    ensure!(batch.num_columns() > 0, EmptyBatchSnafu);
```

query/src/util.rs
```diff
@@ -198,6 +216,55 @@ impl<'a> ExprRewriter for MissingColumnsToNull<'a> {
...
+        .context(TimeCastingSnafu {})?;
+
+    let min = time_col
+        .iter()
+        .min()
+        .expect("Time column must have values")
+        .expect("Time column cannot be NULL");
+    let max = time_col
+        .iter()
+        .max()
```
shepmaster, line 261: I don't know how many values may be present in `iter`, but it could be inefficient to iterate it twice. I like to use things like [`minmax`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.minmax) when possible.

compactor/src/query.rs
NGA-TRAN: @shepmaster: I have made it `impl Into<String>`. Can you have a quick look if I understand it correctly?

query/src/util.rs
NGA-TRAN: I have used `minmax`.

compactor/src/query.rs
alamb: FWIW I think @crepererum  is actively working on removing the use of `DbChunk` -- so perhaps either he can use what is in this file in the querier or replace this as part of the work to not rely on `Db` / OG in the querier

compactor/src/query.rs
```diff
@@ -0,0 +1,216 @@
...
+        let mut merger = SchemaMerger::new();
+        for chunk in chunks {
+            merger = merger
+                .merge(&chunk.data.schema())
+                .expect("schemas compatible");
+        }
+        Arc::new(merger.build())
+    }
+
+    /// Returm min sequence number
```
alamb, line 75:
```suggestion
    /// Return min sequence number
```

compactor/src/query.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn min_sequence_number(&self) -> SequenceNumber {
+        self.iox_metadata.min_sequence_number
+    }
+
+    /// Returm max sequence number
```
alamb, line 80:
```suggestion
    /// Return max sequence number
```

compactor/src/query.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    fn delete_predicates(&self) -> &[Arc<DeletePredicate>] {
+        self.delete_predicates.as_ref()
+    }
+}
+
+impl QueryChunk for QueryableParquetChunk {
+    type Error = Error;
+
+    // Todo: This function should not be used in this NG chunk context
```
alamb, line 107: I don't really follow the logic here here about why the order of chunk ids doesn't matter. Maybe when the compactor creates a `QueryableParquetChunk` it should create a ChunkId with the correct relative order to the other chunks being compacted

compactor/src/utils.rs
```diff
@@ -0,0 +1,117 @@
...
+        partition_key: String,
+    ) -> QueryableParquetChunk {
+        let decoded_parquet_file = DecodedParquetFile::new((*self.data).clone());
+        let root_path = IoxObjectStore::root_path_for(&object_store, self.data.object_store_id);
+        let iox_object_store = IoxObjectStore::existing(object_store, root_path);
+        let parquet_chunk = new_parquet_chunk(
+            &decoded_parquet_file,
+            Arc::from(table_name.clone()),
+            Arc::from(partition_key),
+            ChunkMetrics::new_unregistered(), // TODO: need to add metrics
```
alamb, line 87: yeah this just needs to be threaded through

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
-struct ParquetFileWithTombstone {
-    data: Arc<ParquetFile>,
-    tombstones: Vec<Tombstone>,
-}
+        // Keep the fist IoxMetadata to reuse same IDs and names
+        let iox_metadata = overlapped_files[0].iox_metadata();
+
+        // Verify if the given files belong to the same partition
+        // Note: we can ignore this verification if we assume this is a must-have condition
+        if let Some((head, tail)) = overlapped_files.split_first() {
```
alamb, line 266: TIL https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_first 👍

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
+                    }
+                )
+            }
+        }
+
+        // Convert the input files into QueryableBatches for making query plan
+        let query_chunks: Vec<_> = overlapped_files
+            .iter()
+            .map(|f| {
+                f.to_queryable_parquet_chunk(
```
alamb, line 290: here I wonder if you could add a `chunk_id` to ensure the chunks were applied in the same order

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
+        // Merge schema of the compacting chunks
+        let merged_schema = QueryableParquetChunk::merge_schemas(&query_chunks);
+
+        // Compute the sorted output of the compacting result
+        let sort_key = compute_sort_key_for_chunks(&merged_schema, &query_chunks);
+
+        // Build compact query plan
+        let plan = ReorgPlanner::new()
+            .compact_plan(
+                Arc::new((*merged_schema).clone()),
```
alamb, line 317: I wonder if this could be written like `Arc::clone(&merged_schema)` to avoid a copy?

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
+        assert_batches_sorted_eq!(
+            &[
+                "+-----------+------+-----------------------------+",
+                "| field_int | tag1 | time                        |",
+                "+-----------+------+-----------------------------+",
+                "| 70        | UT   | 1970-01-01T00:00:00.000020Z |",
+                "| 1000      | WA   | 1970-01-01T00:00:00.000008Z |",
+                "+-----------+------+-----------------------------+",
+            ],
+            &batches
```
alamb, line 452: 👍

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
+
+        let lp1 = vec![
+            "table,tag1=WA field_int=1000 8000", // will be eliminated due to duplicate
+            "table,tag1=VT field_int=10 10000",  // will be deleted
+            "table,tag1=UT field_int=70 20000",
+        ]
+        .join("\n");
+
+        let lp2 = vec![
+            "table,tag1=WA field_int=1500 8000", // latest duplicate and kept
```
alamb, line 468: is it latest because the sequence number is greater?

compactor/src/compact.rs
```diff
@@ -167,47 +241,399 @@ impl Compactor {
...
+            .unwrap()
+            .unwrap()
+            .data;
+        // Should have 4 rows left
+        assert_batches_sorted_eq!(
+            &[
+                "+-----------+------+-----------------------------+",
+                "| field_int | tag1 | time                        |",
+                "+-----------+------+-----------------------------+",
+                "| 10        | VT   | 1970-01-01T00:00:00.000006Z |",
```
alamb, line 529: 👍  for not deleting VT (the tombstone is on the other chunk)

alamb (approved):
LGTM -- the only thing I am not totally sure of is the use of `0` for chunk id -- otherwise this looks great

alamb:
nice work @NGA-TRAN

compactor/src/compact.rs
NGA-TRAN: This is not necessary  @alamb with 2 reasons:
1. When we plan and run the query, the actual chunk order that matters is computed here https://github.com/influxdata/influxdb_iox/blob/b3911d0bdd595dfa71b96850b5d279beb39ec0ec/compactor/src/query.rs#L209 and based on sequence number.
2. @carols10cents is working on #3946 and #3949 to group overlapped files and they are not necessarily sorted in any order

compactor/src/query.rs
NGA-TRAN: @alamb If you look at his code here https://github.com/influxdata/influxdb_iox/blob/f2126c76d3bfd8f553f82e7baf45383c996b4c69/query/src/provider.rs#L604 it sorts the chunk on `order` and then `chunk id`. The order is based on sequence number which is always different for different parquet file by Ingester & Compactor design  so the chunk id does not matter much.

compactor/src/compact.rs
alamb: I thought it was  also important that when files are compacted together their relative order is also maintained (so that the correct rows are kept).

PR #4015, opened 2022-03-11
feat: extract "backend" from querier cache

crepererum:
The backend will implement pruning policies like LRU and TTL as well as
where/how the data is stored. Having a proper interface for that
simplifies the implementation since we don't need to have one massive
`Cache` object with a super complex mechanism.

This is for #3985.


querier/src/cache_system/backend/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub mod hash_map;
+
+#[cfg(test)]
+mod test_util;
+
+/// Backend to keep and manage stored entries.
+///
+/// A backend might remove entries at any point, e.g. due to memory pressure or expiration.
+pub trait Backend: Debug + Send + 'static {
```
domodwyer, line 11: What do you think to `CacheBackend`?

That way the consumer usage is nice and descriptive:

```rust
pub struct MyServer<T> where T: CacheBackend ...
```

vs

```rust
pub struct MyServer<T> where T: Backend ...
```

querier/src/cache_system/backend/test_util.rs
```diff
@@ -0,0 +1,93 @@
...
+    B: Backend<K = u8, V = String>,
+{
+    backend.set(1, String::from("a"));
+    backend.remove(&1);
+
+    assert_eq!(backend.get(&1), None);
+}
+
+/// Test setting a new value after removing it.
+fn test_readd<B>(mut backend: B)
```
domodwyer, line 84: typo: 
```suggestion
fn test_read<B>(mut backend: B)
```

domodwyer (approved):
I love composing up small traits ❤️

querier/src/cache_system/backend/test_util.rs
crepererum: It's actually not a type. It's "re-add" :sweat_smile:

domodwyer: ha fair enough!

querier/src/cache_system/backend/mod.rs
crepererum: done

PR #4016, opened 2022-03-11
feat: add addressable heap for query cache

crepererum:
This will be used as a helper data structure for TTL and LRU. It's
probably not the most performant implementation but it's good enough for
now.

This is for #3985.


querier/src/cache_system/backend/addressable_heap.rs
```diff
@@ -0,0 +1,361 @@
...
+            self.key_to_order_and_value
+                .insert(k.clone(), (v, o.clone()))
+                .is_none(),
+            "entry should have been removed by now"
+        );
+
+        match self.queue.binary_search_by_key(&(&o, &k), project_tuple) {
+            Ok(_) => unreachable!("entry should have been removed by now"),
+            Err(index) => {
+                self.queue.insert(index, (o, k));
```
domodwyer, line 64: This might be a little expensive if we're inserting elements into the middle of the queue:

```text
┌───┬───┐ ┌───┬───┐
│ A │ B │ │ D │ E │
└───┴───┘▲└───┴───┘
         │         
       ┌───┐       
       │ C │       
       └───┘       
```

In this case, the `D` and `E` elements need to be copied to move them to the right. If the `VecDeque` buffer needs to grow we'll have to copy the whole lot (though that's obviously less of a problem as it amortises nicely).

What about using a [`BTreeSet`](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html) keyed by `(O, K)` here? We'd have to add a `Hash` bounds to `O` but that'd give us `log2 N` insert even in the middle of the queue while still being ordered and maintaining [peek](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.first) and [pop](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.pop_first) functionality.

querier/src/cache_system/backend/addressable_heap.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    /// Peek first element (by smallest `O`).
+    pub fn peek(&self) -> Option<(&K, &V, &O)> {
+        if let Some((o, k)) = self.queue.front() {
+            let (v, o2) = self
+                .key_to_order_and_value
+                .get(k)
+                .expect("value is in queue");
+            assert!(o == o2);
```
domodwyer, line 78: Nice validation 👍 I assume it's not `assert_eq!` because it'd been a `Debug` bound? Might not be a problem to add for a helpful error if this assert were to ever be hit - I suspect all `O` we use will probably impl `Debug` already.

domodwyer (approved):
Looks good to me!

I dunno if we have any property testing already in the wider code base, but if we've already paid the dependency cost I find it to be a useful tool for validating these kind of double-referenced data structures (especially as they evolve). Something to consider!

querier/src/cache_system/backend/addressable_heap.rs
domodwyer: Ah I just realised the peek/pop functionality is `nightly` at the moment which seems like a big shame :( 

https://github.com/rust-lang/rust/issues/62924

The ticket describes how we can emulate the functionality (call `.iter().next()` / `remove()`).

querier/src/cache_system/backend/addressable_heap.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    #[test]
+    fn test_peek_some() {
+        let mut heap = AddressableHeap::new();
+
+        heap.insert(1, "a", 4);
+        heap.insert(2, "b", 3);
+        heap.insert(3, "c", 5);
+
+        assert_eq!(heap.peek(), Some((&2, &"b", &3)));
```
alamb, line 155: 👍

querier/src/cache_system/backend/addressable_heap.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    #[test]
+    fn test_override() {
+        let mut heap = AddressableHeap::new();
+
+        assert_eq!(heap.insert(1, "a", 4), None);
+        assert_eq!(heap.insert(2, "b", 3), None);
+        assert_eq!(heap.insert(1, "c", 5), Some(("a", 4)));
+    }
+}
```
alamb, line 361: I may have missed it, but it may be wirth testing / spelling out what happens if there is a tie on the order `O` -- like what happens if you insert

```rust
        heap.insert(1, "a", 4);
        heap.insert(2, "b", 4);

        // what happens here? Is it arbitrary? 
        heap.pop(), Some((2, "b", 3)));
```

alamb (approved):
I didn't review the code (I think @domodwyer  gave that a good once over) but looked at all the test cases. Nice work

querier/src/cache_system/backend/addressable_heap.rs
crepererum: I thought about that as well. Let's agree on: write a benchmark first, then change the data structure. I'm pretty sure thought that the tree will pay off.

querier/src/cache_system/backend/addressable_heap.rs
crepererum: Exactly. Yeah, I could probably just assume `O: Debug`, because why would you not implement it.

crepererum:
> I dunno if we have any property testing already in the wider code base, but if we've already paid the dependency cost I find it to be a useful tool for validating these kind of double-referenced data structures (especially as they evolve). Something to consider!

Will do so in a follow-up PR.

querier/src/cache_system/backend/addressable_heap.rs
crepererum: Ties are broken via `K`. I've adjusted the docs and added a two new tests.

PR #4019, opened 2022-03-11
feat: object store instrumentation

domodwyer:
This PR adds a decorator over [`ObjectStoreApi`](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L68) implementations that records call latency and bytes up/down.

As of this PR the decorator is not used (outside of tests) as it will take some refactoring of the object store code to enable us to use it effectively which I shall do in a separate PR; while the object store is abstacted as a trait, we use the concrete [`ObjectStore`](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L113) type throughout the codebase so we can't easily overlay the decorator without introducing generics / dyn at all consumers. Currently the `ObjectStore` type achieves runtime selection of the correct backend by internally dispatching to the correct impl based on the value of `pub` [enum field](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L115), of which all variants impl the `ObjectStoreApi`.

I plan on removing the internal [enum dispatch](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L462-L475) and replacing it with `dyn ObjectStoreApi` and wrapping that, assuming all existing implementations satisfy it (I've not looked in depth yet). This will allow all the existing code to continue using the concrete `ObjectStore` type which will internally take care of instrumentation transparently for the caller.

I believe there's a small performance benefit of performing enum-based dispatch vs. true `dyn` dynamic dispatch I intend to switch to, but I suspect this is of negligent value once network round trips for the actual call are factored in - I am unaware of any other downside of removing this enum approach.

Once this refactor is complete, the ingestor and all other consumers of the `object_store` crate will (optionally) emit the metrics introduced in this PR.

---

* feat: object store instrumentation (bfc473b8c)

      Implements a decorator of the ObjectStoreAPI trait, recording:

          * Bytes uploaded / downloaded through the instrumented API
          * Call latencies, broken down by operation & success / error state

      All the current implementations that return a Stream from the get operation
      actually return a "fake" stream containing all the data in one go rather than
      streaming chunks from the upstream. I've instrumented the Stream to
      future-proof any actual streaming impls in the future.


domodwyer:
Interestingly the failed test doesn't fail locally: https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/16718/workflows/c9ef9622-4332-49b2-8b1c-bf7610eb3571/jobs/134736 and the identical test that drives the `list_with_delimiter` variant does fail as expected in CI....

I'm not sure what the deal is here, but I'll switch to using the mock [`DummyObjectStore`](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/dummy.rs#L26) rather an the [`new_failing_store()`](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L181) which is simply a misconfigured real impl.

alamb (commented):
> I plan on removing the internal [enum dispatch](https://github.com/influxdata/influxdb_iox/blob/bfc473b8cd63bf67b34fcb57a338093760b590bf/object_store/src/lib.rs#L462-L475) and replacing it with dyn ObjectStoreApi

I think that sounds like a good idea

alamb:
> I believe there's a small performance benefit of performing enum-based dispatch vs. true dyn dynamic dispatch I intend to switch to, but I suspect this is of negligent value once network round trips for the actual call are factored in - I am unaware of any other downside of removing this enum approach.

I don't think the runtime overhead of a dynamic function call will be measurable at all. I am all for `dyn ObjectStore` all over the place if possible

I vaguely remember the enum approach was related to different types for `Path` maybe (though @carols10cents  is probably the best person to comment here)

alamb:
I'll review this later today if no one else gets to it

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+    }
+
+    async fn get(&self, location: &Self::Path) -> Result<GetResult<Self::Error>, Self::Error> {
+        let started_at = self.time_provider.now();
+
+        let res = self.inner.get(location).await;
+
+        match res {
+            Ok(GetResult::File(file, path)) => {
+                // Record a the file size in bytes and time the inner call took.
```
carols10cents, line 188: Nit, extra word:

```suggestion
                // Record the file size in bytes and time the inner call took.
```

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+
+    use futures::stream;
+    use metric::Attributes;
+    use tokio::io::AsyncReadExt;
+
+    use crate::{dummy, ObjectStore};
+
+    use super::*;
+
+    fn assert_histogram_hit<const N: usize>(
```
carols10cents, line 456: Oooh fancy const generics!

carols10cents (approved):
LGTM! Tiny, tiny comment fixes you may take or leave.

carols10cents:
> I vaguely remember the enum approach was related to different types for Path maybe (though @carols10cents is probably the best person to comment here)

Ooooh good memory. Yeah switching to `dyn ObjectStoreApi` might lose the compile-time checking that the right object store only uses the right kind of path.... but if there are advantages to the `dyn` way that outweigh that checking, I'm totally cool with that :)

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+/// # Backwards Clocks
+///
+/// If the system clock is observed as moving backwards in time, call durations
+/// are not recorded. The bytes transferred metric is not affected.
+#[derive(Debug)]
+pub struct ObjectStoreMetrics<T, P = SystemProvider> {
+    inner: T,
+    time_provider: P,
+
+    put_success_duration: U64Histogram,
```
alamb, line 64: I would recommend naming all these fields to include the unit. Such as 

```suggestion
    put_success_duration_ms: U64Histogram,
```

A common source of errors I have found is mis interpreting time units and putting the units in the names removes all possible misinterpretations

I see the actual metric names reported have the `_ms` postfix, so adding them to the code would also make things a little more consistent

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+                }
+                Ok(GetResult::File(file, path))
+            }
+            Ok(GetResult::Stream(s)) => {
+                // Wrap the object store data stream in a decorator to track the
+                // yielded data / wall clock, inclusive of the inner call above.
+                Ok(GetResult::Stream(Box::pin(Box::new(
+                    StreamMetricRecorder::new(
+                        s,
+                        started_at,
```
alamb, line 203: 👍

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+        &'a self,
+        prefix: Option<&'a Self::Path>,
+    ) -> Result<BoxStream<'a, Result<Vec<Self::Path>, Self::Error>>, Self::Error> {
+        let t = self.time_provider.now();
+
+        let res = self.inner.list(prefix).await;
+
+        // Avoid exploding if time goes backwards - simply drop the measurement
+        // if it happens.
+        if let Some(delta) = self.time_provider.now().checked_duration_since(t) {
```
alamb, line 248: I am not sure it is important, but does this record the time taken to get the first part of the response back from `LIST` or the entire time of the list operation?

I am asking because `list` returns a stream rather than a single result

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+
+    // The timestamp at which the read request began, inclusive of the work
+    // required to acquire the inner stream (which may involve fetching all the
+    // data if the result is only pretending to be a stream).
+    started_at: Time,
+    // The time at which the last part of the data stream (or error) was
+    // returned to the caller.
+    //
+    // The total get operation duration is calculated as this timestamp minus
+    // the started_at timestamp.
```
alamb, line 311: This implies that if a consumer (like DataFusion) opens a stream but doesn't read from it as fast as data is available, then I wonder if the metrics might give the mistaken impression that the stream took a long time to read rather than "stream was ready but the consumer took its sweet time reading"

I don't have any specific changes to suggest other than maybe a note in the metric description

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+                    false => this.get_error_duration,
+                };
+
+                if let Some(d) = this
+                    .last_yielded_at
+                    .expect("no last_yielded_at value for fused stream")
+                    .checked_duration_since(*this.started_at)
+                {
+                    hist.record(d.as_millis() as _)
+                }
```
alamb, line 400: I am not sure it is clearer, but you can avoid the explicitt assignment to `None` if you wanted via `take()`


```suggestion
                if let Some(d) = this
                    .last_yielded_at
                    .take()
                    .expect("no last_yielded_at value for fused stream")
                    .checked_duration_since(*this.started_at)
                {
                    hist.record(d.as_millis() as _)
                }
```

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+
+    #[tokio::test]
+    async fn test_list_fails() {
+        let metrics = Arc::new(metric::Registry::default());
+        let store = dummy::new_failing_s3().expect("cannot init failing store");
+        let store = ObjectStoreMetrics::new(store, &metrics);
+
+        assert!(store.list(None).await.is_err(), "mock configured to fail");
+
+        assert_histogram_hit(
```
alamb, line 554: is it worth asserting that there are no `"success"` results reported?

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+            .await
+            .expect("should yield data")
+            .expect("should succeed");
+        assert_eq!(got.len(), 1);
+        assert_counter_value(&metrics, "object_store_transfer_bytes", [], 1);
+
+        // Sleep at least 10ms to assert the recorder to captures the wall clock
+        // time.
+        const SLEEP: Duration = Duration::from_millis(20);
+        tokio::time::sleep(SLEEP).await;
```
alamb, line 737: Given you are using a time provider, I wonder if rather than using `tokio::time::sleep` you could use a `MockTimeProvider` 

like
```rust
let time_provider = MockTimeProvider::new();
....
time_provider.inc(Duration::from_millis(20))
```

Doing so would avoid potential race conditions in the test as well as run the tests faster (though a sleep of 20ms is unlikely to cause problems on either front I suppose)

alamb (approved):
Beautiful tests 👍 

It would be cool to remove the calls to `tokio::time:sleep` but otherwise this PR is 👨‍🍳 ❤️  to me

alamb:
Nice work @domodwyer

object_store/src/instrumentation.rs
```diff
@@ -0,0 +1,1102 @@
...
+#[async_trait]
+impl<T, P> ObjectStoreApi for ObjectStoreMetrics<T, P>
+where
+    T: ObjectStoreApi,
+    P: TimeProvider,
+{
+    type Path = T::Path;
+    type Error = T::Error;
+
+    fn new_path(&self) -> Self::Path {
```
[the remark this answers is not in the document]
domodwyer: 💯

object_store/src/instrumentation.rs
domodwyer: Can't argue with that!

object_store/src/instrumentation.rs
domodwyer: I hadn't noticed that LIST was also a stream.... I shall shoehorn it into the existing stream instrumentation.

object_store/src/instrumentation.rs
domodwyer: fd6d115cef37df2069f9ebee9567419b3c7220c1

object_store/src/instrumentation.rs
domodwyer: The test should never race due to the fact the duration of the actual sleep should never be _less than_ the 20ms. It may well be more than 20ms under load, and so the test checks all the buckets >20ms to find the hit - that way the assert is tolerant of any load-induced extra long sleeps!

I was going to use the mock time provider, but then to advance it I need to be holding the mock time provider, which means I have to stick a pointless `Arc` in so I can share it with the actual stream decorator and the test code and it all just seemed a bit pointless - we'll never notice the 20ms sleep anyway.

domodwyer:
Thanks @carols10cents / @alamb for the keen eyes 👀 

I'll decorate the LIST stream later today and merge this 👍

object_store/src/instrumentation.rs
alamb: sounds good -- we can always switch to TimeProvider in the future if this is causing issues so leaving as is sounds good to me 👍

PR #3965, opened 2022-03-07
feat: all in 1 IOx NG mode

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/3926

With this we now have a IOx NG that looks externally like the current IOx. Namely you run a command such as the following, load line protocol, and query the data

```shell
./influxdb_iox run all-in-one ....
```
And you get an HTTP api on port 8080 and a gRPC API port on 8082 that behaves eerily like `./influxdb_iox run`

# Depends on PRs
I have broken this down into smaller PRs to avoid a single massive PR:
- [x] refactor: decouple router2 setup and creation logic from the config struct use `Arc<dyn ServerType>` rather than generics over ServerType https://github.com/influxdata/influxdb_iox/pull/3969
- [x] https://github.com/influxdata/influxdb_iox/pull/4017
- [x] https://github.com/influxdata/influxdb_iox/pull/4018
- [x] https://github.com/influxdata/influxdb_iox/pull/4020
- [x] https://github.com/influxdata/influxdb_iox/pull/3969

# Locally Testing

## Clean catalog:

```shell
$ psql
psql (14.2, server 14.1)
Type "help" for help.

alamb=# drop schema iox_catalog cascade;
NOTICE:  drop cascades to 12 other objects
DETAIL:  drop cascades to table iox_catalog._sqlx_migrations
drop cascades to table iox_catalog.kafka_topic
drop cascades to table iox_catalog.query_pool
drop cascades to table iox_catalog.namespace
drop cascades to table iox_catalog.table_name
drop cascades to table iox_catalog.column_name
drop cascades to table iox_catalog.sequencer
drop cascades to table iox_catalog.sharding_rule_override
drop cascades to table iox_catalog.partition
drop cascades to table iox_catalog.parquet_file
drop cascades to table iox_catalog.tombstone
drop cascades to table iox_catalog.processed_tombstone
DROP SCHEMA
```

### Startup in one mode:
```shell
LOG_FILTER=info ./target/debug/influxdb_iox run -vv  all-in-one \
  --write-buffer=file\
  --write-buffer-addr /tmp\
  --catalog-dsn postgresql://localhost:5432/alamb\
  --write-buffer-partition-range-start 0\
  --write-buffer-partition-range-end 0\
  --pause-ingest-size-bytes 10000\
  --persist-memory-threshold-bytes 1000
```

### Write data to port 8080:

In the IOx directory
```shell
[...]

PR #4020, opened 2022-03-11
refactor: decouple ingester setup and creation logic from the config structs

alamb:
# Rationale:

Apply the same rationale as described in https://github.com/influxdata/influxdb_iox/pull/3969 for the `router` to the `ingester.`

Bascially I want to be able to start a ingester without the command line config

For a big picture view, you can see how this is used in the WIP PR for the all in one server https://github.com/influxdata/influxdb_iox/pull/3965


clap_blocks/src/ingester.rs
```diff
@@ -0,0 +1,65 @@
+/// CLI config for catalog ingest lifecycle
+#[derive(Debug, Clone, clap::Parser)]
+pub struct IngesterConfig {
```
alamb, line 3: This code was moved straight up out of influxdb_iox/src/commands/run/ingester.rs

influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -81,75 +61,16 @@ pub struct Config {
     #[clap(flatten)]
     pub(crate) write_buffer_config: WriteBufferConfig,
 
-    /// Write buffer partition number to start (inclusive) range with
```
alamb, line 84: moved to clap_blocks/src/ingester.rs

influxdb_ioxd/src/server_type/ingester.rs
```diff
@@ -96,3 +140,73 @@ impl HttpApiErrorSource for IoxHttpError {
         HttpApiError::new(self.status_code(), self.to_string())
     }
 }
+
+/// Instantiate an ingester server type
+pub async fn create_ingester_server_type(
```
alamb, line 145: The point of this PR is to create this function which I will call from the all in one entrypoint!

pauldix (approved):

PR #4021, opened 2022-03-11
feat: Add catalog/compactor functions for getting level 0 and level 1 files; marking files as level 1

### LINKED ISSUE #3950: NG Compactor: Update the catalog of parquet files of a given list of parquet IDs to level 1
# Rationale
This ticket implements step 8 of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). The input is a list of parquet file IDs 

```rust
fn update_to_level_1(&[ParquetFileId]) -> Result<()>
``` 
Each update can be in its own transaction but for better performance, we prefer all the updates are in one transaction

carols10cents:
I chose to fix #3950 and #3946 together because it was hard to test each without the other :)

This will conflict with https://github.com/influxdata/influxdb_iox/pull/4001 because I see that moved `TablePartition` too; I'm happy to take care of the merge conflicts.

You'll see a reference to a sqlx bug in here near some icky code; I'm working on submitting an upstream fix.

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -409,6 +409,9 @@ pub trait TombstoneRepo: Send + Sync {
 /// The starting compaction level for parquet files is zero.
 pub const INITIAL_COMPACTION_LEVEL: i16 = 0;
 
+/// The maximum size, in bytes, of a file to be considered for compaction.
+pub const MAX_COMPACT_SIZE: i64 = 1024 * 1024;
+
 /// Functions for working with parquet file pointers in the catalog
 #[async_trait]
 pub trait ParquetFileRepo: Send + Sync {
@@ -433,6 +436,10 @@ pub trait ParquetFileRepo: Send + Sync {
         &mut self,
         namespace_id: NamespaceId,
     ) -> Result<Vec<ParquetFile>>;
+
+    /// List parquet files for a given sequencer with compaction level 0 and other criteria that
+    /// define a file as a candidate for compaction
+    async fn level_0(&mut self, sequencer_id: SequencerId) -> Result<Vec<ParquetFile>>;
 
     /// Verify if the parquet file exists by selecting its id
     async fn exist(&mut self, id: ParquetFileId) -> Result<bool>;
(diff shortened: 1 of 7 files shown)


iox_catalog/src/interface.rs
```diff
@@ -409,6 +410,9 @@ pub trait TombstoneRepo: Send + Sync {
 /// The starting compaction level for parquet files is zero.
 pub const INITIAL_COMPACTION_LEVEL: i16 = 0;
 
+/// The maximum size, in bytes, of a file to be considered for compaction.
+pub const MAX_COMPACT_SIZE: i64 = 1024 * 1024;
```
carols10cents, line 414: This is a totally arbitrary value; what should this be set to?

iox_catalog/src/postgres.rs
```diff
@@ -1267,6 +1267,109 @@ WHERE table_name.namespace_id = $1 AND parquet_file.to_delete = false;
...
+            UPDATE parquet_file
+            SET compaction_level = 1
+            WHERE id = ANY($1)
+            RETURNING id
+            ;"#,
+        )
+        .bind(&ids[..])
+        .fetch_all(&mut self.inner)
+        .await
+        .map_err(|e| Error::SqlxError { source: e })?;
```
NGA-TRAN, line 1367: I am not sure how sqlx handle transaction for this. Does it auto-commit after all updates or commit after every update?  I think it is worth finding this out. We prefer the former. Maybe @domodwyer and @crepererum already know the answer

NGA-TRAN (approved):
Very nice @carols10cents . Only need to remove < max size conditions.

iox_catalog/src/postgres.rs
carols10cents: I hope that `sqlx` is passing this query through to Postgres as it is; if that's not how `sqlx` works, then I have a lot more reading to do! This is one SQL statement that Postgres executes in one operation.

domodwyer: You get to choose! It depends how we're using it - if the caller asked for a txn:

```rust
let txn = postgres.start_transaction().await.expect("txn start");
txn.parquet_files().update_to_level_1(...).await;
```

Then normal transaction rules apply - the whole batch of changes in the txn needs explicitly commiting or it is rolled back when dropped.

If however they didn't ask for a txn:

```rust
postgres.repositories().parquet_files().update_to_level_1(...).await;
```

Then each call to `update_to_level_1()` will be commited / become visible as it executes.

carols10cents: Right, but *within* one call of `update_to_level_1`, that's running one SQL `UPDATE` statement with potentially multiple parquet file IDs, either all parquet files will be updated or none will, even without a transaction. Correct?

domodwyer: Yep! It of course may later be rolled back, but that'd only happen if the caller explicitly wanted that behaviour.

PR #4027, opened 2022-03-14
refactor: add TTL cache backend

crepererum:
This is based on the new `AddressableHeap`, which simplifies the
implementation quite a lot.


querier/src/cache_system/backend/ttl.rs
```diff
@@ -0,0 +1,445 @@
...
+    ) -> Self {
+        Self {
+            inner_backend,
+            ttl_provider,
+            time_provider,
+            expiration: Default::default(),
+        }
+    }
+
+    fn update(&mut self, now: Time) {
```
domodwyer, line 88: Perhaps

```suggestion
    fn evict_expired(&mut self, now: Time) {
```

So usage below clearly shows what is happening:

```rust
fn get(&mut self, k: &Self::K) -> Option<Self::V> {
        self.evict_expired(self.time_provider.now());
        self.inner_backend.get(k)
    }
```

Could also inline acquiring the time from the time provider? Every call is `self.update(self.time_provider.now())`

querier/src/cache_system/backend/ttl.rs
```diff
@@ -0,0 +1,445 @@
...
+}
+
+impl<K, V> TtlBackend<K, V>
+where
+    K: Clone + Eq + Debug + Hash + Ord + Send + 'static,
+    V: Clone + Debug + Send + 'static,
+{
+    /// Create new backend w/o any known keys.
+    ///
+    /// The inner backend SHOULD NOT contain any data at this point, otherwise we will not track any TTLs for these entries.
```
domodwyer, line 74: Are you CAPITALISING the RFC words? If so I love it :heart:

I used to do a lot of IETF / RFC work and I started doing this in code comments too a while back - I have my editor highlight the lines that have `SHOULD`, `MUST` and `MUST NOT` and so on in code comments and it makes for great easily spotting documentation of usually critical invariants - commenting like this should be more popular 🎉

querier/src/cache_system/backend/ttl.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    /// When should the given key-value pair expire?
+    ///
+    /// Return `None` for "never".
+    ///
+    /// The function is only called once for a newly cached key-value pair. This means:
+    /// - There is no need in remembering the time of a given pair (e.g. you can safely always return a constant).
+    /// - You cannot change the TTL after the data was cached.
+    ///
+    /// Expiration is set to take place AT OR AFTER the provided duration.
```
domodwyer, line 23: Are you having a ask-once TTL oracle like this to avoid having to store a map of `K` -> `expiry_time`? 

An alternative design would be a decorator that impls the `CacheBackend` trait, wrapping an underlying cache to compose on TTL functionality - you'd only have one trait, no glue code and all the TTL behaviour (this kind of logic in the comment) could be internalised. You would need that map though which would increase memory consumption, even if only for a hashed representation of K and a deadline. I'm not saying it's a better way of doing things as I'm still building a picture of what you're building - I'm more asking what yoru thinking is / what your prioritising 👍

querier/src/cache_system/backend/ttl.rs
```diff
@@ -0,0 +1,445 @@
...
+impl<K, V> TtlProvider for NeverTtlProvider<K, V> {
+    type K = K;
+    type V = V;
+
+    fn expires_in(&self, _k: &Self::K, _v: &Self::V) -> Option<Duration> {
+        None
+    }
+}
+
+/// Cache backend that implements Time To Life.
```
domodwyer, line 54: Might be worth mentioning that calls to the cache methods trigger GC of keys - mainly because it could cause unpredictable latency of the calls.

I think it would be awesome to add a metric that tracks the duration of GC work done for each call eventually - that'd help us understand the latency of the various subsytems in the querier and build up a solid picture of what is going on 👍 

Let me know if you want me to PR anything!

domodwyer (approved):
Looks good - interesting to see how this builds out 👍

querier/src/cache_system/backend/ttl.rs
crepererum: Renamed. Didn't inline it though, because we need the time from the time provider also during `set` and I wanted to avoid to call it twice.

querier/src/cache_system/backend/ttl.rs
crepererum: `TtlBackend` is a wrapper that uses an inner backend. The reason why I'm storing the TTL data in an extra data structure is so we can clean up any expired key, even when it's not touched via `get`. That is important if the inner backend is for example an memory-capped LRU (i.e. it doesn't make sense to prune keys due to memory pressure if half of your cache is already "TTL-expired").

Another reason I'm using another trait (`TtlProvider`) is because it kinda makes the usage of the cache setup a bit clearer, because the cache loader (which actually does the work) doesn't need to care about the cache policy.

PR #4031, opened 2022-03-14
feat: add ingester lifecycle metrics

pauldix:


ingester/src/lifecycle.rs
```diff
@@ -657,6 +726,15 @@ mod tests {
...
         assert_eq!(stats.partition_stats.len(), 0);
+
+        let persist_counter: Metric<U64Counter> = metric_registry
+            .get_instrument("ingester_lifecycle_persist_count")
+            .unwrap();
+        let mem_counter = persist_counter
+            .get_observer(&Attributes::from(&[("trigger", "memory")]))
+            .unwrap()
+            .fetch();
+        assert_eq!(mem_counter, 1);
```
crepererum, line 737: could be worth to factor this into a helper function to reduce the noise a bit, essentially this is is the same code again and again except the trigger trigger and maybe the count itself.

crepererum (approved):
This will be helpful.

ingester/src/lifecycle.rs
```diff
@@ -244,11 +268,22 @@ impl LifecycleManager {
...
                 .map(|age| age > self.config.partition_cold_threshold)
                 .unwrap_or(false);
+            if is_cold {
+                self.persist_cold_counter.inc(1);
+            }
+
             let sized_out = s.bytes_written > self.config.partition_size_threshold;
+            if sized_out {
+                self.persist_size_counter.inc(1);
+            }
```
domodwyer, line 286: A single partition that matches multiple persistence trigger conditions could cause more than one increment to the persist counters (a big, old, cold partition would +3 to the counter) - we should remember the persist counters != number of persisted partitions!

Perhaps it's easier to pick one if there are multiple valid conditions?

domodwyer (approved):
📈 

I'll push more metrics in too as I go through the code 👍

ingester/src/lifecycle.rs
pauldix: Even with one, every time the manager says to persist doesn't necessarily mean that something will get persisted because it's an estimator. I think this is  useful because it tells us how many times each condition would have triggered a persist.

PR #4038, opened 2022-03-15
fix: race condition in `Cache::set`

crepererum:
In theory on a multi-threaded tokio executor, the following could have
happened:

| Thread 1              | Thread 2                            |
| --------------------- | ----------------------------------- |
|                       | Running query begin                 |
|                       | ...                                 |
|                       | `loader.await` finished             |
| `Cache::set` begin    |                                     |
| state locked          |                                     |
|                       | try state lock, blocking            |
| running query removed |                                     |
| ...                   |                                     |
| state unlocked        |                                     |
| `Cache::set` end      |                                     |
|                       | state locked                        |
|                       | panic because running query is gone |

Another issue that could happen is if we:
1. issue a get request, loader takes a while, this results in task1
2. side-load data into the running query (task1 still running)
3. the underlying cache backend drops the result very quickly (task1
   still running)
4. we request the same data again, resulting in yet another query task
   (task2), task1 is still running at this point

In this case the original not-yet-finalized query task (task1) would
remove the new query task (task2) from the active query set, even
though task2 is actually not done.

We fix this by the following measures:

- **task tagging:** tasks are tagged so if two tasks for the same key
  are running, we can tell them apart
- **task->backend propagation:** let the query task only write to the
  underlying backend if it is actually sure that it is running
- **prefer side-loaded results:** restructure the query task to strongly
  prefer side-loaded data over whatever comes from the loader
- **async `Cache::set`:** Let `Cache::set` wait until a running query
  task completes. This has NO correctness implications, it's probably
  just nicer for resource management.

For #3985.


domodwyer (approved):
👍 

These concurrent state transitions are getting a little hairy 🙈

PR #4039, opened 2022-03-15
feat: add "dual" cache pattern

crepererum:
This will be useful for certain parts that are addressed internally via
ID but where the user-facing APIs use names.

For #3985.


domodwyer (commented):
It might be easier to decorate a pair of `CacheBackend` rather than wrap the loader, then it can maintain two caches by _owning_ the two inner caches, presenting two `CacheBackend` impls with different `K` to interact with one vs. the other and allowing the owner to synchronise in one lock, instead of having a loader that pushes into two caches in two goes.

That would eliminate the circular refs / weak ptrs, and make it possible for the dual cache to control the concurrency and atomicity of an insert - currently I think it would be possible to observe the k1 but find the k2 key is not yet there. It could get increasingly confusing when we start composing together these lack of synchronisations into bigger cache components.

It depends on how this code is used, but from the impl here it would be possible to have two different `V` in each cache for the same key if two calls to `load()` raced too:

```
T1: load("bananas") that will yield 42
T2: load("bananas") that will yield 100

T1: self.inner.load("bananas").await -> 42
T2: self.inner.load("bananas").await -> 100

T2: other.set("bananas", 100);
T1: other.set("bananas", 42);

Now you have:

	inner: 100
	other: 42
```

If the `V` values never change between `load()` calls then the values obviously won't diverge, but that's not an invariant I've seen mentioned - would that be enforcible for all usage of the cache, even in the face of outside processes mutating bits and bobs like compaction? If so, we should doc all the caveats we can think of and merge it so we can get going, but I'm starting to get the sync terror feels 👻

crepererum:
@domodwyer done, PTAL:

```
refactor: rework "dual" cache construct to be backend based

Pros:
- easiser to reason about the locking and consistency, esp. in
  concurrent applications

Cons:
- we are not canceling running queries for the dual cache any longer
```

querier/src/cache_system/backend/dual.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Create two backends that cross-populate [`set`](CacheBackend::set) options each other.
+///
+/// This can be helpful if data is requested via two keys, e.g. via ID and via name.
+///
+/// Note that other operations like [`remove`](CacheBackend::remove) are NOT cross-populated.
+///
+/// Takes two backend and a mapper that can extract the keys for the other backends. If these mappers return `None`, the
+/// data will NOT cross-populated. This can be helpful in cases where the other key cannot be extracted, e.g. for
+/// "missing" values. Be careful however since this can lead to inconsistent states in both caches.
```
domodwyer, line 26: Do you have an example of when this might happen? I'm trying to understand when we'd want this behaviour.

querier/src/cache_system/backend/dual.rs
```diff
@@ -0,0 +1,222 @@
...
+
+/// Boxed mapper function.
+type BoxedMapper<K1, K2, V> =
+    Box<dyn for<'k, 'v> Fn(&'k K1, &'v V) -> Option<K2> + Send + Sync + 'static>;
+
+/// Create two backends that cross-populate [`set`](CacheBackend::set) options each other.
+///
+/// This can be helpful if data is requested via two keys, e.g. via ID and via name.
+///
+/// Note that other operations like [`remove`](CacheBackend::remove) are NOT cross-populated.
```
domodwyer, line 22: Also interesting - how come?

querier/src/cache_system/backend/dual.rs
```diff
@@ -0,0 +1,222 @@
...
+    };
+    (dual1, dual2)
+}
+
+/// First backend created by [`dual_backends`].
+pub struct DualBackend1<K1, K2, V>
+where
+    K1: Clone + Eq + Hash + std::fmt::Debug + Ord + Send + 'static,
+    K2: Clone + Eq + Hash + std::fmt::Debug + Ord + Send + 'static,
+    V: Clone + std::fmt::Debug + Send + 'static,
```
domodwyer, line 57: I know it's more restrictive, but would it be worth requiring `V` to be `Arc<V>` or arc-ing it internally? I imagine that's how it'll be used and it's fairly easy for a naive user of the dual cache to accidentally wind up storing all `V` twice.

querier/src/cache_system/backend/dual.rs
```diff
@@ -0,0 +1,222 @@
...
+    type K = K1;
+    type V = V;
+
+    fn get(&mut self, k: &Self::K) -> Option<Self::V> {
+        self.shared.lock().0.get(k)
+    }
+
+    fn set(&mut self, k1: Self::K, v: Self::V) {
+        let mut shared = self.shared.lock();
+        if let Some(k2) = (self.mapper)(&k1, &v) {
```
domodwyer, line 93: I imagine these mapper fns are expected to be fast (as opposed to involving I/O or anything)?

We could always map, then lock if we're not certain of the mapper latency to guard against it.

domodwyer (approved):
Looks much simpler 🎉

querier/src/cache_system/backend/dual.rs
crepererum: Because `remove` only gets `K1`, to extract `K2` you also need `V` which might be gone from the inner backend by this time (e.g. if TTLs are applied or a to-be-created memory-pressure-LRU-thing). Same applies vice versa (i.e. getting `K1` from `K2`).

We could store the `K1<>K2` dual mapping internally though :thinking:

querier/src/cache_system/backend/dual.rs
crepererum: Have a look at the test ("Mappers CAN lead to inconsistent states..." at the end).

querier/src/cache_system/backend/dual.rs
crepererum: done

crepererum:
Merging it now, we can always improve it later (e.g. we could decide to store `K1<>K2` as a dual mapping, so that `remove` is also synced).

querier/src/cache_system/backend/dual.rs
domodwyer: That separate TTL-per-cache sounds weird/unintended - is there a reason to keep K->V around but by asking for it by a different name says it's gone?

I think K1 -> K2 sounds good - i think it's less of a cloned-memory-footgun and provides the same properties :+1:

querier/src/cache_system/backend/dual.rs
domodwyer: I understand that caching `Copy` types doesn't need an `Arc`, but I imagine they're relatively rare - the `Arc<str>` would be fine with an `Arc<V>` bound though!

> I think the API user should decide what they prefer.

That's a perfectly valid design choice, but we should definitely _document_ that `V` is stored twice.

PR #4040, opened 2022-03-15
feat: `OptionalValueTtlProvider`

crepererum:
Quite a few caches will request data from the catalog w/o knowing if it
exists (e.g. a table by name). We should have different TTLs for "exists"
and "unknown" w/o writing much boilerplate code.

For #3985.


domodwyer (approved):
Does what it says 👍

PR #4043, opened 2022-03-15
fix: propagate shutdown into QuerierHandlerImpl

domodwyer:
I noticed this while playing with @alamb 's all-in-one mode 🎉 

---

* fix: propagate shutdown into QuerierHandlerImpl (22eab934c)

      Prior to this commit, calling shutdown() on the QuerierServer (the server
      layer run by the iox binary) would cancel it's own CancellationToken, while
      the QuerierHandlerImpl (the actual querier workload entrypoint) would be
      watching it's own, different token.

      This commit removes the redundant CancellationToken in the QuerierServer,
      instead using the inner QueryHandlerImpl for cancellation notification &
      completion.

### DIFF
--- a/influxdb_ioxd/src/server_type/querier/mod.rs
+++ b/influxdb_ioxd/src/server_type/querier/mod.rs
@@ -15,7 +15,6 @@ use querier::{
 };
 use query::exec::Executor;
 use time::TimeProvider;
-use tokio_util::sync::CancellationToken;
 use trace::TraceCollector;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


crepererum (approved):
Good catch!

influxdb_ioxd/src/server_type/querier/mod.rs
```diff
@@ -43,7 +41,6 @@ impl<C: QuerierHandler> QuerierServerType<C> {
         Self {
             server,
             database,
-            shutdown: CancellationToken::new(),
```
alamb, line 46: 👍

PR #4047, opened 2022-03-16
refactor: make `QueryDatabase::chunks` async

crepererum:
For OG we can determine the chunks w/o any IO, for NG however this might
require a few catalog queries.

This is likely not the last change of this sort, i.e. the whole schema
handling is currently sync as well.


query/src/frontend/influxrpc.rs
```diff
@@ -1885,87 +1886,113 @@ mod tests {
...
 
-    /// Given a `TestDatabase` plans a InfluxRPC query
-    /// (e.g. read_filter, read_window_aggregate, etc). The test below
-    /// ensures that predicates are simplified during query planning.
-    type PlanRPCFunc = dyn Fn(&TestDatabase, InfluxRpcPredicate) + Send + Sync;
-
     /// Runs func() and checks that predicates are simplified prior to
     /// sending them down to the chunks for processing.
-    async fn run_test(func: &'static PlanRPCFunc) {
+    async fn run_test<T>(func: T)
```
alamb, line 1992: Is there a reason to rewrite this to be generic rather than `dyn` dynamic dispatch? 

The reason I originally wrote it with dynamic dispatch was part of the "query_tests` were already really slow to compile

I don't think it really matters but i was curious about the rationale

alamb (approved):

e-dard (approved):
Change seems fine.

query/src/frontend/influxrpc.rs
crepererum: I think it was borrowing and async functions that made this a PITA.

crepererum: I think it was borrowing and async functions that made this a PITA.

PR #4048, opened 2022-03-16
feat: Group files by whether they overlap in time

### LINKED ISSUE #3949: NG Compactor: Group overlapped files for a given set of files

carols10cents:
Fixes #3949.

Probably not the absolute most efficient, but looks correct and would be worth profiling. Let me know if you'd like to see any other test cases!


pauldix (approved):
Looks like you hit the right cases. I really don't think we need to worry about profiling and performance on it since a single table is only likely to have at most a few dozen files within a partition.

compactor/src/compact.rs
```diff
@@ -456,13 +455,48 @@ impl Compactor {
 
         Ok(Some(compacted_data))
     }
+
+    /// Given a list of parquet files that come from the same Table Partition, group files together
+    /// if their (min_time, max_time) ranges overlap. Does not preserve or guarantee any ordering.
+    fn overlapped_groups(mut parquet_files: Vec<ParquetFile>) -> Vec<Vec<ParquetFile>> {
+        let mut groups = Vec::with_capacity(parquet_files.len());
```
domodwyer, line 462: First thing I looked for! Pre-allocation of known sized outputs 👍

domodwyer (approved):
```
       ┌─┐ 
       ║"│ - looks great!!
       └┬┘ 
       ┌┼┐ 
        │  
       ┌┴┐ 
```

compactor/src/compact.rs
```diff
@@ -456,13 +455,48 @@ impl Compactor {
...
+                if in_group.iter().any(|group_file| {
+                    (file.min_time <= group_file.min_time && file.max_time >= group_file.min_time)
+                        || (file.min_time > group_file.min_time
+                            && file.min_time <= group_file.max_time)
+                }) {
+                    in_group.push(file);
+                } else {
+                    out_group.push(file);
+                }
+            }
```
NGA-TRAN, line 485: I think this algorithm will think file 2 is not a part of the overlapped group
``` text
. file 1:  |-----|
. file 2:                |--------|
. file 3:       |----------|
```

So if we have 4 files in the order like this:
``` text
. file 1:  |-----|
. file 2:                |--------------|
. file 3:       |----------|
. file 4:                           |-------|
```
They should be in one group but this algorithm will split them into 2 groups: [file1, file3], [file2, file4]
@alamb worked on grouping overlapped chunks before and may have some hints

NGA-TRAN (commented):
I think we may want to change the overlap algorithm a bit to avoid missing overlapped files

compactor/src/compact.rs
alamb: follow on work / discussion is in https://github.com/influxdata/influxdb_iox/pull/4082

Issue #3948, opened 2022-03-04
NG Compactor: get tombstones for a list of given parquet files


NGA-TRAN:
# Rationale
This I=ticket implements step 4 of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). The input is a lift of [parquet files](https://github.com/influxdata/influxdb_iox/blob/6ba5e518976a1feadb09b27b74967ac57b0e9971/data_types2/src/lib.rs#L624) and the output is a list of QueryableParquetChunk

```rust
fn tombstones(&[ParquetFile]) -> Result(Vector<QueryableParquetChunk>)

struct QueryableParquetChunk {
    data: Arc<ParquetChunk>, // data of the parquet file  
    delete_predicates: Vec<Arc<DeletePredicate>>,  // converted from tombstones
    table_name: String,  // needed to build query plan
}
```

NGA-TRAN:
@carols10cents I would say you want to build  `GroupWithTombstones` first (or something like that as an intermediate result). Then, you need to find the appropriate tombstones for each parquet file. So you will end up to have something like `ParquetFileWithTombstone` after that.

carols10cents:
@NGA-TRAN ooooh wait, looking into this more, the tombstones should still have the same sequencer_id and table_id as the group of parquet files, right? What about the sequence number? Should the tombstones' sequence numbers be greater than... the minimum max_sequence_number of the parquet files? The maximum max_sequence_number?

NGA-TRAN:
@carols10cents Yeah, all good points
1. the tombstones should still have the same sequencer_id and table_id as the group of parquet files. Yeah, we need to be careful here.
2. We only apply tombstones whose sequencer number > both min and max of the parquet files. Because of this, we may need to get tombstones per parquet file. Hah, this is tricky. I let you play with this and decide what is the best algorithm to have us read the catalog the least

PR #4049, opened 2022-03-16
feat: Group files by whether they overlap in time

### LINKED ISSUE #3949: NG Compactor: Group overlapped files for a given set of files

carols10cents:
Builds on #4048 Review that first, then don't review the first commit here because it's from the other PR.

Fixes #3949.


compactor/src/compact.rs
```diff
@@ -237,19 +242,35 @@ impl Compactor {
...
-            let overlapped_file_with_tombstones: Vec<ParquetFileWithTombstone> = vec![]; // TODO: #3948
-            overlapped_file_with_tombstones_groups.push(overlapped_file_with_tombstones);
+        let mut repo = self.catalog.repositories().await;
+        let tombstone_repo = repo.tombstones();
+        let mut overlapped_file_with_tombstones_groups =
+            Vec::with_capacity(overlapped_file_groups.len());
+
+        // For each group of overlapping parquet files,
+        for parquet_files in overlapped_file_groups {
+            let mut parquet_files_with_tombstones = Vec::with_capacity(parquet_files.len());
```
carols10cents, line 260: This also feels like it could be inefficient with all the `Vec` allocations, and feels like we're re-iterating through collections we've already iterated through. Is there any reason not to query for the tombstones for each parquet file earlier? Orrrrr we could even be fancy and query the catalog for both the parquet files *and* their joining tombstones at once. Maybe something to think about and profile after we're confident about correctness.

NGA-TRAN: We do not want to query tombstone for each file. Reading PG catalog is not cheap. The idea here is we can do it by either:
1. For each group of overlapped files, compute min time  & max time. Then query the catalog to get all tombstones that overlap with that min-max range of the group. Then apply those tombstones to each file of the group they overlap with. Note that we need to re-check if the tombstone overlaps with the file because overlapping with the file's group does not mean overlap with the file.
2. Compute min time and max time for all the groups. Then query the catalog for tombstones overlapping with that time range. Then applied to each file (again, need to check if the tombstone overlaps with that file)

. #2 will read catalog for tombstones only once but the number of returned tombstones will be likely more. It can be done even before grouping overlapped files.
. #1 will read catalog for tombstones per overlapped group but the return files will be more likely to get applied to all files in the group.

I think performance is more important here and I would choose to go with #2

NGA-TRAN: And I do no think using join to get parquet and tombstone at the same time will provide us good performance but it is an option for us to try/compare if we hit performance issue here

iox_catalog/src/interface.rs
```diff
@@ -405,6 +405,12 @@ pub trait TombstoneRepo: Send + Sync {
         sequencer_id: SequencerId,
         sequence_number: SequenceNumber,
     ) -> Result<Vec<Tombstone>>;
+
+    /// Return all tombstones relevant to a particular parquet file. Used during compaction.
+    async fn list_tombstones_for_parquet_file(
+        &mut self,
+        parquet_file: &ParquetFile,
+    ) -> Result<Vec<Tombstone>>;
 }
```
NGA-TRAN, line 414: I think we prefer `list_tombstones_for_time_range` here and a time range can cover time ranges of many parquet files. Each returned tombstone will be checked if it overlaps with each file in that time range.

iox_catalog/src/interface.rs
```diff
@@ -1249,7 +1256,219 @@ pub(crate) mod test_helpers {
...
+            matching_tombstone3,
+        ];
+        let mut expected_ids: Vec<_> = expected.iter().map(|t| t.id).collect();
+        expected_ids.sort();
+
+        assert_eq!(
+            tombstones_ids, expected_ids,
+            "\ntombstones: {:#?}\nexpected: {:#?}\nparquet_file: {:#?}",
+            tombstones, expected, parquet_file
+        );
```
NGA-TRAN, line 1471: After we modify and only read catalog once per group (or once for all groups), we want to add tests in which one or all only overlap with a subset of files in the group

NGA-TRAN (changes requested):
The idea is there, @carols10cents . We only need to modify to read catalog for tombstones only either once per overlapped group or even once for all groups

carols10cents:
@NGA-TRAN Ok, getting tombstones is now only doing one catalog query per overlap group... please let me know if this is correct!

compactor/src/compact.rs
```diff
@@ -664,6 +664,90 @@ impl Compactor {
...
+            // For a tombstone to be relevant to any parquet file, the tombstone must have a
+            // sequence number greater than the parquet file's max_sequence_number. If we query
+            // for all tombstones with a sequence number greater than the smallest parquet file
+            // max_sequence_number in the group, we'll get all tombstones that could possibly
+            // be relevant for this group.
+            let overall_min_max_sequence_number = parquet_files
+                .iter()
+                .map(|pf| pf.max_sequence_number)
+                .min()
+                .expect("The group was checked for emptiness above");
```
carols10cents, line 704: I hope this maybe clears up [what I was trying to say](https://github.com/influxdata/influxdb_iox/issues/3948#issuecomment-1074185211) about "the `min` of the `max_sequence_number`s"

compactor/src/compact.rs
```diff
@@ -664,6 +664,90 @@ impl Compactor {
...
+
+            // Query the catalog for the tombstones that could be relevant to any parquet files
+            // in this group.
+            let tombstones = tombstone_repo
+                .list_tombstones_for_time_range(
+                    // We've previously grouped the parquet files by sequence and table IDs, so
+                    // these values will be the same for all parquet files in the group.
+                    parquet_files[0].sequencer_id,
+                    parquet_files[0].table_id,
+                    overall_min_max_sequence_number,
```
carols10cents, line 714: The min of the max_sequence_number sets the floor that *all* tombstones must exceed in order to be considered as part of the group, so it's passed to the one catalog query done for the group

compactor/src/compact.rs
```diff
@@ -664,6 +664,90 @@ impl Compactor {
...
+            let parquet_files = parquet_files
+                .into_iter()
+                .map(|data| {
+                    // Filter the set of tombstones relevant to any file in the group to just those
+                    // relevant to this particular parquet file.
+                    let relevant_tombstones = tombstones
+                        .iter()
+                        .cloned()
+                        .filter(|t| {
+                            t.sequence_number > data.max_sequence_number
```
carols10cents, line 730: And then here, in memory, we assign the relevant tombstones to each parquet file based on whether the tombstone's sequence number is greater than this parquet file's max_sequence_number in particular.

compactor/src/compact.rs
NGA-TRAN: 👍 for min of all max

compactor/src/compact.rs
NGA-TRAN: I think this approach of reading PG catalog once per group and then, in-memory, assigning the accurate ones to each file will give us the best performance 👍

NGA-TRAN (approved):
Looks great @carols10cents

PR #4051, opened 2022-03-17
fix: docker server fixture had mistake in run command

lukebond:
It seems I'm the only one who doesn't have an iox binary on my laptop for the test fixtures to use, so it uses Docker, and that code path either never worked or got stale.

The addition of the `container` term is because `docker run` has long been deprecated but will probably never be removed. The real fix is the removal of the superfluous `database` term.

### DIFF
--- a/influxdb2_client/tests/common/server_fixture.rs
+++ b/influxdb2_client/tests/common/server_fixture.rs
@@ -214,8 +214,8 @@ impl TestServer {
             let container_name = format!("influxdb2_{}", http_port);
 
             Command::new("docker")
+                .arg("container")
                 .arg("run")
-                .arg("database")
                 .arg("--name")
                 .arg(&container_name)
                 .arg("--publish")


carols10cents (approved):
I swear this did work at one point, but I definitely haven't been running them regularly 😅 Sorry about that!! LGTM!!

crepererum:
`test_perf` is broken

carols10cents:
```
rustup-init.sh: OK
info: downloading installer
curl: (35) OpenSSL SSL_connect: Connection reset by peer in connection to static.rust-lang.org:443 
rustup: command failed: downloader https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init /tmp/tmp.GTpiqnhmOG/rustup-init x86_64-unknown-linux-gnu

```

Looks intermittent; I'm rerunning

crepererum (approved):

PR #4052, opened 2022-03-17
refactor: lower pg idle connection timeout

domodwyer:
Can't hurt! 60s is still quite a lot but will definitely reduce connection hogging.

---

* refactor: lower pg idle connection timeout (0d4949cd1)

      Configure the postgres catalog to close unused connections after 1 minute,
      rather than 500s to introduce a bit of fluidity to pool of connection
      acquires.


lukebond (approved):

mkmik (approved):
LGTM!

Issue #3934, opened 2022-03-04
NG Querier: Epic for Run query_tests against querier


alamb:
The idea here is to update the `query_test` crate to add coverage of the new querier component 

the definition of done is that some (exact subset tbd) of the query tests will also load data via the new ng flow (aka parquet files in the new Catalog, perhaps via the ingester), run queries against it, and validate that the answers are correct

Some possible steps:
- [x] Decouple `query_test` from using `Db` directly (either use a trait or an Enum) to prepare to use new Querier
- [x] For each existing scenario, put the data into a parquet file, create the appropriate entry in the data in the `iox_catalog` and instantiate a querier, and then runs the query tests against it
- [x] Add additional data setup to query_tests: Data in Multiple Parquet Files
- [ ] #4296
- [ ] Add additional data setup to query_tests: Data in Ingester structures + Parquet
- [ ] Add additional data setup to query_tests: different cache contents, etc
- [ ] #4459

alamb:
cc @tustvold  and @crepererum  -- hopefully this captures what we discussed face to face

alamb:
Assigning to @crepererum  as he is cranking along on this

crepererum:
I pause that for now, because I think it would be helpful to have a least a 2nd chunk stage (= ingester) before fully polishing the test framework.

alamb:
I think this is basically not tracking any planned work any more and the querier runs query_tests,  so closing

PR #4053, opened 2022-03-17
refactor: dyn-typed DB for `query_tests`

crepererum:
To test the `db::Db` as well as the `querier` with the same test
framework, they require a shared interface. Ideally this interface is
dynamically typed instead of static dispatched via generics because:

- `query_tests` already take ages to compile
- we often hold a list of scenarios and a single scenario should (in a
  future PR) be able to represent both OG as well as NG

The vision here is that we basically keep the whole test setup but add
new scenarios which are NG-specific later on.

Now the issue w/ many query-related types is that they are NOT
object-safe because methods that don't take `&self` or they have
associated types that we cannot specify in general for OG and NG at the
same time.

So we need a bunch of wrappers that make dynamic dispatch possible. They
mostly call to an internal "interface" crate which is the actual `dyn`
part. The interface is currently only implemented for OG.

The scenarios currently also only contain OG databases. However,
creating a dynamic interface that can be used in all `query_tests` is
already a huge step.

Note that there are two places where we downcast the dynamic/abstract
database to `db::Db` again:

1. To create one scenario based on another and where we need to
   manipulate `db::Db` with OG-specific semantics.
2. `server_benchmarks`. These contain OG databases only and there is no
   point in benchmarking throw the dynamic dispatch interface because
   prod (`influxdb_ioxd`) also uses static dispatch.

Ref #3934.


query_tests/src/db.rs
```diff
@@ -0,0 +1,421 @@
...
+        f.debug_tuple("Error").field(&self.0).finish()
+    }
+}
+
+impl Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(&self.0, f)
+    }
+}
+
```
alamb, line 41: We could almost avoid the newtype wrapper and do something like

```
type Error = String + Send + Sync;
```

😆 
this is fine too , I was just wondering if the extra level of wrapping was making things unecessairly complicated, perhaps

query_tests/src/db.rs
```diff
@@ -0,0 +1,421 @@
...
+impl std::error::Error for Error {}
+
+/// Abstract database used during testing.
+///
+/// This is required to make all the involved traits object-safe.
```
alamb, line 46: 👍  Maybe as a follow on PR we can make the underlying traits object-safe and then we can remove this particular abstraction. I am not sure what can't be made object safe (other than errors 🤔 )

server_benchmarks/benches/read_filter.rs
```diff
@@ -77,6 +77,9 @@ fn execute_benchmark_group(c: &mut Criterion, scenarios: &[DbScenario]) {
         let DbScenario { scenario_name, db } = scenario;
         let mut group = c.benchmark_group(format!("read_filter/{}", scenario_name));
 
+        // downcast Db for performance
+        let db = db.old_db().unwrap();
```
alamb, line 81: I doubt it really matters for performance (hopefully this benchmark wouldn't be sensitive to an extra dynamic dispatch) but I also don't think it is terrible either

🚀

query_tests/src/db.rs
```diff
@@ -0,0 +1,421 @@
...
+    fn chunk_type(&self) -> &str {
+        self.0.chunk_type()
+    }
+
+    fn order(&self) -> ChunkOrder {
+        self.0.order()
+    }
+}
+
+impl QueryChunkMeta for AbstractChunk {
```
alamb, line 189: The extra level of indirection would be great to remove eventually.

alamb (approved):
Thank you for this

query_tests/src/db.rs
crepererum: I think mostly this would be:

- the error
- Returning `Arc<dyn Chunk>` or `Box<dyn Chunk>` instead of having a real chunk type
- Not accepting `self: &Arc<Self>` but `&self` (otherwise vtable dispatch doesn't work)

NGA-TRAN (commented):
So nice. Thanks @crepererum

PR #4054, opened 2022-03-17
refactor: rework `TableCache

crepererum:
- dual cache that can also map table names to IDs
- deal w/ missing tables w/o panics
- set proper timeouts to missing data

For #3974.


querier/src/cache/table.rs
```diff
@@ -75,31 +168,60 @@ mod tests {
...
+
+        let cache = TableCache::new(
+            catalog.catalog(),
+            BackoffConfig::default(),
+            catalog.time_provider(),
+        );
+
+        let none = cache.name(TableId::new(i32::MAX)).await;
+        assert_eq!(none, None);
+        assert_histogram_metric_count(&catalog.metric_registry, "table_get_by_id", 1);
```
alamb, line 210: Since this test is checking `table_by_name` I wonder if the metrics should be checking `table_get_by_name` rather than `table_get_by_id`?

alamb (approved):

querier/src/cache/table.rs
crepererum: > Since this test is checking `table_by_name`...

The test is checking `TableCache::name(&self, id: TableId)` which is "get name by ID", which in turn maps to the catalog operation "get table by ID" (we extract the name from the table object).

Issue #3951, opened 2022-03-04
NG Cmpactor: Persist a given set of record batch and its corresponding IoxMetadata


NGA-TRAN:
This ticket implements step 6.b.ii of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx). 

```rust
fn persist(&[RecordBatch], IoxMetadata) -> Result<ParquetFile>
``` 
I think this one will be a simpler version of https://github.com/influxdata/influxdb_iox/blob/7b436d37cd1d32f224cb18d54b9b3ff1881ccc6f/ingester/src/data.rs#L171

carols10cents:
This issue now includes inserting the new parquet file into the catalog because that's straightforward to do

carols10cents:
Actually actually, [I misunderstood](https://github.com/influxdata/influxdb_iox/pull/4058#discussion_r829487915), so this issue will now only cover persisting the files to object storage and no change to the catalog.

Issue #3952, opened 2022-03-04
NG Compactor: Update Catalog for processed compaction


NGA-TRAN:
This ticket implements step 6.b.iii of [Find and Compact files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.y3uqgzkpk0vx).

carols10cents:
Actually, doing this and #3951 together will be straightforward.

carols10cents:
Actually actually, [I misunderstood](https://github.com/influxdata/influxdb_iox/pull/4058#discussion_r829487915), so this issue will now cover all updates to the catalog after a compaction.

PR #4058, opened 2022-03-17
feat: For each compacted data set, persist new parquet file to object store

carols10cents:
Fixes https://github.com/influxdata/influxdb_iox/issues/3951.

The code added here shares some code with ingester persistence; I'm happy to extract the duplication into... a persister crate? Wasn't sure if it was big enough to warrant at this point.

~~Also note I decided to make *insertion* of the new parquet file into the catalog part of the scope of this issue; #3952 is now only for *updating* the old parquet file data in the catalog.~~

I misunderstood; this now only covers persistence to object storage and all updates to the catalog will be done for #3952.


compactor/src/compact.rs
```diff
@@ -269,17 +280,16 @@ impl Compactor {
...
+                let (split_output1, split_output2): (CompactedData, CompactedData) =
+                    Self::split_compacted(compacted)?;
 
-            for _file in output_parquet_files {
-                // persist the file
-                // TODO: #3951
-
-                // update the catalog
-                // TODO: #3952
+                self.persist_and_update_catalog(split_output1).await?;
```
pauldix, line 288: Writing the file to object storage should be decoupled from updating the catalog. The process should be something like: 
1. Persist files to object storage
2. In a single database transaction:
    1. Add new parquet files
    2. Mark replaced parquet files to delete

The thing is that these new Parquet files should replace the old files. So we need to add them all to object storage first and then do the create and replace in a single transaction to ensure that the queriers have a consistent view.

compactor/src/compact.rs
```diff
@@ -456,6 +466,97 @@ impl Compactor {
         Ok(Some(compacted_data))
     }
 
+    fn split_compacted(_compacted_data: CompactedData) -> Result<(CompactedData, CompactedData)> {
+        // split the compacted data into 2 files 90/10
```
pauldix, line 470: This could actually amount to more in theory. If we have five overlapping input files and the result would be too large for two files to fit because of our size limits per file, it might be 3 output files each with non-overrlapping times.

compactor/src/compact.rs
```diff
@@ -456,6 +477,78 @@ impl Compactor {
...
+    /// Returns the persisted file size (in bytes) and metadata if a file was created.
+    async fn persist(
+        metadata: &IoxMetadata,
+        record_batches: Vec<RecordBatch>,
+        object_store: Arc<DynObjectStore>,
+    ) -> Result<Option<(usize, IoxParquetMetaData)>> {
+        if record_batches.is_empty() {
+            return Ok(None);
+        }
+        let schema = record_batches
```
NGA-TRAN, line 497: I would add a comment here saying all given `record_batches` have the same schema

NGA-TRAN (approved):

PR #4063, opened 2022-03-18
refactor: `querier` w/o `db`

### LINKED ISSUE #3974: NG Querier: Move aways from `db`

crepererum:
On top of #4054.

Get the `querier` to work w/o relying on `db`. A few notes:

- Testing is kinda shallow, we really need to get `query_tests` working
  w/ `querier` (see https://github.com/influxdata/influxdb_iox/issues/3934).
- We still run a sync loop for namespaces, tables and schemas. This will
  be a replaced by "update namespace incl. tables and schemas on demand".
  Note however that we cannot fetch single tables and schemas on demand
  at the moment, because DataFusion doesn't implement async schema
  inspection (only `scan` / "give me all the chunks" is async). I think
  that's OK for now and we can address this later.
- There is NO cache for parquet files and tombstones at the moment. For
  correctness, they need to be fetched in a single transaction (or we
  need a kinda tricky sequence number / logical clock tracking) and I am
  not sure yet how this makes sense when we have the ingester data wired
  up and predicates pushed down to the catalog (see next point). So
  let's measure first and then decide on a caching strategy for this.
- Predicates are currently NOT pushed down to the catalog. I'll need to
  figure out how to extract time range from generic DataFusion
  expressions to make that work (it's easier for InfluxRPC queries, but
  they are not tested at the moment, see first point).

Sorry that this commit is kinda huge. I initially planned to only
migrate the chunks away from `db` and leave the tables and schemas for a
follow-up PR, but the DataFusion trait structure (chunks are bound to
their tables) makes this kinda pointless.

Closes #3974.


querier/Cargo.toml
```diff
@@ -12,13 +12,11 @@ bytes = "1.0"
 client_util = { path = "../client_util" }
 data_types2 = { path = "../data_types2" }
 datafusion = { path = "../datafusion" }
-db = { path = "../db" }
```
alamb, line 15: 🎉

querier/src/chunk/query_access.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn apply_predicate_to_metadata(
+        &self,
+        predicate: &predicate::Predicate,
+    ) -> Result<predicate::PredicateMatch, QueryChunkError> {
+        let pred_result = match &self.storage {
+            ChunkStorage::Parquet { chunk, .. } => {
+                if predicate.has_exprs() || chunk.has_timerange(predicate.range.as_ref()) {
+                    PredicateMatch::Unknown
+                } else {
+                    PredicateMatch::Zero
```
alamb, line 71: I am not 100% sure about this logic (there are also table filters and column selection/filters that can be in `Predicates`). However, I think the metadata queries are well covered by `query_tests` so as we wire those tests up, that will shake out any bugs in this code too

querier/src/chunk/query_access.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn column_names(
+        &self,
+        _ctx: query::exec::IOxExecutionContext,
+        predicate: &predicate::Predicate,
+        columns: schema::selection::Selection<'_>,
+    ) -> Result<Option<query::exec::stringset::StringSet>, QueryChunkError> {
+        match &self.storage {
+            ChunkStorage::Parquet { chunk, .. } => {
+                if !predicate.is_empty() {
+                    return Ok(None);
```
alamb, line 88:
```suggestion
                    // if there is anything in the predicate, bail for now and force a full plan
                    return Ok(None);
```

querier/src/chunk/query_access.rs
```diff
@@ -0,0 +1,150 @@
...
+            .delete_predicates()
+            .iter()
+            .map(|pred| Arc::new(pred.as_ref().clone().into()))
+            .collect();
+        ctx.set_metadata("delete_predicates", delete_predicates.len() as i64);
+
+        // merge the negated delete predicates into the select predicate
+        let mut pred_with_deleted_exprs = predicate.clone();
+        pred_with_deleted_exprs.merge_delete_predicates(&delete_predicates);
+        debug!(?pred_with_deleted_exprs, "Merged negated predicate");
```
alamb, line 126: I know this is just following the model in DB, but I don't like it. Some day I'll figure out how to clean it up

querier/src/table/query_access.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    async fn scan(
+        &self,
+        projection: &Option<Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+        // build provider out of all chunks
+        // TODO: push down some predicates to catalog
+        let mut builder = ProviderBuilder::new(self.name(), Arc::clone(self.schema()));
```
alamb, line 43: I have also found it invaluable while debugging query_test bugs in OG to have `debug!` logging  here that say how many chunks are pruned (because that helps isolate where the data is incorrectly removed) -- we can add it as a follow on PR

querier/src/namespace/query_access.rs
```diff
@@ -1,41 +1,112 @@
...
+                return vec![];
+            }
+        };
+
+        let mut chunks = table.chunks().await;
+
+        // if there is a field restriction on the predicate, only
+        // chunks with that field should be returned. If the chunk has
+        // none of the fields specified, then it doesn't match
+        // TODO: test this branch
```
alamb, line 50: is covered in `query_tests`

querier/src/namespace/query_access.rs
```diff
@@ -1,41 +1,112 @@
...
-        ctx: &IOxExecutionContext,
-        query_type: &str,
-        query_text: QueryText,
+        _ctx: &IOxExecutionContext,
+        _query_type: &str,
+        _query_text: QueryText,
     ) -> QueryCompletedToken {
-        self.catalog_access
-            .record_query(ctx, query_type, query_text)
+        // TODO: implement query recording again
```
alamb, line 71: this may be worth a tracking ticket (it is likely just a plumbing exercise)

querier/src/namespace/query_access.rs
```diff
@@ -1,41 +1,112 @@
...
+    }
+}
+
+impl CatalogProvider for QuerierCatalogProvider {
+    fn as_any(&self) -> &dyn Any {
+        self as &dyn Any
+    }
+
+    fn schema_names(&self) -> Vec<String> {
+        // TODO: system tables
```
alamb, line 95: likewise, getting the system tables back is probably worth a tracking ticket as well

querier/src/namespace/mod.rs
```diff
@@ -163,484 +119,36 @@ impl QuerierNamespace {
...
-
-                    let chunk = chunk.upgradable_read();
-                    if chunk.delete_predicates() != predicates_filtered {
-                        let mut chunk = RwLockUpgradableReadGuard::upgrade(chunk);
-                        chunk.set_delete_predicates(predicates_filtered);
-                    }
-                }
-            }
-        }
+        *self.tables.write() = Arc::new(tables);
```
alamb, line 141: I think making a snapshot of the current set of tables makes sense to me

alamb (approved):

alamb:
Nice work @crepererum  -- I think we need to get this in and then update `query_tests` to get coverage, but the basics look good here

querier/src/chunk/query_access.rs
crepererum: Yeah, we would have a chat about delete predicate handling at some point because I think it's a bit weird (e.g. we include the delete predicates here somewhat manually but at the same time `QueryChunk` also provides a way for the query engine to read out the list of predicates).

querier/src/table/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+                let mut delete_predicates = Vec::with_capacity(tombstones.len());
+                for tombstone in tombstones {
+                    // check conditions that don't need catalog access first to avoid unnecessary catalog load
+
+                    // check if tombstone even applies to the sequence number range within the parquet file
+                    if chunk.meta().min_sequence_number() > tombstone.sequence_number() {
+                        continue;
+                    }
+
+                    // TODO: also consider time ranges
```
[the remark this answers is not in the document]
crepererum: added a tracking issue

querier/src/namespace/query_access.rs
crepererum: added  a tracking issue

querier/src/namespace/query_access.rs
crepererum: done

querier/src/table/query_access.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl ChunkPruner for QuerierTableChunkPruner {
+    fn prune_chunks(
+        &self,
+        _table_name: &str,
+        table_schema: Arc<Schema>,
+        chunks: Vec<Arc<dyn QueryChunk>>,
+        predicate: &Predicate,
+    ) -> Vec<Arc<dyn QueryChunk>> {
+        // TODO: bring back metrics
```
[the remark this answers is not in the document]
crepererum: added a tracking issue

alamb:
🎉

PR #4069, opened 2022-03-18
test: assert uniform-ish sharder distribution

domodwyer:
Everyone loves a uniform sharder and a good test.

---

* test: assert uniform-ish sharder distribution (370cc4c15)

      I added this test to confirm the sharder produces a uniform distribution when
      investigating non-uniform sequencer writes - may as well keep it.


router2/src/sharder/jumphash.rs
```diff
@@ -219,4 +219,25 @@ mod tests {
...
+            let bucket = hasher.shard(format!("{}", i).as_str(), &namespace, &0);
+            *mapping.entry(bucket).or_default() += 1;
+        }
+
+        let (min, max) = mapping.values().fold((usize::MAX, 0), |acc, &v| {
+            let (min, max) = acc;
+            (min.min(v), max.max(v))
+        });
+
+        // ± 0.05%
```
alamb, line 240:
```suggestion
        // Expect that the number of values of each bucket are all within   ± 0.05%
        // of the total
```

alamb (approved):

router2/src/sharder/jumphash.rs
```diff
@@ -219,4 +219,25 @@ mod tests {
...
+        assert!(max - min < 5000, "min: {}, max: {}", min, max);
```
NGA-TRAN, line 241: The assert to discover uniform-ish looks great.
Question: How do we classify different input data to know that the hash algorithm always returns uniform-ish for all of them?

NGA-TRAN (approved):

router2/src/sharder/jumphash.rs
domodwyer: So long as the type implements the [`Hash`](https://doc.rust-lang.org/std/hash/trait.Hash.html) trait in a sensible way, the hashing should be reliable - the properties of the algorithm are evaluated in the paper I took it from here: https://arxiv.org/ftp/arxiv/papers/1406/1406.2294.pdf

PR #4080, opened 2022-03-21
refactor: dyn-dispatch chunks in query subsystem

crepererum:
- this is what DataFusion is doing as well; it's also fast enough
  because the number of chunks in a query is not THAT massive (it's not
  like we are doing row-level dyn dispatching)
- it simplifies abstracting over different databases
- it allows us to drop our enum-based dispatching that we have for
  `DbChunk` and that we would also need for the querier (e.g. depending
  on if a chunk is backed by a parquet file or ingester data)
- it likely speeds up compile times because the `query` is no longer
  contains massive amounts of generic code

For #3934.


db/src/access.rs
```diff
@@ -271,9 +280,7 @@ impl ChunkPruner<DbChunk> for ChunkAccess {
 }
 
 impl PruningObserver for TableAccessMetrics {
-    type Observed = DbChunk;
```
alamb, line 274: ❤️

query/src/provider/overlap.rs
```diff
@@ -101,39 +100,33 @@ where
...
-    C: QueryChunkMeta,
-{
+struct KeyStats<'a> {
     /// The index of the chunk
     index: usize,
 
     /// The underlying chunk
     #[allow(dead_code)]
-    chunk: &'a C,
+    chunk: &'a dyn QueryChunk,
```
alamb, line 119: Nice -- I got hung up on this particular structure when I was messing with the same idea 👍

query/src/pruning.rs
```diff
@@ -543,35 +539,35 @@ mod test {
...
             Some(10),
-        ));
+        )) as Arc<dyn QueryChunk>;
 
         let c2 = Arc::new(TestChunk::new("chunk2").with_i64_field_column_with_stats(
             "column1",
             Some(0),
             Some(1000),
-        ));
+        )) as Arc<dyn QueryChunk>;
```
alamb, line 548: Having to explicitly downcast definitely makes dynamic dispatch in rust a little more annoying than in other languages, but hopefully it will get sorted out soonish

query_tests/src/db.rs
```diff
@@ -83,87 +80,6 @@ impl QueryDatabaseMeta for AbstractDb {
     }
 }
 
-#[derive(Debug)]
```
alamb, line 86: 🎉

Issue #4055, opened 2022-03-17
Add API endpoint for fetching schema for a namespace


lukebond:
Should be added to router2.

Should return:
- namespace id, name, retention duration
- table names and ids
- columns (name, id, type)
- (later, when we have them:) sharding and partitioning rules (strings on namespace)

Provide CLI command to call this endpoint.

PR #4081, opened 2022-03-21
feat: schema grpc server & proto in router2

lukebond:
Part of #4055 

- adds proto for a schema service and one fn: get_schema
- adds schema service to router2
- adds get_schema impl that uses the existing `get_schema_by_name()` fn used by router2 for schema validation
- adds a test

Note there is no HTTP endpoint for this. If we need one then I'll add one (I figure we need a HTTP write/delete API but don't necessarily need everything else to be in both gRPC/HTTP? LMK in the review if you disagree). Also note that the proto doesn't return the same fields listed in the issue; it's because I'm reusing an existing fn and those fields were relatively arbitrary anyway. I figure if those are the fields needed for schema validation then it's probably going to meet the needs of consumers of this API.


generated_types/protos/influxdata/iox/schema/v1/service.proto
```diff
@@ -0,0 +1,34 @@
...
+
+message GetSchemaResponse {
+  NamespaceSchema schema = 1;
+}
+
+message NamespaceSchema {
+  int32 id = 1;
+  int32 kafka_topic_id = 2;
+  int32 query_pool_id = 3;
+  map<string, TableSchema> tables = 4;
```
mkmik, line 23: it may be worth stating in a comment what is the key of this map (likely the table name, right?)

e-dard (approved):

PR #4082, opened 2022-03-21
fix: incorrect overlapped grouping

carols10cents:
This is a failing test to demonstrate the problem @NGA-TRAN found in this comment: https://github.com/influxdata/influxdb_iox/pull/4048/files#r830582335

I wanted to make sure this didn't get lost!

@alamb Nga mentioned you've done work similar to this before, do you have any pointers to existing code I could look at?


alamb (commented):
@carols10cents  -- All the overlap test I know of are in https://github.com/influxdata/influxdb_iox/blob/main/query/src/provider/overlap.rs#L269-L764

alamb:
I see now that -- `overlapped_groups` implements its own overlap optic maybe could be rewritten to use the `overlap` module in the query engine (aka fix the bug by deleting some code :) )

carols10cents:
Hmmmmm, reading the `overlap` module in the query engine, it looks like it operates on `QueryChunk`s, and [`QueryChunk` has a lot of behavior](https://github.com/influxdata/influxdb_iox/blob/9ec45f5aec9eedd7bf61934018c8b17d9478d8f7/query/src/lib.rs#L157-L232) that the `ParquetFile`s in the compactor won't have? And the `overlap` module is looking for overlaps in all columns, not only `time`? Am I missing a way that reusing this code could work?

carols10cents:
Also uhhhhhhh bad news, query overlap has the same problem 😬

alamb:
> Hmmmmm, reading the overlap module in the query engine, it looks like it operates on QueryChunks, and [QueryChunk has a lot of behavior](https://github.com/influxdata/influxdb_iox/blob/9ec45f5aec9eedd7bf61934018c8b17d9478d8f7/query/src/lib.rs#L157-L232) that the ParquetFiles in the compactor won't have? And the overlap module is looking for overlaps in all columns, not only time? Am I missing a way that reusing this code could work?


The querier (as of this morning, thanks to @crepererum )) now implements `QueryChunk` for parquet files 

https://github.com/influxdata/influxdb_iox/blob/55643945a1a654d0db0bcc5eb0a42d7c3185efa9/querier/src/chunk/mod.rs#L102-L115

And you can make a `ParquetChunk` from a `ParquetFile` like this:

https://github.com/influxdata/influxdb_iox/blob/c6cab3538f9153d67d3f908c1a9f12e0894895e5/parquet_file/src/chunk.rs#L113-L150

An alternate is to pull out just what the overlap code needs (a few of the methods on `QueryChunk`) into its own trait and implement that for the compactor, perhaps?

alamb:
> Also uhhhhhhh bad news, query overlap has the same problem 😬

That is good news! We found a bug!

NGA-TRAN:
You have 2 beautiful sets of tests @carols10cents  😄

NGA-TRAN:
Regarding what @alamb said about `QueryChunk` and `ParquetChunk`, the compactor has a similar structure QueryableParquetChunk https://github.com/influxdata/influxdb_iox/blob/886f9dc8c1ab3fa7be4672e7457ed8d5fd07b6b0/compactor/src/query.rs#L40. But I think since we do not care about their sequence number at this step in NG compactor (it was taken care after that  by the this [order](https://github.com/influxdata/influxdb_iox/blob/886f9dc8c1ab3fa7be4672e7457ed8d5fd07b6b0/compactor/src/query.rs#L216) function), I think all information in the `ParquetFile` or `ParquetChunk` is good enough for grouping overlap files if you can make it that way.

carols10cents:
@NGA-TRAN @alamb Ok, so I found [a blog post about a way to find overlapping groups](https://towardsdatascience.com/overlapping-time-period-problem-b7f1719347db), and I have a little implementation that works for the test case in the compactor, where there's *one* non-null value for the min/max.... my brain is hurting trying to see if/how this can be generalized for the potentially-null, potentially-multi-column case in the query overlap code..... what do you think?

NGA-TRAN:
@carols10cents Oh man, this will be tricky. I need a new fresh day to read this and think about it. My brain also hurt when I was testing `find_and_compact`. So many details about overlaps on min max time, min max sequence, old new files, ... LOL

alamb:
I still don't understand why we need a new `overlap` implementation (I feel like we should be using the one in `overlap.rs` and if it has a bug fixing that)

carols10cents:
Well, ideally they would share an implementation, but the code in overlap.rs is much more complex given that it supports calculating overlap with multiple columns that may have null values. I wanted to start with the simpler case of the parquet files that only need to be checked for overlap on time and always have a min time and max time, to make sure this algorithm I found fixes the bug Nga found, and it does, but my brain hurts to generalize it more to the query chunk case.

NGA-TRAN:
@carols10cents The reading is great. I think the NULL and missing column case will be way more complicated than this only non-null min and max. If It is fine with @alamb, we just commit this one. Then have a different ticket to think about Null and missing columns and consolidate this code with `overlap.rs` in that ticket

alamb:
I think merging this PR and filing a ticket to combine the logic would be good 👍

NGA-TRAN (approved):

NGA-TRAN:
@carols10cents : I have merged main into this branch. Can you have a quick look to see if I made any silly mistake while resolving conflicts?

NGA-TRAN:
I am merging this to continue with #4158 which will use this

PR #4092, opened 2022-03-21
fix: teach semantic commits script about GH revert PRs

lukebond:
Github creates revert PRs that don't match the patterns in the semantic commit checker and I don't think you can customise them. This PR teaches the script to handle those commits (includes tests!).


jacobmarble (approved):

.github/workflows/semantic_check.sh
```diff
@@ -25,6 +25,7 @@ if [[ $1 == "test" ]]; then
   expect_ok="chore: foo
 chore(hello): foo
 CHORE: foo
+Revert \"fix: certain this fix is correct!\"
```
alamb, line 28: 😆

PR #4095, opened 2022-03-21
fix: have ingester use single mutable batch for buffer

### LINKED ISSUE #4072: Update ingester to use a single mutable batch
I think the ingesters take considerably more memory than they need to at this point because they're not combining mutable batches on ingest. Since they're fairly idle, we should see what they look like if they collapse data into a single mutable batch within a partition.

pauldix:
Removed some unnecessary tests as they no longer apply with the new buffer structure. This will hopefully reduce the memory footprint of the ingesters significantly.

Closes #4072


mkmik (approved):
looks reasonable

PR #4098, opened 2022-03-22
test: run SOME query tests for querier

crepererum:
This includes some type changes to dispatch between OG and NG and allows
some tests to be run against the NG querier. This only contains parquet
files though, so it's somewhat a limited scope.

For #3934.


query_tests/cases/in/delete_multi_expr_one_chunk.expected
```diff
@@ -56,7 +56,7 @@
 +--------------+
 | 1            |
 +--------------+
--- SQL: SELECT foo from cpu;
+-- SQL: SELECT foo from cpu order by foo;
```
alamb, line 59: 👍  the classic. 

FWIW you can avoid changing the sql by using 
```
-- IOX_COMPARE: sorted
```

For example:
https://github.com/influxdata/influxdb_iox/blob/c4a234e83cfe480a5b0b19446ed60119dadb1d40/query_tests/cases/in/sql_information_schema.sql#L8

query_tests/src/scenarios/util.rs
```diff
@@ -8,22 +8,39 @@ use db::{
...
     /// Line protocol data of this chunk
     pub lp_lines: Vec<&'a str>,
     /// which stage this chunk will be created
-    pub chunk_stage: ChunkStage,
+    pub chunk_stage: ChunkStageOld,
 }
 
 #[derive(Debug, Clone)]
-pub enum ChunkStage {
+pub struct ChunkDataNew<'a> {
```
alamb, line 29: this is a very clever mechanism. 👍

query_tests/src/scenarios/util.rs
```diff
@@ -36,35 +53,111 @@ pub enum ChunkStage {
...
 }
 
 impl ChunkStage {
     /// return the list of all chunk types
     pub fn all() -> Vec<Self> {
-        vec![Self::Mubo, Self::Mubf, Self::Rub, Self::RubOs, Self::Os]
+        ChunkStageOld::all()
+            .into_iter()
+            .map(ChunkStage::Old)
+            .chain(ChunkStageNew::all().into_iter().map(ChunkStage::New))
```
alamb, line 117: 😍

query_tests/src/scenarios/util.rs
```diff
@@ -410,14 +602,154 @@ pub async fn make_chunk_with_deletes_at_different_stages(
...
+        }
+    }
+
+    // create chunks
+    match chunk_stage {
+        ChunkStageNew::Parquet => {
+            // need to use a temporary vector because BTree iterators ain't `Send`
+            let lp_lines_grouped: Vec<_> = lp_lines_grouped.into_iter().collect();
+
+            for (table_name, lp_lines) in lp_lines_grouped {
```
alamb, line 686: I think reading this code that it will only ever create a single parquet file for each table, but I am not 100% sure. Is that also your understanding?

query_tests/src/scenarios/util.rs
```diff
@@ -410,14 +602,154 @@ pub async fn make_chunk_with_deletes_at_different_stages(
...
+    DbScenario { scenario_name, db }
+}
+
 /// Build many chunks which are in different stages
 //  Note that, after a lot of thoughts, I decided to have 2 separated functions, this one and the one above.
 //  The above tests delete predicates before and/or after a chunk is moved to different stages, while
 //  this function tests different-stage chunks in various stages when one or many deletes happen.
 //  Even though these 2 functions have some overlapped code, merging them in one
 //  function will created a much more complicated cases to handle
 pub async fn make_different_stage_chunks_with_deletes_scenario(
```
alamb, line 741:
```suggestion
pub async fn make_different_stage_chunks_with_deletes_scenario_new(
```

?

alamb (approved):
This is a very clever refactoring. 👍  nice work @crepererum 

FYI @NGA-TRAN  -- your chunk stage scenarios have been updated

query_tests/src/scenarios/util.rs
```diff
@@ -80,26 +173,84 @@ pub enum DeleteTime {
...
+    /// Return all DeleteTime at and after the given chunk stage
+    pub fn all_from_and_before(chunk_stage: ChunkStageOld) -> Vec<Self> {
+        match chunk_stage {
+            ChunkStageOld::Mubo => vec![Self::Mubo],
+            ChunkStageOld::Mubf => vec![Self::Mubo, Self::Mubf],
+            ChunkStageOld::Rub => {
+                vec![Self::Mubo, Self::Mubf, Self::Rub]
+            }
+            ChunkStageOld::RubOs => vec![Self::Mubo, Self::Mubf, Self::Rub, Self::RubOs],
+            ChunkStageOld::Os => vec![Self::Mubo, Self::Mubf, Self::Rub, Self::RubOs, Self::Os],
```
NGA-TRAN, line 186: 👍

query_tests/src/scenarios/util.rs
crepererum: This is the generic one. The "new" version is missing because `delete.rs` currently creates OG-specific data. So this is more a reminder for myself to fix `delete.rs`, implement OG/NG-dispatching in the generic version and add the NG variant.

query_tests/src/scenarios/util.rs
crepererum: Yes. As discussed in Slack I think our interpretation of `all_scenarios_for_one_chunk` and `make_chunk_...` is "1 chunk per table" (there are many tests that use multiple tables). This matches the behavior what OG (i.e. LP lines are grouped by table).

NGA-TRAN (approved):
Very nice. Thanks @crepererum for making OG tests available to NG. This tests are valuable to us.

query_tests/cases/in/delete_multi_expr_one_chunk.expected
crepererum: That's super helpful. I've changed it to `IOX_COMPARE` (w/o modifying the actual SQL query to use `order by`).

PR #4103, opened 2022-03-22
refactor: generalize query test scenarios a bit

crepererum:
Some query test scenarios are duplicates and are very OG specific. Let's
use generic scenarios (i.e. the ones that contain all chunk stages
instead of a specific one) where applicable.

For #3934.


NGA-TRAN (approved):
It looks ok if you have considered the case in my comment

query_tests/src/table_schema.rs
```diff
@@ -93,17 +80,17 @@ async fn list_schema_cpu_all_rub() {
         .unwrap();
 
     run_table_schema_test_case(
-        TwoMeasurementsRubScenario {},
+        TwoMeasurements {},
         Selection::All,
         "cpu",
         expected_schema,
-        Some(&sort_key),
+        None,
```
alamb, line 87: seems like the sort key was lost. Was that intentional? Maybe it should be updated to refer to columns in the new scenario?

query_tests/src/table_schema.rs
```diff
@@ -93,17 +80,17 @@ async fn list_schema_cpu_all_rub() {
         .unwrap();
 
     run_table_schema_test_case(
-        TwoMeasurementsRubScenario {},
+        TwoMeasurements {},
```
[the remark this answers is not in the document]
alamb: I think @NGA-TRAN  may not have implemented the fancier scenario generation when these tests were first written, which perhaps explains the duplication

query_tests/src/table_schema.rs
```diff
@@ -49,7 +50,12 @@ async fn run_table_schema_test_case<D>(
                     actual_schema
                 );
 
-                assert_eq!(chunk.sort_key(), expected_sort_key);
+                // There are a few cases where we don't care about the sort key:
```
alamb, line 53: something doesn't feel quite right to me about this

alamb: Maybe it was the change in expected sort key

query_tests/src/table_schema.rs
crepererum: ups, you're right :+1:

query_tests/src/table_schema.rs
crepererum: It's quite simple:

- MUBs in OG are never sorted
- all other representations in OG are sorted (but in some tests we don't care about that fact)

Actually it's only one test that deals w/ sort keys at all, otherwise I could just delete the whole block.

alamb: adding that rationale into the comments might help me when I see this in the future

alamb (approved):

query_tests/src/table_schema.rs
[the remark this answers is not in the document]
NGA-TRAN: That right @alamb. That todo has never been a high priority. Thanks @crepererum for making it happen

query_tests/src/table_schema.rs
crepererum: I've extended the code comment a bit.

PR #4105, opened 2022-03-22
feat: schema client and CLI

### LINKED ISSUE #4055: Add API endpoint for fetching schema for a namespace

lukebond:
Closes #4055 

Following on from #4081, this adds a client and CLI command to fetch the schema for a namespace.

To test against IOx all-in-one mode:
- launch a postgres (i use a docker container)
- run sqlx setup against it (see catalog docs)
- run psql to put some test data in:

```
psql $ set search_path to iox_catalog, public;
psql $ insert into kafka_topic (name) values ('franz');
psql $ insert into query_pool (name) values ('swimming');
psql $ insert into namespace (name, retention_duration, kafka_topic_id, query_pool_id) values ('namespace_table_test', 'inf', 1, 1);
psql $ insert into table_name (namespace_id, name) values (1, 'schema_test_table');
psql $ insert into column_name (table_id, name, column_type) values (1, 'schema_test_column', 1);
```

- run IOx in all-in-one mode (adjust dsn where appropriate):

```
$ LOG_FILTER=debug cargo run -- run   all-in-one   --write-buffer=file  --write-buffer-addr /tmp  --catalog-dsn postgresql://postgres:password@localhost:5432/iox_shared  --write-buffer-partition-range-start 0  --write-buffer-partition-range-end 0  --pause-ingest-size-bytes 10000  --persist-memory-threshold-bytes 1000
```

- run CLI:

```
$ cargo run -- -h http://localhost:8081 schema get namespace_table_test
    Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/influxdb_iox -h 'http://localhost:8081' schema get namespace_table_test`
{
  "id": 1,
  "kafkaTopicId": 1,
  "queryPoolId": 1,
  "tables": {
    "schema_test_table": {
      "id": 1,
      "columns": {
        "schema_test_column": {
          "id": 2,
          "columnType": 1
        }
      }
    }
  }
}
```


influxdb_iox/src/commands/schema.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("JSON Serialization error: {0}")]
+    Serde(#[from] serde_json::Error),
+
+    #[error("Client error: {0}")]
+    ClientError(#[from] influxdb_iox_client::error::Error),
+}
+
+/// Various commands for catalog manipulation
```
alamb, line 17:
```suggestion
/// Various commands for catalog schema inspection
```

influxdb_iox/src/commands/schema.rs
```diff
@@ -0,0 +1,52 @@
...
+    Get(Get),
+}
+
+pub async fn command(connection: Connection, config: Config) -> Result<(), Error> {
+    match config.command {
+        Command::Get(command) => {
+            let mut client = schema::Client::new(connection);
+            let schema = client.get_schema(&command.namespace).await?;
+            println!("{}", serde_json::to_string_pretty(&schema)?);
+        } // Deliberately not adding _ => so the compiler will direct people here to impl new
```
alamb, line 47: 👍

alamb (approved):
Makes sense to me. All it is missing is end to end tests -- lol - but we don't have a harness (yet) to add them. Perhaps you can file a follow on ticket to add an end to end test?

influxdb_iox/src/commands/schema.rs
lukebond: thanks, copy pasta error!

lukebond:
> Makes sense to me. All it is missing is end to end tests -- lol - but we don't have a harness (yet) to add them. Perhaps you can file a follow on ticket to add an end to end test?

Sure! I'm not fully abreast of what we have e2e-wise in IOx. If nothing, then I can create a ticket to add this test to the suite when the harness is there. I think that's what you meant?

alamb:
>> Makes sense to me. All it is missing is end to end tests -- lol - but we don't have a harness (yet) to add them. Perhaps you can file a follow on ticket to add an end to end test?

> Sure! I'm not fully abreast of what we have e2e-wise in IOx. If nothing, then I can create a ticket to add this test to the suite when the harness is there. I think that's what you meant?

There are end to end CLI tests for OG here: https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_cases/read_cli.rs

Here is the start of end to end tests for NG: https://github.com/influxdata/influxdb_iox/pull/4114. So once we have that merged, I think it should be a "simple" matter of porting the same CLI type test over for this command

lukebond:
Created here: https://github.com/influxdata/influxdb_iox/issues/4121

PR #4107, opened 2022-03-22
feat: add compactor configuration for kafka topic and sequencers

pauldix:


lukebond (approved):
LGTM!

NGA-TRAN (approved):
I am not familiar with this code but it looks like it does something similar to the one of ingester so 👍

alamb:
LOL turns out this changhe broke the all in one mode:

```shell
target/debug/influxdb_iox run -vv all-in-one --write-buffer=file --write-buffer-addr /tmp --catalog-dsn 'postgresql://localhost:5432/alamb' --write-buffer-partition-range-start 0 --write-buffer-partition-range-end 0 --pause-ingest-size-bytes 10000 --persist-memory-threshold-bytes 1000`
thread 'main' panicked at 'Command all-in-one: Argument names must be unique, but 'topic' is in use by more than one argument or group', /Users/alamb/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.1.6/src/build/debug_asserts.rs:76:9

```

(I only found it because my new snazzy end to end test in https://github.com/influxdata/influxdb_iox/pull/4114 started failing and merge from master)

alamb:
(I will fix the breakage, btw)

alamb:
https://github.com/influxdata/influxdb_iox/pull/4115

Issue #3968, opened 2022-03-07
ingester: Store consistent sort key for all parquet files of the same partition (and the same table, too?)


NGA-TRAN:
# Parts

- [x] https://github.com/influxdata/influxdb_iox/issues/4194
- [x] https://github.com/influxdata/influxdb_iox/issues/4195
- [x] https://github.com/influxdata/influxdb_iox/issues/4196

# Rationale

- Since sort order of a parquet file is an important performance factor in Querier, we will store the sort order in the parquet file.
- To avoid resort parquet files (chunks) for deduplication at read time, we want to have consistent sort order for all parquet files. To do this, the Ingester will compute the sort order during persistence of first parquet of a table based on the column cardinality from low to high. The following persistence of the same table will reuse that sort order. 

### TBD : 
Because different parquet files may contain different columns, the latest parquet file may not include all columns of the sort order. For example:

- Persist 1:  parquet file includes `tag1, tag2, tag3, field1, field2, time`.  The computed sort order: `tag3, tag1, tag2, time`
- Persist 2: parquet file includes `tag2, tag 3, tag4, field1, field2, time`.  This file does not have `tag1` but contains a new `tag4`. What will be the sort order of this file based on the sort order above? Either?
    1. `tag3, tag1, tag2, tag4, time` (Keep `tag1` and add `tag4` infant of `time`): If so, the sort order contains column `tag1` that is not in the file.
    2. `tag3, tag2, tag4, time` (ignore `tag1` because it is not in the file): If so, in order for us to know the sort order to reuse for future persistence, we need to read both parquet files.

### Design update: 
The ingester uses this [compact](https://github.com/influxdata/influxdb_iox/blob/5a29d070ea6f5d8e3e961bf4deea15ef29be463c/ingester/src/compact.rs#L113) to build a query plan to compact its writes.  Since the sort key is not provided, the plan will sort the column lexicographically for deduplication. In order to support consistent sort key and that sort key should be computed before compacting, this compact should be replaced with this [compact_plan](https://github.com/influxdata/influxdb_iox/blob/26a7a61d0a38785ad5c413fefec6950640b3da70/query/src/frontend/reorg.rs#L105)

alamb:
https://github.com/influxdata/influxdb_iox/pull/4110 may be related from @crepererum

NGA-TRAN:
@carols10cents search for `compute_sort_key` that we use for OG. You may want to modify that function to make it work for both OG and NG

[...]

PR #4110, opened 2022-03-23
feat: store sort key in NG metadata

crepererum:
The sort key is optional and currently only produced by `iox_tests`.
Writing it within the ingester/compactor is tracked by #3968. The sort
key is read by the querier (and this will be verified by the query tests
and is required to merge #4103).


iox_tests/src/util.rs
```diff
@@ -435,3 +443,60 @@ impl TestTombstone {
 pub fn now() -> Time {
     Time::from_timestamp(0, 0)
 }
+
+/// Sort mutable batch into arrow record batch and sort key.
+fn sort_mutable_batch(batch: MutableBatch) -> (RecordBatch, SortKey) {
+    // build dummy sort key
```
alamb, line 449: I know this is just for testing, but I think you might be able to make this a lot simpler by using a datafusion plan.

Something like (untested)

```rust
    let record_batch = batch.to_arrow(Selection::All).unwrap();
    let mem_table = MemTable::new(record_batch.schema(), vec![vec![record_batch]});
    let mut ctx = SessionContext::new();
    let sorted_results = ctx.register_table(mem_table)
      .sort(exprs)
      .collect()
```

I am happy to help with a follow on PR if needed

(making a API in datafusion to scan a set of batches would be great too 🤔 )

querier/src/chunk/mod.rs
```diff
@@ -310,10 +309,11 @@ mod tests {
             catalog.time_provider(),
         );
 
+        // set up catalog
         let lp = vec![
-            "table,tag1=WA field_int=1000 8000",
-            "table,tag1=VT field_int=10 10000",
-            "table,tag1=UT field_int=70 20000",
+            "table,tag1=WA field_int=1000i 8000",
```
alamb, line 314: why this change?

alamb (approved):

querier/src/chunk/mod.rs
crepererum: Because the test helpers (`iox_tests`) now sort the data before storing it, so the "sort key" isn't a lie (that's the code path you commented on above). And we don't run a full datafusion query here, it's just a simple "scan" operation, so "orderIn = orderOut".

iox_tests/src/util.rs
crepererum: I leave that to you then. TBH just the simple sort operation for testing was way more complicated than I wanted it to be, but as expressed below I didn't want the sort key to be a lie, so on the long run doing it proper is worth it.

NGA-TRAN: But eventually, we need that `compute_sort_key` to be smart because we want all files sort on the same sort key. I think that is a bigger task listed in #3968

PR #4111, opened 2022-03-23
refactor: make query tests less OG-specific

crepererum:
Small change.

### DIFF
--- a/query_tests/cases/in/sql_system_tables.expected
+++ b/query_tests/cases/in/sql_system_tables.expected
@@ -1,10 +1,10 @@
--- Test Setup: TwoMeasurementsManyFieldsOneChunk
+-- Test Setup: OldTwoMeasurementsManyFieldsOneRubChunk
 -- SQL: SELECT partition_key, table_name, storage, memory_bytes, row_count from system.chunks;
 -- Results After Sorting
 +---------------+------------+-------------------+--------------+-----------+
 | partition_key | table_name | storage           | memory_bytes | row_count |
 +---------------+------------+-------------------+--------------+-----------+
-| 1970-01-01T00 | h2o        | OpenMutableBuffer | 1831         | 3         |
+| 1970-01-01T00 | h2o        | ReadBuffer        | 4536         | 3         |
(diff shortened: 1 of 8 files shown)


query_tests/src/scenarios/library.rs
```diff
@@ -642,19 +642,17 @@ impl DbSetup for TwoMeasurementsManyFieldsOneChunk {
             "o2,state=CA temp=79.0 300",
         ];
 
-        write_lp(&db, &lp_lines.join("\n"));
-        vec![DbScenario {
-            scenario_name: "Data in open chunk of mutable buffer".into(),
-            db,
-        }]
+        all_scenarios_for_one_chunk(vec![], vec![], lp_lines, "h2o", partition_key).await
```
crepererum, line 645: This is the core change: many tests use `TwoMeasurementsManyFieldsOneChunk` but neither care about OG nor about a specific chunk stage (in fact they work w/ all chunk stages and even NG).

query_tests/cases/in/sql_system_tables.expected
```diff
@@ -1,10 +1,10 @@
--- Test Setup: TwoMeasurementsManyFieldsOneChunk
+-- Test Setup: OldTwoMeasurementsManyFieldsOneRubChunk
```
NGA-TRAN, line 1: Did the original  test MUB? Should we keep it?

query_tests/cases/in/sql_system_tables.sql
```diff
@@ -1,4 +1,4 @@
--- IOX_SETUP: TwoMeasurementsManyFieldsOneChunk
+-- IOX_SETUP: OldTwoMeasurementsManyFieldsOneRubChunk
```
NGA-TRAN, line 1: Same here

NGA-TRAN (approved):
Looks good except 2 tests I think were for MUB but you changed to RUB

query_tests/cases/in/sql_system_tables.expected
crepererum: Both tests don't really care. They test the system tables. Actually I find the new setup even better because it creates to chunks in different states.

Issue #4102, opened 2022-03-22
NG Querier: End to End testing


alamb:
This is a tracking ticket to get basic end to end tests setup for NG

# Purpose 
The purpose of the end to end tests is to be the highest level "integration" test that can run just within the influxdb_iox repository and makes sure all the various plumbing is connected correctly. It is NOT meant to cover corner cases in implementation, which are better tested in the tests of the various modules

# Structure
Each of these tests starts up IOx as a sub process (aka runs the `influxdb_iox` binary) and manipulates it either via client or APIs. These tests should not manipulate or use contents of any influxdb_iox crates. 

# Current state of affairs:
* `end_to_end`: https://github.com/influxdata/influxdb_iox/blob/alamb%2Ffix_return/influxdb_iox/tests/end_to_end.rs#L1
* `end_to_end_ng`: https://github.com/influxdata/influxdb_iox/blob/alamb%2Ffix_return/influxdb_iox/tests/end_to_end_ng.rs#L1

The [current end to end NG](https://github.com/influxdata/influxdb_iox/blob/alamb%2Ffix_return/influxdb_iox/tests/end_to_end_ng_cases/ng.rs#L15) test verifies that the ingester can return data written by the router. Kudos to @carols10cents  for getting the foundation setup.

# High level plan
I reviewed the scenarios in end_to_end that we need to port, and here is the initial list I came up with

### Basic plumbing tests
- [ ] End to end smoke  test for all-in-one mode (write data, query it back), with CLI and client
- [ ] End to end smoke test for querier: write data to a router, ensure the data gets to parquet, read it back from the querier
- [ ] End to end coverage for grpc write endpoint: https://github.com/influxdata/influxdb_iox/issues/3941
- [ ] End to end  gRPC coverage (https://github.com/influxdata/influxdb_iox/issues/3958) -- port the existing OG tests https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_cases/influxrpc.rs#L1
- [ ] End to end flight API (https://github.com/influxdata/influxdb_iox/issues/3958) -- port the existing OG tests https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_cases/flight_api.rs#L1
[...]

PR #4114, opened 2022-03-23
test: add end to end test for all in one mode, restructure fixture

alamb:
~Draft as it builds off https://github.com/influxdata/influxdb_iox/pull/4115~

re: https://github.com/influxdata/influxdb_iox/issues/4102

I am sorry it is so large... Revamping the harness is quite intrusive and I don't have any great ideas on how to make it smaller. 

# Rationale
We need "end to end tests" for NG

# Changes
1. Move common fixture setup for ng into its own crate (avoid recompiling it each time a test is changed, and encourage better code isolation)
2. Rework common fixtures to support all in one mode (including port assignment, etc)
3. Use common fixture to write an end to end test for `all-in-one` mode 
4. Add docs on how to run the tests


test_helpers_end_to_end_ng/src/client.rs
```diff
@@ -0,0 +1,90 @@
...
+        .body(Body::from(line_protocol.into()))
+        .expect("failed to construct HTTP request");
+
+    client
+        .request(request)
+        .await
+        .expect("http error sending write")
+}
+
+/// Runs a query using the flight API on the specified connection
```
alamb, line 37: this is somewhat fascinating -- we probably need a better way to programatically wait for data to become readable, but this works initially

test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -0,0 +1,579 @@
...
+use tempfile::{NamedTempFile, TempDir};
+use tokio::sync::Mutex;
+
+use crate::server_type::AddAddrEnv;
+
+use super::{addrs::BindAddresses, ServerType, TestConfig};
+
+/// Represents a server that has been started and is available for
+/// testing.
+pub struct ServerFixture {
```
alamb, line 21: I am not really sure why this wrapper exists, but it existed and I chose not to mess with it in this PR

test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -0,0 +1,579 @@
...
+
+    /// connection to router gRPC, if available
+    router_grpc_connection: Option<Connection>,
+
+    /// connection to ingester gRPC, if available
+    ingester_grpc_connection: Option<Connection>,
+
+    /// connection to querier gRPC, if available
+    querier_grpc_connection: Option<Connection>,
+}
```
alamb, line 106: will add compactor when there is something to test

test_helpers_end_to_end_ng/src/server_type.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub trait AddAddrEnv {
+    /// add the relevant bind addreses for the server type
+    fn add_addr_env(&mut self, server_type: ServerType, addrs: &BindAddresses) -> &mut Self;
+}
+
+impl AddAddrEnv for Command {
+    fn add_addr_env(&mut self, server_type: ServerType, addrs: &BindAddresses) -> &mut Self {
+        match server_type {
```
alamb, line 30: 🤮

influxdb_iox/tests/end_to_end_ng_cases/all_in_one.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    // Set up all_in_one ====================================
+
+    let test_config = TestConfig::new(ServerType::AllInOne)
+        .with_postgres_catalog(&database_url)
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_TYPE", "file")
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS", n_sequencers)
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_ADDR", &write_buffer_string)
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START", "0")
+        .with_env("INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END", "0")
```
crepererum, line 29: I think I've said this in the original PR: the end2end test should not be expected to set any env variables at all (at least not when it's not some weird special snowflake). This should be done by the test fixture/framework and if there is anything that is tunable the framework should expose a typed API for that (similar to the `.with_postgres_catalog` method).

Since this PR is already quite massive and this problem existed before, I think that's more something for a follow-up PR.

test_helpers_end_to_end_ng/src/client.rs
crepererum: This is a general issue we also have in the test bench. I think the only sane fix would be if the router would return some watermark -- e.g. the sequence number numberS (plural!) that it just produced -- and the querier would offer an API to ask or check for that watermark.

test_helpers_end_to_end_ng/src/server_type.rs
crepererum: TBH: could be worse. IMHO to hide that ugliness is exactly the job of the test framework.

test_helpers_end_to_end_ng/src/server_type.rs
```diff
@@ -0,0 +1,72 @@
...
+    AllInOne,
+    Ingester,
+    Router2,
+}
+
+impl ServerType {
+    /// returns the name of the 'run' subcommand of the `influxdb_iox` binary
+    pub fn run_command(&self) -> &'static str {
+        match self {
+            Self::AllInOne => "all-in-one",
```
crepererum, line 16: Just to understand the direction here: if we wanna test the querier later do we use the all-in-one mode or do we actually use multiple servers (similar to the smoke test you have ported over)?

I'm wondering how much the all-in-one mode counts as an integration test beyond testing the all-in-one mode itself. I know that technically the mode is implemented in a way that we try not to share data structures, but there is still the risk that having the same `Arc<dyn Catalog>`, `Arc<ObjectStore>`, `Arc<dyn WriteBuffer>` somewhat leaks state. I mean if we use the all-in-one-mode, then we also don't need a file-based object store and write buffer as well as a PostgreSQL. You could just use the in-mem variants for all three.

So what I wanna say: I think the all-in-one-mode should only be used to test the all-in-one-mode, not to test the other modes or their interaction.

test_helpers_end_to_end_ng/src/server_fixture.rs
crepererum: 1. What's the alternative? Looking at the methods in `impl ServerFixture` you would need to implement some extension trait for `Arc<TestServer>` which is probably more work.
2. I think the expectation on an API-level is that you do NOT clone `ServerFixture` because why would you request a "single use" server and then clone it? The fact that this is internally an `Arc<_>` is a mere implementation detail to provide a single interface for single-use and shared.
3. Following (2) I think that `ServerFixture::server` should NOT be public. We may instead implement `Deref` or `fn server(&self) -> &TestServer` though.

crepererum (approved):
Solid framework, I like it. Some minor API concerns (see `ServerFixture`) and a roadmap question (see all-in-one mode).

test_helpers_end_to_end_ng/src/server_type.rs
alamb: > I think the all-in-one-mode should only be used to test the all-in-one-mode, not to test the other modes or their interaction.

I 100% agree. I plan to write most tests using individually configured ingester, querier, etc

influxdb_iox/tests/end_to_end_ng_cases/all_in_one.rs
alamb: Makes sense and I will do it

test_helpers_end_to_end_ng/src/server_fixture.rs
alamb: The alternative I was thinking was to straight up use `Arc<TestServer>` 

At least one thing that the old tests do is to put a client on `Arc<TestServer>` that sends extra headers. 🤔  maybe that type of thing.

I'll add a `fn server(&self) -> &TestServer` 👍  to make this less public

alamb: https://github.com/influxdata/influxdb_iox/pull/4122

lukebond (approved):

docs/testing.md
```diff
@@ -2,6 +2,43 @@
...
+better tested with targeted tests in the various sub modules that make
+up IOx.
+
+Each of these tests starts up IOx as a sub process (aka runs the
+`influxdb_iox` binary) and manipulates it either via a client
+CLI. These tests should *not* manipulate or use the contents of any
+subsystem crate.
+
+### Prerequisites
+The end to end tests currently require a connection to postgres specified by a `DSN`, such as `postgresql://localhost:5432/alamb`.
```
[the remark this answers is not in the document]
alamb: This harness sets up the schema (by running `influxdb_iox catalog setup`); I will clarify in the docs.

test_helpers_end_to_end_ng/src/client.rs
alamb: that is a good idea 🤔  there is https://docs.rs/reqwest-retry/latest/reqwest_retry/ which may do what we want.

One challenge is that the definition of "need to retry" can be non obvious (for example, in this case the request succeeding at the HTTP level,  returning a `RecordBatch` but then no rows are in the batch so it still needs to be retried).

I'll see how it goes as I flesh out the tests

docs/testing.md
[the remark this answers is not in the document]
alamb: in https://github.com/influxdata/influxdb_iox/pull/4114/commits/466de784419f438d01d5d2c4af94211a69e0d980

influxdb_iox/tests/end_to_end_ng_cases/all_in_one.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/4126

PR #4115, opened 2022-03-23
fix: fix all in one mode argument handling so it can start

alamb:
# Rationale
https://github.com/influxdata/influxdb_iox/pull/4107 broke all in one mode. If you try to run it now you get:

```shell
target/debug/influxdb_iox run -vv all-in-one --write-buffer=file --write-buffer-addr /tmp --catalog-dsn 'postgresql://localhost:5432/alamb' --write-buffer-partition-range-start 0 --write-buffer-partition-range-end 0 --pause-ingest-size-bytes 10000 --persist-memory-threshold-bytes 1000`
thread 'main' panicked at 'Command all-in-one: Argument names must be unique, but 'topic' is in use by more than one argument or group', /Users/alamb/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.1.6/src/build/debug_asserts.rs:76:9
```

However, we don't have test coverage so it is my own fault (and I am adding coverage in https://github.com/influxdata/influxdb_iox/pull/4114)

# Changes:
Fix all in one mode


influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -199,33 +196,50 @@ pub struct Config {
...
-            .run_config
+        let compactor_run_config = run_config
             .clone()
-            .with_grpc_bind_address(self.compactor_grpc_bind_address);
+            .with_grpc_bind_address(compactor_grpc_bind_address);
+
+        // create a CompactorConfig for the all in one server based on
+        // settings from other configs. Cant use `#clap(flatten)` as the
+        // parameters are redundant with ingesters
+        let compactor_config = CompactorConfig {
```
alamb, line 238: Here is where the compactor config gets created now

lukebond (approved):

PR #4116, opened 2022-03-23
feat: delete fully processed tombstones &  integration tests for  find_and_compact

### LINKED ISSUE #3953: NG Compactor: remove fully processed tombstones

NGA-TRAN:
Closes #3953 

This PR includes:
1. Implemented and tested `remove_fully_processed_tombstones`
2. Integration tests for `find_and_compact`


compactor/src/compact.rs
```diff
@@ -351,7 +360,7 @@ impl Compactor {
         }
 
         // Remove fully processed tombstones
-        // TODO: #3953 - remove_fully_processed_tombstones(tombstones)
+        self.remove_fully_processed_tombstones(tombstones).await?;
```
NGA-TRAN, line 363: This is the last piece of `find_and_compact` which triggers the (integration) tests for find_and_compact

compactor/src/compact.rs
```diff
@@ -627,11 +641,20 @@ impl Compactor {
...
             // Now that the parquet file is available, create its processed tombstones
-            for tombstone_id in catalog_update.tombstone_ids {
-                txn.processed_tombstones()
-                    .create(parquet.id, tombstone_id)
-                    .await
-                    .context(UpdateSnafu)?;
+            for (_, tombstone) in catalog_update.tombstones {
+                // Becasue data may get removed and split during compaction, a few new files
+                // may no longer overlap with the delete tombstones. Need to verify whether
+                // they are overlap before adding process tombstones
```
NGA-TRAN, line 640: This is the bug and its fix @carols10cents . In the original design, we did not see this at all 😄

compactor/src/compact.rs
```diff
@@ -685,6 +708,133 @@ impl Compactor {
         min_time + (max_time - min_time) * SPLIT_PERCENTAGE / 100
     }
 
+    // remove fully processed tombstones
+    async fn remove_fully_processed_tombstones(
```
NGA-TRAN, line 705: This is the main purpose of this PR

compactor/src/compact.rs
```diff
@@ -778,9 +928,333 @@ mod tests {
...
+    use querier::{
+        cache::CatalogCache,
+        chunk::{collect_read_filter, ParquetChunkAdapter},
+    };
     use query::test::{raw_data, TestChunk};
     use time::SystemProvider;
 
+    #[tokio::test]
+    // This is integration test to verify all pieces are put together correctly
+    async fn test_find_and_compact() {
```
NGA-TRAN, line 936: Since the last piece is connected, this integration test helps verify the last piece in a full context of compaction

iox_catalog/src/postgres.rs
```diff
@@ -1469,8 +1509,8 @@ WHERE parquet_file.sequencer_id = $1
   AND parquet_file.partition_id = $3
   AND parquet_file.compaction_level = 1
   AND parquet_file.to_delete IS NULL
-  AND ((parquet_file.min_time <= $5 AND parquet_file.max_time >= $4)
-      OR (parquet_file.min_time > $5 AND parquet_file.min_time <= $5));
+  AND ((parquet_file.min_time <= $4 AND parquet_file.max_time >= $4)
+      OR (parquet_file.min_time > $4 AND parquet_file.min_time <= $5));
```
NGA-TRAN, line 1513: Fix another bug here. The tests for this are in  `test_find_and_compact`

iox_catalog/src/postgres.rs
```diff
@@ -1284,6 +1305,25 @@ ORDER BY id;
...
+    async fn remove(&mut self, tombstone_ids: &[TombstoneId]) -> Result<usize> {
+        let ids: Vec<_> = tombstone_ids.iter().map(|t| t.get()).collect();
+
+        let sql = format!(
+            "DELETE FROM tombstone WHERE id in({}) RETURNING id;",
+            ids.iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>()
+                .join(",")
+        );
```
NGA-TRAN, line 1317: @carols10cents : I tried `WHERE id IN ($1)` and then `.bind(&ids[..])`  like you did with `= ANY($1)` but it did not work. They said array should work for Postgres and not sure what I missed. I use this as a workaround

querier/src/chunk/mod.rs
```diff
@@ -285,15 +292,30 @@ impl ParquetChunkAdapter {
     }
 }
 
+/// collect data for the given chunk
+pub async fn collect_read_filter(chunk: &QuerierChunk) -> Vec<RecordBatch> {
```
NGA-TRAN, line 296: @crepererum : I moved this out of the tests module to be able to use it in tests of the compactor crate

compactor/src/compact.rs
```diff
@@ -685,6 +701,136 @@ impl Compactor {
...
+                    tombstone_id
+                );
+                return false;
+            }
+        };
+
+        // Get number of non-deleted parquet files of the same tableId & sequencerId that overlap with the tombstone time range
+        let count_pf = repos
+            .parquet_files()
+            .count_by_overlaps(
```
pauldix, line 755: in order for the parquet file to overlap with the tombstone, it must have a `min_sequence_number` < the tombstone's sequence number. Otherwise, this tombstone was created before the parquet file in question so it shouldn't be applied.

iox_catalog/src/mem.rs
```diff
@@ -857,6 +863,18 @@ impl TombstoneRepo for MemTxn {
         Ok(tombstones)
     }
 
+    async fn remove(&mut self, tombstone_ids: &[TombstoneId]) -> Result<usize> {
```
pauldix, line 866: shouldn't this also remove the processed tombstone records associated with these tombstones?

pauldix: Oh, I see you have this as two separate operations, one against tombstones and one against processed tombstones. Seems like it would be better to have a single catalog function to call that does both in a single transaction.

pauldix (changes requested):
Found one logic bug in counting how many files a tombstone applies to. Also a question about removing tombstones and processed tombstones in a single transaction. Other than that it looks good!

compactor/src/compact.rs
```diff
@@ -685,6 +701,136 @@ impl Compactor {
...
+
+        Ok(())
+    }
+
+    // Return true if the given tombstones is fully processed
+    async fn fully_processed(&self, tombstone_id: TombstoneId) -> bool {
+        let mut repos = self.catalog.repositories().await;
+
+        // Get tablbeId, sequencerId, min_time, max-time of the tombstone
+        let tombstone = repos.tombstones().get_by_id(tombstone_id).await;
```
carols10cents, line 740: This is doing one query per tombstone, and this query in particular is getting the rest of the tombstone columns for a tombstone ID, but we had all the tombstone data in the GroupWithTombstones in `groups_with_tombstones`. Could we keep the whole `Tombstone` around, dedupe the collection by `TombstoneId` instead, and then avoid all these queries to get the data again?

compactor/src/utils.rs
```diff
@@ -63,6 +66,16 @@ impl ParquetFileWithTombstone {
...
 
+    /// Return all tombstones in hash map format
+    pub fn tombstones(&self) -> BTreeMap<TombstoneId, Tombstone> {
+        let mut ts_map = BTreeMap::new();
+        for ts in &self.tombstones {
+            ts_map.insert(ts.id, (*ts).clone());
+        }
+
+        ts_map
+    }
```
carols10cents, line 77: We could store the tombstones as a `BTreeMap` to begin with, if that's more useful! I just picked the simplest data structure based on what I knew at the time :)

If this stays, the code could be simplified as:

```suggestion
    pub fn tombstones(&self) -> BTreeMap<TombstoneId, Tombstone> {
        self.tombstones
            .iter()
            .map(|ts| (ts.id, ts.clone()))
            .collect()
    }
```

carols10cents (changes requested):
Suggestions!

iox_catalog/src/mem.rs
NGA-TRAN: The their code are in different places but they are in one transaction but let me put them on one place to make it easier to read and maintenance

compactor/src/compact.rs
NGA-TRAN: Done: I have added sequence number and compare that while reading the catalog

compactor/src/compact.rs
```diff
@@ -685,6 +696,81 @@ impl Compactor {
...
+        // Remove the tombstones
+        // Note (todo maybe): if the list of deleted tombstone_ids is long and
+        // make this transaction slow, we need to split them into many smaller lists
+        // and each will be removed in its own transaction.
+        let mut repos = self.catalog.repositories().await;
+        repos
+            .tombstones()
+            .remove(&to_be_removed)
+            .await
+            .context(RemoveTombstonesSnafu)?;
```
NGA-TRAN, line 721: @carols10cents äter a lot of thoughts, I decided we do not need to have any warning about which rows not deleted. The catalog will throw error if something wrong happens and we already capture it

NGA-TRAN (commented):
@carols10cents and @pauldix The pR is ready for second review

pauldix (approved):

iox_catalog/src/postgres.rs
```diff
@@ -1305,23 +1305,36 @@ ORDER BY id;
...
-        let deleted = sqlx::query(&sql)
-            .fetch_all(&mut self.inner)
-            .await
-            .map_err(|e| Error::SqlxError { source: e })?;
+        // Remove tombstones
+        sqlx::query(
+            r#"
+DELETE
+FROM tombstone
+WHERE id = ANY($1);
```
carols10cents, line 1329: Should these two DELETEs be done in a transaction?

NGA-TRAN: Good catch. I removed the transaction on purpose assuming sqlx/postgres will auto commit but I forgot there are 2 delete statements in here. Working not to add transaction back

compactor/src/compact.rs
NGA-TRAN: This comment is no longer valid here because new code has just been added

compactor/src/compact.rs
```diff
@@ -685,6 +696,87 @@ impl Compactor {
...
+            if self.fully_processed(ts).await {
+                to_be_removed.push(ts_id);
+            }
+        }
+
+        // Remove the tombstones
+        // Note (todo maybe): if the list of deleted tombstone_ids is long and
+        // make this transaction slow, we need to split them into many smaller lists
+        // and each will be removed in its own transaction.
+        let mut txn = self
```
NGA-TRAN, line 716: @carols10cents I have made the caller to start the transaction

iox_catalog/src/postgres.rs
NGA-TRAN: Have made the caller to start the transaction

carols10cents (approved):
Ok, I think this is good to go now. I have a tiny optimization that I'll submit in a new PR 👍🏻

PR #4117, opened 2022-03-23
fix: Change to_delete column on parquet_files to be a time

### LINKED ISSUE #4059: catalog: change to_delete on parquet_file to be a time
The `to_delete` field on `paruqet_file` should be changed to be a nullable time. Use an int like in other time values. If a file is not marked to be deleted, this value will be null. When it is marked to be deleted, it should have the timestamp of now. This will enable the compactor to later decide to actually delete the file from object storage and to delete the record from the catalog. We could do this after this time is more older than an hour (no query should take longer than an hour) or if we want more safety, we could increase this.

carols10cents:
Set to_delete to the time the file was marked as deleted rather than true.

Fixes #4059.


carols10cents:
oops lol. I was only running the catalog tests locally 🤦🏻‍♀️

NGA-TRAN (approved):
LGTM

PR #4126, opened 2022-03-24
refactor: use typed `TestConfig` rather than environment variable names for NG end to end tests

alamb:
re https://github.com/influxdata/influxdb_iox/issues/4102

Follow up on @crepereum's comment: https://github.com/influxdata/influxdb_iox/pull/4114#discussion_r834083863

# Rationale
Use a typed `config` object for configuring end to end tests rather than environment variables to minimize the maintenance burden.

# Changes
Set configuration values with the test fixture/framework rather than the tests

PR #4128, opened 2022-03-24
refactor: move object store configuration to TestConfig

alamb:
~Draft until https://github.com/influxdata/influxdb_iox/pull/4126 is merged~

re https://github.com/influxdata/influxdb_iox/issues/4102

# Rationale:
The services are supposed to share an object store, but the fixture at the moment doesn't allow that

# Changes:
Move configuration changes into `TestConfig`


test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -86,10 +81,6 @@ pub struct TestServer {
     /// Which ports this server should use
     addrs: BindAddresses,
 
-    /// The temporary directory **must** be last so that it is
-    /// dropped after the database closes.
-    dir: TempDir,
```
alamb, line 91: this PR moves the handling of `TempDir` into the `TestConfig` so that the different services actually share the same object store directory (which will be important of an actual end to end test involving the querier)

crepererum (approved):

domodwyer:
CircleCI seems to be sad about this PR and needs a kick - I've removed the automerge label to unblock the other PRs waiting in the merge queue.

PR #4130, opened 2022-03-24
fix: propagate shutdown requests & compactor join() fix

### LINKED ISSUE #4089: Investigate shutdown behaviour of all-in-one mode after PR reverts
I have created PRs to revert #4043 and #4044 which touched on propagating shutdown info. Perhaps those changes alone weren't enough to fix what @domodwyer was fixing, so I'm creating this issue to track it so he can pick it up when he gets back later this week.

domodwyer:
This PR restores #4044 & #4043 which were reverted by #4090 & #4091 and fixes an issue in the compactor that was incorrectly exiting immediately when `join()`-ed - I've left this as a TODO as the compactor worker loop does not yet exist.

Closes #4089.

---

* fix: propagate shutdown into QuerierHandlerImpl (011d3fbf2)

      Prior to this commit, calling shutdown() on the QuerierServer (the server
      layer run by the iox binary) would cancel it's own CancellationToken, while
      the QuerierHandlerImpl (the actual querier workload entrypoint) would be
      watching it's own, different token.

      This commit removes the redundant CancellationToken in the QuerierServer,
      instead using the inner QueryHandlerImpl for cancellation notification &
      completion.

* fix: propagate shutdown into CompactorHandler (a4aedcf2a)

      Prior to this commit, calling shutdown() on the CompactorServerType (the 
      server layer run by the iox binary) would cancel it's own CancellationToken,
      while the CompactorHandler (the actual compaction workload entrypoint) would
      be watching it's own, different token.

      This commit removes the redundant CancellationToken in the 
      CompactorServerType, instead using the inner CompactorHandler for cancellation
      notification & completion.

* fix: compactor early shutdown (bf782de42)

      The compactor stub code would wait on nothing when the caller waited on 
      join()-ing the compactor handler, and this meant any caller who blocked on
      join() would immediately return.

### DIFF
--- a/compactor/src/handler.rs
+++ b/compactor/src/handler.rs
@@ -5,8 +5,7 @@ use backoff::BackoffConfig;
 use data_types2::SequencerId;
 use futures::{
     future::{BoxFuture, Shared},
-    stream::FuturesUnordered,
-    FutureExt, StreamExt, TryFutureExt,
+    FutureExt, TryFutureExt,
 };
 use iox_catalog::interface::Catalog;
 use object_store::DynObjectStore;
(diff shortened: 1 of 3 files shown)


lukebond (approved):
thanks!

alamb:
Looks reasonable to me

alamb:
Thanks @domodwyer

alamb:
removing label as github actions seem not to have run here. Hoping by putting it at the end of the line it will merge and rerun

PR #4131, opened 2022-03-24
feat: Clean up old parquet files in the catalog and object storage

### LINKED ISSUE #3954: NG Compactor: Remove soft deleted parquet files
# Rationale
This ticket implements [Remove soft deleted Parquet Files](https://docs.google.com/document/d/1lfILhw5ckzTlXFOk7VXKQOiz3TvDDlx-tdCgQ0YiG8E/edit#heading=h.r1i52kutaqb8)

carols10cents:
Fixes #3954.

I'm happy to change names if "garbage collector" is too overloaded ;)

This does not actually call the function to clean up old files, nor does it set what "old" means.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -1499,6 +1499,7 @@ mod tests {
         ParquetFile {
             id: ParquetFileId::new(0),
             sequencer_id: SequencerId::new(0),
+            namespace_id: NamespaceId::new(0),
             table_id: TableId::new(0),
             partition_id: PartitionId::new(0),
             object_store_id: Uuid::new_v4(),
@@ -1691,6 +1692,7 @@ mod tests {
 
         let p1 = ParquetFileParams {
             sequencer_id: sequencer.id,
+            namespace_id: namespace.id,
             table_id: table.id,
             partition_id: partition.id,
             object_store_id: Uuid::new_v4(),
@@ -1993,6 +1995,7 @@ mod tests {
         // Prepare metadata in form of ParquetFileParams to get added with tombstone
         let parquet = ParquetFileParams {
             sequencer_id: sequencer.id,
+            namespace_id: namespace.id,
             table_id: table.id,
             partition_id: partition.id,
             object_store_id: Uuid::new_v4(),
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -868,7 +868,7 @@ mod tests {
     use arrow_util::assert_batches_sorted_eq;
     use data_types2::{KafkaPartition, NamespaceId, ParquetFileParams, SequenceNumber};
     use iox_tests::util::TestCatalog;
-    use object_store::{path::Path, ObjectStoreTestConvenience};
+    use object_store::ObjectStoreTestConvenience;
     use querier::{
         cache::CatalogCache,
         chunk::{collect_read_filter, ParquetChunkAdapter},
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -748,7 +748,7 @@ impl Compactor {
             Ok(count_pf) => count_pf,
             _ => {
                 warn!(
-                    "Error getting parquet file count for table ID {}, sequencer ID {}, min time {:?}, max time {:?}. 
+                    "Error getting parquet file count for table ID {}, sequencer ID {}, min time {:?}, max time {:?}.
                     Won't be able to verify whether its tombstone is fully processed",
                     tombstone.table_id, tombstone.sequencer_id, tombstone.min_time, tombstone.max_time
                 );
@@ -765,7 +765,7 @@ impl Compactor {
             Ok(count_pt) => count_pt,
             _ => {
                 warn!(
-                    "Error getting processed tombstone count for tombstone ID {}. 
+                    "Error getting processed tombstone count for tombstone ID {}.
                     Won't be able to verify whether the tombstone is fully processed",
                     tombstone.id
                 );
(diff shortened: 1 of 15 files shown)


compactor/src/garbage_collector.rs
```diff
@@ -0,0 +1,398 @@
...
+    use data_types2::{KafkaPartition, ParquetFile, ParquetFileParams, SequenceNumber};
+    use iox_object_store::ParquetFilePath;
+    use iox_tests::util::TestCatalog;
+    use object_store::ObjectStoreTestConvenience;
+    use std::time::Duration;
+    use uuid::Uuid;
+
+    /// Test helper to put an empty object store file at the expected location for a parquet file
+    /// tracked by the catalog, but without having to process data because the garbage collector
+    /// isn't concerned with the contents of the file, only its existence or lack thereof
```
NGA-TRAN, line 113: 👍

compactor/src/garbage_collector.rs
```diff
@@ -0,0 +1,398 @@
...
+            object_store,
+            time_provider,
+        }
+    }
+
+    /// Perform a pass of garbage collection, querying the catalog for all files marked to be
+    /// deleted earlier than the specified time. Remove the catalog entries, then remove the
+    /// associated object store files.
+    /// Meant to be invoked in a background loop.
+    pub async fn cleanup(&self, older_than: Timestamp) -> Result<()> {
```
NGA-TRAN, line 55: Since @pauldix has recently raised a concern that there may be a lot of files to clean up, I wonder we should do this clean up per (sequencer_id) or per (sequencer_id, namespace_id) or even smaller per (sequencer_id, namespaceId, table_id) or even a lot smaller per (sequencer_id, namespace_id, table_id, partition_id)

The namespace_id can be either gotten from your newly added field in this PR or we can read all namespace_ids for a  sequencer_id and its Kafka topic from the catalog

compactor/src/compact.rs
```diff
@@ -1085,6 +1084,7 @@ mod tests {
         ParquetFile {
             id: ParquetFileId::new(0),
             sequencer_id: SequencerId::new(0),
+            namespace_id: NamespaceId::new(0),
```
NGA-TRAN, line 1087: I think storing `namespace_id` in parquet file will be very convenient for us but I do not remember why we did not do it in the first place. Any particular reason for it @pauldix ?

I may miss something but I do not spot in the code the reason why this PR needs `namespace_id` in catalog parquet file.

NGA-TRAN (commented):
@carols10cents This PR does what it means to do. I need your help to point out why we need to add `namespace_id` into the parquet_file catalog table. 

We also need @pauldix's feedback on whether we want to delete all old soft deleted files in on step or we need to do that per sequencer or smaller grain.

compactor/src/compact.rs
NGA-TRAN: Sounds good. Thanks @carols10cents

NGA-TRAN (approved):
I approve this for Carol to move forward. If @pauldix wants to delete smaller subset of files in every call, we can change that easily, too.

And we need a task to schedule this, too. Not sure if @pauldix wants to do it as part of his scheduler or we need an extra todo ticket for this

iox_catalog/src/postgres.rs
```diff
@@ -1438,6 +1444,20 @@ WHERE table_id = $1 AND to_delete IS NULL;
         .map_err(|e| Error::SqlxError { source: e })
     }
 
+    async fn delete_old(&mut self, older_than: Timestamp) -> Result<Vec<ParquetFile>> {
+        sqlx::query_as::<_, ParquetFile>(
+            r#"
+DELETE FROM parquet_file
+WHERE to_delete < $1
```
domodwyer, line 1451: Might be worth adding an index on `to_delete` so it doesn't have to do a full table scan here 👍

domodwyer (approved):
Looks good to me - I think we'll need an index on the parquet table to support deletes given the unbounded growth of the table:

```
                                  QUERY PLAN
-------------------------------------------------------------------------------
 Delete on parquet_file  (cost=0.00..827571.22 rows=1 width=6)
   ->  Seq Scan on parquet_file  (cost=0.00..827571.22 rows=1 width=6)
         Filter: (to_delete < 1)
```

iox_catalog/src/postgres.rs
carols10cents: Ok, I've added the index, I don't have enough data locally to get postgres to actually *use* said index... could you maybe give it a quick look @domodwyer ? It doesn't cause any errors 🤷🏻‍♀️

domodwyer: Looks good to me! I'll let it roll into prod and then (hopefully remember to) check it is using the indexes in prod where there's ~5mil parquet files - that should convince the planner to use take all the help it can 😂

iox_catalog/migrations/20220324152729_add_namespace_id_to_parquet_file.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE
+  IF EXISTS parquet_file
+  ADD COLUMN namespace_id INT NOT NULL;
```
mkmik, line 3: hindsight: we should test our migrations against non-empty databases.
This statement can only be done on an empty table. If the table has any rows it fails with `column "namespace_id" of relation "parquet_file" contains null values`

(more context at https://github.com/influxdata/influxdb_iox/issues/4173)

mkmik: thanks; let's continue the discussion in https://github.com/influxdata/influxdb_iox/issues/4173

PR #4134, opened 2022-03-24
fix: maybe_auto_create_directories infinite loop

domodwyer:
This PR fixes an inifinite loop when re-using an existing dir for the file-backed write buffer. Before this PR, running this twice would get stuck in an infine loop the second time:

```console
$ cargo run -- run all-in-one \
--write-buffer=file \
--write-buffer-addr=./testwb \
--write-buffer-partition-range-start 0 \
--write-buffer-partition-range-end 0 \
--pause-ingest-size-bytes 10000000000 \
--persist-memory-threshold-bytes 1000000000
```

---

* fix: maybe_auto_create_directories infinite loop (0948a4af2)

      Fixes symlinking in maybe_auto_create_directories() - previously it would
      create a symlink specifying the target path relative to the working dir, and
      not relative to the symlink.

      If the working dir != the path in the WriteBufferCreationConfig, subsequent
      calls would get stuck in an infinite loop attempting to resolve the bad
      symlink.

### DIFF
--- a/write_buffer/src/file.rs
+++ b/write_buffer/src/file.rs
@@ -606,12 +606,19 @@ async fn maybe_auto_create_directories(
                 directories.insert(sequencer_id, sequencer_path_in_active);
             }
 
+            // A symlink target is resolved relative to the parent directory of
+            // the link itself.
+            let target = version
+                .strip_prefix(&root)
+                .expect("symlink target not in root workspace");
+
             // symlink active->version
-            if tokio::fs::symlink(&version, active).await.is_ok() {
+            if tokio::fs::symlink(target, active).await.is_ok() {
                 // linking worked
                 return Ok(directories);
             } else {
-                // linking did not work, assuming a concurrent initialization process. Remove version and and try again.
+                // linking did not work, assuming a concurrent initialization
+                // process. Remove version and and try again.
                 tokio::fs::remove_dir_all(&version).await?;
                 continue;
             }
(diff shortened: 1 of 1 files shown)


write_buffer/src/file.rs
```diff
@@ -606,12 +606,19 @@ async fn maybe_auto_create_directories(
                 directories.insert(sequencer_id, sequencer_path_in_active);
             }
 
+            // A symlink target is resolved relative to the parent directory of
+            // the link itself.
+            let target = version
+                .strip_prefix(&root)
+                .expect("symlink target not in root workspace root");
```
alamb, line 613: not sure what a `root workspace root` is -- maybe too many roots 🤔

write_buffer/src/file.rs
```diff
@@ -860,4 +867,25 @@ mod tests {
 
         assert_write_op_eq(&stream.next().await.unwrap().unwrap(), &w2);
     }
+
+    #[tokio::test]
+    async fn test_maybe_auto_create_dirs() {
+        let path = Path::new("./test-file-write-buffer");
+        let config = WriteBufferCreationConfig::default();
+
+        tokio::time::timeout(Duration::from_secs(5), async {
```
alamb, line 876: 👍

alamb (approved):

write_buffer/src/file.rs
domodwyer: 🥕

crepererum (approved):
Good catch!

domodwyer:
Rebased to kick circleci - no funcitonal changes.

Issue #4123, opened 2022-03-24
Querier: remove sync loop altogether


crepererum:
The querier now fetches chunks on-demand, but still syncs namespaces, tables and schemas in a tight sync loop. DataFusion needs sync access to the schema once we've picked the namespace, but while getting the namespace we can actually sync tables/schemas on-demand (probably shielded by a cache). This would result in a "does nothing until asked" querier.

PR #4135, opened 2022-03-25
refactor: clean up querier public interface

crepererum:
Small clean ups before I start w/ #4123 .

### DIFF
--- a/influxdb_iox/tests/end_to_end_ng_cases/ng.rs
+++ b/influxdb_iox/tests/end_to_end_ng_cases/ng.rs
@@ -34,7 +34,7 @@ async fn router2_through_ingester() {
     let ingester = ServerFixture::create(ingester_config).await;
 
     let mut querier_flight =
-        querier::flight::Client::new(ingester.server().ingester_grpc_connection());
+        querier::QuerierFlightClient::new(ingester.server().ingester_grpc_connection());
 
     let query = IngesterQueryRequest::new(
         namespace,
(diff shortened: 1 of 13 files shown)


influxdb_iox/tests/end_to_end_ng_cases/ng.rs
```diff
@@ -34,7 +34,7 @@ async fn router2_through_ingester() {
     let ingester = ServerFixture::create(ingester_config).await;
 
     let mut querier_flight =
-        querier::flight::Client::new(ingester.server().ingester_grpc_connection());
+        querier::QuerierFlightClient::new(ingester.server().ingester_grpc_connection());
```
alamb, line 37: 💯  I was thinking about this name in the morning. Much better.

I think we could maybe even call it `QuerierClient` as the flight part is maybe an implementation detail

querier/src/lib.rs
```diff
@@ -11,17 +10,21 @@
...
 #![allow(dead_code)]
 
-pub use client_util::connection;
-
-pub mod cache;
+mod cache;
 mod cache_system;
 mod chunk;
-pub mod database;
+mod database;
```
alamb, line 16: this is a nice change to be more deliberate about the interface to the querier

alamb (approved):
LGTM

alamb:
I am going to take the liberty of merging this PR so that it doesn't hang out and bitrot

PR #3807, opened 2022-02-21
feat: routing service protection

lukebond:
Part of #3778 

This lays the groundwork for service protection / limits for the routing tier by adding new columns to the catalog namespace schema and providing a way to update them via the catalog.

Next PR will impose those limits.

PR #4136, opened 2022-03-25
fix: service limit http code

### LINKED ISSUE #4066: Return 4xx HTTP error codes for column / table limit hits
Currently the router will return a 500 error when the catalog returns a colum/table limit error - it should be a 4xx.

domodwyer:
The service limits added in #3807 should return a "HTTP 429 Too Many Requests" error when they're reached.

Closes #4066.

---

* refactor: emit HTTP 429 for service limits (1edc4ca56)

      When a user hits the service limit, emit a "HTTP 429 Too Many Requests" error
      with no "Retry-After" header. Reference:

         
      https://docs.influxdata.com/influxdb/cloud/account-management/limits/#api-error-responses

* refactor: explicit schema validation errors (c6af2f593)

      This commit adds new service limit & schema conflict error variants to the
      SchemaError enum, allowing less error-prone matching higher up the stack and
      more explicit log messages for both.

      HTTP status codes are now correctly reported as 500 for internal catalog 
      errors during schema validation.

* refactor: lowercase error messages (8e85846db)

      Lowercases the error messages in the big iox_catalog Error enum for better
      composition of messages (no random capitalisation in glued-together strings,
      which is common with wrapped errors).

### DIFF
--- a/router2/src/dml_handlers/schema_validation.rs
+++ b/router2/src/dml_handlers/schema_validation.rs
@@ -4,7 +4,7 @@ use async_trait::async_trait;
 use data_types2::{DatabaseName, DeletePredicate};
 use hashbrown::HashMap;
 use iox_catalog::{
-    interface::{get_schema_by_name, Catalog},
+    interface::{get_schema_by_name, Catalog, Error as CatalogError},
     validate_or_insert_schema,
 };
 use mutable_batch::MutableBatch;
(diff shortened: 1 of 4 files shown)


router2/src/dml_handlers/schema_validation.rs
```diff
@@ -157,8 +167,33 @@ where
...
+                        existing_column_type=%existing,
+                        request_column_type=%new,
+                        "schema conflict"
+                    );
+                    SchemaError::Conflict(e)
+                }
+                // Service limits
+                CatalogError::ColumnCreateLimitError { .. }
+                | CatalogError::TableCreateLimitError { .. } => {
+                    warn!(%namespace, error=%e, "service protection limit reached");
```
domodwyer, line 189: Logging of schema conflict/service limit events are now more explicit 👍

mkmik (approved):
nice!

router2/src/dml_handlers/schema_validation.rs
```diff
@@ -323,6 +358,76 @@ mod tests {
         assert_cache(&handler, "bananas", "time", ColumnType::Time);
     }
 
+    #[tokio::test]
+    async fn test_write_table_service_limit() {
```
domodwyer, line 362: These unit tests cover hitting the service limits in the schema validator layer itself.

router2/tests/http.rs
```diff
@@ -231,3 +232,111 @@ async fn test_write_ok() {
     let hit_count = histogram.buckets.iter().fold(0, |acc, v| acc + v.count);
     assert_eq!(hit_count, 1);
 }
+
+#[tokio::test]
+async fn test_schema_conflict() {
```
domodwyer, line 237: These tests cover the `router2` stack as a whole, and ensure the schema validator error is mapped to the correct HTTP code.

PR #4138, opened 2022-03-25
feat: Compute a sort key in the ingester

### LINKED ISSUE #3968: ingester: Store consistent sort key for all parquet files of the same partition (and the same table, too?)

### LINKED ISSUE #4194: In the ingester, compute a sort key based on primary key cardinality
Part of #3968.

In the ingester, when persisting a `PersistingBatch`, compute a sort key for the batch based on the primary key's columns' cardinalities from low -> high, with the time column last.

Store this sort key in the parquet file's metadata.

carols10cents:
Fixes #4194.

I have now SPLIT #3968 into 3 parts, so this PR now only contains the *computation* of the sort key. It does *not* cover storing the sort key in the catalog (now #4195) or checking and potentially updating the sort key (now #4196).

### DIFF
--- a/ingester/src/compact.rs
+++ b/ingester/src/compact.rs
@@ -1,6 +1,9 @@
 //! This module is responsible for compacting Ingester's data
 
-use crate::data::{PersistingBatch, QueryableBatch};
+use crate::{
+    data::{PersistingBatch, QueryableBatch},
+    sort_key::compute_sort_key,
+};
 use arrow::record_batch::RecordBatch;
 use data_types2::NamespaceId;
 use datafusion::{error::DataFusionError, physical_plan::SendableRecordBatchStream};
@@ -10,8 +13,9 @@ use query::{
     exec::{Executor, ExecutorType},
     frontend::reorg::ReorgPlanner,
     util::compute_timenanosecond_min_max,
-    QueryChunkMeta,
+    QueryChunk, QueryChunkMeta,
 };
+use schema::sort::SortKey;
 use snafu::{ResultExt, Snafu};
 use std::sync::Arc;
 use time::{Time, TimeProvider};
@@ -67,8 +71,11 @@ pub async fn compact_persisting_batch(
         return Ok(None);
     }
 
+    // Get sort key based on cardinality
+    let sort_key = compute_sort_key(&batch.data);
+
     // Compact
-    let stream = compact(executor, Arc::clone(&batch.data)).await?;
+    let stream = compact(executor, Arc::clone(&batch.data), sort_key.clone()).await?;
     // Collect compacted data into record batches for computing statistics
     let output_batches = datafusion::physical_plan::common::collect(stream)
         .await
@@ -106,7 +113,7 @@ pub async fn compact_persisting_batch(
         max_sequence_number: max_seq,
         row_count,
         compaction_level: INITIAL_COMPACTION_LEVEL,
-        sort_key: None,
+        sort_key: Some(sort_key),
     };
 
     Ok(Some((output_batches, meta)))
@@ -116,11 +123,12 @@ pub async fn compact_persisting_batch(
 pub async fn compact(
     executor: &Executor,
     data: Arc<QueryableBatch>,
+    sort_key: SortKey,
 ) -> Result<SendableRecordBatchStream> {
     // Build logical plan for compaction
     let ctx = executor.new_context(ExecutorType::Reorg);
     let logical_plan = ReorgPlanner::new()
-        .scan_single_chunk_plan(data.schema(), data)
+        .compact_plan(data.schema(), [data as Arc<dyn QueryChunk>], sort_key)
         .context(LogicalPlanSnafu {})?;
 
     // Build physical plan
@@ -322,9 +330,11 @@ mod tests {
         let expected_pk = vec!["tag1", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -360,9 +370,11 @@ mod tests {
         let expected_pk = vec!["tag1", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -398,9 +410,11 @@ mod tests {
         let expected_pk = vec!["tag1", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -441,9 +455,11 @@ mod tests {
         let expected_pk = vec!["tag1", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -481,9 +497,11 @@ mod tests {
         let expected_pk = vec!["tag1", "tag2", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -535,9 +553,11 @@ mod tests {
         let expected_pk = vec!["tag1", "tag2", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -597,9 +617,11 @@ mod tests {
         let expected_pk = vec!["tag1", "tag2", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
@@ -652,9 +674,11 @@ mod tests {
         let expected_pk = vec!["tag1", "tag2", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
+
         // compact
         let exc = Executor::new(1);
-        let stream = compact(&exc, compact_batch).await.unwrap();
+        let stream = compact(&exc, compact_batch, sort_key).await.unwrap();
         let output_batches = datafusion::physical_plan::common::collect(stream)
             .await
             .unwrap();
--- a/ingester/src/compact.rs
+++ b/ingester/src/compact.rs
@@ -301,7 +301,7 @@ mod tests {
             seq_num_end,
             3,
             INITIAL_COMPACTION_LEVEL,
-            None, // todo: this  will have value when #3968 is done
+            None, // todo: this will have value when #3968 is done
         );
         assert_eq!(expected_meta, meta);
     }
--- a/ingester/src/compact.rs
+++ b/ingester/src/compact.rs
@@ -309,7 +309,7 @@ mod tests {
             seq_num_end,
             3,
             INITIAL_COMPACTION_LEVEL,
-            None, // todo: this will have value when #3968 is done
+            Some(SortKey::from_columns(["tag1", "time"])),
         );
         assert_eq!(expected_meta, meta);
     }
--- a/ingester/src/compact.rs
+++ b/ingester/src/compact.rs
@@ -331,6 +331,7 @@ mod tests {
         assert_eq!(expected_pk, pk);
 
         let sort_key = compute_sort_key(&compact_batch);
+        assert_eq!(sort_key, SortKey::from_columns(["tag1", "time"]));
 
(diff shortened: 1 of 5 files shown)


ingester/src/compact.rs
```diff
@@ -66,8 +74,13 @@ pub async fn compact_persisting_batch(
         return Ok(None);
     }
 
+    // Get the sort order
+    let schema = batch.data.schema();
+    let sort_key =
+        compute_sort_key_for_chunks(&schema, &[Arc::clone(&batch.data) as Arc<dyn QueryChunk>]);
```
NGA-TRAN, line 80: This is the right function to use here. However, we need to build statistics so it return the sort key based on the column cardinality instead

NGA-TRAN (commented):
>  I'm using [compute_sort_key_for_chunks](https://github.com/influxdata/influxdb_iox/blob/cabaf7c0e277927d290362c24cc7174c794df0d6/query/src/lib.rs#L271), which does work with NG because QueryableBatch implements QueryChunk and QueryChunkMeta, buuuut [its summary is always None](https://github.com/influxdata/influxdb_iox/blob/cabaf7c0e277927d290362c24cc7174c794df0d6/ingester/src/query.rs#L106-L108) so the SortKey is the schema primary key sorted lexicographically with time last.... is actually implementing summary for QueryableBatch part of this issue? Or is what I have in this PR so far correct? Just wanted to check before going down that path.

[compute_sort_key_for_chunks](https://github.com/influxdata/influxdb_iox/blob/cabaf7c0e277927d290362c24cc7174c794df0d6/query/src/lib.rs#L271) is the right one to use but we need to build statistics so it returns the sort key based on column cardinality instead. How and where to build statistic (summary or something like that) is up to you to design.

If this is the first batch of the partition (or the first batch of the table. We want to discuss this with the team), we will use the computed sort key. But if it is not the first batch, we have to reuse the sort key of the first batch. This becomes complicated in your question below

> The "TBD" part of the issue... it sounds like the potentially different sort orders based on presence or absence of columns isn't something the ingester should try to solve? That maybe the compactor can resort if it compacts two parquet files that have different orders? Or that the querier can resort if it gets data from multiple parquet files that are different? Otherwise, won't reading existing parquet files to find the sort order slow down the ingester? I'm not at all sure :)

We want sort key of the all files are the same. However, not all files have the same columns.    This is the design part you want to propose. A few examples:

- File 1's computed sort key:  `tag3, tag1, tag2, time`. There are other field columns that we do not care.
- So what is the sort key of file 2 if it has these tag columns: tag1, tag3,  tag4, time? I think it should be `tag3, tag1, tag4, time`. This means new tag column (tag4) will be put last and right before time and missing tag column (tag2) is ignored.
- If there are >= 2 new tag column, even though they are still last but which column before the other will be based on their cardinality, too.

Update: consistent sort key is important for  performance of ingester, compactor and querier and we want to have a proposal that the team agrees :)

@pauldix @alamb @e-dard thoughts?

carols10cents:
> If this is the first batch of the partition (or the first batch of the table. We want to discuss this with the team), we will use the computed sort key. But if it is not the first batch, we have to reuse the sort key of the first batch. This becomes complicated in your question below

I'm also confused because there are many things called "batch" :) Each parquet file only gets one PersistingBatch, so the "first batch" and "not first batch" you're talking about here... can't be a parquet file's batch because a parquet file only ever has one batch? So do you mean across the multiple parquet files, so that the sort key has to be decided on and stored higher up in the ingester, say in `TableData`, then passed down into `compact_persisting_batch` instead of creating it there as I am in the current state of this PR?

NGA-TRAN:
> I'm still confused. What if we're starting up a new ingester process for a partition that has already persisted some parquet files to object storage from an earlier ingester process that's now shut down? Are you saying that on ingester startup, it should read existing parquet files for the partition and tables to know what the sort order should be for incoming data?

This is a good point and I think we will run into this situation a lot even if the integer does not go down.  Do we store the sort key in the catalog's `parquet_file` table?  If so, we only need to read the catalog for it. This is a part of TBD and  why this needs to put a lot of thinking into and have a design out for the team to discuss :)

carols10cents:
new columns go at the end of the sort order (but before time)

carols10cents:
I'm feeling lost about how best to create the SortKey (for the initial persisting of a partition in the ingester)... here are my assumptions and the things I've found/considered:

Assumptions:

- To create a sort key based on cardinality, I need a TableSummary in the ingester for QueryableBatch. QueryableBatch has a Vec<Arc<SnapshotBatch>> where SnapshotBatch is a wrapper around an arrow RecordBatch (so essentially I want to make a TableSummary from a Vec<RecordBatch>). 

- When the QueryableBatch is built up from `BufferBatch`es (which is a wrapper around a `MutableBatch`), I can [use `MutableBatch`.columns() to build `ColumnSummary`s and `TableSummary`s](https://github.com/influxdata/influxdb_iox/pull/4138/files#diff-3309f5f122691950693127ab1489844b15eedeb286e74a3aec94dda167d7ef35R1174-R1195) and then [combine all those into one `TableSummary`](https://github.com/influxdata/influxdb_iox/pull/4138/files#diff-9186914d18a86fc9a44ff2288e98ad88f4fd6372b57839e93c046078bfd6d93fR65-R72). This seems to work, I think.

- When [a `SnapshotBatch` is created by `buffer_tombstone`](https://github.com/influxdata/influxdb_iox/blob/80b7e9cce16b465f882505738f1919980d8a148f/ingester/src/data.rs#L930-L934), there aren't `BufferBatch`es involved: the `SnapshotBatch` is created by [concatenating `RecordBatch`es from a `query` that might be filtering out data](https://github.com/influxdata/influxdb_iox/blob/80b7e9cce16b465f882505738f1919980d8a148f/ingester/src/data.rs#L899-L929), so I need a way to create a `TableSummary` from an arrow `RecordBatch` (or a list of `RecordBatch`es). 

- There is no existing code in iox to create a `TableSummary` from an arrow `RecordBatch`.

Are all these assumptions correct? 

What I've found/considered:

- I see there's [code to convert `TableSummary` to `datafusion::physical_plan::Statistics`](https://github.com/influxdata/influxdb_iox/blob/80b7e9cce16b465f882505738f1919980d8a148f/query/src/statistics.rs#L52-L53) but not vice versa, but doing the inverse looks to be trivial. So if I could get `datafusion::physical_plan::Statistics` for an arrow `RecordBatch`, that would work.
- I don't see any way to get a `datafusion::physical_plan::Statistics` from a `RecordBatch`. I see [`compute_record_batch_statistics`, but it explicitly only sets some of the statistics](https://github.com/apache/arrow-datafusion/blob/22fdca3e594d6b49bd867584a27e6258c52be831/datafusion/src/physical_plan/common.rs#L201-L204), and crucially doesn't set `distinct_count`, which is what we actually need.
- I don't see any other statistic-gathering code for a `RecordBatch` in arrow except for in parquet writer, and that looks like [it expects you to provide the statistics from somewhere else](https://github.com/apache/arrow-rs/blob/00accc7ad79f4c2fbdc3f471e20c04f477c64807/parquet/src/column/writer.rs#L384-L387)?

So it looks like I need to write something that iterates over a RecordBatch and gathers these statistics? Ideally, should this code live in iox, arrow, datafusion?

alamb:
>There is no existing code in iox to create a TableSummary from an arrow RecordBatch.

I think this is accurate. 

> So it looks like I need to write something that iterates over a RecordBatch and gathers these statistics? Ideally, should this code live in iox, arrow, datafusion?

I would suggest putting it in IOx initially. You can probably use the min/max kernels in https://docs.rs/arrow/11.0.0/arrow/compute/kernels/aggregate/index.html to do the calculations. 

The way OG works is that the statistics are incrementally kept up to date as data is ingested. I believe @pauldix  had some reason he didn't want to incrementally update statistics during ingest.

NGA-TRAN:
@carols10cents All your assumptions are correct. A few more things for you to know and consider:

1. Because data may be removed (deduplicated and tombstones applied) during compaction, statistics after compaction will be highly different from the statistics before compaction. I think the `sort_key` that is stored in the `partition` catalog table should be computed after compacting. This is the place I computed [row count and min and max for column time](https://github.com/influxdata/influxdb_iox/blob/80b7e9cce16b465f882505738f1919980d8a148f/ingester/src/compact.rs#L82) for a given set of record batches. I think your function that computes statistics and `sort_key` should be invoked there, too.

2. The funny thing is the compaction itself does need `sort_key` to deduplicate data. And you know very well, due to the lack of statistics right now, the sort key is the schema primary key sorted lexicographically with time last. I do not think computing statistics before compacting in the Ingester is extremely critical but if you think you can do that easily, it is worth considering build statistics before compaction, too.

3. I think the optimization way to build statistics for a RecordBatch is from its MutableBatches (as in your second assumption). If the RecordBatch is not built from MutableBatches (as in your third assumption), then we need to build it using arrow's aggregate functions pointed out by @alamb

alamb:
> The funny thing is the compaction itself does need sort_key to deduplicate data. 

I do think it is important to avoid re-sorting the data during compaction if possible, which I think will require the sort key

NGA-TRAN:
@carols10cents and @alamb 
After a good night sleep, I remember the `sort_key` must be computed **before**  compaction so the data is sorted accordingly during compaction to deduplicate data. I agree we should not resort data. So I think sort_key should be computed only **once** before compacting. My point #1 on previous comment should be ignored

carols10cents:
@NGA-TRAN Thank you for writing this up! This is very helpful 👍🏻

NGA-TRAN:
@carols10cents I forgot one step. When the `sort_key` has a new tag column added to the end, it needs to get updated in the `partition` catalog table, too. Example`:

1. First data set: `sort_key` =  `tag2, tag1, tag3, time`. `tag2, tag1, tag3, time` will be added to the `partition` catalog table.
2. Second data set has no  column `tag1`, the sort key of this dat set `sort_key` will be `tag2, tag3, time`. **Nothing** will be updated to the `partition` table because the sort key of both data set 1 and data set 2 is the sort key of the data set 1.
3. Third data set has no column tag3 but has a new column `tag4`, its `sort_key` will be `tag2, tag1, tag4, time`. Since there is an extra column,  we need to update the `partition` table with sort key `tag2, tag1, tag3, tag4, time`.

Note that the sort key of the partition now covers all tags of all 3 sets.

ingester/src/compact.rs
```diff
@@ -301,7 +309,7 @@ mod tests {
             seq_num_end,
             3,
             INITIAL_COMPACTION_LEVEL,
-            None, // todo: this  will have value when #3968 is done
+            Some(SortKey::from_columns(["tag1", "time"])),
```
NGA-TRAN, line 312: Nice ❤️

ingester/src/sort_key.rs
```diff
@@ -0,0 +1,233 @@
...
+    collections::{HashMap, HashSet},
+    num::NonZeroU64,
+    sync::Arc,
+};
+
+/// Given a `QueryableBatch`, compute a sort key based on:
+///
+/// - The columns that make up the primary key of the schema of this batch
+/// - Order those columns from low cardinality to high cardinality based on the data
+/// - Always have the time column last
```
NGA-TRAN, line 25: 👍

ingester/src/sort_key.rs
```diff
@@ -0,0 +1,233 @@
...
+    let schema = queryable_batch.schema();
+    let primary_key = schema.primary_key();
+
+    let cardinalities = distinct_counts(&queryable_batch.data, &primary_key);
+
+    trace!(cardinalities=?cardinalities, "cardinalities of of columns to compute sort key");
+
+    let mut cardinalities: Vec<_> = cardinalities.into_iter().collect();
+    // Sort by (cardinality, column_name) to have deterministic order if same cardinality
+    cardinalities.sort_by_cached_key(|x| (x.1, x.0.clone()));
```
NGA-TRAN, line 36: 👍

ingester/src/sort_key.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        // The hashmap should contain the distinct values for "host" and "env" only
+        assert_eq!(distinct.len(), 2);
+
+        // Return unique values
+        assert_eq!(
+            *distinct.get("host").unwrap(),
+            HashSet::from(["a".into(), "b".into(), "c".into()]),
+        );
+        // TODO: do nulls count as a value?
```
NGA-TRAN, line 182: I think NULL can be ignored while computing cardinality to decide sort order. What do you think @pauldix?

ingester/src/sort_key.rs
```diff
@@ -0,0 +1,233 @@
...
+        "#;
+        let lp3 = r#"
+            cpu,host=a,env=dev val=23 5
+            cpu,host=b,env=dev val=2 6
+        "#;
+        let qb = lp_to_queryable_batch(&[lp1, lp2, lp3]);
+
+        let sort_key = compute_sort_key(&qb);
+
+        assert_eq!(sort_key, SortKey::from_columns(["host", "env", "time"]));
```
NGA-TRAN, line 231: 👍

ingester/src/compact.rs
```diff
@@ -322,9 +330,11 @@ mod tests {
         let expected_pk = vec!["tag1", "time"];
         assert_eq!(expected_pk, pk);
 
+        let sort_key = compute_sort_key(&compact_batch);
```
NGA-TRAN, line 333: I suggest to add an assert_eq! here and many tests below to verify the sort key. I think there are tests with 2 tags that can help us verify if the sort key is correct. If the data is not good for the verification, feel free to modify or add more

ingester/src/sort_key.rs
```diff
@@ -0,0 +1,233 @@
...
+    primary_key: &[&str],
+) -> HashMap<String, NonZeroU64> {
+    let mut distinct_values_across_batches = HashMap::with_capacity(primary_key.len());
+
+    for batch in batches {
+        for (column, distinct_values) in distinct_values(&batch.data, primary_key) {
+            let set = distinct_values_across_batches
+                .entry(column)
+                .or_insert_with(HashSet::new);
+            set.extend(distinct_values.into_iter());
```
NGA-TRAN, line 65: Does this only extend new distinct value?

NGA-TRAN (approved):
Look good @carols10cents . Only suggest to verify the sort keys in available tests to give us more test coverage.

@tustvold Can you skim this quickly when you have time. I guess Carol should merge even before Raphael review it

ingester/src/sort_key.rs
carols10cents: Yes, sets do that automatically.

carols10cents:
I'm going to merge this; @tustvold and @pauldix I'm happy to fix anything in future PRs!

PR #4139, opened 2022-03-27
feat: wire up compactor scheduler and config

### LINKED ISSUE #4120: NG Compactor: break compaction check up to more bite size pieces
The compaction code currently looks for all possible things to compact, constructs a view of all of them, then tries to do all those compactions and then update for all compactions against the catalog (if I'm understanding the current code correctly).

This should be broken up more so that the compaction for a single partition can be run separately. I'm thinking of something like this:
1. For each sequencer:
2. Identify the best partition to compact
    1. Based on the most level 0 files
    2. Or the most tombstones
    3. Break tie against either criteria by picking the oldest (i.e. created_at is oldest for the level 0 files)
    4. Or if nothing is left to compact, return none
3. Return that info to the caller so that it can schedule compaction for the selected partition

The promotion of older level 0 files with no time overlap to level 1 could happen as part of step 2 if we want to save on work. But the running of compaction should be possible on the single identified partition per sequencer. Then the loop can just check for the next best partition after it completes the last one.

At the very least we need separation of identifying which partitions are highest priority to compact vs. actually running the compaction. So we can limit the scope of work and decide how many compactions to run concurrently.

pauldix:
* Add configuration options for compactor for the max size of level 0 files and split percentage.
* Add metrics for compaction to track the number of candidates, compactions, and durations.
* Add functions to separate identifying partitions to compact from running compaction.
* Make compaction run in smaller chunks, specifically per partition.
* Update compaction to automatically promote level 0 files that are non-overlapping without waiting some period of time.

Closes #4120


iox_catalog/src/postgres.rs
```diff
@@ -1483,6 +1483,24 @@ WHERE parquet_file.sequencer_id = $1
...
+    async fn list_by_partition_not_to_delete(
+        &mut self,
+        partition_id: PartitionId,
+    ) -> Result<Vec<ParquetFile>> {
+        sqlx::query_as::<_, ParquetFile>(
+            r#"
+SELECT *
+FROM parquet_file
+WHERE parquet_file.partition_id = $1
+  AND parquet_file.to_delete IS NULL;
```
domodwyer, line 1495: I think we're going to want at least one index on `partition_id`:

```
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Gather  (cost=1000.00..791136.96 rows=56 width=1143)
   Workers Planned: 2
   ->  Parallel Seq Scan on parquet_file  (cost=0.00..790131.36 rows=23 width=1143)
         Filter: ((to_delete IS NULL) AND (partition_id = 1))
(7 rows)
```

Depending on how many deleted rows we expect per partition, it might be worth adding an index to `to_delete` too?

compactor/src/handler.rs
```diff
@@ -61,22 +65,126 @@ impl CompactorHandlerImpl {
...
+            if used_size > max_size {
+                break;
+            }
+        }
+
+        let compactions_run = handles.len();
+
+        if let Err(e) = futures::future::try_join_all(handles).await {
+            warn!("error compacting: {}", e);
+        }
```
domodwyer, line 183: If one compaction job fails, it'll immediately abort the rest of them - you might want to use something like [`FuturesUnordered`](https://docs.rs/futures/0.3.21/futures/stream/struct.FuturesUnordered.html) here, or the non-short-circuiting [`join_all`](https://docs.rs/futures/0.3.21/futures/future/fn.join_all.html) and inspect the returned array of `Result` for errors directly to allow the other jobs a change to complete.

clap_blocks/src/compactor.rs
```diff
@@ -23,4 +23,34 @@ pub struct CompactorConfig {
...
+    /// size of the input files to be compacted. Currently this only takes into account the
+    /// level 0 files, but should later also consider the level 1 files to be compacted. This
+    /// number should be less than 1/10th of the available memory to ensure compactions have
+    /// enough space to run. Default is 100,000,000 (100MB).
+    #[clap(
+        long = "--compaction-concurrent-size-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_CONCURRENT_SIZE_BYTES",
+        default_value = "100000000"
+    )]
+    pub max_concurrent_compaction_size_bytes: i64,
```
domodwyer, line 55: Perhaps `usize` for these configs? It doesn't make sense to have a maximum of -10 bytes or -10% split and then the CLI will automatically reject invalid (negative) values.

compactor/src/handler.rs
```diff
@@ -61,22 +65,126 @@ impl CompactorHandlerImpl {
...
             object_store,
             exec,
             time_provider,
             BackoffConfig::default(),
+            config.clone(),
+            registry,
         ));
 
+        let shutdown = CancellationToken::new();
+        let runner_handle = tokio::task::spawn(run_compactor(Arc::clone(&compactor_data)));
```
domodwyer, line 83: We can plumb through the shutdown signal / proper cleanup rather simply from here! If we change the `run_compactor` call to include the token above:

```suggestion
        let runner_handle = tokio::task::spawn(run_compactor(
            Arc::clone(&compactor_data),
            shutdown.child_token(),
        ));
```

And update the fn signature to:

```rust
async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken)
```

Then the compactor loop can stop when the shutdown signal is given by changing the unconditional `loop` to:

```rust
while !shutdown.is_cancelled() { /* compaction stuff */ }
```

At this point, calling shutdown on the compactor will cause it to finish up the compactions it is doing, and then gracefully stop. We just need a way for the caller to know when the compactor has finished up with the in-flight compactions - you've done most of that plumbing already! We just need to change the `join()` method in the `CompactorHandler` in this file to:

```rust
    async fn join(&self) {
        self.runner_handle
            .clone()
            .await
            .expect("compactor task failed");
    }
```


And now graceful shutdown is all set up! 🎉 

We might want a different shutdown behaviour (like stop ASAP and throw away in-progress compactions) but they're all easy to tweak from here 👌

domodwyer (approved):

iox_catalog/src/postgres.rs
pauldix: I wouldn't expect that many deleted rows on a per partition basis. But yeah, we'll likely need to add some indexes in. I think we should get a little farther down the path and then figure out what makes the most sense.

clap_blocks/src/compactor.rs
pauldix: Yeah, I just had them as i64 because that's what we have as the underlying data type from Postgres. Lack of unsigned support strikes again.

compactor/src/handler.rs
pauldix: I think finishing the running compactions is probably the best behavior. Each run of this loop is meant to be something that doesn't take a great deal of time. I'd hope that the compactions finish within a few seconds, which seems a reasonable wait time for a shutdown if work is in progress. 

Thanks for the help!

iox_catalog/src/postgres.rs
domodwyer: We wound up adding an index on `to_delete` as part of #4131 anyway as it runs a query against the parquet table with no other predicates that would be able narrow it down from a full table scan: https://github.com/influxdata/influxdb_iox/pull/4131/commits/196507a235b050f1bc796834cbdc00aea8112169

compactor/src/compact.rs
```diff
@@ -177,26 +179,80 @@ pub struct Compactor {
...
+    level_promotion_counter: Metric<U64Counter>,
+
+    /// Gauge for the number of compaction candidates
+    compaction_candidate_gauge: Metric<U64Gauge>,
+
+    /// Gauge for the number of bytes across compaction candidates
+    compaction_candidate_bytes_gauge: Metric<U64Gauge>,
+
+    /// Histogram for tracking the time to compact a partition
+    compaction_duration_ms: Metric<U64Histogram>,
```
NGA-TRAN, line 200: Question about these metrics: Do their values always increase? Will we ever reset them?

compactor/src/compact.rs
```diff
@@ -248,7 +304,160 @@ impl Compactor {
...
+            let duration_ms = self.compaction_duration_ms.recorder(attributes.clone());
+            duration_ms.record(delta.as_millis() as _);
+        }
+
+        let compaction_counter = self.compaction_counter.recorder(attributes);
+        compaction_counter.inc(file_count as u64);
+
+        Ok(())
+    }
+
```
NGA-TRAN, line 460: This comment is for the line below.
When this PR is merged, `find_and_compact` will not be used anymore and we should modify tests of it and then remove it?

compactor/src/compact.rs
```diff
@@ -2196,4 +2412,136 @@ mod tests {
...
+            PartitionCompactionCandidate {
+                sequencer_id: sequencer.id,
+                table_id: table.id,
+                partition_id: partition2.id,
+                level_0_file_count: 1,
+                file_size_bytes: 1337,
+                oldest_file: pf3.created_at,
+            },
+        ];
+        assert_eq!(expect, candidates);
```
NGA-TRAN, line 2545: Very nice test

NGA-TRAN (approved):
Very nice, @pauldix 
After your merge, I will modify current tests to replace `find_and_compact` with `compact_partition` and then remove it. You do not need to worry about tests for `compact_partition`. I think I will add metric tests wile testing that too.

We may need some integration/concurrency tests that I wonder if you are planning to add.

compactor/src/compact.rs
```diff
@@ -248,7 +304,160 @@ impl Compactor {
...
+    /// Returns a list of partitions that have level 0 files to compact along with some summary
+    /// statistics that the scheduler can use to decide which partitions to prioritize. Orders
+    /// them by the number of level 0 files and then size.
+    pub async fn partitions_to_compact(&self) -> Result<Vec<PartitionCompactionCandidate>> {
+        let mut candidates = vec![];
+
+        for sequencer_id in &self.sequencers {
+            // Read level-0 parquet files
+            let level_0_files = self.level_0_parquet_files(*sequencer_id).await?;
+            println!("level 0 files {}", level_0_files.len());
```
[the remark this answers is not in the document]
pauldix: Ah, accidental left over

compactor/src/compact.rs
```diff
@@ -248,7 +304,160 @@ impl Compactor {
...
+            }
+
+            let mut partitions: Vec<_> = partitions.into_values().collect();
+
+            let total_size = partitions.iter().fold(0, |t, p| t + p.file_size_bytes);
+            let attributes =
+                Attributes::from([("sequencer_id", format!("{}", *sequencer_id).into())]);
+
+            let number_gauge = self.compaction_candidate_gauge.recorder(attributes.clone());
+            number_gauge.set(partitions.len() as u64);
```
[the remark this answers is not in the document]
pauldix: How many candidate partitions there are. Either a partition is a candidate for compaction or not. What would happen in the compaction isn't what's being captured. The idea is that we should be able to see if the compactor is getting behind because the number of candidates keeps going up.

compactor/src/compact.rs
pauldix: yeah, I was keeping it in for now to keep from trampling on any in-flight work.

PR #4142, opened 2022-03-28
feat: cache namespace schemas 

crepererum:
In preparation to remove the querier sync loop completely.

This is similar to how the table caches are handled. The biggest portion of this change are tests. As a bonus, we can now also handle the case where a namespace is removed from the catalog.

Ref #4123.


querier/src/cache/namespace.rs
```diff
@@ -1,28 +1,52 @@
...
+    },
+    driver::Cache,
+    loader::FunctionLoader,
+};
+
+/// Duration to keep existing namespaces.
+pub const TTL_EXISTING: Duration = Duration::from_secs(10);
+
+/// Duration to keep non-existing namespaces.
+pub const TTL_NON_EXISTING: Duration = Duration::from_secs(60);
```
alamb, line 22: I don't think it matters but the asymmetry here is confusing to me (as in why aren't these timeouts the same)?

querier/src/cache/namespace.rs
```diff
@@ -74,18 +186,234 @@ mod tests {
...
+                                    column_type: ColumnType::Time,
+                                },
+                            ),
+                        ]),
+                    },
+                ),
+            ]),
+        };
+        assert_eq!(schema1_a.as_ref(), &expected_schema_1);
+        assert_histogram_metric_count(&catalog.metric_registry, "namespace_get_by_name", 1);
```
alamb, line 327: would it be valuable to fetch this schema again and validate it was served from the cache (as is done in the tets below)?

alamb (approved):
👍  looks good to me

querier/src/cache/namespace.rs
crepererum: I found it more likely to change a schema of a namespace than creating a new one.

querier/src/cache/namespace.rs
crepererum: This is done a few lines below (`schema1_b`).

PR #4146, opened 2022-03-28
refactor: remove querier sync loop

### LINKED ISSUE #4123: Querier: remove sync loop altogether

crepererum:
Namespaces are now created on demand and contain their full schema. Tombstones/chunks are created on demand during the query.

Closes #4123.


querier/src/handler.rs
```diff
@@ -63,14 +60,7 @@ impl QuerierHandlerImpl {
...
 
-        let join_handles = vec![(
-            String::from("database sync"),
-            shared_handle(tokio::spawn(database_sync_loop(
-                Arc::clone(&database),
-                shutdown.clone(),
-                Arc::clone(&poison_cabinet),
-            ))),
-        )];
+        let join_handles = vec![];
```
alamb, line 63: are join_handles used at all anymore? Maybe they can be removed entirely

querier/src/handler.rs
```diff
@@ -98,6 +88,8 @@ impl QuerierHandler for QuerierHandlerImpl {
                 panic!("Background worker '{name}' exited early!");
             }
         }
+
+        self.shutdown.cancelled().await;
```
alamb, line 92: I don't understand why `join()`ing the querier would also trigger the cancellation token. I expect that the call to `cancelled()` happens at a higher level.

Specifically,  I expected the sequence to be:

1. call `shutdown.cancelled()`
2. call `join()`

By calling `shutdown.cancelled()` doesn't it means that `join()`ing the querier will eventually cause everything else to shutdown?

querier/src/namespace/mod.rs
```diff
@@ -19,71 +17,72 @@ mod test_util;
...
-    chunk_adapter: Arc<ParquetChunkAdapter>,
-
     /// ID of this namespace.
     id: NamespaceId,
 
     /// Name of this namespace.
     name: Arc<str>,
 
+    /// Tables in this namespace.
+    tables: Arc<HashMap<Arc<str>, Arc<QuerierTable>>>,
```
alamb, line 33: ❤️

alamb (approved):
Love it

querier/src/handler.rs
crepererum: `cancelled()` doesn't trigger, it waits. Don't confuse that with `.cancel()`. Your higher-level API intuition is right and tested by the shutdown test in the same module. The reason we need this additional `.await` call here is that we no longer have any background workers, this method here will return immediately which would cause the querier process to exit.

querier/src/handler.rs
crepererum: I wanted to keep them for now because I'm not sure yet if we need some background loop for some cache maintenance or warm-up work.

querier/src/handler.rs
alamb: 👍

PR #4147, opened 2022-03-28
docs: improve description of `INFLUXDB_..._RANGE_START` and `INFLUXDB_...._RANGE_END`

alamb:
# Rationale
I was confused about what this parameter did in NG initially. As I have been learning more about the various terms in IOx NG I think I have a better understanding now, so I wanted to encode that understanding in the help text

# Changes
Update CLI text


clap_blocks/src/compactor.rs
```diff
@@ -10,14 +10,18 @@ pub struct CompactorConfig {
     )]
     pub topic: String,
 
-    /// Write buffer partition number to start (inclusive) range with
+    /// Start (inclusive) of the write buffer partition ids (sequencer
+    /// ids) this Compactor will compact for. Each Compactor handles
+    /// data from sequencer ids [start, end]
```
crepererum, line 15: I think we should avoid the term "partition" here since technically we have have two data splits -- sharding and partitioning -- and this parameter has nothing to do with the latter.

clap_blocks/src/compactor.rs
```diff
@@ -10,14 +10,18 @@ pub struct CompactorConfig {
...
     #[clap(
         long = "--write-buffer-partition-range-start",
         env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
     )]
     pub write_buffer_partition_range_start: i32,
 
-    /// Write buffer partition number to end (inclusive) range with
+    /// End of the write buffer partition ids (sequencer ids) this
+    /// Compactor will compact for. Each Compactor handles data from
+    /// sequencer ids [start, end]
```
crepererum, line 24: Same comment as above. Would also be nice if the word `inclusive` could stay since some people might not fully grasp the mathematical notation.

clap_blocks/src/compactor.rs
alamb: 🤔  good point -- the `partition` in this case is `kafka` partitions, confusingly, not IOx partitions.


However, the actual parameter name has `partition` in it.... 🤔

clap_blocks/src/compactor.rs
```diff
@@ -10,14 +10,18 @@ pub struct CompactorConfig {
...
+    /// ids) this Compactor will compact for. Each Compactor handles
+    /// data from sequencer ids [start, end]
     #[clap(
         long = "--write-buffer-partition-range-start",
         env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
     )]
     pub write_buffer_partition_range_start: i32,
 
-    /// Write buffer partition number to end (inclusive) range with
+    /// End of the write buffer partition ids (sequencer ids) this
```
alamb, line 22:
```suggestion
    /// End (inclusive) of the write buffer partition ids (sequencer ids) this
```

clap_blocks/src/ingester.rs
```diff
@@ -1,14 +1,18 @@
...
+    /// ids) this Ingester will read from. Each Ingester reads from
+    /// sequencer ids [start, end]
     #[clap(
         long = "--write-buffer-partition-range-start",
         env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
     )]
     pub write_buffer_partition_range_start: i32,
 
-    /// Write buffer partition number to end (inclusive) range with
+    /// End of the write buffer partition ids (sequencer
```
alamb, line 13:
```suggestion
    /// End (inclusive) of the write buffer partition ids (sequencer
```

clap_blocks/src/compactor.rs
crepererum: Oh, could we rename the parameter :sweat_smile:

alamb: I am going to start a conversation with the larger group about renaming the parameter

domodwyer (approved):
I am in favour of renaming it too 👍

alamb:
Update here is that I plan to rename the parameters, but will try and do so in a backwards compatible way (so that I can roll out the changes, update our configs, and then remove the old parameters) as a test of doing so in the Continuous Deployment world

alamb:
I will create a new ticket / PR for this work later today

alamb:
I found another confusingly named one too `INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS` really is auto create partitions...

PR #4149, opened 2022-03-28
refactor: Extract common, OG database and router out of influxdb_ioxd

alamb:
# Rationale

Continues the quest to make `influxdb_ioxd` less monolithic (and thus compile faster)

# Changes
Extracts code from `influxdb_ioxd`:
1. `ioxd_common`
2. `ioxd_database`
3. `ioxd_router`

Removes unneeded dependencies in `influxdb_ioxd`

I next plan to give the same treatment to `router2`, `querier`, `ingester` and `compactor` and possibly fold remaining `influxdb_ioxd` into `influxdb_iox` (depending on what is left)


influxdb_ioxd/src/lib.rs
```diff
@@ -1,78 +1,36 @@
...
-    #[snafu(display("Early RPC shutdown"))]
-    LostRpc,
-
-    #[snafu(display("Early server shutdown"))]
-    LostServer,
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// On unix platforms we want to intercept SIGINT and SIGTERM
```
alamb, line 55: moved to `ioxd_common`

domodwyer (approved):

alamb:
I tested locally that this PR does not conflict with https://github.com/influxdata/influxdb_iox/pull/4139 which I think @pauldix was possibly worried about

I purposely didn't extract the NG components yet -- will do as a follow on PR

PR #4156, opened 2022-03-29
refactor: small query tests clean up

crepererum:
See commits. Will help with system table tests for NG.


query_tests/src/scenarios/util.rs
```diff
@@ -30,13 +30,15 @@ pub struct ChunkDataOld<'a> {
...
+#[derive(Debug, Clone, Default)]
 pub struct ChunkDataNew<'a, 'b> {
     /// Line protocol data of this chunk
     pub lp_lines: Vec<&'a str>,
 
-    /// which stage this chunk will be created
-    pub chunk_stage: ChunkStageNew,
+    /// which stage this chunk will be created.
+    ///
+    /// If not set, this will be randomized.
```
alamb, line 40: I am not convinced that randomized stages is a good idea -- I would rather explicitly list out the stages to test

query_tests/src/scenarios/util.rs
```diff
@@ -1049,20 +1041,37 @@ async fn make_two_chunk_scenarios_new(
...
+            }
 
-        for (chunk_stage, chunk_data) in stages.into_iter().zip(chunks) {
-            let chunk_data = chunk_data.clone().with_chunk_stage(chunk_stage);
             let name = make_ng_chunk(Arc::clone(&ns), chunk_data).await;
 
             write!(&mut scenario_name, "{}", name).unwrap();
         }
 
+        assert!(stages_it.next().is_none(), "generaged too many stages");
```
alamb, line 1073: "generaged" --> "generated"

alamb (approved):

query_tests/src/scenarios/util.rs
crepererum: "All" is actually the better wording and is technically what will happen.

crepererum: Adjusted the wording and extended the explanation to make that clear.

alamb: Thank you

PR #4157, opened 2022-03-29
feat: query log system table for querier

### LINKED ISSUE #4084: Querier: bring back query log
Query log was removed in #4063 but should be brought back.

crepererum:
The code is mostly taken from the OG `db` crate. The PR is that big because it needs some infrastructure for the system tables in general.

Differences to the `db` version are:

- we only have one system table (for now)
- the query log is global, NOT per namespace. ALL our data structures are global so we can scale easier for large number of namespaces (free tier) and configure size constraints in a nicer way, e.g. the standard is to have 10k queries stored globally which doesn't get multiplied by the number of namespaces
- the query log entries also have a namespace ID

Note that the system tables for a specific namespace only show the queries for that particular namespace (i.e. we don't leak information of other users). The plan is to add a system db/namespace that has an admin view on the global data structures and that can read the query log for ALL namespaces.

Closes #4084.


querier/src/database.rs
```diff
@@ -11,6 +14,11 @@ use service_common::QueryDatabaseProvider;
 use std::{collections::HashMap, sync::Arc};
 use time::TimeProvider;
 
+/// The number of entries to store in the circular query buffer log.
+///
+/// That buffer is shared between all namespaces.
```
alamb, line 19:
```suggestion
/// That buffer is shared between all namespaces, and filtered on query
```

querier/src/system_tables/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+
+mod queries;
+
+pub const SYSTEM_SCHEMA: &str = "system";
+
+const QUERIES_TABLE: &str = "queries";
+
+const ALL_SYSTEM_TABLES: &[&str] = &[QUERIES_TABLE];
+
+pub struct SystemSchemaProvider {
```
alamb, line 32: As much as the copy/paste/modify pains me intellectually, I think this approach makes sense 👍

querier/src/system_tables/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        Ok(Arc::new(SystemTableExecutionPlan {
+            table: Arc::clone(&self.table),
+            projection: projection.clone(),
+            projected_schema,
+        }))
+    }
+}
+
+struct SystemTableExecutionPlan<T> {
```
alamb, line 122: I think this looks a lot like `MemoryExec` from datafusion: 

https://github.com/apache/arrow-datafusion/blob/41b4e491663029f653e491b110d0b5e74d08a0b6/datafusion/core/src/physical_plan/memory.rs#L40-L57

querier/src/system_tables/queries.rs
```diff
@@ -0,0 +1,264 @@
...
+use data_types2::NamespaceId;
+use observability_deps::tracing::error;
+use std::{collections::VecDeque, sync::Arc};
+
+/// Implementation of system.queries table
+#[derive(Debug)]
+pub(super) struct QueriesTable {
+    schema: SchemaRef,
+    query_log: Arc<QueryLog>,
+    namespace_id_filter: Option<NamespaceId>,
```
alamb, line 22: I would think we should *always* have a namespace id filter, right?

I am concerned because if it is optional then we may forget to specify it, resulting in a potential data leak

querier/src/system_tables/queries.rs
```diff
@@ -0,0 +1,264 @@
...
+                offset,
+                len,
+                namespace_id_filter.is_none(),
+            ) {
+                Ok(batch) => {
+                    offset += len;
+                    Some(Ok(batch))
+                }
+                Err(e) => {
+                    error!("Error system.chunks table: {:?}", e);
```
alamb, line 71:
```suggestion
                    error!("Error system.queries table: {:?}", e);
```

alamb (approved):
Nice work 👍 


I think the code duplication is ok, but I can't wait to decomission OG and remove it. It is going to be complicated when looking in the code to find multiple things called `QueryLog` I think

I think it would be better to avoid an optional namespace_id filter (and always require one) in order to minimize the potential for cross namespace data leakage, but I don't consider that a deal breaker for this PR

crepererum:
> I think it would be better to avoid an optional namespace_id filter (and always require one) in order to minimize the potential for cross namespace data leakage, but I don't consider that a deal breaker for this PR

My plan is to use the same system table for the system "namespace" (i.e. the admin view that can see everything).

querier/src/system_tables/queries.rs
crepererum: See comment below.

querier/src/system_tables/mod.rs
crepererum: I think the difference is that the system table plan delays the actual record batch creation until `ExecutionPlan::execute` is called, while the `MemoryExec` would require us to create the batches when `TableProvider::scan` is called. TBH I'm not sure if this really makes a difference.

querier/src/system_tables/queries.rs
alamb: I think your plan makes sense

PR #4163, opened 2022-03-29
fix: make sure the QueryableParquetChunks are always sorted correctly

### LINKED ISSUE #4159: Make QueryableParquetChunk's order work correctly even if they have the same sequence number
# Rationale
Since we now split compacted data into 2 files who will have the same sequence number, we need to consider `min_sequence_number` while making order for the `QueryableParquetChunk` to have them ordered correctly

NGA-TRAN:
Closes #4159


compactor/src/query.rs
```diff
@@ -110,15 +110,22 @@ impl QueryChunkMeta for QueryableParquetChunk {
...
+    // Note: parquet_file's id is an uuid which is also the datatype of the ChunkId. However,
+    // it is not safe to use it for sorting chunk
     fn id(&self) -> ChunkId {
-        // always return id 0 for debugging mode and with reason above
-        ChunkId::new_test(0)
+        let timestamp_nano = self.iox_metadata.time_of_first_write.timestamp_nanos();
+        let timestamp_nano_u128 =
+            u128::try_from(timestamp_nano).expect("Cannot convert timestamp nano to u128 ");
+
+        ChunkId::new_test(timestamp_nano_u128)
```
NGA-TRAN, line 128: I am not super happy to return `min_time` here but this function has not been used so far and this is the safest way for sorting the chunks for compacting right now.

compactor/src/query.rs
```diff
@@ -110,15 +110,22 @@ impl QueryChunkMeta for QueryableParquetChunk {
 }
 
 impl QueryChunk for QueryableParquetChunk {
-    // Todo: This function should not be used in this NG chunk context
-    // For now, since we also use scan for both OG and NG, the chunk id
-    // is used as second key in build_deduplicate_plan_for_overlapped_chunks
-    // to sort the chunk to deduplicate them correctly.
-    // Since we make the first key, order, always different, it is fine
-    // to have the second key the sames and always 0
+    // In NG, this function is needed to distiguish the ParquetChunks further if they happen to have the same order.
```
alamb, line 113:
```suggestion
    // In NG, this function is needed to distinguish the ParquetChunks further if they happen to have the same creation order.
```

compactor/src/query.rs
```diff
@@ -110,15 +110,22 @@ impl QueryChunkMeta for QueryableParquetChunk {
...
+    // Ref: chunks.sort_unstable_by_key(|c| (c.order(), c.id())); in provider.rs
+    // Note: The order of this QueryableParquetChunk is the parquet file's min_sequence_number which
+    // will be the same for parquet files of splitted compacted data.
+    //
+    // Thia function returns the parquet file's min_time which will be always different for the parquet files of
```
alamb, line 118:
```suggestion
    // This function returns the parquet file's min_time which will be always different for the parquet files of
```

compactor/src/query.rs
alamb: I think we should rename `ChunkId::new_test`to `ChunkId::new` or something as this is now used outside of test code.

compactor/src/query.rs
```diff
@@ -110,15 +110,22 @@ impl QueryChunkMeta for QueryableParquetChunk {
...
+    // same order/min_sequence_number and is good to order the parquet file
+    //
+    // Note: parquet_file's id is an uuid which is also the datatype of the ChunkId. However,
+    // it is not safe to use it for sorting chunk
     fn id(&self) -> ChunkId {
-        // always return id 0 for debugging mode and with reason above
-        ChunkId::new_test(0)
+        let timestamp_nano = self.iox_metadata.time_of_first_write.timestamp_nanos();
```
alamb, line 124: I thought the order of the insert data in the parquet files was actually the order of `min_sequence_number` (rather than the timestamp of the first write)?

NGA-TRAN: Yes, the `order` is min_sequence_number. This is the `id` which is only effective when the order is the same

compactor/src/query.rs
NGA-TRAN: Let me create a new function name in ChunkId and use it

alamb (approved):

PR #4164, opened 2022-03-29
feat: Support 'SHOW NAMESPACES' in sql repl

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/4178~
 
# Rationale:
I want to be able to explore the data in the querier using my fancy SQL repl without having to know all namespaces up front. OG knows how to do this using `SHOW DATABASES`

# Change
1. Add a metadata gRPC service to the querier
2. use new API to implement `SHOW NAMESPACES` in the sql repl

Muchos kudos to @lukebond  whose `schema` implementation I ~copied~ took a friendly look at

# In action:
```
(arrow_dev) alamb@MacBook-Pro-2:~/Software/influxdb_iox$ ./target/debug/influxdb_iox sql
Connected to IOx Server
Set output format format to pretty
Ready for commands. (Hint: try 'help;')
> show namespaces;
+--------------+-----------------------------------+
| namespace_id | name                              |
+--------------+-----------------------------------+
| 1            | 26f7e5a4b7be365b_917b97a92e883afc |
+--------------+-----------------------------------+
> use 26f7e5a4b7be365b_917b97a92e883afc;
You are now in remote mode, querying database 26f7e5a4b7be365b_917b97a92e883afc
26f7e5a4b7be365b_917b97a92e883afc> select count(*) from cpu;
+-----------------+
| COUNT(UInt8(1)) |
+-----------------+
| 5312            |
+-----------------+
Returned 1 row in 118.927403ms
26f7e5a4b7be365b_917b97a92e883afc>
```


influxdb_ioxd/src/server_type/querier/rpc/namespace.rs
```diff
@@ -0,0 +1,63 @@
...
+        // TODO real error handling
+
+        // Get catalog namespaces
+        let namespaces = self.catalog
+            .repositories()
+            .await
+            .namespaces()
+            .list()
+            .await
+            .unwrap();
```
crepererum, line 51: Maybe we should move this to the querier crate to have a nice interface separation? I think wildly accessing the catalog from the grpc service layer has a bit of a bad code smell.

This could also use a cache, but I'm also fine w/o it.

alamb: Sounds good -- I will move it

alamb: in fded6452f

querier/src/database.rs
```diff
@@ -116,6 +117,16 @@ impl QuerierDatabase {
...
+
+    /// Return all namespaces this querier knows about
+    pub async fn namespaces(&self) -> Result<Vec<Namespace>, iox_catalog::interface::Error> {
+        self.catalog.repositories().await.namespaces().list().await
+    }
+
+    /// return the catalog that is used for this database
+    pub fn catalog(&self) -> Arc<dyn Catalog> {
+        Arc::clone(&self.catalog)
+    }
```
crepererum, line 129: Who is using this method? If it's not required I would opt to remove it so we can discourage accessing the catalog directly and encourage to have proper APIs attached to `QuerierDatabase`/`QuerierNamespace`.

querier/src/database.rs
```diff
@@ -116,6 +117,16 @@ impl QuerierDatabase {
             Arc::clone(&self.query_log),
         )))
     }
+
+    /// Return all namespaces this querier knows about
+    pub async fn namespaces(&self) -> Result<Vec<Namespace>, iox_catalog::interface::Error> {
+        self.catalog.repositories().await.namespaces().list().await
+    }
```
crepererum, line 124: I'm wondering if we should retry IO errors here as well as we do for basically all other catalog interaction :thinking:

crepererum (approved):
Minor nitpicks/questions.

querier/src/database.rs
alamb: Yeah I think it was a left over -- will remove.

PR #4167, opened 2022-03-30
refactor: remove partition key from `ParquetChunk`

crepererum:
The parquet chunk is always wrapped into some higher-level data
structure (e.g. a catalog chunk, a partition, ...) that knows exactly
"where" the chunk is located. There is no need for the parquet chunk to
back-reference container-level attributes. In the contrary:
double-bookkeeping makes the code more complex and costs additional
memory.


tustvold (approved):

NGA-TRAN (commented):
👍  The code clearly shows we do not need it.

PR #4168, opened 2022-03-30
feat: Capture more extreme tail latency buckets for dml handler

mkmik:
when the router performs a rolling restart we currently witness a dramatic increase in tail latencies:

image

however since the majority of request fall into the "+inf" bucket, the actual range is not clear. 

This PR adds a couple of more buckets to get the sense of where we're at.


tustvold (approved):
I assume the cardinality is not a problem?

mkmik:
> I assume the cardinality is not a problem?
yeah it's not a problem, at least not now. We run our own prometheus and we only have a handful of router replicas

PR #4170, opened 2022-03-30
refactor: remove table name from `TableSummary`

crepererum:
This allows us to remove the table name from the low-level chunk
representations (like `ParquetFile`, RUB, ...) since table names are
already tracked by the higher-level data structures (e.g. catalog,
catalog chunk) that manage the low-level chunk representations.

This is similar to #4167.


domodwyer (approved):
Makes sense to me.

alamb (approved):
👍

NGA-TRAN (commented):
👍 for this change.

PR #4174, opened 2022-03-30
feat: `CacheBackend::is_empty`

crepererum:
Adds a simple way to check if a cache backend is empty and use it for some sanity checks.


domodwyer (approved):
Ah that's nice - the "must be empty" invariant for the dual cache thing had me uneasy with no way to check/enfroce it, I'm sad I didn't think of `is_empty()` at the time 😂

Issue #4161, opened 2022-03-29
Add Keysort and binary format keys capabilities to InfluxRPC API


e-dard:
IOx currently supports returning special tag keys for measurement and field in either a textual format: `_measurement` and `_field`, or in a binary format: `\x00`, `\xff`.

However, InfluxCloud query clients won't request the binary format unless there is support for a specific storage capability that ensures series keys are returned in lexicographic order.

The following capability needs to be present:

```json
"KeySortCapability": Capability { features: ["ReadFilter"] }
```

In order to implement this we need to determine that IOx can return all series frames in lexicographic order according to series keys whose measurement and field tag keys use the binary format (which means the tag key for measurement is first and the tag key for field is last).

PR #4176, opened 2022-03-30
feat: add KeySortCapability capability

e-dard:
Closes https://github.com/influxdata/influxdb_iox/issues/4161

This PR teaches IOx to expose a particular capability that encourages Flux/InfluxQL engines to request series keys sorted and with the special binary tag key values for `measurement` and `field`.

No `influxrpc` logic had to change because:

1) we already return series keys in lexicographic order as if they used the binary versions of measurement/field name.
2) we already can support text/binary selection in a `ReadFilter` request. 

The changes here are about signalling to a client that we support this.


alamb (commented):
❤️

PR #4177, opened 2022-03-30
fix: Backfill namespace_id in schema migration

mkmik:
Closes #4173

This PR fixes a bad migration (see ticket for more context)

Unfortunately it's not possible to fix this migration by appending new migrations so we have to modify an existing migration.

Migration steps are meant to be immutable, and the sqlx migration tooling will detect when the hash of a migration step script changes and it will throw an error.

However, there are only two possible scenarios:

1. the catalog had no recorded persisted files and thus the broken migration script executed successfully
2. the catalog had at least one persisted file and thus the migration step failed and is thus not recorded as "done" (our current test environment is stuck in this state)

In the first scenario, that's likely just a test db, so it's reasonable to ask the developer to just re-create the db.

In the second scenario, the fixed sql step will run and move forward without errors.


NGA-TRAN (approved):
Nice fix. Thanks @mkmik for jumping into this quickly

iox_catalog/migrations/20220324152729_add_namespace_id_to_parquet_file.sql
```diff
@@ -1,10 +1,19 @@
...
   ON DELETE NO ACTION
   NOT VALID;
+
+UPDATE parquet_file p
+  SET namespace_id=t.namespace_id
+  FROM table_name t
+  WHERE t.id=p.table_id;
+
+ALTER TABLE parquet_file
+  ALTER COLUMN namespace_id SET NOT NULL;
```
NGA-TRAN, line 19: FYI @carols10cents

PR #4179, opened 2022-03-30
fix: create_or_get_multi for column in catalog now enforces limits

### LINKED ISSUE #4137: Column service limit not enforced
The service limit validation added in #3807 does not protect the `ColumnRepo::create_or_get_many()` code path which is used to [insert the columns during schema validation](https://github.com/influxdata/influxdb_iox/blob/8e85846db6c2e16a6367eea3feb49fc12401ba1e/iox_catalog/src/lib.rs#L156).

This probably slipped by as the insert-many code path in the in-memory catalog [delegates to `ColumnRepo::create_or_get()`](https://github.com/influxdata/influxdb_iox/blob/8e85846db6c2e16a6367eea3feb49fc12401ba1e/iox_catalog/src/mem.rs#L567) and therefore hits the correctly limited code path.

The current catalog tests assert that the limits [can be updated](https://github.com/influxdata/influxdb_iox/blob/8e85846db6c2e16a6367eea3feb49fc12401ba1e/iox_catalog/src/interface.rs#L725-L739) but I guess we should have tests that assert the limit is reached too to try and help prevent behavioural divergence between the various impls.

lukebond:
Fixes #4137 

Column limits were not enforced when inserted via `create_or_get_many()`.

Warning: the SQL is not pretty!

Note that the batch insert can do a partial update in which case it will return an error saying so.

Some noise created by reordering the functions in the columns repo to put the two create_or_get() fns next to each other.


iox_catalog/src/postgres.rs
```diff
@@ -919,7 +903,25 @@ WHERE table_name.namespace_id = $1;
         let out = sqlx::query_as::<_, Column>(
             r#"
 INSERT INTO column_name ( name, table_id, column_type )
-SELECT name, table_id, column_type FROM UNNEST($1, $2, $3) as a(name, table_id, column_type)
+SELECT column_name, input.table_id, column_type
+FROM UNNEST($1, $2, $3) AS input(column_name, table_id, column_type)
+INNER JOIN
+(
+    SELECT table_id, COUNT(table_id) AS input_per_table_count
+    FROM UNNEST($1, $2, $3) AS input2(column_name, table_id, column_type)
```
NGA-TRAN, line 911: This UNNEST only needs $2. We can remove $1 and $3

iox_catalog/src/postgres.rs
```diff
@@ -919,7 +903,25 @@ WHERE table_name.namespace_id = $1;
...
+RIGHT JOIN
+(
+    SELECT max_columns_per_table, namespace.id, table_name.id AS table_id, COUNT(column_name.*) AS count_existing
+    FROM namespace
+    LEFT JOIN table_name ON namespace.id = table_name.namespace_id
+    LEFT JOIN column_name ON table_name.id = column_name.table_id
+    GROUP BY namespace.max_columns_per_table, namespace.id, table_name.id
+) AS get_existing_count(max_columns_per_table, namespace_id, table_id, count_existing)
+ON input.table_id = get_existing_count.table_id
+WHERE count_existing + input_per_table_count <= max_columns_per_table
```
NGA-TRAN, line 924: I do not really look into the semantic/meaning of this insert query with many-join subqueries but if the array values of $1, $2, and $3 are only a few, and I think other tables namespace, table_name and column_name are also quite small, the performance of this insert should be fine.
If this query happens to be very slow, I can help understand the requirements and see if there are other ways around.

NGA-TRAN (approved):

iox_catalog/src/postgres.rs
```diff
@@ -919,7 +903,25 @@ WHERE table_name.namespace_id = $1;
...
 INSERT INTO column_name ( name, table_id, column_type )
-SELECT name, table_id, column_type FROM UNNEST($1, $2, $3) as a(name, table_id, column_type)
+SELECT column_name, input.table_id, column_type
+FROM UNNEST($1, $2, $3) AS input(column_name, table_id, column_type)
+INNER JOIN
+(
+    SELECT table_id, COUNT(table_id) AS input_per_table_count
+    FROM UNNEST($1, $2, $3) AS input2(column_name, table_id, column_type)
+    GROUP BY table_id
+) AS get_input_count(table_id, input_per_table_count)
```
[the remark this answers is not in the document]
lukebond: 2022-03-31-112915_515x512_scrot

yep that's right. see image for it visualised. i'll add a comment :+1:

iox_catalog/src/postgres.rs
```diff
@@ -919,7 +903,25 @@ WHERE table_name.namespace_id = $1;
...
+ON input.table_id = get_input_count.table_id
+RIGHT JOIN
+(
+    SELECT max_columns_per_table, namespace.id, table_name.id AS table_id, COUNT(column_name.*) AS count_existing
+    FROM namespace
+    LEFT JOIN table_name ON namespace.id = table_name.namespace_id
+    LEFT JOIN column_name ON table_name.id = column_name.table_id
+    GROUP BY namespace.max_columns_per_table, namespace.id, table_name.id
```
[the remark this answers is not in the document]
lukebond: yes this gets the limit and the current count. the other columns just there to make the joins work. i'll add a comment.

in hindsight this doesn't need to be a `RIGHT JOIN` - i think that's a remnant from an earlier version of the query where that subquery would not return rows that don't have capacity for the new rows within the limit. it no longer works like that. if i change it to an inner join it still works and maybe will be slightly faster or at least one won't be scratching your head asking "why is it a RIGHT JOIN?"

lukebond:
> think we need to clearly document the partial-write semantics now `create_or_get_many()` can partially fail, and given the complexity of the query needed to support this restriction

have pushed some comments to clarify this behaviour

iox_catalog/src/postgres.rs
```diff
@@ -2436,4 +2478,225 @@ mod tests {
...
+        calls = {
+            [
+                ("test1", 1) => ColumnType::Tag,
+                ("test2", 1) => ColumnType::Tag,
+                ("test3", 1) => ColumnType::Tag,
+                ("test4", 1) => ColumnType::Tag,
+            ]
+        },
+        want = Ok(_)
+    );
```
domodwyer, line 2605: Very cool! I love that you managed to drive all the various edge cases for this impl in tests 👍

domodwyer (approved):
Thanks for taking the time to thoroughly document & test this - great test coverage!

PR #4180, opened 2022-03-30
feat: add protobuf-compiler to CI image

tustvold:
The latest version of prost no longer bundles protobuf - https://github.com/tokio-rs/prost/pull/610

It will currently fallback to build using cmake, but this isn't installed in the images which is causing CI to fail - https://github.com/influxdata/pbjson/pull/38.

Whilst installing cmake will fix this problem, it seems better to just install the protobuf-compiler itself


tustvold:
Will test CI image, then merge

tustvold:
https://github.com/influxdata/pbjson/pull/39 is now green so merging this

PR #4182, opened 2022-03-30
refactor: extract querier into ioxd_querier

alamb:
# Rationale
Follow on to https://github.com/influxdata/influxdb_iox/pull/4149

Continues the quest to make `ioxd` less monolithic (and thus compile faster)

# Changes
Extracts code from `ioxd`:
1. `ioxd_querier`

Removes unneeded dependencies in `influxdb_ioxd`


carols10cents (approved):
Love it!

PR #4184, opened 2022-03-30
fix: have the compaction level set correctly

### LINKED ISSUE #4165: Compactor: compacted files should always be level 1
I noticed that the compactor creates files that are level 0, which isn't the correct behavior. When compaction runs for a partition the result should be that:
* Compacted files (i.e. ones that are rewritten) are marked with to_delete timestamps
* Newly created files are:
    * Non-overlapping in time
    * Level 1
    * Have processed tombstone records if appropriate

Essentially, if compaction is called on a partition and no new files are created by the ingester while compaction is running, the end result should be a partition with the minimum number of files that are non-overlapping in time that have all tombstones processed.

NGA-TRAN:
Closes #4165 

Pass the right compaction level to create parquet_files 


iox_catalog/src/postgres.rs
```diff
@@ -1412,7 +1413,7 @@ RETURNING *;
         .bind(file_size_bytes) // $9
         .bind(parquet_metadata) // $10
         .bind(row_count) // $11
-        .bind(INITIAL_COMPACTION_LEVEL) // $12
+        .bind(compaction_level) // $12
```
NGA-TRAN, line 1416: This is the core of the fix

pauldix (approved):
LGTM!

PR #4185, opened 2022-03-30
test: replace find_and_compact with compact_partition in tests

NGA-TRAN:
Tests for `compact_partition` (follow-up of #4139 ) and remove no-longer-used function `find_and_compact`


compactor/src/compact.rs
```diff
@@ -437,6 +437,9 @@ impl Compactor {
             txn.commit().await.context(TransactionCommitSnafu)?;
         }
 
+        // Remove fully processed tombstones
+        self.remove_fully_processed_tombstones(tombstones).await?;
```
NGA-TRAN, line 441: This is a fix. I think @pauldix missed this because I merged this functionality after he had started creating this new function. Thanks to the available tests that discovered this in a blink

pauldix (approved):

PR #4187, opened 2022-03-31
fix: consider "in-between" tombstones as processed

crepererum:
Abstract
========
We need to be careful w/ tombstone that fall exactly in sequence number range of a parquet file.

Current Bug
===========
Imagine the following order of events:

1. Router creates write at sequence number 1:
   - `table,selector=1 payload=1 1`
   - `table,selector=2 payload=2 2`
2. Ingester pulls write, waits a bit and persists it to parquet file 1:
   - `table,selector=1 payload=1 1`
   - `table,selector=2 payload=2 2`
4. Router creates write at sequence number 2:
   - `table,selector=1 payload=3 3`
   - `table,selector=2 payload=4 4`
5. Ingester pulls write
6. Router create delete at sequencer number 3: full time range, `selector=1`
7. Ingeser pulls delete and creates tombstone 1
8. Router creates write at sequence number 4:
   - `table,selector=1 payload=5 5`
   - `table,selector=2 payload=6 6`
9. Ingester pulls write
10. Ingester persists parquet file 2:
    - `table,selector=2 payload=4 4`
    - `table,selector=1 payload=5 5`
    - `table,selector=2 payload=6 6`

When reading parquet file 2, the tombstone MUST NOT be applied. Otherwise `table,selector=1 payload=5 5` will be
deleted.

Notes
=====
Technically this issue also applies to files created by the compactor, however the compactor marks tombstones as
processed that fall into the sequence number range. It even does that in a single transaction:

https://github.com/influxdata/influxdb_iox/blob/fc4635a33476dcc69880e4db48e553b2a75329b0/compactor/src/compact.rs#L821-L861

Alternative
===========
An alternative solution would be if the ingester would mark tombstones that it materialized during persistence as
"processed" (tombstone 1 for parquet file 2 in the example above). However "processed" markers are currently a mere
optimization and don't affect correctness, which is nice for caching on the querier side as well as reasoning.


alamb:
I think this is one that will need @NGA-TRAN  to review carefully. I don't understand the scenario under consideration

querier/src/table/mod.rs
```diff
@@ -120,8 +120,24 @@ impl QuerierTable {
...
+                    //    further down).
+                    // 3. Tombstone is in the min-max sequencer number range of the chunk:
+                    //    Technically the querier has NO way to determine the rows that are affected by the tombstone
+                    //    since we have no row-level sequence numbers. Such a file can be created by two sources -- the
+                    //    ingester and the compactor. The ingester must have materialized the tombstone while creating
+                    //    the parquet file, so the querier can skip it. The compactor also materialized the tombstones,
+                    //    so we can skip it as well. In the compactor case the tombstone will even be marked as
+                    //    processed.
+                    //
+                    // So the querier only needs to consider the tombstone in case 2.
```
NGA-TRAN, line 139: 👍

querier/src/table/mod.rs
```diff
@@ -288,17 +318,23 @@ mod tests {
...
         );
 
         // check delete predicates
         // file112: marked as processed
         assert_eq!(chunks[0].delete_predicates().len(), 0);
         // file113: has delete predicate
-        assert_eq!(chunks[1].delete_predicates().len(), 1);
-        // file114: came after in sequencer
+        assert_eq!(chunks[1].delete_predicates().len(), 2);
+        // file114: predicates are directly within the chunk range => assume they are materialized
```
NGA-TRAN, line 333: 👍

NGA-TRAN (approved):
Very good scenario. Great found @crepererum 
I will look into Ingester and Compactor to make sure this works correctly there. Must need some changes

NGA-TRAN:
> Very good scenario. Great found @crepererum
> I will look into Ingester and Compactor to make sure this works correctly there. Must need some changes

I have checked. The Ingester and Compactor work correctly. The Ingester snapshots and applies deletes when a tombstone arrives so we do not need to worry about tombstone's sequence_number that falls in between  min and max sequence numbers of the parquet file of many Ingester's snapshots. The Compactor and Querier only need to apply tombstones whose sequence number greater than the parquet_file's max sequence number.  We are all good :)

@alamb: This is a bit tricky. I can walk you through each step of the scenario in this PR to see why they end up like that. @pauldix and I had to talk several times to come up with this design of Ingester to cover all cases that simplify the work for Compactor & Querier :)

PR #4189, opened 2022-03-31
feat: LRU cache infrastructure

crepererum:
This allows us to build LRU caches on top of memory pools (e.g. RAM, on-disk, somewhere else). The beauty is that multiple caches can share a single pool, which vastly simplifies composing the whole cache infrastructure.

The shared pool adds a bit of complexity, but I've tried to document everything as good as possible. But let me know if anything is unclear.

The actual usage within the querier will follow in another PR. I'll likely use this layer for metrics as well, so we don't need to account sizes twice.

Another possible extension (which I don't want to put in this PR here) would be priorities in addition to "last used timestamps" so that we evict less performance-sensitive data first.


querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+impl<K, V, S> CacheBackend for LruBackend<K, V, S>
+where
+    K: Clone + Eq + Debug + Hash + Ord + Send + 'static,
+    V: Clone + Debug + Send + 'static,
+    S: MemorySize,
+{
+    type K = K;
+    type V = V;
+
+    fn get(&mut self, k: &Self::K) -> Option<Self::V> {
```
alamb, line 324: This API will be subject to thundering herd  (aka things will race to update the LRU) which is probably fine; I just wanted to point it out 

It may be worth thinking about in the future

crepererum:
> Specifically, how important is the design of multiple backends? I am guessing this is basically done so some other system can retain the ownership of cached items rather than having this cache own them?

There is a requirement to have multiple OUTER backends. Let's say we wanna cache namespace schemas and RUB chunks. Both consume RAM, i.e. the same memory pool. The keys and values for both are different though.

I think there isn't really a requirement to have multiple INNER backends. However the only alternative design I see is to do some kind of `Any` backend hackery to type-erase the different outer backends (so that `K1, V1` and `K2, V2` can use the same `Any, Any` backend). I think this is something that you're proposing. I actually had something like this locally and it works, however it comes with some costs:

- all hash and compare operations for the type erased keys go throw dyn-dispatch
- the code is quite wild (no unsafe code, but I think the solution in this PR here is easier to understand).
- You have lock contention on the untyped `Any` backend also for `get` operation, while for this PR here you only need to lock the memory pool for modifications

I've opened #4200 as a draft which illustrates that code. BTW: even there you still need locks, because the outer backends now use a single inner untyped backend.

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+        ));
+        let size_provider = Arc::new(TestMemorySizeProvider {});
+
+        let backend1 = LruBackend::new(
+            Box::new(HashMap::new()),
+            Arc::clone(&pool),
+            String::from("id"),
+            Arc::clone(&size_provider) as _,
+        );
+        drop(backend1);
```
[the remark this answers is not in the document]
crepererum: Compare that w/ `test_panic_id_collission` which doesn't drop. The ID is then registered twice, leading to a panic (which is actually what we want). This test here just makes sure that once you drop a backend w/ an ID, you're free to recreate it (I think we won't use this property, but it makes things nicer to reason about).

querier/src/cache_system/backend/lru.rs
crepererum: I agree, but at the same time I think we could work around it. At the moment the addressable heap that is the core data structure requires `mut` access when you want to "bump" an entry. There are at least to ways to workaround id:

- use some lock-free/nearly-lock-free "update log" that captures the "touch" events and only materialize it into the heap based on size bounds or when we want to evaluate the heap (i.e. due to memory pressure)
- make the heap smarter so you don't need to lock the whole thing (probably needs some more advanced data structure)

We can also easily instrument these locks, I just don't want to add more complexity to the initial PR.

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+//! - **Cyclic Structure:** The LRU backends communicate with the pool, but the pool also neeeds to communicate with
+//!   all the backends. This creates some form of cyclic data structure.
+//! - **Type Erasure:** The pool is only specific to the memory size type, not the key and value types of the
+//!   participating backends. So at some place we need to perform type erasure.
+//!
+//! ## Data Structures
+//!
+//! ```text
+//!               .~~~~~~~~~~~~.            .~~~~~~~~~~~~~~~~~.
+//! ------------->: MemoryPool :--(mutex)-->: MemoryPoolInner :-----------------------------------+
```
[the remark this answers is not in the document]
crepererum: Good point. I will add an explicit test before merging (i.e. "dropping the outer backend drops the inner one, even when the pool is still around") but this is already solved by `impl Drop for LruBackend` which calls `MemoryPoolInner::unregister_member`. The user can only hold `LruBackend` and `MemoryPool`, the other types are private.

alamb:
I think in general I am missing "the forest for the trees" so to speak -- as in I don't really get the higher level plan that this PR fits into. 

I am a bit worried that we are building up a complicated caching system without fully understanding what the first performance issue we will hit in the querier. 

What would you think of focusing for a while on getting the querier rolled out into production talking with the ingester and putting additional caching on pause for a while?

Maybe when you are back we can have a chat and figure out what to do next.

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+    fn as_any(&self) -> &dyn Any {
+        self as &dyn Any
+    }
+}
+
+/// A member of a [`MemoryPool`]/[`MemoryPoolInner`].
+///
+/// Must be [locked](Self::lock) to gain access.
+///
+/// The only implementation of this is [`PoolMemberImpl`]. This indiction is required to erase `K` and `V` from specific
```
alamb, line 397:
```suggestion
/// The only implementation of this is [`PoolMemberImpl`]. This indirection is required to erase `K` and `V` from specific
```

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+    fn lock(&self) -> Box<dyn PoolMemberGuard<S = Self::S> + '_> {
+        Box::new(PoolMemberGuardImpl {
+            inner: self.inner.lock(),
+        })
+    }
+}
+
+/// Locked [`MemoryPool`]/[`MemoryPoolInner`] member.
+///
+/// The only implementation of this is [`PoolMemberGuardImpl`]. This indiction is required to erase `K` and `V` from
```
alamb, line 437:
```suggestion
/// The only implementation of this is [`PoolMemberGuardImpl`]. This indirection is required to erase `K` and `V` from
```

querier/src/cache_system/backend/lru.rs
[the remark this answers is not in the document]
alamb: I think it would help future readers to add that rationale to the test

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
+//! LRU (Least Recently Used) cache system.
+//!
+//! # Internals
```
alamb, line 3: In addition to an internals document, a high level "here is the high level public API for users of this system" type document (example in docstring??) would likely help

querier/src/cache_system/backend/size.rs
```diff
@@ -0,0 +1,28 @@
+use std::{
+    fmt::Debug,
+    ops::{Add, Sub},
+};
+
+/// Strongly-typed memory size.
+///
+/// Can be used to represent in-RAM memory as well as on-disc memory.
+pub trait MemorySize:
```
alamb, line 9: Maybe this could be called "ResourceConsumption" or something to make it more general -- the fact that this is called `Memory` definitely would be confusing to me if it were used for disk / storage accounting as well

alamb (approved):
I went through this code again. I still think it *feels* overly complicated to me, but I have no specific reasons we shouldn't merge it. 

I think we should merge this PR, try to hook it up in the querier, and then see how much, if any, any additional finagling to use. 


# Follow on (another PR) suggestion for readability
I think putting `cache_system` into its own crate would help make clear what the external API is meant to be and also enforce the notion it is not tied to any particular implementation

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+    fn is_empty(&self) -> bool {
+        self.inner.lock().last_used.is_empty()
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self as &dyn Any
+    }
+}
+
+/// A member of a [`MemoryPool`]/[`MemoryPoolInner`].
```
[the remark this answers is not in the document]
alamb: What I am having trouble with is mapping "PoolMember" (specifically as it appears in the diagram) and "thing as end user of this system would understand -- like "memory bytes in ReadBuffer". 

Is `PoolMember a wrapper around that? 

I am sorry for my confusion

querier/src/cache_system/backend/lru.rs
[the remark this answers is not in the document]
crepererum: done

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,968 @@
...
+            String::from("id1"),
+            Arc::clone(&size_provider) as _,
+        );
+
+        backend.set(String::from("a"), 2usize);
+        assert_eq!(pool.current().0, 0);
+    }
+
+    #[test]
+    fn test_generic() {
```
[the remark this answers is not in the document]
crepererum: done

querier/src/cache_system/backend/lru.rs
crepererum: added a high-level usage example, let me know if that helps

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,1198 @@
+//! LRU (Least Recently Used) cache system.
```
alamb, line 1: 😍

querier/src/cache_system/backend/lru.rs
```diff
@@ -0,0 +1,1198 @@
...
+//!
+//! // set up a memory pool
+//! let limit = RamSize(50);
+//! let pool = Arc::new(MemoryPool::new(limit, time_provider));
+//!
+//! // set up first pool user: a u64->String map
+//! #[derive(Debug)]
+//! struct Provider1 {}
+//!
+//! impl MemorySizeProvider for Provider1 {
```
alamb, line 55: this is super helpful 👍  thank you

querier/src/cache_system/backend/lru.rs
alamb: it is great -- thank you -- very helpful

querier/src/cache_system/backend/lru.rs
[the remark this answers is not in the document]
alamb: The example shows me what I was missing -- ❤️

alamb (approved):

querier/src/cache_system/backend/size.rs
crepererum: done

alamb:
🎉

PR #4190, opened 2022-03-31
refactor: rename Sequence::id and Sequence::number field names

alamb:
refactor: rename Sequence::id and Sequence::number field names

# Rationale:
1. Using consistent names for the same concept in the codebase is :thumbsup:

# Changes
1. Rename fields


data_types/src/sequence.rs
```diff
@@ -1,14 +1,14 @@
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct Sequence {
-    pub id: u32,
-    pub number: u64,
+    pub sequencer_id: u32,
+    pub sequence_number: u64,
```
crepererum, line 4: It feels a bit repetitive, but if it helps people to understand the code and avoids confusion, I'm OK with it.

crepererum (approved):

Issue #4191, opened 2022-03-31
Release build doesn't include git_hash


mkmik:
Our release build doesn't seem to include the `git_hash` commit anymore

__Expected behaviour:__

```
level=info msg="InfluxDB IOx server starting" git_hash=<some git hash> num_cpus=32 build_malloc_conf="\u{0}" target="ioxd" location="ioxd/src/lib.rs:104" time=1648737910972101624
```

__Actual behaviour:__

```
level=info msg="InfluxDB IOx server starting" git_hash=UNKNOWN num_cpus=32 build_malloc_conf="\u{0}" target="ioxd" location="ioxd/src/lib.rs:104" time=1648737910972101624
```

__Environment info:__

* happens in our internal testing cluster with image: quay.io/influxdb/iox@sha256:5bc74fa39fff3277ae6ddf874aea98b17bb3abebe74015e1574a2b1a03460748

alamb:
What happened is that the GIT_SHA env is being set in the `influxdb_iox` crate  but the message is being created in the `ioxd` (formerly `influxdb_ioxd`) crate which thus can't find ir.

I plan a two step fix:
1. Move the relevant code out of `ioxd`  which aligns with https://github.com/influxdata/influxdb_iox/issues/4145
2. Cause build to fail if the value can't be found, as suggested by @domdwyer

Issue #4145, opened 2022-03-28
NG Querier: APIs for exposing DataLifecycle / sequence numbers


alamb:
When trying to query NG, it will be important in tests to understand how far a particular write has made it in the [NG DataLifecycle](https://docs.google.com/document/d/19fA0XtXHPBnBxB2yc6jFMu3cBnv1jsVpovbFL1NqVzc/edit#) and become readable. It may also be helpful to be able to wait for it to be 'persisted / in parquet'

The definition of done for this ticket is
1. Defined an API for the router (and ingester / querier) that can be used to wait for data to become readable
2. Implemented said API and ported over tests

PR #4193, opened 2022-03-31
fix: include git sha (again) in release build

alamb:
resolves https://github.com/influxdata/influxdb_iox/issues/4191

# Backstory
What happened is that the GIT_SHA env is being set in the `influxdb_iox` crate  but the message is being created in the `ioxd` (formerly `influxdb_ioxd`) crate which thus can't find it.

# Changes
1. Move the code that uses `GIT_SHA` out of `ioxd` and into `influxdb_iox` (which aligns with https://github.com/influxdata/influxdb_iox/issues/4145 as well) 
2. Use same `git` invocation as in `iox_gitops_adapter`
3. Change the code to use `env!` rather than `option_env!` that will cause the build to fail if the `GIT_HASH` value can't be found, as suggested by @domodwyer 

# Example of it working

Here is an image (build from https://github.com/influxdata/influxdb_iox/pull/4193/commits/9757b22d7629c690947b3b9b81f5b58731668ecf) that shows the git sha again:
```shell
(arrow_dev) alamb@MacBook-Pro-2:~/Software/influxdb_iox$ docker run -it --entrypoint /bin/bash quay.io/influxdb/iox:9757b22d 
....
Status: Downloaded newer image for quay.io/influxdb/iox:9757b22d
iox@dd236825c987:/$ /usr/bin/influxdb_iox run -vv
WARNING: Not specifying the run-mode is deprecated. Defaulting to 'database'.
2022-03-31T18:07:46.886820Z  WARN clap_blocks::object_store: NO PERSISTENCE: using Memory for object storage
2022-03-31T18:07:46.886951Z  INFO server::application: using specified number of threads per thread pool num_threads=15
2022-03-31T18:07:46.887190Z DEBUG datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
....
```

🎉


influxdb_iox/build.rs
```diff
@@ -1,14 +1,29 @@
...
 //
 // https://stackoverflow.com/questions/43753491/include-git-commit-hash-as-string-into-rust-program
 use std::process::Command;
-fn main() {
-    let output = Command::new("git").args(&["rev-parse", "HEAD"]).output();
 
-    if let Ok(output) = output {
-        if let Ok(git_hash) = String::from_utf8(output.stdout) {
-            println!("cargo:rustc-env=GIT_HASH={}", git_hash);
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
alamb, line 7: copied from https://github.com/influxdata/influxdb_iox/blob/main/iox_gitops_adapter/build.rs#L26

influxdb_iox/src/commands/run/main.rs
```diff
@@ -0,0 +1,173 @@
+use ioxd::Service;
```
alamb, line 1: This was simply moved from `ioxd/src/lib.rs` -- it is not new

influxdb_iox/src/commands/run/main.rs
```diff
@@ -0,0 +1,173 @@
...
+/// This is the entry point for the IOx server.
+///
+/// This entry point ensures that the given set of Services are
+/// started using best practice, e.g. that we print the GIT-hash and
+/// malloc-configs, that a panic handler is installed, etc.
+///
+/// Due to its invasive nature (install global panic handling,
+/// logging, etc) this function should not be used during unit tests.
+pub async fn main(common_state: CommonServerState, services: Vec<Service>) -> Result<()> {
+    let git_hash = env!("GIT_HASH", "starting influxdb_iox server");
```
alamb, line 70: The change is to use `env!` here rather than `option_env!` to fail build if githash is not set

ioxd/src/lib.rs
```diff
@@ -1,74 +1,15 @@
-use clap_blocks::run_config::RunConfig;
```
alamb, line 1: THis code was just moved into `influxdb_iox`

ioxd/Cargo.toml
```diff
@@ -47,16 +47,12 @@ num_cpus = "1.13.0"
 pprof = { version = "0.7", default-features = false, features = ["flamegraph", "prost-codec"], optional = true }
 snafu = "0.7"
 thiserror = "1.0.30"
-tikv-jemalloc-ctl = { version = "0.4.0", optional = true }
```
alamb, line 50: I strongly suspect there are more dependencies in this crate that is not used, but I plan to remove this crate eventually shortly, so I don't want to spend too much time

influxdb_iox/src/main.rs
```diff
@@ -46,7 +46,7 @@ static VERSION_STRING: Lazy<String> = Lazy::new(|| {
     format!(
         "{}, revision {}",
         option_env!("CARGO_PKG_VERSION").unwrap_or("UNKNOWN"),
-        option_env!("GIT_HASH").unwrap_or("UNKNOWN")
+        env!("GIT_HASH", "Can not find find GIT HASH in environment")
```
domodwyer, line 49: I doubt we'll hit this, but can't hurt:

```suggestion
        env!("GIT_HASH", "Can not find find GIT HASH in build environment")
```

domodwyer (approved):

alamb:
Rebased because semantic pull request is cruel:

Screen Shot 2022-03-31 at 2 55 43 PM

PR #4198, opened 2022-03-31
fix: add indexes to parquet_file

pauldix:
* Add indexes so compactor can find candidate partitions and specific partition files quickly.
* Limit number of level 0 files returned for determining candidates. This should ensure that if comapction is very backed up, it will be able to work through the backlog without evaluating the entire world.


domodwyer (approved):

domodwyer:
I'm going to merge this so we can look at PG load with these changes applied and track down any other hot spots.

PR #4200, opened 2022-04-01
feat: shared cache

crepererum:
Just a draft, see discussion in #4189.


crepererum:
Closing since we've accepted #4189 as a solution.

PR #4201, opened 2022-04-01
test: add benchmarks for addressable heap

crepererum:
When introducing the addressable heap, it was suggested to replace the `VecDequeue` with a `BTreeSet`. Let's have some benchmarks to see if that makes actually sense (hint: it's not that easy).

### DIFF
--- a/querier/benches/addressable_heap.rs
+++ b/querier/benches/addressable_heap.rs
@@ -1,0 +1,303 @@
+use criterion::{
+    criterion_group, criterion_main, measurement::WallTime, AxisScale, BatchSize, BenchmarkGroup,
+    BenchmarkId, Criterion, PlotConfiguration, SamplingMode,
+};
+use querier::AddressableHeap;
+use rand::{prelude::SliceRandom, thread_rng, Rng};
+
+type TestHeap = AddressableHeap<u64, (), u64>;
+
+const TEST_SIZES: &[usize] = &[0, 1, 10, 100, 1_000, 10_000];
+
+#[derive(Debug, Clone)]
+struct Entry {
+    k: u64,
+    o: u64,
+}
+
+impl Entry {
+    fn new_random<R>(rng: &mut R) -> Self
+    where
+        R: Rng,
+    {
+        Self {
+            k: rng.gen(),
+            o: rng.gen(),
+        }
+    }
+
+    fn new_random_n<R>(rng: &mut R, n: usize) -> Vec<Self>
+    where
+        R: Rng,
+    {
+        (0..n).map(|_| Self::new_random(rng)).collect()
+    }
+}
+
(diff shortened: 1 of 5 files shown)


querier/benches/addressable_heap.rs
```diff
@@ -0,0 +1,303 @@
...
+    setup_group(&mut g);
+
+    let mut rng = thread_rng();
+
+    for n in TEST_SIZES {
+        g.bench_with_input(BenchmarkId::from_parameter(n), &n, |b, &_n| {
+            b.iter_batched(
+                || {
+                    let (heap, _entries) = create_filled_heap(&mut rng, *n);
+                    let entry = Entry::new_random(&mut rng);
```
alamb, line 151: 👍

querier/benches/addressable_heap.rs
```diff
@@ -0,0 +1,303 @@
...
+                    (heap, entry)
+                },
+                |(mut heap, entry)| {
+                    heap.remove(&entry.k);
```
alamb, line 241: I wonder if this would be a more realistic test if  `entry` would be inserted first before removing it?

alamb (approved):

querier/benches/addressable_heap.rs
```diff
@@ -0,0 +1,303 @@
+use criterion::{
+    criterion_group, criterion_main, measurement::WallTime, AxisScale, BatchSize, BenchmarkGroup,
+    BenchmarkId, Criterion, PlotConfiguration, SamplingMode,
+};
+use querier::AddressableHeap;
+use rand::{prelude::SliceRandom, thread_rng, Rng};
+
+type TestHeap = AddressableHeap<u64, (), u64>;
```
domodwyer, line 8: Is moving zero-sized values around during a resize a realistic test?

domodwyer (approved):
It might be more useful to measure insert/remove at fixed points (start, middle, end) as that is what would highlight the difference in structures - what was the outcome with the current setup?

I also don't know enough about the expected workload to know if the key/value size and numbers of elements are representative, but trust what you've put together 👍  

Benchmarks are always nice to have!

querier/benches/addressable_heap.rs
crepererum: This is covered by `bench_remove_existing_after_n_elements`.

querier/benches/addressable_heap.rs
crepererum: Fixed

crepererum:
> It might be more useful to measure insert/remove at fixed points (start, middle, end) as that is what would highlight the difference in structures - what was the outcome with the current setup?

The element element is picked randomly, so if the data structure is tailored for start/middle/end you'll see that in the test distribution. For "remove min" (which we'll use often) there is a special OP called "pop" which is also tested here.

PR #4202, opened 2022-04-01
feat: return write_token from HTTP writes to router2

alamb:
# Rationale:
part of https://github.com/influxdata/influxdb_iox/issues/4145

We want to be able to know when data is readable and when it is persisted. The first step is knowing information about the sequence numbers

See https://docs.google.com/document/d/1VFB7xmVn_ezet0gzFxE9x5pNS-1IhXi26OQOWsjn5OU/edit#heading=h.pr4zatesz5xi

# Changes:
1. Add `write_summary` crate
1. Add protobuf encoding / decoding for write summaries
3. Thread `DmlMeta` and `WriteSummary` into dml handler chain (DONE)
2. return encoded write token as part of /api/v2/write (DONE)
3. End to end tests for same

Follow on work
- [ ] Add an API to the ingester that will report when all data for a write becomes readable and update end to end test to use that
- [ ] Add write tokens to the grpc write endpoint - https://github.com/influxdata/influxdb_iox/issues/4208
- [ ] Add write token support to the delete end point- https://github.com/influxdata/influxdb_iox/issues/4209


Cargo.toml
```diff
@@ -78,6 +78,7 @@ members = [
     "trogging",
     "workspace-hack",
     "write_buffer",
+    "write_summary",
```
alamb, line 81: Added in new crate because:
1. `WriteSummary` will be shared between router2 and ingester
2. Has generated_types dependency (as it uses protobuf for encoding / decoding) so I didn't want to add it to `data_types` or `data_types2` for example

influxdb_iox/tests/end_to_end_ng_cases/all_in_one.rs
```diff
@@ -26,12 +26,14 @@ async fn smoke() {
     let response = write_to_router(lp, org, bucket, all_in_one.server().router_http_base()).await;
 
     assert_eq!(response.status(), StatusCode::NO_CONTENT);
+    let write_token = get_write_token(&response);
 
     // run query in a loop until the data becomes available
     let sql = format!("select * from {}", table_name);
-    let batches = query_until_results(
+    let batches = query_when_readable(
```
alamb, line 33: the vision is to be able to wait on the write token to become readable rather than running sql queries until you get more than one row back

ioxd_router2/src/lib.rs
```diff
@@ -248,7 +251,7 @@ pub async fn create_router2_server_type(
         .and_then(InstrumentationDecorator::new(
             "parallel_write",
             &*metrics,
-            FanOutAdaptor::new(write_buffer),
+            parallel_write,
```
alamb, line 254: here is where the new adapter is inserted int the DML handler chain

router2/src/dml_handlers/fan_out.rs
```diff
@@ -51,18 +51,17 @@ where
...
+        let results = input
             .into_iter()
             .map(|v| {
                 let namespace = namespace.clone();
                 let span_ctx = span_ctx.clone();
                 async move { self.inner.write(&namespace, v, span_ctx).await }
             })
             .collect::<FuturesUnordered<_>>()
-            .try_for_each(|_| future::ready(Ok(())))
+            .try_collect::<Vec<_>>()
```
alamb, line 62: here we change fanout to pass its input to output rather than drop them

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -162,30 +161,37 @@ where
...
-        .collect::<FuturesUnordered<_>>()
-        .filter_map(|v| {
-            if v.is_ok() {
-                successes += 1;
-            }
-            future::ready(v.err())
-        })
-        .collect::<Vec<WriteBufferError>>()
-        .await;
+    let mut successes = vec![];
```
alamb, line 170: This code changes the dml handler that actually sends to the write buffer so that it returns the resulting DMLMeta structure, rather than discarding it

router2/src/server/http.rs
```diff
@@ -274,10 +277,17 @@ where
             (&Method::POST, "/api/v2/delete") => self.delete_handler(req).await,
             _ => return Err(Error::NoHandler),
         }
-        .map(|_| response_no_content())
+        .map(|summary| {
+            println!("AAL todo add headers here: {:?}", summary);
+            Response::builder()
+                .status(StatusCode::NO_CONTENT)
+                .header(WRITE_TOKEN_HTTP_HEADER, summary.to_token())
```
alamb, line 283: Here is where the HTTP header is added

test_helpers_end_to_end_ng/src/client.rs
```diff
@@ -34,20 +34,41 @@ pub async fn write_to_router(
...
+pub async fn query_when_readable(
     sql: impl Into<String>,
     namespace: impl Into<String>,
+    write_token: impl Into<String>,
     connection: Connection,
 ) -> Vec<RecordBatch> {
     let namespace = namespace.into();
     let sql = sql.into();
 
+    println!(
```
alamb, line 67: The goal in a follow on PR will be to use this token to  wait for the data to become readable.

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -162,30 +161,37 @@ where
...
+    // Use FuturesUnordered so the futures can run in parallel
+    .collect::<FuturesUnordered<_>>()
+    .collect::<Vec<_>>()
+    .await
+    // Sort the result into successes/failures upon completion
+    .into_iter()
+    .for_each(|v| match v {
+        Ok(meta) => successes.push(meta),
+        Err(e) => errs.push(e),
+    });
```
domodwyer, line 187: Perhaps [`partition()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) is a good candidate here:

```suggestion
    let (successes, errs): (Vec<_>, Vec<_>) = v
        .map(|(sequencer, op)| async move {
            tokio::spawn(async move { sequencer.enqueue(op).await })
                .await
                .expect("shard enqueue panic")
        })
        // Use FuturesUnordered so the futures can run in parallel
        .collect::<FuturesUnordered<_>>()
        .collect::<Vec<_>>()
        .await
        // Sort the result into successes/failures upon completion
        .into_iter()
        .partition(Result::is_ok);
```

db/src/pred.rs
```diff
@@ -14,7 +14,7 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// Converts a [`predicate::predicate::Predicate`] into [`read_buffer::Predicate`],
+/// Converts a [`predicate::Predicate`] into [`read_buffer::Predicate`],
```
domodwyer, line 17: ❤️

router2/benches/e2e.rs
```diff
@@ -66,8 +67,9 @@ fn e2e_benchmarks(c: &mut Criterion) {
             parts: vec![TemplatePart::TimeFormat("%Y-%m-%d".to_owned())],
         });
 
-        let handler_stack =
-            schema_validator.and_then(partitioner.and_then(FanOutAdaptor::new(write_buffer)));
+        let handler_stack = schema_validator.and_then(
+            partitioner.and_then(WriteSummaryAdapter::new(FanOutAdaptor::new(write_buffer))),
```
domodwyer, line 71: 👍

router2/src/dml_handlers/write_summary.rs
```diff
@@ -0,0 +1,60 @@
...
+
+#[async_trait]
+impl<T> DmlHandler for WriteSummaryAdapter<T>
+where
+    T: DmlHandler<WriteOutput = Vec<Vec<DmlMeta>>>,
+{
+    type WriteInput = T::WriteInput;
+    type WriteOutput = WriteSummary;
+    type WriteError = T::WriteError;
+    type DeleteError = T::DeleteError;
```
domodwyer, line 34: Lovely 👌

router2/src/dml_handlers/instrumentation.rs
```diff
@@ -205,10 +206,14 @@ mod tests {
         );
     }
 
+    fn summary() -> WriteSummary {
+        WriteSummary::new(vec![])
```
domodwyer, line 210:
```suggestion
        WriteSummary::default()
```

Or could inline into the one use 👍

write_summary/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    pub fn to_token(self) -> String {
+        let proto_write_summary: proto::WriteSummary = self.into();
+        base64::encode(
+            serde_json::to_string(&proto_write_summary)
+                .expect("unexpected error serializing token to json"),
+        )
+    }
+}
+
+impl From<WriteSummary> for proto::WriteSummary {
```
domodwyer, line 38: What about a `From<proto::WriteSummary> for WriteSummary` and a test that asserts `original == From(From(original))` - or I guess that'll be while we're plumbing it into the ingester 👍

router2/src/server/http.rs
```diff
@@ -376,7 +386,8 @@ where
 
         self.delete_metric_body_size.inc(body.len() as _);
 
-        Ok(())
+        // TODO pass back write summaries for deletes as well?
```
domodwyer, line 389: I think this will be a good idea to know when "it has gone" - it will use all the same plumbing I imagine so shouldn't be difficult, but equally I'm sure no one would notice if it wasn't implemented either.

domodwyer (approved):
Looks spot on! Very nice 👍

router2/src/dml_handlers/sharded_write_buffer.rs
alamb: Thank you for the suggestion -- I actually tried `partition` in an earlier version of this code. The issue with using it is that then you end up with 2 `Vec<Result<WriteSummary>>` rather than a `Vec<WriteSummary>` and ` Vec<Error>`.

It is of course possible to then `unwrap` / `unwrap_err` each `Vec` , but I thought this way was a bit cleaner

db/src/pred.rs
alamb: drive by cleanup (showed up locally for me when I did `cargo doc ...` for some reason

write_summary/src/lib.rs
alamb: Yeah that was my ~excuse~ / thinking -- that I would add the code to go the other way when adding a `is_this_write_readable` to the ingester API.

router2/src/server/http.rs
alamb: I'll take a middle ground and file a ticket to track the work

alamb: Tracked in https://github.com/influxdata/influxdb_iox/issues/4209 (and I left a link as a comment)

PR #4203, opened 2022-04-01
refactor: handle errors in ingester stream handler

domodwyer:
This PR is aims to refactor the [`stream_in_sequenced_entries()`](https://github.com/influxdata/influxdb_iox/blob/d393ec6c257800508e33a6486c27d78b85b400fa/ingester/src/handler.rs#L273) fn that currently handles reading from kafka and applying the operation down through to the table buffer (and other things) with the intent being primarily to refine error handling and observability, and secondarily decouple various responsibilities within the ingest process to improve encapsulation / make testing easier.

This PR does NOT add the code within it to the ingest pipeline, instead the old handler remains in place - I've a little more to add to this refactor (specifically instrumentation, kafka watermark fetching, and a sink to connect the handler to the existing op processing impl) but felt this was more than big enough to review already.

---

* feat(tests): future timeout helper (ea3983e86)

      Adds a timeout test helper for futures - this lets us easily write tests that
      await on futures for a bounded duration of time.

      Optional feature to avoid dragging tokio into existing consumers of the 
      test_helpers crate that don't need it.

* refactor: ingest stream handler (e787e700d)

      Refactors the stream_in_sequenced_entries() into a new impl in the 
      SequencedStreamHandler type, decoupling the reading / decoding of ops from
      Kafka (and associated error handling) from the "what happens to those ops"
      concern to ease testing, encapsulate the specifics of "how to get an op" and
      improve flexibility.

      This is intended to provide robust error handling within what is reasonably
      possible (unexpected errors are always unexpected!) while retaining the
      existing metrics and functionality. I've also separated out code that exists
      in the current impl specifically to drive tests from the prod code path,
      instead driving those behaviours through mocks.

      As of this commit, the handler is not used - this commit simply adds the new
      impl.

* feat: expose ingest sequencer errors (0e8e46d9a)

      Instruments the SequencedStreamHandler with a series of new metrics that 
      record the various error classes observable in the stream handler.

      These metrics are labelled with potential_data_loss=true where relevant to
      surface potential data loss events for alerting & further review.

### DIFF
--- a/ingester/src/stream_handler/handler.rs
+++ b/ingester/src/stream_handler/handler.rs
@@ -1,0 +1,616 @@
+use std::{fmt::Debug, time::Duration};
+
+use data_types2::KafkaPartition;
+use dml::DmlOperation;
+use futures::{pin_mut, FutureExt, Stream, StreamExt};
+use metric::{Attributes, U64Counter, U64Gauge};
+use observability_deps::tracing::*;
+use time::{SystemProvider, TimeProvider};
+use tokio_util::sync::CancellationToken;
+use write_buffer::core::{WriteBufferError, WriteBufferErrorKind};
+
+use crate::lifecycle::LifecycleHandle;
+
+use super::DmlSink;
+
+/// When the [`LifecycleManager`] indicates that ingest should be paused because
+/// of memory pressure, the sequencer will loop, sleeping this long between
+/// calls to [`LifecycleHandle::can_resume_ingest()`] with the manager if it
+/// can resume ingest.
+///
+/// [`LifecycleManager`]: crate::lifecycle::LifecycleManager
+/// [`LifecycleHandle::can_resume_ingest()`]: crate::lifecycle::LifecycleHandle::can_resume_ingest()
+const INGEST_POLL_INTERVAL: Duration = Duration::from_millis(100);
+
+/// A [`SequencedStreamHandler`] consumes a sequence of [`DmlOperation`] from a
+/// sequencer stream and pushes them into the configured [`DmlSink`].
+///
+/// Ingest reads are rate limited by the [`LifecycleManager`] it is initialised
+/// by, pausing until the [`LifecycleHandle::can_resume_ingest()`] obtained from
+/// it returns true, and TTBR / error metrics are emitted on a per-sequencer
+/// basis.
+///
+/// [`LifecycleManager`]: crate::lifecycle::LifecycleManager
+/// [`LifecycleHandle::can_resume_ingest()`]: crate::lifecycle::LifecycleHandle::can_resume_ingest()
+#[derive(Debug)]
+pub struct SequencedStreamHandler<I, O, T = SystemProvider> {
+    /// A input stream of DML ops
+    stream: I,
+    /// An output sink that processes DML operations and applies them to
+    /// in-memory state.
+    sink: O,
+
+    /// A handle to the [`LifecycleManager`] singleton that may periodically
+    /// request ingest be paused to control memory pressure.
+    ///
+    /// [`LifecycleManager`]: crate::lifecycle::LifecycleManager
+    lifecycle_handle: LifecycleHandle,
+
+    // Metrics
+    time_provider: T,
+    time_to_be_readable_ms: U64Gauge,
+    /// Duration of time ingest is paused at the request of the LifecycleManager
+    pause_duration_ms: U64Counter,
+
+    /// Log context fields - otherwise unused.
+    kafka_topic_name: String,
+    kafka_partition: KafkaPartition,
+}
+
+impl<I, O> SequencedStreamHandler<I, O> {
+    /// Initialise a new [`SequencedStreamHandler`], consuming from `stream` and
+    /// dispatching successfully decoded [`DmlOperation`] instances to `sink`.
+    ///
+    /// A [`SequencedStreamHandler`] starts actively consuming items from
+    /// `stream` once [`SequencedStreamHandler::run()`] is called, and
+    /// gracefully stops when `shutdown` is cancelled.
+    pub fn new(
+        stream: I,
+        sink: O,
+        lifecycle_handle: LifecycleHandle,
+        kafka_topic_name: String,
+        kafka_partition: KafkaPartition,
+        metrics: &metric::Registry,
+    ) -> Self {
+        let attributes = Attributes::from([
+            ("sequencer_id", kafka_partition.to_string().into()),
+            ("kafka_topic", kafka_topic_name.clone().into()),
+        ]);
+        let time_to_be_readable_ms = metrics.register_metric::<U64Gauge>(
+            "ingester_ttbr_ms",
+            "duration of time between producer writing to consumer putting into queryable cache in milliseconds",
+        ).recorder(attributes.clone());
+
+        let pause_duration_ms = metrics.register_metric::<U64Counter>(
+            "ingest_paused_duration_ms_total",
+            "duration of time ingestion has been paused by the lifecycle manager in milliseconds",
+        ).recorder(&[]);
+
+        Self {
+            stream,
+            sink,
+            lifecycle_handle,
+            time_provider: SystemProvider::default(),
+            time_to_be_readable_ms,
+            pause_duration_ms,
+            kafka_topic_name,
+            kafka_partition,
+        }
+    }
+
+    /// Switch to the specified [`TimeProvider`] implementation.
+    #[cfg(test)]
+    pub(crate) fn with_time_provider<T>(self, provider: T) -> SequencedStreamHandler<I, O, T> {
+        SequencedStreamHandler {
+            stream: self.stream,
+            sink: self.sink,
+            lifecycle_handle: self.lifecycle_handle,
+            time_provider: provider,
+            time_to_be_readable_ms: self.time_to_be_readable_ms,
+            pause_duration_ms: self.pause_duration_ms,
+            kafka_topic_name: self.kafka_topic_name,
+            kafka_partition: self.kafka_partition,
+        }
+    }
+}
+
+impl<I, O, T> SequencedStreamHandler<I, O, T>
+where
+    I: Stream<Item = Result<DmlOperation, WriteBufferError>> + Unpin + Send + Sync,
+    O: DmlSink,
+    T: TimeProvider,
+{
+    /// Run the stream handler, consuming items from [`Stream`] and applying
+    /// them to the [`DmlSink`].
+    ///
+    /// This method blocks until gracefully shutdown by cancelling the
+    /// `shutdown` [`CancellationToken`]. Once cancelled, this handler will
+    /// complete the current operation it is
+    ///
+    /// # Panics
+    ///
+    /// This method panics if the input stream ends (yields a `None`).
+    pub async fn run(mut self, shutdown: CancellationToken) {
+        let shutdown_fut = shutdown.cancelled().fuse();
+        pin_mut!(shutdown_fut);
+
+        loop {
+            // Wait for a DML operation from the sequencer, or a graceful stop
+            // signal.
+            let maybe_op = futures::select!(
+                next = self.stream.next().fuse() => next,
+                _ = shutdown_fut => {
+                    info!(
+                        kafka_topic=%self.kafka_topic_name,
+                        kafka_partition=%self.kafka_partition,
+                        "stream handler shutdown",
+                    );
+                    return;
+                }
+            );
+
+            // Read a DML op from the write buffer, logging and emitting metrics
+            // for any potential errors to enable alerting on potential data
+            // loss.
+            //
+            // If this evaluation results in no viable DML op to apply to the
+            // DmlSink, return None rather than continuing the loop to ensure
+            // ingest pauses are respected.
+            let maybe_op = match maybe_op {
+                Some(Ok(op)) => Some(op),
+                Some(Err(e)) if e.kind() == WriteBufferErrorKind::UnknownSequenceNumber => {
+                    error!(
+                        error=%e,
+                        kafka_topic=%self.kafka_topic_name,
+                        kafka_partition=%self.kafka_partition,
+                        potential_data_loss=true,
+                        "unable to read from desired sequencer offset"
+                    );
+                    // TODO: emit metric
+                    None
+                }
+                Some(Err(e)) if e.kind() == WriteBufferErrorKind::IO => {
+                    warn!(
+                        error=%e,
+                        kafka_topic=%self.kafka_topic_name,
+                        kafka_partition=%self.kafka_partition,
+                        "I/O error reading from sequencer"
+                    );
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                    None
+                }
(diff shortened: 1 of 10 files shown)


ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+            &kafka_topic_name,
+            Some("sequencer_unknown_error"),
+            true,
+        ));
+        let sink_apply_error_count = ingest_errors.recorder(metric_attrs(
+            kafka_partition,
+            &kafka_topic_name,
+            Some("sink_apply_error"),
+            true,
+        ));
```
domodwyer, line 120: These new metrics record error occurrences in the stream reader, and clearly label possible data loss events with a metric label `potential_data_loss=true` so we can easily monitor their frequency, alert on them & investigate.

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+                    error!(
+                        error=%e,
+                        kafka_topic=%self.kafka_topic_name,
+                        kafka_partition=%self.kafka_partition,
+                        potential_data_loss=true,
+                        "unable to read from desired sequencer offset"
+                    );
+                    self.seq_unknown_sequence_number_count.inc(1);
+                    None
+                }
```
domodwyer, line 212: Here and below all (useful) write buffer / sequencer errors are enumerated and logged (both with a field `potential_data_loss=true`) to help surface them.

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+                    );
+                    self.seq_unknown_error_count.inc(1);
+                    tokio::time::sleep(Duration::from_secs(1)).await;
+                    None
+                }
+                None => {
+                    panic!(
+                        "sequencer {:?} stream for topic {} ended without graceful shutdown",
+                        self.kafka_partition, self.kafka_topic_name
+                    );
```
domodwyer, line 256: In the current impl a stream ending abnormally silently ends the reader task - it is possible this is intended if it happens during graceful shutdown, in which case I will replace this with a `WARN` but want to see if it is hit first.

If it is not something that can happen during shutdown, the ingester should be bounced to resume ingesting from the affected sequencer and a panic makes sense to me to drive this.

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+            if let Some(delta) =
+                produced_at.and_then(|ts| self.time_provider.now().checked_duration_since(ts))
+            {
+                self.time_to_be_readable_ms.set(delta.as_millis() as u64);
+            }
+
+            if should_pause {
+                // The lifecycle manager may temporarily pause ingest - wait for
+                // persist operations to shed memory pressure if needed.
+                self.pause_ingest().await;
```
domodwyer, line 304: Currently this is only triggered once an op is successfully applied as this is what the current impl does - I intend to refactor this to move the pause to include the error path.

ingester/src/stream_handler/sink.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use async_trait::async_trait;
+use dml::DmlOperation;
+
+/// A [`DmlSink`] handles [`DmlOperation`] instances read from a sequencer.
+#[async_trait]
+pub trait DmlSink: Debug + Send + Sync {
+    /// Apply `op` read from a sequencer, returning `Ok(true)` if ingest should
+    /// be paused.
+    async fn apply(&self, op: DmlOperation) -> Result<bool, crate::data::Error>;
```
domodwyer, line 11: I intend to remove the returned bool, fully decoupling the ingest from op application and instead rely on the ingesters to stop automatically after refactoring how this pause signal is propagated.

test_helpers/Cargo.toml
```diff
@@ -12,3 +12,9 @@ tracing-log = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 observability_deps = { path = "../observability_deps" }
 workspace-hack = { path = "../workspace-hack"}
+async-trait = { version = "0.1.53", optional = true }
+tokio = { version = "1.17.0", optional = true, default_features = false, features = ["time"] }
```
domodwyer, line 16: `tokio` is not dragged into `test_helpers` unless it is opted into by enabling the `future_timeout` feature flag.

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+                    for op in $stream_ops {
+                        tx.send(op)
+                            .with_timeout_panic(Duration::from_secs(5))
+                            .await
+                            .expect("early handler exit");
+                    }
+                    // Wait for the handler to read the last op, restoring the
+                    // capacity to 1.
+                    let _permit = tx.reserve()
+                        .with_timeout_panic(Duration::from_secs(5))
```
domodwyer, line 488: Here is an example of using the `FutureTimeout` extension trait to bound the duration of an await for testing purposes so it doesn't block forever - it is used throughout this test.

ingester/src/stream_handler/handler.rs
alamb: makes sense to me

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+                    self.sink_apply_error_count.inc(1);
+                    return;
+                }
+            };
+
+            // Update the TTBR metric before potentially sleeping.
+            if let Some(delta) =
+                produced_at.and_then(|ts| self.time_provider.now().checked_duration_since(ts))
+            {
+                self.time_to_be_readable_ms.set(delta.as_millis() as u64);
```
alamb, line 298: 👍  for ttbr calculation

ingester/src/stream_handler/handler.rs
alamb: It is not 100% clear to me that errors should pause ingestion (as they didn't add to the memory pressure). That being said I don't think it is a problem if they do either

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+    use tokio_stream::wrappers::ReceiverStream;
+
+    lazy_static::lazy_static! {
+        static ref TEST_TIME: Time = SystemProvider::default().now();
+        static ref TEST_KAFKA_PARTITION: KafkaPartition = KafkaPartition::new(42);
+    }
+    static TEST_KAFKA_TOPIC: &str = "kafka_topic_name";
+
+    // Return a DmlWrite with the given namespace and a single table.
+    fn make_write(name: impl Into<String>, ttbr: u64) -> DmlWrite {
```
alamb, line 397:
```suggestion
    fn make_write(name: impl Into<String>, write_time: u64) -> DmlWrite {
```

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+            Sequence::new(1, 2),
+            TEST_TIME.checked_sub(Duration::from_millis(ttbr)).unwrap(),
+            None,
+            42,
+        );
+        DmlWrite::new(name, tables, sequence)
+    }
+
+    // Return a DmlDelete with the given namespace.
+    fn make_delete(name: impl Into<String>, ttbr: u64) -> DmlDelete {
```
alamb, line 409:
```suggestion
    fn make_delete(name: impl Into<String>, write_time: u64) -> DmlDelete {
```

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+        );
+        DmlDelete::new(name, pred, None, sequence)
+    }
+
+    // Generates a test that ensures that the handler given $stream_ops makes
+    // $want_sink calls.
+    //
+    // Additionally all test cases assert the handler does not panic, and the
+    // handler gracefully shuts down after the test input sequence is exhausted.
+    macro_rules! test_stream_handler {
```
alamb, line 428: I wonder if it would be possible to use an actual function (rather than macro) here ? I have found macros harder to debug and change -- if it is due to type shenanigans, perhaps it is worth considering dynamic dispatch to avoid generics 🤔

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+    );
+    test_stream_handler!(
+        non_fatal_stream_invalid_data,
+        stream_ops = [
+            Err(WriteBufferError::new(WriteBufferErrorKind::InvalidData, "explosions")),
+            Ok(DmlOperation::Write(make_write("bananas", 50)))
+        ],
+        sink_rets = [Ok(true)],
+        want_ttbr = 50,
+        want_err_metrics = [
```
alamb, line 620: 💯  for testing all the possible error metrics / paths

alamb (approved):
Makes sense to me -- this structure is very much in the spirit of the router's DML handler, which makes sense to me.  I also found the code very well commented and a pleasure to read.

The error handling certainly looks 👨‍🍳 ❤️ 

I found the DML Handler  easy to read, but relatively harder to change, which may not be a bad tradoff. 

cc @pauldix I think this has the potential to change the code structure of the ingester significantly

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+    I: Stream<Item = Result<DmlOperation, WriteBufferError>> + Unpin + Send + Sync,
+    O: DmlSink,
+    T: TimeProvider,
+{
+    /// Run the stream handler, consuming items from [`Stream`] and applying
+    /// them to the [`DmlSink`].
+    ///
+    /// This method blocks until gracefully shutdown by cancelling the
+    /// `shutdown` [`CancellationToken`]. Once cancelled, this handler will
+    /// complete the current operation it is
```
pauldix, line 169: it is...?

ingester/src/stream_handler/handler.rs
pauldix: I agree with this, I've found the macro test generation stuff to be hard to follow when things fail.

pauldix (approved):
Looks like it'll be a good improvement

ingester/src/stream_handler/handler.rs
domodwyer: Yeah this was more to highlight a change I plan on making but poorly worded - I will be moving this pause into the main loop and eliminating the op having to tell us the current state when it's applied 👍

ingester/src/stream_handler/handler.rs
```diff
@@ -0,0 +1,741 @@
...
+            sink_apply_error_count: self.sink_apply_error_count,
+            kafka_topic_name: self.kafka_topic_name,
+            kafka_partition: self.kafka_partition,
+        }
+    }
+}
+
+impl<I, O, T> SequencedStreamHandler<I, O, T>
+where
+    I: Stream<Item = Result<DmlOperation, WriteBufferError>> + Unpin + Send + Sync,
```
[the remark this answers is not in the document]
domodwyer: Well rightly or wrongly, I guess it's mostly personal preference if I'm honest - I don't think we should just default to `Arc dyn`-ing everything, it imposes some restrictions on the traits and implementations and I feel like the rust language is more geared towards generics with `dyn` being almost an after thought at times, though it is definitely getting better and obviously has its use cases with heterogeneous collections or runtime impls. As an example, this generic bound allows you to pass a `Box<dyn Stream...>` if you want.

I'm certainly not making the argument for performance either, but something that is often overlooked in the "it's just a cheap dynamic dispatch" argument is that while at runtime that is true, at compile time it effectively shoots the optimiser in the foot as it cannot infer/assume anything about what is being called - a good example is the delete path in the router where most of the `DmlHandler`'s do not touch the delete op and instead immediately return `Ok(())` - I would expect these useless function calls (only there to satisfy the trait) to be optimised away to nothing because it's not `dyn` and the compiler knows they do nothing.

Obviously readability & maintainability is more important, so I try to balance the two the best I can 👍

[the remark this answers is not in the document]
domodwyer: Also sorry for throwing a book of text at you, but I think this might genuinely help as you called it out explicitly above 📖 📖 📖 

Currently in the code base we explicitly pass the `Arc<dyn TimeProvider>` around everywhere in all the type constructors that use it, and therefore plumb it through all the call chains to get the provider to where it is needed but **we only ever use the one impl `SystemProvider` outside of tests** .

The ONLY other impl is the `MockProvider` for testing. You might notice the `SequencedStreamHandler` DOES NOT need the time provider plumbing into [the constructor](https://github.com/influxdata/influxdb_iox/blob/0e8e46d9aac9b4788a130d25700bcc89081137b2/ingester/src/stream_handler/handler.rs#L126) - it is generic over a `TimeProvider` type but specifies the default type for it [like we do here](https://github.com/influxdata/influxdb_iox/blob/0e8e46d9aac9b4788a130d25700bcc89081137b2/ingester/src/stream_handler/handler.rs#L36):

```rust
pub struct SequencedStreamHandler<I, O, T = SystemProvider> {
    time_provider: T,
    // ...
}
```

Now because there's a default type specified, we never have to mention it! It's effectively free with no cognitive overhead. When we're specifying the `SequencedStreamHandler` type we can write:

```rust
SequencedStreamHandler<I, O> // no time provider!
```

And the beauty is we never have to pass the time provider around the code or push it through call chains. By default, it just uses the system clock and in tests we can  [explicitly provide](https://github.com/influxdata/influxdb_iox/blob/0e8e46d9aac9b4788a130d25700bcc89081137b2/ingester/src/stream_handler/handler.rs#L145) a `TimeProvider`. You may like this approach to help tidy up new code, but again it's personal preference - up to you! 👍

ingester/src/stream_handler/handler.rs
domodwyer: The easier it is to write test cases, the more test cases people will write! 🎉

ingester/src/stream_handler/handler.rs
domodwyer: doh

ingester/src/stream_handler/handler.rs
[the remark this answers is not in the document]
alamb: > Well rightly or wrongly, I guess it's mostly personal preference if I'm honest 

Yes I agree this is best summarized as personal preference (and indeed what is more "readable" is definitely in the eye of the beholder). 

> it is generic over a TimeProvider type but specifies the default type for it 

This is a  good example -- if fact one (aka yours truly) would argue that the code with the explicit `Arc<dyn TimeProvider>` is actually  easier to understand as it is explicitly (though somewhat verbosely) plumbed through the structures where it is needed. It also may make tests that need to share the same `MockTimeProvider` instance easier to write.

My plan will be to follow the approach taken with the existing code when modifying things 👍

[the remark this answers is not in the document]
alamb: To be 100 % clear -- I think the generics in this PR are just fine and get the job done well. While I likely would write it differently, I will say no more and adapt to the existing conventions when working on this code 👍

domodwyer:
Rebased onto master - no functional changes in existing commits 👍

PR #4204, opened 2022-04-01
feat: replace a compactor constant with an CLI config param

NGA-TRAN:
A follow-up of #4197 


compactor/src/compact.rs
```diff
@@ -1387,7 +1395,7 @@ mod tests {
             Arc::new(Executor::new(1)),
             Arc::new(SystemProvider::new()),
             BackoffConfig::default(),
-            CompactorConfig::new(90, 10000000),
+            CompactorConfig::new(90, 100000, 100000),
```
NGA-TRAN, line 1398: Make the params 100KB for the tests. No need 100MB

clap_blocks/src/compactor.rs
```diff
@@ -43,4 +44,15 @@ pub struct CompactorConfig {
...
+    /// The compactor will compact overlapped files no matter how much large they are.
+    /// For non-overlapped and contiguous files, compactor will also compact them into
+    /// a larger file of max size defined by the config value.
+    /// Default is 100,000,000 (100MB).
+    #[clap(
+        long = "--compaction-max-size-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_MAX_SIZE_BYTES",
+        default_value = "100000000"
+    )]
+    pub compaction_max_size_bytes: i64,
```
alamb, line 57:
```suggestion
    pub compaction_max_size_bytes: u64,
```

compactor/src/handler.rs
```diff
@@ -107,16 +107,25 @@ pub struct CompactorConfig {
     /// number should be less than 1/10th of the available memory to ensure compactions have
     /// enough space to run.
     max_concurrent_compaction_size_bytes: i64,
+    /// The compactor will compact overlapped files no matter how much large they are.
+    /// For non-overlapped and contiguous files, compactor will also compact them into
+    /// a larger file of max size defined by this config value.
+    compaction_max_size_bytes: i64,
```
alamb, line 113:
```suggestion
    compaction_max_size_bytes: u64,
```

alamb (approved):

clap_blocks/src/compactor.rs
NGA-TRAN: @alamb I have to keep this `i64` to make it consistent with other config params we have created that fall into the same category in which postgres return i64 (instead of possible u64) and we need to compare this value with them. `max_concurrent_compaction_size_bytes` is other example.

alamb: I just figured that a negative number never made sense -- so we could use the rust type system / parser to ensure no one ever provided one

PR #4206, opened 2022-04-01
feat: Add optional sort_key column to partition table in the catalog

### LINKED ISSUE #4195: Store a partition's sort key in the catalog
- Add a nullable column to the partition table in the catalog to store the sort key computed by the ingester. 
- Add a method to the catalog to get the sort key for a partition
- Add a method to update the sort key for a partition

Part of #3968

carols10cents:
Closes #4195.

On creation of a partition, does not set a sort key. There's a function for updating the sort key. If the sort key has a value, it will be included when a partition is requested from the catalog.

The migration adds a new, nullable column so it should not affect existing data.

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -45,6 +45,9 @@ pub enum Error {
     #[snafu(display("table {} not found", id))]
     TableNotFound { id: TableId },
 
+    #[snafu(display("partition {} not found", id))]
+    PartitionNotFound { id: PartitionId },
+
     #[snafu(display(
         "couldn't create column {} in table {}; limit reached on namespace",
         column_name,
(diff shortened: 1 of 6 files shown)


iox_catalog/src/interface.rs
```diff
@@ -1316,9 +1327,50 @@ pub(crate) mod test_helpers {
...
+        let updated_other_partition = repos
+            .partitions()
+            .get_by_id(other_partition.id)
+            .await
+            .unwrap()
+            .unwrap();
+        assert_eq!(
+            updated_other_partition.sort_key.unwrap(),
+            "tag2,tag1,tag3,time"
+        );
```
NGA-TRAN, line 1373: 👍

NGA-TRAN (approved):
LGTM

PR #4207, opened 2022-04-01
chore: add compactor logging

pauldix:


pauldix:
going to push this in so I can see it deployed

PR #4183, opened 2022-03-30
refactor: extract router2 into ioxd_router2

alamb:
Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/4182

# Rationale
re https://github.com/influxdata/influxdb_iox/issues/3926
Follow on to https://github.com/influxdata/influxdb_iox/pull/4149, https://github.com/influxdata/influxdb_iox/pull/4182

Continues the quest to make `ioxd` less monolithic (and thus compile faster)

# Changes
Extracts code from `ioxd`:
1. `ioxd_router2`

Removes unneeded dependencies in `influxdb_ioxd`

PR #4210, opened 2022-04-02
refactor: Extract `ioxd_test`, `ioxd_compactor`, `ioxd_ingester`; remove `ioxd`

alamb:
Completes split of `ioxd` into smaller parts

# Rationale
Each of these modules ends up with a non trivial amount of code due to the generics and macros used for `tonic::service` so recompiling them all when one of the individual service implementations is changed, is annoying

Follow on https://github.com/influxdata/influxdb_iox/pull/4183

# changes
1. Extract `ioxd_test`, `ioxd_compactor`, `ioxd_ingester` from `ioxd`, 
2. remove `ioxd` crate


mkmik (approved):
LHTm

PR #4211, opened 2022-04-02
fix: compactor use join_all

pauldix:
I forgot to address this in #4139. Have the compactor use join and make sure the error gets logged.


pauldix:
going to YOLO this through to see things in prod

PR #4226, opened 2022-04-04
feat: applying tombstones to cold partitions and compact them

### LINKED ISSUE #4205: NG Compactor: Apply tombstones on cold partitions
# Rationale
 Compactor prioritizes to compact hot partitions which are the ones with recent write files (aka level-0 files). This leads to the tombstones on cold partitions (aka the ones with only level-1 files) never applied. This ticket is to have tombstones apply to cold partition and then remove them if possible.

NGA-TRAN:
Closes #4205 

The main changes of this PR:
1. Create function `partitions_with_files_to_compact` to find tombstones and their overlapped files. They are then grouped into partitions.
2. Modify `compact_partition` function to either
    -  Compact a given hot partition (existing)
    - Compact given files with their tombstones (new). The given files are gotten from the step 1.
    
Due to the changes in number 2, there are a few refactors and a new trait. 

The plan for the following PR is to create a scheduler to invoke `partitions_with_files_to_compact` and then spawn many Tokio to run `compact_partition` on files of each partition.

@pauldix: let me know what you think about this plan. I also have a few questions in the code.
@carols10cents: Your rust input is highly appreciated 


compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
         Ok(candidates)
     }
 
+    /// Get files that need to apply the available tombstones
+    /// Return a list of partitions with tombstones-attached parquet files to be compacted
+    pub async fn partitions_with_files_to_compact(&self) -> Result<Vec<PartitionWithParquetFiles>> {
```
NGA-TRAN, line 371: This functions find candidate `cold` partitions which are partitions that files with overlapped tombstones

compactor/src/compact.rs
```diff
@@ -824,12 +941,13 @@ impl Compactor {
             // the group. If not, add it to the non-overlapping group.
             for file in parquet_files {
                 if in_group.iter().any(|group_file| {
-                    (file.min_time <= group_file.min_time && file.max_time >= group_file.min_time)
-                        || (file.min_time > group_file.min_time
-                            && file.min_time <= group_file.max_time)
+                    (file.min_time() <= group_file.min_time()
+                        && file.max_time() >= group_file.min_time())
+                        || (file.min_time() > group_file.min_time()
+                            && file.min_time() <= group_file.max_time())
```
NGA-TRAN, line 947: The reason to to invoke function `min_time()` rather than its member `min_time` because `file` and `group_file` are now traits

iox_catalog/src/interface.rs
```diff
@@ -1323,34 +1331,32 @@ pub(crate) mod test_helpers {
...
             )
             .await
             .unwrap();
         assert!(t1.id > TombstoneId::new(0));
         assert_eq!(t1.sequencer_id, sequencer.id);
         assert_eq!(t1.sequence_number, SequenceNumber::new(1));
-        assert_eq!(t1.min_time, min_time);
-        assert_eq!(t1.max_time, max_time);
+        assert_eq!(t1.min_time, Timestamp::new(1));
+        assert_eq!(t1.max_time, Timestamp::new(10));
```
NGA-TRAN, line 1350: Replace with values  right here to make it easier to see it overlaps with which files.

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+                for ts in tombstones {
+                    // Read parquet files that are not soft deleted, not tombstone-processed, same table and
+                    // overlaps with this tombstone
+                    // TODO - QUESTIONS:
+                    //    1. Should we only read level-1 files only (right now, we read all files that overlap with the tombstone)?
+                    //    2. Need to handle the case that one or many of these files are also included in other background cycle
+                    //       that compact hot partitions.
+                    let parquet_files = repos
+                        .parquet_files()
+                        .list_by_tombstone(ts.clone())
```
carols10cents, line 409: Looking at the implementations of `list_by_tombstone`, they don't look like they need to own the `Tombstone` (which doesn't implement `Copy` like a lot of the other interface types do, because it holds a serialized predicate `String`)

I think if `list_by_tombstone` takes `&Tombstone` instead, then this `clone` can be removed--

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+
+                    // Attach this tombstone to all parquet files and add it to the parquet_file_with_tombstones if not yet
+                    for pf in parquet_files {
+                        let file = parquet_file_with_tombstones
+                            .entry(pf.id)
+                            .or_insert_with(|| ParquetFileWithTombstone {
+                                data: Arc::new(pf),
+                                tombstones: vec![],
+                            });
+                        file.add_tombstones(vec![ts.clone()]);
```
carols10cents, line 421: -- and then this clone can be removed too, because `ts` can be moved into `add_tombstones` here, right?

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+                        |(partition_id, files_with_tombstones)| PartitionWithParquetFiles {
+                            sequencer_id: *sequencer_id,
+                            table_id,
+                            partition_id,
+                            files_with_tombstones,
+                        },
+                    )
+                    .collect::<Vec<_>>();
+
+                partitions.extend(ps);
```
carols10cents, line 449: `Vec` implements [the `Extend` trait](https://doc.rust-lang.org/core/iter/trait.Extend.html), so it can be extended with an iterator, and we can save some time by not collecting into an intermediate vector here:

```suggestion
                    );

                partitions.extend(ps);
```

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+    }
+
     /// Runs compaction in a partition resolving any tombstones and compacting data so that parquet
     /// files will be non-overlapping in time.
+    /// If no files are provided, the files will be gotten from the catalog
     pub async fn compact_partition(
         &self,
         partition_id: PartitionId,
         compaction_max_size_bytes: i64,
+        files: Option<Vec<Arc<dyn FileMeta>>>,
```
carols10cents, line 463: `Option<Vec<_>>` is kind of a rare type to see-- this might be what you want though, I'm not sure. It depends on what you want to happen if this function gets `Some(vec![])` - should it fetch files from the catalog? Right now it won't, it'll hit the `Some` case and then the early `is_empty` return. 🤷🏻‍♀️  

If providing an empty list of files *should* cause a fetch from the catalog, then the `Option` can be removed here and the `match` can be changed to instead check if the vector is empty or not (the existing `is_empty` check will still need to be there to make sure we don't do anything if no files were provided *and* the catalog doesn't return any files).

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+                    .repositories()
+                    .await
+                    .parquet_files()
+                    .list_by_partition_not_to_delete(partition_id)
+                    .await
+                    .context(ListParquetFilesSnafu)?;
+
+                let files = files
+                    .iter()
+                    .map(|f| Arc::clone(&(*f)) as Arc<dyn FileMeta>)
```
carols10cents, line 484: This can be a little simpler, `f` is already `&Arc`:

```suggestion
                    .map(|f| Arc::clone(f) as Arc<dyn FileMeta>)
```

I wish there was a way to coerce this whole `Vec` at once, but I don't know of a way :(

compactor/src/compact.rs
```diff
@@ -949,69 +1067,84 @@ impl Compactor {
...
-                        })
-                        .collect();
-
-                    ParquetFileWithTombstone {
-                        data: Arc::new(data),
-                        tombstones: relevant_tombstones,
-                    }
-                })
-                .collect();
+                        .map(|f| f.parquet_files_with_timbstone())
```
carols10cents, line 1142: Nit, typo in method name here:

```suggestion
                        .map(|f| f.parquet_files_with_tombstone())
```

compactor/src/compact.rs
```diff
@@ -1386,6 +1663,266 @@ mod tests {
...
+            tombstones: vec![ts4.tombstone.clone()],
+        });
+        assert!(files.contains(&pfwts2));
+        assert!(files.contains(&pfwts3));
+        assert!(files.contains(&pfwts4));
+
+        // cast to FileMeta
+        let files = files
+            .iter()
+            .map(|f| Arc::clone(&(*f)) as Arc<dyn FileMeta>)
```
carols10cents, line 1848: This can be a bit simpler:

```suggestion
            .map(|f| Arc::clone(f) as Arc<dyn FileMeta>)
```

compactor/src/query.rs
```diff
@@ -89,6 +94,44 @@ impl QueryableParquetChunk {
     pub fn max_time(&self) -> i64 {
         self.iox_metadata.time_of_last_write.timestamp_nanos()
     }
+
+    /// Convert to a QueryableParquetChunk
+    pub fn from_parquet_file_with_tombstones(
+        object_store: Arc<DynObjectStore>,
+        table_name: String,
+        parquet_file_with_tobstones: &ParquetFileWithTombstone,
```
carols10cents, line 102: Nit: typo in parameter name through here:

```suggestion
        parquet_file_with_tombstones: &ParquetFileWithTombstone,
```

compactor/src/utils.rs
```diff
@@ -47,75 +65,30 @@ pub struct GroupWithMinTimeAndSize {
...
+        {
+            return false;
+        }
 
-    /// Return iox metadata of the parquet file
-    pub fn iox_metadata(&self) -> IoxMetadata {
-        let decoded_parquet_file = DecodedParquetFile::new((*self.data).clone());
-        decoded_parquet_file.iox_metadata
+        for f in &other.parquet_files {
+            if !self.contains(f) {
```
carols10cents, line 87: So [`PartialEq`](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) has particular guarantees implementations are expected to uphold, namely that "equal" and "not equal" are consistent with each other... I'm sure this would be rare, but I don't see anything in the type itself preventing one `GroupWithMinTimeAndSize` from having duplicate items in its `parquet_files`, which could lead to `self` and `other`'s lengths being equal but not having the same items.

Because this is only used for tests, I think it would be better to make a little helper function in the tests that use this that would check the min time and sizes, then extract out a vec of parquet file IDs from both self and other, then sort and compare those vecs directly.

carols10cents (changes requested):
A few suggestions and questions!

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+                let mut parquet_file_with_tombstones: BTreeMap<
+                    ParquetFileId,
+                    ParquetFileWithTombstone,
+                > = BTreeMap::new();
+                for ts in tombstones {
+                    // Read parquet files that are not soft deleted, not tombstone-processed, same table and
+                    // overlaps with this tombstone
+                    // TODO - QUESTIONS:
+                    //    1. Should we only read level-1 files only (right now, we read all files that overlap with the tombstone)?
+                    //    2. Need to handle the case that one or many of these files are also included in other background cycle
```
pauldix, line 405: I think the same scheduler that runs the other compactions should do this. So it's something like:
1. Look for level 0 compactions, perform them
2. Look for level 1 tombstone compactions

We could have it perform one level 0, then a level 1, then level 0, etc.

compactor/src/compact.rs
```diff
@@ -356,28 +366,132 @@ impl Compactor {
...
+            for (table_id, tombstones) in table_tombstones {
+                let mut parquet_file_with_tombstones: BTreeMap<
+                    ParquetFileId,
+                    ParquetFileWithTombstone,
+                > = BTreeMap::new();
+                for ts in tombstones {
+                    // Read parquet files that are not soft deleted, not tombstone-processed, same table and
+                    // overlaps with this tombstone
+                    // TODO - QUESTIONS:
+                    //    1. Should we only read level-1 files only (right now, we read all files that overlap with the tombstone)?
```
pauldix, line 404: I think if we have the level 0 being handled elsewhere, we should limit it to partitions with only level 1 files. See my next comment on the L0 vs. L1.

compactor/src/compact.rs
pauldix: This construction feels odd to me. I think it would make more sense to split into two functions. Right now `compact_partition` looks up the files and then goes through the compaction process. You could change it so that there is a function `compact_files` which takes the files and move all the compaction code into there. Then keep the function signature for `compact_partition` the same, but have it do the file lookup and call the new `compact_files` function.

compactor/src/compact.rs
pauldix: Why use traits for a data object? I think it makes more sense to have simple data objects, which can be made in testing scenarios and use traits only for things that have some sort of behavior or remote connection you want to mock out.

data_types2/src/lib.rs
```diff
@@ -719,6 +720,40 @@ pub fn tombstones_to_delete_predicates_iter(
     })
 }
 
+/// Collection of funtions a ParquetFile or its wrapper can provide
+#[async_trait]
+pub trait FileMeta: Debug + Sync + Send + 'static {
```
pauldix, line 725: Not sure why this is a trait as it seems like it should just be a plain data object.

iox_catalog/src/postgres.rs
```diff
@@ -1614,11 +1628,40 @@ WHERE parquet_file.sequencer_id = $1
...
+  AND parquet_file.table_id = $2
+  AND parquet_file.max_sequence_number < $3
+  AND parquet_file.to_delete IS NULL
+  AND ((parquet_file.min_time <= $4 AND parquet_file.max_time >= $4)
+      OR (parquet_file.min_time > $4 AND parquet_file.min_time <= $5))
+  AND id NOT IN (
+      SELECT parquet_file_id
+      FROM processed_tombstone
+      WHERE tombstone_id = $6
+  );
```
pauldix, line 1646: I wonder how expensive this query is going to be 😬

pauldix (commented):
Had a look through, I think it's worth having a quick discussion tomorrow morning.

alamb:

PR #4229, opened 2022-04-04
chore: update datafusion

alamb:
# Rationale
* Make it easy to use IOx's tests for some changes like https://github.com/apache/arrow-datafusion/pull/2132
* Also get fix from https://github.com/apache/arrow-datafusion/pull/2135

# Changes
1. Update to https://github.com/apache/arrow-datafusion/commit/2a4a835bd727b00c58631a1c807bb598a0a12a93


Cargo.lock
```diff
@@ -5631,15 +5631,6 @@ dependencies = [
  "unicode_categories",
 ]
 
-[[package]]
-name = "sqlparser"
```
alamb, line 5635: removed old version of sqlparser too, which is nice

tustvold (approved):

PR #4235, opened 2022-04-05
feat: periodic high watermark fetcher

domodwyer:
Implements [fetching the maximum write buffer offset](https://github.com/influxdata/influxdb_iox/blob/b08540064cd2f9f614d23e532587ba61da8295cb/ingester/src/handler.rs#L356-L379) as a singly responsible type that can be mocked out for testing.

This is work towards replacing the existing stream handler in the ingester with `SequencedStreamHandler`:

- [ ] **THIS PR:** High watermark fetcher (feeds into instrumentation)
- [ ] Instrumentation of op apply
- [ ] Impl DmlSink for existing write buffer code path (via compatibility adaptor)
- [ ] Plumb in the `SequencedStreamHandler` and use it in prod

---

* feat: periodic kafka max watermark offset fetcher (891d2e136)

      Adds the PeriodicWatermarkFetcher type responsible for querying write buffer /
      Kafka for the maximum sequence number / offset, surfacing any errors via both
      logs & metrics.

      This high watermark / max offset value is used within the ingest 
      instrumentation metrics. This use case is tolerant of caching / stale values,
      and as such the value is periodically updated to minimise load on the write
      buffer.

* refactor: remove manual Debug impl (506cdebf3)

      Derive the debug impl so it prints all the fields (specifically the
      "number of sequencers configured" is pretty helpful in a test).

      Manual impls drift over time and are more effort than the derive!

### DIFF
--- a/ingester/src/stream_handler/periodic_watermark_fetcher.rs
+++ b/ingester/src/stream_handler/periodic_watermark_fetcher.rs
@@ -1,0 +1,256 @@
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc,
+    },
+    time::{Duration, Instant},
+};
+
+use data_types2::KafkaPartition;
+use metric::U64Counter;
+use observability_deps::tracing::*;
+use tokio::task::JoinHandle;
+use write_buffer::core::WriteBufferReading;
+
+/// Periodically fetch and cache the maximum known write buffer offset
+/// (watermark) from the write buffer for a given sequencer.
+///
+/// Uses [`WriteBufferReading::fetch_high_watermark()`] internally to obtain the
+/// latest offset in the background, updating the cached value when it
+/// successfully returns a value.
+///
+/// Emits an error metric named `write_buffer_watermark_fetch_errors` that
+/// increments once per fetch error.
+#[derive(Debug)]
+pub struct PeriodicWatermarkFetcher {
+    last_watermark: Arc<AtomicU64>,
+    poll_handle: JoinHandle<()>,
+}
+
+impl PeriodicWatermarkFetcher {
+    /// Instantiate a new [`PeriodicWatermarkFetcher`] that polls `write_buffer`
+    /// every `interval` period for the maximum offset for `partition`.
+    pub fn new(
+        write_buffer: Arc<dyn WriteBufferReading>,
+        partition: KafkaPartition,
+        interval: Duration,
+        metrics: &metric::Registry,
+    ) -> Self {
+        // Initialise a new poller to update the watermark in the background.
+        let (poller, last_watermark) = Poller::new(write_buffer, partition, metrics);
+
+        // Start the poller and retain a handle to stop it once this fetcher is
+        // dropped.
+        let poll_handle = tokio::spawn(async move {
+            poller.run(interval).await;
+        });
+
+        Self {
+            last_watermark,
+            poll_handle,
+        }
+    }
+
+    /// Returns the last observed watermark value, if available.
+    ///
+    /// The [`Instant`] in the returned tuple is the approximate instant the
+    /// watermark was observed, allowing the caller to determine the staleness
+    /// of the value.
+    ///
+    /// If the watermark value has yet to be observed (i.e. due to continuous
+    /// fetch errors) `None` is returned.
+    pub async fn cached_watermark(&self) -> Option<u64> {
+        match self.last_watermark.load(Ordering::Relaxed) {
+            // A value of 0 means "never observed a watermark".
+            0 => None,
+            v => Some(v),
+        }
+    }
+}
+
+impl Drop for PeriodicWatermarkFetcher {
+    fn drop(&mut self) {
+        // Stop the background poller task.
+        self.poll_handle.abort();
+    }
+}
+
+/// The internal poller half of the [`PeriodicWatermarkFetcher`] that performs
+/// the write buffer request and updates the watermark.
+///
+/// Logs any errors that occur and emits an error count metric.
+///
+/// # Encoding
+///
+/// The atomic watermark value encodes "no watermark ever observed" as 0. All
+/// non-zero values encode the last maximum watermark observed.
+#[derive(Debug)]
+struct Poller {
+    write_buffer: Arc<dyn WriteBufferReading>,
+
+    // The sequencer / kafka partition to ask the write buffer the max offset
+    // for.
+    sequencer_id: u32,
+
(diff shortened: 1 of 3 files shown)


ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+        metrics: &metric::Registry,
+    ) -> Self {
+        // Initialise a new poller to update the watermark in the background.
+        let (poller, last_watermark) = Poller::new(write_buffer, partition, metrics);
+
+        // Start the poller and retain a handle to stop it once this fetcher is
+        // dropped.
+        let poll_handle = tokio::spawn(async move {
+            poller.run(interval).await;
+        });
```
domodwyer, line 46: Polling for new watermark values now occurs outside of the ingest code path, in a background task.

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+    ///
+    /// If the watermark value has yet to be observed (i.e. due to continuous
+    /// fetch errors) `None` is returned.
+    pub async fn cached_watermark(&self) -> Option<u64> {
+        match self.last_watermark.load(Ordering::Relaxed) {
+            // A value of 0 means "never observed a watermark".
+            0 => None,
+            v => Some(v),
+        }
+    }
```
domodwyer, line 68: This impl explicitly encodes the case of "I've not seen a watermark value yet" as a `None` to let the caller decide if this is important.

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+        metrics: &metric::Registry,
+    ) -> (Self, Arc<AtomicU64>) {
+        let last_watermark = Arc::new(AtomicU64::new(0));
+        let rx = Arc::clone(&last_watermark);
+
+        let error_count = metrics
+            .register_metric::<U64Counter>(
+                "write_buffer_watermark_fetch_errors",
+                "The total number of errors observed while trying to query for the high watermark",
+            )
```
domodwyer, line 121: This error counter is incremented each time fetching the watermark fails - this will help us identify when the rest of the write buffer metrics may be incorrect (as they depend on a roughly-correct watermark value).

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+                Err(e) => {
+                    // Log how long ago the watermark was last updated
+                    let cache_staleness_seconds =
+                        now.duration_since(last_successful_poll).as_secs();
+
+                    warn!(
+                        error=%e,
+                        %cache_staleness_seconds,
+                        "failed to read write buffer watermark"
+                    );
```
domodwyer, line 171: Failing to fetch the watermark is now a `WARN` rather than debug to help us identify when/if it occurs.

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+            0 => None,
+            v => Some(v),
+        }
+    }
+}
+
+impl Drop for PeriodicWatermarkFetcher {
+    fn drop(&mut self) {
+        // Stop the background poller task.
+        self.poll_handle.abort();
```
alamb, line 74: 👍

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    /// Returns the last observed watermark value, if available.
+    ///
+    /// The [`Instant`] in the returned tuple is the approximate instant the
+    /// watermark was observed, allowing the caller to determine the staleness
+    /// of the value.
+    ///
+    /// If the watermark value has yet to be observed (i.e. due to continuous
```
alamb, line 60: I assume that offset 0 is not a valid kafka offset?

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -0,0 +1,256 @@
...
+        let metrics = Arc::new(metric::Registry::default());
+        let fetcher = PeriodicWatermarkFetcher::new(
+            write_buffer,
+            KafkaPartition::new(0),
+            Duration::from_millis(10),
+            &*metrics,
+        );
+
+        assert_eq!(fetcher.cached_watermark().await, None);
+        tokio::time::sleep(Duration::from_millis(50)).await;
```
alamb, line 203: Rather than a static sleep here (which is subject to the whims of many things, perhaps we can change this to be a  loop that polls `metrics.get_instrument()...` until it sees a result (or timeout occurs)

Aka how you have done it using `with_timeout_panic` in the other test

write_buffer/src/mock.rs
```diff
@@ -347,6 +347,7 @@ impl WriteBufferWriting for MockBufferForWritingThatAlwaysErrors {
     }
 }
 
+#[derive(Debug)]
```
alamb, line 350: 👍

alamb (approved):

ingester/src/stream_handler/periodic_watermark_fetcher.rs
domodwyer: Unlike the other test, this one asserts _nothing changes_ so adding a loop would wind up functionally the same to this, but more complex.

alamb: I always cringe when I see naked `sleep` in a test -- as brings back memories of chasing down non deterministic errors

ingester/src/stream_handler/periodic_watermark_fetcher.rs
alamb: I don't follow this logic 🤔  but then again I am not at all sure it will matter in practice

ingester/src/stream_handler/periodic_watermark_fetcher.rs
domodwyer: Yep I too have had the delight of a codebase full of racy tests so I understand.

However if this test flakes, I shall provide you with 5 of the finest United States Dollars at JOIN 💵 🤑

alamb: deal!

Issue #4237, opened 2022-04-05
Different meanings of `sequencer_id` in ingester and router2 / write_buffer is super confusing


alamb:
__Use case / Issue:__

* In the ingester, `sequencer_id` means auto assigned id from the catalog
* In  `router2`, `write_buffer`, `dml` and elsewhere, `sequencer_id` means "kafka partition"

This is *very* confusing (I spent several hours trying to debug why a sequencer_id was not present in the ingester while working on https://github.com/influxdata/influxdb_iox/pull/4227) 🤯 

@domodwyer and I believe that a sequencer is always 1-to-1 with kafka partitions. That is to say there should only ever be one entry in the catalog per kafka partition, as the sequencer entry in the catalog “owns” that kafka partition

__Proposal:__
Make use of `sequencer_id` consistent in the codebase  by changing the catalog code to force `kafka_partition_id` == `sequencer_id` (as they should be 1-to-1). See alternates below

__Desired behaviour:__
`sequencer_id` means the same thing in different crates

__Alternatives considered:__
1. Change catalog code to enforce `kafka_partition_id` == `sequencer_id` (as they should be 1-to-1)
2. Change ingester (and catalog structures) to use `kafka_partition_id` instead of `sequencer_id`

alamb:
I believe @pauldix  would prefer option 3 ("Change router / writebuffer to use kafka_partition_id instead of sequencer_id to identify which partition was targeted")

that is to update the rest of the code to be consistent with the ingester.

alamb:
After discussion with @NGA-TRAN @pauldix and @domodwyer this morning, the consensus here appears to be that:
1. Wait to do anything until after we have https://github.com/influxdata/influxdb_iox/issues/4245
2. Rename `sequencer_id` in `dml` and `write_buffer` crates to use the same terminology as the ingester

The second item is purposely worded that way because the exact term to use may change (e.g. https://github.com/influxdata/influxdb_iox/issues/4246) before this ticket is worked

PR #4243, opened 2022-04-06
feat: ingester op instrumentation

domodwyer:
This PR adds a `DmlSink` implementation that decorates other implementations to emit metrics & traces during the `DmlSink::apply()` call.

This type currently emits both the write buffer metrics and op call metrics - these responsibilities will likely need splitting if we wind up composing together `DmlSink` implementations in the future and would like to record latencies for each.

There are some differences in the metrics this PR adds compared to what we [currently have](https://github.com/influxdata/influxdb_iox/blob/b6f9c9bfd4b138d71d5440b372dc5652223ac53f/db/src/write_buffer/metrics.rs#L29-L60):

* All metrics were [labelled with `db_name`](https://github.com/influxdata/influxdb_iox/blob/b6f9c9bfd4b138d71d5440b372dc5652223ac53f/db/src/write_buffer/metrics.rs#L77) - this is now [`kafka_topic`](https://github.com/influxdata/influxdb_iox/blob/b6f9c9bfd4b138d71d5440b372dc5652223ac53f/ingester/src/stream_handler/sink_instrumentation.rs#L89).
* Removed `write_buffer_last_min_ts` and `write_buffer_last_max_ts` - I'm not sure these are very useful but happy to put them back if they are important to someone.
* `write_buffer_ingest_request_duration` was confusingly named - it recorded the duration of the op apply after being read from the write buffer and was independent of any write buffer latency. This is now `ingester_op_apply_duration_ms` and has been changed to a latency histogram to capture the latency distribution of op handling.
* `write_buffer_ingest_requests` has been removed as it is redundant - the ingest duration histogram counts the number of handled ops/requests.
* All write buffer metrics are emitted immediately after being read from the write buffer and _before_ the op apply is performed so they're kept as up-to-date as possible and record what has been read from the write buffer irrespective of what happens with the op.

These metrics are NOT currently wired in - this is part of a series of PRs to replace the [`stream_in_sequenced_entries()`](https://github.com/influxdata/influxdb_iox/blob/b6f9c9bfd4b138d71d5440b372dc5652223ac53f/ingester/src/handler.rs#L273) fn:

- [x] High watermark fetcher (feeds into instrumentation) (#4235)
- [ ] **THIS PR:** Instrumentation of op apply
- [ ] Impl DmlSink for existing write buffer code path (via compatibility adaptor)
- [ ] Plumb in the `SequencedStreamHandler` and use it in prod

---

* feat: DmlSink instrumentation (436da19d9)

      This commit adds the SinkInstrumentation type that decorates an inner DmlSink
      with call latency and write buffer metrics.

      The write buffer / sink call metrics may be split apart into two separate
      responsibilities in the future if there are multiple DmlSink that need
      instrumentation, but deferring adding more types until it is needed.

* refactor: impl WatermarkFetcher (f6c65f52a)

      Implement WatermarkFetcher for PeriodicWatermarkFetcher and remove unnecessary
      async.

* feat: emit tracing span for op apply (091640bb2)

      This commit uses the tracing metadata within the DmlOperation to emit a 
      tracing span from the ingester covering the DmlSink::apply() operation.

### DIFF
--- a/ingester/src/stream_handler/periodic_watermark_fetcher.rs
+++ b/ingester/src/stream_handler/periodic_watermark_fetcher.rs
@@ -12,6 +12,8 @@ use observability_deps::tracing::*;
 use tokio::task::JoinHandle;
 use write_buffer::core::WriteBufferReading;
 
+use super::sink_instrumentation::WatermarkFetcher;
+
 /// Periodically fetch and cache the maximum known write buffer offset
 /// (watermark) from the write buffer for a given sequencer.
 ///
@@ -59,7 +61,7 @@ impl PeriodicWatermarkFetcher {
     ///
     /// If the watermark value has yet to be observed (i.e. due to continuous
     /// fetch errors) `None` is returned.
-    pub async fn cached_watermark(&self) -> Option<u64> {
+    pub fn cached_watermark(&self) -> Option<u64> {
         match self.last_watermark.load(Ordering::Relaxed) {
             // A value of 0 means "never observed a watermark".
             0 => None,
@@ -177,6 +179,12 @@ impl Poller {
     }
 }
 
+impl WatermarkFetcher for PeriodicWatermarkFetcher {
+    fn watermark(&self) -> Option<u64> {
+        self.cached_watermark()
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use data_types2::Sequence;
@@ -199,9 +207,9 @@ mod tests {
             &*metrics,
         );
 
-        assert_eq!(fetcher.cached_watermark().await, None);
+        assert_eq!(fetcher.cached_watermark(), None);
         tokio::time::sleep(Duration::from_millis(50)).await;
-        assert_eq!(fetcher.cached_watermark().await, None);
+        assert_eq!(fetcher.cached_watermark(), None);
 
         let hits = metrics
             .get_instrument::<Metric<U64Counter>>("write_buffer_watermark_fetch_errors")
@@ -232,7 +240,7 @@ mod tests {
         async {
             loop {
                 // Read the watermark until the expected value is observed.
-                match fetcher.cached_watermark().await {
+                match fetcher.cached_watermark() {
                     Some(42) => break,
                     // The mock is configured to return 42 - any other value
                     // is incorrect.
--- a/ingester/src/stream_handler/sink_instrumentation.rs
+++ b/ingester/src/stream_handler/sink_instrumentation.rs
@@ -1,3 +1,4 @@
+//! Instrumentation for [`DmlSink`] implementations.
 use std::fmt::Debug;
 
 use async_trait::async_trait;
@@ -5,6 +6,7 @@ use data_types2::KafkaPartition;
 use dml::DmlOperation;
 use metric::{Attributes, U64Counter, U64Gauge, U64Histogram, U64HistogramOptions};
 use time::{SystemProvider, TimeProvider};
+use trace::span::SpanRecorder;
 
 use super::DmlSink;
 
@@ -21,7 +23,7 @@ pub trait WatermarkFetcher: Debug + Send + Sync {
 
 /// A [`SinkInstrumentation`] decorates a [`DmlSink`] implementation and records
 /// write buffer metrics and the latency of the decorated [`DmlSink::apply()`]
-/// call.
+/// call, and emits a tracing span covering the call duration.
 ///
 /// # Panics
 ///
@@ -216,6 +218,12 @@ where
             );
         }
 
+        // Create a tracing span covering the inner DmlSink call.
+        let mut span_recorder = SpanRecorder::new(
+            meta.span_context()
+                .map(|parent| parent.child("DmlSink::apply()")),
+        );
+
         // Call into the inner handler to process the op and calculate the call
         // latency.
         let started_at = self.time_provider.now();
@@ -225,8 +233,14 @@ where
         // latency.
         if let Some(delta) = self.time_provider.now().checked_duration_since(started_at) {
             let metric = match &res {
-                Ok(_) => &self.op_apply_success_ms,
-                Err(_) => &self.op_apply_error_ms,
+                Ok(_) => {
+                    span_recorder.ok("success");
+                    &self.op_apply_success_ms
+                }
+                Err(e) => {
+                    span_recorder.error(e.to_string());
+                    &self.op_apply_error_ms
+                }
             };
             metric.record(delta.as_millis() as _);
         }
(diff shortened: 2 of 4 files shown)


ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+{
+    async fn apply(&self, op: DmlOperation) -> Result<bool, crate::data::Error> {
+        let meta = op.meta();
+
+        // Immediately increment the "bytes read" metric as it records the
+        // number of bytes read from the sequencer, irrespective of the op
+        // apply call.
+        self.write_buffer_bytes_read.inc(
+            meta.bytes_read()
+                .expect("entry from write buffer should have size") as u64,
```
domodwyer, line 183: Missing metadata causes a panic - this is the same behaviour as the current instrumentation.

ingester/src/stream_handler/periodic_watermark_fetcher.rs
```diff
@@ -59,7 +61,7 @@ impl PeriodicWatermarkFetcher {
     ///
     /// If the watermark value has yet to be observed (i.e. due to continuous
     /// fetch errors) `None` is returned.
-    pub async fn cached_watermark(&self) -> Option<u64> {
+    pub fn cached_watermark(&self) -> Option<u64> {
```
domodwyer, line 64: This didn't need to be async 🎉

mkmik (approved):
LGTM

ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+where
+    F: WatermarkFetcher,
+    T: DmlSink,
+{
+    /// Construct a new [`SinkInstrumentation`] layer that decorates `inner`
+    /// with logic that records metrics from the [`DmlOperation`] instances it
+    /// observes.
+    ///
+    /// The current high watermark is read from `watermark_fetcher` and used to
+    /// derive some metric values (such as lag). This impl is tolerant of
```
alamb, line 78: I am not sure what "tolerant of cached/stale watermark values..." means (as in what am I as a reader of this code supposed to conclude from this information? Is it that the sink itself calculates a high water mark from the watermark fetcher?)

ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+                .timestamp_nanos() as u64,
+        );
+
+        // Extract the sequence number from the op before giving up ownership
+        // to the inner DmlSink (avoiding a clone of the large op).
+        let sequence = meta
+            .sequence()
+            .expect("entry from write buffer must be sequenced");
+        assert_eq!(
+            sequence.sequencer_id as i32, self.sequencer_id,
```
alamb, line 202: Having just gone though a similar exercise, I think `sequence.sequencer_id` *actually* equals `kafka_partition`, not the ingester's sequencer_id

https://github.com/influxdata/influxdb_iox/issues/4237 strikes again!

ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+                .map(|parent| parent.child("DmlSink::apply()")),
+        );
+
+        // Call into the inner handler to process the op and calculate the call
+        // latency.
+        let started_at = self.time_provider.now();
+        let res = self.inner.apply(op).await;
+
+        // If the clocks go backwards, skip recording the (nonsense) call
+        // latency.
```
alamb, line 233: I don't understand how this can happen. Is is it worth a `warn!`?

ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+    // This test asserts the various metrics are set in the happy path.
+    #[tokio::test]
+    async fn test_call_inner_ok() {
+        let metrics = metric::Registry::default();
+        let traces: Arc<dyn TraceCollector> = Arc::new(RingBufferTraceCollector::new(5));
+        let span = SpanContext::new(Arc::clone(&traces));
+
+        let meta = DmlMeta::sequenced(
+            // Op is offset 100 for sequencer 42
+            Sequence::new(SEQUENCER_ID, 100),
```
alamb, line 356: as mentioned above, I think this should be kafka partition id

ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -0,0 +1,641 @@
...
+        );
+        assert_matches!(
+            get_metric::<U64Gauge>(&metrics, "write_buffer_last_ingest_ts", &*DEFAULT_ATTRS),
+            // 12345 - 100 - 1
+            Observation::U64Gauge(t) => {
+                assert_eq!(t, TEST_TIME.timestamp_nanos() as u64);
+            }
+        );
+
+        // Validate the success histogram was hit
```
alamb, line 467:
```suggestion
        // Validate the histogram was hit even on error
```

alamb (approved):

ingester/src/stream_handler/sink_instrumentation.rs
domodwyer: I think this code is correct - the `SinkInstrumentation` constructor [takes a `KafkaPartition`](https://github.com/influxdata/influxdb_iox/blob/24eeddce8a1d21ec78bae8e2712bdea465600b86/ingester/src/stream_handler/sink_instrumentation.rs#L84) which is then [stored in the `sequencer_id` field](https://github.com/influxdata/influxdb_iox/blob/24eeddce8a1d21ec78bae8e2712bdea465600b86/ingester/src/stream_handler/sink_instrumentation.rs#L154), and here we're checking the [value in the `DmlOperation`'s metadata](https://github.com/influxdata/influxdb_iox/blob/24eeddce8a1d21ec78bae8e2712bdea465600b86/data_types/src/sequence.rs#L4) which is documented as being the kafka partition ID - I think these should be the same values?

This is all a bit silly though - I hope we can rename everything sooner rather than later to put this to rest.

ingester/src/stream_handler/sink_instrumentation.rs
domodwyer: It uses the wall clock, so it could drift backwards during this fn call.

I'm not sure it's useful to log it - we don't do it anywhere else where we use this pattern and I'm not sure what an operator would do with that information.

ingester/src/stream_handler/sink_instrumentation.rs
domodwyer: See comment above 👍

ingester/src/stream_handler/sink_instrumentation.rs
domodwyer: I wanted to convey that this `SinkInsturmentation` can accept a not-exactly-the-right-value value - it is tolerant of being given a cached/stale watermark from the watermark fetcher impl.

If we're processing op 100, but the cached watermark is old and says "the highest known offset is 50" then obviously the "how far behind are we" maths will give the wrong answer (50-100= beind by -50 ops?!) - this could cause an underflow leading to a ridiculous lag value, it could panic, or as in this impl, we can tolerate it by using [a saturating subtraction](https://github.com/influxdata/influxdb_iox/blob/24eeddce8a1d21ec78bae8e2712bdea465600b86/ingester/src/stream_handler/sink_instrumentation.rs#L214-L218) so that 50-100=0 - we are indeed 0 ops behind our example high watermark of 50.

ingester/src/stream_handler/sink_instrumentation.rs
alamb: Perhaps -- probably best proceeding like this and then you can change the code if it starts asserting when you hook it up

PR #4250, opened 2022-04-07
refactor: compactor debug logging

domodwyer:
This PR simply adds more debug logging - can't fix what you can't see 👀 

---

* refactor: print parquet path in Debug impl (2607151ec)

      Print the actual path being used when debug-printing a ParquetFilePath.

* refactor: add debug in compaction path (3706ac042)

      Adds debug!() and friends through the compaction path.

### DIFF
--- a/iox_object_store/src/paths/parquet_file.rs
+++ b/iox_object_store/src/paths/parquet_file.rs
@@ -26,8 +26,21 @@ enum Variant {
 
 /// Location of a Parquet file within a database's object store.
 /// The exact format is an implementation detail and is subject to change.
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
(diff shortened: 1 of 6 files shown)


mkmik (approved):

iox_object_store/src/paths/parquet_file.rs
```diff
@@ -26,9 +26,22 @@ enum Variant {
...
+            Variant::Old { .. } => self.relative_dirs_and_file_name().to_string(),
+            Variant::New { .. } => self.absolute_dirs_and_file_name().to_string(),
+        };
+        f.debug_struct("ParquetFilePath")
+            .field("inner", &self.0)
+            .field("resolved_path", &path)
+            .finish()
+    }
+}
+
```
domodwyer, line 44: This manual debug impl includes all the information the derived debug impl did, but resolves the actual path it represents too - this way, we don't have to reverse engineer the path from the fields in the debug output.

PR #4256, opened 2022-04-08
refactor: extra debug in compactor

domodwyer:
More debug logging to narrow down where compaction hangs - the main compaction loop is now covered between each major event.

---

* refactor: extra debug in compactor (6131381b8)

      Continues pushing more debug through the compaction processing loop.

### DIFF
--- a/compactor/src/utils.rs
+++ b/compactor/src/utils.rs
@@ -100,8 +100,13 @@ impl ParquetFileWithTombstone {
         );
 
         debug!(
-            parquet_file=?decoded_parquet_file.parquet_file,
-            "generated parquet chunk from object store"
+            parquet_file_id=?decoded_parquet_file.parquet_file.id,
+            parquet_file_sequencer_id=?decoded_parquet_file.parquet_file.sequencer_id,
+            parquet_file_namespace_id=?decoded_parquet_file.parquet_file.namespace_id,
+            parquet_file_table_id=?decoded_parquet_file.parquet_file.table_id,
+            parquet_file_partition_id=?decoded_parquet_file.parquet_file.partition_id,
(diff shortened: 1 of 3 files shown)


compactor/src/utils.rs
```diff
@@ -100,8 +100,13 @@ impl ParquetFileWithTombstone {
...
 
         debug!(
-            parquet_file=?decoded_parquet_file.parquet_file,
-            "generated parquet chunk from object store"
+            parquet_file_id=?decoded_parquet_file.parquet_file.id,
+            parquet_file_sequencer_id=?decoded_parquet_file.parquet_file.sequencer_id,
+            parquet_file_namespace_id=?decoded_parquet_file.parquet_file.namespace_id,
+            parquet_file_table_id=?decoded_parquet_file.parquet_file.table_id,
+            parquet_file_partition_id=?decoded_parquet_file.parquet_file.partition_id,
+            parquet_file_object_store_id=?decoded_parquet_file.parquet_file.object_store_id,
```
domodwyer, line 108: I changed this so we don't print out the metadata bytes for each parquet file, instead just the useful fields.

I am looking forward to when tracing fields are decoupled from the debug repr.

parquet_file/src/storage.rs
```diff
@@ -300,7 +300,7 @@ impl Storage {
             }
         }
 
-        debug!("Completed parquet download & scan");
+        debug!(?path, "Completed parquet download & scan");
```
domodwyer, line 303: Correlation! I forgot to add the path to this, so I couldn't see which files had completed.

e-dard (approved):
LGTM

PR #4257, opened 2022-04-08
feat(ingester): use SequencedStreamHandler

domodwyer:
This PR switches out the old `stream_in_sequenced_entries()` kafka consumer for the `SequencedStreamHandler` introduced in #4203.

If this swap causes problems, reverting dce939c58 (or this whole PR) should switch back to the old code paths - if anything goes wrong, I'll be doing this to unblock prod once I've grabbed enough problem context. As such, depending on what time this PR is reviewed, I may **hold off merging** until Monday to ensure I am around when it is deployed.

Once merged the ingesters will emit slightly different metrics, and I will ensure the dashboards are updated.

- [x] High watermark fetcher (feeds into instrumentation) (#4235)
- [x] Instrumentation of op apply (#4243)
- [x] **THIS PR:** Impl DmlSink for existing write buffer code path (via compatibility adaptor)
- [x] **THIS PR:** Plumb in the `SequencedStreamHandler` and use it in prod

---

* feat: impl DmlSink for IngesterData (c2236fa3f)

      This commit adds an adaptor (IngestSinkAdaptor) that provides a DmlSink 
      implementation for the existing write path (IngesterData). With this, the
      existing write path becomes compatible with the new op stream handler
      (SequencedStreamHandler).

* refactor: accept !Sync write buffer streams (71a278ac7)

      Removes the Sync bound SequencedStreamHandler input stream type, as the 
      BoxStream returned by the WriteBufferStreamHandler is not Sync.

      This change means the SequencedStreamHandler is not Sync either, but is still
      Send and therefore can be moved into tokio tasks.

* refactor: use SequencedStreamHandler (dce939c58)

      Removes the old stream_in_sequenced_entries() write buffer handler, replacing
      it with the SequencedStreamHandler introduced in #4203.

      This change will affect the metrics emitted by an ingester as outlined in
      #4243.

### DIFF
--- a/ingester/Cargo.toml
+++ b/ingester/Cargo.toml
@@ -17,7 +17,6 @@ data_types2 = { path = "../data_types2" }
 futures = "0.3"
 generated_types = { path = "../generated_types" }
 chrono = { version = "0.4", default-features = false }
-db = { path = "../db" }
 dml = { path = "../dml" }
 hyper = "0.14"
 iox_catalog = { path = "../iox_catalog" }
--- a/ingester/src/handler.rs
+++ b/ingester/src/handler.rs
@@ -2,39 +2,33 @@
 
 use crate::{
     data::{IngesterData, IngesterQueryResponse, SequencerData},
-    lifecycle::{run_lifecycle_manager, LifecycleConfig, LifecycleHandle, LifecycleManager},
-    poison::{PoisonCabinet, PoisonPill},
+    lifecycle::{run_lifecycle_manager, LifecycleConfig, LifecycleManager},
+    poison::PoisonCabinet,
     querier_handler::prepare_data_to_querier,
+    stream_handler::{
+        sink_adaptor::IngestSinkAdaptor, sink_instrumentation::SinkInstrumentation,
+        PeriodicWatermarkFetcher, SequencedStreamHandler,
+    },
 };
 use async_trait::async_trait;
 use backoff::BackoffConfig;
-use data_types2::{IngesterQueryRequest, KafkaPartition, KafkaTopic, Sequencer, SequencerId};
-use db::write_buffer::metrics::{SequencerMetrics, WriteBufferIngestMetrics};
-use dml::DmlOperation;
+use data_types2::{IngesterQueryRequest, KafkaPartition, KafkaTopic, Sequencer};
+
 use futures::{
     future::{BoxFuture, Shared},
-    pin_mut,
     stream::FuturesUnordered,
     FutureExt, StreamExt, TryFutureExt,
 };
 use iox_catalog::interface::Catalog;
-use metric::Attributes;
-use metric::U64Counter;
-use metric::U64Gauge;
 use object_store::DynObjectStore;
-use observability_deps::tracing::{debug, error, info, warn};
+use observability_deps::tracing::*;
 use query::exec::Executor;
 use snafu::{ResultExt, Snafu};
-use std::{
-    collections::BTreeMap,
-    sync::Arc,
-    time::{Duration, Instant},
-};
-use time::{SystemProvider, TimeProvider};
+use std::{collections::BTreeMap, sync::Arc, time::Duration};
+use time::SystemProvider;
 use tokio::task::{JoinError, JoinHandle};
 use tokio_util::sync::CancellationToken;
-use trace::span::SpanRecorder;
-use write_buffer::core::{WriteBufferReading, WriteBufferStreamHandler};
+use write_buffer::core::WriteBufferReading;
 use write_summary::SequencerProgress;
 
 #[derive(Debug, Snafu)]
@@ -113,9 +107,6 @@ pub struct IngestHandlerImpl {
     /// A token that is used to trigger shutdown of the background worker
     shutdown: CancellationToken,
 
-    /// Poison pills for testing.
-    poison_cabinet: Arc<PoisonCabinet>,
-
     /// The cache and buffered data for the ingester
     data: Arc<IngesterData>,
 }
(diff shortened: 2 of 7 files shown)


ingester/Cargo.toml
```diff
@@ -17,7 +17,6 @@ data_types2 = { path = "../data_types2" }
 futures = "0.3"
 generated_types = { path = "../generated_types" }
 chrono = { version = "0.4", default-features = false }
-db = { path = "../db" }
```
domodwyer, line 20: 👋 one less dependency

ingester/src/stream_handler/handler.rs
```diff
@@ -157,7 +157,7 @@ impl<I, O> SequencedStreamHandler<I, O> {
 
 impl<I, O, T> SequencedStreamHandler<I, O, T>
 where
-    I: Stream<Item = Result<DmlOperation, WriteBufferError>> + Unpin + Send + Sync,
+    I: Stream<Item = Result<DmlOperation, WriteBufferError>> + Unpin + Send,
```
domodwyer, line 160: The `BoxStream` of ops was not sync - not a problem though 👍

ingester/Cargo.toml
alamb: one less OG dependency 👍

ingester/src/handler.rs
```diff
@@ -152,7 +142,6 @@ impl IngestHandlerImpl {
 
         let ingester_data = Arc::clone(&data);
         let kafka_topic_name = topic.name.clone();
-        let ingest_metrics = WriteBufferIngestMetrics::new(&metric_registry, &topic.name);
```
alamb, line 155: Just to confirm I understand this change, the old ingest metrics have now been entirely replaced by the new stream of Dml StreamHandlers (which have their own metric reporting), right?

ingester/src/handler.rs
```diff
@@ -576,23 +401,24 @@ mod tests {
 
         let observation = ingester
             .metrics
-            .get_instrument::<Metric<U64Counter>>("write_buffer_ingest_requests")
+            .get_instrument::<Metric<U64Histogram>>("ingester_op_apply_duration_ms")
```
alamb, line 404: I assume some dashboards will need updating (but that is ok)?

ingester/src/handler.rs
```diff
@@ -616,72 +442,29 @@ mod tests {
...
             .get_observer(&Attributes::from(&[
-                ("db_name", "whatevs"),
+                ("kafka_topic", "whatevs"),
                 ("sequencer_id", "0"),
             ]))
             .unwrap()
             .fetch();
         assert_eq!(observation, 0);
 
-        let observation = ingester
```
alamb, line 626: Is the assessment is that `write_buffer_last_min_ts` and `write_buffer_last_max_ts` are not important metrics to track anymore?

ingester/src/handler.rs
```diff
@@ -616,72 +442,29 @@ mod tests {
...
-                ("db_name", "whatevs"),
-                ("sequencer_id", "0"),
-            ]))
-            .unwrap()
-            .fetch();
-        assert_eq!(observation, ingest_ts2.timestamp_nanos() as u64);
-
-        let observation = ingester
-            .metrics
-            .get_instrument::<Metric<U64Gauge>>("ingester_ttbr_ms")
```
alamb, line 664: It seems like the `ingester_ttbr_ms` is an important metric to report -- is there a reason it is no longer tested? I think I remember seeing a `ttbr` metric in one of the other refactoring PRs  -- perhaps the test could be updated to use that metric?

ingester/src/handler.rs
```diff
@@ -696,56 +479,6 @@ mod tests {
             .unwrap();
     }
 
-    #[tokio::test]
```
alamb, line 699: What is the reason for removing these tests (and the PosionPill functionality that enabled them)? They seem to test valuable cases (that the ingest handler shuts down cleanly) that would be hard to test otherwise

alamb (commented):
Code looks 👨‍🍳 ❤️ 

I had two questions on test changes, but otherwise this looks very nice 👍

ingester/src/handler.rs
domodwyer: I think the poison pill thing is a long winded way of getting to test the [`join()`](https://github.com/influxdata/influxdb_iox/blob/dce939c5801b06eeff3cb8a7688ca9dc0f4a8225/ingester/src/handler.rs#L247-L265) functionality that checks if a job has exploded or stopped early. I have pushed 5c3cbb14b that tests the same behaviour without it - it seems like a bit of a code smell to pass through testing "please explode" flags into prod code, perhaps trying to work around the tight coupling of the old handler.

A stream that stops early [now panics](https://github.com/influxdata/influxdb_iox/blob/dce939c5801b06eeff3cb8a7688ca9dc0f4a8225/ingester/src/stream_handler/handler.rs#L254-L257) by itself and [is tested here](https://github.com/influxdata/influxdb_iox/blob/dce939c5801b06eeff3cb8a7688ca9dc0f4a8225/ingester/src/stream_handler/handler.rs#L718).

ingester/src/handler.rs
domodwyer: Correct! This also allows us to remove the `db` dep, which was only used for these metrics.

ingester/src/handler.rs
domodwyer: Yeah, as per PR description I will ensure this happens once we're confident the change doesn't need rolling back

ingester/src/handler.rs
domodwyer: No, as mentioned in https://github.com/influxdata/influxdb_iox/pull/4243 I'm not sure if they're actually useful/used, though they can be easily replaced if they turn out to be useful.

ingester/src/handler.rs
domodwyer: Testing the TTBR emitted by the handler moved over to the handler file where it is easier to various code paths (error/success) 👍

alamb (approved):

alamb:
Looks good -- thanks

PR #4255, opened 2022-04-07
feat: MVP include un-persisted results from the ingester in query results

alamb:
~Draft / WIP / YOLO PR: once I get it kind of working I will split off particular parts into their own PRs and list below.~

# Rationale
The first "could minimally claim to work" PR for https://github.com/influxdata/influxdb_iox/issues/3996

See design doc: https://docs.google.com/document/d/1LJZyw-Ik3_EMpXwEFHu3a9c2RJIR3uxpnDmEfL56Sl4/edit#heading=h.fe76eikpw09c

This PR allows the querier to include un-persisted results from the ingester to drastically reduce "time to be readable" 

# Changes
1. Make a flight call from querier to ingester
2. Implement query interface for`IngesterPartition` for data returned from the querier 
3. (Very) basic end to end test coverage

# PR sequence leading up to this one
- [x] Add `Error` back to `QueryDatabase::chunks` trait (so ingester_connection can return error): https://github.com/influxdata/influxdb_iox/pull/4260
- [x]  Wire in basic "specify ingester address" configuration: https://github.com/influxdata/influxdb_iox/pull/4259
- [x] Support multiple ingester addresses https://github.com/influxdata/influxdb_iox/pull/4271
- [x] Clean up flight APIs https://github.com/influxdata/influxdb_iox/pull/4273

# Follow on tasks / PRs:
- [ ] Update ingester request to querier to ask about all ingester ids (rather than just one)
- [ ] Update ingester response to querier to include data per partition rather than data combined for all partitions
- [ ] Implement  logic that detects when the max persisted data is inconsistent with known parquet
- [ ] Do something useful (like retry) when inconsistency is hit
- [ ] File ticket for why the `RecordBatches` batches sent from the ingester are not Dictionary encoded: https://github.com/influxdata/influxdb_iox/issues/4275

PR #4258, opened 2022-04-08
test: Add schema validation to end to end querier test

alamb:
This is an end to end test I wrote while working on https://github.com/influxdata/influxdb_iox/pull/4255 and trying to debug what was going on with results from the querier. I figured I would pull it into its own PR


influxdb_iox/tests/end_to_end_ng_cases/namespace.rs
```diff
@@ -3,7 +3,7 @@ use test_helpers_end_to_end_ng::{maybe_skip_integration, MiniCluster, TestConfig
 
 /// Test the namespacea client
 #[tokio::test]
-async fn qerier_namespace_client() {
```
alamb, line 6: drive by typo fix

carols10cents (approved):
LGTM!

Issue #4144, opened 2022-03-28
NG Querier: Basic API / Configuration for Querier to find relevant ingester(s)


alamb:
The idea here is that to run a query the querier will have to directly communicate with the ingesters that hold the data. The first question is "how will it do so", aka "what url will be used"?

```
                                                                                 
             How will the querier find                                           
             the appropriate ingesters                                           
                     to query?                                                   
                                                                                 
┌─────────────┐                    ┌─────────────┐                               
│ Ingester 1  │◀─ ─ ─ ─ ─ ─ ─ ─ ─ ─│   Querier   │                               
└─────────────┘         ┌ ─ ─ ─ ─ ─└─────────────┘                               
      ...                                ││                                      
                        │                ││                                      
┌─────────────┐                          ││                                      
│ Ingester N  │◀─ ─ ─ ─ ┘                │└─────────────┐                        
└─────────────┘                          └───┐          │                        
                                             │          │                        
                                             │          ▼                        
                                             │       .─────.                     
                                             ▼      (       )                    
                                         ┌.─────.┐  │`─────'│    Object Store    
                         Catalog         (       )  │       │      (e.g S3)      
                        Database         │`─────'│  │.─────.│                    
                     (e.g. postgres)     │.─────.│  (       )                    
                                         (       )   `─────'                     
                                          `─────'                                
```

~Definition of done is:~
~1. A written plan reviewed by at least @pauldix and @mkmik~
~2. An implementation (or follow on ticket) to implement the plan~

Update: This ticket will track the basic configuration and https://github.com/influxdata/influxdb_iox/issues/4297 now tracks the more sophisticated service discovery that handles ingester replicas, etc

alamb:
Service discovery proposal: https://docs.google.com/document/d/1Rosq-KnXDeDcVxvWYPw9mLWktSMAgjHZUqBpizE6VGg/edit

PR #4259, opened 2022-04-08
feat: Add basic Querier <--> Ingester "Service Configuration"

alamb:
# Rationale
Part of allowing the querier to include not-yet-persisted data from the ingester: https://github.com/influxdata/influxdb_iox/pull/4255

# Changes
1. Add a new configuration setting to tell queriers how to communicate with the specified ingesters.
2. Add a stub `QuerierConnection` in the ingester to encapsulate communicating with the ingester(s).
3. Add basic end to end test coverage (which I will expand on in several upcoming PRs)

# Follow on work
I plan an immediate follow on PR to add support for multiple ingester addresses but wanted to keep this PR smaller

This type of configuration clearly doesn't support uptime during rollout or dynamically changing ingester locations, but I think it will be ok for MVP. @mkmik  / @tustvold  have a more full featured configuration plan in https://github.com/influxdata/influxdb_iox/issues/4144 that we can do over time.


test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -91,6 +108,11 @@ impl TestConfig {
             .with_env("INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END", "0")
     }
 
+    /// Adds the ingester address
```
alamb, line 111: test plumbing

querier/src/ingester/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+
+/// Create a new ingester suitable for testing
+pub fn create_ingester_connection_for_testing() -> Arc<dyn IngesterConnection> {
+    Arc::new(MockIngesterConnection::new())
+}
+
+/// Handles communicating with the ingester(s) to retrieve
+/// data that is not yet persisted
+#[async_trait]
+pub trait IngesterConnection: std::fmt::Debug + Send + Sync + 'static {
```
alamb, line 52: Here is the key new structure -- `IngesterConnection` that handles encapsulating the communication with the ingester.

influxdb_iox/src/commands/run/querier.rs
```diff
@@ -59,6 +59,18 @@ pub struct Config {
     /// If not specified, defaults to the number of cores on the system
     #[clap(long = "--num-query-threads", env = "INFLUXDB_IOX_NUM_QUERY_THREADS")]
     pub num_query_threads: Option<usize>,
+
+    /// gRPC address for the querier to talk with the
```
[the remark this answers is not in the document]
alamb: @mkmik  if this configuration supported a comma separated list of hostnames such as the following:

```
INFLUXDB_IOX_INGESTER_ADDRESS=http://10.10.50.10:8082,http://10.10.50.20:8082,http://10.10.50.30:8082
```

Would that be sufficient to configure this in our current cloud environment (prior to https://github.com/influxdata/influxdb_iox/issues/4144 being done)?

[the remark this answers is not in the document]
mkmik: I assumed that we'd need to tell the querier for each sequencer ranges each ingester is responsible for.

How does it work with a "flat" list of urls?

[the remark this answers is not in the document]
alamb: > I assumed that we'd need to tell the querier for each sequencer ranges each ingester is responsible for.

Yes that is the long term plan as I understand it

> How does it work with a "flat" list of urls?

My plan was to query *all* the ingesters and then combine the results somehow (the somehow being somewhat handwavy at the momement -- likely it would involve picking the sequencer that had the most up to date data (or that answered first)

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -184,4 +206,10 @@ impl TestConfig {
     pub fn client_headers(&self) -> &[(HeaderName, HeaderValue)] {
         self.client_headers.as_ref()
     }
+
+    /// Get a reference to the test config's addrs.
+    #[must_use]
```
carols10cents, line 211: Why is this `#[must_use]`?

carols10cents (approved):
Some small thoughts you can take or leave, I don't feel a need to rereview this but might have more comments when this code is used more :)

querier/src/ingester/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+#[async_trait]
+impl IngesterConnection for IngesterConnectionImpl {
+    /// Retrieve chunks from the ingester for the particular table and
+    /// predicate
+    async fn partitions(
+        &self,
+        _namespace_name: Arc<str>,
+        table_name: &str,
```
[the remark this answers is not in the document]
alamb: Its a good idea -- I will do so

test_helpers_end_to_end_ng/src/config.rs
alamb: I have been using `rust_analyzer` (via `lsp-mode` in emacs) generate accessors for me -- it automatically adds the `#[must_use]` annotation. I am not sure why but I figured it wasn't a bad idea so I left it in

alamb:
Thanks for the review @carols10cents  -- I want to run the basic idea past @crepererum  on Monday before I merge this, but I will address your comments before doing so

influxdb_iox/src/commands/run/querier.rs
```diff
@@ -59,6 +59,18 @@ pub struct Config {
...
+    /// ingester. For example "http://127.0.0.1:8083"
+    ///
+    /// Note this is a workround for
+    /// <https://github.com/influxdata/influxdb_iox/issues/3996>
+    #[clap(
+        long = "--ingester-address",
+        env = "INFLUXDB_IOX_INGESTER_ADDRESS",
+        default_value = "http://127.0.0.1:8083"
+    )]
+    pub ingester_address: String,
```
crepererum, line 73: You can get clap to handle lists for you:

https://github.com/influxdata/influxdb_iox/blob/91991eaecf08f93218a1e42845bd1f04f8d0961d/clap_blocks/src/write_buffer.rs#L30-L45

test_helpers_end_to_end_ng/src/config.rs
crepererum: I think having `must_use` for getters makes sense, IIRC this is also what the stdlib is aiming for.

querier/src/ingester/mod.rs
```diff
@@ -0,0 +1,196 @@
...
+
+        Ok(vec![])
+    }
+}
+
+/// A wrapper around the unpersisted data in a partition returned by
+/// the ingester that (will) implement the `QueryChunk` interface
+#[allow(missing_copy_implementations)]
+#[derive(Debug, Clone)]
+pub struct IngesterPartition {}
```
crepererum, line 118: Maybe state the scope and granularity of a "partition" here, because it slowly gets a bit confusing:

- it is bound to a specific `QuerierTable` (which is bound to a table but NOT a sequencer)
- there are 0 or 1 partitions for every catalog partition
- as a reminder: every catalog partition is bound to a catalog table AND sequencer

influxdb_iox/src/commands/run/querier.rs
crepererum: And this should probably be a plural form (addressES)

crepererum (approved):
Minor nitpicks, infra looks good. Thank your for already including the mock abstraction :+1:

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -91,6 +108,11 @@ impl TestConfig {
             .with_env("INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END", "0")
     }
 
+    /// Adds the ingester address
+    fn with_ingester_address(self, ingester_address: &str) -> Self {
```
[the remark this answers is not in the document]
alamb: Yes; My plan will be to get this PR polished up / merged and then make a fast follow on PR that extends the config system to handle lists

querier/src/namespace/query_access.rs
```diff
@@ -45,11 +48,16 @@ impl QueryDatabase for QuerierNamespace {
...
+                trace!(%table_name, "No entry for table");
                 return Ok(vec![]);
             }
         };
 
-        let mut chunks = table.chunks().await;
+        let mut chunks = table
+            .chunks(predicate)
+            .await
+            // TODO QuerierNamespace trait needs to be updated to return Result
```
alamb, line 59: https://github.com/influxdata/influxdb_iox/pull/4260

querier/src/ingester/mod.rs
alamb: I tried to clarify this in 4c8f7e895 but I am not sure I got it 100% right (as it will depend a bit on how you choose to actually implement the interface):

```
/// A wrapper around the unpersisted data in a partition returned by
/// the ingester that (will) implement the `QueryChunk` interface
///
/// Given the catalog heirarchy:
///
/// ```text
/// (Catalog) Sequencer -> (Catalog) Table --> (Catalog) Partition
/// ```
///
/// An IngesterPartition contains the unpersisted data for a catalog
/// partition from a sequencer. Thus, there can be more than one
/// IngesterPartition for each table the ingester knows about.
```

alamb:
rebased as merge confused semantic PR check

test_helpers_end_to_end_ng/src/config.rs
carols10cents: Neat, today i learned! 👍🏻

PR #4260, opened 2022-04-08
feat: `QuerierDatabase::chunks` returns `Result`

alamb:
# Rationale
Part of allowing the querier to include not-yet-persisted data from the ingester: https://github.com/influxdata/influxdb_iox/pull/4255

The querier needs to ask the ingester for data and that asking can fail (e.g. network communication problem). However the existing trait is infallible

# Changes
1. Change QuerierDatabase::chunks to return `Result`
2. Update code to handle this new possibility


query/src/lib.rs
```diff
@@ -140,7 +143,11 @@ pub trait QueryDatabase: QueryDatabaseMeta + Debug + Send + Sync {
     /// Returns a set of chunks within the partition with data that may match
     /// the provided predicate. If possible, chunks which have no rows that can
     /// possibly match the predicate may be omitted.
-    async fn chunks(&self, table_name: &str, predicate: &Predicate) -> Vec<Arc<dyn QueryChunk>>;
+    async fn chunks(
+        &self,
+        table_name: &str,
+        predicate: &Predicate,
+    ) -> Result<Vec<Arc<dyn QueryChunk>>, QueryDatabaseError>;
```
alamb, line 150: here is the change this PR exists for -- everything else is a consequence of this change

query/src/lib.rs
```diff
@@ -130,6 +130,9 @@ impl Drop for QueryCompletedToken {
 /// This avoids storing potentially large strings
 pub type QueryText = Box<dyn std::fmt::Display + Send + Sync>;
 
+/// Error type for [`QueryDatabase`] operations.
+pub type QueryDatabaseError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
alamb, line 134: Follows the model of `QueryChunkError` later in this file

query/src/frontend/influxrpc.rs
```diff
@@ -245,7 +250,11 @@ impl InfluxRpcPlanner {
         for (table_name, predicate) in &table_predicates {
             // Identify which chunks can answer from its metadata and then record its table,
             // and which chunks needs full plan and group them into their table
-            for chunk in database.chunks(table_name, predicate).await {
+            let chunks = database
+                .chunks(table_name, predicate)
+                .await
+                .context(GettingChunksSnafu)?;
```
carols10cents, line 256: Given that `GettingChunksSnafu` is used in multiple places, would any of the surrounding information be useful trying to figure out what was going on if this fails? Like the table name, the predicate, or the operation happening (in this case `table_names`)? Or would that information not be helpful? This can also wait til we have these errors and are trying to debug them if we don't know yet ;)

carols10cents (approved):
Just one question that may or may not be relevant right now, doesn't necessarily need to be resolved in this PR.

query/src/frontend/influxrpc.rs
alamb: I think more context would be useful. I will create a new PR to add them.

alamb: https://github.com/influxdata/influxdb_iox/pull/4263

PR #4261, opened 2022-04-08
fix: fix pprof

alamb:
Turns out when I reorganized the code in `ioxd` I broke pprof (or maybe it had been broken for a while). Noticed when testing internally

The [code for pprof](https://github.com/influxdata/influxdb_iox/blob/alamb%2Ffix_pprof/ioxd_common/src/http/mod.rs#L238) is in ioxd_common crate, but the flag is defined in `influxdb_iox` crate. 

When I ran:
```shell
LOG_FILTER="info,sqlx=warn,h2=warn" cargo run --features=pprof -- run compactor --write-buffer-partition-range-start=0 --write-buffer-partition-range-end=0 --catalog-dsn="postgresql://localhost:5432/alamb"
```

And then visited http://localhost:8080/debug/pprof/profile?seconds=5

I got a

```
{"code":"internal error","message":"pprof support is not compiled"}
```

Despite clearly saying "features=pprof" lol

# Changes
Use feature flags correctly

Now it works great:
Screen Shot 2022-04-08 at 1 37 39 PM


NGA-TRAN (approved):
The change seems reasonable to me

mkmik (commented):
Thanks,!

Issue #4124, opened 2022-03-24
Querier: Remove binary parquet metadata from catalog


crepererum:
The querier currently needs the binary (thrift + zstd) parquet file metadata to create parquet chunks. However these binary blobs are quite large and it would be nice if the querier could work w/o it. There are multiple options:

- the querier fetches (and caches) the blobs from the actual parquet files on-demand
- we serialize enough metadata/stats within the catalog so the querier could work reasonably well

It would be helpful if we would have some real world perf testing in place before working on this issue so that we can judge how bad "fetch on demand" would be.

carols10cents:
The first step I'm going to take here is to continue *storing* the parquet metadata in the catalog, but not *fetching* it when doing catalog queries.

domodwyer:
I have discovered the compactor actually depends on this metadata too:

* There is an explicit catalog method that returns parquet info with the metadata: [`list_by_partition_not_to_delete_with_metadata()`][catalog]
* This is called by the compactor [here][call]
* Before passing the result into [`compact()`][compact]
* Where the [`IoxMetadata` is resolved][ioxmeta]
* By calling [`ParquetFileWithTombstone::iox_metadata()`][withtombstone]
* Which [uses the embedded metadata](https://github.com/influxdata/influxdb_iox/blob/df7b3c3a88091cd920a3d16cd3d37eabb77e2075/parquet_file/src/chunk.rs#L214-L216)

At the time of writing, the following fields in the resolved `IoxMetadata` appear to be used:

* [partition ID](https://github.com/influxdata/influxdb_iox/blob/df7b3c3a88091cd920a3d16cd3d37eabb77e2075/compactor/src/compact.rs#L607)
* [table name + sort key](https://github.com/influxdata/influxdb_iox/blob/df7b3c3a88091cd920a3d16cd3d37eabb77e2075/compactor/src/compact.rs#L614-L615)
* [sequencer ID, namespace ID, namespace name, partition key](https://github.com/influxdata/influxdb_iox/blob/df7b3c3a88091cd920a3d16cd3d37eabb77e2075/compactor/src/compact.rs#L689-L695)

All of this information can be accessed via the catalog directly - it appears using this emebeded metadata is unnecessary.


[catalog]: https://github.com/influxdata/influxdb_iox/blob/df7b3c3a88091cd920a3d16cd3d37eabb77e2075/iox_catalog/src/postgres.rs#L1699-L1715

[...]

PR #4262, opened 2022-04-08
feat: Store parquet_metadata in catalog, but don't fetch it by default

carols10cents:
Connects to #4124.

This is an incremental step to stop fetching the parquet_metadata column from the parquet_files table for most parquet_files queries, which should help with performance.

The places that do use the metadata are still fetching it and decoding it for now. For one place, I made a special `_with_metadata` catalog method. For the other place, I'm doing more queries to get each parquet file's metadata column. It was slightly easier to do each place this way, but not impossible to change either place to use the other technique, so let me know if you'd like me to make these consistent.

However, the next steps should be trying to move to using other catalog data instead of the parquet_metadata, so that change may be moot. These next steps are not trivial 😅 

If we end up removing all uses of the `parquet_metadata` column, then we can stop saving it in the catalog altogether, but this PR isn't there yet. If we decide to keep using `parquet_metadata`, I might recommend storing it in a different catalog table than `parquet_files` so that we don't have to remember not to do `SELECT *` on `parquet_files`.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -10,8 +10,8 @@ use arrow::record_batch::RecordBatch;
 use backoff::{Backoff, BackoffConfig};
 use bytes::Bytes;
 use data_types2::{
-    ParquetFile, ParquetFileId, PartitionId, SequencerId, TableId, TablePartition, Timestamp,
-    Tombstone, TombstoneId,
+    ParquetFile, ParquetFileId, ParquetFileWithMetadata, PartitionId, SequencerId, TableId,
+    TablePartition, Timestamp, Tombstone, TombstoneId,
 };
 use datafusion::error::DataFusionError;
 use iox_catalog::interface::{Catalog, Transaction};
@@ -19,10 +19,7 @@ use iox_object_store::ParquetFilePath;
 use metric::{Attributes, Metric, U64Counter, U64Gauge, U64Histogram, U64HistogramOptions};
 use object_store::DynObjectStore;
 use observability_deps::tracing::{debug, info, warn};
-use parquet_file::{
-    chunk::DecodedParquetFile,
-    metadata::{IoxMetadata, IoxParquetMetaData},
-};
+use parquet_file::metadata::{IoxMetadata, IoxParquetMetaData};
 use query::{
     compute_sort_key_for_chunks, exec::ExecutorType, frontend::reorg::ReorgPlanner,
     util::compute_timenanosecond_min_max,
@@ -379,7 +376,7 @@ impl Compactor {
             .repositories()
             .await
             .parquet_files()
-            .list_by_partition_not_to_delete(partition_id)
+            .list_by_partition_not_to_delete_with_metadata(partition_id)
             .await
             .context(ListParquetFilesSnafu)?;
         if parquet_files.is_empty() {
@@ -491,7 +488,7 @@ impl Compactor {
     fn group_small_contiguous_groups(
         mut file_groups: Vec<GroupWithMinTimeAndSize>,
         compaction_max_size_bytes: i64,
-    ) -> Vec<Vec<ParquetFile>> {
+    ) -> Vec<Vec<ParquetFileWithMetadata>> {
         let mut groups = Vec::with_capacity(file_groups.len());
         if file_groups.is_empty() {
             return groups;
@@ -558,19 +555,8 @@ impl Compactor {
             return Ok(compacted);
         }
 
-        // Fetch the parquet metadata for the first file to reuse IDs and names
-        let parquet_metadata_bytes = {
-            let mut repos = self.catalog.repositories().await;
-
-            repos
-                .parquet_files()
-                .parquet_metadata(overlapped_files[0].data.id)
-                .await
-                .context(ParquetMetadataSnafu)?
-        };
-        let decoded_metadata =
-            DecodedParquetFile::new(*overlapped_files[0].data, parquet_metadata_bytes);
-        let iox_metadata = decoded_metadata.iox_metadata;
+        // Save the parquet metadata for the first file to reuse IDs and names
+        let iox_metadata = overlapped_files[0].iox_metadata();
 
         //  Collect all unique tombstone
         let mut tombstone_map = overlapped_files[0].tombstones();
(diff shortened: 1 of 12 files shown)


alamb:
I plan to review this later this afternoon

data_types2/src/lib.rs
```diff
@@ -720,7 +720,7 @@ pub fn tombstones_to_delete_predicates_iter(
 }
 
 /// Data for a parquet file reference that has been inserted in the catalog.
-#[derive(Debug, Clone, PartialEq, sqlx::FromRow)]
+#[derive(Debug, Clone, Copy, PartialEq, sqlx::FromRow)]
```
alamb, line 723: 👍  for making it copy once the metadata is no lonegr inlined

iox_catalog/src/postgres.rs
```diff
@@ -1502,9 +1502,13 @@ RETURNING *;
         sequencer_id: SequencerId,
         sequence_number: SequenceNumber,
     ) -> Result<Vec<ParquetFile>> {
+        // Deliberately doesn't use `SELECT *` to avoid the performance hit of fetching the large
```
alamb, line 1505: 👍

iox_catalog/src/metrics.rs
```diff
@@ -271,10 +271,12 @@ decorate!(
         "parquet_list_by_table_not_to_delete" = list_by_table_not_to_delete(&mut self, table_id: TableId) -> Result<Vec<ParquetFile>>;
         "parquet_delete_old" = delete_old(&mut self, older_than: Timestamp) -> Result<Vec<ParquetFile>>;
         "parquet_list_by_partition_not_to_delete" = list_by_partition_not_to_delete(&mut self, partition_id: PartitionId) -> Result<Vec<ParquetFile>>;
+        "parquet_list_by_table_not_to_delete_with_metadata" = list_by_table_not_to_delete_with_metadata(&mut self, table_id: TableId) -> Result<Vec<(ParquetFile, Vec<u8>)>>;
         "parquet_level_0" = level_0(&mut self, sequencer_id: SequencerId) -> Result<Vec<ParquetFile>>;
         "parquet_level_1" = level_1(&mut self, table_partition: TablePartition, min_time: Timestamp, max_time: Timestamp) -> Result<Vec<ParquetFile>>;
         "parquet_update_to_level_1" = update_to_level_1(&mut self, parquet_file_ids: &[ParquetFileId]) -> Result<Vec<ParquetFileId>>;
         "parquet_exist" = exist(&mut self, id: ParquetFileId) -> Result<bool>;
+        "parquet_parquet_metadata" = parquet_metadata(&mut self, id: ParquetFileId) -> Result<Vec<u8>>;
```
alamb, line 279: such `parquet`!

I wonder if this should be `parquet_metadata` rather than the double `parquet_parquet_metadata`?

alamb (approved):
I really like the idea of this incremental evolution 👍 nice work @carols10cents 

Someone with more experience with the catalog  and compactor code should probably review this, but it looks great to me

crepererum (approved):
Good first step! :)

compactor/src/compact.rs
```diff
@@ -581,15 +600,23 @@ impl Compactor {
         }
 
         // Convert the input files into QueryableParquetChunk for making query plan
-        let query_chunks: Vec<_> = overlapped_files
-            .iter()
-            .map(|f| {
-                f.to_queryable_parquet_chunk(
+        let mut query_chunks = Vec::with_capacity(overlapped_files.len());
```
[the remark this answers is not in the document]
carols10cents: Yeah, that would be annoying to `zip` back together again though. I think I'm going to do the same thing here for `list_by_partition_not_to_delete` that I did for  `list_by_table_not_to_delete`; adding a `_with_metadata` version.

iox_catalog/src/metrics.rs
carols10cents: Sure! I think this is just the metric name so it can be whatever.

compactor/src/compact.rs
```diff
@@ -371,7 +376,7 @@ impl Compactor {
             .repositories()
             .await
             .parquet_files()
-            .list_by_partition_not_to_delete(partition_id)
+            .list_by_partition_not_to_delete_with_metadata(partition_id)
```
NGA-TRAN, line 379: We do not need `column metadata` during compaction. I think most places do not need them. Do you think you can modify the catalog (Postgres and memory) to replace `select *` with select specific columns? Otherwise, the compactor performance will suffer here.

compactor/src/compact.rs
```diff
@@ -819,7 +824,9 @@ impl Compactor {
 
     /// Given a list of parquet files that come from the same Table Partition, group files together
     /// if their (min_time, max_time) ranges overlap. Does not preserve or guarantee any ordering.
-    fn overlapped_groups(mut parquet_files: Vec<ParquetFile>) -> Vec<GroupWithMinTimeAndSize> {
+    fn overlapped_groups(
+        mut parquet_files: Vec<ParquetFileWithMetadata>,
```
NGA-TRAN, line 828: Same here and all other places in the compactor, we do not want `ParquetFileWithMetadata`. `ParquetFile` is good enough help with better performance and less memory usage

compactor/src/compact.rs
```diff
@@ -951,7 +958,7 @@ impl Compactor {
 
     async fn add_tombstones_to_groups(
         &self,
-        groups: Vec<Vec<ParquetFile>>,
+        groups: Vec<Vec<ParquetFileWithMetadata>>,
```
NGA-TRAN, line 961: Same

compactor/src/compact.rs
```diff
@@ -1339,7 +1342,9 @@ mod tests {
             .unwrap();
 
         // Should have 3 non-soft-deleted files: pf1 not compacted and stay, and 2 newly created after compacting pf2, pf3, pf4
-        let files = catalog.list_by_table_not_to_delete(table.table.id).await;
+        let mut files = catalog
+            .list_by_table_not_to_delete_with_metadata(table.table.id)
```
NGA-TRAN, line 1346: I know this is only  a test but if you have `list_by_table_not_to_delete`, use it instead

compactor/src/compact.rs
```diff
@@ -1702,16 +1703,16 @@ mod tests {
...
     }
 
     fn arbitrary_parquet_file_with_size(
         min_time: i64,
         max_time: i64,
         file_size_bytes: i64,
-    ) -> ParquetFile {
-        ParquetFile {
+    ) -> ParquetFileWithMetadata {
+        ParquetFileWithMetadata {
```
NGA-TRAN, line 1715: same here. No metadata needed

NGA-TRAN (changes requested):
The PR looks good. We only need to not use `withMetadata` in the Compactor

NGA-TRAN (approved):
Ah, It makes sense. Looks good then

PR #4264, opened 2022-04-08
fix: not to add IOxReadFilterNode for empty non-duplicated chunks

NGA-TRAN:
This bug was introduced [here](https://github.com/influxdata/influxdb_iox/commit/8a8de19fb5935045b09477dd71dd51a0d38b2c65#diff-e02d4dfd33c9ea99b3b34d4be7da3b133cca551d23f64f89bcf7b29b6cc33d80R444) that adds `IOxReadFilterNode` for empty non-duplicated chunks.

Note that the plan must have one `IOxReadFilterNode` (scan node) if there is nothing to scan. However, in the case there are already `IOxReadFilterNode` to scan some data, there is no reason to add extra empty scan which will make unnecessary work and may confuse DF.

**A compact plan before the fix:**
```
StreamSplitExec
  SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC]
    UnionExec
      DeduplicateExec: [tag1@1 ASC,time@2 ASC]
        SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC]
          UnionExec
            SortExec: [tag1@1 ASC,time@2 ASC]
              CoalesceBatchesExec: target_batch_size=500
                FilterExec: time@2 < 6000 OR time@2 > 12000 OR NOT CAST(tag1@1 AS Utf8) = VT
                  IOxReadFilterNode: table_name=table, chunks=1 predicate=Predicate
      IOxReadFilterNode: table_name=table, chunks=0 predicate=Predicate
``` 

**The same compact plan after the fix:**
```
StreamSplitExec
  SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC]
    DeduplicateExec: [tag1@1 ASC,time@2 ASC]
      SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC]
        UnionExec
          SortExec: [tag1@1 ASC,time@2 ASC]
            CoalesceBatchesExec: target_batch_size=500
              FilterExec: time@2 < 6000 OR time@2 > 12000 OR NOT CAST(tag1@1 AS Utf8) = VT
                IOxReadFilterNode: table_name=table, chunks=1 predicate=Predicate
```

Before this fix, there will be always extra `IOxReadFilterNode: table_name=table, chunks=0 predicate=Predicate` and `UnionExec` in ALL query plans (yes all queries in Ingester, Compactor, Querier and even OG's Query) if some input chunks overlap/duplicate. After this fix, they will go away.

Extra/unrelated note for @tustvold, @alamb and myself: Why there are 2 `SortPreservingMergeExec`. I think the one on the top is unnecessary


query/src/provider.rs
```diff
@@ -476,15 +478,20 @@ impl Deduplicater {
...
-                Arc::clone(&table_name),
-                Arc::clone(&output_schema),
-                self.no_duplicates_chunks.to_vec(),
-                predicate,
-                output_sort_key.as_ref(),
-            )?;
-            plans.append(&mut non_duplicate_plans);
+            if !self.no_duplicates_chunks.is_empty() {
+                trace!(
+                    "Build one scan node for the rest of neither-duplicated-nor-ovelapped chunks."
```
carols10cents, line 483: Typo in this trace message, could affect searching for particular logs:

```suggestion
                    "Build one scan node for the rest of neither-duplicated-nor-overlapped chunks."
```

query/src/provider.rs
```diff
@@ -426,6 +427,7 @@ impl Deduplicater {
             )?;
             plans.append(&mut non_duplicate_plans);
         } else {
+            trace!("Some of the chunks either overlap or duplicate or both. There should be one scan node for each overlapped or duplicated chunk.");
```
alamb, line 430: maybe you can combine the `trace` calls

alamb (commented):
Given this is so tricky -- I wonder if we can make a test that covers it?

pauldix (approved):

query/src/provider.rs
```diff
@@ -1009,6 +1016,7 @@ impl Deduplicater {
         // Only chunks without delete predicates should be in this one IOxReadFilterNode
         // if there is no chunk, we still need to return a plan
```
carols10cents, line 1017: Hm, I wonder why this comment says "if there is no chunk, we still need to return a plan"? And there's a check for `chunks.is_empty()` on the next line? Is this comment still valid?

query/src/provider.rs
```diff
@@ -416,6 +416,7 @@ impl Deduplicater {
         let mut plans: Vec<Arc<dyn ExecutionPlan>> = vec![];
         if self.no_duplicates() {
             // Neither overlaps nor duplicates, no deduplicating needed
+            trace!("All chunks neither overlap nor duplicate. Build only one scan node for all of them.");
             let mut non_duplicate_plans = Self::build_plans_for_non_duplicates_chunks(
```
carols10cents, line 420: This is the other place `build_plans_for_non_duplicates_chunks` is called, do we need to check for `chunks.is_empty()` here too?

Could the other methods that look similar to `build_plans_for_non_duplicates_chunks`, like `build_deduplicate_plan_for_overlapped_chunks` and `build_deduplicate_plan_for_chunk_with_duplicates` cause this same problem?

query/src/provider.rs
NGA-TRAN: Done

carols10cents (commented):
If this is super urgent to get deployed to get some metrics on, feel free to punt my comments/questions to future PRs. I don't know the answers to these questions, maybe Raphael does?

I agree with Andrew that a test to check the query plan exactly would be nice if it's possible?

NGA-TRAN:
@alamb 
> Given this is so tricky -- I wonder if we can make a test that covers it?

The tests are tricky and I cannot think about a quick one to add. Let me have this merged first and think how to capture tests with and without `chunks=0`. I am thinking about adding these tests:

1. Similar to this test but have no [last line](https://github.com/influxdata/influxdb_iox/blob/bbbdcc75a86e5ebae5a9a3db7501c4f23ceb7bf1/query_tests/cases/in/duplicates.expected#L67) if there are no non-duplicated chunks
2. Mark [this test](https://github.com/influxdata/influxdb_iox/blob/bbbdcc75a86e5ebae5a9a3db7501c4f23ceb7bf1/query_tests/cases/in/pushdown.expected#L283) is correct with `chunks=0`

query/src/provider.rs
NGA-TRAN: @carols10cents 
> This is the other place build_plans_for_non_duplicates_chunks is called, do we need to check for chunks.is_empty() here too?

No, this is the place we accept empty chunks because we need at least one scan node. There is a comment inside `build_plans_for_non_duplicates_chunks` for that.

> Could the other methods that look similar to build_plans_for_non_duplicates_chunks, like build_deduplicate_plan_for_overlapped_chunks and build_deduplicate_plan_for_chunk_with_duplicates cause this same problem?

These plans should only be built on non-empty chunks and, so far, we have done that through `for` loop. This bug was introduced when we remove the 'for' loop :)

We can make the comments clearer as needed

query/src/provider.rs
NGA-TRAN: Yes, it is valid. We can make this clearer by create 2 different functions for this

NGA-TRAN:
@carols10cents and @alamb : I  will create a  ticket to add tests and comments or even make a new function to make this clearer and avoid similar bugs in the future

NGA-TRAN:
Created #4265 for this

PR #4271, opened 2022-04-11
feat: allow the querier to talk with multiple ingesters

alamb:
# Rationale
The long term plan is that we will configure the querier with the sequencer ranges each ingester is responsible for as described in https://github.com/influxdata/influxdb_iox/issues/4144 and the [design doc](https://docs.google.com/document/d/1Rosq-KnXDeDcVxvWYPw9mLWktSMAgjHZUqBpizE6VGg/edit#heading=h.zfpde3tjzb0i).

In order to unblock querying unpersisted data in production faster, in the interim the idea is to query *all* the ingesters and then combine the results somehow. This process will be optimized as part of https://github.com/influxdata/influxdb_iox/issues/4144.

See related discussion: https://github.com/influxdata/influxdb_iox/pull/4259#discussion_r846133584

# Changes

(kudos to @crepererum https://github.com/influxdata/influxdb_iox/pull/4259#discussion_r847185655 for pointing me at an example of how to do this)

1. Changes `INFLUXDB_IOX_INGESTER_ADDRESS` to `INFLUXDB_IOX_INGESTER_ADDRESSES` (plural)
2. Allow multiple addresses (e.g. `INFLUXDB_IOX_INGESTER_ADDRESSES=http://10.10.50.10:8082,http://10.10.50.20:8082,http://10.10.50.30:8082`)
3. Tests for same

I plan a new end to end test as a follow on PR


clap_blocks/src/querier.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Note this is a workround for
+    /// <https://github.com/influxdata/influxdb_iox/issues/3996>
+    #[clap(
+        long = "--ingester-address",
+        env = "INFLUXDB_IOX_INGESTER_ADDRESSES",
+        default_value = "http://127.0.0.1:8083",
+        multiple_values = true,
+        use_value_delimiter = true
+    )]
+    pub ingester_addresses: Vec<String>,
```
crepererum, line 24: One potential config error that can happen here (but that we have in other places of our stack as well) is that we would name the same ingester multiple times and the querier silently accepts it and potentially returns duplicate rows (because it won't perform any cross-ingester de-dup). If we wanna fix that then I would opt for bailing out instead of silently converting this vector to a set.

crepererum (approved):

clap_blocks/src/querier.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[clap(long = "--num-query-threads", env = "INFLUXDB_IOX_NUM_QUERY_THREADS")]
+    pub num_query_threads: Option<usize>,
+
+    /// gRPC address for the querier to talk with the ingester. For
+    /// example "http://127.0.0.1:8083" or
+    /// "http://10.10.10.1:8083,http://10.10.10.2:8083" for multiple
+    /// addresses.
+    ///
+    /// Note this is a workround for
+    /// <https://github.com/influxdata/influxdb_iox/issues/3996>
```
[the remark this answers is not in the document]
alamb: Good point -- will update wording

clap_blocks/src/querier.rs
alamb: I will add some error checking

clap_blocks/src/querier.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Get the querier config's num query threads.
+    #[must_use]
+    pub fn num_query_threads(&self) -> Option<usize> {
+        self.num_query_threads
+    }
+
+    /// Return the querier config's ingester addresses. Returns `Err`
+    /// if any of the addresses is repeated
+    #[must_use]
+    pub fn ingester_addresses(&self) -> Result<Vec<String>, String> {
```
alamb, line 39: I couldn't find any pre-existing way in clap to do this -- the way it handled custom `parse` implementations appears to be value-by-value so there is no way to do the validation until the entire structure is parsed.

Thus I made the fields of this struct non `pub` and did the validation in the accessor

https://github.com/clap-rs/clap/blob/v3.1.8/examples/derive_ref/README.md#arg-types

clap_blocks/src/querier.rs
alamb: 6e57fee1a2a84516f53c4ed9e46baf4e5f062291

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -108,9 +108,9 @@ impl TestConfig {
             .with_env("INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END", "0")
     }
 
-    /// Adds the ingester address
-    fn with_ingester_address(self, ingester_address: &str) -> Self {
-        self.with_env("INFLUXDB_IOX_INGESTER_ADDRESS", ingester_address)
+    /// Adds the ingester addresses (comma separated hostnames)
+    fn with_ingester_addresses(self, ingester_addresses: &str) -> Self {
```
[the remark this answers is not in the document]
alamb: in 609f186330eeebe69f981f57b613b7f09bd6fe89

clap_blocks/src/querier.rs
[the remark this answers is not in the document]
alamb: 44045cb

PR #4273, opened 2022-04-11
refactor: use a single flight client implementation

crepererum:
"end-user -> querier" and "querier -> ingester" should use a single
Flight client implementation. The difference is just the request and
response metadata.

This changes our default Flight client to use protobuf instead of JSON
for the ticket format.


generated_types/protos/influxdata/iox/querier/v1/flight.proto
```diff
@@ -0,0 +1,14 @@
...
+message ReadInfo {
+  // Namespace(/database) name.
+  string namespace_name = 1;
+
+  // SQL query.
+  string sql_query = 2;
+}
+
+// Response in "end-user to querier" flight response.
+message AppMetadata {}
```
jacobmarble, line 14: Please briefly describe usage of this empty message type for external client implementations.

influxdb_iox/tests/end_to_end_cases/write_api.rs
```diff
@@ -583,15 +602,21 @@ async fn test_write_routed_no_shard() {
 
     assert!(target_1
         .flight_client()
-        .perform_query(&db_name_1, "select * from disk")
+        .perform_query(ReadInfo {
```
alamb, line 605: maybe it would be ergonomic to provide a constructor so this could be `ReadInfo::new(&db_name_1, "select * from disk")`

jacobmarble:
@crepererum do you anticipate this will break an outside client implementation?

FYI [there exists a golang client](https://github.com/influxdata/influxdb-iox-client-go) that I maintain. For the next few months, I think it's appropriate to "move fast and break things", but do at-mention @jacobmarble and @wolffcm so we are aware of changes to the protocol. We can update the golang client asynchronously.

influxdb_iox_client/src/client/flight.rs
```diff
@@ -55,14 +63,42 @@ pub enum Error {
...
+    #[error(transparent)]
+    Serialization(#[from] prost::EncodeError),
+
+    /// Deserializing the protobuf structs from bytes failed.
+    #[error(transparent)]
+    Deserialization(#[from] prost::DecodeError),
+}
+
+/// Metadata that can be send during flight requests.
+pub trait ClientMetadata: Message {
```
alamb, line 77: I wonder if we could retain backwards compatibility with the old Json encoding by making this trait not derive from `prost` but unstead have functions:

```rust
trait ClientMetadata {
  fn to_bytes(&self) -> Vec<u8>;
  fn from_bytes(&[u8]) -> Self;
}
```

Which would allow us to retain backwards compatibility with the old message format (and potentially not have to change `queryd` / `flux` for a while)

influxdb_iox_client/src/client/flight.rs
```diff
@@ -84,26 +123,30 @@ pub enum Error {
 /// # }
 /// ```
```
alamb, line 124: It might be nice to explain here what the parameter `ClientMetada` is used for (namely the format of metadata exchanged with the server) and that users of this code should probably use `ReadInfo` unless they have their own specialized needs

influxdb_iox_client/src/client/flight.rs
```diff
@@ -130,40 +173,40 @@ impl Client {
...
+{
+    pub(crate) async fn new<R>(flight: &mut Client<R>, request: R) -> Result<Self, Error>
+    where
+        R: ClientMetadata<Response = T>,
+    {
+        let mut bytes = bytes::BytesMut::new();
+        prost::Message::encode(&request, &mut bytes)?;
         let t = Ticket {
-            ticket: serde_json::to_string(&query)?.into(),
+            ticket: bytes.to_vec(),
```
alamb, line 201: this is the non backwards compatible change, right? How the bytes in `ReadInfo` are serialized

querier/src/flight.rs
```diff
@@ -1,227 +0,0 @@
-use arrow::{
```
alamb, line 1: ❤️

alamb (changes requested):
I really like this reduction in redundancy 🏅  -- mostly as the flight API is complicated enough with having two sort-of-but-not-quite-the-same implementations 👍 

My only concern with this PR is that it may break existing users (e.g.  the flux team and maybe others who have done some SQL implementation). @e-dard  and maybe @wolffcm probably know more. 

Before we merge it I think we need to update the code to continue to accept the old JSON encoded metadata format as well so that we can roll this out in backwards compatible way: ensuring all clients are migrated to the new format, and then removing support for the old format once all clients are upgraded. 

(as an aside My longer term hope / plan is to use FlightSQL as the public API to IOx, but that has no timeline yet)

alamb (approved):
Approving per @jacobmarble 's note that it is ok to break backwards compatibility 👍

influxdb_iox_client/src/client/flight.rs
crepererum: Yes, we change it from JSON to protobuf.

influxdb_iox_client/src/client/flight.rs
crepererum: done

influxdb_iox_client/src/client/flight.rs
crepererum: queryd and flux don't use the flight interface. They use the InfluxRPC / storage API which is NOT affected by this change.

generated_types/protos/influxdata/iox/querier/v1/flight.proto
crepererum: done

influxdb_iox_client/src/client/flight.rs
alamb: I think there was some experimental work to send SQL queries directly to IOx from flux (though you are correct that the main interface uses Influx storage RPC API)

PR #4279, opened 2022-04-12
feat: tokio console support

domodwyer:
This PR adds support for the [tokio-console](https://github.com/tokio-rs/console).

> wow! whoa! it's like `top(1)` for tasks!

Screenshot 2022-04-12 at 12 02 03

This screenshot expertly shows nothing is happening on this IOx instance running the all-in-one mode.

This is part of my plan to help diagnose the compactor woes, and I will follow up with another PR to add docs once I've got it running in prod :+1:

---

* refactor: enable log / tracing composition (48b15a26)

      Changes the trogging::Builder to return a tracing Layer impl when build() is
      called, so that it can be composed together with other Layer impls rather than
      forcing complete control of the tracing pipeline.

      Callers of the Builder::install_global() (basically everything) are unaffected
      by this change.

* feat: tokio-console support (f08c8373)

      This commit adds an optional feature (disabled by default) to the IOx server
      binary named "tokio_console". When enabled, it adds support for the
      tokio-console to IOx:

          https://github.com/tokio-rs/console

      Enabling this feature drops the compile-time log filter to TRACE, and enables
      the necessary dependencies to support the instrumentation needed for the
      console to function.

      Unfortunately, this feature uses tonic 0.7 (latest) while we use tonic 0.6, so
      we wind up with two tonic versions being compiled in when this feature is
      enabled.

* ci: enable tokio_console feature (3c89d2a3)

      Configures the release CI pipeline to generate a binary with tokio-console
      support.

      This drops the compile-time log filter to TRACE, which may have some minor
      overhead. All our log filters in prod are debug or higher, so the runtime
      filter will remove the TRACE events and no additional logs will be emitted
      because of this change.

* chore: workspace-hack update (a4df93e6)


.circleci/config.yml
```diff
@@ -422,7 +422,7 @@ jobs:
       # https://github.com/influxdata/influxdb_iox/pull/2119
       ROARING_ARCH: "haswell"
       # Use avx512 instructions to take full advantage of the CPUs instruction set
-      RUSTFLAGS: "-C target-feature=+avx2 -C link-arg=-fuse-ld=lld"
+      RUSTFLAGS: "-C target-feature=+avx2 -C link-arg=-fuse-ld=lld --cfg tokio_unstable"
```
crepererum, line 425: I'm not entirely happy about the usage of the unstable flag here, but I guess the advantage of having a debug tool outweighs the risk.

crepererum (approved):

alamb:
Thanks @domodwyer  -- looking forward to seeing how this works ❤️

PR #4281, opened 2022-04-12
fix: accept legacy flight requests

crepererum:
This removes the SERVER-side breaking change introduced in #4273 by also
accepting JSON-based tickets.

Closes #4280.

### DIFF
--- a/service_grpc_flight/src/lib.rs
+++ b/service_grpc_flight/src/lib.rs
@@ -21,6 +21,7 @@ use generated_types::influxdata::iox::querier::v1 as proto;
 use pin_project::{pin_project, pinned_drop};
 use prost::Message;
 use query::{QueryCompletedToken, QueryDatabase};
+use serde::Deserialize;
 use service_common::QueryDatabaseProvider;
 use snafu::{ResultExt, Snafu};
 use tokio::task::JoinHandle;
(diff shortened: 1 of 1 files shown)


service_grpc_flight/src/lib.rs
```diff
@@ -111,6 +125,34 @@ impl Error {
...
 
+#[derive(Deserialize, Debug)]
+/// Body of the `Ticket` serialized and sent to the do_get endpoint.
+struct ReadInfo {
+    database_name: String,
+    sql_query: String,
+}
+
+impl ReadInfo {
+    fn decode_json(ticket: &[u8]) -> Result<Self> {
```
alamb, line 136:
```suggestion
    /// decode "legacy" format ReadInfo that was encoded as JSON (not protobuf)
    fn decode_json(ticket: &[u8]) -> Result<Self> {
```

alamb (commented):
Looks like it should work to me 👍  might be worth a test to document the expected behavior (perhaps as a follow on PR). I think we should merge this one asap to unblock IDPE / flux stuff

alamb (approved):

alamb:
It might also be worth a ticket to track remove this old legacy format (filed under tech debt, perhaps)

crepererum:
Removal ticket: https://github.com/influxdata/influxdb_iox/issues/4282

PR #4284, opened 2022-04-12
feat: static jaeger span tags

domodwyer:
Allows us to emit more descriptive spans by annotating them with a per-process set of static key=value pairs. 

This functionality is commonly used to emit tags like `hostname=myserver.local` or `build=<sha>`. It's also commonly [exposed via the `JAEGER_TAGS` env](https://www.jaegertracing.io/docs/1.32/client-features/), but we've already broken away from common envs it seems.

---

* feat: support static tracing tags (edc52181)

      Adds support for annotating all Jaeger tracing spans with a set of static
      process tags.

* feat: enable static Jaeger span tags (aa80776d)

      Exposes the static process tags via a CLI flag ("--traces-jaeger-tags") or env
      var ("TRACES_EXPORTER_JAEGER_TAGS") that accepts a key=value, comma-delimited
      tag string.

### DIFF
--- a/trace_exporters/src/jaeger.rs
+++ b/trace_exporters/src/jaeger.rs
@@ -1,8 +1,11 @@
-use std::net::{SocketAddr, ToSocketAddrs, UdpSocket};
+use std::{
+    net::{SocketAddr, ToSocketAddrs, UdpSocket},
+    str::FromStr,
+};
 
 use async_trait::async_trait;
 
-use observability_deps::tracing::{error, info};
+use observability_deps::tracing::*;
 use trace::span::Span;
 
 use crate::export::AsyncExport;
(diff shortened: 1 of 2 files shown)


trace_exporters/src/jaeger.rs
```diff
@@ -61,16 +110,25 @@ impl JaegerAgentExporter {
...
             client,
             next_sequence: 0,
+            tags: None,
         })
     }
 
+    /// Annotate all spans emitted by this exporter with the specified static
+    /// tags.
+    pub fn with_tags(self, tags: &[JaegerTag]) -> Self {
+        debug!(?tags, "setting static jaeger span tags");
```
alamb, line 120:
```suggestion
        trace!(?tags, "setting static jaeger span tags");
```

alamb (approved):
Only thing I can think of would be to add an end to end test in https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_cases/tracing.rs

mkmik (approved):

domodwyer:
I attempted to add an e2e test, but inspecting the thrift-encoded binary to prove the tag was added is... difficult.

alamb:
>  but inspecting the thrift-encoded binary to prove the tag was added is... difficult.

You can see how I did it in the existing one -- the equivalent of `grep`!

domodwyer:
The tag strings appeared to be mangled in the binary output 🤷

alamb:
giphy

PR #4287, opened 2022-04-12
refactor: disable max_debug static tracing level

domodwyer:
This was fun.

Usually features are (read: should be) additive, but this is not the case with tracing, and the "trace" level logs were being suppressed in the final binary. Hakari only added to the fun by enabling the debug feature even when it was removed from the iox binary.

---

* refactor: disable max_debug static tracing level (63e2a522)

      The tracing "release_max_level_*" feature flags are not additive, and are hard
      to use - especially when the workspace-hack crate enables features behind your
      back!


alamb (approved):
We can always crank it back to debug if we think the overhead of having `trace` in the code is too high

domodwyer:
That is my plan once I'm done faffing around, and I'll include "drop static level to trace" in the docs I'll write up about the console.

PR #4289, opened 2022-04-12
fix: Do not specify a querier address by default

alamb:
Fixes https://github.com/influxdata/conductor/issues/929

# Rationale
It turns out that I broke production queries with https://github.com/influxdata/influxdb_iox/pull/4259 as now the querier defaults to `http://localhost:8082` as querier address. This is not correct on production.

# Changes
1. Default to no ingester address
2. Tests for same


influxdb_iox/tests/end_to_end_ng_cases/querier.rs
```diff
@@ -104,3 +104,53 @@ async fn basic_on_parquet() {
     ];
     assert_batches_sorted_eq!(&expected, &batches);
 }
+
+#[tokio::test]
+async fn basic_no_ingster_connection() {
```
alamb, line 109: I am making a follow on PR to refactor the test setup to reduce this duplication

PR #4300, opened 2022-04-13
fix: proper executor shutdown in querier

crepererum:
This is not a huge issue but might drain resources like file descriptors
during tests. The dedicated exuector also logs a warning.


alamb:
Thank you -- this warning has been annoying me

alamb (approved):

PR #4301, opened 2022-04-13
refactor: Use declarative steps to reduce duplication in end to end testing

alamb:
# Rationale

Second attempt / follow on from https://github.com/influxdata/influxdb_iox/pull/4290

There are several tests that are all mostly the same. "Don't Repeat Yourself" is a standard software engineering maxim. I plan to use this setup more as I write more tests (e.g. with multiple ingesters) 

A  list-like declarative test definition makes it easy for a test reader to follow along.

Also, who doesn't like a good framework?

# changes
Implement plan from @crepereum here: https://github.com/influxdata/influxdb_iox/pull/4290/files#r849312364


influxdb_iox/tests/end_to_end_ng_cases/querier.rs
```diff
@@ -24,37 +19,31 @@ async fn basic_ingester() {
...
+    StepTest::new(
+        &cluster,
+        vec![
+            Step::WriteLineProtocol(format!(
+                "{},tag1=A,tag2=B val=42i 123456\n\
+             {},tag1=A,tag2=C val=43i 123457",
+                table_name, table_name
+            )),
+            Step::WaitForReadable,
+            Step::AssertNotPersisted,
```
alamb, line 31: Note this is a new part of the test (to validate that the rows are not persisted), addressing @crepererum 's comment in https://github.com/influxdata/influxdb_iox/pull/4290#discussion_r849302783

influxdb_iox/tests/end_to_end_ng_cases/querier.rs
```diff
@@ -24,37 +19,31 @@ async fn basic_ingester() {
...
-    let batches = run_query(
-        sql,
-        cluster.namespace(),
-        cluster.querier().querier_grpc_connection(),
+    StepTest::new(
+        &cluster,
+        vec![
+            Step::WriteLineProtocol(format!(
+                "{},tag1=A,tag2=B val=42i 123456\n\
+             {},tag1=A,tag2=C val=43i 123457",
```
carols10cents, line 27: Nit: Maybe indent this second line so it lines up with the first line, to better see the differences in the data?

```suggestion
                "{},tag1=A,tag2=B val=42i 123456\n\
                 {},tag1=A,tag2=C val=43i 123457",
```

carols10cents (approved):
LGTM! One small nit you can take or leave, one idea for the future.

influxdb_iox/tests/end_to_end_ng_cases/querier.rs
alamb: fixed in 525634962

influxdb_iox/tests/end_to_end_ng_cases/querier.rs
```diff
@@ -24,37 +19,31 @@ async fn basic_ingester() {
...
+                table_name, table_name
+            )),
+            Step::WaitForReadable,
+            Step::AssertNotPersisted,
+            Step::Query {
+                sql: format!("select * from {}", table_name),
```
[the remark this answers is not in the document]
alamb: I actually think I might prefer inlining the tablename everywhere manually rather than programmatically constructing the queries. 🤔  I will think about this some more

PR #4303, opened 2022-04-13
fix: Revert "fix: Revert "feat: Use the sort key stored in the catalog during compaction

carols10cents:
Do not merge before wiping the catalog!

This reverts commit 7ddbf7c02516a2524d0d5d5c53aa531cd5ea1807.

Closes #4299 again.


alamb:
Marking as a draft to avoid someone getting excited and merging it by accident... Not that I would ever do that 🙄  lol

NGA-TRAN:
Since @mkmik is wiping the catalog, I am going to merge this so when we start running the clusters again, this will get picked up immediately

Issue #4298, opened 2022-04-13
Querier connected to multiple ingesters errors when one does not have information for the table


alamb:
__Steps to reproduce:__
Configure an IOx querier to talk to several ingesters, only one of which has data for that ingester

__Expected behaviour:__
I expect the query to complete

__Actual behaviour:__

I get a query error
```sql
797a0180da80497c_ce59a0ee2de3ed17> select * from statuses;
Error running remote query: status: InvalidArgument, message: "Error while planning query: Execution error: Join Error: channel closed", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Wed, 13 Apr 2022 11:50:09 GMT"} }
```

And a panic i the log
```
level=error msg="Thread panic" panic_info="panicked at 'error performing query: GrpcError(Status { code: Internal, message: \"Error while performing query: No Table Data found for the given namespace name 797a0180da80497c_ce59a0ee2de3ed17, table name statuses\", metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"server\": \"envoy\", \"x-envoy-upstream-service-time\": \"1\", \"date\": \"Wed, 13 Apr 2022 11:50:22 GMT\"} }, source: None })', querier/src/ingester/mod.rs:152:18" target="panic_logging" location="panic_logging/src/lib.rs:94" time=1649850622284618910
thread 'IOx Query Executor Thread' panicked at 'error performing query: GrpcError(Status { code: Internal, message: "Error while performing query: No Table Data found for the given namespace name 797a0180da80497c_ce59a0ee2de3ed17, table name statuses", metadata: MetadataMap { headers: {"content-type": "application/grpc", "server": "envoy", "x-envoy-upstream-service-time": "1", "date": "Wed, 13 Apr 2022 11:50:22 GMT"} }, source: None })', querier/src/ingester/mod.rs:152:18
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at ./rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at ./rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/result.rs:1749:5
[...]

PR #4309, opened 2022-04-13
test: add end to end querier test with multiple ingesters

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/4341~

# Rationale
Add an end to end test using a setup that is similar to how we have production configured: multiple ingesters and a single querier.

Re https://github.com/influxdata/influxdb_iox/issues/3958

Also is a test for https://github.com/influxdata/influxdb_iox/issues/4298

# Changes
1. Add an end to end that writes data to two different tables / ingesters and then reads the results back in the querier
2. Add a `Step::Custom` mechanism in the `StepTest` framework where we can add custom steps and this avoid many types of `Step`s that are only used in a single place


test_helpers_end_to_end_ng/src/steps.rs
```diff
@@ -57,8 +108,10 @@ impl<'a> StepTest<'a> {
...
         let Self { cluster, steps } = self;
 
-        // Tokens for all writes performed in this test
-        let mut write_tokens = vec![];
+        let mut state = StepTestState {
+            cluster,
+            write_tokens: vec![],
+        };
 
         for step in steps {
```
crepererum, line 116: Since you're now auto-generating steps, you probably should print out the step index (starting at 1) as well, otherwise debugging gets a bit painful.

test_helpers_end_to_end_ng/src/steps.rs
```diff
@@ -16,10 +17,52 @@ pub struct StepTest<'a> {
...
+/// use futures::future::BoxFuture;
+/// use test_helpers_end_to_end_ng::{MiniCluster, StepTestState};
+///
+/// let custom_function = Box::new(|cluster: &mut MiniCluster, state: &mut StepTestState| {
+///   async move {
+///     // Your potentially async code here
+///   }.boxed()
+/// });
+/// ```
+pub type FCustom = Box<dyn for<'b> Fn(&'b mut StepTestState) -> BoxFuture<'b, ()>>;
```
crepererum, line 65: I really like the type change and the usage example :+1:

test_helpers_end_to_end_ng/src/client.rs
```diff
@@ -162,6 +194,49 @@ pub fn all_persisted(res: &GetWriteInfoResponse) -> bool {
...
+
+    let kafka_partition_infos = partition_infos
+        .into_iter()
+        .map(|(_kafka_partition_id, info)| info)
+        .collect();
+
+    GetWriteInfoResponse {
+        kafka_partition_infos,
+    }
+}
```
crepererum, line 223: Who tests the tester? I think this tiny function (which also calls `merge_info` but doesn't do any IO) could have a small unit test as well.

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut cluster = MiniCluster::new()
+        .with_router2(router2_config)
+        .await
+        .with_ingester(ingester_config)
+        .await
+        // second ingester
+        .with_other(ingester2_config)
+        .await
+        .with_querier(querier_config)
+        .await;
```
crepererum, line 44: This is not specific for this PR but a more general concern regarding NG: We're currently not re-using clusters. Esp. when we have many CLI and API tests, this likely increases test times quite a lot. For OG we basically use just a few cached server processes and randomized databases (which would be namespaces for NG). Let's see if we need such an approach for NG as well. IMHO it would be better if the NG cluster creation would be fast enough to avoid that hack.

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
```diff
@@ -0,0 +1,160 @@
...
+    // read all the data back out
+    .chain((0..100).map(|i| Step::VerifiedQuery {
+        sql: format!("select * from table_{}", i),
+        verify: Box::new(move |batches: Vec<RecordBatch>| {
+            // results look like this:
+            // "+------+------+--------------------------------+-----+",
+            // "| tag1 | tag2 | time                           | val |",
+            // "+------+------+--------------------------------+-----+",
+            // "| A    | B    | 1970-01-01T00:00:00.000123456Z | val |",
+            // "+------+------+--------------------------------+-----+",
```
crepererum, line 107: I ran into this issue in the past as well. While I like the readability of the text-based tables, I would really like to see some DSL/macro magic instead that is properly typed, is as easy to read as the text-based approach and accepts variables.

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
```diff
@@ -0,0 +1,160 @@
...
+                } else {
+                    println!("retrying, not yet readable: {:?}", combined_response);
+                }
+            }
+            Err(e) => {
+                println!("retrying, error getting token status: {}", e);
+            }
+        }
+        interval.tick().await;
+    }
```
crepererum, line 159: I would prefer a `tokio::time::timeout` instead of an endless loop. "retries forever" tests are a bit of a pain because they never fail and in the end you just have to hope for a CI timeout.

crepererum (approved):

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
alamb: 👍  created https://github.com/influxdata/influxdb_iox/issues/4366 to track

test_helpers_end_to_end_ng/src/client.rs
alamb: Done in bfc7de1f8

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
alamb: It is quite subtle but the call to this function has a timeout when this is used above, added by:

>                     .with_timeout_panic(Duration::from_secs(10))

I think it would be clearer if the timeout was in this function, however. I'll do that. 1e9a3c1487c9c5981f4094b23de1466b1dd8bd35

test_helpers_end_to_end_ng/src/steps.rs
alamb: Although it had a bug 🤦  - fixed in b69d7428e

test_helpers_end_to_end_ng/src/steps.rs
alamb: Good idea -- done in a49bad8948680301fcc1417e398b8407e2b406c3

influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
alamb: Update: fixed!

Issue #4288, opened 2022-04-12
Investigate the issues and health of Compactor in prod1


NGA-TRAN:
# Rationale 
The ticket is to keep track of the investigation of Compactor on `prod1`
- [Compactor Dashboards](https://grafana.prod01-eu-central-1.aws.influxdata.io/d/z7-FhLsnk/compactor?orgId=1)
- Get log command:  kubectl --context dex@prod01-eu-central-1-aws -n iox logs statefulset/iox-shared-compactor-0

# Status

**2022.04.12:**
- First query execution of the very first partition is stuck. 
      - This major step [built result stream for partition](https://github.com/influxdata/influxdb_iox/blob/bd4566cbe0e791bbc24c4738b52cba4b86e93bef/compactor/src/compact.rs#L654) happens
      - But its next major step c[ollected record batches from partition exec stream](https://github.com/influxdata/influxdb_iox/blob/bd4566cbe0e791bbc24c4738b52cba4b86e93bef/compactor/src/compact.rs#L664) never happens
- More info:  
     - The compactor generates 248 to-be-compacted groups of overlapped files for (sequencer: 90, namespace: 1, table: 116, partition: 5529587)
     - The first group with 7 parquet files are planned and executed
     - The plan includes **449 columns** (too many)
     - From available debug log, all 7 files were successfully downloaded from S3
     - Query plan:
```
level=debug msg="prepare_plan: plan to run" text="StreamSplitExec
    SortPreservingMergeExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]  <= [This is the reason to have this operator](https://github.com/influxdata/influxdb_iox/blob/85f3e696e89b7f298ca6e8821e7c5bf735bbf94d/query/src/provider.rs#L520)
      DeduplicateExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
        SortPreservingMergeExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
          UnionExec
            SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]. -- all sortExec will go away after epic #4283 is done
              IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
            SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
              IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
[...]

PR #4310, opened 2022-04-13
chore: add more compactor debug info

NGA-TRAN:
For on-going investigation of #4288 


query/src/exec/split.rs
```diff
@@ -199,7 +199,7 @@ impl ExecutionPlan for StreamSplitExec {
         partition: usize,
         context: Arc<TaskContext>,
     ) -> Result<SendableRecordBatchStream> {
-        trace!(partition, "SplitExec::execute");
+        debug!(partition, "Start SplitExec::execute");
```
NGA-TRAN, line 202: Since @domodwyer suggests to use debug instead of trace, I convert all related trace! into debug!

query/src/provider.rs
```diff
@@ -822,8 +829,8 @@ impl Deduplicater {
         ctx: IOxSessionContext,
         table_name: Arc<str>,
         output_schema: Arc<Schema>,
-        chunk: Arc<dyn QueryChunk>, // This chunk is identified having duplicates
```
NGA-TRAN, line 825: This comment is no longer accurate after a few changes in the past

compactor/src/compact.rs
```diff
@@ -353,6 +353,12 @@ impl Compactor {
             },
         );
 
+        let candidate_num = candidates.len();
+        debug!(
+            candidate_num,
+            "Number of candidate partitions to be considered to compact"
+        );
+
```
alamb, line 361: I don't think it makes any difference, but you can do something like this for less code (and only invoke `candidates.len()` if debug level logging is enabled)

```suggestion
        debug!(
            candidate_num=candidates.len();,
            "Number of candidate partitions to be considered to compact"
        );

```

compactor/src/compact.rs
```diff
@@ -616,6 +622,11 @@ impl Compactor {
             .map(|c| Arc::new(c) as Arc<dyn QueryChunk>)
             .collect();
         let merged_schema = QueryableParquetChunk::merge_schemas(&query_chunks);
+        let num_cols = merged_schema.as_arrow().fields().len();
+        debug!(
+            num_cols,
+            "Number of columns in the merged schema to build query plan"
+        );
```
alamb, line 629:
```suggestion
        debug!(
            num_cols=merged_schema.as_arrow().fields().len(),
            "Number of columns in the merged schema to build query plan"
        );
```

query/src/provider.rs
```diff
@@ -270,7 +270,7 @@ impl TableProvider for ChunkTableProvider {
 
         // This debug shows the self.arrow_schema() includes all columns in all chunks
         // which means the schema of all chunks are merged before invoking this scan
-        trace!("all chunks schema: {:#?}", self.arrow_schema());
+        debug!("All chunks schema: {:#?}", self.arrow_schema());
```
alamb, line 273:
```suggestion
        debug!(schema=?self.arrow_schema(), "All chunks schema: {:#?}");
```

query/src/provider.rs
```diff
@@ -717,7 +724,7 @@ impl Deduplicater {
         // Add DeduplicateExec
         // Sort exprs for the deduplication
         let sort_exprs = arrow_sort_key_exprs(sort_key, &plan.schema());
-        trace!(Sort_Exprs=?sort_exprs, chunk_ID=?chunks[0].id(), "Sort Expression for the deduplicate node of chunk");
+        debug!(Sort_Exprs=?sort_exprs, chunk_ID=?chunks[0].id(), "Sort Expression for the deduplicate node of chunk");
```
alamb, line 727:
```suggestion
        debug!(?sort_exprs, chunk_id=?chunks[0].id(), "Sort Expression for the deduplicate node of chunk");
```

query/src/provider.rs
```diff
@@ -953,7 +962,7 @@ impl Deduplicater {
         let input_schema = input.schema();
         let sort_exprs = arrow_sort_key_exprs(output_sort_key, &input_schema);
 
-        trace!(Sort_Exprs=?sort_exprs, Chunk_ID=?chunk.id(), "Sort Expression for the sort operator of chunk");
+        debug!(Sort_Exprs=?sort_exprs, Chunk_ID=?chunk.id(), "Sort Expression for the sort operator of chunk");
```
alamb, line 965:
```suggestion
        debug!(?sort_exprs, chunk_id=?chunk.id(), "Sort Expression for the sort operator of chunk");
```

Issue #4311, opened 2022-04-13
The configuration parameters for IOx are inconsistently and confusingly named


alamb:
This is a follow up for https://github.com/influxdata/influxdb_iox/pull/4147

# Rationale / Issue
I have been sorely confused by the names used for various configuration settings in IOx NG

Specifically:
* Write buffer config: `INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS` actually is how many kafka partitions (also called sequencers) are created
* Ingester config: `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START` refers to Kafka partitions (not the Partitions in the data heirarchy "Table" -> "Partition" -> Data. This is made more confusing as the ingester also has parameters that refer to the data partitions such as `INFLUXDB_IOX_PERSIST_PARTITION_AGE_THRESHOLD_SECONDS`
* Ingester config: `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END` also refers to Kafka partitions

So I propose change parameter names to reflect this.
* `INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_COUNT`
* `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_RANGE_START`
* `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_RANGE_END`

Because I would like to avoid  avoid bricking production which uses the existing configuration names, I plan to do this work in in three PRs:
- [ ] https://github.com/influxdata/influxdb_iox/pull/4312
- [ ] https://github.com/influxdata/k8s-idpe/pull/7112
- [ ] https://github.com/influxdata/influxdb_iox/pull/4313

alamb:
no one else seems to care about this so closing as won't do

Issue #4246, opened 2022-04-06
Rename Sequencer to Shard in Catalog / Codebase


alamb:
@NGA-TRAN  noted, and it was agreed by @domodwyer and @pauldix that the notion of "shard" and "sequencer" in the NG architecture are actually the same entity

Thus the proposal would be to:
- [ ] Replace `Sequencer` with `Shard` in the `ingester` (and replace `sequencer_id` with `shard_id`)
- [ ] Update the postgres catalog / schema to use the term `shard` rather than `sequencer`

domodwyer:
We should also rename the associated metrics which also use a spread of names for the same thing (from @pauldix) 👍

carols10cents:
I think this is completely moot now that shards are going away 👍🏻

alamb:
❤️

PR #4312, opened 2022-04-13
fix: Make configuration names less confusing

alamb:
# Rationale:
Part 1 / 3 of https://github.com/influxdata/influxdb_iox/issues/4311 (detailed "why" is listed there)

This looks like a large PR but it is largely moving code around / tests

# Changes:

1. Add the following aliases:
* `INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_COUNT`
* `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_RANGE_START`
* `INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END` --> `INFLUXDB_IOX_WRITE_BUFFER_KAFKA_PARTITION_RANGE_END`

2. Tests to show setting the old names continues to work (so we can roll this out to production without breaking things)
3. Consolidate the `WRITE_BUFFER_PARTITION_... ` into a single struct shared by `IngesterConfig` and `CompactorConfig`

Note that the existing NG end to end tests  use the old configuration parameters so I think this change is well covered

### DIFF
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -1,3 +1,5 @@
+use crate::kafka_partitions::KafkaPartitionConfig;
+
 /// CLI config for compactor
 #[derive(Debug, Clone, clap::Parser)]
 pub struct CompactorConfig {
@@ -10,19 +12,9 @@ pub struct CompactorConfig {
     )]
     pub topic: String,
 
-    /// Write buffer partition number to start (inclusive) range with
-    #[clap(
-        long = "--write-buffer-partition-range-start",
-        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
-    )]
-    pub write_buffer_partition_range_start: i32,
-
-    /// Write buffer partition number to end (inclusive) range with
-    #[clap(
-        long = "--write-buffer-partition-range-end",
-        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END"
-    )]
-    pub write_buffer_partition_range_end: i32,
+    /// Kafka Partition options
+    #[clap(flatten)]
+    pub kafka_partition_config: KafkaPartitionConfig,
 
     /// Percentage of least recent data we want to split to reduce compacting non-overlapped data
     /// Must be between 0 and 100. Default is 100, which won't split the resulting file.
--- a/clap_blocks/src/kafka_partitions.rs
+++ b/clap_blocks/src/kafka_partitions.rs
@@ -1,0 +1,155 @@
+/// CLI config for Kafka partition information
+#[derive(Debug, Clone, clap::Parser)]
+pub struct KafkaPartitionConfig {
+    /// DEPRECATED: Write buffer partition number to start (inclusive) range with
+    #[clap(
+        long = "--write-buffer-partition-range-start",
+        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
+    )]
+    write_buffer_partition_range_start: Option<i32>,
+
+    /// DEPRECATED: Write buffer partition number to end (inclusive) range with
+    #[clap(
+        long = "--write-buffer-partition-range-end",
+        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END"
+    )]
+    write_buffer_partition_range_end: Option<i32>,
+
+    /// First write buffer kafka partition number (inclusive) this
+    /// service is responsible for
+    #[clap(
+        long = "--write-buffer-kafka-partition-range-start",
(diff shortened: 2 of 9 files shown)


test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -60,6 +60,7 @@ impl TestConfig {
     pub fn new_router2(dsn: impl Into<String>) -> Self {
         Self::new(ServerType::Router2, dsn, random_catalog_schema_name())
             .with_new_write_buffer()
+            .with_kafka_partition_count(1)
```
alamb, line 63: made this explicit so we don't try and configure the all in one mode with the kafka partition count as well

clap_blocks/src/compactor.rs
```diff
@@ -10,19 +12,9 @@ pub struct CompactorConfig {
     )]
     pub topic: String,
 
-    /// Write buffer partition number to start (inclusive) range with
```
alamb, line 13: was copy pasted, so unified into its own struct

clap_blocks/src/kafka_partitions.rs
```diff
@@ -0,0 +1,155 @@
...
+        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_START"
+    )]
+    write_buffer_partition_range_start: Option<i32>,
+
+    /// DEPRECATED: Write buffer partition number to end (inclusive) range with
+    #[clap(
+        long = "--write-buffer-partition-range-end",
+        env = "INFLUXDB_IOX_WRITE_BUFFER_PARTITION_RANGE_END"
+    )]
+    write_buffer_partition_range_end: Option<i32>,
```
alamb, line 16: My next PR will delete these fields and most of the tests in this one

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -168,26 +169,41 @@ impl TestConfig {
...
             .with_env("INFLUXDB_IOX_WRITE_BUFFER_ADDR", &write_buffer_string)
     }
 
+    /// Configures the write buffer to use the specified kafka partition count
+    fn with_kafka_partition_count(self, kafka_partition_count: usize) -> Self {
+        self.with_env(
+            // change to:
+            // "INFLUXDB_IOX_WRITE_BUFFER_KAFAKA_PARTITION_COUNT,
+            // https://github.com/influxdata/influxdb_iox/issues/4311
+            "INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS",
```
alamb, line 188: note the end to end tests still use the old configuration names

crepererum:
Why again are we using Kafka-specific terminology here? To me that's super confusing because the write-buffer might be file-based (which is what we actually use during the end2end tests).

alamb:
>  Why again are we using Kafka-specific terminology here? To me that's super confusing because the write-buffer might be file-based (which is what we actually use during the end2end tests).


The short answer is "because that is what the ingester / NG code calls the concept, and now the configuration is inconsistent"

The longer answer is that @pauldix /  NG used a different meaning of `sequencer_id` in the ingester than exists in the writer buffer. The ingester uses `KafkaPartition` ([source link](https://github.com/influxdata/influxdb_iox/blob/main/data_types2/src/lib.rs#L155-L158) as this concept and so I am just carrying that through to the configuration values

I also find this super confusing and wish that the ingester had not reused the same term as OG for a different concept -- see  https://github.com/influxdata/influxdb_iox/issues/4237 for more of my thoughts on the matter. 

There was some consensus that renaming "sequencer" in the ingester to "shard" is the ideal outcome https://github.com/influxdata/influxdb_iox/issues/4246 but I didn't think it was time to do a wholesale rename in the codebase (hoping we can do so after OG is gone)

So to conclude, I think naming is still a mess, but this PR at least makes it more consistent

crepererum:
The explanation makes sense, but TBH I'm not really sure if that PR is an improvement, but I'm also not opposing it. So let's @mkmik decide.

mkmik:
I agree that the current flags are confusing and that we should do something.

I'm a bit confused about where why can't we just call it "--write-buffer-sequencer-range-start" ? Aren't we effectively calling these "shards"  as "sequencers" already? (I'm also favourable of going one step further and just embrace the "shard" terminology)

alamb:
So it sounds like we should  punt / postpone all of this until we have retired  OG and renamed (if we have the stamina to do so) `sequencer` to `shard`

Until then I do not care enough to push this initiative forward. I thought it was going to help but if it doesn't then it is just a waste of time for all involved 🤷 .

Issue #4315, opened 2022-04-13
Add tooling to pull data from prod


pauldix:
This comes through in a number of parts. Here's what I'm planning:
- [x] #4316 - Add gRPC service to get parquet file records by partition id
- [x] Add gRPC service to router to get parquet file bytes from object store by object store id
- [x] Add client and CLI to get parquet file from object store and write into local file
- [x] Add calls to the iox client with end-to-end tests
- [ ] Add CLI to connect to remote and inject parquet files into local catalog and object store

PR #4316, opened 2022-04-13
feat: add initial catalog service to router

pauldix:
Create new crate for iox_catalog_service.
Add rpc to return parquet_file records by partition id.
Add CatalogService to router2.

The catalog service will be added to over time to provide access to the catalog over gRPC.

This is the first step in #4315


iox_catalog_service/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        sequencer_id: p.sequencer_id.get() as i32,
+        namespace_id: p.namespace_id.get(),
+        table_id: p.table_id.get(),
+        partition_id: p.partition_id.get(),
+        object_store_id: p.object_store_id.to_string(),
+        min_sequence_number: p.min_sequence_number.get(),
+        max_sequence_number: p.max_sequence_number.get(),
+        min_time: p.min_time.get(),
+        max_time: p.max_time.get(),
+        to_delete: p.to_delete.map(|t| t.get()).unwrap_or(0),
```
pauldix, line 75: This should be an optional value, but everything is required in the autogenerated code. Not sure what the right thing to do here is.

domodwyer: We decided that we wouldn't have explicit optional fields, and would instead use "zero values" to represent it - I'm not sure this is right, because 0 might be a perfectly valid value 🤷 

I guess in this case, 0 would not be a valid value and so 0 == not deleted?

domodwyer (approved):
Looks good.

We need [one more line](https://github.com/influxdata/influxdb_iox/blob/9d72ede1cf5a81729700bf4a423cd25c312b7089/ioxd_router2/src/lib.rs#L108) to tell the grpc server about the new service:

```rust
add_service!(builder, self.server.grpc().catalog_service());
```

iox_catalog_service/src/lib.rs
pauldix: Yeah, 0 is not deleted. Given that it represents the time at which the file was marked for deletion, anything before whenever the system starts in production would be invalid. So 0 is safe to use as the invalid one.

PR #4317, opened 2022-04-14
fix: deadlock in compactor

pauldix:
The asserts won't succeed, but this reproduces the hang we're seeing in prod. I picked the number of rows arbitrarily. I haven't had time to actually start digging in, but I figured I'd let you at first since you'll be up way before me 😄


pauldix:
Closing for #4319

PR #4319, opened 2022-04-14
test: deadlock test and add more debug log

NGA-TRAN:
This is test reproduced in #4317 and fixed in #4325

The command below was from investigation yesterday which is still valid but the problem was already fixed in #4325

-----

Command to run the test to get debug log
`RUST_LOG=debug cargo test --package compactor --lib -- compact::tests::test_compact_two_big_files --exact --nocapture`

A lot of useful log info but I need to add more log into some Tokio spawn (which may be inside DF) to see where it is stuck. It is a bit tricky because data is streaming

From the log, it looks to me:
1. Two data sets are sorted. I think those 2 set are data of 2 files but that needs to verify because DF does split one file into many set/partitions if the set is large which I do not think in this case2. 
2. The deduplication happens and it seems complete too but I need to verify that, too. There are tokios and send receive here which make things a bit tricky to follow3. 

**Here is the plan with a bit of analysis**
```
SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC] <= this one has only one input and simply returns the execute plan below
    DeduplicateExec: [tag1@1 ASC,time@2 ASC]. , <=  seems finished but NOT SURE the send/receive between Tokio task finishes. It seems to me the deduplication is finished at the send side. Need to verify the whether they are all received at the receiver.
      SortPreservingMergeExec: [tag1@1 ASC,time@2 ASC]
        UnionExec
          SortExec: [tag1@1 ASC,time@2 ASC]       <= seems finished
            IOxReadFilterNode: table_name=table, chunks=1 predicate=Predicate.   <= finished
          SortExec: [tag1@1 ASC,time@2 ASC].    <= seems finished
            IOxReadFilterNode: table_name=table, chunks=1 predicate=Predicate.   <= finished
```


Cargo.lock
```diff
@@ -1168,7 +1168,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "7.0.0"
-source = "git+https://github.com/NGA-TRAN/arrow-datafusion.git?rev=215db4c07a5cd1c18088bd602adfc1f4ec9dfbbc#215db4c07a5cd1c18088bd602adfc1f4ec9dfbbc"
+source = "git+https://github.com/NGA-TRAN/arrow-datafusion.git?rev=7d7c8255082069c7dd47c7bbf4051d30cde91edb#7d7c8255082069c7dd47c7bbf4051d30cde91edb"
```
NGA-TRAN, line 1171: This is another DF commit on my fork

compactor/src/compact.rs
```diff
@@ -2770,4 +2770,97 @@ mod tests {
         ];
         assert_eq!(expect, candidates);
     }
+
+    #[tokio::test]
+    async fn test_compact_two_big_files() {
+        test_helpers::maybe_start_logging();
```
NGA-TRAN, line 2776: Needs this line to see debug info

compactor/src/compact.rs
```diff
@@ -2770,4 +2770,97 @@ mod tests {
...
+        // File 2 without tombstones
+        let pf2 = ParquetFileWithTombstone {
+            data: Arc::new(parquet_file2),
+            tombstones: vec![],
+        };
+        println!("pf1 & [f2 created");
+
+        // Compact them
+        let batches = compactor.compact(vec![pf1, pf2]).await.unwrap();
+        println!("Finish");
```
NGA-TRAN, line 2835: Never gets here

query/src/provider/deduplicate.rs
```diff
@@ -238,8 +242,10 @@ impl ExecutionPlan for DeduplicateExec {
                     debug!("deduplicate receiver hung up");
                 }
             }
+            debug!("no error watching worker task");
         });
 
+        // Note: the debug message ends before the above tokio tasks executed
         debug!(partition, "End DeduplicationExec::execute");
 
         Ok(AdapterStream::adapt(self.schema(), rx))
```
NGA-TRAN, line 251: Todo: add log into `AdapterStream::adapt` to see if the `rx` receives results or it is stuck here

alamb:
Related to https://github.com/influxdata/influxdb_iox/pull/4317

pauldix (approved):

PR #4314, opened 2022-04-13
chore: use Nga's DF branch with debug information for compactor

NGA-TRAN:
Debug for #4288 

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #4322, opened 2022-04-14
chore: Update datafusion (and point back to apache)

alamb:
Draft until @NGA-TRAN  is done with debugging production (see https://github.com/influxdata/influxdb_iox/pull/4314)

# Rationale
Pick up new changes, but also use apache rather than @NGA-TRAN's branc

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/8d5bb47f59c9aebf8bead6593ccd56e4eff4affc


alamb:
Thanks @NGA-TRAN

Issue #4324, opened 2022-04-14
Support streaming compaction in NG


domodwyer:
Currently the compactor reads all record batches into memory during compaction - it would be nice to stream directly into parquet files on disk to reduce the memory usage of the compactor.

NGA-TRAN:
I hope you will cover #4101

alamb:
Additional thoughts: the success critera would be to avoid the unbounded stream that was introduced: https://github.com/influxdata/influxdb_iox/pull/4325

domodwyer:
Yes removing the unbounded channels would be my intention as part of this ticket 👍

alamb:
Another potential issue is that we buffer the entire plan output into memory prior to writing to parquet

https://github.com/influxdata/influxdb_iox/blob/f9e3495/compactor/src/compact.rs#L731-L734

domodwyer:
Yes this is what I'll be addressing - I'll also undo the buffering I added in https://github.com/influxdata/influxdb_iox/pull/4325 to get us compacting.

PR #4325, opened 2022-04-14
fix: compactor deadlock

### LINKED ISSUE #4306: Compaction does not progress
During initial testing, we have discovered the compactor does not complete a compaction operation.

The generated physical plan is:

```
StreamSplitExec
  SortPreservingMergeExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
    DeduplicateExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
      SortPreservingMergeExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
        UnionExec
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
          SortExec: [cloud@4 ASC,host@195 ASC,server@362 ASC,time@384 ASC]
            IOxReadFilterNode: table_name=mysql, chunks=1 predicate=Predicate
```

* Compaction stalls during plan execution and never progresses. 
* CPU profiling shows no activity. 
* The parquet files appear to be fetched from object store and read into record batches successfully.

domodwyer:
This PR fixes the compaction deadlock we have observed during testing.

The deadlock is caused by the `StreamSplitExec` background worker that pushes results into two channels - one for each partition. The channels have a [buffer size of 2], and the worker [gets stuck] trying to push results into the second partition while the compactor is reading the first partition.

Because the worker is stuck, it cannot push the rest of the records into the channel for the first partition, nor close the channel (indicating completion, and signalling to the compactor that it can move on). Because the channel is never closed, the compaction loop [blocks waiting for the first partition](https://github.com/influxdata/influxdb_iox/blob/31fdeaaabc21d2e7d24b801b8503d77aee7853f8/compactor/src/compact.rs#L671) to complete, and can't drain the second partition to unblock the worker 💣 

[gets stuck]:https://github.com/influxdata/influxdb_iox/blob/351b0d0c152a36f03dd12514df56fa50c143cc75/query/src/exec/split.rs#L399
[buffer size of 2]:https://github.com/influxdata/influxdb_iox/blob/main/query/src/exec/split.rs#L263

This change prevents us from performing a streaming compaction - records are split, buffered in memory, and then read by the compactor. Currently the compactor buffers the resulting record batches it reads from datafusion in memory anyway, so this isn't a big deal. I will address this in the future (#4324).

Closes #4306.

---

* feat: unbounded channel support for AdaptorStream (5ac4785e)

      Allows the AdaptorStream to work with either a bounded, or unbounded channel
      internally.

* fix: compaction deadlock (00b5c1b2)

      This commit resolves the compaction deadlock described in #4306.

      The deadlock occurs during StreamSplitExec execution, where a background 
      worker is spawned to read input record batches and partition them into two
      groups. This code pushes the resulting split record batches into two channels
      - one for records that match a given predicate, and another channel for those
      that do not. These channels buffer at most 2 record batches each.

      The compactor that executes this plan reads the resulting partitions 
      sequentially to completion. Completion is indicated by reading until the 
      results stream ends, which ends when the underlying channel is closed, and
      therefore the split worker task must have finished and closed the results
      channel for the partition to be successfully read.

      While the compactor is reading from the first partition, the worker is 
      attempting to push record batches into the second partition and blocks due to
      the channel capacity being reached. The worker never drops the channel for the
      first partition, so the compactor never finishes reading the first partition,
      and nothing is reading the second partition to unblock the worker. Deadlock!

* refactor: log split worker panics at error level (31fdeaaa)

      When the split background worker panics, it now causes an ERROR level log to
      be emitted.

### DIFF
--- a/datafusion_util/src/lib.rs
+++ b/datafusion_util/src/lib.rs
@@ -16,8 +16,8 @@ use datafusion::{
     scalar::ScalarValue,
 };
 use futures::{Stream, StreamExt};
-use tokio::sync::mpsc::Receiver;
-use tokio_stream::wrappers::ReceiverStream;
+use tokio::sync::mpsc::{Receiver, UnboundedReceiver};
+use tokio_stream::wrappers::{ReceiverStream, UnboundedReceiverStream};
 
 /// Traits to help creating DataFusion [`Expr`]s
 pub trait AsExpr {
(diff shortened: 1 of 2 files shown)


NGA-TRAN (approved):
Thanks @domodwyer

alamb:
I kept getting confused because when I looked at the link,  PR had already merged and fixed the issues

alamb:
Nice work @domodwyer  -- a great team effort between you, @NGA-TRAN  and @pauldix  🏆

PR #4328, opened 2022-04-14
fix: Revert "chore: Revert "fx: Revert "fix: Revert "feat: Use the sort key stored in the catalog during compaction" (#4299)" (#4303)" (#4327)"

### LINKED ISSUE #4249: compactor: Use the sort_key in the compactor
For example here: https://github.com/influxdata/influxdb_iox/blob/e6c07809872edec0974d5fac9502cd6ed3e30411/compactor/src/compact.rs#L684
And here: https://github.com/influxdata/influxdb_iox/blob/e6c07809872edec0974d5fac9502cd6ed3e30411/compactor/src/query.rs#L104

NGA-TRAN:
Believe or not, we have to put this back in after the wipe catalog is postponed
This reverts commit 7e5d71902722caaee257455921d4f372e58dc536.

Closes #4249 

@mkmik , @alamb and @carols10cents : This PR must be merged AFTER catalog is wiped but BEFORE the clusters start running again when the catalog is till empty. 
Why: If we have Compactor runs on inserted data after the catalog is wiped out without this PR, the Compactor will produce sorted data that is NOT consistent with what we want or data is not even recorded sorted and will cause performance issue and maybe even panic.


alamb:
Its reverts all the way down

image

alamb (approved):

NGA-TRAN:
@pauldix, @jacobmarble and I have talked today. Since no one uses the NG IOx cluster (all Ingesters and Compactor are not running due to some issues), we plan to wipe the catalog even before restarting the cluster, and I need to use the changes in this PR to continue other work, we decided to merge this PR in today. Hopefully, we do not revert it the third time 😄

PR #4227, opened 2022-04-04
feat: Add API to ingester for write readability and persistence

alamb:
Draft as it builds on the following PRs for easier review:
- [x] https://github.com/influxdata/influxdb_iox/pull/4232
- [x] https://github.com/influxdata/influxdb_iox/pull/4238
- [ ] `WriteInfoService` implementation and + end to end tests (this PR)

Closes https://github.com/influxdata/influxdb_iox/issues/4145

# Rationale:
I want to answer the question "is my data readable yet"?

Design doc: https://docs.google.com/document/d/1VFB7xmVn_ezet0gzFxE9x5pNS-1IhXi26OQOWsjn5OU/edit#heading=h.pr4zatesz5xi

# New API:

Given the write token retrieved from /api/v2/write API (API added in https://github.com/influxdata/influxdb_iox/pull/4202), returns a message like:

Response:
```json
{
  "readable": true,
  "persisted": false
}
```

# Follow on work:
- [ ] File an issue related to summarizing write summaries across multiple ingesters

PR #4329, opened 2022-04-14
feat: add catalog client and remote command

pauldix:
Adds the catalog gRPC service to influxdb_iox_client.
Adds a new remote command to execute commands against a remote IOx host.
Adds partition subcommand to remote to get the details of a partition by id.

This is one step of #4315. Next steps will be to add an option to import the information into the client's catalog and to get the actual parquet files from the remote.

I'm not sure how to include and end-to-end test at this point. I'd need an ingester spun up and I may need to add an API to the ingester to force persistence of Parquet files. Although this can also probably be done via config so I'll have to play around with it. The timing of it will be tricky since we'd be waiting for the ingester persist.

Any thoughts on the best approach for that @alamb?


influxdb_iox/src/commands/remote.rs
```diff
@@ -0,0 +1,43 @@
+//! This module implements the `remote` CLI command
```
alamb, line 1:
```suggestion
//! This module implements the `remote` CLI command (NG)
```

It is getting somewhat confusing what CLI commands are for NG and what are for OG. The sooner we can remove the OG code the better I think

influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -0,0 +1,59 @@
...
+#[allow(clippy::enum_variant_names)]
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("JSON Serialization error: {0}")]
+    Serde(#[from] serde_json::Error),
+
+    #[error("Client error: {0}")]
+    ClientError(#[from] influxdb_iox_client::error::Error),
+
+    #[error("Error updating catalog: {0}")]
```
alamb, line 16: Is UpdateCatalogError used? It isn't clear to me what updating would be happening...

influxdb_iox/src/main.rs
```diff
@@ -169,6 +170,9 @@ enum Command {
     // Clippy recommended boxing this variant because it's much larger than the others
     Run(Box<commands::run::Config>),
 
+    /// Commands to run against remote IOx APIs
```
alamb, line 173:
```suggestion
    /// Commands to run against remote IOx APIs (NG)
```

influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -0,0 +1,59 @@
...
+enum Command {
+    Show(Show),
+}
+
+pub async fn command(connection: Connection, config: Config) -> Result<(), Error> {
+    match config.command {
+        Command::Show(show) => {
+            // let metrics = Arc::new(metric::Registry::new());
+            // let catalog = show.catalog_dsn.get_catalog("cli", metrics).await?;
+            // let mut repos = catalog.repositories().await;
```
alamb, line 51: Did you mean to leave these commented out?

influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/// Manage IOx chunks
+#[derive(Debug, clap::Parser)]
+pub struct Config {
+    #[clap(subcommand)]
+    command: Command,
+}
+
+/// Create or update a topic
```
alamb, line 30:
```suggestion
/// Show a particular partition_id
```

alamb (approved):
Looks good to me. 

I tried it out locally:

```shell
# Run all in one mode:
cargo run -- run -vv  all-in-one   --write-buffer=file  --write-buffer-addr /tmp  --catalog-dsn postgresql://localhost:5432/alamb  --write-buffer-partition-range-start 0  --write-buffer-partition-range-end 0  --pause-ingest-size-bytes 10000  --persist-memory-threshold-bytes 1000

# write data
curl -v -X POST "http://127.0.0.1:8080/api/v2/write?bucket=bucket&org=org" --data-binary @test_fixtures/lineproto/metrics.lp

```

But then I got a `Operation is not implemented or supported: ` error because by default the CLI connects to the querier gRPC

```shell

alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ./target/debug/influxdb_iox remote partition show 1
Error in partition subcommand: Client error: Operation is not implemented or supported: 
alamb@MacBook-Pro-3:~/Software/influxdb_iox$ 
```

However, when I aimed it at the router2 address it worked great: 

```shell
alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ./target/debug/influxdb_iox -h http://localhost:8081 remote partition show 1
[
  {
    "id": "1",
    "sequencerId": 1,
    "namespaceId": 1,
    "tableId": 1,
    "partitionId": "1",
    "objectStoreId": "06d8343e-b8c9-4b19-89fd-a854e4e9146b",
    "minTime": "123456",
    "maxTime": "123456",
    "fileSizeBytes": "2021",
    "rowCount": "1",
    "compactionLevel": 1,
    "createdAt": "1649794891912144000"
  }
]
```

alamb:
> I'm not sure how to include and end-to-end test at this point. I'd need an ingester spun up and I may need to add an API to the ingester to force persistence of Parquet files. Although this can also probably be done via config so I'll have to play around with it. The timing of it will be tricky since we'd be waiting for the ingester persist.

> Any thoughts on the best approach for that @alamb?

You are right that it is tricky. Conveniently, I have been working to do all that trickery the last week so it is now fairly straightforward (especially with some great ideas and help from @crepererum). Here is a PR with an end to end test in case that is helpful for you: https://github.com/influxdata/influxdb_iox/pull/4336

p.s. the "The timing of it will be tricky since we'd be waiting for the ingester persist." is exactly what https://github.com/influxdata/influxdb_iox/pull/4227 was all about 👍

influxdb_iox/src/commands/remote/partition.rs
pauldix: Ah, nope, will remove.

influxdb_iox/src/commands/remote/partition.rs
pauldix: Ah, that's accidentally left behind from something I'm including later, which is to inject the pulled down catalog data into the local catalog. Will remove and add in later in a follow on PR.

PR #4336, opened 2022-04-15
test: add end to end test for `remote partition` CLI

alamb:
Here is an end to end test for the new `remote partition` cli from @pauldix  added in https://github.com/influxdata/influxdb_iox/pull/4329

The more I see this `StepTest` framework going the better I like it


pauldix:
Awesome, this looks great! Super helpful to have that step framework.

PR #4338, opened 2022-04-15
feat: add object store service to router

pauldix:
Add method to catalog to get parquet file by object store id.
Add gRPC service for object store to get a file from by its uuid.
Add the object store service to router2 with object store config.

I'll be adding the client, CLI and end-to-end tests in a following PR.


iox_object_store_service/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        let res = self
+            .object_store
+            .get(&path)
+            .await
+            .map_err(|e| Status::unknown(e.to_string()))?;
+
+        let mut stream = res.into_stream();
+
+        let (tx, rx) = mpsc::channel(4);
+        tokio::spawn(async move {
```
tustvold, line 90: You should be able to avoid an additional task, and buffer, by doing something like

```
Box::pin(res.into_stream.map(|next| {
    Ok(data) => Ok(GetParquetFileByObjectStoreIdResponse {
          data: data.to_vec(),
      }),
      Err(e) => Err(Status::unknown(e.to_string())),
}))
```

And changing `GetParquetFileByObjectStoreIdStream` to be `BoxStream<'static, Result<GetParquetFileByObjectStoreIdResponse, Status>>`

generated_types/protos/influxdata/iox/object_store/v1/service.proto
```diff
@@ -0,0 +1,17 @@
...
+package influxdata.iox.object_store.v1;
+
+service ObjectStoreService {
+    // Get the parquet file from the object store by its uuid
+    rpc GetParquetFileByObjectStoreId(GetParquetFileByObjectStoreIdRequest) returns (stream GetParquetFileByObjectStoreIdResponse);
+}
+
+message GetParquetFileByObjectStoreIdRequest {
+    // the parquet file object store uuid
+    string uuid = 1;
```
tustvold, line 11: At least in the older APIs, we represented UUIDs as bytes and not string

alamb: For example: https://github.com/influxdata/influxdb_iox/blob/2c04215c1fa082bbe0c812f764931fc3a32cf18a/generated_types/protos/influxdata/iox/management/v1/chunk.proto#L69

I remember there was some discussion about the merits of `bytes` vs `string` and the protobuf gurus decided that `bytes` was the way to go

generated_types/protos/influxdata/iox/object_store/v1/service.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+package influxdata.iox.object_store.v1;
+
+service ObjectStoreService {
```
alamb, line 4: I like the idea of an `ObjectStoreService` (that presumably can eventually be extended to expose more of the object store API, as needed 👍 )

influxdb_iox/src/commands/run/router2.rs
```diff
@@ -74,10 +78,17 @@ pub async fn command(config: Config) -> Result<()> {
         .get_catalog("router2", Arc::clone(&metrics))
         .await?;
 
+    let object_store = ObjectStoreImpl::try_from(config.run_config.object_store_config())
```
alamb, line 81: Ping @domodwyer  -- I can't remember if there was some architectural reason to avoid requiring the object store in the router. If there is, perhaps we can more the object store service to either the querier or the ingester processes

Cargo.toml
```diff
@@ -30,6 +30,7 @@ members = [
     "iox_data_generator",
     "iox_gitops_adapter",
     "iox_object_store",
+    "iox_object_store_service",
```
alamb, line 33: There are similiar grpc service implementations in crates called `service_grpc_influxrpc`, etc. Maybe it would make sense to rename this crate `service_grpc_object_store`

For example:

https://github.com/influxdata/influxdb_iox/pull/4338/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L70-L72

iox_object_store_service/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let path = ParquetFilePath::new_new_gen(
+            parquet_file.namespace_id,
+            parquet_file.table_id,
+            parquet_file.sequencer_id,
+            parquet_file.partition_id,
+            parquet_file.object_store_id,
+        )
+        .absolute_dirs_and_file_name();
+        let path = Path::InMemory(path);
```
alamb, line 79: I am surprised to see `InMemory` here

alamb (approved):
Looks like it does what is intended. I think there are a few good comments / suggestions on the PR reviews that are worth considering

Is it worth a ticket to track the work (I am happy to file one) where we can also list the other follow on tasks?

Adding to the router2 service makes sense 👍  Though I wonder about how, longer term, this functionality will be secured. If we allow end users to make requests to router2 we'll need some way to enforce authorization / security so they can't fetch and make off with parquet files. The same is true for the querier too

influxdb_iox/src/commands/run/router2.rs
pauldix: There isn't, it just wasn't necessary before because there wasn't an API yet that needed it. But the router is the front facing API for the service as a whole, so it makes sense to have it here.

generated_types/protos/influxdata/iox/object_store/v1/service.proto
pauldix: What happens when it's converted into JSON?

pauldix:
Authentication and authorization will be happening in the Gateway service, which is also where we do things like rate limiting and billing metering. Once the request gets to the router it should have been validated.

iox_object_store_service/src/lib.rs
pauldix: So I wasn't sure how to convert the parquet file path into the object store path I needed. This worked, but I'd definitely like to change it to a better way.

Cargo.toml
pauldix: Ah, I was doing the same thing I had done with `iox_catalog_service`. Should I rename this one and also rename that one to `service_grpc_<thing>`? I picked the name so that it would show up in the project right below the library.

pauldix:
@alamb I have #4315 to track work related to this.

iox_object_store_service/src/lib.rs
pauldix: Ah, looks like this: `

pauldix:
I've addressed the two primary comments so I'll set to merge for now. But let me know if you want me to rename the service and also the catalog service as well.

pauldix:
Oh, and I'm still not sure if we should be using string or bytes for the UUIDs. I can change that as well in a follow up PR. Just let me know if that's what I should be doing. Just curious about how that ends up looking when converted to JSON.

Cargo.toml
alamb: I guess I was thinking having things consistently named would be ideal -- I don't have a strong preference on which way to go

influxdb_iox/src/commands/run/router2.rs
domodwyer: I think I made a passing comment before that we should not add object store functionality to the router _if it was avoidable_ simply to segregate object store access away from the network "front door" - the object store will contain everything and be a high value target, and the router is the easiest component in the architecture to attack. 

It's fine if we need to add it to support functionality like this (yay rust safety!) 👍

PR #4348, opened 2022-04-19
feat: add sequencer ID and correct partition key to `IngesterPartition`

crepererum:
- simplifies debugging (parquet chunks and ingester chunks now use the
  same partition key naming)
- the sequencer ID is required to correctly reason about tombstones (to
  be implemented in a later PR)


querier/src/cache/partition.rs
```diff
@@ -56,11 +57,17 @@ impl PartitionCache {
     pub async fn old_gen_partition_key(&self, partition_id: PartitionId) -> Arc<str> {
         self.cache.get(partition_id).await.old_gen_partition_key
     }
+
```
alamb, line 60: Are the point of the changes in this module is to route all catalog access in the querier through the catalog cache?

alamb (approved):

querier/src/cache/partition.rs
crepererum: yes, because now the ingester connection needs it as well and I wanted to make sure that all querier parts use the same cache.

PR #4351, opened 2022-04-19
fix: revert column service limits (#4179)

domodwyer:
This PR reverts #4179 which added column limit enforcement to the `ColumnRepo::create_or_get_many()` code path.

This method is [called as part of the schema validation][call] in the router when an unknown column is observed in a write request and the result is cached. At router startup, the schema cache is empty and all requests invoke `create_or_get_many()` at least once per table.

Currently this query is taking ~5s to execute, generating the following plan showing multiple sequential scans:

```
 Insert on column_name  (cost=85347.08..112559.28 rows=1027 width=42)
   Conflict Resolution: UPDATE
   Conflict Arbiter Indexes: column_name_unique
   ->  Nested Loop  (cost=85347.08..112559.28 rows=1027 width=42)
         Join Filter: (input2.table_id = input.table_id)
         ->  Function Scan on input  (cost=0.01..0.02 rows=1 width=12)
         ->  Hash Join  (cost=85347.07..112536.16 rows=1027 width=8)
               Hash Cond: (table_name.id = input2.table_id)
               Join Filter: (((count(column_name_1.*)) + (count(input2.table_id))) <= namespace.max_columns_per_table)
               ->  HashAggregate  (cost=85347.02..104742.21 rows=616026 width=20)
                     Group Key: namespace.id, table_name.id
                     Planned Partitions: 32
                     ->  Hash Right Join  (cost=1769.37..15466.57 rows=616026 width=58)
                           Hash Cond: (table_name.namespace_id = namespace.id)
                           ->  Hash Right Join  (cost=1498.49..13576.96 rows=616026 width=54)
                                 Hash Cond: (column_name_1.table_id = table_name.id)
                                 ->  Seq Scan on column_name column_name_1  (cost=0.00..10461.26 rows=616026 width=50)
                                 ->  Hash  (cost=920.44..920.44 rows=46244 width=8)
                                       ->  Seq Scan on table_name  (cost=0.00..920.44 rows=46244 width=8)
                           ->  Hash  (cost=219.28..219.28 rows=4128 width=8)
                                 ->  Seq Scan on namespace  (cost=0.00..219.28 rows=4128 width=8)
               ->  Hash  (cost=0.04..0.04 rows=1 width=12)
                     ->  HashAggregate  (cost=0.02..0.03 rows=1 width=12)
                           Group Key: input2.table_id
                           ->  Function Scan on input2  (cost=0.01..0.02 rows=1 width=4)
 JIT:
   Functions: 44
   Options: Inlining false, Optimization false, Expressions true, Deforming true
(28 rows)
```

Removing this functionality is a temporary measure and I intend to reintroduce this enforcement either via a new implementation/algorithm or by reducing the cost of this query to an acceptable level - this is tracked as part of https://github.com/influxdata/influxdb_iox/issues/4349.

[call]:https://github.com/influxdata/influxdb_iox/blob/320f1073e0ddbe5ade3ba4ba07ab038a22bb2e9f/iox_catalog/src/lib.rs#L156

---

* fix: revert column service limits (#4179) (320f1073)

      This reverts commit ea865b63f42899645b51464e83d67b19f21b6852.

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -59,16 +59,6 @@ pub enum Error {
     },
 
     #[snafu(display(
-        "partial failure: inserted {} of {} column names in multi insert.",
-        input_column_count,
-        inserted_column_count,
-    ))]
-    PartialColumnCreateError {
-        input_column_count: usize,
-        inserted_column_count: usize,
-    },
-
-    #[snafu(display(
         "couldn't create table {}; limit reached on namespace {}",
         table_name,
         namespace_id
@@ -357,13 +347,6 @@ pub trait ColumnRepo: Send + Sync {
     /// Implementations make no guarantees as to the ordering or atomicity of
     /// the batch of column upsert operations - a batch upsert may partially
     /// commit, in which case an error MUST be returned by the implementation.
-    ///
-    /// Partial failure will occur when column limits per table are hit. For example, if namespace
-    /// A has a limit of 10 columns per table and so does namespace B, the DB already has 5 columns
-    /// for table A in namespace A and nothing for namespace B, and a batch is passed to this
-    /// function containing 10 inserts for table A and 10 inserts for some table in namespace B,
-    /// ALL of the inserts for table A will fail and all of the inserts for the table in namespace
-    /// B will succeed, and an error will be returned telling the user it was a partial failure.
     async fn create_or_get_many(
         &mut self,
         columns: &[ColumnUpsertRequest<'_>],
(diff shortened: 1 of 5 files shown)


alamb (approved):
I am approving this to unstick production -- is it worth filing a ticket to track the "schema  validation is exploding the router" issue ?

domodwyer:
Thanks - I am using https://github.com/influxdata/influxdb_iox/issues/4349 to track these changes to the service limit behaviour 👍

Issue #4352, opened 2022-04-19
Invalid logfmt output for URI field


domodwyer:
Our logfmt tracing formatter is not quoting the URI:

```
level=error msg="Error while handling request" error="internal error: dml handler error: unhandled sqlx error: pool timed out while waiting for an open connection" method=POST uri=/api/v2/write?bucket=06fddb4f912a0d7f&org=9df0256628d1f506&orgID=9df0256628d1f506&precision=ns content_length="Some(\"61548\")" target="ioxd_common::http" location="ioxd_common/src/http/mod.rs:147" time=1650382056819807317
```

PR #4353, opened 2022-04-19
fix(logfmt): quote strings containing '='

domodwyer:
Fixes https://github.com/influxdata/influxdb_iox/issues/4352

---

* fix(logfmt): quote strings containing '=' (8ec9d0ea)

      Logfmt uses '=' as a key/value separator, and prior to this commit the 
      formatter did not escape strings containing '='.

### DIFF
--- a/logfmt/src/lib.rs
+++ b/logfmt/src/lib.rs
@@ -285,9 +285,17 @@ fn needs_quotes_and_escaping(value: &str) -> bool {
         return true;
     }
 
-    let has_not_printable = value.bytes().any(|b| b <= b' ');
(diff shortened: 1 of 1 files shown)


pauldix (approved):

crepererum (approved):
I mean fluentbit could read `foo=bar=baz` as `foo="bar=baz"` but hey, why not do something funny instead :shrug:

domodwyer:
logfmt is ridiculously under-specified (or at least I've never found a solid spec)

I really think we should switch to something else which is a bit less ambiguous (JSON!)

PR #4355, opened 2022-04-19
feat: fuse ingester and catalog states in querier

crepererum:
This now correctly combines the data we get from the ingester w/ the
data we get from the catalog. Right now it bails out if during the very
small time windows between asking the ingester and querying the catalog
the compactor combines the newest files w/ "too new" files (see tests).


querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
+//! Logic to fuse the state that the querier got the from catalog and from the ingester.
```
crepererum, line 1: This is the core logic pulled out w/ some trait-based interface for simpler testing.

querier/src/table/mod.rs
```diff
@@ -425,4 +434,128 @@ mod tests {
         // file121: wrong sequencer
         assert_eq!(chunks[4].delete_predicates().len(), 0);
     }
+
+    #[tokio::test]
```
crepererum, line 438: The new tests here are deliberately shallow since we can rely on the tested core logic. It basically just confirms the error handling and that the filters of the core logic are really applied.

querier/src/table/mod.rs
```diff
@@ -251,29 +265,24 @@ impl QuerierTable {
     }
 }
 
-/// Validate that our current cache state is not stale given the
```
alamb, line 254: 👍

querier/src/table/mod.rs
```diff
@@ -425,4 +434,128 @@ mod tests {
...
+    async fn test_compactor_collision() {
+        let pred = Predicate::default();
+        let catalog = TestCatalog::new();
+
+        let ns = catalog.create_namespace("ns").await;
+        let table = ns.create_table("table").await;
+        let sequencer = ns.create_sequencer(1).await;
+        let partition = table.with_sequencer(&sequencer).create_partition("k").await;
+        table.create_column("foo", ColumnType::I64).await;
+        partition
```
alamb, line 448: Could you add. few comments here about why this partition will collide with the information from the `Ingester`? Is it because it has a greater max persisted number/

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    fn sequence_number(&self) -> SequenceNumber {
+        self.sequence_number
+    }
+}
+
+#[derive(Snafu, Debug)]
+pub enum FilterParquetError {
+    #[snafu(display("Compactor processed file that the querier would need to split apart"))]
```
alamb, line 101:
```suggestion
    #[snafu(display("Compactor processed file that the querier would need to split apart which is not yet implemented"))]
```

I think it is helpful to include in this message that this is a known limitation (rather than an unexpected error. But maybe I am over thinking things 🤔

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+    CompactorConflict,
+}
+
+/// Filter out parquet files that are "too new".
```
alamb, line 105: Could you please also add a description here of what "too new" means? As in summarize what the algorithm is trying to accomplish (not how it does so, the code is well commented in that regard)

querier/src/table/state_fuser.rs
alamb: I like the trait based interface for testing

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+                    && (file.min_sequence_number() <= persisted_max)
+                {
+                    return Err(FilterParquetError::CompactorConflict);
+                }
+                if file.max_sequence_number() > persisted_max {
+                    // filter out, file is newer
+                    continue;
+                }
+            } else {
+                // ingester thinks it doesn't have any data persisted yet => can safely throw away everything
```
alamb, line 141:
```suggestion
                // ingester thinks it doesn't have any data persisted yet => can safely ignore file
```

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+
+    Ok(result)
+}
+
+/// Generates "exclude" filter for tombstones.
+///
+/// Since tombstones are sequencer-wide but data persistence is partition-based (which are sub-units of sequencers), we
+/// cannot just remove tombstones entirely but need to decide on a per-partition basis. This function generates a lookup
+/// table of partition-tombstone tuples that later need to be EXCLUDED/IGNORED when pairing tombstones with chunks.
+pub fn filter_tombstones<I, T>(
```
alamb, line 159: the fact it is named similarly to `filter_parquet_files` might be confusing as the output means something different (aka an exclude list rather than include list)

How about calling this function something like `tombstone_exclude_list` to try and highlight the difference?

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+        .collect();
+
+    // we assume that we filter out a minimal amount of files, so we can use the same capacity
+    let mut result = Vec::with_capacity(parquet_files.len());
+
+    for file in parquet_files {
+        if let Some(ingester_partition) = lookup_table.get(&file.partition_id()) {
+            if let Some(persisted_max) = ingester_partition.parquet_max_sequence_number() {
+                if (file.max_sequence_number() > persisted_max)
+                    && (file.min_sequence_number() <= persisted_max)
```
alamb, line 132: I spent some time 🤔  about this and I think it makes sense that the file need to contain data that the ingester hadn't (yet) persisted as well as data that the ingester had already persisted

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+            partition_id: PartitionId::new(1),
+            min_sequence_number: SequenceNumber::new(3),
+            max_sequence_number: SequenceNumber::new(9),
+        };
+        let pf12 = MockParquetFileInfo {
+            partition_id: PartitionId::new(1),
+            min_sequence_number: SequenceNumber::new(10),
+            max_sequence_number: SequenceNumber::new(10),
+        };
+        let pf13 = MockParquetFileInfo {
```
alamb, line 261:
```suggestion
        // filtered because it was persisted after ingester sent response (11 > 10)
        let pf13 = MockParquetFileInfo {
```

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+            partition_id: PartitionId::new(1),
+            min_sequence_number: SequenceNumber::new(11),
+            max_sequence_number: SequenceNumber::new(20),
+        };
+        let pf2 = MockParquetFileInfo {
```
alamb, line 266: Shouldn't pf2 be included in `expected`? 

It seems to me like the scenario is that the ingester hasn't yet see new data from Partition 2 but `pf2` was  persisted on a prior startup (I may be misunderstanding the scenario, but I am trying to turn the code into words as a way to double check the logic)

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+            partition_id: PartitionId::new(2),
+            min_sequence_number: SequenceNumber::new(0),
+            max_sequence_number: SequenceNumber::new(0),
+        };
+        let pf31 = MockParquetFileInfo {
+            partition_id: PartitionId::new(3),
+            min_sequence_number: SequenceNumber::new(1),
+            max_sequence_number: SequenceNumber::new(3),
+        };
+        let pf32 = MockParquetFileInfo {
```
alamb, line 276:
```suggestion
        // filtered because it was persisted after ingester sent response (4 > 3)
        let pf32 = MockParquetFileInfo {
```

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+            partition_id: PartitionId::new(3),
+            min_sequence_number: SequenceNumber::new(4),
+            max_sequence_number: SequenceNumber::new(5),
+        };
+        let pf4 = MockParquetFileInfo {
```
alamb, line 281:
```suggestion
        // passed because it came from a partition (4) the ingester didn't know about 
        let pf4 = MockParquetFileInfo {
```

querier/src/table/state_fuser.rs
```diff
@@ -0,0 +1,472 @@
...
+                sequence_number: SequenceNumber::new(8),
+            },
+            MockTombstoneInfo {
+                id: TombstoneId::new(10),
+                sequencer_id: SequencerId::new(3),
+                sequence_number: SequenceNumber::new(10),
+            },
+        ];
+
+        let actual = filter_tombstones(ingester_partitions, tombstones);
```
alamb, line 387: It would be great if @NGA-TRAN  could double check the tombstone logic -- I haven't been fully following along how they are implemented so I don't feel like I can give this a great review

alamb (approved):
Looks good 🏅 

I found the name `fuse` confusing, but perhaps that is because I have seen it in [`Futures::fuse`](https://docs.rs/futures/latest/futures/future/struct.Fuse.html) 

Another possible name for your consideration is `reconcile` as in `state_reconcile`

NGA-TRAN:
I will look into this first thing tomorrow. My brain is full now. Feel free to commit and address my comments if any as a follow-up PR

querier/src/table/state_fuser.rs
crepererum: There is a difference between pf2 and pf4:

- pf2: the ingester HAS seen the partition but has NOT persisted anything yet (the max sequencer number is `None`). Since no data was persisted yet, we must assume that the parquet file here was persisted after we've talked to the ingester. Remember that the "max persisted parquet sequence number" is inclusive and this edge case is the reason why (within protobuf and Rust) this is optional (we could also use `-1`, but I would find that more confusing).
- pf4: the ingester doesn't know about the partition at all, i.e. no data was flowing throw it => we can (somewhat safely, it's not 100% bullet-proof) assume that the ingester has fully persisted this partition its previous "life".

querier/src/table/state_fuser.rs
crepererum: done (extended the doc string)

querier/src/table/mod.rs
crepererum: done (added some ascii art)

crepererum:
> Another possible name for your consideration is `reconcile` as in `state_reconcile`

:+1: done

querier/src/table/state_fuser.rs
crepererum: done

querier/src/table/state_reconciler.rs
```diff
@@ -0,0 +1,485 @@
...
+pub enum FilterParquetError {
+    #[snafu(display("Compactor processed file that the querier would need to split apart which is not yet implemented"))]
+    CompactorConflict,
+}
+
+/// Filter out parquet files that contain "too new" data.
+///
+/// The caller may only use the returned parquet files.
+///
+/// This will remove files that are part of the catalog but that contain data that the ingester persisted AFTER the
```
alamb, line 109: ❤️

alamb: thank you -- this is great

querier/src/table/state_fuser.rs
alamb: > `pf2` ... we must assume that the parquet file here was persisted after we've talked to the ingester. 

In my mind `pf2` models the scenario where the ingester:
1.  had persisted the parquet file during a previous startup
2. was restarted
3. had not yet read any data for his partition from the write buffer

In this case, I would expect that the parquet file would be included in query results

I may be misunderstanding how the ingester works

crepererum: > 1. had persisted the parquet file during a previous startup
> 2. was restarted
> 3. had not yet read any data for his partition from the write buffer

This is pf4. The ingester won't include that partition as "unpersisted partition" (or anywhere else in the gRPC response). We're free to read all parquet files in that case.

I think you might confuse that with the following scenario:

1. had read data but has NOT persisted yet
2. was restarted
3. had not yet read any data for his partition from the write buffer

That's an interesting edge case. The ingester doesn't have an explicit replay step (it just seeks the write buffer and then does its normal thing module some on-demand data skipping). So the ingester doesn't "remember" that it had anything buffered up in memory before it died/restarted, because otherwise every read from the write buffer would require it to perform some write to the catalog (or any other watermark memory). So from the querier POV the "readable" watermark can jump backwards. (the persisted watermark however will never go backwards!)

Side-note: OG also doesn't remember in-mem states in all cases. It just persist watermarks on every catalog write but if it doesn't persist for a while, then the in-mem state can also jump backwards between restarts (even if replay is performed).

alamb: Ah yes, I see now -- the fact that partition 2 was returned from the ingester meant that it knew about that partition but didn't have any data persisted yet. Makes sense. Thank you for the clarification

querier/src/chunk/mod.rs
```diff
@@ -64,6 +67,11 @@ impl ChunkMeta {
         self.sequencer_id
     }
 
+    /// Partition ID.
+    pub fn partition_id(&self) -> PartitionId {
+        self.partition_id
+    }
```
NGA-TRAN, line 73: Hah, yesterday I added this function into my PR #4360 (that will be ready for review later today/tomorrow) but later I replaced it with `partition_sort_key`. Anyway, We both see we need this partition_id

querier/src/table/mod.rs
```diff
@@ -120,15 +132,10 @@ impl QuerierTable {
...
-        // Validate that the cache contents is consistent with the
-        // parquet files we know about
-        if !validate_cache(&ingester_partitions, &parquet_files) {
-            // TODO real error / retry
-            panic!("returned ingester partitions are inconsistent with our parquet files");
-        }
+        // fuse ingester and catalog state
+        let parquet_files =
+            filter_parquet_files(&ingester_partitions, parquet_files).context(StateFusionSnafu)?;
+        let tombstone_exclusion = tombstone_exclude_list(&ingester_partitions, &tombstones);
```
NGA-TRAN, line 138: 👍

PR #4357, opened 2022-04-19
chore: Update arrow, arrow-flight, parquet, tonic, prost, etc

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/4358~

# Rationale
Pull in changes from arrow 12 including updates to prost and tonic and dependents

Closes https://github.com/influxdata/influxdb_iox/pull/4318
Closes https://github.com/influxdata/influxdb_iox/pull/4362
Closes https://github.com/influxdata/influxdb_iox/pull/4363
Closes https://github.com/influxdata/influxdb_iox/pull/4364

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/7548e961b58f75fc26fd4fb16c99377f6f468fc9
2. Upgrade to arrow / parquet / arrow-flight 12
3. Update heappy: https://github.com/mkmik/heappy/pull/38 (thanks @mkmik !)

PR #4358, opened 2022-04-19
chore: Install custom protoc in perf-builder image

alamb:
As part of updating to a newer version of prost / tonic in https://github.com/influxdata/influxdb_iox/pull/4357 it turns out the protoc installed by `apt-get install protobuf-compiler` is too old. We need a newer one in the test_perf builder

Install version 3.19.4

Without this fix, #4357 fails with https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18209/workflows/39a6d02d-d680-42d3-baf3-ba17e3938601/jobs/151180

```
--- stderr
  Error: Custom { kind: Other, error: "protoc failed: Unknown flag: --experimental_allow_proto3_optional\n" }
warning: build failed, waiting for other jobs to finish...
error: build failed

Exited with code exit status 101
CircleCI received exit code 101
```


.circleci/config.yml
```diff
@@ -304,6 +306,14 @@ jobs:
           command: sudo apt install curl clang lld pkg-config libssl-dev --yes --no-install-recommends
       - checkout
 
+      - run:
+          name: Install protoc 3.19.4 to ~/protoc (stock in ubunutu is too old)
```
alamb, line 310:
```suggestion
          name: Install protoc 3.19.4 to ~/protoc (stock in Ubuntu is too old)
```

tustvold (approved):
LGTM, although it makes me somewhat sad that we're relying on functionality not in mainline versions of protoc...

alamb:
> LGTM, although it makes me somewhat sad that we're relying on functionality not in mainline versions of protoc...

I believe this is a tonic requirement, no?

alamb:
@tustvold  has a suggestion that would avoid needing this: https://github.com/influxdata/influxdb_iox/pull/4357#issuecomment-1103050828

alamb:
> @tustvold has a suggestion that would avoid needing this: https://github.com/influxdata/influxdb_iox/pull/4357#issuecomment-1103050828

Tried and we still probably need it

crepererum (approved):
Look OK. I however think we should probably get rid of the `test_perf` special snowflake at some point. We now have 3 CI environments to build IOx (and two should be enough):

- prod image build: uses top-level multi-stage `Dockerfile` (which IMHO is good because it allows everyone to produce clean IOx builds using a single CLI command)
- `test_perf`: completely hand-rolled
- all other steps: use [`docker/Dockerfile.ci`](https://github.com/influxdata/influxdb_iox/blob/main/docker/Dockerfile.ci)-based images (which contain a bit more than what we would need for a prod build and are InfluxData/CI-specific and not meant to be used by end users, even though they contain no secrets)

alamb:
Removing this special ❄️  would be ideal

Issue #4158, opened 2022-03-29
Support split_overlap_chunks using only min and max of time column


NGA-TRAN:
# Rationale

In OG, we use full statistics to find overlap chunks in [split_overlapped_chunks](https://github.com/influxdata/influxdb_iox/blob/886f9dc8c1ab3fa7be4672e7457ed8d5fd07b6b0/query/src/provider.rs#L413) for deduplication. In NG, we do not have full statistics but we do have min and max for time column which we think is good enough find overlaps. This  ticket is to support `split_overlapped_chunks` but on limited min_time and max_time only.

We suspect this can be done by modifying `overlap.rs` as discussed in #4082

PR #4360, opened 2022-04-19
feat: use stored partition sort key to deduplicate data

### LINKED ISSUE #4285: querier: use the stored sort key to deduplicate data
# Rationale
This is a feature of epic #4283.
The NG data chunks (parquet files) are sorted consistently and the `super sort key` is stored in their partition catalog. This ticket is when deduplicate data, we need to use that sort key instead.

### Design
The deduplication happens during scan and the `sort key` is provided for that scan here https://github.com/influxdata/influxdb_iox/blob/581d277f75de4e145083f8c8407766191df7349d/query/src/provider.rs#L282
However, the `sort key` is actually set earlier to the Scans' **ProviderBuilder** here https://github.com/influxdata/influxdb_iox/blob/581d277f75de4e145083f8c8407766191df7349d/query/src/provider.rs#L168
I suggest to read the `sort_key` from the catalog (or pass it if it is already available at this time) while building the ProviderBuilder. We do want to remove columns from the sort_key that are not included in the scan data.

NGA-TRAN:
Closes #4285 

# Rationale
This PR is an attempt to avoid re-sorting parquet files that are in the same partition. The case it addresses is something like:

```
Schema: A, B, C, D
Partition Sort Key: A, B, C, D
```

```
Chunk 1: Sort Key A, B (contains only columns A and B)
Chunk 2: Sort Key C, D (contains only columns C and D)
```

In this case, prior to this PR the chunks will be resorted prior to deduplication. 

This PR threads through the Partition Sort Key so that the chunks can avoid resorting prior to deduplication. 


NGA-TRAN:
Need to add some tests before turning this on for review

query/src/frontend/influxrpc.rs
```diff
@@ -1424,6 +1424,10 @@ impl InfluxRpcPlanner {
     ///   Filter(predicate) [optional]
     ///     Scan
     /// ```
+    // TODO: this function is only invokded by InfluxRpcPlanner's functions. Ask Edd to see
+    //   whether we have any knowledge about sortkey of those chunks. If yes, this function
+    //   needs to get refactored to provide sort_key to its `builder` (ProviderBuilder) to
+    //   make sure we do not add unecessary sort operator (needed for dara deduplication) into the plan during planning
```
NGA-TRAN, line 1430: @e-dard : Question for you

querier/src/chunk/mod.rs
```diff
@@ -230,10 +238,22 @@ impl ParquetChunkAdapter {
         let order = ChunkOrder::new(1 + iox_metadata.min_sequence_number.get() as u32)
             .expect("cannot be zero");
 
+        // Read partition sort key
+        // TODO: Ask Marco whether where to move this code and how to read it from CatalogCache (it seesms the cache has enough info for me to read this yet)
```
NGA-TRAN, line 242: @crepererum : Question for you

query/src/frontend/reorg.rs
```diff
@@ -57,6 +57,17 @@ impl ReorgPlanner {
     }
 
     /// Creates an execution plan for a scan and filter data of a single chunk
+    // Notes on 2022.04.8 for developers: So far this function is only invoked by either
+    //   1. OG lifecycle's load_chunk which will go way
+    //   2. NG ingester
+    //  In both places, data is coming from kafka which is not yet sorted and may contain duplicates.
+    //   Hence, this function will create a plan with a sort operator added on top of the scan operator for the deduplication work.
+    // This function should NOT be used on sorted chunks. However, if you need to use it for chunks that are sorted on the SAME key,
```
alamb, line 65: do I read this comment as it is making assumptions on how the output is sorted? If the output needs to be sorted a certain way I suggest passing the sort key into this function rather than relying on the output sortedness of `LogicalPlanBuilder::scan_with_filters`

query/src/frontend/reorg.rs
```diff
@@ -221,7 +232,7 @@ impl ReorgPlanner {
         let mut builder = ProviderBuilder::new(table_name, schema)
             // There are no predicates in these plans, so no need to prune them
             .add_no_op_pruner()
-            // Tell the scan of this provider to sort its output on the chunks' PK
+            // Tell the scan of this provider to sort its output on the given sort_key
```
alamb, line 235: 👍

query/src/provider.rs
```diff
@@ -436,61 +437,114 @@ impl Deduplicater {
...
                     overlapped_chunks,
                     predicate.clone(),
-                    &dedup_sort_key,
+                    &chunks_dedup_sort_key,
                 )?);
             }
 
             // Go over each in_chunk_duplicates_chunks, build deduplicate plan for each
             for chunk_with_duplicates in self.in_chunk_duplicates_chunks.iter().cloned() {
+                // Set dedup sort key
```
alamb, line 507: I recommend pulling this into a function to reuse

query/src/provider.rs
```diff
@@ -436,61 +437,114 @@ impl Deduplicater {
...
+                    // Use output _sort_key if provided
+                    Some(sort_key) => sort_key.clone(),
+                    None => {
+                        // Use the chunk's partition key if they were persisted chunks
+                        if let Some(sort_key) =
+                            Self::sort_key_of_overlapped_chunks(&overlapped_chunks)
+                        {
+                            sort_key
+                        } else {
+                            // This is a valid case but we are trying to not let this happen.
```
alamb, line 478: I find this confusing -- if it is a valid case, it shouldn't be warning. If it not a valid case then we should `panic!` right?

query/src/provider.rs
```diff
@@ -436,61 +437,114 @@ impl Deduplicater {
...
-                None => compute_sort_key_for_chunks(&pk_schema, chunks.as_ref()),
-            };
-
-            debug!(
-                ?dedup_sort_key,
-                ?output_sort_key,
-                ?pk_schema,
-                "Sort keys while building"
-            );
+            // Verify that output_sort_key must cover PK
```
alamb, line 440: I don't understand the reference to `"building deduplicate plan"` in the debug messages -- this code is used to create scans for normal queries as well, right?

query/src/provider.rs
```diff
@@ -531,6 +585,58 @@ impl Deduplicater {
...
+            return None;
+        }
+
+        // At least one of the overlapped chunks are not sorted
+        if chunks.iter().any(|c| c.sort_key() == None) {
+            debug!("At least one of the overlapped chunks are not sorted");
+            return None;
+        }
+
+        // If reach here, all overlapped chunks must be sorted
```
alamb, line 600: I don't understand why / how this should be the case -- if something above this code is trying to avoid resorting chunks that span partitions, why doesn't it just pass in the partition sort key as the desired output sort

alamb (commented):
I don't understand the need for this code. Is it possible to write some test in the compactor (or maybe an end to end test) that fails without these changes? The sort key logic is getting super convoluted which doesn't seem right to me....

query/src/provider.rs
```diff
@@ -531,6 +585,58 @@ impl Deduplicater {
...
+        // --> they must come from persisted parquet files of the same partition
+        // Get their partition's sort_key
+        let partition_sort_key = chunks[0].partition_sort_key();
```
alamb, line 603: Maybe as a extra check we can assert that the partition key for all chunks are the same (i believe after talking with @NGA-TRAN  this is an assumption that the code is making, but it would be nice to assert it)

NGA-TRAN:
@tustvold : if you have sometime, can you help review this? You knows this code very well

alamb (approved):
I went over the intent of this PR with @NGA-TRAN  in person -- (thanks for that @NGA-TRAN)

My gut feel is that this PR is  more complicated than it needs to be -- the assumptions the planner has to make  about chunks and `partitions`s and how they can/can't overlap is quite complicated

After we remove OG, perhaps we can teach the Deduplicator about partitions (rather than give it a Vec<Chunk> we could give it a Vec<(Partition, Chunk)>` or something and then the `partition_sort_key` would not be duplicated across partitions

querier/src/ingester/mod.rs
```diff
@@ -244,6 +244,10 @@ impl QueryChunkMeta for IngesterPartition {
         Arc::clone(&self.schema)
     }
 
+    fn partition_sort_key(&self) -> Option<&SortKey> {
+        None // data comes from Ingester is not persisted yet and should not yet attached to any catalog partition
```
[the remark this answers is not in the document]
NGA-TRAN: And this `None` value will signal the deduplication planning to add a sort operator to sort data

query/src/frontend/influxrpc.rs
NGA-TRAN: I will create a ticket to verify if the necessary sort is added

query/src/frontend/reorg.rs
NGA-TRAN: This is a good point. I think when we implemented before, DF/IOx did not provide a way for us to see if the output of a plan is sorted or not and on which sort_key. If we have that knowledge now, this will be simpler

query/src/provider.rs
NGA-TRAN: Yes, this code is used while scanning and compacting/deduplicating data for Ingester, Compactor, and Querier. They all do deduplicating

query/src/provider.rs
NGA-TRAN: It is valid but we want to optimize it out. I put the debug message here just for us to explore while we optimize the work which I am doing right now

query/src/provider.rs
NGA-TRAN: They are not exactly the same but I will refactor whatever I can

query/src/provider.rs
NGA-TRAN: I will do this check. FYI, I at first wanted to do that but then I sued the code below to check this is a super sort key of all the chunk sort key instead

query/src/provider.rs
NGA-TRAN: The problem is we need to cover both OG and NG now. MUB of OG is not sorted, right?

NGA-TRAN:
@alamb 

> My gut feel is that this PR is more complicated than it needs to be -- the assumptions the planner has to make about chunks and partitionss and how they can/can't overlap is quite complicated

> After we remove OG, perhaps we can teach the Deduplicator about partitions (rather than give it a Vec we could give it a Vec<(Partition, Chunk)>or something and then thepartition_sort_key` would not be duplicated across partitions

I have the same gut feelings and agree with the plan

querier/src/ingester/mod.rs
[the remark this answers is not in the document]
crepererum: The ingester already de-duplicated the data is I would expect that the data is sorted. It's just that we don't transmit this information in the flight response, no?

querier/src/chunk/mod.rs
crepererum: We already read the entire partition here:

https://github.com/influxdata/influxdb_iox/blob/c084785bc38edde4766292e343d848d8ae6b5d46/querier/src/cache/partition.rs#L24

Just add the sort key here

https://github.com/influxdata/influxdb_iox/blob/c084785bc38edde4766292e343d848d8ae6b5d46/querier/src/cache/partition.rs#L67-L71

and:

- add a new method `sort_key` method to the partition cache
- add a test concerning the `sort_key`
- extend `test_cache_sharing` in the same module by a 3rd partition and demonstrate that fetching the sort key also caches the two other attributes (old gen partition key, sequencer ID) and the other way around (so the 2x2 "matrix" becomes a 3x3 "matrix").

Note that the cache currently doesn't handle partition deletion, but this doesn't need to concern you (I'll fix that at some point).

crepererum (commented):
My review concerns mostly the `querier` part, for which I have two nitpicks (see comments): the "is the ingester data not sorted" and the  cache thingy.

query/src/frontend/influxrpc.rs
e-dard: Given all the extra discussion on the PR is this question still valid, or is there a new question? :-)

query/src/provider.rs
alamb: maybe leave a ticket reference for future readers benefit as well

query/src/provider.rs
alamb: I can't remember

querier/src/ingester/mod.rs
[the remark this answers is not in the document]
NGA-TRAN: This is a good point @crepererum. I think here is a few things I will need to consider in this case:
1. Data may be already sorted but on different sort key and we do not know its sort key at this time, either. I will create a ticket for this optimization.
2. If this chunk overlaps with others, we still need to deduplicate its data with other overlapped chunks.
3. If this chunk does overlap with any, we do not need to deduplicate again it. There is aviable code for this, we just need some extra code to put this chunk into the   right category.

I will add numbers 2 and 3 in either in this PR or the PR to do the work for #4158

query/src/frontend/influxrpc.rs
NGA-TRAN: I think we are all set for this for now @e-dard

query/src/provider.rs
NGA-TRAN: @alamb: yes, will do
The support both OG and NG make things complicated here. I will reevaluate this PR and see what extra or refactor should be done before merging it

NGA-TRAN (commented):
@alamb and @crepererum 
Given your comments and what I have thought a lot last night, this PR needs more love. Here is what I will do
1. This PR depends on #4158. I will work on that first either in the same PR or highly different PR.
2. I may need to point out clearly which part need OG only, NG only and both then the code will be cleaer
3. I may need to refactor using (partition, list of chunks) as Andrew suggests

query/src/frontend/influxrpc.rs
```diff
@@ -1424,6 +1424,9 @@ impl InfluxRpcPlanner {
     ///   Filter(predicate) [optional]
     ///     Scan
     /// ```
+    // TODO: Verify if the input chunks are sorted on the same super-key. If so,
+    // provide that sort key to the ProviderBuilder to make sure we do not add unecessary
+    // sort operator (needed for dara deduplication) into the plan during planning
```
NGA-TRAN, line 1429: Will create a ticket and link it here after this PR is merged

querier/src/chunk/mod.rs
NGA-TRAN: Done
@crepererum : The steps are clear but if you have time, can you have a look?

NGA-TRAN:
Overlaps this PR has fully used #4389

query/src/frontend/influxrpc.rs
NGA-TRAN: I have verify and we do not need to provide sort key. If output sort key is not provided, the query plan will go and find the best sort key for this using the chunks and partition sort key.

PR #4367, opened 2022-04-20
feat: panic metrics

domodwyer:
This PR adds a `thread_panic_count_total` metric to the various server types so we can measure the frequency of, and alert on panics.

I was going to keep the logging-to-tracing and metric recorder concerns separate and compose the impls as needed, but the drop ordering becomes important when setting multiple panic handlers and it was overly brittle / easy to cock up, so it's an optional behaviour of the existing custom panic handler instead 👍 

---

* feat: emit thread panic count metric (49c38c297)

      Allows the panic handler to be optionally configured to emit a thread panic
      metric, increasing by 1 each time a panic is observed.

* refactor: emit panic metrics for server types (c36324290)

      Configures the long-running / server modes to emit panic metrics.

### DIFF
--- a/panic_logging/src/lib.rs
+++ b/panic_logging/src/lib.rs
@@ -11,6 +11,7 @@
 
 use std::{fmt, panic, sync::Arc};
 
+use metric::U64Counter;
 use observability_deps::tracing::{error, warn};
 use panic::PanicInfo;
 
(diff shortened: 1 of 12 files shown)


mkmik (approved):
Nice!

crepererum (approved):

panic_logging/src/lib.rs
```diff
@@ -97,3 +110,36 @@ fn tracing_panic_hook(other_hook: &PanicFunctionPtr, panic_info: &PanicInfo<'_>)
...
+
+        assert_count(&metrics, 0);
+
+        std::thread::spawn(|| {
+            panic!("it's bananas");
+        })
+        .join()
+        .expect_err("wat");
+
+        assert_count(&metrics, 1);
```
alamb, line 143: 👍

PR #4371, opened 2022-04-20
test: include materialized delete predicates in NG query tests

crepererum:
Testing that we can write pre-filtered parquet files might seem like a boring test, but is the first step to mock/include the ingester into the NG query tests.


datafusion_util/src/lib.rs
```diff
@@ -288,6 +292,33 @@ pub async fn test_collect_partition(
         .unwrap()
 }
 
+/// Filter data from RecordBatch
```
alamb, line 295: this is just moved, as I understand it

iox_tests/src/util.rs
```diff
@@ -435,12 +435,43 @@ impl TestPartition {
...
+            max_seq,
+            min_time,
+            max_time,
+            None,
+            creation_time,
+        )
+        .await
+    }
+
+    /// Create a parquet for the partition
```
alamb, line 457:
```suggestion
    /// Create a parquet with the data in the specified `record_batch` for the partition
```

iox_tests/src/util.rs
```diff
@@ -511,68 +542,23 @@ impl TestPartition {
...
+        let record_batch = batch.to_arrow(Selection::All).unwrap();
+
+        self.create_parquet_file_with_batch(
+            record_batch,
+            schema,
+            min_seq,
+            max_seq,
+            min_time,
+            max_time,
+            Some(file_size_bytes),
```
alamb, line 558: I don't understand the usecase of passing in `file_size_bytes` -- it seems like the only thing that could do is be inconsistent with the actual parquet file size that is created.

query_tests/src/scenarios/util.rs
```diff
@@ -1188,52 +1231,126 @@ async fn make_ng_chunk(ns: Arc<TestNamespace>, chunk: ChunkDataNew<'_, '_>) -> S
...
-
-                    let min_time = pred.predicate.range.start();
-                    let max_time = pred.predicate.range.end();
-                    let predicate = pred.predicate.expr_sql_string();
-                    table
-                        .with_sequencer(&sequencer)
-                        .create_tombstone(sequence_number as i64, min_time, max_time, &predicate)
+        match chunk_stage {
+            ChunkStageNew::Parquet => {
+                // process delete predicates that are materialized before/during parquet file creation
```
alamb, line 1245:
```suggestion
                // model delete predicates that are materialized (applied) by the ingester,
                // during parquet file creation
```

query_tests/src/scenarios/util.rs
```diff
@@ -1188,52 +1231,126 @@ async fn make_ng_chunk(ns: Arc<TestNamespace>, chunk: ChunkDataNew<'_, '_>) -> S
...
+                        let delete_table = tables.get(delete_table_name).unwrap();
+
+                        for (i, pred) in chunk.preds.iter().enumerate() {
+                            match pred.delete_time {
+                                DeleteTimeNew::Ingester {
+                                    also_in_catalog: false,
+                                } => {
+                                    // was already materialized
+                                }
+                                DeleteTimeNew::Ingester {
```
alamb, line 1308:
```suggestion
                                // model deletes that were already applied by ingester prior to writing
                                // parquet file but ALSO still in the catalog
                                DeleteTimeNew::Ingester {
```

query_tests/src/scenarios/util.rs
```diff
@@ -238,27 +241,59 @@ impl DeleteTimeOld {
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum DeleteTimeNew {
-    /// Delete  predicate is added to chunks at their parquet stage
+    /// Delete predicate is added while chunk is was still in ingester memory.
+    Ingester {
+        /// Flag if the tombstone also exists in the catalog.
+        ///
+        /// If this is set to `false`, then the tombstone was already pruned.
```
alamb, line 248:
```suggestion
        /// If this is set to `false`, then the tombstone was already removed from the
        /// catalog by the ingester
```

alamb (approved):
LGTM

iox_tests/src/util.rs
crepererum: Just had a look: The compactor uses that in its unit tests because there's compactor file size threshold. I agree that it is potentially dangerous but I think the concrete application is OK.

query_tests/src/scenarios/util.rs
crepererum: The ingester wouldn't remove them, but the compactor could. I'll adjust the comment.

crepererum: done

PR #4380, opened 2022-04-21
refactor: use `info!` rather `println!` in end to end tests

alamb:
# Rationale
So I can get timestamps to correlate the harness output with what was happening on the server

# Changes:
see subject

You can now see logs with a command like this:
```shell
RUST_LOG=debug,sqlx=warn,h2=warn,hyper=warn TEST_INTEGRATION=1 TEST_INFLUXDB_IOX_CATALOG_DSN=postgresql://localhost:5432/alamb RUST_BACKTRACE=1 cargo test --test end_to_end_ng 
```


carols10cents (changes requested):
It looks like the commit message is backwards and incomplete, currently it says "refactor: use println rather" as opposed to the PR title that says "refactor: use info! rather println!"

I also see a few more printlns that I think should be changed? On this branch I searched for `println` in influxdb_iox/tests/end_to_end_ng_cases/* and found:

```
influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
148:                        println!("All data is readable: {:?}", combined_response);
151:                        println!("retrying, not yet readable: {:?}", combined_response);
155:                    println!("retrying, error getting token status: {}", e);
```

I didn't find any printlns in `test_helpers_end_to_end_ng`, but there were some `eprintln`s? Not sure if these should be changed too.

```
test_helpers_end_to_end_ng/src/lib.rs
51:                eprintln!("skipping NG end-to-end integration tests - set TEST_INTEGRATION to run");
55:                eprintln!(
```

alamb:
I think the `eprintln` should stay to hint to anyone who runs this what is happening even if they don't enable logging. 

The println is a good call -- will fix. Thank you

crepererum (approved):

Issue #4368, opened 2022-04-20
End to end tests: Consider re-using clusters.


alamb:
This is not specific for this PR but a more general concern regarding NG: We're currently not re-using clusters. Esp. when we have many CLI and API tests, this likely increases test times quite a lot. For OG we basically use just a few cached server processes and randomized databases (which would be namespaces for NG). Let's see if we need such an approach for NG as well. IMHO it would be better if the NG cluster creation would be fast enough to avoid that hack.

_Originally posted by @crepererum in https://github.com/influxdata/influxdb_iox/pull/4309#discussion_r853904975_

PR #4381, opened 2022-04-21
refactor: Move Connection in ServerFixture to make TestServer shareable

alamb:
# Rationale
A part of https://github.com/influxdata/influxdb_iox/issues/4368 (not starting new processes for each test). 

It turns out that you can't try and make requests over the same `Connection` object, so when I tried to share `TestServer`s which have cached `Connection`s this turned out to be a problem.

# Changes
1. Cache connections on `ServerFixture` rather than `TestServer`

This is the same design as in the OG end to end tests [code link](https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/common/server_fixture.rs#L84-L87) which I did not appreciate when I ported this over

* Shared server clusters: alamb/share_server_clusters https://github.com/influxdata/influxdb_iox/issues/4368


test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -18,10 +18,12 @@ use crate::{database::initialize_db, server_type::AddAddrEnv};
 use super::{addrs::BindAddresses, ServerType, TestConfig};
 
 /// Represents a server that has been started and is available for
-/// testing.
+/// testing. This structure can not be shared between tests as the
+/// connections are not shareable
```
carols10cents, line 22: This sentence sounds a bit like a tautology but I'm not sure how to fix it... not a huge deal

test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -293,31 +328,32 @@ impl TestServer {
...
-                    return;
-                }
-                ServerState::Ready => {
-                    return;
-                }
-                ServerState::Error => {
-                    panic!("Server was previously found to be in Error, aborting");
                 }
+                ServerState::Ready => {}
+                ServerState::Error => {}
```
carols10cents, line 347: Is this correct? The old code would panic if the server was in an error state, and now it continues? Is it possible for the server to get into an error state? Is this not completed yet?

carols10cents (approved):
Couple of thoughts/questions but nothing blocking.

test_helpers_end_to_end_ng/src/server_fixture.rs
alamb: I'll try and make it clearer.

test_helpers_end_to_end_ng/src/server_fixture.rs
alamb: that is an excellent spot -- restored the panic in6fdab5acf

PR #4384, opened 2022-04-21
refactor: less `async` shenanigans in the end to end tests

alamb:
This is a 10 line change if whitespace is ignored: https://github.com/influxdata/influxdb_iox/pull/4384/files?w=1

# Rationale
I found there were some unnatural `async move` required while porting the influxrpc test -- this is because the custom closure function was marked `Fn` ( in theory be called multiple times) even though it is only ever called once

# Changes
Make the custom function `FnOnce` (as it is only called once) and remove the shenanigans


test_helpers_end_to_end_ng/src/steps.rs
```diff
@@ -64,7 +64,7 @@ impl<'a> StepTestState<'a> {
 ///   }.boxed()
 /// });
 /// ```
-pub type FCustom = Box<dyn for<'b> Fn(&'b mut StepTestState) -> BoxFuture<'b, ()>>;
+pub type FCustom = Box<dyn for<'b> FnOnce(&'b mut StepTestState) -> BoxFuture<'b, ()>>;
```
alamb, line 67: here is the change

influxdb_iox/tests/end_to_end_ng_cases/querier/influxrpc/read_filter.rs
```diff
@@ -25,35 +25,35 @@ use super::{data::DataGenerator, exprs};
...
-            "SeriesFrame, tags: _measurement=cpu_load_short,host=server01,region=us-west,_field=value, type: 0",
-            "FloatPointsFrame, timestamps: [ns0, ns4], values: \"0.64,0.000003\"",
-            "SeriesFrame, tags: _measurement=swap,host=server01,name=disk0,_field=in, type: 1",
-            "IntegerPointsFrame, timestamps: [ns6], values: \"3\"",
-            "SeriesFrame, tags: _measurement=swap,host=server01,name=disk0,_field=out, type: 1",
-            "IntegerPointsFrame, timestamps: [ns6], values: \"4\""
-        ]);
-
         async move {
+            let mut storage_client =
```
alamb, line 29: previously the tests had to be split up between "run once" part and the "could in theory run many times" part which is silly

influxdb_iox/tests/end_to_end_ng_cases/querier/influxrpc/read_filter.rs
```diff
@@ -179,9 +179,6 @@ async fn do_read_filter_test(
             Step::WriteLineProtocol(line_protocol),
             Step::WaitForReadable,
             Step::Custom(Box::new(move |state: &mut StepTestState| {
-                let range = range.clone();
```
alamb: this is what was triggering my suspicion -- having to clone this a few times...

carols10cents (approved):

PR #4386, opened 2022-04-21
feat: Port end-to-end tracing tests to NG end-to-end tests

carols10cents:
Connects to #4102.

Adds some infrastructure to the test helper crate; there is some more that could maybe be shared with the NG querier RPC tests, but I wasn't sure if that was worth it or not.

The port is done in 2 commits: one where I just ran `cp influxdb_iox/tests/end_to_end_cases/tracing.rs influxdb_iox/tests/end_to_end_ng_cases` and committed even though it doesn't compile at all, and then another commit that actually does the port, so that if you review this PR commit-by-commit it should hopefully be easier to see what actually changed in the port.

### DIFF
--- a/influxdb_iox/tests/end_to_end_ng_cases/tracing.rs
+++ b/influxdb_iox/tests/end_to_end_ng_cases/tracing.rs
@@ -1,0 +1,162 @@
+use super::scenario::Scenario;
+use crate::common::{
+    server_fixture::{ServerFixture, ServerType, TestConfig},
+    udp_listener::UdpCapture,
+};
+use futures::TryStreamExt;
+use generated_types::{storage_client::StorageClient, ReadFilterRequest};
+use influxdb_iox_client::flight::generated_types::ReadInfo;
+use std::num::NonZeroU32;
+
+async fn setup() -> (UdpCapture, ServerFixture) {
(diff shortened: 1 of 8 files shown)


influxdb_iox/tests/end_to_end_ng_cases/tracing.rs
```diff
@@ -0,0 +1,181 @@
...
+    // wait for the UDP server to shutdown
+    udp_capture.stop().await;
+}
+
+#[tokio::test]
+pub async fn test_tracing_create_trace() {
+    let database_url = maybe_skip_integration!();
+    let table_name = "the_table";
+    let udp_capture = UdpCapture::new().await;
+    let test_config = TestConfig::new_all_in_one(database_url)
```
alamb, line 140: An in one cluster is probably reasonable for now 👍  I think in general we would like to avoid using all in one to test as much as possible as it is not *quite* the same as what is deployed in production.

FYI @crepererum

alamb (approved):
Wunderbar -- thank you @carols10cents

alamb:
Since there are several end to end tests outstanding I want to get this one merged -- so hopefully that is not a problem @carols10cents

PR #4387, opened 2022-04-21
feat: share server processes in end to end test

alamb:
# Rationale
closes https://github.com/influxdata/influxdb_iox/issues/4368 and hopefully reduces some of the flakiness we see from #4395 

1. Creating 3 new processes for each test is quite slow
2. Sometimes locally I run into a port collision when running a bunch of the end to end tests
3.  Also I have evidence that starting so many processes that try to connect to postgres at once causes us to DDOS it

Example failure for postgres DDOS: https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18654/workflows/fa092f61-6512-4b78-bbf1-77970730b5ee/jobs/156142
```
2022-04-28T15:10:26.073681Z ERROR ingester::stream_handler::handler: failed to apply dml operation error=Error accessing catalog: unhandled sqlx error: error returned from database: sorry, too many clients already kafka_topic=iox-shared kafka_partition=1 potential_data_loss=true
```

# Changes
1 . Reuse ServerProcesses when possible
2. Stop pretending there is any difference between "fast_parquet_generation" and "non_fast_parquet_generation"

# Timing if the `influxrpc` tests:

## Before this PR 
```
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.79s
```

## With this PR: 🚀 

```
test result: ok. 31 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.73s
```

## Example runs from circle CI showing the tests doing better:
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18655/workflows/03166175-82a4-4961-971c-2d1dc1bb574b/jobs/156152


alamb:
I probably need to tune the ingester somehow

alamb:
picking this back up and getting it ready

influxdb_iox/tests/end_to_end_ng_cases/namespace.rs
```diff
@@ -10,7 +10,7 @@ async fn querier_namespace_client() {
     let table_name = "the_table";
 
     let router2_config = TestConfig::new_router2(&database_url);
-    let ingester_config = TestConfig::new_ingester(&router2_config).with_fast_parquet_generation();
```
alamb, line 13: this parameter doesn't do anything, so I removed it (will comment inline below)

influxdb_iox/tests/end_to_end_ng_cases/querier.rs
```diff
@@ -51,7 +51,7 @@ async fn basic_on_parquet() {
     let table_name = "the_table";
 
     // Set up the cluster  ====================================
-    let mut cluster = MiniCluster::create_quickly_persisting(database_url).await;
```
alamb, line 54: again there was no difference between the configurations for these clusters, so don't pretend there is

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -100,13 +100,6 @@ impl TestConfig {
             .with_new_write_buffer()
             .with_new_object_store()
             .with_default_ingester_options()
-            .with_fast_parquet_generation()
-    }
-
-    /// Create a configuraton for aggressive creation of parquet files
-    pub fn with_fast_parquet_generation(self) -> Self {
-        self.with_env("INFLUXDB_IOX_PAUSE_INGEST_SIZE_BYTES", "2")
-            .with_env("INFLUXDB_IOX_PERSIST_MEMORY_THRESHOLD_BYTES", "1")
```
alamb, line 109: note that the memory threshold is "10" in `with_default_ingester_options` so setting it to 1 doesn't really do anything

test_helpers_end_to_end_ng/src/config.rs
```diff
@@ -140,7 +133,7 @@ impl TestConfig {
 
     /// Adds default ingester options
     fn with_default_ingester_options(self) -> Self {
-        self.with_env("INFLUXDB_IOX_PAUSE_INGEST_SIZE_BYTES", "20")
+        self.with_env("INFLUXDB_IOX_PAUSE_INGEST_SIZE_BYTES", "2000000")
```
alamb, line 136: I picked the magic `20` initially because I didn't know any better. When I started sharing the clusters, I found that this would effectively serialize writes / persist because every write was larger than 20

So I cranked it up so the ingester can process multiple writes in pararallel

test_helpers_end_to_end_ng/src/mini_cluster.rs
```diff
@@ -41,34 +51,78 @@ impl MiniCluster {
...
 
-        // Set up the cluster  ====================================
-        Self::new()
-            .with_router2(router2_config)
-            .await
-            .with_ingester(ingester_config)
-            .await
-            .with_querier(querier_config)
-            .await
+        // try to reuse existing server processes
```
alamb, line 93: here is the code that does the actual reuse

test_helpers_end_to_end_ng/src/mini_cluster.rs
```diff
@@ -200,3 +254,108 @@ impl MiniCluster {
         self.other_servers.as_ref()
     }
 }
+
+/// holds shared server processes to share across`
+#[derive(Clone)]
+struct SharedServers {
```
alamb, line 260: The basic idea is to keep a bunch of `Weak` pointers around and try to upgrade them when a new cluster is needed.

This means that if there is already a concurrent test running, the `TestServer`s can be found and reused but if not, then the `TestServer`s are shutdown normally

test_helpers_end_to_end_ng/src/mini_cluster.rs
```diff
@@ -41,34 +51,78 @@ impl MiniCluster {
...
+
+        // Update the shared servers to point at the newly created server proesses
+        *shared_servers = Some(SharedServers::new(&new_cluster));
+        new_cluster
     }
 
-    /// return a "standard" MiniCluster that has a router, ingester,
-    /// querier and quickly persists files to parquet
-    pub async fn create_quickly_persisting(database_url: String) -> Self {
+    /// Create a non shared "standard" shared MiniCluster that has a router, ingester,
```
carols10cents, line 121: I think there's an extra "shared" in here?

```suggestion
    /// Create a non shared "standard" MiniCluster that has a router, ingester,
```

test_helpers_end_to_end_ng/src/mini_cluster.rs
```diff
@@ -200,3 +254,108 @@ impl MiniCluster {
         self.other_servers.as_ref()
     }
 }
+
+/// holds shared server processes to share across`
```
carols10cents, line 258: Looks like an unfinished comment with an extra backtick?

```suggestion
/// holds shared server processes to share across tests
```

carols10cents (approved):
A few comment nits. I was trying to reproduce the other flaky test locally and kept hitting the "port already in use" problem, so I tried out this branch and the "port already in use" problem appears to be fixed!

alamb:
> "port already in use" problem appears to be fixed!

I suspect this PR doesn't actually *fix* the underlying symptom of "port already in use" but I think it will reduce the liklihood of the symptoms (because there are now fewer times the end to end tests try to listen on ports)

test_helpers_end_to_end_ng/src/mini_cluster.rs
alamb: I probably should note that this is the approach used in OG and it works reasonably well:

https://github.com/influxdata/influxdb_iox/blob/944f628e293e2d10d8d1109cb50e44df53c1b09a/influxdb_iox/tests/common/server_fixture.rs#L112-L113

PR #4389, opened 2022-04-21
feat: group chunk overlaps based on time range only

### LINKED ISSUE #4158: Support split_overlap_chunks using only min and max of time column

NGA-TRAN:
Closes #4158

The main change is to return time range for each `QueryChunk` and use it to group overlapped chunks.

- In this implementation, only Chunks in the Ingester have NO time range which means we will consider all of them overlapped while compacting and this is the preferred behavior for  now.
- Parquet Chunks already have time range attached.
- Chunks that are sent from the Ingester to Querier are assumed to be small and their min max are computed at the Querier to ensure we do not do unnecessary compaction due to missing knowledge of time range


ingester/src/query.rs
```diff
@@ -97,6 +98,10 @@ impl QueryableBatch {
     pub fn is_empty(&self) -> bool {
         self.data.is_empty()
     }
+
+    fn timestamp_min_max(&self) -> Option<TimestampMinMax> {
```
alamb, line 102: I don't think this function is needed as QueryableBatch::timestamp_min_max is included in the `impl QueryChunkMeta` below 

Perhaps wherever needs to call `timestamp_min_max` could `use QueryableBatch`

querier/src/ingester/mod.rs
```diff
@@ -439,6 +441,16 @@ impl QueryChunkMeta for IngesterPartition {
     fn delete_predicates(&self) -> &[Arc<data_types2::DeletePredicate>] {
         &[]
     }
+
+    fn timestamp_min_max(&self) -> Option<TimestampMinMax> {
+        // TODO: may want to ask the Ingester to send this value instead of computing it here.
```
alamb, line 446: I agree this is good for now (compute in the querier)

query/src/lib.rs
```diff
@@ -49,6 +50,9 @@ pub trait QueryChunkMeta {
     /// return a reference to the sort key if any
     fn sort_key(&self) -> Option<&SortKey>;
 
+    /// Return time range of the data
+    fn timestamp_min_max(&self) -> Option<TimestampMinMax>;
```
alamb, line 54: This makes me somewhat sad as it is a subset of the information in `summary()` (which is full stats) so now there are redundant and inconsistent ways to get at the min / max timestamp ranges.

I wonder if we can combine your change with @crepererum 's in  https://github.com/influxdata/influxdb_iox/pull/4391 to use `TableSummary` consistently ?

If left like this, I predict we are going to chase down a bunch of bugs / confusion due to the inconsistency with where the min/max timestamps are calculated and provided.

query/src/provider.rs
```diff
@@ -536,12 +536,45 @@ impl Deduplicater {
     ///  2. vector of non-overlapped chunks, each have duplicates in itself
     ///  3. vectors of non-overlapped chunks without duplicates
     fn split_overlapped_chunks(&mut self, chunks: Vec<Arc<dyn QueryChunk>>) -> Result<()> {
+        // Only OG has this statitics
```
alamb, line 539: I wonder if this is true after https://github.com/influxdata/influxdb_iox/pull/4391

query/src/provider.rs
```diff
@@ -536,12 +536,45 @@ impl Deduplicater {
...
+
+        self.split_overlapped_chunks_ng(chunks)
+    }
+
+    /// discover overlaps and split them into three groups:
+    ///  1. vector of vector of overlapped chunks
+    ///  2. vector of non-overlapped chunks, each have duplicates in itself
+    ///  3. vectors of non-overlapped chunks without duplicates
+    fn split_overlapped_chunks_ng(&mut self, chunks: Vec<Arc<dyn QueryChunk>>) -> Result<()> {
+        // Find all groups based on thier time ranges
```
alamb, line 552: I am still confused -- should we be grouping on partition_id or on timestamp range? https://github.com/influxdata/influxdb_iox/pull/4360 assumes that the data is grouped according to partition_id but this code is grouping based on timestamp range.

I realize this happens to be the same at the moment, but I would like to see the assumption made explicit in the code if possible (like perhaps asserting that each member of each overlapping group has the same partition id)

query/src/provider/overlap.rs
```diff
@@ -107,6 +107,70 @@ pub fn group_potential_duplicates(
...
+pub fn group_potential_duplicates(
+    chunks: Vec<Arc<dyn QueryChunk>>,
+) -> Result<Vec<Vec<Arc<dyn QueryChunk>>>> {
+    // If at least one of the chunks has no time range,
+    // all chunks are considered to overlap with each other.
+    if chunks.iter().any(|c| c.timestamp_min_max().is_none()) {
+        return Ok(vec![chunks]);
+    }
+
+    // Use ethis algorithm to group them
```
alamb, line 121:
```suggestion
    // Use this algorithm to group them
```

alamb (commented):
This code looks very similar to code in the compactor (maybe written by @carols10cents )
 https://github.com/influxdata/influxdb_iox/blob/c7a1c496cf54c4671e5b63ae26de4b42bfaf1545/compactor/src/compact.rs#L868-L925 

Could you please try to unify the  implementations -- I am worried that they will drift / become inconsistent over time.

(perhaps the compactor can call the implementation in `query`)?

query/src/provider.rs
NGA-TRAN: Now after several days thinking and doing this with #4360, I will change this function to do 2 steps:
1. Spit provided chunks into groups of partition-chunks
2. Then apply this algorithm on each group of partition-chunks

query/src/lib.rs
NGA-TRAN: I feel like we need better communication because we are working on the same thing at the same time

query/src/provider.rs
NGA-TRAN: This function will go away after we remove OG. It does not matter we have full stats of not, we only need to use time min/max here

NGA-TRAN:
@alamb 
The function you asked me to refactor in Compactor https://github.com/influxdata/influxdb_iox/pull/4389#pullrequestreview-949915272, I should have left a message in this PR saying I have planned to refactor it after this PR is merged because the overlaps in this PR iare more general and I suspected I would need to make it more general after review comment which actually happens here https://github.com/influxdata/influxdb_iox/pull/4389#discussion_r856129529

query/src/provider.rs
NGA-TRAN: This will avoid all confusions and the needs to check, verify, and panic

query/src/lib.rs
NGA-TRAN: @alamb After reviewing #4391 and looking back at the code implemented here, I **partially** agree on your statement. I think we can avoid inconsistent and bugs:
1. We will need this `timestamp_min_max` function anyway. What I have implemented here is just return available min and max. I neither compute anything (except the one sent from the Ingester to Querier) nor store anything extra.
2. No we have available min max (which is a design of NG), the `summary` (from OG) is somehow redundant in NG. If we only need `row_count` for each chunk, then instead of using `summary`, using `row_count`. If we want to have full stats for every single column, `summary` is good to keep for each chunk but we may need to change design earlier to capture min max in there, too.
3. If we want to change the design to keep `summary` and **move** min max  in there, we will **naturally** have to modify the code here to return those min max instead.

In short, I do not think there will be inconsistency if we want move current min max inside a `summary`. The key here is to design the right `summary`, not just use the current OG summary implementation

db/src/chunk.rs
```diff
@@ -590,6 +592,10 @@ impl QueryChunk for DbChunk {
     fn order(&self) -> ChunkOrder {
         self.order
     }
+
+    fn ng_chunk(&self) -> bool {
```
NGA-TRAN, line 596: After a bit of thinking to distinguish OG with NG, I decided the clearest and simplest is to use this trait function. Easy to understand and remove

query/src/frontend/reorg.rs
```diff
@@ -357,15 +357,15 @@ mod test {
...
             "| 100       |            | AL   | 1970-01-01T00:00:00.000000050Z |",
+            "| 70        |            | CT   | 1970-01-01T00:00:00.000000100Z |",
+            "| 1000      |            | MT   | 1970-01-01T00:00:00.000001Z    |",
             "| 5         |            | MT   | 1970-01-01T00:00:00.000005Z    |",
+            "| 10        |            | MT   | 1970-01-01T00:00:00.000007Z    |",
             "+-----------+------------+------+--------------------------------+",
         ];
 
-        assert_batches_eq!(&expected, &batches);
+        assert_batches_sorted_eq!(&expected, &batches);
```
NGA-TRAN, line 368: The output is the same but sorted on tag1 and time per new NG setup and the chunk stats. I did not have to update the output because the new `assert_batches_sorted_eq` will sort them before comparing but this new sort order make it easier to read so I update it.

query/src/provider.rs
```diff
@@ -1776,22 +1939,22 @@ mod test {
...
-            "| 70        | CT   | 1970-01-01T00:00:00.000000100Z |",
             "| 100       | AL   | 1970-01-01T00:00:00.000000050Z |",
+            "| 70        | CT   | 1970-01-01T00:00:00.000000100Z |",
+            "| 1000      | MT   | 1970-01-01T00:00:00.000001Z    |",
             "| 5         | MT   | 1970-01-01T00:00:00.000005Z    |",
+            "| 10        | MT   | 1970-01-01T00:00:00.000007Z    |",
             "+-----------+------+--------------------------------+",
         ];
-        assert_batches_eq!(&expected, &raw_data(&chunks).await);
+        assert_batches_sorted_eq!(&expected, &raw_data(&chunks).await);
```
NGA-TRAN, line 1949: Same as above. Output the same but sorted on different columns

query/src/provider.rs
```diff
@@ -532,17 +533,72 @@ impl Deduplicater {
...
+        } else {
+            self.split_overlapped_chunks_og(chunks)
+        }
+    }
+
+    /// discover overlaps and split them into three groups:
+    ///  1. vector of vector of overlapped chunks
+    ///  2. vector of non-overlapped chunks, each have duplicates in itself
+    ///  3. vectors of non-overlapped chunks without duplicates
+    fn split_overlapped_chunks_ng(&mut self, chunks: Vec<Arc<dyn QueryChunk>>) -> Result<()> {
```
NGA-TRAN, line 553: @alamb Here is the change to group chunks by partition first. Since the partition_id can be accessed easily from the chunk itself, I just simply create a trait function `partition_id()` to access it. No need to teach TableProvider or Dedupplicater for that info.

This will simplify the work in #4360 which I will continue after this is merged

query/src/frontend.rs
```diff
@@ -198,6 +199,7 @@ mod test {
     fn get_test_chunks() -> (Arc<Schema>, Vec<Arc<dyn QueryChunk>>) {
         let chunk1 = Arc::new(
             TestChunk::new("t")
+                .with_partition_id(1)
```
NGA-TRAN, line 202: Needs this to tell NG these 2 chunks are attached to an assigned (same) partition to avoid making them overlap. All current **implicit** tests on TestChunk are now running on NG and no need to rewrite

NGA-TRAN (commented):
@alamb I have addressed all your comments but the compactor refactor that I have open #4403 to do it right after this is merged

query/src/provider.rs
NGA-TRAN: Use `ng_chunk()` to identify this

query/src/provider.rs
NGA-TRAN: Done

query/src/lib.rs
alamb: I like the idea of using a summary that is designed for NG rather than OG. I still don't have a full understanding of what stats will be needed for reasonable query plans (I know @pauldix  has a strong aversion to them ;0 but I don't feel like we have enough data yet to know)

alamb (approved):

db/src/chunk.rs
alamb: I hope it is temporary;

ingester/src/query.rs
```diff
@@ -119,6 +115,12 @@ impl QueryChunkMeta for QueryableBatch {
         merge_record_batch_schemas(&batches)
     }
 
+    fn partition_id(&self) -> Option<PartitionId> {
+        // Ingetser data is not officially attached to and parittion id yet.
```
alamb, line 119: "and parittion" --> `any partition`

alamb:
Thanks @NGA-TRAN  -- I think this looks much better

query/src/lib.rs
NGA-TRAN: After all OG related code are removed, we many want to create `summary` for the QueryChunk when it us created. What kind of statistics we want to put in there can be decided later. I have added a todo checkbox in https://github.com/influxdata/influxdb_iox/issues/4392 for this

PR #4391, opened 2022-04-22
feat: Provide basic summaries for `IngesterPartition`

crepererum:
This is required to pass the query tests with ingester data. (the actual query tests will follow in a separate PR)


alamb (approved):
Looks good -- I changed the name of this PR to be "provide basic summaries" as all that gets provided is `total_count` (which is fine). 👍

alamb:
Marking this one to auto merge to keep the PR queue down

NGA-TRAN (commented):
This PR has semantic-conflict with #4389, especially it will make the function `chunks_have_stats` returns true. I will see what I need to modify in #4389 to avoid this conflict.

PR #4398, opened 2022-04-22
fix: handle LP lines with duplicated fields/tags

### LINKED ISSUE #4326: Assert firing in MutableBatch::column_mut()
[An assert] validating the number of rows in a mutable batch is being hit in prod:

```
level=error msg="Thread panic" panic_info="panicked at 'assertion failed: `(left == right)`\n  left: `1`,\n right: `0`: expected 0 rows in column \"InputPower\" got 1 when performing write of 1 rows', mutable_batch/src/writer.rs:638:9" target="panic_logging" location="panic_logging/src/lib.rs:94" time=1649950675785265461
thread 'tokio-runtime-worker' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: expected 0 rows in column "InputPower" got 1 when performing write of 1 rows', mutable_batch/src/writer.rs:638:9
stack backtrace:
```

(there was no backtrace)

[An assert]:https://github.com/influxdata/influxdb_iox/blob/fe8d9948d5cc235e51811f2f06773bc6f970b28c/mutable_batch/src/writer.rs#L638

domodwyer:
This PR changes how IOx handles line-protocol lines that reference the same column name more than once, as described in #4326.

Previously a write such as `table A=1,A=1` would cause IOx to panic in the `MutableBuffer`. After this PR is merged, IOx will accept writes with duplicate fields according to the following semantics:

* duplicate fields, same value - do nothing/coalesce
* duplicate fields, different value - last write wins
* duplicate fields, different types - reject the request
* duplicate tags, same value - reject the request
* duplicate tags, different value - reject the request
* duplicate tags, different type - reject the request
* same name for tag and field - reject the request
* same name for tag and field, different type  - reject the request

This retains parity with TSM, but differs when a line contains internal type conflicts for a given field.

The LP parsing behaviour is interesting in that it has many stakeholders/interested parties, so sticking with TSM behaviour gives the least surprises / headaches, even if it's a bit more work in IOx. 

Thanks @philjb / @8none1 / @jacobmarble for talking through these tedious edge cases 👍 

Fixes #4326.

---

* fix: handle LP lines with duplicated fields/tags (cf200e509)

        This commit allows the LP consumer to correctly handle line protocol
        writes that duplicate one or more fields or tags within a single line:

                table v=2,bananas=42,v=3,platanos=24
                        ▲              ▲
                        └───────┬──────┘
                                │
                        duplicate field "v"

        This change implements the following logic when processing fields:

            IF field is duplicated
                IF all duplicate field values are of the same data type
                    use last occurrence ("last write wins")
                ELSE
                    return an error

        Any duplication of tags is rejected, and use of a field name as both a
        field and a tag is remains forbidden (unchanged in this PR, a previously
        agreed breaking change from TSM).

        See https://github.com/influxdata/influxdb_iox/issues/4326 for context.

### DIFF
--- a/influxdb_line_protocol/src/lib.rs
+++ b/influxdb_line_protocol/src/lib.rs
@@ -345,13 +345,7 @@ pub enum FieldValue<'a> {
 impl<'a> FieldValue<'a> {
     /// Returns true if `self` and `other` are of the same data type.
     pub fn is_same_type(&self, other: &Self) -> bool {
-        match self {
-            FieldValue::I64(_) => matches!(other, FieldValue::I64(_)),
-            FieldValue::U64(_) => matches!(other, FieldValue::U64(_)),
-            FieldValue::F64(_) => matches!(other, FieldValue::F64(_)),
-            FieldValue::String(_) => matches!(other, FieldValue::String(_)),
-            FieldValue::Boolean(_) => matches!(other, FieldValue::Boolean(_)),
-        }
+        std::mem::discriminant(self) == std::mem::discriminant(other)
     }
 }
 
--- a/influxdb_line_protocol/src/lib.rs
+++ b/influxdb_line_protocol/src/lib.rs
@@ -2231,11 +2231,22 @@ her"#,
         assert!(FieldValue::I64(0).is_same_type(&FieldValue::I64(42)));
         assert!(FieldValue::U64(0).is_same_type(&FieldValue::U64(42)));
         assert!(FieldValue::F64(0.0).is_same_type(&FieldValue::F64(4.2)));
+        // String & String
         assert!(
             FieldValue::String(EscapedStr::CopiedValue("bananas".to_string())).is_same_type(
                 &FieldValue::String(EscapedStr::CopiedValue("platanos".to_string()))
             )
         );
+        // str & str
+        assert!(FieldValue::String(EscapedStr::SingleSlice("bananas"))
+            .is_same_type(&FieldValue::String(EscapedStr::SingleSlice("platanos"))));
+        // str & String
+        assert!(
+            FieldValue::String(EscapedStr::SingleSlice("bananas")).is_same_type(
+                &FieldValue::String(EscapedStr::CopiedValue("platanos".to_string()))
+            )
+        );
+
         assert!(FieldValue::Boolean(true).is_same_type(&FieldValue::Boolean(false)));
 
         // Some false cases
--- a/influxdb_line_protocol/src/lib.rs
+++ b/influxdb_line_protocol/src/lib.rs
@@ -342,6 +342,19 @@ pub enum FieldValue<'a> {
     Boolean(bool),
 }
 
+impl<'a> FieldValue<'a> {
+    /// Returns true if `self` and `other` are of the same data type.
+    pub fn is_same_type(&self, other: &Self) -> bool {
+        match self {
+            FieldValue::I64(_) => matches!(other, FieldValue::I64(_)),
+            FieldValue::U64(_) => matches!(other, FieldValue::U64(_)),
+            FieldValue::F64(_) => matches!(other, FieldValue::F64(_)),
+            FieldValue::String(_) => matches!(other, FieldValue::String(_)),
+            FieldValue::Boolean(_) => matches!(other, FieldValue::Boolean(_)),
+        }
+    }
+}
+
 /// Converts FieldValue back to LineProtocol
 /// See <https://docs.influxdata.com/influxdb/v2.0/reference/syntax/line-protocol/>
 /// for more detail.
@@ -369,7 +382,7 @@ impl<'a> Display for FieldValue<'a> {
 /// For example the 8 character string `Foo\\Bar` (note the double
 /// `\\`) is parsed into the logical 7 character string `Foo\Bar`
 /// (note the single `\`)
-#[derive(Debug, Clone, Eq)]
+#[derive(Debug, Clone, Eq, Hash)]
 pub enum EscapedStr<'a> {
     SingleSlice(&'a str),
     CopiedValue(String),
@@ -2216,5 +2229,29 @@ her"#,
         let vals = parse(input).unwrap();
 
         assert_eq!(vals[0].tag_value("asdf"), None);
+    }
+
+    #[test]
+    fn test_field_value_same_type() {
+        // True cases
+        assert!(FieldValue::I64(0).is_same_type(&FieldValue::I64(42)));
+        assert!(FieldValue::U64(0).is_same_type(&FieldValue::U64(42)));
+        assert!(FieldValue::F64(0.0).is_same_type(&FieldValue::F64(4.2)));
+        assert!(
+            FieldValue::String(EscapedStr::CopiedValue("bananas".to_string())).is_same_type(
+                &FieldValue::String(EscapedStr::CopiedValue("platanos".to_string()))
+            )
+        );
+        assert!(FieldValue::Boolean(true).is_same_type(&FieldValue::Boolean(false)));
+
+        // Some false cases
+        assert!(!FieldValue::I64(0).is_same_type(&FieldValue::U64(42)));
+        assert!(!FieldValue::U64(0).is_same_type(&FieldValue::I64(42)));
+        assert!(!FieldValue::F64(0.0).is_same_type(&FieldValue::U64(42)));
+        assert!(
+            !FieldValue::String(EscapedStr::CopiedValue("bananas".to_string()))
+                .is_same_type(&FieldValue::U64(42))
+        );
+        assert!(!FieldValue::Boolean(true).is_same_type(&FieldValue::U64(42)));
     }
 }
--- a/mutable_batch_lp/src/lib.rs
+++ b/mutable_batch_lp/src/lib.rs
@@ -531,6 +531,7 @@ m b=t 1639612800000000000
             });
         }
 
+        // NOTE: All tags are strings, so this should never be a type conflict.
         #[test]
         fn test_duplicate_tags_different_type() {
             let lp = "m1,tag=1,tag=2.0 val=1i 0";
@@ -546,6 +547,9 @@ m b=t 1639612800000000000
             });
         }
 
+        // NOTE: disallowed in IOx but accepted in TSM
+        //
+        // https://github.com/influxdata/influxdb_iox/issues/3150
         #[test]
         fn test_duplicate_is_tag_and_field() {
             let lp = "m1,v=1i v=1i 0";
--- a/mutable_batch_lp/src/lib.rs
+++ b/mutable_batch_lp/src/lib.rs
@@ -491,7 +491,7 @@ m b=t 1639612800000000000
             let lp = "m1 val=1i,val=2.0 0";
 
             let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
-            assert_matches!( err,
+            assert_matches!(err,
                 Error::Write {
                     source: LineWriteError::ConflictedFieldTypes { name },
                     line: 1
@@ -505,8 +505,8 @@ m b=t 1639612800000000000
         fn test_duplicate_tags_same_value() {
             let lp = "m1,tag=1,tag=1 val=1i 0";
 
-            let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
-            assert_matches!( err,
+            let err = lines_to_batches(lp, 5).expect_err("duplicate tag write should fail");
+            assert_matches!(err,
                 Error::Write {
                     source: LineWriteError::DuplicateTag { name },
                     line: 1
@@ -520,8 +520,8 @@ m b=t 1639612800000000000
         fn test_duplicate_tags_different_value() {
             let lp = "m1,tag=1,tag=2 val=1i 0";
 
-            let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
-            assert_matches!( err,
+            let err = lines_to_batches(lp, 5).expect_err("duplicate tag write should fail");
+            assert_matches!(err,
                 Error::Write {
                     source: LineWriteError::DuplicateTag { name },
                     line: 1
@@ -536,7 +536,7 @@ m b=t 1639612800000000000
             let lp = "m1,tag=1,tag=2.0 val=1i 0";
 
             let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
-            assert_matches!( err,
+            assert_matches!(err,
                 Error::Write {
                     source: LineWriteError::DuplicateTag { name },
                     line: 1
--- a/mutable_batch_lp/src/lib.rs
+++ b/mutable_batch_lp/src/lib.rs
@@ -11,7 +11,7 @@
     clippy::clone_on_ref_ptr
 )]
 
-use hashbrown::HashMap;
+use hashbrown::{hash_map::Entry, HashMap, HashSet};
 use influxdb_line_protocol::{parse_lines, FieldValue, ParsedLine};
 use mutable_batch::writer::Writer;
 use mutable_batch::MutableBatch;
@@ -28,10 +28,7 @@ pub enum Error {
     },
 
     #[snafu(display("error writing line {}: {}", line, source))]
-    Write {
-        source: mutable_batch::writer::Error,
-        line: usize,
-    },
+    Write { source: LineWriteError, line: usize },
 
     #[snafu(display("empty write payload"))]
     EmptyPayload,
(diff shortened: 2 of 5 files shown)


mutable_batch_lp/src/lib.rs
```diff
@@ -139,38 +154,155 @@ pub fn lines_to_batches_stats(
...
+    //                        table v=2i,v=3u
+    //
+    // In this instance we break from established TSM behaviour and return an
+    // error. IOx features schema enforcement, and as such is expected to reject
+    // conflicting types for writes. See the github issue[1] for the desired
+    // semantics.
+    //
+    // Tests below codify each of the scenarios described in the ticket.
+    //
+    // [1]: https://github.com/influxdata/influxdb_iox/issues/4326
```
domodwyer, line 257: I have tried to capture the desired outcome and explain how the code goes about it so it's not lost to the ages

router2/src/server/http.rs
```diff
@@ -931,4 +959,119 @@ mod tests {
...
+            want_dml_calls = [MockDmlHandlerCall::Write{namespace, write_input}] => {
+                assert_eq!(namespace, "bananas_test");
+                let table = write_input.get("whydo").expect("table not in write");
+                let col = table.column("InputPower").expect("column missing");
+                assert_matches!(col.data(), ColumnData::I64(data, _) => {
+                    // Last value wins
+                    assert_eq!(data.as_slice(), [42]);
+                });
+            }
+        );
```
domodwyer, line 999: These tests (and more below it) exercise each of the edge cases in the description from the HTTP write API perspective.

Here writes containing fields of the same value are coalesced into one write (first test) and writes containing fields with differing values (but same types) are reduced according to the "last write wins" rule (second test).

router2/src/server/http.rs
```diff
@@ -829,6 +831,32 @@ mod tests {
...
+
+    test_write_handler!(
+        column_named_time,
+        query_string = "?org=bananas&bucket=test",
+        body = "test field=1u,time=42u 100".as_bytes(),
+        dml_handler = [],
+        want_result = Err(_),
+        want_dml_calls = []
+    );
+
```
domodwyer, line 859: These tests are unrelated to the LP parsing change, but I wanted to assert they did not cause any panics as they're similar inputs.

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+                &[
+                    "+----------------------+-----+",
+                    "| time                 | val |",
+                    "+----------------------+-----+",
+                    "| 1970-01-01T00:00:00Z | 2   |",
+                    "+----------------------+-----+",
+                ],
+                &[batches["m1"].to_arrow(Selection::All).unwrap()]
+            );
+        }
```
domodwyer, line 467: These tests exercise the edge cases by inspecting the resulting batches from the parsed LP writer.

influxdb_line_protocol/src/lib.rs
```diff
@@ -342,6 +342,19 @@ pub enum FieldValue<'a> {
...
+impl<'a> FieldValue<'a> {
+    /// Returns true if `self` and `other` are of the same data type.
+    pub fn is_same_type(&self, other: &Self) -> bool {
+        match self {
+            FieldValue::I64(_) => matches!(other, FieldValue::I64(_)),
+            FieldValue::U64(_) => matches!(other, FieldValue::U64(_)),
+            FieldValue::F64(_) => matches!(other, FieldValue::F64(_)),
+            FieldValue::String(_) => matches!(other, FieldValue::String(_)),
+            FieldValue::Boolean(_) => matches!(other, FieldValue::Boolean(_)),
+        }
```
shepmaster, line 354:
```suggestion
        std::mem::discriminant(self) == std::mem::discriminant(other)
```

[`std::mem::discriminant`](https://doc.rust-lang.org/std/mem/fn.discriminant.html)

mutable_batch_lp/src/lib.rs
```diff
@@ -139,38 +154,155 @@ pub fn lines_to_batches_stats(
...
+        for (tag_key, tag_value) in tags.iter() {
+            // Check if a field with this name has been observed previously.
+            if !seen.insert(tag_key) {
+                // This tag_key appears more than once, with differing values.
+                //
+                // This is always an error.
+                return Err(LineWriteError::DuplicateTag {
+                    name: tag_key.to_string(),
+                });
+            }
```
shepmaster, line 211: Can use `snafu::ensure`

```suggestion
            // This tag_key appears more than once, with differing values.
            //
            // This is always an error.
            ensure!(seen.insert(tag_key), DuplicateTagSnafu {
                name: tag_key
            });
```

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+                &[batches["m1"].to_arrow(Selection::All).unwrap()]
+            );
+        }
+
+        #[test]
+        fn test_duplicate_fields_different_type() {
+            let lp = "m1 val=1i,val=2.0 0";
+
+            let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
+            assert_matches!( err,
```
shepmaster, line 494: extra space snuck in before the `err` (and other places this was copied)

mutable_batch_lp/src/lib.rs
```diff
@@ -81,7 +78,25 @@ impl LinesConverter {
         self.timestamp_base = timestamp_base
     }
 
-    /// Write some line protocol data
+    /// Write some line protocol data.
+    ///
+    /// If a field / tag name appears more than once in a single line, the
+    /// following semantics apply:
+    ///
+    ///   * duplicate fields, same value: do nothing/coalesce
```
NGA-TRAN, line 86: What exactly `do nothing` means? Will, at some step later, they be coalesced and treated as one

mutable_batch_lp/src/lib.rs
NGA-TRAN: Awesome 👍

NGA-TRAN (approved):
Looks great @domodwyer

philjb:
> duplicate tags, different type - reject the request

In LP, tags can only be string types so even if the tag k-v pair is `mynumbertag=3.1415` that should be typed as a string of `3.1415`.

philjb:
> same name for tag and field, different type - reject the request

I don't get this one: example? tag names and field names are only strings. There are a few rules for the names for starting characters but I can write a tag of `3.1415=pi`. Not great schema design but it is accepted currently.

philjb:
> duplicate fields, different types - reject the request

Cloud2 does rejects this scenario `weather,location=london temp=24i,temp=24.6` but not at the `/write` endpoint synchronously. Because cloud2 writes are asynchronous (isn't this true for the iox model too?), field type information isn't available at the api request time. So cloud2 would accept the above and then later issue an error to our dead letter queue (_monitoring bucket for the customer) for the point that didn't fit the existing type for the field. Interesting, it might the int field that errors or the float field depending. If the `temp` field is already established as float then `temp=24i` will be rejected, if the `temp` field is already established as int then `temp=24.6` will be rejected, and if `temp` is already established as string, then both will be rejected. 

Now if the customer has defined a schema for the bucket, the field types can be rejected synchronously (but I would accept the point that matches the schema and reject only the rejecting points - again accepting as much as possible instead of throwing out the whole request).

domodwyer:
> duplicate tags, same value .... I would accept this request.

I have no strong feelings either way, though my preference would be to keep this code as simple as possible, adding behaviours only if needed (especially as it is in the hot path). If we want to support this, it is a small change 👍 

> In LP, tags can only be string types

Yeah I got carried array enumerating the combinations of `{field, tag} x {same value, diff value, diff type}`. When parsing LP, tags are always string typed in IOx too. I don't think there's any harm including the tests for this though.

> same name for tag and field, different type - reject ... I don't get this one: example?

In InfluxDB tags/fields are stored independently - In IOx a tag effectively becomes a column in an SQL table alongside the fields - there's no distinction between tags/fields once it hits datafusion for query execution which uses [this data model][model]. This has been previously documented as a breaking change between TSM & IOx (by @e-dard).

[model]:https://docs.rs/arrow/12.0.0/arrow/datatypes/enum.DataType.html

> Because cloud2 writes are asynchronous (isn't this true for the iox model too?),

Writes are asynchronous, but schema validation is performed synchronously at the edge, irrespective of whether or not the schema is implicitly derived or explicitly set. This lets the client get instant "yay or nay" feedback for the common use case to help reduce confusion - it's a design goal of IOx to support this 🎉 

As mentioned in the issue for this PR (#4326) I think rejecting `duplicate fields, different types` straight up when the user submits it makes sense as it avoids adding a bunch of special case logic in the LP consumer for a request that would never pass schema validation later in the handler stack anyway.

influxdb_line_protocol/src/lib.rs
domodwyer: Thanks - I didn't know about this 👍

mutable_batch_lp/src/lib.rs
domodwyer: Just that we take no special action - `field=1,field=1` becomes a write of `field=1` - could have been clearer, sorry!

alamb:
> This has been previously documented as a breaking change between TSM & IOx (by @e-dard).

Documentation and previous discussion on this issue are on https://github.com/influxdata/influxdb_iox/issues/3150

domodwyer:
> Documentation and previous discussion on this issue are on #3150

Thanks! @alamb! I couldn't find anything to reference 👍

influxdb_line_protocol/src/lib.rs
```diff
@@ -342,6 +342,13 @@ pub enum FieldValue<'a> {
     Boolean(bool),
 }
 
+impl<'a> FieldValue<'a> {
+    /// Returns true if `self` and `other` are of the same data type.
+    pub fn is_same_type(&self, other: &Self) -> bool {
+        std::mem::discriminant(self) == std::mem::discriminant(other)
```
alamb, line 348: TIL: https://doc.rust-lang.org/std/mem/fn.discriminant.html 👍

influxdb_line_protocol/src/lib.rs
```diff
@@ -2217,4 +2224,28 @@ her"#,
...
+        // True cases
+        assert!(FieldValue::I64(0).is_same_type(&FieldValue::I64(42)));
+        assert!(FieldValue::U64(0).is_same_type(&FieldValue::U64(42)));
+        assert!(FieldValue::F64(0.0).is_same_type(&FieldValue::F64(4.2)));
+        assert!(
+            FieldValue::String(EscapedStr::CopiedValue("bananas".to_string())).is_same_type(
+                &FieldValue::String(EscapedStr::CopiedValue("platanos".to_string()))
+            )
+        );
+        assert!(FieldValue::Boolean(true).is_same_type(&FieldValue::Boolean(false)));
```
alamb, line 2239: maybe worth testing a `EscapedStr::CopiedValue` and `EscapedStr::SingleSlice`? for equality too?

mutable_batch_lp/src/lib.rs
alamb: maybe `ensure!(!seen` rather than `ensure!(seen` 🤔

mutable_batch_lp/src/lib.rs
```diff
@@ -139,38 +154,155 @@ pub fn lines_to_batches_stats(
...
+    // error. IOx features schema enforcement, and as such is expected to reject
+    // conflicting types for writes. See the github issue[1] for the desired
+    // semantics.
+    //
+    // Tests below codify each of the scenarios described in the ticket.
+    //
+    // [1]: https://github.com/influxdata/influxdb_iox/issues/4326
+
+    let mut seen = HashMap::<_, &FieldValue<'_>>::with_capacity(line.field_set.len());
+    for (field_key, field_value) in line.field_set.iter().rev() {
```
alamb, line 260: 👍  for rev

mutable_batch_lp/src/lib.rs
```diff
@@ -139,38 +154,155 @@ pub fn lines_to_batches_stats(
...
+    // In this instance we break from established TSM behaviour and return an
+    // error. IOx features schema enforcement, and as such is expected to reject
+    // conflicting types for writes. See the github issue[1] for the desired
+    // semantics.
+    //
+    // Tests below codify each of the scenarios described in the ticket.
+    //
+    // [1]: https://github.com/influxdata/influxdb_iox/issues/4326
+
+    let mut seen = HashMap::<_, &FieldValue<'_>>::with_capacity(line.field_set.len());
```
alamb, line 259: this code is now on the hot path for writes, correct? This looks pretty efficient but probably something to keep an eye on when this code is deployed

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+            => {
+                assert_eq!(name, "val");
+            });
+        }
+
+        #[test]
+        fn test_duplicate_tags_same_value() {
+            let lp = "m1,tag=1,tag=1 val=1i 0";
+
+            let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
```
alamb, line 508:
```suggestion
            let err = lines_to_batches(lp, 5).expect_err("duplicate tag write should fail");
```

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+            => {
+                assert_eq!(name, "tag");
+            });
+        }
+
+        #[test]
+        fn test_duplicate_tags_different_value() {
+            let lp = "m1,tag=1,tag=2 val=1i 0";
+
+            let err = lines_to_batches(lp, 5).expect_err("type conflicted write should fail");
```
alamb, line 523:
```suggestion
            let err = lines_to_batches(lp, 5).expect_err("duplicate tag write should fail");
```

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+                    line: 1
+                }
+            => {
+                assert_eq!(name, "tag");
+            });
+        }
+
+        #[test]
+        fn test_duplicate_tags_different_type() {
+            let lp = "m1,tag=1,tag=2.0 val=1i 0";
```
alamb, line 536: I think tags are always "strings" -- so this is basically the same code as above where `tag` is provided the two string values as `"1"` and `"2.0"`. Not that it is a bad test but the test name is misleading I think

mutable_batch_lp/src/lib.rs
```diff
@@ -309,4 +442,136 @@ m b=t 1639612800000000000
...
+                    source: LineWriteError::DuplicateTag { name },
+                    line: 1
+                }
+            => {
+                assert_eq!(name, "tag");
+            });
+        }
+
+        #[test]
+        fn test_duplicate_is_tag_and_field() {
```
alamb, line 550:
```suggestion
        // See https://github.com/influxdata/influxdb_iox/issues/3150
        fn test_duplicate_is_tag_and_field() {
```

alamb (approved):
Beautiful tests 🏅 

Very nicely done

philjb:
@domodwyer and I discussed offline. All is clear to me.

> duplicate tags, same value .... I would accept this request.

This is your call; I like permissive acceptance for writes but it's a minor case.

mutable_batch_lp/src/lib.rs
domodwyer: To be honest, I find this more confusing that the `if...return err()` but I guess I've not learnt snafu enough for it to be natural.

mutable_batch_lp/src/lib.rs
domodwyer: Very much so - it's right in the hot path! I've been careful to make this as cheap as possible with this in mind 👍 

I'll watch over this (and the panic counter!) when we deploy it, and for that reason I'll probably do it on Monday 👍

mutable_batch_lp/src/lib.rs
domodwyer: Yeah that's a good point - I will leave a comment for this specific test but I don't think there's any harm in covering with a test case.

domodwyer:
Thanks for the reviews - I intend to push final touch-up commits and merge on Monday when I can keep an eye on changes it causes as it deploys.

Enjoy your weekends 🎉

mutable_batch_lp/src/lib.rs
alamb: I agree the current formulation (without `ensure!`) is quite clear and I don't think it needs to be changed

mutable_batch_lp/src/lib.rs
alamb: good plan

crepererum (approved):
What a nasty bug :grin:

mutable_batch_lp/src/lib.rs
domodwyer: I just benchmarked this as incurring a 4% overhead - I also noticed a fair bit of low hanging fruit to optimise this call path and I'll add it to my TODO. Perhaps one to keep me busy on the flight to JOIN 👍

alamb: > I just benchmarked this as incurring a 4% overhead 

A wise mentor once told me "If you don't constrain me to make correct results, I can make the code as fast as you want"

PR #4383, opened 2022-04-21
chore: port remaining read_filter influxrpc tests to NG

alamb:
Draft until https://github.com/influxdata/influxdb_iox/pull/4382 is merged

# Rationale
Part ot https://github.com/influxdata/influxdb_iox/issues/3958
I am porting influxrpc tests to NG end to end harnesses in preparation of removing NG

# Changes:
1. Port more read_filter tests

PR #4404, opened 2022-04-22
feat: Rewrite the NG end-to-end metrics test 

shepmaster:
This is still broken because the the row timestamp metrics don't exist
in NG.


influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
```diff
@@ -0,0 +1,72 @@
+use test_helpers_end_to_end_ng::{maybe_skip_integration, MiniCluster, Step, StepTest, TestConfig};
+
+#[tokio::test]
+pub async fn test_row_timestamp() {
+    let database_url = maybe_skip_integration!();
+    let test_config =
+        TestConfig::new_all_in_one(database_url).with_row_timestamp_metrics(&["system"]);
+    let mut cluster = MiniCluster::create_all_in_one(test_config).await;
+
+    let lp = crate::end_to_end_ng_cases::querier::influxrpc::data::DataGenerator::new()
```
shepmaster, line 10: @carols10cents mentioned that she also wanted some of the data created by this type. Instead of just making it public, perhaps we should move it up to a common location?

influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
```diff
@@ -0,0 +1,72 @@
...
+    StepTest::new(
+        &mut cluster,
+        vec![
+            Step::WriteLineProtocol(lp),
+            Step::WaitForReadable,
+            Step::VerifiedMetrics(Box::new(|state, metrics| {
+                let lines: Vec<_> = metrics
+                    .trim()
+                    .split('\n')
+                    .filter(|x| x.starts_with("catalog_row_time_seconds_bucket"))
```
shepmaster, line 23: There are no metrics that meet this criteria anymore. It appears [only OG code uses the environment variable](https://github.com/influxdata/influxdb_iox/blob/117569184e48750bdc5906c0ebc1407d99efc709/db/src/catalog/metrics.rs#L10-L16) to enable these metrics.

Should we switch to checking some other metric? Here's [a list](https://gist.github.com/shepmaster/6c413932c4600fc93de6d92621dba783) of all the metrics If no data is inserted and if a little bit of LP data is ingested.

influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
```diff
@@ -0,0 +1,72 @@
...
+                    .iter()
+                    .all(|x| x.contains(r#"table="system""#) && x.contains(&db_name_attribute)));
+            })),
+        ],
+    )
+    .run()
+    .await;
+}
+
+#[cfg(feature = "jemalloc_replacing_malloc")]
```
shepmaster, line 43: Notably, the old tests used `cfg(jemalloc_replacing_malloc)`, which never gets set as far as I saw...

test_helpers_end_to_end_ng/src/steps.rs
```diff
@@ -101,6 +101,13 @@ pub enum Step {
         verify: Box<dyn Fn(Vec<RecordBatch>)>,
     },
 
+    /// Retrieve the metrics and verify the results using the provided
```
alamb, line 104: 👍

influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
alamb: Perhaps somewhere in the `test_helpers_end_to_end_ng` crate would be good

https://github.com/influxdata/influxdb_iox/tree/main/test_helpers_end_to_end_ng/src

influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
alamb: This test appears to have been introduced as part of https://github.com/influxdata/influxdb_iox/pull/2135 from @tustvold 

I don't think testing `INFLUXDB_IOX_ROW_TIMESTAMP_METRICS` in NG is needed as it isn't a relevant setting (as you have discovered)


I suggest
1. Renaming the test to `test_metrics` and 
2. pick another metric for your list (maybe `catalog_op_duration_ms_bucket{op="table_list_by_namespace_id",result="success",le="1280"} 1`? would be a good idea.

alamb (commented):
Thanks @shepmaster  this is looking good

alamb:
Th PR looks like all test pass now ✅

influxdb_iox/tests/end_to_end_ng_cases/querier/influxrpc/metadata.rs
```diff
@@ -68,16 +59,12 @@ async fn tag_keys() {
...
-                let range = generator.timestamp_range();
-                let predicate = exprs::make_tag_predicate("host", "server01");
-                let predicate = Some(predicate);
 
-                let tag_keys_request = tonic::Request::new(TagKeysRequest {
-                    tags_source: read_source,
-                    range,
-                    predicate,
-                });
+                let tag_keys_request = GrpcRequestBuilder::new()
```
alamb, line 63: `GrpcRequestBuilder` 👍

test_helpers_end_to_end_ng/src/grpc.rs
```diff
@@ -0,0 +1,165 @@
+use crate::MiniCluster;
+use generated_types::{
+    node::{Comparison, Type as NodeType, Value},
+    MeasurementFieldsRequest, MeasurementNamesRequest, MeasurementTagKeysRequest,
+    MeasurementTagValuesRequest, Node, Predicate, ReadFilterRequest, ReadSource, TagKeysRequest,
+    TagValuesRequest, TimestampRange,
+};
+use prost::Message;
+
+pub struct GrpcRequestBuilder {
```
alamb, line 10: this is very nice 👍  

*goes to update https://github.com/influxdata/influxdb_iox/pull/4383*

alamb (approved):
wunderbar!

PR #4405, opened 2022-04-22
feat: Write end-to-end tests for the gRPC write API in NG

### LINKED ISSUE #3941: NGQuerier: Add end to end test for gRPC write endpoint
Idea is to test the gRPC write service added in 
 https://github.com/influxdata/influxdb_iox/pull/3936 as part of the end to end NG test

### LINKED ISSUE #4208: Querier: Add write tokens to the grpc write endpoint
The idea is to add write tokens to the response of the gRPC write endpoint so that users can then ask the ingester when their data is readable. 

# Open question
1. Should the write token be in a http header or as part of the [WriteResponse](https://github.com/influxdata/influxdb_iox/blob/32d7c4c/generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto#L120-L122)

In theory the WriteResponse is shared by other systems (e.g. golang) so some additional care might be needed to add a new field to the response

# Definition of Done
- [ ] Design
- [ ] Code
- [ ] End to end test

# Context
write tokens added to http write in https://github.com/influxdata/influxdb_iox/pull/4202

carols10cents:
Fixes #3941, and as a bonus fixes #4208 because the end-to-end test needs the write token info to be able to tell if the data is available for querying yet :)


router2/src/server/grpc.rs
```diff
@@ -18,15 +17,16 @@ use mutable_batch::MutableBatch;
...
-use std::ops::DerefMut;
-use tonic::{Request, Response, Status};
+use std::{ops::DerefMut, sync::Arc};
+use tonic::{metadata::AsciiMetadataValue, Request, Response, Status};
 use trace::ctx::SpanContext;
 use write_summary::WriteSummary;
 
-use crate::dml_handlers::{DmlError, DmlHandler, PartitionError};
+// This header value needs to be all lowercase as required by the HTTP/2.0 spec! tonic does not
+// lowercase this value for you like the http crate does!
```
alamb, line 26: I think HTTP headers are supposed to be case insensitive. 

https://www.w3.org/Protocols/rfc2616/rfc2616.html

It is not clear to me why this needs to be lower case (nor that it is a problem if it is -- I just find the comment to be confusing)

test_helpers_end_to_end_ng/src/client.rs
```diff
@@ -53,6 +71,18 @@ pub fn get_write_token(response: &Response<Body>) -> String {
         .to_string()
 }
 
+/// Extracts the write token from the specified response (to the gRPC write API)
+pub fn get_write_token_from_grpc(response: &tonic::Response<WriteResponse>) -> String {
```
alamb, line 75: ❤️

alamb (approved):
This is great @carols10cents  -- I had some comments but nothing I think would prevent merging.

router2/src/server/grpc.rs
carols10cents: In HTTP/2, [header names must be lowercase](https://httpwg.org/specs/rfc7540.html#HttpHeaders):

> Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed

This code eventually calls [`http::HeaderName::from_static` which panics if this isn't lowercase](https://docs.rs/http/latest/http/header/struct.HeaderName.html#method.from_static) :-/

carols10cents: [I've changed this comment to perhaps be less confusing and only convey the important information](https://github.com/influxdata/influxdb_iox/pull/4405/commits/35690d390bb747fa2cf67fd44832ab22d200bbcc), if you'd like me to change it again or do more investigation into why this is important here but not in the HTTP API, let me know!

alamb: This is great -- thank you @carols10cents

router2/src/server/grpc.rs
```diff
@@ -22,8 +22,8 @@ use tonic::{metadata::AsciiMetadataValue, Request, Response, Status};
 use trace::ctx::SpanContext;
 use write_summary::WriteSummary;
 
-// This header value needs to be all lowercase as required by the HTTP/2.0 spec! tonic does not
-// lowercase this value for you like the http crate does!
+// HERE BE DRAGONS: Uppercase characters in this constant cause a panic. Insert them and
+// investigate the cause if you dare.
```
alamb, line 26: lolol -- what a cliff hanger

PR #4415, opened 2022-04-25
test: chunks in ingester stage

crepererum:
We probably need to port some more very specific scenarios (e.g. "delete all"), but this should cover quite a lot already.


NGA-TRAN (approved):
I understand the needs for this because you want to create parquet files from LP without going thru Ingester but want data to meed all requirements as it went thru the Ingester. Because of that I agree to have this PR merge.

However, I feel the PR is confusing as it adds checks for those rules that only needs for tests. I wonder if it is worth using Ingester to build the test data instead

query_tests/src/scenarios/util.rs
```diff
@@ -1245,6 +1343,59 @@ async fn make_ng_chunk(ns: Arc<TestNamespace>, chunk: ChunkDataNew<'_, '_>) -> S
...
+                                DeleteTimeNew::Begin | DeleteTimeNew::End => {
+                                    unreachable!("Begin/end cases should have been replaced with concrete instances at this point")
+                                }
+                            }
+                        }
+                    }
+                }
+
+                // create ingester partition
+                if batch.num_rows() > 0 {
```
alamb, line 1367: > However, I feel the PR is confusing as it adds checks for those rules that only needs for tests. I wonder if it is worth using Ingester to build the test data instead

I agree with @NGA-TRAN  that the closer we can keep the query_tests testing the same code paths as it is in the real system,  the better the coverage we have and the less likely the `query_tests` are to diverge from what is happening. 

For example, I wonder if it would be possible to 
1. instantiate a [IngesterData](https://github.com/influxdata/influxdb_iox/blob/main/ingester/src/data.rs#L107)
2. Create the relevant `DmlOperation` for the line protocol to be inserted
3. Call `IngesterData::buffer_operation` with that operation
4. Create a `IngesterQueryRequest` , invoke the right think in the Ingester to get back a QueryResponse / IngesterPartition

I know this would be more code, but I feel like the coverage added in this PR may be of limited value as it mocks out so much of the interaction (e.g. sort keys, partition keys, serialization, etc)

alamb (approved):

query_tests/src/scenarios/util.rs
crepererum: > the coverage added in this PR may be of limited value

I disagree here because we're still testing the querier part and the "limited value" coverage in this PR actually found some issues (see #4391 and #4379).

That said, I agree that using more ingester code would be nice, but that's a story for another PR.

query_tests/src/scenarios/util.rs
```diff
@@ -157,23 +195,17 @@ pub struct PredOld<'a> {
...
+    pub predicate: &'a DeletePredicate,
+
     /// At which chunk stage this predicate is applied
-    delete_time: DeleteTimeNew,
+    pub delete_time: DeleteTimeNew,
 }
 
 impl<'a> PredNew<'a> {
-    pub fn begin(predicate: &'a DeletePredicate) -> Self {
+    fn replace_begin_and_end_delete_times(self, stage: ChunkStageNew) -> Self {
```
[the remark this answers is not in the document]
crepererum: done

query_tests/src/scenarios/util.rs
```diff
@@ -99,24 +122,39 @@ impl ChunkStageOld {
...
+    }
+}
+
+impl PartialOrd for ChunkStageNew {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        match (self, other) {
+            (Self::Parquet, Self::Parquet) => Some(Ordering::Equal),
+            (Self::Parquet, Self::Ingester) => Some(Ordering::Less),
+            (Self::Ingester, Self::Parquet) => Some(Ordering::Greater),
+            (Self::Ingester, Self::Ingester) => None,
```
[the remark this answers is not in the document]
crepererum: done

query_tests/src/scenarios/util.rs
alamb: > I disagree here because we're still testing the querier part and the "limited value" coverage in this PR actually found some issues (see https://github.com/influxdata/influxdb_iox/pull/4391 and https://github.com/influxdata/influxdb_iox/pull/4379).

Excellent point

> That said, I agree that using more ingester code would be nice, but that's a story for another PR.

Sounds great. Thank you

Issue #4350, opened 2022-04-19
Implement router2 request throttle / back pressure


domodwyer:
Currently the router will happily accept an unbounded number of parallel requests.

However, this lack of back pressure leads to OOMs when the PG connection pool is saturated and request handling takes a long time - the router happily accepts more and more requests, blocked waiting for a PG connection, until it OOMs.

This ticket is to add a configurable (via flag & env) parallel request limit to the router, and size it appropriately such that it is only hit when the memory usage of the router nears the limit in prod.

domodwyer:
We're proceeding with limiting parallel requests in the istio config to achieve this 👍

PR #4419, opened 2022-04-25
feat(router2): http service protection limit

domodwyer:
This PR adds functionality to limit the number of simultaneous HTTP requests in the router. Once the configured limit is reached, further requests get a `HTTP 503 Service Unavailable` response.

This should mitigate a cascading failure we observe in prod:

* A bunch of routers are happily working away 
* A subset of routers crash / restart / are re-deployed
* The remaining routers therefore handle more requests per second while the other routers are down
* This request rate creates too much memory pressure on the router, causing it to OOM
* This adds more workload onto the remaining routers, which repeats the problem
* All routers eventually OOM
* One router starts up, and immediately gets hit with an unmanageable number of requests/second
* Router crashes
* Another router starts up and....

This causes an excessively long outage while routers try, and fail, to restart and handle the full request workload until (through luck) enough of them start up at the same time to satisfy the current request rate. We observed outages lasting hours in this loop due to requests blocking on the PG connection pool before #4351 was merged.

This PR addresses the issue by rejecting requests once a pre-configured maximum number of simultaneous requests is reached. Allowing us to avoid crashing in the first place, and more easily recover if an outage occurs.

Additionally a metric `http_request_limit_rejected` is incremented once per request rejection in order to monitor and alert when this limit is reached, and use the rate/second to help inform correct capacity sizing.

Closes https://github.com/influxdata/influxdb_iox/issues/4350

---

* feat(router2): simultaneous HTTP request limit (5b09253dd)

      This commit adds a service protection limit, allowing the router to restrict
      the number of simultaneous HTTP requests to a hard-coded limit of 200.

      By dropping additional requests once the router is serving the configured
      maximum, the router is prevented from accepting more and more requests that
      stack up waiting on the catalog / postgres connection pool. Currently
      buffering all these stalled connections consumes so much memory the router is
      OOM killed. After this commit, a portion of requests will be satisfied and the
      rest will be dropped early with a HTTP 503 Service Unavailable response.

* feat(router2): configurable max HTTP requests (0b66b0728)

      Adds a CLI / env configuration option controlling the maximum number of 
      simultaneous HTTP requests in the router.

* feat: emit http_request_limit_rejected metric (d78ea50b3)

      Add the "http_request_limit_rejected" metric that is incremented once each
      time a request is dropped due to the simultaneous request service protection
      limit being exceeded.

      This metric will allow us to effectively alert on router saturation, and the
      increase rate/second will help inform us in how much capacity needs adding.


influxdb_iox/src/commands/run/router2.rs
```diff
@@ -67,6 +67,22 @@ pub struct Config {
...
+    /// This number of requests, multiplied by the maximum request body size the
+    /// HTTP server is configured with gives the rough amount of memory a HTTP
+    /// server will use to buffer request bodies in memory.
+    ///
+    /// A default maximum of 200 requests, multiplied by the default 10MiB
+    /// maximum for HTTP request bodies == ~2GiB.
+    #[clap(
+        long = "--max-http-requests",
+        env = "INFLUXDB_IOX_MAX_HTTP_REQUESTS",
+        default_value = "200"
```
domodwyer, line 83: I did the maths and this is a sensible default value for our env (and in general I imagine).

router2/src/server/http.rs
```diff
@@ -272,6 +302,24 @@ where
...
+        // aims to protect.)
+        let _permit = match self.request_sem.try_acquire() {
+            Ok(p) => p,
+            Err(TryAcquireError::NoPermits) => {
+                error!("simultaneous request limit exceeded - dropping request");
+                self.request_limit_rejected.inc(1);
+                return Err(Error::RequestLimit);
+            }
+            Err(e) => panic!("request limiter error: {}", e),
+        };
```
domodwyer, line 320: This is actually the entire functional change of this PR.

The rest is plumbing / testing

router2/src/server/http.rs
```diff
@@ -1074,4 +1125,130 @@ mod tests {
...
+    #[derive(Debug, Error)]
+    enum MockError {
+        #[error("bad stuff")]
+        Terrible,
+    }
+
+    // This test ensures the request limiter drops requests once the configured
+    // number of simultaneous requests are being serviced.
+    #[tokio::test]
+    async fn test_request_limit_enforced() {
```
domodwyer, line 1138: This (perhaps excessive) test ensures the limiter works correctly.

ioxd_router2/src/lib.rs
```diff
@@ -151,6 +151,7 @@ pub async fn create_router2_server_type(
     object_store: Arc<DynObjectStore>,
     write_buffer_config: &WriteBufferConfig,
     query_pool_name: &str,
+    request_limit: usize,
```
alamb, line 154:
```suggestion
    concurrent_request_limit: usize,
```

Maybe would make it more self-describing? Not required

router2/benches/e2e.rs
```diff
@@ -71,7 +71,7 @@ fn e2e_benchmarks(c: &mut Criterion) {
             partitioner.and_then(WriteSummaryAdapter::new(FanOutAdaptor::new(write_buffer))),
         );
 
-        HttpDelegate::new(1024, Arc::new(handler_stack), &metrics)
+        HttpDelegate::new(1024, 100, Arc::new(handler_stack), &metrics)
```
alamb, line 74:
```suggestion
        HttpDelegate::new(1024, 200, Arc::new(handler_stack), &metrics)
```

For consistency?

router2/tests/http.rs
```diff
@@ -117,7 +117,7 @@ impl TestContext {
 
         let handler_stack = InstrumentationDecorator::new("request", &*metrics, handler_stack);
 
-        let delegate = HttpDelegate::new(1024, Arc::new(handler_stack), &metrics);
+        let delegate = HttpDelegate::new(1024, 100, Arc::new(handler_stack), &metrics);
```
alamb, line 120:
```suggestion
        let delegate = HttpDelegate::new(1024, 200, Arc::new(handler_stack), &metrics);
```

router2/src/server/http.rs
alamb: it is very nice in my opinion 🏆

influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -339,6 +339,7 @@ pub async fn command(config: Config) -> Result<()> {
         Arc::clone(&object_store),
         &write_buffer_config,
         query_pool_name,
+        1_000, // max 1,000 concurrent HTTP requests
```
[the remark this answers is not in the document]
domodwyer: I just stuck "don't be a problem" numbers around - in tests, 100 was never going to be hit (1 would suffice). In the all-in-one, I had no idea what is acceptable 🤷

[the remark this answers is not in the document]
domodwyer: I pushed 46714b4b13ec72bed4dea9ca72fe7ab63047c21c to keep you sane :)

domodwyer:
> Often service protection limits like this are implemented by putting the service behind a webserver like nginx that offers such limits (and many more). I wonder if we have considered that approach in this case?

Check out #4350 - we've been doing exactly this! 

We (me and @mkmik) have been using istio (in front of IOx) to enforce limits there, but it's causing problems of it's own and is poorly documented. Given how easy it is to implement, I figured it was worth adding so I can try and rule out request throttling as a problem in prod - should be fairly easy to determine what is going on now we have metrics too.

I would like to change this in the future and move to limiting the amount of memory allocated for HTTP bodies rather than connections. Really, we care about memory pressure / OOMs and the connections are an very-easy-to-measure-and-limit, but inaccurate, proxy for the memory usage. I imagine we'll get best utilisation by limiting memory directly and this would have to be an internal limit in IOx.

mkmik:
> Often service protection limits like this are implemented by putting the service behind a webserver like `nginx` that offers such limits (and many more). I wonder if we have considered that approach in this case?

@alamb we indeed set the limits in our reverse proxy fabric (envoy proxy). The behaviour is not to return 503 error straight away but to apply backpressure by not accepting connections (although I'm a bit thin on the details of how does this work when multiplexing requests on a single http/2 connection; will investigate).

The effect of the backpressure will be to spread the request spikes over a longer period of time (thus increasing latency), possibly hitting timeouts on the client side and thus surfacing as 504 errors.

I think this PR doesn't hurt as it performs a hard limit; we still can employ another layer of limiting in front of it if it that's the behaviour we end up preferring.

alamb:
Thank you @domodwyer  and @mkmik  for the explanation -- makes sense to me.

PR #4420, opened 2022-04-25
refactor: grouping overlaps now uses the same overlap function in both compactor and deduplication

### LINKED ISSUE #4403: Refactor compactor's overlapped_group to use new NG overlap function
Refactor this function https://github.com/influxdata/influxdb_iox/blob/117569184e48750bdc5906c0ebc1407d99efc709/compactor/src/compact.rs#L868 to use the new one in PR #4389 4389

NGA-TRAN:
Closes #4403 


query/src/lib.rs
```diff
@@ -286,6 +286,35 @@ where
     }
 }
 
+/// Implement ChunkMeta for Arc<dyn QueryChunk>
+impl QueryChunkMeta for Arc<dyn QueryChunk> {
```
NGA-TRAN, line 290: This is needed because the parameter of `group_potential_duplicates` function is now `QueryChunkMeta` rather than `QueryChunk`

query/src/provider/overlap.rs
```diff
@@ -107,11 +111,42 @@ pub fn group_potential_duplicates_og(
     Ok(groups)
 }
 
-/// Groups [`QueryChunk`] objects into disjoint sets of overlapped time range.
+impl QueryChunkMeta for ParquetFileWithMetadata {
```
NGA-TRAN, line 114: Implement this so the`group_potential_duplicates` can work on `ParquetFileWithMetadata`. The `unimplemented` is done on purpose those specific functions

query/src/provider/overlap.rs
```diff
@@ -415,6 +450,7 @@ mod test {
         let c4 = Arc::new(TestChunk::new("chunk4").with_timestamp_min_max(25, 35));
 
         let groups = group_potential_duplicates(vec![c1, c2, c3, c4]).expect("grouping succeeded");
+        let groups = to_group_query_chunks(groups);
```
NGA-TRAN, line 453: From `Vec<Vec<Arc<TestChunk>>>` to `Vec<Vec<Arc<dyn QueryChunk>>>`

query/src/lib.rs
alamb: I think this is fine

query/src/provider/overlap.rs
alamb: maybe you can lave a comment with the rationale of *why* it was left unimplemented? It is not immediately clear to me the reason (is it because we don't want people to actually query the data there?)

query/src/provider/overlap.rs
alamb: Maybe this function signature could be changes to take `Vec<Arc<T>>` to avoid the need for the cast or change the tests to use `TestChunk` directly rather than `Arc<TestChunk>`

compactor/src/compact.rs
```diff
@@ -868,60 +869,30 @@ impl Compactor {
...
-        for gr in grouper {
-            cumulative_sum += match gr.start_end {
-                StartEnd::Start => 1,
-                StartEnd::End(_) => -1,
+        // group overlap files
+        let overlapped_groups =
+            group_potential_duplicates(parquet_files).expect("Error grouping overlapped chunks");
+
+        // Compute min time and total size for each overllaped group
+        let mut groups_with_min_time_and_size = Vec::with_capacity(overlapped_groups.len());
```
alamb, line 877: ❤️

query/src/provider/overlap.rs
```diff
@@ -968,4 +1012,17 @@ mod test {
...
+    fn to_group_query_chunks(groups: Vec<Vec<Arc<TestChunk>>>) -> Vec<Vec<Arc<dyn QueryChunk>>> {
+        groups
+            .iter()
+            .map(|chunks| {
+                chunks
+                    .iter()
+                    .map(|c| Arc::clone(c) as Arc<dyn QueryChunk>)
+                    .collect::<Vec<_>>()
+            })
+            .collect::<Vec<_>>()
```
alamb, line 1026: I wonder if you could do something like this (untested):

```suggestion
        groups
            .into_iter()
            .map(|chunks| {
                chunks
                    .into_iter()
                    .map(|c| as _)
                    .collect::<Vec<_>>()
            })
            .collect::<Vec<_>>()
```

alamb (approved):
Thank you @NGA-TRAN  -- looks very nice 👍

query/src/provider/overlap.rs
NGA-TRAN: Which function do you mean to change the signature, @alamb? I have tried to change signature of this function `to_string` https://github.com/influxdata/influxdb_iox/blob/e3caf24954a9f18b1a03dce1f702c3559c51d7b4/query/src/provider/overlap.rs#L963
as it is used in the macro to get `table_name` for a chunk but it does not work because OG returns `Arc<dyn QueryChunk` and NG returns `Arc<TestChunk>`

query/src/provider/overlap.rs
NGA-TRAN: It works. Thanks @alamb

query/src/provider/overlap.rs
NGA-TRAN: Done

Issue #4394, opened 2022-04-22
ScalarUDFs can't be serialized to ingester (Panicked at 'not implemented', generated_types/src/ingester.rs:652:35)


alamb:
When porting some of the influxrpc tests over to NG in https://github.com/influxdata/influxdb_iox/pull/4383 I get a panic in the deserialization of UDFs in the ingester

Note this is in code we copied from DataFusion / Ballista which has subsequently implemented the necessary  code (see https://github.com/apache/arrow-datafusion/pull/2130) but since we have our own fork we did not get the same upgrade.  The work to complete this is tracked in https://github.com/influxdata/influxdb_iox/pull/3997

The code that is panic'ing is https://github.com/influxdata/influxdb_iox/blob/e8bfd7a537a1df4fc3ed87f190f32a52af05d50d/generated_types/src/ingester.rs#L652-L653

Here is the stacktrace

```
2022-04-22T11:30:30.973731Z  INFO service_grpc_influxrpc::service: read filter db_name=1485094707946523_2383090981010620 req.range=Some(TimestampRange { start: 0, end: 2001 }) predicate=(TagRef:host RegEx RegEx:^b.+)
2022-04-22T11:30:31.029532Z ERROR panic_logging: Thread panic panic_info=panicked at 'not implemented', generated_types/src/ingester.rs:652:35
thread 'IOx Query Executor Thread' panicked at 'not implemented', generated_types/src/ingester.rs:652:35
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:48:5
   3: generated_types::ingester::from_expr
             at /Users/alamb/Software/influxdb_iox/generated_types/src/ingester.rs:652:35
   4: core::ops::function::FnMut::call_mut
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:150:5
   5: core::iter::adapters::map::map_try_fold::{{closure}}
[...]

PR #4421, opened 2022-04-25
fix: Use datafusion serialization code rather than our own copy of it

alamb:
~Draft as this needs a few subtasks to be completed:~
- [x] https://github.com/apache/arrow-datafusion/issues/2340 / https://github.com/apache/arrow-datafusion/pull/2341
- [x] https://github.com/influxdata/influxdb_iox/pull/4428
- [x] https://github.com/influxdata/influxdb_iox/pull/4431
- [x] https://github.com/influxdata/influxdb_iox/pull/4432
- [x] https://github.com/influxdata/influxdb_iox/pull/4440

Follow on work
- [ ] https://github.com/influxdata/influxdb_iox/issues/4448 (not necessary for now as querier doesn't send aggregates to  ingester)

# Rationale:
Resolves https://github.com/influxdata/influxdb_iox/issues/4394

The lack of support for serializing UDFs is blocking https://github.com/influxdata/influxdb_iox/pull/4383 which has since been fixed in DataFusion https://github.com/apache/arrow-datafusion/pull/2130 but the code has not been ported to IOx. Instead of doing so, this PR switches IOx to use DataFusion's serialization directly

# Changes:
Rather than try to port over datafusion serialization code (and also keep up with it in the future) this PR changes IOx to use the datafusion serialization code directly and deletes a bunch of forked code in the process

Pros of this approach 😄 : 
1. NG -> OG migration is unblocked
1. Less code in IOx
2. We will stay up to date with serialization in datafusion with our standard upgrades, and no additional effort on our part

Drawbacks of this approach 😢  :
1. Serialized exprs become opaque to anyone trying to observe raw GRPC messages without additional decoding.

We discussed this previously on https://github.com/influxdata/influxdb_iox/pull/3997 https://github.com/influxdata/influxdb_iox/pull/3997#issuecomment-1077949298 and @tustvold brought up the "exprs are opaque" issue.

I would like to proceed with this approach and we can revisit / try to have native protobuf support in the future if/when that would be helpful for debugging and operating the system.


datafusion/src/lib.rs
```diff
@@ -1 +1,4 @@
 pub use upstream::*;
+
+// Publically re-export datafusion-proto crate as well
```
alamb, line 3: Pull in the datafusion code

generated_types/protos/influxdata/iox/ingester/v1/query.proto
```diff
@@ -106,429 +108,9 @@ message TimestampRange {
...
-message Union {
-  repeated Field union_types = 1;
-  UnionMode union_mode = 2;
-}
-
-message Dictionary {
-  ArrowType key = 1;
-  ArrowType value = 2;
+  // Encoded using DataFusion's Expr serialization code
+  bytes expr = 1;
```
alamb, line 115: again, I don't really want to deal with backwards compatibility at this point as it doesn't affect anything yet. Just making the protobuf "compatible" will still result in errors without a bunch more code so I don't want to pretend we are compatible even at the protobuf layer.

domodwyer (approved):
🧹

PR #4428, opened 2022-04-26
refactor: Move IOx UDfs into a Function Registry (1/3)

alamb:
This is PR 1 of 3 I have planned. This one simply moves some code around mechanically so should be straightforward to review.

# Rationale

NG's design calls for serializing `Expr`s to send them from the querier to the ingester. Serialization currently doesn't work for User Defined Functions (UDFs) https://github.com/influxdata/influxdb_iox/issues/4394 , as the deserialization requires some way to map strings back to function pointers.

This mapping is done using a [`FunctionRegistry`](https://github.com/apache/arrow-datafusion/blob/41b4e491663029f653e491b110d0b5e74d08a0b6/datafusion/core/src/logical_plan/registry.rs#L23) in DataFusion.

Previously we have been playing fast and loose with creating one off UDFs with custom closures as needed. However, since we can't serialize closures across the network we need to make this smarter so we can deserialize `Expr`s)

# Changes
1. Move all IOx user defined functions to query_udfs crate (so they can be used in the generated_types code).

My next PR will refactor the functions and create a `FunctionRegistry`, rather than being created with a custom closure each time.


query_functions/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+//! DataFusion User Defined Functions (UDF/ UDAF) for IOx
+#![deny(rustdoc::broken_intra_doc_links, rust_2018_idioms)]
+#![warn(
```
alamb, line 3: I added the "standard" lints to this crate which provoked some additional cleanup in regex.rs

domodwyer (approved):
I like where this is going 🧹

alamb:
Thank you @domodwyer

PR #4429, opened 2022-04-26
fix: extend + harden querier `ensure_schema`

crepererum:
- only convert dictionary types that we really want to convert (instead
  of blindly converting all types)
- handle missing / NULL columns


querier/src/ingester/mod.rs
```diff
@@ -558,27 +566,71 @@ impl QueryChunk for IngesterPartition {
...
+                            actual_data_type: actual_type.clone(),
+                            desired_data_type: desired_type.clone(),
+                        }
+                        .fail()
+                    } else {
+                        Ok(Arc::clone(col))
+                    }
+                }
+                None => Ok(arrow::array::new_null_array(desired_type, batch.num_rows())),
+            }
```
NGA-TRAN, line 633: I wonder if we need this because this is already handled when we create the query plan. If you test data that is read from a parquet file that does not include a column in the query, that missing column is  handled during the planning. Should we let this  handled there, too?
My point is it is fine if the batch missing a column. If you allow it here, the NULL column will be added during the plan instead of extra work here.

NGA-TRAN (approved):
In general, adding NULL columns at this stage will avoid them to get added later and I am fine with that. However, I highly suspect if you allow the missing columns here and only modify the code somehow to make it work, the NULL columns will be automatically added with the rest of the existing query planning.

querier/src/ingester/mod.rs
alamb: > My point is it is fine if the batch missing a column. If you allow it here, the NULL column will be added during the plan instead of extra work here.

Agree -- the adding of null columns for missing schemas happens in the provider --

https://github.com/influxdata/influxdb_iox/blob/main/query/src/provider.rs#L103-L116

What test is failing without this code? Perhaps I can help debug it further?

querier/src/ingester/mod.rs
```diff
@@ -558,27 +566,71 @@ impl QueryChunk for IngesterPartition {
...
 /// `DataType::Utf8` even when they were sent as
 /// `DataType::Dictionary(Int32, Utf8)`.
+///
+/// # NULL-column Creation
+/// If a column is absent in an ingester partition it will not be part of record batch even when the querier requests
+/// it. In that case we create it as "all NULL" column with the appropriate type.
+///
+/// An alternative would be to remove the column from the schema of the appropriate [`IngesterPartition`]. However since
+/// a partition may contain multiple record batches and we do not want to assume that the presence/absence of columns is
+/// identical for all of them, we fix this here.
```
crepererum, line 593: @alamb this may answer your question. Just messing w/ the schema for the record batch is not sufficient because then the `IngesterPartition` misreports the schema, leading to:

`ArrowError(ExternalError(Internal("Internal error creating SchemaAdapterStream: field 'temp' does not appear in the output schema")))`

alamb: 🤔

NGA-TRAN: @crepererum and @alamb 
I have done with the sortkey PR and can help look into this issue. If @crepererum point me to the place before this code, I will debug it.

Since I plan to look into query plans built in querier to make sure overlap and sort are done correctly (to avoid bad performance), this can be a part of that work, too

crepererum: Could we agree on merging this since it is functionally correct? I'll add tracking issue to optimize this. The problem here is that the bug only shows up with the extended query tests, which I need to file as a PR later today.

alamb: I think merging this PR to unblock your other work, and filing a ticket to look into the underlying bug (any maybe reverting this change if it turns out it is unnecessary) that maybe @NGA-TRAN can look into would be a good idea

alamb (approved):

querier/src/ingester/mod.rs
crepererum: #4438.

PR #4431, opened 2022-04-26
feat: Implement IOx Function Registry + `regex_match` / `regex_not_match`  (2/3)

alamb:
~Draft until https://github.com/influxdata/influxdb_iox/pull/4428 is merged~

Follow on to https://github.com/influxdata/influxdb_iox/pull/4428

# Rationale

NG's design calls for serializing `Expr`s to send them from the querier to the ingester. This process currently doesn't work for User Defined Functions (UDFs), as the deserialization requires some way to map strings back to function pointers.

This is done using a [`FunctionRegistry`](https://github.com/apache/arrow-datafusion/blob/41b4e491663029f653e491b110d0b5e74d08a0b6/datafusion/core/src/logical_plan/registry.rs#L23) in DataFusion.

Previously we have been playing fast and loose with creating one off UDFs with custom closures as needed. However, since we can't serialize closures across the network we need to make this smarter so we can deserialize Exprs.

# Changes
1. Create a global function registry for IOx
2. Port `regex` / `regex_not_match` to use that registry


query_functions/src/lib.rs
```diff
@@ -8,13 +8,54 @@
...
+/// Function registry
+mod registry;
+
+/// Return an Expr that invokes a InfluxRPC compatible regex match to
+/// determine which values satisfy the pattern. Equivalent to:
+///
+/// ```text
+/// col ~= /pattern/
+/// ```
+pub fn regex_match_expr(input: Expr, pattern: String) -> Expr {
```
alamb, line 38: This is now the public interface.

The whole point of this PR is that the `Expr::FunctionCall` gets used in our plans is created by looking up a pre-existing `ScalarUDF` rather than creating a new one each time.

query_functions/src/regex.rs
```diff
@@ -29,19 +56,52 @@ pub const REGEX_NOT_MATCH_UDF_NAME: &str = "RegexNotMatch";
...
-    // the golang regexp library which is different than Rust's regexp
-    let pattern = clean_non_meta_escapes(pattern);
+fn regex_match_expr_impl(matches: bool) -> ScalarFunctionImplementation {
+    // N.B., this function does not utilise the Arrow regexp compute
+    // kernel because in order to act as a filter it needs to return a
+    // boolean array of comparison results, not an array of strings as
+    // the regex compute kernel does and it needs to implement the
+    // regexp syntax for influxrpc.
+
+    let func = move |args: &[ColumnarValue]| {
```
alamb, line 66: The signature of the closure changed to take ColumnarValue (which is either a scalar or array) so we could use the same implementation of `regex_match_expr` for any pattern. Previously there were diferent `ScalarFunctionImplementation`s for each different pattern that was used

query_functions/src/regex.rs
```diff
@@ -29,55 +56,74 @@ pub const REGEX_NOT_MATCH_UDF_NAME: &str = "RegexNotMatch";
...
+                        // Here we decide to make our UDF to return null when either base or exponent is null.
+                        row.map(|v| pattern.is_match(v) == matches)
+                    })
+                    .collect::<BooleanArray>();
+
+                Ok(ColumnarValue::Array(Arc::new(results) as ArrayRef))
+            }
+            ColumnarValue::Scalar(ScalarValue::Utf8(row)) => {
+                let res = row.as_ref().map(|v| pattern.is_match(v) == matches);
+                Ok(ColumnarValue::Scalar(ScalarValue::Boolean(res)))
```
alamb, line 115: I don't think it matters, but this code now natively handles the `scalar` case (rather than converting such data into a single row `BooleanArray`) 🤷

alamb (commented):
The existing test coverage for this function is quite significant (as I discovered when changing it 😆 )

NGA-TRAN (approved):
This is not my area but I have gone over the code, the functions and the changes. They make sense to me

PR #4432, opened 2022-04-26
feat: Add window_bounds to IOx Function Registry (2.5/3)

alamb:
~Draft until https://github.com/influxdata/influxdb_iox/pull/4431 is merged~

Follow on to https://github.com/influxdata/influxdb_iox/pull/4428

# Rationale

Apply same treatment to `window_bounds` as was applied to `regex_match`/`regex_not_match` in https://github.com/influxdata/influxdb_iox/pull/4431

See rationale on https://github.com/influxdata/influxdb_iox/pull/4431 for more details

# Changes
1. Use a single `ScalarUDF` for `window_bounds` rather than creating new `ScalarUDFs` for each set of arguments. 
2. Add `EncodedWindowDuration` for converting back/forth between WindowDuration and `ScalarValue`s


query_functions/src/lib.rs
```diff
@@ -55,6 +57,30 @@ pub fn regex_not_match_expr(input: Expr, pattern: String) -> Expr {
...
 }
 
+/// Create a DataFusion `Expr` that invokes `window_bounds` with the
+/// appropriate every and offset arguments at runtime
+pub fn make_window_bound_expr(
+    time_arg: Expr,
+    every: WindowDuration,
+    offset: WindowDuration,
+) -> Expr {
+    let encoded_every: EncodedWindowDuration = every.into();
```
alamb, line 67: It is somewhat unfortunate to have to encode the `WindowDuration` structure as 2 sets of 3 `ScalarValues` but the support in DataFusion for proper struct types is not yet all that good: https://github.com/apache/arrow-datafusion/issues/2326

I also suspect the overhead for single (scalar) values is likely to be quite high

predicate/src/lib.rs
```diff
@@ -327,18 +327,16 @@ impl PredicateBuilder {
 
     /// Builds a regex matching expression from the provided column name and
     /// pattern. Values not matching the regex will be filtered out.
-    pub fn build_regex_match_expr(self, column: &str, pattern: impl Into<String>) -> Self {
-        self.regex_match_expr(column, pattern, true)
+    pub fn build_regex_match_expr(mut self, column: &str, pattern: impl Into<String>) -> Self {
+        let expr = query_functions::regex_match_expr(col(column), pattern.into());
```
domodwyer, line 331: This `regex_match_expr` and `regex_not_match_expr` is much clearer than having some stray bool in the args (pet dislike - we have descriptive enums!) 👍

query_functions/src/lib.rs
domodwyer: I'm intrigued - how come using a `ScalarValue` representation has relatively high overhead?

query_functions/src/lib.rs
```diff
@@ -8,13 +8,80 @@
...
+        .expect("WindowBounds function not registered")
+        .call(vec![
+            time_arg,
+            lit(encoded_every.ty),
+            lit(encoded_every.field1),
+            lit(encoded_every.field2),
+            lit(encoded_offset.ty),
+            lit(encoded_offset.field1),
+            lit(encoded_offset.field2),
+        ])
```
domodwyer, line 81: Do we have tests that drive/cover this fns, primarily to validate argument order? I imagine there's probably some indirect coverage?

query_functions/src/regex.rs
```diff
@@ -1,21 +1,50 @@
...
+pub(crate) const REGEX_NOT_MATCH_UDF_NAME: &str = "RegexNotMatch";
+
+lazy_static::lazy_static! {
+    /// Implementation of regexp_match
+    pub(crate) static ref REGEX_MATCH_UDF: Arc<ScalarUDF> = Arc::new(
+        create_udf(
+            REGEX_MATCH_UDF_NAME,
+            // takes two arguments: regex, pattern
+            vec![DataType::Utf8, DataType::Utf8],
+            Arc::new(DataType::Boolean),
```
domodwyer, line 28: Having to `Arc` an enum discriminant is rather unfortunate.

I imagine it is so we can share data in some `DataType` variants (perhaps variants like `DataType::List(Box<Field>)`) right?

Would it be possible to move the `Arc` _into_ variants that need it (so, `DataType::List(Arc<Box<Field>>)` or even `DataType::List(Arc<Field>)`) and then making `DataType` copy? Almost all `DataType` variants could be copy and avoid the `Arc` into the heap entirely! I'd be happy to do this if it's a go-er 👍

query_functions/src/regex.rs
```diff
@@ -235,7 +283,14 @@ mod test {
...
 
         for (pattern, matches, expected) in cases.into_iter() {
-            let regex_expr = super::regex_match_expr(col("words"), pattern.to_string(), matches);
+            let args = vec![col("words"), lit(pattern)];
+
+            let regex_expr = if matches {
+                REGEX_MATCH_UDF.call(args)
+            } else {
+                REGEX_NOT_MATCH_UDF.call(args)
+            };
```
domodwyer, line 292: Perhaps we could drive these tests through the registry to check the parameters are serialised and passed through correctly?

domodwyer (approved):
Keep it coming!

query_functions/src/lib.rs
alamb: Yes, the `influxrpc` tests in query_tests cover it as do the end to end tests (it is used in `read_aggregate` tests)

query_functions/src/lib.rs
alamb: Well in this case I was thinking that a single rust enum with a `i64` and possibly a `bool` was converted into 3 `ScalarValue`s, each of which I think is 48 bytes (to hold the largest variant) along with an owned `String`

So in the grand scheme of things I doubt it matters. In the "percentage overhead" category its like a 1000% which makes me irrationally sad

query_functions/src/regex.rs
alamb: The reason for the `Arc` on return type *here* is that it is part of the DataFusion signature: https://github.com/apache/arrow-datafusion/blob/5f0b61b0db9849336e2e83b23c8a45508a85fb38/datafusion/expr/src/function.rs#L43

I am not sure why the `return_type` in DataFusion is `Arc`d in the first place. 

The work to make `DataType` implement `Copy` would start in the arrow-rs crate I think -- it would likely result in a large amount of code churn (but result in a better representation)

query_functions/src/regex.rs
alamb: That is how it worked originally and I thought it was kind of messy to have `lib.rs` depend on `regex.rs` then call back into `lib.rs`.

I'll add an explicit test in `lib.rs` for coverage

alamb: Added in b34e858c4

query_functions/src/lib.rs
```diff
@@ -55,7 +57,139 @@ pub fn regex_not_match_expr(input: Expr, pattern: String) -> Expr {
...
+        array::{ArrayRef, StringArray, TimestampNanosecondArray},
+        record_batch::RecordBatch,
+    };
+    use datafusion::{assert_batches_eq, logical_plan::col};
+    use datafusion_util::context_with_table;
+    use std::sync::Arc;
+
+    use super::*;
+
+    /// plumbing test to validate registry is connected. functions are
```
alamb, line 101: these tests were added since @domodwyer 's review

alamb:
I added some more tests to this PR, which I don't think will be controversial and thus I am going to merge this PR to keep the train rolling. I will be happy to update / make additional changes as a follow on PR if anything else is desired.

PR #4433, opened 2022-04-27
feat: add CLI command `remote pull partition`

### LINKED ISSUE #4315: Add tooling to pull data from prod

pauldix:
This adds the functionality to pull a remote partition and its parquet files by namespace name, table name and partition key down to a local catalog and object store. It injects the schema into the catalog along with the partition and parquet files.

* Add lookup of partitions by table id to catalog.
* Add API to catalog to return partitions by table id.
* Add to client to return partitions by table id.
* Add CLI to pull remote schema, partition, and parquet files into a local catalog and object store.

Closes #4315


influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -49,5 +105,489 @@ pub async fn command(connection: Connection, config: Config) -> Result<(), Error
...
+                    parquet_file.sequencer_id,
+                    parquet_file.partition_id,
+                    parquet_file.object_store_id,
+                )
+                .absolute_dirs_and_file_name();
+                let path = object_store.path_from_dirs_and_filename(path);
+                match object_store.get(&path).await {
+                    Ok(_) => {
+                        println!(
+                            "skipping file {} already in the object store",
```
domodwyer, line 165:
```suggestion
                            "skipping file {} already in the local object store",
```

domodwyer: Nice idea de-duplicating requests 👍

influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -49,5 +105,489 @@ pub async fn command(connection: Connection, config: Config) -> Result<(), Error
...
+                                parquet_file.object_store_id
+                            );
+                        });
+                        handles.push(task);
+                    }
+                    e => return Err(Error::ObjectStore(e.unwrap_err())),
+                }
+            }
+
+            join_all(handles).await;
```
domodwyer, line 201: If one of the tokio tasks panics, it is not propagated to the main thread and so the program will still return 0 / OK which can be annoying if scripting!

This code iterates over the task handles and panics if any of the tokio tasks have panicked (due to an `unwrap()` or `expect()`):

```suggestion
            join_all(handles)
                .await
                .into_iter()
                .try_collect::<Vec<_>>()
                .expect("worker thread crashed");
```

With the above, a failed task will return non-0 👍

domodwyer (approved):
SUCH POWER 🎉

alamb (commented):
I tried it out:

Set up a port forward:

```
(arrow_dev) alamb@MacBook-Pro-3:~/Software/influxdb_iox$ kubectl --context dex@stag03-us-east-1-aws -n iox port-forward iox-router-5f4d5c67fd-5bjtl 8082
```

And then:

```
(arrow_dev) alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ./target/debug/influxdb_iox remote partition pull
error: The following required arguments were not provided:
    <NAMESPACE>
    <TABLE>
    <PARTITION_KEY>

USAGE:
    influxdb_iox remote partition pull [OPTIONS] <NAMESPACE> <TABLE> <PARTITION_KEY>

For more information try --help
```

I had to go digging in the staging postgres catalog to find the correct ids, but sadly looks like I can't actually test until the serve change is rolled out (unsurprisingly)

```shell
./target/debug/influxdb_iox remote partition pull 3bd2399eac571807_8dc736eef15fc59d  cpu 2022-04-25 --catalog-dsn postgresql://localhost:5432/alamb
getting schema from remote for namespace 3bd2399eac571807_8dc736eef15fc59d
Error in partition subcommand: Client error: Operation is not implemented or supported: 
```

So all in all, can't wait to see this live 👍

PR #4440, opened 2022-04-27
refactor: Use DataFusion `Expr` instead of our own custom wrapper for `ValueExpr`

alamb:
# Rationale
I am switching to use DataFusion serialization for Exprs https://github.com/influxdata/influxdb_iox/issues/4394

BinaryExpr currently contains subparts of `Exprs `(e.g. `Column` and `Operator`). I would like to remove all such custom serialization code in https://github.com/influxdata/influxdb_iox/issues/4394

# Changes:
1.  Rename `BinaryExpr` to `ValueExpr` to better describe what it is and not to shadow `Expr::BinaryExpr`
2. Change representation to a wrapper around an `Expr`
3. Clean up serialization code


generated_types/src/ingester.rs
```diff
@@ -121,16 +121,14 @@ impl TryFrom<proto::Predicate> for Predicate {
         let value_expr = value_expr
             .into_iter()
             .map(|ve| {
-                let left = ve.left.unwrap_field("left")?;
-                let right = ve.right.unwrap_field("right")?;
-
-                Ok(BinaryExpr {
-                    left: from_proto_column(left),
-                    op: from_proto_binary_op("op", &ve.op)?,
-                    right: from_proto_expr(right)?,
```
alamb, line 130: The changes in this file are the reason for this PR -- to stop calling `from_proto_column` and `from_proto_binary_op` which I am removing in https://github.com/influxdata/influxdb_iox/pull/4421

predicate/src/rpc_predicate.rs
```diff
@@ -287,39 +286,28 @@ impl ExprRewriter for MeasurementRewriter<'_> {
...
 }
 
 struct FieldValueRewriter<'a> {
-    value_exprs: &'a mut Vec<BinaryExpr>,
+    value_exprs: &'a mut Vec<ValueExpr>,
 }
 
 impl<'a> ExprRewriter for FieldValueRewriter<'a> {
     fn mutate(&mut self, expr: Expr) -> DataFusionResult<Expr> {
-        Ok(match expr {
```
alamb, line 303: moved to `ValueExpr::try_new`

predicate/src/lib.rs
```diff
@@ -450,12 +451,45 @@ impl PredicateBuilder {
...
+            if let Expr::Column(inner) = left.as_ref() {
+                if inner.name == VALUE_COLUMN_NAME {
+                    return Ok(Self { expr });
+                }
+            }
+        }
+        Err(expr)
+    }
+
+    /// Returns a new Expr with th reference to `_value` replaced with the specified value and
```
carols10cents, line 480: This comment looks incomplete and has a typo :)

```suggestion
    /// Returns a new Expr with the reference to `_value` replaced with the specified value and...???
```

predicate/src/lib.rs
```diff
@@ -450,12 +451,45 @@ impl PredicateBuilder {
...
-    pub op: Operator,
-    pub right: Expr,
+pub struct ValueExpr {
+    expr: Expr,
+}
+
+impl ValueExpr {
+    /// tries to create a new ValueExpr. If `expr` follows the
+    /// expected pattrn, returns Ok(Self). If not, returns Err(expr)
+    pub fn try_new(expr: Expr) -> std::result::Result<Self, Expr> {
```
carols10cents, line 464: This `try_new` function looks a lot like `TryFrom::try_from`... any reason this function couldn't be in an `impl TryFrom<Expr> for ValueExpr`?

predicate/src/lib.rs
```diff
@@ -450,12 +451,45 @@ impl PredicateBuilder {
...
+            binary_expr(col(name), *op, right.as_ref().clone())
+        } else {
+            unreachable!("Unexpected content in ValueExpr")
+        }
+    }
+
+    /// return the inner Expr, consuming `self`
+    pub fn into_expr(self) -> Expr {
+        self.expr
+    }
```
carols10cents, line 492: Similarly, it looks like this method could be:

```
impl From<ValueExpr> for Expr {
    fn from(value_expr: ValueExpr) -> Self {
        value_expr.expr
    }
}
```

Any reason I'm missing why not?

carols10cents (approved):
Just a few small thoughts, overall looks good!

predicate/src/lib.rs
alamb: No reason at all -- good call -- will do

predicate/src/lib.rs
alamb: good call -- done

PR #4446, opened 2022-04-27
feat(router2): pre-warm namespace schema cache

domodwyer:
This should significantly reduce the PG load induced by deployments and significantly reduce PG pool saturation / resulting back-pressure / rejected requests 🎉

Closes https://github.com/influxdata/influxdb_iox/issues/3577

---

* feat(iox_catalog): TableRepo::list() (eb5abce99)

      Allow all tables in the catalog to be fetched.

* feat(iox_catalog): ColumnRepo::list() (874521da8)

      Allow all columns in the catalog to be fetched.

* feat(iox_catalog): list_schemas() (bb8a19b57)

      Adds a function to resolve an atomic snapshot of all NamespaceSchema in the
      catalog with minimal query overhead.

* feat(router2): pre-warm schema cache (d7d4de514)

      Fetch all NamespaceSchema from the catalog and pre-cache them during router
      initialisation.

      This pre-warming happens in 3 queries, and should significantly reduce the
      load on the Postgres instance backing the catalog during deployments/after
      router crashes/etc.

      Pre-warming occurs during init, before the HTTP endpoints are bound, and 
      therefore before the /health HTTP endpoint can return OK. This ensures the
      router pre-warms the cache before it begins receiving traffic.

### DIFF
--- a/ioxd_router2/src/lib.rs
+++ b/ioxd_router2/src/lib.rs
@@ -6,7 +6,7 @@ use std::{
 
 use async_trait::async_trait;
 use clap_blocks::write_buffer::WriteBufferConfig;
-use data_types2::{PartitionTemplate, TemplatePart};
+use data_types2::{DatabaseName, PartitionTemplate, TemplatePart};
 use hashbrown::HashMap;
 use hyper::{Body, Request, Response};
 use iox_catalog::interface::Catalog;
@@ -19,7 +19,9 @@ use router2::{
         NamespaceAutocreation, Partitioner, SchemaValidator, ShardedWriteBuffer,
         WriteSummaryAdapter,
     },
-    namespace_cache::{metrics::InstrumentedCache, MemoryNamespaceCache, ShardedCache},
+    namespace_cache::{
+        metrics::InstrumentedCache, MemoryNamespaceCache, NamespaceCache, ShardedCache,
+    },
     sequencer::Sequencer,
     server::{grpc::GrpcDelegate, http::HttpDelegate, RouterServer},
     sharder::JumpHash,
@@ -174,6 +176,10 @@ pub async fn create_router2_server_type(
         &*metrics,
     ));
 
+    pre_warm_schema_cache(&ns_cache, &*catalog)
+        .await
+        .expect("namespace cache pre-warming failed");
+
     // Initialise and instrument the schema validator
     let schema_validator =
         SchemaValidator::new(Arc::clone(&catalog), Arc::clone(&ns_cache), &*metrics);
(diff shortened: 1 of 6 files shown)


mkmik (approved):
> Pre-warming occurs during init, before the HTTP endpoints are bound, and 
therefore before the /health HTTP endpoint can return OK. This ensures the
router pre-warms the cache before it begins receiving traffic.

the downside of that is that we cannot expose metrics or /debug/pprof data while we're initializing. It's not a big deal, but it's confusing.

Ideally we should serve the health check endpoint but have it return an error code until it's fully initialized.

domodwyer:
This can be done in the future, but the HTTP server framework is from OG and exposes no way for the actual application logic to pass/fail the health check so needs some work...

mkmik:
(yes, my comment was in passing and not a blocker for this PR)

ioxd_router2/src/lib.rs
```diff
@@ -318,3 +324,68 @@ async fn init_write_buffer(
...
+
+/// Pre-populate `cache` with the all existing schemas in `catalog`.
+async fn pre_warm_schema_cache<T>(
+    cache: &T,
+    catalog: &dyn Catalog,
+) -> Result<(), iox_catalog::interface::Error>
+where
+    T: NamespaceCache,
+{
+    iox_catalog::interface::list_schemas(catalog)
```
alamb, line 336: I wonder if there is / should be any limit on the number of items in this cache / fetch the "most recently used" items?

As it is, it will load all schemas for all time even though only schemas receiving writes will be needed. So I worry that this may delay the rollout of new versions unnecessarily. Maybe it doesn't matter all that much or we can optimize it more later, but I wanted to mention the possibility.

ioxd_router2/src/lib.rs
```diff
@@ -174,6 +176,10 @@ pub async fn create_router2_server_type(
         &*metrics,
     ));
 
+    pre_warm_schema_cache(&ns_cache, &*catalog)
```
alamb, line 179: Is it worth a note here about why prewarming the cache is needed? Like

```suggestion
    // load schemas / tables / columns into the table prior to serving
    // to avoid dropping writes
    // https://github.com/influxdata/influxdb_iox/issues/3577
    pre_warm_schema_cache(&ns_cache, &*catalog)
```

alamb (approved):
looks like something good to try

alamb:
The end to end test failures may actually be related to this PR -- I haven't seen that same failure signature today 🤔

alamb:
But then again that is the problem with flaky end to end tests -- it is hard to know what is causing the issue

PR #4452, opened 2022-04-27
docs: document for deduplication and sort key

NGA-TRAN:


query/src/frontend/reorg.rs
```diff
@@ -57,17 +57,6 @@ impl ReorgPlanner {
...
-    //   1. OG lifecycle's load_chunk which will go way
-    //   2. NG ingester
-    //  In both places, data is not yet sorted and may contain duplicates.
-    //   Hence, this function will create a plan with a sort operator added on top of the scan operator for the deduplication work.
-    // This function should NOT be used on sorted chunks. However, if you need to use it for chunks that are sorted on the SAME key,
-    //   you need to refactor this function to provide that sort_key and add this line
-    //    `builder.with_sort_key(your-newly-added-sortkey)` below `builder = builder.add_no_op_pruner()`
-    //   to make sure the plan is built using that sort_key and avoid resorting already sorted data.
-    //  Also note that if all of your provided chunks are not sorted on your provided sort_key, IOx plan builder
-    //   is smart enough to resort only chunks that are not yet sorted or sorted on different key.
```
NGA-TRAN, line 70: This comment is not accurate. The deduplicate is smart enough to discover sort keys of the chunks and use them

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
         assert_eq!(count, 2);
     }
 
+    #[tokio::test]
+    async fn alread_sorted_scan_plan_with_two_partitions_explain() {
```
NGA-TRAN, line 3171: A test on 2 different partitions which is needed and for me to verify while writing the document

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+║     └──────────────────┘ └─────────────────┘ └─────────────────┘ ║
+╚══════════════════════════════════════════════════════════════════╝
+Figure 1: Two partitions each has four chunks
+```
+
+### Data Deduplication
+
+Figure 2 shows a simplified query plan that IOx builds to read all eight chunks of these two parttions and deduplicate them to return only non-duplicated data. This diagram should be read bottom up as follows:
+- First, IOx scans eight chunks of the two partitions.
+- Chunks C1.1 and C1.2 overlap, they need to get deduplicated. Same to chunks C2.1 and C2.2. To do this, chunks C1.1 and C1.2 are sorted on the same sortt key K1.1_2 and the Dedup (deduplication) is applied to eliminate duplicated values. Similarly, chunks C2.1 and C2.2 are sorted on the same K2.1_2 sort key and then deduplicated.
```
lukebond, line 42: does it know that c1.1 & c1.2 overlap only by scanning them, or is there some data structure for a chunk which captures the time range of the chunk? might be obvious to some but isn't clear to me.

typo in this line: "sortt" should be "sort".

when you say "To do this, chunks C1.1 and C1.2 are sorted on the same sortt key K1.1_2" do you mean they are already sorted by the same key so dedup can happen easily, or are you saying that we now choose a key and sort the chunks so that we are able to dedup them? likewise with K2*

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+- Chunk C1.3 does not overlap with any chunks but potentially has duplicates in itself, hence it also needs to get sorted on its sort key K1.3 and the deduplicated. Same to chunk C2.3.
```
lukebond, line 43: i think this line answers my previous question. we are doing the sorting here. i think perhaps clarifying that in the previous line would be helpful?

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+- Chunk C1.4 and C2.4 do not overlap with any other chunks and each contain no duplicates, no extra work needed for them.
```
lukebond, line 44: does this mean that they may not be sorted and it's okay to leave them like that?

(edit: reading on later i got my answer when coming to figures 3&4)

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+    │          │           │          │            │         │         │        │                      
+┌──────┐   ┌──────┐    ┌──────┐   ┌──────┐     ┌──────┐  ┌──────┐   ┌──────┐ ┌──────┐               
+│ C1.1 │   │ C1.2 │    │ C2.1 │   │ C2.1 │     │ C1.3 │  │ C2.3 │   │ C1.4 │ │ C2.4 │               
+└──────┘   └──────┘    └──────┘   └──────┘     └──────┘  └──────┘   └──────┘ └──────┘                
+
+Figure 2: A simplified query plan that scans all eight chunks of two partitions and deduplicate their data                 
+```
+
+If many queries read these two partitions, IOx has to build the same query plan above which we think suboptimal because of sort and deduplication. To avoid this suboptimal query plan, IOx is designed to:
+1. Data in each partition is frequently compacted to only contains a few non-overlaped and non-duplicated chunks as shown in Figure 3 which will help scan fewer chunks and eliminate the `Sort` and `Dedup` operators as illustrated in Figure 4.
```
lukebond, line 74: `s/only contains/only contain/`
`s/non-overlaped/non-overlapping/`

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+2. All **persisted chunks of the same parttition are sorted on the same key** to avoid resorting before deduplication. So the plan in Figure 2 is only seen sometimes in IOx, instead, you will mostly see the plan in Figure 4 (for data in Fugure 3) or in Figure 6 (for data of Figure 5)
```
lukebond, line 75: typo here: "parttition" should be "partition"

"only seen sometimes in IOx"- when? i'm guessing for not-yet-persisted chunks?

lukebond (approved):
this really cleared things up for me, thank you.

as you will see, most of my questions got answered later on. wonder if there needs to be some reordering, or maybe a few "see below" comments so the reader can either jump around if they need, or relax because the answers will come when they keep reading?

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
+# Deduplication
+
+In an IOx table (aka measurement), data with the same primary key are considered duplicates. Primary key of a table is a set of all `tag` columns and the `time` column of that table. Duplicates can have different values of `field` columns and deduplication process is needed to only keep the latest field values.
+
+Data **deduplication** happens in `Ingester`, `Compactor`, and `Querier` on the [same code base](https://github.com/influxdata/influxdb_iox/blob/fa2c1febf4b9dfb2d3049d9840ac087d9c4eb427/query/src/provider.rs#L410). In order to deduplicate chunks, their data must be sorted on `the same sort order of their primany key` named **sort key**. Since existing chunks can be deduplicated again and again with new ingesting chunks, having the chunks sorted on a consistent sort order will help avoid resorting them.
```
alamb, line 5:
```suggestion
Data **deduplication** happens in `Ingester`, `Compactor`, and `Querier` using the [same code base](https://github.com/influxdata/influxdb_iox/blob/fa2c1febf4b9dfb2d3049d9840ac087d9c4eb427/query/src/provider.rs#L410). In order to deduplicate chunks, their data must be sorted on `the same sort order`, and that sort order must include all columns of their primary key. We define this sort order to be the **sort key**. Since existing chunks can be deduplicated again and again with newly ingested chunks, maintaining a consistent sort order for chunks which may contain duplicated data helps avoid resorting them.
```

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+
+This document explains how data chunks are deduplicated and sorted consistently to avoid resort.
+
+### Overlapped Chunks
+Since only chunks with overlapped time range may potentially have duplicates, we only deduplicate data of overlapped chunks. IOx partitions data by time (e.g. one day per partition), chunks in different partitions are guaranteed not to overlap. Figure 1 shows an example of two partitions, each has four chunks respectively. Two of them overlap with each other, the other two do not overlap with any but one has duplicates within the chunk itself and the other has no duplicats at all.
```
alamb, line 10:
```suggestion
Only chunks that might contain the same primary key might contain duplicates. 
   
Within the same table partition, this means that chunks with overlapped time ranges may potentially have duplicates. Since IOx partitions data by time (e.g. one day per partition), chunks in different partitions are guaranteed not to overlap. Figure 1 shows an example of two partitions. Each has four chunks respectively. Two of them overlap with each other, the other two do not overlap with any but one has duplicates within the chunk itself and the other has no duplicates at all.
```

docs/dedup_and_sort.md
alamb, line 42:
```suggestion
- Chunks C1.1 and C1.2 overlap, they need to get deduplicated. Same to chunks C2.1 and C2.2. To do this, chunks C1.1 and C1.2 are sorted on the same sort key K1.1_2 and the Dedup (deduplication) is applied to eliminate duplicated values. Similarly, chunks C2.1 and C2.2 are sorted on the same K2.1_2 sort key and then deduplicated.
```

docs/dedup_and_sort.md
alamb: I thought that Chunk C1.3, since it was in the same partition as C1.1 and C1.2, would be sorted using the same key as was used to deduplicate -- namely K1.1_2 as explained above

docs/dedup_and_sort.md
alamb: I think it may make the explanation clearer to skip / gloss over the theoretically different sort keys maybe, and focus on the case when the sort key is always the same, which should be the case almost all the time

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+└──────┘   └──────┘          
+Figure 6: Scan plan of data in Figure 5
+
+```
+
+### Sorted and Unsorted Chunks
+- Persisted chunks are always sorted and contains no duplicates in itself
+- Non-persisted chunks are not sorted and
+  - Every non-persited chunk in the Ingester may contain duplicates
+  - Every non-persisted chunk sent from Ingester to Querier contain no duplicates
```
alamb, line 137:
```suggestion
  - Every non-persisted chunk in the Ingester may contain duplicates
  - Every non-persisted chunk sent from Ingester to Querier contain no duplicates
```

alamb: I wonder why "Every non-persisted chunk sent from Ingester to Querier contain no duplicates"  -- is this required? Or is it an accident that we run a plan to apply filters in the ingester, resulting in deduplicated chunks?

Taking this one step farther, maybe we should *also* sort the data prior to sending from the ingester --> querier to maintain the invariant that the data is always sorted on the partition key 🤔

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Sort Keys
+Since IOx allows different sets of columns for various chunks of the same table, to keep the sort key consistent for all chunks in the same partition, IOx stores two kinds of sort keys:
+1. **Chunk Sort Key**: this is the sort key of the chunk and only includes tag and time columns of that chunk.
+2. **Super Sort Key**: this is the sort key of all chunks of the same partition. This sort key will include time and all tag columns of all chunks.
```
alamb, line 142: I wonder if this was (or should it be) called `partition sort key` rather than `super sort key`? I vaguely remember that term in one of the PRs this week

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+║     │       C2.2       │ │      C2.3       │ │      C2.4       │ ║
+║     │Overlaps with C2.1│ │Duplicates within│ │  No Duplicates  │ ║
+║     └──────────────────┘ └─────────────────┘ └─────────────────┘ ║
+╚══════════════════════════════════════════════════════════════════╝
+Figure 1: Two partitions each has four chunks
+```
+
+### Data Deduplication
+
+Figure 2 shows a simplified query plan that IOx builds to read all eight chunks of these two parttions and deduplicate them to return only non-duplicated data. This diagram should be read bottom up as follows:
```
carols10cents, line 40: This query plan-- is it happening in the querier? Or in the ingester/compactor when compacting too?

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
...
+        //         IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     DeduplicateExec: [tag1@1 ASC,time@2 ASC]
+        //         IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     DeduplicateExec: [tag1@1 ASC,time@2 ASC]
+        //         IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        let plan_str = format!("{}", displayable(plan.as_ref()).indent());
+        //println!("{}", plan_str);
```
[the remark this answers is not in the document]
alamb:
```suggestion
        println!("{}", plan_str);
```

docs/dedup_and_sort.md
carols10cents, line 42: The sort keys should come from the partition record in the catalog at this point, right? So should they be called something like P1.K and P2.K?

alamb (approved):
Really nice @NGA-TRAN -- thank you. While reading this I left a bunch of style suggestions (mostly as a way to actively read) -- none are required

docs/dedup_and_sort.md
carols10cents, line 75:
```suggestion
2. All **persisted chunks of the same parttition are sorted on the same key** to avoid resorting before deduplication. So the plan in Figure 2 is only seen sometimes in IOx, instead, you will mostly see the plan in Figure 4 (for data in Figure 3) or in Figure 6 (for data of Figure 5)
```

Another typo with "Fugure 3"

docs/dedup_and_sort.md
carols10cents, line 137: "Every non-persisted chunk sent from Ingester to Querier contain no duplicates" makes it sound like when the querier requests data from the ingester and the ingester has non-persisted data that matches the query, the ingester deduplicates the chunk before sending it to the querier. Am I reading that right? (I forget what the code does). If so, and if the query plan above is done in the querier, why would non-overlapping chunks like C1.3 and C2.3 (that I'm guessing come from the ingester?) need to do dedup in the query plan in Figure 2?

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+```
+
+### Sorted and Unsorted Chunks
+- Persisted chunks are always sorted and contains no duplicates in itself
+- Non-persisted chunks are not sorted and
+  - Every non-persited chunk in the Ingester may contain duplicates
+  - Every non-persisted chunk sent from Ingester to Querier contain no duplicates
+
+# Sort Keys
+Since IOx allows different sets of columns for various chunks of the same table, to keep the sort key consistent for all chunks in the same partition, IOx stores two kinds of sort keys:
```
carols10cents, line 140: Which part of IOx is "IOx stores two kinds of sort keys" referring to here? The catalog? The parquet files? I *think* the Chunk Sort Key is only computed once data is persisted to a Parquet File, right? That is, the ingester doesn't have a chunk sort key for a non-persisted chunk? Or am I misremembering/outdated with what the code is doing?

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+1. **Chunk Sort Key**: this is the sort key of the chunk and only includes tag and time columns of that chunk.
+2. **Super Sort Key**: this is the sort key of all chunks of the same partition. This sort key will include time and all tag columns of all chunks.
+
+### Compute Super Sort Key and Chunk Sort Key
+- For the first chunk of the partition, its Chunk Sort Key will be computed based on the ascending of cardinality of the tag columns. Time column is always last. This Chunk Sort Key will be stored with the chunk. Since there is only one chunk at this time, the Super Sort Key is the same with the Chunk Sort Key and stored in the Partition catalog of the chunk.
```
carols10cents, line 145: I would clarify when this sort key is computed and when the sort key is stored -- it's when the ingester persists the data to a parquet file, right?

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+- For the following chunks, its Chunk Sort Key will be the same as the Super Sort Key but
+  - If the chunk does not include certain tags in the Super Sort Key, its Chunk Sort Key will be the Super Sort Key minus the missing columns.
+  - If the chunk include extra tags that do not exist in the Super Sort Key, those tags will be added to the end but in front of the time column. In this case, the Super Sort Key is changed and must be updated to the Partition catalog of the chunk.
```
carols10cents, line 148: I would clarify that this chunk sort key is computed and the super sort key is potentially updated when the ingester persists the chunk to parquet

docs/dedup_and_sort.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Example**
+
+In the example below, chunks in lower numbers are created before the ones with higher numbers.
```
carols10cents, line 152: I would say "persisted to parquet files" rather than "created" here, if that's accurate

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
...
+
+        // chunk2_5 no overlap, duplicates within
+        let chunk2_5 = Arc::new(
+            TestChunk::new("t")
+                .with_id(4)
+                .with_time_column()
+                .with_timestamp_min_max(28000, 30000)
+                .with_tag_column("tag1")
+                .with_i64_field_column("field_int")
+                .with_partition_id(2) // signal the chunk in partition 1 while grouping overlaps
```
carols10cents, line 3284: Copypasta, I think this should be:

```suggestion
                .with_partition_id(2) // signal the chunk in partition 2 while grouping overlaps
```

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
...
+
+        // chunk2_6 no overlap, duplicates within
+        let chunk2_6 = Arc::new(
+            TestChunk::new("t")
+                .with_id(4)
+                .with_time_column()
+                .with_timestamp_min_max(38000, 40000)
+                .with_tag_column("tag1")
+                .with_i64_field_column("field_int")
+                .with_partition_id(2) // signal the chunk in partition 1 while grouping overlaps
```
carols10cents, line 3297: Same:

```suggestion
                .with_partition_id(2) // signal the chunk in partition 2 while grouping overlaps
```

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
...
+                .with_timestamp_min_max(28000, 220000)
+                .with_tag_column("tag1")
+                .with_i64_field_column("field_int")
+                .with_partition_id(1) // signal the chunk in partition 1 while grouping overlaps
+                .with_may_contain_pk_duplicates(true) // signal having duplicates within this chunk
+                .with_sort_key(sort_key.clone()), // signal the chunk is sorted
+        ) as Arc<dyn QueryChunk>;
+
+        // ------------
+        // Partition 2: Same 4 kinds of chunks on 6 chunks
```
carols10cents, line 3228: What functionality does having 6 chunks exercise that 4 chunks doesn't exercise? That is, I think this comment would be better if it explained why there are 6 chunks?

query/src/provider.rs
```diff
@@ -3167,6 +3167,201 @@ mod test {
...
+        //         IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     DeduplicateExec: [tag1@1 ASC,time@2 ASC]
+        //         IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        //     IOxReadFilterNode: table_name=t, chunks=1 predicate=Predicate
+        let plan_str = format!("{}", displayable(plan.as_ref()).indent());
+        //println!("{}", plan_str);
+
+        //  Verify no SortExec because chunks are not yet sorted on the specified sort_key
```
carols10cents, line 3345: This comment seems to conflict with the comment above that says:

```
        // The plan should look like this. No SortExec at all because
        // all chunks are already sorted on the same requested sort key
```

So I think this comment should say: 

```suggestion
        //  Verify no SortExec because chunks are already sorted on the specified sort_key
```

?

carols10cents (commented):
Thoughts!

docs/dedup_and_sort.md
NGA-TRAN: @lukebond: I have added a line to link to the code that identifies overlapped and duplicates-within chunks and also says the document will explain how it works.

I have added "(see below for how it is computed)" next to K1.1_2

docs/dedup_and_sort.md
NGA-TRAN: @lukebond : yes, added a clarification above

@alamb : yes, C1.3 is same partition as C1.1 and C1.2 but since it does not overlap with any, we only need to sort it on its Chunk Sort Key rather than going to look for the Partition Sort Key. I do not mention the case if the output of the whole scan is sorted on a given key, then yes, that sort key will be use to sort C1.3 instead.

docs/dedup_and_sort.md
NGA-TRAN: I have modify the sentence to make it clear that they may not be sorted and it's okay to leave them like that

docs/dedup_and_sort.md
NGA-TRAN: I have added a sentence to tell when "only seen sometimes in IOx"

docs/dedup_and_sort.md
NGA-TRAN: I have replaced `Super` with `Partition`

docs/dedup_and_sort.md
NGA-TRAN: I have added a sentence saying it is built in Ingester, Compactor and Querier using the same code based pointed out above

docs/dedup_and_sort.md
NGA-TRAN: Right, the sort keys should come from the partition record in the catalog. I have added after K1.1_2 (see below for how it is computed) and add paragraph at the end to explain in the example the value got from partition and then eliminate missing columns

docs/dedup_and_sort.md
NGA-TRAN: > "Every non-persisted chunk sent from Ingester to Querier contain no duplicates" makes it sound like when the querier requests data from the ingester and the ingester has non-persisted data that matches the query, the ingester deduplicates the chunk before sending it to the querier. Am I reading that right? 

That is right

> If so, and if the query plan above is done in the querier, why would non-overlapping chunks like C1.3 and C2.3 (that I'm guessing come from the ingester?) need to do dedup in the query plan in Figure 2?

Data from Ingester to Querier will be in either:
 1. C1.4 (or C2.4) which does not overlap and has no duplicates
 2. C1.1 (or C1.2 or C2.1 or C2.2) which overlaps with some other chunks
 
C1.3 or C2.3 is the case of data in the Ingester that may not overlap but may have duplicates.

Your questions clearly show you understand the system very well @carols10cents 🎉

docs/dedup_and_sort.md
NGA-TRAN: > Which part of IOx is "IOx stores two kinds of sort keys" referring to here? The catalog? The parquet files?

I meant the Catalog but the Chunk Sort Key is stored both in the `Parquet File` catalog table and in the parquet file in S3 itself

> I think the Chunk Sort Key is only computed once data is persisted to a Parquet File, right? 

That is right

> That is, the ingester doesn't have a chunk sort key for a non-persisted chunk? 

That is right

> Or am I misremembering/outdated with what the code is doing?

You are up to date, Carol

docs/dedup_and_sort.md
NGA-TRAN: Done

query/src/provider.rs
NGA-TRAN: Have added comments why 6 chunks

Issue #4457, opened 2022-04-28
flake: e2e test failures


domodwyer:
PR #3577 slightly changes the timing of router2 startup (lengthening it by a very small amount) - [in CI I have observed][1] lots of flaky tests and have collated them below:

```
end_to_end_ng_cases::cli::remote_partition_and_get_from_store
end_to_end_ng_cases::ingester::ingester_flight_api
end_to_end_ng_cases::ingester::ingester_flight_api_namespace_not_found
end_to_end_ng_cases::ingester::ingester_flight_api_table_not_found
end_to_end_ng_cases::metrics::test_metrics
end_to_end_ng_cases::namespace::querier_namespace_client
end_to_end_ng_cases::querier::basic_no_ingester_connection
end_to_end_ng_cases::querier::basic_on_parquet
end_to_end_ng_cases::querier::influxrpc::metadata::capabilities
end_to_end_ng_cases::querier::influxrpc::metadata::measurement_fields
end_to_end_ng_cases::querier::influxrpc::metadata::measurement_names
end_to_end_ng_cases::querier::influxrpc::metadata::measurement_tag_keys
end_to_end_ng_cases::querier::influxrpc::metadata::measurement_tag_values
end_to_end_ng_cases::querier::influxrpc::metadata::offsets
end_to_end_ng_cases::querier::influxrpc::metadata::tag_keys
end_to_end_ng_cases::querier::influxrpc::read_filter::read_filter
end_to_end_ng_cases::querier::multi_ingester::basic_multi_ingesters
end_to_end_ng_cases::querier::table_not_found_on_ingester
end_to_end_ng_cases::router::write_via_grpc
end_to_end_ng_cases::schema::ingester_schema_cli
end_to_end_ng_cases::schema::ingester_schema_client
```

The only constant is `end_to_end_ng_cases::metrics::test_metrics`

[1]: https://app.circleci.com/pipelines/github/influxdata/influxdb_iox?branch=dom%2Fschema-cache-warm&filter=all

domodwyer:
These e2e tests rarely flake locally but seems to always be preceded by:

```
Server command failed: Error in ingester subcommand: Run: Unable to bind to listen for HTTP requests on 127.0.0.1:8205: 2022-04-28T10:35:44.377822Z  WARN executor: DedicatedExecutor dropped without calling shutdown()
error creating server listener: Address already in use (os error 48)
```

[...]

PR #4458, opened 2022-04-28
test: remove hard-coded metric count

domodwyer:
Fixes an e2e failure first noticed in #3577.

I'm not sure if this was intentional - I can't think of a reason why we'd want to assert this one metric appears an exact number of times, though I am happy to change it back to an == if needed 👍 

Relates to https://github.com/influxdata/influxdb_iox/issues/4457

---

* test: remove hard-coded metric count (246af0c3c)

      Prior to this commit, adding any metric to the catalog (and only the catalog)
      would cause the end_to_end_ng_cases::metrics::test_metrics test to fail due to
      asserting an exact number of metrics observed.

      This commit changes the check condition to a more permisive >= rather than ==.

### DIFF
--- a/influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
+++ b/influxdb_iox/tests/end_to_end_ng_cases/metrics.rs
@@ -16,13 +16,13 @@ pub async fn test_metrics() {
             Step::WriteLineProtocol(lp),
             Step::WaitForReadable,
             Step::VerifiedMetrics(Box::new(|_state, metrics| {
-                assert_eq!(
+                assert!(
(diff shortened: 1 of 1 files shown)


carols10cents (approved):
Yeah, I was similarly surprised that [the old test did this](https://github.com/influxdata/influxdb_iox/blob/420c306caa511cfec5919a85d93b49f4b146378f/influxdb_iox/tests/end_to_end_cases/metrics.rs#L40), but figured it was worth a shot if that test was passing 😂 This is much better though :)

alamb:
Thanks -- looks good.

Issue #4395, opened 2022-04-22
Flaky test    end_to_end_ng_cases::querier::multi_ingester::basic_multi_ingesters


alamb:
Test fails https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18376/workflows/25b4b161-0bbb-4d65-9a36-c900548caea0/jobs/153044

But succeeded on a rerun: https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18376/workflows/3e951e32-e802-41c2-bdce-cb769874c718/jobs/153090

Failure was
```
failures:

---- end_to_end_ng_cases::querier::multi_ingester::basic_multi_ingesters stdout ----
thread 'end_to_end_ng_cases::querier::multi_ingester::basic_multi_ingesters' panicked at 'assertion failed: `(left == right)`
  left: `["+------+------+--------------------------------+-----+", "| tag1 | tag2 | time                           | val |", "+------+------+--------------------------------+-----+", "| A    | B    | 1970-01-01T00:00:00.000123456Z | 5   |", "+------+------+--------------------------------+-----+"]`,
 right: `["++", "++"]`: 

expected:

[
    "+------+------+--------------------------------+-----+",
    "| tag1 | tag2 | time                           | val |",
    "+------+------+--------------------------------+-----+",
    "| A    | B    | 1970-01-01T00:00:00.000123456Z | 5   |",
    "+------+------+--------------------------------+-----+",
]
actual:

[
    "++",
    "++",
]

', /home/rust/project/test_helpers_end_to_end_ng/src/steps.rs:177:21
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:182:5
   4: test_helpers_end_to_end_ng::steps::StepTest::run::{{closure}}
             at /home/rust/project/test_helpers_end_to_end_ng/src/steps.rs:177:21
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
[...]

PR #4461, opened 2022-04-28
refactor: start using `with_timeout_panic` in end to end tests

alamb:
Very minor PR on the path to  https://github.com/influxdata/influxdb_iox/issues/4395


test_helpers_end_to_end_ng/src/server_fixture.rs
```diff
@@ -365,9 +366,9 @@ impl TestServer {
                 }
                 interval.tick().await;
             }
-        })
-        .await
-        .expect("Timed out waiting to connect clients");
+        }
+        .with_timeout_panic(Duration::from_secs(10))
```
alamb, line 370: The idea is to start using this very nice decorator from @domodwyer

PR #4465, opened 2022-04-28
chore: add some logging to flaky end to end test

alamb:
I am trying to chase down https://github.com/influxdata/influxdb_iox/issues/4395

The test output appears to show that the test thinks the data is readable when it really isn't. Adding some debug logging so we can see the output if/when the test fails again


lukebond (approved):
will this work without `--nocapture`?

e-dard (approved):
👍

alamb:
Danke

PR #4466, opened 2022-04-28
fix: Possible fix for flaky end to end test `end_to_end_ng_cases::querier::multi_ingester::basic_multi_ingesters`

alamb:
Draft while testing

# Rationale
Closes https://github.com/influxdata/influxdb_iox/issues/4395 (I hope)

I discovered in https://github.com/influxdata/influxdb_iox/pull/4387#discussion_r860997349 that the configuration options that claim to make a test fixture not persist quicky was a sham (all configurations ended up persisting quickly)

The failure mode on #4395 is that there is no data in the ingester even though we had previously observed that all the data was readable. This could have happened if the ingester *persisted* the data between when the test checked for readable and when it ran the query. 

# Changes
1. Configures the ingester to (practically speaking) not persist its data

# Testing:
I ran a sophisticated script (below) to run `cargo test --test end_to_end_ng -- multi_ingester` 100 times.

* Before this change: TODO
* with this change: IN PROGRESS

Sophisticated script:
```bash
set -e

for i in `seq 1 100`; do
    echo "iteration $i"
    TEST_INTEGRATION=1 TEST_INFLUXDB_IOX_CATALOG_DSN=postgresql://localhost:5432/alamb RUST_BACKTRACE=1 CARGO_TARGET_DIR=/Users/alamb/Software/iox-target cargo test --test end_to_end_ng -- multi_ingester
done
```

### DIFF
--- a/influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
+++ b/influxdb_iox/tests/end_to_end_ng_cases/querier/multi_ingester.rs
@@ -21,9 +21,13 @@ async fn basic_multi_ingesters() {
     let router2_config =
         TestConfig::new_router2(&database_url).with_new_write_buffer_kafka_partitions(2);
 
-    // ingester gets partition 0
-    let ingester_config = TestConfig::new_ingester(&router2_config).with_kafka_partition(0);
-    let ingester2_config = TestConfig::new_ingester(&router2_config).with_kafka_partition(1);
+    let ingester_config = TestConfig::new_ingester(&router2_config)
+        .with_kafka_partition(0)
+        .with_no_persist();
+
+    let ingester2_config = TestConfig::new_ingester(&router2_config)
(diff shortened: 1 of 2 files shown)


alamb:
My plan is to add some more logging (via https://github.com/influxdata/influxdb_iox/pull/4465) and then see if I can gather any evidence that this would fix the root cause

alamb:
> 
@alamb I've been seeing this test fail a bunch in CI lately. Is there anything I can do to help move this fix forward?


@carols10cents  I can't think of anything unless you want to look into finding some way to make the failure reproduceable -- I'll put this on my list of things to review

alamb:
related: https://github.com/influxdata/influxdb_iox/pull/4580

alamb:
I don't think this is the issue -- see https://github.com/influxdata/influxdb_iox/pull/4609#issuecomment-1128146435 for something that looks much more promising

PR #4472, opened 2022-04-29
feat: expose catalog timeouts via CLI/env

crepererum:
This is useful for local instances that run against a prod system,
because port forwarding can lead to long connection delays.


clap_blocks/src/catalog_dsn.rs
```diff
@@ -36,17 +62,44 @@ pub struct CatalogDsnConfig {
...
     /// Schema name for PostgreSQL-based catalogs.
     #[clap(
         long = "--catalog-postgres-schema-name",
         env = "INFLUXDB_IOX_CATALOG_POSTGRES_SCHEMA_NAME",
-        default_value = iox_catalog::postgres::SCHEMA_NAME,
+        default_value = PostgresConnectionOptions::DEFAULT_SCHEMA_NAME,
     )]
     pub postgres_schema_name: String,
+
+    /// Set the amount of time to attempt connecting to the database.
```
alamb, line 77: Adding an example value `30s` might make it easier for users to understand they can add flexible times to these parameters (rather than a numeric value in seconds)

clap_blocks/src/catalog_dsn.rs
```diff
@@ -36,17 +62,44 @@ pub struct CatalogDsnConfig {
...
+    #[clap(
+        long = "--catalog-connection-timeout",
+        env = "INFLUXDB_IOX_CATALOG_CONNECT_TIMEOUT",
+        default_value = default_connection_timeout(),
+        parse(try_from_str = humantime::parse_duration),
+    )]
+    pub connect_timeout: Duration,
```
alamb, line 84:
```suggestion
    pub connection_timeout: Duration,
```

Maybe to make the struct field more consistent with the parameter name?

clap_blocks/src/catalog_dsn.rs
```diff
@@ -36,17 +62,44 @@ pub struct CatalogDsnConfig {
...
+    /// Set a maximum idle duration for individual connections.
+    #[clap(
+        long = "--catalog-idle-timeout",
+        env = "INFLUXDB_IOX_CATALOG_IDLE_TIMEOUT",
+        default_value = default_idle_timeout(),
+        parse(try_from_str = humantime::parse_duration),
+    )]
+    pub idle_timeout: Duration,
+
+    /// The file pointed to by a `dsn-file://` DSN is polled for change every `HOTSWAP_POLL_INTERVAL`.
```
alamb, line 95: I don't understand what this parameter does / means or when I might set it. Can you possibly add a link to documentation / pointer to more help?

iox_catalog/src/postgres.rs
```diff
@@ -23,15 +23,65 @@ use sqlx::{migrate::Migrator, postgres::PgPoolOptions, Acquire, Executor, Postgr
...
+    /// Maximum number of concurrent connections.
+    pub max_conns: u32,
+
+    /// Set the amount of time to attempt connecting to the database.
+    pub connect_timeout: Duration,
+
+    /// Set a maximum idle duration for individual connections.
+    pub idle_timeout: Duration,
+
+    /// The file pointed to by a `dsn-file://` DSN is polled for change every `HOTSWAP_POLL_INTERVAL`.
```
alamb, line 51: same here -- I don't know what this means so adding some documentation would be great

alamb (approved):
LGTM -- onward to production

clap_blocks/src/catalog_dsn.rs
crepererum: Oh, good catch. It should be `connect` timeout for both since it esp. concerns the "connect" action, not the entire connection lifecycle.

clap_blocks/src/catalog_dsn.rs
crepererum: There is a default value which should explain that, no?

```text
        --catalog-connection-timeout <CONNECT_TIMEOUT>
            Set the amount of time to attempt connecting to the database

            [env: INFLUXDB_IOX_CATALOG_CONNECT_TIMEOUT=]
            [default: 2s]
```

clap_blocks/src/catalog_dsn.rs
crepererum: I rewrote the docs, should be a bit clearer now.

alamb: I reviewed them -- much clearer now. 👍

PR #4480, opened 2022-04-29
feat: Export metrics about acquire wait time

mkmik:
This is draft.

This PR adds a crate called `sqlx-acquire-meter` which provides a `AcquireDurationMeasuringPool` pool that receives as paramters a sqlx pool and an `AcquireDurationMeter` "callback trait", that will be called every time a connection is acquired from the pool and will be given the time it took to wait on acquisition.

TODO: instantiate this `AcquireDurationMeasuringPool` in our `iox_catalog` crate and implement an `AcquireDurationMeter` that will emit a histogram with our metric framework.

### DIFF
--- a/sqlx-acquire-meter/src/lib.rs
+++ b/sqlx-acquire-meter/src/lib.rs
@@ -1,0 +1,318 @@
+use async_stream::try_stream;
+use either::Either;
+use futures::future::BoxFuture;
+use futures::prelude::stream::BoxStream;
+use futures_util::{pin_mut, TryStreamExt};
+use iox_time::{SystemProvider, TimeProvider};
+use sqlx::database::HasStatement;
+use sqlx::pool::PoolConnection;
+use sqlx::{Acquire, Database, Describe, Error, Execute, Executor, Pool, Transaction};
+use std::fmt::Debug;
+use std::time::Duration;
+
+pub trait AcquireDurationMeter: Debug + Send + Sync {
+    fn record_acquire_duration(&self, d: Duration);
+}
+
+#[derive(Debug)]
+pub struct AcquireDurationMeasuringPool<DB, M, P = SystemProvider>
+where
+    DB: Database,
+    M: AcquireDurationMeter,
+{
+    pool: Pool<DB>,
+    meter: M,
+    time_provider: P,
+}
+
+impl<DB, M> AcquireDurationMeasuringPool<DB, M>
+where
+    DB: Database,
+    M: AcquireDurationMeter,
+{
+    pub fn new(pool: Pool<DB>, meter: M) -> Self {
+        Self {
+            pool,
+            meter,
+            time_provider: Default::default(),
+        }
+    }
+
+    pub fn time_provider<P: TimeProvider>(
+        self,
+        time_provider: P,
+    ) -> AcquireDurationMeasuringPool<DB, M, P> {
+        AcquireDurationMeasuringPool {
+            pool: self.pool,
+            meter: self.meter,
+            time_provider,
+        }
+    }
+}
+
+impl<'a, DB, M, P> Acquire<'a> for &'a AcquireDurationMeasuringPool<DB, M, P>
+where
+    DB: Database,
+    M: AcquireDurationMeter,
+    P: TimeProvider,
+{
+    type Database = DB;
+
+    type Connection = PoolConnection<DB>;
+
+    fn acquire(self) -> BoxFuture<'a, Result<Self::Connection, Error>> {
+        Box::pin(async move {
+            let start = self.time_provider.now();
+            let conn = self.pool.acquire().await?;
+            // Avoid exploding if time goes backwards - simply drop the measurement
+            // if it happens.
+            if let Some(delta) = self.time_provider.now().checked_duration_since(start) {
+                self.meter.record_acquire_duration(delta);
+            }
+            Ok(conn)
+        })
+    }
+
+    fn begin(self) -> BoxFuture<'a, Result<Transaction<'a, DB>, Error>> {
+        // We can't intercept "acquire" independently of the actual begin transaction overhead.
+        // That's not a big deal since pool.begin() is just a practical helper around `acquire()` + `begin()`.
+        //
+        // The sqlx internals make it impossible. It's due to the way transactions can be created:
+        // You can create a transaction either from a pool or from a connection.
+        // If you create the transaction from a connection (by calling `conn.begin()`),
+        // the transaction borrows the connection and thus the connection needs outlive the transaction.
+        // This means that when the transaction is dropped there is no way to "drop" the connection.
+        //
+        // Pool connections OTOH rely on the ability to intercept the connection drop in order to put the
+        // connection back into the pool. For that to work, `Transaction` needs to also implement `Drop` and
+        // drop the pool connection when dropped (i.e. it needs to "own" the connection).
+        //
+        // `Transaction` handles those two cases (owning and reference) by using an internal "MaybePoolConnection"
+        // enum which we can't use outside sqlx.
+        unimplemented!("Use `pool.acquire()` to get a connection and then call `begin()` on it")
+    }
+}
+
+impl<'p, DB, M> Executor<'p> for &'p AcquireDurationMeasuringPool<DB, M>
+where
+    DB: Database,
+    M: AcquireDurationMeter,
+    for<'c> &'c mut DB::Connection: Executor<'c, Database = DB>,
+{
+    type Database = DB;
+
+    fn fetch_many<'e, 'q: 'e, E: 'q>(
+        self,
+        query: E,
+    ) -> BoxStream<'e, Result<Either<DB::QueryResult, DB::Row>, Error>>
+    where
+        E: Execute<'q, Self::Database>,
+        'p: 'e,
+    {
+        Box::pin(try_stream! {
+          let mut conn = self.pool.acquire().await?;
+          let s = conn.fetch_many(query);
+          pin_mut!(s);
+          while let Some(v) = s.try_next().await? {
+            yield v;
+          }
+        })
+    }
+
+    fn fetch_optional<'e, 'q: 'e, E: 'q>(
+        self,
+        query: E,
+    ) -> BoxFuture<'e, Result<Option<DB::Row>, Error>>
+    where
+        E: Execute<'q, Self::Database>,
+        'p: 'e,
+    {
+        Box::pin(async move { self.pool.acquire().await?.fetch_optional(query).await })
+    }
+
+    fn prepare_with<'e, 'q: 'e>(
+        self,
+        sql: &'q str,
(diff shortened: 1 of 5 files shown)


sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+{
+    type Database = DB;
+
+    type Connection = PoolConnection<DB>;
+
+    fn acquire(self) -> BoxFuture<'a, Result<Self::Connection, Error>> {
+        Box::pin(async move {
+            let start = tokio::time::Instant::now();
+            let conn = self.pool.acquire().await?;
+            self.meter.record_acquire_duration(start.elapsed());
```
alamb, line 49: 👍

sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+where
+    DB: Database,
+    M: AcquireDurationMeter,
+{
+    pub fn new(pool: Pool<DB>, meter: M) -> Self {
+        Self { pool, meter }
+    }
+}
+
+impl<'a, DB, M> Acquire<'a> for &'a AcquireDurationMeasuringPool<DB, M>
```
alamb, line 36: https://docs.rs/sqlx/0.5.13/sqlx/trait.Acquire.html may be helpful

for anyone else reviewing this code

sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+        // the transaction borrows the connection and thus the connection needs outlive the transaction.
+        // This means that when the transaction is dropped there is no way to "drop" the connection.
+        //
+        // Pool connections OTOH rely on the ability to intercept the connection drop in order to put the
+        // connection back into the pool. For that to work, `Transaction` needs to also implement `Drop` and
+        // drop the pool connection when dropped (i.e. it needs to "own" the connection).
+        //
+        // `Transaction` handles those two cases (owning and reference) by using an internal "MaybePoolConnection"
+        // enum which we can't use outside sqlx.
+        unimplemented!("Use `pool.acquire()` to get a connection and then call `begin()` on it")
```
alamb, line 70: I was thinking it might make sense to return an error here (well really more a future that returns an error). Perhaps something like

```rust
async {
  Err(Error("Unsupported begin transaction in `AcquireDurationMeter`"))
}.boxed()
```


But then again I can see the benefit of failing quickly with a precise stacktrace via `unimplemented!`

sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+        Box::pin(async move {
+            let start = tokio::time::Instant::now();
+            let conn = self.pool.acquire().await?;
+            self.meter.record_acquire_duration(start.elapsed());
+            Ok(conn)
+        })
+    }
+
+    fn begin(self) -> BoxFuture<'a, Result<Transaction<'a, DB>, Error>> {
+        // We can't intercept "acquire" independently of the actual begin transaction overhead.
```
alamb, line 55: I wonder if using an `Arc<dyn Database>` might allow for full support with less lifetime drama?

sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+    use std::sync::{Arc, Mutex};
+    use std::time::Duration;
+
+    use super::*;
+    use rand::{distributions::Alphanumeric, Rng};
+    use sqlx::{postgres::PgPoolOptions, Postgres};
+
+    // Helper macro to skip tests if TEST_INTEGRATION and TEST_INFLUXDB_IOX_CATALOG_DSN environment variables
+    // are not set.
+    macro_rules! maybe_skip_integration {
```
alamb, line 154: It would be sweet to move this macro to `test_helpers` or something so it isn't replicated all over the codebase

sqlx-acquire-meter/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+            })
+            .collect();
+        futures::future::join_all(tasks).await;
+
+        let mut durations = durations.lock().expect("not poisoned");
+        durations.sort();
+        for i in 0..3 {
+            let delta = durations[i + 1] - durations[i];
+            // we should at least measure the pg_sleep(0.1).
+            // (measuring actual sleeps is brittle, but using fake time is hard here so let's hope this test doesn't cause headache)
```
alamb, line 287: The way much of the rest of the codebase handles this kind of thing is to measure time via `iox_time`
https://github.com/influxdata/influxdb_iox/blob/main/iox_time/src/lib.rs

So rather than call 
```rust
            let start = tokio::time::Instant::now();
```

The code would call

```
            let start = time_provider.now()
```

And you can manipulate the time provider however you want

alamb (commented):
Looks like a neat idea -- I left some thoughts; hopefully they are helpful

sqlx-acquire-meter/src/lib.rs
mkmik: Ideally I'd like to fail statically but I don't think that's possible.

sqlx-acquire-meter/src/lib.rs
mkmik: That might be useful indeed.

Probably I should not do the sleeping with pg_sleep but do the sleeping entirely with fake time.

sqlx-acquire-meter/src/lib.rs
mkmik: also, the Error type of the Acquire trait cannot be overridden and none of the Error enum variables appears to be applicable for this

sqlx-acquire-meter/src/lib.rs
mkmik: The lifetime of the database is not the problem here.

There is a non trivial amount of logic in the [`Pool`](https://docs.rs/sqlx/0.5.13/sqlx/struct.Pool.html) implementation, that deals with keeping a list of active connections, closing the idle ones etc etc. We thus have strong incentives to layer something on top of the `Pool` struct implementation rather than reimplementing our own.

The `Pool` struct implementation of the `Acquire` yields `PoolConnection` values (which implement the `Connection` trait). The `PoolConnection` struct is parameterized by a `Database` type, so we could parameterize it by an `Arc<dyn Database`, but as I'll show below that wouldn't help.

The `PoolConnection` is a lazy wrapper around a `Connection`. It points to a connection through a private `live: Option<Live<DB>>` field. `Live` private struct is:

```
pub(super) struct Live<DB: Database> {
    pub(super) raw: DB::Connection,
    pub(super) created: Instant,
}
```

so as you can see, even if we pass `Arc<impl Database>` as `DB`, we'd still end up using whatever associated type our underlying postgres database`DB::Connection`. The lifetime of the connection will be bound by the lifetime of the database and that's fine.

The tricky part is the `Acquire` trait implementation of the `Pool` structure:

```
    fn begin(self) -> BoxFuture<'static, Result<Transaction<'a, DB>, Error>> {
        let conn = self.acquire();

        Box::pin(async move {
            Transaction::begin(MaybePoolConnection::PoolConnection(conn.await?)).await
        })
    }
```

We cannot call `Transaction::begin` because it's a crate private method:

```
impl<'c, DB> Transaction<'c, DB>
where
    DB: Database,
{
    pub(crate) fn begin(
        conn: impl Into<MaybePoolConnection<'c, DB>>,
    ) -> BoxFuture<'c, Result<Self, Error>> {
        let mut conn = conn.into();
...
```

So the only thing we can do from our Pool wrapper's `begin` method is to grab a connection and call `.begin` on the connection.

However the `.begin` call on the connection takes the connection by reference and so never calls `drop` on the connection when the transaction is done.

The `drop` method is called only when the transaction is created from the `Pool` implementation, and that happens because there is a special private enum called `MaybePoolConnection` that either takes a connection by reference or owns it and this `MaybePoolConnection` is treated specially by the Transaction trait (as shown above).

---

Some refactoring in the upstream `sqlx` crate would be necessary to allow this. We could do this, but probably it's nit the highest priority right now

alamb: > We could do this, but probably it's nit the highest priority right now

Agree entirely -- thank you for the detailed analysis. It makes sense to me

domodwyer:
I took a look at this but I'm not sure it gives us the properties we were looking for? Perhaps it answers a different question that what I was assuming.

Originally I wanted to measure the duration of time taken to _get a connection from the connection pool_ as a way of measuring the **saturation** of the request pool (USE metric style). A high pool wait duration would indicate we are attempting to handle more requests than the pool can satisfy, and we should look at increasing capacity/speeding up queries/reducing workload/etc to bring the pool wait time back down.

I might be mistaken, but I thought hooking the acquire method like this would conflate the following scenarios:

* **Connection is available in pool:** pool wait time (negligible), connection test time
* **Connection is not available in pool, pool below capacity:** pool wait time (negligible), connection dial time, handshake time, connection test time
* **Connection is not available in pool, pool at capacity:** pool wait time (long), connection test time

That is because the [`Pool<DB>::acquire()` impl][pool] calls into the `SharedPool<DB>::acquire()` where it:

* [Waits for a pool permit](https://github.com/launchbadge/sqlx/blob/fa5c436918664de112677519d73cf6939c938cb0/sqlx-core/src/pool/inner.rs#L172)
* [Liveness checks the connection](https://github.com/launchbadge/sqlx/blob/fa5c436918664de112677519d73cf6939c938cb0/sqlx-core/src/pool/inner.rs#L182)
* [Potentially expands the pool](https://github.com/launchbadge/sqlx/blob/fa5c436918664de112677519d73cf6939c938cb0/sqlx-core/src/pool/inner.rs#L191)
* And finally, [dials a whole new connection, handshakes & liveness checks it](https://github.com/launchbadge/sqlx/blob/fa5c436918664de112677519d73cf6939c938cb0/sqlx-core/src/pool/inner.rs#L201)

Hooking the `acquire()` method answers a subtly different version of question of "how long do we wait to get a connection" which is a less reliable indicator of the pool saturation because of the large difference of latency between these different paths. A pool can return an existing connection, test it, decide it is "bad" and dial a new one, and repeat those steps multiple times, bounded only by the connection timeout - it doesn't measure the saturation of the pool as it could be the only request the pool is handling.

---

I actually have a PR ready to go that I knocked up on the flight back that measures just the pool wait time - I'll open it up shortly. I like the callback idea so we can capture the distribution in latency, but my PR currently only measures the total time spent in waiting on the pool per pool - would that work do you think? I realise we'll miss the outliers but we could derive the wait/second or average wait / connection, or perhaps simply "it's non-zero" would be a good enough indicator we need to do something about it.

[pool]:https://github.com/launchbadge/sqlx/blob/fa5c436918664de112677519d73cf6939c938cb0/sqlx-core/src/acquire.rs#L88

alamb:
What is the status of this PR? Shall we close it?

alamb:
This PR has been open for 9 months so closing it for now -- please reopen if it is ready to merge

PR #4486, opened 2022-04-29
feat: Remove outdated management APIs / CLI commands

### LINKED ISSUE #4474: Remove outdated management APIs / CLI commands
Depends on https://github.com/influxdata/influxdb_iox/issues/4473

The idea is to remove the entry points for OG code in the `influxdb_iox` CLI so that we can remove the implementation code. However, we still need the NG entry points

# To Move
Here are the commands that we should move elsewhere

- [ ] `influxdb_iox database write` -->  `influxb_iox write` (grpc write)
- [ ] `influxdb_iox database query` -->  `influxb_iox query` (sql query)
- [ ] `influxdb_iox schema` -->  `influxb_iox debug schema` (schema listing)

# To Delete
Here are the commands that we should remove

- [ ] `database` (other than `write`, `query`, see above) -- there are a lot of subcommands here
- [ ] `router`  Router-related commands
- [ ] `run router` and `run database` mode
- [ ] `schema` (move, see above)
- [ ] `server`  IOx server configuration commands

# To Keep
Here are the commands that we should keep

- [ ] `catalog` Various commands for catalog manipulation
- [ ] `debug`  Interrogate internal database data
- [ ] `help` Print this message or the help of the given subcommand(s)
- [ ] `operation` Manage long-running IOx operations
- [ ] `remote` Commands to run against remote IOx APIs
- [ ] `run` Run the InfluxDB IOx server
- [ ] `sql`  Start IOx interactive SQL REPL loop
- [ ] `storage` Initiate a read request to the gRPC storage service

shepmaster:
Closes #4474


alamb:
@crepererum  -- this PR's CI test is failing due to `perf.py` not being able to start IOx. Perhaps this is because perf.py is using OG?

Do we know how important it is to keep perf.py operating?

https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/18737/workflows/d9730357-a749-470c-907f-2fe8663ae927/jobs/157093

```shell
Redpanda service is ready
waiting for iox_router to become ready
service iox_router stopped unexpectedly, check logs/iox_router.log
removing containers: ioxperf-redpanda
removing networks: ioxperf
cat: perf/logs/test.log: No such file or directory

```

alamb:
>  Just a reminder that this PR is blocked on an admin making the test_perf CI check not required :)

@carols10cents I believe @tkyocum  has made the change 👍 🏅

Cargo.toml
```diff
@@ -37,8 +37,6 @@ members = [
     "ioxd_compactor",
     "ioxd_ingester",
     "ioxd_querier",
-    "ioxd_database",
```
alamb, line 40: 👍

influxdb_iox/src/commands/query.rs
```diff
@@ -0,0 +1,83 @@
...
+    // ReadingFile {
+    //     file_name: PathBuf,
+    //     source: std::io::Error,
+    // },
+    #[error("Error formatting: {0}")]
+    FormattingError(#[from] influxdb_iox_client::format::Error),
+
+    #[error("Error querying: {0}")]
+    Query(#[from] influxdb_iox_client::flight::Error),
+    // #[error("Error in chunk subcommand: {0}")]
```
alamb, line 22: I wonder if we should remove the commented out code here

influxdb_iox/src/main.rs
```diff
@@ -170,6 +164,12 @@ enum Command {
 
     /// Initiate a read request to the gRPC storage service.
     Storage(commands::storage::Config),
+
+    /// Write data into the specified database
+    Write(commands::write::Config),
+
+    /// Query the data with SQL
```
alamb, line 171: 👍

ioxd_database/src/rpc/delete.rs
```diff
@@ -1,54 +0,0 @@
-use std::sync::Arc;
```
alamb, line 1: Hmm we may need to ressurect the delete service  -- I'll file a ticket to track that

alamb: https://github.com/influxdata/influxdb_iox/issues/4527

alamb (approved):
So this PR I believe will definitely break the IDPE fluxtest. However, I think this is ok because:

1. This iox/fluxtest is not part of main IDPE CI (it runs as a nightly job)
2. The automated test was broken for unrelated reasons the last time I looked at it anyways (see https://github.com/influxdata/conductor/issues/933#issuecomment-1113405923)
3. The work to resurrect the test is tracked in https://github.com/influxdata/conductor/issues/933

onward!

influxdb_iox/src/commands/query.rs
carols10cents: Good call and done: https://github.com/influxdata/influxdb_iox/pull/4528

Issue #4489, opened 2022-04-29
Remove OG generated_types proto definitions


carols10cents:
These protos are OG-only:

 generated_types/protos/influxdata/iox/deployment/v1/service.proto
 generated_types/protos/influxdata/iox/management/v1/chunk.proto
 generated_types/protos/influxdata/iox/management/v1/database_rules.proto
 generated_types/protos/influxdata/iox/management/v1/jobs.proto
 generated_types/protos/influxdata/iox/management/v1/partition.proto
 generated_types/protos/influxdata/iox/management/v1/partition_template.proto
 generated_types/protos/influxdata/iox/management/v1/server_config.proto
 generated_types/protos/influxdata/iox/management/v1/service.proto
 generated_types/protos/influxdata/iox/preserved_catalog/v1/catalog.proto
 generated_types/protos/influxdata/iox/preserved_catalog/v1/parquet_metadata.proto
 generated_types/protos/influxdata/iox/remote/v1/remote.proto
 generated_types/protos/influxdata/iox/remote/v1/service.proto
 generated_types/protos/influxdata/iox/router/v1/router.proto
 generated_types/protos/influxdata/iox/router/v1/service.proto
 generated_types/protos/influxdata/iox/router/v1/shard.proto

These crates use them, so this is blocked on removing these:

- router
- parquet_catalog
- server

PR #4490, opened 2022-04-29
fix: Remove OG gRPC client code and APIs

### LINKED ISSUE #4475: Remove OG gRPC APIs and clients

carols10cents:
Closes #4475.

This doesn't remove the proto definitions yet because all the uses of them need to be removed first. I opened https://github.com/influxdata/influxdb_iox/issues/4489 for that.


influxdb_iox/src/commands/database.rs
```diff
@@ -211,158 +195,15 @@ struct Restart {
...
-    /// Create a new database
-    Create(Create),
-
-    /// Get list of databases
-    List(List),
-
-    /// Return configuration of specific database
-    Get(Get),
-
     /// Write data into the specified database
```
alamb, line 198: 👍

alamb (approved):
Stellar 🌟

PR #4491, opened 2022-04-29
feat: shard delete ops with no table predicate

### LINKED ISSUE #3629: Shard delete DML ops with no specified table across all sequencers

domodwyer:
This PR adds special-case logic to delete op handling in router2, sharding deletes that do not specify a table name across all shards/sequencers/kafka topics.

This work completes the delete op support in router2.

Closes #3629.

---

* feat: shard delete ops with no table predicate (9a876c557)

      Changes the JumpHash sharder and modifies the ShardedWriteBuffer's 
      DmlHandler::delete() impl in order to enqueue delete ops across all shards if
      no table name is specified.

      If a table name is specified, it is sharded as before: a delete for a given
      table & namespace always maps to the same shard as a write to the same table &
      namespace.

### DIFF
--- a/router2/src/dml_handlers/sharded_write_buffer.rs
+++ b/router2/src/dml_handlers/sharded_write_buffer.rs
@@ -74,7 +74,7 @@ impl<S> ShardedWriteBuffer<S> {
 impl<S> DmlHandler for ShardedWriteBuffer<S>
 where
     S: Sharder<MutableBatch, Item = Arc<Sequencer>>
-        + Sharder<DeletePredicate, Item = Arc<Sequencer>>,
+        + Sharder<DeletePredicate, Item = Vec<Arc<Sequencer>>>,
 {
     type WriteError = ShardError;
     type DeleteError = ShardError;
@@ -98,7 +98,7 @@ where
         // per shard to maximise the size of each write, and therefore increase
         // the effectiveness of compression of ops in the write buffer.
         for (table, batch) in writes.into_iter() {
-            let sequencer = Arc::clone(self.sharder.shard(&table, namespace, &batch));
+            let sequencer = self.sharder.shard(&table, namespace, &batch);
 
             let existing = collated
                 .entry(sequencer)
@@ -135,9 +135,7 @@ where
         span_ctx: Option<SpanContext>,
     ) -> Result<(), ShardError> {
         let predicate = predicate.clone();
-        let sequencer = self.sharder.shard(table_name, namespace, &predicate);
-
-        trace!(sequencer_id=%sequencer.id(), %table_name, %namespace, "routing delete to shard");
+        let sequencers = self.sharder.shard(table_name, namespace, &predicate);
 
         let dml = DmlDelete::new(
             namespace,
@@ -146,13 +144,14 @@ where
             DmlMeta::unsequenced(span_ctx),
         );
 
-        sequencer
-            .enqueue(DmlOperation::from(dml))
-            .await
-            .map_err(|e| ShardError::WriteBufferErrors {
-                successes: 0,
-                errs: vec![e],
-            })?;
+        let iter = sequencers.into_iter().map(|s| {
+            trace!(sequencer_id=%s.id(), %table_name, %namespace, "routing delete to shard");
+
+            (s, DmlOperation::from(dml.clone()))
+        });
+
+        // TODO: return sequencer metadata
+        parallel_enqueue(iter).await?;
 
         Ok(())
     }
@@ -200,7 +199,7 @@ mod tests {
     use super::*;
     use crate::{
         dml_handlers::DmlHandler,
-        sharder::mock::{MockSharder, MockSharderCall},
+        sharder::mock::{MockSharder, MockSharderCall, MockSharderPayload},
     };
     use assert_matches::assert_matches;
     use data_types2::TimestampRange;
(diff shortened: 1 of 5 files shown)


router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -146,13 +144,14 @@ where
...
-                successes: 0,
-                errs: vec![e],
-            })?;
+        let iter = sequencers.into_iter().map(|s| {
+            trace!(sequencer_id=%s.id(), %table_name, %namespace, "routing delete to shard");
+
+            (s, DmlOperation::from(dml.clone()))
+        });
+
+        // TODO: return sequencer metadata
```
alamb, line 153: Is this TODO still relevant? If so is it worth a ticket (is it so a client can track `readability` if a delete across sequencers)?

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -464,4 +463,187 @@ mod tests {
...
+
+        let w = ShardedWriteBuffer::new(sharder);
+
+        // Call the ShardedWriteBuffer and drive the test
+        let ns = DatabaseName::new("namespace").unwrap();
+        w.delete(&ns, TABLE, &predicate, None)
+            .await
+            .expect("delete failed");
+
+        // The write buffer for shard 1 should observe 1 write containing 3 rows.
```
alamb, line 584:
```suggestion
        // The write buffer for shard 1 should observe the delete
```

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -464,4 +463,187 @@ mod tests {
...
+        let mut got = write_buffer1_state.get_messages(shard1.id() as _);
+        assert_eq!(got.len(), 1);
+        let got = got
+            .pop()
+            .unwrap()
+            .expect("write should have been successful");
+        assert_matches!(got, DmlOperation::Delete(_));
+
+        // The second shard should observe 1 write containing 1 row.
```
alamb, line 593:
```suggestion
        // The second shard should observe the delete as well
```

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -464,4 +463,187 @@ mod tests {
...
+        // Configure the first shard to write to one write buffer
+        let write_buffer1 = init_write_buffer(1);
+        let write_buffer1_state = write_buffer1.state();
+        let shard1 = Arc::new(Sequencer::new(
+            0,
+            Arc::new(write_buffer1),
+            &Default::default(),
+        ));
+
+        // Configure the second shard to write to a write buffer that always fails
```
alamb, line 621: 💯  for testing this case

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -464,4 +463,187 @@ mod tests {
...
+        let err = w
+            .delete(&ns, TABLE, &predicate, None)
+            .await
+            .expect_err("delete should fail");
+        assert_matches!(err, ShardError::WriteBufferErrors{successes, errs} => {
+            assert_eq!(errs.len(), 1);
+            assert_eq!(successes, 1);
+        });
+
+        // The write buffer for shard 1 should observe 1 write containing 3 rows.
```
alamb, line 645:
```suggestion
        // The write buffer for shard 1 will still observer the delete.
```

router2/src/dml_handlers/sharded_write_buffer.rs
```diff
@@ -464,4 +463,187 @@ mod tests {
...
+        let shard1 = Arc::new(Sequencer::new(
+            0,
+            Arc::new(write_buffer1),
+            &Default::default(),
+        ));
+
+        // Configure the second shard to write to a write buffer that always fails
+        let write_buffer2 = init_write_buffer(1);
+        // Non-existant sequencer ID to trigger an error.
+        let shard2 = Arc::new(Sequencer::new(
```
alamb, line 624: It is not 100% clear to me if the Delete gets to all shards that don't error, if the handler stops after the first error

I think this test would succeed in either case but if you swapped the order (writing to shard1 fails but shard2 succeeds) the  intended behavior may be more clear.

router2/src/sharder/jumphash.rs
```diff
@@ -110,24 +117,62 @@ struct HashKey<'a> {
...
+{
+    type Item = Vec<Arc<T>>;
+
+    fn shard(
+        &self,
+        table: &str,
+        namespace: &DatabaseName<'_>,
+        _payload: &DeletePredicate,
+    ) -> Self::Item {
+        // A delete
```
alamb, line 161: Not sure what this comment is meant to convey

router2/src/sharder/jumphash.rs
```diff
@@ -241,4 +334,46 @@ mod tests {
         // of the total 10M values
         assert!(max - min < 5000, "min: {}, max: {}", min, max);
     }
+
+    #[test]
+    fn test_delete_with_table() {
+        let namespace = DatabaseName::try_from("bananas").unwrap();
```
alamb, line 340: Now all I can think of when reading your tests is https://www.bostonmagazine.com/news/2022/04/12/mit-banana-lounge/

🍌

router2/src/sharder/mock.rs
```diff
@@ -78,45 +74,41 @@ where
...
         payload: &MutableBatch,
-    ) -> &Self::Item {
+    ) -> Self::Item {
         let mut guard = self.0.lock();
         guard.record_call(MockSharderCall {
             table_name: table.to_string(),
             namespace: namespace.to_string(),
             payload: MockSharderPayload::MutableBatch(payload.clone()),
         });
-        Box::leak(Box::new(
```
alamb, line 88: 👍

alamb (approved):
Looks good to me -- I had one comment about the test that might be worth considering, but all in all 🏅

router2/src/dml_handlers/sharded_write_buffer.rs
domodwyer: The ordering of which shard is wrote to first is non-deterministic.

An attempt to write to all shards will always be made, even if the first fails, as the `parallel_enqueue` [collects the results](https://github.com/influxdata/influxdb_iox/blob/2a15b525a90c2b5d214b16b58ec6489c4e05f521/router/src/dml_handlers/sharded_write_buffer.rs#L180) before [evaluating them](https://github.com/influxdata/influxdb_iox/blob/2a15b525a90c2b5d214b16b58ec6489c4e05f521/router/src/dml_handlers/sharded_write_buffer.rs#L183-L187).

Truthfully the semantics of this partial failure are not well defined for the user and therefore not all combinations of ordering are exhaustively tested - the intent is (like writes) they "just retry" if it is not successful to converge the state of the system.

router2/src/sharder/jumphash.rs
domodwyer: This was at the "awake for 21 hours" mark, so your guess is as good as mine 🔴 👁️ ✈️

router2/src/sharder/jumphash.rs
domodwyer: This is fantastic! 🍌

I demand our own banana room.

router2/src/sharder/trait.rs
```diff
@@ -19,5 +19,5 @@ pub trait Sharder<P>: Debug + Send + Sync {
     type Item: Debug + Send + Sync;
 
     /// Map the specified `payload` to a shard.
-    fn shard(&self, table: &str, namespace: &DatabaseName<'_>, payload: &P) -> &Self::Item;
+    fn shard(&self, table: &str, namespace: &DatabaseName<'_>, payload: &P) -> Self::Item;
```
domodwyer, line 22: Commmmoooooon GATs

alamb:
> If they're not needed I can close the ticket / remove the TODOs?

I am not sure -- feel free to remove or keep.

router2/src/dml_handlers/sharded_write_buffer.rs
alamb: Maybe some comments explaining the expected behavior (aka on failure, the write / delete may have partially succeeded and thus it is on the client to retry until success)? And that since the system handles updates it should eventually converge

domodwyer: Sounds like a sensible idea to me 👍

PR #4505, opened 2022-05-02
fix: Add go_package to protos

carols10cents:
I tried to use these proto definitions in the Go client and it was sad because they didn't have `go_package` defined.


alamb:
Sorry I missed this

PR #4506, opened 2022-05-02
feat: Remove the server_benchmarks crate

carols10cents:


server_benchmarks/benches/read_filter.rs
```diff
@@ -1,133 +0,0 @@
-use criterion::{BenchmarkId, Criterion};
```
alamb, line 1: while it would be nice to have `influxrpc` benchmarks for `read_filter` and other things, looking at this implementation I realize it is quite OG specific. 

If/when we get to the point of wanting benchmarks for NG I think recreating them is probably the mot efficient plan.

alamb (approved):

Issue #4450, opened 2022-04-27
Remove OG from codebase


alamb:
Document: https://docs.google.com/document/d/1-cl3K-fNEJIoBNKeDhTlNqK5yt4XpWhbO1v-oPBDyU0/edit#

The idea here is to remove the OG code from the IOx codebase

 We will take a “walk up the dependency tree” type approach

- [x] #4473
- [x] #4474
- [x] #4475
- [x] https://github.com/influxdata/influxdb_iox/pull/4507
- [x] https://github.com/influxdata/influxdb_iox/pull/4506
- [x] https://github.com/influxdata/influxdb_iox/pull/4515
  - [x] Remove `db` crate
  - [x] Remove `internal_types` crate
  - [x] Remove `lifecycle` crate
  - [x] Remove `mutable_buffer` crate
  - [x] Remove `parquet_catalog` crate
  - [x] Remove `persistence_windows` crate
  - [x] Remove `router` crate
  - [x] Remove `server` crate
  - [x] Remove `job_registry` crate
- [x] https://github.com/influxdata/influxdb_iox/pull/4522
  - [x] Disentangle and remove `iox_object_store` crate
  - [x] Move any `data_types` used by NG to `data_types2`
  - [x] Remove `data_types` crate
  - [x] #4451
  - [x] Remove all "2" suffixes (router2, data_types2, etc)
- [x] Moar TBD

Then there will be a period of time searching for dead code / cleaning it up. Here is a partial list

- [x] https://github.com/influxdata/influxdb_iox/issues/4489
- [x] https://github.com/influxdata/influxdb_iox/pull/4584
- [x] https://github.com/influxdata/influxdb_iox/pull/4586

alamb:
cc @carols10cents  I tried to start writing up some of the main initial tasks here

alamb:
Tracking all remaining tasks in  https://github.com/influxdata/influxdb_iox/issues/4597 so I can claim success on this task!

PR #4515, opened 2022-05-04
fix: Remove lots of OG crates

carols10cents:
Connects to #4450.

Depends on #4486.

I started trying to detangle crates like parquet_file from OG before doing this, but I was updating a lot of code slated for deletion, so... I decided to first delete the majority of the code instead :)


alamb:
Screen Shot 2022-05-04 at 8 12 24 AM
😮 👍

alamb:
rebased on master

influxdb_iox/Cargo.toml
```diff
@@ -29,27 +27,21 @@ ioxd_ingester = { path = "../ioxd_ingester"}
...
-mutable_buffer = { path = "../mutable_buffer" }
 object_store = { path = "../object_store" }
 observability_deps = { path = "../observability_deps" }
 panic_logging = { path = "../panic_logging" }
-parquet_catalog = { path = "../parquet_catalog" }
 parquet_file = { path = "../parquet_file" }
 predicate = { path = "../predicate" }
 querier = { path = "../querier" }
 query = { path = "../query" }
-read_buffer = { path = "../read_buffer" }
```
alamb, line 48: I think we still need the read buffer -- that should not be deleted

query_tests/src/scenarios/library.rs
```diff
@@ -37,127 +29,6 @@ impl DbSetup for MeasurementWithMaxTime {
     }
 }
 
-// TODO: rewrite this for NG. Need to see whether we will need to test empty DB and if so how
-// <https://github.com/influxdata/influxdb_iox/issues/4488>
```
alamb, line 41: What do we think about getting these tests ported over? Should we just leave a link in the relevant ticket so the task isn't lost?

@NGA-TRAN  do you have any timeline / plans for porting this test over?

query_tests/src/scenarios/library.rs
```diff
@@ -1309,68 +1150,6 @@ impl DbSetup for MeasurementForWindowAggregate {
     }
 }
 
-pub struct MeasurementForWindowAggregateMonths {}
-#[async_trait]
-impl DbSetup for MeasurementForWindowAggregateMonths {
```
alamb, line 1314: @NGA-TRAN or @crepererum  can you confirm this test doesn't need to be ported to an NG scenario? Or is it still a todo?

query_tests/src/cancellation.rs
```diff
@@ -1,129 +0,0 @@
...
-use std::{
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
-    time::Duration,
-};
-
-#[tokio::test]
-async fn test_query_cancellation_slow_store() {
```
alamb, line 18: cc @crepererum  is this an important test to port over (a throttled object store?)

query_tests/generate/src/main.rs
```diff
@@ -58,10 +58,10 @@ fn find_sql_files(root: &Path) -> Vec<PathBuf> {
     sqls
 }
 
-/// writes out what will be the rust test file that lists out .sqls
+/// Writes out what will be the Rust test file that lists out .sqls
 fn make_cases_rs(sqls: &[PathBuf]) -> Vec<String> {
     let mut output_lines: Vec<String> = vec![r#"
-//! This file is auto generated by build.rs
+//! This file is auto generated by query_tests/generate.
```
alamb, line 64: 👍

query_tests/README.md
```diff
@@ -1,27 +1,22 @@
-This crate contains "integration" tests for the query
-engine. Specifically, it runs queries against a fully created `Db`
-instance, records the output, and compares it to expected output.
+This crate contains "integration" tests for the query engine. Specifically, it runs queries against
```
alamb, line 1: 👍  for readme updates

alamb (approved):
Mostly looks great to me -- Very nice @carols10cents 

# Concern 1: `query_test`
Some of the `query_test` coverage that this PR removes -- but as long as we have tickets tracking the work I think it is ok to merge this PR and do the porting as part of a subsequent PR (by reviving the relevant parts from this PR, perhaps)

# Concern 2: `read_buffer`
We are almost certainly will need the ReadBuffer in the near term for query performance, and even though it isn't used by NG yet I don't think we should delete it

I am approving this PR as it is so large and thus will likely collect merge conflicts if we leave it open for long but I think the two concerns above should be addressed prior to merging

influxdb_iox/Cargo.toml
carols10cents: Absolutely!

query_tests/src/cancellation.rs
carols10cents: Issue here: https://github.com/influxdata/influxdb_iox/issues/4529

query_tests/src/scenarios/library.rs
carols10cents: Ticket here: https://github.com/influxdata/influxdb_iox/issues/4530

carols10cents:
Ok, I put back the read_buffer and packers crates, and filed issues for those tests that were removed that didn't already have an issue. So merging!!!

alamb:
🔥  whoop whoop

query_tests/src/scenarios/library.rs
NGA-TRAN: @alamb 
I was actually planning to ask @carols10cents to take care of this because I am not sure whether it is useful to test queries on an empty DB in NG and, at this point, I think @carols10cents knows better where in our tests an empty DB is created. I let Carol decide whether to add emptyDB tests or not

PR #4516, opened 2022-05-04
chore: Update datafusion and upgrade arrow/parquet/arrow-flight to 13

alamb:
# Rationale
* Upgrade to arrow 13
* Get basic subquery support: https://github.com/apache/arrow-datafusion/pull/2421
* get new scheduler, https://github.com/apache/arrow-datafusion/pull/2226

# Changes:
1. Add a script to update arrow versions automatically (so I don't have to do it manually)
1. Update to https://github.com/apache/arrow-datafusion/commit/b7bb2cfba13cc04a08c2f687102dd14a8dedc7b6
2. Upgrade arrow/parquet/arrow-flight to 13
3. Update code for changes in API


scripts/update_arrow_deps.py
```diff
@@ -0,0 +1,74 @@
...
+# python update_arrow_deps.py 13
+#
+
+import argparse
+from pathlib import Path
+
+# use tomlkit as it preserves comments and other formatting
+import tomlkit
+
+def update_version_cargo_toml(cargo_toml, new_version):
```
alamb, line 36: Regarding this script, @crepererum  asked "why can't you use `cargo update`"

The answer as I understand it is that `cargo update` will update the contents of `Cargo.lock` files rather than `Cargo.toml` files -- aka it will only do semver compatible updates, which updating arrow to `13` is not

PR #4518, opened 2022-05-05
chore: Update datafusion (again)

alamb:
~Draft until https://github.com/influxdata/influxdb_iox/pull/4520 is merged~

# Rationale
* Get API change: https://github.com/apache/arrow-datafusion/pull/2434 from @tustvold

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/7304719bb4830c873af32f873ce22f205fef4c77
2. Update APIs


tustvold:
I'd be happy to pick this up given I made the breaking change

alamb:
>  I'd be happy to pick this up given I made the breaking change

Thanks -- it will be good for me to do it. Maybe you can review :)

 It turns out we actually have some operators that do their work in `execute` which the change uncovered 🎉

db/src/system_tables.rs
```diff
@@ -198,7 +198,6 @@ impl<T> std::fmt::Debug for SystemTableExecutionPlan<T> {
     }
 }
 
-#[async_trait]
```
alamb, line 201: this is the core change -- `ExecutionPlan` is no longer is async and so the code for `ExecutionPlan::execute` needs to be updated

query/src/provider/deduplicate.rs
```diff
@@ -1157,19 +1154,19 @@ mod test {
 
             debug!(partition, "Start DummyExec::execute");
 
-            // ensure there is space to queue up the channel
-            let (tx, rx) = mpsc::channel(self.batches.len());
+            // queue them all up
+            let (tx, rx) = mpsc::unbounded_channel();
```
tustvold, line 1158: Drive by fix?

tustvold (approved):

query/src/provider/deduplicate.rs
alamb: The original `channel` has  an `async` send -- so I switched to using `unbounded_channel` here to avoid that nonsense I suppose I could have used `try_send(batch).unwrap()` or something but this felt more natural

Issue #2386, opened 2021-08-23
SchemaPivot (used for tag_keys, tag_values gRPC queries) is not streaming


alamb:
__Steps to reproduce:__
Run a storage gRPC tag_keys or tag_values query

__Expected behaviour:__
The call to `ExecutionPlan::execute()` will return quickly and then all computation will happen when `next()` is called on the resulting stream

__Actual behaviour:__
`execute()` may take a long time to execute, and then the stream's `next()` method will return immediately

Note this is because I didn't really know what I was doing when I wrote this code:
https://github.com/influxdata/influxdb_iox/blob/7602bde850976f3447d18fa1af558d7bc2c79f79/query/src/exec/schema_pivot.rs#L223-L252

The loop that pulls the next input should happen in some other `tokio::task` rather than the main call to `execute()`.

I doubt this is a real issue at the moment because immediately after the call to `execute()` the call to `collect()` will be invoked anyways so it would be hard to tell the difference between the two implementations

PR #4520, opened 2022-05-05
fix: Move schema pivot work out of `execute`

alamb:
# Rationale
Closes https://github.com/influxdata/influxdb_iox/issues/2386

https://github.com/apache/arrow-datafusion/pull/2434  changes the DataFusion execution plan API to make it harder to do work during `execute` (rather than the stream phase).

While working to integrate this change into IOx in https://github.com/influxdata/influxdb_iox/pull/4518 it turns out that the DeduplicateExec does its work during `execute` (which was a known issue https://github.com/influxdata/influxdb_iox/issues/2386 :tada:):

# Changes
1. Refactor a pattern used in deduplicate and is_null_check into `datafusion_util`
2. Use the refactored code to run the schema pivot logic in a separate tokio task

There may be a better pattern than a task waiting to do something, but perhaps @tustvold can weign in there


tustvold:
It is possible to do this without needing to spawn a task, or any buffering, which will also ensure this will work correctly with the new scheduler.

The "trick" I've applied elsewhere is to use `futures::stream::once()` to enclose the async work that was being performed before, that yields a Stream of `ArrowResult<RecordBatch>`.

Then call flatten() (or try_flatten() if the closure is fallible) to remove the nested streams

Then use the RecordBatchAdapterStream to make this stream into a SendableRecordBatchStream.

FWIW this was the crux of my objection to SendableRecordBatchStream - the stream combinators don't work correctly, but RecordBatchAdapterStream works to plug this so :shrug:

alamb:
> It is possible to do this without needing to spawn a task, or any buffering, which will also ensure this will work correctly with the new scheduler.


@tust  Would you be ok with doing this as a follow on PR? 

I have some other stuff (DataFusion reviews, and maybe work on Merging) I want to work on with the balance of my work week but maybe could look into it next week

Or perhaps you would be interested in preparing IOx for the new DF scheduler?

tustvold (approved):

alamb:
Filed https://github.com/influxdata/influxdb_iox/issues/4521 for follow on work

PR #4517, opened 2022-05-04
fix: Remove OG proto definitions

carols10cents:
Closes #4489.

~~Depends on #4515.~~

PR #4522, opened 2022-05-05
fix: Delete even more 

carols10cents:
~~Depends on #4517.~~

Takes care of most of the remaining items in https://github.com/influxdata/influxdb_iox/issues/4450

I'd highly recommend reviewing this  commit-by-commit.

I opted to move everything into the top level of data_types (neé data_types2), even though it's getting really long, because I don't think the organization that was in the old data_types was right anymore. However, I'm not sure how best to organize it now :)


tustvold (approved):
This looks good to me, I think there are some things that could be further removed - e.g. I don't think anything is actually using `TimestampSummary` anymore, but I'm going to take the liberty of merging this as is because the conflict potential is so high and I think it moves us in the right direction

PR #4531, opened 2022-05-06
test: POC of new group by implementation in DataFusion (DO NOT MERGE)

alamb:
I do not plan to merge this PR. 

Rather this PR demonstrates the new grouping from @yjshen in https://github.com/apache/arrow-datafusion/pull/2375 in IOx  so that we can run the IOx test suite against it, which has turned up some issues that were not covered by the existing datafusion tests.


alamb:
The tests passed which was the whole point of this PR. whoop whoop!

PR #4532, opened 2022-05-06
fix: Remove or rescope dead code warnings

carols10cents:
Depends on https://github.com/influxdata/influxdb_iox/pull/4522

I noticed some `allow(dead_code)`s that were applied at the crate level, which was good while the new crates were under heavy churn. Now, though, they masked a few things like actually unused functions and a test that didn't have the test annotations. So I removed the crate-level ones, and for the code that looked like it should stay, I moved the annotations to be more specific.

There was also some code marked with `allow(dead_code)` that wasn't actually dead, so I removed those annotations.


querier/src/poison.rs
```diff
@@ -30,6 +30,7 @@ impl PoisonCabinetInner {
 
 #[derive(Debug)]
 pub struct PoisonCabinet {
+    #[allow(dead_code)]
```
tustvold, line 33: I wonder if this should actually be #[cfg(test)] around the module?

tustvold (approved):

querier/src/poison.rs
domodwyer: It's test code, but in prod code paths - https://github.com/influxdata/influxdb_iox/blob/9a876c5577b9c7aa083cdc767ee788070dbbb246/ingester/src/lifecycle.rs#L423

I plan on removing this when I get round to refactoring this bit of the ingester - it seems like a bit of a code smell to conflate test/prod paths like thsi.

domodwyer (approved):

PR #4534, opened 2022-05-07
refactor: simplify object_store path handling

tustvold:
Whilst reviewing the object store abstraction from the perspective of using it in Datafusion (https://github.com/apache/arrow-datafusion/issues/2445) a couple of pain points cropped up that would make it quite painful to use for this purpose:

* The use of enumerations would make out-of-crate extension impossible, e.g. to add HDFS support
* Paths can't be created without knowledge of the object store they are for

Fortunately with the change made by @crepererum in https://github.com/influxdata/influxdb_iox/pull/3393 we treat all paths the same, and no longer need any notion of directory vs file. Not only does this solve the above problem, but is also eliminates a whole host of complexity. Unfortunately we use paths in a LOT of places... I think this is definitely an "oh lawd he comin"...


iox_object_store/src/lib.rs
```diff
@@ -654,11 +643,12 @@ mod tests {
         assert_eq!(file_count, 0);
     }
 
-    fn make_owner_path(object_store: &DynObjectStore, uuid: Uuid) -> Path {
-        let mut p = object_store.new_path();
-        p.push_all_dirs(&[ALL_DATABASES_DIRECTORY, uuid.to_string().as_str()]);
-        p.set_file_name("owner.pb");
-        p
+    fn make_owner_path(_object_store: &DynObjectStore, uuid: Uuid) -> Path {
+        Path::from_iter([
```
alamb, line 647: `from_iter` certainly looks good to me 👍

alamb (commented):
I skimmed a bunch of the changes in usage of the ObjectStore and I think it looks great

I did not scrutinize the implementation but overall I like where this is headed 👍

clap_blocks/src/object_store.rs
```diff
@@ -222,152 +226,157 @@ pub enum ObjectStoreType {
...
 }
 
-pub fn warn_about_inmem_store(config: &ObjectStoreConfig) {
-    match config.object_store {
-        Some(ObjectStoreType::Memory) | None => {
-            warn!("NO PERSISTENCE: using Memory for object storage");
-        }
-        Some(store) => {
-            info!("Using {:?} for object storage", store);
+#[cfg(feature = "gcp")]
```
tustvold, line 229: This logic is moved into here to simplify the object_store crate, and make it easier for external consumption. If a feature isn't enabled, you will get a compile error

iox_object_store/src/paths/parquet_file.rs
```diff
@@ -229,70 +207,73 @@ pub enum ParquetFilePathParseError {
...
             "got {:?}",
             result
         );
 
-        let mut extra = df.clone();
-        extra.push_dir("nope");
-        let result = ParquetFilePath::from_relative_dirs_and_file_name(&extra);
+        let extra = df.clone();
+        let df = extra
+            .child("00000000-0000-0000-0000-00000000000a.parquet")
```
tustvold, line 247: Because we no longer distinguish between directories and filenames, we need an additional level to get the UnexpectedDirectory error, otherwise we would just get InvalidChunkId.

object_store/src/aws.rs
```diff
@@ -506,12 +479,12 @@ impl AmazonS3 {
         }
         use ListState::*;
 
-        let raw_prefix = prefix.map(|p| p.to_raw());
+        let raw_prefix = prefix.map(|p| format!("{}{}", p.to_raw(), DELIMITER));
```
tustvold, line 497: We add the trailing delimiter to the path name, so that we don't need to perform additional filtering afterwards

object_store/src/aws.rs
```diff
@@ -344,10 +323,6 @@ impl ObjectStoreApi for AmazonS3 {
                                 size,
                             })
                         })
-                        .filter(move |meta_res| match meta_res {
```
tustvold, line 347: The trailing delimiter is included in the request, so we don't need to do this anymore

object_store/src/cache.rs
```diff
@@ -1,63 +0,0 @@
-//! This module contains a trait and implementation for caching object storage objects
```
tustvold, line 1: The cache functionality is not implemented and so I just took the opportunity to remove it

object_store/src/buffer.rs
```diff
@@ -1,126 +0,0 @@
-//! This module contains a `Stream` wrapper that fully consumes (slurps) a
```
tustvold, line 1: This doesn't seem to be used, so just removed it

object_store/src/disk.rs
```diff
@@ -192,69 +171,50 @@ impl ObjectStoreApi for File {
...
-        let mut search_path = resolved_prefix.clone();
-        search_path.unset_file_name();
-
-        let walkdir = WalkDir::new(&search_path.to_path_buf())
-            .min_depth(1)
-            .max_depth(1);
+    async fn list_with_delimiter(&self, prefix: &Path) -> Result<ListResult> {
+        let resolved_prefix = self.path_to_filesystem(prefix);
+
+        let walkdir = WalkDir::new(&resolved_prefix).min_depth(1).max_depth(1);
```
tustvold, line 192: This technically could just list a directory, but I kept it using WalkDir for consistency

object_store/src/dummy.rs
```diff
@@ -1,130 +1,63 @@
...
-        &self,
-        _prefix: &Self::Path,
-    ) -> crate::Result<crate::ListResult<Self::Path>, Self::Error> {
-        NotSupportedSnafu { name: &self.name }.fail()
+    async fn list_with_delimiter(&self, _prefix: &Path) -> Result<ListResult> {
+        Ok(NotSupportedSnafu { name: self.name }.fail()?)
     }
 }
-
-/// Stub when s3 is not configured
```
tustvold, line 92: This code is moved into clap_blocks as it is really a quirk of the IOx binary build, and would be confusing in an externally published object_store crate

object_store/src/gcp.rs
```diff
@@ -387,115 +362,55 @@ mod test {
...
-        } = err
-        {
-            assert!(matches!(source, cloud_storage::Error::Google(_)));
-            assert_eq!(bucket, config.bucket);
-            assert_eq!(path, NON_EXISTENT_NAME);
-        } else {
-            panic!("unexpected error type: {:?}", err)
-        }
+        let err = integration.delete(&location).await.unwrap_err().to_string();
+        assert!(err.contains("Unable to DELETE data"), "{}", err)
```
tustvold, line 400: Interestingly this is a different behaviour from S3

object_store/src/lib.rs
```diff
@@ -18,595 +18,97 @@
...
     /// `foo/bar_baz/x`.
-    async fn list_with_delimiter(
-        &self,
-        prefix: &Self::Path,
-    ) -> Result<ListResult<Self::Path>, Self::Error>;
-}
-
-/// Universal interface to multiple object store services.
-#[derive(Debug)]
-pub struct ObjectStoreImpl {
```
tustvold, line 123: ObjectStoreApi is now object-safe, and so this isn't needed

object_store/src/lib.rs
```diff
@@ -647,40 +150,18 @@ impl GetResult<Error> {
...
-
-    #[snafu(display("Azure Blob storage-based Object Store error: {}", source))]
-    AzureObjectStoreError { source: azure::Error },
-
-    #[snafu(display("In-memory-based Object Store error: {}", source))]
-    InMemoryObjectStoreError { source: memory::Error },
-
-    #[snafu(display("{}", source))]
-    DummyObjectStoreError { source: dummy::Error },
+    #[snafu(display("Generic {} error: {}", store, source))]
```
tustvold, line 160: This enumeration is made generic, so that ObjectStoreApi can be implemented outside of this crate

object_store/src/lib.rs
```diff
@@ -18,595 +18,97 @@
 //! Future compatibility will include Azure Blob Storage, Minio, and Ceph.
 
 #[cfg(feature = "aws")]
-mod aws;
+pub mod aws;
```
tustvold, line 21: This structure is made public to make it easier to extend with object store specific structures, e.g. config

object_store/src/lib.rs
```diff
@@ -18,595 +18,97 @@
...
-mod memory;
+pub mod gcp;
+pub mod memory;
 pub mod path;
-mod throttle;
+pub mod throttle;
 
-pub mod cache;
-pub mod dummy;
+#[cfg(test)]
```
tustvold, line 31: This is only needed in tests now

object_store/src/instrumentation.rs
```diff
@@ -81,9 +82,13 @@ pub struct ObjectStoreMetrics<T, P = SystemProvider> {
     list_error_duration_ms: U64Histogram,
 }
 
-impl<T> ObjectStoreMetrics<T> {
+impl ObjectStoreMetrics {
     /// Instrument `T`, pushing to `registry`.
-    pub fn new(inner: T, registry: &metric::Registry) -> Self {
+    pub fn new(
+        inner: Arc<dyn ObjectStoreApi>,
```
tustvold, line 88: This needed to be changed to `Arc<dyn ObjectStoreApi>` as previously it worked because `ObjectStoreImpl` was a concrete type. This is no longer the case with, the clap_blocks::object_store returning `Arc<dyn ObjectStoreApi>`

This would mean either implementing `ObjectStoreApi` for `Arc<dyn ObjectStoreApi>` or just passing this `Arc<dyn ObjectStoreApi>` directly. The latter was less code and seemed simpler.

As we made this change, it made sense to just do the same for `TimeProvider`.

object_store/src/path/parts.rs
```diff
@@ -47,40 +44,33 @@ const INVALID: &AsciiSet = &CONTROLS
...
             // We don't want to encode `.` generally, but we do want to disallow parts of paths
             // to be equal to `.` or `..` to prevent file system traversal shenanigans.
-            "." => Self(String::from("%2E")),
-            ".." => Self(String::from("%2E%2E")),
-
-            // Every string except the empty string will be percent encoded.
-            // The empty string will be transformed into a sentinel value EMPTY
-            // which can safely be a prefix of an encoded value since it will be
-            // fully matched at decode time (see impl Display for PathPart).
-            "" => Self(String::from(EMPTY)),
```
tustvold, line 62: We no longer allow empty path segments within `Path`

object_store/src/path/parts.rs
```diff
@@ -1,23 +1,20 @@
...
-const EMPTY: &str = "%";
-
 /// The PathPart type exists to validate the directory/file names that form part
 /// of a path.
 ///
 /// A PathPart instance is guaranteed to be non-empty and to contain no `/`
 /// characters as it can only be constructed by going through the `from` impl.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
-pub struct PathPart(pub(super) String);
+pub struct PathPart<'a> {
```
tustvold, line 15: This is made `Cow` to allow `Path` to split up its owned `String` into an iterator of `PathPart` without copying

object_store/src/path/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /// Creates a cloud storage location by joining this `CloudPath`'s
+    /// parts with `DELIMITER`
+    pub fn to_raw(&self) -> &str {
+        &self.raw
+    }
+
+    /// Returns the [`PathPart`] of this [`Path`]
+    pub fn parts(&self) -> impl Iterator<Item = PathPart<'_>> {
```
tustvold, line 35: There is a subtlety to this design worth highlighting.

Constructing a `PathPart` from a string escapes the provided string, `Path::from_iter` takes an iterator of things convertible to `PathPart`. By returning an iterator of `PathPart` instead of an iterator of `&str` we allow constructions like

`Path::from_iter(other_path.parts().take(1))`

Without this resulting in double-escaping

object_store/src/path/parsed.rs
```diff
@@ -1,388 +0,0 @@
...
-    ($file:expr) => {
-        parsed_path!([], $file)
-    };
-    () => {
-        parsed_path!([])
-    };
-}
-
-#[cfg(test)]
-mod tests {
```
tustvold, line 175: These tests are adapted and moved into mod.rs

object_store/src/path/cloud.rs
```diff
@@ -1,312 +0,0 @@
...
-                    directories: parts,
-                    file_name: maybe_file_name,
-                }
-            }
-            Parsed(dirs_and_file_name) => dirs_and_file_name,
-        }
-    }
-}
-
-#[cfg(test)]
```
tustvold, line 177: These tests are moved into mod.rs

object_store/src/path/file.rs
```diff
@@ -1,507 +0,0 @@
...
-    if let Some(s) = path.to_str() {
-        if let Some(c) = s.chars().last() {
-            return is_separator(c);
-        }
-    }
-    false
-}
-
-#[cfg(test)]
-mod tests {
```
tustvold, line 341: These tests are no longer relevant as we don't distinguish between directories and files

object_store/src/path/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+    #[test]
+    fn cloud_prefix_with_trailing_delimiter() {
+        // Use case: files exist in object storage named `foo/bar.json` and
+        // `foo_test.json`. A search for the prefix `foo/` should return
+        // `foo/bar.json` but not `foo_test.json'.
+        let prefix = Path::from_iter(["test"]);
+
+        let converted = prefix.to_raw();
+        assert_eq!(converted, "test");
+    }
```
carols10cents, line 128: These two tests are the same now, since adding a trailing delimiter is now part of each object store's implementation. So I think one can be deleted?

object_store/src/path/parts.rs
carols10cents: What do you mean by "no longer allow"? I can do `let part: PathPart<'_> = "".into();` and I don't get any errors or panics, and the test case for that was removed.

carols10cents (changes requested):
This generally looks ok to me, there's one failing test for me and a few things that look out of date or redundant now.

Changing from the enum to be trait-based to be extensible makes sense!

Some things to think about in regards to open-sourcing this and using it outside of IOx, for future PRs:

- There's still use of the environment variable `INFLUXDB_IOX_BUCKET` in the integration tests
- It'd be really good to actually do integration testing with the services in CI, but that gets... interesting...
- It'd be good to test the local file system object store with Windows in CI too
- To be friendly to people who want to extend this, it'd be nice to write up a guide with instructions and explanation of the semantics of implementors of the trait
- Should the object_store crate continue to live in this repo or should it get its own? Either way, it should probably have its own README

object_store/src/gcp.rs
```diff
@@ -387,115 +377,55 @@ mod test {
...
-        if let Some(ObjectStoreError::GcsObjectStoreError {
-            source: Error::UnableToStreamListData { source, bucket },
-        }) = err.downcast_ref::<ObjectStoreError>()
-        {
-            assert!(matches!(source, cloud_storage::Error::Google(_)));
-            assert_eq!(bucket, &config.bucket);
-        } else {
-            panic!("unexpected error type: {:?}", err);
-        }
+        assert!(err.contains("Unable to list data"), "{}", err)
```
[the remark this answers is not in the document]
tustvold: Aah... I did not realise we don't test that in CI...

object_store/src/path/mod.rs
```diff
@@ -0,0 +1,317 @@
...
+        let haystack = haystack.child("foo.segment");
+
+        assert!(
+            !haystack.prefix_matches(&needle),
+            "{:?} should not have started with {:?}",
+            haystack,
+            needle
+        );
+
+        // All directories match but file name is not a prefix
```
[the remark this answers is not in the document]
tustvold: :eagle: :eye:

tustvold:
> Some things to think about

Yup, there definitely will be further changes to get this ready for publication. Thanks for enumerating some of them for me :grin:

PR #4547, opened 2022-05-10
refactor: split out ObjectStoreMetrics

tustvold:
Splits out ObjectStoreMetrics so that we can publish object_store without needing to also publish `metrics` and `time`.


object_store/Cargo.toml
```diff
@@ -36,9 +36,6 @@ reqwest = { version = "0.11", optional = true, default-features = false, feature
 walkdir = "2"
 tempfile = "3.1.0"
 workspace-hack = { path = "../workspace-hack"}
-metric = { version = "0.1.0", path = "../metric" }
```
alamb, line 39: 👍

alamb (approved):

PR #4548, opened 2022-05-10
chore: rename INFLUXDB_IOX_BUCKET to OBJECT_STORE_BUCKET in object_store tests

tustvold:
This renames the environment variable used by the tests within object_store to OBJECT_STORE_BUCKET

As far as I can tell we don't run any integration tests outside of the object_store crate against mocked stores


alamb:
I think our prod system uses `INFLUXDB_IOX_BUCKET`:

https://github.com/influxdata/k8s-idpe/search?q=INFLUXDB_IOX_BUCKET

alamb:
If/when we get around to changing the names, perhaps we can add the `IOX parameter in  k8s-idpe, commit and deploy this change, and then remove the old param

alamb (approved):

PR #4550, opened 2022-05-10
fix: feature flags

tustvold:
#4534 moved the feature optional-ness to clap_blocks, however, I forgot to update the influxdb_ioxd binary. I thought I did, but perhaps got lost in the merge conflict of doom resulting from #4522


influxdb_iox/Cargo.toml
```diff
@@ -77,9 +77,9 @@ test_helpers_end_to_end_ng = { path = "../test_helpers_end_to_end_ng" }
 [features]
 default = ["jemalloc_replacing_malloc"]
 
-azure = ["object_store/azure"] # Optional Azure Object store support
-gcp = ["object_store/gcp"] # Optional GCP object store support
-aws = ["object_store/aws"] # Optional AWS / S3 object store support
+azure = ["clap_blocks/azure"] # Optional Azure Object store support
```
alamb, line 80: Doesn't it also need to activate the `object_store` features as well?

https://github.com/influxdata/influxdb_iox/blob/main/object_store/Cargo.toml#L40-L44

Or maybe that happens transitively in clap_blocks 🤔 

https://github.com/influxdata/influxdb_iox/blob/main/clap_blocks/Cargo.toml#L25-L27

PR #4553, opened 2022-05-10
feat: support HEAD requests against object storage

tustvold:
~Draft as builds on #4549~

Adds support to perform a HEAD request against object storage, this returns the object metadata without its body. This important for interacting with parquet in object storage, as Azure does not support range requests of the form `-<bytes_suffix>` and so to read a file footer you need to know how large it is. It is also occasionally used to verify object existence.


object_store/src/aws.rs
```diff
@@ -260,6 +270,40 @@ impl ObjectStoreApi for AmazonS3 {
...
+            key: key.clone(),
+            ..Default::default()
+        };
+        let s = self
+            .client()
+            .await
+            .head_object(head_request)
+            .await
+            .map_err(|e| match e {
+                rusoto_core::RusotoError::Service(rusoto_s3::HeadObjectError::NoSuchKey(_)) => {
```
tustvold, line 286: I'm not a massive fan of how verbose the error handling is in this crate, I may take some time to see if I can't clean this up in a follow up PR

alamb:
tests appear to be failing

tustvold:
💩 too many plates, will look into tomorrow

object_store/src/aws.rs
```diff
@@ -260,6 +270,56 @@ impl ObjectStoreApi for AmazonS3 {
...
+                    }
+                }
+                _ => Error::UnableToHeadData {
+                    bucket: self.bucket_name.to_owned(),
+                    path: key.clone(),
+                    source: e,
+                },
+            })?;
+
+        // Note: GetObject and HeadObject return a different date format from ListObjects
```
alamb, line 305: Maybe worth a ticket / url reference and examples

This is 🤯  lol

```suggestion
        // Note: GetObject and HeadObject return a different date format from ListObjects
        //
        // S3 List returns timestamps in the form
        //     <LastModified>2013-09-17T18:07:53.000Z</LastModified>
        // S3 GetObject returns timestamps in the form
        //            Last-Modified: Sun, 1 Jan 2006 12:00:00 GMT
```

alamb (approved):
LGTM

PR #4557, opened 2022-05-10
feat: Interrogate schema from querier (as well as router)

alamb:
# Rationale
While working on https://github.com/influxdata/influxdb_iox/pull/4556, I wanted to be able to get schema information (added in https://github.com/influxdata/influxdb_iox/pull/4105 and https://github.com/influxdata/influxdb_iox/pull/4081 by @lukebond ) by talking to the querier

The querier is the default service that the CLI connects to in all-in-one mode and the one we connect to in production. Turns out that it is not possible because the schema service is implemented on the router (not the querier).

# Review Suggestion
I suggest reviewing this PR by commit:
- [ ] the first commit simply moves code out of the `router` crate and into the `service_grpc_schema` crate
- [ ] the second commit connects the code up in `querier` and adds a test

# Current Behavior
```shell
(arrow_dev) alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ../iox-target/debug/influxdb_iox debug schema get 26f7e5a4b7be365b_917b97a92e883afc
Error in schema subcommand: Client error: Operation is not implemented or supported:
```

The work around is to talk directly to the router, by adding `--host http://localhost:8081` to the command

# Behavior after this PR
```
(arrow_dev) alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ../iox-target/debug/influxdb_iox  debug schema get 26f7e5a4b7be365b_917b97a92e883afc
{
  "id": "1",
  "kafkaTopicId": "1",
  "queryPoolId": "1",
  "tables": {
    "mem": {
      "id": "2",
      "columns": {
        "time": {
          "id": "10",
          "columnType": 6
        },
        "host": {
          "id": "16",
          "columnType": 7
        },
        "available": {
          "id": "17",
          "columnType": 1
        },
        "wired": {
...
```


NGA-TRAN (approved):
Nice.
I wonder if you can document  this and the similar previous PR? They will be super helpful

alamb:
> I wonder if you can document this and the similar previous PR? They will be super helpful

Good idea @NGA-TRAN  -- where would you go for documentation? (aka where do you suggest I write it up -- in the help text of the CLI? In a separate doc?)

NGA-TRAN:
@alamb 
>  where would you go for documentation? (aka where do you suggest I write it up -- in the help text of the CLI? In a separate doc?)

I think IOx doc https://github.com/influxdata/influxdb_iox/tree/main/docs would be a good place. We have SQL tips and tricks https://github.com/influxdata/influxdb_iox/blob/main/docs/sql.md. Maybe a new similar link these CLI. I am thinking to pull system table out in its own link, too.

alamb:
> I think IOx doc https://github.com/influxdata/influxdb_iox/tree/main/docs would be a good plac

I will do it @NGA-TRAN  -- thank you

alamb:
Docs PR in https://github.com/influxdata/influxdb_iox/pull/4559

Issue #4567, opened 2022-05-11
Support streaming ingester state to queriers


domodwyer:
Currently when an ingester attempts to execute a query on in-memory state for a querier, it buffers all the results in memory before returning all record batches to the querier.

This causes each query to use a (potentially) large amount of RAM to buffer all the records. A query with no predicate pushed down will copy all the table data in memory before returning it to the querier.

This ticket is to implement streaming record batches back to the querier instead of buffering them in memory.

alamb:
I think we do a similar thing for actual query results -- more streaming more better as my mother always says 👍

alamb:
Assigning this to myself as I believe @domodwyer  is tracking down another issue at the moment

alamb:
So there are at least two copies of the data that are buffered (in addition to the `RecordBatch`es the ingester has already):

```text
                  Buffering location             Buffering location #2:
                  #1: turned into a               turned into a single
┌─────────────┐   single RecordBatch              RecordBatch prior to
│┌───────────┐│   to feed into query               sending to querier
││RecordBatch││         engine
│└───────────┘├─────┐
│             │     │
│SnapshotBatch│     │      ┌────────────────────────┐       ┌ ─ ─ ─ ─ ─ ┐        ┌────────────────────────┐
└─────────────┘     │      │                        │                            │                        │
                    ├─────▶│      Query Engine      │──────▶│    ...    │────┬──▶│ IngesterQueryResponse  │
      ...           │      │                        │                        │   │                        │
                    │      └────────────────────────┘       └ ─ ─ ─ ─ ─ ┘    │   └────────────────────────┘
┌─────────────┐     │                                                        │
│┌───────────┐│     │     Runs query engine, which                           │
││RecordBatch││     │     deduplicates and applies                           │
│└───────────┘│─────┘    filters, produces a Stream                          │
│             │                                                            ──┤
│SnapshotBatch│                                                              │
└─────────────┘                                                              │
                                                                     ...     │
                                                                             │
                                                                             │
                                                            data from other  │
                                                               partitions    │
                                                                          ───┘

```

# Plan
I plan remove the extra buffering in at least two stages:
1. Remove buffering #1: by implementing `QueryChunk` for SnapshotBatch rather than `QueryBatch` and let the query engine do the schema adapting
2. Remove buffering #2: find some alternate way to pass partition information along to the querier that doesn't require knowing all the output record batches up front

# Alternates considered
[...]

PR #4569, opened 2022-05-11
feat(ingester): emit query latency metrics

domodwyer:
This will help quantify the volume of queries hitting the ingester for debugging.

The RPC call doesn't seem to be tested, but I'll add coverage as part of #4567.

---

* feat(ingester): emit query latency metrics (3890a8986)

      Adds a histogram metric "flight_query_duration_ms" that records the duration a
      flight RPC query takes to complete. Broken down by query result
      (success/error).


alamb (commented):
Should we write at least a basic test?

alamb (approved):

PR #4570, opened 2022-05-11
refactor(ingester): emit per-op debugging info

domodwyer:
This will be helpful to identify which of the kafka messages contains malformed data.

---

* refactor(ingester): emit per-op debugging info (a188d3948)

      Emit a DEBUG level log containing the op offset & other helpful fields.

      This will allow us to identify which messages were last successfully decoded,
      and which caused errors so we can pull them from analysis.


domodwyer:
Ok - good to go! 👍

alamb (approved):

PR #4573, opened 2022-05-11
feat: Querier can get write status from ingesters

carols10cents:
Connects to influxdata/influxdb-iox-client-go#27.

Now you can ask a querier the status of a particular write, and the querier will ask its ingesters what they know about that write then merge the responses.

I've updated the end-to-end tests in this repo to ask the querier if data is readable/persisted instead of the ingester, to verify this new functionality is working. The rest of influxdata/influxdb-iox-client-go#27 is having the Go client ask the querier for this information in a similar fashion, to know when data is visible for tests instead of sleeping.

I'm still not quite sure this is the exact right way/place to do this; I added on to the trait used for queriers to talk to ingesters, which seems right, but there's a `MockIngesterConnection` that implements the trait, and I didn't implement anything for the write info method on the mock because the method it would be mocking just delegates to the ingester connection.

This also feels weird because I've brought a bunch of gRPC types into code that didn't used to need them, but because this code is merging multiple gRPC responses into one, I didn't really see a point in converting them to plain Rust objects and then right back again?


querier/src/database.rs
```diff
@@ -123,6 +127,11 @@ impl QuerierDatabase {
             .expect("retry forever")
     }
 
+    /// Get write info for the given token
+    pub async fn get_write_info(&self, write_token: &str) -> Result<GetWriteInfoResponse, Error> {
```
alamb, line 131: I think you could avoid the exposing of generated types here by passing the `ingester_connection` directly back to the `QuerierWriteInfoServiceImpl`-- and that would keep the generated code a little more isolated

alamb (approved):
makes sense to me -- thank you @carols10cents  

I agree it feels a little messy with the generated_types proliferating. The only thing I can think of is to move more of the logic into `data_types` and avoid importing generated_types as much as possible

alamb:
I basically think this branch does what is necessary, and I don't see any reason it can't be merged (even if the code might be able to be polished a bit)

alamb:
Thank you @carols10cents 🙏

alamb:
Thank you @carols10cents

Issue #4568, opened 2022-05-11
cache / reuse gRPC connections from querier to ingester


alamb:
__Use case:__
During investigation of our internal cluster, we believe something about the number of queries made from querier to ingester is causing a problem. 

__Proposal:__
We should reuse the connections

__Current behaviour:__
Currently each query will (effectively) open a connection to each ingester. 

__Desired behaviour:__
Use the same connection for each ingester

__Alternatives considered:__
There is an implementation of this in OG code somewhere -- I will resurrect that implementation and use it in the querier.

PR #4574, opened 2022-05-11
feat: cache / reuse gRPC connections from querier to ingester

alamb:
Draft until:
- [ ] Test for LRU eviction
- [ ] Implement / test max TTL in cache

# Rationale
During investigation of our internal cluster, we believe something about the number of queries made from querier to ingester is causing a problem. Currently each query will (effectively) open a connection to each ingester.

Closes https://github.com/influxdata/influxdb_iox/issues/4568

# Changes:
1. Add a grpc_client_pool crate to reuse connections, partially based on the  grpc connection pool introduced in https://github.com/influxdata/influxdb_iox/pull/3040 by @crepererum
2. Use said connection pool for querier --> ingester communication

# FAQ
## Why not reuse OG connection pool

I did think about Ressurectung grpc connection pool introduced in https://github.com/influxdata/influxdb_iox/pull/3040 by @crepererum but I ~got lazy~ didn't want to deal with all the templates and it was not clear that we actually had connected ttl or error handling correctly connected up anyways

## Why not make this more general?

I also thought about making this more general (for example,  putting it into the `influxdb_iox_client`) and decided that was a much larger project than I wanted to invest in, so I just followed the existing model of specialized cache for IOx's use

## Why not use [`deadpool`](https://docs.rs/deadpool/latest/deadpool/index.html)

`deadpool` does not seem to be in our dependency stack yet https://github.com/influxdata/influxdb_iox/blob/main/Cargo.lock

It also seems somewhat light on testing: https://github.com/bikeshedder/deadpool/search?q=test

(the timeout testing in particular is quite lacking: https://github.com/bikeshedder/deadpool/blob/84ecea31785ec8fd9aa490ee0c33894a924b40d2/tests/managed_timeout.rs#L45-L61)


grpc_client_pool/src/pool.rs
```diff
@@ -0,0 +1,398 @@
...
+    },
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// PoolOptions
+#[derive(Debug)]
+pub struct ClientPoolOptions {
+    /// Where to get times from
+    time_provider: Arc<dyn TimeProvider>,
```
crepererum, line 40:
```suggestion
    pub time_provider: Arc<dyn TimeProvider>,
```

grpc_client_pool/src/pool.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    /// maximum number of entries to store in the cache
+    pub max_entries: usize,
```
crepererum, line 43: I find the logic in this crate very confusing: this number describes the size of the non-leased clients that are kept within the pool but not the maximum number of clients that the pool stores + hands out at any given client. I think this is very different to what pool sizes normally do. Esp. in an async context this basically means that you can have an unbound number of clients (or in other words: the size is now limited by the number of concurrent queries that we allow). Is there are reason for this semantic?

alamb: As discussed I will change the logic so that the pool will never hand out more than `max_entries` clients and will wait (with timeout) to get a new connection if that limit has been reached.

alamb: In an attempt to make the behavior "least surprising as possible" I plan to implement the three settings from the .NET connection pool, as described in: https://www.stevejgordon.co.uk/httpclient-connection-pooling-in-dotnet-core

PR #4579, opened 2022-05-12
refactor: Rearrange delegation to remove one usage of generated types

carols10cents:
Just a small improvement suggested by @alamb https://github.com/influxdata/influxdb_iox/pull/4573#discussion_r870752406

The only other thought I had was making `influxdb_iox_client` able to support multiple connections (to the multiple ingesters) and then move the merging into there too? I'm not sure that's better or desired. 

I also see the connection caching that should change code around here too


alamb (approved):
I like it (though if course I am biased 😆 )

> The only other thought I had was making influxdb_iox_client able to support multiple connections (to the multiple ingesters) and then move the merging into there too? I'm not sure that's better or desired.

TLDR "no I don't think it is desired to have the client make connections to multiple ingesters" 

The reason is that for any setup like our production environment, the ingesters will be on different pods / hosts and so different network connections would be needed.

Since the client already needs access to the querier and the querier already needs to be configured to talk to the ingester, reusing that pre-configured communication mechanism I think is the easiest way to determine the visibility of a current write.

PR #4580, opened 2022-05-12
feat: Add more logging to understand the flaky multi ingester test better

carols10cents:
Connects to #4395.

I still don't fully understand what's happening when the multi ingester test fails in CI. I can't get it to fail locally. I haven't had success getting it to fail in CI so far either, so I'm proposing to merge this that turns up logging in CI to get more information next time it happens.


alamb (approved):
I like it 👍

carols10cents:
I handled it with a hack that I pinky swear to revert when we figure out what's going on 😅

PR #4582, opened 2022-05-12
feat: Demonstrate reusing connections (NOT FOR MERGING)

alamb:
@tustvold mentioned in our chat:

> I'm surprised you can't just have one connection and be done with it
>
> If you can't share the connection that sounds like a tonic bug

However, when I tried sharing the same `Connection` across multiple clients in the end to end tests, I remember the clients being very confused. 

However, when I hacked connection re-use into the querier, it seems to have gone just find 👍


querier/src/ingester/flight_client.rs
```diff
@@ -52,7 +54,9 @@ pub trait FlightClient: Debug + Send + Sync + 'static {
 
 /// Default [`FlightClient`] implemenetation that uses a real client.
 #[derive(Debug, Default)]
-pub struct FlightClientImpl {}
+pub struct FlightClientImpl {
+    connections: Mutex<HashMap<String, Connection>>,
```
alamb, line 58: this doesn't ever release connections, but otherwise this demonstrates that using a `Connection` pool would likely work just fine

PR #4583, opened 2022-05-12
fix: Use one (shared) http2 connection per querier and ingester pair

alamb:
# Rationale
During investigation of our internal cluster, we believe something about the number of connections made from querier to ingester is causing a problem. Currently each query will open a new connection to each ingester.

I started working on a full featured client pool in https://github.com/influxdata/influxdb_iox/pull/4574, but then @tustvold noted that the existing `tonic::Connection` is already  shareable (and handles multiplexing multiple concurrent requests

Closes https://github.com/influxdata/influxdb_iox/issues/4568
Closes https://github.com/influxdata/influxdb_iox/pull/4574
Closes https://github.com/influxdata/influxdb_iox/pull/4582

# Changes:
1. Create `Connections` from the querier to the ingester on first use and reuse them for subsequent communications

# FAQ

## Why not connect on startup?

Doing so might effectively gate the startup of the querier pod to successful connection with the ingesters (so queriers won't be healthy if they can't connect to ingester). While this might be a good idea, I think we could/should do it as a follow on PR. 

## How does this handle DNS updates (aka ingester pod rollovers)?

Quoting @tustvold: presumably a DNS change would be coupled with dropping of the connection (aka when a new ingester is added, existing ones will also be killed). The tower reconnect layer will kick in and create a new connection in this case

## How will this handle querying multiple ingesters behind the same address?
Honest answer: "I am not ure"

Perhaps this is what the `istio` magic we have deployed in kubernetes will accomplish?


querier/src/ingester/flight_client.rs
```diff
@@ -153,3 +156,53 @@ impl QueryData for PerformQuery<IngesterQueryResponseMetadata> {
...
+        let ingester_address = self.ingester_address.as_ref();
+
+        if let Some(connection) = maybe_connection.as_ref() {
+            debug!(%ingester_address, "Reusing connection to ingester");
+
+            Ok(connection.clone())
+        } else {
+            debug!(%ingester_address, "Connecting to ingester");
+
+            let connection = connection::Builder::new()
```
alamb, line 188: The functional part of this is tested in the end to end tests -- the actual caching behavior (aka that the connections are only created one) is not.  Any suggestions for testing welcome

querier/src/ingester/flight_client.rs
```diff
@@ -61,40 +71,33 @@ impl FlightClientImpl {
...
 
 #[async_trait]
 impl FlightClient for FlightClientImpl {
     async fn query(
         &self,
-        ingester_addr: &str,
+        ingester_addr: Arc<str>,
         request: IngesterQueryRequest,
     ) -> Result<Box<dyn QueryData>, Error> {
-        // TODO maybe cache this connection
```
alamb, line 96: ✅

crepererum:
> How will this handle querying multiple ingesters behind the same address?

What? Behind the same IPv4 address? Is that even possible? I mean each ingester has a unique service name (DNS) which should resolve to a unique IP, at least as long as K8s assumes that TCP connection are opaque by default. Sure for HTTP istio & Co can inspect the hostname sent by the client (IIRC the client must send this, because this is how people host multiple domains on a single IP+port since forever).

crepererum (approved):

tustvold:
I think Andrew is referring to if say you had a single k8s service fronting multiple ingesters. In this case IOx would only see the service IP and not the unique pod IPs of the ingesters.

The solutions here are:

* Rely on a proxy such as Envoy to handle the load balancing
* Use a headless service and use tower's load balancing, to distribute amongst the endpoints
* Don't put multiple ingesters behind the same network identity

All are workable, I think the last one is probably the simplest to understand and get working

alamb:
> I think Andrew is referring to if say you had a single k8s service fronting multiple ingesters. In this case IOx would only see the service IP and not the unique pod IPs of the ingesters.

Yes, thank you @tustvold, this is what I was inelegantly trying to say

alamb:
YOLO let's get this into production and see if the querier get happier

Issue #4585, opened 2022-05-12
Investigation: Compactor cannot catch up with Ingester


NGA-TRAN:
# Summary of issues:
1. Compact a partition of large files is so slow
2. Compactor buffers data of parquet files instead of streaming which lead to huge memory needed 
4. The current setup makes the concurrency level  is very low 1

# TODOs

- [x] #4883
- [x] #4869 
- [x] influxdata/conductor#969 
- [x] influxdata/influxdb_iox#4761 
- [x] influxdata/influxdb_iox#4762
- [x] influxdata/influxdb_iox#4695
- [x] influxdata/influxdb_iox#4696 
- [x] influxdata/influxdb_iox#4671 
- [x] influxdata/influxdb_iox#4596 - @NGA-TRAN 
- [x] influxdata/influxdb_iox#4324  - @domodwyer 
- [x] influxdata/influxdb_iox#4594 - @mkmik , @NGA-TRAN , @domodwyer 

# Observation
[Full Dashboard](https://grafana.prod01-eu-central-1.aws.influxdata.io/d/z7-FhLsnk/compactor?orgId=1)

Our current tests on one Compactor that compacts all files of 100 sequencers ingested by 4 Ingesters shows that the Compactor cannot catch up with the Ingester and, every compact cycle, there are more and more files to get compacted for each partition. This leads to a lot of deduplication work that makes the situation worse. Her are the numbers:

**This chart shows:**
1. More files per partition/sequencer for later cycle2. 
5. The cycle takes longer and longer to finish
![Screen Shot 2022-05-12 at 3 13 11 PM](https://user-images.githubusercontent.com/20850014/168153812-937691af-05f9-4c17-9b56-abbc5218b962.png)

**This charts is similar to the one above but shows the file size to get compacted per sequencer**
![Screen Shot 2022-05-12 at 3 16 25 PM](https://user-images.githubusercontent.com/20850014/168154105-beace22f-1431-4e2d-98e1-ce8d021216b8.png)

**This chars shows input compacted file rate for each sequencer every minute**
![Screen Shot 2022-05-12 at 3 17 06 PM](https://user-images.githubusercontent.com/20850014/168154435-f4a0e792-157e-4506-b9fb-b1be49a66f6a.png)

**Similarly, this chart shows how longer compacted output is created because it takes longer to finish a cycle**
[...]

PR #4589, opened 2022-05-12
chore: convert some debug to trace to reduce noises

NGA-TRAN:
This is for the investigation of #4585 


query/src/provider/deduplicate/algo.rs
```diff
@@ -72,15 +72,15 @@ impl RecordBatchDeduplicator {
         };
 
         let mut dupe_ranges = self.compute_ranges(&batch)?;
-        debug!("Finish computing range");
```
NGA-TRAN, line 75: Three debug lines in this function overwhelme Compactor log and make it less useful

query_tests/cases/in/several_chunks.sql
```diff
@@ -2,13 +2,14 @@
 
 -- validate we have access to information schema for listing system tables
 -- IOX_COMPARE: sorted
-SELECT * from h2o order by temp, other_temp, time;
+SELECT * from h2o;
```
NGA-TRAN, line 5: no need order by because we already have `-- IOX_COMPARE: sorted`

query_tests/cases/in/several_chunks.sql
```diff
@@ -2,13 +2,14 @@
...
 -- Plan will look like:
 --  . Two chunks (one parquet and one from ingester) neither overlap nor contain duplicate
 --     --> scan in one scan node
 --  . Three parquet chunks overlapped -> dedup without resort
 EXPLAIN SELECT * from h2o;
 
 -- Only selct fields and time
-select temp, other_temp, time from h2o order by 1, 2;
+-- IOX_COMPARE: sorted
```
NGA-TRAN, line 13: replace the order by with this `-- IOX_COMPARE: sorted`

mkmik (approved):

PR #4587, opened 2022-05-12
chore: Enable pprof feature back

mkmik:
We used to use the `/debug/pprof` endpoint in our testing environments. We likely lost that feature from the release image when we refactored our release build steps.

PR #4590, opened 2022-05-12
chore: Enable pprof feature back

mkmik:
In https://github.com/influxdata/influxdb_iox/pull/4587 we enabled the `pprof` feature in our release builds but this resulted in a build failure because the pprof crate needed `backtrace-rs` dep.

It turns out that we didn't catch this in the normal CI pipeline because, despite actually exercising a build with the `pprof` feature on, we also include the `heappy` feature, which just happens to also depend on `backtrace-rs` and thus the build succeeds.

This PR fixes the release build (which only gets excercised in `main`) by adding explicitly the `backtrace-rs` feature.

---

Tested manually with:

```
cargo build --package="influxdb_iox" --profile="release"  --no-default-features --features=aws,gcp,azure,jemalloc_replacing_malloc,tokio_console,pprof
```


NGA-TRAN (approved):

alamb:
Thank you for sorting this out @mkmik

PR #4620, opened 2022-05-17
fix: partition null column serialisation

### LINKED ISSUE #4272: Invalid data observed in write buffer
The ingester in NG is observing `WriteBufferErrorKind::InvalidData` errors reading from the kafka stream and dropping the message.

A quick logs scan shows:

```
level=error msg="unable to deserialise dml operation" error="WriteBufferError(InvalidData): failed to decode database batch: column with no values: ingestId" kafka_topic=iox-shared kafka_partition=10 potential_data_loss=true target="ingester::stream_handler::handler" location="ingester/src/stream_handler/handler.rs:230" time=1649682432928789433
```

repeating infrequently. Later:

```
level=error msg="unable to deserialise dml operation" error="WriteBufferError(InvalidData): failed to decode database batch: column with no values: f1v" kafka_topic=iox-shared kafka_partition=10 potential_data_loss=true target="ingester::stream_handler::handler" location="ingester/src/stream_handler/handler.rs:230" time=1649683885431231728
level=error msg="unable to deserialise dml operation" error="WriteBufferError(InvalidData): failed to decode database batch: column with no values: rpwr" kafka_topic=iox-shared kafka_partition=10 potential_data_loss=true target="ingester::stream_handler::handler" location="ingester/src/stream_handler/handler.rs:230" time=1649683885431400464
```

And:

```
level=error msg="unable to deserialise dml operation" error="WriteBufferError(InvalidData): failed to decode database batch: column with no values: f4l" kafka_topic=iox-shared kafka_partition=10 potential_data_loss=true target="ingester::stream_handler::handler" location="ingester/src/stream_handler/handler.rs:230" time=1649685303878728023
```

domodwyer:
As mentioned in slack, this is a "quick fix" for #4272.

The commit message explains the problem and implemented fix, but there are a few notable points:

1. The proto wire format allows a column to contain values of many types at the same time, as it is [expressed as a record/struct][proto] of all available types, rather than an enumeration which would constrain the wire format to match the semantics of the system internals - a record batch column can have only one type, but our wire protocol column can have many types.

2. Because we do not differentiate between NULL and "empty"/"zero value" in the proto messages, the current message is ambiguous - I don't think conflating zero values with NULL is the right strategy, and explicit NULLs would have prevented this bug. Fundamentally 0 != NULL when processing data.

3. Partitioning can yield a record batch that contains a column composed entirely of NULLs - should this column even exist in the resulting partitioned record batch? The overhead of keeping it around is low (effectively just a bitmap of N rows and some fixed, low cost memory overhead) but semantically it seems incorrect: if the original LP input was sent line-by-line, the NULL column would never have been created in the partition - it's an artefact born from the interaction of batching+partitioning.

## Suggestions / discussion points:

1. Make a breaking change to the wire encoding by swapping the `Column` [`Values`][proto] type to use [`Oneof`][oneof]. 

This would cause our wire proto to match the semantics of record batches - a column can have only one type, and it is specified even if no values are serialised.

We can probably do this gracefully by encoding the "message codec version" as a header in the kafka message, where the new version sends something like `version=2`, and defaulting to the current codec if the header is not specified during changeover. Or we can do a hard cutover which would be less faff.

2. Don't allow the partitioner to construct partitions composed of entirely NULL columns.

If the user sent two LP lines as a single batch:

```
cpu A=1 <day 1 timestmap>
cpu B=2 <day 2 timestamp>
```

This leads to the partitioned batch containing A also having an entirely NULL column B, and vice-versa, but if the lines were sent one-at-a-time, the "ghost column" B would never have been created. This seems semantically incorrect, but this is a more in-depth change to the partitioner / writer. It also means encoding/decoding to the wire protocol is lossy/not round-trippable, which can be a further source of confusion and inconsistencies, and makes me uncomfortable!  

I'm happy to take this up if it's what we decide, but it would be a low priority in my backlog of work.

Fixes #4272.

[proto]:https://github.com/influxdata/influxdb_iox/blob/1dc6d6deda83ee0583f90682ed1732bdbc8a5efa/generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto#L79-L88
[oneof]: https://developers.google.com/protocol-buffers/docs/proto3#oneof

---

* feat: is all set/unset BitSet query methods (f5753d7ea)

      Adds is_all_set() and is_all_unset() methods to the BitSet, to query the state
      of all bits in the bitmap.

* fix(pb): encoding entirely NULL columns (#4272) (1dc6d6ded)

      This commit changes the protobuf record batch encoding to skip entirely NULL
      columns when serialising. This prevents the deserialisation from erroring due
      to a column type inference failure.

      Prior to this commit, when the system was presented a record batch such as
      this:

                 | time       | A    | B    |
                 | ---------- | ---- | ---- |
                 | 1970-01-01 | 1    | NULL |
                 | 1970-07-05 | NULL | 1    |

      Which would be partitioned by YMD into two separate partitions:

                 | time       | A    | B    |
                 | ---------- | ---- | ---- |
                 | 1970-01-01 | 1    | NULL |

      and:

                 | time       | A    | B    |
                 | ---------- | ---- | ---- |
                 | 1970-07-05 | NULL | 1    |

      Both partitions would contain an entirely NULL column.

      Both of these partitioned record batches would be successfully encoded, but
      decoding the partition fails due to the inability to infer a column type from
      the serialised format which contains no values, which on the wire, looks like:

                 Column {
                     column_name: "B",
                     semantic_type: Field,
                     values: Some(
                         Values {
                             i64_values: [],
                             f64_values: [],
                             u64_values: [],
                             string_values: [],
                             bool_values: [],
                             bytes_values: [],
                             packed_string_values: None,
                             interned_string_values: None,
                         },
                     ),
                     null_mask: [
                         1,
                     ],
                 },

      In a column that is not entirely NULL, one of the "Values" fields would be
      non-empty, and the decoder would use this to infer the type of the column.

      Because we have chosen to not differentiate between "NULL" and "empty" in our
      proto encoding, the decoder cannot infer which field within the
      "Values" struct the column belongs to - all are valid, but empty.

      This commit prevents this type inference failure by skipping any columns that
      are entirely NULL during serialisation, preventing the deserialiser from
      having to process columns with ambiguous types.

### DIFF
--- a/arrow_util/src/bitset.rs
+++ b/arrow_util/src/bitset.rs
@@ -196,6 +196,36 @@ impl BitSet {
     pub fn bytes(&self) -> &[u8] {
         &self.buffer
     }
+
+    /// Return `true` if all bits in the [`BitSet`] are currently set.
+    pub fn is_all_set(&self) -> bool {
+        // An empty bitmap has no set bits.
+        if self.len == 0 {
+            return false;
+        }
+
+        // Check all the bytes in the bitmap that have all their bits considered
+        // part of the bit set.
+        let full_blocks = (self.len / 8).saturating_sub(1);
+        if !self.buffer.iter().take(full_blocks).all(|&v| v == u8::MAX) {
+            return false;
+        }
+
+        // Check the last byte of the bitmap that may only be partially part of
+        // the bit set, and therefore need masking to check only the relevant
+        // bits.
+        let mask = match self.len % 8 {
+            1..=8 => !(0xFF << (self.len % 8)), // LSB mask
+            0 => 0xFF,
+            _ => unreachable!(),
+        };
+        *self.buffer.last().unwrap() == mask
+    }
+
+    /// Return `true` if all bits in the [`BitSet`] are currently unset.
+    pub fn is_all_unset(&self) -> bool {
+        self.buffer.iter().all(|&v| v == 0)
+    }
 }
 
 /// Returns an iterator over set bit positions in increasing order
@@ -480,5 +510,78 @@ mod tests {
 
         v.get(5);
         v.set(5);
+    }
+
+    #[test]
+    fn test_all_set_unset() {
+        for i in 1..100 {
+            let mut v = BitSet::new();
+            v.append_set(i);
+            assert!(v.is_all_set());
+            assert!(!v.is_all_unset());
+        }
+    }
+
+    #[test]
+    fn test_all_set_unset_multi_byte() {
+        let mut v = BitSet::new();
+
+        // Bitmap is composed of entirely set bits.
+        v.append_set(100);
+        assert!(v.is_all_set());
+        assert!(!v.is_all_unset());
+
+        // Now the bitmap is neither composed of entirely set, nor entirely
+        // unset bits.
+        v.append_unset(1);
+        assert!(!v.is_all_set());
+        assert!(!v.is_all_unset());
+
+        let mut v = BitSet::new();
+
+        // Bitmap is composed of entirely unset bits.
+        v.append_unset(100);
+        assert!(!v.is_all_set());
+        assert!(v.is_all_unset());
+
+        // And once again, it is neither all set, nor all unset.
+        v.append_set(1);
+        assert!(!v.is_all_set());
+        assert!(!v.is_all_unset());
+    }
+
+    #[test]
+    fn test_all_set_unset_single_byte() {
+        let mut v = BitSet::new();
+
+        // Bitmap is composed of entirely set bits.
+        v.append_set(2);
+        assert!(v.is_all_set());
+        assert!(!v.is_all_unset());
+
+        // Now the bitmap is neither composed of entirely set, nor entirely
+        // unset bits.
+        v.append_unset(1);
+        assert!(!v.is_all_set());
+        assert!(!v.is_all_unset());
+
+        let mut v = BitSet::new();
+
+        // Bitmap is composed of entirely unset bits.
+        v.append_unset(2);
+        assert!(!v.is_all_set());
+        assert!(v.is_all_unset());
+
+        // And once again, it is neither all set, nor all unset.
+        v.append_set(1);
+        assert!(!v.is_all_set());
+        assert!(!v.is_all_unset());
+    }
+
+    #[test]
+    fn test_all_set_unset_empty() {
+        let v = BitSet::new();
+        assert!(!v.is_all_set());
+        assert!(v.is_all_unset());
     }
 }
--- a/mutable_batch_pb/tests/encode.rs
+++ b/mutable_batch_pb/tests/encode.rs
@@ -1,5 +1,6 @@
 use arrow_util::assert_batches_eq;
-use mutable_batch::MutableBatch;
+use data_types::{PartitionTemplate, TemplatePart};
+use mutable_batch::{writer::Writer, MutableBatch, PartitionWrite, WritePayload};
 use mutable_batch_pb::{decode::write_table_batch, encode::encode_batch};
 use schema::selection::Selection;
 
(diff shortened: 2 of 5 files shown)


mkmik:
I like both (1) (switch to oneof) and (2) avoid sending unnecessary null columns as part of the partition-aware batcher.

> We can probably do this gracefully by encoding the "message codec version" as a header in the kafka message, where the new version sends something like version=2, and defaulting to the current codec if the header is not specified during changeover. Or we can do a hard cutover which would be less faff.

or we can do without an explicit version field, and put both the current fields and the new oneof block in the same message and some logic to convert old messages in the new format right after parse (and then drop this adapter code after a month)

arrow_util/src/bitset.rs
```diff
@@ -192,6 +192,32 @@ impl BitSet {
...
+        let full_blocks = (self.len / 8).saturating_sub(1);
+        if !self.buffer.iter().take(full_blocks).all(|&v| v == u8::MAX) {
+            return false;
+        }
+
+        // Check the last byte of the bitmap that may only be partially part of
+        // the bit set, and therefore need masking to check only the relevant
+        // bits.
+        let mask = (1 << (self.len % 8)) - 1; // LSB mask
+        self.buffer[self.buffer.len().saturating_sub(1)] == u8::MAX & mask
```
mkmik, line 214: no need and with u8::MAX since `self.len % 8` is  `< 8`

```suggestion
        self.buffer.last.unwrap() == (1 << (self.len % 8)) - 1;
```

arrow_util/src/bitset.rs
```diff
@@ -192,6 +192,32 @@ impl BitSet {
...
+    }
+
+    /// Return `true` if all bits in the [`BitSet`] are currently unset.
+    pub fn is_all_unset(&self) -> bool {
+        self.buffer.iter().all(|&v| v == 0)
```
alamb, line 219: Should this also mask off the last byte (as in if the `BitSet` length is 4, can the last byte be `0b10000000` for example?

mkmik (approved):
looks good to me

(minor bikeshedding comment notwithstanding)

mutable_batch_pb/tests/encode.rs
```diff
@@ -36,3 +37,141 @@ fn test_encode_decode() {
...
+    write_table_batch(&mut batch, &encoded).unwrap();
+
+    let expected = &[
+        "+---+--------------------------------+",
+        "| A | time                           |",
+        "+---+--------------------------------+",
+        "| 1 | 1970-01-01T00:00:00.000000160Z |",
+        "+---+--------------------------------+",
+    ];
+    assert_batches_eq!(expected, &[batch.to_arrow(Selection::All).unwrap()]);
```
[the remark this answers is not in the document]
alamb: What do you mean "disappear"? This seems correct and consistent to me

mutable_batch_pb/tests/encode.rs
```diff
@@ -36,3 +37,141 @@ fn test_encode_decode() {
...
+//
+//    | time       | A    | B    |
+//    | ---------- | ---- | ---- |
+//    | 1970-07-05 | NULL | 1    |
+//
+// In both partitions, one column is composed entirely of NULL values.
+//
+// Encoding each of these partitions succeeds, but decoding the partition fails
+// due to the inability to infer a column type from the serialised format which
+// contains no values:
```
alamb, line 72:
```suggestion
// Encoding each of these partitions succeeds, but decoding the partition could
// fail if the all NULL column was included, duue to the inability to infer a column 
// type from the serialised format which contains no values:
```

alamb (approved):
1. I think the approach of skipping entirely NULL columns when encoding is a good strategy. 👍 

As for my thoughts discussion topics:

# Nicest Invariant 
I think the invariant that the  data in the ingester partitions is the same (e.g. columns in a partition) are the same if the line protocol is fed in one line at a time or multiple lines at a time is the easiest to reason about and makes the most sense.

It seems like this PR changes the code to actually ensure this is the case. 

# Specific comments on questions in PR description

> Make a breaking change to the wire encoding by swapping the Column

This is what I would suggest if it makes encoding/decoding simpler for column based clients. However, I think encoding all NULL columns isn't a great idea in general when converting from line protocol (because as you have noted there is no corresponding line protocol representation of such columns)

> This leads to the partitioned batch containing A also having an entirely NULL column B, and vice-versa, but if the lines were sent one-at-a-time, the "ghost column" B would never have been created.

I don't understand this comment. Column `B` would get created, just not in all partitions, which is what I think would happen if you fed this data in one line at a time


cc @tustvold  -- it would if you could review this code (especially the `BitSet` changes) , even if it is after merge

mutable_batch_pb/tests/encode.rs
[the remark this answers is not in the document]
domodwyer: If you were to iterate over the columns in `batch`, you would observe `time`, `A` and column `B` - this differs from the rendered output, as `B` is all NULL. I'm just referring to the

domodwyer:
> I don't understand this comment. Column B would get created, just not in all partitions, which is what I think would happen if you fed this data in one line at a time

This is true if the record batch passes across the API boundary - that is, if we serialise it to protobuf in the router and send it to the ingester. It is a property of the _encoding_ and not of the data structures within the system.

If the batch were never to be serialised to protobuf (i.e. different serialisation format, sent as flight record batch response(?), or simply never transmitted across an API boundary) then column `B` would still exist in the partition. Put another way, if the all-in-one mode skipped serialising messages to proto and passed batches through a channel to the "ingester" side instead, column `B` would never be cleaned up and would make it's way into parquet files.

I guess I'm saying it would be nice if not having entirely NULL columns in a record batch was a property of the data structures, rather than relying on it as a side effect of passing through encoding.

arrow_util/src/bitset.rs
domodwyer: Good point - I originally had this check before the "full bytes" check, but moved it to the end to maintain a linear memory access and didn't change the masking 👍

arrow_util/src/bitset.rs
alamb: The reasoning I believe is to be compatible with the arrow bitset (aka use the same underlying format as the arrow validity bitmap)

At the time @tustvold  created this code I think the idea of reusing what was in Arrow was not feasible (due to API or something)

alamb:
> I guess I'm saying it would be nice if not having entirely NULL columns in a record batch was a property of the data structures, rather than relying on it as a side effect of passing through encoding.

Got it -- I think the core impedence mismatch is mapping line protocol to a tabular representation.  Specifically, it is not possible to have "an all null column" in line protocol but it is possible to have such a column in a tabular representation. So somewhere we'll have to deal with the discremancy.

Issue #4616, opened 2022-05-17
Querier Cache Metrics


crepererum:
The querier should emit metrics about its cache state (number of cached entries) and as well cache changes (e.g. load and evict).

crepererum:
done

PR #4632, opened 2022-05-18
feat: add LRU cache metrics

crepererum:
Ref #4616 (doesn't close the issue, we need MORE metrics!)


cache_system/src/backend/lru.rs
```diff
@@ -44,15 +54,23 @@
...
+//!     "my_pool",
+//!     limit,
+//!     time_provider,
+//!     metric_registry,
+//! ));
 //!
 //! // set up first pool user: a u64->String map
 //! #[derive(Debug)]
-//! struct Provider1 {}
+//! struct Estimator1 {}
```
alamb, line 71: 👍  I like the name change

cache_system/src/backend/lru.rs
```diff
@@ -239,31 +258,106 @@ use super::{
     CacheBackend,
 };
 
+#[derive(Debug)]
+struct MeasuredT<T> {
```
alamb, line 262: Maybe it is worth an explanatory comment here (to save readers from having to read the `impl` to get the gist. Something like :
```suggestion
/// Wrapper around something that can be converted into `u64`
/// to enable emitting metrics.
struct MeasuredT<T> {
```

cache_system/src/backend/lru.rs
```diff
@@ -239,31 +258,106 @@ use super::{
     CacheBackend,
 };
 
+#[derive(Debug)]
```
alamb, line 261: does this now require that all resources are u64 (or maybe only convertible into u64)? If so, perhaps we could simplify the code by removing the generics and simply saying `u64` in the various places

cache_system/src/backend/lru.rs
```diff
@@ -390,18 +491,34 @@ where
...
     ) -> Self {
         assert!(inner_backend.is_empty(), "inner backend is not empty");
 
+        let metric_count = pool
+            .metric_registry
+            .register_metric::<U64Gauge>(
+                "cache_lru_member_count",
+                "Number of entries for a given LRU cache pool member",
+            )
+            .recorder(&[("pool", pool.name), ("member", id)]);
```
alamb, line 505: Am I right in assuming the metric that will come out will have `"pool" == "Ram Butes"` or something similar?

cache_system/src/backend/lru.rs
```diff
@@ -1112,6 +1260,122 @@ mod tests {
...
+                .unwrap(),
+            &Observation::U64Gauge(6)
+        );
+        assert_eq!(
+            reporter
+                .metric("cache_lru_member_count")
+                .unwrap()
+                .observation(&[("pool", "pool"), ("member", "id")])
+                .unwrap(),
+            &Observation::U64Gauge(2)
```
alamb, line 1367: The 2 are `a` and `d`, correct?

alamb (approved):

cache_system/src/backend/lru.rs
alamb: (this is a comment for some future PR)

crepererum: Convertible, yes, but this conversion can be lossy since it's solely for metrics.

Let's wait a bit w/ that simplification until we have the on-disk pools (for parquet files) so we can see what we need from the resources. I can imagine that either we end up with "everything is bytes, counted as u64 anyways" or "we actually need some features in the `Resource` trait".

cache_system/src/backend/lru.rs
crepererum: `pool = "Ram". The unit will only currently be attached as a label, but I could try to pack this into the name as well. That would involve some nasty Box-leaking, since the names must be `static` and rustc doesn't offer a way to concat static strings.

cache_system/src/backend/lru.rs
crepererum: Added a comment: it's b and d, since a got evicted (see comment a few lines up where we add d).

cache_system/src/backend/lru.rs
alamb: The important thing is that `"Ram"` is included in the metric. Sounds good 👍

PR #4636, opened 2022-05-18
refactor: remove unused max_row_group_size

domodwyer:
This isn't used 🧹 

---

* refactor: remove unused max_row_group_size (7a8e6d1a3)

      The Parquet writer references an unused max_row_group_size property in the
      parquet file metadata.


alamb:
> We may want to revisit this once https://github.com/influxdata/influxdb_iox/issues/4324 is complete, as IIRC the default is rather large and will cause the ParquetWriter to buffer up potentially large amounts of data.

To be clear, what should be revisited is the actual max row group size. Having a setting not connected to anything is misleading and I am fully in support of removing it

Issue #4634, opened 2022-05-18
Avoid querying catalog for each query


alamb:
__Use case:__
We have noticed internally (https://github.com/influxdata/k8s-idpe/pull/7577) that the queriers are making a large number of requests to the catalog.

__Proposal:__
This ticket covers:
1. Design some way to avoid contacting the catalog on each query (aka avoid at least the metadata) -- "cache"
2. Test / implement said cache

__Current behaviour:__
At least one issue  is that the queriers are running a catalog query to determine parquet files for each query.

__Desired behaviour:__
Catalog requests from the querier to postgres significantly reduced

Implementation notes from @crepererum 

> Here is the code:

https://github.com/influxdata/influxdb_iox/blob/4bd899369e0427b14a7b1cdbad502a227952e3e2/querier/src/table/mod.rs#L134
mod.rs

> IIRC all other access to the catalog are shielded by some cache

>this needs to consider the tombstones as well (hence the single transaction)

__Alternatives considered:__
Describe other solutions or features you considered.

alamb:
Proposed design document: https://docs.google.com/document/d/1gszVCQkvY8A40dsNRWJThxdPGV2jsh3u1DolsGpj9XU/edit#

Spoiler:

```text
        Information about                                    
        parquet files and                                    
        Tombstones fetched                                   
        from Cache                                           
                                                             
┌─────────────┐             ┌────────────────┐               
│Catalog Cache│─ ─ ─ ─ ─ ─ ▶│  IOx Querier   ├ ─ ─ ─ ─ ─ ─ ─▶
└─────────────┘             └────────────────┘               
       ▲                             ▲                       
                                                             
       │ Cache is refreshed          ┃                       
         from Catalog                                        
       │                             ┃                       
┌─────────────┐              ┌───────────────┐               
│   Catalog   │              │ IOx Ingester  │               
└─────────────┘              └───────────────┘               
```

PR #4637, opened 2022-05-18
chore: Remove dead code from QueryDatabase

alamb:
While working on https://github.com/influxdata/influxdb_iox/issues/4634 I was confused about what appeared to be a `QuerierNamespace` cache.

Then it turned out it was never read. So remove it to avoid future confusion.


carols10cents (approved):

crepererum (approved):
Good catch, namespaces are constructed on-demand these days.

domodwyer (approved):
🧹

PR #4645, opened 2022-05-19
chore: Update to Rust 1.61

carols10cents:
🎉

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,6 +131,7 @@ jobs:
   doc:
     docker:
       - image: quay.io/influxdb/rust:ci
+    resource_class: medium+ # use of a smaller executor runs out of memory
     environment:
       # Disable incremental compilation to avoid overhead. We are not preserving these files anyway.
       CARGO_INCREMENTAL: "0"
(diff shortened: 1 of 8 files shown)


carols10cents:
OOPS forgot about clippy 😅

carols10cents:
Ok, it looks like there's something wrong with... jemalloc? in the doc job? I don't know how... that's a tomorrow problem...

shepmaster:
> cc: fatal error: Killed signal terminated program cc1
>  compilation terminated.

This feels a lot like "ran out of memory and the OS killed the process". There's another error as well which [also is suggested as a memory issue](https://stackoverflow.com/a/40132053/155423):

> {standard input}: Assembler messages:
> {standard input}:308625: Warning: end of file not at end of a line; newline inserted
> {standard input}: Error: open CFI at the end of file; missing .cfi_endproc directive

We could try:
1. reducing build parallelism
2. increasing memory on the builder

shepmaster:
And after bumping to 6GB:

image

alamb:
Doc builder change makes sense to me. I say :shipit:

alamb (approved):

PR #4650, opened 2022-05-20
feat: avoid catalog access on each query

alamb:
~Draft as I am breaking this PR up into smaller pieces for easier review. Will mark as ready for review when we need to turn on the code~

~Waiting on https://github.com/influxdata/influxdb_iox/pull/4701~

# Rationale
Closes: https://github.com/influxdata/influxdb_iox/issues/4634

See design doc for more details: https://docs.google.com/document/d/1gszVCQkvY8A40dsNRWJThxdPGV2jsh3u1DolsGpj9XU/edit#heading=h.p96gfd4b8tda

Basically, the catalog (unsurprisingly) is a bottleneck during query time so we shouldn't be querying it over and over again. 

# Changes
This PR adds caching for parquet files and tombstones in the querier along with the appropriate invalidation logic based on the response from the ingesters. 

# PR Chain
To avoid one massive PR I am breaking this one up into several pieces. Here is a list:

- [x] https://github.com/influxdata/influxdb_iox/pull/4644
- [x] https://github.com/influxdata/influxdb_iox/pull/4652
- [x] https://github.com/influxdata/influxdb_iox/pull/4661
- [x] https://github.com/influxdata/influxdb_iox/pull/4663
- [x] https://github.com/influxdata/influxdb_iox/pull/4685
- [x] https://github.com/influxdata/influxdb_iox/pull/4701

# Follow on work related to deletes
- [ ] https://github.com/influxdata/influxdb_iox/issues/4694
- [ ] https://github.com/influxdata/influxdb_iox/issues/4705

# Manual Testing

Here is an end to end manual test showing the behavior we want

I added some logging to the catalog locally:

```diff
diff --git a/iox_catalog/src/postgres.rs b/iox_catalog/src/postgres.rs
index fd2e88519..46679499e 100644
--- a/iox_catalog/src/postgres.rs
+++ b/iox_catalog/src/postgres.rs
@@ -1590,6 +1590,8 @@ WHERE table_id = $1 AND to_delete IS NULL;
         &mut self,
         table_id: TableId,
     ) -> Result<Vec<ParquetFileWithMetadata>> {
+        warn!(%table_id, "Asking postgres for catalog files");
+
         sqlx::query_as::<_, ParquetFileWithMetadata>(
             r#"
 SELECT *
```

Then I ran IOx like this
```shell
LOG_FILTER=warn nice cargo run -- run all-in-one --catalog-dsn=postgresql://localhost:5432/alamb
```

And ran 5 queries like this in another shell
```
[...]

PR #4652, opened 2022-05-20
feat: Add `SharedBackend` to cache system

alamb:
# Rationale
Part of https://github.com/influxdata/influxdb_iox/issues/4634

I need some way to tell the cache system that an entry should be removed when the querier discovers from the ignester that there are new parquet files it doesn't yet know about

Please see see https://github.com/influxdata/influxdb_iox/pull/4650 for how this is used.

# Changes
* Add `SharedBackend` that allows a backend to be used by a Cache but also directly


cache_system/src/backend/shared.rs
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    /// "remove" a key (aka remove it from the shared backend) if the
+    /// specified predicate is true. If the key is removed return
+    /// true, otherwise return false
+    ///
+    /// Note that the predicate function is called while the lock is
+    /// held (and thus the inner backend can't be concurrently accessed
+    pub fn remove_if<P>(&self, k: &K, predicate: P) -> bool
```
alamb, line 39: This is the mechanism I plan to use to clear the cache if the querier discovers the ingester has created new files and its cache is stale

cache_system/src/backend/shared.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use super::*;
+
+    #[test]
+    fn test_generic() {
+        crate::backend::test_util::test_generic(|| SharedBackend::new(test_backend()))
```
alamb, line 91: ❤️  for this generic test @crepererum

crepererum (approved):
I like the new generic building block :)

PR #4656, opened 2022-05-23
test: add query test scenario w/ missing columns in different chunks

crepererum:
I originally hoped that this would trigger a panic but it doesn't. The test is useful anyways.


query_tests/src/cases.rs
```diff
@@ -134,28 +134,34 @@ async fn test_cases_two_sql() {
 // Tests from "several_chunks.sql",
 async fn test_cases_several_chunks_sql() {
     let input_path = Path::new("cases").join("in").join("several_chunks.sql");
-    let output_path = make_output_path(&input_path).unwrap();
-    let expected_path = input_path.with_extension("expected");
+    let mut runner = Runner::new();
+    runner
+        .run(input_path)
+        .await
+        .expect("test failed");
```
mkmik, line 141: I'm always confused about the polarity of the expect message. Does this mean that we expected the test to fail but it didn't?

mkmik (approved):
Test looks good. I have one question which can be answered async

query_tests/src/cases.rs
crepererum: :shrug: I have the same question :smile:

query_tests/cases/in/two_chunks_missing_columns.sql
```diff
@@ -0,0 +1,5 @@
+-- Basic query tests
+-- IOX_SETUP: TwoChunksMissingColumns
+
+-- query data
+SELECT * from "table" order by time;
```
NGA-TRAN, line 5: I am sorry but I do not understand the purpose of the test. It seems from the name of `TwoChunksMissingColumns`, we want to test how things work with 2 missing columns but the `select * from "table" order by time;` seems not do that job. Do you mean to select specific column names that do not exist in the table or the `where` condition includes missing columns?

PR #4659, opened 2022-05-23
feat: enable debugging of failed querier->ingester requests

crepererum:
- extend `query-ingester` CLI to allow usage of predicates
- on failed requests: log all information that required for the CLI
- test the "ingester fails" scenario


clap_blocks/src/ingester.rs
```diff
@@ -62,4 +62,12 @@ pub struct IngesterConfig {
         default_value = "300"
     )]
     pub persist_partition_cold_threshold_seconds: u64,
+
+    /// Sets how often `do_get` flight requests should panic for testing purposes.
```
alamb, line 66: can you please clarify what metric is used? Is it "every N tests panic"?

influxdb_iox/src/commands/query_ingester.rs
```diff
@@ -27,8 +34,18 @@ pub struct Config {
...
     /// The columns to request
+    #[clap(
+        long = "--columns",
+        default_value = "",
+        multiple_values = true,
+        use_value_delimiter = true
+    )]
     columns: Vec<String>,
 
+    /// Predicate in base64 protobuf encoded form.
```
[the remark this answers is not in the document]
alamb: It might be nice to give some hints to someone in the future about how to obtain such a value

I also wonder if it would be useful to use the code in the `storage` command to parse sql-like predicates and encode them into Predicates

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -159,3 +161,102 @@ async fn table_not_found_on_ingester() {
...
+#[tokio::test]
+async fn ingester_panic() {
+    test_helpers::maybe_start_logging();
+    let database_url = maybe_skip_integration!();
+
+    let table_name = "the_table";
+
+    // Set up the cluster  ====================================
+    let router_config = TestConfig::new_router(&database_url);
+    let ingester_config =
```
alamb, line 174:
```suggestion
    // can't use standard mini cluster here as we setup the querier to panic
    let ingester_config =
```

ingester/src/server/grpc.rs
```diff
@@ -32,19 +39,26 @@ use write_summary::WriteSummary;
 #[derive(Debug, Default)]
 pub struct GrpcDelegate<I: IngestHandler> {
     ingest_handler: Arc<I>,
+
+    /// Sets how often `do_get` flight requests should panic for testing purposes.
```
alamb, line 43: again, here I think it would help to explain how the units are to be interpreted ("every N requests"?)

querier/src/ingester/mod.rs
```diff
@@ -247,9 +247,25 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<Arc<Ingeste
...
     }
-    let mut perform_query = query_res.context(RemoteQuerySnafu {
-        ingester_address: ingester_address.as_ref(),
-    })?;
+    let mut perform_query = query_res
+        .context(RemoteQuerySnafu {
+            ingester_address: ingester_address.as_ref(),
+        })
+        .map_err(|e| {
+            // generate a warning that is sufficient to replicate the request using CLI tooling
```
alamb, line 255: ❤️

querier/src/ingester/mod.rs
```diff
@@ -322,6 +338,28 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<Arc<Ingeste
     Ok(ingester_partitions)
 }
 
+fn encode_predicate_as_base64(predicate: &Predicate) -> String {
```
alamb, line 341: it is somewhat unfortunate that the encoding / decoding of Predicate to base64 proto are in entirely different crates / modules.

i wonder if we could put them together somewhere, But I admit there is no obviously good crate -- maybe make a new one `grpc_predicate` or something?

alamb (approved):
👍  looks useful. Nice work

The tests in this PR are 🏆

clap_blocks/src/ingester.rs
crepererum: The first N, fixed.

influxdb_iox/src/commands/query_ingester.rs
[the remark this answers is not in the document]
crepererum: > I also wonder if it would be useful to use the code in the storage command to parse sql-like predicates and encode them into Predicates

Yeah, that would be helpful, but I didn't want to got down that rabbit hole.

querier/src/ingester/mod.rs
crepererum: moved (at least from/to the protobuf version, the last mile and a bit of plumming remains)

alamb:
Reviewer hall of shame: I missed the error wasn't actually logged https://github.com/influxdata/influxdb_iox/pull/4672 😞

PR #4661, opened 2022-05-23
feat: Add `ParquetFiles` cache and memory size estimation for ParquetMetadata

alamb:
feat: Add `ParquetFiles` cache

# Rationale
Part of https://github.com/influxdata/influxdb_iox/issues/4634

The goal is to avoid asking the catalog for parquet files on each

Please see https://github.com/influxdata/influxdb_iox/pull/4650 for how this is used.

# Changes
* Add parquet file cache, following the model of existing caches in the querier `CatalogCache`
* Tests for same


parquet_file/src/metadata.rs
```diff
@@ -668,6 +688,14 @@ impl DecodedIoxParquetMetaData {
 
         Ok(column_summaries)
     }
+
+    /// Estimate the memory consumption of this object and its contents
+    pub fn size(&self) -> usize {
+        // This is likely a wild under count as it doesn't include
+        // memory pointed to in the `ParquetMetaData` structues.
+        // Feature tracked in arrow-rs: https://github.com/apache/arrow-rs/issues/1729
```
alamb, line 696: FYI @domodwyer  -- I am not sure what precisely you have planned for parquet metadata, but here is another practical implication of caching so much metadata

iox_catalog/src/metrics.rs
```diff
@@ -149,6 +149,7 @@ macro_rules! decorate {
 
             $(
                 async fn $method(&mut self, $($arg : $t),*) -> Result<$out> {
+                    //println!("Calling {}::{}", std::any::type_name::<Self>(), stringify!($method));
```
mkmik, line 152: this looks like a leftover from some local debugging sessions

parquet_file/src/chunk.rs
```diff
@@ -225,4 +225,17 @@ impl DecodedParquetFile {
             iox_metadata,
         }
     }
+
+    /// Estimate the memory consumption of this object and its contents
+    pub fn size(&self) -> usize {
+        // Note size_of_val is the size of the Azrc
```
mkmik, line 231:
```suggestion
        // Note size_of_val is the size of the Arc
```

test_helpers/src/lib.rs
```diff
@@ -166,3 +166,32 @@ macro_rules! assert_error {
...
+        {
+            let actual = $ACTUAL;
+            let expected = $EXPECTED;
+            let epsilon = $EPSILON;
+            // determine how far apart they actually are
+            let delta = if actual < expected {
+                expected - actual
+            } else {
+                actual - expected
+            };
```
mkmik, line 186:
```suggestion
            // determine how far apart they actually are
            let delta = actual.abs_diff(expected);
```

parquet_file/src/chunk.rs
```diff
@@ -225,4 +225,17 @@ impl DecodedParquetFile {
             iox_metadata,
         }
     }
+
+    /// Estimate the memory consumption of this object and its contents
+    pub fn size(&self) -> usize {
```
crepererum, line 230: Does this include `self`? Throughout the codebase we normally do that, so I'm wondering why `size_of_val(self)` is not included in the sum :thinking:

querier/src/cache/parquet_file.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Clear the parquet file cache if it does not know about data up
+    /// to max_parquet_sequence_number.
+    ///
+    /// If it does not know about max_parquet_sequence_number it means
+    /// the ingester has written new data to the catalog and we need
+    /// to update our knowledge of that.
+    ///
+    /// Returns true if the cache was cleared.
+    pub fn expire_if_unknown(
```
crepererum, line 186: "if unknown" confuses me. Maybe `expire_if_outdated` or `expire_on_newly_persisted_files` or something.

querier/src/cache/parquet_file.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Mark the entry for table_id as expired (and needs a refresh)
+    pub fn expire(&self, table_id: TableId) {
+        self.backend.remove_if(&table_id, |_| true);
+    }
+
+    /// Clear the parquet file cache if it does not know about data up
+    /// to max_parquet_sequence_number.
+    ///
+    /// If it does not know about max_parquet_sequence_number it means
```
crepererum, line 181: Who's "it" in "If it does not know"? The caller (providing `max_parquet_sequence_number`) or the cache? The reason I'm wondering is that I'm looking for a documentation why the `max_parquet_sequence_number` is optional and what `None` would mean in that context (I think you're directly passing the ingester information here and want a single place were the invalidation logic is handled, but it would be nice to document it anyways).

querier/src/cache/parquet_file.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    /// Get list of cached parquet files, by table id
+    pub async fn get(&self, table_id: TableId) -> Arc<CachedParquetFiles> {
+        self.cache.get(table_id).await
+    }
+
+    /// Mark the entry for table_id as expired (and needs a refresh)
+    pub fn expire(&self, table_id: TableId) {
+        self.backend.remove_if(&table_id, |_| true);
+    }
```
crepererum, line 176: Do we need this?

querier/src/cache/parquet_file.rs
```diff
@@ -0,0 +1,436 @@
...
+
+        let cache = make_cache(&catalog);
+        let table_id = table.table.id;
+        assert_eq!(
+            cache.get(table_id).await.ids(),
+            ids(&[&tfile1_2, &tfile1_3])
+        );
+
+        // simulate request with sequence number 2
+        // should not expire anything
```
crepererum, line 338: > should not expire anything

How's this asserted? You assert the cache content but not that the catalog was used again. I think you might need to assert the metric counter here.

iox_catalog/src/metrics.rs
alamb: Yes it was -- I will remove it. I vacillated about leaving it in.

crepererum (approved):
Few minor nits, looks good overall.

I see you made your way throw the cache framework as well as `iox_tests::util`. We've talked about the first but this also holds for the second one: if you need to extend / tune / refactor it, feel free to do so. I've designed both and I might have used some weird/complicated/undocumented patterns here and there.

parquet_file/src/chunk.rs
alamb: I guess my thinking was that size_of_val(self) will include the actual inlined size of `ParquetFile`, `DecodedIoxParquetMetaData`, and `IoxMetadata`, so the code would end up looking like:

```rust
  size_of_val(self) + 
  self.parquet_file.size() -
  size_of::<ParquetFile>() +
  self.decoded_metadata.size() -
  size_of::<IoxMetadata> + 
  self.iox_metadata.size() -
  size_of::<IoxMetadata>
```

While this is somewhat messy, it is probably best to be consistent, so I will make it so

querier/src/cache/parquet_file.rs
alamb: good eyes -- it is used in a test that is part of https://github.com/influxdata/influxdb_iox/pull/4650 -- I will remove it here for nwo

parquet_file/src/chunk.rs
alamb: Improved (made consistent to use `size_of_val(self)` in ed61e0363)

querier/src/cache/parquet_file.rs
alamb: made test only in e17d8b446

querier/src/cache/parquet_file.rs
alamb: yes, I will do so

querier/src/cache/parquet_file.rs
alamb: I like `expire_on_newly_persisted_files` 👍

parquet_file/src/chunk.rs
mkmik: There are tools like https://github.com/Aeledfyr/deepsize

I've been experimenting with such crates when I was exploring our options for heap profiling

alamb:
> as well as iox_tests::util. We've talked about the first but this also holds for the second one: if you need to extend / tune / refactor it, feel free to do so. I've designed both and I might have used some weird/complicated/undocumented patterns here and there.

Thanks @crepererum ! 👍 

What I was thinking would be helpful would be to the knowledge about the metric name into the `TestCatalog` eventually

I was actually thinking about changing tests that look like:

```rust
 assert_histogram_metric_count(&catalog.metric_registry, METRIC_NAME, 1);
```

To look more like:
```rust
assert_eq!(catalog.num_parquet_file_loads(), 1);
```

And embed the metric name in  `TestCatalog::num_parquet_file_loads()`

But I then saw that `assert_histogram_metric_count` was what was used in the rest of the cache tests and figured that consistency was better than brevity. Maybe I'll refactor all call sites at some point

querier/src/cache/parquet_file.rs
```diff
@@ -0,0 +1,449 @@
...
+            self.files.iter().map(|f| f.size()).sum::<usize>()
+    }
+
+    /// Returns the greatest parquet sequence number stored in this cache entry
+    pub(crate) fn max_parquet_sequence_number(&self) -> Option<SequenceNumber> {
+        self.files
+            .iter()
+            .map(|f| f.parquet_file.max_sequence_number)
+            .max()
+    }
```
crepererum, line 83: @alamb Sorry that I'm a bit late, but I think this solution here might lead to cache trashing under certain conditions involving deletes. To repeat what I've said [here](https://docs.google.com/document/d/1gszVCQkvY8A40dsNRWJThxdPGV2jsh3u1DolsGpj9XU/edit?disco=AAAAZZc2MNA):

1. sequ. number 1: ingester gets write
2. ingester persists, max persisted sequ. number is 1
3. querier executes some queries, caches parquet file at sequence number 1.
4. sequ. number 2: ingester gets write
5. sequ. number 3: ingester gets delete, covering all the data from step 4
6. ingester persists, max persisted sequ. number is 3, but no parquet file was written (step 5 deletes step 4)
7. querier executes query, gets max perstisted sequ. number 3, looks at its cache, determines it only knows data up to sequ. number 1 and refreshes the cache
8. repeat step 7, cache will be invalidated again and again

The workaround is simple: in the same transaction that you use to fetch the parquet files, also fetch the max persisted sequ. number and use that as a marker for cache freshness.

parquet_file/src/chunk.rs
alamb: That looks interesting -- though I wonder if you need to derive `DeepSizeOf` for all sub types as well (like you have to do for `Derive(Serialize)`, for example.

If so we would still end up having to do stuff like https://github.com/apache/arrow-rs/issues/1729 🤔

mkmik: yeah, one of the problems with this kind of tools is that they can (due to how traits work in rust):

1. give you a macro so it can generate the trait for your types
2. the crate can generate the trait for a fixed number of external crates

Often people re-implement such tools (e.g. the first one I saw was part of servo), so that they can implement the trait for all the dependencies they end up using.

alamb: 🤔  Defining our own trait and implementing it for external code is an intriguing idea 🤔

PR #4662, opened 2022-05-23
refactor: do not embed row count & min/max timestamps in IOxMetadata

domodwyer:
This PR removes the row count, min timestamp, and max timestamp from the `IoxMetadata` type serialised with protobuf, and embedded within a Parquet file's metadata.

These fields are redundant, being all fully derivable from the Parquet metadata / column statistics, and preclude a reasonable streaming compaction implementation as you must first inspect the data to determine their values _before_ serialising the data to correctly populate the `IoxMetadata` type embedded within the file.

This PR starts by adding two integration tests to ensure correct behaviour was not violated by these changes. Of particular interest is the derivation of the Parquet file's metadata that is inserted into the catalog, covered by `test_derive_parquet_file_params`.

---

* test: Parquet metadata integration test (71555ee55)

      Adds two integration tests covering validation of the embedded IOx metadata
      within the Parquet file metadata, and validation of the derived 
      ParquetFileParams metadata used to populate the catalog.

* refactor: remove row_count from IoxMetadata (a142a9eb5)

      Remove the redundant row_count from the IoxMetadata structure that is 
      serialised into the Parquet file.

      The reasoning is twofold:

         * The Parquet file's native metadata already contains a row count
         * Needing to know the number of rows up-front precludes streaming

* refactor: remove IoxMetadata min & max timestamp (2e6c49be8)

      Removes the min/max timestamp fields from the IoxMetadata proto structure
      embedded within a Parquet file's metadata.

      These values are redundant as they already exist within the Parquet column
      statistics, and precluded streaming serialisation as these removed min/max
      values were needed before serialising the file.


parquet_file/src/metadata.rs
```diff
@@ -389,11 +366,52 @@ impl IoxMetadata {
...
+
+        // Extract the min/max timestamps.
+        let (min_time, max_time) = match time_summary.stats {
+            Statistics::I64(stats) => {
+                let min = Timestamp::new(stats.min.expect("no min time statistic"));
+                let max = Timestamp::new(stats.max.expect("no max time statistic"));
+                (min, max)
+            }
+            _ => panic!("unexpected physical type for timestamp column"),
+        };
```
domodwyer, line 413: This block derives the values removed from the `IoxMetadata` type when generating the `ParquetFileParams` instance (used to populate the catalog).

parquet_file/tests/metadata.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+#[tokio::test]
+async fn test_derive_parquet_file_params() {
+    // A representative IOx data sample (with a time column, an invariant upheld
+    // in the IOx write path)
+    let data = vec![
+        to_string_array(&["bananas", "platanos", "manzana"]),
+        to_timestamp_array(&[
+            // NOTE: not ordered to ensure min/max extracted, not head/tail
```
alamb, line 93: 👍

parquet_file/tests/metadata.rs
```diff
@@ -0,0 +1,174 @@
...
+    let meta = IoxMetadata {
+        object_store_id: Default::default(),
+        creation_timestamp: Time::from_timestamp_nanos(1234),
+        namespace_id: NamespaceId::new(1),
+        namespace_name: "bananas".into(),
+        sequencer_id: SequencerId::new(2),
+        table_id: TableId::new(3),
+        table_name: "platanos".into(),
+        partition_id: PartitionId::new(4),
+        partition_key: "potato".into(),
```
alamb, line 111: No 🍌 s?

alamb (approved):
👌 

For historical perspective, I believe the reason these fields are on `IOxMetadata `is that `IOxMetadata` used to be (also) stored separately from the Parquet files themselves (in the OG catalog)  so things like RowCount could not be gathered without actually fetching the parquet bytes.

Now that there is a new way to store parquet metadata in the NG postgres catalog, removing these fields makes sense to me)

cc @crepererum

parquet_file/src/metadata.rs
NGA-TRAN: Nice.

NGA-TRAN (approved):
This is really great that we do not have to compute row count, min and max time when storing (they are done implicitly for us) but we have them ready when reading.

PR #4664, opened 2022-05-23
fix: Make Kafka Partition related metric names potentially less confusing 

carols10cents:
While this couples us a little bit more than we'd ultimately like to Kafka, at least it's being clear about what this ID is or isn't, to hopefully fend off debugging confusion.

Connects to #4246. Connects to #4237.


data_types/src/lib.rs
```diff
@@ -1920,17 +1920,17 @@ impl TableSummary {
 /// Kafka partition ID plus offset
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub struct Sequence {
-    /// The sequencer id (kafka partition id)
-    pub sequencer_id: u32,
+    /// The kafka partition id
+    pub kafka_partition: KafkaPartition,
```
crepererum, line 1924: I understand the METRIC renames but this one here seems wrong. Why is this renamed? Now the write buffer (and everyone interacting w/ it) deals with a "Kafka" even though it doesn't touch Kafka at all.

write_buffer/src/mock.rs
```diff
@@ -151,7 +151,7 @@ impl MockBufferSharedState {
         let mut guard = self.writes.lock();
         let writes = guard.as_mut().expect("no sequencers initialized");
         let writes_vec = writes
-            .get_mut(&sequence.sequencer_id)
+            .get_mut(&(sequence.kafka_partition.get() as u32))
```
crepererum, line 154: Here's one example where the `Sequence.sequencer_id` doesn't make any sense at all.

crepererum (changes requested):

carols10cents:
@crepererum this is ready for re-review!

crepererum (approved):

PR #4677, opened 2022-05-24
test: panic handling, add compactor to end to end test harness

crepererum:
Let's ensure that our panic handling is wired up.

(Background: I was experimenting w/ a few new features in that domain and first wanted to make sure that the status quo works.)


test_helpers_end_to_end/src/mini_cluster.rs
```diff
@@ -123,6 +123,7 @@ impl MiniCluster {
         let router_config = TestConfig::new_router(&database_url);
         let ingester_config = TestConfig::new_ingester(&router_config);
         let querier_config = TestConfig::new_querier(&ingester_config);
+        let compactor_config = TestConfig::new_compactor(&ingester_config);
```
alamb, line 126: here comes the compactor!

PR #4678, opened 2022-05-24
feat: streaming compaction

### LINKED ISSUE #4324: Support streaming compaction in NG

domodwyer:
This PR changes our compaction process to eliminate any in-memory buffering of resultant `RecordBatch` instances, instead directly streaming the compaction plan execution into a temporary Parquet file on disk before uploading to object storage.

Closes #4324.

---

* refactor: retry object store upload (6aa626ef8)

      Changes the Storage::upload() method to endlessly retry uploading the 
      generated Parquet file.

* feat: steaming compaction (8f05250c9)

      This commit changes the Compactor::compact() method to stream the RecordBatch
      instances directly to the parquet serialiser, before being uploaded directly
      to object storage.

* refactor: concurrent StreamSplitExec execution (c885b845d)

      Changes the compactor to consume both StreamSplitExec output partitions 
      concurrently.

      Practically speaking this means both Parquet files will be generated 
      concurrently, and uploaded to object store concurrently.

* revert: fix: compaction deadlock (8ff1a7379)

      This reverts commit 00b5c1b296fb7b6dbabd28744721e56b05061d0c.

      This change reverts the StreamSplitExec plan to using bounded, blocking 
      channels, with the possibility of deadlock added to the docs.

      This is now tolerable because of the concurrent consumption of both output
      partitions in the compactor.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -14,6 +14,7 @@ use data_types::{
     Timestamp, Tombstone, TombstoneId,
 };
 use datafusion::error::DataFusionError;
+use futures::stream::{FuturesUnordered, TryStreamExt};
 use iox_catalog::interface::{Catalog, Transaction, INITIAL_COMPACTION_LEVEL};
 use iox_query::{
     exec::{Executor, ExecutorType},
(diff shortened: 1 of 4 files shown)


compactor/src/compact.rs
```diff
@@ -1154,6 +1067,13 @@ mod tests {
         let count = catalog.count_tombstones_for_table(table.table.id).await;
         assert_eq!(count, 0);
 
+        // Verify the files were pushed to the object store
+        let object_store = catalog.object_store();
+        let list = object_store.list(None).await.unwrap();
+        let object_store_files: Vec<_> = list.try_collect().await.unwrap();
+        // Original + 2 compacted
+        assert_eq!(object_store_files.len(), 3);
```
NGA-TRAN, line 1075: 👍

compactor/src/compact.rs
```diff
@@ -1477,6 +1397,19 @@ mod tests {
...
+        let batches = batches
+            .into_iter()
+            .map(|v| async {
+                datafusion::physical_plan::common::collect(v.data)
+                    .await
+                    .expect("failed to collect record batches")
+            })
+            .collect::<FuturesOrdered<_>>()
+            .collect::<Vec<_>>()
+            .await;
```
NGA-TRAN, line 1411: This is really nice. Data is streaming!!!!

compactor/src/compact.rs
```diff
@@ -2878,14 +2756,21 @@ mod tests {
...
```
NGA-TRAN, line 2768: This is just minor. Since this data collection repeats a lot in many tests, I wonder whether we can refactor this to make it one line function call here.

compactor/src/compact.rs
```diff
@@ -2366,85 +2323,6 @@ mod tests {
         assert_eq!(groups.len(), 1);
     }
 
-    #[tokio::test]
-    async fn persist_adds_to_object_store() {
```
domodwyer, line 2370: This assertions made by this test are still made, they have moved into a different test, line 1070.

iox_query/src/exec/split.rs
```diff
@@ -257,8 +263,8 @@ impl StreamSplitExec {
         trace!("Setting up SplitStreamExec state");
 
         let input_stream = self.input.execute(0, context)?;
-        let (tx0, rx0) = tokio::sync::mpsc::unbounded_channel();
-        let (tx1, rx1) = tokio::sync::mpsc::unbounded_channel();
+        let (tx0, rx0) = tokio::sync::mpsc::channel(2);
+        let (tx1, rx1) = tokio::sync::mpsc::channel(2);
```
NGA-TRAN, line 267: Didn't you change from channel(2) to unbounded channels to avoid deadlock before?

parquet_file/src/storage.rs
```diff
@@ -94,6 +95,11 @@ impl ParquetStorage {
 
     /// Push `batches`, a stream of [`RecordBatch`] instances, to object
     /// storage.
+    ///
+    /// # Retries
+    ///
+    /// This method retries forever in the presence of object store errors. All
+    /// other errors are returned as they occur.
```
NGA-TRAN, line 102: This is fine for now and I have no other solution but this may be something for us to think in the future to scope with the fact that retrying forever may not work because of outage or the like in object store

NGA-TRAN (approved):
Very nice, @domodwyer . I just have a question

alamb (commented):
Screen Shot 2022-05-24 at 10 08 54 AM

I take the net removal of lines as a sign we are on the right path

compactor/src/compact.rs
```diff
@@ -462,33 +408,49 @@ impl Compactor {
...
+                    debug!(%object_store_id, "file uploaded to object store");
+
+                    Ok(CatalogUpdate::new(
+                        meta,
+                        file_size,
+                        parquet_meta,
+                        tombstones,
+                    ))
+                })
+                .collect::<FuturesUnordered<_>>()
```
alamb, line 451: This is quite neat, actually, that it will create as many output files as there are streams out of the `split` plan (which at the moment is 2) 👍  nice

parquet_file/src/storage.rs
domodwyer: The caller can still decide not to retry forever - if you were to call this with a timeout, for example, you still bound the duration it can take:

```rust
timeout(storage.upload(...), Duration::from_secs(5 * 60);
```

This would allocate 5 minutes to produce the parquet file and upload it, or fail. This may include multiple object store put retries in that time. The existing code performed an unbounded number of retries, just with `Backoff` - I wanted to avoid plumbing in the dependency here too as it is effectively just a loop in that case :+1:

Issue #1694, opened 2021-06-11
Better Panic Handling


tustvold:
Spawned out of a series of discussions with @alamb around the current pain points of error handling in IOx.

Within Rust there are two major mechanisms for bailing out if something has gone wrong, [Result](https://doc.rust-lang.org/std/result/) and panic. 

This is often portrayed as a binary choice, but perhaps it is more reasonable to consider them complementary techniques. For one thing you cannot not use panic, it's baked into the language.

I'm personally a fan of the distinction drawn up [here](https://blog.rust-lang.org/2016/05/26/Rust-1.9.html#controlled-unwinding) - `Result` is for predictable error scenarios that are expected in standard operation, normally due to bad user input, whereas panic is indicative of a bug. Or to put it another way, Result is for control flow and panic for debug.

# Pain Points

Currently within IOx we tend to avoid panic and favour error enumerations, and this causes some pain points

## Impossible Code Paths

It is fairly common to come across error propagation code that could never be encountered in the absence of a bug. 

For example, the catalog chunk has an error enumeration for when its state doesn't match what was expected for the given transition. However, the callers always check the state, perform the necessary work to convert the data, and then call the method to transition the state. If this function ever returned an error it would be a bug. There is therefore limited value gained from making this function fallible, over just documenting the invariant and panicking if is violated.

Other similar examples might be encountering an error building an arrow schema for a chunk, or creating a record batch from a set of columns, neither of which should be possible by construction.

## Error Agnostic Code Paths

A number of code paths within IOx are what I'm going to call error agnostic, that is, they have no mechanism to recover from an erroneous situation, and will just print any encountered error to stderr and return. 

Most background tasks fall into this category - if say a compaction job fails, not only is this almost certainly a bug, but all that is going to happen is it is going to get logged to stderr. 

I'd imagine query execution to also likely fall into this category, if the query encounters an IO error reading a parquet file, or encounters some invalid dictionary index, it is just going to abort and return this error to the user. Having additional plumbing to represent all the possible failure scenarios as separate enumerations, only to just serialize them blindly, is not adding a great deal

## Error Cardinality

[...]

PR #4686, opened 2022-05-25
fix: return gRPC error on panic

mkmik:
The tonic gRPC library doesn't convert panics into gRPC errors. Instead, in case of a panic, the request is abruptly dropped resulting in various effects, depending on how the gRPC client library surfaces the error:

For example, a rust client:

```console
$ ./target/debug/influxdb_iox query  foobar_weather 'show tables'
Error querying: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }
```

A test app using our https://github.com/influxdata/influxdb-iox-client-go client:

```
code = Canceled desc = stream terminated by RST_STREAM with error code: CANCEL
```

----

This PR adds a tower middleware that intercepts the panic and responds with a gRPC error.

The error code is modelled after the choice done by the popular Go gRPC middleware used for the same purpose: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0/recovery

----

I tested this manually by injecting a panic in a function called by the query path:

```diff
diff --git querier/src/database.rs querier/src/database.rs
index c80c49b65..7f6af94ba 100644
--- querier/src/database.rs
+++ querier/src/database.rs
@@ -84,6 +84,7 @@ impl QuerierDatabase {
 
     /// Get namespace if it exists
     pub async fn namespace(&self, name: &str) -> Option<Arc<QuerierNamespace>> {
+        panic!("FOO");
```

The rust and Go applications mentioned before now return:

```console
$ ./target/debug/influxdb_iox query  foobar_weather 'show tables'
Error querying: status: Internal, message: "internal error, sad kittens", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Wed, 25 May 2022 00:22:23 GMT"} }
```

```
code = Internal desc = internal error, sad kittens
```

----

I didn't add any automated test since it would require a lot of work for a relatively little return.


ioxd_common/src/rpc.rs
```diff
@@ -139,6 +146,16 @@ macro_rules! serve_builder {
     }};
 }
 
+pub fn handle_panic(_err: Box<dyn Any + Send + 'static>) -> http::Response<BoxBody> {
+    http::Response::builder()
+        .status(http::StatusCode::OK)
+        .header(http::header::CONTENT_TYPE, "application/grpc")
+        .header("grpc-status", "13" /* tonic::status::Code::Internal */)
+        .header("grpc-message", "internal error, sad kittens") // we don't want to leak the panic message
```
tustvold, line 154: Perhaps this could be configurable, if IOx isn't external facing as will often be the case, getting the full error in the caller can be useful? I am of course assuming gateway sanitises errors

tustvold (approved):
Big fan, if I'm not mistaken this covers https://github.com/influxdata/influxdb_iox/issues/1694 .

I would perhaps vote for returning the panic message, but don't feel strongly

ioxd_common/src/rpc.rs
mkmik: Yeah ideally it should be configurable. I'm too too lazy to do it here

ioxd_common/src/rpc.rs
```diff
@@ -139,6 +146,16 @@ macro_rules! serve_builder {
     }};
 }
 
+pub fn handle_panic(_err: Box<dyn Any + Send + 'static>) -> http::Response<BoxBody> {
+    http::Response::builder()
+        .status(http::StatusCode::OK)
+        .header(http::header::CONTENT_TYPE, "application/grpc")
+        .header("grpc-status", "13" /* tonic::status::Code::Internal */)
```
mkmik, line 153: Is there a way to create a const string out integer constant?

Or should I just use format! here?

PR #4689, opened 2022-05-25
fix: segfault/crash handling

crepererum:
See individual commits.

I actually just wanted to test that our segfault/crash handler works but it turned out it wasn't reliable, so I got nerdsniped and redesigned it :sweat_smile: 

Also log printing in our test infrastructure was somewhat broken.

### DIFF
--- a/influxdb_iox/Cargo.toml
+++ b/influxdb_iox/Cargo.toml
@@ -40,15 +40,16 @@ backtrace = "0.3"
 bytes = "1.0"
 clap = { version = "3", features = ["derive", "env"] }
 console-subscriber = { version = "0.1.6", optional = true, features = ["parking_lot"] }
+crash-handler = "0.3"
 dotenv = "0.15.0"
 futures = "0.3"
 hashbrown = "0.12"
 http = "0.2.7"
 humantime = "2.1.0"
 itertools = "0.10.1"
-libc = { version = "0.2" }
 num_cpus = "1.13.0"
 once_cell = { version = "1.12.0", features = ["parking_lot"] }
+parking_lot = "0.12"
 rustyline = { version = "9.0", default-features = false }
 serde_json = "1.0.81"
 snafu = "0.7"
--- a/influxdb_iox/src/main.rs
+++ b/influxdb_iox/src/main.rs
@@ -520,7 +520,7 @@ unsafe impl CrashEvent for CrashEventImpl {
         // https://github.com/rust-lang/backtrace-rs/pull/265#discussion_r352654555 ) be done by an ordinary thread.
         eprintln!("!!! CRASH REPORT START !!!",);
 
-        eprintln!("Signal: {}", context.siginfo.ssi_signo,);
+        print_signal(context);
 
         if let Some(name) = std::thread::current().name() {
             eprintln!("Thread: {}", name);
@@ -562,6 +562,19 @@ unsafe impl CrashEvent for CrashEventImpl {
         crash_handler::CrashEventResult::Handled(false)
     }
 }
+
+#[cfg(target_os = "linux")]
+fn print_signal(context: &crash_handler::CrashContext) {
+    eprintln!("Signal: {}", context.siginfo.ssi_signo);
+}
+
+#[cfg(target_os = "macos")]
+fn print_signal(context: &crash_handler::CrashContext) {
+    eprintln!("Exception: {:?}", context.exception);
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "macos")))]
+fn print_signal(context: &crash_handler::CrashContext) {}
 
 /// A ':' separated key value pair
 #[derive(Debug, Clone)]
--- a/influxdb_iox/src/main.rs
+++ b/influxdb_iox/src/main.rs
@@ -13,16 +13,26 @@ use crate::commands::{
     run::all_in_one,
     tracing::{init_logs_and_tracing, init_simple_logs, TroggingGuard},
 };
+use crash_handler::{CrashEvent, CrashHandler};
 use dotenv::dotenv;
 use influxdb_iox_client::connection::Builder;
 use iox_time::{SystemProvider, TimeProvider};
 use observability_deps::tracing::warn;
-use once_cell::sync::Lazy;
-use std::time::Duration;
+use once_cell::sync::{Lazy, OnceCell};
+use parking_lot::Mutex;
 use std::{
     collections::hash_map::DefaultHasher,
     hash::{Hash, Hasher},
     str::FromStr,
+};
+use std::{
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        mpsc::{sync_channel, SyncSender},
+        Arc,
+    },
+    thread::JoinHandle,
+    time::Duration,
 };
 use tokio::runtime::Runtime;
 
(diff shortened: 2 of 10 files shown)


influxdb_iox/Cargo.toml
```diff
@@ -40,15 +40,16 @@ backtrace = "0.3"
 bytes = "1.0"
 clap = { version = "3", features = ["derive", "env"] }
 console-subscriber = { version = "0.1.6", optional = true, features = ["parking_lot"] }
+crash-handler = "0.3"
```
alamb, line 43: https://crates.io/crates/crash-handler
Screen Shot 2022-05-25 at 11 14 49 AM
🤔 

How much do we want to invest in helping make this crate stable ?

alamb: The crate is less than a month old:
Screen Shot 2022-05-25 at 11 16 04 AM
https://crates.io/crates/crash-handler/versions

test_helpers_end_to_end/src/server_fixture.rs
```diff
@@ -669,9 +671,9 @@ fn dump_log_to_stdout(server_type: ServerType, log_path: &Path) {
     let mut f = std::fs::File::open(log_path).expect("failed to open log file");
     let mut buffer = [0_u8; 8 * 1024];
 
-    info!("****************");
-    info!("Start {:?} TestServer Output", server_type);
-    info!("****************");
+    println!("****************");
```
alamb, line 674: 👍

influxdb_iox/tests/end_to_end_cases/error.rs
```diff
@@ -43,18 +43,107 @@ pub async fn test_panic() {
...
+}
+
+async fn assert_segfault_logging(connection: Connection, log_path: Box<Path>) {
+    // trigger segfault
+    let mut client = influxdb_iox_client::test::Client::new(connection);
+    let err = client
+        .error(influxdb_iox_client::test::ErrorType::Segfault)
+        .await
+        .unwrap_err();
+    match err {
```
alamb, line 117: Why are these errors ignored? Wouldn't the test be invalid if we got a `transport error` (b/c the client couldn't connect, for example)?

service_grpc_testing/src/lib.rs
```diff
@@ -9,16 +9,31 @@ struct IOxTestingService {}
...
+        let req = req.into_inner();
+        match req.error_type() {
+            ErrorType::Unspecified | ErrorType::Panic => {
+                warn!("Got a test_error request. About to panic");
+                // Purposely do not use a static string (so that the panic
+                // code has to deal with aribtrary payloads). See
+                // https://github.com/influxdata/influxdb_iox/issues/1953
+                panic!("This {}", "is a test panic");
+            }
+            ErrorType::Segfault => {
```
alamb, line 23: Given that segfault brings down the entire process, I am not sure what I think about an admin API that can do so. Can we perhaps compile this out of release builds?

influxdb_iox/src/main.rs
```diff
@@ -371,50 +381,185 @@ fn load_dotenv() {
...
-    use libc::{sigaction, sigfillset, sighandler_t};
-    let mut sigset = std::mem::zeroed();
+unsafe impl CrashEvent for CrashEventImpl {
+    fn on_crash(&self, context: &crash_handler::CrashContext) -> crash_handler::CrashEventResult {
+        let _guard = self.currently_handles_signal.lock();
+
+        // We MUST do as little as possible in this context. Esp. we must avoid allocations. For that we only try to use
+        // stack-allocated data structures and let the actual stack frame printing (which allocates, see
+        // https://github.com/rust-lang/backtrace-rs/pull/265#discussion_r352654555 ) be done by an ordinary thread.
+        eprintln!("!!! CRASH REPORT START !!!",);
```
alamb, line 521: will this information end up in the logs?

alamb (dismissed):
I am concerned about:
1. Use of a brand new crate
2. Adding ability to `sefault` from admin API in release builds

However, I think it is a step forward and anything that improves our ability to diagnose issues is good 👍

influxdb_iox/src/main.rs
crepererum: yes and no. It ends up in the stdout, which is the best we can try. From there it will be handled by our logging infra. Plain prints obviously are very limited when it comes to log handling and formatting (e.g. this will never send logs to a remote endpoint -- however we don't do that in prod anyways). If you already segfaulted and everything is on fire, you have to be quick and keep it simple. So we try our best here w/ the limited resources.

influxdb_iox/tests/end_to_end_cases/error.rs
crepererum: You get an transport error because if you kill the server using a segfault, it dies and won't shut down the connection properly, hence the transport error. There's no real way for us to know when this transport error occurred though, but since we later will find the error backtrace in the log file, we'll know that it worked.

alamb:
Good news; it now compiles on mac!

Somewhat bad news is that it seems to hang on crash

When I apply this patch to simulate an crash in the compactor:

```diff
(arrow_dev) alamb@MacBook-Pro-6:~/Software/influxdb_iox2$ git diff
diff --git a/compactor/src/handler.rs b/compactor/src/handler.rs
index acde6daf4..bd326ff83 100644
--- a/compactor/src/handler.rs
+++ b/compactor/src/handler.rs
@@ -172,6 +172,9 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
         let max_file_size = compactor.config.compaction_max_size_bytes();
         let mut handles = vec![];
 
+        // crash and burn
+        unsafe { std::ptr::null_mut::<i32>().write(42) };
+
         for c in candidates {
             let compactor = Arc::clone(&compactor);
             let handle = tokio::task::spawn(async move {
```

And then run iox like

```shell
(arrow_dev) alamb@MacBook-Pro-6:~/Software/influxdb_iox2$ CARGO_TARGET_DIR=/Users/alamb/Software/iox2-target cargo run -- -vv
    Finished dev [unoptimized + debuginfo] target(s) in 0.41s
     Running `/Users/alamb/Software/iox2-target/debug/influxdb_iox -vv`
2022-05-25T16:56:58.569200Z  INFO clap_blocks::object_store: Object store: In-memory
2022-05-25T16:56:58.571538Z  INFO clap_blocks::write_buffer: Write buffer: File-based in `/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmpnusbvY`
2022-05-25T16:56:58.571692Z  INFO clap_blocks::catalog_dsn: Catalog: In-memory
2022-05-25T16:56:58.587907Z  INFO influxdb_iox::commands::run::all_in_one: running db migrations
2022-05-25T16:56:58.589893Z  INFO influxdb_iox::commands::run::all_in_one: Creating shared query executor num_threads=16
2022-05-25T16:56:58.593797Z DEBUG datafusion::execution::memory_manager: Creating memory manager with initial size 11744051.2 TB    
2022-05-25T16:56:58.600099Z  INFO influxdb_iox::commands::run::all_in_one: starting router
2022-05-25T16:56:58.613639Z  INFO ioxd_router: connected to write buffer topic topic="iox-shared" shards=1
2022-05-25T16:56:58.622259Z  INFO influxdb_iox::commands::run::all_in_one: starting ingester
2022-05-25T16:56:58.627196Z  INFO ingester::handler: ingester handler and lifecycle started with config LifecycleConfig { pause_ingest_size: 107374182400, persist_memory_threshold: 1073741824, partition_size_threshold: 314572800, partition_age_threshold: 1800s, partition_cold_threshold: 300s }
2022-05-25T16:56:58.627275Z DEBUG ingester::handler: Seek stream kafka_partition=0 min_unpersisted_sequence_number=0
2022-05-25T16:56:58.627998Z  INFO influxdb_iox::commands::run::all_in_one: starting compactor
2022-05-25T16:56:58.628725Z DEBUG ingester::stream_handler::periodic_watermark_fetcher: fetching write buffer watermark
2022-05-25T16:56:58.630484Z  INFO compactor::handler: compactor started with config CompactorConfig { split_percentage: 90, max_concurrent_compaction_size_bytes: 100000, compaction_max_size_bytes: 100000 }
2022-05-25T16:56:58.630717Z  INFO influxdb_iox::commands::run::all_in_one: starting querier ingester_addresses=["http://127.0.0.1:8083"]
2022-05-25T16:56:58.631247Z DEBUG compactor::compact: Number of candidate partitions to be considered to compact candidate_num=0
2022-05-25T16:56:58.631272Z DEBUG compactor::handler: found compaction candidates n_candidates=0
!!! CRASH REPORT START !!!
Exception: Some(ExceptionInfo { kind: 1, code: 0, subcode: Some(8) })
Thread: <unknown>
Stack trace:


```

It just hangs here (though I can kill it with `Ctrl-C`)

alamb:
I haven't had a chance to test it on linux yet

alamb:
Worked very nicely on linux 👌

alamb:

alamb:

PR #4692, opened 2022-05-25
test: add end to end test for deletes via `/api/v2/delete`

alamb:
# Rationale
Draft as the test is failing now (I think there is a bug in the code)

Closes https://github.com/influxdata/influxdb_iox/issues/4691 by adding a basic end to end test of the delete endpoint

# Changes
1. end to end test to ensure deletes are wired up correctly through the router, ingester and querier.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5324,6 +5324,8 @@ dependencies = [
  "prost",
  "rand",
  "reqwest",
+ "serde",
+ "serde_json",
  "sqlx",
  "tempfile",
  "test_helpers",
--- a/influxdb_iox/tests/end_to_end_cases/querier.rs
+++ b/influxdb_iox/tests/end_to_end_cases/querier.rs
@@ -1,11 +1,15 @@
 pub(crate) mod influxrpc;
 mod multi_ingester;
 
+use std::time::Duration;
+
 use assert_cmd::Command;
 use futures::FutureExt;
 use predicates::prelude::*;
+use test_helpers::timeout::FutureTimeout;
 use test_helpers_end_to_end::{
-    maybe_skip_integration, try_run_query, MiniCluster, Step, StepTest, StepTestState, TestConfig,
+    maybe_skip_integration, run_query, try_run_query, MiniCluster, Step, StepTest, StepTestState,
+    TestConfig,
 };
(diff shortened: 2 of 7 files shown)


alamb:
This PR currently fails like this:

```
running 1 test
test end_to_end_cases::querier::basic_delete ... FAILED

failures:

---- end_to_end_cases::querier::basic_delete stdout ----
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
Got 2 rows, waiting for 1
thread 'end_to_end_cases::querier::basic_delete' panicked at 'timeout waiting for task to join: Elapsed(())', /Users/alamb/Software/influxdb_iox/test_helpers/src/timeout.rs:22:14
stack backtrace:
```

(aka the delete is never applied)

I can't find evidence in the full testing log [full_log.txt](https://github.com/influxdata/influxdb_iox/files/8770251/full_log.txt) that the delete ever makes it to the ingester. Not sure where it gets lost

domodwyer:
When running this locally, I see:


```
2022-05-25T11:13:49.462232Z ERROR ingester::stream_handler::handler: failed to apply dml operation error=Table "the_table" not found in buffer kafka_topic=iox-shared kafka_partition=0 potential_data_loss=true
```

This error comes from [the ingester][handler] so it made it over the gap - it looks like we need to populate the catalog correctly.

The error is triggered due to a catalog call [`TableRepo::get_table_persist_info()`][catalog] returning no results. The router creates tables in the catalog for writes, but not deletes as it is assumed to already exist (which should be the case as the test issues writes to the table first).

[handler]:https://github.com/influxdata/influxdb_iox/blob/6ed5b8c5813be129b05621a368791632baad325b/ingester/src/stream_handler/handler.rs#L344-L350
[catalog]:https://github.com/influxdata/influxdb_iox/blob/6ed5b8c5813be129b05621a368791632baad325b/iox_catalog/src/interface.rs#L309

PR #4693, opened 2022-05-25
fix: Return panic message in internal error

mkmik:
In #4686 we turned panics into gRPC errors, but we were decided to play it safe and not return the actual panic message.

Since most of the current usage of IOx is likely going to be internal (i.e. IOx will always be exposed to the external world via some kind of API gateway), returning the panic text is probably a saner default.

(Ideally we should make this behaviour configurable, but I don't fancy spending that time now.)


mkmik:
@tustvold already voted for this in the original PR

alamb:
🎉

PR #4697, opened 2022-05-25
refactor: de-duplicate low-level arrow code

crepererum:
It seems that during prototyping NG we've copied low level code (w/o
tests!) and never cleaned up. Let's not have this functionality twice.


alamb (approved):
👨‍🍳 👌 

Nice find -- thank you @crepererum

PR #4699, opened 2022-05-25
refactor: warn for silly object store configs

domodwyer:
I'll let you guess why I am proposing this change.

---

* refactor: warn for silly object store configs (0e5c86fd8)

      Warn when downloading files to an in-memory object store.

      The "remote partition pull" command downloads parquet files from an object
      store via a router, and saves them locally. It's pretty unlikely the user
      intends to download those files to memory of the CLI process which then exits
      when the pull is complete, throwing away the downloaded files, but this is the
      default.

* test(e2e): ensure "partition pull" writes files (1fc95ba72)

      Adds a test case covering the "remote partition pull" command configured with
      file-based object storage.

### DIFF
--- a/influxdb_iox/src/commands/remote/partition.rs
+++ b/influxdb_iox/src/commands/remote/partition.rs
@@ -1,7 +1,7 @@
 //! This module implements the `remote partition` CLI subcommand
 
 use bytes::Bytes;
-use clap_blocks::object_store::make_object_store;
+use clap_blocks::object_store::{make_object_store, ObjectStoreType};
 use clap_blocks::{catalog_dsn::CatalogDsnConfig, object_store::ObjectStoreConfig};
 use data_types::{
     ColumnType, KafkaPartition, NamespaceId, NamespaceSchema as CatalogNamespaceSchema,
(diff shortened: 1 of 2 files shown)


influxdb_iox/src/commands/remote/partition.rs
```diff
@@ -133,6 +139,14 @@ pub async fn command(connection: Connection, config: Config) -> Result<(), Error
             let partition_mapping =
                 load_partition(&catalog, &schema, &pull.table, &partition).await?;
 
+            #[cfg(not(test))]
```
carols10cents, line 142: Yeah this isn't going to work because the end-to-end tests [essentially do `cargo run`](https://github.com/influxdata/influxdb_iox/blob/2d5a327bf40cf3b257841ee496a042deef1d9c6b/influxdb_iox/tests/end_to_end_cases/cli.rs#L75) and compile the code in non-test mode, so that's why this is getting triggered in tests. 

Could the tests use a temp directory and the file object store instead?

domodwyer: Yeah as soon as I saw the failure I knew I'd have to rejig the test! Doh!

PR #4701, opened 2022-05-25
fix: expire empty parquet_files cache and empty tombstones cache

alamb:
# Rationale
As part of https://github.com/influxdata/influxdb_iox/pull/4650 I found there was a bug in the caches introduced in https://github.com/influxdata/influxdb_iox/pull/4661 and https://github.com/influxdata/influxdb_iox/pull/4663

Specifically, the cache was not invalidated if it contained no files / tombstones. 

# Changes
This PR fixes the issue and adds test coverage


querier/src/cache/parquet_file.rs
```diff
@@ -206,7 +216,9 @@ impl ParquetFileCache {
                 if let Some(max_cached) = max_cached {
                     max_cached < max_parquet_sequence_number
                 } else {
-                    false
+                    // a max sequence was provided but there were no
+                    // files in the cache. Means we need to refresh
+                    true
```
alamb, line 221: this (and the corresponding fix in tombstones.rs) is the actual code change. The rest is tests

NGA-TRAN (approved):
Yeah, I saw the 2 `true` lines with the tests :)

alamb:
Thanks @NGA-TRAN !

PR #4703, opened 2022-05-25
feat: add concurrency limit for ingester queries

### LINKED ISSUE #4657: Limit ingester query concurrency
Apply a configurable service protection limit to the number of concurrent queries an ingester will execute at any one time to indirectly control the amount of memory & CPU used to satisfy them.

Ideally we should expose a metric that does an `inc(1)` for each dropped query so we can quantify query saturation and alert/correlate/scale accordingly - especially important if we start seeing "hot shards" for queries :+1:

pauldix:
I picked 20 as the arbitrary default concurrency limit, we can adjust as needed.

Closes #4657


ingester/src/handler.rs
```diff
@@ -744,6 +786,27 @@ mod tests {
         .await;
     }
 
+    #[tokio::test]
+    async fn limits_concurrent_queries() {
```
pauldix, line 790: This test is kind of lame, but at the very least it'll be a warning to whoever refactors it later that this concurrency must come along for the ride.

ingester/src/handler.rs
```diff
@@ -109,6 +112,18 @@ pub struct IngestHandlerImpl<T = SystemProvider> {
...
+    /// Query request rejected due to concurrency limits
+    query_request_limit_rejected: U64Counter,
+
+    // A request limiter to restrict the number of simultaneous requests this
+    // router services.
+    //
+    // This allows the router to drop a portion of requests when experiencing an
+    // unusual flood of requests (i.e. due to peer routers crashing and
+    // depleting the available instances in the pool) in order to preserve
+    // overall system availability, instead of OOMing or otherwise failing.
```
alamb, line 125: I think this may need to be updated to refer to `ingester` rather than `router`

alamb:
Looks like the default may need to be increased in the end to end tests (or maybe change default to 50) as it concurrently runs  several tests in parallel

https://circleci.com/gh/influxdata/influxdb_iox/167212?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

alamb (approved):
LGTM

domodwyer (approved):
Looks spot on to me :+1:

Issue #4681, opened 2022-05-24
Add Querier caches for parquet files in read buffer


alamb:
# Background / Abstract:
On every query today the querier might request one or more parquet files directly from the object store, decodes them and discards them at the end of the query

this is especially wasteful for common workloads that repeatedly query the same table

We would like to cache the results of fetching and decoding the parquet files using the ReadBuffer (👋  @e-dard ) in the querier. 

The read buffer has had significant effort on memory efficiency and runtime performance. 

# Acceptance Criteria / Definition of Done
- [ ] Create a new type of [`Cache`](https://github.com/influxdata/influxdb_iox/blob/8c537c94672586a0aadd437a024e92c3c3c3664b/cache_system/src/driver.rs#L17-L29) that maps from [`ParquetFileId`](https://github.com/influxdata/influxdb_iox/blob/4d8ece5524bd8927df830a578186fbbdc8f028ea/data_types/src/lib.rs#L296-L299) to a [ReadBuffer `Chunk`](https://github.com/influxdata/influxdb_iox/blob/8c537c94672586a0aadd437a024e92c3c3c3664b/read_buffer/src/chunk.rs#L53-L60) 
- [ ] The cache should properly account for memory consumption and have a [LRU](https://github.com/influxdata/influxdb_iox/blob/main/cache_system/src/backend/lru.rs) eviction policy
- [ ] When the querier runs a query, if the parquet data is in the cache, the data should be read from the read buffer in the cache rather than fetching the parquet file (see "ReadBuffer Chunk" section below)

## ReadBuffer Chunk
To wire up the ReadBuffer into the querier (and ultimately making it visible to the query engine) it needs to implement [`QueryChunk`](https://github.com/influxdata/influxdb_iox/blob/95e6a8ed4679552d90ffdb60b5ae40880577652b/iox_query/src/lib.rs#L177-L249), which in turn requires [`QueryChunkMeta`](https://github.com/influxdata/influxdb_iox/blob/95e6a8ed4679552d90ffdb60b5ae40880577652b/iox_query/src/lib.rs#L38-L91). The querier does this already for parquet chunks ([`
[...]

PR #4704, opened 2022-05-25
feat: Start of a read buffer chunk cache

carols10cents:
Connects to #4681.

This is a new querier cache that eventually will cache Read Buffer chunks created from Parquet files as they're requested.

The new querier cache isn't called anywhere yet, and it doesn't work completely: I have a bunch of functions sketched out that show the steps I'm planning on taking, but all except the last one in the sequence are `unimplemented!` in this PR.

However, because this isn't called anywhere yet, this can be safely merged in and iterated on!

### DIFF
--- a/cache_system/src/driver.rs
+++ b/cache_system/src/driver.rs
@@ -1,30 +1,35 @@
 //! Main data structure, see [`Cache`].
-use std::{collections::HashMap, hash::Hash, sync::Arc};
 
+use super::{backend::CacheBackend, loader::Loader};
 use futures::{
     future::{BoxFuture, Shared},
     FutureExt, TryFutureExt,
 };
 use observability_deps::tracing::debug;
 use parking_lot::Mutex;
+use std::{collections::HashMap, hash::Hash, sync::Arc};
 use tokio::{
     sync::oneshot::{error::RecvError, Sender},
     task::JoinHandle,
 };
 
-use super::{backend::CacheBackend, loader::Loader};
-
 /// High-level cache implementation.
 ///
 /// # Concurrency
-/// Multiple cache requests for different keys can run at the same time. When data is requested for the same key the
-/// underlying loader will only be polled once, even when the requests are made while the loader is still running.
+///
+/// Multiple cache requests for different keys can run at the same time. When data is requested for
+/// the same key the underlying loader will only be polled once, even when the requests are made
+/// while the loader is still running.
 ///
 /// # Cancellation
-/// Canceling a [`get`](Self::get) request will NOT cancel the underlying loader. The data will still be cached.
+///
+/// Canceling a [`get`](Self::get) request will NOT cancel the underlying loader. The data will
+/// still be cached.
 ///
 /// # Panic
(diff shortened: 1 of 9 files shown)


cache_system/src/driver.rs
```diff
@@ -187,9 +199,10 @@ where
                 // it's OK when the receiver side is gone (likely panicked)
                 running_query.set.send(v.clone()).ok();
 
-                // When we side-load data into the running task, the task does NOT modify the backend, so we have to do
-                // that. The reason for not letting the task feed the side-loaded data back into `cached_entries` is that we
-                // would need to drop the state lock here before the task could acquire it, leading to a lock gap.
+                // When we side-load data into the running task, the task does NOT modify the
```
alamb, line 202: Thank you for re-indenting the comments ;) I ❤️  80 columns !

querier/src/cache/read_buffer.rs
```diff
@@ -0,0 +1,201 @@
...
+use read_buffer::RBChunk;
+use snafu::{ResultExt, Snafu};
+use std::{collections::HashMap, mem, sync::Arc};
+
+const CACHE_ID: &str = "read_buffer";
+
+/// Cache for parquet file data decoded into read buffer chunks
+#[derive(Debug)]
+pub struct ReadBufferCache {
+    cache: Cache<ParquetFileId, Arc<RBChunk>>,
```
alamb, line 27: 👍

querier/src/cache/read_buffer.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// Get read buffer chunks by Parquet file id
+    pub async fn get(&self, parquet_file_id: ParquetFileId) -> Arc<RBChunk> {
+        self.cache.get(parquet_file_id).await
+    }
+}
+
+fn parquet_file_by_id(_parquet_file_id: ParquetFileId) -> ParquetFile {
+    unimplemented!()
```
alamb, line 94: it might be interesting / worth it to consider looking at the `parquet_file` cache to see can reuse some / all of that information and avoid re-fetching it. On the other hand, maybe that is an optimization best done as a follow on 🤔

alamb (approved):
🚆  Looks like a good start!

cache_system/src/driver.rs
carols10cents: You're welcome! (It's actually 100 😉 )

querier/src/cache/read_buffer.rs
carols10cents: Yeah, I definitely was punting on this until your PR got in. Also, it looks like in the code we currently have, we'll have a whole `ParquetFileWithMetadata` at the point we'll be asking for the file's contents, so while the cache key should only need to be the `ParquetFileId`, it'd be nice if we could ask for the chunk by passing the whole `ParquetFileWithMetadata` and just use that if we need to fetch the file contents. That might need some rearranging though, and I decided to start on easier stuff first :)

Issue #4392, opened 2022-04-22
Querier to Prod


crepererum:
TODO

crepererum:
I think we currently keep the epics in our board for better "vision" planning, but I don't have a strong opinion on that :shrug:

alamb:
Makes sense -- I only now saw 
<img width="737" alt="Screen Shot 2022-06-29 at 6 35 10 AM" src="https://user-images.githubusercontent.com/490673/176449684-568ef89e-c778-437c-a7c6-fff561a5279c.png">
👍

PR #4709, opened 2022-05-26
feat: Make a QuerierRBChunk wrapper that implement QueryChunk and QueryChunkMeta

carols10cents:
Connects to #4681.

This brings back [a bunch of code from the OG `DbChunk`](https://github.com/influxdata/influxdb_iox/blob/37c7ce793cc12a7f5dc29b3bc3d575c6c500f2bf/db/src/chunk.rs) when the chunk's storage was in the read buffer.

I also added a method for the read buffer cache in the catalog cache, but it's deliberately unimplemented so that it won't get used in production yet. That's so I could call it in `ChunkAdapter::new_chunk` and start to get an idea of what data I'll have when, but I'm not calling `new_chunk` anywhere yet either. 

Is everything in the place you'd expect it to be?

### DIFF
--- a/querier/src/cache/read_buffer.rs
+++ b/querier/src/cache/read_buffer.rs
@@ -85,7 +85,7 @@ impl ReadBufferCache {
         Self { cache, _backend }
     }
 
-    /// Get read buffer chunks by Parquet file id
+    /// Get read buffer chunks from the cache or the Parquet file
     pub async fn get(&self, decoded_parquet_file: &DecodedParquetFile) -> Arc<RBChunk> {
         let parquet_file = &decoded_parquet_file.parquet_file;
 
--- a/querier/src/cache/read_buffer.rs
+++ b/querier/src/cache/read_buffer.rs
@@ -15,6 +15,7 @@ use data_types::{ParquetFile, ParquetFileId};
 use datafusion::physical_plan::SendableRecordBatchStream;
 use futures::StreamExt;
 use iox_time::TimeProvider;
+use parquet_file::chunk::DecodedParquetFile;
 use read_buffer::RBChunk;
 use snafu::{ResultExt, Snafu};
 use std::{collections::HashMap, mem, sync::Arc};
@@ -85,8 +86,10 @@ impl ReadBufferCache {
     }
 
     /// Get read buffer chunks by Parquet file id
-    pub async fn get(&self, parquet_file_id: ParquetFileId) -> Arc<RBChunk> {
-        self.cache.get(parquet_file_id).await
+    pub async fn get(&self, decoded_parquet_file: &DecodedParquetFile) -> Arc<RBChunk> {
+        let parquet_file = &decoded_parquet_file.parquet_file;
+
+        self.cache.get(parquet_file.id).await
     }
 }
 
--- a/querier/src/chunk/mod.rs
+++ b/querier/src/chunk/mod.rs
@@ -4,7 +4,7 @@ use crate::cache::CatalogCache;
 use arrow::record_batch::RecordBatch;
 use data_types::{
     ChunkId, ChunkOrder, DeletePredicate, ParquetFileId, ParquetFileWithMetadata, PartitionId,
-    SequenceNumber, SequencerId, TimestampMinMax,
+    SequenceNumber, SequencerId, TableSummary, TimestampMinMax, TimestampRange,
 };
 use futures::StreamExt;
 use iox_catalog::interface::Catalog;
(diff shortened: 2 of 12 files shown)


alamb:
I will review this tomorrow morning when I am fresher. Thanks @carols10cents

querier/src/cache/read_buffer.rs
```diff
@@ -85,8 +86,10 @@ impl ReadBufferCache {
     }
 
     /// Get read buffer chunks by Parquet file id
```
alamb, line 88: This comment appears to be out of date (it is now fetched by decoded _parquet_file)

querier/src/chunk/mod.rs
```diff
@@ -80,6 +81,98 @@ impl ChunkMeta {
     }
 }
 
+/// Chunk representation of `read_buffer::RBChunk`s for the querier.
```
alamb, line 84: 👍

querier/src/chunk/mod.rs
```diff
@@ -276,6 +369,69 @@ impl ParquetChunkAdapter {
             partition_sort_key,
         ))
     }
+
+    /// Create new cached chunk.
+    ///
+    /// Returns `None` if some data required to create this chunk is already gone from the catalog.
+    pub async fn new_chunk(
```
alamb, line 376:
```suggestion
    pub async fn new_rb_chunk(
```

Or maybe 

```suggestion
    pub async fn new_cached_chunk(
```

There are already so many "chunk"s in the system, making it a little more specific I think will help

querier/src/chunk/mod.rs
```diff
@@ -276,6 +369,69 @@ impl ParquetChunkAdapter {
...
+            .table()
+            .name(parquet_file.table_id)
+            .await?;
+
+        let iox_metadata = &decoded_parquet_file.iox_metadata;
+
+        // Somewhat hacky workaround because of implicit chunk orders, use min sequence number and
+        // hope it doesn't overflow u32. Order is non-zero, se we need to add 1.
+        let order = ChunkOrder::new(1 + iox_metadata.min_sequence_number.get() as u32)
+            .expect("cannot be zero");
```
alamb, line 407: I wonder if @crepererum  has any thoughts on this

querier/src/chunk/query_access.rs
```diff
@@ -135,3 +158,328 @@ impl QueryChunk for QuerierParquetChunk {
...
+
+        let mut values = self.rb_chunk.column_values(
+            rb_predicate,
+            schema::selection::Selection::Some(&[column_name]),
+            BTreeMap::new(),
+        )?;
+
+        // The InfluxRPC frontend only supports getting column values
+        // for one column at a time (this is a restriction on the Influx
+        // Read gRPC API too). However, the Read Buffer support multiple
```
alamb, line 286:
```suggestion
        // Read gRPC API too). However, the Read Buffer supports multiple
```

querier/src/chunk/query_access.rs
```diff
@@ -135,3 +158,328 @@ impl QueryChunk for QuerierParquetChunk {
...
+        let rb = self.read_results.next();
+        if let Some(rb) = &rb {
+            ctx.set_metadata("output_rows", rb.num_rows() as i64);
+        }
+
+        Poll::Ready(Ok(rb).transpose())
+    }
+
+    // TODO is there a useful size_hint to pass?
+}
```
alamb, line 485: The code looks good in this module, but I think we need to add test coverage. The fullest coverage is probably by adding ReadBufferChunks to the setups created in `query_tests` -- maybe we can add a ticket to track that task (I think @crepererum has some ideas)

alamb (approved):
The only significant concern I have is lack of test coverage, but I think it would be ok to add such coverage as a follow on PR

querier/src/chunk/mod.rs
carols10cents: This was copy-pasted from [`new_querier_parquet_chunk` above](https://github.com/influxdata/influxdb_iox/blob/16e7a6d5962f610e4b21432f568865719f494edb/querier/src/chunk/mod.rs#L249-L252), I'm not sure if it applies anymore though?

querier/src/chunk/query_access.rs
carols10cents: Filed https://github.com/influxdata/influxdb_iox/issues/4720 and added it to the list in https://github.com/influxdata/influxdb_iox/issues/4392 !

querier/src/chunk/mod.rs
crepererum: I think it's not too hacky. We might want to extend the chunk order integer width to 64bits now that OG is gone.

PR #4717, opened 2022-05-27
refactor: use consistent spelling of serialize in logging and code

alamb:
# Rationale
I want to be able to `grep` for a single string to find serialization.

Found while debugging https://github.com/influxdata/conductor/issues/956

 I don't believe myself to be an 🇺🇸 imperialist, simply there are already many more uses of the spelling `serialize` than `serialise` in the code

* https://github.com/influxdata/influxdb_iox/search?q=serialise (10 now)
* https://github.com/influxdata/influxdb_iox/search?q=serialize (35 now)

This stuff was mostly added in b9a745d42d6e72705a927aee34baf17572313015 / https://github.com/influxdata/influxdb_iox/pull/4649 from @domodwyer

# Changes
1. Use consistent spelling of `serialize` as suggested in https://github.com/influxdata/influxdb_iox/pull/4649#discussion_r878348122


shepmaster (approved):
I don't see any `serialise` in the code after this change.

PR #4721, opened 2022-05-27
fix: Support `_field != <name>` predicates

alamb:
# Rationale
Closes https://github.com/influxdata/conductor/issues/956

Basically predicates like `_field != "foo"` are not supported. Predicates like `_field = "foo"` are (the difference is the equality predicate).

This was a todo the original code, which I will highlight below

# Changes:
1. Support `_field != "foo"` and chains of them like `_field != "foo" AND _field != "bar"`
2. Throw errors now if we see some a pattern of _field predicate we don't support (rather than silently ignoring it)

# TODO:
- [x] add query-test for unsupported _field predicate
- [x] add tests for normalize_predicate (and code that translates exclusion lists
- [x] Tests for multiple invocations against two exprs pick up invalid combinations


predicate/src/rpc_predicate.rs
```diff
@@ -312,54 +322,10 @@ impl<'a> ExprRewriter for FieldValueRewriter<'a> {
...
-    fn mutate(&mut self, expr: Expr) -> DataFusionResult<Expr> {
-        Ok(match expr {
-            Expr::BinaryExpr {
-                ref left,
-                op,
-                ref right,
-            } => {
-                if let Expr::Column(inner) = &**left {
-                    if inner.name != FIELD_COLUMN_NAME || op != Operator::Eq {
-                        // TODO(edd): add support for !=
```
alamb, line 343: "TODO" lol

alamb (commented):
I am making progress here, but my "is this predicate supported" logic is not yet sophisticated enough

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,587 @@
...
+                        Ok(())
+                    }
+                })?;
+                Ok(NoPred)
+            }
+        }
+    }
+}
+
+/// searches for unsupported patterns of predicates that we can't
```
alamb, line 257: This checker was by far the most complicated part of this PR

predicate/src/rpc_predicate.rs
```diff
@@ -1,18 +1,21 @@
-//! Interface logic between IOx ['Predicate`] and predicates used by the
-//! InfluxDB Storage gRPC API
+//! Interface logic between IOx ['Predicate`] and the predicates used
+//! by the InfluxDB Storage gRPC API
+mod field_rewrite;
```
alamb, line 3: I moved the field rewriting logic into its own module, as this one was already pretty large

predicate/src/rpc_predicate.rs
```diff
@@ -186,14 +196,32 @@ fn normalize_predicate(
...
-        match &mut predicate.field_columns {
-            Some(field_columns) => field_columns.extend(field_projections.into_iter()),
-            None => predicate.field_columns = Some(field_projections),
-        };
+    match field_projections.into_projection() {
+        FieldProjection::None => {}
+        FieldProjection::Include(include_field_names) => {
+            predicate.add_field_names(include_field_names)
+        }
+        FieldProjection::Exclude(exclude_field_names) => {
```
alamb, line 208: This is part of supporting != -- ned to rewrite it into an exclude list

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
-                    binary_expr(col(FIELD_COLUMN_NAME), Operator::Eq, lit("f1")),
-                    Operator::Or,
-                    binary_expr(col(FIELD_COLUMN_NAME), Operator::Eq, lit("f2")),
-                ),
-                binary_expr(lit(true), Operator::Or, lit(true)),
-                vec!["f1", "f2"],
-            ),
-        ];
+    #[test]
+    fn test_normalize_predicate_field_rewrite_multi_field() {
```
alamb, line 408: New tests to cover exclusion list creation and that rewriting across exprs is hooked up

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,579 @@
...
+        match as_col_name(expr) {
+            Some(FIELD_COLUMN_NAME) => Err(DataFusionError::Plan(
+                "Unsupported _field predicate, could not rewrite for IOx".into(),
+            )),
+            _ => Ok(Recursion::Continue(self)),
+        }
+    }
+}
+
+/// What type of structurex been seen in children
```
alamb, line 171: I recommend reading the tests first in this module to understand what they are doing

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -727,6 +727,43 @@ async fn test_read_filter_on_field() {
...
+    let predicate = InfluxRpcPredicate::new(None, predicate);
+
+    let expected_results = vec![
+        "Series tags={_measurement=h2o, city=Boston, state=CA, _field=other_temp}\n  FloatPoints timestamps: [350], values: [72.4]",
+        "Series tags={_measurement=h2o, city=Boston, state=MA, _field=moisture}\n  FloatPoints timestamps: [100000], values: [43.0]",
+        "Series tags={_measurement=h2o, city=Boston, state=MA, _field=other_temp}\n  FloatPoints timestamps: [250], values: [70.4]",
+        "Series tags={_measurement=o2, city=Boston, state=MA, _field=reading}\n  FloatPoints timestamps: [50], values: [51.0]",
+    ];
+
+    run_read_filter_test_case(TwoMeasurementsManyFields {}, predicate, expected_results).await;
```
alamb, line 747: Without the code change in this PR, this new tests has the same symptoms as reported on https://github.com/influxdata/conductor/issues/956

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -727,6 +727,43 @@ async fn test_read_filter_on_field() {
...
+    // (_field != 'temp')
+    let p1 = col("_field")
+        .not_eq(lit("temp"))
+        .or(col("_field").eq(lit("other_temp")));
+    let predicate = PredicateBuilder::default().add_expr(p1).build();
+    let predicate = InfluxRpcPredicate::new(None, predicate);
+
+    let expected_error = "Unsupported _field predicate";
+
+    run_read_filter_error_case(TwoMeasurementsManyFields {}, predicate, expected_error).await;
```
alamb, line 764: I spent a long time on this PR trying to ensure errors were caught rather than silently generating incorrect answers or cryptic error messages. I hope that pays off

NGA-TRAN:
@alamb I will review this first thing tomorrow if no one else reviews it yet

predicate/src/lib.rs
```diff
@@ -99,6 +99,21 @@ impl Predicate {
...
+        I: IntoIterator<Item = S>,
+        S: Into<String>,
+    {
+        let mut new_field_columns = self.field_columns.take().unwrap_or_default();
+
+        let new_names = new_names.into_iter().map(|n| n.into());
+
+        new_field_columns.extend(new_names);
+
+        self.field_columns = Some(new_field_columns);
```
crepererum, line 114: This always set `field_columns` to `Some(...)`, even when the iterator was empty. I'm wondering two things:

- is that expected? If so, this should probably be documented as part of the `add_field_names` docstring.
- what's the difference between `self.field_columns = None` and `self.field_columns = Some(<empty>)`? Is it "ALL" VS "no columns at all"?

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
+                .add_expr(col("_field").eq(lit("f1")))
+                .add_expr(col("_field").not_eq(lit("f2")))
+                .build(),
+        )
+        .unwrap_err();
+
+        assert_contains!(
+            err.to_string(),
+            "Unsupported _field predicates: both '=' and '!=' present"
+        );
```
crepererum, line 461: I was wondering what would happen in this case. I think "unsupported" is OK for now.

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
 
-        for (input, exp_expr, field_columns) in cases {
-            let rewritten = input.rewrite(&mut rewriter).unwrap();
+    #[test]
+    fn test_normalize_predicate_field_rewrite_not_eq() {
+        let predicate = normalize_predicate(
+            "table",
+            Some(schema()),
+            &PredicateBuilder::new()
+                .add_expr(col("_field").not_eq(lit("f1")))
```
crepererum, line 432: While for `=` you have a test case with multiple expressions linked w/ an OR, for `!=` such a test case is missing. From looking at the implementation I'm not really sure what would happen in that case, even though `_field != foo AND _field != bar` would make sense.

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
-                vec!["f1", "f2"],
-            ),
-        ];
+    #[test]
+    fn test_normalize_predicate_field_rewrite_multi_field() {
+        let predicate = normalize_predicate(
+            "table",
+            Some(schema()),
+            &PredicateBuilder::new()
+                .add_expr(col("_field").eq(lit("f1")).or(col("_field").eq(lit("f2"))))
```
crepererum, line 413: What happens if these to EQs are linked via AND instead of OR?

predicate/src/rpc_predicate.rs
```diff
@@ -312,54 +340,14 @@ impl<'a> ExprRewriter for FieldValueRewriter<'a> {
...
-                    }
-                }
-                expr
-            }
-            _ => expr,
-        })
-    }
-}
-
 #[cfg(test)]
```
crepererum, line 343: As you might guess from my questions below: I'm having a bit of a hard time untangling the test setup/coverage, esp. which cases are covered here and which cases are covered in `mod field_rewrite`.

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
...
+            right: _,
+        } = expr
+        {
+            self.pop_stack_and_validate(op)?;
+        }
+        Ok(self)
+    }
+}
+
+#[cfg(test)]
```
crepererum, line 361: Ahh, here are all the nasty cases :tada:

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
+//! Logic for rewriting _field = "val" expressions to projections
```
crepererum, line 1: :+1: for having a proper rewriter w/ all the edge cases handled.

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
...
+            op_stack: vec![],
+        }
+    }
+
+    /// validates that we have not seen both != and == operators against _field
+    fn validate_operators(&self) -> DataFusionResult<()> {
+        if self.seen_eq && self.seen_not_eq {
+            Err(DataFusionError::Plan(
+                "Unsupported _field predicates: both '=' and '!=' present".into(),
+            ))
```
crepererum, line 284: Why do we have a generic error here and more specific errors in `FieldProjection::try_add_{include,exclude}`? Esp. under which condition does the user get the generic error and when do they get the more specific ones?

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
...
+fn as_str_lit(expr: &Expr) -> Option<&str> {
+    if let Expr::Literal(ScalarValue::Utf8(Some(name))) = &expr {
+        Some(name.as_str())
+    } else {
+        None
+    }
+}
+
+/// Searches for `_field` and errors if any are found
+struct FieldRefFinder {}
```
crepererum, line 158: How's this visitor different to `FieldRefChecker`?

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
+//! Logic for rewriting _field = "val" expressions to projections
+
+use super::FIELD_COLUMN_NAME;
+use datafusion::error::{DataFusionError, Result as DataFusionResult};
+use datafusion::logical_plan::{
+    lit, Expr, ExprRewritable, ExprRewriter, ExprVisitable, ExpressionVisitor, Operator, Recursion,
```
crepererum, line 6: High-level question: what's the difference between "visit" and "rewrite"? Is "visit" only for checking and "rewrite" can also mutate?

crepererum (commented):
I think I understand the high-level idea and very much welcome the more excessive tests. I don't fully follow the expr visit/rewrite part.

predicate/src/rpc_predicate.rs
NGA-TRAN: This is a tricky question. I guess similar question to `_field != col_a OR _field != col_b`. Should we throw syntax error?

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
+            &PredicateBuilder::new()
+                .add_expr(col("_field").eq(lit("f1")).or(col("_field").eq(lit("f2"))))
+                .build(),
+        )
+        .unwrap();
+
+        let expected = PredicateBuilder::new()
+            .field_columns(vec!["f1", "f2"])
+            .add_expr(lit(true))
+            .build();
```
NGA-TRAN, line 421: 👍

predicate/src/rpc_predicate.rs
NGA-TRAN: Yeah, I ask the same question above. Syntax error?

predicate/src/rpc_predicate.rs
```diff
@@ -395,56 +379,97 @@ mod tests {
...
+            &PredicateBuilder::new()
+                .add_expr(col("_field").not_eq(lit("f1")))
+                .build(),
+        )
+        .unwrap();
+
+        let expected = PredicateBuilder::new()
+            .field_columns(vec!["f2"])
+            .add_expr(lit(true))
+            .build();
```
NGA-TRAN, line 440: 👍

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -0,0 +1,573 @@
...
+                    .or(field_ref().eq(lit("f2"))),
+                "Unsupported structure with _field reference",
+            ),
+            (
+                // mismatched OR
+                // (_field != f1) OR (_field != f3)
+                field_ref()
+                    .not_eq(lit("f1"))
+                    .or(field_ref().not_eq(lit("f3"))),
+                "Unsupported structure with _field reference",
```
NGA-TRAN, line 504: 👍

NGA-TRAN (approved):
The PR does what it means 👍

predicate/src/rpc_predicate.rs
```diff
@@ -23,11 +26,18 @@ use std::sync::Arc;
...
+/// projections in influx RPC predicates.
 ///
-/// This is required to support predicates like
-/// `_field` = temperature
+/// For example, a predicate like
+/// ```text
+/// _field = temperature
+/// ```
+///
+/// Means to select only the (field) column named "temperature"
```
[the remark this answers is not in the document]
alamb: that is correct

And even more fun `_field = col_a OR _field = col_b` means `select col_a, col_b` 🤯

predicate/src/rpc_predicate.rs
alamb: > What happens if these to EQs are linked via AND instead of OR?

Yes, I agree  it should throw an error (which is also covered in the field_rewrite module) but I will add a test demonstrating that here as well.

predicate/src/rpc_predicate.rs
alamb: My intention was to more fully cover the cases in the tests of predicate/src/rpc_predicate/field_rewrite.rs -- the tests in this module were mostly meant to test that the logic is connected correctly (especially for multiple exprs / various patterns of connection with AND and OR)

predicate/src/rpc_predicate/field_rewrite.rs
alamb: yes, that is correct.

predicate/src/rpc_predicate/field_rewrite.rs
alamb: It searches for things like `_field ~ 'foo'` (aka that weren't matched by the rewriter) rather than invalid AND/OR structures. I couldn't find an easy way to combine them -- I add some comments trying to clarify

predicate/src/rpc_predicate/field_rewrite.rs
alamb: The reason that I wrote `try_add_...` first and then during testing I realized during the rewrite (when `try_add_...` is called) it is too late to catch all the errors, so I implemented `FieldRefChecker` 

Now these errors are caught before `try_add_...`  is even called; I will change `try_add_...` from an error to a `unreachable!`

predicate/src/lib.rs
```diff
@@ -99,6 +99,21 @@ impl Predicate {
...
+    pub fn add_field_names<I, S>(&mut self, new_names: I)
+    where
+        I: IntoIterator<Item = S>,
+        S: Into<String>,
+    {
+        let mut new_field_columns = self.field_columns.take().unwrap_or_default();
+
+        let new_names = new_names.into_iter().map(|n| n.into());
+
+        new_field_columns.extend(new_names);
```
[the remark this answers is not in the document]
alamb: I think it is actually (non obviously) using a `BTree` structure already

```rust
    /// Optional field restriction. If present, restricts the results to only
    /// tables which have *at least one* of the fields in field_columns.
    pub field_columns: Option<BTreeSet<String>>,
```

I will add some type annotations to make this clearer and some tests to validate

PR #4722, opened 2022-05-27
feat: Cache read buffer chunks

carols10cents:
Connects to #4681. 

This adds a new type parameter to the cache loader so that, if we need to load the chunk from the parquet file, we can use the other information we already have about that parquet file rather than needing to look it up in the catalog again.

The cache still stores the chunk by parquet file ID.

This still isn't actually hooked up yet as there's only the most rudimentary tests. ~~Namely, I haven't verified the LRU behavior at all.~~

### DIFF
--- a/querier/src/cache/read_buffer.rs
+++ b/querier/src/cache/read_buffer.rs
@@ -11,21 +11,23 @@ use cache_system::{
     driver::Cache,
     loader::{metrics::MetricsLoader, FunctionLoader},
 };
-use data_types::{ParquetFile, ParquetFileId};
+use data_types::ParquetFileId;
 use datafusion::physical_plan::SendableRecordBatchStream;
 use futures::StreamExt;
 use iox_time::TimeProvider;
-use parquet_file::chunk::DecodedParquetFile;
+use parquet_file::{chunk::DecodedParquetFile, storage::ParquetStorage};
 use read_buffer::RBChunk;
 use snafu::{ResultExt, Snafu};
 use std::{collections::HashMap, mem, sync::Arc};
 
 const CACHE_ID: &str = "read_buffer";
 
+type ExtraFetchInfo = (Arc<DecodedParquetFile>, Arc<str>, ParquetStorage);
+
 /// Cache for parquet file data decoded into read buffer chunks
 #[derive(Debug)]
 pub struct ReadBufferCache {
-    cache: Cache<ParquetFileId, Arc<RBChunk>, ()>,
+    cache: Cache<ParquetFileId, Arc<RBChunk>, ExtraFetchInfo>,
 
     /// Handle that allows clearing entries for existing cache entries
     _backend: SharedBackend<ParquetFileId, Arc<RBChunk>>,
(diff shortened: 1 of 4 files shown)


cache_system/src/driver.rs
```diff
@@ -297,19 +300,19 @@ mod tests {
     async fn test_answers_are_correct() {
         let (cache, _loader) = setup();
 
-        assert_eq!(cache.get(1).await, String::from("1"));
-        assert_eq!(cache.get(2).await, String::from("2"));
+        assert_eq!(cache.get(1, ()).await, String::from("1"));
+        assert_eq!(cache.get(2, ()).await, String::from("2"));
```
crepererum, line 304: Would be nice if the extra would also be tested. Maybe make it a `bool` change the the loader from `k.to_string()` to `format!("{k}_{extra}")`?

cache_system/src/loader/mod.rs
```diff
@@ -11,46 +11,51 @@ pub trait Loader: std::fmt::Debug + Send + Sync + 'static {
     /// Cache key.
     type K: Send + 'static;
 
+    /// Extra data needed when loading a missing entry. Specify `()` if not needed.
+    type Extra: Send + 'static;
```
crepererum, line 15: @alamb and I talked about this feature before and I think it is kinda helpful, not only for your application but later also for cache updates (i.e. load a new value using the old value as a baseline).

One thing that I do NOT like about this approach is that at least until now, in theory (!!!) knowing the key was sufficient to construct the value. This property will be very helpful for debugging, since you could construct a cache that resembles a prod system locally just by iterating the keys -- or at least this was an idea I had. I know that in practice this property is already kinda violated right now because we cache object that are mutable (e.g. a namespace schema). So I probably need to rethink how to handle cache "transfers" / "reconstructions".

querier/src/chunk/mod.rs
```diff
@@ -257,7 +257,7 @@ pub struct ChunkAdapter {
     /// Cache
     catalog_cache: Arc<CatalogCache>,
 
-    /// Object store.
+    /// Object store. Wrapper around an Arc; cheap to clone.
     store: ParquetStorage,
```
crepererum, line 261: Where is this `Arc`?

crepererum (commented):
Could you split of the `cache_system` change into its own PR for easier review? The actual core change (= "cache read buffer chunks") kinda gets lost in the noise, esp. with all the other changes that the `cache_system` change requires (= adding the "extra" parameter everywhere).

querier/src/chunk/mod.rs
carols10cents: https://github.com/influxdata/influxdb_iox/blob/988bd38e939be177ee09299185f05f34ab64e683/parquet_file/src/storage.rs#L76-L80

carols10cents:
> Could you split of the `cache_system` change into its own PR for easier review?

Sure! Switching this PR into draft mode as it'll depend on the cache system one I'm about to open.

querier/src/chunk/mod.rs
crepererum: Ahh, then maybe write:

```
/// Object store.
///
/// Internally `ParquetStorage` wraps the actual store implementation into an `Arc`, so `ParquetStorage` is cheap to clone.
```

carols10cents:
This is ready for review, and also now has a test for the LRU behavior of the read buffer cache.

querier/src/cache/read_buffer.rs
```diff
@@ -11,21 +11,23 @@ use cache_system::{
...
 use iox_time::TimeProvider;
-use parquet_file::chunk::DecodedParquetFile;
+use parquet_file::{chunk::DecodedParquetFile, storage::ParquetStorage};
 use read_buffer::RBChunk;
 use snafu::{ResultExt, Snafu};
 use std::{collections::HashMap, mem, sync::Arc};
 
 const CACHE_ID: &str = "read_buffer";
 
+type ExtraFetchInfo = (Arc<DecodedParquetFile>, Arc<str>, ParquetStorage);
```
crepererum, line 25: Since `Arc<str>` is basically untyped, either document clearly that it is a `table_name` or (IMHO better) use a struct instead.

crepererum (approved):
Small nit, rest looks good.

I would say: even when our current prod querier is unstable and OOMes all the time, let's fail forward and deploy this thing :shipit:

Issue #4696, opened 2022-05-25
Possible compactor deadlock


domodwyer:
Logs for compactors 0, 2 & 3 in prod show they progress up until they begin executing the query plan, and then appear to deadlock - they emit no further logs, do not use any significant amounts of CPU or RAM, or upload/download any files from the object store.

NGA-TRAN:
C0: possibly deadlock deadlock

. Compacting partition candidates: n_candidates=8098
. Actuall selected partitions to compact: n_compactions=7
. Num files and columns of those selected partitions
  -- num_files: 126, 236, 114, 136, 373, 257, 2
  -- num_cols: 9, 12, 15, 15, 14, 9, 51

C1: panic 
. Compacting partition candidates: n_candidates=10438
. Actuall selected partitions to compact: n_compactions=1
. Num files and columns of those selected partitions
  -- num_files: 218
  -- num_cols: 72 

[...]

PR #4727, opened 2022-05-30
fix: do NOT block in parquet file IO

crepererum:
I think for historical reason we were using blocking IO to read parquet
files. With the current streaming `SendableRecordStream` approach this
is technically NOT required anymore.

Now one might think that the sync-async dance that we did is kinda
harmless, but looking at our producition querier I think it is really
bad. The querier seems to be stuck but looking at `strace` and other
health signal it seems it is not entirely dead. Looking at GDB
backtraces it seems that nearly all threads are busy in
`download_and_scan_parquet`. Looking at the tokio docs
(<https://docs.rs/tokio/1.18.2/tokio/task/fn.spawn_blocking.html>)
for `spawn_blocking` (which is used to start the sync download) this
makes sense: tokio only starts replacement threads for the current
runtime thread (which calls `spawn_blocking`) if this does NOT exceed the
runtime thread limit. However we set the runtime thread limit to the
number of CPU cores available to IOx, so this is a limiting factor. This
means that there are only a few threads left to do actual work (I've
seen postgres data flowing back and forth for example) but tokio is not
able to use its full potential anymore. This is esp. bad because the
sync code in `download_and_scan_parquet` then uses `futures` `block_on`
functionality to call back into async code, so it waits for tokio
itself.

The change is rather simple: just use async task spawns.


crepererum:
I forgot: as a bonus, query cancellation will now properly cancel downloads as well.

parquet_file/src/storage.rs
```diff
@@ -191,20 +191,18 @@ impl ParquetStorage {
...
+        let object_store = Arc::clone(&self.object_store);
+        let handle = tokio::task::spawn(async move {
+            let download_result =
+                download_and_scan_parquet(projection, path, object_store, tx.clone()).await;
+
+            // If there was an error returned from download_and_scan_parquet send it back to the receiver.
+            if let Err(e) = download_result {
+                warn!(error=%e, "Parquet download & scan failed");
+                let e = ArrowError::ExternalError(Box::new(e));
+                if let Err(e) = tx.blocking_send(ArrowResult::Err(e)) {
```
tustvold, line 203: ?

tustvold (approved):
Makes sense to me :+1:

parquet_file/src/storage.rs
```diff
@@ -283,7 +280,7 @@ fn download_and_scan_parquet(
     let record_batch_reader = arrow_reader.get_record_reader_by_columns(projection, batch_size)?;
 
     for batch in record_batch_reader {
```
tustvold, line 282: Possibly worth highlighting this will perform blocking IO on a tokio worker thread, I think this is probably fine

parquet_file/src/storage.rs
crepererum: fixed

parquet_file/src/storage.rs
crepererum: Now I remember where the problem was and why we have this construct. I'll try to fix that...

alamb (commented):
Nice find

alamb:
FWIW I also added this to the list of items on https://github.com/influxdata/influxdb_iox/issues/4392

domodwyer (approved):
For what it's worth, this code was my primary suspicion for https://github.com/influxdata/influxdb_iox/issues/4696 :+1:

parquet_file/src/storage.rs
crepererum: done, it uses a better (but slightly more complex) solution now

parquet_file/src/storage.rs
```diff
@@ -243,55 +260,105 @@ fn column_indices(selection: Selection<'_>, schema: SchemaRef) -> Vec<usize> {
...
+impl ParquetBlockingReader {
+    fn new(
+        file: std::fs::File,
+        tx: tokio::sync::mpsc::Sender<ArrowResult<RecordBatch>>,
+        projection: Vec<usize>,
+        thread_pool: Arc<ThreadPool>,
+    ) -> Result<Self, ReadError> {
+        let (tx_finished, finished) = tokio::sync::oneshot::channel();
+
+        thread_pool.spawn(move || {
```
domodwyer, line 320: What happens if all the threads are blocked doing I/O already - does this spawn more threads, or does it queue jobs up?

crepererum: From reading <https://docs.rs/rayon/1.5.3/rayon/fn.spawn.html> I assume it would NOT block if the pool is used. Playground confirms this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=356904780ed7d8a6d529da3facd7470b

However I just figured out that we may not cancel jobs that are stuck within the queue, I'll fix that. (you cannot really cancel sync threaded jobs, but we can at least try our best)

crepererum: done

domodwyer: An interesting note is that it appears tasks are executed in LIFO order unless stolen by a free thread, so it could penalise some I/O tasks with higher latency / uneven latency distribution.

I don't think this is likely/a problem given the orders of magnitude higher latency we'll get from the actual I/O itself. There is a [FIFO](https://docs.rs/rayon-core/1.9.3/rayon_core/struct.ThreadPool.html#method.spawn_fifo) mode though.

alamb: I think @tustvold 's longer term work in DataFusion / ObjectStore will address many  of the concerns in this thread so it is not worth over optimizing at this poin

PR #4729, opened 2022-05-30
refactor(compactor): consistent sort key

domodwyer:
Tidies up the compactor to avoid repeatedly querying for the parition `SortKey` (or using any potentially different values across calls) and asserts a whole bunch of invariants the system assumed to be true.

---

* refactor: consistent SortKey source (364cbc0b2)

      Changes the compaction logic to always reference the same SortKey instance,
      rather than repeatedly querying for it.

      The Partition metadata is always read from the catalog as part of 
      compact_partition(), where it previously threw away all metadata except the
      sort key, which was passed into compact(). Then compact() would always
      re-query the catalog to look up just the sort key again, and mix up the two
      instances during use - one passed into the fn, one freshly queried within the
      fn.

      Now the Partition metadata is resolved in compact_partition() as it was 
      previously, but the entire Partition reference is passed to compact(), and
      this is consistently used do access the sort key. This also removes a catalog
      query per compaction call.

* refactor: assert consistent parquet file metadata (6f75667b1)

      Assert consistent metadata when evaluating candidate parquet files for 
      compaction.

      Asserts all files have the same:
         * Sequencer ID
         * Namespace ID
         * Table ID
         * Partition ID
         * Sort key


compactor/src/compact.rs
```diff
@@ -438,13 +437,7 @@ impl Compactor {
         info!("compacting partition {}", partition_id);
         let start_time = self.time_provider.now();
 
-        let sort_key_from_catalog = self
-            .sort_key_from_catalog(partition_id)
-            .await?
```
domodwyer, line 443: This used to query the catalog once per `compact()` call - the sort key was always passed into the func though 🤷

compactor/src/compact.rs
```diff
@@ -665,22 +666,26 @@ impl Compactor {
...
                         partition_id_2: file.data.partition_id
                     }
-                )
+                );
+
+                assert_eq!(file.data.sequencer_id, sequencer_id);
+                assert_eq!(file.data.namespace_id, namespace_id);
+                assert_eq!(file.data.table_id, table_id);
+                assert_eq!(file.data.partition_id, partition.id);
+                assert_eq!(file.iox_metadata().sort_key, parquet_sort_key);
```
domodwyer, line 675: These asserts ensure the assumptions made in the rest of this code hold - all files must have the same sequencer/namespace/table/sort key.

Partition ID was always (and still is) checked above in a slightly different way, but can't hurt to assert here too :+1:

compactor/src/compact.rs
alamb: `get_partition_from_catalog` still queries the catalog once per compact() call, right? The difference is that now it returns the entire partition.

compactor/src/compact.rs
```diff
@@ -665,22 +663,25 @@ impl Compactor {
                         table_id_2: file.data.table_id,
                         partition_id_2: file.data.partition_id
                     }
-                )
+                );
+
+                assert_eq!(file.data.sequencer_id, sequencer_id);
```
alamb, line 668: 👍

alamb (approved):
Makes sense to me -- I think it would be good if @NGA-TRAN  could also give it a look

compactor/src/compact.rs
domodwyer: No - it's called once per call to `compact_partition()`, whereas it used to do this, plus one call per file group (`compact()` call).

alamb: I was confusing `compact` and `compact_partition` 👍

PR #4737, opened 2022-05-31
feat: limit tmp parquet file count and size

### LINKED ISSUE #4736: Limit parquet IO tmp disk usage
Our prod pods get evicted due to disk pressure. Let's make sure that the `ParquetStorage` not only limits the number of threads it uses (already done) but also:

- the number of open files
- the total size of all tmp files

crepererum:
Fixes #4736.


crepererum:
Removing the hardcoded values is tracking in #4738.

mkmik (approved):
LGTM.

It would be nice to have some sort of observability for when we're hitting this semaphore

alamb:
FWIW I think this code was removed in https://github.com/influxdata/influxdb_iox/pull/4741

Issue #4740, opened 2022-05-31
Do not buffer parquet files to "disk" for reading


alamb:
__Use case:__
Now that we have a significant query load on our internal production cluster, @crepererum is  seeing kubernetes report: "node had condition: [DiskPressure]. events".

Upon investigation this disk pressure is coming from the temporary parquet file created here https://github.com/influxdata/influxdb_iox/blob/2bf03e57bfb5d722701b005ccc4045fec5f58d0b/parquet_file/src/storage.rs#L243-L271

There are a myriad of options and tradeoffs related to Kubernetes and local disk (see @mkmik 's https://github.com/influxdata/k8s-idpe/pull/7775#discussion_r885393717) , but the summary I think is that "k8s works best when the application does not use significant amount of locally attached storage"

These files are simply written, read immediately  and then deleted. 

Note there is a larger plan, driven by @tustvold, and described variously in   https://github.com/apache/arrow-datafusion/issues/2489 https://github.com/apache/arrow-datafusion/issues/2504 https://github.com/apache/arrow-datafusion/issues/2205 and others to more fully integrate object store reads into query execution and minimize buffering needed

__Proposal:__

The proposal therefore is to:
1. Download the parquet files into a memory buffer rather a temp file
2. Read directly from that memory buffer (with something such as https://docs.rs/parquet/15.0.0/parquet/file/serialized_reader/struct.SliceableCursor.html)

__Alternatives considered:__

This ticket does not cover caching or potentially more sophisticated behaviors only the removal of using tmpfs

Note that without additional work, it is likely using memory buffers will cause other issues (like OOMs, etc). 

However, it will at least  unify the resource needs and better prepare us for the longer term plans

alamb:
I believe @crepererum  said he plans to work on this

PR #4741, opened 2022-05-31
refactor: do NOT use ANY file IO for parquet reading

crepererum:
Resolves https://github.com/influxdata/influxdb_iox/issues/4740


mkmik (approved):
LGTM.

Eventually we'll also want to limit the total amount of memory used for this and possibly apply some backpressure, but let's defer that

alamb (commented):
FWIW it looks good to me 👍  Less code

Issue #4695, opened 2022-05-25
Should not write no-data to parquet file which will cause no column meta data


domodwyer:
In prod (compactor1) we observed a panic introduced in #4678 caused by missing column statistics. The changes in #4678 depend on these statistics - it's likely they have always been missing, but this went previously undetected.

Specifically the call to [`read_statistics()`][read] after a successful compaction yields no column statistics and this causes a panic. As [documented] this _can be_ triggered when the `Schema` embedded within the Parquet metadata is missing IOx-specific metadata.

<details><summary>
Panic
</summary>

```
thread 'tokio-runtime-worker' panicked at 'invalid statistics: NoRowGroup', parquet_file/src/metadata.rs:397:14
stack backtrace:
   0: rust_begin_unwind
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::expect
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1035:23
   4: parquet_file::metadata::IoxMetadata::to_parquet_file
             at ./influxdb_iox/parquet_file/src/metadata.rs:395:28
   5: compactor::utils::CatalogUpdate::new
             at ./influxdb_iox/compactor/src/utils.rs:173:28
   6: compactor::compact::Compactor::compact_partition::{{closure}}::{{closure}}::{{closure}}
             at ./influxdb_iox/compactor/src/compact.rs:444:24
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
   8: <futures_util::stream::futures_unordered::FuturesUnordered<Fut> as futures_core::stream::Stream>::poll_next
             at ./usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/futures-util-0.3.21/src/stream/futures_unordered/mod.rs:514:17
   9: <S as futures_core::stream::TryStream>::try_poll_next
[...]

PR #4744, opened 2022-05-31
fix: let us not compact no-data

NGA-TRAN:
This will avoid panics in #4695 


domodwyer (approved):
🥳

PR #4747, opened 2022-05-31
chore: let aggressively increase compactor job size and concurrency level

NGA-TRAN:


clap_blocks/src/compactor.rs
```diff
@@ -36,30 +36,30 @@ pub struct CompactorConfig {
...
         long = "--compaction-max-size-bytes",
         env = "INFLUXDB_IOX_COMPACTION_MAX_SIZE_BYTES",
-        default_value = "10000000"
+        default_value = "100000000"
     )]
     pub compaction_max_size_bytes: i64,
 
     /// Limit the number of files per compaction
-    /// Default is 50
+    /// Default is 10
```
NGA-TRAN, line 58:
```suggestion
    /// Default is 100
```

NGA-TRAN:
@alamb 
There is no reason. I will figure out how to do it. Currently, we only do for one parameter which is here https://github.com/influxdata/k8s-idpe/pull/7780

alamb:
I didn't understand this was in addition to rather than instead of the change to production config

It would be fine in my opinion to change the defaults too -- whatever you think is best

alamb (approved):

Issue #4739, opened 2022-05-31
Expose semaphore metrics


crepererum:
~#4737 introduced two semaphores (one for file size, one for number of files).~ (this code is now gone but the generic problem still exists for other semaphores)

As @mkmik [pointed out](https://github.com/influxdata/influxdb_iox/pull/4737#pullrequestreview-990067371) it would be nice if we would have some metrics detecting that we hit these semaphores.

alamb:
I believe the relevant code was removed in https://github.com/influxdata/influxdb_iox/pull/4741 so this ticket can probably be closed

alamb:
I believe the relevant code was removed in https://github.com/influxdata/influxdb_iox/pull/4741 so this ticket can probably be closed

crepererum:
I think a similar problem applies to other async semaphores in our code base and I'm soooo close to implement a generic wrapper (similar to what we have for sync locks), so I would steal this ticket and change the scope a bit.

crepererum:
this is done (the only remaining async semaphore is the query concurrency limiter)

PR #4751, opened 2022-06-01
feat: add tooling to instrument async semaphores

crepererum:
Ref #4739.


alamb (approved):
LGTM -- the tests specially are 👌 very nice

PR #4752, opened 2022-06-01
feat: limit querier concurrency by limiting number of active namespaces

crepererum:
This is a rather quick fix for prod. On the mid-term we probably wanna
rethink our deployment strategy, e.g. by using "one query per pod" and
by deploying queryd w/ IOx into the same pod.


querier/src/database.rs
```diff
@@ -79,11 +102,21 @@ impl QuerierDatabase {
             exec,
             ingester_connection,
             query_log,
+            namespace_semaphore,
         }
     }
 
-    /// Get namespace if it exists
+    /// Get namespace if it exists.0usize
```
alamb, line 109:
```suggestion
    /// Get namespace if it exists.
```

alamb (commented):
I think the code does what it claims to do, but I am not so sure about what it wants to do.

Questions:
1. Does anything free namespaces (and thus drops permits)? 
2. Have you thought about the tradeoffs between waiting for a new permit vs erroring if the permit count is reached?

crepererum:
> 1. Does anything free namespaces (and thus drops permits)?

Yes, `QuerierNamespaces` are short-lived objects (i.e. they are dropped when the query ends), so dropping them at the end of the query will free the permit.


> 2. Have you thought about the tradeoffs between waiting for a new permit vs erroring if the permit count is reached?

We already error if the request times out (either client- or server-side).

clap_blocks/src/querier.rs
```diff
@@ -45,6 +45,16 @@ pub struct QuerierConfig {
         default_value = "1073741824"
     )]
     pub ram_pool_bytes: usize,
+
+    /// Number of namespaces that can be used at once.
+    ///
+    /// This limits the concurrent query load.
+    #[clap(
```
alamb, line 52: I recommend changing the name of this configuration to "max_concurrent_queries"

I think using `namespace` is confusing as it implies to me the limit applies to how many concurrent namspaces can exist (aka how may `org/buckets`) even though. two queries will actually count 2 against the limit

I think using `semaphore` is also confusing as it is an implementation detail in terms of some concept that doesn't exist elsewhere in IOx's external description that I know of

querier/src/database.rs
```diff
@@ -43,6 +44,13 @@ pub struct QuerierDatabase {
 
     /// Query log.
     query_log: Arc<QueryLog>,
+
+    /// Semaphore that limits the number of namespaces in used at the time by the query subsystem.
+    ///
+    /// This should be a 1-to-1 relation to the number of active queries.
```
alamb, line 50: this is a key point -- maybe you can name the parameter `query_execution_semaphore` or something

I wonder if it actually is the number of concurrent queries, or if this really limits the number of concurrent queries that can be planned (aka is the `QurerierDatabase` dropped once query execution begins?)

clap_blocks/src/querier.rs
crepererum: renamed.

querier/src/database.rs
crepererum: renamed.

PR #4754, opened 2022-06-01
feat: Add an Extra type to Cacher Loader to specify extra information…

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/4681.

This is only the cache system changes from #4722 to add the `Extra` type and test it.


crepererum (approved):
Cool stuff :sunglasses:

PR #4743, opened 2022-05-31
chore: Update datafusion + `arrow`/`parquet`/`arrow-flight` to `15.0.0`

alamb:
~Draft until https://github.com/apache/arrow-datafusion/pull/2674 is merged~

# Rationale
* Get the sweet sweet upstream code

# Changes:
1. ~Update to https://github.com/apache/arrow-datafusion/commit/acb245a9c1d4cf30a484a26152ba0489886a8474~
2. ~Update to TBD (after https://github.com/apache/arrow-datafusion/pull/2674 is merged)~
3. Update to https://github.com/apache/arrow-datafusion/commit/8ddd99c8432fdac2c236040973f984a4146f18b7
4. Update `arrow`/`parquet`/`arrow-flight` to `15.0.0`
2. Update APIs

Closes https://github.com/influxdata/influxdb_iox/pull/4733
Closes https://github.com/influxdata/influxdb_iox/pull/4734
Closes https://github.com/influxdata/influxdb_iox/pull/4735

PR #4755, opened 2022-06-01
chore: Remove uneeded workaround `cast_aggregates`

alamb:
Draft until https://github.com/influxdata/influxdb_iox/pull/4743 is merged as it requires code from there. 

This is a cleanup I noticed while working on https://github.com/influxdata/influxdb_iox/pull/4743

Basically Influx storage RPC wants `Int64` counts, but before https://github.com/apache/arrow-datafusion/pull/2636 DataFusion made `UInt64` counts so IOx had to cast counts to `I64`. Now that  DataFusion makes `I64` counts natively, the casts are no longer needed


alamb:
Tests were failing for some reason. I don't care enough to debug more. Closing

PR #4756, opened 2022-06-01
test(e2e): do not mangle prod database

domodwyer:
I had `INFLUXDB_IOX_CATALOG_DSN` set in my env while working on reproducing the compactor deadlock, and the database kept getting populated with unexpected values - turns out these tests were running against this DSN rather than the test DSN, executing migrations and writing stuff! :fire:

---

* test(e2e): do not mangle prod database (6b46f45f0)

      Adds maybe_skip_integration!() calls to the following CLI e2e tests:

         * default_mode_is_run_all_in_one
         * default_run_mode_is_all_in_one

      This prevents them from executing against the "prod" catalog, running 
      migrations and inserting values to the prod database specified in the prod DSN
      env (INFLUXDB_IOX_CATALOG_DSN). These tests are now consistent with the rest
      of the e2e tests in that they use the test DSN env
      (TEST_INFLUXDB_IOX_CATALOG_DSN) to source the catalog DSN.


influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -10,9 +10,10 @@ use test_helpers_end_to_end::{maybe_skip_integration, MiniCluster, Step, StepTes
 
 #[tokio::test]
 async fn default_mode_is_run_all_in_one() {
+    let catalog_dsn = maybe_skip_integration!();
     Command::cargo_bin("influxdb_iox")
         .unwrap()
-        .args(&["-v"])
+        .args(&["-v", "--catalog-dsn", catalog_dsn.as_str()])
```
alamb, line 16: This test also ensured that one mode runs against an in-memory database by default -- though of course setting in the environment will change that 😰 

This PR will change the test to run against a postgres db.

Can we also please add a test that checks that by default all in one runs against in-memory? Perhaps asserting / failing of the `INFLUXDB_IOX_CATALOG_DSN` is configured?


You can see what happens locally for me when running all in one (using memory catalog)
```shell
(arrow_dev) alamb@MacBook-Pro-3:~/Software/influxdb_iox$ ./target/debug/influxdb_iox run -v
2022-06-01T14:54:06.689912Z  INFO clap_blocks::object_store: Object store: In-memory
2022-06-01T14:54:06.692051Z  INFO clap_blocks::write_buffer: Write buffer: File-based in `/var/folders/s3/h5hgj43j0bv83shtmz_t_w400000gn/T/.tmpMOj7LG`
2022-06-01T14:54:06.692748Z  INFO clap_blocks::catalog_dsn: Catalog: In-memory
2022-06-01T14:54:06.710694Z  INFO influxdb_iox::commands::run::all_in_one: running db migrations
2022-06-01T14:54:06.712667Z  INFO influxdb_iox::commands::run::all_in_one: Creating shared query executor num_threads=16
2022-06-01T14:54:06.723332Z  INFO influxdb_iox::commands::run::all_in_one: starting router
2022-06-01T14:54:06.738685Z  INFO ioxd_router: connected to write buffer topic topic="iox-shared" shards=1
2022-06-01T14:54:06.748612Z  INFO influxdb_iox::commands::run::all_in_one: starting ingester
2022-06-01T14:54:06.754334Z  INFO ingester::handler: ingester handler and lifecycle started with config LifecycleConfig { pause_ingest_size: 107374182400, persist_memory_threshold: 1073741824, partition_size_threshold: 314572800, partition_age_threshold: 1800s, partition_cold_threshold: 300s }
2022-06-01T14:54:06.756804Z  INFO influxdb_iox::commands::run::all_in_one: starting compactor
2022-06-01T14:54:06.759779Z  INFO compactor::handler: compactor started with config CompactorConfig { split_percentage: 90, max_concurrent_compaction_size_bytes: 100000, compaction_max_size_bytes: 100000, compaction_max_file_count: 10 }
2022-06-01T14:54:06.759848Z  INFO influxdb_iox::commands::run::all_in_one: starting querier ingester_addresses=["http://127.0.0.1:8083"]
2022-06-01T14:54:06.772159Z  INFO influxdb_iox::commands::run::all_in_one: starting all in one server
2022-06-01T14:54:06.772966Z  INFO influxdb_iox::commands::run::main: InfluxDB IOx server starting git_hash="cb2e74ad4886881fab5d9933b5350a92c269e932" num_cpus=16 build_malloc_conf=
```

domodwyer: I don't want tests to fail if `INFLUXDB_IOX_CATALOG_DSN` is set - that's actually how I discovered it was the tests that were mangling my DB. 

Running IOx and passing everything via flags is very verbose/repetitive/annoying so I usually export them in one terminal where I'm running IOx over-and-over, and then forget about it. I'd still like to be able to run the tests though! This is why we have a separate test DSN.

I can change this to specify the in-memory catalog explicitly to ensure it works - is it important to assert that it is the default? Should it even be the default? Maybe we should opt into memory-based state to prevent situations like https://github.com/influxdata/influxdb_iox/pull/4699. I could also unset the prod env in the test before running to check the default is in-mem though if it's important 🤷

domodwyer:
Personal env workflows / what should be default aside, this now resolves my original issue and should keep @alamb happy I believe :+1:

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -10,6 +10,11 @@ use test_helpers_end_to_end::{maybe_skip_integration, MiniCluster, Step, StepTes
 
 #[tokio::test]
 async fn default_mode_is_run_all_in_one() {
+    // Do not attempt to connect to the real DB if the prod DSN is set - all
+    // other tests use TEST_INFLUXDB_IOX_CATALOG_DSN but this one will use the
+    // real env.
+    std::env::remove_var("INFLUXDB_IOX_CATALOG_DSN");
```
alamb, line 16: Now won't this affect all tests? So like some might run with a postgres catalog and some might run without it?

alamb: @domodwyer and I had a chat -- and I think the agreement was that using https://docs.rs/assert_cmd/2.0.4/assert_cmd/cmd/struct.Command.html#method.env_clear would be a less racy and less error prone solution

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -10,6 +10,11 @@ use test_helpers_end_to_end::{maybe_skip_integration, MiniCluster, Step, StepTes
...
+
     Command::cargo_bin("influxdb_iox")
         .unwrap()
         .args(&["-v"])
```
alamb, line 20:
```suggestion
        .args(&["-v"])
        // Ensure no environment variable (like postgres DSN) are passed
        .env_clear()
```

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -21,6 +26,11 @@ async fn default_mode_is_run_all_in_one() {
...
 #[tokio::test]
 async fn default_run_mode_is_all_in_one() {
+    // This test is designed to assert the default running mode is using
+    // in-memory state, so ensure that any outside config does not influence
+    // this.
+    std::env::remove_var("INFLUXDB_IOX_CATALOG_DSN");
+
     Command::cargo_bin("influxdb_iox")
         .unwrap()
         .args(&["run", "-v"])
```
alamb, line 36:
```suggestion
        .args(&["run", "-v"])
        // Ensure no environment variable (like postgres DSN) are passed
        .env_clear()

```

alamb (commented):
Thank you

PR #4759, opened 2022-06-01
refactor: always panic for empty parquet files

domodwyer:
@NGA-TRAN as discussed :+1:

---

* refactor: always panic for empty parquet files (0bfc11f4a)

      Moves the panic into the child call to_parquet() so all code paths are covered
      (i.e. not serialising into memory via to_parquet_bytes()).

* test: assert panic behaviour (f8b83c508)

      Modifies the existing test added as part of #4695 to ensure a panic is emitted
      when serialising an empty parquet file.

### DIFF
--- a/parquet_file/src/serialize.rs
+++ b/parquet_file/src/serialize.rs
@@ -107,7 +107,12 @@ where
         writer.write(&maybe_batch?)?;
     }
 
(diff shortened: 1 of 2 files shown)


NGA-TRAN (approved):
Look great. Thanks @domodwyer

PR #4770, opened 2022-06-02
feat: add owned versions of instrumented semaphores

crepererum:
Owned versions will be required to instrument the query concurrency
limiter.


tracker/src/async_semaphore.rs
```diff
@@ -99,7 +99,9 @@ impl AsyncSemaphoreMetrics {
 #[derive(Debug)]
 pub struct InstrumentedAsyncSemaphore {
     /// Underlying sempahore implementation.
-    inner: Semaphore,
+    ///
+    /// This is wrapped into an [`Arc`] so we can use a single implementation for the owned and non-owned implementation.
+    inner: Arc<Semaphore>,
```
alamb, line 104: I wonder how important is it to have owned/not owned -- maybe we can just always use owned for code simplicity

tracker/src/async_semaphore.rs
```diff
@@ -257,13 +289,25 @@ pub struct InstrumentedAsyncSemaphorePermit<'a> {
...
         self.metrics.permits_acquired.dec(self.n as u64);
     }
 }
 
+/// An instrumented wrapper around [`tokio::sync::SemaphorePermit`].
+#[derive(Debug)]
+pub struct InstrumentedAsyncSemaphorePermit<'a> {
+    /// Use the owned variant so we can use a single implementation.
+    #[allow(dead_code)]
+    owned_permit: InstrumentedAsyncOwnedSemaphorePermit,
```
alamb, line 304: I don't really understand the usecase of having something that pretends it doesn't own the underling semaphore but really does 🤔

alamb (approved):

crepererum:
Regarding "why the heck do we need a non-owned version":

Normally you want the non-owned version because the semaphore is usually attached to *something* and shouldn't be moved around independently to avoid state confusion. In certain distributed scenarios it is easier though detach the permit from its origin (with the risk that this introduces state confusion) and use an owned version.

tracker/src/async_semaphore.rs
crepererum: The fact that it uses an owned version under the hood is a mere implementation detail.

alamb:
> Regarding "why the heck do we need a non-owned version":

Got it -- makes sense -- thank you for the clarification. I propose adding this information as comments to https://github.com/influxdata/influxdb_iox/pull/4775 in the hopes it will help a future reader (maybe even just my future self)

PR #4779, opened 2022-06-03
feat: compact all overlapped files no matter how large they are

### LINKED ISSUE #4596: Improve performance of compact_partition on large input files
This is a task of #4594 and #4585

# Rationale
The two charts below show that it takes around 2 hours to compact a large partition. This ticket is to investigate and improve it. Refer to #4594 for initial analysis when there is only one large partition is compacted

Screen Shot 2022-05-12 at 3 17 06 PM

Screen Shot 2022-05-12 at 3 40 33 PM

### How to see performance flame chart
 kubectl --context dex@prod01-eu-central-1-aws -n iox port-forward iox-shared-compactor-0-0 8080 
 http://localhost:8080/debug/pprof/profile?seconds=30.  (to see 30 seconds)

NGA-TRAN:
Close #4596 
Add "compact all overlapped files no matter how large they are" back but as an option for us to choose to do it or not. This is on by default


compactor/src/compact.rs
```diff
@@ -361,10 +361,13 @@ impl Compactor {
 
     /// Group files to be compacted together and level-0 files that will get upgraded
     /// for a given partition.
-    /// The number of compacting files per group will be limited by thier total size and number of files
+    /// If split_large_overlaps is false, all overlapped files will be in the same group no matter
```
NGA-TRAN, line 364:
```suggestion
    /// If compaction_split_large_overlaps is false, all overlapped files will be in the same group no matter
```

compactor/src/compact.rs
```diff
@@ -361,10 +361,13 @@ impl Compactor {
...
+    /// how many and how larg ethey are. Otherwise, he number of compacting files per group will be
```
NGA-TRAN, line 365:
```suggestion
    /// how many and how large they are. Otherwise, the number of compacting files per group will be
```

compactor/src/compact.rs
```diff
@@ -361,10 +361,13 @@ impl Compactor {
...
+    /// limited by thier total size and number of files
```
NGA-TRAN, line 366:
```suggestion
    /// limited by their total size and number of files
```

compactor/src/compact.rs
```diff
@@ -2326,6 +2342,76 @@ mod tests {
         assert!(groups[4].contains(&alone));
     }
 
+    // This tests
```
NGA-TRAN, line 2345: This test is similar to the one right above it but with config split_large_overlaps false and thus the output is different. This is the main test of this PR.

pauldix (commented):
Not sure I understand what the goal of this functionality is, can you offer a bit of explanation of why this is needed?

clap_blocks/src/compactor.rs
```diff
@@ -62,4 +62,14 @@ pub struct CompactorConfig {
...
+
+    /// If true split a group of many or large overlapped files into smaller groups based on
+    /// INFLUXDB_IOX_COMPACTION_MAX_SIZE_BYTE and
+    /// Default is 0 which means false
+    #[clap(
+        long = "--compaction-split-large-overlaps",
+        env = "INFLUXDB_IOX_COMPACTION_SPLIT_LARGE_OVERLAPS",
+        default_value = "0"
+    )]
+    pub compaction_split_large_overlaps: i32,
```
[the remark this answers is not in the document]
NGA-TRAN: When the flag is 0 (false) which is the default, we always compact overlapped files no mater how large they are.

I have a suspicion that in near future, when we run into customer use cases that want large partitions but each has many files. This flag is for that purpose since w already have the code to support that. However, I do not mind to remove this flag and always compact them.

compactor/src/compact.rs
```diff
@@ -387,14 +388,9 @@ impl Compactor {
 
         // Group overlapped files
         // Each group will be limited by thier size and number of files
-        let overlapped_file_groups = Self::overlapped_groups(
-            parquet_files,
-            compaction_max_size_bytes,
-            compaction_max_file_count,
-        );
+        let overlapped_file_groups = Self::overlapped_groups(parquet_files);
```
NGA-TRAN, line 391: Since overlapped files are now always in the the same group to get compacted, no need to consider size and file count

compactor/src/compact.rs
```diff
@@ -832,71 +828,18 @@ impl Compactor {
         Ok(())
     }
 
-    // Split overlapped groups into smaller groups if there are so mnay files in each group or
-    // their size are too large. The files are sorted by their min time before splitting so files
-    // are guarannteed to be overlapped in each new group
-    fn split_overlapped_groups(
```
NGA-TRAN, line 838: This is the purpose of this PR: do not split overlapped files anymore

compactor/src/compact.rs
```diff
@@ -2185,77 +2115,8 @@ mod tests {
         assert_eq!(groups, vec![vec![pf1, pf2]]);
     }
 
-    // This is specific unit test for split_overlapped_groups but it is a subtest of test_limit_size_and_num_files
-    // Keep all the variables the same names in both tests for us to follow them easily
-    #[test]
-    fn test_split_overlapped_groups() {
```
NGA-TRAN, line 2200: Remove corresponding tests

compactor/src/compact.rs
```diff
@@ -2286,16 +2147,14 @@ mod tests {
...
-        let max_size_bytes = 1000;
         let max_file_count = 2;
-        let overlapped_groups = Compactor::overlapped_groups(all, max_size_bytes, max_file_count);
-        // Must be 5
-        assert_eq!(overlapped_groups.len(), 5);
-        assert_eq!(overlapped_groups[0].parquet_files.len(), 2); // reach limit file count
-        assert_eq!(overlapped_groups[1].parquet_files.len(), 2); // reach limit file count
-        assert_eq!(overlapped_groups[2].parquet_files.len(), 1); // reach limit file size
-        assert_eq!(overlapped_groups[3].parquet_files.len(), 1); // reach limit file size
-        assert_eq!(overlapped_groups[4].parquet_files.len(), 1); // reach limit file size
```
NGA-TRAN, line 2307: No longer appropriate since all overlapped files will go into the same group

NGA-TRAN:
@pauldix: all overlapped files are in the same group now. This PR mostly remove code that split them before

pauldix (approved):
Great, is there a follow on issue to track having large groups output to multiple files based on the configured max size?

NGA-TRAN:
@pauldix 
> Great, is there a follow on issue to track having large groups output to multiple files based on the configured max size?

We no longer limit overlapped group by size. The configured max size is only used to limit the merge of continuous non-overlapped groups.

PR #4780, opened 2022-06-03
docs: Remove outdated references to the old MUB -> RUB -> OS chunk cycle

carols10cents:
Found these in query tests and then searched elsewhere while working on adding the read buffer cache.

Where possible, I tried to future-proof these comments by either removing them or making them a little more vague so maybe they won't need to be continually updated to remain accurate (famous last words, I know!)

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -6,7 +6,7 @@ Here are useful metrics
 
 ### Requests to IOx Server including Routers and Query Servers
 | Metric name |  Code Name | Description |
-| --- | --- | --- | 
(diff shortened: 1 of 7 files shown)


docs/metrics.md
```diff
@@ -26,9 +26,9 @@ Here are useful metrics
 ### Chunks
 | Metric name |  Code Name | Description |
```
alamb, line 27: my guess is that many more of these metrics have likely changed.

alamb (approved):
Nice cleanups

PR #4799, opened 2022-06-07
refactor: consolidate `PredicateBuilder` into `Predicate`

alamb:
# Rationale

`PredicateBuilder` was a wrapper over `Predicate`, and yet `Predicate` also has modification APIs. This was 

1. confusing as is not clear why some code is in `Predicate` and some is in `PredicateBuilder` and why
2. unecessairly verbose code

# Changes
1. Combines `PredicateBuilder`  into `Predicate`

Note I plan to consolidate the code / APIs some more as follow on PRs but this one was already pretty big


predicate/src/lib.rs
```diff
@@ -277,49 +299,17 @@ pub enum PredicateMatch {
     Unknown,
 }
 
-/// Structure for building [`Predicate`]s
```
alamb, line 280: This is now just inlined into `Predicate`

mkmik (approved):
Builder Lite pattern FTW :-)

https://matklad.github.io/2022/05/29/builder-lite.html

alamb:
> Builder Lite pattern FTW :-)

TIL a name for this. It works well for objects that have a natural `Default` I guess.

PR #4802, opened 2022-06-08
refactor: make `Cache` a trait

crepererum:
To insert more high-level metrics (e.g. cache misses/hits) it would be
helpful if we could easily instrument the layer right above the cache
driver (that combines the backend and the loader). To do that without
polluting the types too much, let's introduce a trait that describes the
driver interface and that we could later wrap with intrumentation.

This also pulls out the test into a generic setup, similar to how this
is done for the cache storage backends.

This does NOT include any functionality changes.


cache_system/src/cache/mod.rs
```diff
@@ -0,0 +1,50 @@
+//! Top-level trait ([`Cache`]) that provides a fully functional cache.
+//!
+//! Caches usually combine a [backend](crate::backend) with a [loader](crate::loader). The easiest way to archive that
```
alamb, line 3:
```suggestion
//! Caches usually combine a [backend](crate::backend) with a [loader](crate::loader). The easiest way to achieve that
```

cache_system/src/cache/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+///
+/// Canceling a [`get`](Self::get) request will NOT cancel the underlying loader. The data will
+/// still be cached.
+///
+/// # Panic
+///
+/// If the underlying loader panics, all currently running [`get`](Self::get) requests will panic.
+/// The data will NOT be cached.
+#[async_trait]
+pub trait Cache: Debug + Send + Sync + 'static {
```
alamb, line 33: 👍  makes sense to me

alamb (approved):

PR #4804, opened 2022-06-08
refactor: expose `CacheGetStatus` (and improve tests)

crepererum:
- add a `CacheGetStatus` which tells the user if the request was a hit
  or miss (or something inbetween)
- adapt some tests to use the status (only the tests where this could be
  relevant)
- move the test suite from using `sleep` to proper barriers (more stable
  under high load, more correct, potentially faster


cache_system/src/cache/mod.rs
```diff
@@ -12,6 +12,20 @@ pub mod driver;
 #[cfg(test)]
 mod test_util;
 
+/// Status of a [`Cache`] GET request.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CacheGetStatus {
+    /// The requested entry was present in the storage backend.
+    Hit,
+
+    /// The requested entry was NOT present in the storage backend and there was the loader had no query running.
```
alamb, line 21:
```suggestion
    /// The requested entry was NOT present in the storage backend and the loader had no previous query running.
```

cache_system/src/cache/mod.rs
```diff
@@ -12,6 +12,20 @@ pub mod driver;
...
+    Miss,
+
+    /// The requested entry was NOT present in the storage backend, but there was already a loader query running for
+    /// this particular key.
+    Queried,
```
alamb, line 26: What do you think about `Loading`,  or maybe `MissAlreadyLoading` as a possibly more precise name for this state. I normally think about the `Cache` as being queried, not doing the querying...

However, I also see the cache code already uses the term `Query` in several places so using `Queried` here for consistency is OK too

cache_system/src/cache/test_util.rs
```diff
@@ -60,16 +79,39 @@ where
...
     let n_blocked = loader.unblock();
     assert_eq!(n_blocked, 1);
 
-    assert_eq!(handle_1.await.unwrap(), String::from("1_true"));
-    assert_eq!(handle_2.await.unwrap(), String::from("1_true"));
+    assert_eq!(
+        handle_1.await.unwrap(),
+        (String::from("1_true"), CacheGetStatus::Miss),
+    );
+    assert_eq!(
```
alamb, line 111: 👍  nice

cache_system/src/cache/test_util.rs
```diff
@@ -285,3 +392,59 @@ impl Loader for TestLoader {
         format!("{k}_{extra}")
     }
 }
+
+#[async_trait]
+trait FutureExt2 {
```
alamb, line 397: Maybe `EnsurePending` would be a more descriptive name

cache_system/src/cache/test_util.rs
```diff
@@ -285,3 +392,59 @@ impl Loader for TestLoader {
...
+    type Out;
+
+    /// Ensure that futures is pending. In the pending case, try to pass the given barrier. Afterwards await the future again.
```
alamb, line 400:
```suggestion
    /// Ensure that the future is pending. In the pending case, try to pass the given barrier. Afterwards await the future again.
```

cache_system/src/cache/test_util.rs
```diff
@@ -285,3 +392,59 @@ impl Loader for TestLoader {
...
+}
+
+#[async_trait]
+trait AbortAndWaitExt {
+    /// Abort handle and wait for completion.
+    async fn abort_and_wait(&self);
+}
+
+#[async_trait]
+impl<T> AbortAndWaitExt for JoinHandle<T>
```
alamb, line 431: This looks a lot like  the existing `FutureTimeout` trait (provided by @domodwyer ) -- you can probably use that one instead

https://github.com/influxdata/influxdb_iox/blob/faa85a04ad2d65ae7807dedd2696829e6ccc6223/test_helpers/src/timeout.rs#L12-L30

alamb (approved):

cache_system/src/cache/test_util.rs
crepererum: They are NOT the same:

- this thing here is specific to `JoinHandle<T>`
- we call `abort()` to abort the task
- we do some dance to ensure that the task was actually cancelled (because calling `abort()` may not immediately cancel the task under all circumstances).

cache_system/src/cache/test_util.rs
crepererum: fixed

cache_system/src/cache/mod.rs
crepererum: done

cache_system/src/cache/test_util.rs
crepererum: done

PR #4806, opened 2022-06-08
feat: Log time spent requesting ingester partitions

### LINKED ISSUE #4558: Add Ingester information into `queries` system table
# Rationale
The `queries` system table have very useful information

```
./target/debug/influxdb_iox query 797a0180da80497c_4d2743db995d4bce 'select * from system.queries limit 5' 
+--------------------------------+------------+----------------------------------------------+--------------------+---------+----------+
| issue_time                     | query_type | query_text                                   | completed_duration | success | trace_id |
+--------------------------------+------------+----------------------------------------------+--------------------+---------+----------+
| 2022-05-10T19:06:50.643568525Z | sql        | select count(*) from cpu                     | 159.630996ms       | true    |          |
| 2022-05-10T19:07:14.039043931Z | sql        | select * from cpu order by time desc limit 1 | 10.882624993s      | true    |          |
| 2022-05-10T19:07:51.278477949Z | sql        | select * from cpu limit 1                    | 217.244192ms       | true    |          |
| 2022-05-10T19:10:08.592315942Z | sql        | show tables                                  | 916.615µs          | true    |          |
| 2022-05-10T19:11:20.872665428Z | sql        | show columns                                 | 239.367µs          | false   |          |
+--------------------------------+------------+----------------------------------------------+--------------------+---------+----------+
```

It would be great to add 3 more columns:
1. Time to communicate with Ingesters
2. Which Ingesters it communicates with
3. Which Ingesters return data (not empty)

carols10cents:
Fixes #4558.

First draft for @crepererum 's thoughts :)


carols10cents:
Ok, I've added another commit with a metrics-based approach to contrast with the `info!` log.

Please let me know if this is or is not what you were thinking!

@NGA-TRAN @pauldix @alamb @crepererum

alamb:
I defer to @NGA-TRAN  and @crepererum  as this is their request

NGA-TRAN (approved):
Looks good to me but @crepererum and @alamb want to review this because they know Querier better where to record the time.

I think when this is available you will be able to add this Ingester time (and total runtime) into this Querier Grafana dashboard https://grafana.prod01-eu-central-1.aws.influxdata.io/d/HhX2h_wnz/querier?orgId=1

querier/src/table/mod.rs
```diff
@@ -90,6 +94,32 @@ impl QuerierTable {
...
+                    640,
+                    1280,
+                    2560,
+                    5120,
+                    10240,
+                    20480,
+                    u64::MAX,
+                ])
+            },
+        );
```
crepererum, line 120: I think you wanna use a `DurationHistogram` here. That one automatically does the time conversion for you (IIRC the underlying metric is then also based on `ms`). The `DurationHistorgram` also comes w/ default options which we probably should use until we have a specific reason to diverge from.

crepererum (commented):
Apart from the histogram type nitpick: a test for the metric would be nice.

I BTW like the dual approach of issuing `debug` logs w/ a bit more info and pumping the data into a proper metric :+1:

querier/src/table/mod.rs
carols10cents: Oh neat!

carols10cents:
> Apart from the histogram type nitpick: a test for the metric would be nice.
> 
> I BTW like the dual approach of issuing `debug` logs w/ a bit more info and pumping the data into a proper metric 👍

Done and done!

querier/src/table/mod.rs
```diff
@@ -568,6 +605,16 @@ mod tests {
...
         assert_eq!(chunks[2].delete_predicates().len(), 0);
+
+        let histogram = metrics
+            .get_instrument::<Metric<DurationHistogram>>("ingester_ms")
+            .expect("failed to read metric")
+            .get_observer(&Attributes::from(&[]))
+            .expect("failed to get observer")
+            .fetch();
+
+        let hit_count = histogram.buckets.iter().fold(0, |acc, v| acc + v.count);
```
crepererum, line 616:
```suggestion
        let hit_count = histogram.sample_count();
```

querier/src/table/mod.rs
```diff
@@ -568,6 +605,16 @@ mod tests {
         assert_eq!(chunks[1].delete_predicates().len(), 2);
         // ingester in-mem chunk doesn't need predicates, because the ingester has already materialized them for us
         assert_eq!(chunks[2].delete_predicates().len(), 0);
+
+        let histogram = metrics
+            .get_instrument::<Metric<DurationHistogram>>("ingester_ms")
+            .expect("failed to read metric")
+            .get_observer(&Attributes::from(&[]))
+            .expect("failed to get observer")
+            .fetch();
```
crepererum, line 614: Could you maybe set up a test that only concerns the ingester chunks? This test here mostly concerns the state reconciling (i.e. how chunks from the ingester are factored into the state that we get from the catalog) and I'm also not a huge fan of adding more and more asserts and themes to a single test. 

Also see my note further up regarding failed requests on why a dedicated test might be helpful.

querier/src/table/mod.rs
```diff
@@ -191,6 +211,23 @@ impl QuerierTable {
             .await
             .context(GettingIngesterPartitionsSnafu)?;
```
crepererum, line 212: If I look at our current prod deployment, I think it would be helpful if the new metric would have an additional "dimension" (i.e. attribute) noting if the request was successful or not. The current implementation in this PR only takes successful requests into account.

Because this PR is open for quite a while, I leave it up to you if this should be fixed here or if you wanna delay it (using a ticket).

alamb:
I am going to be adding some more debugging so I want to get this code in. Thus merging. 🤠

PR #4811, opened 2022-06-09
feat: expose hit-miss metrics for querier caches

crepererum:
This should help us figuring out how our caches are doing.

You may wonder how/why we already have "cache miss" metrics in our prod dashboards. These are somewhat "hacky" since they are based on the timing histogram for the cache loader function (which conveniently also creates a counter). The loader calls and the "miss" metric should roughly match (they don't for canceled GET requests because we do NOT cancel the running loader calls).


cache_system/src/cache/metrics.rs
```diff
@@ -0,0 +1,353 @@
...
+            metric_get_miss_already_loading,
+            metric_get_cancelled,
+            metric_set,
+        }
+    }
+}
+
+/// Wraps given cache with metrics.
+#[derive(Debug)]
+pub struct MetricsCache<K, V, Extra>
```
alamb, line 62: A name such as `InstrumentedCache` or `CacheWithMetrics` would make it clearer to me that this structure wrapped a cache with metrics rather than was a cache for metrics. I don't feel strongly, just a suggestion.

cache_system/src/cache/metrics.rs
```diff
@@ -0,0 +1,353 @@
...
+        test_util::{AbortAndWaitExt, EnsurePendingExt, TestLoader},
+    };
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_generic() {
+        use crate::cache::test_util::test_generic;
+
+        test_generic(|loader| TestMetricsCache::new_with_loader(loader).cache).await;
```
alamb, line 175: It is so nice

alamb (approved):
BTW the more I see the Cache System in action the better I like it 🏆  -- I was definitely skeptical initially if the infrastructure was worth the effort at the beginning but as the last few weeks have demonstrated, I think the design is really working well. Nicely done.

cache_system/src/cache/metrics.rs
crepererum: done

Issue #4787, opened 2022-06-06
Partitions contain writes that don't map to the partition's partition key


domodwyer:
All writes in a partition should be for the same partition key.

We have 659,328 partitions that contain writes that span more than one partition key (at time of writing) :fire:

```sql
SELECT COUNT(id)
FROM
        parquet_file
WHERE
        to_timestamp(min_time / 1000000000) :: date != to_timestamp(max_time / 1000000000) :: date;
 count
--------
 659328
(1 row)
```

Lots of these seem to be potential rounding/timezone/off-by-one issues:

```
    id    | sequencer_id | table_id | partition_id | namespace_id | compaction_level |    min     |    max
----------+--------------+----------+--------------+--------------+------------------+------------+------------
 13908550 |           28 |   333125 |     21288956 |      4630294 |                0 | 2022-03-01 | 2022-03-02
 13908905 |           28 |   333125 |     21166535 |      4630294 |                0 | 2022-05-02 | 2022-05-03
 13908721 |           28 |   333125 |     21288972 |      4630294 |                0 | 2022-03-09 | 2022-03-10
 13908807 |           28 |   333125 |     21288985 |      4630294 |                0 | 2022-03-20 | 2022-03-21
```

However here are some more worrying examples:

```
    id    | sequencer_id | table_id | partition_id | namespace_id | compaction_level |    min     |    max
----------+--------------+----------+--------------+--------------+------------------+------------+------------
 10623921 |           36 |    13350 |     19051402 |          541 |                0 | 2022-04-28 | 2022-05-11
 14840098 |           21 |   105176 |     22112396 |       824781 |                1 | 1859-04-16 | 2022-06-03
 14876809 |           21 |   105176 |     22112396 |       824781 |                0 | 1859-04-16 | 2022-06-03
 15452719 |           13 |    99163 |     22277696 |          604 |                1 | 1798-10-17 | 2022-06-05
 15472059 |           13 |    99163 |     22277696 |          604 |                0 | 1798-10-17 | 2022-06-05
```

These queries show examples where the YMD derived from the min/max timestamps in the parquet file are not equal (they should always be equal for the current YMD partitioning scheme). There may be more examples where min/max in parquet files derive to the same, but incorrect, YMD partition key r.w.t the partition they are part of.

domodwyer:
I suspect solving this will also fix #[4761](https://github.com/influxdata/influxdb_iox/issues/4761).

domodwyer:
[...]

PR #4812, opened 2022-06-09
refactor(data_types): no Timestamp wraparound

domodwyer:
Minor hardening while looking into the timestamp issue #4787 - we'd never want to wrap around, though I think these are only used in tests.

---

* refactor(data_types): no Timestamp wraparound (d1436c9f0)

      This commit changes addition/subtraction of Timestamp values to panic if they
      would trigger under/overflow rather than silently wrapping around.

### DIFF
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -265,7 +265,7 @@ impl Sub<i64> for SequenceNumber {
     }
 }
 
-/// A time in nanoseconds from epoch
+/// A time in nanoseconds from epoch.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, sqlx::Type)]
 #[sqlx(transparent)]
 pub struct Timestamp(i64);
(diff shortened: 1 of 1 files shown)


data_types/src/lib.rs
```diff
@@ -280,19 +280,35 @@ impl Timestamp {
...
 
 impl Sub<i64> for Timestamp {
     type Output = Self;
 
-    fn sub(self, other: i64) -> Self {
-        Self(self.0 - other)
+    fn sub(self, rhs: i64) -> Self::Output {
+        self - Self(rhs)
     }
 }
```
crepererum, line 313: I know it existed before, but do we really want untyped additions to `Timestamp`? What does `timestamp + 100` even mean? Which unit is that the integer? (I guess the same as timestamp which means NS, but I still find it confusing).

Thinking about it more: we do we have yet another timestamp type here and don't just use `iox_time`? If it's just to prevent sqlx from leaking into `iox_time`, then IMHO just impl. `From`/`Into` and don't provide any additional functionality (esp. NO arithmetic) for the `data_types::Timestamp`.

PR #4813, opened 2022-06-09
feat: measure "seen" cache loads

crepererum:
To roughly gauge how much data we re-load into cached (i.e. data that
was already loaded but was later evicted due to LRU pressure or TTL
eviction) this change introduces a new metric that estimates if a cache
entry that is requested from the loader was already seen before (using a
probabilistic filter).


cache_system/Cargo.toml
```diff
@@ -10,6 +10,7 @@ iox_time = { path = "../iox_time" }
 metric = { path = "../metric" }
 observability_deps = { path = "../observability_deps" }
 parking_lot = "0.12"
+pdatastructs = { git = "https://github.com/crepererum/pdatastructs.rs.git", rev = "08f8a38ea8bc0444525cc4e1c9ab9dc2ad9ee82e" }
```
alamb, line 13: What is your long term plan for this pin -- do you want to release a real version of pbdstastructs and rely on it?

Three new dependencies 😢  I wonder if if it possible to copy in just the `BloomFilter` and `Filter parts to IOx to avoid the extra dependency?

alamb:
> @alamb I think this is what this metric does, since this is implemented in the loader and the loader is only hit for keys that are not currently stored (either because they are new or were p

Maybe renaming to `reloads` might make that clear

cache_system/Cargo.toml
crepererum: I'll cut a release and also trim the dependencies (e.g. see https://github.com/crepererum/pdatastructs.rs/pull/119) before merging this PR.

cache_system/src/loader/metrics.rs
```diff
@@ -79,7 +120,16 @@ where
...
 
     async fn load(&self, k: Self::K, extra: Self::Extra) -> Self::V {
-        self.metric_calls.inc(1);
+        {
+            let mut seen_guard = self.seen.lock();
+
+            if seen_guard.insert(&k).expect("bloom filter cannot fail") {
+                &self.metric_calls_new
+            } else {
+                &self.metric_calls_seen
```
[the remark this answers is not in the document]
crepererum: renamed.

PR #4817, opened 2022-06-09
refactor: Make `DMLWrite::sequence_number` a `SequenceNumber`

alamb:
# Rationale
OG used u64 for sequence numbers, NG uses `i64` (wrapped in `SequenceNumber` new type). There was still code that used the old `u64` style. 

# Changes
1. Make `DMLWrite::sequence_number` a `SequenceNumber` and then make compiler happy


data_types/src/lib.rs
```diff
@@ -1966,12 +1966,12 @@ pub struct Sequence {
     /// The sequencer id (kafka partition id)
     pub sequencer_id: u32,
     /// The sequence number (kafka offset)
-    pub sequence_number: u64,
+    pub sequence_number: SequenceNumber,
```
alamb, line 1969: This is the core change, and everything else follows from that

domodwyer (approved):
What's not to love?

PR #4818, opened 2022-06-09
refactor: Rename `Sequence::sequencer_id` to  `Sequence::kafka_partition` and use proper type

alamb:
Draft until  https://github.com/influxdata/influxdb_iox/pull/4817 is merged

# Rationale
Related to https://github.com/influxdata/influxdb_iox/issues/4237

NG uses the term `kafka_partition` for what OG called  `sequencer_id`. Even more confusing NG introduced a new concept called `Sequencer` / `sequencer_id` that is similar, but *different* than what it was in OG. Some of the code in DML still uses the OG naming which is super confusing.

Eventually, we may want to rename `kafka_partition` (see https://github.com/influxdata/influxdb_iox/issues/4246) but for now at least using consistent terminology is better.

# Changes
1. Rename `Sequence::sequencer_id` to  `Sequence::kafka_partition`
2. Change type to `KafkaPartition`

### DIFF
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -155,7 +155,7 @@ pub struct KafkaPartition(i32);
 
 #[allow(missing_docs)]
 impl KafkaPartition {
-    pub fn new(v: i32) -> Self {
+    pub const fn new(v: i32) -> Self {
         Self(v)
     }
     pub fn get(&self) -> i32 {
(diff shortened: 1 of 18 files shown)


domodwyer (approved):
I ❤️ types.

alamb:
@carols10cents  mentioned that she had tried something similar to this PR earlier, and @crepererum  had some opinions https://github.com/influxdata/influxdb_iox/pull/4664#discussion_r880274862

I will wait for him to return and comment prior to merging

crepererum:
I'm still of the opinion that Kafka shouldn't leak throw the whole code base. There's a write buffer abstraction in place and that should be used. Even having the word "kafka" within the catalog was -- IMHO -- a mistake. It's going to be a substantial waste of work if we replace Kafka with something else (and since there are good managed solutions out there that a fully sufficient for what the write buffer needs, I actually expect us to move away from a self-managed Kafka at a not-too-distant point).

That said, I'm not going to die on that hill :shrug:

alamb:
> I'm still of the opinion that Kafka shouldn't leak throw the whole code base.

I happen to agree -- the issue is that we used `sequencer / sequencer_id` to mean something different #4237  -- maybe @pauldix  has some more thoughts.

Right now we it is worse than kafka through the whole codebase I think as `sequencer_id` is inconsistent 🤦

PR #4819, opened 2022-06-09
refactor: remove unused error variants

alamb:
Minor cleanup I noticed while working on https://github.com/influxdata/influxdb_iox/issues/4567

This code is dead, so remove it.


domodwyer (approved):
It annoys me that the linter can't "see" that these are unused - I suspect it is because of the magic context stuff snafu generates.

PR #4820, opened 2022-06-09
refactor: Use DurationHistogram in more places

carols10cents:
In https://github.com/influxdata/influxdb_iox/pull/4806 I learned about the`DurationHistogram` metric type. I didn't find that originally because I based my code off another spot tracking durations that was using a `U64Histogram`. So this PR changes more places tracking durations to use `DurationHistogram`, which makes the code a lot nicer as appropriate buckets are already converted and recording takes a `Duration`!


router/src/dml_handlers/instrumentation.rs
```diff
@@ -106,11 +81,11 @@ where
             match &res {
                 Ok(_) => {
                     span_recorder.ok("success");
-                    self.write_success.record(delta.as_millis() as _)
+                    self.write_success.record(delta)
```
alamb, line 84: I agree this is much nicer

alamb (approved):
I think it looks lovely and a nice code improvement 🧹 👍

alamb:
When this is merged, I suspect some of the dashboards will need to be updated as the metric names have been updated (to remove the `_ms` suffix):

carols10cents:
> When this is merged, I suspect some of the dashboards will need to be updated as the metric names have been updated (to remove the `_ms` suffix):

Yep... cc @domodwyer can you take care of updating the dashboards?

PR #4821, opened 2022-06-09
WIP feat: split compacted output to many non-overlapped but size limited files

### LINKED ISSUE #4790: Compactor: Split a large file into smaller non-overlapped ones, each smaller than a configured max size
# Rationale
Currently, we compact all overlapped files into one file which can be very large. This ticket is to find those large files and split them into many non-overlapped ones, each smaller than a configured max size

# Initial Solution
Include 2 separated implementations:
1. While compacting several overlapped files, if they are huge, we will split them at that time.
2. A different background task (or add this into a compact background loop) to find large files and split them.

If we discover large files in #1 then #2 almost does nothing and may consider not to be implemented

NGA-TRAN:
Closes #4790 


NGA-TRAN:
Close this PR since #4918 has covered it

PR #4823, opened 2022-06-09
refactor: Move `scan_and_filter` into a `common` module for reuse

alamb:
# Rationale
Related to https://github.com/influxdata/influxdb_iox/issues/4567

See https://github.com/influxdata/influxdb_iox/pull/4822 for the next step. This is a preparatory PR that just moves some code around 

# Changes:
1. Move `scan_and_filter` into a `common`


ingester/src/querier_handler.rs
```diff
@@ -28,9 +26,6 @@ use std::{collections::BTreeMap, sync::Arc};
 #[derive(Debug, Snafu)]
 #[allow(missing_copy_implementations, missing_docs)]
 pub enum Error {
-    #[snafu(display("Failed to select columns: {}", source))]
```
alamb, line 31: these were unused

iox_query/src/frontend/common.rs
```diff
@@ -0,0 +1,130 @@
...
+use observability_deps::tracing::trace;
+use predicate::Predicate;
+use schema::Schema;
+use snafu::{ResultExt, Snafu};
+
+use crate::{
+    exec::IOxSessionContext, provider::ProviderBuilder, util::MissingColumnsToNull, QueryChunk,
+};
+
+#[derive(Debug, Snafu)]
```
alamb, line 13: This code / errors are extracted from influxrpc.rs -- no functional change intended

iox_query/src/frontend/common.rs
```diff
@@ -0,0 +1,130 @@
...
+        .with_execution_context(ctx.child_ctx("provider_builder"));
+
+    // Since the entire predicate is used in the call to
+    // `database.chunks()` there will not be any additional
+    // predicates that get pushed down here
+    //
+    // However, in the future if DataFusion adds extra synthetic
+    // predicates that could be pushed down and used for
+    // additional pruning we may want to add an extra layer of
+    // pruning here.
```
domodwyer, line 84: This sounds like it relied on the specific usage previously - is it still correct?

domodwyer (approved):

iox_query/src/frontend/common.rs
alamb: Not really -- I will update this to be correct

alamb: Fixed in 2594255f3

iox_query/src/frontend/common.rs
```diff
@@ -0,0 +1,128 @@
...
+/// evaluation, returns Ok(None)
+///
+/// The created plan looks like:
+///
+/// ```text
+///   Filter(predicate) [optional]
+///     Scan
+/// ```
+///
+/// NOTE: This function assumes the chunks have already been "pruned"
```
alamb, line 63: updated comments

PR #4832, opened 2022-06-10
refactor: Consolidate `Selection` creation, add tests

alamb:
# Rationale
Related to https://github.com/influxdata/influxdb_iox/issues/4567
Various parts of DataFusion encode selections like `Option<Vec<usize>>` and IOx has a `Selection` enum that encodes selections as `[&str]`.  `usize` only makes sense with respect to a specific schema so there is translation logic back and forth in several places

# Changes
1. Move some translation  logic to `Selection` with some doc comments that hopefully helps with readability
2. Add some tests


querier/src/ingester/mod.rs
```diff
@@ -645,21 +645,13 @@ impl QueryChunk for IngesterPartition {
...
-                    // TODO real error
-                    .expect("error with selection");
-
-                self.batches
-                    .iter()
-                    .map(|batch| batch.project(&projection))
-                    .collect::<std::result::Result<Vec<_>, ArrowError>>()
-                    .expect("error with projection to batches")
-            }
+        let batches = match self.schema.df_projection(selection)? {
```
alamb, line 648: bonus is that we get a real error as well rather than panic on error

ingester/src/querier_handler.rs
```diff
@@ -285,19 +285,7 @@ pub async fn query(
...
-            let schema = data.schema();
-            Some(
-                columns
-                    .iter()
-                    .flat_map(|&column_name| schema.find_index_of(column_name))
-                    .collect(),
-            )
-        }
-    };
+    let indices = data.schema().df_projection_present(selection);
```
alamb, line 288: there is a very subtle difference here that it ignores columns that are not present. I hope the naming makes it clearer what the difference is

NGA-TRAN: I think it will be even clearer if you add a comment saying what this line does as in your comment here 😄

schema/src/lib.rs
```diff
@@ -304,26 +304,50 @@ impl Schema {
...
+        match selection {
+            Selection::All => None,
+            Selection::Some(columns) => {
+                let projection = columns
+                    .iter()
+                    .flat_map(|&column_name| self.find_index_of(column_name))
+                    .collect();
+                Some(projection)
+            }
+        }
```
NGA-TRAN, line 350: I do not know what  the difference between this `df_projection_present` and `df_projection` is beside one returns Result the other does not

NGA-TRAN (approved):
I just wonder why we need two almost the same functions

schema/src/lib.rs
alamb: Yeah, that is the (very subtle) difference -- one errors if a named column is not present and one just ignores it. I will make this clearer in the comments.

alamb: AAAND now I realize I removed this code in https://github.com/influxdata/influxdb_iox/pull/4822 🤦  -- will remove it from this PR too.... Sorry about that @NGA-TRAN

ingester/src/querier_handler.rs
alamb: this code has since been removed, so removed it too

Issue #4621, opened 2022-05-17
Minimize the number of ingesters contacted by the querier to increase robustness to failure


alamb:
__Use case:__
There will be cases where one or more ingesters required for the query's answers can not be contacted by the querier. We are debating the finer details of what to do in this case on https://github.com/influxdata/influxdb_iox/issues/4608.

However, we can reduce the likelihood of needing to deal with an unavailable ingester by making the querier smarter about which ingesters might have *any* data for the query. At the time of writing , the querier asks *all* ingesters that are configured

@pauldix  notes that 

> Based on the current sharding, the querier should only have to hit a single shard per table in the query. In storage gRPC this means that every query should hit a single ingester. 

__Proposal:__

Teach the queriers enough about sharding that they can determine which ingesters might have any data for the query -- this may be part of  #4297 

__Current behaviour:__

Querier asks all ingesters it knows about

__Desired behaviour:__
Querier asks minimal set of ingesters

__Alternatives considered:__
N/A

carols10cents:
I think there are going to be a few parts to this, please correct me if anything here is incorrect, too much, or too little:

- Queriers need to use [the same sharding as the router](https://github.com/influxdata/influxdb_iox/blob/46de8d6cb3b34230bc982f94687b68d5434ddcd2/router/src/sharder/jumphash.rs#L117-L138) to be able to take a table name and namespace and get a sequencer id
- Queriers need to know, in addition to [the address for each ingester](https://github.com/influxdata/influxdb_iox/blob/46de8d6cb3b34230bc982f94687b68d5434ddcd2/clap_blocks/src/querier.rs#L20-L39), which [write buffer partition numbers they're responsible for](https://github.com/influxdata/influxdb_iox/blob/46de8d6cb3b34230bc982f94687b68d5434ddcd2/clap_blocks/src/ingester.rs#L4-L16)
- The [request for partitions](https://github.com/influxdata/influxdb_iox/blob/46de8d6cb3b34230bc982f94687b68d5434ddcd2/querier/src/ingester/mod.rs#L365-L399) needs to take either a sequencer id (and then know the ingester-to-sequencer-range mapping) or an ingester address

alamb:
(note sequencer_id == kafaka partition in some parts of the code, which makes things super confusing)

> Queriers need to use  to be able to take a table name and namespace and get a sequencer id

Yes, I think this is a key part. 

[...]

PR #4835, opened 2022-06-10
feat: Make a sharder available in the querier

carols10cents:
Connects to #4621.

This PR:

- Moves the `JumpHash`/`Sharder` code out of the `router` into its own crate so the same logic can be used in the `querier`
- The querier will ask the catalog for the sequencers to create its sharder
- The querier will now refuse to start up if there aren't any sequencers in the catalog

Let me know if you'd like to see that before merging, but this was getting long, and merging this shouldn't break anything.

If this code looks like it's going in the right direction, then I have a question:

- Should the [`QuerierTable`](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/table/mod.rs#L52-L75) hold a `JumpHash` sharder (in an `Arc`) to use [in `ingester_partitions`](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/table/mod.rs#L181-L184)? Or, because a table/namespace should have one consistent sequencer assigned by the sharder, would it be sufficient for `QuerierTable` to store its assigned `SequencerId` [upon creation](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/namespace/mod.rs#L64-L71)?


alamb (commented):
Thanks @carols10cents  -- I think this PR does what it's description says

I like splitting out the sharder into its own crate 👍 

I am not sure about adding "OptionalWriteBufferConfig" to the querier -- it looks like `maybe_sharder` only needs the number of shards (rather than also the write buffer address, etc)?

I think @domodwyer  and @crepererum  should give this a look.

crepererum:
> * Should the [`QuerierTable`](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/table/mod.rs#L52-L75) hold a `JumpHash` sharder (in an `Arc`) to use [in `ingester_partitions`](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/table/mod.rs#L181-L184)? Or, because a table/namespace should have one consistent sequencer assigned by the sharder, would it be sufficient for `QuerierTable` to store its assigned `SequencerId` [upon creation](https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/querier/src/namespace/mod.rs#L64-L71)?

It's probably the `QuerierDatabase` that stores the sharder config, no? For all the other objects that are generated by it (table -> partitions / chunks whatever) you can assume that they are short-lived, i.e. that they are thrown away after a query completes (modulo caching).

clap_blocks/src/write_buffer.rs
```diff
@@ -32,7 +32,57 @@ pub struct WriteBufferConfig {
...
+        env = "INFLUXDB_IOX_WRITE_BUFFER_AUTO_CREATE_TOPICS"
+    )]
+    pub(crate) auto_create_topics: Option<NonZeroU32>,
+}
+
+/// For use by the querier. If these options are specified, the querier can use the same sharding
+/// that the  router uses to know the subset of ingesters to query. If these options are not
+/// specified, the querier will ask all ingesters.
+#[derive(Debug, clap::Parser)]
+pub struct OptionalWriteBufferConfig {
```
crepererum, line 61: Does this mean that the querier needs to connect to Kafka as well just to discover the shards? I'm not sure if we want that for deployment complexity and scalability. I think the querier could just look at the catalog. The shards should be in there.

carols10cents: Ah, ok. I was thinking it should work just like the router does, but I can make that change!

domodwyer:
Thanks @carols10cents I shall take a look at this shortly :+1:

sharder/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "sharder"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
```
domodwyer, line 6: I think the following can be dev-deps :+1:

* hashbrown
* mutable_batch_lp

sharder/src/lib.rs
```diff
@@ -0,0 +1,27 @@
...
+    unreachable_pub
+)]
+#![warn(
+    missing_docs,
+    clippy::todo,
+    clippy::dbg_macro,
+    clippy::clone_on_ref_ptr,
+    clippy::future_not_send
+)]
+#![allow(clippy::missing_docs_in_private_items)]
```
domodwyer, line 18: THANKS FOR LINTZ 🎉

domodwyer (approved):
:+1:

sharder/Cargo.toml
carols10cents: Ooh good catch!

querier/src/database.rs
```diff
@@ -163,6 +182,29 @@ impl QuerierDatabase {
...
 }
 
+pub async fn create_sharder(catalog: &dyn Catalog) -> Result<JumpHash<Arc<KafkaPartition>>, Error> {
+    let sequencers = catalog
+        .repositories()
+        .await
+        .sequencers()
+        .list()
+        .await
+        .context(CatalogSnafu)?;
```
crepererum, line 192: Should we wrap this into a retry loop? Otherwise our deploys might be prone to long convergence times, because the cold querier fleet (= all pods) caches usually result in higher catalog (= postgres) load and therefore catalog request may easily time out. This might be esp. annoying when we do emergency deploys.

querier/src/database.rs
```diff
@@ -191,12 +233,40 @@ mod tests {
...
             create_ingester_connection_for_testing(),
             QuerierDatabase::MAX_CONCURRENT_QUERIES_MAX.saturating_add(1),
-        );
+        )
+        .await
+        .unwrap();
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "cannot initialise sharder with no shards")]
```
crepererum, line 242: Since `QuerierDatabase::new` can already fail, should this be an ordinary error? If I understand this correctly, this may happen when the routers / ingesters didn't initialize the catalog state yet (i.e. the list of shards is empty) so we would panic for initial cluster deploys. It's not a huge deal, but also not super clean, since (at least IMHO) panics should only occur for unexpected errors while error propagation should occur for expected error cases.

crepererum (approved):
Only two error handling nits, rest looks good. Also thanks for the drive-by clean-ups :+1:

carols10cents:
Resolved the two error handling suggestions, so I'm merging!

PR #4837, opened 2022-06-13
refactor: consolidate plan creation from `QueryChunk`s in `iox_query`

alamb:
# Rationale

@NGA noticed similarities between various methods of `ReorgPlanner` and what the querier was doing https://github.com/influxdata/influxdb_iox/pull/4822#discussion_r894747713

Related to https://github.com/influxdata/influxdb_iox/issues/4567 as this helps  to make a consistent way to build scans across multiple chunks

# Changes
1. Consolidates creation of scan plans to `ScanPlanBuilder`
2. Have ingester use `ScanPlanBuilder` directly rather than a wrapper in `ReorgPlanner`
3. Removes some unnecessary wrapping with `Option`


iox_query/src/frontend/reorg.rs
```diff
@@ -351,45 +230,6 @@ mod test {
         (Arc::new(schema), vec![chunk1, chunk2])
     }
 
-    #[tokio::test]
-    async fn test_sorted_scan_plan() {
```
alamb, line 355: The function this was testing is no longer used (I think it was used in OG)

domodwyer (approved):
This is much nicer :+1:

ingester/src/querier_handler.rs
```diff
@@ -295,12 +294,30 @@ pub(crate) async fn query(
...
+    // Creates an execution plan for a scan and filter data of a single chunk
+    let schema = data.schema();
+    let table_name = data.table_name().to_string();
+
+    debug!(%table_name, ?predicate, "Creating single chunk scan plan");
+
+    let logical_plan = ScanPlanBuilder::new(schema)
+        .with_session_context(ctx.child_ctx("scan_and_filter planning"))
+        .with_predicate(&predicate)
+        .with_chunks([data as _])
```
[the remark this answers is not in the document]
NGA-TRAN: I like the `with_..` calls. Make it easier to follow

iox_query/src/frontend/common.rs
```diff
@@ -40,18 +43,31 @@ pub enum Error {
 
 pub(crate) type Result<T, E = Error> = std::result::Result<T, E>;
 
-pub(crate) struct TableScanAndFilter {
-    /// Represents plan that scans a table and applies optional filtering
+/// Represents scanning one or more [`QueryChunk`]s.
+pub struct ScanPlan {
```
NGA-TRAN, line 47: 👍

iox_query/src/frontend/common.rs
```diff
@@ -64,65 +80,138 @@ pub(crate) struct TableScanAndFilter {
...
+            // so no need to prune them
+            .add_no_op_pruner();
+
+        if let Some(ctx) = ctx {
+            builder = builder.with_execution_context(ctx.child_ctx("provider_builder"));
+        }
+
+        if let Some(sort_key) = sort_key {
+            // Tell the scan of this provider to sort its output on the given sort_key
+            builder = builder.with_sort_key(sort_key);
```
NGA-TRAN, line 163: Later at execution step, if the chunk is sorted (somehow if we know it), will we not add the SortExec?

NGA-TRAN (approved):
Very nice. Thanks @alamb

iox_query/src/frontend/common.rs
alamb: I am not sure. Under the covers it uses all the common logic in the chunk providers which I think go out of their way to avoid sorts.

The idea is that if the sort key is specified, then the output of the scan is guaranteed to be sorted -- I'll make that clearer with some comments

https://github.com/influxdata/influxdb_iox/blob/main/iox_query/src/provider.rs#L414-L427

iox_query/src/frontend/common.rs
```diff
@@ -64,65 +80,138 @@ pub(crate) struct TableScanAndFilter {
...
-    // Use a filter node to add general predicates + timestamp
-    // range, if any
-    if let Some(filter_expr) = predicate.filter_expr() {
-        // Rewrite expression so it only refers to columns in this chunk
-        trace!(table_name, ?filter_expr, "Adding filter expr");
-        let mut rewriter = MissingColumnsToNull::new(&schema);
-        let filter_expr = filter_expr
-            .rewrite(&mut rewriter)
-            .context(RewritingFilterPredicateSnafu { table_name })?;
+    /// Sets the desired output sort key
```
alamb, line 114:
```suggestion
    /// Sets the desired output sort key. If the output of this plan
    /// is not already sorted this way, it will be re-sorted to conform
    /// to this key
```

iox_query/src/frontend/common.rs
```diff
@@ -64,65 +80,138 @@ pub(crate) struct TableScanAndFilter {
...
-
-    for chunk in chunks {
-        // check that it is consistent with this table_name
-        assert_eq!(
-            chunk.table_name(),
-            table_name,
-            "Chunk {} expected table mismatch",
-            chunk.id(),
-        );
+    sort_key: Option<SortKey>,
```
alamb, line 92:
```suggestion
    /// The sort key that describes the desired output sort order
    sort_key: Option<SortKey>,
```

PR #4839, opened 2022-06-13
refactor: Remove some `Option` in `iox_query`

alamb:
Draft until  https://github.com/influxdata/influxdb_iox/pull/4837 is merged

I noticed a possible cleanup while working on https://github.com/influxdata/influxdb_iox/pull/4837

Several functions in the `InfluxRPC` frontend return Option that no (longer) need to (the non plan path has been handled higher up)

Changes:
1. remove `Option` from several signatures


alamb (commented):
the core of the change is to this change:

```rust
    ) -> Result<Option<LogicalPlan>> {
```

to 
```rust
    ) -> Result<LogicalPlan> {
```

crepererum (approved):

PR #4852, opened 2022-06-14
refactor: PartitionKey type

domodwyer:
This PR adds a new `PartitionKey` type so we can accept a typed input in fn calls, instead of a bare `String`, taking advantage of type checking to eliminate errors. It also ref-counts the underlying string to reduce memory usage, but this wasn't the primary motivation.

The actual functional code change is small, but quite a few tests needed updating.

---

* refactor: PartitionKey type (b41ea1d71)

      This commit changes the code base to use a new reference-counted PartitionKey
      type wrapper, instead of passing a bare String around.

      This allows the compiler to type check & verify usage of the partition key,
      instead of passing a bare string around. By reference counting the underlying
      string, we reduce memory usage for some use cases.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -771,7 +771,7 @@ impl Compactor {
                 table_id: iox_metadata.table_id,
                 table_name: Arc::<str>::clone(&iox_metadata.table_name),
                 partition_id: iox_metadata.partition_id,
-                partition_key: Arc::<str>::clone(&iox_metadata.partition_key),
+                partition_key: iox_metadata.partition_key.clone(),
                 min_sequence_number,
                 max_sequence_number,
                 compaction_level: 1, // compacted result file always have level 1
@@ -2590,7 +2590,7 @@ mod tests {
             .unwrap();
         let partition = txn
             .partitions()
-            .create_or_get("one", sequencer.id, table.id)
+            .create_or_get(&"one".into(), sequencer.id, table.id)
             .await
             .unwrap();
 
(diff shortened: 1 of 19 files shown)


data_types/src/lib.rs
```diff
@@ -676,6 +676,58 @@ pub struct Sequencer {
...
+    }
+}
+
+impl From<&str> for PartitionKey {
+    fn from(s: &str) -> Self {
+        Self(s.into())
+    }
+}
+
+impl<DB> sqlx::Type<DB> for PartitionKey
```
domodwyer, line 702: This teaches sqlx to (de)serialise `Arc<str>` as `VARCHAR` - I will upstream it and we'll be able to remove these impls :+1:

data_types/src/lib.rs
```diff
@@ -687,7 +739,7 @@ pub struct Partition {
     /// the table the partition is under
     pub table_id: TableId,
     /// the string key of the partition
-    pub partition_key: String,
+    pub partition_key: PartitionKey,
```
domodwyer, line 742: This is the jist of the PR - where we used `String`, we now use `PartitionKey`.

pauldix (approved):

compactor/src/compact.rs
```diff
@@ -2590,7 +2590,7 @@ mod tests {
             .unwrap();
         let partition = txn
             .partitions()
-            .create_or_get("one", sequencer.id, table.id)
+            .create_or_get(&"one".into(), sequencer.id, table.id)
```
alamb, line 2593: I don't know why but these `&` bother me - proposed PR to remove then: https://github.com/influxdata/influxdb_iox/pull/4862

data_types/src/lib.rs
```diff
@@ -676,6 +676,58 @@ pub struct Sequencer {
     pub min_unpersisted_sequence_number: SequenceNumber,
 }
 
+/// A reference-counted partition key, serialisable to the Postgres VARCHAR data
```
alamb, line 679: This is very nice -- thank you @domodwyer

Issue #4849, opened 2022-06-14
Replace ingester<>querier flight protocol


crepererum:
# What
Replace the current Apache Flight protocol used for ingester<>querier communication w/ a custom protocol that better suits our needs.

# Why
The current Apache Flight protocol was introduced because it was a familiar protocol and simplified the initial implementation. However it has certain drawbacks:

- **schema-less batches:** Flight works by transmitting a schema and then schema-less batches. This requires the ingester to create null-columns. See #4700.
- **metadata-less batches:** The record batches in Flight don't have any metadata attached. The querier needs to know which batch belongs to which partition for efficient de-duplication. So we need to encode all this data in the schema prior to sending the batches which limits our streaming capabilities. See #4567.
- **IOx-flavored flight:** The flight protocol we use here is a special IOx flavor. It's the same that we use for sending queries to the querier (i.e. iox client to querier). However the latter use case is probably going to be replaced with a proper Flight SQL protocol. Flight SQL is definitely NOT what we wanna support for the ingester<>querier path. This leaves us with a single application for our hand-rolled protocol.

# How
Build a simple gRPC protcol that is similar to flight (i.e. we probably wanna use the flight encoding for record batches) but only has a single method and groups each batch with the required metadata (instead of sending the metadata once).

For now, transmit the schema for each record batch as well (this can be optimized later).

alamb:
Closed in https://github.com/influxdata/influxdb_iox/pull/4875

PR #4853, opened 2022-06-14
feat: extend flight client to accept multiple (changing) schemas

crepererum:
See #4849.

Originally I intended not to use Flight at all for the new
ingester<>querier protocol. However since flight also deals with
dictionary batches and multiple batches and the gRPC protocol that I
would write would look very similar, I will use Flight with a bit more
flexible message types.

The rough idea for the protocol is the following stream:

- for each partition:
  1. "none" message with partition metadata
  2. for each chunk (can have different schemas under certain
     circumstances):
     1. "schema" message (resets dictionary state)
     2. (optional) dictionary batch messages
     3. one or more "record batch" message

The nice thing about it is that the same arrow client works also for the
existing client<>querier protocol since there we just send:

1. "schema" message (no app metadata)
2. (optional) dictionary batch messages
3. zero, one or more "record batch" message (no app metadata)


querier/src/ingester/flight_client.rs
```diff
@@ -142,18 +158,25 @@ where
     }
 }
 
+#[derive(Debug)]
+struct PerformQueryAdapter {
+    inner: PerformQuery<IngesterQueryResponseMetadata>,
+    app_metadata: IngesterQueryResponseMetadata,
+    schema: Arc<Schema>,
+}
```
crepererum, line 166: This adapter will be gone when I'm done (i.e. I'll change the `QueryData` interface to the new protocol approach), but I need to make the changes step by step, otherwise this would result in a massive, messy PR.

influxdb_iox_client/src/client/flight.rs
```diff
@@ -182,6 +186,53 @@ where
     }
 }
 
+#[derive(Debug)]
+struct PerformQueryState {
+    schema: Arc<Schema>,
+    dictionaries_by_field: HashMap<i64, ArrayRef>,
+}
+
+/// Low-level message returned by the flight server.
```
alamb, line 195: This is a very clever approach -- nice work @crepererum 

I think documenting this "low level" message approach somewhere is important -- given the message types are defined in the client/flight.rs perhaps we could define it here (via a diagram showing what messages were wrapped and how this connects to flight)
?

influxdb_iox_client/src/client/flight.rs
```diff
@@ -182,6 +186,53 @@ where
...
+pub enum LowLevelMessage {
+    /// None.
+    None,
+
+    /// Schema.
+    Schema(Arc<Schema>),
+
+    /// Record batch.
+    RecordBatch(RecordBatch),
+}
```
alamb, line 206: Eventually for the ingester -> querier, there may be a need for custom metadata message as well (though perhaps we can hijack the `RecordBatch` metadata for that

alamb (approved):
I think this change will have implications for existing clients (aka the golang client library which is used to integrate IOx with flux and fluxtest) -- so we may want to figure out what changes are needed there before merging this one. 

While writing this, I think it is the time to split the protocol that clients use to run queries on the querier with the protocol used by the querier to talk to the ingester (they can still be flight based and share much of the same code) so that we can change the querier <--> ingester communications without other clients. 

Approving so you can make progress, but the backwards compatibility with other clients should be considered I think

crepererum:
> I think this change will have implications for existing clients (aka the golang client library which is used to integrate IOx with flux and fluxtest) -- so we may want to figure out what changes are needed there before merging this one.

The client->querier interface won't change. It's just the the IOx / Rust client is now able to consume the more advanced flight patterns that I wanna use for ingester<>querier.

crepererum:
@alamb I've re-organized the high- and low-level client and added docs. PTAL.

influxdb_iox/src/commands/query_ingester.rs
```diff
@@ -54,7 +58,7 @@ pub struct Config {
 }
 
 pub async fn command(connection: Connection, config: Config) -> Result<()> {
-    let mut client = flight::Client::new(connection);
+    let mut client = flight::low_level::Client::new(connection);
```
alamb, line 61: I like this

influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+            match self.inner.next().await? {
+                None => return Ok(None),
+                Some((LowLevelMessage::Schema(_), _)) => {
+                    if self.got_schema {
+                        return Err(Error::UnexpectedSchemaChange);
+                    }
+                    self.got_schema = true;
+                }
+                Some((LowLevelMessage::RecordBatch(batch), _)) => return Ok(Some(batch)),
+                Some(_) => (),
```
alamb, line 171: What is this case for? It might be nice to complete the enumeration (`Some(LowLevelMessage::None)`) so if new message types are added the compiler will draw attention to this `match`

alamb (approved):
I think it looks very nice -- thank you @crepererum  

It is especially nice that the code structure now mirrors the protocol structure (the lower level messages are handled by a lower level client) 👍

influxdb_iox_client/src/client/flight/mod.rs
crepererum: done

PR #4854, opened 2022-06-14
feat: extend duration histograms down to 1ms

crepererum:
5ms are quite long considering that many requests take way below 100ms in
total. Let's add two more levels on the lower end of the spectrum.

Since we only do not use data-dependent histograms (i.e. do not include
table or namespace names), the overhead should be acceptable.


alamb (commented):
Thanks @crepererum

PR #4855, opened 2022-06-14
refactor: remove one buffer copy in the ingester

alamb:
Part of https://github.com/influxdata/influxdb_iox/issues/4567

# Rationale
This saves at least one copy / combination of `RecordBatch`s while the ingester is processing data for the querier. 

# changes
1. Run only a single query plan for persisting and snapshot batches


ingester/src/querier_handler.rs
```diff
@@ -202,55 +202,24 @@ async fn prepare_data_to_querier_for_partition(
...
+    // No data!
+    if queryable_batch.data.is_empty() {
+        return Ok(None);
     }
 
-    // ------------------------------------------------
-    // Combine record batches into one batch and pad null values as needed
-
-    // Schema of all record batches after merging
-    let schema = merge_record_batch_schemas(&filter_applied_batches);
```
alamb, line 249: This `merge_record_batches` call is removed -- I am not sure how much difference this makes to peak memory load (the input batches are dropped right after), but it is more efficient and simplifies the code

NGA-TRAN (approved):
I have to rely on available/new tests to verify if this works because the buffer is no longer used

alamb:
> I have to rely on available/new tests to verify if this works because the buffer is no longer used

yes I am relying on the tests (both the extensive tests in the querier as well as in `query_tests`)

PR #4862, opened 2022-06-15
refactor: remove some `&` use in iox_catalog

alamb:
Minor follow on to https://github.com/influxdata/influxdb_iox/pull/4852

For some reason, the pattern `&"foo".into()` was really bothering me (getting old, most likely).

This PR removes the need for `&`

I freely admit there is no rationale for this change other than some inherent idea of "code beauty" that others might not share


data_types/src/lib.rs
```diff
@@ -676,8 +676,11 @@ pub struct Sequencer {
     pub min_unpersisted_sequence_number: SequenceNumber,
 }
 
-/// A reference-counted partition key, serialisable to the Postgres VARCHAR data
-/// type.
+/// Defines an partition as an arbitrary string within a table within
+/// a namespace.
```
alamb, line 680:
```suggestion
/// Defines an partition via an arbitrary string within a table within
/// a namespace.
```

iox_catalog/src/mem.rs
```diff
@@ -671,22 +671,22 @@ impl SequencerRepo for MemTxn {
...
+            p.partition_key == key && p.sequencer_id == sequencer_id && p.table_id == table_id
         }) {
             Some(p) => p,
             None => {
                 let p = Partition {
                     id: PartitionId::new(stage.partitions.len() as i64 + 1),
                     sequencer_id,
                     table_id,
-                    partition_key: key.clone(),
+                    partition_key: key,
```
alamb, line 689: this needs an owned copy of the key anyways in some paths

alamb:
For posterity, @domodwyer  has pointed out that this is a new level of stylistic bike-shedding -- and I agree. Not sure why I chose to spend the 5 minutes on this... But I did...

PR #4863, opened 2022-06-15
refactor: prepare new ingester<>querier protocol on the querier side

crepererum:
This changes the querier internals to work with the new protocol. The
wire protocol stays the same (for now). There's a (somewhat hackish)
adapter in place on the querier side that converts the old to the new
protocol on-the-fly. This is an intermediate step before we actually
change the wire protocol (and in a step after that also take advantage
of the new possibilites on the ingester side).

Ref #4849.


ingester/src/server/grpc.rs
```diff
@@ -398,6 +398,7 @@ impl GetStream {
                 })
                 .collect(),
             batch_partition_ids: batch_partition_ids.into_iter().map(|id| id.get()).collect(),
+            ..Default::default()
```
crepererum, line 401: Getting rid of this "dual" state (i.e. the new fields are used by the querier internally but the old fields are actually used by the wire protocol) will be the job of the next PR.

querier/src/ingester/flight_client.rs
```diff
@@ -162,31 +207,139 @@ struct PerformQueryAdapter {
...
+        };
+        (LowLevelMessage::None, meta)
     }
+}
 
-    fn schema(&self) -> Arc<Schema> {
-        Arc::clone(&self.schema)
+#[async_trait]
+impl QueryData for PerformQueryAdapter {
+    async fn next(
```
crepererum, line 252: This whole adapter is obviously a bit of a mess and only tested by the integration test. It will be gone after the next PR, but I wanted to keep the PR size minimal because every change (querier internals, wire protocol, ingester internals) also requires touching a bunch of tests.

querier/src/ingester/mod.rs
```diff
@@ -271,50 +265,54 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
     debug!(%ingester_address, num_batches, "Received batches from ingester");
-    trace!(?partitions, schema=?perform_query.schema(), "Detailed from ingester");
-    ensure!(
-        num_batches == partition_ids.len(),
-        TooManyPartitionIdsSnafu {
-            actual: partition_ids.len(),
-            expected: num_batches,
-            ingester_address: ingester_address.as_ref(),
-        },
-    );
```
crepererum, line 317: This is the new core logic to dissect the ingester response on the querier side (or rather at the moment: the ingester response that went through the adapter).

querier/src/ingester/mod.rs
```diff
@@ -954,142 +936,189 @@ mod tests {
...
             .await,
         );
         let ingester_conn = mock_flight_client.ingester_conn().await;
         let err = get_partitions(&ingester_conn).await.unwrap_err();
-        assert_matches!(err, Error::UnknownPartitionId { .. });
+        assert_matches!(err, Error::BatchWithoutPartition { .. });
     }
 
     #[tokio::test]
     async fn test_flight_many_batches() {
```
crepererum, line 1032: This test is a nice example of new wire protocol with every message and metadata.

generated_types/protos/influxdata/iox/ingester/v1/query.proto
```diff
@@ -65,6 +64,18 @@ message IngesterQueryResponseMetadata {
 
   // Map each record batch to a partition ID.
   repeated int64 batch_partition_ids = 6;
+
+  // Partition id for this batch.
+  //
+  // This field is currently NOT used by the ingester but will be soon.
+  int64 partition_id = 7;
```
alamb, line 71: I don't fully understand why per-batch metadata (`partition_id`) is on a seemingly per-request structure (`IngesterQueryResponseMetadata`) but I suspect it will soon be made clear

querier/src/ingester/flight_client.rs
```diff
@@ -162,31 +207,139 @@ struct PerformQueryAdapter {
...
                 }
                 // ignore all other message types for now
                 Some((LowLevelMessage::None | LowLevelMessage::Schema(_), _)) => (),
             }
         }
     }
 
-    fn app_metadata(&self) -> &proto::IngesterQueryResponseMetadata {
-        &self.app_metadata
+    fn announce_partition(
```
alamb, line 232: A docstring here would help I think

querier/src/ingester/flight_client.rs
```diff
@@ -162,31 +207,139 @@ struct PerformQueryAdapter {
...
+        &mut self,
+        partition_id: PartitionId,
+    ) -> (LowLevelMessage, proto::IngesterQueryResponseMetadata) {
+        let meta = proto::IngesterQueryResponseMetadata {
```
alamb, line 236: I wonder if making `IngesterQueryResponseMetadata` an enum would make it clearer that it contains per-connection data as well as is used to transmit per batch partition ids

querier/src/ingester/flight_client.rs
```diff
@@ -162,31 +207,139 @@ struct PerformQueryAdapter {
...
     schema: Arc<Schema>,
+    batch_counter: usize,
+    state: PerformQueryAdapterState,
 }
 
-#[async_trait]
-impl QueryData for PerformQueryAdapter {
-    async fn next(&mut self) -> Result<Option<RecordBatch>, FlightError> {
+impl PerformQueryAdapter {
+    async fn next_batch(&mut self) -> Result<Option<(RecordBatch, usize)>, FlightError> {
```
alamb, line 215: I recommend documenting, in docstring, what the returned values are (specifically that `size` is the batch counter)

querier/src/ingester/flight_client.rs
```diff
@@ -128,32 +126,79 @@ impl FlightClient for FlightClientImpl {
...
+        &mut self,
+    ) -> Result<Option<(LowLevelMessage, proto::IngesterQueryResponseMetadata)>, FlightError> {
         self.deref_mut().next().await
     }
+}
 
-    fn app_metadata(&self) -> &proto::IngesterQueryResponseMetadata {
-        self.deref().app_metadata()
-    }
+/// Protocol state.
```
alamb, line 148: 👍  makes sense -- it might also help to document somewhere that the reason for this state is to allow sending per-partition metadata as well as record batches with (potentially) different schemas

Though given your description that this adapter is destined to be deleted, maybe no changes are needed

querier/src/ingester/mod.rs
```diff
@@ -330,6 +328,7 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
             Arc::clone(&table_name),
             partition_id,
             sequencer_id,
+            // TODO(marco): project schema to the columns that are present within this partition
```
alamb, line 331: https://github.com/influxdata/influxdb_iox/blob/eca3b6b9a10c7674c7c9405da029b2b3dfe56542/arrow_util/src/util.rs#L26-L57 may help

querier/src/ingester/mod.rs
```diff
@@ -954,142 +936,189 @@ mod tests {
...
                 "addr1",
                 Ok(MockQueryData {
-                    results: vec![Ok(record_batch)],
-                    app_metadata: IngesterQueryResponseMetadata {
-                        unpersisted_partitions: BTreeMap::from([(
-                            1,
-                            PartitionStatus {
-                                parquet_max_sequence_number: Some(1),
-                                tombstone_max_sequence_number: Some(2),
+                    results: vec![
```
alamb, line 967: These tests are great (in that they document in code how the querier is supposed to be reacting to low level messages from the ingester, as well as what the querier expects from the ingester)

querier/src/ingester/mod.rs
```diff
@@ -954,142 +936,189 @@ mod tests {
...
+                            )),
+                            Ok((
+                                LowLevelMessage::Schema(Arc::clone(&schema_1_2)),
+                                IngesterQueryResponseMetadata::default(),
+                            )),
+                            Ok((
+                                LowLevelMessage::RecordBatch(record_batch_1_2),
+                                IngesterQueryResponseMetadata::default(),
+                            )),
+                            Ok((
```
alamb, line 1076: very cool

alamb (approved):
Looks great to me ❤️

querier/src/ingester/mod.rs
crepererum: The TODO is meant for a follow-up. There is no regression here, it's more a note that I should fix/improve that.

querier/src/ingester/flight_client.rs
crepererum: done

Issue #3603, opened 2022-02-01
Update router to send mutable batches belonging to a single partition


pauldix:
The setting for how data is partitioned in a namespace will be a property of the namespace. The router should automatically partition the data as it comes in (along with sharding it) based on the partition settings in the namespace. It should add the partition key (a string) to the mutable batch so that it can be read by the ingester.

For now, partitioning can just be hard coded by day. We'll need to come up with some expression to specify how to partition data later.

The ingester will be responsible for creating any partition records in the catalog.

The code in `data.rs` in #3533 will need to be updated so that it doesn't generate the partition key itself.

domodwyer:
Update: 

* Router partitions the data using a hard coded YMD date string
* Key is not propagated across the API boundary to the ingester - currently the ingester is inferring it from the data it gets

PR #4864, opened 2022-06-15
refactor: add partition key to DmlWrite

domodwyer:
This change allows us to propagate the derived partition key through the system, but stops short of passing it over the kafka boundary - I'll follow up with another PR to add this shortly (tracked in #3603).

Prerequisite for fixing https://github.com/influxdata/influxdb_iox/issues/4787.

---

* refactor: emit PartitionKey from partitioner (61182f506)

      Changes the partitioning code to emit a PartitionKey, instead of a bare 
      String.

* refactor: always generate a partition key (0da8ec87d)

      Changes the partitioner to always generate a partition key, even if the column
      being used to partition doesn't exist. This doesn't functionally change the
      batch partitioning output, but ensures we always have a non-empty string for
      the partition key.

* refactor: store PartitionKey in DmlWrite (4df296456)

      Carry the PartitionKey in the DmlWrite, allowing the batch to be associated
      with a specific partition key.

### DIFF
--- a/write_buffer/src/codec.rs
+++ b/write_buffer/src/codec.rs
@@ -196,6 +196,8 @@ pub fn decode(
                     Ok(DmlOperation::Write(DmlWrite::new(
                         headers.namespace,
                         tables,
+                        // TODO(3603): propagate partition key through kafka
+                        None,
                         meta,
                     )))
                 }
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -692,12 +692,14 @@ impl Display for PartitionKey {
 
 impl From<String> for PartitionKey {
     fn from(s: String) -> Self {
+        assert!(!s.is_empty());
         Self(s.into())
     }
 }
 
 impl From<&str> for PartitionKey {
     fn from(s: &str) -> Self {
+        assert!(!s.is_empty());
         Self(s.into())
     }
 }
--- a/dml/src/lib.rs
+++ b/dml/src/lib.rs
@@ -11,7 +11,7 @@
     clippy::clone_on_ref_ptr
 )]
 
-use data_types::{DeletePredicate, NonEmptyString, Sequence, StatValues, Statistics};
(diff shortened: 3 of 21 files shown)


write_buffer/src/codec.rs
```diff
@@ -196,6 +196,8 @@ pub fn decode(
                     Ok(DmlOperation::Write(DmlWrite::new(
                         headers.namespace,
                         tables,
+                        // TODO(3603): propagate partition key through kafka
+                        None,
```
carols10cents, line 200: It seems like we actually want the partition key to be required rather than an `Option`... is that coming with the next PR? Or am I misunderstanding something? When would the partition key legitimately be missing in production?

carols10cents (approved):
One question but I'm probably misunderstanding something, otherwise looks fine!

write_buffer/src/codec.rs
domodwyer: Ideally we would make it to required!

Unfortunately we currently construct a `DmlWrite` in the `influxdb_iox_client` [here](https://github.com/influxdata/influxdb_iox/blob/4df2964566f770f6914c7d252231f4ec24ebacc2/influxdb_iox_client/src/client/write.rs#L65) and serialise it to the proto representation to send over grpc. We use the same proto representation of a `DmlWrite` for the public grpc write interface, and the internal kafka message.

Since the end user doesn't partition their writes they don't specify the partition key, so it has to be optional for now :( It's annoying, because the grpc write handler doesn't do anything with the provided partition key either! The writes still pass through the partitioner in the router downstream of the write handler. 

I'm not sure the public grpc write API should use the same types/representation as our internal kafka message - it leads to silly coupling problems like this.

aierui:
In this place, partitioner [TemplatePart](https://github.com/influxdata/influxdb_iox/blob/7c60edd38c/ioxd_router/src/lib.rs#L195) maybe need to custom configuration. @domodwyer

PR #4870, opened 2022-06-15
fix(ingester): make ingester metrics start with `ingester`

alamb:
# Rationale:
Most the other ingester metrics start with `ingester` rather than `ingest` so I almost missed some ingester metrics while working on https://github.com/influxdata/conductor/issues/970:

Screen Shot 2022-06-15 at 12 12 31 PM

# Changes
1. Change metrics in ingester from `ingest_` to `ingester_`


ingester/src/stream_handler/handler.rs
```diff
@@ -94,14 +94,14 @@ impl<I, O> SequencedStreamHandler<I, O> {
         // Lifecycle-driven ingest pause duration
         let pause_duration = metrics
             .register_metric::<DurationCounter>(
-                "ingest_paused_duration_total",
+                "ingeste_paused_duration_total",
```
alamb, line 97:
```suggestion
                "ingester_paused_duration_total",
```

crepererum (commented):
maybe use `iox_ingester` to make it even clearer that this is an IOx specific metric?

domodwyer:
I think `ingester` is sufficient - no other component prefixes with `iox` 🤷

PR #4871, opened 2022-06-15
fix(ingester): ensure all ingester metrics are prefixed with `ingester_`

alamb:
Follow on to https://github.com/influxdata/influxdb_iox/pull/4870

# Rationale
As part of https://github.com/influxdata/conductor/issues/970 I was looking in grafana for metrics that showed how far behind the most recent message in kafka was. I couldn't find such metrics (see #4870 showing how I searched for "ingester" in the grafana metrics browser). After some digging, I found https://github.com/influxdata/influxdb_iox/pull/4235 which had already added it 👍 

# Changes 
Add `ingester_` prefix to all metric names to ensure that they are more easily discoverable in grafana


alamb (commented):
I will update the ingester dashboard after these changes

domodwyer (approved):
These were named to keep parity with OG - this is much nicer 🎉

domodwyer:
It's worth noting there have been problems with these metrics reporting impossible lag values, etc. Maybe this is no longer a problem as I haven't looked in a long time 👍

alamb:
> 
It's worth noting there have been problems with these metrics reporting impossible lag values, etc. Maybe this is no longer a problem as I haven't looked in a long time 👍


The `kafka_lag` looked reasonable to me during investigation into https://github.com/influxdata/conductor/issues/970 -- 🤷

PR #4874, opened 2022-06-16
fix(ingester): Distinguish between "not found" and other flight errors

alamb:
# Rationale:
I want to understand the health of the ingesters when providing data to the queriers. 

At the moment the ingesters may respond with "unknown table" when asked by the queriers and the [queriers ignore this](https://github.com/influxdata/influxdb_iox/blob/main/querier/src/ingester/mod.rs#L231-L244). However the unknown error still appears in the ingester dashboard as an error:

Screen Shot 2022-06-16 at 7 02 16 AM

Thus, I want to be able to distinguish between "not found" errors and other errors on the ingesters.

# Changes
1. Add "failure-unknown" as another result type metrics so I can distinguish this (in addition to "success" and "failure")

I will follow on with a change to the ingester dashboard for the changed metric names


crepererum (approved):

domodwyer (commented):
:+1:

PR #4875, opened 2022-06-16
refactor: stream query results from ingester to querier

crepererum:
Do NOT collect record batches on the ingester side before sending them over to the querier.


domodwyer (approved):
This looks sensible to me :+1:

ingester/src/querier_handler.rs
```diff
@@ -164,16 +133,42 @@ pub async fn prepare_data_to_querier(
...
-    Ok(IngesterQueryResponse::new(Box::pin(futures::stream::iter(
-        unpersisted_partitions,
-    ))))
+    let ingest_data = Arc::clone(ingest_data);
+    let request = Arc::clone(request);
+    let partitions = futures::stream::iter(unpersisted_partitions).then(move |partition| {
+        let ingest_data = Arc::clone(&ingest_data);
+        let request = Arc::clone(&request);
+
+        async move {
```
alamb, line 142: Does this processes partitions one at a time rather than concurrently?  I suspect that is just fine (and it is what the old code did too), but I wanted to ask as it is unclear this seems to return a stream of `IngesterQueryPartition` which each have a stream inside

ingester/src/querier_handler.rs
```diff
@@ -197,38 +192,14 @@ async fn prepare_data_to_querier_for_partition(
...
         Arc::new(queryable_batch),
         predicate.clone(),
         selection,
     )
     .await
-}
-
-/// Query a given Queryable Batch, applying selection and filters as appropriate
-/// Return one record batch
-async fn run_query(
```
alamb, line 211: 🔥

alamb (commented):
I think this also closes https://github.com/influxdata/influxdb_iox/issues/4849 and https://github.com/influxdata/influxdb_iox/issues/4567

Very nice 👌  -- thank you for breaking this up into multiple PRs

alamb (approved):

ingester/src/querier_handler.rs
crepererum: Yes, it's one partition at the time. We could change the internals to something like "a set/vector of futures" instead of a stream for the partitions, or poll the partition stream more aggressively.

However for both solution the question would be how many partitions we process in parallel.

And the wire format requires us to serialize one partition after the other. We could easily extend the format to remove this restriction (e.g. by sending `app_metadata` that contains the partition ID also for the schema and record batch messages). Since we have a somewhat solid test coverage now, I think the complexity would be acceptable.

PR #4878, opened 2022-06-16
fix: respect partition key when batching dml ops

### LINKED ISSUE #4787: Partitions contain writes that don't map to the partition's partition key

domodwyer:
This PR fixes #4787.

The router is hard-coded to [partition writes by YMD][scheme] (meaning a partition per day, today's being `2022-06-16`), breaking the writes into a single [DML write request per partition][router-partition]. The router then dispatches each of these partitioned DML writes into kafka via the write buffer abstraction.

On the other side of kafka, the ingester reads these messages and uses the same hard-coded partitioning scheme to [infer the partition key][ingest] of the batch. This inference solution was a temporary workaround until OG was deprecated and removed, which made changing the kafka message format feasible (#3603). 

Correctness of this scheme relies on the atomicity of the DML batches being preserved through the system - a batch is split up by the router according to the partition rules, and then sent to the ingester which uses the same logic to infer it.

Unfortunately the kafka write buffer client (and _only the kafka impl_...) [merges together DML operations][agg] before writing the resulting merged batch to kafka. This undid the partitioning the router performed by glueing together independent, per partition DML ops without respecting the partitioning, causing one DML op to contain many distinct partitions. This would lead to the ingester incorrectly inferring the partition key for the batch which would contain data for multiple desired partitions. This caused both the per-partition catalog metadata, and actual resulting parquet files to incorrectly contain data for multiple partitions.

This led to at least 902,241 partitions which broke the "writes for one day" invariant after being merged together with writes for different dates.

## Further work

* Affected partitions need to be fixed, or removed. This includes both the catalog references, and the parquet files.
* Pass the derived partition key over the API boundary (from router -> ingester) removing the need for inference (follow-up PR incoming!)
* Reconsider the DML aggregation strategy:
  * Aggregation should occur "above" the kafka client - aggregating should be independent of the client used, both for consistency across all write buffer impls, and separation of concerns. This made debugging much harder as it didn't happen when using the all-in-one mode, nor did any unit tests trigger the issue as there is no aggregation for the mock impl used in tests.
  * Consider preserving the individual DML ops, but batching them up (effectively `Vec<DmlOp>` rather than merging them into a single `DmlOp`). The potential loss of compression is likely to be negligible, but both the impl & mental model would be much simpler.

[router-partition]:https://github.com/influxdata/influxdb_iox/blob/43b3f224116f5a886ec7c77e70ed479f21721a17/router/src/dml_handlers/partitioner.rs#L76-L98
[scheme]:https://github.com/influxdata/influxdb_iox/blob/43b3f224116f5a886ec7c77e70ed479f21721a17/ioxd_router/src/lib.rs#L198
[ingest]:https://github.com/influxdata/influxdb_iox/blob/43b3f224116f5a886ec7c77e70ed479f21721a17/ingester/src/partioning.rs#L25-L39
[agg]:https://github.com/influxdata/influxdb_iox/blob/43b3f224116f5a886ec7c77e70ed479f21721a17/write_buffer/src/kafka/aggregator.rs#L128-L147
---

* fix: respect partition key when batching dml ops (43b3f2241)

      This commit changes the kafka write aggregator to only merge DML ops destined
      for the same partition.

      Prior to this commit the aggregator was merging DML ops that had different
      partition keys, causing data to be persisted in incorrect partitions:

          https://github.com/influxdata/influxdb_iox/issues/4787

### DIFF
--- a/write_buffer/src/kafka/aggregator.rs
+++ b/write_buffer/src/kafka/aggregator.rs
@@ -1,5 +1,5 @@
 use crate::codec::{ContentType, IoxHeaders};
-use data_types::{Sequence, SequenceNumber};
+use data_types::{PartitionKey, Sequence, SequenceNumber};
 use dml::{DmlMeta, DmlOperation, DmlWrite};
 use hashbrown::{hash_map::Entry, HashMap};
 use iox_time::{Time, TimeProvider};
@@ -34,6 +34,9 @@ struct WriteAggregator {
     /// Data for every table.
     tables: HashMap<String, MutableBatch>,
 
+    /// The partition key derived for these batches.
+    partition_key: PartitionKey,
+
     /// Span recorder to link spans from incoming writes to aggregated write.
     span_recorder: Option<SpanRecorder>,
 
@@ -49,9 +52,15 @@ impl WriteAggregator {
         let mut span_recorder = None;
         Self::record_span(&mut span_recorder, write.meta().span_context(), &collector);
 
+        let partition_key = write
+            .partition_key()
+            .cloned()
+            .expect("enqueuing unpartitioned write into kafka");
+
         Self {
             namespace: write.namespace().to_owned(),
             tables: write.into_tables().collect(),
+            partition_key,
             span_recorder,
             tag,
             collector,
@@ -90,6 +99,11 @@ impl WriteAggregator {
     /// Check if we can push the given write to this aggregator (mostly if the schemas match).
     fn can_push(&self, write: &DmlWrite) -> bool {
         assert_eq!(write.namespace(), self.namespace);
+
+        // Only batch together writes for the same partition.
+        if write.partition_key() != Some(&self.partition_key) {
+            return false;
+        }
 
         for (table, batch) in write.tables() {
             if let Some(existing) = self.tables.get(table) {
@@ -151,7 +165,12 @@ impl WriteAggregator {
         };
 
         let meta = DmlMeta::unsequenced(ctx);
-        DmlWrite::new(self.namespace.clone(), self.tables.clone(), None, meta)
+        DmlWrite::new(
+            self.namespace.clone(),
+            self.tables.clone(),
+            Some(self.partition_key.clone()),
+            meta,
+        )
     }
 }
 
--- a/write_buffer/src/core.rs
+++ b/write_buffer/src/core.rs
@@ -372,14 +372,14 @@ pub mod test_utils {
         writer: &impl WriteBufferWriting,
         lp: &str,
         sequencer_id: u32,
-        partition_key: Option<PartitionKey>,
+        partition_key: PartitionKey,
         span_context: Option<&SpanContext>,
     ) -> DmlWrite {
         let tables = mutable_batch_lp::lines_to_batches(lp, 0).unwrap();
         let write = DmlWrite::new(
             namespace,
             tables,
-            partition_key,
+            Some(partition_key),
             DmlMeta::unsequenced(span_context.cloned()),
         );
         let operation = DmlOperation::Write(write);
@@ -431,7 +431,7 @@ pub mod test_utils {
             &writer,
             entry_1,
             sequencer_id,
-            Some(PartitionKey::from("bananas")),
+            PartitionKey::from("bananas"),
             None,
         )
         .await;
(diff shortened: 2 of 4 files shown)


crepererum:
> Aggregation should occur "above" the kafka client - aggregating should be independent of the client used, both for consistency across all write buffer impls, and separation of concerns. This made debugging much harder as it didn't happen when using the all-in-one mode, nor did any unit tests trigger the issue as there is no aggregation for the mock impl used in tests.

The issue is that Kafka (or any sequencer) is NOT a data stream but a message stream and the message size depends on the concrete write buffer type (e.g. Kafka) as well as on the infrastructure. So performing this aggregation on a higher level is IMHO as messy as the current solution.


> Consider preserving the individual DML ops, but batching them up (effectively Vec<DmlOp> rather than merging them into a single DmlOp). The potential loss of compression is likely to be negligible, but both the impl & mental model would be much simpler.

At least for the current serialization format this is not true, because now you need to store column names and string dictionaries for every single tiny write operation.

crepererum (approved):
Good catch!

write_buffer/src/kafka/aggregator.rs
```diff
@@ -91,6 +100,11 @@ impl WriteAggregator {
     fn can_push(&self, write: &DmlWrite) -> bool {
         assert_eq!(write.namespace(), self.namespace);
 
+        // Only batch together writes for the same partition.
+        if write.partition_key() != Some(&self.partition_key) {
```
alamb, line 104: this is the actual fix, right?

alamb (approved):

write_buffer/src/kafka/aggregator.rs
domodwyer: Yes sorry, I was going to comment around the place but got meeting sniped.

tustvold: If I understand this correctly, any write for a different partition will effectively poison the aggregation logic for all batches. This seems not ideal...

tustvold:
> Consider preserving the individual DML ops, but batching them up

FWIW rskafka has provides an Aggregator that does this, the implementation in IOx just exploits the fact we can do better and aggregate the actual messages themselves - https://docs.rs/rskafka/latest/rskafka/client/producer/aggregator/struct.RecordAggregator.html.

write_buffer/src/kafka/aggregator.rs
alamb: Context is https://github.com/influxdata/conductor/issues/971 where we are trying to get more throughput from kafka and the current theory is that small kafka batch sizes are a contributing factor

crepererum: Should be fixed by #5113

PR #4879, opened 2022-06-16
refactor: make querier->ingester request metrics per-ingester

crepererum:
The metrics and logs introduced in #4806 will be emitted once for all
ingesters instead of per request. The accumulated view makes it pretty
hard to judge the actual request-response timings and the number of
requests.

Instead we now measure the data per request.


alamb (approved):
LTGM

querier/src/ingester/mod.rs
```diff
@@ -151,13 +153,104 @@ pub trait IngesterConnection: std::fmt::Debug + Send + Sync + 'static {
...
+    ingester_duration_cancelled: DurationHistogram,
+}
+
+impl IngesterConnectionMetrics {
+    fn new(metric_registry: &metric::Registry) -> Self {
+        let ingester_duration: Metric<DurationHistogram> = metric_registry.register_metric(
+            "ingester_duration",
+            "ingester request query execution duration",
+        );
+        let ingester_duration_success = ingester_duration.recorder(&[("result", "success")]);
```
alamb, line 175: Something that might also be interesting is to add in the ingester address to which the request is made (so we can get per-ingester breakdown). As a follow on PR perhaps

crepererum: Yeah, I thought about that as well. Will do that in a follow-up :+1:

Issue #4761, opened 2022-06-01
Panic: Cannot convert timestamp nano to u128 in iox-shared-compactor-3-0 


NGA-TRAN:
This has been only seen in `iox-shared-compactor-3-0`

```
level=error msg="Thread panic" panic_info="panicked at 'Cannot convert timestamp nano to u128 : TryFromIntError(())', compactor/src/query.rs:162:44" target="panic_logging" location="panic_logging/src/lib.rs:107" time=1654095962454205924
   0: rust_begin_unwind
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::expect
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1035:23
   4: <compactor::query::QueryableParquetChunk as iox_query::QueryChunk>::id
             at ./influxdb_iox/compactor/src/query.rs:162:13
   5: iox_query::provider::Deduplicater::build_deduplicate_plan_for_overlapped_chunks::{{closure}}
             at ./influxdb_iox/iox_query/src/provider.rs:762:57
   6: core::slice::<impl [T]>::sort_unstable_by_key::{{closure}}
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/mod.rs:2606:38
   7: core::slice::sort::shift_tail
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/sort.rs:99:24
   8: core::slice::sort::insertion_sort
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/sort.rs:177:9
   9: core::slice::sort::recurse
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/sort.rs:739:13
  10: core::slice::sort::quicksort
             at ./rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/slice/sort.rs:820:5
[...]

PR #4881, opened 2022-06-16
chore: more info for i64-to-u128 panic message

NGA-TRAN:
More debug info for #4761 


compactor/src/query.rs
```diff
@@ -158,8 +158,11 @@ impl QueryChunk for QueryableParquetChunk {
     // it is not safe to use it for sorting chunk
     fn id(&self) -> ChunkId {
         let timestamp_nano = self.min_time.get();
-        let timestamp_nano_u128 =
-            u128::try_from(timestamp_nano).expect("Cannot convert timestamp nano to u128 ");
+        let error_msg = format!(
+            "Cannot convert timestamp nano to u128. Timestamp nano: {}, Paritition id: {}",
+            timestamp_nano, self.partition_id
+        );
+        let timestamp_nano_u128 = u128::try_from(timestamp_nano).expect(&error_msg);
```
domodwyer, line 165: This will always allocate & generate the string message, even when it's not needed.

If we do this instead:

```suggestion
        let timestamp_nano_u128 = u128::try_from(timestamp_nano).unwrap_or_else(|_| panic!(
            "Cannot convert timestamp nano to u128. Timestamp nano: {}, Paritition id: {}",
            timestamp_nano, self.partition_id
        ));
```

The string will only be generated when it is needed :+1:

domodwyer (changes requested):

compactor/src/query.rs
domodwyer: We could also include the error, if that is useful:

```rust
let timestamp_nano_u128 = u128::try_from(timestamp_nano).unwrap_or_else(|e| panic!(
    "Cannot convert timestamp nano to u128. Timestamp nano: {}, Paritition id: {}, error: {}", timestamp_nano, self.partition_id, e
));
```

domodwyer (approved):

PR #4882, opened 2022-06-16
refactor: Remove redundant `MemoryStream`

alamb:
# Rationale
@crepererum  points out on https://github.com/influxdata/influxdb_iox/pull/4873#discussion_r898990360 that IOx has a redundant copy of `MemoryStream`

# Changes
1. Remove code from IOx and use upstream version


alamb:
Marking as draft while I sort out what is wrong

crepererum:
@alamb why is this not working?

PR #4887, opened 2022-06-16
feat: Only query the ingesters relevant to the table 

### LINKED ISSUE #4621: Minimize the number of ingesters contacted by the querier to increase robustness to failure

carols10cents:
Fixes #4621.

This adds a new CLI option, `--sequencer-to-ingester-file`, that takes as a value a path to a JSON file with content mapping each sequencer ID (aka Kafka Partition ID) to an ingester address:

```json
{
  "sequencers": {
    "0": {
      "ingesters": [
        {"addr": "http://ingester-1:8082"},
        {"addr": "http://ingester-3:8082"}
      ]
    },
    "1": { "ingesters": [{"addr": "http://ingester-1:8082"}] }
  }
}
```

Then when the querier gets a query for a particular table and namespace, it uses the sharder to get the relevant sequencer ID. The sequencer ID is looked up to find the corresponding ingester address, and the querier only makes a request to that ingester.

If the sequencer ID doesn't appear in this mapping, the querier falls back to asking all ingesters as it does currently.

To make this PR mergeable and deployable, the existing `--ingester-address` CLI option remains. I tried to make it so that one and only one of `--sequencer-to-ingester-file` and `--ingester-address` was required, but I'm having some issues with clap. Ultimately, `--ingester-address` can be removed and `--sequencer-to-ingester-file` can be required. If both styles are specified, the sequencer-to-ingester mapping is preferred. A future PR can remove all `--ingester-address` support.


crepererum:
Regarding the config:

at least at the moment ingester are configured with ranges, so you don't really need a file but something like `0-100=http://ingester-1:8082,101-120=http://ingester-1:8082` could be parsed from a CLI arg or env var.

querier/src/ingester/mod.rs
```diff
@@ -457,10 +536,10 @@ fn encode_predicate_as_base64(predicate: &Predicate) -> String {
 
 #[async_trait]
 impl IngesterConnection for IngesterConnectionImpl {
-    /// Retrieve chunks from the ingester for the particular table and
-    /// predicate
+    /// Retrieve chunks from the ingester for the particular table, sequencer, and predicate
     async fn partitions(
         &self,
+        sequencer_id: KafkaPartition,
```
crepererum, line 542: You are now creating one request per sequencer ID. This is problematic because:

**Performance:**
An ingester can easily answer a request for ALL its sequencer IDs at once. There is no need for 100 requests (and responses) if the ingester spans a range of 100 sequencers.

**Correctness:**
The querier<>ingester protocol deliberately does NOT contain sequencer IDs (for exactly the performance reason discussed above) so you're now creating duplicate data (might be filtered out by the query engine but I'm not entirely sure and don't want to rely on that). So you now get the full range of sequencer data for every request.

I suggest you change the abstraction to not reason about sequencer IDs per request, but to iterate over the possible ingesters and filter the ones that are relevant for the this query (instead of iterating over the sequencer IDs).

crepererum (changes requested):

mkmik:
> > Regarding the config:
> > at least at the moment ingester are configured with ranges, so you don't really need a file but something like `0-100=http://ingester-1:8082,101-120=http://ingester-1:8082` could be parsed from a CLI arg or env var.
> 
> The config I've implemented is what @mkmik proposed: [#4621 (comment)](https://github.com/influxdata/influxdb_iox/issues/4621#issuecomment-1156709681) @mkmik can you provide more detail on why you chose that format rather than what @crepererum is suggesting here?

my first thought was to define a format like `0..99=http://...,100..199=http` etc
but then I opted against it because that would effectively be a custom format, which we have to parse in iox, and generate with custom string formatting operations in jsonnet. I thus preferred using  simple stoopid json format (for which we can even provide a protobuf schema, hence the lack of the root array). It uses an object instead of an array because this makes it more directly overrideable on jsonnet (so we don't have to do jsonnet contortions to convert an object into an array of objects)

querier/src/ingester/mod.rs
crepererum: > I thought a table would only be on one ingester? 

For now yes, but we probably wanna change that rather soon and the querier shouldn't make assumptions about that fact (and rather treat the sharding system as a blackbox).

> And that each query was for one table?

That's correct. Within querier->ingester RPC a request has the following parameters:

- namespace
- table
- column set
- predicate (for filter pushdown)

It does NOT have a sequencer as a parameter. So for a single-table user query we send at most (prior to this PR: exactly one) request to every ingester. For a multi-table user query (that only works for SQL queries) we can have as many requests as there are tables (however this is not your concern here, see summary below).

> Or otherwise missing how one query would need to contact more than one ingester?

To summarize:

- you need to contact more than one ingester when the sharding system gets improved (I expect this to happen rather soon)
- IOx will contact more than one ingester for multi-table queries. However you don't need to care about this fact here because the whole "give me all the chunks for this table" code you're modifying here is only about a single table. The query engine above that will take care of the multi-table aspect.

crepererum:
@mkmik could we -- in addition to a path to a JSON file -- also accept "inline" JSON within the env variable? IMHO that makes testing and quick&dirty deploys sooo much easier.

mkmik:
Yes, I wanted to mention this too!

I propose:

1. `--flag value`. inline `value`
2. `--flag @filename`. load value from filename

carols10cents:
> I suggest you change the abstraction to not reason about sequencer IDs per request, but to iterate over the possible ingesters and filter the ones that are relevant for the this query (instead of iterating over the sequencer IDs).

@crepererum @mkmik In this case, if the code should be iterating over each ingester and checking if the sequencer IDs that ingester is responsible for are relevant to the query, then wouldn't it make more sense for the JSON format to be a map of ingester addresses to an array of sequencer IDs, like:

```
{
  "sequencers": {
    "http://ingester-1:8082": ["0", "1"],
    "http://ingester-2:8082": ["25"],
    ...
  }
}
```

?

mkmik:
The actual format is not very important.

Your proposal is more compact and easier to write.

My proposal can be a) directly mapped into a datastructure that can be directly looked up and b) we can apply per-sequencer overrides in jsonnet without having a jsonnet PhD (my assumption is that we might want to manually divert traffic for hot sequencers)

But both rationales are rather minimal. I don't bave strong feelings about any format.

mkmik:
Ah, another detail.

Sometimes having a config file that can " evolve" seamlessly (without version numbers, or coordinated image+config updates) can make things simpler.

For example, in my proposal each sequencer value is an object, with an addr field. If we ever want to extend this to add some other parameter we can just add the fields and make sure that a missing field has a sensible backward compatible interpretation.

In your proposal the value of each investor is an array and the value of each array is a number. If you want to evolve that config with some new per ingester or per sequencer config it must be added in a more convoluted way.

carols10cents:
> Your proposal

> My proposal can be a) directly mapped into a datastructure that can be directly looked up

I don't think I was particularly clear-- I prefer your proposal too, *because* it clearly maps to a `HashMap` where I can look up ingester addresses given a sequence number, but @crepererum is saying this code should iterate through each ingester and filter each in or out based on whether the ingester's sequence IDs are relevant to the sequencer IDs returned by the sharder... but that would require reversing the config from your proposal. 

It seems to me that the more straightforward solution is to use your proposal, and for each sequencer ID returned by the sharder, look it up in the hashmap of sequencer IDs to ingester addresses and collect into a `HashSet` to deduplicate. That's what I've implemented in the latest revision of this PR; we'll see what @crepererum thinks of it tomorrow I guess!

mkmik:
yeah sorry I was focusing too much on the config shaping giving it for granted that "one sequencer -> one ingester".

Clearly if we end up having to contact several ingesters because they could be responsible for the given table, we could revise the schema to be:

```json
{
  "sequencers": {
     "0": { "ingesters": [{"addr": "http://ingester-0:8082"},{"addr": "http://ingester-3:8082"}]},
     "1": { "ingesters": [{"addr": "http://ingester-1:8082"}]},
  }
}
```

crepererum:
TBH I don't care too much about the JSON format, I just think that if you have 1000 sequencers that are mapped to 10 ingesters, you probably don't want to have 1000 ingester addresses in this config. The reason for this thought is that we probably want to have quite a large number of sequencers so we can easily re-assign them to ingesters to rebalance the load.

mkmik:
No strong opinions either. I assumed we won't have such a huge discrepancy in number of sequencers and ingesters.

querier/src/ingester/mod.rs
```diff
@@ -576,36 +655,53 @@ impl IngesterConnection for IngesterConnectionImpl {
...
-                    }
+            .flat_map(|sequencer_id| self.sequencer_to_ingester.get(sequencer_id).map(Arc::clone))
+            .collect();
 
-                    res
-                }
-            })
+        let relevant_ingester_addresses = if relevant_ingester_addresses.is_empty() {
+            // If specific ingesters for the sequencers aren't found, query all ingesters
+            self.unique_ingester_addresses.clone()
```
crepererum, line 697: This must error because it's clearly a configuration error. It might even produce wrong results (or panic) because when we start to implement read-only replicas for ingesters, then multiple ingesters may have the same data so you then get the same overlapping information twice and the querier (at least currently) doesn't know how to deal with this. The more I think about it, the more I'm even convinced that the querier should not even try to deal with this because "read-only replicas" will never provide the exact same data as their primary/read-write clone because everything is "eventual".

crepererum: You could probably sanity-check the sharder and the ingester map on startup, so that the sharder cannot produce ingester IDs that are not mapped.

querier/src/ingester/mod.rs
```diff
@@ -1592,6 +1759,34 @@ mod tests {
...
+            let sequencer_to_ingester = ingester_addresses
+                .into_iter()
+                .enumerate()
+                .map(|(sequencer_id, ingester_address)| {
+                    (
+                        sequencer_id as i32 + 1,
+                        Arc::from(ingester_address.as_str()),
+                    )
+                })
+                .collect();
```
crepererum, line 1777: Since you're only using this method once (in `test_flight_per_sequencer_querying`) and the primary goal of that test is to ensure that the sequencer mapping works, I would wish that this map is more hand-written or at least that this code is inlined.

querier/src/ingester/mod.rs
crepererum: Also I think this case is not tested.

crepererum: Afterthought: This case might also happen when the sharder (for some reason) just returns an empty list of sequencers. I think this function here should assert that the sequencer list is not empty (because that really smells like a bug).

carols10cents: > Also I think this case is not tested.

This case was tested by the tests in this file other than the ones added in this PR, [as illustrated by these test failures when I changed this code to assert that the list isn't empty](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/20665/workflows/fcd99691-df89-43fb-9cbe-7bfa49364be5/jobs/176724)

querier/src/ingester/mod.rs
carols10cents: I'm now using this code in almost all of the tests :)

carols10cents:
Ok, I think I have this working now-- please feel free to continue editing and/or merge if you want this to move forward before the end of the week as I'm out Thurs and Fri.

I deleted one end-to-end test that was testing what would happen if the querier wasn't connected to the ingester (which tested that it would get the results once they had been persisted to parquet), I don't think that scenario is valid anymore. It's a bit hard to get the test fixtures to panic with the right error message at the right time because of how the end-to-end tests work...

crepererum:
> I deleted one end-to-end test that was testing what would happen if the querier wasn't connected to the ingester (which tested that it would get the results once they had been persisted to parquet), I don't think that scenario is valid anymore. It's a bit hard to get the test fixtures to panic with the right error message at the right time because of how the end-to-end tests work...

I think we still want to be able to run queriers w/o any ingester connection for testing, for certain deployment scenarios (e.g. a "cold data only" query pool) or emergency.

crepererum:
@mkmik and I had a chat about the config and I have to admit that I was making a misleading suggestion about the sanity checking, sorry.

# Constraints
- config should be checked as good as possible
- when the system recognizes a config error, it should signal that and not silently ignore that or fall back to some silly default
- we should prevent ourselves from DDoSing our own system. This esp. means that we don't just contact "all" ingesters as a default (whatever "all" means because the querier cannot know that set when we forgot to tell it about a specific ingester)
- the config should be easily extendable in the future, but at the same time not provide any knobs that aren't wired up yet
- the config should be unambiguous 

# Scenarios
Here are the scenarios we want to support:

- **boring prod:** querier is connected to ingesters. If the sharder produces a shard that is not mapped to any ingester, the querier should error.
- **degraded prod:** an ingester is unhealthy but we still want to run with the persisted data. In that case, the ingester should marked as "skipped" (or "ignored" or similar) in the config explicitly.
- **development and "cold data only"-clusters:** It should be easy to set up a querier that doesn't talk to ingester at all (i.e. marks all as "ignored" without the need to manually listing them)

# Proposed Config:

```yaml
{
    // flag to ignore all ingesters
    //
    // if this is set to `true` but `ingesters` or `sequencer_mapping` is NOT empty,
    // this should probably error when parsing the config.
    //
    // default: false
    "ignore_all": false,

    // mapping of ingester name to config
    //
    // default: {}
    "ingesters": {
        "i1": {
            // ingester addr as URL
            //
            // if this is `null` but `ignore` is false, this should error during config parsing
            // (for now, we may extend this config later)
            //
            // default: null
            "addr": "http://ingester-1:1234",
        },
        "i2": {
            // flag to ignore this ingester during query time and not contact it
            //
            // default: false
            "ignore": true,
        },
    },

    // mapping of sequencer IDs to ingester names
    //
    // keys are strings because JSON doesn't allow integer keys
    //
    // default: {}
    "sequencer_mapping": {
        "1": {
            // ingester from `ingester` dict
            //
            // use `null` to not map, will error during query (no panic)
            //
            // default: null
            "ingester": "i1",
        },
        "2": {
            "ingester": "i1",
        },
        "3": {
            "ingester": "i2",
        },
        // oh oh, number 4 will fail during query time (return an error, NOT panic) which is OK
        "5": {
            // do NOT fetch data from any ingester
            //
            // default: false
            "ignore": true,
        },
        "6": {
            // "i3" is not a valid ingester. I think this should error while parsing the config.
            "ingester": "i3",
        },
    },
}
```

mkmik:
The config LGTM.

Conventionally our JSON configs are camelCase, I think this one also should follow that.

carols10cents:
@crepererum in [this comment](https://github.com/influxdata/influxdb_iox/pull/4887#issuecomment-1161449248), you said:

> While reading throw the code, I found one important of the config design that @carols10cents proposes here: it's currently impossible to map the same sequencer to multiple ingesters. The reason why this is important is because the code doesn't deal with this fact.

but it looks like the latest proposed config only allows mapping one sequencer to one ingester? Should the `ingester` key of a sequencer instead be `ingesters` and accept an array of ingester names? That is:

```
"2": {
  "ingesters": ["i1", "i2"]
},
```

Or is mapping a sequencer to multiple ingesters no longer a requirement?

mkmik:
It's no longer a requirement.

carols10cents:
Ooookay. I think this is ready for re-review!

crepererum (approved):
Small nit, but that's for a future PR. Rest looks good, so let's merge it and have rather small follow-up clean up PRs (also removing the old CLI option and `TemporaryMigrationSupport`).

I'm really happy about this feature btw. It means we could scale ingester and query tier more easily now.

carols10cents:
I have filed https://github.com/influxdata/influxdb_iox/issues/5002 and https://github.com/influxdata/influxdb_iox/issues/5003 to track the follow-on tasks!

mkmik:
❤️

mkmik:
One of our test clusters after we enabled the change (4 ingester pods):

image

PR #4888, opened 2022-06-16
fix: make ChunkOrder u64 data type to accept min sequence number 0

### LINKED ISSUE #4883: Compactor: New panic 'Sequence number should have been converted to chunk order successfully' after wiping the catalog
After wiping catalog and redeploying (https://github.com/influxdata/k8s-idpe/pull/8060), all 4 compactors hit the same panic a lot: 
`level=error msg="Thread panic" panic_info="panicked at 'Sequence number should have been converted to chunk order successfully', compactor/src/query.rs:263:14" target="panic_logging" location="panic_logging/src/lib.rs:107" time=1655399309075348768`

NGA-TRAN:
Closes #4883 


data_types/src/lib.rs
```diff
@@ -1077,16 +1077,15 @@ impl From<Uuid> for ChunkId {
 /// 1. **upsert order:** chunks with higher order overwrite data in chunks with lower order
 /// 2. **locking order:** chunks must be locked in consistent (ascending) order
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ChunkOrder(NonZeroU32);
+pub struct ChunkOrder(u64);
 
 impl ChunkOrder {
     /// The minimum ordering value a chunk could have. Currently only used in testing.
-    // TODO: remove `unsafe` once https://github.com/rust-lang/rust/issues/51999 is fixed
```
NGA-TRAN, line 1084: The bug is fixed

compactor/src/query.rs
```diff
@@ -257,17 +257,17 @@ impl QueryChunk for QueryableParquetChunk {
     // Order of the chunk so they can be deduplicate correctly
     fn order(&self) -> ChunkOrder {
         let seq_num = self.min_sequence_number.get();
-        let seq_num_u32 = u32::try_from(seq_num).unwrap_or_else(|_| {
+        let seq_num_u64 = u64::try_from(seq_num).unwrap_or_else(|_| {
```
NGA-TRAN, line 260: seq_num is i64. make chunk order u64 to avoid future overflow

data_types/src/lib.rs
```diff
@@ -1077,16 +1077,15 @@ impl From<Uuid> for ChunkId {
 /// 1. **upsert order:** chunks with higher order overwrite data in chunks with lower order
 /// 2. **locking order:** chunks must be locked in consistent (ascending) order
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ChunkOrder(NonZeroU32);
+pub struct ChunkOrder(u64);
```
alamb, line 1080: I wonder if we should just make this `i64` to match sequence number directly?

NGA-TRAN: Good point. Let me make it i64

compactor/src/query.rs
NGA-TRAN: It is now i64 and matches with the type of sequence number

NGA-TRAN:
I am waiting to here from @crepererum

data_types/src/lib.rs
```diff
@@ -1077,16 +1077,15 @@ impl From<Uuid> for ChunkId {
...
 impl ChunkOrder {
     /// The minimum ordering value a chunk could have. Currently only used in testing.
-    // TODO: remove `unsafe` once https://github.com/rust-lang/rust/issues/51999 is fixed
-    pub const MIN: Self = Self(unsafe { NonZeroU32::new_unchecked(1) });
+    pub const MIN: Self = Self(0);
 
     /// Create a ChunkOrder from the given value.
-    pub fn new(order: u32) -> Option<Self> {
-        NonZeroU32::new(order).map(Self)
+    pub fn new(order: i64) -> Option<Self> {
```
alamb, line 1087: I think this could potentially be changed to return `Self` (rather than `Option<Self>`) further simplifying the codebase -- perhaps that could be done as a follow on PR

alamb (approved):

compactor/src/query.rs
crepererum: Technically the sequence number is unsigned, it's just the catalog that treats it as signed because of Postgres. I kinda get the feeling that we're leaking that fact all over the place now just because we didn't find a proper solution to deal with it in the catalog crate.

alamb: How about we add an `assert!(seq_num >= 0)` in `ChunkOrder::new()` and call it a day? Of all the things we could be working on, trying to change the underlying representation of sequence number seems relatively low on the list

alamb: I clarified with @crepererum over slack -- he would prefer if `ChunkOrder` remained an unsigned number (even though SequenceNumber is signed).

I am sorry for suggesting this change @NGA-TRAN 😢  -- I would be happy to help make `ChunkOrder` unsigned again if that would help

NGA-TRAN: @crepererum and @alamb Here is my response on slack 
"Yesterday, when I addressed 
[@alamb](https://influxdata.slack.com/team/U013Y563J5P)
’s comment to make ChunkOrder integer, I was aware that  sequence number was unsigned but since we used postgres, we made it i64. I did think there should not be any problem to make ChunkOrder integer. As long as we have the order, they will be correct. I really prefer to keep it that way for long term solution without adding any panics.  We use i64 for all our unsigned IDs anyway and we are able to keep them the right values. We should be able to keep ChunkOrder the same.
Moreover, there is an on-going discussion here https://docs.google.com/document/d/1AqW73MjWFqkNXwbJ3ZdfEuvJbqOSoh26pdkBVk3NZ0E/edit?disco=AAAAa3abCSg about sequence number. Right now they are still positive but in the future we may want to support negative if it works better for bulk loading old data"

@alamb I can convert it back u64 myself but I prefer to keep it i64

alamb: > @alamb I can convert it back u64 myself but I prefer to keep it i64

I suggest we take @crepererum  at his word that he doesn't think this is ideal but also won't oppose it. 

I vote we merge

NGA-TRAN: I am going to merge this. If we discover any problem, we will change it but I do not think so

PR #4893, opened 2022-06-17
refactor(querier): split ingester partitions into chunks

crepererum:
With the new wire protocol the ingester can now transmit multiple
snapshots per partition with different schemas. This changes the querier
to reflect this and and splits uses the individual snapshots as chunks
for the query engine instead of a single partition.

The schema handling was changed so that instead of a table-wide schema
enforcement, we now use the snapshot-specific projections. This means we
do not need to create all-NULL columns any longer because the batches
within the chunks now always have the correct schema.


querier/src/ingester/mod.rs
```diff
@@ -263,82 +269,187 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
+    }
+
+    disassembler.finalize()
+}
+
+/// Helper to disassemble the data from the ingester Apache Flight arrow stream.
+///
+/// This should be used AFTER the stream was drained because we will perform some catalog IO and this should likely not
+/// block the ingester.
+struct IngesterStreamDisassembler {
```
alamb, line 304: A minor choice of words, but I find the word "Disassembler" to be confusing as logically in my mind this is putting multiple messages together (in a sense).

What about something like `IngesterStreamDecoder` ?

querier/src/ingester/mod.rs
```diff
@@ -263,82 +269,187 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
+                .take()
+                .expect("Partition should have been checked before chunk creation");
+            self.current_partition =
+                Some(current_partition.try_add_chunk(ChunkId::new(), Arc::new(schema), batches)?);
+        }
+
+        Ok(())
+    }
+
+    /// FLush current partition, if any.
```
alamb, line 347:
```suggestion
    /// Flush current partition, if any.
```

querier/src/ingester/mod.rs
```diff
@@ -263,82 +269,187 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
+            current_partition: None,
+            current_chunk: None,
+            ingester_address,
+            table_name,
+            catalog_cache,
+            expected_schema,
+        }
+    }
+
+    /// FLush current chunk, if any.
```
alamb, line 333:
```suggestion
    /// Flush current chunk, if any.
```

querier/src/ingester/mod.rs
```diff
@@ -263,82 +269,187 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
+    fn flush_chunk(&mut self) -> Result<()> {
+        if let Some((schema, batches)) = self.current_chunk.take() {
+            let current_partition = self
+                .current_partition
+                .take()
+                .expect("Partition should have been checked before chunk creation");
```
alamb, line 339:
```suggestion
                .expect("Partition should have been created before chunk creation");
```

querier/src/ingester/mod.rs
```diff
@@ -263,82 +269,187 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
+                self.flush_chunk()?;
+                ensure!(
+                    self.current_partition.is_some(),
+                    ChunkWithoutPartitionSnafu {
+                        ingester_address: self.ingester_address.as_ref()
+                    }
+                );
+
+                // don't use the transmitted arrow schema to construct the IOx schema because some metadata might be
+                // missing. Instead select the right columns from the expected schema.
```
alamb, line 414: 👍

querier/src/ingester/mod.rs
```diff
@@ -1325,62 +1481,21 @@ mod tests {
...
-            vec![batch],
         )
+        .try_add_chunk(ChunkId::new(), Arc::clone(&expected_schema), vec![batch])
         .unwrap_err();
 
         assert_matches!(err, Error::RecordBatchType { .. });
     }
 
-    #[test]
-    fn test_ingester_partition_null_column() {
```
alamb, line 1347: why isn't this test relevant anymore? Can't the ingester still send a record batch without all the needed columns?

querier/src/ingester/mod.rs
```diff
@@ -551,15 +659,45 @@ impl IngesterPartition {
...
+impl IngesterChunk {
+    pub(crate) fn with_partition_sort_key(self, partition_sort_key: Arc<Option<SortKey>>) -> Self {
+        Self {
+            partition_sort_key,
+            ..self
+        }
+    }
+}
+
+impl QueryChunkMeta for IngesterChunk {
```
alamb, line 694: 👍  nice

alamb (approved):
This is a really cool idea -- thank you @crepererum 

I had actually tried to something like this on the ingester side in earlier PRs but couldn't get it to work

querier/src/ingester/mod.rs
crepererum: The ingester can (and does so), but it works differently now on the querier side:

**Before:**
- ingester sends snapshot[^1] A and B with different column sets
- querier puts both into the same query-engine chunk (`IngesterPartition`). Within that unit, all record batches MUST have the schemas.
- Therefore, the querier needs to add all-NULL columns to equalize the schemas.

**After:**
- ingester sends snapshot[^1] A and B with different column sets
- querier puts each of them into its own query-engine chunk (`IngesterChunk`). The query engine handles missing columns for us (this dates back to OG).
- Therefore, all the querier needs to do is to use the right projection for each chunk (I'll mark the place where this is happening)

Note that within a single ingester snapshot, the record batches have the same schema by design. As a reminder, here's the hierarchy on the ingester side:

- namespace:
  - table:
    - sequencer (this is NOT represented in the wire protocol):
      - partition
        - snapshot (e.g. "persisting", "unpersisted")
          - record batch

[^1]: Or "chunk", depending on which side or the wire you sit :grin:

querier/src/ingester/mod.rs
crepererum: done

Issue #4866, opened 2022-06-15
Remove gRPC write API


domodwyer:
The routers expose a [gRPC write API][router] which uses the same wire format as the kafka messages. This coupling of external/public API with internal message formatting creates [problems](https://github.com/influxdata/influxdb_iox/pull/4864#discussion_r898095113).

@pauldix has stated this current gRPC API is unnecessary and should be removed.

[router]:https://github.com/influxdata/influxdb_iox/blob/4df2964566f770f6914c7d252231f4ec24ebacc2/router/src/server/grpc.rs#L148

Currently this API is used by:

1. [`influxdb_iox_client::write_lp()`][lp]
2. [`influxdb_iox_client::write_pb()`][pb]

For 1 this call should be switched to use the HTTP API. 2 should be removed.

[lp]:https://github.com/influxdata/influxdb_iox/blob/4df2964566f770f6914c7d252231f4ec24ebacc2/influxdb_iox_client/src/client/write.rs#L55
[pb]:https://github.com/influxdata/influxdb_iox/blob/4df2964566f770f6914c7d252231f4ec24ebacc2/influxdb_iox_client/src/client/write.rs#L80

crepererum:
I thought we wanted to offer the protobuf format as a more efficient alternative to LP? IIRC the format is/was actually public (why is https://github.com/influxdata/influxdb-pb-data-protocol archived?).

crepererum:
> Having the internal kafka message format be a public API is never going to be a good idea. Ideally we would like to change the kafka messages to reference by object id, rather than object name/strings, for example instead of sending the namespace/table/column/partition key strings, send their catalog IDs. That's not going to be a very good public API.

Depends. Having a Kafka format for which we have proper tooling and that is widely used has it benefits. I'm not sure about the strings, to me it smells like premature optimization w/o numbers.

domodwyer:
It was an example of why evolving the internal message format is made difficult if it's also the public API - any internal changes you want to make to optimise/improve/replace/deprecate/etc instantly affects all users of the public API which is a silly position to put ourselves in.

> Having a Kafka format for which we have proper tooling

This is not really a reason for having a public API use the internals.

domodwyer:
Unfortunately the IOx API client's `write_lp()` method seems to be incompatible with the HTTP write API, as it accepts a single "database name" and not an org/bucket pair (needed for the HTTP API):

[...]

PR #4907, opened 2022-06-20
feat: propagate partition key through kafka

domodwyer:
Pass the `PartitionKey` over the Kafka boundary (router -> ingester).

The ingester currently does nothing with this value, in order to facilitate a graceful deployment. The next PR in this series will change the ingester to consume this `PartitionKey` once this PR has been merged and fully deployed.

---

* feat: propagate partition key through kafka (c1f715403)

      Changes the kafka message wire format to include the partition key for 
      serialised DML writes on the wire.

      After this commit, the kafka messages will contain the partition key for each
      op, but this information will go unused in the ingester - this enables us to
      roll out the producer side, before making the value's presence necessary on
      the consumer side.

      A follow-up PR will change the ingester to utilise this embedded partition
      key.

      This has the unfortunate side effect of making the partition key part of the
      public gRPC write API:

          https://github.com/influxdata/influxdb_iox/issues/4866

### DIFF
--- a/write_buffer/src/codec.rs
+++ b/write_buffer/src/codec.rs
@@ -1,7 +1,7 @@
 //! Encode/Decode for messages
 
 use crate::core::WriteBufferError;
-use data_types::{NonEmptyString, Sequence};
+use data_types::{NonEmptyString, PartitionKey, Sequence};
 use dml::{DmlDelete, DmlMeta, DmlOperation, DmlWrite};
 use generated_types::{
     google::FromOptionalField,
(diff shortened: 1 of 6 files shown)


write_buffer/src/codec.rs
```diff
@@ -1,7 +1,7 @@
 //! Encode/Decode for messages
```
crepererum, line 1: This requires an additional test (or an extension of an existing one) in `write_buffer/src/core.rs`.

crepererum (changes requested):

write_buffer/src/codec.rs
crepererum: Clarified w/ @domodwyer via Slack:

the tests already use the partition key but because the assertion wasn't there, the `DmlWrite` roundtrip "error" (= loosing the partition key) was accepted.

crepererum (approved):

PR #4908, opened 2022-06-21
refactor: store per-file column set in catalog

crepererum:
Together with the table-wide schema and the partition-wide sort key, this should
be everything we need to read a parquet file directly into memory
without peeking any file-level metadata.

The querier will use this to directly load parquet files into the read
buffer.

**WARNING: This requires a catalog wipe!**

Ref #4124.


mkmik:
Is a catalog wipe a string necessity or is it just because we decide it's not worth the effort to implement a graceful migration?

crepererum:
I could implement some migration but it would require actual client code to run because I would need to decode the binary parquet metadata to extract the column set. So it won't just work on postgres.

The wipe doesn't have to be in-sync with the deploy of this change though, it can be done later. This is because this change only writes the set but doesn't use it anywhere.

data_types/src/lib.rs
```diff
@@ -837,6 +837,21 @@ pub struct ParquetFile {
...
+    ///
+    /// Columns that are NOT part of the table-wide schema must be ignored. It is likely that these
+    /// columns were originally part of the table but were later removed.
+    ///
+    /// # Column Types
+    /// Column types are identical to the table-wide types.
+    ///
+    /// # Column Order & Sort Key
+    /// The columns that are present in the table-wide schema are sorted according to the partition
+    /// sort key. The occur in the parquet file according to this order.
```
domodwyer, line 853:
```suggestion
    /// sort key. They occur in the parquet file ordered by the partition sort key.
```

data_types/src/lib.rs
```diff
@@ -837,6 +837,21 @@ pub struct ParquetFile {
...
+    pub column_set: Vec<String>,
```
domodwyer, line 854: Can I convince you a type wrapper is a good idea? 

I've already been trying to trim down the number of bare `String` we pass around. Perhaps a `ColumnSet(Vec<String>)` with a `#[sqlx(transparent)]` annotation so it's compatible with sqlx with no additional work :+1:

mkmik:
There have been a few things we didn't get right in the last wipe so on one hand I'm looking forward wiping again (e.g. the topic config)

parquet_file/src/metadata.rs
```diff
@@ -406,9 +406,11 @@ impl IoxMetadata {
         let schema = decoded
             .read_schema()
             .expect("failed to read encoded schema");
-        let time_summary = decoded
+        let stats = decoded
             .read_statistics(&*schema)
-            .expect("invalid statistics")
+            .expect("invalid statistics");
+        let columns = stats.iter().map(|v| v.name.clone()).collect();
```
domodwyer, line 412: Deriving it from the result rather than passing it in gets a big :+1: from me.

domodwyer (approved):

iox_catalog/migrations/20220620160136_parquet_file_columns.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE parquet_file ADD COLUMN column_set TEXT[];
```
domodwyer, line 1: It would be nice to follow this up by making it `NOT NULL` once the migration is complete as we did with the sort key.

data_types/src/lib.rs
crepererum: done

alamb:
It will be interesting to see if/how the routers can keep up after the next catalog wipe 👍

data_types/src/lib.rs
```diff
@@ -837,6 +880,21 @@ pub struct ParquetFile {
     pub compaction_level: i16,
     /// the creation time of the parquet file
     pub created_at: Timestamp,
+    /// Set of columns within this parquet file.
```
alamb, line 883: This is a clever idea to not store the entire schema but instead store the columns. 👍 Double bonus for storing in human readable (string array) format

PR #4910, opened 2022-06-21
refactor: remove table name from read buffer

crepererum:
The low-level chunk storage shouldn't care about the table name (this is
also true for parquet chunks btw). In fact, the table name is already
only a partial information since it misses the namespace.

If we need a table name, then the high-level chunk/data management is
responsible for that.


read_buffer/src/chunk.rs
```diff
@@ -29,23 +29,14 @@ pub enum Error {
...
-    #[snafu(display("column '{}' does not exist in table '{}'", column_name, table_name))]
-    ColumnDoesNotExist {
-        column_name: String,
-        table_name: String,
-    },
+    #[snafu(display("no data to build chunk"))]
+    ChunkBuilderNoInput,
+
+    #[snafu(display("error building chunk: {}", source))]
+    ChunkBuilderError { source: arrow::error::ArrowError },
```
alamb, line 36: I do think long term having the namespace / table information in errors / logs will be critical to being able to debug IOx in production. If we need to plumb it in a different way in some future PR that would be ok too

alamb (approved):

read_buffer/src/chunk.rs
crepererum: Then this information should be added by the higher level by wrapping the low-level errors correctly.

PR #4912, opened 2022-06-21
refactor: assorted fixes and prep work for #4124

crepererum:
See commits. Mostly smaller fixes and feature additions so I can actually load read buffer chunks w/o decoding the parquet metadata first (this will be the next PR).

Ref #4124.


alamb:
Test failure seems real: https://circleci.com/gh/influxdata/influxdb_iox/176146?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

parquet_file/src/storage.rs
```diff
@@ -276,6 +288,11 @@ async fn download_and_scan_parquet(
     let file_reader = SerializedFileReader::new(Bytes::from(data))?;
     let mut arrow_reader = ParquetFileArrowReader::new(Arc::new(file_reader));
 
+    let schema = arrow_reader.get_schema()?;
+    if expected_schema.as_ref() != &schema {
+        return Err(ReadError::SchemaMismatch);
```
alamb, line 293: I recommend including the `path` in this error so we have some idea of what file is causing the problem

querier/src/cache/namespace.rs
```diff
@@ -119,6 +127,29 @@ impl NamespaceCache {
...
+            if let Some(namespace) = cached_namespace.as_ref() {
+                if let Some(table) = namespace.schema.tables.get(table_name) {
+                    let covered: HashSet<_> = table.columns.keys().cloned().collect();
+                    columns.iter().any(|col| !covered.contains(col.as_str()))
+                } else {
+                    // table unknown => need to update
+                    true
+                }
+            } else {
+                // nomespace unknown => need to update
```
alamb, line 148:
```suggestion
                // namespace unknown => need to update
```

querier/src/cache/namespace.rs
```diff
@@ -119,6 +127,29 @@ impl NamespaceCache {
...
+    pub fn expire_if_schema_does_not_cover(
+        &self,
+        namespace_name: Arc<str>,
+        table_name: &str,
+        columns: &HashSet<String>,
+    ) {
+        self.backend.remove_if(&namespace_name, |cached_namespace| {
+            if let Some(namespace) = cached_namespace.as_ref() {
+                if let Some(table) = namespace.schema.tables.get(table_name) {
+                    let covered: HashSet<_> = table.columns.keys().cloned().collect();
```
alamb, line 141: I the `cloned` necessary? It seems like refs would be just fine as `covered` is not used again -- it would be cool to avoid  copying the string

alamb (approved):

parquet_file/src/storage.rs
crepererum: done

parquet_file/src/storage.rs
```diff
@@ -288,9 +291,12 @@ async fn download_and_scan_parquet(
     let file_reader = SerializedFileReader::new(Bytes::from(data))?;
     let mut arrow_reader = ParquetFileArrowReader::new(Arc::new(file_reader));
 
-    let schema = arrow_reader.get_schema()?;
+    // Check schema but ignore the metadata
+    let schema = arrow_reader
+        .get_schema()?
+        .with_metadata(expected_schema.metadata().clone());
```
alamb, line 297: 👍

alamb (approved):

querier/src/cache/namespace.rs
crepererum: fixed

Issue #4886, opened 2022-06-16
Research what requires so much memory on the ingester during persist


alamb:
We saw in https://github.com/influxdata/conductor/issues/970 that something (we think it is the ingester persist operation) on a partition that had less than `300MB` of write data consumed over 30GB of RAM and caused our ingesters to crash.

This ticket tracks investigating the root cause of what was consuming so much memory

alamb:
My high level plan for this is to add tracing to the plans run by the ingester (so I can see what the plans are doing)

I am trying to understand what is taking so much memory in the ingester (see https://github.com/influxdata/conductor/issues/970). I suspect it is the datafusion plans, and I want to confirm that. To do so I am hope to use the existing tracing mechanism

We already have a pretty sophisticated tracing mechanism to see what the plans are doing (connected to the datafusion statistics), but they are not hooked up to the ingester (either for user queries or for automated background tasks like `persist` and `compact` operations.

Possible subtasks: 
1. Add ingester information to querier tracing (to see where time is spent)
3. A way to create spans for internal operations like persist
2. Add a way to generate spans for the ingester's persist operations with some random frequency

alamb:
See it happening again https://github.com/influxdata/conductor/issues/980

alamb:
I  spent some time trying to reproduce the large memory explosion we were seeing in the ingester on prod but was not able to.

# What I tried

What I did was use the code on the https://github.com/influxdata/influxdb_iox/tree/alamb/memory_profile branch, which effectively turns off compaction with something like

```diff
diff --git a/influxdb_iox/src/commands/run/all_in_one.rs b/influxdb_iox/src/commands/run/all_in_one.rs
index 9fa3e623b..8b00407fa 100644
--- a/influxdb_iox/src/commands/run/all_in_one.rs
+++ b/influxdb_iox/src/commands/run/all_in_one.rs
@@ -502,17 +502,17 @@ pub async fn command(config: Config) -> Result<()> {
     )
     .await?;

-    info!("starting compactor");
-    let compactor = create_compactor_server_type(
-        &common_state,
-        Arc::clone(&metrics),
-        Arc::clone(&catalog),
-        Arc::clone(&object_store),
-        Arc::clone(&exec),
-        Arc::clone(&time_provider),
-        compactor_config,
-    )
-    .await?;
+    // info!("starting compactor");
+    // let compactor = create_compactor_server_type(
+    //     &common_state,
+    //     Arc::clone(&metrics),
[...]

PR #4917, opened 2022-06-21
fix: fix heappy + update docs

alamb:
re https://github.com/influxdata/influxdb_iox/issues/4886

I want to find out what is taking so much memory during persist. I wanted to use heappy from @mkmik 

turns out it was broken and the docs were out of date.

This PR fixes both

See the rendered instructions at https://github.com/influxdata/influxdb_iox/blob/alamb/heappy_instructions/docs/profiling.md#iox--heap-profiling


mkmik (approved):
Keep in mind that the in_use metric exposed by ghe measure_free feature doesn't make much sense since the stack traces of the places where memory is deallocated doesn't match the places where is allocated, except in very special circumstances. I have a prototype of a fix in a branch

PR #4921, opened 2022-06-22
fix: column handling when reading parquet files

crepererum:
This improves/fixes/tests a few aspects when reading parquet files:

- fix usage of `Selection::Some(...)`. This was broken since #4912 but
  apparently no test caught that.
- ensure that the order of `Selection::Some(...)` is preserved
- ensure that schema metadata is attached to output batches
- ignore parquet columns that we don't care about (i.e. do not select)
- allow parquet file to have a different column order than our internal
  bookkeeping, this makes it way simpler to read parquet files w/o
  scanning the metadata first
- extend the test coverage

Ref #4124.


parquet_file/src/storage.rs
```diff
@@ -234,31 +223,40 @@ impl ParquetStorage {
 
 /// Return indices of the schema's fields of the selection columns
 fn column_indices(selection: Selection<'_>, schema: SchemaRef) -> Vec<usize> {
```
alamb, line 225: I wonder if you might be able to use Schema::df_projection for this:

https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/schema/src/lib.rs#L303-L332

parquet_file/src/storage.rs
```diff
@@ -234,31 +223,40 @@ impl ParquetStorage {
...
+                .collect();
+            cols.iter()
+                .filter_map(|c| fields_lookup.get(c).cloned())
+                .collect()
+        }
+        Selection::All => (0..schema.fields().len()).collect(),
     }
 }
 
+fn project_schema(schema: &Schema, projection: &[usize]) -> Schema {
```
alamb, line 243: Perhaps Schema::select_by_indices would work here

https://github.com/influxdata/influxdb_iox/blob/ddf61c5e984724449a1038409d9573c003946a67/schema/src/lib.rs#L335-L350

parquet_file/src/storage.rs
```diff
@@ -291,18 +289,49 @@ async fn download_and_scan_parquet(
...
+        };
+        let file_field = file_schema.field(file_idx);
+        if field != file_field {
+            return Err(ReadError::SchemaMismatch { path });
+        }
+        mask.push(file_idx);
     }
 
-    let mask = ProjectionMask::roots(arrow_reader.parquet_schema(), projection);
+    // for some weird reason, the parquet-rs projection system only filters columns but ignores the mask order, so we
```
alamb, line 315: I think the reason is that supporting reorder during parquet decoding significantly complicates complex struct decoding -- https://github.com/apache/arrow-rs/pull/1716 I think is related

I thought a very similar set of code for projecting / reordering was also in DataFusion -- though I couldn't  find it when I looked maybe @tustvold  can remember.


In general, I think this code might benefit from being extracted (ideally into `Schema` somewhere) and tested individually

parquet_file/src/storage.rs
```diff
@@ -465,26 +532,46 @@ mod tests {
...
+        upload(&store, &meta, batch).await;
 
-        // Drain the retrieved record batch stream
-        datafusion::physical_plan::common::collect(rx)
+        download(&store, &meta, Selection::All, schema)
             .await
             .unwrap();
     }
 
+    #[tokio::test]
```
alamb, line 544: The test coverage is 👌  very nice

parquet_file/src/storage.rs
```diff
@@ -465,26 +532,46 @@ mod tests {
...
+        ])
+        .unwrap();
+        let expected_batch =
+            RecordBatch::try_from_iter([("a", to_string_array(&["value"]))]).unwrap();
+        let schema = expected_batch.schema();
+        assert_roundtrip(file_batch, Selection::All, schema, expected_batch).await;
+    }
+
+    #[tokio::test]
+    async fn test_schema_check_ignores_type_for_unselected_column() {
```
alamb, line 558: I am not sure why this is a good idea (ignoring schema mismatches) but it seems a reasonable behavior

alamb (approved):

parquet_file/src/storage.rs
crepererum: The problem is: the whole crate uses the arrow schema type, not the IOx schema type. The reason for that is (I think) that `read_filter` ultimately comes from DF which uses the arrow type. I could convert back and forth be between the two though, but I'm not sure if that's cleaner.

parquet_file/src/storage.rs
crepererum: Same argument as [here](https://github.com/influxdata/influxdb_iox/pull/4921#discussion_r903596747).

parquet_file/src/storage.rs
alamb: Yeah it will be great when @tustvold  can get the DataFusion parquet reader to the point where it can be used by IOx and we don't have to maintain two copied

parquet_file/src/storage.rs
alamb: I guess I was thinking of the code that gets a logical description of "projection on `arrow::data_types::Schema` and correctly maps that to the projection required for the `parquet-rs` reader

crepererum: done

parquet_file/src/storage.rs
```diff
@@ -343,6 +328,73 @@ async fn download_and_scan_parquet(
...
+    FieldTypeMismatch {
+        /// Expected field.
+        expected: Field,
+
+        /// Actual field.
+        actual: Field,
+    },
+}
+
+/// Calculate project for the parquet-rs reader.
```
alamb, line 350: 👍 

thank you

alamb (commented):
love it

PR #4922, opened 2022-06-22
refactor: make querier internals private

crepererum:
Queries internals are not meant to be used by other crates. Only a
handful selected interfaces should be used by IOxD and the query tests.

The compactor only used a very small subset just to read parquet files
back into memory. It shall rather use the official `parquet_file`
interface instead.


alamb (approved):
looks like a nice cleanup to me

PR #4924, opened 2022-06-22
feat: do not rely on encoded parquet metadata for RB chunks

crepererum:
~Requires #4921 and #4922.~

Ref #4124.


querier/src/chunk/mod.rs
```diff
@@ -382,50 +385,126 @@ impl ChunkAdapter {
...
+            .schema(namespace_name)
+            .await?;
+        self.catalog_cache
+            .partition()
+            .expire_if_sort_key_does_not_cover(parquet_file.partition_id, &file_columns);
+        let partition_sort_key = self
+            .catalog_cache
+            .partition()
+            .sort_key(parquet_file.partition_id)
+            .await;
```
alamb, line 419: Reading this code I worry that someone in the future (like me 😢 ) might forget to call `expire_if_sort_key_does_not_cover` prior to calling `sort_key` -- what would you think about automatically calling `expire_if_sort_key_does_not_cover` as part of `sort_key` or maybe making a function that does both expire and sort key lookup?

Same question goes for `expire_if_schema_does_not_cover`

querier/src/chunk/mod.rs
```diff
@@ -382,50 +385,126 @@ impl ChunkAdapter {
...
+            .as_ref()
+            .as_ref()
+            .expect("partition sort key should be set when a parquet file exists");
+        let table_schema: Schema = namespace_schema
+            .tables
+            .get(table_name.as_ref())?
+            .clone()
+            .try_into()
+            .expect("Invalid table schema in catalog");
+        let table_columns: HashSet<String> = table_schema
```
alamb, line 432: Could be `HashSet<&str>` potentially

querier/src/table/mod.rs
```diff
@@ -756,11 +760,12 @@ mod tests {
 
     /// Adds a "foo" column to the table and returns the created schema
     async fn make_schema(table: &Arc<TestTable>) -> Arc<Schema> {
-        table.create_column("foo", ColumnType::I64).await;
+        table.create_column("foo", ColumnType::F64).await;
```
alamb, line 763: what is the reason to change this type?

alamb (approved):

querier/src/chunk/mod.rs
```diff
@@ -382,50 +385,126 @@ impl ChunkAdapter {
...
         &self,
-        decoded_parquet_file: Arc<DecodedParquetFile>,
+        namespace_name: Arc<str>,
+        parquet_file: Arc<ParquetFile>,
     ) -> Option<QuerierRBChunk> {
-        let parquet_file_id = decoded_parquet_file.parquet_file_id();
-        let schema = decoded_parquet_file.schema();
-        let chunk_id = ChunkId::from(Uuid::from_u128(parquet_file_id.get() as _));
+        // prepare lookup data structures
+        let file_columns: HashSet<String> = parquet_file.column_set.iter().cloned().collect();
```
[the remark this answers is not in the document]
crepererum: done

querier/src/chunk/mod.rs
crepererum: done.

It doesn't just help you (or any future contributor). This change required my to think about the the cache usage a bit and I needed to adjust the partition sort key handling during the ingester response decoding (see [this commit](https://github.com/influxdata/influxdb_iox/pull/4924/commits/01da953d60c7e906c362c7c24d234c42368f9197)). I think the existing behavior didn't lead to a bug because the partition primary keys are [synced anyways](https://github.com/influxdata/influxdb_iox/blob/c899c3a0f4b78e336572f6037d1ebde3181277ae/querier/src/table/state_reconciler.rs#L211) but any refactoring could have exposed stale partition sort keys.

So overall: good call :muscle: :brain:

crepererum: Merging this even though there are a few changes on top of the reviewed state because I wanna see the impact of this endeavour for the querier today. I can always open a follow-up PR to fix/adjust stuff.

Issue #4928, opened 2022-06-22
Ingester should pause ingesting whenever persisting can't keep up with ingesting


alamb:
# Use case: 
There are times (during replay especially) that we saw the ingester ingesting faster than the `persist` operation. If this continues unabated eventually the ingester will be OOM killed (due to out of memory) as each persist gets bigger. 

This meant that despite setting `INFLUXDB_IOX_PERSIST_PARTITION_SIZE_THRESHOLD_BYTES` to limit the maximum size of an individual persist operation, we were persisting larger partitions -- see here for an example https://github.com/influxdata/conductor/issues/980#issuecomment-1159094137 

# Proposal
Pause ingest if a partition exceeds the `INFLUXDB_IOX_PERSIST_PARTITION_SIZE_THRESHOLD_BYTES` setting and a persist is already underway.

alamb:
>  Just to clarify, you'll only pause if a persist is already underway for the partition that has crossed the threshold?

Yes,  only pause if a persist is already underway for the partition that has crossed the threshold (again)

alamb:
> Just to clarify, you'll only pause if a persist is already underway for the partition that has crossed the threshold?

In coding this up, @pauldix mentioned to me that persists happen "in batches" meaning no new persist operation happens while there is an existing persist operation outstanding. 

Thus as soon as any partition exceeds the partition threshold it can't be persisted until the next persist batch is calculated.  

Once we make persist happen per-partition (rather than in batches) changing the algorithm to pause only when a partition with an outstanding persist goes over makes sense.

alamb:
Per discussions, this goal has been lowered in priority, so unassigning myself https://influxdata.slack.com/archives/CRK9M8L5Q/p1657812384568659?thread_ts=1657739394.189219&cid=CRK9M8L5Q

The ingester I think is "stable enough" for the time being

alamb:
Resetting this as "nice to have" status as I am not sure it is actually needed / useful

jacobmarble:
Related to "bimodal workload priority".

alamb:
I think github got over eager closing this ticket -- I believe @domodwyer 's intention on https://github.com/influxdata/influxdb_iox/pull/6316 was that when we switch IOx to use ingester2 this issue will be solved

PR #4930, opened 2022-06-22
refactor: Minor refactor in ingester lifecycle tests

alamb:
# Rationale
I am studying the code for https://github.com/influxdata/influxdb_iox/issues/4928 and playing with it is a good way to learn; less code is better code

# Changes:
1. Remove `TestLifeCycleManager`

### DIFF
--- a/ingester/src/job.rs
+++ b/ingester/src/job.rs
@@ -25,6 +25,7 @@ impl Job {
 /// The global job registry
 #[derive(Debug)]
 pub struct JobRegistry {
+    pub metric_registry: Arc<metric::Registry>,
     inner: Mutex<TaskRegistryWithMetrics<Job, TaskRegistryWithHistory<Job, TaskRegistry<Job>>>>,
 }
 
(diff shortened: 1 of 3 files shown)


ingester/src/job.rs
```diff
@@ -25,6 +25,7 @@ impl Job {
 /// The global job registry
 #[derive(Debug)]
 pub struct JobRegistry {
+    pub metric_registry: Arc<metric::Registry>,
```
tustvold, line 28: Why does this need to be pub?

iox_time/src/lib.rs
```diff
@@ -182,6 +182,9 @@ impl Time {
 }
 
 pub trait TimeProvider: std::fmt::Debug + Send + Sync + 'static {
+    /// Returns `self` as [`std::any::Any`] for runtime downcasts
+    fn as_any(&self) -> &dyn std::any::Any;
```
tustvold, line 186: This is only needed for tests, and so it seems a little strange to have it leak out in this way...

tustvold (commented):
I'm not sure I agree with this change, it effectively seems to leak test harness code into non-test code, which seems undesirable...

alamb:
not worth pursuing

PR #4931, opened 2022-06-22
fix: Pause ingester if `persist` can't keep up with ingest

alamb:
# Rationale
Closes https://github.com/influxdata/influxdb_iox/issues/4928

There are times (during replay especially) that we saw the ingester ingesting faster than the `persist` operation could keep up (perhaps due to https://github.com/influxdata/influxdb_iox/issues/4886). If this continues unabated eventually the ingester will be OOM killed (due to out of memory) as each persist gets bigger.

This meant that even after setting `INFLUXDB_IOX_PERSIST_PARTITION_SIZE_THRESHOLD_BYTES` to limit the maximum size of an individual persist operation, we were persisting larger partitions -- see here for an example https://github.com/influxdata/conductor/issues/980#issuecomment-1159094137

Bonus is that this doesn't require any new knobs

# Changes:
Pause ingest if any partition exceeds the `INFLUXDB_IOX_PERSIST_PARTITION_SIZE_THRESHOLD_BYTES` setting while a persist to that same partition is underway

# Expected change in behavior:
1. There should be fewer OOM errors in production, and we can increase the overall "ingest pause" threshold to get more memory bufferd
2. The ingester will pause more frequently (especially when recovering large periods of time)

### DIFF
--- a/ingester/src/lifecycle.rs
+++ b/ingester/src/lifecycle.rs
@@ -89,23 +89,34 @@ impl LifecycleHandle for LifecycleHandleImpl {
             info!(
                 total_bytes = s.total_bytes,
                 pause_ingest_size = self.config.pause_ingest_size,
-                "Total ingested is over size threshold, pausing"
+                "Total ingested size is over size threshold, pausing"
             );
             return true;
         }
 
-        // if this partition's size exceeds limits and it already has
-        // an outstanding persist, pause further ingest
-        // https://github.com/influxdata/influxdb_iox/issues/4928
+        // If this partition's size exceeds its threshold and the
+        // ingester has outstanding persist, pause further ingest
+        // until outstanding persists are complete.
+        //
+        // This is designed to protect the ingester from OOMing by
+        // when partitions have higher write throughput than persist
+        // throughput (e.g. during a long recovery when data in kafka
+        // is plentify or if there is something about a particular
+        // persist operation that is abnormally slow).
+        //
+        // We pause all ingest if *any* partition exceeds its limit
+        // because persists happen in batches so *no* new persist will
+        // start until outstanding persists are complete.
+        //
+        // More details: https://github.com/influxdata/influxdb_iox/issues/4928
         if partition_bytes_written > self.config.partition_size_threshold
-            && s.persists.contains_key(&partition_id)
+            && s.persist_state.persisting()
         {
             info!(%partition_id,
-                      partition_bytes_written=partition_bytes_written,
-                      partition_size_threshold=self.config.partition_size_threshold,
-                      "Partition with ongoing persist is over size threshold, pausing");
-            s.persists.insert(partition_id, true);
-
+                  partition_bytes_written=partition_bytes_written,
+                  partition_size_threshold=self.config.partition_size_threshold,
+                  "Partition exceeded size threshold with ongoing persist, pausing");
+            s.persist_state = s.persist_state.pause();
             return true;
         }
 
@@ -117,8 +128,8 @@ impl LifecycleHandle for LifecycleHandleImpl {
         let s = self.state.lock();
         // resume ingest if
         // 1. the total size is under cap and
-        // 2. we didn't pause due to outstanding persists
-        s.total_bytes < self.config.pause_ingest_size && s.persists.values().all(|f| !f)
+        // 2. we didn't pause due to partition size with outstanding persists
+        s.total_bytes < self.config.pause_ingest_size && !s.persist_state.paused()
     }
 }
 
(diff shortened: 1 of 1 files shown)


ingester/src/lifecycle.rs
```diff
@@ -352,7 +476,7 @@ impl LifecycleManager {
         let persist_tasks: Vec<_> = to_persist
             .into_iter()
             .map(|s| {
-                self.remove(s.partition_id);
+                self.mark_persisting(s.partition_id);
```
alamb, line 479: Function was renamed to better describe what it is doing

ingester/src/lifecycle.rs
```diff
@@ -495,6 +630,95 @@ mod tests {
         }
     }
 
+    #[derive(Debug, Clone)]
+    /// Synchronizes waiting on some test event
+    struct EventBarrier {
```
alamb, line 635: This is copy/paste from elsewhere in the ingester tests -- Given it is like 10 lines long I thought that was ok, but if others feel strongly I can put it somewhere reusable (like `test_utils`)

ingester/src/lifecycle.rs
```diff
@@ -729,6 +953,88 @@ mod tests {
         assert_eq!(size_counter, 1);
     }
 
+    #[tokio::test]
+    /// Ensure ingest is paused if partition ingests faster than persist can keep up
```
alamb, line 957: here is the test

ingester/src/lifecycle.rs
```diff
@@ -78,13 +80,56 @@ impl LifecycleHandle for LifecycleHandleImpl {
...
+        if s.total_bytes > self.config.pause_ingest_size {
+            info!(
+                total_bytes = s.total_bytes,
+                pause_ingest_size = self.config.pause_ingest_size,
+                "Total ingested size is over size threshold, pausing"
+            );
+            return true;
+        }
+
+        // If this partition's size exceeds its threshold and the
```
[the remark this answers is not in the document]
alamb: Here is further discussion on this topic: https://github.com/influxdata/influxdb_iox/issues/4928#issuecomment-1164264033

ingester/src/lifecycle.rs
```diff
@@ -78,13 +80,56 @@ impl LifecycleHandle for LifecycleHandleImpl {
...
+        {
+            info!(%partition_id,
+                  partition_bytes_written=partition_bytes_written,
+                  partition_size_threshold=self.config.partition_size_threshold,
+                  "Partition exceeded size threshold with ongoing persist, pausing");
+            s.persist_state = s.persist_state.pause();
+            return true;
+        }
+
+        // can keep consummng
```
mkmik, line 123:
```suggestion
        // can keep consuming
```

ingester/src/lifecycle.rs
```diff
@@ -411,10 +537,18 @@ impl LifecycleManager {
         }
     }
 
-    /// Removes the partition from the state
-    fn remove(&self, partition_id: PartitionId) -> Option<PartitionLifecycleStats> {
+    /// Removes the partition from the state and marks it peristing
```
mkmik, line 540:
```suggestion
    /// Removes the partition from the state and marks it persisting
```

ingester/src/lifecycle.rs
```diff
@@ -495,6 +630,95 @@ mod tests {
...
+            sequencer_id: SequencerId,
+            sequence_number: SequenceNumber,
+        ) {
+            self.inner
+                .update_min_unpersisted_sequence_number(sequencer_id, sequence_number)
+                .await
+        }
+    }
+
+    impl PausiblePersister {
```
mkmik, line 685: your code is applausable but perhaps the word "Pausable" is more plausible than "Pausible"

mkmik (approved):
LGTM

domodwyer:
I intend to review this shortly, but I suspect it will significantly increase the time to recover TTBR after a deployment, due to the increased pausing.

ingester/src/lifecycle.rs
```diff
@@ -78,13 +80,56 @@ impl LifecycleHandle for LifecycleHandleImpl {
...
+        }
+
+        // If this partition's size exceeds its threshold and the
+        // ingester has outstanding persist, pause further ingest
+        // until outstanding persists are complete.
+        //
+        // This is designed to protect the ingester from OOMing by
+        // when partitions have higher write throughput than persist
+        // throughput (e.g. during a long recovery when data in kafka
+        // is plentify or if there is something about a particular
```
domodwyer, line 104:
```suggestion
        // is plentiful or if there is something about a particular
```

ingester/src/lifecycle.rs
```diff
@@ -78,13 +80,56 @@ impl LifecycleHandle for LifecycleHandleImpl {
...
+        // persist operation that is abnormally slow).
+        //
+        // We pause all ingest if *any* partition exceeds its limit
+        // because persists happen in batches so *no* new persist will
+        // start until outstanding persists are complete.
```
domodwyer, line 109:
```suggestion
        // We pause all ingest if *any* partition exceeds its limit whilst there
        // is an ongoing persist operation, because persists happen in batches
        // so *no* new persist will start until outstanding persists are
        // complete.
```

alamb:
Thank you for the excellent reviews @mkmik  and @domodwyer 

> While reading this PR (unrelated to the content of the PR, but perhaps related to the problem it is addressing) I noticed that when we trigger a persist operation, we immediately stop accounting for the memory the partition is using:

This is an excellent point. Filed https://github.com/influxdata/influxdb_iox/issues/4936 to track and I will fix that in a follow on PR

> It might be worth trying this to see how much it helps the OOMing before merging this change, as this will also negatively impact the TTBR recovery rate.

This change (along with increasing the overall memory limit) I think could potentially significantly *decrease* the TTBR recovery rate because it will reduce the number of parquet files being written (and thus catalog load).  I think the code in this PR is needed as a self-protection mechanism against OOMs in general, but I hope after we sort out https://github.com/influxdata/influxdb_iox/issues/4886 it will be triggered rarely in practice

Thus my plan currently is:
1. Wait for @mkmik  to complete his experiments with the horde of 50 ingesters (https://github.com/influxdata/conductor/issues/988)
2. Merge this PR in and reduce the force-ingest limits on the ingesters
3. Work on https://github.com/influxdata/influxdb_iox/issues/4936

ingester/src/lifecycle.rs
alamb: Fixed in bdfeb7334 🤦

domodwyer:
> This change (along with increasing the overall memory limit) I think could potentially significantly decrease the TTBR recovery rate because it will reduce the number of parquet files being written (and thus catalog load)

I'm not sure I understand the reasoning - I don't believe the catalog is currently the bottleneck for ingest, but rather the single core utilisation we have observed.

Perhaps a quick call to compare notes would be good before going ahead with this?

alamb:
Converting to draft so we don't accidentally merge this while working on https://github.com/influxdata/influxdb_iox/issues/4936

alamb:
https://github.com/influxdata/influxdb_iox/issues/4936 is merged, I will now attempt to crank up the limits on prod ingesters and see what happens

alamb:
Update here: after https://github.com/influxdata/influxdb_iox/issues/4936 was merged and we updated the total memory limits on the compactor we haven't seen OOMs yet. 

However, I still see evidence of some partitions growing beyond their per-partition limit. For example, the following log from `iox-shared-ingester-2-0` shows a partition had 70MB in it when the limit was set to 50MB. 

```
level=info msg="Partition is over size threshold, persisting" sequencer_id=65 partition_id=1310399 bytes_written=70490886 partition_size_threshold=52428800 target="ingester::lifecycle" location="ingester/src/lifecycle.rs:299" time=1656058324585497017
```

@domodwyer  any thoughts about why that might be or what we should do about it (aka should we also merge this PR?)

domodwyer:
(copying from slack for posterity)

We may have a bug causing the per-partition sizes to not trigger a persist operation - over the last 12h we have seen 0 partition size triggers:

image

This might be a red herring, or indicative of a bug causing us to not trigger based on the partition size limit.

alamb:
I am fairly confident that the partition limit is being hit and triggering persistence as seen in this screenshot:

Screen Shot 2022-06-24 at 8 41 15 AM

alamb:
Summary here:
* After merging https://github.com/influxdata/influxdb_iox/issues/4936 the overall pause mechanism now appears to be working as expected and will make it easier to protect the ingesters from OOMs 
* I still believe that this PR would further protect the ingester from OOMs (by slowing down ingest)
* As @domodwyer  has pointed out, merging this PR ay slow down recovery after ingester restart, which already seems to be an issue, see https://github.com/influxdata/conductor/issues/971
* This PR makes complicated code more complicated

Thus, my plan is to hold off on merging this PR unless we get into another situation where the pause limit is not sufficient to stop the ingester from OOOming.

alamb:
We recently saw two OOMs when replaying a significant historical backlog from kafka (9 hours of ingest) even with conservative settings (20G pause limit with 55GB available). Details on https://github.com/influxdata/conductor/issues/1006#issuecomment-1172526513

alamb:
While I think this PR improves the ingesters behavior under replay we are lowering the priority of improving the ingestion at the moment and so I am closing this PR for the time being. see https://github.com/influxdata/influxdb_iox/issues/4928#issuecomment-1184675245

PR #4933, opened 2022-06-23
refactor: remove unused errors

domodwyer:
These don't get identified by the unused lint, I assume because of the macro generated magic code.

---

* refactor: remove unused errors (87af3848d)

      These errors are not referenced, but are hidden from the "unused" lint because
      of the macro magic code generation.

### DIFF
--- a/ingester/src/stream_handler/sink_instrumentation.rs
+++ b/ingester/src/stream_handler/sink_instrumentation.rs
@@ -411,11 +411,11 @@ mod tests {
         let got = test(
             op,
             &metrics,
-            Err(crate::data::Error::PersistingEmpty),
+            Err(crate::data::Error::TableNotPresent),
             Some(12345),
         )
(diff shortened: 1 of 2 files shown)


ingester/src/stream_handler/sink_instrumentation.rs
```diff
@@ -411,11 +411,11 @@ mod tests {
         let got = test(
             op,
             &metrics,
-            Err(crate::data::Error::PersistingEmpty),
+            Err(crate::data::Error::TableNotPresent),
```
domodwyer, line 414: This is a test that asserts an inner DML error is propagated out of the instrumentation error, so the actual error value doesn't make any meaningful difference.

shepmaster:
> I assume because of the macro generated magic code.

Yep. https://github.com/shepmaster/snafu/issues/238

PR #4918, opened 2022-06-21
feat: Compact now can split compacted results into multiple non-overlapped files based on config max file size

NGA-TRAN:
Closes #4591 #4790
This  PR includes
- [x] Teach `split_the_stream` to know how to split data into n streams using n-1 input PhysicalExprs
- [x] Make `StreamSplitNode` support a list of expressions
- [x] Teach `compute_split_time` to split a time range into many contiguous ranges
- [x] split compacted result into multiple non-overlapped files based on the config comapction_max_size_bytes

And others
- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #4935, opened 2022-06-23
refactor: do not use in-parquet IOx metadata in compactor (*)

crepererum:
This reduces the usage of the binary metadata to a single point: the creation of the chunks for the query engine. I'll tackle that next.

Ref #4124.


compactor/src/utils.rs
```diff
@@ -142,12 +142,6 @@ impl ParquetFileWithTombstone {
             partition_sort_key,
         )
     }
-
-    /// Return iox metadata of the parquet file
```
crepererum, line 146: This was the main goal of this PR.

alamb: 🔥

compactor/src/compact.rs
```diff
@@ -1100,20 +1164,34 @@ pub struct PartitionCompactionCandidate {
     pub oldest_file: Timestamp,
 }
 
+/// [`PartitionCompactionCandidate`] with some information about its table and namespace.
```
alamb, line 1167:
```suggestion
/// [`PartitionCompactionCandidate`] with information from the catalog
/// about its table and namespace.
```

compactor/src/compact.rs
```diff
@@ -1128,14 +1206,17 @@ mod tests {
         let catalog = TestCatalog::new();
 
         let lp = vec![
-            "table,tag1=WA field_int=1000 8000",
-            "table,tag1=VT field_int=10 10000",
-            "table,tag1=UT field_int=70 20000",
+            "table,tag1=WA field_int=1000i 8000",
```
alamb, line 1209: Any particular reason not to use `1000` and `ColumnType::U64` rather than switch to `I64`?

alamb (approved):
The code changes made sense to me -- though I think @NGA-TRAN  should also review this prior to merge as it touches the compactor

Note it may conflict with https://github.com/influxdata/influxdb_iox/pull/4918 from @NGA-TRAN as well

compactor/src/compact.rs
crepererum: `field_int=1000` produces float, not a signed or unsigned integer. I personally find this a bit confusing as well, but is indeed how the [official docs](https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_tutorial/#data-types) specify it. These docs don't even mention unsigned integers. So I felt there are two options:

- renaming the field (because `field_int` clearly means "integer field")
- fixing the payload type

I felt that the 2nd option has less fallout, because otherwise I would also need to adjust the record-batch tests (the ones that use string representations to compare arrow batches against a baseline). But I can go for the other option if that's desirable.

alamb: this PR is fine -- I was just curious@

compactor/src/compact.rs
```diff
@@ -628,13 +693,10 @@ impl Compactor {
...
-        // Save the parquet metadata for the first file to reuse IDs and names
-        let iox_metadata = overlapped_files[0].iox_metadata();
-
         // Hold onto various metadata items needed later.
-        let sequencer_id = overlapped_files[0].sequencer_id;
-        let namespace_id = overlapped_files[0].namespace_id;
-        let table_id = overlapped_files[0].table_id;
+        let sequencer_id = partition.sequencer_id;
+        let namespace_id = table.namespace_id;
+        let table_id = table.id;
```
NGA-TRAN, line 699: 🎉

compactor/src/compact.rs
```diff
@@ -342,15 +359,59 @@ impl Compactor {
...
+                .context(QueryingTableSnafu)?
+                .context(TableNotFoundSnafu { table_id: id })?;
+            tables.insert(id, Arc::new(table));
+        }
+
+        let mut namespaces = HashMap::with_capacity(namespace_ids.len());
+        for id in namespace_ids {
+            let namespace = repos
+                .namespaces()
+                .get_by_id(id)
```
NGA-TRAN, line 387: This is unrelated to this PR but it reminds me we read catalog and store in memory the full row of namespace, table, ... everywhere but we actually only need a few fields of each of them. This is another place to improve in Ingester, Compactor and Querier. Let me open a ticket

NGA-TRAN (approved):
Nice. Thanks @crepererum

PR #4941, opened 2022-06-23
fix: account for partition memory until persist is completed

### LINKED ISSUE #4936: Ingester's "total memory accounting" does not account for memory being persisted
As pointed out by @domodwyer  on https://github.com/influxdata/influxdb_iox/pull/4931#issuecomment-1164402251

The ingester maintains a rough "total memory in use" counter it uses to try and limit the amount of memory the ingester is using overall:

https://github.com/influxdata/influxdb_iox/blob/fe235d0b0e3e2166d50b72c0b2c30146747a0971/ingester/src/lifecycle.rs#L317-L319

When the ingester triggers a persist operation, it immediately stops accounting for the memory the partition is using:

https://github.com/influxdata/influxdb_iox/blob/fe235d0b0e3e2166d50b72c0b2c30146747a0971/ingester/src/lifecycle.rs#L307-L310

Which is called in `mark_persisting()` at the start of the persist [here][1], while it is very much still using the memory (and soon more as the parquet is generated). This would cause the ingester to falsely believe it has memory headroom, and it will continue ingesting, pushing memory usage over the limit.

The ingester should move this decrement to be _after_ the persist is complete so it accounts for the partition memory until it has been actually freed, which should help the OOMs as we will correctly account for the in-use memory and pause ingest.

[1]: https://github.com/influxdata/influxdb_iox/blob/fe235d0b0e3e2166d50b72c0b2c30146747a0971/ingester/src/lifecycle.rs#L479

domodwyer:
Fixes #4936.

---

* fix: account for partition memory until persisted (9a79d1658)

      The ingester maintains a rough "total memory in use" counter it uses to try
      and limit the amount of memory the ingester is using overall.

      When a partition is persisted, this total memory usage value is adjusted to
      account for releasing the partition memory. Prior to this commit, the ordering
      was:

      * Writes increase the memory counter
      * maybe_persist() is called to trigger persistence
      * A partition is identified for persistence
      * Partition memory usage is released back to the total memory counter
      * Persistence starts

      This meant that the partitions in the process of being persisted were not
      accounted for in the ingester's total memory counter, and therefore we could
      significantly overrun the configured memory limit.

      After this commit, the ordering is:

      * Writes increase the memory counter
      * maybe_persist() is called to trigger persistence
      * A partition is identified for persistence
      * Persistence starts
      * Persistence completes
      * Partition memory usage is released back to the total memory counter

      This ensures persisting partitions are sill tracked in the total memory 
      counter, causing pauses to correctly fire.

### DIFF
--- a/ingester/src/lifecycle.rs
+++ b/ingester/src/lifecycle.rs
@@ -180,10 +180,7 @@ struct LifecycleState {
 
 impl LifecycleState {
     fn remove(&mut self, partition_id: &PartitionId) -> Option<PartitionLifecycleStats> {
-        self.partition_stats.remove(partition_id).map(|stats| {
-            self.total_bytes -= stats.bytes_written;
-            stats
-        })
+        self.partition_stats.remove(partition_id)
     }
 }
 
@@ -352,14 +349,25 @@ impl LifecycleManager {
         let persist_tasks: Vec<_> = to_persist
             .into_iter()
             .map(|s| {
-                self.remove(s.partition_id);
+                // Mark this partition as being persisted, and remember the
+                // memory allocation it had accumulated.
+                let partition_memory_usage = self
+                    .remove(s.partition_id)
+                    .map(|s| s.bytes_written)
+                    .unwrap_or_default();
                 let persister = Arc::clone(persister);
 
                 let (_tracker, registration) = self.job_registry.register(Job::Persist {
                     partition_id: s.partition_id,
                 });
+
+                let state = Arc::clone(&self.state);
                 tokio::task::spawn(async move {
                     persister.persist(s.partition_id).await;
+                    // Now the data has been uploaded and the memory it was
+                    // using has been freed, released the memory capacity back
+                    // the ingester.
+                    state.lock().total_bytes -= partition_memory_usage;
                 })
                 .track(registration)
             })
@@ -459,6 +467,7 @@ mod tests {
     use iox_time::MockProvider;
     use metric::{Attributes, Registry};
     use std::collections::BTreeSet;
+    use test_helpers::timeout::FutureTimeout;
 
     #[derive(Default)]
     struct TestPersister {
@@ -537,8 +546,8 @@ mod tests {
         assert_eq!(p2.first_write, Time::from_timestamp_nanos(10));
     }
 
-    #[test]
-    fn pausing_and_resuming_ingest() {
+    #[tokio::test]
+    async fn pausing_and_resuming_ingest() {
         let config = LifecycleConfig {
             pause_ingest_size: 20,
             persist_memory_threshold: 10,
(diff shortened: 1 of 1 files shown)


ingester/src/lifecycle.rs
```diff
@@ -352,14 +349,25 @@ impl LifecycleManager {
...
                     partition_id: s.partition_id,
                 });
+
+                let state = Arc::clone(&self.state);
                 tokio::task::spawn(async move {
                     persister.persist(s.partition_id).await;
+                    // Now the data has been uploaded and the memory it was
+                    // using has been freed, released the memory capacity back
+                    // the ingester.
+                    state.lock().total_bytes -= partition_memory_usage;
```
alamb, line 370: Given how error handling works in the ingester (fail fast!), I can't think of any way we might lose track of `partition_memory_usage` but I just wanted to bring it up

ingester/src/lifecycle.rs
```diff
@@ -537,38 +546,55 @@ mod tests {
...
         assert!(!h.can_resume_ingest());
 
-        m.remove(PartitionId::new(1));
-        assert!(h.can_resume_ingest());
-        assert!(!h.log_write(PartitionId::new(1), sequencer_id, SequenceNumber::new(3), 3));
+        // Trigger a persistence job that succeeds.
+        let persister = Arc::new(TestPersister::default());
+        m.maybe_persist(&persister).await;
+
+        // Wait for the persist op to complete
```
alamb, line 574: I don't see how this test change covers the logic change. I am probably missing something

I think the key thing to assert is that while the persist is outstanding, subsequent calls to `h.log_write` that *will* be accepted when the persist is complete won't be accepted while persist is outstanding

In fact I backed out the code changes in this PR and the test still passes:

```diff
diff --git a/ingester/src/lifecycle.rs b/ingester/src/lifecycle.rs
index 18f5ead6d..9679f7d73 100644
--- a/ingester/src/lifecycle.rs
+++ b/ingester/src/lifecycle.rs
@@ -180,7 +180,10 @@ struct LifecycleState {
 
 impl LifecycleState {
     fn remove(&mut self, partition_id: &PartitionId) -> Option<PartitionLifecycleStats> {
-        self.partition_stats.remove(partition_id)
+        self.partition_stats.remove(partition_id).map(|stats| {
+            self.total_bytes -= stats.bytes_written;
+            stats
+        })
     }
 }
 
@@ -349,25 +352,14 @@ impl LifecycleManager {
         let persist_tasks: Vec<_> = to_persist
             .into_iter()
             .map(|s| {
-                // Mark this partition as being persisted, and remember the
-                // memory allocation it had accumulated.
-                let partition_memory_usage = self
-                    .remove(s.partition_id)
-                    .map(|s| s.bytes_written)
-                    .unwrap_or_default();
+                self.remove(s.partition_id);
                 let persister = Arc::clone(persister);
 
                 let (_tracker, registration) = self.job_registry.register(Job::Persist {
                     partition_id: s.partition_id,
                 });
-
-                let state = Arc::clone(&self.state);
                 tokio::task::spawn(async move {
                     persister.persist(s.partition_id).await;
-                    // Now the data has been uploaded and the memory it was
-                    // using has been freed, released the memory capacity back
-                    // the ingester.
-                    state.lock().total_bytes -= partition_memory_usage;
                 })
                 .track(registration)
             })
```

```
running 1 test
test lifecycle::tests::pausing_and_resuming_ingest ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 71 filtered out; finished in 0.00s

   Doc-tests ingester

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

The `PausiblePersister` from https://github.com/influxdata/influxdb_iox/pull/4931 may be helpful for writing this kind of test

alamb (commented):
The code looks good to me -- I have a question about the test

alamb:
Update: I'll fix the tests as @domodwyer  is currently tied up

ingester/src/lifecycle.rs
```diff
@@ -537,38 +631,80 @@ mod tests {
...
+        let persister = Arc::new(TestPersister::default());
+        m.maybe_persist(&persister).await;
+
+        // ingest can resume
+        assert!(h.can_resume_ingest());
+        assert!(!h.log_write(partition_id, sequencer_id, SequenceNumber::new(3), 3));
+    }
+
+    #[tokio::test]
+    async fn pausing_ingest_waits_until_persist_completes() {
```
alamb, line 665: Without the code changes in this PR, this new test fails like:

```
---- lifecycle::tests::pausing_ingest_waits_until_persist_completes stdout ----
2022-06-23T15:51:54.264419Z  INFO ingester::lifecycle: Partition is over size threshold, persisting sequencer_id=1 partition_id=1 bytes_written=25 partition_size_threshold=5
thread 'lifecycle::tests::pausing_ingest_waits_until_persist_completes' panicked at 'assertion failed: !h.can_resume_ingest()', ingester/src/lifecycle.rs:691:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
```

alamb:
I am merging this one and will begin seeing what happens as we allow ingesters to accumulate more in memory

alamb (approved):
YOLO approving a PR I half wrote ;)

PR #4948, opened 2022-06-27
refactor: remove `ParquetFileWithMetadata` usage outside the catalog

crepererum:
On top of #4938.

Next and final step will be to remove the metadata from the catalog itself.

Ref #4124.


NGA-TRAN (approved):
Very nice. Thanks @crepererum

PR #4964, opened 2022-06-28
fix: parquet reader sort order

crepererum:
I really dislike the parquet-rs interface.


tustvold (approved):
I think this code would be less confusing if it made use of the ProjectionMask exposed by the parquet crate, instead of `Vec<usize>`.

FWIW the reason it is a mask is because of nested types, DataFusion may switch at some point also - https://github.com/apache/arrow-datafusion/issues/2581

(also also when we switch to using DataFusion for parquet access this problem will go away, as it already handles schema adaption)

PR #4965, opened 2022-06-28
refactor: use IDs for `parquet_file.column_set`

### LINKED ISSUE #4959: Use IDs instead of strings for catalog `parquet_file.column_set`
We currently use a list of strings to store the per-parquet-file column set. Since this is the larges table in prod, it may be worth to consider moving that to column IDs (primary key of the `column_name` table).

crepererum:
Closes #4959.


aierui:
👍

compactor/src/utils.rs
```diff
@@ -134,10 +135,23 @@ impl ParquetFileWithTombstone {
...
-        let selection: Vec<_> = self.column_set.iter().map(|s| s.as_str()).collect();
+        let column_id_lookup: HashMap<ColumnId, &str> = table_schema
+            .columns
+            .iter()
+            .map(|(name, c)| (c.id, name.as_str()))
+            .collect();
+        let selection: Vec<_> = self
+            .column_set
+            .iter()
+            .flat_map(|id| column_id_lookup.get(id).copied())
```
carols10cents, line 149: This looks like it's ignoring any columns in the column set that aren't in the schema... I'm trying to think of how that would happen and can't really think of anything beyond catalog corruption... would we want to log, perhaps, if `selection.len() != self.column_set.len()`? Or is that overly defensive?

(if anything is done here, it can be done in a future PR)

iox_tests/src/util.rs
```diff
@@ -748,13 +750,19 @@ impl TestParquetFile {
 
     /// Get Parquet file schema.
     pub async fn schema(&self) -> Arc<Schema> {
-        let table_schema = self.table.schema().await;
+        let table_schema = self.table.catalog_schema().await;
+        let column_id_lookup: HashMap<ColumnId, &str> = table_schema
```
carols10cents, line 754: Given this building of the reverse index happens in a few places, maybe it's worth being a method or field on `TableSchema`? 

Just an idea for a future refactoring potentially, not blocking this PR

carols10cents (approved):
Looks good! A few future suggestions, but nothing blocking.

compactor/src/utils.rs
crepererum: overly defensive, just in case that we ever wanna implement column deletion.

iox_tests/src/util.rs
crepererum: done

PR #4966, opened 2022-06-28
fix: ensure panic is propagated throw `AdapterStream`

crepererum:
Prior to this change background tasks that we feed into `AdapterStream`
can panic but that would just end the stream without any user-visible
error (except for the panic message on stdout/stderr).


crepererum:
Looking at the code again, I think there's a race condition. I'll need a different approach.

PR #4967, opened 2022-06-28
chore: unsuppress RUSTSEC-2020-0159, which was withdrawn

jacobmarble:
Helps #2884


jacobmarble:
@domodwyer how do I test this locally? Should CI pick up an unsuppressed deny rule?

domodwyer:
You should install [cargo-deny][deny] locally if you want to test it:

[deny]:https://github.com/EmbarkStudios/cargo-deny

```console
$ cargo install --locked --force cargo-deny
```

Then you can run:

```console
$ cargo deny check -s

<snip>

 advisories ok: 0 errors, 2 warnings, 3 notes
       bans ok: 0 errors, 27 warnings, 0 notes
   licenses ok: 0 errors, 0 warnings, 515 notes
    sources ok: 0 errors, 0 warnings, 10 notes
```

No advisory errors means it is OK :+1:

The two advisory warnings are because of this line: 

https://github.com/influxdata/influxdb_iox/blob/b8e7071f8229ec729369816430ac5c23d46b2f2e/deny.toml#L25

Which shouldn't be there - I dunno why it is, but it should be removed - it's not an advisory ID, so it generates warnings saying as much:

```
warning[A008]: advisory not found in any advisory database
   ┌─ /Users/dom/Documents/rust/influxdb_iox/deny.toml:25:5
   │
25 │     "query",
   │     ^^^^^^^ unknown advisory

warning[A007]: advisory was not encountered
   ┌─ /Users/dom/Documents/rust/influxdb_iox/deny.toml:25:5
   │
25 │     "query",
   │     ^^^^^^^ no crate matched advisory criteria

```

And to answer the other question, yes CI would fail - the [cargo_audit](https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/20831/workflows/c987ab04-a063-4c15-b5d7-1cfdeda87e42/jobs/178362) step would fail if there was an advisory for any of our dependencies that has not been whitelisted. As you've removed one, and it didn't fail, that means we're no longer using affected code 🎉

domodwyer (approved):

PR #4968, opened 2022-06-28
fix: Split overlapped files based on the order of sequence numbers and only group non-overlapped contigous small files

### LINKED ISSUE #4963: Compactor: Split and group contiguous time-range files must meet contiguous sequence numbers
# Rationale

This functions [split_overlapped_groups](https://github.com/influxdata/influxdb_iox/blob/106d84c6e1808eabd52f4446f65f556298981f98/compactor/src/compact.rs#L938) and [group_small_contiguous_groups](https://github.com/influxdata/influxdb_iox/blob/106d84c6e1808eabd52f4446f65f556298981f98/compactor/src/compact.rs#L654) most consider contiguous time ranges. Sequence number ranges must be considered as well to avoid wrong deduplication

NGA-TRAN:
Closes #4963

This PR includes 2 fixes:
1. If a group of overlapped file was too large, we would split them to smaller groups using `min_time` order which would produce wrong result in deduplication (thanks to @Joe-Blount who pointed that out). The fix is to sort the chunks using `min_sequence_number` instead.
2. If non-overlapped contiguous files were small, we would group them before compacting. The intention was to group non-overlapped groups only (because of the same sequence number/deduplication issue above)  but due to the change that split a large overlapped group further, we did group overlapped groups. This fix is to avoid doing so. 

The fix is not that complicated but the tests are a bit tricky. I will point to the code of each fix


compactor/src/compact.rs
```diff
@@ -933,21 +934,78 @@ impl Compactor {
...
         for group in groups {
             let total_size_bytes: i64 = group.iter().map(|f| f.file_size_bytes).sum();
-            if group.len() <= max_count && total_size_bytes <= max_size_bytes {
-                overlapped_groups.push(group.to_vec());
+            if group.len() == 1 || (group.len() <= max_count && total_size_bytes <= max_size_bytes)
+            {
+                overlapped_groups.add_internal_overalapped_group(group.to_vec());
             } else {
-                group.sort_by_key(|f| f.min_time);
+                group.sort_by_key(|f| f.min_sequence_number);
```
NGA-TRAN, line 1007: Here is the first fix to sort the chunk by sequence number

compactor/src/compact.rs
```diff
@@ -668,7 +668,8 @@ impl Compactor {
         let mut current_size = 0;
         let mut current_num_files = 0;
         for g in file_groups {
-            if current_size + g.total_file_size_bytes < compaction_max_size_bytes
+            if !g.overlapped_with_other_groups // only group non-overlapped ones
```
NGA-TRAN, line 671: Here is the second fix that do not group further groups that overlap with other groups

compactor/src/compact.rs
pauldix: Any reason to use min sequence number here rather than max?

compactor/src/compact.rs
```diff
@@ -933,21 +934,78 @@ impl Compactor {
...
+    // Before splitting, the files are sorted on their sequence numbers to guarantee
+    // the deduplication is still correct
+    //
+    // Example:
+    // The input `groups` consists of 4 groups of 9 chunks with time ranges as below.
+    // The chunks are named using their sequence number. `C4` means this chunk inlcudes
+    // data of sequence numner 4 (Note: a chunk can include data of a range of sequence numbers
+    // but to simplify the example, we make chunk contain only one sequence number)
+    //   . G1: includes one chunk C4
+    //   . G2: cinlcudes 4 overlapped C1, C2, C3, C6
```
domodwyer, line 947:
```suggestion
    //   . G2: inlcudes 4 overlapped C1, C2, C3, C6
```

compactor/src/compact.rs
domodwyer: It might be worth sorting the array by min (or max) and then linearly walking the array to assert the max sequence number always increases, and the previous max is always less-than the next group's min. That would assert correct ordering, alert us to the case of "gaps" having been compacted and should be relatively fast (`O(n)` + CPU cache friendly).

This check might have alerted us to the problems being solved in this PR if we came to compact two previously compacted chunks, and discovered they contained overlapping sequence ranges.

compactor/src/compact.rs
```diff
@@ -1231,6 +1270,48 @@ impl Compactor {
...
+
+        for group in &self.internal_overlapped_groups {
+            groups.push(GroupWithMinTimeAndSize::new(group.to_vec(), false));
+        }
+        for group in &self.external_overlapped_groups {
+            groups.push(GroupWithMinTimeAndSize::new(group.to_vec(), true));
+        }
+
+        groups
+    }
```
domodwyer, line 1312: Calling `to_vec()` will clone all the parquet files (which could be a lot!).

We never re-use this `OverlappedGroups` after we have called `groups_with_min_time_and_size()` so we can _move_ the vecs, and not have to clone them. 

We do this by changing the fn sig to take `self` (not `&mut self`) and moving the vecs:

```suggestion
    pub fn groups_with_min_time_and_size(self) -> Vec<GroupWithMinTimeAndSize> {
        let mut groups = Vec::with_capacity(
            self.internal_overlapped_groups.len() + self.external_overlapped_groups.len(),
        );

        for group in self.internal_overlapped_groups {
            groups.push(GroupWithMinTimeAndSize::new(group, false));
        }
        for group in self.external_overlapped_groups {
            groups.push(GroupWithMinTimeAndSize::new(group, true));
        }

        groups
    }
```

compactor/src/compact.rs
```diff
@@ -933,21 +934,78 @@ impl Compactor {
...
-    ) -> Vec<Vec<ParquetFile>> {
-        let mut overlapped_groups: Vec<Vec<ParquetFile>> = Vec::with_capacity(groups.len() * 2);
+    ) -> OverlappedGroups {
+        let mut overlapped_groups = OverlappedGroups::new(groups.len() * 2);
         let max_count = max_file_count.try_into().unwrap();
         for group in groups {
             let total_size_bytes: i64 = group.iter().map(|f| f.file_size_bytes).sum();
-            if group.len() <= max_count && total_size_bytes <= max_size_bytes {
-                overlapped_groups.push(group.to_vec());
+            if group.len() == 1 || (group.len() <= max_count && total_size_bytes <= max_size_bytes)
```
domodwyer, line 1003: In this check we will process a single file that exceeds the `total_size_bytes` - is that the desired behaviour? It could lead to a crash/OOM loop but the alternative is to ignore the file forever. At least a crash would alert us to the issue.

compactor/src/utils.rs
```diff
@@ -45,6 +45,28 @@ pub struct GroupWithMinTimeAndSize {
...
+impl GroupWithMinTimeAndSize {
+    /// Make GroupWithMinTimeAndSize for a given set of parquet files
+    pub fn new(files: Vec<ParquetFile>, overlaped: bool) -> Self {
+        let mut group = Self {
+            parquet_files: files,
+            min_time: Timestamp::new(i64::MAX),
+            total_file_size_bytes: 0,
+            overlapped_with_other_groups: overlaped,
+        };
+
```
domodwyer, line 62: Do you want to make sure this always has at least one file?

```suggestion

assert!(!files.is_empty(), "invalid empty group")
```

Otherwise calling `GroupWithMinTimeAndSize::new(vec![], false)` will be accepted and will result in a `min_time` of `i64::MAX` being observable to the caller.

domodwyer (commented):
The complexity of this split/group/selection code still terrifies me a little 😅 I hope to see if we can refactor it one day!

---

I'm a little concerned about the `list_by_partition_not_to_delete()` call that produces the set of files this PR groups - it [returns an unbounded number of results][query] and is already the [most expensive query we run][report] against the catalog. In the medium term, we will have to change it / limit results before it starts causing issues, but that's obviously not in-scope for this PR. I created https://github.com/influxdata/influxdb_iox/issues/4981 to track :+1:

I mention it because my understanding is that this grouping solution assumes it always has the full set of all parquet files for the partition, therefore sorting by sequence number will always derive a total ordering per partition. If we have to change the query to `LIMIT 1000` or some such, this assumption will not hold true as we may have "gaps" in the files - something to keep an eye on with any potential changes made to the query @NGA-TRAN.

[report]:https://gist.github.com/domodwyer/c4232143467f8f0fa0e11b081eae2b01#most-expensive
[query]:https://github.com/influxdata/influxdb_iox/blob/8dcaec30275cfa6d10bdb6d944b26a9b3a1e9360/iox_catalog/src/postgres.rs#L1673-L1678

compactor/src/compact.rs
```diff
@@ -933,21 +934,78 @@ impl Compactor {
     }
 
     // Split overlapped groups into smaller groups if there are so mnay files in each group or
-    // their size are too large. The files are sorted by their min time before splitting so files
-    // are guarannteed to be overlapped in each new group
+    // their size are too large.
+    // Before splitting, the files are sorted on their sequence numbers to guarantee
+    // the deduplication is still correct
+    //
```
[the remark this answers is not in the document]
NGA-TRAN: Added

compactor/src/compact.rs
NGA-TRAN: Open #4984

NGA-TRAN: And in #4984 I explain the situation of one super large file will be unlikely because we limit persisted file size. After all related tickets are closed, the fix for this is simply panic when we hit one super large file

Issue #1231, opened 2021-04-16
Speed up compilation by reducing the number of Rust dependencies


e-dard:
There is nothing specifically actionable from this ticket until someone goes and pokes around a bit and figures out what we could do to help reduce our dependencies.

### Problem

I'm already starting to see huge number of dependencies being pulled into many different crates when building at the crate or project level. Rust is slow enough to compile as it is, but we can control our dependencies to try and speed things up that way.

### Initial Steps

I'm open to ideas but maybe:

- [ ] determine how many dependencies we have and where they stem from (`cargo tree | wc -l` gives me 1233 lines).
- [ ] look for any dependencies that we are using little of, but are accounting for a lot of the bloat? We could just not use those and instead implement the dependency ourselves (clients libraries and SDKs can be a good example of these, e.g., aws/gcp/azure object store clients).
- [ ] can we reduce any cross-crate dependencies by maybe just replicating types from one crate in another and implementing `From` where appropriate to convert from one to the other?

alamb:
I think the `prost` / protobuf handling libraries are an obvious place to target to slim down dependencies as they bring in a lot

We could also make the arrow dependencies more fine grained (like maybe make `arrow_deps` `arrow_flight_deps` and `datafusion_deps`) as arrow has significantly fewer dependencies than datafusion or flight

crepererum:
I think the issue is that we have too many different dependencies, rather than multiple versions of them (or at least the latter one isn't a massive issue). For [`7b67ae498bc5f89306cefc9946a9363759dbefa6`](https://github.com/influxdata/influxdb_iox/tree/7b67ae498bc5f89306cefc9946a9363759dbefa6)

```shell
cargo tree --workspace | sed -e 's/^[^a-zA-Z0-9_-]*//' -e 's/\s*([^)]*)\s*//g' -e '/^$/d' | sort -u
```

we get the following version list (which has 372 entries):

<details>

```text
addr2line v0.14.1
adler v1.0.2
ahash v0.4.7
ahash v0.7.2
aho-corasick v0.7.15
alloc-no-stdlib v2.0.1
alloc-stdlib v0.2.1
ansi_term v0.11.0
ansi_term v0.12.1
anyhow v1.0.40
arrayvec v0.5.2
arrow_deps v0.1.0
arrow-flight v4.0.0-SNAPSHOT
arrow v4.0.0-SNAPSHOT
assert_cmd v1.0.3
assert-json-diff v1.1.0
[...]

PR #4979, opened 2022-06-29
refactor: Replace all uses of lazy_static with once_cell

Marwes:
Went through and replaced all lazy_static uses with once_cell (while waiting for the project to compile). There are still dependencies using lazy_static so it is still in the crate graph but at least there isn't an explicit dependency on it (and it is easier to update to `std::lazy::Lazy` once that is stable).

cc https://github.com/influxdata/influxdb_iox/issues/1231


crepererum (approved):
Thank you! :)

PR #4980, opened 2022-06-29
fix: ensure panic is propagated through `AdapterStream`

crepererum:
Prior to this change background tasks that we feed into `AdapterStream`
can panic but that would just end the stream without any user-visible
error (except for the panic message on stdout/stderr).

This was found while developing #4964. I have proposed another fix in #4966
but found that I actually developed an existing solution a 2nd time:
`watch_task`. But I also see a major issue with the existing API: one
can create `AdapterStream` with ordinary tokio tasks that are not
watched at all, leaving the burden to the implementor to check for that
(and actually we forgot that in `parquet_file`).

So this change takes a slightly different approach:
The `AdapterStream` does NOT accept ordinary join handles any longer but
requires that you pass a "watched task". The newly introduced
`WatchedTask` does the same as we did manually before: wrapping a future
into a tokio task, watch it and wrap the watcher into a task.

It is now way more difficult to do anything stupid (sure you can still
mix up the tasks and the channels, but we need at least some flexibility
here to allow for "split" and potential future fan-in/out constructs).


alamb:
I will review this shortly

datafusion_util/src/lib.rs
```diff
@@ -390,4 +364,24 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_adapter_stream_panic_handling() {
+        let schema = SchemaBuilder::new().timestamp().build().unwrap().as_arrow();
+        let (tx, rx) = tokio::sync::mpsc::channel(2);
+        let tx_captured = tx.clone();
+        let fut = async move {
+            let _tx = tx_captured;
+            if true {
+                panic!("epic fail");
```
alamb, line 376: 😆

datafusion_util/src/sender.rs
```diff
@@ -0,0 +1,33 @@
+use async_trait::async_trait;
+use tokio::sync::mpsc::{error::SendError, Sender, UnboundedSender};
+
+#[async_trait]
+pub trait AbstractSender: Clone + Send + Sync + 'static {
```
alamb, line 5: I think some docstrings explaining what an AbstractSender is (and what it should be used for) would be helpful

iox_query/src/exec/split.rs
```diff
@@ -310,47 +309,16 @@ impl StreamSplitExec {
...
-                    }
-                }
-                // Input task completed successfully
-                Ok(Ok(())) => {
-                    trace!("All input tasks completed successfully");
-                }
-            }
-        });
-        let handle = Arc::new(AutoAbortJoinHandle::new(handle));
+        let fut = split_the_stream(input_stream, split_exprs, txs.clone(), baseline_metrics);
```
alamb, line 312: That is certainly nicer 👍

parquet_file/src/storage.rs
```diff
@@ -192,27 +192,28 @@ impl ParquetStorage {
...
+                if let Err(e) = tx_captured.send(ArrowResult::Err(e)).await {
                     // if no one is listening, there is no one else to hear our screams
                     debug!(%e, "Error sending result of download function. Receiver is closed.");
                 }
             }
-        });
+
+            Ok(())
+        };
+        let handle = WatchedTask::new(fut, vec![tx], "download and scan parquet");
```
alamb, line 213: This is an actual bug fix, right?

alamb (approved):
👍 Thank you -- this is really nice

parquet_file/src/storage.rs
crepererum: Correct, this is the fix that is now enforced by the type system.

datafusion_util/src/sender.rs
crepererum: done

Issue #4827, opened 2022-06-10
Ensure "All time" metadata queries are appropriately special cased


alamb:

domodwyer:
I spoke to @wolffcm earlier this week and we came to the conclusion we can use the schema API within the routers to satisfy this instead of a query 🎉 

I will polish up the API and implement it in the Go client in the next week or two :+1:

alamb:
To be clear I think we also need to keep supporting the existing `influxrpc` APIs as well (which can contain arbitrary predicates)

pauldix:
@crepererum Probably best to just use whatever now is.

pauldix:
@crepererum I'm saying whatever system time.now returns, nothing related to the parquet files.

alamb:
FWIW there is already some special case handling for max time (introduced in https://github.com/influxdata/influxdb_iox/issues/3439). Tests are here:

https://github.com/influxdata/influxdb_iox/blob/bc107aff1d25f2ad52580e4013dff2612edc7da5/query_tests/src/scenarios/library.rs#L13-L28

But basically the code simply removes the timestamp predicate, rather than special casing such predicates to be schema only.

lesam:
Would a better option here be:

* Check postgres metadata for each chunk/parquet file - if metadata has 'enough' for us to compute tag keys (i.e. the time range completely covers the partition / is missing, and there is no other restrictions on the predicate) we take the tags from the metadata
* Otherwise, pull the chunks and take the tags from the chunks.

Regardless, there will be a behaviour change: currently, we never delete tags (or any columns) from the metadata. E.g. this parquet file:

```
row group 1: RC:1 TS:406 OFFSET:4
--------------------------------------------------------------------------------
host:         BINARY ZSTD DO:4 FPO:39 SZ:92/74/0.80 VC:1 ENC:RLE,PLAIN,RLE_DICTIONARY ST:[min: server01, max: server01, num_nulls not defined]
in:           DOUBLE ZSTD DO:152 FPO:183 SZ:88/70/0.80 VC:1 ENC:RLE,PLAIN,RLE_DICTIONARY ST:[min: 3.0, max: 3.0, num_nulls not defined]
name:         BINARY ZSTD DO:296 FPO:319 SZ:56/38/0.68 VC:1 ENC:RLE,PLAIN,RLE_DICTIONARY ST:[num_nulls: 1, min/max not defined]
[...]

PR #4985, opened 2022-06-29
fix: optimize tag_keys to go only to schema when predicate is empty

lesam:
Helps #4827 

When the predicate is empty (note `_measurement = "foo"` and `time range all time` predicates are simplified away), satisfy tag_keys queries from the metadata store instead of reading all the parquet data.

This means deletes are not honoured for this query.

### DIFF
--- a/iox_query/src/frontend/influxrpc.rs
+++ b/iox_query/src/frontend/influxrpc.rs
@@ -49,6 +49,9 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
+    #[snafu(display("gRPC planner table {} not found", table_name))]
+    GettingTableSchema { table_name: String },
+
     #[snafu(display("gRPC planner got error finding column values: {}", source))]
     FindingColumnValues {
         source: Box<dyn std::error::Error + Send + Sync>,
@@ -345,6 +348,19 @@ impl InfluxRpcPlanner {
             .table_predicates(database.as_meta())
             .context(CreatingPredicatesSnafu)?;
         for (table_name, predicate) in &table_predicates {
+            if predicate.is_empty() {
+                // special case - return the columns from metadata only.
+                // Note that columns with all rows deleted will still show here
+                known_columns.extend(
+                    database
+                        .table_schema(table_name)
+                        .context(GettingTableSchemaSnafu { table_name })?
+                        .tags_iter()
+                        .map(|f| f.name().clone()),
+                );
+                continue;
+            }
+
             let chunks = database
                 .chunks(table_name, predicate)
                 .await
--- a/predicate/src/rpc_predicate.rs
+++ b/predicate/src/rpc_predicate.rs
@@ -6,6 +6,7 @@ mod value_rewrite;
 
 use crate::{rewrite, Predicate};
 
+use datafusion::common::ScalarValue;
 use datafusion::error::{DataFusionError, Result as DataFusionResult};
 use datafusion::execution::context::ExecutionProps;
 use datafusion::logical_plan::{
@@ -198,6 +199,11 @@ fn normalize_predicate(
                 })
                 .and_then(rewrite::simplify_predicate)
         })
+        // Filter out literal true so is_empty works correctly
+        .filter(|f| match f {
+            Err(_) => true,
+            Ok(expr) => !expr.eq(&Expr::Literal(ScalarValue::Boolean(Option::Some(true)))),
+        })
         .collect::<DataFusionResult<Vec<_>>>()?;
 
     // Store any field value (`_value`) expressions on the `Predicate`.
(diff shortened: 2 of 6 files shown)


iox_query/src/frontend/influxrpc.rs
```diff
@@ -350,10 +350,6 @@ impl InfluxRpcPlanner {
                 .await
                 .context(GettingChunksSnafu { table_name })?;
             for chunk in chunks {
-                // If there are delete predicates, we need to scan (or do full plan) the data to eliminate
-                // deleted data before getting tag keys
-                let mut do_full_plan = chunk.has_delete_predicates();
```
lesam, line 355: TODO: instead of just removing delete predicates, do a full plan only when the following are true:
* Our predicate can be satisfied only from metadata (column names, not column values)
* There is no time range bound

We also, of course, need to avoid fetching the full chunk when just the metadata will do.

query_tests/src/influxrpc/tag_keys.rs
```diff
@@ -175,8 +188,21 @@ async fn list_tag_name_end_to_end_with_delete() {
         .with_range(0, 10000)
         .with_expr(col("host").eq(lit("server01")));
     let predicate = InfluxRpcPredicate::new(None, predicate);
-    let expected_tag_keys = vec!["host", "region"];
-    run_tag_keys_test_case(EndToEndTestWithDelete {}, predicate, expected_tag_keys).await;
+    // We now ignore deletes for tag keys - if the tag is in the metadata, we return it.
+    // However the deletes would already be applied by the ingester, so two cases are necessary.
```
crepererum, line 192: If the delete was applied the ingester, the column also exists. It's just that none of the chunks contains the column. So the changed behavior is kinda inconsistent. Esp. the "place" (e.g. ingester, parquet) of the data should be invisible to the user.

lesam: I'll have to look at the ingester in more detail, but my assumption was that if the delete is fully applied in memory before the parquet file is persisted, we would never write the column to the metadata store. I'll have to confirm that assumption.

crepererum: > I'll have to look at the ingester in more detail, but my assumption was that if the delete is fully applied in memory before the parquet file is persisted, we would never write the column to the metadata store. I'll have to confirm that assumption.

Columns are created by the router, long before the ingester actually gets the data. The existence of a column does NOT require the existence of any payload row/cell.

lesam: Cool, I was able to see this by poking around in Postgres, thanks!

I've re-worked to be much simpler, the existing tests all pass. Some tests already test the empty predicate case:

```
    influxrpc::tag_keys::list_tag_columns_measurement_name
    influxrpc::tag_keys::list_tag_columns_no_predicate
    influxrpc::tag_keys::list_tag_columns_with_no_tags
    influxrpc::tag_keys::list_tag_name_max_i64
    influxrpc::tag_keys::list_tag_name_max_time
```

service_grpc_influxrpc/src/expr.rs
```diff
@@ -1020,8 +1020,8 @@ mod tests {
             .build();
         let predicate = table_predicate(predicate);
 
-        // predicate is rewritten to true, and projection is added
-        let expected_exprs = vec![lit(true)];
+        // predicate is rewritten to true (which is simplified to an empty predicate list), and projection is added
```
lesam, line 1023: This comment is existing, but I don't see a projection or similar being added here?

query_tests/src/influxrpc/tag_keys.rs
```diff
@@ -179,6 +179,14 @@ async fn list_tag_name_end_to_end_with_delete() {
     run_tag_keys_test_case(EndToEndTestWithDelete {}, predicate, expected_tag_keys).await;
 }
 
+#[tokio::test]
+async fn list_tag_name_end_to_end_with_delete() {
+    let predicate = Predicate::default().with_expr(col("_measurement").eq(lit("swap")));
+    let predicate = InfluxRpcPredicate::new(None, predicate);
+    let expected_tag_keys = vec!["host", "name"];
```
lesam, line 186: Before this change, we would have returned an empty list (as all the values for the `swap` measurement have been deleted).

predicate/src/rpc_predicate.rs
```diff
@@ -198,6 +199,11 @@ fn normalize_predicate(
                 })
                 .and_then(rewrite::simplify_predicate)
         })
+        // Filter out literal true so is_empty works correctly
```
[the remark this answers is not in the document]
crepererum: I thought about that one as well at some point and I think this is OK, @alamb what do you think?

crepererum (approved):
Looks good to me, wanna have a 2nd pair of eyes though, @alamb maybe?

alamb (approved):
Looks great -- thank you @lesam 

I think a similar optimization might apply to `tag_values` as well

service_grpc_influxrpc/src/expr.rs
lesam: There are actually no fields selected - I figured out this morning this is because our test table here doesn't have a schema.

PR #4988, opened 2022-06-29
refactor: Clean up timestamp handling logic and avoid a conversion

alamb:
Related to https://github.com/influxdata/conductor/issues/1000

While trying to see if I could find an issue with timestamp conversion by inspection I came across some code that seemed to do an unnecessary conversion. I can't really see how this will would cause an error, but hope does spring eternal


write_buffer/src/kafka/mod.rs
```diff
@@ -191,18 +191,10 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
                 sequence_number: SequenceNumber::new(record.offset),
             };
 
-            let timestamp_millis =
```
alamb, line 194: I didn't understand why this code was truncating from nanoseconds to milliseconds simply to recreate a nanosecond timestamp from milliseconds 🤔

crepererum: Looks like a leftover from a refactoring. The code was originally written for rdkafka, then copied and modified for rskafka which uses different types.

I wanna point out however that kafka only roundtrips millisecond precision. The write buffer test suits "knows" that and truncates timestamps before asserting roundtrips.

crepererum (approved):

PR #4995, opened 2022-06-30
feat: extract timestamp predicates from generic expressions

crepererum:
Includes #4994.

Found while working on #4976.

It turns out that most of our chunk filtering looks at the timestamp
ranges within `Predicate`. While this if filled for InfluxRPC queries,
we don't have this information for SQL queries.

I see two ways to fix that:

- **Advanced Chunk Filter:** Teach our chunks to evaluate generic
  datafusion expressions. This would have the advantage that we don't
  rely on IOx-specific magic too much. On the other hand, I am unsure
  how much complexity that would require.
- **Lower Generic Expressions for IOx:** Try to scan datfusion
  expressions for simple timestamp filters. Perform this conversion
  early (i.e. when building `Predicate`) so that all chunks can use
  simple timestamp lookups.

Here I try the 2nd approach. Tests are missing for now, since this is
rather an RFC.

### DIFF
--- a/predicate/src/lib.rs
+++ b/predicate/src/lib.rs
@@ -311,7 +311,11 @@ impl Predicate {
 
     /// Adds an expression to the list of general purpose predicates
     pub fn with_expr(mut self, expr: Expr) -> Self {
-        self.exprs.push(expr);
+        if let Some(tsrange) = expr_to_tsrange(&expr) {
+            self.range = Some(self.range.map(|r| r.intersect(&tsrange)).unwrap_or(tsrange));
+        } else {
+            self.exprs.push(expr);
+        }
         self
     }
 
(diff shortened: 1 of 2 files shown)


predicate/src/lib.rs
```diff
@@ -311,7 +311,11 @@ impl Predicate {
 
     /// Adds an expression to the list of general purpose predicates
     pub fn with_expr(mut self, expr: Expr) -> Self {
-        self.exprs.push(expr);
+        if let Some(tsrange) = expr_to_tsrange(&expr) {
+            self.range = Some(self.range.map(|r| r.intersect(&tsrange)).unwrap_or(tsrange));
+        } else {
+            self.exprs.push(expr);
+        }
```
crepererum, line 318: This (and the two functions below) is the actual change.

predicate/src/lib.rs
```diff
@@ -442,6 +446,67 @@ impl Predicate {
...
+        Operator::LtEq if !flip => Operator::LtEq,
+        Operator::Gt if flip => Operator::LtEq,
+        Operator::Gt if !flip => Operator::Gt,
+        Operator::LtEq if flip => Operator::Gt,
+        _ => {
+            return None;
+        }
+    };
+
+    let range = match normalized_op {
```
alamb, line 499: 👍

predicate/src/lib.rs
```diff
@@ -311,7 +311,11 @@ impl Predicate {
 
     /// Adds an expression to the list of general purpose predicates
     pub fn with_expr(mut self, expr: Expr) -> Self {
```
alamb, line 313: I think a very common pattern to capture is a predicate connected by `AND`s and I am pretty sure a SQL predicate is going to come in as `Expr::Binary { left: .., op: AND, right: ...}` rather than repeated calls to `with_expr` despite what this code might imply. 

So like `tag = 'foo' AND time < 100 AND time > 50` can have a timestamp range `50, 100` extracted

There is a function that splits an `Expr` into its conjuncts which I can dig up for you if you want.

alamb (approved):
This is a neat idea -- if we have min/max timestamp ranges for chunks and wish to prune on them it might also be worth looking into using the DataFusion `PruningPredicate` logic, which handles arbitrary expressions (conservatively) -- and so could in theory also handle expressions with `OR`s

https://github.com/apache/arrow-datafusion/blob/5e937abceff5a2b4b7e7981fdabe3220ebd81c7f/datafusion/core/src/physical_optimizer/pruning.rs#L94

So in this case it would be to create a `PruningPredicate` and then feed it a `min/max` value for the `time` column (if that was known)

crepererum:
Alright, let me try to opposite route than and instead of using IOx-special casing use the datafusion pruner and feed it with the IOx-specific knowledge.

PR #4996, opened 2022-06-30
feat: filter chunks before requesting read buffers

crepererum:
~Includes #4994.~

Fixes #4976.


query_tests/cases/in/several_chunks.expected
```diff
@@ -61,3 +61,23 @@
 |               |     IOxReadFilterNode: table_name=h2o, chunks=2 predicate=Predicate                   |
 |               |                                                                                       |
 +---------------+---------------------------------------------------------------------------------------+
+-- SQL: EXPLAIN SELECT * from h2o where time >= to_timestamp('1970-01-01T00:00:00.000000250+00:00');
```
crepererum, line 64: I want to point out that this test here proofs nothing (it's more included as an end2end case). It would pass without this change because after we create the RB chunks here

https://github.com/influxdata/influxdb_iox/blob/40a852552045ff6929c1f31e7d2890a0b4e41f88/querier/src/table/query_access.rs#L53-L60

we apply the predicate to them

https://github.com/influxdata/influxdb_iox/blob/40a852552045ff6929c1f31e7d2890a0b4e41f88/querier/src/table/query_access.rs#L67

So the query engine never "sees" that we created read buffers that we were pruned right away. The extended unit test in `querier` however checks that we don't create the RB chunks.

querier/src/table/mod.rs
```diff
@@ -158,6 +158,41 @@ impl QuerierTable {
...
+            })
+            .try_collect()
+            .await
+            .unwrap();
+        debug!(
+            namespace=%self.namespace_name,
+            table_name=%self.table_name(),
+            n_parquet_files_pre_filter,
+            n_parquet_files_post_filter=parquet_files.len(),
+            "Applied predicate-based filter to parquet file"
```
alamb, line 193: 👍  I found this kind of debugging logic critical to debugging query bugs in the past

querier/src/table/mod.rs
```diff
@@ -158,6 +158,41 @@ impl QuerierTable {
...
+            })
+            .filter_map(|chunk| {
+                let res = chunk
+                    .apply_predicate_to_metadata(predicate)
+                    .map(|pmatch| {
+                        let keep = !matches!(pmatch, PredicateMatch::Zero);
+                        keep.then(|| chunk)
+                    })
+                    .transpose();
+                async move { res }
```
alamb, line 183: is there any reason for the call to `apply_predicate_to_meta` to appear outside the `async` closure? Not that it is a problem but it is different than the closure above (that calls `new_parquet_chunk`, which is `async` and thus needs to be in the closure). 

Basically I am trying to understand for my own edification if it matters

querier/src/table/mod.rs
```diff
@@ -811,15 +851,22 @@ mod tests {
         /// Invokes querier_table.chunks modeling the ingester sending the partitions in this table
         async fn chunks(&self) -> Result<Vec<Arc<dyn QueryChunk>>> {
             let pred = Predicate::default();
+            self.chunks_with_predicate(&pred).await
```
alamb, line 854: 👍

querier/src/table/mod.rs
```diff
@@ -367,8 +405,10 @@ mod tests {
         // now we have some files
         // this contains all files except for:
         // - file111: marked for delete
+        // - file122: filtered by predicate
         // - file221: wrong table
-        let mut chunks = querier_table.chunks().await.unwrap();
+        let pred = Predicate::new().with_range(0, 100);
+        let mut chunks = querier_table.chunks_with_predicate(&pred).await.unwrap();
         chunks.sort_by_key(|c| c.id());
         assert_eq!(chunks.len(), 5);
```
alamb, line 413: I would have expected that `chunks.len()` changed if a new chunk was actually filtered by the predicate (though I do see that you changed the timerange so perhaps I missed something) 🤔 

Update: I see now, there was a new chunk added to the test 👍

alamb (approved):
Looks good to me

PR #5006, opened 2022-06-30
fix: Fix test bug for InfluxRPC `<unknown field> != <value>`

### LINKED ISSUE #4786: Queries for column names (tag keys) with predicates containing columns that don't exist aren't quite right
# What happens

Given a table that contains columns `region` and `time`, (and say all region values are "west"), if you do an `InfluxRPC` query for column names with the predicate `region = "west" AND host != "server01"`, you'll get the empty set for the column names.

In other words, this test currently passes:

```
#[tokio::test]
async fn list_tag_columns_predicate_negative_nonexistent_column() {
    let predicate = PredicateBuilder::default()
        .add_expr(col("state").eq(lit("MA"))) // state=MA
        .add_expr(col("host").not_eq(lit("server01"))) // nonexistent column with !=; always true
        .build();
    let predicate = InfluxRpcPredicate::new(None, predicate);
    let expected_tag_keys = vec!["city", "county", "state"];
    // This currently returns nothing, which is incorrect, it should return "city", "county",
    // "state" because a nonexistent column is always not equal to anything.
    let expected_tag_keys = vec![];
    run_tag_keys_test_case(TwoMeasurementsManyNulls {}, predicate, expected_tag_keys).await;
}
```

# What I expected to happen

Because the table doesn't contain the `host` column at all, then `host != "server01"` should always evaluate to `true`. The `region = "west"` expression should be applied, and because in this example all region values in the table are "west", then the result should be all columns, or `[region, time]`.

In other words, this test should pass instead:

```
#[tokio::test]
async fn list_tag_columns_predicate_negative_nonexistent_column() {
    let predicate = PredicateBuilder::default()
        .add_expr(col("state").eq(lit("MA"))) // state=MA
        .add_expr(col("host").not_eq(lit("server01"))) // nonexistent column with !=; always true
        .build();
    let predicate = InfluxRpcPredicate::new(None, predicate);
    let expected_tag_keys = vec!["city", "county", "state"];
    run_tag_keys_test_case(TwoMeasurementsManyNulls {}, predicate, expected_tag_keys).await;
}
```

# More context

- https://github.com/influxdata/influxdb_iox/pull/4767
- https://github.com/influxdata/influxdb_iox/pull/4776

crepererum:
Fixes #4786.


predicate/src/rpc_predicate/ignore_unknown.rs
```diff
@@ -0,0 +1,42 @@
...
+
+struct IgnoreRewriter {
+    schema: Arc<Schema>,
+}
+
+impl ExprRewriter for IgnoreRewriter {
+    fn mutate(&mut self, expr: Expr) -> DataFusionResult<Expr> {
+        if let Expr::BinaryExpr {
+            left,
+            op: Operator::NotEq,
```
crepererum, line 25: Should we only ignore `!=` or should we also ignore other operators? (esp. "regex NOT match")

alamb (commented):
This is a neat idea @crepererum  -- the code is quite beautiful. However, TLDR is I don't think the `list_tag_columns_predicate_negative_nonexistent_column`  test is quite right.

Specifically, rather than 
```rust
.with_expr(col("host").not_eq(lit("server01")))
``` 

it should be something like

```rust
.with_expr(make_empty_tag_ref_expr("host").not_eq(lit("server01")))
``` 

Using `make_empty_tag_ref_expr` defined in https://github.com/influxdata/influxdb_iox/blob/master/query_tests/src/influxrpc/read_filter.rs#L805

I suspect if you adjust the test the code changes will not be needed. Specifically, the way a predicate like `host != 'server1'` from influxRPC is supposed to work is:

# Step 1
The predicate comes encoded as `TagRef('host') != 'server1'` on the storage grpc proto

# Step 2
`TagRef('host')` is transformed to `CASE WHEN host IS NULL THEN '' ELSE host END` resulting in the predicate like

```sql
CASE WHEN host IS NULL THEN '' ELSE host END != 'server1'
```

This happens in https://github.com/influxdata/influxdb_iox/blob/master/service_grpc_influxrpc/src/expr.rs#L511


# Step 3
The predicate rewrite code replaces references to any non existent columns with `NULL`. So in this case, the predicate will be rewritten to

```sql
CASE WHEN NULL IS NULL THEN '' ELSE host END != 'server1'
```
-->
```sql
 '' != 'server1'
```

-->
```sql
true
```

crepererum:
@alamb You're right, then the test passes. So I've removed the new code again and just adjusted the test. I think that still counts as "closing the issue", because it turned out to be a test bug.

alamb:
>  think that still counts as "closing the issue", because it turned out to be a test bug.

I agree -- thank you

query_tests/src/influxrpc/tag_keys.rs
```diff
@@ -7,6 +7,8 @@ use iox_query::{
 };
 use predicate::{rpc_predicate::InfluxRpcPredicate, Predicate};
 
+use super::util::make_empty_tag_ref_expr;
```
alamb, line 10: 👍

alamb (approved):
Thanks @crepererum . FYI @carols10cents

PR #5009, opened 2022-06-30
refactor: Move sharding query from namespace to table

### LINKED ISSUE #5003: Move querier sharding logic down further to where ingesters are queried
Follow-on to https://github.com/influxdata/influxdb_iox/issues/4621 / https://github.com/influxdata/influxdb_iox/pull/4887

From [this comment](https://github.com/influxdata/influxdb_iox/pull/4887#discussion_r904769712):

> Can we move this code further down were the ingesters are actually queried? The current layout makes a quite strong assumption that we only shard based on tables, which with the current shard_for_table function is true but will require an unnecessary large refactoring when we want to extend the sharding to for example include time or tag columns.

carols10cents:
Better supports future work. Fixes #5003.


crepererum (approved):
Thank you!

PR #5010, opened 2022-06-30
ci: Verify checksum of `protoc` binary

pierwill:
Adds an integrity check before unzipping and using this binary.


jacobmarble:
Let's just close this PR and delete perf.py

PR #5016, opened 2022-07-01
refactor: remove `Predicat::partition_key`

crepererum:
There is no way a user can filter for partition keys (neither via
InfluxRPC nor via SQL) and the query engine doesn't use this field at
all. So let's remove it.


alamb (approved):
Makes sense to me -- nice cleanup

PR #5017, opened 2022-07-01
feat: easy tests of table summary against predicate

crepererum:
Helps with #4976.

Alternative to #4995.


predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+        &self,
+        table_summary: &TableSummary,
+        schema: SchemaRef,
+    ) -> PredicateMatch {
+        let summary = SummaryWrapper {
+            summary: table_summary,
+        };
+
+        let timestamp_expr = self.range.map(|range| {
+            col(TIME_COLUMN_NAME)
```
alamb, line 248: I think this is the same as `Self::make_timestamp_predicate_expr`

predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+                .gt_eq(lit_timestamp_nano(range.start()))
+                .and(col(TIME_COLUMN_NAME).lt(lit_timestamp_nano(range.end())))
+        });
+
+        for expr in self.exprs.iter().chain(timestamp_expr.as_ref()) {
```
alamb, line 253: It might be worth checking the timestamp predicate first rather than last ( the theory is it will be the most selective and this pruning is on the critical query planning path)

predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+            match PruningPredicate::try_new(expr.clone(), Arc::clone(&schema)) {
+                Ok(pp) => {
+                    match pp.prune(&summary) {
+                        Ok(matched) => {
+                            assert_eq!(matched.len(), 1);
+                            if matched[0] {
+                                // might match => test next expr
+                                continue;
+                            } else {
+                                // does not match => since expressions are `AND`ed, we know that we will have zero matches
```
alamb, line 263: I think this logic is correct, but I also that correctly handling `AND` exprs  is covered by the logic in `PruningPredicate` -- you could use something like (untested);

```rust
if let Some(filter_expr) = self.filter_expr() {
  match PruningPredicate::try_new(filter_expr, Arc::clone(&schema)) {
    ..
  }           
}
```

However, I think this approach is correct and fine as well (though perhaps more code)

predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+
+        PredicateMatch::Unknown
+    }
+}
+
+struct SummaryWrapper<'a> {
+    summary: &'a TableSummary,
+}
+
+impl<'a> PruningStatistics for SummaryWrapper<'a> {
```
alamb, line 296: ❤️

predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+    ) -> Option<arrow::array::ArrayRef> {
+        let col = self.summary.column(&column.name)?;
+        let stats = &col.stats;
+
+        // special handling for timestamps
+        if col.influxdb_type == Some(InfluxDbType::Timestamp) {
+            if let data_types::Statistics::I64(val) = stats {
+                return Some(Arc::new(TimestampNanosecondArray::from(vec![val.min])));
+            } else {
+                // wrong data type
```
alamb, line 309: This is hopefully impossible so could perhaps be a `panic!` but silently returning `None` is also fine (other things are going to explode if a `Timestamp` is not an `i64`).

predicate/src/lib.rs
```diff
@@ -673,4 +819,112 @@ mod tests {
...
+                stats: data_types::Statistics::I64(StatValues {
+                    min: Some(300),
+                    max: Some(300),
+                    null_count: Some(0),
+                    total_count: 1_000,
+                    distinct_count: None,
+                }),
+            }],
+        };
+        assert_eq!(
```
alamb, line 907: nice tests

alamb (approved):
Nice work @crepererum

predicate/src/lib.rs
```diff
@@ -225,6 +233,140 @@ impl Predicate {
...
+    fn max_values(
+        &self,
+        column: &datafusion::logical_plan::Column,
+    ) -> Option<arrow::array::ArrayRef> {
+        let col = self.summary.column(&column.name)?;
+        let stats = &col.stats;
+
+        // special handling for timestamps
+        if col.influxdb_type == Some(InfluxDbType::Timestamp) {
+            if let data_types::Statistics::I64(val) = stats {
```
shepmaster, line 336: This appears duplicated; you could add

```rust
impl Statistics {
    fn into_i64(self) -> Option<StatValues<i64>>
}
```

making this code into

```rust
let val = stats.into_i64()?;
return Some(Arc::new(TimestampNanosecondArray::from(vec![val.max])));
```

predicate/src/lib.rs
crepererum: done

predicate/src/lib.rs
crepererum: done

predicate/src/lib.rs
crepererum: done

predicate/src/lib.rs
crepererum: done

PR #5019, opened 2022-07-01
chore: Upgrade to Rust 1.62

carols10cents:
🎉


alamb (approved):
Thanks @carols10cents  -- LGTM -- though it looks like `cargo fmt` isn't super happy

PR #5030, opened 2022-07-04
refactor: unify `apply_predicate_to_metadata`

crepererum:
Instead of using some hand-rolled timestamp-based logic (or just
"unknown") all over the place, just use logic introduced in #5017.

This requires slightly improved table summaries within the querier that
at least has min/max for the timestamp column. For that, the former
`IngesterChunk`-specific `calculate_summary` method was extended to
`create_basic_summary` to include that data and is now also used by
`QuerierParquetChunk`.

Note: `QuerierRBChunk` already has detailled metrics that are provided
by the read buffer implementation.

Should we ever need even better pruning for `QuerierParquetChunk` (or
`IngesterChunk`) then we _only_ need add extra data to the table
summaries.

Closes #4976.


iox_query/src/util.rs
```diff
@@ -235,15 +236,18 @@ impl<'a> ExprRewriter for MissingColumnsToNull<'a> {
 }
 
 /// Return min and max for column `time` of the given set of record batches
-pub fn compute_timenanosecond_min_max(batches: &[RecordBatch]) -> Result<(i64, i64)> {
+pub fn compute_timenanosecond_min_max(batches: &[RecordBatch]) -> Result<TimestampMinMax> {
```
alamb, line 239: 👍

querier/src/chunk/util.rs
```diff
@@ -0,0 +1,296 @@
...
+        let row_count = 0;
+        let ts_min_max = TimestampMinMax { min: 10, max: 20 };
+
+        let actual = create_basic_summary(row_count, &schema, ts_min_max);
+        let expected = TableSummary { columns: vec![] };
+        assert_eq!(actual, expected);
+    }
+
+    #[test]
+    fn test_create_basic_summary_no_rows() {
```
alamb, line 103: It isn't clear to me that this test adds any additional coverage over `test_create_basic_summary`, not that it is a problem.

querier/src/ingester/mod.rs
```diff
@@ -1030,13 +1038,9 @@ impl QueryChunkMeta for IngesterChunk {
     }
 
     fn timestamp_min_max(&self) -> Option<TimestampMinMax> {
-        // TODO: may want to ask the Ingester to send this value instead of computing it here.
         // Note: if we return None here, this chunk will be considered ovelapped with all other chunks
         // even if it does not and lead to unecessary deduplication
```
alamb, line 1042: It seems like this comment about `None` is now out of date

alamb (approved):
I really like this change @crepererum  -- unifying all the predicate application / pruning to use `TableSummary` is a very nice structure 👌

iox_query/src/lib.rs
```diff
@@ -197,7 +197,12 @@ pub trait QueryChunk: QueryChunkMeta + Debug + Send + Sync + 'static {
     fn apply_predicate_to_metadata(
         &self,
         predicate: &Predicate,
-    ) -> Result<PredicateMatch, QueryChunkError>;
+    ) -> Result<PredicateMatch, QueryChunkError> {
+        Ok(self
+            .summary()
+            .map(|summary| predicate.apply_to_table_summary(summary, self.schema().as_arrow()))
```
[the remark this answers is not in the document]
crepererum: IMHO there's nothing wrong if a chunk WANTS to override the behavior because it has fancy internal data structures that are not purely based on a summary.

querier/src/chunk/util.rs
crepererum: I left it because the original test (that was moved) used record batches as input and there it indeed made a difference. IMHO it's OK if the test doesn't make too many assumption about the implementation and because we may want to extend `create_basic_summary` in the future again.

querier/src/ingester/mod.rs
crepererum: removed

PR #5035, opened 2022-07-05
fix: handle NULLs for delete predicates

### LINKED ISSUE #3582: Wrong result: delete predicate on columns that do not exist in the batch will remove rows of that batch
__Steps to reproduce:__
This happens during compaction of the new Ingester. This is the test that reproduce this:
`test_compact_many_batches_different_columns_different_order_with_duplicates_with_many_deletes_2` (currently only in the PR #3581). I let the test on with wrong results which will signal us when the bug is fixed.

### LINKED ISSUE #5023: Delete predicate applies incorrectly to empty column
@lesam  notes, [in slack](https://influxdata.slack.com/archives/CRK9M8L5Q/p1656625129563529?thread_ts=1656624676.184329&cid=CRK9M8L5Q):

Seems possibly related to https://github.com/influxdata/influxdb_iox/issues/3582

__Steps to reproduce:__

Given this input;
```shell
> ./target/debug/influxdb_iox query 0000000000000001_0000000000000001 "select * from swap"
+----------+----+-------+-----+----------------------+
| host     | in | name  | out | time                 |
+----------+----+-------+-----+----------------------+
| server01 | 3  |       | 4   | 2020-06-11T16:54:10Z |
| server01 | 3  | disk0 | 4   | 2020-06-11T16:54:20Z |
| server01 | 3  | disk1 | 4   | 2020-06-11T16:54:20Z |
+----------+----+-------+-----+----------------------+
```

Run this delete:

```shell
> curl -XPOST http://localhost:8080/api/v2/delete\?org\=0000000000000001\&bucket\=0000000000000001 -d '{"predicate":"_measurement=swap AND name=\"disk0\"","start":"1970-01-01T00:00:00Z","stop":"2070-01-02T00:00:00Z"}'
```

Then look at the data again:
```shell
> ./target/debug/influxdb_iox query 0000000000000001_0000000000000001 "select * from swap"
+----------+----+-------+-----+----------------------+
| host     | in | name  | out | time                 |
+----------+----+-------+-----+----------------------+
| server01 | 3  | disk1 | 4   | 2020-06-11T16:54:20Z |
+----------+----+-------+-----+----------------------+
```

__Expected behaviour:__

Expect these two rows:
```shell
+----------+----+-------+-----+----------------------+
| host     | in | name  | out | time                 |
+----------+----+-------+-----+----------------------+
| server01 | 3  |       | 4   | 2020-06-11T16:54:10Z |
| server01 | 3  | disk1 | 4   | 2020-06-11T16:54:20Z |
+----------+----+-------+-----+----------------------+
```

@lesam  says "Looks like the delete predicate is incorrectly applying to the empty row"

crepererum:
Fixes #3582.
Fixes #5023.


query_tests/src/scenarios/library.rs
```diff
@@ -1348,3 +1350,98 @@ impl DbSetup for TwoChunksMissingColumns {
...
+            "swap,host=server01,name=disk1 in=3,out=4,field_f64=60 1591894470000000000",
+            "swap,host=server01,name=disk1 in=3,out=4,field_f64=70 1591894480000000000",
+            "swap,host=server01,name=disk1 in=3,out=4,field_i64=1i 1591894490000000000",
+            "swap,host=server01,name=disk1 in=3,out=4,field_u64=1u 1591894500000000000",
+            "swap,host=server01,name=disk1 in=3,out=4,field_bool=true 1591894510000000000",
+            "swap,host=server01,name=disk1 in=3,out=4,field_string=\"foo\" 1591894520000000000",
+        ];
+
+        let delete_table_name = "swap";
+        let pred1 = DeletePredicate {
```
alamb, line 1375: I originally thought that it might be worth testing is `Op::NotEq` in addition to `Op::Eq` however, upon reading the docs for the cloud it appears that only `eq` is supported:

https://docs.influxdata.com/influxdb/cloud/reference/syntax/delete-predicate/#:~:text=InfluxDB%20uses%20the%20delete%20predicate,evaluate%20to%20false%20are%20preserved.

read_buffer/src/row_group.rs
```diff
@@ -1605,12 +1606,16 @@ impl TryFrom<&DfExpr> for BinaryExpr {
         match df_expr {
             DfExpr::BinaryExpr { left, op, right } => {
                 match (&**left, &**right) {
-                    (DfExpr::Column(c), DfExpr::Literal(scalar)) => Ok(Self::new(
-                        &c.name,
-                        Operator::try_from(op)?,
-                        Literal::try_from(scalar)?,
-                    )),
-                    (DfExpr::Literal(_), DfExpr::Column(_)) => {
+                    (col @ (DfExpr::Column(_) | DfExpr::Case { .. }), DfExpr::Literal(scalar)) => {
```
alamb, line 1609: I am worried about this transformation as it effectively implies that the read buffer knows how to handle `'foo' = CASE WHEN foo IS NULL THEN '' ELSE foo END` type expressions (aka treat null as empty string) which I do not think is the case.

Or put another this way, unwrapping `CASE WHEN foo IS NULL then '' ELSE foo END` is only valid if there are no null values in `foo`

querier/src/namespace/query_access.rs
```diff
@@ -333,24 +333,24 @@ mod tests {
...
+                "|               |   Projection: #mem.host, #mem.perc, #mem.time                                                             |",
+                "|               |     TableScan: mem projection=[host, perc, time]                                                          |",
+                "| physical_plan | SortExec: [host@0 ASC NULLS LAST,time@2 ASC NULLS LAST]                                                   |",
+                "|               |   CoalescePartitionsExec                                                                                  |",
+                "|               |     ProjectionExec: expr=[host@0 as host, perc@1 as perc, time@2 as time]                                 |",
+                "|               |       UnionExec                                                                                           |",
+                "|               |         CoalesceBatchesExec: target_batch_size=500                                                        |",
+                "|               |           FilterExec: time@2 < 1 OR time@2 > 13 OR NOT CASE WHEN host@0 IS NULL THEN  ELSE host@0 END = d |",
```
alamb, line 347: this plan change makes sense to me

datafusion_util/src/lib.rs
```diff
@@ -347,6 +348,58 @@ pub fn context_with_table(batch: RecordBatch) -> SessionContext {
...
+                }
+                _ => None,
+            }
+        }
+        _ => None,
+    }
+}
+
+/// Check if a scalar looks like a default value for NULL values.
+fn is_null_scalar(scalar: &ScalarValue) -> bool {
```
alamb, line 393: A more specific name might be:

```suggestion
fn is_influxrpc_default_null_scalar(scalar: &ScalarValue) -> bool {
```

predicate/src/delete_expr.rs
```diff
@@ -1,19 +1,45 @@
...
+            ScalarValue::Utf8(Some("".to_owned())),
+        ),
+    };
+
+    let column = Box::new(Expr::Column(datafusion::logical_plan::Column {
         relation: None,
         name: expr.column,
-    };
+    }));
+    let column = Box::new(Expr::Case {
```
alamb, line 30: I think some comments here would be helpful for future readers -- namely that this is converting column references in delete predicates to the form `CASE WHEN col IS NULL THEN default_value ELSE col END` to map the semantics of influx RPC exprs into SQL

predicate/src/delete_expr.rs
```diff
@@ -1,19 +1,45 @@
...
+use datafusion::scalar::ScalarValue;
+use datafusion_util::extract_null_wrapped_column;
 use snafu::{ResultExt, Snafu};
 use std::ops::Deref;
 
 pub(crate) fn expr_to_df(expr: DeleteExpr) -> datafusion::logical_plan::Expr {
     use datafusion::logical_plan::Expr;
 
-    let column = datafusion::logical_plan::Column {
+    let (value_scalar, null_scalar) = match expr.scalar {
```
alamb, line 10: As this code needs to remain in sync with `datafusion_util::is_null_scalar`, it would be nice to keep the two functions in the same module.

I recommend encapsulating this logic in a function like the following:

```rust
fn null_default_scalar(data_type: &DataType) -> ScalarValue {
  match data_type {
    DataType::Boolean => ScalarValue::Boolean(Some(false)), 
    ..
   }
}
```

predicate/src/delete_expr.rs
alamb: Or maybe refactor the code (and comments) in `build_tag_ref` https://github.com/influxdata/influxdb_iox/blob/main/service_grpc_influxrpc/src/expr.rs#L511-L534 to `datafusion_util::influxrpc_column_ref` or something

read_buffer/src/row_group.rs
alamb: Although I see that the tests fail with:

```
thread 'influxrpc::tag_keys::list_tag_name_end_to_end_with_delete' panicked at 'converted plan to strings successfully: Execution("Error creating scan for table swap chunk ChunkId(4): Error translating predicate: unsupported expression CASE WHEN #name IS NULL THEN Utf8(\"\") ELSE #name END Eq Utf8(\"disk0\"), Predicate {\n    field_columns: None,\n    range: Some(\n        TimestampRange {\n            start: 6000,\n            end: 6000,\n        },\n    ),\n    exprs: [\n        CASE WHEN #name IS NULL THEN Utf8(\"\") ELSE #name END = Utf8(\"disk0\"),\n    ],\n    value_expr: [],\n}")', query_tests/src/influxrpc/tag_keys.rs:40:14
```

without this code

datafusion_util/src/lib.rs
```diff
@@ -347,6 +348,58 @@ pub fn context_with_table(batch: RecordBatch) -> SessionContext {
...
 
+/// Extract the column name from a an expression that might be treated with special NULL handling.
+///
+/// Normally we would just use the column itself:
+///
+/// ```text
+/// #col
+/// ```
+///
+/// However for certain interfaces we need to handle NULL values in a special way, e.g.:
```
alamb, line 359:
```suggestion
/// However to properly evaluate influxrpc semantics,  which does not have NULLs, 
/// we map NULL values in IOx in the following  way, e.g.:
```

read_buffer/src/row_group.rs
alamb: I looked more into it -- I think what is happening is that the read buffer is (correctly) reporting it can't evaluate the delete predicate during the scan. However, evaluating the delete predicate during the scan isn't required for correctness (as there is another filter applied):


```
(sort)
  (filter with all exprs (including deletes))
    (read filter with pushed down exprs)
``

The exprs pushed to the filter are not required for correctness, but rather they are a performance optimization


Thus I think we should either:
1. Ignore predicates the read filter can't evaluate (don't try and push them down)
2. Apply this transformation only if there are no null values in the column (the read buffer has detailed statistics)

alamb:
marking as draft until we have time to reasses

alamb:

PR #5041, opened 2022-07-05
refactor: convert a panic into an error and throw a warning if we choose non-actionable compacting candidates

### LINKED ISSUE #5038: Convert a panic of overlaps in both time ranges and sequence number ranges an into an error instead
# Rationale

Before we want to [panic in case of data corruption](https://github.com/influxdata/influxdb_iox/blob/16bd3e67c0330e88ba1c0c9b89da2657d3bb8a18/compactor/src/compact.rs#L1036), overlaps in both time ranges and sequence number ranges, for  us to look into immediately. But since we can let compactor compact on other partition, we want to change this to error and log only

### LINKED ISSUE #5039: Reduce OOM: Do not read catalog for partition candidates that the Compactor cannot handle
# Rationale

This may be related to https://github.com/influxdata/influxdb_iox/issues/4939 and  https://github.com/influxdata/conductor/issues/1014

This function [groups_to_compact_and_files_to_upgrade] right there (https://github.com/influxdata/influxdb_iox/blob/1de022136c1f89faf456d54c3478f6542e475a87/compactor/src/handler.rs#L204) shows that we do read catalog for partitions that the compactor cannot compact due to memory limit. This will lead to OOM out of our control. That function should be moved inside the function [compaction_parttion](https://github.com/influxdata/influxdb_iox/blob/1de022136c1f89faf456d54c3478f6542e475a87/compactor/src/handler.rs#L224) that is only invoked if the Compactor has estimated memory for it

NGA-TRAN:
Closes #5038 
Closes #5039
   It turned out the issue in #5039 is not an issue because there are always actions for a chosen candidate to either compact some files or upgrade a few or both. I have added a warning if that is not the case though


compactor/src/compact.rs
```diff
@@ -1074,19 +1096,18 @@ impl Compactor {
...
-            )
+            return Err(crate::compact::Error::OverlapTimeAndSequenceNumber {
+                file_id_1: file_1.id,
+                file_id_2: file_2.id,
+                min_seq_1: file_1.min_sequence_number.get(),
+                max_seq_1: file_1.max_sequence_number.get(),
+                min_seq_2: file_2.min_sequence_number.get(),
+                max_seq_2: file_2.max_sequence_number.get(),
+                partition_id: file_1.partition_id,
+            });
```
NGA-TRAN, line 1107: This is the panic converting to error

compactor/src/handler.rs
```diff
@@ -243,6 +243,10 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
                             debug!(%max_size, %used_size, n_compactions=%handles.len(), "reached maximum concurrent compaction size limit");
                             break;
                         }
+                    } else {
+                        // All candidates should be compactable (have files to compact and/or upgrade)
+                        // Reach here means we do not choose the right cnadidates and waste our time to repeat this cycle
+                        warn!("The candidate partition {} has no files to be either compacted or upgraded", c.candidate.partition_id);
```
NGA-TRAN, line 249: This is the warning in case we choose non-actionable candidates which we have never verified

compactor/src/handler.rs
```diff
@@ -243,6 +243,10 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
                             debug!(%max_size, %used_size, n_compactions=%handles.len(), "reached maximum concurrent compaction size limit");
                             break;
                         }
+                    } else {
+                        // All candidates should be compactable (have files to compact and/or upgrade)
+                        // Reach here means we do not choose the right cnadidates and waste our time to repeat this cycle
```
alamb, line 248:
```suggestion
                        // Reaching here means we do not choose the right candidates and 
                        // it would be a waste of time to repeat this cycle
```

compactor/src/handler.rs
alamb: I think it makes sense to `warn!` on a case that "should not happen" and needs human investigation

alamb (approved):

PR #5043, opened 2022-07-05
test: add test for "duration since production" in kafka `write_buffer` implementation

alamb:
# Rationale
re  https://github.com/influxdata/conductor/issues/1000

I am tracking down what seems to be incorrectly reported TTBR and one theory was that the kafka read buffer was to blame, so I wrote a test that validated the timestamps. Good news is that things appear ok. Bad news is I still don't know what is going on with https://github.com/influxdata/conductor/issues/1000

# Changes
Add a test demonstrating timestamp calculations in kafka writes

Next up I will look into the metric reporting infrastructure


write_buffer/src/kafka/mod.rs
```diff
@@ -717,6 +718,73 @@ mod tests {
             .unwrap();
     }
 
+    // Coverage of https://github.com/influxdata/conductor/issues/1000
+    #[tokio::test]
+    async fn test_batching_timestamps() {
```
crepererum, line 723: This doesn't seem to be Kafka-specific, so could this be a generic test?

alamb: in d6c41ac7b

crepererum (approved):
Thank you!

PR #5046, opened 2022-07-05
fix: optimize field columns for all-time predicates

### LINKED ISSUE #4827: Ensure "All time" metadata queries are appropriately special cased

lesam:
Also fix timestamp range to allow selecting points at MAX_NANO_TIME

Closes #4827 

### DIFF
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -2085,14 +2085,14 @@ impl TimestampRange {
     ///
     /// Takes an inclusive start and an exclusive end. You may create an empty range by setting `start = end`.
     ///
-    /// Clamps to [`MIN_NANO_TIME`]/[`MAX_NANO_TIME`].
+    /// Clamps start to [`MIN_NANO_TIME`].
+    /// end is unclamped - end may be set to i64:MAX == MAX_NANO_TIME+1 to indicate no restriction on time.
     ///
     /// # Panic
     /// Panics if `start > end`.
     pub fn new(start: i64, end: i64) -> Self {
         assert!(end >= start, "start ({start}) > end ({end})");
         let start = start.max(MIN_NANO_TIME);
-        let end = end.min(MAX_NANO_TIME);
         Self { start, end }
     }
 
@@ -3104,8 +3104,8 @@ mod tests {
     fn test_timestamp_nano_min_max() {
         let cases = vec![
             (
-                "MIN/MAX Nanos",
-                TimestampRange::new(MIN_NANO_TIME, MAX_NANO_TIME),
+                "MIN / MAX Nanos",
+                TimestampRange::new(MIN_NANO_TIME, MAX_NANO_TIME + 1),
             ),
             ("MIN/MAX i64", TimestampRange::new(i64::MIN, i64::MAX)),
         ];
@@ -3116,7 +3116,7 @@ mod tests {
             assert!(range.contains(MIN_NANO_TIME));
             assert!(range.contains(MIN_NANO_TIME + 1));
             assert!(range.contains(MAX_NANO_TIME - 1));
-            assert!(!range.contains(MAX_NANO_TIME));
+            assert!(range.contains(MAX_NANO_TIME));
             assert!(!range.contains(i64::MAX));
         }
     }
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -2097,6 +2097,11 @@ impl TimestampRange {
         Self { start, end }
     }
 
+    /// Returns true if this range contains all representable timestamps
+    pub fn contains_all(&self) -> bool {
+        self.start <= MIN_NANO_TIME && self.end > MAX_NANO_TIME
+    }
+
     #[inline]
     /// Returns true if this range contains the value v
     pub fn contains(&self, v: i64) -> bool {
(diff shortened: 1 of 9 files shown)


iox_query/src/frontend/influxrpc.rs
```diff
@@ -670,9 +668,25 @@ impl InfluxRpcPlanner {
...
+            if predicate.is_empty() {
+                // optimization: just get the field columns from metadata.
+                // note this both ignores field keys, and sets the timestamp data 'incorrectly'.
+                let schema = database
+                    .table_schema(table_name)
+                    .context(TableRemovedSnafu { table_name })?;
+                let fields = schema.fields_iter().map(|f| Field {
+                    name: f.name().clone(),
+                    data_type: f.data_type().clone(),
+                    last_timestamp: 0,
```
lesam, line 683: I could make this the system `now` time, but I think a constant signalling value would be better since it is much more clear to our users that it does not represent the time of any particular field value. `MIN_NANO_TIME` or `MAX_NANO_TIME` might also be good but zero was nicely 'default-y'.

data_types/src/lib.rs
```diff
@@ -2085,14 +2085,14 @@ impl TimestampRange {
     ///
     /// Takes an inclusive start and an exclusive end. You may create an empty range by setting `start = end`.
     ///
-    /// Clamps to [`MIN_NANO_TIME`]/[`MAX_NANO_TIME`].
+    /// Clamps start to [`MIN_NANO_TIME`].
+    /// end is unclamped - end may be set to i64:MAX == MAX_NANO_TIME+1 to indicate no restriction on time.
```
lesam, line 2089: Note this change!

predicate/src/lib.rs
```diff
@@ -220,7 +220,7 @@ impl Predicate {
     /// existing storage engine
     pub(crate) fn with_clear_timestamp_if_max_range(mut self) -> Self {
         self.range = self.range.take().and_then(|range| {
-            if range.start() <= MIN_NANO_TIME && range.end() >= MAX_NANO_TIME {
+            if range.start() <= MIN_NANO_TIME && range.end() == i64::MAX {
```
lesam, line 223: Note this change!

predicate/src/lib.rs
```diff
@@ -220,7 +220,7 @@ impl Predicate {
     /// existing storage engine
     pub(crate) fn with_clear_timestamp_if_max_range(mut self) -> Self {
         self.range = self.range.take().and_then(|range| {
-            if range.start() <= MIN_NANO_TIME && range.end() >= MAX_NANO_TIME {
+            if range.start() <= MIN_NANO_TIME && range.end() > MAX_NANO_TIME {
```
lesam, line 223: Note also this change!

data_types/src/lib.rs
crepererum: This sounds really hackish to me. If you wanna allow no time restriction at all, then use `Option<TimestampRange>`.

We clamp it because `i64` is NOT a valid nanosecond range (now that I write that I think the clamping is actually incomplete, `start` must also be clamped by `MAX_NANO_TIME` and `end` must also be clamped by `MIN_NANO_TIME`). Otherwise other parts of the stack will likely explode when constructing filter predicates based on this data. Since this type is used for other requests as well, I suspect a user can now send out-of-bounds values and either get an error or (even worse) get the querier to panic.

service_grpc_influxrpc/src/service.rs
```diff
@@ -2908,7 +2908,7 @@ mod tests {
         let request = MeasurementFieldsRequest {
             source: source.clone(),
             measurement: "TheMeasurement".into(),
-            range: None,
+            range: Some(make_timestamp_range(i64::MIN, i64::MAX - 1)),
```
crepererum, line 2911: This is what I meant with typing: Use `None` if you actually mean "no restriction", not some magic out-of-bounds values.

crepererum (changes requested):

service_grpc_influxrpc/src/service.rs
lesam: This is actually just to avoid the 'optimized' case so the test doesn't change. This filter says 'range is all time except `i64::MAX-1`. Passing `None` still works fine and would actually mean 'all time'.

data_types/src/lib.rs
lesam: I disagree about clamping by MAX_NANO_TIME (though I agree we probably should clamp end by MIN_NANO_TIME for safety).

The definition for MAX_NANO_TIME:

```
/// maximum time that can be represented.
///
/// 2262-04-11 23:47:16.854775806 +0000 UTC
///
/// The highest time represented by a nanosecond needs to be used for an
/// exclusive range in the shard group, so the maximum time needs to be one
/// less than the possible maximum number of nanoseconds representable by an
/// int64 so that we don't lose a point at that one time.
/// Source: [influxdb](https://github.com/influxdata/influxdb/blob/540bb66e1381a48a6d1ede4fc3e49c75a7d9f4af/models/time.go#L12-L34)
pub const MAX_NANO_TIME: i64 = i64::MAX - 1;
```

Filter expressions need to be able to express the time range 'all representable timestamps above 1000', since the end of the range is exclusive this is the range `{1000, MAX_NANO_TIME+1}` == `{1000, i64::MAX}`.

IDPE unfortunately gets this right for tag keys:

https://github.com/influxdata/idpe/blob/master/storage/store/store.go#L1395-L1397

But wrong for ReadFilter:

https://github.com/influxdata/idpe/blob/master/storage/store/store.go#L1086-L1088

lesam: Flux does have the bounds correct:

https://github.com/influxdata/flux/blob/master/execute/bounds.go#L15-L28

lesam: Filed https://github.com/influxdata/influxdb/issues/23519 for OSS2

lesam: > now that I write that I think the clamping is actually incomplete, start must also be clamped by MAX_NANO_TIME and end must also be clamped by MIN_NANO_TIME

I agree end should also be clamped by MIN_NANO_TIME, fixed now.

crepererum: We had a quick 1:1 chat, I agree w/ @lesam 's statements here.

lesam:
Created https://github.com/influxdata/influxdb_iox/issues/5051

Issue #5032, opened 2022-07-04
Lazy parquet chunk => RB chunk conversion


crepererum:
# Why
Certain parts of our query stack assume that gathering chunks does not pull data. The querier somewhat violates that pattern by instancing RBs for every requested parquet file. We should delay this conversion as long as possible.

The assumption of the query stack is both practical (i.e. it simplifies the implementation of some of our InfluxRPC routines) as well as necessary (imagine a `SELECT * from table limit 1`). #4996 improved the situation a bit but does not cover all use cases (e.g. https://github.com/influxdata/conductor/issues/1012 )

# What
The implementation of `QuerierParquetChunk` and `QuerierRBChunk` are very close.

| Method                                  | identical | RB :arrow_up: quality | RB faster |
| --------------------------------------- | --------- | --------------------- | --------- |
| `QueryChunkMeta::summary`               |           | X[^su]                |           |
| `QueryChunkMeta::schema`                | X         |                       |           |
| `QueryChunkMeta::partition_sort_key`    | X         |                       |           |
| `QueryChunkMeta::partition_id`          | X         |                       |           |
| `QueryChunkMeta::sort_key`              | X         |                       |           |
| `QueryChunkMeta::delete_predicates`     | X         |                       |           |
| `QueryChunkMeta::timestamp_min_max`     | X         |                       |           |
| `QueryChunk::id`                        | X         |                       |           |
| `QueryChunk::table_name`                | X         |                       |           |
| `QueryChunk::may_contain_pk_duplicates` | X         |                       |           |
| `QueryChunk::column_names`              |           | X                     |           |
| `QueryChunk::column_values`             |           | X                     |           |
| `QueryChunk::read_filter`               |           |                       | X         |
| `QueryChunk::chunk_type`                | X[^ct]    |                       |           |
| `QueryChunk::order`                     | X         |                       |           |

[^su]: as of #5030 this also improves `QueryChunk::apply_predicate_to_metadata`
[^ct]: both return a constant value, but the value is different

All methods are sync, except for `QueryChunk::read_filter` which contains a `SendableRecordBatchStream` and therefore is somewhat async. Therefore this is the only method that would allow lazy loading of data.

# How
As shown in the section above, there are only a few methods that actually differentiate the RB version from the parquet version. Hence, the following schema should be implemented:

[...]

PR #5049, opened 2022-07-06
feat: add "peek" functionality for caches

crepererum:
Required for #5032.


cache_system/src/cache/driver.rs
```diff
@@ -99,37 +99,16 @@ where
                 let k_captured = k.clone();
                 let handle = tokio::spawn(async move {
                     let loader_fut = async move {
-                        // need to clone K and bind it so rustc doesn't require `K: Sync`
-                        let k_for_loader = k_captured.clone();
+                        let submitter =
+                            ResultSubmitter::new(state_captured, k_captured.clone(), tag);
```
crepererum, line 103: Slight refactoring, this now also ensures that loader panics do NOT poison the affected cache key forever (`ResultSubmitter::drop` ensures that the active request is removed during panic).

cache_system/src/cache/mod.rs
```diff
@@ -27,6 +27,17 @@ pub enum CacheGetStatus {
     MissAlreadyLoading,
 }
 
+/// Status of a [`Cache`] [PEEK](Cache::peek_with_status) request.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CachePeekStatus {
+    /// The requested entry was present in the storage backend.
+    Hit,
+
```
alamb, line 35: it took me a while to figure out why there was no `Miss` status here (because the signature of `peek_with_status` will return `None`). Short of returning the actual possible value in a templated `CachePeekStatus<T>` I can't think of any better encoding.

alamb (approved):

cache_system/src/cache/mod.rs
crepererum: Yeah, I thought about this issue as well but decided that the status is rather a by-product than the actual return value so I decided against wrapping the actual value of interest into the status. But of course, that's an opinionated design.

alamb: I think this design is 👍

PR #5052, opened 2022-07-06
fix: avoid  combing groups that overlap with other groups  even if they are small

NGA-TRAN:
Closes #https://github.com/influxdata/conductor/issues/1013


NGA-TRAN:
The original code did "avoid combing groups that overlap with other groups even if they are small" but it missed to handle a corner case

PR #5056, opened 2022-07-06
fix: Correctly set up querier-to-ingester config, sequencers, and catalog for all-in-one ephemeral mode

### LINKED ISSUE #5022: Error while querying using all-in-one after querier -> ingester config change
From chat:

> Running with plain all-in-one is worse, it errors during the write query, not sure why:
> ```
> ➜  influxdb_iox git:(main) ✗ ./target/debug/influxdb_iox run all-in-one
> 2022-06-30T21:27:59.246813Z  WARN service_grpc_flight: err=Planning { source: External(GettingIngesterPartitions { 
> source: NoIngesterFoundForSequencer { sequencer_id: KafkaPartition(1) } }) } msg="Error handling Flight gRPC request"
```

> I wonder if all-in-one error is related to the change in https://github.com/influxdata/influxdb_iox/pull/4887
> If you revert prior to that PR does it work as expected?

> Reverting makes all-in-one work (:raised_hands:)

carols10cents:
Fixes #5022.

Since https://github.com/influxdata/influxdb_iox/pull/4887, the all-in-one ephemeral mode, used by some integration tests, has been intermittently failing.

The problem was that the all-in-one mode was assuming 1 Kafka partition/sequencer in most places, but if an ephemeral in-memory catalog was being used, `create_or_get_default_records` was making 2 sequencer records in the catalog. That meant the router was sharding using 1 shard, and the querier was sharding using the 2 shards in the catalog (but only had 1 sequencer -> ingester mapping configured). If the querier happened to shard the query to the same shard the router used, everything worked fine, but if the querier sharded to the other shard, the query would error.

The `create_or_get_default_records` method was also starting the Kafka Partition IDs at 1, when the rest of the all-in-one setup was assuming the IDs start at 0.

Most people using all-in-one mode locally, as well as production, are using a postgres catalog that's already set up, so we haven't been hitting these issues.

This PR also adds a test for this situation directly.

### DIFF
--- a/clap_blocks/src/catalog_dsn.rs
+++ b/clap_blocks/src/catalog_dsn.rs
@@ -175,7 +175,7 @@ impl CatalogDsnConfig {
                 let mem = MemCatalog::new(metrics);
 
                 let mut txn = mem.start_transaction().await.context(CatalogSnafu)?;
-                create_or_get_default_records(2, txn.deref_mut())
+                create_or_get_default_records(1, txn.deref_mut())
                     .await
                     .context(CatalogSnafu)?;
                 txn.commit().await.context(CatalogSnafu)?;
--- a/influxdb_iox/tests/end_to_end_cases/all_in_one.rs
+++ b/influxdb_iox/tests/end_to_end_cases/all_in_one.rs
@@ -17,7 +17,46 @@ async fn smoke() {
 
     // Set up all_in_one ====================================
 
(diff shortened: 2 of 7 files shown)


alamb (approved):
I tested all in one mode out locally for me and it works great. @lesam  perhaps you can test it out too

lesam:
Tested, works for me. Thanks!

PR #5058, opened 2022-07-06
HACK disable ingesting (NOT FOR MERGING)

alamb:
Related to https://github.com/influxdata/conductor/issues/971

As in biological experiments, I am disabling parts of IOx to see what happens to the ingesters performance

Specifically, I would like to understand the maximum speed the ingester can read  from kafka (and throw the data away) to see if applying some more parallelism to the problem is possible


alamb:
For the record, this approach did not appear to improve the performance of ingest very much -- see #5066  for a more promising approach

PR #5064, opened 2022-07-07
feat: `ReadBufferCache::peek`

crepererum:
For #5032.


querier/src/cache/read_buffer.rs
```diff
@@ -130,6 +130,12 @@ impl ReadBufferCache {
             )
             .await
     }
+
+    /// Get existing or "loading" read buffer chunk from cache.
```
alamb, line 134: It only returns the chunk if it is in the cache, right? If it is loading, `None` will be returned?

The comment implies to me it will return the chunk even if it is also still loading.

alamb (approved):

querier/src/cache/read_buffer.rs
crepererum: If it is actively loading (i.e. someone already requested it via `get`), then it will return `Some(...)`. This is the same functionality as introduced in #5049 and I decided that we are going to optimize of work/memory avoidance, not for the slight latency advantage. We can however change that later if we want (it shouldn't change anything in the code that uses this function).

alamb: As long as the comment is clear about what it is doing, I think the approach sounds good

PR #5069, opened 2022-07-07
docs: add consensus for the desired final output of the compactor

### LINKED ISSUE #5067: Consensus on what the desired final output of the compactor
# Rationale

This is a task of https://github.com/influxdata/conductor/issues/1023 to have consensus on what the desired final output of the compactor is and add it as comments in the main function of the `compact_partition`

This needs @pauldix to review:

The eventually desired output of a compacted partition is to include N non-overlapped files, each does not exceed a config param`desired_max_file_size_byte`.

We will adjust the setting `desired_max_file_size_byte` as we go to have:
  1. Good compaction performance. Which means
         i.  `Compact a partition` should finish in a reasonable time (TODO: quantify `reasonable time` in in seconds or minutes)
        ii. The Compactors catches up well with Ingesters (TODO: define what this means. For examples, if for some reason, the Compactors are down for several hours/days, when they come back on, how soon they bring the latest partitions to the desired output, how soon they bring the rest of the partitions to the desired output)
  2. Good query performance (TODO: define what good query performance means in the context of compaction such as max file size, number of non-overlapped files per partition, acceptance overlapped files or not) 

Note: the above consensus does not consider deletes that happens on cold partitions (aka partitions that have no ingesting data). This has lower priority and will be done in different (background) compaction cycle that needs new design.

Update: and we do hope N will be small (<= 10 or so)

NGA-TRAN:
Closes #5067 


compactor/src/compact.rs
```diff
@@ -521,12 +521,14 @@ impl Compactor {
         Ok(compact_and_upgrade)
     }
 
-    /// Runs compaction in a partition resolving any tombstones and compacting data so that parquet
-    /// files will be non-overlapping in time.
-    /// If the compacted result is too large, it will be plit into many non-overlapped files, each is
-    /// likely smaller than max_desired_file_size. There is still possibility the file sizes are larger
-    /// than the desired becasue the process to split the result is estimated using percentage size based
-    /// on the input time range
+    /// Runs compaction in a partition resolving any tombstones and deduplicating data
```
alamb, line 524: I recommend putting this in `compactor/README.md` so it is more easily discoverable and then referring to that file in these comments

alamb (approved):

compactor/src/compact.rs
```diff
@@ -521,12 +521,14 @@ impl Compactor {
...
+    ///
+    /// Expectation: After a partition of a table has not received any writes for some
+    /// amount of time, the compactor will ensure it is stored in object store as N parquet files which:
+    ///   . have non overlapping time ranges
+    ///   . each does not exceed a size specified by config param max_desired_file_size_byte.
```
alamb, line 529:
```suggestion
    ///   . each does not exceed a size specified by config param max_desired_file_size_bytes.
```

compactor/src/compact.rs
NGA-TRAN: Done. The readme is super simple now but I expect we will add more as we go

compactor/README.md
```diff
@@ -0,0 +1,3 @@
+After a partition of a table has not received any writes for some amount of time, the compactor will ensure it is stored in object store as N parquet files which:
```
alamb, line 1: We should ensure that @pauldix  agrees with this goal as well when he gets back

PR #5070, opened 2022-07-07
fix: Compact all partitions to 1 file

carols10cents:
Closes #5068

I see further simplifications we could make, but given that we'll probably be bringing some sort of splitting back, I decided to stop here.

This seems to be working? Are there any other logs/assertions that would be helpful for verifying this is working as expected?


NGA-TRAN:
Do you mean `compact all files of a partition into one file`? 😄

NGA-TRAN (approved):
Looks really nice. 
- I have gone over the tests and they look good.
- I really want keep `fn compute_split_time` and its tests because they are orthogonal and will be useful. Maybe move them to utils or something like that.
- I feel a bit heartbroken to see all the optimization code and tests deleted but I agree it is better its way.  We will be better to rewrite them (even similar) to add more sanity checks and be careful to recreate the tests from simple to sophisticated as needed. The current tests have grown way complicated with the optimizations.

Two things I think we can do in next PRs:
1. Remove unused flags to avoid confusions. I think they are `split_percentage`, `compaction_max_size_bytes`, `compaction_max_desired_file_size_bytes` and `compaction_max_file_count`.  I twill be a lot better to re-create them when needed. The `compaction_max_size_bytes` also exist in the IDPE repo that I do not know whether we have to delete as a consequence of deleting this. @alamb may know this better
2. To avoid hitting the same bug as https://github.com/influxdata/conductor/issues/1023 in the future, if we have to split and group overlapped files in next PRs, I suggest we add sanity checks right after reading the files and after each split/group. Something similar to the function `verify_contiguous_files` that was deleted in this PR

NGA-TRAN:
BTW,  after you address all the comments and merge this; and the merge test deployed, I will lock parquet_file table, remove all data and let the compactors run instead of shutting them down. I think I will have a way to play with the catalog to see how ingested data look like before compacting even though the compactors are  running.

carols10cents:
> * I feel a bit heartbroken to see all the optimization code and tests deleted

Git will keep them for us forever! ❤️

Issue #5037, opened 2022-07-05
Introduce chunk priority


crepererum:
# Why
For certain queries (mostly InfluxRPC) we scan table chunks until we find SOME specific data, e.g.:

https://github.com/influxdata/influxdb_iox/blob/16bd3e67c0330e88ba1c0c9b89da2657d3bb8a18/iox_query/src/frontend/influxrpc.rs#L244-L292

For that it would be helpful if chunks are scanned first that:

- have most accurate statistics
- have data already in-mem

This way we could avoid loading unnecessary data.

# What
Introduce a chunk priority and use it for queries.

# How
Add `QueryChunk::priority` which returns a new-type wrapper `ChunkPriority(i64)` (allows positive and negative values for better flexibility).

The querier assigns priorities as following (**descending** order):

1. ingester chunk
2. RB chunk
3. parquet chunk

This is somewhat related to #5032 where the priority then depends on the state of new fused chunk.

Note that the priority is different from the chunk order, because a chunk order usually depends on when the data was ingested, but there might be cases where old data was recently queried and therefore is available as RB and should be priorities.

The priority should then be used for the InfluxRPC scan plans that follow the "find SOME data" pattern.

PR #5075, opened 2022-07-08
refactor: try to query "cheap" chunks first

crepererum:
This should help a lot once #5032 is implemented. Currently it doesn't
really make a difference.

See #5037, which also proposes a more advanced but more complex system.
The team however agreed to try something simple first.


mkmik (approved):
it seems like a reasonable thing to do. LGTM

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1714,6 +1715,15 @@ fn make_selector_expr<'a>(
         .alias(col_name))
 }
 
+/// Orders chunks so it is likely that the ones that already have cached data are pulled first.
```
alamb, line 1718: 👍

alamb (approved):

Issue #5000, opened 2022-06-30
Improve Ingester speed to catch up / replay


tustvold:
__Steps to reproduce:__

When catching up from Kafka, the ingesters only manage ~1MB/s per Kafka partition

![image](https://user-images.githubusercontent.com/1781103/176664324-6962662d-7597-44a6-9225-d715e54cf0b7.png)

This is significantly below the 10-20 MB/s we were seeing with OG

The ingesters never appear to use more than 1 CPU, so it is possible there is lock contention somewhere

![image](https://user-images.githubusercontent.com/1781103/176663801-257d429b-36da-45f5-9207-ee73a0fd14cf.png)

Also looking at the sequence number graphs

![image](https://user-images.githubusercontent.com/1781103/176664690-ee8e44a0-f3bd-4843-ac9e-c890c3a651e4.png)

It would appear they are catching up ~16M messages. This might suggest that the aggregation in the routers is not working effectively, and we are ending up with a lot of very small message payloads.

__Expected behaviour:__

We should be able to catchup at many multiples of the steady state ingest speed. This would allow buffering more data in memory before persisting it, helping to reduce over-all system churn due to small parquet files

tustvold:
We also see the rate of consumption from individual partitions varying wildly during replay

![image](https://user-images.githubusercontent.com/1781103/176738443-101d8fca-5067-44c3-bef5-e62e4b34cca2.png)

![image](https://user-images.githubusercontent.com/1781103/176738498-85f5f1f1-09cc-4403-a6e3-11b9bb124f1b.png)

This would suggest to me that we are bottlenecking somewhere, as I would expect it to saturate the throughput for each partition

alamb:
Related to https://github.com/influxdata/conductor/issues/971

alamb:
I believe @tustvold  is working on this one so assigning to him

alamb:
What if one partition has most of the data writes (and thus dominates the max TTBR and limits the maximum amount of concurrency we can get -- aka even with 20 partitions the TTBR will be dominated by the single large partition). 

🤔 I wonder if we could pipeline the "read next message from kafka and decode it"  from the actual insertion. Or maybe even use more than one core if decoding the messages are the bottneck.

But that still doesn't seem to explain why only a single cor eis busy

alamb:
@tustvold  is working on https://github.com/apache/arrow-rs/issues/1764 which should also help significantly

alamb:
[...]

PR #5078, opened 2022-07-08
refactor: Improve code in `RSKafkaStreamHandler`

alamb:
re https://github.com/influxdata/influxdb_iox/issues/5000

# Rationale 

@shepmaster  had some great suggestions to improve the readability of the code from https://github.com/influxdata/influxdb_iox/pull/5066: https://github.com/influxdata/influxdb_iox/pull/5066#pullrequestreview-1033021217

# Changes
1. Implement changes from https://github.com/influxdata/influxdb_iox/pull/5066#pullrequestreview-1033021217 ❤️

### DIFF
--- a/write_buffer/src/kafka/mod.rs
+++ b/write_buffer/src/kafka/mod.rs
@@ -13,7 +13,7 @@ use crate::{
 use async_trait::async_trait;
 use data_types::{Sequence, SequenceNumber};
 use dml::{DmlMeta, DmlOperation};
-use futures::{stream::BoxStream, StreamExt};
+use futures::{stream::BoxStream, StreamExt, TryStreamExt};
 use iox_time::{Time, TimeProvider};
 use observability_deps::tracing::warn;
 use parking_lot::Mutex;
@@ -142,19 +142,13 @@ pub struct RSKafkaStreamHandler {
 /// there is an error decoding the data in the record, but not if
 /// there was an error reading the record in the first place.
 async fn try_decode(
-    record: Result<RecordAndOffset, WriteBufferError>,
+    record: RecordAndOffset,
     sequencer_id: u32,
     trace_collector: Option<Arc<dyn TraceCollector>>,
-) -> (Option<i64>, Result<DmlOperation, WriteBufferError>) {
-    let offset = match &record {
-        Ok(record) => Some(record.offset),
-        Err(_) => None,
-    };
-
+) -> Result<DmlOperation, WriteBufferError> {
     // launch a task to try and do the decode (which is CPU intensive)
     // in parallel
     let result = tokio::task::spawn(async move {
-        let record = record?;
         let kafka_read_size = record.record.approximate_size();
 
         let headers = IoxHeaders::from_headers(record.record.headers, trace_collector.as_ref())?;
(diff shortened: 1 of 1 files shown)


write_buffer/src/kafka/mod.rs
```diff
@@ -270,28 +258,39 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
...
+                    // kafka). This is to ensure that if a new stream is
+                    // created, we do not lose records that were never
+                    // consumed.
+                    //
+                    // Note that we update the offset as long as a record was
+                    // read (even if there was an error decoding) so we don't
+                    // get stuck on invalid records
                     *next_offset.lock() = Some(offset + 1);
+                    dml_result
                 }
```
crepererum, line 293: Why is this `and_then` + async now? I don't see any async operation within the async block, so `map` should be sufficient

alamb: I also found this confusing (and it is what required the somewhat tortured use of try_decode). The `dml_result` is a `Result` and then processing it can also return an error.

I found this easier to express with explicit types in the original formulation. @shepmaster  found a way to express it using `and_then`

Thus, in this formulation `and_then` is needed so a `Result` can be returned (`dml_result`) is itself a `Result`

It is async because `and_then` requires a future (as this is an async stream, not an iterator) https://docs.rs/futures/0.3.21/futures/stream/trait.TryStreamExt.html#method.and_then

Running
```diff
diff --git a/write_buffer/src/kafka/mod.rs b/write_buffer/src/kafka/mod.rs
index 9fa61ddf5..273843f9b 100644
--- a/write_buffer/src/kafka/mod.rs
+++ b/write_buffer/src/kafka/mod.rs
@@ -275,9 +275,8 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
             // the decode jobs in parallel
             // (`buffered` does NOT reorder, so the API user still gets an ordered stream)
             .buffered(CONCURRENT_DECODE_JOBS)
-            .and_then(move |(offset, dml_result)| {
+            .map(move |(offset, dml_result)| {
                 let next_offset = Arc::clone(&next_offset);
-                async move {
                     // but only update the offset when a decoded recorded
                     // is actually returned to the consumer of the stream
                     // (not when it was decoded or when it was read from
@@ -290,7 +289,6 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
                     // get stuck on invalid records
                     *next_offset.lock() = Some(offset + 1);
                     dml_result
-                }
             });
         stream.boxed()
     }
```

Results in `rustc` saying

```
cd /Users/alamb/Software/influxdb_iox && LOG_FILTER=debug,sqlx=warn,h2=warn,hyper=warn TEST_INTEGRATION=1 TEST_INFLUXDB_IOX_CATALOG_DSN=postgresql://localhost:5432/alamb RUST_BACKTRACE=1 CARGO_TARGET_DIR=/Users/alamb/Software/iox-target cargo test -p write_buffer
   Compiling write_buffer v0.1.0 (/Users/alamb/Software/influxdb_iox/write_buffer)
error[E0308]: mismatched types
   --> write_buffer/src/kafka/mod.rs:278:24
    |
278 |             .map(move |(offset, dml_result)| {
    |                        ^^^^^^^^^^^^^^^^^^^^
    |                        |
    |                        expected enum `std::result::Result`, found tuple
    |                        expected due to this
    |
    = note: expected enum `std::result::Result<(i64, std::result::Result<DmlOperation, WriteBufferError>), WriteBufferError>`
              found tuple `(_, _)`

```

alamb: I would also be fine with closing this PR and leaving the code as is if others think this makes things more confusing. I am ambivalent (though my obviously biased opinion is that this change makes things somewhat harder to reason about)

Issue #2188, opened 2021-08-04
User should be able to group by all tags (GROUP BY *)


pauldix:
Our SQL dialect should have an easy way for a user to specify that they want to `GROUP BY` all the tags in a table. Users frequently don't know all of the tags that exist ahead of time and it's a very common thing for them to want to group by all tags so we should make it easy rather than forcing them to go look up the tags and write them out individually. In InfluxQL this looked like this:

```
SELECT *
FROM foo
GROUP BY *
```

I'm not sure what the easiest way to add an extension to Datafusion to do this. Maybe it's a UDF? 

```sql
SELECT *
FROM foo
GROUP BY tags('foo')

-- or maybe simpler
SELECT *
FROM foo
GROUP BY tags()

-- or would this even parse?
SELECT *
FROM foo
GROUP BY foo.tags()
```

I think I'm hesitant to do it as a new keyword. Mainly because I don't think we want to modify the SQL parser.

alamb:
The way I would do this is as a rewrite pass prior to datafusion

So like
```
IOx Flavored SQL --> (parse into SQL AST) --> (do rewrite) --> (new SQL AST) --> pass to existing DataFusion planner
```

At the moment, there is no DataFusion extension mechanism I know of for adding a SQL rewrite phase, but I bet that people would welcome that extension (as they will have similar usecases to IOx wanting to support some flavor of SQL and then translating it into DataFusion supported SQL)

We could also do it ourselves (more quickly but likely with potential bugs)  in IOx by directly using `sqlparser` to parse the SQL, do the rewrite and then turn it back into a SQL string)

alamb:
> I think I'm hesitant to do it as a new keyword. Mainly because I don't think we want to modify the SQL parser.

I agree avoiding the need to modify the sql parser would be ideal

NGA-TRAN:
Sine we might support join in the future, I like the syntax  `tag(table1, table5)`

```
SELECT ...
FROM foo, bar, ...
GROUP BY tags(foo, bar)
```

Since this is IOx specific, I think the solution @alamb's suggestion is great
 > in IOx by directly using sqlparser to parse the SQL, do the rewrite and then turn it back into a SQL string

alamb:
FYI here is a datafusion issue that describes the desire for an extension mechanism: https://github.com/apache/arrow-datafusion/issues/533

alamb:
[...]

PR #5080, opened 2022-07-08
feat: allow sending an influxql query string to querier

lesam:
Closes #


alamb (commented):
👍  Let me know if you ever want to chat / brainstorm about how influxQL in IOx -- I have some thoughts 

  https://github.com/influxdata/influxdb_iox/issues/2188 is the largest open item I know if (there are likely others) but I have had thoughts here too

PR #5085, opened 2022-07-11
feat: `DelayedRecordBatchStream`

crepererum:
Required for #5032.

### DIFF
--- a/datafusion_util/src/delayed.rs
+++ b/datafusion_util/src/delayed.rs
@@ -1,0 +1,223 @@
+//! Tooling to delay certain async actions.
+
+use datafusion::{
+    arrow::{
+        datatypes::SchemaRef,
+        error::{ArrowError, Result as ArrowResult},
+        record_batch::RecordBatch,
+    },
+    physical_plan::{RecordBatchStream, SendableRecordBatchStream},
+};
+use futures::Stream;
+use pin_project::pin_project;
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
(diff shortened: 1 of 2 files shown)


datafusion_util/src/delayed.rs
```diff
@@ -0,0 +1,223 @@
...
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+};
+
+/// A [`RecordBatchStream`] that is produced by a future.
+///
+/// This can be used to delay some IO or other operations until the stream is polled.
+#[pin_project]
+pub struct DelayedRecordBatchStream<F>
```
alamb, line 24: `LazyRecordBatchStream` of `DeferredRecordBatchStream` would be other classic names for this pattern

alamb (approved):

datafusion_util/src/delayed.rs
crepererum: will rename

tustvold:
FWIW in DataFusion we use a combination of futures::stream::once, futures::stream::iter, try_flatten and RecordBatchStreamAdapter to avoid needing a custom abstraction

crepererum:
@tustvold I didn't find a way to construct a stream (using existing tooling) from a future that is polled when the stream is polled and returns an error or the actual stream.

tustvold:
@crepererum 

See https://github.com/apache/arrow-datafusion/pull/2428/files#diff-254beeb8f51dccfff404c87a3e5ebed831d62975a33e4464634c99181f2ec450R202

crepererum:
closing because @tustvold 's suggestions works pretty well

PR #5089, opened 2022-07-11
test: do NOT create expensive bloom filters when we do not need them

crepererum:
Esp. during the query tests this is a bit too expensive.


alamb:
>  IMHO they already do too much, not too little.

I have a different opinion, but the argument that this is related to metric reporting

alamb:
Specifically, the challenge of orchestrating the storage layouts in an end to end test would be quite hard

PR #5091, opened 2022-07-11
feat: load read buffer on demand

### LINKED ISSUE #5032: Lazy parquet chunk => RB chunk conversion

crepererum:
Closes #5032.


querier/src/chunk/mod.rs
```diff
@@ -452,62 +530,235 @@ pub mod tests {
...
+
+        // check state
+        assert_eq!(chunk.chunk_type(), "parquet");
 
         // measure catalog access
-        let catalog_metrics1 = get_catalog_access_metrics(&catalog.metric_registry());
+        let catalog_metrics1 = test_data.get_catalog_access_metrics();
 
         // check chunk schema
+        assert_schema(&chunk);
```
alamb, line 592: this is a very nice test structure

querier/src/chunk/query_access.rs
```diff
@@ -400,3 +455,40 @@ impl futures::Stream for ReadFilterResultsStream {
 
     // TODO is there a useful size_hint to pass?
 }
+
+enum OwnedSelection {
+    All,
+    Some(Vec<String>),
+}
+
+enum HalfOwnedSelection<'a> {
```
alamb, line 464: lol

querier/src/chunk/query_access.rs
```diff
@@ -400,3 +455,40 @@ impl futures::Stream for ReadFilterResultsStream {
 
     // TODO is there a useful size_hint to pass?
 }
+
+enum OwnedSelection {
```
alamb, line 459: eventually this might be nice to put into `schema` crate (along side `Selection`)

alamb (approved):

alamb:
This is really nice

parquet_file/src/storage.rs
```diff
@@ -226,35 +223,6 @@ impl ParquetStorage {
     }
 }
 
-/// Return indices of the schema's fields of the selection columns
-fn column_indices(selection: Selection<'_>, schema: SchemaRef) -> Vec<usize> {
```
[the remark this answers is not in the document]
crepererum: yes, see first commit

querier/src/chunk/mod.rs
```diff
@@ -125,6 +127,66 @@ impl ChunkStage {
...
+    fn from(rb_chunk: Arc<RBChunk>) -> Self {
+        let table_summary = Arc::new(rb_chunk.table_summary());
+        Self::ReadBuffer {
+            rb_chunk,
+            table_summary,
+        }
+    }
+}
+
+/// Controls if/how data for a [`QuerierChunk`] is loaded
```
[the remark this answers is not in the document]
alamb: Sounds reasonable to me

querier/src/chunk/query_access.rs
crepererum: done

PR #5093, opened 2022-07-11
feat: Record metric for number of files in a compaction by compaction level

### LINKED ISSUE #4901: Create metric for counting level-0 file compacted files
# Rationale

Currently, we keep track of number of both level-0 and level-1 files compacted for each partition in [compaction_counter and here](https://github.com/influxdata/influxdb_iox/blob/bf6fec24473e7627f45444371d2e806f93bd67a8/compactor/src/compact.rs#L548). Since we want to compare the speed of ingesting (level-0 files) with the speed of compacting those files, we want another counter to track only level-0 files compacted for each partition

carols10cents:
Closes #4901.

There are also two refactorings included:

1. Switch compaction level from using constants to an enum to get additional semantics, type checking, and exhaustiveness checking while still having a numerical value
2. Switch creating test parquet files from using combinatorial-explosion methods on test partition to using the builder pattern to support more flexibility and clearer documentation of what the numerical arguments represent in each test

After the refactorings is one commit where I added a metric for the number of files in a compaction with an attribute "compaction_level" for each observed compaction level, plus tests.


compactor/src/compact.rs
```diff
@@ -1464,12 +1506,37 @@ mod tests {
...
+        let level_non_overlapped_file_counter = metrics
+            .get_instrument::<Metric<U64Counter>>("compactor_compacted_files_total")
+            .unwrap()
+            .get_observer(&Attributes::from(&[
+                ("sequencer_id", "1"),
+                ("compaction_level", "2"),
+            ]))
+            .unwrap()
+            .fetch();
+        assert_eq!(level_non_overlapped_file_counter, 1);
```
NGA-TRAN, line 1539: 🎉

compactor/src/compact.rs
```diff
@@ -1583,15 +1653,25 @@ mod tests {
...
+            .with_file_size_bytes(140_000);
+        let parquet_file1 = partition.create_parquet_file(builder).await.parquet_file;
+
+        let builder = TestParquetFileBuilder::default()
+            .with_line_protocol(&lp2)
+            .with_min_seq(10)
+            .with_max_seq(15)
+            .with_min_time(6_000)
+            .with_max_time(25_000)
+            .with_file_size_bytes(100_000);
```
NGA-TRAN, line 1673: I tell myself I must remember to insert the underscore to make it easier to read next time

NGA-TRAN (approved):
Very cool. Thanks @carols10cents for also refactoring the test functions and the constants 🎉

Issue #5044, opened 2022-07-05
Remove min_sequence_number


NGA-TRAN:
# Rationale

@pauldix and I have talked last week and we decided keeping the range [min_sequence_number, max_sequence_number] is unnecessary and even causes some headache seem-like-a-bug (e.g. https://github.com/influxdata/conductor/issues/1013).

This needs a lot of changes in Ingester, Compactor and Querier.

# Steps:
 Since `min_sequence_number` can be null, the approach is to modify the code to neither use nor set this value. Then when the code is available, go and drop this column in a migration script

- [x] #5152
- [x] #5153
- [x] #5154

alamb:
I haven't thought through the implications of removing min_sequence_number

crepererum:
While I agree that the range itself might be unnecessary for the happy path, it might be a really valuable information for:

- sanity checking
- data lineage (which will be extremely helpful for debugging)

To me it looks like the bugs https://github.com/influxdata/conductor/issues/1013 clearly show that some invariant / state change wasn't understood or that we indeed have a bug. So just removing the value and sanity checks around it smells like YOLO-engineering :shrug: [^1]

[^1]: I'm clearly overplaying it here. I'm not suggesting that any team member is such a kind of engineer, just that the proposal w/o any further investigation sounds a bit bold.

NGA-TRAN:
@crepererum After going thru all scenarios I can think of, I agree with you and I do not (yet) see any value to remove `min_sequnce_number`. I actually found the bug of https://github.com/influxdata/conductor/issues/1013 and will add repro there.

I keep this one open because @pauldix suggested us to do so. Will close this after discussion with him

NGA-TRAN:
After some more discussion, we decided to remove `min_sequence_number` since it is not necessary at all and has introduced confusion so far

NGA-TRAN:
Done

PR #5101, opened 2022-07-12
refactor: use max_sequence_number to sort chunks for deduplication

### LINKED ISSUE #5099: Make chunk order by max sequence number
# Rationale

So far we use `min_sequence_number` to order chunks for deduplication in Ingester, Compactor and Querier. While we do not see the difference between using min or max so far, we highly suspect we will prefer `max_sequence_number` in the long run, especially for the Compactor.

NGA-TRAN:
Closes #5099


NGA-TRAN:
We have available correctness tests for this. The fact no tests fail means either `min_sequnce_number` and `max_sequnce_number` is good to choose for sorting the chunks for deduplication so far

alamb (approved):
This is fine with me, but I agree with @crepererum  that it doesn't seem to be critical.

Maybe a comment or two in the code explaining that the chunks must be ordered so that chunks with newer sequence numbers are ordered after chunks with older sequence numbers is needed for correctness and that the compactor ensures that it doesn't create chunks with overlapping keys (`time`) and sequence numbers

NGA-TRAN:
From our [conversation today](https://influxdata.slack.com/archives/CRK9M8L5Q/p1657641958221029?thread_ts=1657628314.725199&cid=CRK9M8L5Q), we will remove `min_sequence_number` for good in #5044 . Merging this PR before that happens

Issue #5105, opened 2022-07-12
Select partitions to compact first


NGA-TRAN:
# Rationale

Detailed design and implementation for `Selection of partition to compact first` of [Compaction Strategy & Algorithm](https://docs.google.com/document/d/1G4B0YlXyfBx6A_GLXM7FKneu_NE-TuHKkRovpa_XIaI/edit#heading=h.cpiu6bk26nwq)

# Design

This is how a cycle of a compaction looks like

-  For each **sequencer_id** the compactor manages
       - Select  **n** most recent highest ingested throughput partitions (1)
       - In parallel, run **m** `compact a partition` (2), where **m**  <=  **n** and **m** is computed  based on our available memory and the size of the selected partitions.

- For our first version, m = n = 1, which means we will compact everything in serial.

## (1) Select most recent highest ingested throughput partitions:

The below query will be used to select **n** most recent highest ingested throughput partitions. Since we have created an index on `(sequencer_id, compaction_level, to_delete)` this query runs quite fast, around 55ms on 5 or 6 days of data. We will create histogram metrics for this query

```SQL
select partition_id, count(id)
from parquet_file 
where compaction_level = 0 and to_delete is null
    and sequencer_id = seq_id
    and to_timestamp(created_at/1000000000) > now() - interval 'num_hours hour'
group by 1
having count(id) >= num_files
order by 2 DESC
limit n;
```

There are 4 parameters for this query:
- **seq_id**: id of the sequencer we want to compact
- **num_hours**: how much recent we want to measure the throughput. We will start with the last 4 (hours) and if we do not have data back, we will increase it to 24 hours. If still nothing, we will select top **n** partitions with the most level-0 files for that  sequencer and compact them.
- **num_files**: Minimum number of recent ingested files. The default is 2.
- **n**: Number of most recent highest ingested throughput partitions we want to consider to compact in parallel (note that the real number of partitions to run in parallel will be smaller than this value and depending on available memory). The default value is 1 which means we will run only one partition per sequencer in a cycle.

## (2) Compact a partition
See #5106

carols10cents:
> ```sql
> group by 1
> ...
> order by 2 DESC
> ```

What are 1 and 2 here?

carols10cents:
What should be logged when? Are there any metrics other than how long this query takes to run that would be helpful?

NGA-TRAN:
>> ```sql
>> group by 1
>> ...
>> order by 2 DESC
>> ```

[...]

PR #5108, opened 2022-07-12
feat: config params for Compactor

### LINKED ISSUE #5107: Create appropriate config params for Compaction
# Rationale

Create config params specified in [Compaction Strategy & Algorithm](https://docs.google.com/document/d/1G4B0YlXyfBx6A_GLXM7FKneu_NE-TuHKkRovpa_XIaI/edit#heading=h.cpiu6bk26nwq)

NGA-TRAN:
Closes #5107 

This PR:
  . Creates new config params defined in [this doc](https://docs.google.com/document/d/1G4B0YlXyfBx6A_GLXM7FKneu_NE-TuHKkRovpa_XIaI/edit#) 
  . Removes the old params that are no longer used


clap_blocks/src/compactor.rs
```diff
@@ -44,39 +81,18 @@ pub struct CompactorConfig {
     #[clap(
         long = "--compaction-concurrent-size-bytes",
         env = "INFLUXDB_IOX_COMPACTION_CONCURRENT_SIZE_BYTES",
-        default_value = "1000000000",
+        default_value = "100000000",
```
NGA-TRAN, line 84: I will modify this value in next PR of the work for #5105

clap_blocks/src/compactor.rs
```diff
@@ -27,15 +27,52 @@ pub struct CompactorConfig {
     )]
     pub write_buffer_partition_range_end: i32,
 
-    /// Percentage of least recent data we want to split to reduce compacting non-overlapped data
-    /// Must be between 0 and 100. Default is 100, which won't split the resulting file.
+    /// Max number of level-0 files we want to compact with level-1 each time
+    /// Default: 3
+    /// TODO: verify the theory that if the streaming works  as expected,
+    /// we do not need to limit this number
+    #[clap(
```
alamb, line 34: I would recommend putting this TODO note somewhere other than the docstring - when placed here it will appear in the actual message shown to users when they run `influxdb_iox run compactor --help`

Perhaps something like
```suggestion
    // TODO: verify the theory that if the streaming works  as expected,
    // we do not need to limit this number
    #[clap(
```

clap_blocks/src/compactor.rs
```diff
@@ -27,15 +27,52 @@ pub struct CompactorConfig {
     )]
     pub write_buffer_partition_range_end: i32,
 
-    /// Percentage of least recent data we want to split to reduce compacting non-overlapped data
-    /// Must be between 0 and 100. Default is 100, which won't split the resulting file.
+    /// Max number of level-0 files we want to compact with level-1 each time
+    /// Default: 3
```
alamb, line 31:
```suggestion
    /// Max number of level-0 files (written by ingester) to compact with level-1 each time
    /// Default: 3
```

clap_blocks/src/compactor.rs
```diff
@@ -27,15 +27,52 @@ pub struct CompactorConfig {
...
+        env = "INFLUXDB_IOX_COMPACTION_MAX_NUMBER_LEVEL_0_FILES",
+        default_value = "3",
+        action
+    )]
+    pub compaction_max_number_level_0_files: i32,
+
+    /// Desired max size that a compacted result won't be split and written into one file
+    /// It is a `desired` value because we decide whether to split or not based on
+    /// the data before compaction.
+    /// Default is 30,000,000 (30MB)
```
alamb, line 45:
```suggestion
    /// Desired max size of compacted parquet files 
    /// It is a target desired value rather than a guarantee.
    /// Default is 30,000,000 (30MB)
```

I am not sure if this rephrasing is correct; I didn't quite understand the  "won't be split and written into one file" (because if it is split, then wouldn't it be in *more* than one file?)

alamb (approved):
In general, if the goal is to simplify the compactor algorithm, 4 new configuration parameters seems somewhat at odds with that goal. 

I found the split size and percentage parameters quite confusing (as in I would have a hard time explaining how to set them for a particular desired output or explain how changes would affect the compactor operation).

I recommend considering deriving the values from `compaction-split-percentage` and `compaction-max-desired-size-bytes` as described on https://docs.google.com/document/d/1G4B0YlXyfBx6A_GLXM7FKneu_NE-TuHKkRovpa_XIaI/edit?disco=AAAAcrRxixE

clap_blocks/src/compactor.rs
```diff
@@ -27,57 +27,68 @@ pub struct CompactorConfig {
...
+    /// Desired max size of compacted parquet files
+    /// It is a target desired value than a guarantee
+    /// Default is 100,000,000 (100MB)
     #[clap(
-        long = "--compaction-concurrent-size-bytes",
-        env = "INFLUXDB_IOX_COMPACTION_CONCURRENT_SIZE_BYTES",
-        default_value = "1000000000",
+        long = "--compaction-max-desired-size-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_MAX_DESIRED_FILE_SIZE_BYTES",
+        default_value = "100000000",
```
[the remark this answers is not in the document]
NGA-TRAN: @carols10cents I would say yes with your more accurate number. So far, we do not strictly make it the case but I agree we should

alamb (commented):
❤️

PR #5113, opened 2022-07-13
fix: fix overeager Kafka message flushing

crepererum:
I think @tustvold spotted that one, so thank you! :)

The theory is that this will improve kafka message processing throughput -- re https://github.com/influxdata/conductor/issues/971


tustvold (approved):
No idea if this will help in practice, but worth a shot. The ideal would be per-partition lingers and flush triggers, but that is probably out of scope

PR #5114, opened 2022-07-13
chore: update all dependencies (esp. `time`)

crepererum:
Our CI found a potential panic:
https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/21307/workflows/7c1df458-3c6a-4f5d-bfda-51612273bbc7/jobs/183140

See https://github.com/time-rs/time/issues/481 , so let's update
`time` and since we've accumulated quite a backlog, just run
`cargo update --workspace`.


alamb (approved):

crepererum:
> Hm, I wonder why dependabot hasn't been handling these updates...

My theory was that there are too many updates per day and it never catches up, but thinking about this it cannot be right because we've seen many days w/o dependabot PRs. So this might be some bug.

PR #5115, opened 2022-07-13
feat: skip ingester buffering if INFLUXDB_IOX_INGESTER_SKIP_BUFFER is set

alamb:
re https://github.com/influxdata/conductor/issues/971 and https://github.com/influxdata/conductor/issues/1034

Basically, I would like to run the experiment on production, described in  https://github.com/influxdata/conductor/issues/1034  to isolate where the bottleneck in replay is: reading from kafka or something in the processing of messages afterwards.

Closes https://github.com/influxdata/influxdb_iox/pull/5104

# Changes

When the `INFLUXDB_IOX_INGESTER_SKIP_BUFFER` environment is set to `true` the ingester will decode data from kafka and then just throw it away.

For example:
```shell
INFLUXDB_IOX_INGESTER_SKIP_BUFFER=true influxdb_iox run ingester
```

# Testing
The "don't skip" path is well tested in the existing tests. I manually tested the "do skip" path

When I tested this manually, by setting `INFLUXDB_IOX_INGESTER_SKIP_BUFFER=true` on my performance machine,  I saw the log messages as expected 

```
2022-07-13T14:09:26.947291Z  WARN ingester::stream_handler::sink_adaptor: Skipping buffering due to environment request name=INFLUXDB_IOX_INGESTER_SKIP_BUFFER value=true
2022-07-13T14:09:26.947315Z  WARN ingester::stream_handler::sink_adaptor: Skipping buffering due to environment request name=INFLUXDB_IOX_INGESTER_SKIP_BUFFER value=true
2022-07-13T14:09:26.947338Z  WARN ingester::stream_handler::sink_adaptor: Skipping buffering due to environment request name=INFLUXDB_IOX_INGESTER_SKIP_BUFFER value=true
2022-07-13T14:09:26.947383Z  WARN ingester::stream_handler::sink_adaptor: Skipping buffering due to environment request name=INFLUXDB_IOX_INGESTER_SKIP_BUFFER value=true
```


alamb:
I should have noted it in the description originally, but setting `INFLUXDB_IOX_INGESTER_SKIP_BUFFER` will also not update the `min_persisted_sequence_number` (as that is done as part of the persist logic, which is totally skipped)

Thus setting this in production will not cause any permanent data loss -- rather it will simply cease to make any data read from kafka visible. When we restart the ingester without INFLUXDB_IOX_INGESTER_SKIP_BUFFER, the previously ingested data will be read and persisted as normal

PR #5124, opened 2022-07-13
feat: Gather parquet files for a partition compaction operation

### LINKED ISSUE #5118: Given a candidate partition for compaction, gather all parquet files and organize them
- Get all the files in the partition: `select * from parquet_file where partition_id = $`
  - Use the existing `iox_catalog` [`list_by_partition_not_to_delete` method](https://github.com/influxdata/influxdb_iox/blob/89c24dfec0c3316a7c34306722dc2b9f46e4260f/iox_catalog/src/interface.rs#L548)
- In memory, separate out the files into two collections:
  - L0 files ordered by max_sequence_number asc
  - L1 files (their order isn’t important as they’re non-overlapping)
- Log... the partition ID and the number of files in each collection? Anything else?
- Would the number of files in each collection be a helpful metric? Anything else?

carols10cents:
Fixes #5118.

Given a partition ID, look up the non-deleted Parquet files for that
partition. Separate them into level 0 and level 1, and sort the level 0
files by max sequence number.

🚨🚨 This is not called anywhere yet. 🚨🚨

### DIFF
--- a/compactor/src/parquet_file_lookup.rs
+++ b/compactor/src/parquet_file_lookup.rs
@@ -1,0 +1,317 @@
+//! Logic for finding relevant Parquet files in the catalog to be considered during a compaction
+//! operation.
+
+use data_types::{CompactionLevel, ParquetFile, PartitionId};
+use iox_catalog::interface::Catalog;
+use observability_deps::tracing::*;
+use snafu::{ResultExt, Snafu};
+use std::sync::Arc;
+
+#[derive(Debug, Snafu)]
+#[allow(missing_copy_implementations, missing_docs)]
+pub(crate) enum PartitionFilesFromPartitionError {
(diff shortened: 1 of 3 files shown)


compactor/src/parquet_file_lookup.rs
```diff
@@ -0,0 +1,317 @@
...
+            .create_partition("2022-07-12")
+            .await;
+
+        TestSetup {
+            catalog,
+            partition,
+            partition_on_another_sequencer,
+            older_partition,
+        }
+    }
```
NGA-TRAN, line 125: I like this test setup 👍

NGA-TRAN (approved):
Very nice test setup. I have reviewed the tests carefully. Great coverage.

PR #5127, opened 2022-07-14
refactor: ingester seek log debug => info

crepererum:
This message will be printed once per partition on ingester startup and
shouldn't be too noisy, but is very helpful to judge "replay" /
"catch-up".


mkmik:
CC: @primeroz

alamb (commented):
👍

Issue #5129, opened 2022-07-14
Querier tracing improvements


crepererum:
# Status Quo
I look at all our metrics and see the gRPC end2end latencies and I can look at all the sub-metrics like semaphore acquire times and cache requests, but I have a really hard time to add things up and I feel I'm missing something.

The right instrument here would be tracing, however the span we currently produce is quite lacking:

![image](https://user-images.githubusercontent.com/1529400/178978395-5eb51bf9-1e08-45ce-9e8a-cfbf1137d7cf.png)

The reason it is lacking is rather simple: We've invested a lot of effort into instrumenting the query engine itself and this was mostly sufficient for OG where chunk gathering was basically instantaneous, but with NG we need to perform a bit more work to have a complete _picture_ of what we want to query. This chunk gathering (which BTW also includes the ingester gRPC request) is currently NOT included in the trace.

# ACs
Have spans for:

- [x] the semaphore wait
- [x] every single no-hit (i.e. "miss" and "already loading") cache request (`querier::cache`, potentially with proper support from `cache_system`)
- [x] every single ingester request

This ticket does NOT include span propagation to the ingester. We'll later see if this is required/helpful as well.

crepererum:
After #5130 we can now clearly see the semaphore:

![image](https://user-images.githubusercontent.com/1529400/179203656-c023c955-0bb3-43f4-b27d-ab834727d707.png)

In this example, the semaphore is clearly NOT an issue.

crepererum:
With #5134 we now clearly see the table chunk accesses.

Here one example where a single chunk enumeration dominates the query time:
![Screenshot from 2022-07-20 09-53-45](https://user-images.githubusercontent.com/1529400/179928331-80a8462c-7fed-4f28-bd17-7938b61463cf.png)

And here one example were we perform multiple chunk enumerations in a row:
![Screenshot from 2022-07-20 09-55-10](https://user-images.githubusercontent.com/1529400/179928762-449b5845-87b6-43b4-9aac-81ca323face9.png)

crepererum:
I consider this fixed. Let's have concrete follow-up ticket if that's required.

PR #5130, opened 2022-07-14
feat: add tracing to instrumented semaphores

crepererum:
This will allow us to easily see how much time we spend during query
processing waiting for the query semaphore.

Ref #5129.


service_grpc_flight/src/lib.rs
```diff
@@ -197,7 +198,10 @@ where
             }
         };
 
-        let permit = self.server.acquire_semaphore().await;
+        let permit = self
+            .server
+            .acquire_semaphore(span_ctx.as_ref().map(|span| span.child("query semaphore")))
```
alamb, line 203:
```suggestion
            .acquire_semaphore(span_ctx.as_ref().map(|span| span.child("query rate limit semaphore")))
```

tracker/src/async_semaphore.rs
```diff
@@ -319,6 +347,12 @@ pub struct InstrumentedAsyncOwnedSemaphorePermit {
 
     /// Metrics.
     metrics: Arc<AsyncSemaphoreMetrics>,
+
+    /// Span recorder for the entire semaphore interaction.
+    ///
+    /// No direct interaction, will be exported during drop.
+    #[allow(dead_code)]
```
alamb, line 354: is this really needed as `this.span_recorder` is used in the implementation of `Poll` above

alamb (approved):

workspace-hack/Cargo.toml
```diff
@@ -122,31 +122,18 @@ tokio-stream = { version = "0.1", features = ["fs", "net", "time"] }
 uuid = { version = "1", features = ["private_getrandom", "rng", "std", "v4"] }
 
 [target.x86_64-unknown-linux-gnu.dependencies]
-getrandom = { version = "0.2", default-features = false, features = ["std"] }
```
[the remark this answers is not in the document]
crepererum: Me neither :D

tracker/src/async_semaphore.rs
crepererum: Yes, because the export-on-drop will also set the end of the span. So you the happy-path span is `<start>....<acquired>...<end>` (the span has start+end and an event in the _middle_), the very unlikely unhappy path (I've never seen acquire failing) `<start>....<error+end>`.

service_grpc_flight/src/lib.rs
crepererum: done (+ others)

tracker/src/async_semaphore.rs
crepererum: extended docs

Issue #5106, opened 2022-07-12
Compact a partition


NGA-TRAN:
# Rationale

Design the details and implement `With the partition we’re going to compact` in the [Compaction Strategy & Algorithm](https://docs.google.com/document/d/1G4B0YlXyfBx6A_GLXM7FKneu_NE-TuHKkRovpa_XIaI/edit#heading=h.cpiu6bk26nwq) 

# Subtasks

- [x] https://github.com/influxdata/influxdb_iox/issues/5119
- [x] https://github.com/influxdata/influxdb_iox/issues/5118
- [x] https://github.com/influxdata/influxdb_iox/issues/5120
- [x] influxdata/influxdb_iox#5174
- [x] https://github.com/influxdata/influxdb_iox/issues/5121
- [x] https://github.com/influxdata/influxdb_iox/issues/5123
- [ ] influxdata/influxdb_iox#5172
- [x] influxdata/conductor#1078

NGA-TRAN:
Let us close this because delete is not in our near todo

PR #5131, opened 2022-07-14
feat: Select compaction candidates

### LINKED ISSUE #5105: Select partitions to compact first

NGA-TRAN:
Closes #5105

This PR includes:
  1. The new implementation of `partitions_to_compact` that will lead to the needs for stuff in points 2
  2. Implement 2 catalog queries  `recent_highest_throughput_partitions` and `most_level_0_files_partitions`  that are a source of long tests.
  3. Simplify `run_compactor` to make the compaction run in serial until everything works work. 
  4. New related metrics. 


compactor/src/compact.rs
```diff
@@ -225,11 +238,16 @@ pub struct Compactor {
     /// Gauge for the number of compaction partition candidates
     compaction_candidate_gauge: Metric<U64Gauge>,
 
-    /// Gauge for the number of bytes of level-0 files across compaction partition candidates
-    compaction_candidate_bytes_gauge: Metric<U64Gauge>,
-
```
NGA-TRAN, line 230: Per new design, we do not have this value anymore  because we do not care how large all L-0 files are. We only pick to a few L0 to compact. As part of #5106, @carols10cents has planed to add more appropriate metrics.

compactor/src/compact.rs
```diff
@@ -225,11 +238,16 @@ pub struct Compactor {
...
     /// Histogram for tracking the time to compact a partition
     compaction_duration: Metric<DurationHistogram>,
+
+    /// Histogram for tracking the time of a catalog query to get
+    /// recent highest throughput partitions
+    compaction_query_highest_throughput_duration: Metric<DurationHistogram>,
+
+    /// Histogram for tracking the time of a catalog query to get
+    /// most level 0 files partitions
+    compaction_query_most_l0_duration: Metric<DurationHistogram>,
```
NGA-TRAN, line 250: These 2 queries are only 55ms for now but since we have  a concern they will grow, add metrics for them

compactor/src/compact.rs
```diff
@@ -273,6 +287,18 @@ impl Compactor {
             "Compact partition duration",
         );
 
+        let compaction_query_highest_throughput_duration: Metric<DurationHistogram> = registry
+            .register_metric(
+                "compaction_query_highest_throughput_duration",
+                "Duration of catalog query that gets recent highest ingested throughput partitions",
```
NGA-TRAN, line 293:
```suggestion
                "Duration of a catalog query that gets the most recent highest ingested throughput partitions",
```

compactor/src/compact.rs
```diff
@@ -273,6 +287,18 @@ impl Compactor {
...
+            );
+
+        let compaction_query_most_l0_duration: Metric<DurationHistogram> = registry
+            .register_metric(
+                "compaction_query_most_l0__duration",
+                "Duration of catalog query that gets partitions with the most l0 files",
```
NGA-TRAN, line 299:
```suggestion
                "Duration of a catalog query that gets partitions with the most level-0 files",
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,105 @@ impl Compactor {
...
 
-    /// Returns a list of partitions that have level 0 files to compact along with some summary
-    /// statistics that the scheduler can use to decide which partitions to prioritize. Orders
-    /// them by the number of level 0 files and then size.
-    pub async fn partitions_to_compact(&self) -> Result<Vec<PartitionCompactionCandidate>> {
-        let mut candidates = vec![];
+    /// Return a list of the most recent highest ingested throughput partitions.
+    /// The highest throughput partitions are prioritized as follows:
+    ///  1. If there are partitions with new writes within the last 4 hours, pick them.
+    ///     The one with the most files will be first in the list.
```
NGA-TRAN, line 346:
```suggestion
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,105 @@ impl Compactor {
...
+    ///  3. If there are no writes within the last 24 hours, will look for partitions
+    ///     with any new writes in the past.
+    ///
+    /// . New writes means non-deleted L0 files
+    /// . In all cases above, for each sequencer, N partitions with the most new writes will be selected
+    ///     and the return list will include at most, P = N * S, partitions where S is the number of
+    ///     sequencers this compactor handles.
+    pub async fn partitions_to_compact(
+        &self,
+        // Number of most recent highest ingested throughput partitions
```
NGA-TRAN, line 358:
```suggestion
        //  Max number of the most recent highest ingested throughput partitions
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,105 @@ impl Compactor {
...
+        // per sequencer we want to read
+        num_partitions_per_sequencer: i32,
```
NGA-TRAN, line 360:
```suggestion
        max_num_partitions_per_sequencer: i32,
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,105 @@ impl Compactor {
...
+        // Minimum number of most recent writes per partition we want to count
```
NGA-TRAN, line 361:
```suggestion
        // Minimum number of the most recent writes per partition we want to count
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,105 @@ impl Compactor {
...
+            }
 
-        candidates.sort_by(
-            |a, b| match b.level_0_file_count.cmp(&a.level_0_file_count) {
-                Ordering::Equal => b.file_size_bytes.cmp(&a.file_size_bytes),
-                o => o,
-            },
-        );
+            // No active ingesting partitions the last 24 hours,
+            // get partition with the most level-0 file
```
NGA-TRAN, line 406:
```suggestion
            // get partition with the most level-0 files
```

compactor/src/compact.rs
```diff
@@ -1059,31 +1113,11 @@ impl Compactor {
...
-    /// namespace ID
-    pub namespace_id: NamespaceId,
-    /// the number of level 0 files in the partition
-    pub level_0_file_count: usize,
-    /// the total bytes of the level 0 files to compact
-    pub file_size_bytes: i64,
-    /// the created_at time of the oldest level 0 file in the partition
-    pub oldest_file: Timestamp,
-}
-
```
NGA-TRAN, line 1081: This struct includes fields we no longer need. The PartitionParam is used instead

compactor/src/compact.rs
```diff
@@ -273,6 +287,18 @@ impl Compactor {
...
 
+        let compaction_query_highest_throughput_duration: Metric<DurationHistogram> = registry
+            .register_metric(
+                "compaction_query_highest_throughput_duration",
+                "Duration of a catalog query that gets the most recent highest ingested throughput partitions",
+            );
+
+        let compaction_query_most_l0_duration: Metric<DurationHistogram> = registry
+            .register_metric(
+                "compaction_query_most_l0__duration",
```
carols10cents, line 298: It looks like there's an extra underscore in this metric name? 

```suggestion
                "compaction_query_most_l0_duration",
```

compactor/src/compact.rs
```diff
@@ -324,67 +340,104 @@ impl Compactor {
...
     }
 
-    /// Returns a list of partitions that have level 0 files to compact along with some summary
-    /// statistics that the scheduler can use to decide which partitions to prioritize. Orders
-    /// them by the number of level 0 files and then size.
-    pub async fn partitions_to_compact(&self) -> Result<Vec<PartitionCompactionCandidate>> {
-        let mut candidates = vec![];
+    /// Return a list of the most recent highest ingested throughput partitions.
+    /// The highest throughput partitions are prioritized as follows:
+    ///  1. If there are partitions with new writes within the last 4 hours, pick them.
```
carols10cents, line 345: Also through here, I think this "write" means "write of a parquet file to object storage" rather than `DmlWrite`:

```suggestion
    ///  1. If there are partitions with new persisted files within the last 4 hours, pick them.
```

pauldix (approved):
Does what we talked about for partition selection. I'll let others comment on Rust style & code specifics.

compactor/src/compact.rs
```diff
@@ -2229,10 +2264,12 @@ mod tests {
         txn.commit().await.unwrap();
     }
 
+    #[ignore]
```
carols10cents, line 2267: Why is this test ignored?

iox_catalog/src/mem.rs
```diff
@@ -1102,6 +1109,106 @@ impl ParquetFileRepo for MemTxn {
...
+            .filter(|(_, v)| v >= &&min_num_files)
+            .collect::<Vec<_>>();
+
+        // Sort partitions by file count
+        partitions.sort_by(|a, b| b.1.cmp(a.1));
+
+        let mut partitions = partitions.iter().map(|(k, _)| **k).collect::<Vec<_>>();
+
+        // Return top partitions with most file counts
+        partitions.truncate(num_partitions as usize);
```
carols10cents, line 1164: This can be:

```suggestion
        let partitions = partitions.into_iter().map(|(k, _)| k).take(num_partitions as usize).collect::<Vec<_>>();
```

iox_catalog/src/mem.rs
```diff
@@ -1102,6 +1109,106 @@ impl ParquetFileRepo for MemTxn {
...
+        // Count num of files per partition by simply count the number of partition duplicates
+        let mut partition_duplicate_count: HashMap<PartitionParam, i32> =
+            HashMap::with_capacity(partitions.len());
+        for p in partitions {
+            let count = partition_duplicate_count.entry(p).or_insert(0);
+            *count += 1;
+        }
+
+        // Sort partitions by file count
+        // Partitions with select file count > min_num_files
```
carols10cents, line 1200: I don't think this comment applies here because there isn't a `min_num_files` in the `most_level_0_files_partitions` query?

iox_catalog/src/mem.rs
```diff
@@ -1102,6 +1109,106 @@ impl ParquetFileRepo for MemTxn {
...
+        let mut partitions = partition_duplicate_count.iter().collect::<Vec<_>>();
+        partitions.sort_by(|a, b| b.1.cmp(a.1));
+
+        let mut partitions = partitions.iter().map(|(k, _)| **k).collect::<Vec<_>>();
+
+        // Return top partitions with most file counts
+        partitions.truncate(num_partitions as usize);
```
carols10cents, line 1207: Same here with `take` rather than `truncate`

iox_catalog/src/metrics.rs
```diff
@@ -278,6 +278,8 @@ decorate!(
         "parquet_count" = count(&mut self) -> Result<i64>;
         "parquet_count_by_overlaps" = count_by_overlaps(&mut self, table_id: TableId, sequencer_id: SequencerId, min_time: Timestamp, max_time: Timestamp, sequence_number: SequenceNumber) -> Result<i64>;
         "parquet_get_by_object_store_id" = get_by_object_store_id(&mut self, object_store_id: Uuid) -> Result<Option<ParquetFile>>;
+        "recent_highest_throughput_partitions" = recent_highest_throughput_partitions(&mut self, sequencer_id: SequencerId, num_hours: i32, min_num_files: i32, num_partitions: i32 ) -> Result<Vec<PartitionParam>>;
+        "most_level_0_files_partitions" =  most_level_0_files_partitions(&mut self, sequencer_id: SequencerId, num_partitions: i32) -> Result<Vec<PartitionParam>>;
```
carols10cents, line 282: Wait a second, these metrics wrappers are automatically making duration metrics for these catalog operations, right? So the metrics in compactor are recording the same data and aren't really needed?

iox_catalog/src/postgres.rs
```diff
@@ -1678,6 +1679,64 @@ WHERE parquet_file.sequencer_id = $1
...
+        // We have index on (sequencer_id, comapction_level, to_delete)
+        // If this query happens to be a lot slower (>500ms), we might think to add
+        // and index on (sequencer_id, comapction_level, to_delete, created_at)
+        sqlx::query_as::<_, PartitionParam>(
+            r#"
+SELECT partition_id, sequencer_id, namespace_id, table_id, count(id)
+FROM parquet_file 
+WHERE compaction_level = 0 and to_delete is null
+    and sequencer_id = $1
+    and to_timestamp(created_at/1000000000) > now() -  ($2 || 'hour')::interval
```
carols10cents, line 1699: If this query ends up being slow, the first thing I would look at is if it's possible for us to compute the value that `created_at` should be compared to in Rust and pass it once rather than having Postgres do the math *and* the comparison

carols10cents (changes requested):

compactor/src/compact.rs
NGA-TRAN: Will replace `writes` with `ingested files`

compactor/src/compact.rs
```diff
@@ -324,67 +340,104 @@ impl Compactor {
...
+    ///     and the return list will include at most, P = N * S, partitions where S is the number of
+    ///     sequencers this compactor handles.
+    pub async fn partitions_to_compact(
+        &self,
+        //  Max number of the most recent highest ingested throughput partitions
+        // per sequencer we want to read
+        max_num_partitions_per_sequencer: i32,
+        // Minimum number of the most recent writes per partition we want to count
+        // to prioritize partitions
+        minimun_recent_ingested_files: i32,
```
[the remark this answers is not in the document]
NGA-TRAN: The Postgres function `recent_highest_throughput_partitions` require  integer. We can either cast i32 to usize or usize to i32

compactor/src/compact.rs
NGA-TRAN: Good catch. Should not 😄

compactor/src/compact.rs
```diff
@@ -2272,12 +2309,40 @@ mod tests {
             .create_or_get("four".into(), sequencer.id, table.id)
             .await
             .unwrap();
+        txn.commit().await.unwrap();
+
+        // Create a compactor
+        let config = make_compactor_config();
+        let compactor = Compactor::new(
+            vec![sequencer.id],
```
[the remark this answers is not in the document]
NGA-TRAN: will do

compactor/src/handler.rs
```diff
@@ -224,67 +249,36 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
                 }
                 Ok(compact_and_upgrade) => {
                     if compact_and_upgrade.compactable() {
```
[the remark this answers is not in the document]
NGA-TRAN: No, we do not need them. When the new `compaction_partition` available, we should not use these anymore

iox_catalog/src/mem.rs
NGA-TRAN: Good catch

iox_catalog/src/metrics.rs
NGA-TRAN: This is interesting. I will need to read the code a bit more and see if it does what we want to remove other code

iox_catalog/src/postgres.rs
NGA-TRAN: I do not think the concern is the work of Postgres doing the math. The concern is there are a lot of data that lead to a lot of comparison. 
If we compute something here, we can only compute the value for `now() -  ($2 || 'hour')::interval`. We do not know `created_at` and there are a lot of rows of data with different values of `created_at`.
Postgres will rewrite this query to compute `now() -  ($2 || 'hour')::interval` once and reuse for all rows. It is actually very cheap compared with the comparison work

clap_blocks/src/compactor.rs
```diff
@@ -91,4 +91,24 @@ pub struct CompactorConfig {
...
+
+    /// Max number of partitions per sequencer we want to compact per cycle
+    /// Default: 1
+    #[clap(
+        long = "--compaction-max-number-partitions-per-sequencer",
+        env = "INFLUXDB_IOX_COMPACTION_MAX_NUMBER_PARTITION_PER_SEQUENCER",
+        default_value = "1",
+        action
+    )]
+    pub compaction_max_number_partitions_per_sequencer: i32,
```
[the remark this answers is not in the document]
NGA-TRAN: The `COMPACTION` is helpful when I search `INFLUXDB_IOX_COMPACTION`. There are many `INFLUXDB_IOX_` environment flags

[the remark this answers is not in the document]
carols10cents: Yes, the environment variable should have compaction in it, but the field name doesn't need it

iox_catalog/src/metrics.rs
NGA-TRAN: You are right. I am going remove the other metrics

compactor/src/compact.rs
NGA-TRAN: I will address this next. Somehow the time_provide does not match yet

NGA-TRAN (commented):
I have addressed all but one. Will do that next before asking for re-reviewing

NGA-TRAN:
@carols10cents It is ready for second round

clap_blocks/src/compactor.rs
```diff
@@ -91,4 +91,24 @@ pub struct CompactorConfig {
...
+        default_value = "1",
+        action
+    )]
+    pub max_number_partitions_per_sequencer: i32,
+
+    /// Min number of recent writes a partition needs to be considered for compacting
+    /// Default: 1
+    #[clap(
+        long = "--compaction-min-number-recent-writes-per-partition",
+        env = "INFLUXDB_IOX_COMPACTION_MAX_NUMBER_PARTITION_PER_SEQUENCER",
```
carols10cents, line 109: The flag is named "compaction min number recent writes per partition", the environment variable is named "compaction *max* number partition per *sequencer*, and the struct field is named "min number recent ingested files per partition", I think these should all be consistent?

compactor/src/compact.rs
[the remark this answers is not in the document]
carols10cents: It seems to me that casting in one place (in the postgres implementation of that function) is better than casting to usize in multiple places?

iox_catalog/src/postgres.rs
carols10cents: > We do not know created_at and there are a lot of rows of data with different values of created_at.

Right, I was saying compute the value that `created_at` *should be compared to*. That is, don't convert created_at `to_timestamp`, instead compute `now_nanos() - 2_hours_in_nanoseconds` in Rust, once, and make the condition here `created_at > $2` where ` $2 is the Rust-computed value of 2 hours ago in nanoseconds, then Postgres can do an integer comparison rather than a timestamp comparison (which my gut says should be faster). Again, though, I don't see a reason to do this unless we observe this query to be slow.

clap_blocks/src/compactor.rs
```diff
@@ -91,4 +91,24 @@ pub struct CompactorConfig {
...
+
+    /// Min number of recent ingested files a partition needs to be considered for compacting
+    /// Default: 1
+    #[clap(
+        long = "--compaction-min-number-recent-ingested-files-per-partition",
+        env = "INFLUXDB_IOX_COMPACTION_MIN_RECENT_NUMBER_INGESTED_FILES_PER_PARTITION",
+        default_value = "1",
+        action
+    )]
+    pub min_number_recent_ingested_files_per_partition: i32,
```
carols10cents, line 113: The flag and field name are still different from the environment variable? Flag and field name are "min number recent", the environment variable is "min recent number".

compactor/src/compact.rs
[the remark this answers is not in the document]
NGA-TRAN: In addtition to these 2 values, I have made all relevant iSomethign in this struct  to uSomething

iox_catalog/src/postgres.rs
NGA-TRAN: It is actually no that simple. Let us keep this for now. Only do this if  we hit the issue

PR #5134, opened 2022-07-15
feat: wire span all the way to `QuerierTable::chunks`

crepererum:
This is required for #5129.


iox_query/src/exec/context.rs
```diff
@@ -281,6 +301,28 @@ impl fmt::Debug for IOxSessionContext {
...
+
+        let recorder = Arc::new(Mutex::new(recorder));
+        SPAN_RECORDERS.lock().insert(id, Arc::clone(&recorder));
+
+        inner
+            .state
+            .write()
+            .config
+            .config_options
+            .set_u64(CONFIG_OPTIONS_KEY, id);
```
crepererum, line 316: This is a bit messy, but DataFusion doesn't allow us to pass opaque object throw the stack (or at least I didn't find another way), so we use an ID + a global lookup table. Table entries are removed on drop, so this should be rather clean.

compactor/src/query.rs
```diff
@@ -219,7 +219,7 @@ impl QueryChunk for QueryableParquetChunk {
     /// streams from several different `QueryChunk`s.
     fn read_filter(
         &self,
-        mut ctx: IOxSessionContext,
+        ctx: IOxSessionContext,
```
alamb, line 222: 👍

iox_query/src/frontend/influxrpc.rs
```diff
@@ -243,7 +243,7 @@ impl InfluxRpcPlanner {
             // Identify which chunks can answer from its metadata and then record its table,
             // and which chunks needs full plan and group them into their table
             let chunks = database
-                .chunks(table_name, predicate)
+                .chunks(table_name, predicate, ctx.child_ctx("table chunks"))
```
alamb, line 246: 👍  this makes sense

alamb (commented):
I am confused about the global map of `SpanRecorders` -- I thought the idea was that each query runs with its own `IOxSessionContext` (which was attached to the `SpanContext` for the request)

Otherwise the plumbing in this PR looks good to me

iox_query/src/exec/context.rs
```diff
@@ -180,6 +189,18 @@ impl fmt::Debug for IOxSessionConfig {
...
+/// Generator for [`IOxSessionContext`] IDs.
+///
+/// Start at `1` because `0` is used as a default for unknown DataFusion config options.
+static CONTEXT_ID_COUNTER: AtomicU64 = AtomicU64::new(1);
+
+/// Key within the DataFusion config options that is used to pass [`IOxSessionContext`] IDs around.
+const CONFIG_OPTIONS_KEY: &str = "IOX_SESSION_CONTEXT_ID";
+
+/// [`IOxSessionContext`] ID -> [`SpanRecorder`]
+static SPAN_RECORDERS: Lazy<Mutex<HashMap<u64, Arc<Mutex<SpanRecorder>>>>> =
```
tustvold, line 201: I think it would be nicer to serialize the trace context directly, rather than proxying via a global hashmap. Especially since when SpanRecorder is dropped is important

tustvold (commented):
Wiring IOxSessionContext into the planning interfaces is a good change, and makes sense to me :+1:

That being said, I'm less a fan of the` SessionContextIOxExt` wiring. My main reservation is that it effectively side-steps the hierarchy that is established by the plan, and so rather than the spans appearing as children of the other DataFusion nodes, they will appear as a separate entity. I think I would prefer we use DataFusion's metrics plumbing to expose trace information from the physical execution, in the same vein as all the rest of the DataFusion operators. I think this would result in the most consistent output, and would also help us dogfood this interface, potentially improving it if necessary.

crepererum:
For the record: @tustvold and I had a chat and agreed that we extend DF to allow the pass-through of extension objects.

alamb:
> 
For the record: @tustvold and I had a chat and agreed that we extend DF to allow the pass-through of extension objects.

I will be happy to help review such a PR when it is ready

alamb:
Related PR: https://github.com/apache/arrow-datafusion/issues/2939

alamb:
PR to update DataFUsion: https://github.com/influxdata/influxdb_iox/pull/5148

crepererum:
New approach is ready for review.

iox_query/src/exec/context.rs
```diff
@@ -601,3 +614,17 @@ impl IOxSessionContext {
         self.exec.as_ref().map(|e| e.tasks()).unwrap_or_default()
     }
 }
+
+/// Extension trait to pull IOx spans out of DataFusion contexts.
+pub trait SessionContextIOxExt {
```
alamb, line 619: 👌

alamb (approved):
Looks great -- thanks @crepererum

PR #5136, opened 2022-07-15
feat: Select a subset of parquet files for a partition to compact

### LINKED ISSUE #5120: Given all L0 and L1 parquet files to consider for compaction, select and organize a subset to compact
Given the output of https://github.com/influxdata/influxdb_iox/issues/5118, which should be a group of two collections:

- L0 files ordered by max_sequence_number asc
- L1 files (their order isn’t important as they’re non-overlapping)

Then:

- Select the first 3 L0 files 
  - This is `INFLUXDB_IOX_COMPACTION_MAX_NUMBER_LEVEL_0_FILES` and defaults to 3
  - What if there are < 3 L0 files? 
    - Stop compaction attempt and return to find a new candidate partition? 
    - Do we think this is possible/likely? 
    - Should we log/have a metric recording that this has occurred?
- Select all L1 files that have time ranges overlapping with any of the selected L0 files. This could be empty if there are no L1 files or no overlapping ones.
- If the total size of the L0 and L1 files is greater than 350MB (or some threshold), try again, this time selecting fewer L0 files (again from the beginning of the L0 collection)
  - Hardcode 350MB for now? 
  - We should do this step before bothering to combine and sort the files, I think?
  - What if just one L0 file and the L1 files are over the threshold, such that when we continue selecting fewer L0 files, there are 0? Return? Log? Metrics?
- Construct and return a collection where the L1 files come first and the L0 files come after where the L0 files keep their ordering by sequence number. Precedence for overwrites/updates goes from lowest (L1) to highest (L0 ordered by sequence)

carols10cents:
Fixes #5120.

### DIFF
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -37,7 +37,7 @@ pub struct CompactorConfig {
         default_value = "3",
         action
     )]
-    pub compaction_max_number_level_0_files: i32,
+    pub max_number_level_0_files: i32,
 
     /// Desired max size of compacted parquet files.
     /// It is a target desired value, rather than a guarantee.
@@ -48,7 +48,7 @@ pub struct CompactorConfig {
         default_value = "104857600",
         action
     )]
-    pub compaction_max_desired_file_size_bytes: i64,
+    pub max_desired_file_size_bytes: i64,
 
     /// Percentage of desired max file size.
     /// If the estimated compacted result is too small, no need to split it.
@@ -62,7 +62,7 @@ pub struct CompactorConfig {
         default_value = "30",
         action
     )]
-    pub compaction_percentage_max_file_size: i16,
+    pub percentage_max_file_size: i16,
 
     /// Split file percentage
     /// If the estimated compacted result is neither too small nor too large, it will be split
@@ -81,7 +81,7 @@ pub struct CompactorConfig {
         default_value = "80",
         action
     )]
-    pub compaction_split_percentage: i16,
+    pub split_percentage: i16,
 
     /// The compactor will limit the number of simultaneous compaction jobs based on the
     /// size of the input files to be compacted. This number should be less than 1/10th
@@ -93,5 +93,5 @@ pub struct CompactorConfig {
         default_value = "1073741824",
         action
     )]
-    pub max_concurrent_compaction_size_bytes: i64,
+    pub max_concurrent_size_bytes: i64,
 }
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -82,4 +82,13 @@ pub struct CompactorConfig {
         action
     )]
     pub max_concurrent_size_bytes: i64,
+
+    /// TODO: Describe this. Multiplier?
+    #[clap(
+        long = "--new-param",
+        env = "INFLUXDB_IOX_COMPACTION_NEW_PARAM",
+        default_value = "3",
+        action
+    )]
+    pub new_param: i64,
 }
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -27,18 +27,6 @@ pub struct CompactorConfig {
     )]
     pub write_buffer_partition_range_end: i32,
 
-    /// Max number of level-0 files (written by ingester) we want to compact with level-1 each time
-    /// Default: 3
-    #[clap(
-        // TODO: verify the theory that if the streaming works as expected,
-        // we do not need to limit this number
-        long = "--compaction-max-number-level-0-files",
-        env = "INFLUXDB_IOX_COMPACTION_MAX_NUMBER_LEVEL_0_FILES",
-        default_value = "3",
-        action
-    )]
-    pub max_number_level_0_files: i32,
-
     /// Desired max size of compacted parquet files.
     /// It is a target desired value, rather than a guarantee.
     /// Default is 1024 * 1024 * 100 = 104,857,600 bytes (100MB)
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -30,7 +30,7 @@ pub struct CompactorConfig {
     /// Max number of level-0 files (written by ingester) we want to compact with level-1 each time
     /// Default: 3
     #[clap(
-        // TODO: verify the theory that if the streaming works  as expected,
+        // TODO: verify the theory that if the streaming works as expected,
         // we do not need to limit this number
         long = "--compaction-max-number-level-0-files",
         env = "INFLUXDB_IOX_COMPACTION_MAX_NUMBER_LEVEL_0_FILES",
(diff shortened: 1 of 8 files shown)


clap_blocks/src/compactor.rs
```diff
@@ -78,17 +69,26 @@ pub struct CompactorConfig {
...
-        default_value = "100000000",
+        default_value = "1073741824",
+        action
+    )]
+    pub max_concurrent_size_bytes: i64,
+
+    /// TODO: Describe this. Multiplier?
+    #[clap(
+        long = "--new-param",
+        env = "INFLUXDB_IOX_COMPACTION_NEW_PARAM",
```
pauldix, line 89: I think maybe `INFLUXDB_IOX_COMPACTION_INPUT_SIZE_THRESHOLD_BYTES`. Description could be something like: Compaction will gather as many L0 and L1 files to compact together until the total size of input files crosses this threshold. Later compactions will pick up the remaining L0 files.

pauldix (approved):
Looks good, just a few comments to name the parameter and add a parameter for a threshold on the number of input files.

clap_blocks/src/compactor.rs
```diff
@@ -62,14 +50,17 @@ pub struct CompactorConfig {
...
-    pub compaction_percentage_max_file_size: i16,
+    pub percentage_max_file_size: i16,
 
     /// Split file percentage
     /// If the estimated compacted result is neither too small nor too large, it will be split
     /// into 2 files determined by this percentage.
-    ///    . Too small means: < compaction_percentage_max_file_size * compaction_max_desired_file_size_bytes
+    ///
+    ///    . Too small means: < compaction_percentage_max_file_size *
+    ///      compaction_max_desired_file_size_bytes
```
NGA-TRAN, line 60: I also removed `compaction_` in the comments from the other PR 😄

compactor/src/compact.rs
```diff
@@ -225,9 +225,15 @@ pub struct Compactor {
...
 
-    /// Gauge for the number of bytes of level-0 files across compaction partition candidates
+    /// Gauge for the number of bytes of Parquet files across compaction partition candidates
     compaction_candidate_bytes_gauge: Metric<U64Gauge>,
 
+    /// Gauge for the number of Parquet file candidates
+    pub(crate) parquet_file_candidate_gauge: Metric<U64Gauge>,
+
+    /// Gauge for the number of bytes of Parquet file candidates
+    pub(crate) parquet_file_candidate_bytes_gauge: Metric<U64Gauge>,
```
NGA-TRAN, line 235: same here

compactor/src/parquet_file_filtering.rs
```diff
@@ -0,0 +1,655 @@
...
+use crate::parquet_file_lookup::ParquetFilesForCompaction;
+use data_types::ParquetFile;
+use metric::{Attributes, Metric, U64Gauge};
+use observability_deps::tracing::*;
+
+/// Given a list of level 0 files sorted by max sequence number and a list of level 1 files for a
+/// partition, compute a set of files that:
+///
+/// - Has a total size less than `max_bytes`
+/// - Has only level 1 files that overlap in time with the level 0 files
```
NGA-TRAN, line 13: You may want to add only the first L0 files in the sorted L0 list are selected.

compactor/src/parquet_file_filtering.rs
```diff
@@ -0,0 +1,655 @@
...
+///
+/// The returned files will be ordered with the level 1 files first, then the level 0 files ordered /// in ascending order by their max sequence number.
+#[allow(dead_code)] // TODO: Call this during a compaction operation
+pub(crate) fn filter_parquet_files(
```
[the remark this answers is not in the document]
NGA-TRAN: We did create both `size` and `file_count` thresholds before. I wonder should we do that or we simplify by just using the size threshold?

compactor/src/parquet_file_filtering.rs
```diff
@@ -0,0 +1,655 @@
...
+    let num_level_1_compacting = files_to_return.len();
+
+    info!(
+        partition_id = partition_id.get(),
+        num_level_0_considering,
+        num_level_1_considering,
+        num_level_0_compacting,
+        num_level_1_compacting,
+        "filtered Parquet files for compaction",
+    );
```
NGA-TRAN, line 99: 👍 This info will be useful. The metrics below look great, too

clap_blocks/src/compactor.rs
carols10cents: Right now, this parameter isn't a number of bytes-- as @NGA-TRAN [requested](https://github.com/influxdata/influxdb_iox/issues/5120#issuecomment-1184800571):

> I think we can go with max_size = compaction_max_desired_file_size_bytes * new_param then you create the new_param and make it 3 by default?

It's [a multiplier on the `INFLUXDB_IOX_COMPACTION_MAX_DESIRED_FILE_SIZE_BYTES`](https://github.com/influxdata/influxdb_iox/pull/5136/files#diff-76222fba0b3c4bfe27faf1d28dd0c95af52382870950bd035b757ec2ac913a58R50-R51).

@pauldix @NGA-TRAN could you two discuss this and let me know whether you want this parameter to be a multiplier for `INFLUXDB_IOX_COMPACTION_MAX_DESIRED_FILE_SIZE_BYTES` or an independent number of bytes?

compactor/src/parquet_file_filtering.rs
```diff
@@ -0,0 +1,655 @@
...
+        let files = filter_parquet_files(
+            parquet_files_for_compaction,
+            0,
+            &files_metric,
+            &bytes_metric,
+        );
+
+        assert_eq!(files.len(), 2);
+        assert_eq!(files[0].id.get(), 102);
+        assert_eq!(files[1].id.get(), 1);
```
NGA-TRAN, line 335: 🎉

compactor/src/parquet_file_filtering.rs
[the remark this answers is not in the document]
carols10cents: [Over here](https://github.com/influxdata/influxdb_iox/issues/5120#issuecomment-1184776731), @NGA-TRAN said:

> If you go with the size in the pseudocode, we have to ignore `compaction_max_number_level_0_files`

@pauldix @NGA-TRAN could you two discuss these conflicting requirements and let me know what you decide together? Thanks!

NGA-TRAN (approved):
Very nice.

- Can you propose the name for `new_param` on slack and I hope @alamb and @pauldix will suggest a good one? Ot at least we will discuss there until we get one.
- I hope you will let me PR go in first because there a good number of conflicts 😄 

.

carols10cents:
Outcome of the discussion on slack:

> 1. The new_param should not be the coefficient but the real size.  Let us name this `INFLUXDB_IOX_COMPACTION_INPUT_SIZE_THRESHOLD`  and make its default 300MB
> 2. We want to add another input threshold which is file count and let name it `INFLUXDB_IOX_COMPACTION_INPUT_FILE_COUNT_THRESHOLD`  and make it default 100. This is max compacting  (L0 + L1) file count. We need this because we suspect if we compact too many files, we will hit problem.
>
> We will stop adding files when either one of the above  thresholds hit first

PR #5137, opened 2022-07-15
refactor: replace level_2 with level_1

shepmaster:
Follow up to #5122


NGA-TRAN (approved):
Only one comment needs to remove, the rest looks good

PR #5135, opened 2022-07-15
feat: make object store garbage collector into a long-running service

shepmaster:
Closes #5076

PR #5147, opened 2022-07-18
fix: warn if `--data-dir is specified but `--object_store` type is not file

alamb:
# Rationale:
I was quite confused  (see https://github.com/influxdata/influxdb_iox/pull/5135#pullrequestreview-1042492748) when I ran the garbage collector feature with just `--data-dir` (as is supported with `all-in-one` mode).

Since I did not specify `--object-store=file`, it defaulted to a `Memory` object store, and the  `--data-dir` was silently ignored.

# Changes:
Generate a warning if `--data-dir` is specified but not `file` object store type

For example:
```shell
cargo run -- run querier --data-dir /Users/alamb/.influxdb_iox/ --catalog-dsn postgres://localhost/alamb
```

Now produces this warning:
```
2022-07-18T21:07:31.515144Z  WARN clap_blocks::object_store: --data-dir / `INFLUXDB_IOX_DB_DIR` ignored. It only affects 'file' object stores data_dir="/Users/alamb/.influxdb_iox/" object_store_type=None
```

Note that I also verified `all-in-one` mode still works:
```shell
cargo run -- run all-in-one -v --data-dir ~/.influxdb_iox --catalog-dsn postgres://localhost/alamb --persist-partition-cold-threshold-seconds=5
```


philjb (approved):
checked the warning appears with `INFLUXDB_IOX_DB_DIR` set only too.

PR #5149, opened 2022-07-19
chore: Add org and bucket ID parsing to tsm_reader

andrewcharlton:
We want to be able to migrate data from tsm files into iox/parquet. To do this, we need to be able to read a tsm file and extract the org and bucket ID so that we can get the right iox namespace for that data.  This PR just adds the ability to parse those values to the existing key parser.

The initial behaviour was subtly wrong, in that we do have keys in some clusters that contain the bytes that correspond to backslash, space or comma which we escape.  This means that the org & bucket section of the key is not always 16 bytes. I've corrected this so it should work properly.

Note: This only applies to cloud2 TSM files, the key format is different for cloud1.

This is my very first rust PR, any pointers to all the places I'm going horribly wrong would be much appreciated!


influxdb_tsm/src/key.rs
```diff
@@ -499,6 +536,39 @@ fn parse_tsm_tag_value(
...
+    }
+
+    #[test]
+    fn test_parse_id_escaped() {
+        // ID with escaped characters: we escape space (\x20), comma (\x2c) and backslash (\x5c)
+        let mut key = b"\x00\x5c\x20\x5c\x5c\x5c\x2c\x01\x5c\x2c\x03\x04,\x00=cpu"
+            .iter()
+            .copied();
+        let unescaped: [u8; 8] = hex::decode("00205c2c012c0304").unwrap().try_into().unwrap();
+
```
alamb, line 563: For anyone else who is curious, the unescaped value is

```
Unescaped: "\0 \\,\u{1},\u{3}\u{4}"
```

alamb: I am not sure I saw coverage of a double escape (like `"\\\5"` or something)

alamb (approved):
Very nice @andrewcharlton  👍 

I can't speak to if this PR's behavior is correct for legacy TSM (aka that binary keys have spaces, etc. escaped) but given you are an expert in that area I assume it is.

I think they code does what it claims to do and the tests are solid. Testing a double escape might make a nice improvement but I don't think it is strictly necessary

influxdb_tsm/src/key.rs
andrewcharlton: Doubly-escaped values aren't valid, and will actually stop the data being queryable in cloud2.

alamb: Got it -- I guess I saw there was code that handled double escaped values in the reader but I didn't see a test demonstrating what happened if it was fed input in that case (e.g. that it made an error).

Not a big deal I don't think

PR #5155, opened 2022-07-19
fix: tab complete from hints in `sql` relp (itch scratch)

alamb:
# Rationale

There is a cool history feature of the sql repl, which shows you a history command. However, it was driving my nuts that I could not type `tab` and get it to complete for me:
Screen Shot 2022-07-19 at 4 30 04 PM

# Changes
If a user hits `tab` and there is a history hint the hint will be filled in 

cc @Marwes


philjb (approved):
i also want this!

influxdb_iox/src/commands/sql/repl.rs
```diff
@@ -142,6 +142,20 @@ impl rustyline::highlight::Highlighter for RustylineHelper {
...
 impl rustyline::completion::Completer for RustylineHelper {
     type Candidate = String;
+
+    fn complete(
+        &self,
+        line: &str,
+        pos: usize,
+        ctx: &rustyline::Context<'_>,
+    ) -> rustyline::Result<(usize, Vec<Self::Candidate>)> {
+        if let Some(hint) = self.hinter.hint(line, pos, ctx) {
```
alamb, line 152:
```suggestion
        // If there is a hint, use that as the auto-complete when user hits `tab` 
        if let Some(hint) = self.hinter.hint(line, pos, ctx) {
```

PR #5158, opened 2022-07-20
refactor: enforce proper `IOxSessionContext`

crepererum:
- remove `IOxSessionContext::default()` because untracked contexts
  should only be created by tests
- remove `Option<IOxSessionContext>` because it is a typed workaround
  for `IOxSessionContext::default`

Tests should use `IOxSessionContext::testing` and all _normal_ users
should create proper contexts.

I suspect this will help tracing or at least prevent silent regressions.
See #5129.


querier/src/table/query_access.rs
```diff
@@ -44,7 +44,13 @@ impl TableProvider for QuerierTable {
     ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
         // build provider out of all chunks
         // TODO: push down some predicates to catalog
-        let mut builder = ProviderBuilder::new(self.table_name(), Arc::clone(self.schema()));
+        let iox_ctx = self
+            .exec
+            .new_execution_config(ExecutorType::Query)
+            .with_span_context(ctx.span_ctx())
+            .build();
```
crepererum, line 51: This feels a bit weird and I'm open to ideas how to better recover the IOx context here.

iox_query/src/exec/context.rs
```diff
@@ -276,6 +275,18 @@ impl fmt::Debug for IOxSessionContext {
 }
 
 impl IOxSessionContext {
+    /// Constructor for testing.
+    ///
+    /// This is identical to [`Default::default`] but we do NOT implement [`Default`] to make the creation of untracked
+    /// contexts mor explicit.
```
alamb, line 281:
```suggestion
    /// contexts more explicit.
```

iox_query/src/exec/context.rs
```diff
@@ -276,6 +275,18 @@ impl fmt::Debug for IOxSessionContext {
...
+    pub fn testing() -> Self {
```
alamb, line 282: 👍  this is a good idea. Very nice.

Another possible name is `new_for_testing` but maybe that is too wordy

iox_query/src/exec/context.rs
```diff
@@ -627,4 +641,10 @@ impl SessionContextIOxExt for SessionState {
             .get_extension::<Option<Span>>()
             .and_then(|span| span.as_ref().as_ref().map(|span| span.child(name)))
     }
+
+    fn span_ctx(&self) -> Option<SpanContext> {
```
alamb, line 645: 👍

iox_query/src/frontend/common.rs
```diff
@@ -95,9 +95,9 @@ pub struct ScanPlanBuilder<'a> {
 }
 
 impl<'a> ScanPlanBuilder<'a> {
-    pub fn new(table_schema: Arc<Schema>) -> Self {
+    pub fn new(table_schema: Arc<Schema>, ctx: IOxSessionContext) -> Self {
```
alamb, line 98: I like the explicitness of this (and the fact it now requires a `IOxSessionContext` to construct. 👍

querier/src/table/query_access.rs
alamb: The only thought I have is to pass the `IOxSessionContext` to the constructor of `QuerierTable` (as I think I remember that we construct a new `QuerierTable` for each query that comes in. But I may be mis remembering.

I think it would also be fine to add a comment here like "This function is called from within DataFusion, so we need to recover the IOx specific tracing span context to connect it to the original request" to give future readers a hint about what this is doing

alamb (approved):
I think this change makes a lot of sense to me 👍  nice work

querier/src/table/query_access.rs
crepererum: > The only thought I have is to pass the `IOxSessionContext` to the constructor of `QuerierTable` (as I think I remember that we construct a new `QuerierTable` for each query that comes in.

That could work as well. The constructor is called in `QuerierDatabase::namespace` and we wanna have some tracing information in there anyways because for #5129 I want to have a trace for every cache request. I'll try that...

iox_query/src/exec/context.rs
crepererum: done

PR #5159, opened 2022-07-20
feat: trace querier->ingester, stopping at gRPC layer

crepererum:
This adds tracing of querier->ingester request up to the point where we
perform the network request, i.e. the trace will only appear on the
querier side. We may extend this at some point to carry the tracing
information to the ingester as well.

Ref #5129.


querier/src/ingester/mod.rs
```diff
@@ -791,13 +803,18 @@ impl IngesterConnection for IngesterConnectionImpl {
             .map(move |ingester_address| measured_ingester_request(ingester_address))
             .collect::<FuturesUnordered<_>>()
             .try_collect::<Vec<_>>()
-            .await?
+            .await
+            .map_err(|e| {
```
alamb, line 807: 🤔  I wonder if `OberveIngesterRequest::error` should also be called here

alamb (approved):

querier/src/ingester/mod.rs
crepererum: Wrong scope, there are two recorders/spans here:

1. the overall scope that spans ALL ingester requests (see start of the method)
2. the per-ingester scope (created within `measured_ingester_request`)

querier/src/ingester/mod.rs
```diff
@@ -739,12 +744,19 @@ impl IngesterConnection for IngesterConnectionImpl {
             // INFO: create the measurement structure outside of the async block so cancellation is
             // always measured
             let measure_me = ObserveIngesterRequest::new(request.clone(), metrics);
+            let mut span_recorder = span_recorder.child("flight request");
```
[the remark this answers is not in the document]
crepererum: done

PR #5160, opened 2022-07-20
fix: Add TagKeyMetaNamesCapability capability

mkmik:
Closes https://github.com/influxdata/conductor/issues/1054


alamb (approved):
Thanks @mkmik  -- makes sense to me

I also verified that IOx does actually support `TagKeyMetaNames` 👍 

https://github.com/influxdata/influxdb_iox/blob/16768cc53ff88f1890abd0d6cb845ba309d7cf54/generated_types/protos/influxdata/platform/storage/storage_common.proto#L23-L33

PR #5161, opened 2022-07-20
feat: allow passing for extra arguments to `Cache::peek`

crepererum:
This will be used to pass spans down to `CacheWithMetrics` (or a new
wrapper specific to tracing) and will help with #5129.


cache_system/src/cache/metrics.rs
```diff
@@ -114,27 +116,37 @@ where
...
-    Extra: Debug + Send + 'static,
+    GetExtra: Debug + Send + 'static,
+    PeekExtra: Debug + Send + 'static,
 {
     type K = K;
     type V = V;
-    type Extra = Extra;
-
-    async fn get_with_status(&self, k: Self::K, extra: Self::Extra) -> (Self::V, CacheGetStatus) {
+    type GetExtra = GetExtra;
```
alamb, line 128: I think it might help at some point to add docstrings explaining what `GetExtra` and `PeekExtra` are for. `K` and `V` are more or less self explanatory for caches. The `Extra` concept is not as standard

alamb (approved):
Code does what it says it does I think 👍 

The cache system is getting more complicated (4 generic arguments) but I don't have any concrete suggestions on how to improve things.

cache_system/src/cache/metrics.rs
crepererum: see ping in other file

alamb: Thanks

PR #5163, opened 2022-07-20
refactor: improve cache system generic tests

crepererum:
This prepares the test system for the usage of extra args for GET and
PEEK by pulling all generics and the whole interface into a single
trait. This is similar to how the write buffer tests work.

This is needed to introduce extra args into the metrics wrapper (i.e. to
pass down spans) while still being able to use the generic tests in a
follow-up PR. Overall this change is required for #5129.

The tests itself are unchanged.


alamb (approved):
Looks like an improvement to me

PR #5164, opened 2022-07-20
feat: cache tracing 

crepererum:
~On top of #5163.~

Add tracing to the metrics cache wrapper. The extra arguments for GET
and PEEK make this quite simple, because the wrapper can just extend the
inner args with the trace information.

We currently terminate the span in `querier::cache` (i.e. only pass in
`None`, so no tracing will occur) to keep this PR rather small. This
will be changed in subsequent PRs.

For #5129.


cache_system/src/cache/metrics.rs
```diff
@@ -359,12 +419,30 @@ mod tests {
         let hist = get_metric_cache_get(&reporter, "cancelled");
         assert_eq!(hist.sample_count(), n_cancelled);
         assert_eq!(hist.total, (n_cancelled as u32) * d3);
+
+        // check spans
+        assert_n_spans(&traces, "hit", SpanStatus::Ok, n_hit as usize);
```
alamb, line 424: 👍  this is clever testing

cache_system/src/cache/mod.rs
```diff
@@ -27,6 +27,17 @@ pub enum CacheGetStatus {
     MissAlreadyLoading,
 }
 
+impl CacheGetStatus {
+    /// Get human and machine redable name.
```
alamb, line 31:
```suggestion
    /// Get human and machine readable name.
```

querier/src/cache/namespace.rs
```diff
@@ -157,8 +164,9 @@ impl NamespaceCache {
             }
         });
 
+        // TODO(marco): pass span
```
alamb, line 167: is this for a follow on PR?

querier/src/cache/parquet_file.rs
```diff
@@ -73,8 +74,14 @@ impl CachedParquetFiles {
     }
 }
 
-type CacheT =
-    Box<dyn Cache<K = TableId, V = Arc<CachedParquetFiles>, GetExtra = (), PeekExtra = ()>>;
+type CacheT = Box<
+    dyn Cache<
+        K = TableId,
+        V = Arc<CachedParquetFiles>,
+        GetExtra = ((), Option<Span>),
```
alamb, line 81: Why is the extra a tuple of `((), Option<Span>)` and not just an `Option<Span>`? I assume you have a plan, but figured I would ask on the off chance it was an oversight

The same question applies to the other Cache types below

alamb (approved):
LGTM

querier/src/cache/namespace.rs
crepererum: yes

querier/src/cache/read_buffer.rs
```diff
@@ -30,8 +31,14 @@ struct ExtraFetchInfo {
     store: ParquetStorage,
 }
 
-type CacheT =
-    Box<dyn Cache<K = ParquetFileId, V = Arc<RBChunk>, GetExtra = ExtraFetchInfo, PeekExtra = ()>>;
+type CacheT = Box<
+    dyn Cache<
+        K = ParquetFileId,
+        V = Arc<RBChunk>,
+        GetExtra = (ExtraFetchInfo, Option<Span>),
```
crepererum, line 38: @alamb to answer [your question about types](https://github.com/influxdata/influxdb_iox/pull/5164/files#r926512114) let's have a look at this particular example. Here the first tuple entry is actually used.

The construction is the following: The `CacheWithMetrics` wrapper takes a 

```text
Cache <
    K = K1,
    V = V1,
    GetExtra = GetExtra1,
    PeekExtra = PeekExtra1,
>
```

-- where `GetExtra` and `PeekExtra` are `()` in most but not all cases -- and produces a new cache

```text
Cache <
    K = K1,
    V = V1,
    GetExtra = (GetExtra1, Option<Span>),
    PeekExtra = (PeekExtra1, Option<Span>),
>
```

i.e. it uses the extra arguments to spans. Now I could probably come up with some type-level hackery to make the `()`/`((), Option<Span>)` case nicer, but TBH I'm not sure this would be worth it.

To me this type wrapping seemed like the best approach because:

- only the metrics wrapper cares about the span
- the interface (`Cache`) stays the same which:
  - makes combining this with other wrappers simpler
  - the API simpler because the user only needs to know a single unified interface
  - all the generic tests still work

alamb: Sounds good to me 👍

PR #5166, opened 2022-07-20
chore: add todo for missing tasks

NGA-TRAN:
@carols10cents While trying to compile the branch, I found these todos. I will do them next


NGA-TRAN:
I did not try to have anything pass jut like the original branch. I just add some functions that will help you move forward while I work on them. I think it is mergable but you may want to review it first

compactor/src/lib.rs
```diff
@@ -61,13 +64,17 @@ pub(crate) async fn compact_partition(
         &compactor.parquet_file_candidate_bytes_gauge,
     );
 
+    let to_compact_files_with_tombstones: Vec<ParquetFileWithTombstone> = vec![]; // NGA TODO
+
     let stream = parquet_file_combining::compact_parquet_files(
-        to_compact,
+        to_compact_files_with_tombstones,
```
carols10cents, line 70: I thought we weren't worrying about deletes right now?

NGA-TRAN: We do not have to worry about it right now but we do not want to remove what we already support. We want to keep compactor to fully support delete as the design has worked well and so far there are no issues at all (even though we have not done massive tests yet)

NGA-TRAN: After chatting with Carol, we decided to support delete in different PR

PR #5170, opened 2022-07-21
feat: wire up cache tracing (1/2)

crepererum:
Addresses the first half of the TODOs left by #5164 and passes down spans to some caches (tombstone, table, read buffer, processed tombstones). Also creates some intermediate spans where it makes sense.

Ref #5129.


querier/src/chunk/mod.rs
```diff
@@ -245,7 +248,15 @@ impl QuerierChunk {
             }
             QuerierChunkLoadSetting::OnDemand => {
                 // depends on cache state
-                catalog_cache.read_buffer().peek(meta.parquet_file_id).await
+                catalog_cache
```
alamb, line 251: I double checked that all the span descriptions matched what the code was doing 👍

querier/src/table/state_reconciler.rs
```diff
@@ -163,7 +175,13 @@ impl Reconciler {
                         .chunk_adapter
                         .catalog_cache()
                         .processed_tombstones()
-                        .exists(chunk.meta().parquet_file_id(), tombstone.tombstone_id())
+                        .exists(
+                            chunk.meta().parquet_file_id(),
+                            tombstone.tombstone_id(),
+                            span_recorder
+                                .span()
+                                .map(|span| span.child("cache GET processed_tombstone")),
```
alamb, line 183:
```suggestion
                                .map(|span| span.child("cache GET exists processed_tombstone")),
```

alamb (approved):
LGTM -- FWIW I agree that trying to test all these various annotations is not worthwhile as they are a debugging / introspection tool. The end to end test that makes sure the basic plumbing is hooked up is good.

My major contribution in this review was to validate that the descriptions matched the actions. 😂

crepererum:
> FWIW I agree that trying to test all these various annotations is not worthwhile as they are a debugging / introspection tool. The end to end test that makes sure the basic plumbing is hooked up is good.

Yeah, I thought about testing as well, but since we already test the low-level code (i.e. `CacheWithMetrics`) and it's merely a plumbing TODO (i.e. pass in a proper value instead of `None`), I decided that it would be worth the effort. If the code is wrong and the spans do NOT appear in our cluster, then the fix IMHO requires as test.

PR #5171, opened 2022-07-21
fix: Sync ReadWindowAggregate API: TagKeyMetaNames

mkmik:
The storage API has been updated in https://github.com/influxdata/idpe/pull/12868
in January, but since we forked the `.proto` files we never noticed.

Closes https://github.com/influxdata/conductor/issues/1054

### DIFF
--- a/service_grpc_influxrpc/src/service.rs
+++ b/service_grpc_influxrpc/src/service.rs
@@ -320,12 +320,20 @@ where
         let gby_agg = expr::make_read_group_aggregate(aggregate, group, group_keys)
             .context(ConvertingReadGroupAggregateSnafu { aggregate_string })?;
 
-        let results = query_group_impl(Arc::clone(&db), db_name, range, predicate, gby_agg, &ctx)
-            .await
-            .map_err(|e| e.to_status())?
-            .into_iter()
-            .map(Ok)
-            .collect::<Vec<_>>();
+        let results = query_group_impl(
+            Arc::clone(&db),
+            db_name,
+            range,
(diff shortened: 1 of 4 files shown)


service_grpc_influxrpc/src/service.rs
```diff
@@ -320,12 +320,20 @@ where
...
-            .into_iter()
-            .map(Ok)
-            .collect::<Vec<_>>();
+        let results = query_group_impl(
+            Arc::clone(&db),
+            db_name,
+            range,
+            predicate,
+            gby_agg,
+            TagKeyMetaNames::Text,
```
alamb, line 329: This made me wonder if `ReadGroupRequest` also had a `TagKeyMetaNames` field as well -- I looked at the protobuf in idpe and the only two request types I could find that had `TagKeyMetaNames` are `ReadFilterRequest` and `ReadWindowAggregateRequest` 👍 

https://github.com/influxdata/idpe/blob/3d64893b40daac972cfd430c6dee8cefcde4c1d9/storage/storageproto/storage_common.proto

alamb (approved):
LGTM -- the only thing that I see missing is a test (aka if we ever broke this in the future a test would fail). I am happy to help write such a one if needed

service_grpc_influxrpc/src/service.rs
mkmik: yeah I manually sifted through the diffs (a lot of them were noise; different formatting, different import paths, removed gogo annotations and other shenanigans) and found just this one.

PR #5173, opened 2022-07-21
feat: wire up cache tracing (2/2)

crepererum:
Ref #5129.

### DIFF
--- a/querier/src/cache/namespace.rs
+++ b/querier/src/cache/namespace.rs
@@ -146,6 +146,7 @@ impl NamespaceCache {
         &self,
         name: Arc<str>,
         should_cover: &[(&str, &HashSet<ColumnId>)],
+        span: Option<Span>,
     ) -> Option<Arc<NamespaceSchema>> {
         self.backend.remove_if(&name, |cached_namespace| {
             if let Some(namespace) = cached_namespace.as_ref() {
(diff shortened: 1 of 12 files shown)


querier/src/cache/namespace.rs
```diff
@@ -164,9 +165,8 @@ impl NamespaceCache {
             }
         });
 
-        // TODO(marco): pass span
```
alamb, line 167: 🎉

querier/src/ingester/mod.rs
```diff
@@ -514,8 +521,15 @@ async fn execute(request: GetPartitionForIngester<'_>) -> Result<Vec<IngesterPar
...
 
     // reconstruct partitions
-    let mut decoder =
-        IngesterStreamDecoder::new(ingester_address, table_name, catalog_cache, expected_schema);
+    let mut decoder = IngesterStreamDecoder::new(
+        ingester_address,
+        table_name,
+        catalog_cache,
+        expected_schema,
+        span_recorder
```
alamb, line 529: I think you can call `SpanRecorder::child` directly here for slightly less typing 
https://github.com/influxdata/influxdb_iox/blob/7b28fb43660b7294879eeeae9fa1db2c7a4b8a49/trace/src/span.rs#L206-L208

alamb (approved):

alamb:
I figured I would merge and deploy this change so it is running in prod when @crepererum comes back online

querier/src/ingester/mod.rs
alamb: See https://github.com/influxdata/influxdb_iox/pull/5180

PR #5175, opened 2022-07-21
feat: add sort key and partition key into PartitionCompactionCandidateWithInfo

### LINKED ISSUE #5174: Add sort_key and partition_key into PartitionParam
This is a task of #5106. The compact partition needs to know sort key and partition key. We do not want to get them per file but per partition. The `PartitionParams` is currently used for that purpose and it should include `sort_key` and `partition_key` too

NGA-TRAN:
Closes #5174 


carols10cents (approved):
Seems fine for now... I'm a little concerned about the number of catalog queries `add_info_to_partitions` is doing, but I guess we'll see if it's too much load on the catalog when this goes out :)

Issue #5121, opened 2022-07-13
Given an organized subset of L0+L1 files, compact into a set of new L1 files


carols10cents:
Given the output of https://github.com/influxdata/influxdb_iox/issues/5120, which should be a collection of parquet files where the L1 files come first and the L0 files come after and the L0 files keep their ordering by sequence number, first:

- Log the partition ID, number (and IDs?) of L0 input files, number (and IDs?) of L1 input files, total size of input files
- Would the number of files of each level and total size make useful metrics too? Anything else?

Then compact the L0 and L1 files into a new set of L1 files with this criteria:

- If the total size of the files is < 30MB: 
  - Write everything into a single L1 file
  - This value is `INFLUXDB_IOX_COMPACTION_PERCENTAGE_MAX_FILE_SIZE * .01 * INFLUXDB_IOX_COMPACTION_MAX_DESIRED_FILE_SIZE_BYTES`, which defaults to `30 * .01 * 100 MB = 30 MB` (needs to be converted to bytes)
- If the total size is > 30MB and < 130MB: 
  - Write 80% of the data into one L1 file and 20% into another, split based on time.
  - Use the min and max times of the files and figure out the 80% point, assume data is roughly evenly distributed.
  - Percentage is `INFLUXDB_IOX_COMPACTION_SPLIT_PERCENTAGE` that defaults to 80
- If the total size is > 130MB: 
  - Determine a split based on time where files of about 100MB are output with some remainder of 10MB or so is left for the last file. 
  - Again the split should be based on time... **wait, how do we convert between approximate size and time? Just start streaming and stop when the file size goes over the limit?**
  - This value is `(1 + INFLUXDB_IOX_COMPACTION_PERCENTAGE_MAX_FILE_SIZE * .01) * INFLUXDB_IOX_COMPACTION_MAX_DESIRED_FILE_SIZE_BYTES`, which defaults to `(1 + 30 * .01) * 100 MB = 130 MB` (needs to be converted to bytes)

After completion of writing all files to object storage:

Within a transaction:

- Create records for all the new L1 files in the catalog
- Mark the input files (both L0 and L1) as to_delete = 't' in the catalog

Then:

- Log successful compaction of partition ID with the number (and IDs?) of new L1 files and the total size of those files 
- Would the number of new L1 files and their total size? Or each size? Make useful metrics?

carols10cents:
[...]

PR #5176, opened 2022-07-21
feat: Compact filtered parquet files

carols10cents:
Connects to #5121.

This does not include any metrics yet, and it isn't correctly hooked up yet (the `compact_partition` function in compactor/src/lib.rs needs to log and ignore some of the errors).

But while I'm working on those two parts, I wanted to get some feedback on the code and the tests.

### DIFF
--- a/compactor/src/parquet_file_combining.rs
+++ b/compactor/src/parquet_file_combining.rs
@@ -310,6 +310,7 @@ fn to_queryable_parquet_chunk(
         file.max_time,
         sort_key,
         partition_sort_key,
+        file.compaction_level,
     )
 }
 
--- a/compactor/src/parquet_file_combining.rs
+++ b/compactor/src/parquet_file_combining.rs
@@ -374,6 +374,7 @@ mod tests {
         .join("\n");
         let builder = TestParquetFileBuilder::default()
             .with_line_protocol(&lp)
+            .with_max_seq(20) // This should be irrelevant because this is a level 1 file
             .with_compaction_level(CompactionLevel::FileNonOverlapped); // Prev compaction
         let level_1_file = partition.create_parquet_file(builder).await;
 
@@ -399,18 +400,30 @@ mod tests {
             .with_max_seq(2);
         let level_0_max_seq_2 = partition.create_parquet_file(builder).await;
 
+        let lp = vec![
+            "table,tag1=VT field_int=88i 10000", // will be deduplicated with level_0_max_seq_1
+            "table,tag1=OR field_int=99i 12000",
+        ]
+        .join("\n");
+        let builder = TestParquetFileBuilder::default()
+            .with_line_protocol(&lp)
+            .with_max_seq(5) // This should be irrelevant because this is a level 1 file
+            .with_compaction_level(CompactionLevel::FileNonOverlapped); // Prev compaction
+        let level_1_with_duplicates = partition.create_parquet_file(builder).await;
+
         let lp = vec!["table,tag2=OH,tag3=21 field_int=21i 36000"].join("\n");
         let builder = TestParquetFileBuilder::default()
             .with_line_protocol(&lp)
             .with_file_size_bytes(40 * 1024 * 1024); // Really large file
         let large_file = partition.create_parquet_file(builder).await;
 
-        // Order here is important! The Level 1 files are first, then Level 0 files in max seq num
-        // ascending order, as filter_parquet_files would create them
+        // Order here isn't relevant; the chunk order should ensure the level 1 files are ordered
+        // first, then the other files by max seq num.
         let parquet_files = vec![
-            level_1_file.parquet_file,
-            level_0_max_seq_1.parquet_file,
             level_0_max_seq_2.parquet_file,
+            level_1_with_duplicates.parquet_file,
+            level_0_max_seq_1.parquet_file,
+            level_1_file.parquet_file,
             large_file.parquet_file,
         ];
 
@@ -480,7 +493,7 @@ mod tests {
         } = test_setup().await;
 
         compact_parquet_files(
-            parquet_files.into_iter().take(3).collect(),
+            parquet_files.into_iter().take(4).collect(),
             &candidate_partition,
             Arc::clone(&catalog.catalog),
             ParquetStorage::new(Arc::clone(&catalog.object_store)),
@@ -501,13 +514,13 @@ mod tests {
         // 1 large file not included in compaction
         assert_eq!(
             (files[0].id.get(), files[0].compaction_level),
-            (4, CompactionLevel::Initial)
+            (5, CompactionLevel::Initial)
         );
         // 1 newly created CompactionLevel::FileNonOverlapped file as the result of
         // compaction
         assert_eq!(
             (files[1].id.get(), files[1].compaction_level),
-            (5, CompactionLevel::FileNonOverlapped)
+            (6, CompactionLevel::FileNonOverlapped)
         );
 
         // ------------------------------------------------
@@ -528,6 +541,7 @@ mod tests {
                 "| 21        |      | OH   | 21   | 1970-01-01T00:00:00.000036Z |",
                 "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
                 "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "| 99        | OR   |      |      | 1970-01-01T00:00:00.000012Z |",
                 "+-----------+------+------+------+-----------------------------+",
             ],
             &batches
--- a/compactor/src/parquet_file_combining.rs
+++ b/compactor/src/parquet_file_combining.rs
@@ -318,7 +318,7 @@ mod tests {
     use super::*;
     use arrow::record_batch::RecordBatch;
     use arrow_util::assert_batches_sorted_eq;
-    use data_types::{ColumnType, PartitionKey, PartitionParam};
+    use data_types::{ColumnType, PartitionParam};
     use iox_tests::util::{TestCatalog, TestParquetFileBuilder, TestTable};
     use parquet_file::ParquetFilePath;
     use test_helpers::assert_error;
@@ -362,8 +362,8 @@ mod tests {
                 namespace_id: ns.namespace.id,
                 table_id: partition.partition.table_id,
             },
-            sort_key: Some(sort_key),                   //NGA TODO
-            partition_key: PartitionKey::from("to-do"), // NGA TODO
+            sort_key: partition.partition.sort_key(),
+            partition_key: partition.partition.partition_key.clone(),
         };
 
         let lp = vec![
--- a/compactor/src/parquet_file_combining.rs
+++ b/compactor/src/parquet_file_combining.rs
@@ -1,0 +1,579 @@
+use crate::{compact::PartitionCompactionCandidateWithInfo, query::QueryableParquetChunk};
+use data_types::{CompactionLevel, ParquetFile, PartitionId, TableSchema};
+use datafusion::error::DataFusionError;
+use iox_catalog::interface::Catalog;
+use iox_query::{
+    exec::{Executor, ExecutorType},
+    frontend::reorg::ReorgPlanner,
+    QueryChunk,
+};
+use iox_time::TimeProvider;
+use observability_deps::tracing::*;
+use parquet_file::{chunk::ParquetChunk, metadata::IoxMetadata, storage::ParquetStorage};
+use schema::{sort::SortKey, Schema};
+use snafu::{ensure, ResultExt, Snafu};
+use std::{
+    cmp::{max, min},
+    sync::Arc,
+};
+use uuid::Uuid;
+
+#[derive(Debug, Snafu)]
+#[allow(missing_copy_implementations, missing_docs)]
+pub(crate) enum Error {
+    #[snafu(display(
+        "Must specify at least 2 files to compact for {}, got {num_files}", partition_id.get()
+    ))]
+    NotEnoughParquetFiles {
+        num_files: usize,
+        partition_id: PartitionId,
+    },
+
+    #[snafu(display(
+        "Total size of selected files for {} is {total_size}, compacting that much is \
+         not yet implemented.", partition_id.get()
+    ))]
+    TooMuchData {
+        total_size: i64,
+        partition_id: PartitionId,
+    },
+
+    #[snafu(display("Error building compact logical plan  {}", source))]
+    CompactLogicalPlan {
+        source: iox_query::frontend::reorg::Error,
+    },
+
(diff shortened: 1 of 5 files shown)


compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+    store: ParquetStorage,
+    // Executor for running queries, compacting, and persisting
+    exec: Arc<Executor>,
+    time_provider: Arc<dyn TimeProvider>,
+) -> Result<(), Error> {
+    let partition_id = partition.id();
+
+    let num_files = files.len();
+    ensure!(
+        num_files >= 2,
```
NGA-TRAN, line 98: FYI: Since we do not support delete, 2 is a reasonable number. When we support delete, one file with tombstone is good enough to compact

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+
+    // Find the total size of all files. For now, only compact if the total size is under 30MB.
+    let total_size: i64 = files.iter().map(|f| f.file_size_bytes).sum();
+    ensure!(
+        total_size < TEMPORARY_COMPACTION_MAX_BYTES_LIMIT,
+        TooMuchDataSnafu {
+            total_size,
+            partition_id
+        }
+    );
```
NGA-TRAN, line 113: If I turn compactor on after this merge, what will we see if we hit this (we will). A panic? Will the compactor keep compacting next candidate?

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+            table_schema: Arc::new(table_schema),
+            namespace: Arc::new(ns.namespace.clone()),
+            candidate: PartitionParam {
+                partition_id: partition.partition.id,
+                sequencer_id: partition.partition.sequencer_id,
+                namespace_id: ns.namespace.id,
+                table_id: partition.partition.table_id,
+            },
+            sort_key: Some(sort_key),                   //NGA TODO
+            partition_key: PartitionKey::from("to-do"), // NGA TODO
```
NGA-TRAN, line 366:
```suggestion
            sort_key: partition.sort_key(),
            partition_key: partition.partition_key.clone(),
```
I have not checked syntax

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+        Arc::new(parquet_chunk),
+        &[],
+        file.max_sequence_number,
+        file.min_time,
+        file.max_time,
+        sort_key,
+        partition_sort_key,
+    )
+}
+
```
NGA-TRAN, line 315: One major change this PR misses is to order the chunks: all level-1 files must be order 0 and level-0 files must be their max sequence number. This must be set in this [order](https://github.com/influxdata/influxdb_iox/blob/c0ac01c67b0721a77b973f89650d64dc5e2f8a26/compactor/src/query.rs#L242) function of QueryableParquetChunk.
In order to do so, you need to add `compaction_level` field into this [QueryableParquetChunk](https://github.com/influxdata/influxdb_iox/blob/c0ac01c67b0721a77b973f89650d64dc5e2f8a26/compactor/src/query.rs#L42) struct, and make the order function look like:

``` rust
// Order of the chunk so they can be deduplicate correctly
        if self.compaction_level == CompactionLevel::Initial {
              ChunkOrder::new(self.max_sequence_number.get())
        }  else {
              ChunkOrder::new(0)
       } 
    }
```

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+        let lp = vec![
+            "table,tag1=WA field_int=1000i 8000", // will be eliminated due to duplicate
+            "table,tag1=VT field_int=10i 10000",
+            "table,tag1=UT field_int=70i 20000",
+        ]
+        .join("\n");
+        let builder = TestParquetFileBuilder::default()
+            .with_line_protocol(&lp)
+            .with_max_seq(1);
```
NGA-TRAN, line 387: After you add the missing change in order function, we will need to add compaction level here to have it work correctly
```suggestion
            .with_max_seq(1)
            .with_compaction_level(CompactionLevel::Initial);
```

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        let lp = vec![
+            "table,tag1=WA field_int=1500i 8000", // latest duplicate and kept
+            "table,tag1=VT field_int=10i 6000",
+            "table,tag1=UT field_int=270i 25000",
+        ]
+        .join("\n");
+        let builder = TestParquetFileBuilder::default()
+            .with_line_protocol(&lp)
+            .with_max_seq(2);
```
NGA-TRAN, line 398: same here: need to set compaction level to 0

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+        let level_0_max_seq_2 = partition.create_parquet_file(builder).await;
+
+        let lp = vec!["table,tag2=OH,tag3=21 field_int=21i 36000"].join("\n");
+        let builder = TestParquetFileBuilder::default()
+            .with_line_protocol(&lp)
+            .with_file_size_bytes(40 * 1024 * 1024); // Really large file
```
NGA-TRAN, line 404: You want to set level 0 here too

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+        let large_file = partition.create_parquet_file(builder).await;
+
+        // Order here is important! The Level 1 files are first, then Level 0 files in max seq num
+        // ascending order, as filter_parquet_files would create them
+        let parquet_files = vec![
+            level_1_file.parquet_file,
+            level_0_max_seq_1.parquet_file,
+            level_0_max_seq_2.parquet_file,
+            large_file.parquet_file,
+        ];
```
NGA-TRAN, line 414: Order here is not important. The compaction level and their max seq are important. I think you even want to mess this order up to make sure they work well

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+
+        TestSetup {
+            catalog,
+            table,
+            candidate_partition,
+            parquet_files,
+        }
```
NGA-TRAN, line 421: To have a better code coverage, I suggest to add these 2 in this test setup:

1. One more level-1 parquet file with at least 2 rows with `tag1`, `field_int`, and `time` columns. Make one row of it duplicate with one of the rows of one of the level-0 files. Make sure that row is eliminated in the compacted result test.
2. Set seq_num for level-1 files, too to have a complete test setup and also use it as a test that those number won't be used in the chunk's order function because their values are always 0

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,579 @@
...
+                "+-----------+------+------+------+-----------------------------+",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000006Z |",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000010Z |",
+                "| 1500      | WA   |      |      | 1970-01-01T00:00:00.000008Z |",
+                "| 1601      |      | PA   | 15   | 1970-01-01T00:00:00.000030Z |",
+                "| 21        |      | OH   | 21   | 1970-01-01T00:00:00.000036Z |",
+                "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "+-----------+------+------+------+-----------------------------+",
+            ],
```
NGA-TRAN, line 531: I you add one more level-1 file with at least 2 rows and one of them is a duplicate, the result here should reflect that

NGA-TRAN (commented):
The code and test look good for covering the total size < 30MB 🎉 .  These are the two we want to add in this PR:
1. The order function of QueryableParquetChunk must be changed  
2. The test setup should add a duplicate row in level-1

compactor/src/parquet_file_combining.rs
carols10cents: Yes, this is for the current state only

compactor/src/parquet_file_combining.rs
carols10cents: I'm working on that for the next PR, but it'll end up [in a spot like this one](https://github.com/influxdata/influxdb_iox/blob/50186ef5ee76f30b1b58a1890ce10a4ef9cd119d/compactor/src/handler.rs#L290-L295) where a `warn!` logs the error and the compaction continues.

compactor/src/parquet_file_combining.rs
carols10cents: 🤦🏻‍♀️ great catch!

compactor/src/parquet_file_combining.rs
carols10cents: [The default level set in the `TestParquetFileBuilder` is `CompactionLevel::Initial`.](https://github.com/influxdata/influxdb_iox/blob/50186ef5ee76f30b1b58a1890ce10a4ef9cd119d/iox_tests/src/util.rs#L620)

carols10cents:
@NGA-TRAN I think this is ready for rereview!

compactor/src/query.rs
```diff
@@ -49,6 +49,7 @@ pub struct QueryableParquetChunk {
     max_time: Timestamp,
     sort_key: Option<SortKey>,
     partition_sort_key: Option<SortKey>,
+    compaction_level: CompactionLevel,
```
NGA-TRAN, line 52: 👍

compactor/src/query.rs
```diff
@@ -232,7 +235,76 @@ impl QueryChunk for QueryableParquetChunk {
 
     // Order of the chunk so they can be deduplicate correctly
     fn order(&self) -> ChunkOrder {
-        let seq_num = self.max_sequence_number.get();
-        ChunkOrder::new(seq_num)
+        match self.compaction_level {
+            CompactionLevel::Initial => ChunkOrder::new(self.max_sequence_number.get()),
+            CompactionLevel::FileNonOverlapped => ChunkOrder::new(0),
```
NGA-TRAN, line 240: 👍

compactor/src/query.rs
```diff
@@ -232,7 +235,76 @@ impl QueryChunk for QueryableParquetChunk {
...
+        let chunk = test_setup(CompactionLevel::Initial, 2).await;
+
+        assert_eq!(chunk.order(), ChunkOrder::new(2));
+    }
+
+    #[tokio::test]
+    async fn chunk_order_is_0_when_compaction_level_1() {
+        let chunk = test_setup(CompactionLevel::FileNonOverlapped, 2).await;
+
+        assert_eq!(chunk.order(), ChunkOrder::new(0));
```
NGA-TRAN, line 308: Nice tests

compactor/src/parquet_file_combining.rs
```diff
@@ -374,6 +374,7 @@ mod tests {
         .join("\n");
         let builder = TestParquetFileBuilder::default()
             .with_line_protocol(&lp)
+            .with_max_seq(20) // This should be irrelevant because this is a level 1 file
```
NGA-TRAN, line 377: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -399,18 +400,30 @@ mod tests {
             .with_max_seq(2);
         let level_0_max_seq_2 = partition.create_parquet_file(builder).await;
 
+        let lp = vec![
+            "table,tag1=VT field_int=88i 10000", // will be deduplicated with level_0_max_seq_1
+            "table,tag1=OR field_int=99i 12000",
```
NGA-TRAN, line 405: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -399,18 +400,30 @@ mod tests {
...
         let lp = vec!["table,tag2=OH,tag3=21 field_int=21i 36000"].join("\n");
         let builder = TestParquetFileBuilder::default()
             .with_line_protocol(&lp)
             .with_file_size_bytes(40 * 1024 * 1024); // Really large file
         let large_file = partition.create_parquet_file(builder).await;
 
-        // Order here is important! The Level 1 files are first, then Level 0 files in max seq num
-        // ascending order, as filter_parquet_files would create them
+        // Order here isn't relevant; the chunk order should ensure the level 1 files are ordered
+        // first, then the other files by max seq num.
```
NGA-TRAN, line 421: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -0,0 +1,594 @@
...
+                "+-----------+------+------+------+-----------------------------+",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000006Z |",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000010Z |",
+                "| 1500      | WA   |      |      | 1970-01-01T00:00:00.000008Z |",
+                "| 1601      |      | PA   | 15   | 1970-01-01T00:00:00.000030Z |",
+                "| 21        |      | OH   | 21   | 1970-01-01T00:00:00.000036Z |",
+                "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "| 99        | OR   |      |      | 1970-01-01T00:00:00.000012Z |",
+                "+-----------+------+------+------+-----------------------------+",
```
NGA-TRAN, line 545: 🎉

NGA-TRAN (approved):
Very nice

compactor/src/parquet_file_combining.rs
NGA-TRAN: @carols10cents I think we will hit a loop-forever for this case
Let say we have many L0s and L1. We pick the first L0 and it has no overlapped L1. This L0 is large enough and is sent to this function. However, this function does not compact it as it is the only file. Next round we choose it again as it is still L0 and smallest max_seq and the process it repeats again and again

I think we have to do something with this case. Something like:
1. If the L0 file <=  100MB (our config param), upgrade it to L1
2. Else use `split_plan` on it to split it into many files just as compact and split

NGA-TRAN: If we do not want to deal with upgrade right now. A simplest way is to ignore this check and compact it. We can  open a ticket for doing the upgrade later when we finish handle all cases of different sizes. I feel this may be a better  approach because we might be able to process upgrade for all cases in a same way. Let me know what you think. I can open a ticket for this as a subtask of #5106.  We can chat to see if those tasks are independent enough for you and me to implement in parallel or they have to be done in serial. I have a feeling they need to be done in serial

PR #5180, opened 2022-07-21
refactor: Reduce boiler plate calling `SpanRecorder::child`

alamb:
This is a minor code cleanup inspired by https://github.com/influxdata/influxdb_iox/pull/5173#discussion_r927073436

PR #5181, opened 2022-07-21
refactor: Reduce ceremony requried to create a `Span` from `SpanContext`

alamb:
Also inspired by @crepererum 's PR https://github.com/influxdata/influxdb_iox/pull/5173

# Rationale:
The ceremony required to get a `Span` from an `Option<SpanContext>` has always seemed excessive to me. 

# Changes
1. Add a trait so I can add a `child_span()` like in https://github.com/influxdata/influxdb_iox/pull/5180
2. Reduce code duplication


service_grpc_influxrpc/src/service.rs
```diff
@@ -235,23 +235,14 @@ where
         let req = req.into_inner();
         let permit = self
             .db_store
-            .acquire_semaphore(
-                span_ctx
-                    .as_ref()
-                    .map(|span| span.child("query rate limit semaphore")),
-            )
+            .acquire_semaphore(span_ctx.child_span("query rate limit semaphore"))
```
alamb, line 238: I was mostly annoyed at how much extra was needed in this file to pass along tracing information. Hopefully after this PR it is now less.

crepererum (approved):

alamb:
cc @tustvold

PR #5183, opened 2022-07-22
feat: Implement LineProtocol (LP) builder

mkmik:
Helps with https://github.com/influxdata/influxrpc_compare/issues/179

This PR implements a typestate influxdb line protocol builder.

It quotes tags keys, tags values, field keys and field values according to [the spec](https://docs.influxdata.com/influxdb/cloud/reference/syntax/line-protocol).

The new `influxdb_line_protocol::builder` module is documented inline as rustdoc.

---

Thanks to @alamb for the initial work in https://github.com/influxdata/influxdb_iox/compare/main...alamb/export_as_lp; I ended up making significant changes hence I preferred a clean slate. In particular my implementation doesn't return `Results` since it doesn't write to an `io::Write` but to a `bytes::BufMut`. I looked at how such a builder would be used to stream uploads using tokio and hyper, and all doc and exmples I found were centered around explicit chunking with `bytes::Buf` impls and never using blocking `io::Write`. The infallible nature of `bytes::BufMut` helped streamline the code, with no error enums, no snafus etc.


lukebond (approved):
very neat and clever approach. looking forward to trying it out!

influxdb_line_protocol/src/builder.rs
```diff
@@ -0,0 +1,490 @@
...
+/// let lp = LineProtocolBuilder::new()
+///     .measurement("foo")
+///     .tag("bar", "baz")
+///     .field("qux", 42.0)
+///     .close_line();
+///
+/// assert_eq!(lp.build(), b"foo,bar=baz qux=42\n");
+/// ```
+///
+/// [`LineProtocolBuilder`] never returns runtime errors. Instead, it employs as type-level state machine
```
alamb, line 48: ❤️

influxdb_line_protocol/src/builder.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+impl LineProtocolBuilder<Vec<u8>, BeforeMeasurement> {
+    /// Creates a new [`LineProtocolBuilder`] with an empty buffer.
+    pub fn new() -> Self {
+        Self::new_with(vec![])
+    }
+}
+
+impl<B> LineProtocolBuilder<B, BeforeMeasurement>
```
alamb, line 120: This is a neat formulation

influxdb_line_protocol/src/builder.rs
```diff
@@ -0,0 +1,490 @@
...
+        assert_eq!(get_field_key(row, 6), WITH_BACKSLASH);
+
+        let get_field_value = |n: usize, f: usize| format!("{}", parsed_lines[n].field_set[f].1);
+        let row = 3;
+        assert_eq!(get_field_value(row, 0), "false");
+        assert_eq!(get_field_value(row, 1), "51i");
+        assert_eq!(get_field_value(row, 2), "51u");
+        assert_eq!(get_field_value(row, 3), "51");
+        assert_eq!(get_field_value(row, 4), "some value");
+        // TODO(mkm): file an issue for the parser since it incorrectly decodes an escaped double quote (possibly also the Go version).
```
alamb, line 474: ??

alamb (approved):
I think adding some way to do error checking would be nice (like on tag/field names) but we can always extend it in a subsequent PR -- this is a great initial implementation

influxdb_line_protocol/src/builder.rs
mkmik: more info when I file the issue

mkmik:
> > since it doesn't write to an io::Write but to a bytes::BufMut
> 
> I think it is fine to start with BytesMut and then genericize to `std::io::Write` if needed
> 
> `BytesMut` implements `Write` -- so I think it would be possible to make this generic on write and still write directly to BytesByt
> 
> https://docs.rs/bytes/1.2.0/bytes/struct.BytesMut.html#impl-Write-for-BytesMut

yes that btw what I do under the hood so that I can use `write!`.

the main difference is that `std::io::Write` can do IO and thus can fail. I have mixed feelings. On one hand rust error handling makes this relatively invisible. But OTOH, why stop there? what about async Write? etc Perhaps constraing the builder to just populate a buffer and then leave it to the user to decide how to flush that buffer to some IO sink is the simplest thing to do.

mkmik:
> I think adding some way to do error checking would be nice (*like on tag/field names*)

interesting; I still think there is value in distinguishing a syntactically well formed LP from a semantically correct one.

For example, influxdb currently accepts fields and tags with the same name while IOx doesn't; yet both speak the same LP wire format. Having a clear distinction between semantic and syntactic specifications can help avoid confusion.

I.e. we could have a validation mechanism that performs some semantic checks if we want, but they could be applied on top of the basic LP builder; WDYT?

alamb:
> On one hand rust error handling makes this relatively invisible. But OTOH, why stop there? what about async Write? etc Perhaps constraing the builder to just populate a buffer and then leave it to the user to decide how to flush that buffer to some IO sink is the simplest thing to do.

I guess I felt like since rust already had `std::io::Write` and it was well integrated into the ecosystem (and lets people who care avoid double buffing) it would be the "obvious" choice. But maybe not 🤷 We could also just `unwrap()` all the `write!` errors and explain in the docs that this assumes no I/O errors as an incremental step. 

> interesting; I still think there is value in distinguishing a syntactically well formed LP from a semantically correct one.

Yes I agree - one might argue that the syntactic validation can be done at compile time (as you have implemented here) but that semantic validation necessairly has to be delayed until runtime (as it depends on the values of data that are not available at compile time)

PR #5185, opened 2022-07-22
refactor: improve query gRPC logging

crepererum:
- ensure that logging is done BEFORE the DB/namespace is requested (i.e.
  any actual work is done) but AFTER the query semaphore is acquired
- simplify tag key decoding (so that the logging statements are simpler
  to write)


service_grpc_flight/src/lib.rs
```diff
@@ -206,6 +206,11 @@ where
                     .map(|span| span.child("query rate limit semaphore")),
             )
             .await;
+        info!(
+            db_name=%read_info.database_name,
```
alamb, line 210: I wonder if it is possible / advisable to log the span / trace id here as well (to help us correlate with the binary logs)

mkmik (approved):

alamb (approved):

service_grpc_flight/src/lib.rs
mkmik: yeah, that's the intent behind this prep-PR:

context from slack channel, @crepererum said:
> OK, wiring the trace ID through all the layers takes more time, because at the moment we only extract it if sampling is requested (but you probably wanna have it for all requests).
> Some prep work: <this PR>

PR #5187, opened 2022-07-22
feat: log external span for query gRPC requests

crepererum:
This should simplify the correlation with our binlog data.


mkmik (approved):

trace_http/src/ctx.rs
```diff
@@ -278,25 +281,52 @@ fn required_header<T, F: FnOnce(&str) -> Result<T, DecodeError>>(
...
 }
 
+/// Span context from external source together with some flags.
+#[derive(Debug, Clone)]
+pub struct ExternalSpanContext {
+    /// The context.
+    pub ctx: SpanContext,
+
+    /// If we should also sample based on this context (i.e. emit child spans).
+    pub sampled: bool,
```
alamb, line 291: I wonder if the `sampled` flag could/should be put on the `SpanContext` struct directly?

alamb (commented):
Looks fine to me -- I think @tustvold  should have a review of this code as well (though that would be fine to do after merge)

trace_http/src/ctx.rs
crepererum: The context is used for every `Span` and the remaining tracing stack doesn't care about it (we only have one sampling decision at the "top level", not for individual span-sub-trees). So I think this is only relevant at this specific point.

service_grpc_influxrpc/src/service.rs
```diff
@@ -246,6 +248,7 @@ where
             %db_name,
             ?req.range,
             predicate=%req.predicate.loggable(),
+            trace=%external_span_ctx.format_jaeger(),
```
tustvold, line 251: It would be more "normal" to just log the trace ID, and not the span information here. Especially as if the trace isn't sampled, the span information is completely useless

trace_http/src/ctx.rs
```diff
@@ -278,25 +281,52 @@ fn required_header<T, F: FnOnce(&str) -> Result<T, DecodeError>>(
     })
 }
 
+/// Span context from external source together with some flags.
+#[derive(Debug, Clone)]
+pub struct ExternalSpanContext {
```
tustvold, line 286: I think I would prefer something opaque and called something like `RequestLogContext` or something. Otherwise it makes it confusing as to why you have two different SpanContext and when you might use one over the other

trace_http/src/tower.rs
```diff
@@ -111,14 +111,22 @@ where
...
+            Ok(Some(external_ctx)) => {
+                request.extensions_mut().insert(external_ctx.clone());
 
-                // Add context to request for use by service handlers
-                request.extensions_mut().insert(span.ctx.clone());
+                let ExternalSpanContext { ctx, sampled } = external_ctx;
 
-                // Create Span to use to instrument request
-                Some(span)
+                if sampled {
```
tustvold, line 119: Could use Option::then

tustvold (approved):
I think this is probably fine, there is something definitely a bit off here though with how this library handles the sampled flag. It might be worth revisiting the design around this, e.g. to put sampled on SpanContext directly.This would be inline with most other tracing libraries, and is likely a key part of making unilateral sampling and trace propagation work correctly

trace_http/src/tower.rs
```diff
@@ -112,13 +112,18 @@ where
...
         let span = match self.trace_header_parser.parse(collector, request.headers()) {
             Ok(Some(ctx)) => {
-                let span = ctx.child("IOx");
+                request
+                    .extensions_mut()
+                    .insert(RequestLogContext(ctx.clone()));
 
-                // Add context to request for use by service handlers
-                request.extensions_mut().insert(span.ctx.clone());
+                ctx.sampled.then(|| {
```
tustvold, line 119: Theoretically the `SpanRecorder` should be the thing that responds to the `sampled` flag, e.g. if in servicing this request we make external requests we will want to propagate the span context even if it isn't being sampled, but definitely something that can be dealt with later

trace_http/src/ctx.rs
```diff
@@ -278,25 +274,46 @@ fn required_header<T, F: FnOnce(&str) -> Result<T, DecodeError>>(
     })
 }
 
+/// Span context from external source together with some flags.
+#[derive(Debug, Clone)]
+pub struct RequestLogContext(pub SpanContext);
```
tustvold, line 279:
```suggestion
pub struct RequestLogContext(SpanContext);
```

tustvold (approved):

trace/src/ctx.rs
```diff
@@ -54,6 +54,9 @@ pub struct SpanContext {
     pub links: Vec<(TraceId, SpanId)>,
 
     pub collector: Option<Arc<dyn TraceCollector>>,
+
+    /// If we should also sample based on this context (i.e. emit child spans).
+    pub sampled: bool,
```
tustvold, line 59: I wonder if we could make `sampled` implicit in if `collector.is_some()`?

trace_http/src/ctx.rs
crepererum: done

trace_http/src/tower.rs
crepererum: I think we wanna design some kind of policy system at one point that allows us to trace subsystems but that avoid creating all the spans if we don't wanna sample anything. So let's keep the flag for now and later design a policy around it (e.g. also one that allows some "sample always, no matter what the context flag was").

trace/src/ctx.rs
crepererum: See [this comment](https://github.com/influxdata/influxdb_iox/pull/5187#discussion_r932171286).

PR #5188, opened 2022-07-22
feat: Hook up new compaction algorithm

### LINKED ISSUE #5121: Given an organized subset of L0+L1 files, compact into a set of new L1 files

carols10cents:
Closes #5121 

🚨🚨 This will turn on the new-new compaction algorithm, so don't merge this until we're ready to deploy 🚨🚨

TODO:

- There are 2 metrics on `Compactor` that aren't being used on this branch: `compaction_output_counter` and `compaction_duration`. Should they be set up and used or should they be removed? 
- Are there other metrics that we need to get useful information about how compaction is performing?


NGA-TRAN (approved):
Looks great. I have gone over this carefully.

- I have pointed out where the `compaction_duration` should go
-  Let us remove `compaction_output_counter`
- When you are done with the 2 above, you should merge this. I have modified IDPE repo to turn off Compactor. We do not have to worry it will get deployed automatically. When we are ready, I will turn Compactor on in IDPE and after that it will get deployed

NGA-TRAN:
Hmm, clicking "Merge pull request" causes Semantic error. I will need to resolve this first

NGA-TRAN:
And also a test failure. Let me try to create a new branch and do cherry pick and see

carols10cents:
Closing this in favor of https://github.com/influxdata/influxdb_iox/pull/5194 !

PR #5192, opened 2022-07-22
fix: compute_split_time returns one value when min_time = max_time

carols10cents:
@NGA-TRAN I don't know how important this is, but when I was setting up a test for splitting compacted files, I forgot to set min times and max times on the parquet files, so `compute_split_time` returned an empty vector.

[The code I'm working on restoring that used `split_times` looks like it assumes `compute_split_time` won't return an empty vector.](https://github.com/influxdata/influxdb_iox/pull/5070/files#diff-fc1cb6e18f14c778cd65d268b59f3ca5de909ff6112c32ed19d93ffe61579a6eL911)

I don't know how likely this is to happen in production; one scenario I can think of though is a large amount of data (so that there are files to compact) that all has the same timestamp (backfilling, misconfiguration, etc)

I'm not sure the right thing to do in this case. Should `compute_split_time` do something else than what it currently does? Should the code using `compute_split_time` check for empty and... panic? Return an error? Split the data some other way?

I'm not sure if this test is useful to actually merge in or not.


compactor/src/utils.rs
```diff
@@ -314,4 +314,23 @@ mod tests {
...
+
+        let total_size = 200;
+        let max_desired_file_size = 100;
+
+        let result = compute_split_time(min_time, max_time, total_size, max_desired_file_size);
+
+        // what should `compute_split_time_do?
+        // Right now, it returns an empty vector...
+        assert!(result.is_empty());
+    }
```
NGA-TRAN, line 335: This is a good question. The important invariant is all level-1 files (or above if we happen to add more levels) must not overlap. Because of that the idea of splitting is to produce many strictly non-overlapped level-1 output files. If we cannot split them, we have to accept to accept they are large for now.

In the future, we might think about a way to improving this by upgrading it to a higher level but it is way more complicated and we do not want to think about that right now

NGA-TRAN: I think in the code, you throw a log warning for this case

NGA-TRAN: And use compact_plan rather than split_plan

NGA-TRAN: Or we can also change`compute_split_time` to return max_time if it cannot be split but we have to check the output and invoke compact_plan instead. That was that I did before

NGA-TRAN: After some thinking with what we have done, return empty is a bug. We should check if `min_time == max_time` and return `max_time` meaning nothing should be split. Consistent with the check `Too small to split`

carols10cents: That makes sense. I'll roll this into my next PR, thanks!

NGA-TRAN: Done and will merge this

NGA-TRAN (approved):

PR #5195, opened 2022-07-22
feat: split large compactions

carols10cents:
Connects to #5121 

Ok, I think I have splitting large compactions into multiple compacted files working 🤞🏻


NGA-TRAN:
I am reviewing this

compactor/src/parquet_file_combining.rs
```diff
@@ -90,6 +66,19 @@ pub(crate) async fn compact_parquet_files(
...
+    // Percentage of desired max file size. This percentage of `max_desired_file_size_bytes` is
+    // considered "small" and will not be split. 100 + this percentage of
+    // `max_desired_file_size_bytes` is considered "large" and will be split into files roughly of
+    // `max_desired_file_size_bytes`. For amounts of data between "small" and "large", the data
+    // will be split into 2 parts with roughly `split_percentage` in the earlier compacted file and
+    // 1 - `split_percentage` in the later compacted file.
+    percentage_max_file_size: u16,
+    // When data is between a "small" and "large" amount, split the compacted files at roughly this
+    // percentage in the earlier compacted file, and the remainder .in the later compacted file.
+    split_percentage: u16,
```
NGA-TRAN, line 81: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -175,90 +158,127 @@ pub(crate) async fn compact_parquet_files(
...
-        .context(CompactLogicalPlanSnafu)?;
+    let plan = if total_size <= small_cutoff_bytes {
+        // Compact everything into one file
+        ReorgPlanner::new(ctx.child_ctx("ReorgPlanner"))
+            .compact_plan(Arc::clone(&merged_schema), query_chunks, sort_key.clone())
+            .context(CompactLogicalPlanSnafu)?
+    } else if small_cutoff_bytes < total_size && total_size <= large_cutoff_bytes {
+        // Split compaction into two files, the earlier of split_percentage amount of
+        // max_desired_file_size_bytes, the later of the rest
+        let split_time = min_time + ((max_time - min_time) * split_percentage as i64) / 100;
```
NGA-TRAN, line 173: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -175,90 +158,127 @@ pub(crate) async fn compact_parquet_files(
...
+
+        ReorgPlanner::new(ctx.child_ctx("ReorgPlanner"))
+            .split_plan(
+                Arc::clone(&merged_schema),
+                query_chunks,
+                sort_key.clone(),
+                vec![split_time],
+            )
+            .context(CompactLogicalPlanSnafu)?
```
NGA-TRAN, line 182: Here is the situation we ran into before and we have added a panic if that happens but we do not want to hit that panic.
If the split_time includes only one value which is this case and that values is `max_time`, we want to run `compact_plan` instead. The reason is `split_plan` will create 2 streams one with 100% of data and one is empty. The empty one will lead to the creation of an empty parquet file which later will cause issue while reading its non-sense metadata.

So I guess yo may need an if condition here, one for compact_plan and the other for split_plan

NGA-TRAN: Just like you did below

compactor/src/parquet_file_combining.rs
```diff
@@ -424,6 +546,7 @@ mod tests {
             level_1_with_duplicates.parquet_file,
             level_0_max_seq_1.parquet_file,
             level_1_file.parquet_file,
+            medium_file.parquet_file,
```
NGA-TRAN, line 549: I think it will easier to follow if you also add level_0 in the name of the medium and large

compactor/src/parquet_file_combining.rs
```diff
@@ -501,13 +630,14 @@ mod tests {
                 (2, CompactionLevel::Initial),
                 (4, CompactionLevel::FileNonOverlapped),
                 (5, CompactionLevel::Initial),
-                (6, CompactionLevel::FileNonOverlapped),
+                (6, CompactionLevel::Initial),
+                (7, CompactionLevel::FileNonOverlapped),
```
NGA-TRAN, line 634: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -524,28 +654,35 @@ mod tests {
...
         // 1 newly created CompactionLevel::FileNonOverlapped file as the result of
         // compaction
         assert_eq!(
-            (files[1].id.get(), files[1].compaction_level),
-            (6, CompactionLevel::FileNonOverlapped)
+            files_and_levels,
+            vec![
+                (5, CompactionLevel::Initial),
+                (6, CompactionLevel::Initial),
+                (7, CompactionLevel::FileNonOverlapped),
```
NGA-TRAN, line 684: I wonder if sometime next week as in a follow-up PR, we (or I)  create constant or enum in the test for value of 5, 6, 7. Something like ParquetId::Large, ParquetId::Medium, ParquetId::Small1, ParquetId::Small2. 🤔 
I think it will be easier to maintain and follow

compactor/src/parquet_file_combining.rs
```diff
@@ -574,26 +711,188 @@ mod tests {
...
+            .collect();
+        // 1 large files not included in compaction,
+        // 2 newly created CompactionLevel::FileNonOverlapped file as the result of
+        // compaction and splitting
+        assert_eq!(
+            files_and_levels,
+            vec![
+                (6, CompactionLevel::Initial),
+                (7, CompactionLevel::FileNonOverlapped),
+                (8, CompactionLevel::FileNonOverlapped),
```
NGA-TRAN, line 757: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -574,26 +711,188 @@ mod tests {
...
+                "+-----------+------+------+------+-----------------------------+",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000006Z |",
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000010Z |",
+                "| 1500      | WA   |      |      | 1970-01-01T00:00:00.000008Z |",
+                "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "| 99        | OR   |      |      | 1970-01-01T00:00:00.000012Z |",
+                "+-----------+------+------+------+-----------------------------+",
+            ],
+            &batches
```
NGA-TRAN, line 795: 🎉

compactor/src/parquet_file_combining.rs
```diff
@@ -574,26 +711,188 @@ mod tests {
...
+            .map(|f| (f.id.get(), f.compaction_level))
+            .collect();
+        // 3 newly created CompactionLevel::FileNonOverlapped file as the result of
+        // compaction and splitting
+        assert_eq!(
+            files_and_levels,
+            vec![
+                (7, CompactionLevel::FileNonOverlapped),
+                (8, CompactionLevel::FileNonOverlapped),
+                (9, CompactionLevel::FileNonOverlapped),
```
NGA-TRAN, line 841: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -574,26 +711,188 @@ mod tests {
...
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000010Z |",
+                "| 1500      | WA   |      |      | 1970-01-01T00:00:00.000008Z |",
+                "| 1601      |      | PA   | 15   | 1970-01-01T00:00:00.000030Z |",
+                "| 21        |      | OH   | 21   | 1970-01-01T00:00:00.000036Z |",
+                "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "| 99        | OR   |      |      | 1970-01-01T00:00:00.000012Z |",
+                "+-----------+------+------+------+-----------------------------+",
+            ],
+            &batches
```
NGA-TRAN, line 894: 🎉 
The way you add the large lp/parquet file and their time range helps me review this output 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -574,26 +711,188 @@ mod tests {
...
+        );
```
NGA-TRAN, line 895: I suggest to add 2 tests on data of same min and max time that leads to not splitting anything even though they are supposed to split into 2 files or many files

NGA-TRAN (commented):
Looks really good @carols10cents. I just have one suggestion for the case that avoids hitting panic and tests for it

compactor/src/parquet_file_combining.rs
carols10cents: Just to be clear, currently the only way I'm seeing to hit the panic in this case is by setting split_percentage to 100... can you think of other cases that would cause the panic?

NGA-TRAN: I am not aware of any other cases

compactor/src/parquet_file_combining.rs
```diff
@@ -175,27 +175,20 @@ pub(crate) async fn compact_parquet_files(
...
+            vec![min_time + ((max_time - min_time) * split_percentage as i64) / 100]
+        } else {
+            // Split compaction into multiple files
+            crate::utils::compute_split_time(
+                min_time,
+                max_time,
+                total_size,
+                max_desired_file_size_bytes,
+            )
+        };
```
NGA-TRAN, line 191: 👍

compactor/src/parquet_file_combining.rs
```diff
@@ -826,6 +819,85 @@ mod tests {
...
+                "| 10        | VT   |      |      | 1970-01-01T00:00:00.000010Z |",
+                "| 1500      | WA   |      |      | 1970-01-01T00:00:00.000008Z |",
+                "| 1601      |      | PA   | 15   | 1970-01-01T00:00:00.000030Z |",
+                "| 21        |      | OH   | 21   | 1970-01-01T00:00:00.000036Z |",
+                "| 270       | UT   |      |      | 1970-01-01T00:00:00.000025Z |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z |",
+                "| 99        | OR   |      |      | 1970-01-01T00:00:00.000012Z |",
+                "+-----------+------+------+------+-----------------------------+",
+            ],
+            &batches
```
NGA-TRAN, line 897: 👍

NGA-TRAN (approved):
Looks good. Thanks @carols10cents

PR #5210, opened 2022-07-26
test: Add a test for LP quoting

mkmik:
In #5183 I had to skip one test:

```
assert_eq!(get_field_value(row, 5), "some \" value");
```

because when roundtripping the escaped value back via the LP parser, I got back `"some \\\" value"`, which doesn't match the input.

This PR tests the LP builder side of the thing (encoding) independently of the LP parser, so I can be sure that the the issue is on the parsing side (will file an issue)


alamb (approved):
LGTM -- thansk!

PR #5212, opened 2022-07-26
fix: Run compactor streams in parallel to avoid deadlock

alamb:
re https://github.com/influxdata/conductor/issues/1071

# Rationale
@NGA-TRAN and I theorize that the compactor is deadlocking due to the streams not being read in parallel, similarly to https://github.com/influxdata/influxdb_iox/issues/4306

See more details in this comment https://github.com/influxdata/conductor/issues/1071#issuecomment-1195356917

# Changes
1. Apply the subtle change to run concurrently used by @domodwyer in https://github.com/influxdata/influxdb_iox/pull/4325 and https://github.com/influxdata/influxdb_iox/pull/4678 (aka collect into `FuturesOrdered`).
2. Also run the plans in separate `tokio::task`s so the CPU intensive parquet writing can be done in parallel, if possible

# Note
A whitespace blind diff https://github.com/influxdata/influxdb_iox/pull/5212/files?w=1 shows the difference more clearly I think


compactor/src/parquet_file_combining.rs
```diff
@@ -215,60 +219,89 @@ pub(crate) async fn compact_parquet_files(
...
+
+                Ok(parquet_file)
+            })
+        })
+        // NB: FuturesOrdered allows the futures to run in parallel
+        .collect::<FuturesOrdered<_>>()
+        // Check for errors in the task
+        .map(|t| t.context(ExecuteParquetTaskSnafu)?)
+        .try_collect::<Vec<_>>()
+        .await?;
```
NGA-TRAN, line 304: Oh man. Tokio spawn, async and collect FuturesOrdered. So tricky LOL

NGA-TRAN (approved):

compactor/src/parquet_file_combining.rs
alamb: Yes -- it is a small amount of code but the trickiness is very subtle....

alamb:
cc @carols10cents

Issue #5202, opened 2022-07-25
Querier chunk: projected schema interning/caching


crepererum:
While we cache and re-use table-level schemas (shared via an `Arc`), we currently produce a new `Schema` object for every querier chunk due to projection:

https://github.com/influxdata/influxdb_iox/blob/614ca5ca96f0df7fea90b611122dbad0525d056e/querier/src/chunk/mod.rs#L488-L492

In prod we have some schemas w/ >600 columns. Every columns requires at least 3 strings:

1. The column name
2. The per-column metadata key for the IOx data type
3. The per-column metadata value for the IOx data type

We also have one table w/ over 30k chunks (probably due to compaction issue, but let's not play whataboutism now). That adds up to multiple GBs of chunk-level metadata:

![Screenshot from 2022-07-25 13-23-33](https://user-images.githubusercontent.com/1529400/180810134-4f85edda-f937-4a2c-a0e2-a6760266ad47.png)

I propose to add a new `ProjectedSchema` cache for that:

```rust
Cache <
    K = (TableId, Vec<ColumnName>),
    V = Arc<Schema>,  // projected schema
    GetExtra = Arc<Schema>,  // unprojected schema
    PeekExtra = (),  // no peek
>
```

The cache loader is basically `Arc::new(schema.select_by_names(...))` (see code linked above).

alamb:
"whataboutism" --love it

alamb:
For what it is worth, I suggest adding some sort of self preservation check that doesn't even try to plan a query with 30k chunks -- it isn't likely going to work at all, even if you fix this particular problem.

Not that this ticket's suggestion isn't valuable (it is!) I am just reiterating out fixing this will likely still not allow the query with 30k files to actually complete

crepererum:
> For what it is worth, I suggest adding some sort of self preservation check that doesn't even try to plan a query with 30k chunks -- it isn't likely going to work at all, even if you fix this particular problem.

I thought about this again and still agree with the self-preservation, but the check is hard to implement without the change proposed in this ticket because we probably wanna reject the query if there are too many chunks after the query predicates are applied (= filtering via chunk stats), not if there are too many chunks in the table in general. Otherwise the prod querier would currently reject about every query.

Filtering (for the sake of simplicity and effort) is implemented using the same chunk type that the query engine uses, so we must at least be able to create them (even when we filter them out before actually using their payload within the query). But we're probably not even able to do that under all circumstances due to the issue described in this ticket.

[...]

PR #5213, opened 2022-07-26
feat: cache projected chunk schemas in querier

crepererum:
Ref #5202.

This eliminates the `chunk_parts` memory consumption:

image

I don't consider the issue fully closed because profiling the application with this PR reveals another (luckily low-hanging) fruit:

https://github.com/influxdata/influxdb_iox/blob/bbcf4ec64ed2ec6ef11e5709849fb2be6cb72fe9/schema/src/selection.rs#L38

There we unconditionally create schemas even when they can be re-used. This consumes about 50% of the planning-phase RAM in the scenario of #5202:

image

image


querier/src/chunk/mod.rs
```diff
@@ -484,12 +484,18 @@ impl ChunkAdapter {
             .iter()
             .map(|(_t, field)| field.name().as_str())
             .filter(|col| file_columns.contains(*col))
+            .map(|s| s.to_owned())
```
alamb, line 487: it is unfortunate that the projection is represented by a bunch of owned strings

This means for your chunk in a table of 30k chunks with 600 cols, each chunk will be making an additional 1.8M `Strings` mostly to then throw them away immediately.


If we are going to go through the trouble of optimizing for large numbers of chunks (which seems reasonable) how about also:
1. Key this cache on `Vec<Arc<str>>`
2. Get the table schema in the `QuerierTable` and calculate the possible column names as `Arc<Vec<Arc<str>>>` (or probably some struct that hide the mess) and reuse that Vec of strings each time?

querier/src/cache/projected_schema.rs
```diff
@@ -0,0 +1,272 @@
...
+        let backend = Box::new(LruBackend::new(
+            Box::new(HashMap::new()),
+            Arc::clone(&ram_pool),
+            CACHE_ID,
+            Arc::new(FunctionEstimator::new(
+                |k: &(TableId, Vec<String>), v: &Arc<Schema>| {
+                    let arrow_schema = v.as_arrow();
+
+                    RamSize(
+                        size_of_val(k)
```
alamb, line 79: this code is complicated enough I don't really feel confident that if we changed this code in the future it would remain correct.

What would you think about encapsulating some of this logic by using a struct for the key rather than an unnamed tuple?

querier/src/cache/projected_schema.rs
```diff
@@ -0,0 +1,272 @@
...
+                                        + field
+                                            .metadata()
+                                            .map(|md| {
+                                                md.iter()
+                                                    .map(|(k, v)| k.capacity() + v.capacity())
+                                                    .sum::<usize>()
+                                            })
+                                            .unwrap_or_default()
+                                })
+                                .sum::<usize>()
```
alamb, line 100: Given the value is `Arc`d is it it correct that the schema size is charged to this cache?

querier/src/cache/projected_schema.rs
```diff
@@ -0,0 +1,272 @@
...
+        ));
+
+        // add to memory pool
+        let backend = Box::new(LruBackend::new(
+            Box::new(HashMap::new()),
+            Arc::clone(&ram_pool),
+            CACHE_ID,
+            Arc::new(FunctionEstimator::new(
+                |k: &(TableId, Vec<String>), v: &Arc<Schema>| {
+                    let arrow_schema = v.as_arrow();
```
alamb, line 76: I would like to request moving the size estimation of `Schema` into a function like `Schema::estimated_size(&self)` so it is encapsulated better and there is a greater chance of being able to verify its correctness

alamb (approved):
I think it is a step in the right direction -- thanks @crepererum  👍 

I had several substantial suggestions that I think are worth considering prior to merge but I don't see any reason that this also couldn't be merged as is

querier/src/cache/projected_schema.rs
crepererum: Yes, we use the same approach for other caches: the cache "owns" (not in terms of Rust ownership, but semantically) the value and hands out pointers (= `Arc` clones) to cache users. So freeing a cache value may not immediately free the memory (because there still might be users) but lower the mem mid-term (eps. for the next query).

querier/src/cache/projected_schema.rs
crepererum: done

querier/src/cache/projected_schema.rs
crepererum: done

querier/src/chunk/mod.rs
crepererum: The issue is that `table_schema` is already an IOx `schema::Schema` at this point which uses an arrow schema internally. I could somehow change the code to use our own types longer, but it makes it more complicated. Overall I'm not super happy about the fact that arrow uses owned strings for column names and the metadata dict. So I would like to change that upstream and then get rid of owned strings in our stack as well (same btw. goes for the `TableSummary`), so that in the end strings are no longer duplicated.

crepererum: Tracked here: https://github.com/influxdata/influxdb_iox/issues/5221

alamb: I agree the owned strings in `Schema` could be better; I was thinking in this particular case the representation of the projection is `Vec<String>` which is a totally IOx level thing and we could fix without any changes in upstream Arrow

crepererum: The thing is: just using an `Arc` alone gives you nothing if this `Arc` isn't shared and the only places where we share them is within the schemas and the table summaries (and data fusion, but let's not go down that rabbithole for now).

alamb: Right -- that is why we would have to push the creation of `selection` up (so it was done once per `QuerierTable` rather than once per Chunk -- which is fine as the table schema is constant) and pass it down here.

PR #5214, opened 2022-07-26
refactor: avoid schema copies in `select_schema`

crepererum:
This massively helps with #5202.

Together with #5213 this now looks decent:

image

The next elephant is `build_deduplicate_plan_for_overlapping_chunks`:

image

with a similar problem, namely that it creates too many new schemas:

image


schema/src/selection.rs
```diff
@@ -153,3 +156,48 @@ mod test_super {
...
+    fn test_select_schema_all() {
+        let schema = SchemaBuilder::new()
+            .tag("t1")
+            .tag("t2")
+            .timestamp()
+            .build()
+            .unwrap()
+            .as_arrow();
+        let actual = select_schema(Selection::All, &schema);
+        assert!(Arc::ptr_eq(&schema, &actual));
```
alamb, line 176: nice tests 🎉

schema/src/selection.rs
```diff
@@ -153,3 +156,48 @@ mod test_super {
...
+            .tag("t2")
+            .timestamp()
+            .build()
+            .unwrap()
+            .as_arrow();
+
+        // normal selection
+        let actual = select_schema(Selection::Some(&["time", "t2"]), &schema);
+        let expected = schema.project(&[2, 1]).unwrap();
+        assert_eq!(actual.as_ref(), &expected);
```
alamb, line 192: should we also maybe `assert!(!Vec::ptr_eq(&schema, &actual)`?

alamb (approved):
👨‍🍳 👌 

very nice

schema/src/selection.rs
crepererum: done

PR #5215, opened 2022-07-26
feat: intern schemas during query planning

crepererum:
Helps with #5202.

With #5213 and #5214 this looks good now:

Screenshot from 2022-07-26 16-30-02


schema/src/interner.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl SchemaInterner {
+    /// Create new, empty interner.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Intern schema.
+    pub fn intern(&mut self, schema: Schema) -> Arc<Schema> {
+        let key = serde_json::to_string(&schema.as_arrow()).expect("schema should be serializable");
```
alamb, line 24: ouch -- converting to JSON to intern the schema lol

If this turns to be too expensive, another way to avoid this might be to implement a wrapper around `Schema` that implements `Hash` as Schema is already `Eq`: https://docs.rs/arrow/19.0.0/arrow/datatypes/struct.Schema.html#impl-Eq-for-Schema

and then use a `HashSet`

Then we could potentially upstream the Hash implementation.

iox_query/src/provider.rs
```diff
@@ -1233,19 +1255,23 @@ impl Deduplicater {
             }
         }
 
-        Arc::new(schema_merger.build())
+        schema_interner.intern(schema_merger.build())
```
alamb, line 1258: 🤔  I wonder if we could put the `SchemaInterner` as an optional field on `SchemaMerger` -- as a way to make it more discoverable

Something like:
```rust
let mut schema_merger = SchemaMerger::new()
  .with_interner(schema_interner);
...

Arc::new(schema_merger.build())
```

Just a thought -- definitely not required

alamb (approved):

iox_query/src/provider.rs
crepererum: I don't really wanna promote the interner to some standard, because even if we implement `Hash` and `Eq` for the arrow schema, it will always be a rather expensive way of interning (see performance note within the interner) and I would rather use some proxy key if we can (this is for example what the querier projected schema cache does: it uses the table ID as a proxy).

schema/src/interner.rs
crepererum: https://github.com/apache/arrow-rs/pull/2183

iox_query/src/provider.rs
alamb: ok -- I would not describe the querier projected schema cache as  "interning" schemas. The same schema (e.g. that would evaluate to `==`) can be repeated across different tables with the same set of columns.

crepererum:
Converting this to a draft, I'm waiting for a new arrow release that contains the required `Hash` impl. so we can avoid the serialization during interning.

alamb:
ETA for next arrow release to make it into IOx is early next week

schema/src/interner.rs
crepererum: done

iox_query/src/provider.rs
crepererum: done

PR #5219, opened 2022-07-27
fix: Clarify error msg: line number is 1-based

mkmik:
Helps with https://github.com/influxdata/conductor/issues/1053
Helps with https://github.com/influxdata/conductor/issues/1021

When debugging errors like:

```
{"code":"invalid","message":"failed to parse line protocol: error writing line 358: Unable to insert iox::column_type::field::string type into a column of iox::column_type::tag"}
```

it would be nice we don't have to second guess whether line 358 is 0-based or 1-based.


crepererum (approved):

alamb (commented):
👌

PR #5223, opened 2022-07-27
feat: limit per-table chunk data for every query

### LINKED ISSUE #5211: Querier: self-protection against too much data
As suggested by @alamb [here](https://github.com/influxdata/influxdb_iox/issues/5202#issuecomment-1194354077), we should return an error for a query if the querier thinks that too much data is involved (either by number of chunks or by some estimated size).

This is a temporary measure until we really have our query streaming ironed out. With other words: for the time being it's better to return an error than to OOM.

crepererum:
Closes #5211.


clap_blocks/src/querier.rs
```diff
@@ -229,6 +229,17 @@ pub struct QuerierConfig {
         action
     )]
     pub max_concurrent_queries: usize,
+
+    /// Number of bytes that a single table can provide during a query.
+    ///
+    /// The byte counts are estimated values are merely for self-protection reasons.
```
alamb, line 235:
```suggestion
    /// Maximum bytes to scan for a table in a query (estimated).
    ///
    /// If IOx estimates that it will scan more than this many bytes 
    /// in a query, the query will error. This protects against potentially unbounded 
    /// memory growth leading to OOMs in certain pathological queries.
```

The "certain pathological queries" is somewhat aspirational now -- hopefully we'll get to the point where this protection is not needed for almost any query

clap_blocks/src/querier.rs
```diff
@@ -287,6 +298,11 @@ impl QuerierConfig {
     pub fn max_concurrent_queries(&self) -> usize {
         self.max_concurrent_queries
     }
+
+    /// Number of bytes that a single table can provide during a query.
```
alamb, line 302:
```suggestion
    /// Query will error if it estimated that a single table will provide more
    /// than this many bytes.
```

influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -311,6 +311,17 @@ pub struct Config {
         action
     )]
     pub querier_max_concurrent_queries: usize,
+
+    /// Number of bytes that a single table can provide during a query.
+    ///
+    /// The byte counts are estimated values are merely for self-protection reasons.
```
alamb, line 317:
```suggestion
    /// Maximum bytes to scan for a table in a query (estimated).
    ///
    /// If IOx estimates that it will scan more than this many bytes 
    /// in a query, the query will error. This protects against potentially unbounded 
    /// memory growth leading to OOMs in certain pathological queries.
```

iox_query/src/provider.rs
```diff
@@ -72,6 +72,9 @@ pub enum Error {
 
     #[snafu(display("Internal error: Can not group chunks '{}'", source,))]
     InternalChunkGrouping { source: self::overlap::Error },
+
+    #[snafu(display("Too much data."))]
```
alamb, line 76: Since this will be returned to the user, I feel like some more specifics are warranted:

```suggestion
    #[snafu(display("Query would scan {} bytes, more than configured maximum {} bytes. Try adjusting your compactor settings or increasing the per query memory limit.", maximum_size_limit))]
```

alamb (approved):
While not ideal, I think this will be a nice way to get the querier OOM's in production under control in the short term

alamb:
Nice work -- I have only some description / documentation suggestions

PR #5230, opened 2022-07-27
feat: run many compact partitions in parallel

### LINKED ISSUE #5228: Make many `compact a partition` runs in parallel
# Rationale
Now the `compact a partition` works nicely, let us make many of them run concurrently to have a better use CPU & memory. and to speed up the Compactors

NGA-TRAN:
Closes #5228 


NGA-TRAN:
@alamb : I reuse the code before. Do you think it makes sense or we should do something differently?

Test question: We have not added any tests for this so far. I am going to add one while you have a quick review for early feedback

compactor/src/handler.rs
```diff
@@ -283,23 +283,28 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
...
+        // Repeat compacting n partitions in parallel until all candidates are compacted.
+        // Concurrency level calculation (this is estimated from previous experiments. The actual resource
+        // management will be more complicated and a future feature):
+        //   . Each `compact partititon` takes max of this much memory input_size_threshold_bytes
+        //   . We have this memory budget: max_concurrent_compaction_size_bytes
+        // --> num_parallel_partitions = max_concurrent_compaction_size_bytes/ input_size_threshold_bytes
+        let num_parallel_partitions = (compactor.config.max_concurrent_compaction_size_bytes
+            / compactor.config.input_size_threshold_bytes)
+            as usize;
+        while !candidates.is_empty() {
```
alamb, line 297: I think this code will work but it will run jobs in 'batches' and will not move on to the next batch until the largest job is complete. 

I think we could potentially finagle this using tasks / futures magic and let rust handle the "keep at most `num_parallel_partitions`" busy 

I'll give it a try and see if I can make it work

alamb: Here is a proposed way with futures: https://github.com/influxdata/influxdb_iox/pull/5232

I am a little worried about the futures debuggability 🤔

alamb (approved):
I feel bad about approving this as I contributed some to it, but I think since @NGA-TRAN  is reviewing it too when she is happy with the code it will have been sufficiently reviewed

alamb:
🚢 🇮🇹 🥳 🦜

alamb:
Well it seems to be using more CPU 🎉 
Screen Shot 2022-07-27 at 6 10 34 PM

PR #5231, opened 2022-07-27
fix: interpret [MIN_NANO_TIME, MAX_NANO_TIME) range as all time

lesam:
(For optimized metadata queries only)

InfluxQL queries can send (technically incorrect) ranges like this, meaning all time
but excluding the max nanosecond time.

Since this is an important case, we should handle it specially and use the optimized
'all time' handling for meta queries even though this is technically wrong in that
it does not filter out column names / measurement names at MAX_NANO_TIME exactly.

Closes: https://github.com/influxdata/conductor/issues/1072


crepererum (approved):
:shipit: :tada:

alamb:
Thanks @lesam

PR #5232, opened 2022-07-27
refactor: Use rust futures fu to run compactor jobs in parallel

alamb:
Here is a proposed way of implementing compact in parallel using futures: https://github.com/influxdata/influxdb_iox/pull/5230

It may be too fancy (aka opaque / hard to debug) but I wanted to offer it


alamb:
Note this PR is targeted at https://github.com/influxdata/influxdb_iox/pull/5230

compactor/src/handler.rs
```diff
@@ -294,17 +294,42 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
...
+            // num_parallel_partitions compactions in parallel
+            .buffer_unordered(num_parallel_partitions)
+            // report any JoinErrors (aka task panics)
+            .map(|join_result| {
+                if let Err(e) = join_result {
+                    warn!(?e, "compaction task failed");
+                }
+                Ok(())
+            })
+            // Errors are reported during execution
```
alamb, line 326:
```suggestion
            // Errors are reported during execution, so ignore results here
```

compactor/src/handler.rs
```diff
@@ -294,17 +294,42 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
...
+                        }
+                        Ok(_) => {
+                            debug!(?partition_id, "compaction complete");
+                        }
+                    };
+                })
+            })
+            // Assume we have enough resources to run
+            // num_parallel_partitions compactions in parallel
+            .buffer_unordered(num_parallel_partitions)
```
NGA-TRAN, line 318: Does it mean the `futures::stream::iter(candidates)` will continue until all candidates are run? A new candidate is run right after the previous one finishes?
Neat!

NGA-TRAN (approved):
So neat

compactor/src/handler.rs
alamb: > Does it mean the futures::stream::iter(candidates) will continue until all candidates are run? A new candidate is run right after the previous one finishes?

Yes -- that is my understanding, up to `num_parallel_partitions` at any one time

Specifically it means that up to `num_parallel_partitions` will be 'resolved' at any time (since they are all futures 🤯 ) 

https://docs.rs/futures/0.3.21/futures/stream/trait.StreamExt.html#method.buffer_unordered is the docs page

PR #5235, opened 2022-07-28
feat: add a few helpful compactor debug logs

crepererum:


compactor/src/handler.rs
```diff
@@ -289,6 +291,7 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
 
         let n_candidates = candidates.len();
         if n_candidates == 0 {
+            debug!("no compaction candidates found");
```
NGA-TRAN, line 294: Hah, @alamb may not like to see this a lot. But I guess one every second now maybe not that bad.

NGA-TRAN (approved):

compactor/src/handler.rs
alamb: I can deal with 1 a second ;)

alamb (approved):

PR #5249, opened 2022-08-01
fix: do not panic on invalid timestamp ranges

crepererum:
Timestamp ranges come from "untrusted" inputs (via gRPC) and must not
lead to panics. The only case where this could happen is at `start >
end`. Let's just set `start = end` in this case. Reaonsing:

- Semantically this is a sound range, since this is only a somewhat
  degenerated case of "empty".
- We already allow `start = end` to represent "empty" ranges.
- We already clamp (and therefore modify) `start` to the valid range.

Fixes https://github.com/influxdata/conductor/issues/1080.


data_types/src/lib.rs
```diff
@@ -2111,14 +2111,12 @@ impl TimestampRange {
...
+    /// Clamps `start` to [`MIN_NANO_TIME`].
+    /// end is unclamped - end may be set to `i64:MAX == MAX_NANO_TIME+1` to indicate no restriction on time.
     ///
-    /// # Panic
-    /// Panics if `start > end`.
+    /// If `start > end`, this will be interpreted as an empty time range and `start` will be set to `end`.
     pub fn new(start: i64, end: i64) -> Self {
-        assert!(end >= start, "start ({start}) > end ({end})");
-        let start = start.max(MIN_NANO_TIME);
+        let start = start.max(MIN_NANO_TIME).min(end);
```
alamb, line 2119: I worry that this will silently ignore errors that someone should be looking into. 

Should we add a `warn!` in this case? 

My rationale is that this case likely signals a bug in the client. But then maybe we should avoid spam in the iox log 🤔 

However, I don't have a single unified perfect suggestion, so I think this is fine too

alamb (approved):

data_types/src/lib.rs
crepererum: Yeah, I would avoid spamming logs due to client imperfections.

PR #5250, opened 2022-08-01
refactor: increase batch size when reading parquet

crepererum:
This reduced our overhead when reading parquet files quite a lot.

In some internal benchmark, this reduces the size to perform a single
series aggregation of a rather large series with cold caches from 58s to
48s for cold caches. No real difference could be measured for warm
caches (~21ms for both).

This should also help the compactor since the record batches should be
larger.


tustvold (approved):
FWIW you likely want to ensure this number is a divisor of the max row group size, as otherwise there will be additional overheads

crepererum:
> FWIW you likely want to ensure this number is a divisor of the max row group size, as otherwise there will be additional overheads

Done.

alamb (commented):
👌

PR #5252, opened 2022-08-01
refactor: disable querier read buffer cache

crepererum:
Let's try and see how this performs in prod.


alamb (approved):
I think it is acceptable to not have tests for behavior at the moment -- over time it would be nice to figure out how to verify the performance of the querier over time, but that is well outside of the scope of this PR ;)

Issue #5128, opened 2022-07-14
Ingester drops more writes than needed when Kafka offset is missing


crepererum:
If the ingester receives a `WriteBufferErrorKind::UnknownSequenceNumber` it concludes that the retention policy may have deleted the data and automatically jumps to the newest entry:

https://github.com/influxdata/influxdb_iox/blob/512f9850ee575c153ebd89ca8f53d7e290f0a410/ingester/src/stream_handler/handler.rs#L242-L263

Kafka (and other write buffers) however have a closed range of known sequence numbers though. The conclusion is only right on the lower end, i.e. if the known range is above the requested sequence number. If the requested sequence number was too new, then this either means we encountered a write buffer bug (e.g. https://github.com/influxdata/rskafka/issues/147 ) or something else is wrong (catalog data is wrong, some manual Kafka operation, ...). In that case we must bubble up the error.

May require https://github.com/influxdata/rskafka/issues/148.

jacobmarble:
Before implementing a fix, please consider matching [the behavior and log message that TSM uses](https://github.com/influxdata/idpe/blob/45701c0f522d3f0211214bfdf439e8979b0f7aee/storage/store/store.go#L689-L697). A Graylog alert checks for this error message, and notifies the team when it is found.

jacobmarble:
Having slept on the subject, I'm satisfied that we should not try to alert on the log. Alert (loudly) on a metric. The data is likely lost forever, but person receiving the page (1) should notify affected customers via support, and (2) might be able to prevent more such errors with config adjustments.

crepererum:
The metrics part now lives in #5321

PR #5253, opened 2022-08-01
feat: sequencer reset counter.

crepererum:
This is the diagnostics part of #5128 (the actual bug still exists though).

### DIFF
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -164,9 +164,10 @@ pub struct SequencerId(i64);
 
 #[allow(missing_docs)]
 impl SequencerId {
-    pub fn new(v: i64) -> Self {
+    pub const fn new(v: i64) -> Self {
         Self(v)
     }
+
     pub fn get(&self) -> i64 {
         self.0
     }
@@ -185,9 +186,10 @@ pub struct KafkaPartition(i32);
 
 #[allow(missing_docs)]
 impl KafkaPartition {
-    pub fn new(v: i32) -> Self {
+    pub const fn new(v: i32) -> Self {
         Self(v)
     }
+
     pub fn get(&self) -> i32 {
         self.0
     }
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -716,15 +716,23 @@ pub fn column_type_from_field(field_value: &FieldValue) -> ColumnType {
 pub struct Sequencer {
     /// the id of the sequencer
     pub id: SequencerId,
+
     /// the topic the sequencer is reading from
     pub kafka_topic_id: KafkaTopicId,
+
     /// the kafka partition the sequencer is reading from
     pub kafka_partition: KafkaPartition,
+
     /// The minimum unpersisted sequence number. Because different tables
     /// can be persisted at different times, it is possible some data has been persisted
     /// with a higher sequence number than this. However, all data with a sequence number
     /// lower than this must have been persisted to Parquet.
     pub min_unpersisted_sequence_number: SequenceNumber,
+
+    /// How often a sequencer was reset because the ingester was running behind the persistence window.
+    ///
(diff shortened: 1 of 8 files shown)


data_types/src/lib.rs
```diff
@@ -716,15 +718,23 @@ pub fn column_type_from_field(field_value: &FieldValue) -> ColumnType {
...
     /// The minimum unpersisted sequence number. Because different tables
     /// can be persisted at different times, it is possible some data has been persisted
     /// with a higher sequence number than this. However, all data with a sequence number
     /// lower than this must have been persisted to Parquet.
     pub min_unpersisted_sequence_number: SequenceNumber,
+
+    /// How often a sequencer was reset because the ingester was running behind the persistence window.
+    ///
+    /// This indicates data loss.
+    pub reset_count: i32,
```
domodwyer, line 737: How come we store this? It seems weird that we need to track the value of this metric independently of the metric system.

ingester/src/stream_handler/handler.rs
```diff
@@ -60,30 +62,43 @@ pub struct SequencedStreamHandler<I, O, T = SystemProvider> {
...
+
     /// Log context fields - otherwise unused.
     kafka_topic_name: String,
     kafka_partition: KafkaPartition,
 
     skip_to_oldest_available: bool,
+
+    /// Catalog to preserve some metrics accross restarts.
+    catalog: Arc<dyn Catalog>,
+    sequencer_id: SequencerId,
```
domodwyer, line 76: Coupling the catalog into an otherwise stateless dispatcher to the `Catalog` is unfortunate :(

ingester/src/stream_handler/handler.rs
```diff
@@ -135,7 +150,15 @@ impl<I, O> SequencedStreamHandler<I, O> {
             true,
         ));
 
-        Self {
+        // reset count
+        let sequencer_reset_count = metrics
+            .register_metric::<U64Counter>(
+                "sequencer_reset_count",
+                "How often a sequencer was already reset",
```
domodwyer, line 157: nit: all the other metrics have lowercase descriptions except this one, and the "Reset stream" log message

ingester/src/stream_handler/handler.rs
```diff
@@ -170,13 +200,49 @@ impl<I, O> SequencedStreamHandler<I, O> {
...
+                .await
+                .expect("retry forever")
+        } else {
+            Backoff::new(&Default::default())
+                .retry_all_errors("get reset count", || async {
+                    let mut repos = self.catalog.repositories().await;
+                    repos.sequencers().get_reset_count(self.sequencer_id).await
+                })
+                .await
+                .expect("retry forever")
```
domodwyer, line 236: This seems like an init-once bit of code - wouldn't it live in the setup funcs? I couldn't find another call of `update_reset_count(false)` which would make using this fn a clearer (no confusing `update: false` bool).

data_types/src/lib.rs
domodwyer: After speaking to @crepererum on slack, it turns out we want a push-based method to record events that have potential data loss implications.

The current metric system is a pull-based model, and the relatively high scrape interval leaves room for events like this to be lost should the ingester crash before the metrics are scraped.

This PR attempts to solve this problem by persisting this new counter in the catalog - however all the metrics marked `potential_data_loss=true` (and likely more) should also be tracked in a persistent way if we're trying to solve this issue as they all have the same implication: data loss! :fire:

Ideally we split this PR into two - one PR that adds the metric for diagnostics, and another than implements this counter persistence (or a different solution to the metric loss window problem).

crepererum: See #5286.

crepererum:
Closing this for now. The metric was added in #5286 and the persistence will be handled by #5321.

PR #5254, opened 2022-08-01
chore: reduce log info

NGA-TRAN:
To much unnecessary log info


parquet_file/src/storage.rs
```diff
@@ -235,7 +235,7 @@ async fn download_and_scan_parquet(
     object_store: Arc<DynObjectStore>,
     tx: tokio::sync::mpsc::Sender<ArrowResult<RecordBatch>>,
 ) -> Result<(), ReadError> {
-    debug!(?path, "Start parquet download & scan");
+    trace!(?path, "Start parquet download & scan");
```
NGA-TRAN, line 238: We have a complete log which is good enough

crepererum (approved):

PR #5256, opened 2022-08-01
feat: add metric for TTL cache expiration

crepererum:
I'm missing this data point to judge the current querier behavior.


cache_system/src/backend/ttl.rs
```diff
@@ -155,6 +167,7 @@ where
             .unwrap_or_default()
         {
             let (k, _, _t) = self.expiration.pop().unwrap();
+            self.metric_expired.inc(1);
```
alamb, line 170: 👍

alamb (approved):
LGTM -

PR #5257, opened 2022-08-01
feat: add metric for predicate-based cache entry removal

crepererum:
Similar to #5256 but this time for entries removed by predicates. This concern mostly entries that are removed if we think they are outdated/out-of-sync.


cache_system/src/backend/shared.rs
```diff
@@ -24,9 +26,21 @@ where
...
-    pub fn new(inner_backend: Box<dyn CacheBackend<K = K, V = V>>) -> Self {
+    pub fn new(
+        inner_backend: Box<dyn CacheBackend<K = K, V = V>>,
+        name: &'static str,
+        metric_registry: &metric::Registry,
+    ) -> Self {
+        let metric_removed_by_predicate = metric_registry
+            .register_metric::<U64Counter>(
+                "cache_removed_by_predicate",
+                "Number of entries removed from a cache via a filter predicate",
```
alamb, line 37: For someone unfamiliar with how this code works, it may not be clear what a `'filter predicate'` is. 

```suggestion
                "Number of entries removed from a cache via a custom condition",
```

I am not sure 'custom' is really all that much better than `filter` 🤔

alamb (approved):

cache_system/src/backend/shared.rs
crepererum: :+1:

PR #5267, opened 2022-08-02
chore: Expose a C API for the IOx LP parser

mkmik:
Can be useful to call the IOx LP parser from other processes, for example from Go.
I used it to run an online comparison of IOx and influxdb Go LP parser in order to identify compatibility
issues.


influxdb_line_protocol/src/ffi.rs
```diff
@@ -0,0 +1,22 @@
+use super::{parse_lines, Result};
+use std::ffi::{CStr, CString};
+
+#[no_mangle]
+pub unsafe extern "C" fn validate_lines(lp: *const libc::c_char) -> *const libc::c_char {
```
crepererum, line 5: Until https://github.com/rust-lang/rust/issues/74990 is stabilized and rustc will auto-abort when unwinding throw a `"C"` ABI boundary, we should probably wrap the Rust code with [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html).

Also see https://github.com/rust-lang/rust/issues/52652 and https://doc.rust-lang.org/nomicon/ffi.html#ffi-and-panics

influxdb_line_protocol/Cargo.toml
```diff
@@ -4,8 +4,15 @@ version = "0.1.0"
 authors = ["Paul Dix <paul@pauldix.net>"]
 edition = "2021"
 
+[lib]
+crate-type = ["dylib", "staticlib"]
```
crepererum, line 8: I think `staticlib` is not going to work for our default target, because there libc is dynamically linked. You probably wanna use `cdylib` and move the whole FFI code into its own crate (not sure if Rust allows you to easily mix "normal" Rust interfaces and FFI within the same crate).

mkmik: Can I force a static lib crate type from a cli flag?

crepererum: Ah wait, I confused "static lib" and "static binary". `staticlib` should work for the default target.

mkmik: But the CI is unhappy

crepererum: I think the type wrong though. Reading throw https://doc.rust-lang.org/reference/linkage.html I think you want these two:

- `staticlib`: for you C/Go stuff
- `rlib`: to use this crate from other crates within the same workspace

mkmik: That makes sense thanks

mkmik:
@crepererum PTAL

influxdb_line_protocol/src/ffi.rs
mkmik: done; PTAL

influxdb_line_protocol/Cargo.toml
mkmik: done

influxdb_line_protocol/src/ffi.rs
crepererum: Can you please move the arg parsing (`let bytes = ...` and `let str = ...`) into the catch_unwind as well? I suspect they are most likely to panic if some user calls this with some weird arguments.

crepererum (approved):

Issue #5221, opened 2022-07-27
Remove owned strings from schemas and summaries


crepererum:
As pointed out by @alamb [here](https://github.com/influxdata/influxdb_iox/pull/5213#discussion_r930394885), we still have a bunch of owned strings in our stack because the arrow and datafusion ultimately use them.

Let's change the upstream `String` impl. to `Arc<str>` (e.g. in the arrow schema) and adapt the IOx code to use `Arc<str>` as well.

PR #5271, opened 2022-08-02
feat: use two RAM pools in querier

crepererum:
Quick&Dirty implementation of a RAM-pool split to see if this has any
effect. I expect the querier performance to improve due to this because
large read buffers can no longer evict precious metadata.


querier/src/cache/mod.rs
```diff
@@ -114,60 +125,60 @@ impl CatalogCache {
...
             Arc::clone(&metric_registry),
-            Arc::clone(&ram_pool),
+            Arc::clone(&ram_pool_payload),
             testing,
         );
         let projected_schema_cache = ProjectedSchemaCache::new(
             Arc::clone(&time_provider),
             &metric_registry,
-            Arc::clone(&ram_pool),
+            Arc::clone(&ram_pool_metadata),
```
alamb, line 181: Should this be on `ram_pool_payload` as well? I double checked the Querier dashboard and it seems like the projected schemas are quite small so this is probably good I just wanted to check

alamb (approved):

querier/src/cache/mod.rs
crepererum: IMHO it's metadata and is required to create chunks before they are pruned, so this is quite important. And if this gets to large we should probably work on #5221.

PR #5275, opened 2022-08-02
feat: Add information about profiling using Instruments on macOS

stuartcarnie:
Add macOS specific documentation for profiling `influxdb_iox` using Instruments, with additional codesign steps for aarch64 / Si machines.


domodwyer (approved):
Thank you for this :heart:

PR #5283, opened 2022-08-03
feat: make querier RAM pool split a proper feature

crepererum:
Rationale: Splitting the querier cache into "metadata" (to avoid catalog lookups) and "data" (to avoid getting parquet files from object store) improves the cache hit ratio and thus the effectiveness of the metadata caching and query performance

- use propper pool names
- expose sizing via CLI/env

Closes https://github.com/influxdata/conductor/issues/1102.


clap_blocks/src/querier.rs
```diff
@@ -212,14 +212,23 @@ pub struct QuerierConfig {
     )]
     pub sequencer_to_ingesters: Option<String>,
 
-    /// Size of the RAM cache pool in bytes.
+    /// Size of the RAM cache pool for metadata in bytes.
```
alamb, line 215:
```suggestion
    /// Size of the RAM cache used to store catalog metadata information in bytes.
```

clap_blocks/src/querier.rs
```diff
@@ -212,14 +212,23 @@ pub struct QuerierConfig {
...
-        default_value = "1073741824",
+        long = "--ram-pool-metadata-bytes",
+        env = "INFLUXDB_IOX_RAM_POOL_METADATA_BYTES",
+        default_value = "134217728",  // 128MB
         action
     )]
-    pub ram_pool_bytes: usize,
+    pub ram_pool_metadata_bytes: usize,
+
+    /// Size of the RAM cache pool for payload in bytes.
```
alamb, line 224:
```suggestion
    /// Size of the RAM cache used to store data in bytes.
```

clap_blocks/src/querier.rs
```diff
@@ -212,14 +212,23 @@ pub struct QuerierConfig {
...
+    #[clap(
+        long = "--ram-pool-payload-bytes",
+        env = "INFLUXDB_IOX_RAM_POOL_PAYLOAD_BYTES",
```
alamb, line 227:
```suggestion
        env = "INFLUXDB_IOX_RAM_POOL_DATA_BYTES",
```

clap_blocks/src/querier.rs
```diff
@@ -212,14 +212,23 @@ pub struct QuerierConfig {
...
+        env = "INFLUXDB_IOX_RAM_POOL_METADATA_BYTES",
+        default_value = "134217728",  // 128MB
         action
     )]
-    pub ram_pool_bytes: usize,
+    pub ram_pool_metadata_bytes: usize,
+
+    /// Size of the RAM cache pool for payload in bytes.
+    #[clap(
+        long = "--ram-pool-payload-bytes",
```
alamb, line 226:
```suggestion
        long = "--ram-pool-data-bytes",
```

clap_blocks/src/querier.rs
alamb: I personally think `data` would be a more specific and standard name for this type of setting -- so the caches are `metadata` and `data`.

influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -294,14 +294,23 @@ pub struct Config {
     )]
     pub compactor_grpc_bind_address: SocketAddr,
 
-    /// Size of the querier RAM cache pool in bytes.
+    /// Size of the querier RAM cache pool for metadata in bytes.
```
alamb, line 297: Same comments on names apply here too.

We could also consider simply hard coding the values for `all-in-one` mode (as people who are performance testing should probably be running the individual services and we could keep the configuration of all in one mode simpler for users) But I don't see a problem with this way either.

alamb (approved):

influxdb_iox/src/commands/run/all_in_one.rs
crepererum: done

PR #5286, opened 2022-08-03
feat: add sequencer reset count metric and log to ingester

crepererum:
Split out from #5253.

Helps with #5128.


ingester/src/stream_handler/handler.rs
```diff
@@ -772,6 +812,7 @@ mod tests {
         ],
         sink_rets = [Ok(true)],
         want_ttbr = 31,
+        want_reset = 1,
```
domodwyer, line 815: :+1:

domodwyer (approved):
Beautiful - thanks for splitting :+1:

PR #5290, opened 2022-08-03
chore: update datafusion pin

alamb:
chore: Update datafusion

# Rationale
* Keep up with upstream changes

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/9a5f17e6bbc9b8acbb0d0e584ebc9ade291622f2


Cargo.lock
```diff
@@ -4936,15 +4936,6 @@ dependencies = [
  "unicode_categories",
 ]
 
-[[package]]
```
alamb, line 4939: Removed old version of sql parser 🎉

carols10cents (approved):

Issue #5177, opened 2022-07-21
Querier error on regex predicates on `_field` throws "Unsupported _field predicate, could not rewrite for IOx"


alamb:
found with https://github.com/influxdata/conductor/issues/1059

A predicate of the form `regex_match(_field, "foo")` will currently throw an error

For example, here is a test that should pass (the expected output needs to be updated). However, if fails in the following way:

```
thread 'influxrpc::read_filter::test_read_filter_data_pred_using_regex_match_on_field' panicked at 'Unexpected error running read filter: "gRPC planner got error creating predicates: Error during planning: Unsupported _field predicate, could not rewrite for IOx"', query_tests/src/influxrpc/read_filter.rs:48:14
```

```diff
diff --git a/query_tests/src/influxrpc/read_filter.rs b/query_tests/src/influxrpc/read_filter.rs
index 22ca3fbbc..3fbc73b4c 100644
--- a/query_tests/src/influxrpc/read_filter.rs
+++ b/query_tests/src/influxrpc/read_filter.rs
@@ -470,6 +470,20 @@ async fn test_read_filter_data_pred_using_regex_match() {
     run_read_filter_test_case(TwoMeasurementsMultiSeries {}, predicate, expected_results).await;
 }
 
+#[tokio::test]
+async fn test_read_filter_data_pred_using_regex_match_on_field() {
+    let predicate = Predicate::default()
+        .with_regex_match_expr("_field", "temp");
+    let predicate = InfluxRpcPredicate::new(None, predicate);
+
+    // Should see results for temp and other_temp (but not reading)
+    let expected_results = vec![
+        "TODO"
+    ];
+
+    run_read_filter_test_case(TwoMeasurementsManyFields {}, predicate, expected_results).await;
+}
+
 #[tokio::test]
 async fn test_read_filter_data_pred_using_regex_match_with_delete() {
     let predicate = Predicate::default()
```

alamb:
Here is my basic plan to make the existing code more general (but not yet totally general)

The current approach only supports `_field = 'val'` and ``_field != 'val'` ` type predicates. It is at https://github.com/influxdata/influxdb_iox/blob/main/predicate/src/rpc_predicate/field_rewrite.rs#L54-L67

I will break up predicates into conjunctions, e.g. (expr1 AND expr2 AND expr3) --> expr1, expr2, expr3
Then for each `expr` that has a `_field` reference I will attempt to evaluate it at plan time using the DataFusion contant folder. 

[...]

PR #5298, opened 2022-08-03
refactor: simplify creating AND expressions

alamb:
A small code simplification I found while working on https://github.com/influxdata/influxdb_iox/issues/5177

We can use the built `reduce` operation to `AND` expressions together and don't need our own special builder


predicate/src/lib.rs
```diff
@@ -105,20 +105,21 @@ impl Predicate {
...
     pub fn filter_expr(&self) -> Option<Expr> {
-        let mut builder =
-            AndExprBuilder::default().append_opt(self.make_timestamp_predicate_expr());
-
-        for expr in &self.exprs {
-            builder = builder.append_expr(expr.clone());
-        }
-
-        builder.build()
+        let expr_iter = std::iter::once(self.make_timestamp_predicate_expr())
```
alamb, line 114: it is definitely more concise. I think it is also more readable but I may now be too far into Rust to really tell anymore

datafusion_util/src/lib.rs
```diff
@@ -79,51 +79,6 @@ pub fn make_range_expr(start: i64, end: i64, time: impl AsRef<str>) -> Expr {
     ts_low.and(ts_high)
 }
 
-/// Creates a single expression representing the conjunction (aka
```
alamb, line 82: I wrote this code when I was not very familiar with Rust

predicate/src/lib.rs
crepererum: It's not just about "knowing" Rust, IMHO it's also that custom code also requires you to mentally parse the expression + understand the semantics. And from just looking at the prior code it is definitely more black-box like.

crepererum (approved):

PR #5301, opened 2022-08-03
fix: Support RegExMatch and RegExNotMatch predicates on `_field`

alamb:
~Draft until I sort out the TODO items below~

Closes https://github.com/influxdata/influxdb_iox/issues/5177
Closes https://github.com/influxdata/influxdb_iox/issues/3428

# Rationale:
See https://github.com/influxdata/influxdb_iox/issues/5177 but TLDR is it turns out influxql may send predicates like `_field ~= 'count|sum'` (aka pick fields that match the regular expression count or sum)

In general, evaluating `_field_` predicates is non trivial in IOx because it must translate from the influxrpc semantics where they are "row" level predicates (really  series level)  into IOx's data model where  "filtering" _field means "select a subset of columns"

# Changes
I took the special case code for handling `_field` and made it general. The only case it does doesn't handle is mixed `_field` predicates with non `_field` predicates that are not conjuncts. For example

```sql
_field = 'foo' OR tag_col = "host-a.example.com"
```

It handles AND just fine:
```sql
_field = 'foo' AND tag_col = "host-a.example.com"
```

This limitation is tracked in https://github.com/influxdata/influxdb_iox/issues/5310 but I am not sure if this type of predicate is ever actually sent via flux or influxql. 

# TODOs:
- [x] Update run in query tests
- [x] Tests for RegExNotMatch
- [x] Tests for chained AND's with valid _field predicates
- [x] ASCII ART
- [x] Follow on ticket explaining cases we still don't handle: https://github.com/influxdata/influxdb_iox/issues/5310

P.s. I almost abandoned my [grand scheme](https://github.com/influxdata/influxdb_iox/issues/5177#issuecomment-1192924105) and tried to extend the exsiting structure, but it got so complicated that making the logic general was actually significantly simpler and supported more cases. Bonus! :bowtie:


query_tests/src/influxrpc/tag_values.rs
```diff
@@ -354,14 +354,14 @@ async fn list_tag_values_field_col_does_not_exist() {
 async fn list_tag_values_field_col_does_exist() {
     let tag_name = "state";
     let predicate = Predicate::default()
-        .with_range(0, 1000) // get all rows
-        // this field does exist (but only for rows with CA and MA, not NY)
-        .with_expr(col("_field").eq(lit("county")));
+        .with_range(0, 1000000) // get all rows
```
alamb, line 357: This test is incorrect because it was using a `_field` predicate on a `tag` (not a `field` 🤦 )

Specifically, with the original setup, `county` is a tag:

Here 

```rust 
impl DbSetup for TwoMeasurementsManyNulls {
    async fn make(&self) -> Vec<DbScenario> {
        let partition_key = "1970-01-01T00";

        let lp_lines1 = vec![
            "h2o,state=CA,city=LA,county=LA temp=70.4 100",
            "h2o,state=MA,city=Boston,county=Suffolk temp=72.4 250",
            "o2,state=MA,city=Boston temp=50.4 200",
            "o2,state=CA temp=79.0 300",
        ];
        let lp_lines2 = [
            "o2,state=NY temp=60.8 400",
            "o2,state=NY,city=NYC temp=61.0 500",
            "o2,state=NY,city=NYC,borough=Brooklyn temp=61.0 600",
        ];

        make_two_chunk_scenarios(partition_key, &lp_lines1.join("\n"), &lp_lines2.join("\n")).await
    }
}
```

In the new setup, I chose the `moisture` field as it picks out only one of the series (the one defined in `lp_lines2`):

```rust 
impl DbSetup for TwoMeasurementsManyFields {
    async fn make(&self) -> Vec<DbScenario> {
        let partition_key = "1970-01-01T00";

        let lp_lines1 = vec![
            "h2o,state=MA,city=Boston temp=70.4 50",
            "h2o,state=MA,city=Boston other_temp=70.4 250",
            "h2o,state=CA,city=Boston other_temp=72.4 350",
            "o2,state=MA,city=Boston temp=53.4,reading=51 50",
            "o2,state=CA temp=79.0 300",
        ];
        let lp_lines2 = vec!["h2o,state=MA,city=Boston temp=70.4,moisture=43.0 100000"];

        make_two_chunk_scenarios(partition_key, &lp_lines1.join("\n"), &lp_lines2.join("\n")).await
    }
}

```

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -717,14 +734,20 @@ async fn test_read_filter_on_not_field() {
...
+    // _field and the values in the "tag" column, so we can't figure
+    // out which columns ("_field" prediate) at planning time -- we
+    // have to do it at runtime somehow
+    //
+    // https://github.com/influxdata/influxdb_iox/issues/5310
+
+    // (_field != 'temp') OR (city = "Boston")
     let p1 = col("_field")
         .not_eq(lit("temp"))
-        .or(col("_field").eq(lit("other_temp")));
```
alamb, line 727: this kind of predicate is now supported 🎉

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -470,6 +470,23 @@ async fn test_read_filter_data_pred_using_regex_match() {
     run_read_filter_test_case(TwoMeasurementsMultiSeries {}, predicate, expected_results).await;
 }
 
+#[tokio::test]
+async fn test_read_filter_data_pred_using_regex_match_on_field() {
```
alamb, line 474: test case from https://github.com/influxdata/influxdb_iox/issues/5177

datafusion_util/src/lib.rs
```diff
@@ -32,6 +32,32 @@ use tokio::sync::mpsc::{Receiver, UnboundedReceiver};
 use tokio_stream::wrappers::{ReceiverStream, UnboundedReceiverStream};
 use watch::WatchedTask;
 
+/// Split an Expr up into its constituent parts
```
alamb, line 35: This is like `split_conjunction` but takes an owned `Expr`. See  https://github.com/influxdata/influxdb_iox/pull/5308

predicate/src/rpc_predicate.rs
```diff
@@ -216,31 +227,13 @@ fn normalize_predicate(
     // Store any field value (`_value`) expressions on the `Predicate`.
     predicate.value_expr = field_value_exprs;
 
-    let predicate = match field_projections.into_projection() {
```
alamb, line 219: this logic to apply field restrictions to the schema and produce a list of column names is now encapsulated inside `FieldProjectionRewriter`

predicate/src/rpc_predicate.rs
```diff
@@ -339,18 +332,36 @@ mod tests {
...
+
     #[test]
     fn test_normalize_predicate_field_rewrite_multi_field_unsupported() {
         let err = normalize_predicate(
             "table",
             Some(schema()),
             &Predicate::new()
-                // predicate is connected with `and` which is not supported
-                .with_expr(col("_field").eq(lit("f1")).and(col("_field").eq(lit("f2")))),
+                // predicate refers to a column other than _field which is not supported
```
alamb, line 355: the original test case is supported after this PR, so I had to rewrite the test to something that was not supported 😆

predicate/src/rpc_predicate.rs
```diff
@@ -370,21 +381,20 @@ mod tests {
...
                 .with_expr(col("_field").not_eq(lit("f2"))),
         )
-        .unwrap_err();
+        .unwrap();
 
-        assert_contains!(
-            err.to_string(),
-            "Unsupported _field predicates: both '=' and '!=' present"
-        );
+        let expected = Predicate::new().with_field_columns(vec!["f1"]);
```
alamb, line 395: this case now works

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
-//! Logic for rewriting _field = "val" expressions to projections
+use crate::Predicate;
```
alamb, line 1: The logic in this file is largely rewritten -- I tried to add many docstrings (and ASCII art) 

Reading the diff on github might be hard -- pulling this down and just reading it as a new module might be easier for review

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
-                    if !matches!(child_state, NoPred) {
-                        Err(unsupported_op_error())
-                    } else {
-                        Ok(())
-                    }
-                })?;
-                Ok(NoPred)
-            }
-        }
+        // Form an array of strings from the field *names*:
```
alamb, line 111: This is the key of the change. Basically to use the DataFusion expression evaluator to evaluate some arbitrary set of predicates :bowtie:

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
-        // Actually rewrite the expression
-        let expr = expr.rewrite(&mut FieldProjectionRewriterWrapper(self))?;
-
-        // Ensure that there are no remaining _field references
-        expr.accept(FieldRefFinder {})?;
-
-        Ok(expr)
+        // for predicates like `A AND B AND C`
+        // rewrite `A`, `B` and `C` separately and put them back together
+        let rewritten_expr = disassemble_conjuct(expr)
```
alamb, line 63: I am quite pleased with this -- it breaks apart a conjunction (`A AND B AND C`) applies a transformation to each, and then reassembles it with AND

predicate/src/rpc_predicate/field_rewrite.rs
alamb: (this is more comments than code)

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
+            });
+
+        predicate.with_field_columns(new_fields)
     }
 }
 
-/// searches for unsupported patterns of predicates that we can't
-/// translate into an include or exclude list
-#[derive(Debug, Clone)]
-struct FieldRefChecker {
```
alamb, line 259: this structure was getting quite complicated to check for unsupported structures. Now we support many more structures so the check gets much simpler 😆

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -429,116 +358,105 @@ mod tests {
...
-            assert_eq!(rewritten, exp_expr);
-
-            let field_projection = rewriter.into_projection();
-            assert_eq!(field_projection, exp_projection);
-        }
-    }
-
-    #[test]
-    fn test_field_column_rewriter_unsupported() {
-        let cases = vec![
```
alamb, line 454: most of these cases are now supported, so I moved the relevant tests up into `test_field_column_rewriter`

predicate/src/rpc_predicate.rs
```diff
@@ -176,7 +176,13 @@ fn normalize_predicate(
     predicate: &Predicate,
 ) -> DataFusionResult<Predicate> {
     let mut predicate = predicate.clone();
-    let mut field_projections = FieldProjectionRewriter::new();
+
+    // can only rewrite `_field` expressions if we have a schema and
+    // know what the fields are
```
tustvold, line 181: Under what circumstances will we not have a schema? Is this not just an error?

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
+        // └──────┴─────────┘
+        let new_fields = as_boolean_array(&matching)
+            .iter()
+            .zip(as_string_array(&field_names).iter())
+            .filter_map(|(matching, field_name)| {
+                if matching == Some(true) {
+                    // this array was constructed with no nulls, so
+                    // the field_name should not be null
+                    assert!(field_name.is_some());
+                    field_name
```
tustvold, line 203:
```suggestion
                    Some(field_name.unwrap())
```

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
+        // Ceremony to prepare to evaluate the predicates
+        let input_schema = batch.schema();
+        let input_df_schema: DFSchema = input_schema.as_ref().clone().try_into().unwrap();
+        let props = ExecutionProps::default();
+        let exprs = self.field_predicates.into_iter().map(|expr| {
+            create_physical_expr(&expr, &input_df_schema, &input_schema, &props)
+                .expect("Error creating physical expression")
+        });
+
+        // evaluate into a boolean array where each element is true if
```
tustvold, line 140: The comments here are :ok_hand:

predicate/src/rpc_predicate/field_rewrite.rs
```diff
@@ -1,402 +1,333 @@
...
+        let batch = RecordBatch::try_from_iter(vec![(FIELD_COLUMN_NAME, Arc::clone(&field_names))])
+            .expect("Error creating _field record batch");
+
+        // Ceremony to prepare to evaluate the predicates
+        let input_schema = batch.schema();
+        let input_df_schema: DFSchema = input_schema.as_ref().clone().try_into().unwrap();
+        let props = ExecutionProps::default();
+        let exprs = self.field_predicates.into_iter().map(|expr| {
+            create_physical_expr(&expr, &input_df_schema, &input_schema, &props)
+                .expect("Error creating physical expression")
```
tustvold, line 137: Should this error be propagated - if I write a predicate like `_field = 23` will this panic?

tustvold (approved):
Really nice, just some minor nits :+1:

predicate/src/rpc_predicate.rs
alamb: I think you are right -- I will attempt to make the call to schema() infallable in a follow on PR

predicate/src/rpc_predicate/field_rewrite.rs
alamb: I think you are right -- will update to propagate them

alamb: In 08bca0a285546cc32b46ec85eefeee9773dd959e

predicate/src/rpc_predicate.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/5315 (WIP)

PR #1078, opened 2021-03-30
feat: make it easy to get OperationMetadata from Operation

NGA-TRAN:
Closes #1001

@alamb This PR address all your comments including one function list_operation. I will abort the other PR because it was messed up with some unsuccessful rebase stuff 

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #5303, opened 2022-08-03
feat: Compact cold partitions; upgrade a single non-overlapping level 0 file to level 1 without running compaction

carols10cents:
Fixes [#1078](https://github.com/influxdata/conductor/issues/1078).

This PR does somewhat more than the issue describes, but Nga and Paul and I have discussed it and this is ok.

This PR moves the current compaction procedure to `compact_hot_partitions`, and adds a new, slightly different procedure in `compact_cold_partitions` which also runs in the compaction handler loop.

There's a new CLI flag/env var, `--compaction-hot-multiple` (thoughts on a better name?), that controls how many times a compactor loop iteration will run compaction of hot partitions for every 1 time the loop runs compaction of cold partitions, to control how much we bias the compactor towards spending time on hot partitions. I've set the default to 4, so every compaction loop iteration will compact hot partitions 4 times and compact cold partitions once (is this default ok? a good starting place, at least?)

The main purpose of the cold compaction loop as I was thinking about it is to ensure partitions that got starved out of compaction while they were hot (because they weren't as hot as other partitions) eventually get compacted.

The cold compaction logic:

- Looks for partitions that contain any number of level 0 files with the maximum creation time of the level 0 files being more than 24 hours ago (right now that 24 hours is hardcoded but this could be another setting)
- Orders by the number of level 0 files descending and takes `max_num_partitions_per_sequencer` partitions per sequencer (this is something we could make into a separate parameter for different tuning someday after experimentation) 
- Selects all level 1 files in the partition that overlap with the level 0 files
- **Doesn't** limit the number of files or the total file size as the hot compaction does, include all level 0 and overlapping level 1 files for this partition
- If there's only one level 0 file (and no overlapping level 1 files), upgrade that file to level 1 without running compaction (this is [#1078](https://github.com/influxdata/conductor/issues/1078))
- Otherwise, run compaction on all the selected level 0 and level 1 files calling the same `compact_parquet_files` function that hot compaction uses, which splits the files based on `max_desired_file_size_bytes`, `percentage_max_file_size`, and `split_percentage`. We could change this in the future if we wanted to, say, always compact into one file no matter how big it is.

### DIFF
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -69,13 +69,14 @@ pub struct CompactorConfig {
     )]
     pub split_percentage: u16,
 
-    /// The compactor will limit the number of simultaneous compaction jobs based on the
-    /// size of the input files to be compacted. This number should be less than 1/10th
-    /// of the available memory to ensure compactions have
-    /// enough space to run.
-    /// Default is 1,073,741,824 bytes (1GB).
-    /// The number of compact_partititons run in parallel is determined by:
-    ///    max_concurrent_size_bytes/input_size_threshold_bytes
+    /// The compactor will limit the number of simultaneous hot partition compaction jobs based on
+    /// the size of the input files to be compacted. This number should be less than 1/10th of the
+    /// available memory to ensure compactions have enough space to run.
+    ///
+    /// Default is 1024 * 1024 * 1024 = 1,073,741,824 bytes (1GB).
+    //
+    // The number of compact_hot_partititons run in parallel is determined by:
+    //    max_concurrent_size_bytes/input_size_threshold_bytes
     #[clap(
         long = "--compaction-concurrent-size-bytes",
         env = "INFLUXDB_IOX_COMPACTION_CONCURRENT_SIZE_BYTES",
@@ -83,6 +84,22 @@ pub struct CompactorConfig {
         action
     )]
     pub max_concurrent_size_bytes: u64,
+
+    /// The compactor will limit the number of simultaneous cold partition compaction jobs based on
+    /// the size of the input files to be compacted. This number should be less than 1/10th of the
+    /// available memory to ensure compactions have enough space to run.
+    ///
+    /// Default is 1024 * 1024 * 900 = 943,718,400 bytes (900MB).
+    //
+    // The number of compact_cold_partititons run in parallel is determined by:
+    //    max_cold_concurrent_size_bytes/cold_input_size_threshold_bytes
+    #[clap(
+        long = "--compaction-cold-concurrent-size-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_COLD_CONCURRENT_SIZE_BYTES",
+        default_value = "943718400",
+        action
+    )]
+    pub max_cold_concurrent_size_bytes: u64,
 
     /// Max number of partitions per sequencer we want to compact per cycle
     /// Default: 1
@@ -104,14 +121,14 @@ pub struct CompactorConfig {
     )]
     pub min_number_recent_ingested_files_per_partition: usize,
 
-    /// A compaction operation will gather as many L0 files with their overlapping L1 files to
-    /// compact together until the total size of input files crosses this threshold. Later
-    /// compactions will pick up the remaining L0 files.
+    /// A compaction operation for hot partitions will gather as many L0 files with their
+    /// overlapping L1 files to compact together until the total size of input files crosses this
+    /// threshold. Later compactions will pick up the remaining L0 files.
     ///
     /// A compaction operation will be limited by this or by the file count threshold, whichever is
     /// hit first.
     ///
-    /// Default is 1024 * 1024 * 100 = 100,048,576 (100MB).
+    /// Default is 1024 * 1024 * 100 = 100,048,576 bytes (100MB).
     #[clap(
         long = "--compaction-input-size-threshold-bytes",
         env = "INFLUXDB_IOX_COMPACTION_INPUT_SIZE_THRESHOLD_BYTES",
@@ -119,6 +136,19 @@ pub struct CompactorConfig {
         action
     )]
     pub input_size_threshold_bytes: u64,
+
+    /// A compaction operation for cold partitions will gather as many L0 files with their
+    /// overlapping L1 files to compact together until the total size of input files crosses this
+    /// threshold. Later compactions will pick up the remaining L0 files.
+    ///
+    /// Default is 1024 * 1024 * 600 = 629,145,600 bytes (600MB).
+    #[clap(
+        long = "--compaction-cold-input-size-threshold-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_COLD_INPUT_SIZE_THRESHOLD_BYTES",
+        default_value = "629145600",
+        action
+    )]
+    pub cold_input_size_threshold_bytes: u64,
 
     /// A compaction operation will gather as many L0 files with their overlapping L1 files to
     /// compact together until the total number of L0 + L1 files crosses this threshold. Later
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -135,4 +135,17 @@ pub struct CompactorConfig {
         action
     )]
     pub input_file_count_threshold: usize,
+
+    /// The multiple of times that compacting hot partitions should run for every one time that
+    /// compacting cold partitions runs. Set to 1 to compact hot partitions and cold partitions
+    /// equally.
+    ///
+    /// Default is 4.
+    #[clap(
+        long = "--compaction-hot-multiple",
+        env = "INFLUXDB_IOX_COMPACTION_HOT_MULTIPLE",
+        default_value = "4",
+        action
+    )]
+    pub hot_multiple: usize,
 }
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -1047,28 +1047,5 @@ mod tests {
         assert_eq!(candidates[1].sequencer_id, sequencer.id);
         assert_eq!(candidates[2].partition_id, another_partition.id);
         assert_eq!(candidates[2].sequencer_id, another_sequencer.id);
-
-        // Add info to partition
-        let partitions_with_info = compactor.add_info_to_partitions(&candidates).await.unwrap();
-        assert_eq!(partitions_with_info.len(), 2);
-
-        assert_eq!(*partitions_with_info[0].namespace, namespace);
-        assert_eq!(*partitions_with_info[0].table, table);
-        assert_eq!(
-            partitions_with_info[0].partition_key,
-            partition4.partition_key
-        );
-        assert_eq!(partitions_with_info[0].sort_key, partition4.sort_key()); // this sort key is None
-
-        assert_eq!(*partitions_with_info[1].namespace, namespace);
-        assert_eq!(*partitions_with_info[1].table, another_table);
-        assert_eq!(
-            partitions_with_info[1].partition_key,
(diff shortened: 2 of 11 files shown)


alamb:
I have not been keeping the context of all the compactor details in my head (it is already close to 🤯 ) so I will defer review to @NGA-TRAN  and @pauldix . If they can not review this PR in a timely manner I am happy to do my best but I think the level of review I could offer right now is fairly limited

compactor/src/handler.rs
```diff
@@ -248,119 +255,236 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
...
+        .checked_duration_since(start_time)
+    {
+        let duration = compactor
+            .partitions_extra_info_reading_duration
+            .recorder(cold_attributes.clone());
+        duration.record(delta);
+    }
+
+    let n_candidates = candidates.len();
+    if n_candidates == 0 {
```
NGA-TRAN, line 422: I think we should remove this if condition here because if the compactor work well as we expect, there will be time no cold partitions for us to compact but there will be hot one. We just need to exit here to continue compacting hot partitions

compactor/src/handler.rs
```diff
@@ -248,119 +255,236 @@ async fn run_compactor(compactor: Arc<Compactor>, shutdown: CancellationToken) {
...
+    // Done compacting all candidates in the cycle, record its time
+    if let Some(delta) = compactor
+        .time_provider
+        .now()
+        .checked_duration_since(start_time)
+    {
+        let duration = compactor
+            .compaction_cycle_duration
+            .recorder(cold_attributes);
+        duration.record(delta);
```
NGA-TRAN, line 487: This is minor. I think when hot and cold work well, we may want to refactor `compact_hot_partitions` and `compact_cold_partitions` because they look so similar

compactor/src/lib.rs
```diff
@@ -76,9 +82,78 @@ pub(crate) async fn compact_partition(
...
+        ("partition_type", "hot".into()),
+    ]);
+    if let Some(delta) = compactor
+        .time_provider
+        .now()
+        .checked_duration_since(start_time)
+    {
+        let duration = compactor.compaction_duration.recorder(attributes);
+        duration.record(delta);
+    }
```
NGA-TRAN, line 98: 👍

compactor/src/parquet_file_filtering.rs
```diff
@@ -129,6 +129,103 @@ pub(crate) fn filter_parquet_files(
...
+    } = parquet_files_for_compaction;
+
+    if level_0.is_empty() {
+        info!("No level 0 files to consider for compaction");
+        return Vec::new();
+    }
+
+    // Guaranteed to exist because of the empty check and early return above. Also assuming all
+    // files are for the same partition.
+    let partition_id = level_0[0].partition_id;
```
NGA-TRAN, line 163: This makes me think of a future ticket that compact all level-1 files (and L0 if available) into one (or fewer larger file). I will open it

compactor/src/parquet_file_filtering.rs
```diff
@@ -129,6 +129,103 @@ pub(crate) fn filter_parquet_files(
...
+    let num_level_0_compacting = level_0.len();
+    let num_level_1_compacting = files_to_return.len();
+
+    info!(
+        partition_id = partition_id.get(),
+        num_level_0_considering,
+        num_level_1_considering,
+        num_level_0_compacting,
+        num_level_1_compacting,
+        "filtered Parquet files for compaction",
```
NGA-TRAN, line 207:
```suggestion
        "filtered cold Parquet files for compaction",
```

I think the `cold` will help. Otherwise, we will ask why compacting too many files

compactor/src/parquet_file_filtering.rs
```diff
@@ -129,6 +129,103 @@ pub(crate) fn filter_parquet_files(
...
+    // This will start by holding the level 1 files that are found to overlap an included level 0
+    // file. At the end of this function, the level 0 files are added to the end so they are sorted
+    // last.
+    let mut files_to_return = Vec::with_capacity(level_0.len() + remaining_level_1.len());
+    // Running total of the size, in bytes, of level 0 files and level 1 files for inclusion.
+    // The levels are counted up separately for metrics.
+    let mut total_level_0_bytes = 0;
+    let mut total_level_1_bytes = 0;
+
+    for level_0_file in &level_0 {
```
NGA-TRAN, line 177: This mean we will compact all L0 files and their overlapped L1 no matter how large they are? I think we still need to limit here because I remember there is a large partition of 20GB and of they are all overlapped (or we are behind and do not compact L0 files at all), we will hit OOM

NGA-TRAN (commented):
Wow, what a PR! I have reviewed the code carefully and many tests (not all especially the cold ones copied from the hot) carefully.  They look great.

I think if you add a size limit for the cold case to avoid hitting OOM and remove the sleep 1 second for the cold case, it will be great

NGA-TRAN:
> cold compactors benefit from more memory

I think this one sync well with memory and concurrency limit for cold partition. I think we can address this in your current PR by adding 2 more config param

- INFLUXDB_IOX_COMPACTION_COLD_INPUT_SIZE_THRESHOLD_BYTE (we will set it larger than the one for hot partitions in IDPE)
- INFLUXDB_IOX_COMPACTION_COLD_CONCURRENT_SIZE_BYTES (we will set it smaller than the one for the hot one in IDPE)

compactor/src/handler.rs
carols10cents: Yeah, I wasn't sure ultimately how similar or different they would be, but I agree 👍🏻

compactor/src/handler.rs
carols10cents: > I think we should remove this if condition here 

> We just need to exit here to continue compacting hot partitions

> and remove the sleep 1 second for the cold case

It sounds like the `if` condition and the `return` should stay, just the `sleep` should go. Please correct me if I've misunderstood!

NGA-TRAN: I think the debug message would be useful. Yes, the if stays, just remove the sleep then

compactor/src/compact.rs
```diff
@@ -999,6 +999,18 @@ mod tests {
...
+        let candidates = compactor.cold_partitions_to_compact(2).await.unwrap();
+        assert_eq!(candidates.len(), 2);
+        assert_eq!(candidates[0].partition_id, partition4.id);
+        assert_eq!(candidates[1].partition_id, partition2.id);
+
+        // Ask for 3 partitions per sequencer; still get only partition4 and partition2
+        let candidates = compactor.cold_partitions_to_compact(3).await.unwrap();
+        assert_eq!(candidates.len(), 2);
+        assert_eq!(candidates[0].partition_id, partition4.id);
+        assert_eq!(candidates[1].partition_id, partition2.id);
```
NGA-TRAN, line 1016: Nice

compactor/src/compact.rs
```diff
@@ -1024,6 +1036,18 @@ mod tests {
...
+        // another_sequencer
+        let mut candidates = compactor.cold_partitions_to_compact(2).await.unwrap();
+        candidates.sort();
+        assert_eq!(candidates.len(), 3);
+        assert_eq!(candidates[0].partition_id, partition2.id);
+        assert_eq!(candidates[0].sequencer_id, sequencer.id);
+        assert_eq!(candidates[1].partition_id, partition4.id);
+        assert_eq!(candidates[1].sequencer_id, sequencer.id);
+        assert_eq!(candidates[2].partition_id, another_partition.id);
+        assert_eq!(candidates[2].sequencer_id, another_sequencer.id);
```
NGA-TRAN, line 1053: 🎉

compactor/src/compact.rs
```diff
@@ -1047,28 +1047,5 @@ mod tests {
...
-        assert_eq!(*partitions_with_info[1].namespace, namespace);
-        assert_eq!(*partitions_with_info[1].table, another_table);
-        assert_eq!(
-            partitions_with_info[1].partition_key,
-            another_partition.partition_key
-        );
-        assert_eq!(
-            partitions_with_info[1].sort_key,
-            another_partition.sort_key()
-        ); // this sort key is Some(tag1, time)
```
NGA-TRAN: 👍

compactor/src/handler.rs
```diff
@@ -421,8 +421,6 @@ async fn compact_cold_partitions(compactor: Arc<Compactor>) {
     let n_candidates = candidates.len();
     if n_candidates == 0 {
         debug!("no cold compaction candidates found");
-        // sleep for a second to avoid a hot busy loop when the catalog is polled
-        tokio::time::sleep(PAUSE_BETWEEN_NO_WORK).await;
```
NGA-TRAN: 👍

clap_blocks/src/compactor.rs
```diff
@@ -120,6 +137,19 @@ pub struct CompactorConfig {
...
+    /// threshold. Later compactions will pick up the remaining L0 files.
+    ///
+    /// Default is 1024 * 1024 * 600 = 629,145,600 bytes (600MB).
+    #[clap(
+        long = "--compaction-cold-input-size-threshold-bytes",
+        env = "INFLUXDB_IOX_COMPACTION_COLD_INPUT_SIZE_THRESHOLD_BYTES",
+        default_value = "629145600",
+        action
+    )]
+    pub cold_input_size_threshold_bytes: u64,
```
NGA-TRAN, line 151: 👍

compactor/src/handler.rs
```diff
@@ -428,16 +448,17 @@ async fn compact_cold_partitions(compactor: Arc<Compactor>) {
...
-    // --> num_parallel_partitions = max_concurrent_compaction_size_bytes/
-    //     input_size_threshold_bytes
-    let num_parallel_partitions = (compactor.config.max_concurrent_compaction_size_bytes
-        / compactor.config.input_size_threshold_bytes) as usize;
+    //   . Each `compact partititon` takes max of this much memory cold_input_size_threshold_bytes
+    //   . We have this memory budget: max_cold_concurrent_size_bytes
+    // --> num_parallel_partitions = max_cold_concurrent_size_bytes/
+    //     cold_input_size_threshold_bytes
+    let num_parallel_partitions = (compactor.config.max_cold_concurrent_size_bytes
+        / compactor.config.cold_input_size_threshold_bytes)
```
NGA-TRAN, line 460: 👍

compactor/src/parquet_file_filtering.rs
```diff
@@ -193,9 +200,18 @@ pub(crate) fn filter_cold_parquet_files(
         // The remaining level 1 files to possibly include in future iterations are the remaining
         // ones that did not overlap with this level 0 file.
         remaining_level_1 = non_overlaps;
+
+        // Move the level 0 file into the list of level 0 files to return
+        level_0_to_return.push(level_0_file);
+
+        // Stop considering level 0 files if the total size of all files is over or equal to
+        // `max_bytes`
+        if (total_level_0_bytes + total_level_1_bytes) >= max_bytes {
```
NGA-TRAN, line 209: 👍

compactor/src/lib.rs
```diff
@@ -118,6 +118,7 @@ pub(crate) async fn compact_cold_partition(
 
     let to_compact = parquet_file_filtering::filter_cold_parquet_files(
         parquet_files_for_compaction,
+        compactor.config.cold_input_size_threshold_bytes(),
```
NGA-TRAN, line 121: 👍

NGA-TRAN (approved):
Nice. Thanks so much @carols10cents. Can't wait to see this run

PR #5304, opened 2022-08-03
WIP feat: consider cold partitions if the compactor does not have hot partitions to compact in a cycle

### LINKED ISSUE #5302: Compact cold partitions if the Compactor has capacity to compact them
# Rationale

The current setup is to select **n** partitions per sequencer to compact per cycle. However, if we find some active (today or latest) **m** partitions to compact, we do not get non-active (not today) partitions even if m < n. This will lead to cold partitions will never get compacted for that sequencer if there are active partitions.

The ticket is to also select cold partitions to compact. More specific, if m < n, select n-m most-L0-files cold partition to compact with the other n to partitions in the same cycle

NGA-TRAN:
Closes #5302


NGA-TRAN:
I have chatted with @carols10cents and this PR has some design conflict with https://github.com/influxdata/influxdb_iox/pull/5303. Waiting for that PR to get done and we chat with Paul about when to compact cold partitions before coming back to this PR

Issue #5117, opened 2022-07-13
Ingester needs metric for kafka batch size


alamb:
As part of the disucssions on https://github.com/influxdata/conductor/issues/971 there is a lot of theorizing about how the kafka message size distribution could be affecting the ingester performance

It would be great to add a Histogram metric in the ingester that showed the distribution of kafka message sizes read by the ingester to get more information on this topic

crepererum:
Uh, I'm not entirely sure how to implement this in a nice API. The messages themselves are NOT exposed from rskakfa to API users because the API is kinda higher level. I also would like to avoid using the IOx metrics framework within rskafka because it's IOx specific and nobody else out there uses it. So I see the following ways to make this possible:

- **opentelemetry:** Use the opentelemetry SDK in rskafka (optional dependency) and hook that into the IOx metrics exporter (or does the hook already exists?)
- **return information:** return the stats from rskafka API calls. That's a bit messy because how would this map to the stream consumer that consumes records within a single message?
- **stats callback:** that's what rdkafka is doing. They have some kind of callback that you can pass to the client. We probably need multiple callbacks for multiple stats though (since they are produced at different rates) but overall I think this could work quite well and the API user can do with it whatever they like.

@tustvold any thoughts?

alamb:
I guess I was imagining the of the payloads (rather than the actual messages read from kafka). Aka the same value used to report `write_buffer_bytes_read`: in https://github.com/influxdata/influxdb_iox/blob/main/ingester/src/stream_handler/sink_instrumentation.rs#L163

domodwyer:
The current average payload size across all sequencers is ~2103 bytes, with some variation across sequencers:

![Screenshot 2022-08-04 at 15 16 07](https://user-images.githubusercontent.com/9275968/182856448-e1121a7d-4941-45e5-83e5-bd2aa3281ff0.png)

This is the distribution of average payload sizes over sequencers:

![image](https://user-images.githubusercontent.com/9275968/182859422-ae5bda19-f704-4589-952f-48ac28981e35.png)

Clearly a large majority of average message sizes are <3KiB

<details><summary>
Raw sequencer payload sizes
</summary>

| kafka_partition|avg_payload_size |
|---------------:|:----------------|
|               0|988              |
|               1|2.97KiB          |
|              10|7.73KiB          |
|              11|1.72KiB          |
|              12|2.56KiB          |
|              13|2.86KiB          |
|              14|1.62KiB          |
[...]

PR #5307, opened 2022-08-04
feat: instrument kafka aggregated DML batch size

domodwyer:
This will help us understand the effectiveness of the DML write aggregation logic (relates to #5117).

---

* feat: instrument kafka aggregated DML batch size (5c18f13ec)

      The Kafka write buffer implementation (and only the Kafka impl) merges 
      together successive DML writes for the same namespace & partition within a
      window of time.

      This commit records the number of DML writes that have been merged together to
      form a single batched op before it is dispatched to Kafka.


write_buffer/src/kafka/mod.rs
```diff
@@ -676,6 +686,39 @@ mod tests {
...
+            (1, 0),
+            (2, 3),
+            (4, 0),
+            (8, 0),
+            (10, 0),
+            (16, 0),
+            (32, 0),
+            (64, 0),
+            (128, 0),
+        ]
```
alamb, line 713: I may be missing something obvious here, but wouldn't we expect that there was at least one bucket that had a non zero count? This appears to imply all the buckets have a count of 0

write_buffer/src/kafka/aggregator.rs
```diff
@@ -299,14 +322,30 @@ impl DmlAggregator {
...
+        // being pushed to kafka.
+        //
+        // Register here and pass the recorder into each WriteAggregator to
+        // avoid locking the metric repository each time a WriteAggregator is
+        // created.
+        let batch_coalesced_ops: U64Histogram = metric_registry
+            .register_metric_with_options::<U64Histogram, _>(
+                "write_buffer_batch_coalesced_write_ops",
+                "number of dml writes aggregated into a single kafka produce operation",
+                || U64HistogramOptions::new([1, 2, 4, 8, 10, 16, 32, 64, 128]),
```
alamb, line 337: it doesn't really matter, but I found the presence of `10` in this list of numbers that was otherwise powers of two somewhat confusing

https://www.youtube.com/watch?v=rsRjQDrDnY8

write_buffer/src/kafka/mod.rs
alamb: Though then I looked at the code and it looks like `batch_coalesced_ops` is never written to ❓

alamb (approved):

write_buffer/src/kafka/mod.rs
domodwyer:
```
            (2, 3), 
```

This asserts that there are 3 batches of size 2.

write_buffer/src/kafka/aggregator.rs
domodwyer: That is random! It was supposed to be powers of 2 - one sec :+1:

write_buffer/src/kafka/mod.rs
alamb: Ah, missed that somehow 🤦

domodwyer: It happens - like my 10! :joy:

PR #5308, opened 2022-08-04
refactor: remove `split_members` and use `split_conjunction` from upstream DataFusion

alamb:
re https://github.com/influxdata/influxdb_iox/issues/5177

I wanted to reuse this function as part of https://github.com/influxdata/influxdb_iox/pull/5301 but when I tried to refactor it out into `datfusion_utils` I found datafusion already had the same function!


predicate/src/lib.rs
```diff
@@ -510,7 +511,7 @@ impl Predicate {
                 expr_to_columns(&expr, &mut columns)?;
 
                 if columns.len() == 1 && Self::primitive_binary_expr(&expr) {
-                    pushdown_exprs.push(expr);
+                    pushdown_exprs.push(expr.clone());
```
alamb, line 514: this `clone` was previously inside `split_members` as `other.clone()` -- so the same amount of cloning is done as before

NGA-TRAN (approved):
Less code name functionality is always nice. 👍

PR #5309, opened 2022-08-04
fix: `compute_split_time` with small step sizes

crepererum:
Found that while debugging https://github.com/influxdata/conductor/issues/1103 .


compactor/src/utils.rs
```diff
@@ -261,4 +261,18 @@ mod tests {
         assert_eq!(result.len(), 1);
         assert_eq!(result[0], 1);
     }
+
+    #[test]
+    fn compute_split_time_please_dont_explode() {
```
alamb, line 266: I assume without this fix the loop goes forever?

crepererum: Yes, and while doing so fills up the output vector, so depending on your system this may either crash your system or the process dies with SIGKILL (OOM killer).

compactor/src/utils.rs
```diff
@@ -261,4 +261,18 @@ mod tests {
...
+        // degenerated case where the step size is so small that it is < 1 (but > 0). In this case we shall still
+        // not loop forever.
+        let min_time = 10;
+        let max_time = 20;
```
NGA-TRAN, line 270: This is a very interesting case. Nice catch

NGA-TRAN (approved):

PR #5312, opened 2022-08-04
fix: buffer allocation while reading parquet files

crepererum:
Work around https://github.com/apache/arrow-rs/issues/2321 by limiting
reader batch size to number of rows (based on file-level metadata).

Fixes https://github.com/influxdata/conductor/issues/1103 .


tustvold (approved):

NGA-TRAN (commented):
Nice. Thanks @crepererum

PR #5315, opened 2022-08-04
fix: make it clear rpc_predicates are only ever specialized when a schema is known

alamb:
# Rationale
As pointed out by @tustvold on https://github.com/influxdata/influxdb_iox/pull/5301#discussion_r938004439, it makes no sense to be trying to specialize a rpc predicate (which requires schema information) when there is no schema

# Changes
Update the code to make it clear this is an InternalError if this happens

It also makes the code quite a bit cleaner 🧹


alamb:
Cargo audit failure is unrelated to this one

predicate/src/rpc_predicate.rs
```diff
@@ -113,8 +113,17 @@ impl InfluxRpcPredicate {
...
                 let schema = table_info.table_schema(&table);
-                let predicate = normalize_predicate(&table, schema, &self.inner)?;
-
+                let predicate = match schema {
+                    Some(schema) => normalize_predicate(&table, schema, &self.inner)?,
+                    None => {
+                        // if we don't know about this table, we can't
+                        // do any predicate specialization. This can
+                        // happen if there is a request for
+                        // "measurement fields" for a non exitent
```
philjb, line 122: typo: missing `s`

philjb (approved):
take my approval with 🧂

predicate/src/rpc_predicate.rs
alamb:
```suggestion
                        // "measurement fields" for a non existent
```

alamb:
> take my approval with 🧂

I appreciate the review @philjb  ❤️

PR #5319, opened 2022-08-05
fix: Improve chunk pruning stats and extend with was_not_pruned

crepererum:
- emit a warning if we cannot even attempt to prune chunks due to an
  error. This is always either a missing feature or a bug (even though
  it does not impact correctness but _only_ performance). Also see
  https://github.com/influxdata/conductor/issues/1107
- change metrics to clearly differentiate between "could not prune" and
  "not pruned"
- add new "not pruned" observer hook (this was missing for some reason,
  the "pruned" hook existed though)


iox_query/src/pruning.rs
```diff
@@ -52,10 +52,19 @@ impl std::fmt::Display for NotPrunedReason {
...
+    /// Called when the specified chunk was pruned from observation.
     fn was_pruned(&self, _chunk: &dyn QueryChunk) {}
 
-    /// Called when no pruning can happen at all for some reason
+    /// Called when a chunk was not pruned.
+    fn was_not_pruned(&self, _chunk: &dyn QueryChunk) {}
+
+    /// Called when no pruning can happen at all for some reason.
+    ///
+    /// Since pruning is optional and _only_ improves performance but its lack does not affect correctness, this will
```
alamb, line 63: 👍

iox_query/src/pruning.rs
```diff
@@ -114,7 +123,7 @@ where
             for chunk in &chunks {
                 observer.could_not_prune(NotPrunedReason::DataFusionPruningFailed, chunk.as_ref());
             }
-            trace!(%e, ?filter_expr, "DataFusion pruning failed");
+            warn!(%e, ?filter_expr, "DataFusion pruning failed");
```
alamb, line 126: I think DataFusion pruning failure is likely to happen a non trivial amount so this log many be quite noisy -- on the other hand I see no reason we can't try it, and if it generates lots of log errors address them (like fix the underlying issue or reduce the severity of the message)

alamb (approved):
👨‍🍳 👌 

very nice

PR #5322, opened 2022-08-05
fix: do not loose data when Kafka reports that offset is above watermark

### LINKED ISSUE #5128: Ingester drops more writes than needed when Kafka offset is missing

crepererum:
*This can happen in certain cluster rebalance settings.

This is also linked to https://github.com/influxdata/rskafka/issues/147
but for the upstream issue I currently have no idea how to fix it, so
let's at least harden IOx against it.

Fixes #5128.


crepererum:
Need to fix the tests.

alamb:
@domodwyer  is this something you have time to review?

domodwyer:
Without context it is hard to understand what has changed / what the fix is - maybe it is something I missed while I was off and someone else has better understanding?

crepererum:
@domodwyer the problem is described in https://github.com/influxdata/influxdb_iox/issues/5128#issue-1304671998

write_buffer/src/core.rs
```diff
@@ -38,8 +38,16 @@ impl WriteBufferError {
...
+    pub fn sequence_number_after_watermark(
+        e: impl Into<Box<dyn std::error::Error + Sync + Send>>,
+    ) -> Self {
+        Self::new(WriteBufferErrorKind::SequenceNumberAfterWatermark, e)
+    }
+
+    pub fn sequence_number_no_longer_exists(
+        e: impl Into<Box<dyn std::error::Error + Sync + Send>>,
+    ) -> Self {
+        Self::new(WriteBufferErrorKind::SequenceNumberNoLongerExists, e)
```
domodwyer, line 50: So the fix is to split the two cases of "unknown watermark" to distinguish between them in the consumer? 

What should the ingester do if it observes `SequenceNumberAfterWatermark`? I'd imagine panic, but currently it just busy-waits in a loop I think.

ingester/src/stream_handler/handler.rs
```diff
@@ -252,7 +252,7 @@ where
 
                     Some(op)
                 }
-                Some(Err(e)) if e.kind() == WriteBufferErrorKind::UnknownSequenceNumber => {
+                Some(Err(e)) if e.kind() == WriteBufferErrorKind::SequenceNumberNoLongerExists => {
```
domodwyer, line 255: It would be good to handle `SequenceNumberAfterWatermark` too - perhaps by panic-ing to attract the humans.

write_buffer/src/core.rs
crepererum: > So the fix is to split the two cases of "unknown watermark" to distinguish between them in the consumer?

Yes


> What should the ingester do if it observes `SequenceNumberAfterWatermark`? I'd imagine panic, but currently it just busy-waits in a loop I think.

Good call, will fix that :muscle:

crepererum: done

ingester/src/stream_handler/handler.rs
crepererum: done

ingester/src/stream_handler/handler.rs
```diff
@@ -309,6 +309,16 @@ where
                     self.seq_invalid_data_count.inc(1);
                     None
                 }
+                Some(Err(e)) if e.kind() == WriteBufferErrorKind::SequenceNumberAfterWatermark => {
+                    panic!(
+                        "\
+Sequencer {:?} stream for topic {} has a high watermark BEFORE the sequence number we want. This is either a but (see \
```
domodwyer, line 315: Where's my butt emoji?!

> This is either a but

domodwyer (approved):

ingester/src/stream_handler/handler.rs
carols10cents: 🍑

crepererum: s/:peach:/:bug:/

PR #5329, opened 2022-08-05
feat: only sleep when there are neither hot nor cold partitions to compact

### LINKED ISSUE #5324: Move sleep outside compact_hot_partition
# Rationale

Currently, in each compaction cycle, we do 4 times hot compaction and 1 time cold compaction. In each hot compaction, we sleep 1 second if no candidates found. This will lead to sleep 4 seconds before considering to compact cold partitions. The sleep should move outside compact hot partition and only sleep if we have neither hot or cold partition to compact

NGA-TRAN:
Closes #5324 


carols10cents (approved):
LGTM!

PR #5317, opened 2022-08-04
fix: Install cargo deny in ci image

alamb:
# Rationale
While working on https://github.com/influxdata/influxdb_iox/pull/5316 (waiting for the test to complete, actually) I noticed that just __installing__ `cargo deny` takes 2 minutes. 

https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/22172/workflows/f701d1fd-4d66-44f8-af97-491e94b7bf2b/jobs/191926

Screen Shot 2022-08-04 at 4 25 43 PM

# Changes
Let's pay that 2 minutes once a day rather than once a PR by installing it in the image

PR #5331, opened 2022-08-05
chore: force ci build

alamb:
Just here to check that https://github.com/influxdata/influxdb_iox/pull/5317 didn't break anything


alamb:
And this time actually run CI with the change from https://github.com/influxdata/influxdb_iox/pull/5317 🤦

PR #5194, opened 2022-07-22
feat: Different branch to hook up new compaction algorithm

NGA-TRAN:
I messed up with semantic of #5188 and create this new branch for it

The first commit includes cherry picks of the first 3 commits from #5188 

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #5334, opened 2022-08-05
feat: Compactor e2e tests and a "run once" compactor command

carols10cents:
So before this PR, some of the tests were running a compactor, doing its own thing willy-nilly as we want it to do in production. That's pretty inconvenient in the tests, like [in this PR](https://github.com/influxdata/influxdb_iox/pull/5194), as we want to be able to say "ok run compaction now, and once it's done, then assert some things".

So, in the tests, rather than running the compactor as a server, we've created a command `influxdb_iox compactor run-once` that will run the compactor loop one time and then exit. It sets a default of `--compaction-hot-multiple` to 1 so that `run-once` will run one hot and one cold compaction operation. We can change other defaults of other settings if that sounds useful.

I could see this command being useful in development, if you want to run a compaction cycle with particular settings and know it's only going to go through the loop once and then you can examine what it did.

And I can see this command being useful in production-- say there's some sort of incident, a customer dumped in a whole bunch of data, a compactor gets behind, other customers are getting mad or something, and whoever's on call just wants to spin up a new instance and compact a targeted set *right now*-- this command will enable that! 

These other use cases may also inform the setting of defaults for other options. 

So. 

This PR reverted the change made to the e2e test in https://github.com/influxdata/influxdb_iox/pull/5194. That change was necessary because the mini cluster was always running the compactor (and the compactor went from not doing anything to compacting). This PR makes the mini cluster *never* run the compactor.

Then, we added the altered version of the test from that PR, and added in the step `Step::Compact`. This test passes!

I also made some improvements to the test logging along the way, like logging what command exactly the tests are running, both for the server commands and the new `compactor run-once` command.

### DIFF
--- a/clap_blocks/src/compactor.rs
+++ b/clap_blocks/src/compactor.rs
@@ -1,7 +1,11 @@
 //! CLI config for compactor-related commands
 
 macro_rules! gen_compactor_config {
-    ($name:ident $(,)?) => {
+    (
+        $name:ident,
+        hot_multiple_def = $hot_multiple_def:literal
+        $(,)?
+    ) => {
         /// CLI config for compactor
         #[derive(Debug, Clone, clap::Parser)]
         pub struct $name {
@@ -174,11 +178,12 @@ macro_rules! gen_compactor_config {
             /// compacting cold partitions runs. Set to 1 to compact hot partitions and cold partitions
             /// equally.
             ///
-            /// Default is 4.
+            /// Default is
+            #[doc = $hot_multiple_def]
             #[clap(
                 long = "--compaction-hot-multiple",
                 env = "INFLUXDB_IOX_COMPACTION_HOT_MULTIPLE",
-                default_value = "4",
+                default_value = $hot_multiple_def,
                 action
             )]
             pub hot_multiple: usize,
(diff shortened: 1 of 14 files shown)


clap_blocks/src/compactor.rs
```diff
@@ -1,181 +1,220 @@
...
-    #[clap(
-        long = "--compaction-hot-multiple",
-        env = "INFLUXDB_IOX_COMPACTION_HOT_MULTIPLE",
-        default_value = "4",
-        action
-    )]
-    pub hot_multiple: usize,
+#![cfg_attr(rustfmt, rustfmt_skip)] // https://github.com/rust-lang/rustfmt/issues/5489
+
+macro_rules! gen_compactor_config {
```
alamb, line 3: Some comments here might help, especially with the rationale for this macro

For example:

```suggestion
/// Create compactor configuration that can have different defaults
macro_rules! gen_compactor_config {

```

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -69,6 +69,151 @@ async fn remote_partition_and_get_from_store_and_pull() {
...
+                    //     "sequencerId": 1,
+                    //     "namespaceId": 1,
+                    //     "tableId": 1,
+                    //     "partitionId": "1",
+                    //     "objectStoreId": "fa6cdcd1-cbc2-4fb7-8b51-4773079124dd",
+                    //     "minTime": "123456",
+                    //     "maxTime": "123456",
+                    //     "fileSizeBytes": "2029",
+                    //     "rowCount": "1",
+                    //     "createdAt": "1650019674289347000"
```
alamb, line 86: Thank you for the comments -- this would be very hard to review without them 👍

test_helpers_end_to_end/src/lib.rs
```diff
@@ -51,6 +51,56 @@ pub fn rand_id() -> String {
         .collect()
 }
 
+/// Log the [`std::process::Command`] being run in a way that's convenient to copy-paste
+fn log_command(command: &std::process::Command) {
```
alamb, line 55: ❤️

alamb (approved):
Note for other reviewers -- I found whitespace blind diff easier to review: https://influxdata.slack.com/archives/CRK9M8L5Q/p1660051826528419

All in all, I love this -- great work @shepmaster  and @carols10cents 

> So, in the tests, rather than running the compactor as a server, we've created a command influxdb_iox compactor run-once that will run the compactor loop one time and then exit. It sets a default of --compaction-hot-multiple to 1 so that run-once will run one hot and one cold compaction operation. We can change other defaults of other settings if that sounds useful.

This sounds like a great tool for several reasons, not just testing. For example @tustvold  and I can use it to make a mini-benchmark for compaction performance 🎉

influxdb_iox/tests/end_to_end_cases/error.rs
```diff
@@ -28,13 +28,6 @@ pub async fn test_panic() {
                 let querier = state.cluster().querier();
                 assert_panic_logging(querier.querier_grpc_connection(), querier.log_path().await)
                     .await;
-
```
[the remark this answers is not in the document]
carols10cents: Because the minicluster is no longer starting up the compactor.

test_helpers_end_to_end/src/mini_cluster.rs
```diff
@@ -60,7 +65,7 @@ impl MiniCluster {
         router: Option<ServerFixture>,
         ingester: Option<ServerFixture>,
         querier: Option<ServerFixture>,
-        compactor: Option<ServerFixture>,
+        compactor_config: Option<TestConfig>,
```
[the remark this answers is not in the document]
carols10cents: There was only the one test that cared if the compactor ran or not-- the one I talked about in the description that was changed because it started failing when we made the compactor actually run 😅

carols10cents:
I believe I have resolved all suggestions, so I'm merging this!

alamb:
🎉

Issue #5323, opened 2022-08-05
Replace Kafka write aggregation logic


domodwyer:
# TL;DR

The aggregation logic in the write buffer is largely ineffective, costing us performance and likely contributing to the slow ingester recovery time (https://github.com/influxdata/conductor/issues/971) and high latency of kafka writes after rollouts (https://github.com/influxdata/conductor/issues/987). The complexity of this layer has also led hard-to-track-down bugs ([1][partition-bug], [2][flushing]).

We should replace or remove the aggregation logic.

[partition-bug]: https://github.com/influxdata/influxdb_iox/pull/4878
[flushing]: https://github.com/influxdata/influxdb_iox/pull/5113

## Analysis

The write buffer contains an [aggregation layer] that attempts to coalesce multiple DML writes together into one big DML write, before pushing the resulting batch of merged ops to kafka. 

The intent of this aggregation logic to increase the size of each DML write op sent over the wire, thereby increasing throughput (fewer messages, containing more rows) and to increase the effectiveness of compression (dictionary encoding in the `MutableBuffer`, block compression of payloads, etc).

Kafka/IOx is configured to allow a max message size of 10MiB however on average, the messages in Kafka have small payloads - the average uncompressed payload size is ~2103 bytes. We have data per kafka partition, and while there is some variation across partitions, almost all partitions (>75%) have an average message size of **just 2.3KiB**:

![image](https://user-images.githubusercontent.com/9275968/182859422-ae5bda19-f704-4589-952f-48ac28981e35.png)

<details><summary>
Quantiles of partition avg. payload size
</summary>

|     |        x|
|:----|--------:|
|0%   |   815.00|
|25%  |  1278.50|
|50%  |  1606.00|
|75%  |  2318.75|
|100% | 10845.00|
</details>

<details><summary>
Raw partition avg. payload size data
</summary>

| kafka_partition|avg_payload_size |
|---------------:|:----------------|
|               0|988              |
|               1|2.97KiB          |
|              10|7.73KiB          |
|              11|1.72KiB          |
|              12|2.56KiB          |
|              13|2.86KiB          |
|              14|1.62KiB          |
|              15|1.07KiB          |
|              16|1.24KiB          |
|              17|2.24KiB          |
|              18|1.99KiB          |
|              19|2.49KiB          |
|               2|2.04KiB          |
|              20|1.54KiB          |
|              21|1.22KiB          |
|              22|2.11KiB          |
|              23|1005             |
[...]

PR #5341, opened 2022-08-08
feat: instrument Kafka produce latency

domodwyer:
The Kafka write buffer implementation sends DML ops through a bunch of non-trivial [aggregation][agg] logic that (by design) affects the latency of the [high-level DML push call][high], which in turn obscures (and includes) the true latency of the low-level Kafka client.

The metrics in this PR record the true latency of the Kafka client's publish fn in isolation, allowing us to infer if latency spikes originate within the aggregation layer, or are observed by the Kafka client itself.

Relates to https://github.com/influxdata/influxdb_iox/issues/5323 and https://github.com/influxdata/conductor/issues/987

[agg]:https://github.com/influxdata/influxdb_iox/blob/932fecbaff9b113c3681e77fc69d52fe82524a07/write_buffer/src/kafka/aggregator.rs#L29
[high]:https://github.com/influxdata/influxdb_iox/blob/932fecbaff9b113c3681e77fc69d52fe82524a07/write_buffer/src/core.rs#L151-L155

---

* refactor: impl TimeProvider for Arc<TimeProvider> (323788767)

      This allows the MockProvider to be used in tests with consuming code that uses
      generics/static dispatch instead of a dyn TimeProvider, while still retaining
      a ref to the MockProvider instance.

* refactor: const KafkaPartition::new() (d003fe004)

      Allows this fn to be called from const contexts (useful in test setups).

* feat: Kafka client produce() instrumentation (284a3069c)

      Adds a decorator over the underlying kafka client to capture the latency 
      distribution of the low-level kafka writes, independent of the aggregation/DML
      batching framework that sits "above" this client.

      The latency measurements include the serialisation overhead, protocol 
      overhead, and actual network I/O.

* refactor: use kafka produce instrumentation (c133cf22c)

      This commit changes the IOx write buffer initialisation code to add the 
      KafkaProducerMetrics instrumentation to the per-partition Kafka clients.

* refactor(write_buffer): database_name -> topic_name (87e4290e1)

      Previously IOx mapped a single database to a single kafka topic - this is no
      longer the case, so referring to the kafka topic name as the
      "database name" name is confusing.


write_buffer/src/kafka/instrumentation.rs
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /// Call the inner [`ProducerClient`] implementation and record latency in
+    /// the appropriate success/error metric depending on the result.
+    async fn instrument(
+        &self,
+        records: Vec<Record>,
+        compression: Compression,
+    ) -> Result<Vec<i64>, rskafka::client::error::Error> {
+        let t = self.time_provider.now();
```
alamb, line 88: 👍

write_buffer/src/kafka/instrumentation.rs
```diff
@@ -0,0 +1,229 @@
...
+                ("kafka_partition", "42"),
+                ("result", "error"),
+            ]))
+            .expect("failed to get observer")
+            .fetch();
+
+        assert_eq!(histogram.sample_count(), 1);
+        assert_eq!(histogram.total, CALL_LATENCY);
+    }
+}
```
alamb, line 229: I wonder if it is worth testing more than one call (maybe by combing both the success and failure case together) to exercise the "more than one" codepath

I don't think it is critical, just something that I thought of while reviewing this PR

write_buffer/src/kafka/mod.rs
```diff
@@ -52,33 +54,37 @@ pub struct RSKafkaProducer {
 impl RSKafkaProducer {
     pub async fn new(
         conn: String,
-        database_name: String,
+        topic_name: String,
```
alamb, line 57: 👍 (drive by cleanup)

alamb (approved):
LGTM

write_buffer/src/kafka/instrumentation.rs
```diff
@@ -0,0 +1,229 @@
...
+
+        res
+    }
+}
+
+impl<P> rskafka::client::producer::ProducerClient for KafkaProducerMetrics<P>
+where
+    P: TimeProvider,
+{
+    fn produce(
```
crepererum, line 107: We should probably have used an `async_trait` in rskafka :facepalm:

write_buffer/src/kafka/instrumentation.rs
```diff
@@ -0,0 +1,229 @@
...
+/// recording the latency distribution and success/error result of the
+/// underlying [`ProducerClient::produce()`] call, which includes serialisation
+/// & protocol overhead, as well as the actual network I/O.
+///
+/// The metrics created by this instrumentation are labelled with the kafka
+/// topic & partition specified at initialisation.
+#[derive(Debug)]
+pub struct KafkaProducerMetrics<P = SystemProvider> {
+    inner: Box<dyn ProducerClient>,
+    time_provider: P,
```
crepererum, line 22: Do we really need a generic here? `time_provider: Arc<dyn TimeProvider>` should be sufficient. The call isn't that performance critical to be be hurt by a dyn dispatch. In contrast generics increase compile time and code complexity.

crepererum (approved):
Smaller nitpicks, good overall integration :+1:

write_buffer/src/kafka/instrumentation.rs
domodwyer: I'm not sure I agree with the general "generics is bad" position - if using this generic changes the compile time in any significant way, I'll definitely be with you! But dyn by default seems more like a personal choice than a benefit (a case that cannot be made) :shrug:

domodwyer:
image

:tada:

write_buffer/src/kafka/instrumentation.rs
domodwyer: I agree, I also believe this largely boils down to personal choice too - I try to avoid personal choice PR comments to avoid unnecessary friction, though obviously sometimes a valid consideration can be raised :+1:

alamb: I like to think about this topic less as a matter of opinion and more as a matter of "readability" --  most code is read many more times than it is written so investing additional time during writing to make it more readable often is worthwhile

So the question comes down what people feel is most readable (which I suppose can be thought of as an opinion 🤔 )

PR #5342, opened 2022-08-08
chore: update to rusoto-less object_store

tustvold:
Updates to use the rusoto-less object store https://github.com/apache/arrow-rs/pull/2352

I have tested out various parts of this, now lets give it a soak test :smile:


clap_blocks/src/object_store.rs
```diff
@@ -253,29 +257,19 @@ pub enum ObjectStoreType {
...
-            if service_account.is_none() {
-                missing_args.push("google-service-account");
-            }
-            MissingObjectStoreConfigSnafu {
-                object_store: ObjectStoreType::Google,
-                missing: missing_args.join(", "),
-            }
-            .fail()
-        }
+    let mut builder = GoogleCloudStorageBuilder::new();
```
tustvold, line 260: These are the new builders added by @alamb

Cargo.lock
```diff
@@ -339,15 +339,15 @@ dependencies = [
  "bytes",
  "chrono",
  "futures",
- "hmac 0.12.1",
+ "hmac",
```
tustvold, line 342: So many out of date dependencies gone now :smile:

object_store_metrics/src/lib.rs
```diff
@@ -689,7 +705,8 @@ mod tests {
     #[tokio::test]
     async fn test_put_get_delete_file() {
         let metrics = Arc::new(metric::Registry::default());
-        let store = Arc::new(LocalFileSystem::new_with_prefix("./").unwrap());
+        let path = std::fs::canonicalize(".").unwrap();
```
tustvold, line 708: Fix for this in https://github.com/apache/arrow-rs/pull/2371

clap_blocks/src/object_store.rs
```diff
@@ -253,29 +257,19 @@ pub enum ObjectStoreType {
...
+
+    if let Some(bucket) = &config.bucket {
+        builder = builder.with_bucket_name(bucket);
+    }
+    if let Some(account) = &config.google_service_account {
+        builder = builder.with_service_account_path(account);
     }
+
+    Ok(Arc::new(LimitStore::new(
```
domodwyer, line 269: COMPOSITION :heart:

lukebond (approved):
no interface difference with the work i'm doing as i use make_object_store() :+1:

object_store_metrics/src/lib.rs
```diff
@@ -163,6 +166,19 @@ impl ObjectStore for ObjectStoreMetrics {
         res
     }
 
+    async fn put_multipart(
+        &self,
+        location: &Path,
+    ) -> Result<(MultipartId, Box<dyn AsyncWrite + Unpin + Send>)> {
+        // TODO: Add instrumentation of multipart requests
+        self.inner.put_multipart(location).await
```
domodwyer, line 174: Is it hard to instrument this? It would be a shame / confusing if we partially loose observability of our object store interactions.

domodwyer (commented):
Looks good - lack of instrumentation is a bit of bummer though.

You might like to add `rusoto_core` to [`deny.toml`](https://github.com/influxdata/influxdb_iox/blob/b05714f833aa41b23fa20779bbf1144cc88417af/deny.toml#L39) to prevent it getting dragged in as a transitive dependency somewhere like we do with openssl :fire:

object_store_metrics/src/lib.rs
tustvold: It is a pain to instrument as the public interface purposefully hides the actual requests. We aren't using this functionality in IOx, so I didn't think it important to instrument just yet

domodwyer: Perhaps we put a `warn!()` or even a `unimplemented!()` so that if it is accidentally used it is at least known - we should definitely have a ticket to track this :+1:

object_store_metrics/src/lib.rs
```diff
@@ -168,15 +168,13 @@ impl ObjectStore for ObjectStoreMetrics {
...
-        // TODO: Add instrumentation of multipart requests
-        self.inner.put_multipart(location).await
+        unimplemented!()
     }
 
-    async fn abort_multipart(&self, location: &Path, multipart_id: &MultipartId) -> Result<()> {
-        // TODO: Add instrumentation of multipart requests
-        self.inner.abort_multipart(location, multipart_id).await
+    async fn abort_multipart(&self, _location: &Path, _multipart_id: &MultipartId) -> Result<()> {
+        unimplemented!()
```
domodwyer, line 177: Thanks 👍

PR #5281, opened 2022-08-03
feat: support for regex comparisons in storage CLI

crepererum:
Needed this for some debugging and it was a rather simple fix.

PR #5343, opened 2022-08-08
fix: create proper storage regex request via CLI

crepererum:
The literal for regex comparisons within the storage API are NOT
strings but regex nodes. This was a mistake in #5281.


alamb (approved):
LGTM -- eventually it would be sweet to have some tests for this code

crepererum:
> LGTM -- eventually it would be sweet to have some tests for this code


There are tests, but they are basically a re-implementation of the parsing code and not really an integration test.

alamb:
> There are tests, but they are basically a re-implementation of the parsing code and not really an integration test.

Yeah, that is when many tests actually are -- a second encoding of the same logic that computers automatically check for discrepancies 😆

crepererum:
Integration tests are now tracked in #5356.

PR #5344, opened 2022-08-08
feat: initial commit of schema merge bulk import tool

lukebond:
Initial import of tool for merging TSM schemas and highlighting inconsistencies.

Point it at an object storage bucket and it will download schema files according to a configurable convention and try and merge them, outputting the resulting merged schema.


import/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+futures = "0.3"
+object_store = { version = "0.3.0", features = ["aws", "rusoto_s3"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0.82"
+thiserror = "1.0.31"
+tokio = { version = "1.17" }
+tracing = { version = "0.1", features = ["max_level_trace", "release_max_level_debug"] }
+tracing-subscriber = "0.2.0"
```
alamb, line 16: I think we have been using the `observability_deps` (which imports tracing / tracing subscriber) elsewhere in the project

```suggestion
observability_deps = { path = "../observability_deps" }
```

import/src/schema/merge.rs
```diff
@@ -0,0 +1,608 @@
...
+            let mut new_field = Field {
+                name: from_field.name.clone(),
+                types: into_field.types.clone(),
+            };
+            new_field.types.extend(from_field.types.clone().into_iter());
+            into_measurement
+                .fields
+                .insert(from_field.name.clone(), new_field);
+        } else {
+            println!("field else");
```
alamb, line 102: maybe some left over printlns

lukebond: thanks!

influxdb_iox/Cargo.toml
```diff
@@ -11,6 +11,7 @@ clap_blocks = { path = "../clap_blocks" }
 data_types = { path = "../data_types" }
 datafusion = { path = "../datafusion" }
 generated_types = { path = "../generated_types" }
+import = { path = "../import" }
```
alamb, line 14: 🤔  perhaps we could make aws a feature in the `import` crate:

So that  would mean basically follow the model of the `clap_blocks` crate here:

https://github.com/influxdata/influxdb_iox/blob/e8231b2986b549b25955423e958bfe6432cd8e54/clap_blocks/Cargo.toml#L30-L33

And then update `influxdb_iox` / Cargo.toml

https://github.com/influxdata/influxdb_iox/blob/e8231b2986b549b25955423e958bfe6432cd8e54/influxdb_iox/Cargo.toml#L81-L83

So it activated those features. Instead of 

```
aws = ["clap_blocks/aws"] # Optional AWS / S3 object store support
```

It would be like

```
aws = ["clap_blocks/aws", "import/aws"] # Optional AWS / S3 object store support
```

import/src/schema/schema_parse.rs
```diff
@@ -0,0 +1,120 @@
+use serde::de::{Deserialize, Deserializer};
+use serde::ser::{Serialize, Serializer};
+use serde::*;
+use std::collections::{HashMap, HashSet};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Schema {
```
alamb, line 7: I think some comments are in order here -- I would especially like to know how this differs from the `Schema` in the `schema` crate and what this is used for. 

It looks like the plan is that this is the schema for an entire Influxdb 2.0 / influx cloud bucket (aka what will become a `namespace` in IOx parlance). 

Perhaps we could rename this to `BucketSchema` or something to make it clearer what it was

import/src/schema/schema_parse.rs
```diff
@@ -0,0 +1,120 @@
...
+    )]
+    pub tags: HashMap<String, Tag>,
+    #[serde(
+        serialize_with = "serialize_map_values",
+        deserialize_with = "deserialize_fields"
+    )]
+    pub fields: HashMap<String, Field>,
+}
+
+fn serialize_map_values<S, K, V>(value: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::Error>
```
alamb, line 27: I am surprised you have to add a custom serialize implementation for serializing/deserializing `HashMap` - I would have expected this to work by default in `serde` (it works well in arrow-rs for example: https://github.com/apache/arrow-rs/blob/master/arrow/src/datatypes/schema.rs#L34-L40)

import/src/schema/schema_parse.rs
```diff
@@ -0,0 +1,120 @@
+use serde::de::{Deserialize, Deserializer};
+use serde::ser::{Serialize, Serializer};
+use serde::*;
+use std::collections::{HashMap, HashSet};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct Schema {
+    pub org_id: String,
+    pub bucket_id: String,
+    pub measurements: HashMap<String, Measurement>,
```
alamb, line 10: Something you might consider is using the `Schema` here instead of `Measurement`:

```suggestion
    pub measurements: HashMap<String, crate::schema::Schema>,
```

Though that might require some serialization / deserialization dancing

import/src/schema/schema_parse.rs
```diff
@@ -0,0 +1,120 @@
...
+            tag.values,
+            HashSet::from(["server".to_string(), "desktop".to_string()])
+        );
+        let field = &measurement.fields.values().next().unwrap();
+        assert_eq!(field.name, "usage");
+        assert_eq!(field.types, HashSet::from(["float".to_string()]));
+        // exercise the Vec<u8> tryfrom impl too
+        assert_eq!(schema, json.as_bytes().to_vec().try_into().unwrap());
+    }
+}
```
alamb, line 120: I wonder if you want to make a `serialize` test for completeness to (like build up a `Schema` and turn it into json?

import/src/schema/merge.rs
```diff
@@ -0,0 +1,606 @@
...
+    });
+    Schema {
+        org_id: org_id.to_string(),
+        bucket_id: bucket_id.to_string(),
+        measurements: merged_measurements,
+    }
+}
+
+fn do_merge_measurement(into_measurement: &mut Measurement, from_measurement: &Measurement) {
+    // merge tags
```
alamb, line 74: from here on down, the merge logic for `Measurement`s is very similar to https://github.com/influxdata/influxdb_iox/blob/main/schema/src/merge.rs

import/src/schema/merge.rs
```diff
@@ -0,0 +1,606 @@
...
+                Tag {
+                    name: "host".to_string(),
+                    values: HashSet::from(["server".to_string(), "desktop".to_string()]),
+                },
+            )]),
+            fields: HashMap::from([(
+                "usage".to_string(),
+                Field {
+                    name: "usage".to_string(),
+                    types: HashSet::from(["Float".to_string()]),
```
alamb, line 128: I don't know if it matters, but the capitalization is inconsistent. Here you use `"Float"` and in parse.rs it uses `"float"`

https://github.com/influxdata/influxdb_iox/pull/5344/files?w=1#diff-78cf4cbd852a417b122c33eb099759319a49c8212926d5eca9270fbaf22da128R95

alamb (approved):
I think the code does what it sets out to do. I had some comments / suggestions but I also think this could be merged as is.

influxdb_iox/src/commands/import/schema.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[derive(Parser, Debug)]
+pub enum Config {
+    /// Merge schemas produced in the IOx prestep
+    Merge(MergeConfig),
+}
+
+/// Merge schema created in pre-step
+#[derive(Parser, Debug)]
+pub struct MergeConfig {
```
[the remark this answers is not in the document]
lukebond: my thinking is we have commands like so: `input -> schema -> merge` and this is the config for the `merge` step. it might be less odd if there was more then one subcommand at the last level, which i expect there will be someday.

influxdb_iox/src/commands/import/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Debug, clap::Parser)]
+pub struct Config {
+    #[clap(subcommand)]
+    command: Command,
+}
+
+#[derive(clap::Parser, Debug)]
+pub enum Command {
+    /// Operations related to jobs in the k8s controller
```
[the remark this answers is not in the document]
lukebond: makes sense. a long way off implementing this anyway so i'll remove

import/src/schema/schema_parse.rs
lukebond: it's because the schema i'm working with has that data as an array, whereas i want it in a map in memory so i know about duplicates more easily.

import/src/schema/schema_parse.rs
lukebond: fair! will do

import/src/schema/schema_parse.rs
lukebond: i am working against a pre-existing schema so this would be incompatible unless i'm missing something

import/src/schema/schema_parse.rs
lukebond: okay :+1:

import/src/schema/merge.rs
lukebond: i addressed this in one of the PRs before bringing this over from the other repo:

> the schema merging code is similar to what is used in the IOx router but i decided not to use that because A) i'm building schemas that, when merged, are potentially bad (e.g. multiple types - in order to identify that very thing) and B) i don't need the underlying parquet metadata struct for this type, it's just an interchange struct to detect schema anomalies. that may change in the future but for now this simple code will suffice.

perhaps i can comment to make this clear to people because it'll come up again i'm sure.

import/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+pub mod schema {
```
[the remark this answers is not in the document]
lukebond: i'll look into it and follow the style you suggest- i am just refactoring this in the next PR anyway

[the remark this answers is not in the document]
lukebond: marking this as resolved even though it's not fixed in this PR- it's coming in the next one! i had already have addressed this with the refactor and you picked up bits i missed.

influxdb_iox/src/commands/import/schema.rs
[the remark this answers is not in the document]
alamb: Makes sense

import/src/schema/schema_parse.rs
alamb: That rationale makes perfect sense. I recommend adding a `//` comment to hopefully avoid some future reader (like myself in 2 months 😆 ) will not be surprised

import/src/schema/schema_parse.rs
alamb: Perhaps you mean that some other tool is writing out JSON that iox needs to interpret? If so, then yes you would have to provide a way to parse that JSON and convert it to a `Schema` -- you will likely have to do eventually to create parquet files so perhaps that would be the right time to consider unificiation

import/src/schema/merge.rs
alamb: It was meant to go to this test:

```rust
    async fn parses() {
        let json = r#"
        {
          "org_id": "1234",
          "bucket_id": "5678",
          "measurements": {
            "cpu": {
              "tags": [
                { "name": "host", "values": ["server", "desktop"] }
              ],
             "fields": [
                { "name": "usage", "types": ["float"] }
              ]
            }
          }
        }
```

import/src/schema/schema_parse.rs
lukebond: i'm guessing perhaps you meant the `schema` crate in the root of IOx rather than `crate::schema` which from this POV would mean `import::schema`? i think that is addressed in the other comment about why I didn't use that merging code, although your point about needing to make parquet files is a good one if so.

i can't really picture what that code is going to look like until i get closer to implementing it, there certainly needs to be some sort of adapter from this schema-like thing to the main IOx one. i think addressing your other comment that suggests renaming this would go a long way towards addressing this.

import/src/schema/merge.rs
lukebond: thanks. have addressed that as part of the refactor coming in the next PR.

import/src/schema/schema_parse.rs
lukebond: to avoid the pain of conflicts against the next PR that i optimistically based off this one before your review, i'm going to punt this to that other PR too. i have removed this file in the factor so it would be messy. but i've done what you suggest.

import/src/schema/fetch.rs
```diff
@@ -0,0 +1,58 @@
...
+    object_store: Arc<DynObjectStore>,
+    prefix: Option<&Path>,
+    suffix: &str,
+) -> Result<Vec<Schema>, FetchError> {
+    let mut schemas: Vec<Schema> = vec![];
+    let mut results = object_store
+        .list(prefix)
+        .await
+        .map_err(FetchError::Fetching)?;
+    loop {
```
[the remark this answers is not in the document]
lukebond: i tried and i think it's a bit beyond my async fu at the moment. i'll make a note to come back and give it a go!

import/src/schema/schema_parse.rs
lukebond: i think this is resolved now by the rename and the comments.

lukebond:
@tustvold i bumped the object store dep to get the bugfix and i think it's done something weird to `Cargo.lock` as i was resolving conflicts. would you mind taking a look before i merge it?

tustvold:
I think you need hakari to regenerate the workspace hack

alamb:
I took the liberty of merging`main` into this PR -- I think it should pass

import/src/schema/fetch.rs
```diff
@@ -5,7 +5,7 @@ use object_store::{path::Path, DynObjectStore, GetResult};
 use thiserror::Error;
 use tokio::select;
 
-use super::schema_parse::Schema;
+use super::schema_parse::AggregateTSMSchema;
```
alamb, line 8: 👌  Nice

PR #5347, opened 2022-08-08
chore: Update docstrings for Partition::sort_keys

alamb:
# Rationale
As part of https://github.com/influxdata/conductor/issues/1068 I am doing some active reading of the compactor / ingester code and I am trying to improve things along the way

# Changes:
Improve docstrings on Partition


data_types/src/lib.rs
```diff
@@ -796,15 +796,24 @@ pub struct Partition {
...
-    /// and, later on, when data with new tags are ingested into this partition
+    /// Even though it is possible for both the unpersisted data
+    /// and/or multiple parquet files to contain different subsets of
+    /// columns, the partition's sort_key is guaranteed to be
+    /// "compatible" across all files. Compatible means that the
+    /// parquet file is sorted in the same order as the partition
+    /// sort_key after removing any missing columns.
+    ///
+    /// Partitions are initially created before any data is persisted
+    /// with an empty sort_key. The partition sort_key is updated as
```
crepererum, line 813: Maybe add a sentence on how updating works, e.g.:

> Updating inserts new column into the existing order. The order of the existing columns relative to each other is NOT changed (e.g. updating `A,B,C` to `A,D,B,C` is legal but updating to `A,C,D,B` is not).

crepererum (approved):

data_types/src/lib.rs
alamb: Excellent call -- done in 263a89e38

PR #5348, opened 2022-08-08
feat: Add metrics on the size of files created by ingestion and used for compaction

### LINKED ISSUE #5327: Add metrics for upgraded size
# Rationale

We want to record upgraded L0-to-L1 size to evaluate how well the compactors work

carols10cents:
Closes #5327

- I undid the commit that used to be in this PR that added gauges.
- In the compactor, I changed the existing `parquet_file_candidate_bytes_gauge` metric to a `parquet_file_candidate_bytes` histogram - this records the sizes and compaction levels of files selected by the filtering step. This metric may end up looking very similar to the new metric, but it's recorded before compaction is attempted, so it might contain values for compaction operations that fail.
- In the compactor, I added a new `compaction_input_file_bytes` histogram that, after a compaction operation succeeds and the catalog is updated to mark files as deleted, records the file sizes and compaction levels of the files that went into the compaction operation.
- In the ingester, I also added a `persisted_file_size_bytes` histogram that will record the sizes of level 0 files as the ingester persists them to object storage.

I *think* that's everything.


pauldix:
I'm not sure what exactly we want to determine based on the issue #5327. What's the question we're hoping to answer here? Some guesses:
1. Are L0 files getting compacted? (we'd only need two counts: L0 files created, and L0 files compacted (regardless of upgrade or rewrite))
1. How many L0 files are getting rewritten vs. how many upgraded (this would just be a single counter)
1. What size are files getting compacted to? (this would be way harder to effectively measure while compactions are actively running, and I'm not sure how useful it is)

I'm not sure that a gauge is useful for answering any of these questions.

carols10cents:
Yeah I'm not sure either :-/ "to evaluate how well the compactors work" is what Nga said.

carols10cents:
What *would* you find useful, @pauldix ?

pauldix:
I think maybe the following:
1. Counter for L0 files getting created (would be in the ingester)
2. Counter for L0 files getting compacted (in the compactor)
3. Counter for L1 files getting compacted (in the compactor)

Then we'd be able to visualize these as rates, which would be pretty easy to compare. I'm not sure the sizes are very helpful, unless they're histograms, which might be interesting. If doing that then we'd also get the counter from the histogram as part of it. 

So maybe:
1. Histogram of L0 file sizes created
2. Histogram of L0 file sizes compacted
3. Histogram of L1 file sizes compacted

You could start with some buckets of like 500KB, 1MB, 3MB, 10MB, 30MB, Inf.

What do you think?

carols10cents:
> 1. Histogram of L0 file sizes created
> 2. Histogram of L0 file sizes compacted
> 3. Histogram of L1 file sizes compacted

I can do this! I want to be sure I understand what you mean, though-- when you say "L0 file sizes compacted" and "L1 file sizes compacted", do you mean "sizes of the files selected for compacted"/"input files to compaction" or "sizes of the files created by the compaction operation"/"output files from compaction" or both?

carols10cents:
I *think* this is ready for re-review now, @pauldix !

compactor/src/parquet_file_combining.rs
```diff
@@ -317,11 +330,19 @@ pub(crate) async fn compact_parquet_files(
         "sequencer_id",
         format!("{}", partition.sequencer_id()).into(),
     )]);
-    for (compaction_level, file_count) in files_by_level {
+
+    for (compaction_level, file_count) in num_files_by_level {
         let mut attributes = attributes.clone();
         attributes.insert("compaction_level", format!("{}", compaction_level as i32));
-        let compaction_counter = compaction_counter.recorder(attributes);
+        let compaction_counter = compaction_counter.recorder(attributes.clone());
```
pauldix, line 337: The `compaction_counter` should be removed as it's redundant. Having the histogram for sizes will automatically insert a metric to track the count and the sum of the sizes. More info here: https://www.robustperception.io/how-does-a-prometheus-histogram-work/

compactor/src/parquet_file_combining.rs
```diff
@@ -317,11 +330,19 @@ pub(crate) async fn compact_parquet_files(
...
         compaction_counter.inc(file_count);
+
+        let compaction_input_file_bytes = compaction_input_file_bytes.recorder(attributes);
+        if let Some(sizes) = file_sizes_by_level.get(&compaction_level) {
```
pauldix, line 341: Rather than increment the histogram by the total, it's better to increment it per file. So as you're looping through the files higher up you should be adding each one to the histogram. This will give us the correct view of the count of the individual files and the sizes of those files in the histogram.

Separately, if we want a histogram that tracks the total sizes of a compaction, then we can build that up, which might also be worth doing. So one histogram to track the sizes of the individual files (along with an attribute that tracks the level) and then another that tracks the total compaction size (this one shouldn't have level on it).

pauldix (changes requested):
Made some recommendations for changing up the metrics a little bit. Let me know if that makes sense.

compactor/src/parquet_file_combining.rs
carols10cents: This is incrementing it per file, I think-- this hashmap is holding a Vec of sizes, and on the next line I'm iterating through each size and recording each:

```
            for &size in sizes {
                compaction_input_file_bytes.record(size);
            }
```

Am I misunderstanding how the metrics work?

> (this one shouldn't have level on it).

I can take that off :) My thinking of recording these metrics here rather than higher up is that here we're sure the compaction and catalog update has succeeded...?

carols10cents:
Two new commits for you to review @pauldix -- especially in ef716a5b9032b673f037490ad537a0bf150667fd where the tests are updated, it should be easier to see that the individual file sizes are being logged separately, if I'm understanding metrics correctly (there's always a chance I'm not!)

pauldix (approved):
Looks good. We'll also want to track L2 once those get in.

NGA-TRAN (commented):
I went over the conversation between Carol and Paul and skimmed the code. It makes sense. I will use them when I work on dashboard

PR #5349, opened 2022-08-08
chore: make terminology in iox_query::Provider consistent (remove super notation)

alamb:
# Rationale
As part of https://github.com/influxdata/conductor/issues/1068 I am doing some active reading of the compactor / ingester code and I am trying to improve things along the way

The provider sometimes uses `super_sort_key` and sometimes uses `sort_key` or `output_sort_key` to refer to the same concept. Depending on where the query engine is being used, the `super_sort_key` may actually be a sort key that was computed but not yet the partition sort key.

Let's use consistent terminology to make the code less confusing

cc @NGA-TRAN (who I know is on vacation this week but I want her to know this when she gets back)

# Changes
1. Remove references to `super` sort keys
2. Clean up some logging messages
3. Be more specific about *which* sort key is meant by renaming `sort_key` to `output_sort_key` or `chunk_sort_key` as appropriate


iox_query/src/provider.rs
```diff
@@ -639,14 +639,14 @@ impl Deduplicater {
         Some(partition_sort_key)
     }
 
-    // return true if the super_sort_key covers sort_key and same column order
-    fn sort_key_cover_and_same_order(super_sort_key: &SortKey, sort_key: &SortKey) -> bool {
-        if super_sort_key == sort_key {
+    // return true if output_sort_key covers sort_key and has the same column order
+    fn sort_key_cover_and_same_order(output_sort_key: &SortKey, sort_key: &SortKey) -> bool {
```
crepererum, line 643: Is `sort_key` here the `chunk_sort_key`? If so, this could be renamed as well for clarity.

iox_query/src/frontend/reorg.rs
```diff
@@ -80,7 +80,7 @@ impl ReorgPlanner {
     {
         let scan_plan = ScanPlanBuilder::new(schema, self.ctx.child_ctx("compact_plan"))
             .with_chunks(chunks)
-            .with_sort_key(sort_key)
+            .with_output_sort_key(sort_key)
```
crepererum, line 83: Maybe rename `sort_key` here as well to `output_sort_key`.

iox_query/src/frontend/reorg.rs
```diff
@@ -159,7 +159,7 @@ impl ReorgPlanner {
 
         let scan_plan = ScanPlanBuilder::new(schema, self.ctx.child_ctx("split_plan"))
             .with_chunks(chunks)
-            .with_sort_key(sort_key)
+            .with_output_sort_key(sort_key)
```
crepererum, line 162: Maybe rename `sort_key` to `output_sort_key`.

crepererum (approved):
Left suggestions for a few further improvements but I already love it. It's waaaay less confusing (IMHO!).

iox_query/src/frontend/reorg.rs
alamb: in cf913fa12

iox_query/src/frontend/reorg.rs
alamb: in cf913fa12

iox_query/src/provider.rs
alamb: Yes, good spot. Done in 8399a029b

alamb (commented):
Thanks @crepererum  -- those were great comments

PR #5357, opened 2022-08-09
chore: temporary allow http always

tustvold:
The auth system uses the same reqwest client, which is causing the metadata requests to fail with `builder error for url (http://169.254.169.254/latest/api/token): URL scheme is not allowed`


crepererum (approved):

clap_blocks/src/object_store.rs
```diff
@@ -280,7 +280,8 @@ fn new_gcs(_: &ObjectStoreConfig) -> Result<Arc<DynObjectStore>, ParseError> {
 #[cfg(feature = "aws")]
 fn new_s3(config: &ObjectStoreConfig) -> Result<Arc<DynObjectStore>, ParseError> {
     let mut builder = AmazonS3Builder::new()
-        .with_allow_http(config.aws_allow_http)
+        // TODO: Temporary fix
```
alamb, line 283: This needs a ticket tracking what is wrong

lukebond (approved):
i don't have full context but guessing reqwest does auth with http and the new builder doesn't have it on by default. seems reasonable as a quick fix for prod

alamb:
Why are we doing authentication over `http` -- I am sure I am missing something but that seems like a bad idea

tustvold:
> Why are we doing authentication over http -- I am sure I am missing something but that seems like a bad idea

The instance metadata endpoint is a loopback IP that is over http

clap_blocks/src/object_store.rs
tustvold: I'll just fix it upstream, this was just a quick get prod unblocked

tustvold: https://github.com/influxdata/influxdb_iox/pull/5359

PR #5364, opened 2022-08-10
feat: concurrent garbage collector deletes

crepererum:
This should speed up the prod process a bit.


garbage_collector/src/deleter.rs
```diff
@@ -7,20 +8,31 @@ use tokio::sync::mpsc;
...
-        let path = item.location;
-        if dry_run {
-            info!(?path, "Not deleting due to dry run");
-        } else {
-            info!("Deleting {path}");
-            object_store
-                .delete(&path)
-                .await
-                .context(DeletingSnafu { path })?;
+    futures::stream::unfold(items, |mut items| async move {
```
tustvold, line 14: I think a simpler way to write this would be to use a combination of map_ok and buffered.

tustvold (approved):
I think this can be written more simply, but this looks fine

garbage_collector/src/deleter.rs
crepererum: done (needs an additional `try_collect` though which magically works because you can collect `()` into `()`).

tustvold: I meant to remove the unfold...

crepererum: I don't think so: <https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html>

tustvold:
Removed automerge to give chance to respond, feel free to add back when happy

garbage_collector/src/deleter.rs
tustvold: They split it out into a separate crate (which I'm fairly certain we already depend on) - https://tokio.rs/tokio/tutorial/streams

Up to you

crepererum: done

garbage_collector/src/deleter.rs
```diff
@@ -7,20 +8,30 @@ use tokio::sync::mpsc;
...
+                } else {
+                    info!("Deleting {path}");
+                    object_store
+                        .delete(&path)
+                        .await
+                        .context(DeletingSnafu { path })
+                }
+            }
+        })
+        .buffer_unordered(concurrent_deletes)
```
alamb, line 32: `buffer_unordered` is so cool

alamb (commented):
👌 

whitespace blind diff shows what is going on in case anyone is interested: https://github.com/influxdata/influxdb_iox/pull/5364/files?w=1

And I can confirm this has increased the garbage collection rate from 40 ops/sec to 230 ops/sec which is good 🧹  but we need to keep an eye on the catalog and make sure the garbage collector doesn't cause issues (if it does I think we can start batching the postgres checks)

Screen Shot 2022-08-10 at 6 36 46 AM

alamb:
Thanks @crepererum 👍

PR #5365, opened 2022-08-10
chore: enable and fix warnings for `clap_blocks`

crepererum:
Esp. this fixes "unused import" warnings when not all features are
enabled, so developer IDEs don't shout.


domodwyer (approved):
:+1:

lukebond (approved):

PR #5367, opened 2022-08-10
chore: import validate merged schema

lukebond:
Adds some validation code that checks the merged TSM schemas for tags/fields with the same name, and fields with multiple types.

Also some refactoring which I've noted with comments in this PR.

Validation output looks like this (with some crafted schemas dropped into Minio- ask me if you want to see them):

```
$ cargo run --features aws,gcp,azure -- import schema merge --object-store=s3 --org-id=1234 --bucket-id=5678 --bucket=tsm-schemas --prefix=
Merged schema:
AggregateTSMSchema { org_id: "1234", bucket_id: "5678", measurements: {"weather": AggregateTSMMeasurement { tags: {"location": AggregateTSMTag { name: "location", values: {"berlin", "london"} }, "temperature": AggregateTSMTag { name: "temperature", values: {"true"} }}, fields: {"temperature": AggregateTSMField { name: "temperature", types: {"Integer", "Float"} }} }, "cpu": AggregateTSMMeasurement { tags: {"host": AggregateTSMTag { name: "host", values: {"server", "desktop", "gadget"} }}, fields: {"usage": AggregateTSMField { name: "usage", types: {"Float", "Integer"} }} }} }
Schema conflicts:
- Measurement 'weather' has a tag and field with the same name: temperature
- Measurement 'weather' has field 'temperature' with multiple types: ["Integer", "Float"]
- Measurement 'cpu' has field 'usage' with multiple types: ["Float", "Integer"]
```


import/src/lib.rs
```diff
@@ -1,5 +1,159 @@
-pub mod schema {
-    pub mod fetch;
-    pub mod merge;
-    pub mod schema_parse;
+use serde::de::{Deserialize, Deserializer};
```
lukebond, line 1: here i've moved the code from `schema_parse.rs` into here, and done some renaming. also replaced the above mod declaration with a `mod.rs`.

import/src/lib.rs
```diff
@@ -1,5 +1,159 @@
...
+
+pub mod schema;
+
+/// This struct is used to build up structs from TSM snapshots that we are going to use to bulk
+/// ingest. They will be merged, then validated to check for anomalies that will complicate bulk
+/// ingest such as tags/fields with the same name, or fields with different types across the whole
+/// dataset. It is not the same as an IOx schema, although it is similar and some of the merge code
+/// is similar. It's a transient data structure.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct AggregateTSMSchema {
```
lukebond, line 14: previously the types within the schema didn't have the `AggregateTSM` prefix. it's clearer this way to show that they're not the same as IOx schema. but it creates a lot of noise in the PR

import/src/schema/mod.rs
```diff
@@ -1,2 +1,3 @@
-mod merge;
-mod schema_parse;
+pub mod fetch;
```
lukebond, line 1: `fetch` was previously missing, by accident

andrewcharlton (approved):
Looks good to me :+1:

import/src/schema/validate.rs
```diff
@@ -0,0 +1,133 @@
...
+    FieldWithMultipleTypes {
+        measurement: String,
+        name: String,
+        types: Vec<String>,
+    },
+}
+
+pub fn validate_schema(schema: &AggregateTSMSchema) -> Result<(), Vec<ValidationError>> {
+    let mut errors: Vec<ValidationError> = vec![];
+    for measurement in &schema.measurements {
```
alamb, line 23: It might be easier to read if this was something like:

```suggestion
    for (measurement_name, measurement) in &schema.measurements {
```

And then you could use `measurement_name` in the loop body rather than `measurement.0` and `measurement` instead of `measurement.1`

alamb (approved):

import/src/schema/validate.rs
lukebond: > You also consolidated some of code out of schema_parse too, right?

i did, but the hakari bot's subsequent commit relegated by helpful comment to the dustbin of history :cry: 

this code in `lib.rs` you're commenting on is the move out of `schema_parse.rs`

lukebond: took your suggestion, thanks @alamb

Issue #5320, opened 2022-08-05
Querier Cache: Improve backend policies


crepererum:
# Why
Currently multiple cache policies (e.g. TTL, LRU) are combined by stacking multiple backends on top of each other. This however has the weird (but intended) consequence that an outer backend doesn't know about changes an inner backend does. So if you stack an TTL policy on top of an LRU system, the TTL system may still want to prune or refresh values even though the LRU subsystem already evicted the value. If you perform the stacking the other way around, then the LRU layer does not get updated resource consumption after a TTL refresh.

This is somewhat required to implement #5318 properly.

Overall I find this reasoning really hard and the API clunky.

# What
Redesign the policies to easily interact with each other in a single policy _container_.

# How
Rough draft, but I would replace the multiple policy layers with a single one:

```rust
struct PolicyBackend;

impl CacheBackend for PolicyBackend {
   ...
}

impl PolicyBackend {
    /// Create new backend on inner backend.
    fn new(inner: Box<dyn CacheBackend>) -> Self;

    /// Get endpoint to modify the inner backend AND notify all policies.
    fn get_modifier(&self) -> Modifer;

    /// Add new policy to this backend that will be notified for every change.
    fn add_subscriber(&mut self, policy: Box<dyn Subscriber>);
}

struct Modifer;

impl CacheBackend for Modifer {
    ...
}

trait Subscriber {
    fn set(&self, k: &K, v: &V);
    fn remove(&self, k: &K);
}
```

This interface is obviously a bit cyclic, i.e. you'll have the backend that points to a subscriber that points to a modifier that points to the backend again. I'll have to think how to deal with drop orders. Also there's the slight risk of never-ending loops/recursions during updates.

I'll probably design the modifier and subscriber system in two separate modules and just offer the `PolicyBackend` as a convenient and simpler all-in-one wrapper on top of it.

alamb:
The problem of complex policy interactions makes sense and it a good one to address. 

Given that the existing Cache system is fairly complicated (with several levels of generics and some potential cycles already), I wonder if it is time to take a step back and reassess the larger structure?

Like for example, have you considered  removing the concept of multiple  `Backend`s that wrap each others, and instead have a single `Backend` implementation (possibly typed on the value it contains) that takes a `Policy` and then  isolating whatever policy about how to expire the cache contents into the `Policy`?

There are probably other better structures I haven't thought of as well

crepererum:
[...]

Issue #5318, opened 2022-08-05
Querier Cache: Rework TTL


crepererum:
# Why
Currently [TTL (time-to-live)](https://github.com/influxdata/influxdb_iox/blob/main/cache_system/src/backend/ttl.rs) is used for some caches to enforce "freshness", e.g. we want to renew the namespace schema after a while. The way it is implemented however is the following:

> When we request a certain key from the cache and it is cached, we check its age[^1] and delete it if it is too old.

This archives freshness, however it adds substantial latency to requests since now the key is gone and we need to actively fetch the data again.

Now you may wondering if deletes are not a good thing because they gradually fade out unused data. However time alone is not a reason why data should be uncached. The only reason is resource scarcity / pressure and this is already achieved by the [LRU subsystem](https://github.com/influxdata/influxdb_iox/blob/main/cache_system/src/backend/lru.rs).

# What
Change the TTL system from the current lazy deletion system to some active refresher. A value should be refreshed in certain intervals but while it is being loaded, the existing value should still be available. Refreshing must be stopped when the key is evicted from the cache due to other reasons (e.g. [LRU](https://github.com/influxdata/influxdb_iox/blob/main/cache_system/src/backend/lru.rs) or [predicate-based evictions](https://github.com/influxdata/influxdb_iox/blob/main/cache_system/src/backend/shared.rs)).

# How
Not sure yet, this needs some proper API design so it is easy to understand and use.

[^1]: Actually we check the age of all keys using a heap data structure, but from an observer point of view this doesn't make a difference.

alamb:
> Change the TTL system from the current lazy deletion system to some active refresher.

>  A value should be refreshed in certain intervals but while it is being loaded, the existing value should still be available. 

I think this would be a good first step (rather than expire an entry in the cache and wait for it to be reloaded, instead, cause it to be actively refreshed).

The key will be how to decide which elements to proactively refresh without overwhelming the backend (e.g. catalog)

One thought might be to give each querier a 'proactive refresh' budget like "10 catalog ops / second" and then, among keys that  have not been refreshed within some interval (10s?) pick the ones that were most recently accessed

That heuristic intuition is that cache access is a estimate of how important that value is and the more important a value is the more resource should be spent on keeping it up to date.

alamb:
Another possible thing to consider is prioritize negative hits / errors somehow - like try and refresh more actively cache contents that caused an error (e.g. "Table not found")

crepererum:
[...]

PR #5370, opened 2022-08-10
feat: new policy system for caches

crepererum:
This is the framework part for the policy system outlined in #5320. It
does NOT port any existing policies (likely TTL, LRU and shared) over to
the new system. This will be a follow-up.

Looks huge, but the core logic is only about 200 lines, the rest are docs and tests :wink:


cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+/// #[derive(Debug)]
+/// struct EvenNumberPolicy;
+///
+/// type CR = ChangeRequest<&'static str, u64>;
+///
+/// impl Subscriber for EvenNumberPolicy {
+///     type K = &'static str;
+///     type V = u64;
+///
+///     fn set(&mut self, k: &'static str, v: u64) -> Vec<CR> {
```
alamb, line 100:
```suggestion
///     fn set(&mut self, k: &'static str, v: u64) -> Vec<CR> {
///       // When new key `k` is set to value `v` if `v` is odd, 
///       // request a change to set `k` to `v+1`
```

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+/// backend.add_policy(|_callback_backend| EvenNumberPolicy);
+///
+/// backend.set("foo", 8);
+/// backend.set("bar", 9);
+///
+/// assert_eq!(backend.get(&"foo"), Some(8));
+/// assert_eq!(backend.get(&"bar"), Some(10));
+/// ```
+///
+/// And here is a more active backend that regularly writes the current system time to a key:
```
alamb, line 124: I don't see why "regularly write current system time to a key" needs a policy 🤔 

Some other interesting and illustrative policies might be one  that writes the last update time to some special key,  or one that clears the cache after 10 values have been inserted?

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+                subscribers: Vec::new(),
+            }))),
+        }
+    }
+
+    /// Adds new policy.
+    ///
+    /// See documentation of [`PolicyBackend`] for more information.
+    ///
+    /// This is called with a function that receives the "callback backend" to this backend and should return a
```
alamb, line 237: is the `callback_backend` intended to be the backend to which the policy applies? 

I guess I don't totally understand in what cases a Policy would need to be constructed from an instance of the backend it is controlling.

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+        lock_inner!(guard = self.inner);
+        guard.inner.is_empty()
+    }
+
+    fn as_any(&self) -> &dyn std::any::Any {
+        self
+    }
+}
+
+/// Same as [`PolicyBackend`] but with a weak pointer to inner to avoid cyclic references.
```
alamb, line 287: I wonder if we need both? Maybe we could always just use `BackendForPolicy`?

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+        }]));
+
+        backend.remove(&String::from("b"));
+    }
+
+    #[test]
+    fn test_basic_propagation() {
+        let mut backend = PolicyBackend::new(Box::new(HashMap::new()));
+        backend.add_policy(create_test_policy(vec![
+            TestStep {
```
alamb, line 510: this is a pretty clever mock / expect framework

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+            action: TestAction::CallBackendDirectly(ChangeRequest::Remove {
+                k: String::from("b"),
+            }),
+        }]));
+
+        backend.remove(&String::from("a"));
+    }
+
+    #[test]
+    fn test_basic_set_set() {
```
alamb, line 557: This test is more like `basic_set_get` rather than `set_set`? Or maybe `set_get_get`?

cache_system/src/backend/policy/mod.rs
```diff
@@ -0,0 +1,1103 @@
...
+            },
+        ]));
+
+        backend.set(String::from("a"), 11);
+
+        assert_eq!(backend.get(&String::from("a")), Some(13));
+    }
+
+    #[test]
+    fn test_ping_pong() {
```
alamb, line 768: you must have had fun writing this one

alamb (approved):
The documentation ,examples and tests are 👍 

My gut feeling is that the implementation might be less complicated (e.g. could avoid RecursiveMutex'es  weak references and inner locks) if didn't strive so hard to keep the  the `Policy` and `CacheBackend`s separate.

That is just a "feeling" however and I don't really have anything specific to add (and probably won't have anything specific unless I tried to implement something myself) -- and given the quality of the tests I can't say anything other than 🚢 🇮🇹 🥳

cache_system/src/backend/policy/mod.rs
crepererum: > Some other interesting and illustrative policies might be one that writes the last update time to some special key, or one that clears the cache after 10 values have been inserted?

These are both reactive/lazy policies. I wanted to have one example where a policy can do something on their own in the background because this is what new TTL/refresh system (see #5318) will be doing. And just writing a key with some "I did something" marker was the simplest thing I could come up with.

cache_system/src/backend/policy/mod.rs
crepererum: Then who holds/owns the strong reference? If you have only weak pointers, the `Arc` will be freed.

cache_system/src/backend/policy/mod.rs
crepererum: See the reasoning in the "big docs" under "Policies & Recursion": There are two ways a policy can influence a backend:

1. it can react to a change (this is what LRU for example will be doing)
2. it can initiate a change (this is what the new TTL/refresher will be doing, because the refresh operation -- even when it is triggered by a request -- will `SET` the key after a while since it is an async background operation)

For case (2) which is NOT recursive, the policy needs some handle/object/interface (call it how you want) to interact with.

cache_system/src/backend/policy/mod.rs
crepererum: I now realize that I forgot to forward `GET` requests to the subscribers (which is what LRU and TTL will need) and here we only have an initial SET which is "answered" by the policy using another SET. The GET at the end is a mere check/test and not really part of the policy test.

I'll GET support for the subscribers in a follow-up PR.

cache_system/src/backend/policy/mod.rs
alamb: Yeah, I guess I was struggling with the code being called a "Policy" as it seemed more like it is a subscriber of events that then could also influence the backend. Though now that I type that description, policy does sound like a good description 🤔

cache_system/src/backend/policy/mod.rs
alamb: I see -- makes sense -- thank you for the explination

PR #5372, opened 2022-08-10
chore: added helper script for a docker catalog

lukebond:
Made this little helper script to spin up a Postgres container initialised as a catalog. Thought it might be useful for others!


scripts/docker_catalog.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -eu -o pipefail
+
+# bail out if it's already running; let the user delete it if that's what they want
+docker container inspect postgres &> /dev/null && \
+  { echo "Postgres container already running. Remove it with \`docker container rm -f postgres\` and run again."; exit 1; }
+
+# start postgres docker container
+echo "Starting Postres container..."
+docker container run -d \
```
crepererum, line 10: I'm **personally** not a fan of docker containers running in the background and would rather wanna see `--rm` passed here and some "wait for key" blocker at the end of the script so when the script exists, the postgres is gone as well (incl. the port allocation).

scripts/docker_catalog.sh
```diff
@@ -0,0 +1,37 @@
...
+  --name postgres \
```
crepererum, line 11: Maybe name it `iox-postgres` to make it a bit clearer.

crepererum (approved):
Looks good (although I think we really should have a file-based catalog at some point). Some minor nits, but they are partly personal flavor so feel free to ignore them.

PR #5375, opened 2022-08-11
feat: improve cache policies and their testing

crepererum:
1. add GET support for the subscribers (this is needed so that
   TTL/refreshers and LRU systems "know" when a key was used)
2. improve multi-threading test to not rely on a wait loop but use a 2nd
   barrier instead
3. ensure that panics in the testing background thread are propagated
   and make the test fail
4. implement defaults for `Subscriber` methods to reduce boilerplate for
   implementators

Helps with #5320.

The changes in the tests are mostly due to the fact that the `TestSubscriber` now "sees" every `GET` event.


crepererum:
Ah, I forgot one interaction and need one more round on the semantics.

alamb:
I will hold off reviewing until this PR is marked as ready for review again

crepererum:
@alamb you're good to go

cache_system/src/backend/policy/mod.rs
```diff
@@ -72,6 +72,13 @@ macro_rules! lock_inner {
 /// The original requests will return to the caller once all tasks are completed.
 ///
 ///
+/// # `GET`
+/// The return value for [`CacheBackend::get`] is fetched from the inner backend AFTER all changes are applied.
+///
+/// Note [`ChangeRequest::Get`] has no way of returning a result to the [`Subscriber`] that created it. The "changes"
```
alamb, line 78: makes sense

cache_system/src/backend/policy/mod.rs
```diff
@@ -390,20 +400,40 @@ pub trait Subscriber: Debug + Send + 'static {
...
+        // do nothing by default
+        vec![]
+    }
+
     /// Set value for given key.
     ///
     /// It is OK to set and override a key that already exists.
-    fn set(&mut self, k: Self::K, v: Self::V) -> Vec<ChangeRequest<Self::K, Self::V>>;
+    fn set(&mut self, _k: Self::K, _v: Self::V) -> Vec<ChangeRequest<Self::K, Self::V>> {
+        // do nothing by default
```
alamb, line 413: 👍  I like the default impls

cache_system/src/backend/policy/mod.rs
```diff
@@ -390,20 +400,40 @@ pub trait Subscriber: Debug + Send + 'static {
...
+        vec![]
+    }
 }
 
 /// A change request to a backend.
 #[derive(Debug, PartialEq)]
 pub enum ChangeRequest<K, V> {
+    /// Request to `GET` a value.
+    ///
+    /// Since the result is NOT propagated back the the [`Subscriber`], this is mostly useful to send "keep alive" requests.
```
alamb, line 431: thank you for the comments explaining rationale for using the request

cache_system/src/backend/policy/mod.rs
```diff
@@ -817,98 +1025,168 @@ mod tests {
...
+                condition: TestBackendInteraction::Get {
+                    k: String::from("a"),
+                },
+                action: TestAction::ChangeRequests(vec![]),
+            },
+        ]));
+
+        backend.set(String::from("a"), 1);
+        barrier_pre.wait();
+        barrier_post.wait();
```
alamb, line 1189: the test success in this case is that no `panic` is called (because the `TestStep`s are completed successfully), right?

cache_system/src/backend/policy/mod.rs
```diff
@@ -1065,6 +1379,29 @@ mod tests {
...
+                    }
+                    TestSubscriberBackgroundTask::Started(handle) => {
+                        // propagate panics
+                        if let Err(e) = handle.join() {
+                            if let Some(err) = e.downcast_ref::<&str>() {
+                                panic!("Error in background task: {err}")
+                            } else if let Some(err) = e.downcast_ref::<String>() {
+                                panic!("Error in background task: {err}")
+                            } else {
+                                panic!("Error in background task: <unknown>")
```
alamb, line 1397: I think `e` is display, so can't you do:

```suggestion
                                panic!("Error in background task: {}", e)
```

?

alamb (approved):

cache_system/src/backend/policy/mod.rs
crepererum: yes.

cache_system/src/backend/policy/mod.rs
crepererum: This is not `std::error::Error`. It's actually `Box<dyn Any + Send>` which is NOT printable by default. Yeah, Rust allows you to panic with arbitrary types (although nobody does that we should always end up with `&str` or `String` (both can come out of `panic!(...)`).

alamb: I vaguely remember having issues when someone panic'd with something that was Displayable but not `&str` or `&String` (`panic!()` with a format string actually called the panic hook with some crazy typed thing, but perhaps by the time the JoinHandle passes it back it will be a `&str` or `&String`)

More context

https://github.com/influxdata/influxdb_iox/pull/3130/files#r750597059

crepererum: This ain't a panic hook though. We don't have the actual backtrace / panic info at hand (that won't be `Send`). The join handle solely provides us with the value that was "raised" on the other thread.

In any case: this is solely for testing and shall provide a reasonable output. The more important part of this code snipped is to panic no matter what the panic payload on the other thread was so the test function panics and the test fails. This is tested by one of the "meta" tests.

PR #5378, opened 2022-08-11
chore: Upgrade to Rust 1.63

carols10cents:
Clippy learned some new tricks, they seem fine to me!


alamb (approved):
Looks like an improvement to me -- thanks @carols10cents !

PR #5382, opened 2022-08-12
refactor: abstract change requests for cache policies

crepererum:
Tl;Dr; Lets replace the hard-coded enum of change requests with a more
flexible and easier-to-understand function-based approach. It also
unifies the interface for "initial requests" and "reactions".

Story time:

I just wanted to port over the "shared" backend with its "remove if"
functionality to the policy framework. Turns out that this backend does
not just use simple atomic operations but is basically a
"compare&exchange"-like operation (technically: GET+REMOVE) that relies
on a single mutable access lifetime to make sure that this behavior is
sane and nobody messes with the value between the GET and REMOVE. I call
this a "compound request".

The new policy framework did not support this kind of requests so I
thought how I could shoehorn it in. Basically I wanted to extend
`ChangeRequest` to support these kind of operations. But listing
combinations explicitly kinda seemed like some hack. So my brain slowly
came up with an abstract approach how to encode these operations and
after a while I was close to design some kind of mini-VM. I also
considered some ugly workarounds or to not express the "remove if"
functionality" as a policy. My nice vision of a policy framework started
to crack...

This is when sanity kicked in and I realized: every change request is
technically just a function on a (virtual) cache backend. Sure,
all-powerful Rust functions allow you to do ugly stuff that will result
in deadlocks, but with good docs and helpers for the common operations
the risk is very low. Furthermore this is already close to the system I
had for the "initial requests" and which I called a "callback backend".
So I unified both systems and made the change request abstract and all
tests pass and I think it is the better design.

Helps with #5320.


cache_system/src/backend/policy/mod.rs
```diff
@@ -149,12 +166,17 @@ macro_rules! lock_inner {
...
 /// struct NowPolicy {
 ///     cancel: Arc<AtomicBool>,
-///     handle: JoinHandle<()>,
+///     join_handle: Option<JoinHandle<()>>,
 /// };
 ///
 /// impl Drop for NowPolicy {
 ///     fn drop(&mut self) {
 ///         self.cancel.store(true, Ordering::SeqCst);
+///         self.join_handle
```
crepererum, line 175: This is a drive-by fix. Technically this is not required for the example but I didn't want to leave the anti-pattern of leaking thread handles in on of the examples.

cache_system/src/backend/policy/mod.rs
```diff
@@ -445,21 +504,71 @@ pub enum ChangeRequest<K, V> {
...
-    ///
-    /// It is OK to remove a key even when it does not exist.
+    /// [`REMOVE`](CacheBackend::remove).
     Remove {
         /// Key.
         k: K,
     },
 }
 
+/// Specialized [`CacheBackend`] that forwars changes and requests to the underlying backend of [`PolicyBackend`] but
```
alamb, line 514:
```suggestion
/// Specialized [`CacheBackend`] that forwards changes and requests to the underlying backend of [`PolicyBackend`] but
```

cache_system/src/backend/policy/mod.rs
```diff
@@ -1207,6 +1277,133 @@ mod tests {
...
+    /// We've decided to chose option 1. This test ensures that by setting up a compound request (reacting to
+    /// `set("a", 11)`) with a compound of two operations (`set("a", 12)`, `set("a", 13)`) which we call `C1` and `C2`
+    /// (for "compound 1 and 2"). The two policies react two these two compound operations as followed:
+    ///
+    /// |    | Policy 1       | Policy 2       |
+    /// | -- | -------------- | -------------- |
+    /// | C1 | `set("a", 14)` | `set("a", 15)` |
+    /// | C2 | `set("a", 16)` | --             |
+    ///
+    /// For option (1) the outcome will be `"a" -> 16`, for option (2) the outcome would be `"a" -> 15`.
```
alamb, line 1294: 👍  nice explanation -- thank you

alamb (approved):
makes sense to me 👍  I think a little API churn as you try to integrate new logic into existing code is somewhat inevitable. 

It looks to me like you are iterating towards a good solution

cache_system/src/backend/policy/mod.rs
```diff
@@ -50,19 +50,31 @@ macro_rules! lock_inner {
...
-///   interface but shall not hold any locks on internal data structures of the policies. The virtual backend that is
-///   used between [`PolicyBackend`] and the policies is called "callback backend".
-/// - Policies react to changes NOT by using the "callback store" but by returning the changes they want to perform,
-///   dropping all the locks on their internal data structures.
+/// All interaction of the policy with a [`PolicyBackend`] happen throw a proxy object called [`ChangeRequest`]. The
+/// [`ChangeRequest`] encapsulates a single atomic "transaction" on the underlying store. This can be a simple operation
+/// as [`REMOVE`](CacheBackend::remove) but also combound operation like "get+remove" (e.g. to check if a value needs to
+/// be pruned from the cache). The policy has two ways of issuing [`ChangeRequest`]s:
+///
+/// 1. **Initial / self-driven:** Upon creation the policy receives a [`CallbackHandle`] that it can use initiate
```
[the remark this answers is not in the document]
crepererum: Technically it is not direct but indirect mutation because the handle uses the mutexed inner state. A channel may archive the same, but this would just move the locking around into yet another background loop OR we would also implement the "normal" access (e.g. `<PolicyBackend as CacheBackend>::get`) via channels. Then however you need to solve the API concern regarding blocking (at least for the ordinary access, ideally also for the access via a handle) and you need a one-shot back-channel for GET requests. Overall I think it would complicate things more than it helps, but I keep it in mind for future changes since it might also simplify things at some point.

PR #5383, opened 2022-08-12
docs: router writes to partitions

domodwyer:
No one is merging anything and I need deployment-driven data!

---

* docs: router writes to partitions (d3dfc4221)

      Though it can write to different topics, it currently writes to different
      partitions.


lukebond (approved):
stellar work!

PR #5384, opened 2022-08-12
refactor: parallelise Kafka partition client init

domodwyer:
This PR is a simple change - initialise kafka clients in parallel (concurrency=10).

---

* refactor: parallelise Kafka partition client init (bd88ac614)

      Changes the Kafka write buffer impl to parallelise initialisation of the 
      PartitionClient instances.

      Now that the PartitionClient constructor also performs leader discovery
      (using cached metadata, influxdata/rskafka#164) and establishes a broker 
      connection (influxdata/rskafka#166) executing them in parallel will cause a
      proportional decrease in the time taken to bring IOx up.

### DIFF
--- a/write_buffer/src/kafka/mod.rs
+++ b/write_buffer/src/kafka/mod.rs
@@ -14,7 +14,10 @@ use crate::{
 use async_trait::async_trait;
 use data_types::{KafkaPartition, Sequence, SequenceNumber};
 use dml::{DmlMeta, DmlOperation};
-use futures::{stream::BoxStream, StreamExt};
+use futures::{
+    stream::{self, BoxStream},
+    StreamExt, TryStreamExt,
+};
 use iox_time::{Time, TimeProvider};
 use observability_deps::tracing::warn;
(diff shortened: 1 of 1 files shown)


lukebond (approved):

write_buffer/src/kafka/mod.rs
```diff
@@ -435,13 +438,20 @@ async fn setup_topic(
...
+            let client_ref = &client;
+            return stream::iter(topic.partitions.into_iter().map(|p| {
+                let topic_name = topic_name.clone();
+                async move {
+                    let partition = u32::try_from(p).map_err(WriteBufferError::invalid_data)?;
+                    let c = client_ref.partition_client(&topic_name, p).await?;
+                    Ok((partition, c))
+                }
+            }))
+            .buffer_unordered(10)
```
alamb, line 452: such concurrency! 

I wonder if there is some reason not to up this to 20 or higher? Or more to the point why stop at 10?

PR #5385, opened 2022-08-12
feat: import schema updates iox catalog

lukebond:
this PR is the next step of the bulk ingest CLI.

before this PR it would:
- take an object storage location and prefix and fetch a bunch of TSM schema files output by an earlier stage
- merge those schemas into a single schema, and reject it if there are conflicting field types, or any fields & tags that have the same name in the same measurement
- it would report these validation errors and exit

after this PR:
- fetch the IOx schema from the catalog, deriving namespace name from org and bucket, creating the namespace if necessary
- updating the schema for the namespace by merging it with the aggregate TSM schema described above, or creating the schema accordingly if it didn't exist

i'm new to working with the catalog so please check my usage of the repos, i wasn't quite sure if i needed transactions, for e.g..


import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+    let mut repos = catalog.repositories().await;
+    let iox_schema = match get_schema_by_name(namespace_name.as_str(), repos.deref_mut()).await {
+        Ok(iox_schema) => iox_schema,
+        Err(iox_catalog::interface::Error::NamespaceNotFoundByName { .. }) => {
+            // create the namespace
+            let (topic_id, query_id) =
+                get_topic_id_and_query_id(repos.deref_mut(), kafka_topic, query_pool_name).await?;
+            let _namespace = create_namespace(
+                namespace_name.as_str(),
+                retention,
```
andrewcharlton, line 52: Do we need to grab the retention period in this case?  I don't expect we'll get here very often if we're mirroring all writes but I don't want to set it manually.

influxdb_iox/src/commands/import/schema.rs
```diff
@@ -39,6 +69,20 @@ pub struct MergeConfig {
...
+    #[clap(flatten)]
+    write_buffer_config: WriteBufferConfig,
+
+    #[clap(long, default_value = "iox_shared")]
+    /// Retention setting for any namespaces that need to be created
+    query_pool_name: String,
+
+    #[clap(long, default_value = &iox_catalog::INFINITE_RETENTION_POLICY)]
+    /// Retention setting for any namespaces that need to be created
+    retention: String,
```
andrewcharlton, line 84: See the other comment as well, but this scares me - we don't want to set the same retention period for all namespaces surely?

influxdb_iox/src/commands/import/schema.rs
```diff
@@ -39,6 +69,20 @@ pub struct MergeConfig {
...
     object_store: ObjectStoreConfig,
 
+    #[clap(flatten)]
+    catalog_dsn: CatalogDsnConfig,
+
+    #[clap(flatten)]
+    write_buffer_config: WriteBufferConfig,
+
+    #[clap(long, default_value = "iox_shared")]
+    /// Retention setting for any namespaces that need to be created
```
andrewcharlton, line 79: copy-pasta comment?

import/src/lib.rs
```diff
@@ -156,4 +168,70 @@ mod tests {
...
+        {
+          "org_id": "1234",
+          "bucket_id": "5678",
+          "measurements": {
+            "cpu": {
+              "tags": [
+                { "name": "host", "values": ["server", "desktop"] }
+              ],
+             "fields": [
+                { "name": "usage", "types": ["FloatyMcFloatFace"] }
```
andrewcharlton, line 206: :grin:

andrewcharlton (commented):
Looks reasonable to me apart from my concerns about where we source the retention periods from. 

However, I have almost no knowledge of how the catalog works so will defer to someone else to approve.

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+        let mut repos = catalog.repositories().await;
+        let iox_schema = get_schema_by_name("1234_5678", repos.deref_mut())
+            .await
+            .expect("got schema");
+        assert_eq!(iox_schema.tables.len(), 1);
+        let table = iox_schema.tables.get("cpu").expect("got table");
+        assert_eq!(table.columns.len(), 3); // one tag & one field, plus time
+        let tag = table.columns.get("host").expect("got tag");
+        assert!(tag.is_tag());
+        let field = table.columns.get("usage").expect("got field");
```
alamb, line 278: nice tests 👍

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+            // presumably it got created in the meantime?
+            repos
+                .namespaces()
+                .get_by_name(name)
+                .await
+                .map_err(UpdateCatalogError::CatalogError)?
+                .ok_or_else(|| UpdateCatalogError::NamespaceNotFound(name.to_string()))
+        }
+        Err(e) => {
+            eprintln!("Failed to create namespace");
```
alamb, line 123: maybe worth using `error!` macro to log rather than `eprintln`

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+        // check that the schema has all these columns or update accordingly.
+        for tag in measurement.tags.values() {
+            match table.columns.get(tag.name.as_str()) {
+                Some(c) if c.is_tag() => {
+                    // nothing to do, all good
+                }
+                Some(_) => {
+                    // a column that isn't a tag exists; not good
+                    return Err(UpdateCatalogError::SchemaUpdateError(format!(
+                        "a non-tag column with name {} already exists in the schema",
```
alamb, line 171: 👍

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+                        table_id: table.id,
+                        column_type: ColumnType::Tag,
+                    });
+                }
+            }
+        }
+        for field in measurement.fields.values() {
+            // we know there is only one type and it's valid
+            let influx_column_type = InfluxColumnType::Field(
+                InfluxFieldType::try_from(field.types.iter().next().unwrap()).unwrap(),
```
alamb, line 188: As written, this code will `panic!` which probably isn't a huge deal, but might not be a great user experience 

I wonder if it is worth errors in the case if the input is bad (in this case`field` has 0 or 2+ types listed) or it was invalid?

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+                }
+            }
+        }
+        if !column_batch.is_empty() {
+            // add all the new columns we have to create for this table in one batch.
+            // it would have been nice to call this once outside the loop and thus put less
+            // pressure on the catalog, but because ColumnUpsertRequest takes a slice i can't do
+            // that with short-lived loop variables.
+            // since this is a CLI tool rather than something called a lot on the write path, i
+            // figure it's okay.
```
alamb, line 217: 👍  (and further more we can optimize it later)

import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -0,0 +1,386 @@
...
+        assert!(tag.is_tag());
+        let field = table.columns.get("usage").expect("got field");
+        assert_eq!(
+            field.column_type,
+            InfluxColumnType::Field(InfluxFieldType::Float)
+        );
+    }
+
+    #[tokio::test]
+    async fn needs_merging() {
```
alamb, line 286: I think one test exercising schema conflict (a column with a name that already had an alternate type) would probably be good here. Ideally we could hit both errors (tag with existing name as well as a field with existing name)

alamb (approved):
I think the code looks solid to me. Nice work @lukebond 

I do think adding some negative tests for schema import errors would be a good idea, but otherwise I think this is ready to go

import/src/aggregate_tsm_schema/update_catalog.rs
lukebond: we chatted offline. resolution:
- don't provide a default value for retention; if namespace is to be created we want the user to be forced to specify it (for us that means looking it up in Cloud2 for the source bucket)
- fix the confusing comment on the CLI arg that suggests the tool could create multiple namespaces- it can't, we migrate for one org and bucket only, and the schema validation enforces that
- check if the namespace exists, and if so, error if they didn't provide a retention setting. make clear in the CLI arg comment that this setting will be ignored if the namespace already exists

influxdb_iox/src/commands/import/schema.rs
lukebond: captured as TODO in other comment

import/src/aggregate_tsm_schema/update_catalog.rs
lukebond: there is a bit of code after schema validation that checks for this case (`influxdb_iox/src/commands/import/schema.rs` line `146` - couldn't seem to link to LOC in the same PR!):

```rust
            // From here we can happily .unwrap() the field types knowing they're valid
            if !merged_tsm_schema.types_are_valid() {
                return Err(SchemaCommandError::InvalidFieldTypes());
            }
```

this function checks that there is only one type and it's a valid type that can be converted `.into()` the type IOx uses. that's what the comment here is referring to.

although Rust doesn't know that :)

i think it can't hurt to check it again since i have a `fn` available to do it and it's cheap, it'll give the user a nicer message. even though it shouldn't be possible unless someone removes that other code.

import/src/aggregate_tsm_schema/update_catalog.rs
lukebond: as far as i can tell, it's only the `run` command of iox that is set up with log levels, with the relevant clap blocks to configure them. all others just `println` - it may be that `eprintln` isn't the right choice even if this is true. it feels to me that this is right but you know the codebase much better!

import/src/aggregate_tsm_schema/update_catalog.rs
lukebond: the validation happens in the validation module before this update catalog module is called, and all those things are checked there, and there are tests for it. however if something did come through to here we'd still get errors i can test for, such as the catalog erroring on duplicate columns (which is how all these would show up, as a `CatalogError`).

i think instead i should just have one test that i know would trigger an error from the catalog and ensure it's propagated back here in a friendly manner; a failing tests specific to those use cases i think would confuse the user when the logic error is going to be elsewhere- in the validation module. LMK if you disagree.

lukebond: actually, you're right- i would be testing the logic written in this module as well :+1:

lukebond: and found a bug adding that test! thanks :raised_hands:

alamb: Testing for the win!

PR #5396, opened 2022-08-15
refactor: port TTL backend to policy framework

crepererum:
Note that this is "just" a port, it does NOT change how TTL works. This
will be done in #5318.

Helps with #5320.


alamb:
I plan to review this later today in a few hours

cache_system/src/backend/policy/ttl.rs
```diff
@@ -574,63 +508,20 @@ mod tests {
...
     }
 
-    #[test]
-    #[should_panic(expected = "inner backend is not empty")]
-    fn test_panic_inner_not_empty() {
-        let ttl_provider = Arc::new(TestTtlProvider::new());
-        let time_provider = Arc::new(MockProvider::new(Time::MIN));
-        let metric_registry = metric::Registry::new();
-        TtlBackend::new(
-            Box::new(HashMap::<u8, String>::from([(1, String::from("a"))])),
```
alamb, line 608: Is there no equivalent check now that the backend is empty when the `TTL policy` is added? Maybe it could be done in the callbackL

```rust
        |_callback_handle| Self {
           // validate the backend has nothing here???
... 
        }
```

cache_system/src/backend/policy/ttl.rs
```diff
@@ -661,6 +552,36 @@ mod tests {
         }
     }
 
+    struct TestState {
```
alamb, line 555: 👍

cache_system/src/backend/policy/ttl.rs
```diff
@@ -574,63 +508,20 @@ mod tests {
...
-            "my_cache",
-            &metric_registry,
-        );
-    }
-
-    #[test]
-    fn test_generic() {
-        use crate::backend::test_util::test_generic;
-
-        test_generic(|| {
```
alamb, line 620: I don't understand why we no longer test a backend with the generic battery of tests

cache_system/src/backend/policy/ttl.rs
crepererum: done

cache_system/src/backend/policy/ttl.rs
crepererum: a policy is not a backend, so this doesn't really test much, but I added it back because why not having yet another test

cache_system/src/backend/policy/mod.rs
```diff
@@ -558,6 +575,47 @@ where
     }
 }
 
+/// Read-only ref to the inner backend of [`PolicyBackend`] for debugging.
```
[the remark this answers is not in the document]
crepererum: This issue has nothing to with cache types (K and V) but with the fact that this "bypasses" the lock to the inner backend.

And regarding "just use `Any`": I had a proposal up and it was a mess. Without any concrete proposal I'm unwilling to discuss it again.

PR #5398, opened 2022-08-15
refactor: convert `remove_if` feature to policy framework

crepererum:
Decided to drop the "shared" functionality. We only use the small `remove_if` bit which is way easier to reason about.

Also gives `ChangeRequest` a lifetime to avoid cloning the predicates (or make them `'static` which would be equally bad).

For #5320.


cache_system/src/backend/policy/remove_if.rs
```diff
@@ -0,0 +1,196 @@
...
+                "cache_removed_by_custom_condition",
+                "Number of entries removed from a cache via a custom condition",
+            )
+            .recorder(&[("name", name)]);
+
+        let handle1 = RemoveIfHandle {
+            callback_handle: Arc::new(Mutex::new(None)),
+            metric_removed_by_predicate,
+        };
+        let handle2 = handle1.clone();
```
alamb, line 43:
```suggestion
        let handle_captured = handle1.clone();
```

I found myself trying to figure out why there were two handles for a moment -- perhaps a slightly different name would make it clearer the second handle is for the closure

cache_system/src/backend/policy/remove_if.rs
```diff
@@ -0,0 +1,196 @@
...
+        let handle = match guard.as_mut() {
+            Some(handle) => handle,
+            None => return false,
+        };
+
+        let metric_removed_by_predicate = self.metric_removed_by_predicate.clone();
+        let mut removed = false;
+        let removed_captured = &mut removed;
+        let k = k.clone();
+        handle.init_requests(vec![ChangeRequest::from_fn(move |backend| {
```
alamb, line 101: Do I read this right that `init_requests` is actually invoked whenever `remove_if` is called? If so the name `init_requests` may be misleading -- perhaps `push_requests` or something similar might be more specific?

cache_system/src/backend/policy/remove_if.rs
```diff
@@ -0,0 +1,196 @@
...
+        });
+    }
+
+    #[test]
+    fn test_remove_if() {
+        let metric_registry = metric::Registry::new();
+        let mut backend = PolicyBackend::new(Box::new(HashMap::<u8, String>::new()));
+        let (constructor, handle) =
+            RemoveIfPolicy::create_constructor_and_handle("my_cache", &metric_registry);
+        backend.add_policy(constructor);
```
alamb, line 145:
```suggestion
        let (policy, handle) =
            RemoveIfPolicy::create_constructor_and_handle("my_cache", &metric_registry);
        backend.add_policy(policy);
```

Which I realize isn't quite right as the `constructor` isn't a policy but a `Fn` object. But then at least it matches the name `add_policy`  -- I was just thinking that when reading this code the difference between constructor and policy likely isn't relevant to understanding how to use the code. Just a thought I had 🤷

querier/src/cache/namespace.rs
```diff
@@ -56,7 +56,7 @@ type CacheT = Box<
 #[derive(Debug)]
 pub struct NamespaceCache {
     cache: CacheT,
-    backend: SharedBackend<Arc<str>, Option<Arc<CachedNamespace>>>,
+    remove_if_handle: RemoveIfHandle<Arc<str>, Option<Arc<CachedNamespace>>>,
```
alamb, line 59: I like this better than the original idea of a "shared backend" -- an "opaque" handle that can be used to remove entries makes more sense

querier/src/cache/read_buffer.rs
```diff
@@ -44,9 +43,6 @@ type CacheT = Box<
 #[derive(Debug)]
 pub struct ReadBufferCache {
     cache: CacheT,
-
-    /// Handle that allows clearing entries for existing cache entries
```
alamb, line 48: the idea is that the `remove_if` wasn't ever used so just simplify the read buffer code in this case? If so makes sense to me

alamb (approved):
👍

querier/src/cache/read_buffer.rs
crepererum: that's right

cache_system/src/backend/policy/remove_if.rs
crepererum: Went for `policy_constructor` which I think is a good middle-ground between "what it is" and "easy to understand/read".

cache_system/src/backend/policy/remove_if.rs
crepererum: I see how the name is confusing. Renamed it to `execute_requests` which is similar to your proposal but I found "push" slightly ambiguous because it may imply some async / delayed execution which is NOT the case. As the method docs state, the control flow returns to the caller AFTER the requests and ALL follow-ups are resolved.

cache_system/src/backend/policy/remove_if.rs
crepererum: done (also renamed `handle1`).

cache_system/src/backend/policy/remove_if.rs
alamb: I think `execute_requests` sounds very nice 👌

PR #5381, opened 2022-08-12
refactor: Reduce how much LLVM IR is generated for encoders

Marwes:
Reduces how much LLVM IR rustc generates (and thereby how much work LLVM has to do) by separating the transcoder and physical parts of the encoders. 

### Before

````
  Lines          Copies     

```
 Lines          Copies        Function name
  -----          ------        -------------
  624730 (100%)  14732 (100%)  (TOTAL)
   16644 (2.7%)     19 (0.1%)  <read_buffer::column::encoding::scalar::rle::RLE<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter
   16188 (2.6%)     19 (0.1%)  <read_buffer::column::encoding::scalar::rle::RLE<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter_range
   16128 (2.6%)     18 (0.1%)  <read_buffer::column::encoding::scalar::fixed::Fixed<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter_range
   16128 (2.6%)     18 (0.1%)  <read_buffer::column::encoding::scalar::fixed_null::FixedNull<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter
   15666 (2.5%)     18 (0.1%)  <read_buffer::column::encoding::scalar::fixed_null::FixedNull<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter_range
   15660 (2.5%)     18 (0.1%)  <read_buffer::column::encoding::scalar::fixed::Fixed<P,L,T> as read_buffer::column::encoding::scalar::ScalarEncoding<L>>::row_ids_filter
   13408 (2.1%)     72 (0.5%)  read_buffer::column::encoding::scalar::fixed::Fixed<P,L,T>::row_ids_cmp_order
   10436 (1.7%)    200 (1.4%)  core::option::Option<T>::map
    8386 (1.3%)    110 (0.7%)  core::iter::traits::iterator::Iterator::fold
    7644 (1.2%)     36 (0.2%)  read_buffer::column::encoding::scalar::rle::RLE<P,L,T>::new_from_iter_opt
[...]

PR #5401, opened 2022-08-15
fix: `read_buffer` benchmarks array type creation

alamb:
# Rationale

While working on benchmarking https://github.com/influxdata/influxdb_iox/pull/5381 I was having trouble running the benchmark

```shell
cargo bench -p read_buffer
```

Resulted in errors like:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidArgumentError("column types must match schema types, expected Timestamp(Nanosecond, None) but found Int64 at column index 2")', read_buffer/benches/database.rs:114:47
stack backtrace:
```

# Changes:

Update the code to reflect the latest IOx type


read_buffer/benches/database.rs
```diff
@@ -89,21 +90,27 @@ fn generate_row_group(rows: usize) -> RecordBatch {
 
     let data: Vec<ArrayRef> = vec![
         // sorted 2 cardinality column
-        Arc::new(StringArray::from(
+        Arc::new(DictionaryArray::<Int32Type>::from_iter(
```
alamb, line 93: we switched to using dictionaries for tags a long time ago

lukebond (approved):

PR #5404, opened 2022-08-16
refactor: some prep-work for LRU cache policy

crepererum:
Prequel to the actual PR to simplify the review.

See commits.

Ref #5320.


cache_system/src/backend/policy/mod.rs
```diff
@@ -1709,4 +1751,53 @@ mod tests {
             step.action.perform(&mut self.background_task)
         }
     }
+
+    /// Same as [`ChangeRequestFn`] but implements `Send`.
+    type SendableChangeRequestFn =
+        Box<dyn for<'a> FnOnce(&'a mut dyn CacheBackend<K = String, V = usize>) + Send + 'static>;
+
+    /// Same as [`ChangeRequest`] but implements `Send`.
```
alamb, line 1759: I wonder if there is some reason that all `ChangeRequest` and `ChangeRequestFn` shouldn't implement `Send`? I assume there is a good reason as that seems to be the main change in the PR 😆 

The commit message definitely explains *what* this PR does -- I am just not 100% sure *why* you are making this change.

alamb (approved):
👍  looks good to me -- I had a question about the intent of this change, but the change itself 👍

cache_system/src/backend/policy/mod.rs
crepererum: I need this for the LRU work. The essence is: sometimes it's helpful if the `ChangeRequest` that is constructed in the background (i.e. that is used with `CallbackHandle::execute_requests`) can carry around locks to avoid certain lock gaps.

PR #5406, opened 2022-08-16
refactor: port LRU cache backend to policy framework

### LINKED ISSUE #5320: Querier Cache: Improve backend policies

crepererum:
Most headache here is careful lock-dancing. It think this is inherit to he problem that we try to solve as a whole: we have multiple caches with multiple policies each but multiple caches share the same resource pool. The good thing though is that this only needs to be "solved" once in `cache_system` and then the `querier` crate can easily use it without the need to worry about the details.

Closes #5320.


cache_system/src/backend/policy/lru.rs
```diff
@@ -730,41 +778,49 @@ where
...
+            for request in requests {
+                request.eval(backend);
+            }
+        });
+
+        self.callback_handle.execute_requests(vec![combined]);
+    }
 }
 
-impl<'a, K, V, S> Deref for LruBackendInnerBackendHandle<'a, K, V, S>
```
crepererum, line 757: This GIT diff doesn't make any sense. The `Deref` impl and the new `downcast_change_requests` function are NOT related. They just happen to share a similar `where`-clause

image

cache_system/src/backend/mod.rs
```diff
@@ -1,11 +1,8 @@
 //! Storage backends to keep and manage cached entries.
 use std::{any::Any, fmt::Debug, hash::Hash};
 
-pub mod addressable_heap;
```
crepererum, line 4: Took the liberty to move a few modules around to avoid some artificial deeply-nested crates.

cache_system/src/backend/policy/lru.rs
```diff
@@ -520,38 +571,37 @@ where
...
-            resource_estimator,
-        }
-    }
-
-    /// Get underlying / inner backend.
-    pub fn inner_backend(&self) -> LruBackendInnerBackendHandle<'_, K, V, S> {
-        LruBackendInnerBackendHandle {
-            inner: self.inner.lock(),
+        move |mut callback_handle| {
+            callback_handle.execute_requests(vec![ChangeRequest::ensure_empty()]);
```
alamb, line 576: I like the name `execute_requests` as I think it makes it quite clear what is going on

cache_system/src/backend/policy/lru.rs
```diff
@@ -700,12 +742,17 @@ trait PoolMemberGuard: Debug {
...
-    /// Remove oldest entry and return consumption of the removed entry.
+    /// Remove oldest entry and return consumption of the removed entry and an opaque [`ChangeRequest`].
+    ///
+    /// This method is used for pool members that did NOT trigger the removal.
     ///
     /// # Panic
     /// This must only be used if [`could_remove`](Self::could_remove) was used to check if there is anything to check
     /// if there is an entry that could be removed. Panics if this is not the case.
-    fn remove_oldest(&mut self) -> Self::S;
+    fn remove_oldest(&mut self) -> (Self::S, Box<dyn Any>);
```
alamb, line 752: 👍

cache_system/src/backend/policy/lru.rs
```diff
@@ -1108,39 +1183,14 @@ mod tests {
...
-        assert_inner_backend(&backend, [(String::from("a"), 1), (String::from("foo"), 3)]);
+        assert_inner_backend(
+            &mut backend,
+            [(String::from("a"), 1), (String::from("foo"), 3)],
+        );
     }
 
     #[test]
-    fn test_oversized_entries_are_never_added() {
-        #[derive(Debug)]
```
alamb, line 1116: This change is just a drive by cleanup, right? Or did the code change so that the oversized item is actually added (temporarily) to the backend prior to removal?

alamb (approved):
👍 

I wonder if it would be valuable to write some sort of test that showed the interactions between ` TTLPolicy` and a `LRUPolicy`. However, perhaps that would be best done as part of https://github.com/influxdata/influxdb_iox/issues/5318 (forgive me if you have already said you plan to do so)

cache_system/src/backend/policy/lru.rs
crepererum: > Or did the code change so that the oversized item is actually added (temporarily) to the backend prior to removal?

This. I could somewhat achieve the old semantics but that would require some form of "change log compression" within the `PolicyBackend` (i.e. it queues up changes to the underlying backend, and removes set-remove pairs) but that complexity and all the questions around semantics don't seem to be worth it. The oversized value is already in-memory prior to adding it to the `PolicyBackend` so what's the point of reducing its lifetime a bit more. This also considers that oversized values should be a rather niche issue.

We have a new test though that checks if we can deal if oversized entries at all. I've also extended this test (after your review, while writing this answer) to ensure that the semantic is "if we find an oversized entry, don't evict the entire pool but just throw that one entry away".

alamb: I agree that I can think of nothing wrong with adding the oversize entry temporarily to the backend -- the only potential issue I can see is interactions with other policies, but that is probably best left to future tests 

change log compression seems overkill I agree 👍

PR #5409, opened 2022-08-16
docs: Edit catalog API docs

pierwill:
Makes edits to the catalog API docs for clarity, formatting, consistency, and readability.


iox_catalog/src/interface.rs
```diff
@@ -162,24 +163,28 @@ pub(crate) mod sealed {
     }
 }
 
-/// transaction of a [`Catalog`].
+/// Transaction in a [`Catalog`] (similar to a database transaction).
```
pierwill, line 166: "In" makes the most sense to me, as I understand it. 

Maybe "for a catalog"? 🤔

iox_catalog/src/interface.rs
```diff
@@ -162,24 +163,28 @@ pub(crate) mod sealed {
...
 ///
-/// A transaction provides a consistent view on data and stages writes (this normally maps to a database transaction).
-/// Repositories can cheaply be borrowed from it. To finalize a transaction, call [commit](Self::commit) or [abort](Self::abort).
+/// A transaction provides a consistent view on data and stages writes.
+/// To finalize a transaction, call [commit](Self::commit) or [abort](Self::abort).
 ///
-/// Note that after any method in this transaction (including all repositories derived from it) returned an error, the
+/// Repositories can cheaply be borrowed from it.
```
pierwill, line 171: Maybe add an example here?

iox_catalog/src/interface.rs
alamb: I think `Transaction in` makes sense ( though I think `of` is probably also technically accurate)

alamb (approved):
Looks like a great set of improvements to me -- thanks @pierwill !

PR #5411, opened 2022-08-16
feat: table name in schema validation errors

### LINKED ISSUE #5408: Schema verification error reports column name but not table name: DML handler
Helps with https://github.com/influxdata/conductor/issues/1053
Helps with https://github.com/influxdata/conductor/issues/1021

__Steps to reproduce:__

Minimal actions needed to reproduce the behaviour:

1. write `bar,foo=1 v=1`
2. write `bar foo=1,v=1`

__Expected behaviour:__

```
dml handler error: schema conflict: table bar, column foo is type tag but write has type iox::column_type::field::float
```

__Actual behaviour:__

```
dml handler error: schema conflict: column foo is type tag but write has type iox::column_type::field::float
```

__Context:__

https://github.com/influxdata/influxdb_iox/issues/5407 is a similar issue, but in a different layer of the write codepath.

domodwyer:
Changes errors like this:

```
dml handler error: schema conflict: column foo is type tag but write has type iox::column_type::field::float
```

Into:

```
dml handler error: schema conflict: table bar, column foo is type tag but write has type iox::column_type::field::float
```

Closes #5408, closes #5410.

---

* feat: table name in schema validation errors (76c0c3373)

      Scopes all schema validation errors to include the table name in the error
      output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2236,6 +2236,7 @@ dependencies = [
  "sqlx-hotswap-pool",
  "tempfile",
  "test_helpers",
+ "thiserror",
  "tokio",
  "uuid 1.1.2",
  "workspace-hack",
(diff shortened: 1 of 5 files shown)


mkmik (approved):
Nice

PR #5415, opened 2022-08-17
feat: Add influxdb_influxql_parser crate

### LINKED ISSUE #5412: InfluxQL: Create InfluxQL parser crate
In order for IOx to process InfluxQL queries, we must teach IOx how to parse InfluxQL queries. The primordial step is to add a new crate to the IOx repository to serve as the foundation. The suggested name is `influxdb_influxql_parser`.

- [x] Add a new crate to the IOx workspace
- [x] Add a quoted identifier parser to provide some initial code to review and structure the crate
- [x] Add parser for all known InfluxQL keywords

stuartcarnie:
Closes #5412

Teach IOx about a new crate, `influxdb_influxql_parser`, which will evolve to parse a subset of InfluxQL. This PR contains 

* a skeleton crate for the InfluxQL parser;
* quoted identifier parser; and 
* an InfluxQL keywords parser

Both the parsers are expected to be combined with other parsers in time.


influxdb_influxql_parser/src/identifier.rs
```diff
@@ -0,0 +1,134 @@
...
+//! * double quoted identifiers can contain escaped characters, namely `\"`, `\n`, `\t`, `\\` and `\'`
+//! * double quoted identifiers can contain [InfluxQL keywords][keywords]
+//! * unquoted identifiers must start with an upper or lowercase ASCII character or `_`
+//! * unquoted identifiers may contain only ASCII letters, decimal digits, and `_`
+//! * unquoted identifiers must start with either an ASCII letter or `_`
+//!
+//! [identifier]: https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#identifiers
+//! [keywords]: https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#keywords
+
+#![allow(dead_code)]
```
stuartcarnie, line 15: This will be removed once I start tying the functions together

influxdb_influxql_parser/src/identifier.rs
```diff
@@ -0,0 +1,132 @@
...
```
stuartcarnie, line 15: This should be removed once the module is used.

influxdb_influxql_parser/src/keywords.rs
```diff
@@ -0,0 +1,233 @@
+//! # Parse InfluxQL [keywords]
+//!
+//! [keywords]: https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#keywords
+
+#![allow(dead_code)]
```
stuartcarnie, line 5: This should be removed once the module is used.

influxdb_influxql_parser/src/identifier.rs
stuartcarnie, line 15: This should be removed once the module is used.

influxdb_influxql_parser/src/keywords.rs
stuartcarnie, line 5: This should be removed once the module is used.

influxdb_influxql_parser/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+//! # Parse a subset of [InfluxQL]
+//!
+//! [InfluxQL]: https://docs.influxdata.com/influxdb/v1.8/query_language
+
```
crepererum, line 4: Rust doesn't allow us to set lints for the entire workspace (yet), so every crate has to do this manually. Over time we established a somewhat "stricter than default" set of lints:

```suggestion
#![deny(rustdoc::broken_intra_doc_links, rust_2018_idioms)]
#![warn(
    missing_copy_implementations,
    missing_docs,
    clippy::explicit_iter_loop,
    clippy::future_not_send,
    clippy::use_self,
    clippy::clone_on_ref_ptr
)]
```

We should probably write some meta-check at some point to ensure that all our crates use the same set of lints.

influxdb_influxql_parser/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "influxdb_influxql_parser"
+version = "0.1.0"
+authors = ["Stuart Carnie <stuart.carnie@gmail.com>"]
```
crepererum, line 4: Technically the "authors" fields is not longer required (and IMHO a bit pointless in the IOx monorepo setting).

influxdb_influxql_parser/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition = "2021"
+
+[lib]
+crate-type = ["rlib", "staticlib"]
```
crepererum, line 8: This sections is also technically not required EXCEPT when you plan to use the parser from C/Go/Python/whatever and hence need a static "normal" library as well. Is this the plan?

influxdb_influxql_parser/src/keywords.rs
```diff
@@ -0,0 +1,233 @@
+//! # Parse InfluxQL [keywords]
```
crepererum, line 1: The methods in this module could use some docs I guess.

influxdb_influxql_parser/src/identifier.rs
```diff
@@ -0,0 +1,132 @@
+//! # Parse an InfluxQL [identifier]
+//!
+//! Identifiers are parsed using the following rules:
+//!
+//! * double quoted identifiers can contain any unicode character other than a new line
+//! * double quoted identifiers can contain escaped characters, namely `\"`, `\n`, `\t`, `\\` and `\'`
+//! * double quoted identifiers can contain [InfluxQL keywords][keywords]
+//! * unquoted identifiers must start with an upper or lowercase ASCII character or `_`
```
mkmik, line 8: Isn't this line covered by the last ("must start with a an ASCII letter")?

influxdb_influxql_parser/src/keywords.rs
```diff
@@ -0,0 +1,233 @@
...
+    ))(i)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use test_helpers::Result;
+
+    #[test]
+    fn test_keywords() -> Result<()> {
```
crepererum, line 141: We've established a while ago (you couldn't know that) that test do NOT return `Result` and that we just use `.unwrap()` in tests. The reason is rather simple: when a test fails you get a backtrace pointing exactly at the failing bit which -- esp. for these test collections like this one -- spare you a lot of debugging.

crepererum (commented):
Good start :muscle: 

Just for my understanding: Is the plan now to hand-roll our own parser (I'm totally fine with that) and NOT hack around the upstream sqlparser to shoehorn it into InfluxQL?

influxdb_influxql_parser/src/keywords.rs
carols10cents: > We've established a while ago (you couldn't know that)

We *did* document that decision in the style guide at least ;) https://github.com/influxdata/influxdb_iox/blob/f4443f0b3abeb9eb4bcf3f967383d5afd52d0c88/docs/style_guide.md#dont-return-result-from-test-functions

influxdb_influxql_parser/src/identifier.rs
stuartcarnie: Yes, weird that I duplicated it – grabbed it from docs, and it isn't there 🤔

influxdb_influxql_parser/Cargo.toml
stuartcarnie: Not planning on that anytime soon, so removing 👍🏻   

(I copied the `Cargo.toml` from the line protocol parser crate, hence this extraneous stuff)

influxdb_influxql_parser/src/keywords.rs
stuartcarnie: TIL: We have a style guide! 

Thanks @carols10cents 🙇🏻

stuartcarnie: Thanks @crepererum – makes sense too, as I had seen unhelpful output in DataFusion, where that pattern appears frequently also.

Glad to learn we have a style guide!

influxdb_influxql_parser/src/keywords.rs
stuartcarnie: > Just for my understanding: Is the plan now to hand-roll our own parser (I'm totally fine with that) and NOT hack around the upstream sqlparser to shoehorn it into InfluxQL?

That is correct - we discussed this at length in Slack and concluded

* there are many idiosyncrasies to InfluxQL that don't make sense in the `sqlparser` crate;
* using `sqlparser` would result in parsing a superset of SQL for InfluxQL, which is not desired.

stuartcarnie:
Thanks for the review @crepererum, et al. I have completed the PR feedback changes.

stuartcarnie:
I had to force push, as I used the `commit suggestion` feature and was promptly reminded that these do not use semantic commit messages 🤦🏻

crepererum (approved):

PR #5417, opened 2022-08-17
feat: async sleep with `TimeProvider`

crepererum:
This is helpful to mock "ticked" loops or to control certain async time
periods.

Will be used to test the refresh policy developed in #5318.


iox_time/src/lib.rs
```diff
@@ -254,8 +338,34 @@ mod test {
         assert!(b <= c);
     }
 
+    #[tokio::test]
+    async fn test_system_provider_sleep() {
+        let provider = SystemProvider::new();
+
+        let a = provider.now();
+        provider.sleep(Duration::from_secs(1)).await;
```
alamb, line 346: A sleep for 1 second shouldn't noticeably impact the time to run the tests

iox_time/src/lib.rs
```diff
@@ -265,6 +375,142 @@ mod test {
...
+        let fut = provider.sleep(Duration::from_millis(100));
+        let handle = tokio::task::spawn(async move {
+            fut.await;
+        });
+
+        // does not finish immediately
+        tokio::time::sleep(Duration::from_millis(200)).await;
+        assert!(!handle.is_finished());
+
+        // does not finish when time goes backwards
```
alamb, line 423: these are an awesome set of tests

iox_time/src/lib.rs
```diff
@@ -179,6 +182,12 @@ impl Time {
 pub trait TimeProvider: std::fmt::Debug + Send + Sync + 'static {
     /// Returns the current `Time`. No guarantees are made about monotonicity
     fn now(&self) -> Time;
+
+    /// Sleep for the given duration.
+    fn sleep(&self, d: Duration) -> Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
```
alamb, line 187: I wonder if we could provide a default implementation here

```suggestion
    fn sleep(&self, d: Duration) -> Pin<Box<dyn Future<Output = ()> + Send + 'static>> {
        self.sleep_until(self.now() + d)
     }
```

and we might be able to cut down the code in this PR by half 🤔

alamb (approved):
This looks good to me but i wonder if @tustvold  has some time to review the futures fanciness.

iox_time/src/lib.rs
crepererum: I've applied your suggestion but left the tests because I thought it easier to have both functions tested than to reason about "which subset / semantic do I need to test under which circumstances".

alamb (approved):

PR #5419, opened 2022-08-17
chore: Update to (almost) released object_store 0.4.0

alamb:
~Draft until:~
- [x] https://github.com/influxdata/influxdb_iox/pull/5418

# Rationale
https://github.com/influxdata/influxdb_iox/pull/5342 updated to a pre-release on @tustvold's branch

Then 0.4.0 was released but doesn't include the slimmed down S3 impl in https://github.com/apache/arrow-rs/pull/2352

So we are still using a branch from @tustvold 's repo rather than the released version or the one from apache

# Changes
1. Update object_store to `0.4.0`
1. Update pin to arrow-rs to https://github.com/apache/arrow-rs/commit/3f0e12d8d362752181c75836d25d424862acc424, right after https://github.com/apache/arrow-rs/pull/2352 was merged


Cargo.lock
```diff
@@ -207,6 +207,26 @@ version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 
+[[package]]
```
alamb, line 210: I think the azure upgrade brings in a bunch more deps :(

domodwyer (approved):

PR #5420, opened 2022-08-17
feat: import schema override

lukebond:
Closes https://github.com/influxdata/conductor/issues/1115

Part of wider work towards bulk ingest and TSM migration.

Previous PRs (#5344, #5367 and #5385) introduced the schema merge command to merge a collection of TSM schemas pulled from object storage, report validation errors and, if okay, update the catalog with this schema.

This PR allows the user to provide an "override schema" which enables them to resolve one type of schema validation error: multiple types existing for the same field across the collection of TSM schemas. It allows you to tell the tool which type to use for that measurement always. It is expected that this override schema will also be used by the parquet file writing tool to coerce the types. The override happens after the merge and before the validation, and the tool will still stop if the validation finds any further failures after the override was incorporated.

You can see an example of the override schema in the linked-to issue that this PR closes, and numerous examples in the tests of the expected results.


import/src/lib.rs
```diff
@@ -115,7 +174,7 @@ mod tests {
               "tags": [
                 { "name": "host", "values": ["server", "desktop"] }
               ],
-             "fields": [
+              "fields": [
```
lukebond, line 177: whoops found a bunch of misalignments that don't get cleaned up because they're in a `r#`

import/src/aggregate_tsm_schema/merge.rs
```diff
@@ -43,11 +53,42 @@ impl SchemaMerger {
...
+
+fn do_schema_override(
+    merged_schema: &mut AggregateTSMSchema,
+    override_schema: &AggregateTSMSchemaOverride,
+) {
+    for (measurement_name, override_measurement) in &override_schema.measurements {
+        // if the override refers to a measurement not in the schema it will be ignored
+        if let Some(merged_measurement) = merged_schema.measurements.get_mut(measurement_name) {
+            // we only support overrides for field types at this point. later we may support
+            // resolving tags/fields with the same name somehow
```
alamb, line 82: 👍

import/src/aggregate_tsm_schema/merge.rs
```diff
@@ -619,4 +660,224 @@ mod tests {
...
+        {
+          "org_id": "myorg",
+          "bucket_id": "mybucket",
+          "measurements": {
+            "cpu": {
+              "tags": [
+                { "name": "host", "values": ["server", "desktop", "gadget"] }
+              ],
+             "fields": [
+                { "name": "usage", "types": ["Float"] }
```
alamb, line 797: 👍  that both `temperature` and `usage` (lower) are `Float` 👍

import/src/aggregate_tsm_schema/merge.rs
```diff
@@ -619,4 +660,224 @@ mod tests {
...
+        assert_eq!(
+            merged_schema
+                .measurements
+                .get("weather")
+                .unwrap()
+                .fields
+                .get("temperature")
+                .unwrap()
+                .types,
+            HashSet::from(["Integer".to_string()])
```
alamb, line 880: 👍

import/src/lib.rs
alamb: I too am a member of the cult of OCD whitespace 👍

alamb (approved):
I believe the code in this PR does what it says it does. Nice work

> It is expected that this override schema will also be used by the parquet file writing tool to coerce the types. 

This may not be possible in all cases. For example,  if you have both `strings` and `floats`, and the user picks `float` as the override type, there may well be string values like `'foo'` that can't be parsed as Floats

Not that this is any sort of deal breaker, I just wanted to mention it as sorting out what to do in the error cases may be tricky

lukebond:
>This may not be possible in all cases. For example, if you have both strings and floats, and the user picks float as the override type, there may well be string values like 'foo' that can't be parsed as Floats

yes true this is by no means the entire solution to this tricky problem!

PR #5422, opened 2022-08-17
chore: Remove object_store patch

carols10cents:
I noticed in https://github.com/influxdata/influxdb_iox/pull/5418 there were two different versions of object_store in Cargo.lock, which alerted me to the object_store 0.4.0 release being done, so I've removed the `patch` that IOx has been using! This gets rid of *some* dependencies at least?


tustvold:
The 0.4 release doesn't include the rusoto removal...

carols10cents:
> The 0.4 release doesn't include the rusoto removal...

Whoops, nevermind then. I should have checked!

> I'm not sure why we have two object_store versions though, I thought patch was recursive...

It is recursive, but it'll only patch compatible versions: #5418 upgraded datafusion to object_store 0.4 but the IOx patch was still pointing to an 0.3 object_store version, and 0.3 isn't considered compatible with 0.4.

It looks like [this commit updating the patch](https://github.com/influxdata/influxdb_iox/commit/7f0ae53d6fc0b73797ec038f21ed07963fc3856e#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eL672) fixed that problem, so this PR is moot!

alamb:
I believe the reason so many more dependencies came in is because of the azure sdk update, FWIW. 

Eventually hopefully we can get rid of that as well (or maybe we should just stop supporting it in IOx builds as we never deploy to azure anyways??)

tustvold:
> Eventually hopefully we can get rid of that as well

@roeap is already on it :1st_place_medal: - https://github.com/apache/arrow-rs/pull/2509

PR #5423, opened 2022-08-17
docs: Edit API docs for `data_types`

pierwill:
Makes various edits to the `data_types` API docs for clarity, consistence, formatting, and readability.


data_types/src/lib.rs
```diff
@@ -377,32 +377,32 @@ impl std::fmt::Display for ParquetFileId {
     }
 }
 
-/// Data object for a kafka topic
+/// Data object for a kafka topic.
```
pierwill, line 380: Several types in this crate are documented as "Data object for a..." I think higher-level descriptions of these concepts would be better here. For example:

```suggestion
/// A Kafka topic. Kafka topics are used to...
```

pierwill:
I should make these easier to review by putting the most nit picking, repetitive changes in a separate commit.

PR #5427, opened 2022-08-18
feat: simple RecordAggregator for write buffer

domodwyer:
Switches the Kafka write buffer aggregation layer from the existing [`DmlAggregator`](https://github.com/influxdata/influxdb_iox/blob/d75df2b6104c518638ae4810504423e61adfbc1c/write_buffer/src/kafka/aggregator.rs) (which attempts to merge together DML ops to produce one DML op) to a simple implementation that delegates batching to rskafka's [`RecordAggregator`](https://docs.rs/rskafka/0.3.0/rskafka/client/producer/aggregator/struct.RecordAggregator.html), at the Kafka record level.

https://github.com/influxdata/influxdb_iox/issues/5323 suggests the current aggregation strategy is largely ineffective, and this change is an experiment to confirm if removing the DML merging has any impact. I expect little to no change in behaviour w.r.t IOx/Kafka interactions - messages sizes will remain small, and message throughput shouldn't change in any significant way. There may be a decrease in overall latency due to the reduced overhead of aggregating.

**If Kafka falls over, or something breaks, feel free to revert this PR.**

_**NOTE:** as this is an experiment (and will be replaced no matter the outcome) I did not connect up the tracing spans. Nor did I reimplement the batch size metric, as it has little meaning in this implementation._

---

* feat: simple RecordAggregator for write buffer (30e23f6e8)

      This commit adds a new write buffer aggregator used by rskafka to increase the
      size of Kafka messages on the wire. The Kafka write buffer impl is the only
      impl to perform aggregation.

      This Aggregator impl maps IOx-specific DML operations to rskafka Records with
      no additional processing - it can be thought of as an IOx-specific adaptor
      over rskafka's RecordAggregator.

      By delegating batching of Record instances to rskakfa's simple 
      RecordAggregator, we minimise code complexity / bug surface area / LoC.

* refactor: use RecordAggregator (59c2d84d1)

      Replaces the DmlAggregator with the simpler RecordAggregator.

      Metrics gathered as part of #5323 shows there is practically no benefit to the
      additional complexity of the DmlAggregator over the simpler RecordAggregator
      impl.

* refactor: remove redundant timestamp conversions (37727105b)

      Removes the existing, copy-pasted timestamp conversion code to remove 
      redundant conversions.

* refactor: use dyn TimeProvider in RecordAggregator (a66d16576)

      For ease of integration with the existing tests, use dyn TimeProvider in the
      RecordAggregator.

* refactor: assert writes partitioned (312def5ac)

      The previous aggregator impl would assert that writes had been partitioned
      before aggregating them (or rather, that the DML write had a partition key
      assigned).

      This should be true for all writes passing through the write buffer, 
      irrespective of which aggregator is used, therefore this assert is moved
      "up" into the write buffer itself.

* test: reduce timestamp precision for comparisons (6d6fc9a08)

      Reduce the precision of timestamps in tests before comparing the DML metadata
      objects.

      This allows tests to accept different timestamp precisions, such as when ops
      pass "through" Kafka vs. files, etc.

* test: update batching tests for new aggregator (8b054c14a)

      Previously aggregated writes were merged into a single Kafka Record - this
      meant that all merged ops would be placed into the same Record, and therefore
      receive the same sequence number once published to Kafka.

      The new aggregator batches at the Record level, therefore aggregated writes
      now get their own distinct sequence number. This commit updates the batching
      tests to reflect this new sequence number assignment behaviour.


domodwyer:
I will fix the tests tomorrow - code is fine for review :+1:

alamb:
I plan to review this PR in the next hour or two

alamb:
Rather than reviewing the PR with test failures (I often use the fact that all tests pass and review the changes to the tests carefully as part of my review), I figure I will get some hands on experience with this code and try and fix the tests first -- stay tuned...

write_buffer/src/kafka/record_aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+        let record = Record {
+            key: None,
+            value: Some(buf),
+            headers: headers
+                .headers()
+                .map(|(k, v)| (k.to_owned(), v.as_bytes().to_vec()))
+                .collect(),
+            timestamp: rskafka::time::OffsetDateTime::from_unix_timestamp_nanos(
+                timestamp_millis as i128 * MS_NANOSECONDS as i128,
+            )?,
```
alamb, line 114: I don't understand the rationale for truncating the timestamp to millisecond resolution above and then extending back to nanos here.

Why not let rskafka do the truncation internally? I feel like I am probably missing something

domodwyer:
I'm having some trouble running the tests at all locally, which has slowed me down! It also happens on `main` though, so not PR specific.

The `test_batching_respects_partitioning` and `test_enqueue_no_partition_key` tests can be removed entirely with this impl :+1: I will look at the others when I get these tests working locally if you don't beat me to it!

alamb:
Actually, now I see the errors are related to actual kafka -- I'll avoid messing with them and just review the code 😆

write_buffer/src/kafka/record_aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+        DmlOperation::Write(DmlWrite::new(
+            NAMESPACE.to_string(),
+            m,
+            Some("1970-01-01".into()),
+            DmlMeta::unsequenced(Some(span)),
+        ))
+    }
+
+    #[test]
+    fn test_record_aggregate() {
```
alamb, line 262: I may have missed it, but I don't see any test for aggregating / deaggregating multiple `DmlOperation`s -- I think that is an important scenario to cover as it appears to be the key functionality for this code

alamb (commented):
The basic framework looks good to me -- I think there may be some additional testing warranted in addition to fixing the kafka tests

write_buffer/src/kafka/record_aggregator.rs
domodwyer: I dunno, I copy-pasted the [original code](https://github.com/influxdata/influxdb_iox/pull/5427/files#diff-985f95367bf1c6bba9cdf2917cafc759d69d9dfeadaf0f80eec2414ae3cc4dc5L525-L547) and removed some redundant bits / tidied it up. Clearly needed a bit more tidying! 🧹

write_buffer/src/kafka/record_aggregator.rs
domodwyer: The tests are confusing... There's actually many tests that cover batching (and a host of other bits) that are actually called as funcs from [here](https://github.com/influxdata/influxdb_iox/blob/59c2d84d1e0f54fb4e90d09deb8eb5941466dcc4/write_buffer/src/kafka/mod.rs#L588). For example, [this one](https://github.com/influxdata/influxdb_iox/blob/59c2d84d1e0f54fb4e90d09deb8eb5941466dcc4/write_buffer/src/kafka/mod.rs#L677-L828) explicitly drives (de)aggregation.

Additionally the actual Record aggregation bits are delegated to rskafka which should have its own tests that also cover this - I'm looking to make sure the `Dml -> Record` adaptor layer works in the tests in this file :+1:

domodwyer:
I've made some progress on the tests (after banging my head against https://github.com/influxdata/rskafka/issues/169) but will pick this up on Monday.

write_buffer/src/kafka/record_aggregator.rs
domodwyer: I have reduced the amount of timestamp jiggery: https://github.com/influxdata/influxdb_iox/commit/a66d16576d952a87e13d3d0356d40a47ab807673

domodwyer:
Ok - tests are good! This should be ready for a merge whenever :+1:

crepererum (approved):

PR #5430, opened 2022-08-19
feat: Teach IOx to parse `SHOW MEASUREMENTS` statement

### LINKED ISSUE #5416: InfluxQL: Implement `SHOW MEASUREMENTS`
Teach the parser how to recognise the [`SHOW MEASUREMENTS`](https://docs.influxdata.com/influxdb/v1.8/query_language/explore-schema/#show-measurements) statement.

The output of the `show_measurements` parser function should be a structure that captures similar syntactic elements as the Go implementation[^1]:

```go
// ShowMeasurementsStatement represents a command for listing measurements.
type ShowMeasurementsStatement struct {
	// Database to query. If blank, use the default database.
	Database string

	// Retention policy to query. If blank, use all retention policies (do not use default)
	RetentionPolicy string

	WildcardDatabase        bool
	WildcardRetentionPolicy bool

	// Measurement name or regex.
	Source Source

	// An expression evaluated on data point.
	Condition Expr

	// Fields to sort results by
	SortFields SortFields

	// Maximum number of rows to be returned.
	// Unlimited if zero.
	Limit int

	// Returns rows starting at an offset from the first row.
	Offset int
}
```

[^1]: https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/ast.go#L2573-L2599

stuartcarnie:
Closes #5416

This PR teaches the `influxdb_influxql_parser` crate how to parse a complete `SHOW MEASUREMENTS` statement. According to the documentation, [`SHOW MEASUREMENTS`](https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#show-measurements) is defined approximately as the following EBNF:

```ebnf
show_measurements_stmt ::= "SHOW" "MEASUREMENTS" 
  on_clause? 
  with_measurement_clause? 
  where_clause? 
  limit_clause?
  offset_clause?

on_clause ::= "ON" db_name
db_name   ::= 
  IDENTIFIER                | // database
  IDENTIFIER "." IDENTIFIER | // database.retention policy
  "*"                       | // all databases
  "*.*"                       // all databases and retention policies

with_measurement ::= "WITH" "MEASUREMENT" ( "=" measurement_name | "=~" REGEX )
measurement_name ::= ( 
  IDENTIFIER "." IDENTIFIER? "." IDENTIFIER | // db.rp?.name
  IDENTIFIER                                  // name 
  )
  
where_clause  ::= "WHERE"  conditional_expr
limit_clause  ::= "LIMIT"  UNSIGNED_INTEGER
offset_clause ::= "OFFSET" UNSIGNED_INTEGER
```

**Notable Changes**

The Go implementation parses a limited `ORDER BY` clause, defined by the following EBNF:

```ebnf
order_by   ::= "ORDER" "BY" (time_order | order)
order      ::= "ASC | "DESC
time_order ::= "TIME" order?
```

Resulting in the following valid strings:

```sql
ORDER BY ASC
ORDER BY DESC
ORDER BY time
ORDER BY time ASC
ORDER BY time DESC
```

However, this clause is ignored by both the OSS and Cloud 2 implementations of the InfluxQL `SHOW MEASUREMENTS` statement. An analysis of 4.5 million queries from Cloud 2 production showed no evidence of this undocumented clause is present, so it is not recognised by this parser and will result in a parse failure if present.


influxdb_influxql_parser/src/common.rs
```diff
@@ -0,0 +1,302 @@
...
+        ),
+        alt((
+            // "ASC" | "DESC"
+            order(),
+            // "TIME" ( "ASC" | "DESC" )?
+            map(
+                preceded(
+                    verify(preceded(multispace1, unquoted_identifier), |id: &str| {
+                        id.to_lowercase() == "time"
+                    }),
```
crepererum, line 172: Isn't this the same as `tag_no_case("TIME")`?

crepererum (approved):

influxdb_influxql_parser/src/common.rs
stuartcarnie: Yes, absolutely right – thank you! This was a legacy of initially thinking I needed to parse any identifier in an `ORDER BY`, but later finding the implementation only accepts `time`. I'll fix that up before merging.

PR #5431, opened 2022-08-19
feat: refresh policy for caches

crepererum:
For #5318 we want to have a policy that refreshes keys before they are
too old. I initially tried to fold both TTL and the refresh system into
a single policy but than decided that they will basically be two
policies in one with a harder-to-test interface. Semantically TTL and
refresh are also a bit different (but will usually be used together):

- **TTL:** Prevents that a users gets data that is too old. It is some kind
  of "soft correctness". In some sense this is related to the "remove
  if" policy where some part of the system knows for sure (or with
  reasonable likelyhood) that a cache entry is outdated. Note that TTL's
  primary job is NOT to clean up memory from old keys (even though it
  indirectly does that). There is no reason cached entries should be
  removed except for correctness (TTL and remove-if) or resource
  pressure -- and the latter is handled by the LRU policy.
- **Refresh:** While TTL is some kind of deadline, we often have good
  reason to refresh the key before we pull the plug, namely when an
  entry is used and a bit old (but not too old). The concrete mechanism
  to archive this is flexible. At the moment the policy is rather
  simple -- just start a refresh task if a key is old and we receive a
  GET request -- but can be extended in the future.

This also adds some integration tests for TTL+refresh. There will be
follow-up changes to test the interaction with LRU as well, although I am
pretty certain that there won't be any surprises due to the excessive
testing we have in place for the policy backend itself as well as all
the policies.

This change also does NOT integrate the refresh with the querier for the
sake of keeping the changeset "small" (i.e. it is already rather large).


cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+}
+
+impl<K, V> RefreshPolicy<K, V>
+where
+    K: Clone + Eq + Debug + Hash + Ord + Send + 'static,
+    V: Clone + Debug + Send + 'static,
+{
+    /// Create new refresh policy.
+    ///
+    /// Even though this does NOT perform any IO, we require this to be `async` so that we can start a tokio background process.
```
alamb, line 143: I thought it was possible to start a new background `tokio::task` from a normal (non `async` function) -- you can't `await` on it of course 

I actually changed this to be sync locally and it *seems* to have worked 🤷 

```diff
diff --git a/cache_system/src/backend/policy/integration_tests.rs b/cache_system/src/backend/policy/integration_tests.rs
index ac9403fcb..e601f2b8f 100644
--- a/cache_system/src/backend/policy/integration_tests.rs
+++ b/cache_system/src/backend/policy/integration_tests.rs
@@ -147,7 +147,6 @@ impl TestStateTtlAndRefresh {
                 &metric_registry,
                 Arc::clone(&notify_idle),
             )
-            .await,
         );
         backend.add_policy(TtlPolicy::new(
             Arc::clone(&ttl_provider) as _,
diff --git a/cache_system/src/backend/policy/refresh.rs b/cache_system/src/backend/policy/refresh.rs
index 2a7ac66ac..033a52332 100644
--- a/cache_system/src/backend/policy/refresh.rs
+++ b/cache_system/src/backend/policy/refresh.rs
@@ -142,7 +142,7 @@ where
     ///
     /// Even though this does NOT perform any IO, we require this to be `async` so that we can start a tokio background process.
     #[allow(clippy::new_ret_no_self)]
-    pub async fn new(
+    pub fn new(
         refresh_duration_provider: Arc<dyn RefreshDurationProvider<K = K, V = V>>,
         time_provider: Arc<dyn TimeProvider>,
         loader: Arc<dyn Loader<K = K, V = V, Extra = ()>>,
@@ -158,11 +158,10 @@ where
             metric_registry,
             idle_notify,
         )
-        .await
     }
 
     #[allow(clippy::new_ret_no_self)]
-    pub(crate) async fn new_inner(
+    pub(crate) fn new_inner(
         refresh_duration_provider: Arc<dyn RefreshDurationProvider<K = K, V = V>>,
         time_provider: Arc<dyn TimeProvider>,
         loader: Arc<dyn Loader<K = K, V = V, Extra = ()>>,
@@ -677,8 +676,7 @@ mod tests {
             loader,
             "my_cache",
             &metric_registry,
-        )
-        .await;
+        );
         backend.add_policy(|mut handle| {
             handle.execute_requests(vec![ChangeRequest::set(1, String::from("foo"))]);
             policy_constructor(handle)
@@ -702,7 +700,6 @@ mod tests {
                 "my_cache",
                 &metric_registry,
             )
-            .await,
         );
 
         backend.set(1, String::from("a"));
@@ -921,13 +918,13 @@ mod tests {
                 let mut backend = PolicyBackend::new(Box::new(HashMap::<u8, String>::new()));
 
                 let policy_constructor =
-                    tokio::runtime::Handle::current().block_on(RefreshPolicy::new(
+                    tokio::runtime::Handle::current().block_on(async move {RefreshPolicy::new(
                         Arc::clone(&ttl_provider) as _,
                         Arc::clone(&time_provider) as _,
                         loader,
                         "my_cache",
                         &metric_registry,
-                    ));
+                    )});
                 backend.add_policy(policy_constructor);
                 backend
             });
@@ -961,7 +958,6 @@ mod tests {
                     &metric_registry,
                     Arc::clone(&notify_idle),
                 )
-                .await,
             );
 
             Self {
```

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+                // the loop. Even worse, since `FuturesUnodered` is not hooked up into tokio's (somewhat bizarre)
+                // task preemtion system, tokio will poll this method here forever, essentially blocking this
+                // thread.
+                refresh_tasks.push(Box::pin(futures::future::pending()));
+
+                // flag that remembers if we can notify idle observers again
+                let mut can_notify_idle = true;
+
+                loop {
+                    // future that "guards" our idle notification to prevent hot loops (essentially blocking the entire
```
alamb, line 197: I don't fully understand how a hot loop would arise: I would have thought that polling for the next refresh task that was complete or a new one was received from `rx_refresh_tasks` would be sufficient -- and it would wait until something got pushed into `rx_refresh_tasks`

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+
+        let cancelled_captured = cancelled.clone();
+        let fut = async move {
+            // some `let`-dance so that rustc does not complain that `&K` is not `Send`
+            let k_for_loader = k.clone();
+            let v = loader.load(k_for_loader, ()).await;
+
+            let mut callback_handle = callback_handle.lock();
+            callback_handle.execute_requests(vec![ChangeRequest::from_fn(|backend| {
+                // Here we have the PolicyBackend implicit lock. There is now way our Subscriber can be
```
alamb, line 260:
```suggestion
                // Here we have the PolicyBackend implicit lock. There is no way our Subscriber can be
```

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+                // active here, but we need to check if we have been canceled one last time.
```
alamb, line 261: this is the only check on the cancellation token that I saw (the comment implied to me there was some previous check)

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+    type K = K;
+    type V = V;
+
+    fn get(&mut self, k: &Self::K) -> Vec<ChangeRequest<'static, Self::K, Self::V>> {
+        let mut timings = self.timings.lock();
+
+        // Does this entry exists and is there no a refresh operation running?
+        if let Some(RefreshState {
+            t,
+            running_refresh: running_refresh @ None,
```
alamb, line 316: This is a fancy syntax -- 👍

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+                t,
+                running_refresh: None,
+            };
+
+            timings.insert(k, state);
+        } else {
+            // need to remove potentially existing entry that had some refresh set
+            timings.remove(&k);
+
+            // the removal automatically triggered a cancelation for any potentially running refresh operation
```
alamb, line 348:
```suggestion
            // the removal drops the RefreshState which triggers a cancelation for any potentially running 
            // refresh operation
```

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+
+        // Does this entry exists and is there no a refresh operation running?
+        if let Some(RefreshState {
+            t,
+            running_refresh: running_refresh @ None,
+        }) = timings.get_mut(k)
+        {
+            let now = self.time_provider.now();
+
+            // Is it time to refresh?
```
alamb, line 321: Is it correct that this is where the "only refresh keys that were actually used" policy is implemented -- if `get()` is never called for that key then `refresh()` will never be called?

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+
+    #[derive(Debug)]
+    struct TestLoaderResponse {
+        v: String,
+        block: Option<Arc<Barrier>>,
+    }
+
+    /// An easy-to-mock [`Loader`].
+    #[derive(Debug, Default)]
+    pub struct TestLoader {
```
alamb, line 434: 👌

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+        #[tokio::test]
+        #[should_panic(expected = "no mocked response left")]
+        async fn test_loader_panic_no_mocked_reponse_left() {
+            let loader = TestLoader::default();
+            loader.mock_next(1, String::from("foo"));
+            loader.load(1, ()).await;
+            loader.load(1, ()).await;
+        }
+
+        #[test]
```
alamb, line 578: tests for the tests -- very nice

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+            time_provider,
+            loader,
+            metric_registry,
+            notify_idle,
+            ..
+        } = TestState::new().await;
+
+        loader.mock_next(1, String::from("foo"));
+        loader.mock_next(1, String::from("bar"));
+        refresh_duration_provider.set_refresh_in(
```
alamb, line 729: The API for using this policy is quite nice

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+        // perform refresh
+        time_provider.inc(Duration::from_secs(1));
+        assert_eq!(backend.get(&1), Some(String::from("a")));
+        notify_idle.notified_with_timeout().await;
+        assert_eq!(backend.get(&1), Some(String::from("a")));
+
+        backend.set(1, String::from("b"));
+        barrier.wait().await;
+
+        // no refresh
```
alamb, line 903: how does this test verify that no refresh happened? Is it because the `refresh_duration_provider` would panic if it tried to?

cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    // perform refresh
+    time_provider.inc(Duration::from_secs(1));
+    assert_eq!(backend.get(&1), Some(String::from("a")));
+
+    notify_idle.notified_with_timeout().await;
+    assert_eq!(backend.get(&1), Some(String::from("a")));
+
+    time_provider.inc(Duration::from_secs(1));
+    notify_idle.not_notified().await;
```
alamb, line 117: in this case you have modeled a "slow refresh" by never releasing the barrier from `loader.block_next()`, right?

cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -0,0 +1,168 @@
...
+    refresh::{
+        test_util::{TestLoader, TestRefreshDurationProvider},
+        RefreshPolicy,
+    },
+    ttl::{test_util::TestTtlProvider, TtlPolicy},
+    PolicyBackend,
+};
+
+#[tokio::test]
+async fn test_refresh_can_prevent_expiration() {
```
alamb, line 20: 👨‍🍳 ❤️

alamb (approved):
This is really nice work 🏆  - I went through the logic and tests carefully. My only real questions are related to the "hot loop" avoidance in the background_task but I would say the tests in this PR  assuage any potential concern I may have.

cache_system/src/backend/policy/refresh.rs
crepererum: It is possible if somewhere up the callchain you've started a tokio runtime. But if you put this into a normal sync test, this will panic. So I consider "tokio runtime running" an API requirement. However I admit this is a rather soft/personal design choice.

cache_system/src/backend/policy/integration_tests.rs
crepererum: yes

cache_system/src/backend/policy/refresh.rs
crepererum: Let me explain:

1. The `Notify` here acts as some kind of low-high flag, not as a state. I.e. it should be set once we are idling and then left alone until we are changing from "busy" to "idle" again. Otherwise it is pretty much useless for the tests. (it is already a bit of tricky interface, but still WAY better than `tokio::time::sleep`.
2. How do we know that polling `refresh_tasks.next()` or `rx_refresh_tasks` are "pending"? We just use a `select!` with `biased;` and put the notify logic into the last branch (meaning this is only hit when all other branches are pending).
3. However when we would unconditionally evaluate our "notify" branch, then we would hot-loop because there is nothing stopping us (`notify_one()` can be called multiple times, even when the receiver did not yet poll).

cache_system/src/backend/policy/refresh.rs
crepererum: There is one further "down" in this function that is then further "up" in the callstack:

```rust
        let cancelled_captured = cancelled.clone();
        let fut = async move {
            tokio::select! {
                _ = cancelled_captured.cancelled() => {},
                _ = fut => {}
            }
        }
        .boxed();
```

This will cancel the loader, however there is still the slight chance that this is racy because in a multithreaded tokio runtime (which we use in prod) one thread may call `cancel` but this very future here is only canceled when it hits a "yield point" (mostly: `.await`), since there is no sane way to stop a running thread (this is also a reason why the std threads cannot be cancelled). However (as commented in the section you are annotating) when we check the token manually, we are sure that nobody else could `cancel` is from some other thread, because this can only happen throw the `Subscriber` interface and that would require the mentioned `PolicyBackend` internal lock.

I admit this is a bit of a lock-dance, but it is by far the easiest (and most elegant) approach I could come up with (trust me, you don't wanna see the tag/vector-clock version I had at some point :laughing: ).

cache_system/src/backend/policy/refresh.rs
crepererum: I have to admit that I was surprised that this is actually valid and works.

cache_system/src/backend/policy/refresh.rs
crepererum: correct

cache_system/src/backend/policy/refresh.rs
crepererum: Yes, it would panic. Also the loader was configured to write "foo" into the value (not "b") and we assert that `        assert_eq!(backend.get(&1), Some(String::from("b")));`.

cache_system/src/backend/policy/refresh.rs
```diff
@@ -0,0 +1,993 @@
...
+                callback_handle,
+                metric_refreshed,
+                background_worker,
+                timings,
+                tx_refresh_tasks,
+            }
+        }
+    }
+
+    fn refresh(&self, k: K) -> CancellationToken {
```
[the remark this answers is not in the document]
crepererum: done

PR #5433, opened 2022-08-19
chore: Update datafusion (with new sqlparser release) - option 1

alamb:
# Rationale
* Keep up with upstream, specifically sqlparser upgrade, https://github.com/apache/arrow-datafusion/pull/3200

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/3df9f8002cf14939a82e5182bab553ee5bf6e37b
2. Run `cargo update` and check that in
3. Update for `user` becoming a reserved keyword

Follow on work to support `user` without quotes is tracked in https://github.com/influxdata/influxdb_iox/issues/5491


alamb:
The query_tests crate fails due in this way

```
failures:
    cases::test_cases_basic_sql
```

Which is due to https://github.com/sqlparser-rs/sqlparser-rs/pull/561 now treating `user` as though it were a `user()` function as `user` is a special token in postgres.

Here is what postgres does with that query
```sql
ERROR:  syntax error at or near "user"
LINE 1: create table sss(user int);
                         ^
alamb=# create table sss("user" int);
CREATE TABLE
```

We could work around this in IOx by changing `select user from foo" to `select "user" from foo` but given how common user 

Instead of 
```sql
SELECT user, region from cpu;
```

We can write 
```sql 
SELECT "user", region from cpu;
```

However, given how common `user` is (as it is part of the core `cpu` telegraph metrics) I am thinking maybe it is time to implement https://github.com/apache/arrow-datafusion/issues/533

🤔

alamb:
I am going to try an alternate approach to support this and we can figure out how important it is

alamb:
Per @pauldix  we will go with this approach (option 1) and defer supporting unquoted `user` until someone has a more significant need: https://github.com/influxdata/influxdb_iox/issues/5491

NGA-TRAN (approved):
LGTM

PR #5435, opened 2022-08-19
feat: renaming some of the confusing sequencer things

carols10cents:
## This PR renames

- `data_types::Sequencer` to `data_types::Shard`
- `router::Sequencer` to `router::Shard`
- `data_types::SequencerId` to `data_types::ShardId`
- `data_types::KafkaPartition` to `data_types::ShardIndex`
- "write buffer partition" to "shard index" in the CLI/env vars 

## This PR does NOT change

- `Sequence` - the data type that consists of a `ShardIndex` and a `SequenceNumber`.
- `SequenceNumber` (although there were some instances of "sequencer number" that I changed to "sequence number" if that's what it meant). A `Shard` assigns `SequenceNumber`s.
- `KafkaTopic` and `KafkaTopicId`, eventually I think this should be renamed to Topic and TopicId or otherwise abstracted except for in the router/write buffer
- Metrics that use attributes `kafka_partition`, do you want those to stay or be renamed to `shard_index`?
- Postgres column/table names; coming soon in a separate PR

## Merging this PR will break everything unless these are changed first

- [x] CLI flags/env vars used in k8s, PR out https://github.com/influxdata/k8s-idpe/pull/10265
- [ ] dashboards that use metrics:
  - [ ] `sequencer_unknown_sequence_number` is now `shard_unknown_sequence_number`
  - [ ] `sequencer_invalid_data` is now `shard_invalid_data`
  - [ ] `sequencer_unknown_error` is now `shard_unknown_error`
  - [ ] `sequencer_reset_count` is now `shard_reset_count`
  - [ ] the `iox_catalog` `sequencer_*` metrics that are now `shard_`


domodwyer:
A big (the biggest?!) gotcha we keep hitting is the similarity of a partition ID, and a partition ID (see what I did there?)

We have two things with the same name:

* The ID of the Kafka partition you use to say "this Kafka partition please Kafka" using the Kafka client. This is a 0..N range, where N is the number of Kafka partitions.

* The ID of the partition in the database used to refer to the _database record_ for a specific Kafka partition, which is currently happens to be 0..N because it is an autoincrementing ID column and we started at 0, but could be anything (UUIDs, etc).

Confusing the two is easy because they're all called IDs in the code! What's this guy?

https://github.com/influxdata/influxdb_iox/blob/8b054c14a8e45027fbcfd1973526befa8fda5e04/data_types/src/lib.rs#L2065-L2066

Thankfully someone put a comment, which I read as "the number you use when you talk to Kafka" and not "the database row ID that represents the Kafka partition". There's plenty more places where that [distinction isn't so clear](https://github.com/influxdata/influxdb_iox/blob/8b054c14a8e45027fbcfd1973526befa8fda5e04/compactor/src/compact.rs#L80-L81), and we wind up with comments like this that we hope people see:

https://github.com/influxdata/influxdb_iox/blob/8b054c14a8e45027fbcfd1973526befa8fda5e04/data_types/src/lib.rs#L156-L163

What do you think to calling the 0..N ranged number you give to Kafka to tell it which partition to write to, an "shard index"? So you'd have a Kafka index, which indexes into the 0..N kafka partitions, and a shard ID which is the ID of the shard in the database?

carols10cents:
> What do you think to calling the 0..N ranged number you give to Kafka to tell it which partition to write to, an "shard index"? So you'd have a shard index, which indexes into the 0..N Kafka partitions, and a shard ID which is the ID of the shard in the database?

I'm into it! So `Sequence` would change like this?

```diff
- /// Kafka partition ID plus offset
+ /// Shard index plus offset
pub struct Sequence {
-    /// The sequencer id (kafka partition id)
-    pub sequencer_id: u32,
+    /// The shard index
+    pub shard_index: ShardIndex, // I think we could make a type for this now

    /// The sequence number (kafka offset)
    pub sequence_number: SequenceNumber,
}
```

And `Sequencer` would change like this?

```diff
- pub struct Sequencer {
+ pub struct Shard {
-    /// the id of the sequencer
-    pub id: SequencerId,
+    /// the database id of the shard
+    pub id: ShardId,
    /// the topic the sequencer is reading from
    pub kafka_topic_id: KafkaTopicId, // eventually I want to rename this to "topic_id" to get kafka out more
-    /// the kafka partition the sequencer is reading from
-    pub kafka_partition: KafkaPartition,
+    /// the shard index the shard is reading from
+    pub shard_index: ShardIndex,
    /// The minimum unpersisted sequence number. Because different tables
    /// can be persisted at different times, it is possible some data has been persisted
    /// with a higher sequence number than this. However, all data with a sequence number
    /// lower than this must have been persisted to Parquet.
    pub min_unpersisted_sequence_number: SequenceNumber,
}
```

Do I have that right?

domodwyer:
You got it!

> make a type for this now

❤️ 

I hope this helps... I've had enough of ID shenanigans!

carols10cents:
I think I like that, and I think that helps with some of the "sharder" worry I have-- a "sharder" would be something that assigns a "shard index" (rather than a "sequencer ID") here: https://github.com/influxdata/influxdb_iox/pull/5435/files#diff-551c84f9150f40a52e10fcb989bf6e9483662a7ef0ce61355397c7e7de52b9c0R74

domodwyer:
Yeah even reading that as a sentence was clearer!

Thank you for taking this on ❤️

carols10cents:
Ok, I don't think this is quite ready to merge yet because it renames CLI args/env vars and metric names, so it'll break things unless those get updated, but I'd like to make sure people know this is happening....

See the summary in the PR description!

@NGA-TRAN please take a skim to see if this makes sense; this changes some names in the compactor/ingester

@domodwyer please check that i've gotten this right and haven't taken things *too* far

@lukebond please see if this would resolve your confusion without creating additional confusion

@crepererum and @tustvold I *know* you have opinions

@jacobmarble @pauldix fyi mostly for the nomenclature changes

lesam:
FWIW I think the really confusing thing is that IOx has an internal concept called partitions, and also is using Kafka partitions that are completely separate.

Renaming Sequencer to Shard doesn’t help a ton for a newcomer, especially since you need to know that ‘Shard’==‘Kafka Partition’ and an ‘IOx Partition’ is something different

mkmik:
Assuming that the unfortunate overloading of the "partition" term cannot be solved, I think it's much less confusing to talk about "shards" being implemented via Kafka "partitions" rather than talking about "sequencers*

crepererum:
> The ID of the Kafka partition you use to say "this Kafka partition please Kafka" using the Kafka client. This is a 0..N range, where N is the number of Kafka partitions.

This coupling is true for Kafka, but the write buffer deliberately abstracts this and let's the user NOT assume that this range starts at 0 or is gapless so we can easily use solutions that don't have this property or allow partition deletion later.

crepererum:
I like the renaming.

One thing to keep in mind is that "Kafka partition" is slightly ambiguous itself:

- It can mean "the partition within this topic" (basically an integer). My impression is that many people use this terminology because the write buffer always scopes operations to a single topic.
- Within the Kafka protocol however it usually means "this partition within the whole cluster" (now it's the topic String and the integer). This BTW is also what the IOx catalog uses (it links a single global ID to a pair of topic and index within the topic).

domodwyer (approved):
good lord - look at that PR!

```
       ┌─┐ 
       ║"│ - i love it!!
       └┬┘ 
       ┌┼┐ 
        │  
       ┌┴┐ 
```

NGA-TRAN (approved):
What a PR! 🎉 

I like renaming `sequencer` to `shard` because it will make more sense with our architecture and common word `shard` used in the public. I think this will also the start for us to define the difference between `Kafka partition` and `IOx partition`. I think we may want to start eliminating one of them (e.g. when we say `partition`, we really mean IOx partition only).

One think I think a bit confused in this PR is `ShardId` and `ShardIndex`.  I saw there are comments for them at their definition but I think adding example going with the below comment for `ShardIndex` will help readers/reviewers/coders understand better why we need both and when to use one but not the other.

```
/// The index of the shard in the set of shards. When Kafka is used as the write buffer, this is
 /// the Kafka Partition ID.
```

lukebond (approved):

carols10cents:
> This coupling is true for Kafka, but the write buffer deliberately abstracts this and let's the user NOT assume that this range starts at 0 or is gapless so we can easily use solutions that don't have this property or allow partition deletion later.

The write buffer CLI arguments don't abstract this. I think this is best fixed when we have another solution that doesn't have this property.

carols10cents:
> I think we may want to start eliminating one of them (e.g. when we say `partition`, we really mean IOx partition only).

Yeah there are a few spots in the write buffer (outside of the Kafka write buffer implementation) and router that say "partition" and I think they mean "kafka partition", but I think that can be fixed in a future PR.

carols10cents:
> if we had a diagram of the whole collection including shard, namespace, table, partition and annotations to the various types it'd be handy and should clear things up if the newcomer can be pointed to that first.

Great point, I will send a PR to update https://github.com/influxdata/conductor/blob/main/docs/iox_cloud_architecture.md once I have the database columns renamed!

aierui:
When I first started reading I was very confused about these keywords above, this adjustment will make it clearer 🎉🎉🎉

jacobmarble:
I don't mind this rename PR as-is. We can further iterate with low friction, until we decide to change the catalog schema. 👍 

After this PR is merged, I want to explore the clash between InfluxDB/TSM "shards" and IOx "shards", which many IOxers may not be fully aware of.

jacobmarble:
> One thing to keep in mind is that "Kafka partition" is slightly ambiguous itself:
> 
> - It can mean "the partition within this topic" (basically an integer). My impression is that many people use this terminology because the write buffer always scopes operations to a single topic.
> 
> - Within the Kafka protocol however it usually means "this partition within the whole cluster" (now it's the topic String and the integer). This BTW is also what the IOx catalog uses (it links a single global ID to a pair of topic and index within the topic).

My guess is that you @crepererum have permanent brain damage from working in the illegal coal mine that is the Kafka protocol, and I'm very sorry for that. ;-)

From my experience working with the Kafka service and a few Kafka client libraries, it is correct to define a Kafka partition as the tuple `(topic name, partition id)`. The topic name is always a string, like `ingress`, and the partition id is always a zero-based integer index.

@carols10cents I might suggest that the Kafka topic references be string `TopicName` or similar. (cc @crepererum)

cf sarama [TopicMetadata](https://pkg.go.dev/github.com/Shopify/sarama#TopicMetadata) and [PartitionMetadata](https://pkg.go.dev/github.com/Shopify/sarama#PartitionMetadata)
```golang
type TopicMetadata struct {
	Name       string
	Partitions []*PartitionMetadata
	...
}

type PartitionMetadata struct {
	ID              string
	...
}
```

carols10cents:
@jacobmarble 

> We can further iterate with low friction, until we decide to change the catalog schema. 👍

Does that mean I should *not* work on renaming the postgres catalog right now? And wait a little bit once this is merged, to make sure working with the code this way makes sense?

jacobmarble:
Does this help or close https://github.com/influxdata/influxdb_iox/issues/4237 and/or https://github.com/influxdata/influxdb_iox/issues/4246 ?

carols10cents:
> Does this help or close #4237 and/or #4246 ?

I think it helps those, yes. I'll let the team decide if it closes them ;) #4246 specifically calls out changing the columns in the catalog.

carols10cents:
@domodwyer are you really ready for me to merge this and break some metrics? can you take care of fixing them?

@mkmik now that https://github.com/influxdata/k8s-idpe/pull/10265 is merged, is it safe to merge this? Or is there anything this needs to wait for?

mkmik:
LGTM

PR #5436, opened 2022-08-19
fix: Correctly lexigraphically sort `_field` and `_measurement` with upper case tag keys

alamb:
# Rationale

Closes #https://github.com/influxdata/conductor/issues/1098

TSM returns tags in a series 'lexicographically sorted on tag key' -- aka if we have `tagA`, `tagC` and `tagB` TSM will report them `tagA`, `tagB` and `tagC` in the returned series. IOx must do the same thing.

However, while comparing differences between TSM and IOx it turns out IOx was doing something wrong. 

# Changes

The code incorrectly assumed   that `_field` and `_measurement` are always the first two tags, which is true as long as all tags begin with lower case letter (which is the common case),  but it is __not__ true if any of the tags start with capital letters 

asciifull

See details on https://github.com/influxdata/conductor/issues/1098#issuecomment-1220995433 where `GVA_ROUTE`


alamb:
The tests in this PR are sometimes failing due to the flaky test. Once the fix  https://github.com/influxdata/influxdb_iox/pull/5437 is merged we should be good

domodwyer (approved):
TIL: this ordering is madness.

PR #5438, opened 2022-08-19
chore: Update datafusion (with new sqlparser release) - option 2

alamb:
This is an alternate approach than in https://github.com/influxdata/influxdb_iox/pull/5433

Basically this PR parses SQL queries, finds and "special" functions (like `user`) and replaces them with a quoted reference like `"user"`

It isn't complete (I would need to fill out all the structs to properly recurse) but this shows how it would be done

### DIFF
--- a/iox_query/src/frontend/sql_rewrite.rs
+++ b/iox_query/src/frontend/sql_rewrite.rs
@@ -1,0 +1,248 @@
+use sqlparser::ast::{
+    Expr, Fetch, Function, FunctionArg, FunctionArgExpr, Ident, LockType, ObjectName, Offset,
+    OrderByExpr, Query, Select, SelectItem, SetExpr, Statement, WindowSpec, With,
+};
+use sqlparser::dialect::GenericDialect;
+use sqlparser::parser::Parser;
+
+pub fn rewrite_sql(sql: &str) -> String {
+    let dialect = GenericDialect {};
+    match Parser::parse_sql(&dialect, sql) {
+        Ok(statements) if statements.len() == 1 => {
+            let statement = statements.into_iter().next().unwrap();
+            let rewritten = RewriteSpecial {}.rewrite_statement(statement);
+            rewritten.to_string()
+        }
+        _ => sql.to_string(),
+    }
+}
+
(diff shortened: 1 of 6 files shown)


iox_query/src/frontend/sql_rewrite.rs
```diff
@@ -0,0 +1,248 @@
...
+        Ok(statements) if statements.len() == 1 => {
+            let statement = statements.into_iter().next().unwrap();
+            let rewritten = RewriteSpecial {}.rewrite_statement(statement);
+            rewritten.to_string()
+        }
+        _ => sql.to_string(),
+    }
+}
+
+// THIS IS NOT COMPLETE -- would need to fill out the rest of the code
```
alamb, line 20: Here is the magic

iox_query/src/frontend/sql_rewrite.rs
```diff
@@ -0,0 +1,248 @@
...
+/// select user, time from cpu;
+/// ```
+/// To select the amount of time the CPU spent on user queries
+struct RewriteSpecial {}
+
+/// Override only the part we care about
+impl Rewriter for RewriteSpecial {
+    fn rewrite_expr(&mut self, e: Expr) -> Expr {
+        if let Expr::Function(f) = e {
+            if f.special {
```
alamb, line 38: the `special` field is how you can tell a function call like `user()` from `user`

alamb:
This is also the type of interface I envision adding to DataFusion to allow "custom sql rewrites", ala https://github.com/apache/arrow-datafusion/issues/533

alamb:
Per @pauldix  we will go with option 1 (put `"` around `user`) and I am tracking the issue in https://github.com/influxdata/influxdb_iox/issues/5491

lovasoa:
@alamb, to follow up on https://github.com/sqlparser-rs/sqlparser-rs/pull/601 , this would be a great usecase for the visitor pr. If it gets merged, this could be implemented in just three lines of code:

```rust
ast.drive_mut(&mut visitor_enter_fn_mut(|expr: &mut Expr| {
    if let Expr::Function(Function { name, special: true, .. }) = expr {
        *expr = Expr::Identifier(Ident::new(name.to_string()))
    }
}))
```

PR #5449, opened 2022-08-22
feat: refresh namespace cache before expiration

### LINKED ISSUE #5318: Querier Cache: Rework TTL

crepererum:
Closes #5318.


querier/src/cache/namespace.rs
```diff
@@ -111,6 +116,20 @@ impl NamespaceCache {
...
+                Arc::new(OptionalValueRefreshDurationProvider::new(
+                    Some(REFRESH_NON_EXISTING),
+                    Some(REFRESH_EXISTING),
+                )),
+                Arc::clone(&time_provider),
+                Arc::clone(&loader) as _,
+                CACHE_ID,
+                metric_registry,
+            )
+            .await,
```
lesam, line 130: Why is `RefreshPolicy::new` async? It doesn't look to me like it needs to be?

crepererum: There is a comment in the docstring of exactly that method:

> Even though this does NOT perform any IO, we require this to be `async` so that we can start a tokio background process.

I.e. `tokio::spawn` doesn't work if you're in a purely sync context. This esp. hits tests and is really confusing when wiring everything, so I decided that "we have an async runtime running" is part of the API and therefore the method is sync. I admit it is a rather personal flavor.

querier/src/cache/namespace.rs
```diff
@@ -26,7 +27,10 @@ use trace::span::Span;
 use super::ram::RamSize;
 
 /// Duration to keep existing namespaces.
-pub const TTL_EXISTING: Duration = Duration::from_secs(10);
+pub const TTL_EXISTING: Duration = Duration::from_secs(60);
```
lesam, line 30: When do we hit the TTL? Wouldn't the constant refresh from the `REFRESH_EXISTING` keep the TTL alive? Do we have tests to confirm/deny whether a refresh counts as a cache hit?

Also, are we taking this comment from the issue into account? https://github.com/influxdata/influxdb_iox/issues/5318#issuecomment-1207909514 - I'm not quite sure on the details here but this doesn't look like data and catalog caching are currently done differently.

querier/src/cache/namespace.rs
lesam: I don't love the async infection here into many methods that should logically be sync.

Also pedantically, just because a method is async doesn't necessarily mean tokio is running right? There could be a different executor.

If this approach has already been discussed across the IOx team and this is the standard the team has picked, that's fine, but if it is only your personal preference maybe there should be a discussion / decision?

crepererum: > Also pedantically, just because a method is async doesn't necessarily mean tokio is running right? There could be a different executor.

Technically you're right, but whole IOx will explode if you use anything other than tokio.

> If this approach has already been discussed across the IOx team and this is the standard the team has picked, that's fine, but if it is only your personal preference maybe there should be a discussion / decision?

I'm not sure if I care enough for a team-wide style discussion.

querier/src/cache/namespace.rs
crepererum: We only refresh when the value is used. This avoids refreshing values that are basically not used any longer.

The interplay is tested in the `cache_system`, I'm not going to repeat this here again.

> Also, are we taking this comment from the issue into account?

This is exactly what the TTL policy and the refresh policy together achieve. I can give you a 1:1 rundown on the whole policy system if you like.

lesam: > The interplay is tested in the cache_system, I'm not going to repeat this here again.

That's fine, just making sure those tests exist (I need to go look at them to make sure I understand the behaviour)

> I can give you a 1:1 rundown

That would be great, let's do that tomorrow

lesam (approved):
LGTM, thanks for the help to understand

crepererum:
Blocked by #5483 (so I don't need to refactor things twice).

PR #5447, opened 2022-08-22
feat: upsert partition & update sort key for each day in bulk ingest

lukebond:
This is the final PR of the Rust portion of preparatory bulk ingest steps that run before the actual parquet file creation.

- add support for the recently added `earliest_time` & `latest_time` fields in the schema that comes from the prestep, merging them into a single date range that spans the whole import period
- after validating the schema and updating the catalog accordingly, create partitions for each day in the date range, if not already created, and create/update the partitions' sort key based on the tag cardinality of the schema we got from the prestep

Closes https://github.com/influxdata/conductor/issues/1091
Closes https://github.com/influxdata/conductor/issues/1092

Describe your proposed changes here.

- [ ] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #5451, opened 2022-08-23
feat: sharder API definition

domodwyer:
A proposed API definition to expose sharding logic outside of the router.

Ideally the sharding logic lives in one place, rather than multiple components having their own  compiled-in implementation to minimise bugs / simplify keeping everything in sync / support customised sharding keys in the future. This PR exposes the sharding logic in the router to any caller in a future-proof way.

Relates to https://github.com/influxdata/influxdb_iox/pull/5447#pullrequestreview-1080574538

---

* feat: sharder API definition (57bbe6b21)

      This commit adds a gRPC endpoint for callers to map (table, namespace) tuples
      to Sequencer IDs, using the logic internal to the router.

      Reference:
        
      https://github.com/influxdata/influxdb_iox/pull/5447#pullrequestreview-1080574538

### DIFF
--- a/generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
+++ b/generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
@@ -1,0 +1,26 @@
+syntax = "proto3";
+package influxdata.iox.sharder.v1;
+option go_package = "github.com/influxdata/iox/sharder/v1";
+
+service ShardService {
+  // Shard the given inputs to a Catalog ID for the destination Sequencer
+  // (Sequencer ID).
+  rpc ShardToSequencerId(ShardToSequencerIdRequest) returns (ShardToSequencerIdResponse);
+}
+
+message ShardToSequencerIdRequest {
+  // The shard input values to map onto a Sequencer.
+  oneof input {
+    // A table & namespace pair.
+    TableNamespaceTuple table_namespace = 1;
+  }
+}
+
+message ShardToSequencerIdResponse {
+  uint64 sequencer_id = 1;
+}
+
+message TableNamespaceTuple {
+  string table = 1;
+  string namespace = 2;
+}


generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
```diff
@@ -0,0 +1,26 @@
...
+  // (Sequencer ID).
+  rpc ShardToSequencerId(ShardToSequencerIdRequest) returns (ShardToSequencerIdResponse);
+}
+
+message ShardToSequencerIdRequest {
+  // The shard input values to map onto a Sequencer.
+  oneof input {
+    // A table & namespace pair.
+    TableNamespaceTuple table_namespace = 1;
+  }
```
domodwyer, line 16: In the future we intend(ed?) to support multiple sharding strategies - one fine example is the tracing/prometheus use case where one `(table, namespace)` tuple dominates the workload, leading to poor utilisation of Kafka (hot spots).

This `oneof` lets us add more strategies in the future, such as sharding on a trace ID / metric name column in the above example.

generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+package influxdata.iox.sharder.v1;
+option go_package = "github.com/influxdata/iox/sharder/v1";
+
+service ShardService {
+  // Shard the given inputs to a Catalog ID for the destination Sequencer
+  // (Sequencer ID).
+  rpc ShardToSequencerId(ShardToSequencerIdRequest) returns (ShardToSequencerIdResponse);
```
domodwyer, line 8: @carols10cents I believe this will need to be `ShardID` when your PR goes through right?

lukebond (approved):
perfect for my needs, will start building it out

generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
carols10cents: Maybe!!!

domodwyer: I actually wound up returning both the index & ID in the response, maybe we can use `MapToShardRequest` dropping the ID/index part?  (see other comment: https://github.com/influxdata/influxdb_iox/pull/5462#discussion_r953881647)

PR #5452, opened 2022-08-23
refactor(router): use KafkaPartition in sequencer

domodwyer:
A quick tidy-up PR that uses the `KafkaPartition` type wrapper instead of bare `usize`, adds asserts when we cast between types.

I'm not sure why, but the [`KafkaPartition` is an `i32`](https://github.com/influxdata/influxdb_iox/blob/9b920f1cbbfee5b996b5757cd3f855b4a7a72b95/data_types/src/lib.rs#L184), but the [`WriteBuffer` uses a `u32`](https://github.com/influxdata/influxdb_iox/blob/9b920f1cbbfee5b996b5757cd3f855b4a7a72b95/write_buffer/src/core.rs#L167) even though Kafka (and it's clients) [use `i32`](https://github.com/influxdata/influxdb_iox/blob/9b920f1cbbfee5b996b5757cd3f855b4a7a72b95/write_buffer/src/kafka/mod.rs#L452) to identify partitions - this causes us to cast a whole lotta times.

Relates to https://github.com/influxdata/influxdb_iox/pull/5435

---

* refactor(router): use KafkaPartition in sequencer (9b920f1cb)

      The Sequencer (which will be renamed shortly) is a type that represents a
      single sequencer/shard/kafka partition in the router.

      In order to minimise confusion with all the various IDs floating around, we
      have a KafkaPartition - this commit changes the Sequencer to return the Kafka
      partition index as a typed value, rather than a usize to help eliminate any
      inconsistencies.

      As a side effect of these conversion changes, I've tightened up the casting to
      ensure we assert on any overflows - we juggle a lot of numeric types!

### DIFF
--- a/ioxd_router/src/lib.rs
+++ b/ioxd_router/src/lib.rs
@@ -1,6 +1,6 @@
 use async_trait::async_trait;
 use clap_blocks::write_buffer::WriteBufferConfig;
-use data_types::{DatabaseName, PartitionTemplate, TemplatePart};
+use data_types::{DatabaseName, KafkaPartition, PartitionTemplate, TemplatePart};
 use hashbrown::HashMap;
 use hyper::{Body, Request, Response};
 use iox_catalog::interface::Catalog;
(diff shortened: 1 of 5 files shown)


carols10cents:
LOL oh geez... I love this, but I'm going to have one hell of a rebase.......

Issue #5454, opened 2022-08-23
Regression in dotenvy 0.15.2 


mkmik:
__Steps to reproduce:__

```console
$ cp docs/env.example .env  
$ cargo run -- run all-in-one
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `target/debug/influxdb_iox run all-in-one`
FATAL Error loading config from: Error parsing line: '# In the Storage account's Settings > Access keys, one of the Key values
# AZURE_STORAGE_ACCESS_KEY=
#
# To enable Jaeger tracing:
# OTEL_SERVICE_NAME="iox" # defaults to iox
# OTEL_EXPORTER_JAEGER_AGENT_HOST="jaeger.influxdata.net"
# OTEL_EXPORTER_JAEGER_AGENT_PORT="6831"
', error at line index: 274
Aborting
```

Minimal reproducer:

```console
$ echo "# '" >.env
$ cargo run -- run all-in-one
    Finished dev [unoptimized + debuginfo] target(s) in 0.26s
     Running `target/debug/influxdb_iox run all-in-one`
FATAL Error loading config from: Error parsing line: '# '
', error at line index: 4
Aborting
```

__Environment info:__

Commit `84acbd2fa - chore(deps): Bump dotenvy from 0.15.1 to 0.15.2 (#5450) (5 hours ago) <dependabot[bot]>`

---

- [x] https://github.com/influxdata/influxdb_iox/pull/5455
- [ ] https://github.com/influxdata/influxdb_iox/pull/5461

mkmik:
upstream https://github.com/allan2/dotenvy/issues/12

PR #5455, opened 2022-08-23
fix: Revert Bump dotenvy from 0.15.1 to 0.15.2 (#5450)

mkmik:
This reverts commit 84acbd2fad7980fbdd3e2a9b17854e92e370b6ef.

There is an upstream issue; https://github.com/allan2/dotenvy/issues/12

We need a test to make sure dependabot doesn't update the dep back to some version that hasn't fixed the upstream issue yet.

That's why I'm not c-losing #5454 in this PR.


crepererum:
You can add a rule for dependabot here:

https://github.com/influxdata/influxdb_iox/blob/3220c6f88bbacc5e567bed4d9fc2b50f68ebb42e/.github/dependabot.yml#L9-L15

However we shall also prevent people from re-introducing this via a manual `cargo update` (which we often use to unbreak CI after crates were yanked from crates.io). This needs an explicit constraint in our `Cargo.toml` files:

```toml
dotenvy = ">=0.15.1,<0.15.2"
```

(cargo doesn't support "not equal")

crepererum (approved):
Approving to unblock prod.

mkmik:
btw, I'm not sure I'm reading this right:

```
dotenvy = ">=0.15.1,<0.15.2"
```

doesn't this basically mean `exactly equal to 0.15.1` ?

mkmik:
(merging now and following up with either a test or a pin)

PR #5456, opened 2022-08-23
feat: add file_count_threshold for comapcting cold partitions

NGA-TRAN:
add file_count_threshold for comapcting cold partitions to make it consistent with the hot case and help set up to avoid ooms easier

Closes #https://github.com/influxdata/conductor/issues/1122


clap_blocks/src/compactor.rs
```diff
@@ -183,6 +183,23 @@ macro_rules! gen_compactor_config {
...
 
+            /// A compaction operation for cold partitions  will gather as many L0 files with their 
+            /// overlapping L1 files to compact together until the total number of L0 + L1 files 
+            /// crosses this threshold.
+            /// Later compactions will pick up the remaining L0 files.
+            ///
+            /// A compaction operation will be limited by this or by the cold input size threshold,
+            /// whichever is hit first.
+            ///
+            /// Default is 50.
```
pauldix, line 194: Shouldn't this be output automatically by the clap help based on the value just below? Better to not have it in the comments if so.

pauldix (approved):
One small nit on the clap comment, but otherwise LGTM

clap_blocks/src/compactor.rs
NGA-TRAN: Done

PR #5459, opened 2022-08-23
fix: prune chunks earlier

### LINKED ISSUE #5453: Improve pruning performance of cached values in the catalog
__Steps to reproduce:__

1. Run a query with tracing enabled on the last 10s of a measurement that has been around for a few months
2. Look at tracing data that shows 150ms+ spent deciding which of the few hundred chunks in the cache are going to be queried

__Expected behaviour:__
Much faster pruning / planning (~10ms or less?)

__Actual behaviour:__
Long-ish planning time, which contributes to queuing delay / wasted CPU when many queries are being issued.

lesam:
This improves speed significantly for me for a small query (from ~50ms to ~5ms locally).

Closes #5453

### DIFF
--- a/predicate/src/lib.rs
+++ b/predicate/src/lib.rs
@@ -13,13 +13,15 @@ pub mod delete_predicate;
 pub mod rewrite;
 pub mod rpc_predicate;
 
+use crate::Expr::BinaryExpr;
 use arrow::{
     array::{
         BooleanArray, Float64Array, Int64Array, StringArray, TimestampNanosecondArray, UInt64Array,
     },
     datatypes::SchemaRef,
 };
-use data_types::{InfluxDbType, TableSummary, TimestampRange};
+use data_types::{InfluxDbType, TableSummary, TimestampRange, MAX_NANO_TIME, MIN_NANO_TIME};
+use datafusion::common::ScalarValue;
 use datafusion::{
     error::DataFusionError,
     logical_expr::{binary_expr, utils::expr_to_columns},
@@ -31,6 +33,7 @@ use datafusion_util::{make_range_expr, nullable_schema};
 use observability_deps::tracing::debug;
 use rpc_predicate::VALUE_COLUMN_NAME;
 use schema::TIME_COLUMN_NAME;
+use std::cmp::{max, min};
 use std::{
     collections::{BTreeSet, HashSet},
     fmt,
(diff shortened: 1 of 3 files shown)


lesam:
Putting this in draft for now, will self-review tomorrow. Would appreciate an initial look @crepererum

predicate/src/lib.rs
```diff
@@ -240,6 +243,85 @@ impl Predicate {
...
+                                _ => *op,
+                            };
+                            parsed_time = *time;
+                        }
+                        (_, _) => return true,
+                    }
+                }
+                _ => return true,
+            }
+            if parsed_col.relation.is_some() || parsed_col.name != TIME_COLUMN_NAME {
```
lesam, line 296: Not completely sure about this part...

predicate/src/lib.rs
```diff
@@ -240,6 +243,85 @@ impl Predicate {
...
+            range.end()
+        } else {
+            MAX_NANO_TIME + 1
+        };
+
+        self.exprs.retain(|e| {
+            // parse for 'parsed_col op parsed_time'
+            let parsed_col;
+            let parsed_op;
+            let parsed_time: i64;
```
crepererum, line 263: I didn't even know it was legal to postpone variable assignment in Rust :sweat_smile:

predicate/src/lib.rs
crepererum: Could we pass the table name into the method and assert (actually via a real assertion that would panic) that we are only hitting predicates here that are for a single table (i.e. `assert_eq!(parsed_col.relation, Some(table_name);`)? This would ensure two things:

1. the predicate pushdown doesn't pushes down predicates for other tables (e.g. for `SELECT * FROM a JOIN b ON a.time == b.time WHERE a.time > 100` and we are in table `b`, then either we shouldn't have a predicate or -- even better -- the predicate is `b.time > 100`)
2. Nobody calls this method if we're NOT in table scope (InfluxRPC allows predicates to be applied to multiple tables at the same time and I'm kinda afraid that someone will call this very method here too early).

querier/src/table/mod.rs
```diff
@@ -549,7 +554,7 @@ mod tests {
         // this contains all files except for:
         // - file111: marked for delete
         // - file221: wrong table
-        let pred = Predicate::new().with_range(0, 100);
+        let pred = Predicate::new().with_range(0, 101);
```
crepererum, line 557: Why?

querier/src/table/mod.rs
```diff
@@ -233,18 +233,23 @@ impl QuerierTable {
 
         // create parquet files
         let parquet_files: Vec<_> = futures::stream::iter(parquet_files.files.iter())
-            .filter_map(|cached_parquet_file| {
+            .filter_map(|cached_parquet_file| async move {
+                if let Some(r) = predicate.range {
```
crepererum, line 237: We could probably have a test for this that goes like this:

- have a catalog w/ one parquet file
- have a predicate that removes that parquet file
- asssert (via the metric subsystem) that we never peek the catalog cache to request the read buffer for this file

PR #5460, opened 2022-08-24
feat: Teach IOx how to parse InfluxQL literals

### LINKED ISSUE #5444: InfluxQL: Teach IOx how to parse InfluxQL literals
In order to teach InfluxQL how to parse expressions, IOx must first learn how to parse literal values. 

See [InfluxQL expressions](https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#expressions) and the [`ParseExpr`](https://github.com/influxdata/influxql/blob/7e7d61973256ffeef4b99edd0a89f18a9e52fa2d/parser.go#L58-L59) function for documentation.

### Literals

The following rules defines the literals values supported by InfluxQL.

#### String literal

```
string ::= '"' [^"\n] '"'
```

See: https://github.com/influxdata/influxql/blob/df51a45762be9c1b578f01718fa92d286a843fe9/scanner.go#L558-L559

**NOTE:** Per the existing Go code, `"` and several other characters may be escaped with a forward slash, '/'.

#### Integer literal

```
integer ::= [0-9]+
```

#### Number (float) literal

```
number ::= integer '.' integer
```

#### Boolean literal

```
boolean ::= 'true' | 'false'
```

#### Duration literal

```
duration ::= integer ('ns' | 'us' | 'µs' | 'ms' | 's' | 'm' | 'd' | 'w')
```

See: https://github.com/influxdata/influxql/blob/7e7d61973256ffeef4b99edd0a89f18a9e52fa2d/parser.go#L3026-L3051

#### Regular expression literal

```
regex ::= '/' [^/\n] '/'
```

See: https://github.com/influxdata/influxql/blob/df51a45762be9c1b578f01718fa92d286a843fe9/scanner.go#L245-L246

**NOTE:** Per the existing Go code, forward slashes can be escaped as `//`

stuartcarnie:
Closes #5444 

This PR addresses #5444 by teaching IOx how to parse various InfluxQL literal values. The public `literal` parser combinator function will be used by future parsers, like an expression parser.

This PR is dependent on merging PR #5425.


influxdb_influxql_parser/src/literal.rs
```diff
@@ -0,0 +1,404 @@
...
+        Ok(())
+    }
+}
+
+/// Parse an InfluxQL integer.
+///
+/// InfluxQL defines an integer as follows
+///
+/// ```text
+/// INTEGER ::= [0-9]+
```
crepererum, line 111: What about negative numbers? Are they parsed as literals prefixed with an unary operator?

crepererum: For signed and unsigned integers: is `007` a `7` in InfluxQL, not an octal (as it is used by some programming languages)?

influxdb_influxql_parser/src/literal.rs
```diff
@@ -0,0 +1,404 @@
...
+fn unsigned_integer(i: &str) -> IResult<&str, u64> {
+    map_res(digit1, |s: &str| s.parse())(i)
+}
+
+/// Parse an unsigned InfluxQL floating point number.
+///
+/// InfluxQL defines a floating point number as follows
+///
+/// ```text
+/// float   ::= INTEGER "." INTEGER
```
crepererum, line 133: Same question as for signed integers: what about negative values?

Plus: what about NaN/+Inf/-Inf and scientific notation?

influxdb_influxql_parser/src/literal.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+/// Parse the input for a InfluxQL duration fragment and returns the value in nanoseconds.
+fn single_duration(i: &str) -> IResult<&str, i64> {
+    use DurationUnit::*;
+
+    map(
+        pair(
+            integer,
```
crepererum, line 210: This depends on the consequence of my question about negative signed integers, but: can durations be negative?

influxdb_influxql_parser/src/string.rs
```diff
@@ -0,0 +1,266 @@
...
+        assert_eq!(got, "hello\\n".into());
+
+        // Empty regex
+        let (_, got) = regex("//").unwrap();
+        assert_eq!(got, "".into());
+
+        // Fallible cases
+
+        // Missing trailing delimiter
+        assert!(regex(r#"/hello"#).is_err());
```
crepererum, line 257: This is more a hint than something you really have to change:

`assert!(regex(r#"/hello"#).is_err());` will only produce a rather coarse assertion because it basically just tells you "expression was false". However if both your `Ok(...)` and your `Err(...)` variants implement `Debug` (like in this case), then you can also use `regex(r#"/hello"#).unwrap_err();` (w/o any `assert!`). If that fails, than it will also print out the `Ok(...)` value that it found so you quickly understand what it parsed (but apparently shouldn't).

crepererum (approved):
Note that I'm not really an expert in `nom`, so I kinda trust that this idiomatic parsing code :smile: 

I have a couple of semantic questions. Code and tests look very good though.

influxdb_influxql_parser/src/literal.rs
stuartcarnie: Correct - there are no alternative bases in InfluxQL

stuartcarnie: > What about negative numbers? Are they parsed as literals prefixed with an unary operator?

Exactly! For the expression parsing, we have a unary operator to negate a literal

influxdb_influxql_parser/src/literal.rs
stuartcarnie: No support for NaN or INF in InfluxQL and negative numbers are parsed as part of expressions using a unary operator

influxdb_influxql_parser/src/literal.rs
stuartcarnie: Once again, as your surmised earlier, we have a unary operator to negate literals

stuartcarnie:
Thanks for the review @crepererum! I will follow up tomorrow based on your feedback.

influxdb_influxql_parser/src/string.rs
stuartcarnie: Excellent, I'll change to that, which is much better. Thanks!

stuartcarnie: Done!

PR #5462, opened 2022-08-24
refactor(proto): update ShardService gRPC defintion

domodwyer:
An update to the inital proto definition (#5451) for https://github.com/influxdata/influxdb_iox/pull/5447.

---

* refactor(proto): simplify RPC messages & types (f11af90c4)

      Removes the input oneof - a shard caller MUST always provide a 
      table/namespace, and MAY provide an optional payload (which in the future will
      enable sharding using column valuess/etc). As there is currently no
      payload-based sharding, this simplifies the RPC message.

      Changes the returned types to better reflect the types we use internally
      - this should avoid type juggling for both server & client.

* build: generate gRPC bindings for ShardService (c6d4109e0)

      Builds the ShardService proto file in the generated_types package.

### DIFF
--- a/generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
+++ b/generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
@@ -10,17 +10,11 @@ service ShardService {
 
 message ShardToSequencerIdRequest {
   // The shard input values to map onto a Sequencer.
-  oneof input {
-    // A table & namespace pair.
-    TableNamespaceTuple table_namespace = 1;
-  }
+  string table_name = 1;
+  string namespace_name = 2;
 }
 
 message ShardToSequencerIdResponse {
-  uint64 sequencer_id = 1;
-}
-
-message TableNamespaceTuple {
-  string table = 1;
-  string namespace = 2;
+  int64 sequencer_id = 1;
+  int32 sequencer_index = 2;
 }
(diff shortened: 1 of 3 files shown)


lukebond (approved):

generated_types/protos/influxdata/iox/sharder/v1/sharder.proto
```diff
@@ -10,17 +10,11 @@ service ShardService {
...
 
 message ShardToSequencerIdResponse {
-  uint64 sequencer_id = 1;
-}
-
-message TableNamespaceTuple {
-  string table = 1;
-  string namespace = 2;
+  int64 sequencer_id = 1;
+  int32 sequencer_index = 2;
```
carols10cents, line 19: Ok with the rename... is this going to be `ShardToShardId` Request/Response? with fields `shard_id` (the postgres catalog id) and `shard_index` (the kafka partition)?

domodwyer: Hmmmm maybe:

```proto
rpc MapToShardId(MapToShardIdRequest) MapToShardIdRequest;
```

so it is clear we're mapping the input to a shard, rather than having shard in the name twice - sorry, I didn't foresee this!

carols10cents: Love it!

PR #5463, opened 2022-08-24
refactor: infallible JumpHash initialisation

domodwyer:
Bit of a tidy-up whilst working on the gRPC ShardService.

---

* refactor: infallible JumpHash initialisation (13e7b93c1)

      This doesn't really need to be fallible but forces propagation of a ton of
      error handling - no shards is always a sign of something being very wrong, and
      can be caught in the caller if it's for some reason an acceptable state / can
      be recovered from.


lukebond (approved):
nice. i recall encountering this and wondering if it was worth being fallible!

PR #5465, opened 2022-08-24
feat(router): sharding lookup gRPC service 

domodwyer:
This PR completes the implementation of a gRPC shard service for external clients to look up the mapping of `(table, namespace)` to a Kafka shard/partition/whatevs.

I used "shard" (instead of "partition"/etc) because of the impending [changes](https://github.com/influxdata/influxdb_iox/pull/5435).

---

* feat: impl Sharder<()> for JumpHash (baacc015d)

      Implement a payload-less Sharder for the JumpHash concrete type. This models
      the input the gRPC ShardService will provide.

* feat: ShardService gRPC endpoint handler (da31f3140)

      Implements the ShardService to expose the shard mapping produced by routers to
      external clients.

      This impl uses an internal cache to eliminate unnecessary Catalog queries, as
      the Kafka partition/Sequencer/Shard mapping is static once a router has
      initialised.

* refactor: impl Sharder for Arc-wrapped Sharder (3594e5d09)

      Allows any Sharder wrapped in an Arc to be used as a Sharder impl, allowing
      the same Sharder instance to be shared across independent code modules.

* feat: expose ShardService over gRPC (7afa3bfae)

      Plumbs in the ShardService impl, and exposes it over the router's gRPC 
      interface.

### DIFF
--- a/ioxd_router/src/lib.rs
+++ b/ioxd_router/src/lib.rs
@@ -26,9 +26,13 @@ use router::{
         metrics::InstrumentedCache, MemoryNamespaceCache, NamespaceCache, ShardedCache,
     },
     sequencer::Sequencer,
-    server::{grpc::GrpcDelegate, http::HttpDelegate, RouterServer},
+    server::{
+        grpc::{sharder::ShardService, GrpcDelegate},
+        http::HttpDelegate,
+        RouterServer,
+    },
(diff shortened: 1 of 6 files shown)


lukebond (approved):
looks good!

PR #5466, opened 2022-08-24
chore: update object_store pin

tustvold:
Updates to the latest master to pick up the azure removal


crepererum:
@tustvold is this still relevant? IIRC we've merged this upgrade already.

PR #5467, opened 2022-08-24
fix: optimizations for metadata fetch and chunk pruning

### LINKED ISSUE #5453: Improve pruning performance of cached values in the catalog

lesam:
Closes #5453

This implements early pruning of chunks (before fully fetching metadata) without special-casing the time range - we use datafusion's predicate analysis twice, once before metadata fetches and again after.

This PR also hoists quite a lot of metadata fetching work from per-chunk to per-table.


querier/src/chunk/mod.rs
```diff
@@ -429,57 +433,25 @@ impl ChunkAdapter {
...
+        column_id_lookup: &HashMap<ColumnId, &str>,
+        namespace_schema: Arc<NamespaceSchema>,
+        table_schema: Arc<Schema>,
+        table_name: Arc<str>,
         parquet_file: Arc<ParquetFile>,
         span: Option<Span>,
     ) -> Option<ChunkParts> {
         let span_recorder = SpanRecorder::new(span);
-
-        // gather schema information
```
lesam, line 438: this part seems mostly independent of the chunk, so I hoisted it out of `chunk_parts`

querier/src/chunk/mod.rs
```diff
@@ -504,18 +476,30 @@ impl ChunkAdapter {
...
+                    Some(name)
+                } else {
+                    None
+                }
+            })
             .map(|s| s.to_owned())
             .collect();
         let schema = self
             .catalog_cache
             .projected_schema()
```
lesam, line 499: Cache read here is per-chunk still

querier/src/chunk/mod.rs
```diff
@@ -429,57 +433,25 @@ impl ChunkAdapter {
...
+        }
+        // relevant_pk_columns is everything from the primary key for the table, that is actually in some parquet file
+        let relevant_pk_columns : Vec<_> = table_schema
             .primary_key()
             .into_iter()
-            .filter(|c| file_columns.contains(c))
+            .filter(|c| column_id_lookup.contains_key(&table_schema_catalog.columns.get(*c).expect("catalog has all columns").id))
             .collect();
         let partition_sort_key = self
             .catalog_cache
```
lesam, line 457: Cache read here is per-chunk still

querier/src/table/mod.rs
```diff
@@ -231,33 +240,84 @@ impl QuerierTable {
...
                 parquet_files,
                 span_recorder.child_span("reconcile"),
             )
             .await
-            .context(StateFusionSnafu)
+            .context(StateFusionSnafu)?;
+
+        let num_initial_chunks = chunks.len();
+        let chunks = self.chunk_pruner().prune_chunks(
+            self.table_name(), Arc::clone(&self.schema), chunks, &predicate).context(ChunkPruningSnafu)?;
```
lesam, line 318: `self.schema` here is weird to me - we look up the schema from the cache above, but we also already have the schema? Need to track this down (or maybe @crepererum knows what's up). I believe this is not a behaviour change, and the pruning was previously using the schema from `self.schema`.

crepererum: I think in reality this won't matter. We may invalidate the cache during chunk building if we find chunks that have new columns that are not yet in the schema. However the predicate and the whole datafusion planning used the "old" schema so these new columns are in theory not used anywhere and will just be ignored. From a technical perspective I agree with you that this is not ideal and I think we should use the recently cached schema.

BTW: I have a PR nearly done that will change the "cached namespace" from having some catalog version/type of a schema to an actual schema (triggered by the memcopies we've seen during profiling). This should make it easier for you to get the latest schema here quickly.

crepererum (commented):
This makes sense to me, but I would like to get a 2nd opinion by one of the other users/authors of `iox_query`. Maybe @tustvold or @NGA-TRAN ?

NGA-TRAN (dismissed):
I did not look carefully in the changes in the Querier because @crepererum was probably did it carefully but I did review the Table Provider related changes that will be invoked by both Ingester and Compactor. What I understand is the change is trying to remove the special chunk pruning for the grpc queries (queries that came from Influx and not normal SQL), right @lesam? I think it will be ok  because Ingester and Compactor would not have to worry about those special cases. 
@alamb is the one who wrote that pruning code and may have better insight

lesam:
> special chunk pruning for the grpc queries

I think this already got removed (maybe by @crepererum?), but it means we get poor performance for small time range queries since we load all the chunk metadata for every chunk instead of pruning early.

This brings back the early pruning behaviour without special-casing the time range - we use datafusion's predicate analysis twice, once before metadata fetches and again after.

lesam:
This PR also hoists quite a lot of metadata fetching work from per-chunk to per-table.

NGA-TRAN:
Thanks Sam. I see these 2 in the PR and they make sense:

> This brings back the early pruning behaviour without special-casing the time range - we use datafusion's predicate analysis twice, once before metadata fetches and again after.

> This PR also hoists quite a lot of metadata fetching work from per-chunk to per-table.

lesam:
This is now good enough to come out of draft - tests pass, and it is mostly a refactor of existing pieces.

querier/src/table/mod.rs
```diff
@@ -549,7 +649,7 @@ mod tests {
         // this contains all files except for:
         // - file111: marked for delete
         // - file221: wrong table
-        let pred = Predicate::new().with_range(0, 100);
+        let pred = Predicate::new().with_range(0, 101);
```
lesam, line 652: Required because we now actually prune the chunks, so the file with mintime == maxtime == 100 (file 122) is pruned out due to predicate having exclusive end time.

querier/src/namespace/query_access.rs
```diff
@@ -73,11 +73,7 @@ impl QueryDatabase for QuerierNamespace {
                     .any(|col| schema.find_index_of(col).is_some())
             })
         }
-
-        let pruner = table.chunk_pruner();
```
lesam, line 77: Pruning is now done in `table.chunks` above.

querier/src/table/mod.rs
```diff
@@ -231,33 +238,126 @@ impl QuerierTable {
...
+            .iter()
+            .flat_map(|cached_file| cached_file.column_set.iter().copied())
+            .collect();
+        let namespace_schema = self
+            .chunk_adapter
+            .catalog_cache()
+            .namespace()
+            .schema(
+                Arc::clone(&self.namespace_name),
+                &[(&self.table_name, &columns)],
```
lesam, line 252: We used to look up the table_name from the table id in a cache lookup - this was actually the only use of that cache, so I've removed the cache entirely.

querier/src/cache/partition.rs
```diff
@@ -123,17 +119,13 @@ impl PartitionCache {
     pub async fn sort_key(
         &self,
         partition_id: PartitionId,
-        should_cover: &HashSet<&str>,
+        should_cover: &Vec<&str>,
```
lesam, line 122: We don't need this to be a `HashSet` and a `Vec` should be cheaper for callers to construct

querier/src/table/mod.rs
crepererum: Could you add another file that is pruned, so the tests reflect the actual interface?

querier/src/cache/partition.rs
crepererum: I'm a bit surprised that clippy isn't yelling at you: Passing a `&Vec<...>` (immutable reference) in Rust is unidiomatic. You shall pass `&[...]` (`&[&str]` in this case) instead, i.e. a slice (since this is basically what an immutable vector is apart from the capacity that ain't important here). Callsides then can either pass in a vector (`cache.sort_key(id, &vec[..])`) or can easily construct an empty slide out of thin air (`cache.sort_key(id, &[])`).

querier/src/chunk/mod.rs
```diff
@@ -505,17 +479,30 @@ impl ChunkAdapter {
...
-            .map(|(_t, field)| field.name().as_str())
-            .filter(|col| file_columns.contains(*col))
+            .filter_map(|(_, field)| {
+                let name = field.name().as_str();
+                let column_id = match table_schema_catalog.columns.get(name) {
+                    Some(col) => col.id,
+                    None => return None,
+                };
+                // TODO: for large column sets constructing a hash map would be better,
+                // but for small ones the allocation overhead kills us
```
crepererum, line 491: Is this is a TODO for this PR or for future work? In the latter case, either self-assign via `TODO(sam): ` (or whatever short name you wanna use) or create a ticket and link it here.

crepererum (commented):
Looks good, just some minor nits. Big thanks for cleaning up the caching! :)

querier/src/table/mod.rs
lesam: Done

querier/src/cache/partition.rs
lesam: Done

crepererum (approved):

alamb:
I skimmed this PR, for what it is worth, and it looks pretty good. Thanks for the clean ups @lesam  The 'NoOp' pruning stuff that was removed in this PR was related to pruning chunks in SQL queries where as I recall DataFusion needed the table provider prior to knowing what predicates would be pushed down.

So as long as we are still pruning chunks we are all good.

I reviewed how the Querier works now, and I think everything looks copacetic (pruning happens at the right place)
https://github.com/influxdata/influxdb_iox/blob/main/querier/src/table/query_access.rs#L48-L69

👍

Issue #5443, opened 2022-08-22
InfluxQL: Teach IOx how to parse an InfluxQL conditional expression


stuartcarnie:
Dependent on #5440

In order to implement a `WHERE` clause, we must teach IOx how to parse a conditional expression. Conditional expressions is a binary expression using a conditional operator.

### Precedence rules for conditional operators

NOTE: see #5440 for the arithmetic operator precedence table

| Precedence | Operator | Description           | Associativity |
| ---------- | -------- | --------------------- | ------------- |
| 4          | <        | Less than             | Left → Right  |
| 4          | <=       | Less than or equal    | Left → Right  |
| 4          | >        | Greater that          | Left → Right  |
| 4          | >=       | Greater than or equal | Left → Right  |
| 4          | =        | Equal                 | Left → Right  |
| 4          | !=       | Not equal             | Left → Right  |
| 4          | =~       | Regex match           | Left → Right  |
| 4          | !~       | Regex not match       | Left → Right  |
| 4          | IN       | IN list expression       | Left → Right  |
| 5          | AND      | Logical AND           | Left → Right  |
| 6          | OR       | Logical OR            | Left → Right  |

Notable links:

* Precedence rules for binary operators [are codified in source](https://github.com/influxdata/influxql/blob/aa24a6e58d0adc02cee92ec11008d48bb6da0ac8/token.go#L291-L306)

stuartcarnie:
Closed by #5469

PR #5469, opened 2022-08-25
feat: Implementation of expression parsing for InfluxQL

### LINKED ISSUE #5440: InfluxQL: Teach IOx how to parse an InfluxQL expression
Dependent on #5444 

In order to implement a `WHERE` clause, we must first teach IOx how to parse an expression. Expressions may appear in numerous places including the left and right side of conditional expressions, select projections, etc.

Implement an InfluxQL expression parser function to generate an expression AST. This function will be combined with other parsers, such as a conditional expression.

See [InfluxQL expressions](https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#expressions) and the [`ParseExpr`](https://github.com/influxdata/influxql/blob/7e7d61973256ffeef4b99edd0a89f18a9e52fa2d/parser.go#L58-L59) function for documentation.

### Precedence rules for operators

| Precedence  | Operator | Description    | Associativity |
| ----------- | -------- | -------------- | ------------- |
| 1 (highest) | -        | Unary minus    | Right → Left  |
| 1           | +        | Unary plus     | Right → Left  |
| 2           | *        | Multiplication | Left → Right  |
| 2           | /        | Division       | Left → Right  |
| 2           | %        | Modulus        | Left → Right  |
| 2           | &        | Bitwise AND    | Left → Right  |
| 3           | +        | Addition       | Left → Right  |
| 3           | -        | Subtraction    | Left → Right  |
| 3           | \|       | Bitwise OR     | Left → Right  |
| 3           | ^        | Bitwise XOR    | Left → Right  |

#### Notable links

* Precedence rules for binary operators [are codified in source](https://github.com/influxdata/influxql/blob/aa24a6e58d0adc02cee92ec11008d48bb6da0ac8/token.go#L291-L306)

stuartcarnie:
Closes #5440 and #5443

This PR teaches the `influxdb_influxql_parser` crate how to process InfluxQL general and conditional expressions that would appear in the `WHERE` clause of various InfluxQL statements.

Some notable changes in this PR include

* Added missing keywords `AND` and `OR` to the `sql_keyword` parser
* Removed regular expression matching from `literal` parser, as regular expression literals are only valid after the regular expression conditional operators `=~` and `!~`
* Added `literal_regex` parser that only matches regular expression literals
* Added a `BindParameter` type and combinator, to match [bind parameters](https://docs.influxdata.com/influxdb/v1.8/tools/api/#bind-parameters)


influxdb_influxql_parser/src/parameter.rs
```diff
@@ -0,0 +1,116 @@
...
+    map(
+        recognize(many1_count(alt((alphanumeric1, tag("_"))))),
+        str::to_string,
+    )(i)
+}
+
+/// `BindParameter` is a type that represents either a quoted ([`BindParameter::Quoted`]) or unquoted ([`BindParameter::Unquoted`])
+/// InfluxQL bind parameter.
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub enum BindParameter {
```
crepererum, line 35: Are these two cases semantically different? If not, shall they be a single case and `Display` should decide about the quoting depending if the identifier contains a space? This wouldn't "roundtrip" though (not sure if that's a requirement).

influxdb_influxql_parser/src/expression.rs
```diff
@@ -0,0 +1,550 @@
...
+        ($LHS: expr, $OP: ident, $RHS: expr) => {
+            <Expr as std::convert::Into<Box<Expr>>>::into(Expr::BinaryOp {
+                lhs: $LHS.into(),
+                op: BinaryOperator::$OP,
+                rhs: $RHS.into(),
+            })
+        };
+    }
+
+    #[test]
```
crepererum, line 371: Parser testing is a bit of a rabbit hole, esp. when it comes to more complex expressions. So this is more a loose brain dump on that topic instead of a concrete advice:

- **current state:** I think what you do here is basically "text -> AST" (the latter build via macros because it's really cumbersome to manually construct the AST) and "AST -> text"
- **snapshots:** Another approach that I've seen in the wild is "snapshot testing" (e.g. via [insta](https://github.com/mitsuhiko/insta)). This is useful for "text -> AST" but instead of manually constructing it you just "approve" it by comparing a pretty-printed version against a snapshot (see insta docs for an example).
- **fuzzing:** to find really ugly edge cases in the "text -> AST -> text" roundtrip, a fuzzer could be used to generate text, than parse it to AST, and -- if that's successful -- format it into text again, parse it a 2nd time and compare the two ASTs. Alternatively we could start with fuzz-gen the AST, format it, parse it again and compare.

influxdb_influxql_parser/src/expression.rs
```diff
@@ -0,0 +1,550 @@
...
+        assert_eq!(got, *binary_op!(5, Add, param!("foo")));
+
+        // Following two tests validate that operators of higher precedence
+        // are nested deeper in the AST.
+
+        let (_, got) = conditional_expression("5 % -3 | 2").unwrap();
+        assert_eq!(
+            got,
+            *binary_op!(binary_op!(5, Mod, unary!(-3)), BitwiseOr, 2)
+        );
```
crepererum, line 386: Can we test the "flipped" versions of these as well (i.e. `-3 | 2 % 5`, `5 % 2 | -3`, `2 | -3 % 5`) just to make sure that the precedence isn't just a nice side effect of the order?

Edit: this is partly covered by the last successful case (`2 | 5 % 3`), so maybe use that and enumerate all possible "flips".

influxdb_influxql_parser/src/expression.rs
```diff
@@ -0,0 +1,550 @@
...
+        ),
+        map(cut(preceded(multispace0, literal_regex)), From::from),
+    )))(input)?;
+    Ok((input, reduce_expr(f1, exprs)))
+}
+
+/// Parse conditional operators.
+fn conditional(i: &str) -> IResult<&str, Expr> {
+    let (input, f1) = conditional_regex(i)?;
+    let (input, exprs) = many0(tuple((
```
crepererum, line 256: I'll have another look at these tomorrow morning (Berlin time) to really figure out how operator precedence is implemented. I guess I need to understand a few NOM basics to do that.

crepererum (approved):
A few minor questions and recommendations, nothing major (although I still don't fully understand NOM).

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Your current state view is correct 👍🏻 

I plan to evaluate what tests in the existing Go code base are useful here also.

I like the look of snapshots and *insta* – great suggestion! As part of my checkpoint, after completing a single statement, I can look at adding something like insta.

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Will do 👍🏻

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Operator precedence is implemented like a top-down recursive decent parser. Lower-priority combinators call higher-priority ones, until we reach operands (literals, identifiers, etc)

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Done – added all variations. And my implementation of the expected values worked first time in each case, which is a good sign 🥳

influxdb_influxql_parser/src/expression.rs
stuartcarnie: I initially tried using a [crate](https://github.com/mullr/nom-7-precedence) that ports experimental precedence combinators back to nom v7, but it is limited such that you cannot, for example, specify different combinators for different operators. That is, all binary operations must all follow the pattern

```
expr = expr "op1" expr
expr = expr "op2" expr

etc
```

whereas, we need

```
expr = expr "IN" "(" literal ("," literal)* ")"
expr = expr "=~" regex_literal
expr = expr "+" expr

and so on
```

influxdb_influxql_parser/src/parameter.rs
stuartcarnie: I have created #5538 to simplify the `Identifier` and `BindParameter` elements

PR #5470, opened 2022-08-25
feat: Renaming kafka topic types

carols10cents:
Follow up to https://github.com/influxdata/influxdb_iox/pull/5435. Blocks https://github.com/influxdata/influxdb_iox/pull/5475.

This PR:

- Renames `KafkaTopic`, the struct that holds a Kafka topic name and its catalog ID, to `TopicMetadata`
- Renames `KafkaTopicId`, the wrapper type for the Kafka topic catalog IDs, to `TopicId`
- Renames parameters named `kafka_topic_name` that held the Kafka topic name string to `topic_name`
- Renames the catalog `kafka_topics` method to `topics`
- Removes `kafka` from local variable names, constants, test names, error messages

This is still only in the code; this isn't the big scary database rename PR yet.

There are also still some metric names and log messages that have "kafka" in them; I'm happy to take care of those too but I don't think those contribute to confusion or break encapsulation in harmful ways.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -495,14 +495,14 @@ mod tests {
         // empty partition
         let mut txn = catalog.catalog.start_transaction().await.unwrap();
 
-        let kafka = txn.kafka_topics().create_or_get("foo").await.unwrap();
+        let topic = txn.topics().create_or_get("foo").await.unwrap();
         let pool = txn.query_pools().create_or_get("foo").await.unwrap();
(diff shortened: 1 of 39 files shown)


NGA-TRAN (approved):
I am glad we are now able to use `topic` only and remove `kaflka` to have a definition in IOx what a `topic` it is without relying on an external mapping (even though there is a one-one-one mapping with external thingy)

PR #5471, opened 2022-08-25
feat: More accurate memory estimate for compaction

### LINKED ISSUE #5457: More accurate resource allocation for compactor
# Rationale

Currently, we use a lot of config params defined in this [CompactorConfig](https://github.com/influxdata/influxdb_iox/blob/3220c6f88bbacc5e567bed4d9fc2b50f68ebb42e/compactor/src/handler.rs#L89) to identify how many files and max size to compact per compaction and how many of them can run in parallel. This is suboptimal because it does not consider that different tables need deferent resources. The real needed resource depends on how many columns and rows in the table, their column types, and cardinality. If the system happens to have one table that needs a lot of resource and the rest needs little, to compactor today allocates the same large resource for each of them which is very wasteful.

This ticket is to estimate the needed resource based on each table and their data. Here are the formula.

- **compactor_memory_budget**: config param and let say 30GB. We actually have 55GB but we wand to leave plenty of room for error.
- Each compaction should only compact files whose total size should not go over the `compactor_memory_budget`
- **estimated_arrow_bytes_for_file** = value_bytes + dictionary_key_bytes + dictionary_value_bytes + string_bytes + bool_bytes
- **value_bytes** = (float_column_count + int_column_count + time_column_count + uint_count) * row_count * 8
- **average_tag_value_length** = 200
- **average_cardinality** = row_count / 2
- **dictionary_key_bytes** = tag_column_count * average_cardinality * average_tag_value_length
- **dictionary_value_bytes** = tag_column_count * row_count * 8
- **string_bytes** = string_column_count * 1000
- **bool_bytes** = bool_column_count * 1

If all files of a candidate are added and still under budget, we will compact other candidates concurrently computed as above on the leftover budget

NGA-TRAN:
Closes #5457 

@carols10cents and @pauldix : I implemented the estimation for the hot partitions first. After I get all feedback from you, I will do the cold partitions because they are similar.


compactor/src/handler.rs
```diff
@@ -357,49 +391,7 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
-    // Concurrency level calculation (this is estimated from previous experiments. The actual
-    // resource management will be more complicated and a future feature):
-    //   . Each `compact partititon` takes max of this much memory input_size_threshold_bytes
-    //   . We have this memory budget: max_concurrent_size_bytes
-    // --> num_parallel_partitions = max_concurrent_size_bytes/
-    //     input_size_threshold_bytes
-    let num_parallel_partitions = (compactor.config.max_concurrent_size_bytes
-        / compactor.config.input_size_threshold_bytes) as usize;
-
-    futures::stream::iter(candidates)
```
NGA-TRAN, line 370: Since  number of parallel candidates have to be determined by their  estimated budgets, we do not know how many  candidates we can compact in parallel until we accumulate enough budget. Hence this future cannot be used here. I wrote function `compact_hot_partitions_in_parallel` to do this job and let @carols10cents provide feedback

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
     n_candidates
 }
 
+// For a given list of hot partition candidates and a memory budget, compute memory needed to compact each one
+// and compact as many of them in parallel as possible until all candidates are compacted
+async fn compact_hot_partition_candidates(
```
NGA-TRAN, line 411: This is the kind of the start of the major change in this PR. Other changes will go from this function or its children

compactor/src/handler.rs
```diff
@@ -328,14 +351,25 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
         duration.record(delta);
     }
 
-    // Add other compaction-needed info into selected partitions
+    // Get extra needed information for selected partitions
     let start_time = compactor.time_provider.now();
+
+    // Column types and their counts of the tables of the partition candidates
```
NGA-TRAN, line 357: This is needed to estimate the memory needed for a file

compactor/src/lib.rs
```diff
@@ -48,29 +49,14 @@ pub(crate) enum Error {
...
-        .await
-        .context(LookupSnafu)?;
-
-    let to_compact = parquet_file_filtering::filter_hot_parquet_files(
-        parquet_files_for_compaction,
-        compactor.config.input_size_threshold_bytes(),
-        compactor.config.input_file_count_threshold(),
-        &compactor.parquet_file_candidate_gauge,
-        &compactor.parquet_file_candidate_bytes,
-    );
```
NGA-TRAN, line 70: These 2 functions were moved out of this function to estimate the budget.

compactor/src/parquet_file_filtering.rs
```diff
@@ -1,44 +1,145 @@
...
-    // far exceeds this value
-    input_file_count_threshold: usize,
+    // column types and their counts of the table of this partition
+    column_types: &[ColumnTypeCount],
     // Gauge for the number of Parquet file candidates
     parquet_file_candidate_gauge: &Metric<U64Gauge>,
     // Histogram for the number of bytes of Parquet file candidates
     parquet_file_candidate_bytes: &Metric<U64Histogram>,
-) -> Vec<ParquetFile> {
+) -> FilteredFiles {
```
NGA-TRAN, line 134: There is a lot of changes in this function since we change the way to estimate needed memory

compactor/src/parquet_file_filtering.rs
```diff
@@ -600,144 +808,29 @@ mod tests {
                 ],
             };
 
-            // Max size 0; only the first level 0 file and its overlapping level 1 files get
-            // returned
-            let max_size = 0;
+            // total needed budget for one file with a tag, a time and 11 rows = 1176
             let (files_metric, bytes_metric) = metrics();
-            let files = filter_hot_parquet_files(
```
NGA-TRAN, line 607: git does not do a good job in comparing the diff here but in general, 2 kinds of tests are removed:
1. Tests that always return first L0 and its overlapped L1s. This is no longer valid
2. Tests that test file counts because we no longer need it.

NGA-TRAN:
@carols10cents and @pauldix : I implemented the estimation for the hot partitions first. After I get all feedback from you, I will do the cold partitions because they are similar.

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
+        // 5. Repeat
+
+        // --------------------------------------------------------------------
+        // 1. Remove first candidate from the list
+        let partition = candidates.remove(0);
+        let partition_id = partition.candidate.partition_id;
+        let table_id = partition.candidate.table_id;
+
+        // Get column types and their counts for the table of the partition
+        let columns = table_columns.get(&partition.table_id());
```
carols10cents, line 439: Can this be

```suggestion
        let columns = table_columns.get(&table_id);
```

?

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
+            }
+            FilterResult::Proceeed => {
+                remaining_budget_bytes -= to_compact.budget_bytes();
+                parallel_compacting_candidates.push(to_compact);
+            }
+        }
+
+        // --------------------------------------------------------------------
+        // 4. Almost hitting max budget or no more candidates, let compact parallel_compacting_candidates
+        if (remaining_budget_bytes < (compactor.config.memory_budget_bytes as f64 * 0.5) as u64)
```
carols10cents, line 523: Floating point can be expensive... because this is just trying to get 50%, can this be:

```suggestion
        if (remaining_budget_bytes < (compactor.config.memory_budget_bytes / 2) as u64)
```

and if we truncate off a byte, that should be ok?

carols10cents (commented):
I haven't gotten all the way through, but I have to go... will finish later :)

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
 
+// For a given list of hot partition candidates and a memory budget, compute memory needed to compact each one
+// and compact as many of them in parallel as possible until all candidates are compacted
+async fn compact_hot_partition_candidates(
+    compactor: Arc<Compactor>,
+    mut candidates: Vec<PartitionCompactionCandidateWithInfo>,
+    table_columns: HashMap<TableId, Vec<ColumnTypeCount>>,
+) {
+    let mut remaining_budget_bytes = compactor.config.memory_budget_bytes;
+    let mut parallel_compacting_candidates = vec![];
```
carols10cents, line 417: This should use `Vec::with_capacity`... I think this is the right capacity?

```suggestion
    let mut parallel_compacting_candidates = Vec::with_capacity(candidates.len());
```

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
+        //    Otherwise:
+        //      . if the remaining budget is not the full budget, push the candidate back into the `candidates`
+        //        to consider compacting with larger budget.
+        //      . otherwise, log the partition that it is too large to even compact 2 files
+        // 4. If hit the budget, compact all canidates in the compacting_list in parallel
+        // 5. Repeat
+
+        // --------------------------------------------------------------------
+        // 1. Remove first candidate from the list
+        let partition = candidates.remove(0);
```
carols10cents, line 434: I wonder if we should convert `candidates` into a `VecDeque` either in this method or elsewhere, since this code is popping from the front and pushing to the back of the list... Something to consider profiling, anyway.

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
+            &compactor.parquet_file_candidate_bytes,
+        );
+
+        // --------------------------------------------------------------------
+        // 3. Check the compactable status and act provide the right action
+        match to_compact.filter_result() {
+            FilterResult::NothingToCompact => {
+                continue;
+            }
+            FilterResult::ErrorEstimatingBudget => {
```
carols10cents, line 489: This feels a little off to have an enum variant that's named `Error` rather than having `filter_hot_parquet_files` return a `Result` and communicate the error that way? `OverBudget` sounds kind of like an error condition too?

And `NothingToCompact` sounds like `None`, so maybe `filter_hot_parquet_files` should be returning `Result<Option<FilteredFiles>, Error>` rather than `FilteredFiles` having a `FilteredFiles` having a `filter_result` method that computes this state on-the-fly ... what do you think?

compactor/src/handler.rs
```diff
@@ -414,6 +406,167 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
+
+        // --------------------------------------------------------------------
+        // 4. Almost hitting max budget or no more candidates, let compact parallel_compacting_candidates
+        if (remaining_budget_bytes < (compactor.config.memory_budget_bytes as f64 * 0.5) as u64)
+            || (candidates.is_empty() && !parallel_compacting_candidates.is_empty())
+        {
+            let partitions = parallel_compacting_candidates
+                .drain(0..)
+                .collect::<Vec<_>>();
+            compact_hot_partitions_in_parallel(Arc::clone(&compactor), partitions).await;
```
carols10cents, line 529: I was confused about what this `drain` `collect` was doing because it looked like a no-op, then I tried removing it and saw that this is emptying `parallel_compacting_candidates` so that it will be empty for the next iteration of the candidate loop. I think this would be clearer and possibly faster as:

```suggestion
            compact_hot_partitions_in_parallel(Arc::clone(&compactor), parallel_compacting_candidates).await;
            parallel_compacting_candidates = Vec::with_capacity(candidates.len());
```

The ownership of the `Vec` in `parallel_compacting_candidates` gets transferred to `compact_hot_partitions_in_parallel`, but the binding gets reinitialized before it's used again.

compactor/src/parquet_file_filtering.rs
```diff
@@ -52,49 +153,84 @@ pub(crate) fn filter_hot_parquet_files(
...
     // Keep track of level 0 files to include in this compaction operation; maintain assumed
     // ordering by max sequence number.
     let mut level_0_to_return = Vec::with_capacity(level_0.len());
-    // Running total of the size, in bytes, of level 0 files and level 1 files for inclusion.
-    // The levels are counted up separately for metrics but added together when comparing with the
-    // max_size.
-    let mut total_level_0_bytes = 0;
-    let mut total_level_1_bytes = 0;
+    // Estimated memory bytes needed to compact returned L0 files
+    let mut l0_estimated_dudget = Vec::with_capacity(level_0.len());
```
carols10cents, line 162: Typo in variable name:

```suggestion
    let mut l0_estimated_budget = Vec::with_capacity(level_0.len());
```

carols10cents (commented):
Next batch of review done, need another break 😅 Namely, I haven't reviewed the tests yet.

compactor/src/compact.rs
```diff
@@ -356,6 +361,28 @@ impl Compactor {
...
+            let cols = repos
+                .columns()
+                .list_type_count_by_table_id(table_id)
+                .await
+                .context(QueryingColumnSnafu)?;
+            result.insert(table_id, cols);
+        }
+
+        Ok(result)
+    }
```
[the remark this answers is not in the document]
NGA-TRAN: Good catch on `avoid querying the catalog for the same table multiple times`

compactor/src/handler.rs
NGA-TRAN: Good suggestion. Let me try

compactor/src/handler.rs
NGA-TRAN: Yeah, they are the same

compactor/src/handler.rs
NGA-TRAN: Done

compactor/src/handler.rs
NGA-TRAN: @carols10cents 
The `OverBudget` is actually not an error because we still have a possibility to recompact it with larger budget (that is the reason for pop and push). With that and after some thinking, I think it will clearer to have 4 choices in the `FilterResult`  enum rather than splitting error and nothing out

NGA-TRAN:
@carols10cents The PR is ready for a second look

compactor/src/handler.rs
```diff
@@ -357,49 +394,7 @@ async fn compact_hot_partitions(compactor: Arc<Compactor>) -> usize {
...
 
-    // Repeat compacting n partitions in parallel until all candidates are compacted.
-    // Concurrency level calculation (this is estimated from previous experiments. The actual
-    // resource management will be more complicated and a future feature):
-    //   . Each `compact partititon` takes max of this much memory input_size_threshold_bytes
-    //   . We have this memory budget: max_concurrent_size_bytes
-    // --> num_parallel_partitions = max_concurrent_size_bytes/
-    //     input_size_threshold_bytes
-    let num_parallel_partitions = (compactor.config.max_concurrent_size_bytes
-        / compactor.config.input_size_threshold_bytes) as usize;
```
carols10cents, line 368: It doesn't look like `max_concurrent_size_bytes` and `input_size_threshold_bytes` are used anywhere anymore, can they be removed or is there a planned future PR that will use them again?

compactor/src/lib.rs
```diff
@@ -48,29 +49,15 @@ pub(crate) enum Error {
...
-            Arc::clone(&compactor.catalog),
-            partition.id(),
-        )
-        .await
-        .context(LookupSnafu)?;
-
-    let to_compact = parquet_file_filtering::filter_hot_parquet_files(
-        parquet_files_for_compaction,
-        compactor.config.input_size_threshold_bytes(),
-        compactor.config.input_file_count_threshold(),
```
carols10cents, line 67: It looks like `input_file_count_threshold` isn't used anymore either, can it be removed or is there a planned future PR that will use it again?

carols10cents (commented):
Some cleanups maybe, a unit test, and some testing thoughts maybe for the future. Getting close!

compactor/src/lib.rs
NGA-TRAN: This is still used in the cold partition. And yes, after the cold partitions are done, we can move a few params

compactor/src/handler.rs
NGA-TRAN: I am thinking after this PR, I will sync with you to see who will be better to do the cold partition. Either

- I will do the cold similar to this hot or
- You will do the cold by putting this estimate together with the cold PR you are waiting for this PR to merge; or something like that.

How about I let you do this in the follow-up PRs which I think will make more sense with the cold work you will do next, too?

compactor/src/parquet_file_filtering.rs
```diff
@@ -1,44 +1,141 @@
...
 
+const AVERAGE_TAG_VALUE_LENGTH: i64 = 200;
+const STRING_LENGTH: i64 = 1000;
+const DICTIONARY_BYTE: i64 = 8;
+const VALUE_BYTE: i64 = 8;
+const BOOL_BYTE: i64 = 1;
+const AVERAGE_ROW_COUNT_CARDINALITY_RATIO: i64 = 2;
+
+type Error = Box<dyn std::error::Error>;
+fn estimate_arrow_bytes_for_file(
```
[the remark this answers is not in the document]
NGA-TRAN: Good idea, working on it

compactor/src/handler.rs
NGA-TRAN: Done

compactor/src/lib.rs
NGA-TRAN: Actually, you are right. I am removing it

compactor/src/parquet_file_filtering.rs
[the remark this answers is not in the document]
NGA-TRAN: Done

NGA-TRAN:
@carols10cents Can you have a quick look? I have addressed all the comments

carols10cents (approved):
Thank you!!!

compactor/src/handler.rs
NGA-TRAN: Hah, I hit a bug that I need to test this function. Let me try to figure out how to convert  `filter_hot_parquet_files` to closure as you describe

PR #5473, opened 2022-08-26
feat: Implement gRPC binary logger

mkmik:
# Rationale
I n a few cases we wanted to debug some bad query and we wasted hours to change some code and add some logging and wait until it gets deployed etc. Using this functionality (which is standard in the official gRPC SDKs for other languages) we could more easily find the offending query and replay it locally.

# Changes
This PR implements a tower middleware that captures and logs gRPC frames according to https://github.com/grpc/grpc/blob/master/doc/binary-logging.md


alamb:
I plan to review this later today

grpc-binary-logger/src/middleware.rs
```diff
@@ -0,0 +1,532 @@
...
+                    method: uri.path(),
+                    authority: uri.authority(),
+                    headers: request.headers(),
+                });
+
+                // wrap our logger around the request stream.
+                let request = Self::logged_request(request, call.clone());
+
+                // perform the actual request.
+                let response = inner.call(request).await?;
```
alamb, line 171: By using `?` here there will be nothing logged if the inner returns an error. Shouldn't the (error) response be logged if `inner.call` returns an error?

grpc-binary-logger/src/middleware.rs
```diff
@@ -0,0 +1,532 @@
...
+                        // `data` returns an actual gRPC frames, even if it has been sent
+                        // over multiple tcp segments and over multiple HTTP/2 DATA frames.
+
+                        // TODO(mkm): figure out why the client produces a zero length chunk here.
+                        // Ignoring it seems to be the right thing to do.
+                        if !buf.is_empty() {
+                            call.log(LogEntry::ClientMessage(&buf));
+                        }
+                        if sender.send_data(buf).await.is_err() {
+                            sender.abort();
```
alamb, line 210: Is this something we should be logging too (that there was some error sending data to the client?)

grpc-binary-logger/src/middleware.rs
```diff
@@ -0,0 +1,532 @@
...
+                            return;
+                        }
+                    }
+                    Err(_err) => {
+                        sender.abort();
```
alamb, line 215: Here also it seems like not logging `_err` may be obscuring valuable debugging information

alamb (approved):
Looks like a very neat bit of code to me. Nice work 👍

grpc-binary-logger/examples/server.rs
```diff
@@ -0,0 +1,75 @@
+use futures::Stream;
```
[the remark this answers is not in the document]
mkmik: I intentionally wanted the example to not be unnecessarily complicated because of sharing code with e2e test.

But also I didn't want to have yet another .proto file just for the example (since examples can't have their own `build.rs`; so I settled on this trade-off.

grpc-binary-logger/src/sink.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl<W> Sink for FileSink<W>
+where
+    W: io::Write + Send,
+{
+    type Error = ();
+
+    fn write(&self, data: GrpcLogEntry, _error_logger: impl ErrorLogger<Self::Error>) {
+        if let Err(e) = self.write_log_entry(&data) {
+            eprintln!("error writing binary log: {:?}", e);
```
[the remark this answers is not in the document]
mkmik: oops; yes! this predated me adding the error logger; fixed

grpc-binary-logger/src/sink.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Receives [`GrpcLogEntry`] entries capturing all gRPC frames from a [`crate::BinaryLoggerLayer`].
+pub trait Sink: Clone + Send + Sync {
+    /// The type returned in the event of an error.
+    type Error;
+
+    /// The sink receives a [`GrpcLogEntry`] message for every gRPC frame captured by a [`crate::BinaryLoggerLayer`].
+    /// The sink owns the log entry and is encourage to process the log in the background without blocking the logger layer.
+    /// Errors should be handled (e.g. logged) by the sink.
+    fn write(&self, data: GrpcLogEntry, error_logger: impl ErrorLogger<Self::Error>);
```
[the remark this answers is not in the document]
mkmik: I thought about it but then if we want the sink to be async we have to turn this function into an async function, which means the sender will have to spawn a task, which must run in the executor chosen by the middleware. If the sink wants to run in another executor it will have to be more complicated because the sink will have to send the request to another executor, then thread back the error in the async function so it can be logged by the middleware.

I'm trading a little bit of "callback" style API ugliness here in order to make concrete Sink implementations easier. My goal here is to make this binary logging capture as decoupled and best effort as possible; it should have the smallest as possible influence onthe main grpc handler behaviour.

grpc-binary-logger/src/middleware.rs
```diff
@@ -0,0 +1,532 @@
...
+pub struct BinaryLoggerMiddleware<S, K, P, L>
+where
+    K: Sink + Send + Sync,
+    L: ErrorLogger<K::Error>,
+{
+    sink: Arc<K>,
+    inner: S,
+    predicate: P,
+    error_logger: L,
+    next_call_id: Arc<Mutex<u64>>,
```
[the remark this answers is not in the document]
mkmik: PTAL

grpc-binary-logger/src/sink.rs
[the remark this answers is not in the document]
alamb: I agree that this way allows the sink implementation the most flexibility in terms of how it is implemented. 👍

grpc-binary-logger/src/middleware.rs
```diff
@@ -119,9 +121,7 @@ where
     }
 
     fn next_call_id(&self) -> u64 {
-        let mut counter = self.next_call_id.lock().unwrap();
-        *counter += 1;
-        *counter
+        self.next_call_id.fetch_add(1, Ordering::SeqCst) + 1
```
alamb, line 124: 👌

grpc-binary-logger/src/middleware.rs
```diff
@@ -342,17 +342,13 @@ where
...
             error_logger,
         }
     }
     fn log(&self, entry: LogEntry<'_>) {
-        let sequence_id_within_call = {
-            let mut seq = self.sequence.lock().unwrap();
-            *seq += 1;
-            *seq
-        };
+        let sequence_id_within_call = self.sequence.fetch_add(1, Ordering::SeqCst) + 1;
```
alamb, line 351: I think this doesn't need the `+1` as that is done as part of `fetch_add`:

grpc-binary-logger/src/middleware.rs
alamb: Likewise I don't think this needs the `+1` as it is already handled inside of `fetch_add`

alamb (approved):
I think the +1 leaves gaps in the call ids, but otherwise the switch to AtomicUSize looks good to me

grpc-binary-logger/src/middleware.rs
mkmik: Fetch_add returns the previous value. Sequence numbers and call IDs start counting from 1. I could either initialize them from 1 or add 1 to all values

alamb: Ah, I think initializing from 1 might make the intent of the code clearer. This is fine too

grpc-binary-logger/src/middleware.rs
mkmik: that's why code reviews are for! "The ~customer~ reader is always right"

grpc-binary-logger/src/middleware.rs
mkmik: good question; I don't know which kind of low level transport errors would result in the call returning an error.

Normal errors (a handler returning an error) are captured in a `Ok(Response)`.

Added a comment with TODO

grpc-binary-logger/src/middleware.rs
mkmik: similarly here; I can't find a way to induce this kind of error in the Go gRPC framework. It seems that  the Go gRPC framework seems to either log a frame if successful or just not log it.

Documented the behaviour in the code.

grpc-binary-logger/src/middleware.rs
mkmik: same here

grpc-binary-logger/src/middleware.rs
mkmik: that's what code reviews are for! "The customer reader is always right"

PR #5475, opened 2022-08-26
feat: database rename

carols10cents:
Depends on https://github.com/influxdata/influxdb_iox/pull/5435 and https://github.com/influxdata/influxdb_iox/pull/5470

** might cause downtime/errors!!! **

### DIFF
--- a/iox_catalog/migrations/20220819154515_sequencer_to_shard.sql
+++ b/iox_catalog/migrations/20220819154515_sequencer_to_shard.sql
@@ -1,0 +1,17 @@
+-- Rename sequencer to shard
+ALTER TABLE IF EXISTS sequencer RENAME TO shard;
+ALTER TABLE IF EXISTS shard RENAME CONSTRAINT sequencer_unique TO shard_unique;
+ALTER TABLE IF EXISTS partition RENAME sequencer_id TO shard_id;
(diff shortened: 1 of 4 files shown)


iox_catalog/migrations/20220819154515_sequencer_to_shard.sql
```diff
@@ -0,0 +1,17 @@
+-- Rename sequencer to shard
+ALTER TABLE IF EXISTS sequencer RENAME TO shard;
+ALTER TABLE IF EXISTS shard RENAME CONSTRAINT sequencer_unique TO shard_unique;
+ALTER TABLE IF EXISTS PARTITION RENAME sequencer_id TO shard_id;
```
mkmik, line 4: for uniformity perhaps lowercase `partition` ?

carols10cents: Will do-- was copypastaing from the original schema that had partition in all caps for some reason. I like uniformity too 👍🏻

carols10cents:
Ok this is the database rename, I think it's ready to go but I'm scared

NGA-TRAN:
As much as we want this to happen, I think we need to discuss when we have this to go. Let me start a discussion with the team first

domodwyer (commented):
Ship it and bring the (temporary) errors :fire:

domodwyer (approved):

PR #5485, opened 2022-08-29
feat: Kafka payload size distribution metric

### LINKED ISSUE #5117: Ingester needs metric for kafka batch size

domodwyer:
Capture the approximate, uncompressed payload size - this ignores any framing overhead, local/client compression, or message compression by the brokers themselves. It serves as a useful measurement of the effectiveness of aggregation.

Closes #5117, relates to #5323.

---

* refactor: WARN for full aggregators (657efce47)

      Emit a WARN log line whenever an aggregator becomes full - this will help
      identify tuning opportunities.

* feat: capture Kafka message size distribution (c720b29c9)

      Adds instrumentation to the low-level (post-aggregation) Kafka client, 
      capturing the uncompressed, approximate message size (calculated as the sum of
      all Record::approximate_size() returns, ignoring largely static framing
      overhead).


write_buffer/src/kafka/instrumentation.rs
```diff
@@ -39,6 +47,24 @@ impl KafkaProducerMetrics {
             ("kafka_topic", kafka_topic_name.into()),
         ]);
 
+        // Capture the distribution of message sizes (sum of Record size)
+        let msg_size = metrics
+            .register_metric_with_options::<U64Histogram, _>(
+                "write_buffer_client_payload_size",
+                "distribution of approximate uncompressed message \
```
alamb, line 54: 👍

write_buffer/src/kafka/instrumentation.rs
```diff
@@ -225,5 +279,19 @@ mod tests {
...
+            .expect("failed to read metric")
+            .get_observer(&Attributes::from(&[
+                ("kafka_topic", KAFKA_TOPIC),
+                ("kafka_partition", "42"),
+            ]))
+            .expect("failed to get observer")
+            .fetch();
+
+        assert_eq!(histogram.sample_count(), 1);
+        assert_eq!(histogram.total, 0);
```
alamb, line 295: It seems strange that `0` is the correct value here (but maybe the test doesn't save the values 🤔 )

write_buffer/src/kafka/record_aggregator.rs
```diff
@@ -131,7 +132,14 @@ impl Aggregator for RecordAggregator {
             //
             // Map to the original input op this fn was called with, discarding
             // the encoded Record.
-            TryPush::NoCapacity(_) => TryPush::NoCapacity(op),
+            TryPush::NoCapacity(_) => {
+                // Log a warning if this occurs - this allows an operator to
+                // increase the maximum Kafka message size, or lower the linger
+                // time to minimise latency while still producing large enough
+                // batches for it to be worth while.
+                warn!("aggregated batch reached maximum capacity");
```
alamb, line 140:
```suggestion
                warn!("aggregated batch reached maximum capacity. Try increasing maximum Kafka message size or lower the linger time");
```

I think it might be a good idea to include the suggested actions in the message itself so a user doesn't have to know to look in the source code to find them

alamb (approved):

write_buffer/src/kafka/instrumentation.rs
domodwyer: The input was 0 records, this ensures it doesn't do something stupid should that happen :)

write_buffer/src/kafka/record_aggregator.rs
domodwyer: Some settings are in Kafka itself, some are hard-coded, some are exposed via envs (that change) - I fear being more helpful might be less helpful :(

Issue #5464, opened 2022-08-24
More Data-driven Engineering for IOx


crepererum:
# Why

During the IOx development, product scoping and operations, we face the following questions/challenges:

* Did change X improve/regress performance?
* https://github.com/influxdata/conductor/issues/958
* A/B testing for features / algorithms
* Many questions that people seem to wish to extract from query logs (that however are NOT persisted), e.g. https://github.com/influxdata/influxdb_iox/issues/4558

All these questions should be answered based on data. Ideally, this data shall be collected in neutral form (e.g. a CSV file) and then be analyzed with the same methods that we expect from proper medical studies instead of eye-balling[^1]. During this analysis, we want to have the ability to:

* Detect (and potentially remove) outliers
* Trim data ranges (e.g. to “X minutes after process start”)
* Remove / weight different samples (e.g. “their super bizarre query they run every morning”)

The data that we currently have is not suitable for this task, as the following subsections explain.

## Logs
We have logs, but they suffer from multiple drawbacks:

* **Correlation:** A single query has multiple log lines with the important information we need. Correlating / collecting them is cumbersome. This point could be worked around by emitting a single log line per relevant event with all metrics attached.
* **Retention:** Graylog has a very short retention period, so we would need to regularly (and automatically) extract the relevant data for long-time analysis.
* **Extraction:** Graylog is slow and its API rather average.

## Metrics (Prometheus / Grafana)
IOx also emits Prometheus-compatible metrics. However, they are limited as well:

* **Marginal distributions:** All metrics are marginal distributions. A detailed selection/filtering requires many labels with potentially larger cardinality – which is something Prometheus does not cope well with.
* **Binning:** [Histograms](https://prometheus.io/docs/concepts/metric_types/#histogram) (and [summaries](https://prometheus.io/docs/concepts/metric_types/#summary), but we do not use these) are discretized. This requires extreme care during analysis because it easily introduces biases (in mean and variance) and makes the estimation of small effects way harder. Also, our histograms make assumptions about the distribution of the data (= something exponential) to easily capture large ranges of data with minimal operation effort (= tuning the bin ranges).
* **Data Model:** Prometheus data model is based on time series data and while we can squeeze everything into a time series, the extraction of data is rather cumbersome.

## Traces
Traces in prod are so heavily sampled that they are barely existent and when they are, you may easily miss half of the trace.

## Binlog
[...]

PR #5490, opened 2022-08-29
feat: event emission subsystem

crepererum:
Add a framework to simplify event emission. This is required for #5464.
Mostly it allows us to:

- build an event step by step (so we don't need to pass a large amount
  of free variables around)
- integrate some default "data sources" to simplify the data analysis
  based on the events (e.g. one that adds the IOx GIT hash)
- add an abstract "emitter" concept (mostly this will send data to
  InfluxDB but this abstraction is helpful for testing of if we ever
  just want to write the events as logs)

See `event_emitter/README.md` for a short usage example.

The events are type-tagged their measurements. We'll likely have methods
with multiple event types (e.g. the query and an ingester request) and I
just want to avoid that people attach the wrong tags/fields to the wrong
events.


event_emitter/README.md
```diff
@@ -0,0 +1,64 @@
+# Event Emitter
```
alamb, line 1: `event` (as opposed to metrics or logs) is a good name 👍

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+/// Main driver for event emission.
+#[derive(Debug)]
+pub struct EventDriver {
+    _join_handle: JoinHandle<()>,
+    tx: Arc<Sender<Message>>,
+    data_providers: Vec<Box<dyn EventDataProvider>>,
+    time_provider: Arc<dyn TimeProvider>,
+}
+
+impl EventDriver {
```
alamb, line 40: "EventReporter" would be another common name for this type of object. I think EventDriver is also fine, I just wanted to mention it.

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+        mut emitter: Box<dyn EventEmitter>,
+        time_provider: Arc<dyn TimeProvider>,
+        handle: &Handle,
+    ) -> Self {
+        let (tx, mut rx) = channel(1_000);
+        let join_handle = handle.spawn(async move {
+            while let Some(msg) = rx.recv().await {
+                match msg {
+                    Message::Event(event) => {
+                        // TODO: batch events
```
alamb, line 53: So that we don't rediscover this TODO I would recommend that either the emitter API is for a single `Message` (so it is clear the driver is not batching anything) or actually implement the batching. 

To be honest, I almost feel the batching might be best done in the emitter logic as the optimal batching strategy is likely a function on where the events are being sent

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        EventRecorder {
+            event: Some(event.typed()),
+            do_not_send: false,
+            tx: Arc::clone(&self.tx),
+        }
+    }
+
+    /// Flushes queued events.
```
alamb, line 90:
```suggestion
    /// Flushes queued events. Resolves after all events recorded prior
    /// to this call are emitted. 
```

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl EventDriver {
+    /// Create new driver.
+    pub fn new(
+        data_providers: Vec<Box<dyn EventDataProvider>>,
+        mut emitter: Box<dyn EventEmitter>,
+        time_provider: Arc<dyn TimeProvider>,
+        handle: &Handle,
+    ) -> Self {
+        let (tx, mut rx) = channel(1_000);
```
alamb, line 48: I recommend making `1000` a named constant (mostly for visibility / readability to understand that there is a fixed size buffer)

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+where
+    M: TypedMeasurement,
+{
+    fn drop(&mut self) {
+        if !self.do_not_send {
+            let event = self.event.take().expect("not yet dropped").untyped();
+            match self.tx.try_send(Message::Event(event)) {
+                Ok(()) => {}
+                Err(TrySendError::Closed(_)) => {
+                    panic!("Background worker died");
```
alamb, line 156: Given there is a fixed sized channel, I would recommend these be `warn!` rather than `panic!`, as the stated design goal is to have this event system be 'best effort'

event_emitter/src/emitter/influxdb.rs
```diff
@@ -0,0 +1,136 @@
+//! Event emission to InfluxDB.
+use std::{ops::ControlFlow, sync::Arc};
+
+use async_trait::async_trait;
+use backoff::{Backoff, BackoffConfig};
+use influxdb_iox_client::{connection::Builder, error::Error as ClientError, write::Client};
+use influxdb_line_protocol::LineProtocolBuilder;
```
alamb, line 7: cc @mkmik  -- the LineProtocolBuilder is seeing more use

alamb (approved):
I am approving this PR as I think it implements what you have designed it for, but I am not convinced that of the need for all of this code

I just can't help but feel the emitter code is like 95% of what is already included in a system like `tracing`.

Specifically, 
`Event` is analogous to  https://docs.rs/tracing/0.1.36/tracing/#events

 `Emitter` is analogous to  https://docs.rs/tracing/0.1.36/tracing/#subscribers

So I guess I am wondering rather than implementing our own even publish / subscribe loop can you use tracing as the lumbing (and make a `Subscriber` that captures and translates whatever information we want into Line Protocol)?

domodwyer:
I haven't reviewed the code yet (lunch!) but I too thought "tracing crate" when I saw this PR and wanted to give a +1. 

The `tracing` crate can easily annotate functions of interest and emit spans transparently (or explicitly pass scopes around) and the subscriber hook is very flexible - plus it's all maintained / battle tested for us. I'd be surprised if we couldn't make it work for our needs.

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+{
+    measurement: M,
+    time: Time,
+    tags: BTreeMap<&'static str, Cow<'static, str>>,
+    fields: BTreeMap<&'static str, FieldValue>,
+}
+
+impl<M> Event<M>
+where
+    M: Into<&'static str>,
```
tustvold, line 95: What types implement `Into<&'static str>` other than `&'static str`?

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+    tx: Arc<Sender<Message>>,
+}
+
+impl<M> EventRecorder<M>
+where
+    M: TypedMeasurement,
+{
+    /// Do NOT send this event.
+    pub fn do_not_send(mut self) {
+        self.do_not_send = true;
```
tustvold, line 122:
```suggestion
        self.event.take()
```
?

event_emitter/src/driver.rs
```diff
@@ -0,0 +1,215 @@
...
+    event: Option<Event<M>>,
+    do_not_send: bool,
+    tx: Arc<Sender<Message>>,
+}
+
+impl<M> EventRecorder<M>
+where
+    M: TypedMeasurement,
+{
+    /// Do NOT send this event.
```
tustvold, line 120: What is the intended purpose of this?

event_emitter/src/emitter/influxdb.rs
```diff
@@ -0,0 +1,136 @@
...
+                builder_inner = builder_inner.field(k, v);
+            }
+
+            let builder_inner = builder_inner.timestamp(event.time().timestamp_nanos());
+
+            builder = builder_inner.close_line();
+        }
+
+        let lp = builder.build();
+        let lp = Arc::from(String::from_utf8(lp).expect("LP builder produces valid string"));
```
tustvold, line 90: Not related to this PR, but it seems a little strange that the builder doesn't produce a String

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+/// Typed InfluxDB-style event.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Event<M>
+where
+    M: Into<&'static str>,
+{
+    measurement: M,
+    time: Time,
+    tags: BTreeMap<&'static str, Cow<'static, str>>,
+    fields: BTreeMap<&'static str, FieldValue>,
```
tustvold, line 90: I'm curious if we have anything that needs these to be sorted. My experience with BTreeMap is that it is remarkably slow, and has interesting Send bounds on its iterators. TLDR can we just use HashMap and call it a day

event_emitter/src/measurement.rs
```diff
@@ -0,0 +1,33 @@
+//! Typed measurement.
+use std::fmt::Debug;
+
+/// A typed measurement.
+pub trait TypedMeasurement: Clone + Copy + Debug + Default + Into<&'static str> + 'static {}
```
tustvold, line 5: I don't really understand the use-case for this over just `&'static str`

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        self
+    }
+
+    /// Add new tag.
+    ///
+    /// # Panic
+    /// Panics if a tag or a field with the same name already exists. Also panics if a tag called `"time"` is used.
+    pub fn add_tag_move<V>(mut self, name: &'static str, value: V) -> Self
```
tustvold, line 165:
```suggestion
    pub fn with_tag<V>(mut self, name: &'static str, value: V) -> Self
```

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+    /// The iterator elements are sorted by field name.
+    pub fn fields(&self) -> impl Iterator<Item = (&'static str, &FieldValue)> {
+        self.fields.iter().map(|(k, v)| (*k, v))
+    }
+
+    /// Add new tag.
+    ///
+    /// # Panic
+    /// Panics if a tag or a field with the same name already exists. Also panics if a tag called `"time"` is used.
+    pub fn add_tag_mut<V>(&mut self, name: &'static str, value: V) -> &mut Self
```
tustvold, line 135:
```suggestion
    pub fn add_tag<V>(&mut self, name: &'static str, value: V)
```

I don't think this needs to be a builder-style method

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+    {
+        self.add_tag_mut(name, value);
+        self
+    }
+
+    /// Add a new field.
+    ///
+    /// # Panic
+    /// Panics if a tag or a field with the same name already exists. Also panics if a field called `"time"` is used.
+    pub fn add_field_mut<V>(&mut self, name: &'static str, value: V) -> &mut Self
```
tustvold, line 177:
```suggestion
    pub fn add_field<V>(&mut self, name: &'static str, value: V)
```
I don't think this needs to be a builder style method

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+
+        self
+    }
+
+    /// Add a new field.
+    ///
+    /// # Panic
+    /// Panics if a tag or a field with the same name already exists. Also panics if a field called `"time"` is used.
+    pub fn add_field_move<V>(mut self, name: &'static str, value: V) -> Self
```
tustvold, line 207:
```suggestion
    pub fn with_field<V>(mut self, name: &'static str, value: V) -> Self
```

event_emitter/src/event.rs
```diff
@@ -0,0 +1,344 @@
...
+    F64(f64),
+
+    /// Bool.
+    Bool(bool),
+
+    /// String.
+    String(Cow<'static, str>),
+}
+
+impl PartialEq for FieldValue {
```
tustvold, line 30: Is this not derivable?

event_emitter/src/data_provider/process.rs
```diff
@@ -0,0 +1,64 @@
...
+        Self {
+            t_start: time_provider.now(),
+            git_hash,
+            process_uuid,
+        }
+    }
+}
+
+impl EventDataProvider for ProcessEventDataProvider {
+    fn enrich(&self, event: &mut Event<&'static str>) {
```
tustvold, line 34: What further `EventDataProvider` are you intending to have, they appear to be statically registered on the `EventDriver` and so are necessarily global. I wonder if we could just inline this into the `EventEmitter`

tustvold (commented):
As much as it pains me, this does feel very similar to the `tokio::tracing` crate. The major differences I can see:

* The lower-level `tokio::tracing` interfaces are wild and incredibly confusing to understand
* Tracing requires declaring all the fields on the Event up-front, even if values can be provided later using a combination of https://docs.rs/tracing/latest/tracing/field/trait.Visit.html and https://docs.rs/tracing/latest/tracing/field/struct.Empty.html
* Tracing provides no way to distinguish between tags and fields, one hacky way to work around this would be to have a `__tags` meta field with a comma-separated list of tag attributes
* Tracing provides event context "enrichment" through spans and thread-local state which can be incredibly fiddly to debug or get working
* This allows targeting the callsites we are actually interested in, tracing has filtering support, but it is very focused towards the logging use-case and we have run into pain trying to use this for other purposes in the past

There are some concrete benefits to using tracing:

* It would allow integration with logging
* Potentially less to maintain (although likely instead we'd have some impenetrable trait nonsense)

I think both approaches *could* work, and tbh having a simple way to collect and export events is not a huge overhead, and is likely simpler than trying to fight `tokio::tracing` to achieve this. I've left some comments on how I think we can simplify this, in particular I'm not the `EventDataProvider` and `TypedMeasurement` plumbing is especially valuable, and this would make this an easier sell

event_emitter/src/driver.rs
crepererum: Like you build up an event but don't wanna emit it for some error case. Not sure, maybe some premature feature :sweat_smile:

event_emitter/src/event.rs
crepererum: Thinking about it a 2nd time: you're right, as long as the emitter pulls tags->fields->timestamp (which it does).

event_emitter/src/measurement.rs
crepererum: When moving around the events, I don't wanna confuse them when we have multiple at the same time (e.g. a query event and some object store / ingester connection event). I can however remove this if the API seems to be too bloated. @alamb 2nd reviewer opinion on the typing of the measurements?

event_emitter/src/event.rs
crepererum: Measurements.

event_emitter/src/event.rs
crepererum: done

event_emitter/src/event.rs
crepererum: done

event_emitter/src/event.rs
crepererum: it is, fixed

event_emitter/src/event.rs
crepererum: fixed

event_emitter/src/driver.rs
crepererum: done

event_emitter/src/driver.rs
crepererum: done

event_emitter/src/driver.rs
crepererum: done

event_emitter/src/driver.rs
crepererum: removed

event_emitter/src/data_provider/process.rs
crepererum: I found this separation easier to test and reason about. You're right that the bundling currently doesn't allow you to easily nest them throughout the stack.

PR #5492, opened 2022-08-29
refactor: improve namespace caching in querier

crepererum:
1. Cache converted schema instead of catalog schema. This safes a buch
   of memcopies during conversion.
2. Simplify creation of new chunks, we now only need a `CachedTable`
   instead of a namespace and a table schema.

In an artificial benchmark, this removed around 10ms from the query
(although that was prior to #5467 which moved schema conversion one
level up). Still I think it is the cleaner cache design.


querier/src/cache/namespace.rs
```diff
@@ -232,10 +290,10 @@ mod tests {
 
         let col111 = table11.create_column("col1", ColumnType::I64).await;
         let col112 = table11.create_column("col2", ColumnType::Tag).await;
-        let col113 = table11.create_column("col3", ColumnType::Time).await;
+        let col113 = table11.create_column("time", ColumnType::Time).await;
```
alamb, line 293: why did you need to change the name to `time`? Because the creation of an actual `Schema` object will error if you don't include a time column?

alamb (approved):
looks like a good change to me

querier/src/cache/namespace.rs
crepererum: Sort of, the `SchemaBuilder` creates timestamp column with `"time"` and I felt it was more in line with what we usually do and therefore better than working around it.

PR #5508, opened 2022-08-30
refactor: use a singel timestamp in policy backend

crepererum:
Prior to this PR we had at least 1 `TimeProvider::now` calls per GET
request (for caches that only used LRU) and up to 3 calls (caches with
LRU + refresh + TTL). Let's instead use a single timestamp that is
created by the policy backend itself (instead of the policies). This has
the following consequences:

- **efficiency:** `SystemProvider::now` is not free, even though under Linux
  this doesn't result in a syscall, it uses the stdlib time system which
  also checks for monotonicity
- **consistency:** All changes for a single trigger (e.g. a
  GET cache call) now use a single timestamp instead of slightly
  increasing ones. I argue this is the better semantic, simpler to
  understand and better to debug.

For some (slightly artificial) local performance experiment, this shaves
off around 2ms per single-table SQL query. However I expect that there might
be more degenerated cases (e.g. multi-table SQL queries or some
InfluxRPC requests that hit multiple tables).

The majority of this patch is moving the `TimeProvider` from the
policies into the policy backend.


cache_system/src/backend/policy/mod.rs
```diff
@@ -418,23 +441,32 @@ pub trait Subscriber: Debug + Send + 'static {
...
     /// Get value for given key if it exists.
-    fn get(&mut self, _k: &Self::K) -> Vec<ChangeRequest<'static, Self::K, Self::V>> {
+    fn get(&mut self, _k: &Self::K, _now: Time) -> Vec<ChangeRequest<'static, Self::K, Self::V>> {
         // do nothing by default
         vec![]
     }
 
     /// Set value for given key.
     ///
     /// It is OK to set and override a key that already exists.
```
alamb, line 451: I think it might be worth adding a comment here with the rationale for including `Time` as a parameter as it my not be obvious at some later date:

```suggestion
    /// It is OK to set and override a key that already exists.
    ///
    /// The current time `now` is provided as a parameter so that all policies and 
    /// backends use a unified timestamp rather than their own provider, which is 
    /// more consistent and performant.
```

cache_system/src/backend/policy/mod.rs
```diff
@@ -418,23 +441,32 @@ pub trait Subscriber: Debug + Send + 'static {
...
+    ) -> Vec<ChangeRequest<'static, Self::K, Self::V>> {
         // do nothing by default
         vec![]
     }
 
     /// Remove value for given key.
     ///
     /// It is OK to remove a key even when it does not exist.
-    fn remove(&mut self, _k: &Self::K) -> Vec<ChangeRequest<'static, Self::K, Self::V>> {
+    fn remove(
```
alamb, line 465:
```suggestion
    ///
    /// The current time `now` is provided as a parameter so that all policies and 
    /// backends use a unified timestamp rather than their own provider, which is 
    /// more consistent and performant.    fn remove(
```

alamb (commented):
👍  makes sense to me

alamb (approved):

PR #5510, opened 2022-08-30
fix: remove empty parquet panic

domodwyer:
Fixes https://github.com/influxdata/conductor/issues/1121.

While this will stop the code from emitting the panic, it also means we won't notice opportunities to improve our compaction planning - there are a couple cases where we're using `SplitExec` knowing up front that it'll produce no output. We should try to improve our planning logic to eliminate those cases.

Lacking data to prove it, my conjecture is that the situation where this panic was unavoidable is exceedingly rare, and instead was caused by the (far more plausible) cases where we can avoid adding useless split points.

---

* refactor: raise error for no rows in parquet file (769826476)

      Previously when attempting to serialise a stream of one or more RecordBatch
      containing no rows (resulting in an empty file), the parquet serialisation
      code would panic.

      This changes the code path to raise an error instead, to support the compactor
      making multiple splits at once, which may overlap a single chunk:

                       ────────────── Time ────────────▶

                               │                │
                       ┌█████──────────────────────█████┐
                       │█████  │    Chunk 1     │  █████│
                       └█████──────────────────────█████┘
                               │                │

                               │                │

                            Split T1         Split T2

      In the example above, the chunk has an unusual distribution of write 
      timestamps over the time range it covers, with all data having a timestamp
      before T1, or after T2. When a running a SplitExec to slice this chunk at T1
      and T2, the middle of the resulting 3 subsets will contain no rows. Because we
      store only the min/max timestamps in the chunk statistics, it is unfortunately
      impossible to prune one of these split points from the plan ahead of time.

* fix: compactor tolerates empty output (2fc0ddbea)

      Changes the compactor code to tolerate a SplitExec yielding an empty partition
      (with no rows).

      This raises a WARN as the situation in which this is acceptable is very rare,
      and is more likely indicative of an opportunity to improve the SplitExec usage
      (i.e. pruning out unnecessary split points).

### DIFF
--- a/compactor/src/parquet_file_combining.rs
+++ b/compactor/src/parquet_file_combining.rs
@@ -13,12 +13,18 @@ use iox_query::{
 use iox_time::TimeProvider;
 use metric::{Attributes, Metric, U64Histogram};
 use observability_deps::tracing::*;
-use parquet_file::{chunk::ParquetChunk, metadata::IoxMetadata, storage::ParquetStorage};
+use parquet_file::{
+    chunk::ParquetChunk,
+    metadata::IoxMetadata,
+    serialize::CodecError,
+    storage::{ParquetStorage, UploadError},
+};
 use schema::{sort::SortKey, Schema};
 use snafu::{ensure, ResultExt, Snafu};
 use std::{
     cmp::{max, min},
     collections::BTreeMap,
+    future,
     sync::Arc,
 };
 use uuid::Uuid;
@@ -283,8 +289,23 @@ pub(crate) async fn compact_parquet_files(
                 // Stream the record batches from the compaction exec, serialize
                 // them, and directly upload the resulting Parquet files to
                 // object storage.
-                let (parquet_meta, file_size) =
-                    store.upload(data, &meta).await.context(PersistSnafu)?;
+                let (parquet_meta, file_size) = match store.upload(data, &meta).await {
+                    Ok(v) => v,
+                    Err(UploadError::Serialise(CodecError::NoRows)) => {
+                        // This MAY be a bug.
+                        //
+                        // This also may happen legitimately, though very, very
+                        // rarely. See test_empty_parquet_file_panic for an
+                        // explanation.
+                        warn!(
+                            ?partition_id,
+                            %object_store_id,
+                            "SplitExec produced an empty result stream"
+                        );
+                        return Ok(None);
+                    }
+                    Err(e) => return Err(Error::Persist { source: e }),
+                };
 
                 debug!(?partition_id, %object_store_id, "file uploaded to object store");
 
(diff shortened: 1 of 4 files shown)


compactor/src/parquet_file_combining.rs
```diff
@@ -283,8 +289,23 @@ pub(crate) async fn compact_parquet_files(
...
+                        // This also may happen legitimately, though very, very
+                        // rarely. See test_empty_parquet_file_panic for an
+                        // explanation.
+                        warn!(
+                            ?partition_id,
+                            %object_store_id,
+                            "SplitExec produced an empty result stream"
+                        );
+                        return Ok(None);
+                    }
```
domodwyer, line 306: We pick out the "no data in this parquet file we just serialised" error and skip adding the file (or lack of) to the catalog. Everything else continues as normal :tada:

parquet_file/src/serialize.rs
```diff
@@ -25,6 +25,13 @@ pub enum CodecError {
     #[error("no record batches to convert")]
     NoRecordBatches,
 
+    /// The result stream contained at least one [`RecordBatch`] and all
+    /// instances yielded by the stream contained 0 rows.
+    ///
+    /// This would result in an empty file being uploaded to object store.
+    #[error("no rows to serialise")]
+    NoRows,
```
domodwyer, line 33: This is the error that replaces the panic.

parquet_file/tests/metadata.rs
```diff
@@ -105,16 +109,53 @@ async fn test_decoded_iox_metadata() {
...
+//                      Split T1         Split T2
+//
+// It is not possible to use the chunk statistics (min/max timestamps) to
+// determine this empty sub-section will result ahead of time, therefore the
+// parquet encoder must tolerate it and raise a non-fatal error instead of
+// panicking.
+//
+// Relates to:
+//      * https://github.com/influxdata/influxdb_iox/issues/4695
+//      * https://github.com/influxdata/conductor/issues/1121
```
domodwyer, line 157: And this describes the problem for the next person who comes along.

crepererum (approved):
Apart from the split logic, I think we can also have unpredictable empty files when we start to ingest deletes. So I think think this is fine.

It's nice that they don't end up in the catalog though (technically they don't even end up in the object store), so the querier doesn't perform useless work loading them.

domodwyer:
I have wrote up at least one split plan optimisation opportunity we discovered during this fix: https://github.com/influxdata/influxdb_iox/issues/5511

compactor/src/parquet_file_combining.rs
NGA-TRAN: In our alert system, I will propose to have 2 levels of alerts, one is critical and SRE and devs should know, one is something like this that will alert devs only

parquet_file/src/serialize.rs
```diff
@@ -129,16 +161,9 @@ where
...
     // Serialize the record batches into the in-memory buffer
     let meta = to_parquet(batches, meta, &mut bytes).await?;
-    if meta.row_groups.is_empty() {
-        // panic here to avoid later consequence of reading it for statistics
-        panic!("partition_id={}. Created Parquet metadata has no column metadata. HINT a common reason of this is writing empty data to parquet file: {:#?}", partition_id, meta);
-    }
-
-    debug!(?partition_id, ?meta, "Parquet Metadata");
-
     bytes.shrink_to_fit();
```
NGA-TRAN, line 164: Do we need this line if the `meta` has no rows?

parquet_file/tests/metadata.rs
NGA-TRAN: Nice comments. Thanks @domodwyer

PR #5518, opened 2022-08-31
fix: loop forever in compact_hot_partition_candidates

NGA-TRAN:
Closes #https://github.com/influxdata/conductor/issues/1130

This PR includes:

1. Fix the loop-forever reported in https://github.com/influxdata/conductor/issues/11302. 
2. Convert `filter_hot_parquet_files` to a closure as suggested [here](https://github.com/influxdata/influxdb_iox/pull/5471#discussion_r957767635)
3. Move new hot compaction functions to new module compact_hot_partition.rs because they are too long 


compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,713 @@
...
+
+// For a given list of hot partition candidates and a memory budget, compute memory needed to compact each one
+// and compact as many of them in parallel as possible until all candidates are compacted
+//
+// The way this function works is the needed memory to compact each partition will be accumulated until it hits over 90%
+// of the compactor's memory_budget_bytes. Then those partitions are compacted in parallel. Then the process repeats
+// until all partitions are compacted. However, since after leaving some budget for a partition, the remaining budget
+// may be not enough to conpact the next one but the full budget will. In that case, the  considering partition will
+// be pushed back as the last item of the list to be considered later with full budget.
+async fn compact_hot_partition_candidates<F>(
```
NGA-TRAN, line 121: All the fixes are in this function.

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,713 @@
...
+    filter_hot_parquet_files: F,
+    compactor: Arc<Compactor>,
+    mut candidates: VecDeque<PartitionCompactionCandidateWithInfo>,
+    table_columns: HashMap<TableId, Vec<ColumnTypeCount>>,
+) where
+    F: Fn(
```
NGA-TRAN, line 127: @carols10cents : I converted this to a closure as you suggested but I found a way to test data from catalog without the need to provide other test function for this.
I hope this is still useful for the cold partition you are working on

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,713 @@
+//! Collect highest hot candidates and compact them
```
NGA-TRAN, line 1: Besides the fixes in only one function pointed out below and new tests, all other functions in this file is moved without changes from handler.rs

NGA-TRAN:
@carols10cents This is a critical loop-forever bug in to compact hot partitions that we want it to merge before we start working on compacting cold partition

compactor/src/compact_hot_partitions.rs
NGA-TRAN: The fixes:
1. Reset variables for new cycle (near the end of the function)
2. Remove a few `continue`
3. Add `count`

All three avoid different kinds of loop-forever

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,711 @@
...
+        let mut sorted_candidates = candidates.into_iter().collect::<Vec<_>>();
+        sorted_candidates.sort_by_key(|c| c.candidate.partition_id);
+        let sorted_candidates = sorted_candidates.into_iter().collect::<VecDeque<_>>();
+
+        // There are 3 rounds of paralllel compaction
+        //   . Round 1: 3 candidates [P1, P2, P5] and total needed budget 13,500
+        //   . Round 2: 1 candidate [P6] and total needed budget 4,500
+        //   . Round 3: 1 candidate [P3] and total needed budget 11,250
+        //   P4 is not compacted due to overbudget
+        //  Debug info shows all 3 rounds
```
NGA-TRAN, line 673: This test is set up to hit all conditions and reproduce different loop-forevers that are all fixed

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,711 @@
...
+
+        // Todo next: So conveniently, debug log shows this is also a reproducer of
+        // https://github.com/influxdata/conductor/issues/1130
+        // "hot compaction failed: 1, "Could not serialize and persist record batches failed to peek record stream schema"
```
NGA-TRAN, line 677: BTW, this is an unrelated bug that this test happens to hit beause the files have no data. I will add smaller and more specific tests for this bug and `compact_hot_partitions_in_parallel` function after I fix it

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,711 @@
...
+        let columns = table_columns.get(&table_id);
+        if columns == None {
+            warn!(
+                ?partition_id,
+                ?table_id,
+                "hot compaction is skipped due to missing column types of its table"
+            );
+            // todo: add this partition and its info into a new catalog table
+            // https://github.com/influxdata/influxdb_iox/issues/5458
+            continue;
```
NGA-TRAN, line 173: I have just discovered that this `continue` if happens on the last candidate will exit the loop with leads to skipping step 4 below which is wrong. Fixing it in this PR.

compactor/src/compact_hot_partitions.rs
```diff
@@ -0,0 +1,711 @@
...
+            .await;
+
+        if let Err(e) = parquet_files_for_compaction {
+            warn!(
+                ?e,
+                ?partition_id,
+                "hot compaction failed due to error in reading parquet files"
+            );
+            // This may just be a hickup reading object store, skip commpacting it in this cycle
+            continue;
```
NGA-TRAN, line 194: Ditto (see the `continue` above)

compactor/src/compact_hot_partitions.rs
carols10cents: Wait, so this test isn't actually asserting anything? Why not?

compactor/src/compact_hot_partitions.rs
carols10cents: I think I messed up when I was suggesting the closure -- I said `filter_hot_parquet_files` but I actually meant `compact_hot_partitions_in_parallel`. Let me something

carols10cents:
@NGA-TRAN I added a commit that changes the actual function I meant to a closure (😓 ) and in the tests, passes a closure that collects the rounds in a vec. Then we can assert on that, and I added assertions based on what is currently happening. Can you check the assertions and make sure they're what you expect to reflect that the bugs are fixed?

compactor/src/compact_hot_partitions.rs
```diff
@@ -667,23 +652,91 @@ mod tests {
...
+            move |_compactor: Arc<Compactor>,
+                  parallel_compacting_candidates: Vec<FilteredFiles>| {
+                let compaction_groups_for_async = Arc::clone(&compaction_groups_for_closure);
+                async move {
+                    compaction_groups_for_async
+                        .lock()
+                        .unwrap()
+                        .push(parallel_compacting_candidates);
+                }
+            };
```
NGA-TRAN, line 676: This is a nice trick

compactor/src/compact_hot_partitions.rs
```diff
@@ -667,23 +652,91 @@ mod tests {
...
+        // Round 1
+        let group1 = &compaction_groups[0];
+        assert_eq!(group1.len(), 3);
+
+        let g1_candidate1 = &group1[0];
+        assert_eq!(g1_candidate1.budget_bytes(), 4500);
+        assert_eq!(g1_candidate1.partition.id(), partition1.id);
+        let g1_candidate1_pf_ids: Vec<_> =
+            g1_candidate1.files.iter().map(|pf| pf.id.get()).collect();
+        assert_eq!(g1_candidate1_pf_ids, vec![2, 1]);
```
NGA-TRAN, line 703: Do you think this will be deterministic oral test environments (laptops circle CI)? Do we need to sort the vectors before comparing?

compactor/src/compact_hot_partitions.rs
```diff
@@ -667,23 +652,91 @@ mod tests {
...
+        // Round 3
+        let group3 = &compaction_groups[1];
+        assert_eq!(group3.len(), 1);
+
+        let g3_candidate1 = &group3[0];
+        assert_eq!(g3_candidate1.budget_bytes(), 4500);
+        assert_eq!(g3_candidate1.partition.id(), partition6.id);
+        let g3_candidate1_pf_ids: Vec<_> =
+            g3_candidate1.files.iter().map(|pf| pf.id.get()).collect();
+        assert_eq!(g3_candidate1_pf_ids, vec![12, 11]);
```
NGA-TRAN, line 739: This is what exactly I want to verify. Thanks so much @carols10cents

NGA-TRAN (commented):
Your commit does the work I need @carols10cents

compactor/src/compact_hot_partitions.rs
NGA-TRAN: It throws error if the output files are empty (error return) which may or may not right for us depending on cases. This is a bug reported in https://github.com/influxdata/conductor/issues/1130 that  was introduced by https://github.com/influxdata/influxdb_iox/pull/5510 which is a fix we prefer but we need to change a few more places accordingly. I think I know where to there are

compactor/src/compact_hot_partitions.rs
carols10cents: I think this will be deterministic. Famous last words, let's find out :)

compactor/src/compact_hot_partitions.rs
```diff
@@ -728,15 +728,15 @@ mod tests {
         assert_eq!(g2_candidate1_pf_ids, vec![12, 11]);
 
         // Round 3
-        let group3 = &compaction_groups[1];
+        let group3 = &compaction_groups[2];
```
carols10cents, line 731: OOPS great catch!!!

carols10cents (approved):

compactor/src/compact_hot_partitions.rs
NGA-TRAN: Yeah, I went over the implementation that we use VecDeque and with the provided memory budget, this will be deterministic

compactor/src/compact_hot_partitions.rs
NGA-TRAN: 😄 The difference in memory budget caught my eyes. I spent so much time yesterday to come up with this scenario and remember every single action 😆

PR #5521, opened 2022-08-31
feat(ingester): restrict partition row count

domodwyer:
Restrict the number of rows in each partition before marking for persistence - defaulting to 500,000.

---

* feat(ingester): restrict partition row count (2a1960645)

      This limit restricts a single partition to containing at most N rows before it
      is marked for persistence (note: being marked for persistence does not
      currently prevent further ingest for that partition.)

### DIFF
--- a/ingester/src/lifecycle.rs
+++ b/ingester/src/lifecycle.rs
@@ -30,6 +30,7 @@ pub trait LifecycleHandle: Send + Sync + 'static {
         shard_id: ShardId,
         sequence_number: SequenceNumber,
         bytes_written: usize,
+        rows_written: usize,
     ) -> bool;
 
     /// Returns true if the `total_bytes` tracked by the manager is less than the pause amount.
@@ -59,6 +60,7 @@ impl LifecycleHandle for LifecycleHandleImpl {
         shard_id: ShardId,
         sequence_number: SequenceNumber,
         bytes_written: usize,
+        rows_written: usize,
     ) -> bool {
         let mut s = self.state.lock();
         let now = self.time_provider.now();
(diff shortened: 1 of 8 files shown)


pauldix (approved):

ingester/src/lifecycle.rs
```diff
@@ -153,6 +162,13 @@ pub struct LifecycleConfig {
     /// manager will persist it. This is to ensure that cold partitions get cleared out to make
     /// room for partitions that are actively receiving writes.
     partition_cold_threshold: Duration,
+
+    /// The maximum number of rows allowed within a single partition before it
+    /// MUST be persisted.
+    ///
+    /// Reaching this limit pauses ingest while the partition is flushed to
+    /// object storage.
+    partition_row_max: usize,
```
alamb, line 171: I don't really understand why we need both a row limit and a memory limit - if the memory limit wasn't sufficient, isn't that a sign that the memory calculation is incorrect?

Limiting rows seems like it will be hard to set correctly as the resource being consumed (memory) is not linearly correlated with the row count (as it depends on table width, dictionary compression, etc)

pauldix: The problem is that some tables have a very compressed representation in memory, but when we go to persist them, they explode out by a significant amount. This setting accounts for the problem that the in-memory size doesn't necessarily match with how much memory will be required to persist the data.

Having the two together make the system less likely to fall over with the wide range of data shapes that we see.

alamb: 🤔  I still think the best user experience is a setting in terms of memory. However, if we don't have a good way to estimate memory usage for persist (which we probably don't),  I agree that until we do, a row count limit is a reasonable workaround

pauldix: Agree, setting based on a single memory value would be ideal, but getting to that is a much longer process than just putting this in for now.

pauldix: One other thought on this. When thinking about adding row count, I was also considering what impact large partitions might have on query performance. Ideally, the ingester wouldn't have too many rows in a single partition since those all need to be sorted and deduplicated at query time. So keeping a bound on the row count for that purpose is also likely important.

PR #5523, opened 2022-08-31
refactor: Use more iox test utils in compactor tests

carols10cents:
So that the test setup is hopefully clearer on which parts are different and important.


NGA-TRAN (approved):
Nice. Thanks @carols10cents

PR #5525, opened 2022-09-01
refactor: use concrete type in `MetricsLoader`

crepererum:
The API user may still use a `Box<dyn ...>` if they want, but they
technically don't have to.


alamb (approved):
I found https://github.com/influxdata/influxdb_iox/pull/5525/files?diff=split&w=1 easier to review 👍

PR #5527, opened 2022-09-01
feat: querier object store cache

crepererum:
This is the groundwork to remove the read buffer from the querier and solely use parquet files. Technically this will end up in a very similar construct as the read buffer, only that our intermediate representation is an in-mem parquet file instead of a read buffer.

As a reminder: the last naive experiment that tried this ended up in a disaster, because we were requesting files from object store for basically every single request. So we need to cache the parquet-encoded data.

This PR adds a new virtual object store type that is managed by the querier cache system. It is NOT used yet.


querier/src/cache/object_store.rs
```diff
@@ -0,0 +1,465 @@
...
+) -> Result<Option<Bytes>, ObjectStoreError> {
+    let get_result = match store.get(path).await {
+        Ok(get_result) => get_result,
+        Err(ObjectStoreError::NotFound { .. }) => return Ok(None),
+        Err(e) => return Err(e),
+    };
+
+    let data = match get_result.bytes().await {
+        Ok(data) => data,
+        Err(ObjectStoreError::NotFound { .. }) => return Ok(None),
```
tustvold, line 41: FWIW this can't return this variant

querier/src/cache/object_store.rs
```diff
@@ -0,0 +1,465 @@
...
+}
+
+impl ObjectStoreCache {
+    /// Create new empty cache.
+    pub fn new(
+        backoff_config: BackoffConfig,
+        object_store: Arc<dyn ObjectStore>,
+        time_provider: Arc<dyn TimeProvider>,
+        metric_registry: &metric::Registry,
+        ram_pool: Arc<ResourcePool<RamSize>>,
```
tustvold, line 77: Not related to this PR, but it might be nice if this integrated eventually with DataFusion's ResourceManager

querier/src/cache/object_store.rs
```diff
@@ -0,0 +1,465 @@
...
+        testing: bool,
+    ) -> Self {
+        let object_store_captured = Arc::clone(&object_store);
+        let loader = Box::new(FunctionLoader::new(move |key: Path, _extra: ()| {
+            let backoff_config = backoff_config.clone();
+            let object_store = Arc::clone(&object_store_captured);
+
+            async move {
+                Backoff::new(&backoff_config)
```
tustvold, line 86: The object store crate already has a backoff and retry setup FWIW

querier/src/cache/object_store.rs
```diff
@@ -0,0 +1,465 @@
...
+                .await
+                .unwrap()
+                .bytes()
+                .await
+                .unwrap(),
+            bytes_1,
+        );
+        assert_eq!(get_count_hit(&metric_registry), 1);
+
+        // ensure "misses" are cached
```
tustvold, line 336:
```suggestion
        // ensure "not found" are cached
```
Too many overloads :laughing:

tustvold (approved):
Looking nice :+1:

querier/src/cache/object_store.rs
crepererum: Not for the current implementation, but let's better be safe than sorry.

querier/src/cache/object_store.rs
crepererum: Isn't the resource manager session scoped?

querier/src/cache/object_store.rs
crepererum: Yeah, the current querier cache system never fails, it just retries forever. At one point I need to clean this up, also figuring out how to "cache" or not "cache" error states.

querier/src/cache/object_store.rs
alamb: In theory the `RuntimeEnv` (ResourceManager) should be global scoped -- in the IOxquery engine,  it is attached to the top level "Executor" -- https://github.com/influxdata/influxdb_iox/blob/9215a534d0fd241442c40db995884b6d899800eb/iox_query/src/exec.rs#L63

However, since we don't do anything with that runtime env yet, I can not be certain all the plumbing is hooked up correctly.

alamb: In the long term, I envision some higher level querier resource system that controls memory across the executor (DataFusion plans) and the various caches. 

However, since DataFusion doesn't really have mature resource control at the moment, this is a project for the future

querier/src/cache/object_store.rs
alamb: maybe worth a ticket (especially if it is something someone else might be able to do)

crepererum: Filed #5552, from there we can see if we need some more advanced error handing

PR #5530, opened 2022-09-01
refactor: do not always box `FunctionLoader`

crepererum:
While constructing our caches, we rarely spell out types. Instead we
just "box" the end result (`Box<dyn Cache<...>>`) before storing it
within the struct (e.g. `NamespaceCache`).

This means that `FunctionLoader` does NOT need to erase the type of the
function it contains. This saves us a `Box` which means less pointer
chasing and more room for the compiler to optimize.


cache_system/src/loader/mod.rs
```diff
@@ -38,37 +37,91 @@ where
 }
 
 /// Simple-to-use wrapper for async functions to act as a [`Loader`].
-pub struct FunctionLoader<K, V, Extra> {
-    loader: Box<dyn (Fn(K, Extra) -> BoxFuture<'static, V>) + Send + Sync>,
+///
```
alamb, line 40: ❤️  for comments

alamb (approved):

Issue #5397, opened 2022-08-15
Use DataFusion native ParquetReader in IOx


alamb:
__Use case:__
As we invest time and effort into the parquet read performance in  arrow + datafusion (e.g. https://github.com/apache/arrow-datafusion/issues/3147)  we would like to take advantage of those improvements in IOx, ideally with little or no additional effort. 

__Proposal:__
Use the native `ParquetExec` from DataFusion for reading parquet files. This is partially enabled via https://github.com/apache/arrow-datafusion/pull/3051

__Current behaviour:__
IOx uses its own code to download and scan parquet, at the time of writing in  https://github.com/influxdata/influxdb_iox/blob/main/parquet_file/src/storage.rs#L171-L235

__Desired behaviour:__
Use DataFusion's

__Alternatives considered:__
Describe other solutions or features you considered.

alamb:
FYI @tustvold and @crepererum  -- I wrote this ticket to capture something I think I @tustvold  has plans over the next month or two. Hopefully I got it mostly right

PR #5531, opened 2022-09-01
refactor: use DataFusion to read parquet files

crepererum:
Remove our own hand-rolled logic and let DataFusion read the parquet
files.

As a bonus, this now supports predicate pushdown to the deserialization
step, so we can use parquets as in in-mem buffer.

Note that this currently uses some "nested" DataFusion hack due to the
way the `QueryChunk` interface works. Midterm I'll change the interface
so that the `ParquetExec` nodes are directly visible to DataFusion
instead of some opaque `SendableRecordBatchStream`.

Closes https://github.com/influxdata/influxdb_iox/issues/5397


crepererum:
Compactor tests are failing due to the way the "fake" their file sizes (i.e. they pretend files have a different size than they actually have which now messes up reading the files :facepalm: ).

parquet_file/src/storage.rs
```diff
@@ -174,141 +179,76 @@ impl ParquetStorage {
...
-    }
-}
-
-/// Downloads the specified parquet file to a local temporary file
-/// and push the [`RecordBatch`] contents over `tx`, projecting the specified
-/// column indexes.
-///
-/// This call MAY download a parquet file from object storage, temporarily
-/// spilling it to disk while it is processed.
-async fn download_and_scan_parquet(
```
tustvold, line 235: :heart:

tustvold (approved):
This is really awesome, and will open up all sorts of exciting possibilities going forward

parquet_file/src/storage.rs
```diff
@@ -174,141 +179,76 @@ impl ParquetStorage {
     /// object store impl caches the fetched bytes.
     pub fn read_filter(
         &self,
-        _predicate: &Predicate,
+        predicate: &Predicate,
```
alamb, line 182: This is very important (as we will now properly pass down the predicate to the parquet files which will enable row group pruning, as well as various other fun optimizations the DataFusion community has planned) 👍

parquet_file/src/storage.rs
```diff
@@ -174,141 +179,76 @@ impl ParquetStorage {
...
+            limit: None,
+            table_partition_cols: vec![],
+        };
+        let exec = ParquetExec::new(base_config, expr, None);
 
-        // returned stream simply reads off the rx channel
-        Ok(AdapterStream::adapt(schema, rx, handle))
+        // set up "fake" DataFusion session
+        let object_store = Arc::clone(&self.object_store);
+        let session_ctx = SessionContext::new();
```
alamb, line 225: I think this may be problematic as it will not be connected to the main context (or things like  the query span, or temp file manager, if we ever use that). 

I think we should be plumbing the `IOxSessionContext` down here -- perhaps we can do so as a follow on PR.

parquet_file/src/storage.rs
```diff
@@ -174,141 +179,76 @@ impl ParquetStorage {
...
+        let task_ctx = Arc::new(TaskContext::from(&session_ctx));
+        task_ctx
+            .runtime_env()
+            .register_object_store("iox", "iox", object_store);
+
+        Ok(Box::pin(RecordBatchStreamAdapter::new(
+            Arc::clone(&schema),
+            futures::stream::once(execute_stream(Arc::new(exec), task_ctx)).try_flatten(),
```
alamb, line 233: I wonder if we can use the result of calling `execute_stream` directly (aka why do we need to flatten it?)

alamb (commented):
All in all I love it -- I left some comments I think should at least be read and responded to prior to merge, but I think this is a great step forward. Thanks @crepererum 

I also think that this PR closes https://github.com/influxdata/influxdb_iox/issues/5397 and I updated the description to say so.

alamb (approved):

parquet_file/src/storage.rs
crepererum: Yes, this is a temporary measure. Ideally this wouldn't even read the `ParquetExec` into an opaque stream of record batches but we would rather use the `ParquetExec` directly in our physical plan. However this refactoring requires a bit more work and I wanted to avoid "change the world" PRs.

parquet_file/src/storage.rs
crepererum: > aka why do we need to flatten it

Because `execute_stream` is NOT `fn execute_stream(...) -> SendableRecordBatchStream` but `async fn execute_stream(...) -> Result<SendableRecordBatchStream>`, i.e. we have a future here that creates the stream, not the stream itself. So the `once(...).try_flatten()` construct will take the future, treat it as a stream with a single element that produces a stream, and then flattens it.

Or in other words: we have a delayed construction of the actual stream here.

This trick BTW is also used in DataFusion and was suggested by @tustvold for another part of the codebase (`querier::chunk::query_access`).

crepererum:
Blocked by #5541.

parquet_file/src/storage.rs
alamb: I see -- thank you

Maybe it is worth a comment to that effect

alamb:
🎉

PR #5532, opened 2022-09-01
refactor: use concrete loader type within `CacheDriver`

crepererum:
There's no technical reason to dyn-dispatch the loader. The user may
free to do so however.

Note that there's no change within `querier` because the type
inference will now automatically treat the passed `Arc` as `Arc<L>`
instead of `Arc<dyn Loader<...>>`.


cache_system/src/loader/mod.rs
```diff
@@ -1,20 +1,20 @@
...
+use std::{fmt::Debug, future::Future, hash::Hash, marker::PhantomData};
 
 pub mod metrics;
 
 /// Loader for missing [`Cache`](crate::cache::Cache) entries.
 #[async_trait]
 pub trait Loader: std::fmt::Debug + Send + Sync + 'static {
     /// Cache key.
-    type K: Hash + Send + 'static;
+    type K: Debug + Hash + Send + 'static;
```
crepererum, line 11: These `Debug`s weren't required by the `Loader` trait before but basically by every other trait within `cache_system` (e.g. `Cache` and `CacheBackend`) so this is basically just a consistency change.

alamb (approved):
LGTM

Issue #5484, opened 2022-08-29
InfluxDB 2 Rust client: error while reading data


rkudryashov:
I use the library to store and read some data (both the client app and InfluxDB 2.4 instance are running inside a Kubernetes cluster). The former works well, but when I try to test getting data using the following code:

```rust
pub async fn read(influxdb_client: &InfluxDbClient) -> Result<(), CustomError> {
    let query = r#"
        from(bucket: "iot-measurements")
            |> range(start: -1h)
    "#;

    let res = influxdb_client
        .query(
            INFLUXDB_ORG.as_str(),
            Some(Query::new(query.to_string())),
        )
        .await?;

    Ok(())
}
```

The result of a call of the specified function is:

```
[src/processor.rs:61] data = Err(
    InternalError {
        message: "Error while processing the HTTP request: error decoding response body: EOF while parsing a value at line 2 column 0",
    },
)
```

When `trace` logger level is enabled, the output is as follows:

```
2022-08-29T09:17:08.951826Z TRACE hyper::client::pool: checkout waiting for idle connection: ("http", influxdb:8086)
2022-08-29T09:17:08.951862Z DEBUG reqwest::connect: starting new connection: http://influxdb:8086/    
2022-08-29T09:17:08.951880Z TRACE hyper::client::connect::http: Http::connect; scheme=Some("http"), host=Some("influxdb"), port=Some(Port(8086))
2022-08-29T09:17:08.952007Z DEBUG hyper::client::connect::dns: resolving host="influxdb"
2022-08-29T09:17:08.952880Z DEBUG hyper::client::connect::http: connecting to 10.96.179.109:8086
2022-08-29T09:17:08.953513Z TRACE mio::poll: registering event source with poller: token=Token(33554434), interests=READABLE | WRITABLE    
2022-08-29T09:17:08.954749Z DEBUG hyper::client::connect::http: connected to 10.96.179.109:8086
2022-08-29T09:17:08.954830Z TRACE hyper::client::conn: client handshake Http1
2022-08-29T09:17:08.954843Z TRACE hyper::client::client: handshake complete, spawning background dispatcher task
2022-08-29T09:17:08.954861Z TRACE want: signal: Want    
2022-08-29T09:17:08.954864Z TRACE want: signal found waiting giver, notifying    
2022-08-29T09:17:08.954869Z TRACE hyper::proto::h1::conn: flushed({role=client}): State { reading: Init, writing: Init, keep_alive: Busy }
[...]

PR #5533, opened 2022-09-01
fix: Improve debugging ability for influxdb2_client

carols10cents:
Add more error variants with more information and turn on `reqwest` verbose connection logging.

Connects to #5484. Let's wait to merge until we hear whether this is helpful or needs more.


carols10cents:
It looks like this code was helpful in diagnosing the issue (even though this code doesn't *fix* the issue), so I think this can be merged.

mkmik (approved):
This is useful. Thanks!

alamb (commented):
👌

PR #5535, opened 2022-09-01
feat: make compactors to select candidates based on the last n minutes

NGA-TRAN:
This hope to solve the problem reported in https://github.com/influxdata/conductor/issues/1135

make compactors to select candidates based on the last n minutes to reduce workload for postgres catalog query


iox_catalog/src/postgres.rs
```diff
@@ -1705,25 +1705,31 @@ WHERE parquet_file.shard_id = $1
...
         // and index on (shard_id, comapction_level, to_delete, created_at)
+        //
+        // Update: Sept 1, 2022
+        // query is too slow one 4 hours of data (~ 18 munites) if there are a lot of files
+        // Convert 'hour' to 'minute' to query less data
+        // Note: The explain shows postgres does GroupAggregate
         sqlx::query_as::<_, PartitionParam>(
             r#"
-SELECT partition_id, shard_id, namespace_id, table_id, count(id)
+SELECT partition_id, table_id, shard_id, namespace_id, count(id)
```
NGA-TRAN, line 1728: Per a blog post https://www.cybertec-postgresql.com/en/speeding-up-group-by-in-postgresql/, group by highest cardinality first may help PG performance better

compactor/src/compact.rs
```diff
@@ -274,14 +274,15 @@ impl Compactor {
             ]);
 
             // Get the most recent highest ingested throughput partitions within
-            // the last 4 hours. If nothing, increase to 24 hours
+            // the last 1 minutes. If nothing, increase to 10 minutes, 30m minutes, 60 minutes,
+            // 4 * 60 minutes, 24 * 60 minutes
             let mut num_partitions = 0;
-            for num_hours in [4, 24] {
+            for num_minutes in [1, 10, 30, 60, 4 * 60, 24 * 60] {
```
pauldix, line 280: I'd lose the 1 minute query. We want partitions that have at least a few level 0 files to compact.

NGA-TRAN, line 280: I would say this is temporary solution for now. If we keep this for the long run, it might starve many hot partitions

pauldix (approved):
Just one small comment to remove the 1 minute time check. Start with 10 and see how it is.

compactor/src/compact.rs
NGA-TRAN: I remove the one for 1 minute but the query did check at least 2 L0

Issue #5512, opened 2022-08-30
Use same memory estimation calculations for compacting cold partitions


NGA-TRAN:
# Rationale

This is a follow-up of #5471 and #5457 where we use new memory estimation formulas for hot partitions. We need to do the same for cold partitions

PR #5536, opened 2022-09-01
refactor: Prepare for sharing some code between hot and cold compaction

carols10cents:
Connects to #5512.

Does not actually do any code sharing yet, just prepares. I'm not completely happy with this, but I hope with the next step of actually using this code will help guide more improvements.

I *highly* recommend reviewing this commit-by-commit.


NGA-TRAN (approved):
Thanks for suggesting to review per commit. They all look good to me. And yes, the new closure looks so long but I think you will change it when you work on the cold partition

PR #5541, opened 2022-09-02
refactor: fine-grained file size mocking

crepererum:
There are multiple users of the "parquet file size". Some use it to gauge what/when to do things (e.g. the compactor), some need it to actually read the file (see #5531). Blindly overriding the size for all users ain't going to work (again see #5531), so let's have a more selective approach.

The major change is the compactor which now carries its own type for parquet files to propagate size overrides. I think it's useful to have this functionality for testing (instead of carefully crafting an actual file that falls into certain limits).


querier/src/namespace/query_access.rs
```diff
@@ -567,30 +567,34 @@ mod tests {
...
+        let file1 = partition.create_parquet_file(builder).await;
 
         let builder = TestParquetFileBuilder::default()
             .with_line_protocol("table foo=2 22")
             .with_max_seq(4)
             .with_min_time(22)
-            .with_max_time(22)
-            .with_file_size_bytes(200);
-        partition.create_parquet_file(builder).await;
+            .with_max_time(22);
```
alamb, line 577: :thumbsup:

compactor/src/lib.rs
```diff
@@ -92,6 +93,7 @@ pub(crate) async fn compact_hot_partition(
 pub(crate) async fn compact_cold_partition(
     compactor: &Compactor,
     partition: PartitionCompactionCandidateWithInfo,
+    size_overrides: &HashMap<ParquetFileId, i64>,
```
alamb, line 96: I recommend adding comments or some other way to document what `size_overrrides` is for. It would not be obvious to me if I ran across this code for the first time. 

Alternately, perhaps we could encapsulate the size calculation and override in some type that could have the appropriate comments, like

```rust
/// Used to calculate parquet sizes
struct ParquetSizeCalculator {
 overrides: HashMap<ParquetFileId, i64>,
}

impl ParquetSizeCalculcator {
  pub fn size(&self, id: ParquetFileId) -> i64 {
   ...
   }
}
```

compactor/src/parquet_file.rs
```diff
@@ -0,0 +1,83 @@
...
+        this.size_override = Some(size);
+        this
+    }
+
+    pub fn id(&self) -> ParquetFileId {
+        self.inner.id
+    }
+
+    pub fn file_size_bytes(&self) -> i64 {
+        self.size_override.unwrap_or(self.inner.file_size_bytes)
```
alamb, line 26: this is the current reason for `CompactorParquetFile` to exist, right?

alamb (approved):
Looks good to me -- I think @NGA-TRAN  or someone else with more of a compactor focus should give it a review too

NGA-TRAN:
I will review this later today

NGA-TRAN (approved):
I have reviewed this carefully. The changes should work correctly. I do not yet see the benefits in Compactor but they create no harm either.  I trust @crepererum know the benefits this will help him

compactor/src/lib.rs
crepererum: done (via doc-string)

compactor/src/parquet_file.rs
crepererum: Yes, but I thought the type might be helpful in the future in case the compactor wants to pipe additional attributes through.

PR #5543, opened 2022-09-02
chore: Add test for running a query after swapping out a sqlx pool

alamb:
Ports a test from https://github.com/influxdata/conductor/pull/861 to IOx repo to follow the code motion.

# Rationale
I was trying to convince myself that outstanding queries in a hotswap pool would be ok, so I wrote a test. There were some outstanding issues due to my mis-understanding and so this PR languished in the other repo, and I forgot about it until recently, when @mkmik suggested I port it here.

# Changes
Add a test that shows a SQL statement created, but not resolved, before a pool was replaced, works just fine


sqlx-hotswap-pool/src/lib.rs
```diff
@@ -287,4 +291,39 @@ mod tests {
...
+        // create a future from the pool but don't execute it yet
+        let query_future = sqlx::query("CREATE TABLE IF NOT EXISTS test (id int)").execute(&db);
+
+        // hot swap a new pool. This should not deadlock
+        let new_pool = connect_db().await.unwrap();
+        let old_pool = db.replace(new_pool);
+
+        // Ensure there are outstanding references to the old pool
+        // captured in the future
+        assert_eq!(Arc::strong_count(&old_pool), 1);
```
alamb, line 317: This is a modification from the original test to handle @mkmik 's https://github.com/influxdata/conductor/pull/861#discussion_r771423691

mkmik (approved):
Thanks you!

PR #5421, opened 2022-08-17
chore: changed `iox_shared` to `iox-shared`

gulio-python:
Closes #5403

## Describe your proposed changes here:
changed io_shared to iox-shared in the following files: update_catalog.rs, partition.rs, lib.rs (in the service_grpc_catalog folder) and lib.rs (in the service_grpc_object_store folder).

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

Issue #5403, opened 2022-08-16
standardise on using "iox-shared" everywhere to avoid confusion


lukebond:
we use `iox-shared` as the global shared namespace for IOx in cloud2, and then use it as a literal in various tests, comments, docs and examples. there are a number of places where we instead use `iox_shared` which is incorrect, although harmless. i suggest we fix those cases in case people get confused (as i did) about which they should be using for real, outside of tests, etc., and it leads to a misconfiguration of a running environment.

gulio-python:
I would like to help on this problem. If I'm correct these changes have to only apply to documentation files, this includes all .md files that I can find in this repo, right?
I've seen that iox_shared is used in a lot of other cases inside the code for example in .rs and .sql files. I suppose that these should just stay as is?

I'm sorry if these are some stupid questions, but it is just my second time participating on a public project and I just want the things that I work on to be perfect.

lukebond:
not stupid questions at all, @gulio-python! great that you want to help out :)

there are many cases where it is obvious but there will be some that are a little more tricky if one doesn't know the codebase well. for example `iox_shared` is used as a DB schema name for Postgres and we don't want to change that. here's what i think would be a reasonable guide:
- the instances in `README.md` and `iox_catalog/README.md` are Postgres related, leave those as they are
- any instance that is called (or used for) a kafka topic or query pool should be `iox-shared`, but be careful with anything that is not inside the `mod test { }` (usually at the bottom of each file). this means it's part of a test and should be safe to change. often tests use a constant variable declared somewhere else, so if you change that make sure you find all references of it in case it's used outside of a test
- anything that looks like it's going into a postgres DSN or postgres schema, or that lives in a `.sql` file, leave alone

gulio-python:
okay thank you very much for the explanation. I'll dive into it!

PR #5549, opened 2022-09-04
chore: changed `iox_shared` to `iox-shared`

alamb:
Based on https://github.com/influxdata/influxdb_iox/pull/5421 from @gulio-python 

Closes https://github.com/influxdata/influxdb_iox/issues/5403
Closes https://github.com/influxdata/influxdb_iox/issues/5421

### DIFF
--- a/import/src/aggregate_tsm_schema/update_catalog.rs
+++ b/import/src/aggregate_tsm_schema/update_catalog.rs
@@ -945,7 +945,7 @@ mod tests {
             .repositories()
             .await
             .topics()
-            .create_or_get("iox_shared")
+            .create_or_get("iox-shared")
             .await
             .expect("topic created");
         let (connection, _join_handle, requests) = create_test_shard_service(MapToShardResponse {
@@ -984,8 +984,8 @@ mod tests {
         let agg_schema: AggregateTSMSchema = json.try_into().unwrap();
         update_iox_catalog(
             &agg_schema,
-            "iox_shared",
-            Some("iox_shared"),
+            "iox-shared",
+            Some("iox-shared"),
             Some("inf"),
             Arc::clone(&catalog),
             connection,
--- a/import/src/aggregate_tsm_schema/update_catalog.rs
+++ b/import/src/aggregate_tsm_schema/update_catalog.rs
@@ -494,7 +494,7 @@ mod tests {
             .repositories()
             .await
             .topics()
-            .create_or_get("iox_shared")
+            .create_or_get("iox-shared")
             .await
             .expect("topic created");
         let (connection, _join_handle, _requests) = create_test_shard_service(MapToShardResponse {
@@ -524,8 +524,8 @@ mod tests {
         let agg_schema: AggregateTSMSchema = json.try_into().unwrap();
         update_iox_catalog(
             &agg_schema,
-            "iox_shared",
-            Some("iox_shared"),
+            "iox-shared",
+            Some("iox-shared"),
             Some("inf"),
             Arc::clone(&catalog),
             connection,
@@ -580,7 +580,7 @@ mod tests {
             .await
             .expect("started transaction");
         txn.topics()
-            .create_or_get("iox_shared")
+            .create_or_get("iox-shared")
             .await
             .expect("topic created");
         let (connection, _join_handle, _requests) = create_test_shard_service(MapToShardResponse {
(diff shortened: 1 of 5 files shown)


import/src/aggregate_tsm_schema/update_catalog.rs
```diff
@@ -494,7 +494,7 @@ mod tests {
             .repositories()
             .await
             .topics()
-            .create_or_get("iox_shared")
+            .create_or_get("iox-shared")
```
crepererum, line 497: IIRC this is an invalid Kafka topic name and was actually "fixed" a while back to contain an underscore instead of a `-`. Note that our CI may not detect this because we use Redpanda.

lukebond: i have no experience with this, just some drive-by googling here:

https://github.com/apache/kafka/blob/a126e3a622f2b7142f3543b9dbee54b6412ba9d8/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L33

and

https://github.com/apache/kafka/blob/a126e3a622f2b7142f3543b9dbee54b6412ba9d8/clients/src/main/java/org/apache/kafka/common/internals/Topic.java#L55-L56

seems to me that both `-` and `_` are valid characters in a topic name. although maybe that's not the whole story. do you have anything more concrete you can share?

crepererum: > do you have anything more concrete you can share?

Sadly not, maybe just works :shrug:

crepererum: Found it: #3747

querier/src/cache/read_buffer.rs
```diff
@@ -201,7 +201,7 @@ async fn read_buffer_chunk_from_stream(
     let schema = stream.schema();
 
     // create "global" metric object, so that we don't blow up prometheus w/ too many metrics
-    let metrics = ChunkMetrics::new(metric_registry, "iox_shared");
+    let metrics = ChunkMetrics::new(metric_registry, "iox-shared");
```
lukebond, line 204: @alamb i think this one shouldn't be changed- it's a DB name rather than a Kafka topic and it's not in a test so would probably break in prod by pointing to something different to what it did before?

querier/src/cache/read_buffer.rs
```diff
@@ -512,7 +512,7 @@ mod tests {
             .get_instrument("read_buffer_row_group_total")
             .unwrap();
         let v = g
-            .get_observer(&Attributes::from(&[("db_name", "iox_shared")]))
+            .get_observer(&Attributes::from(&[("db_name", "iox-shared")]))
```
lukebond, line 515: likewise this one i believe, see previous comment

lukebond (changes requested):
you know this code better than me @alamb so LMK if i'm mistaken on these comments.

regarding the comment thread with @crepererum, i think it should be fine to merge and find out. as per the issue this PR closes (https://github.com/influxdata/influxdb_iox/issues/5403), we use this topic name in prod so i think it's fine. besides, this PR mostly aligns the tests to match that to avoid confusion.

import/src/aggregate_tsm_schema/update_catalog.rs
lukebond: looks like this PR is updating the tests to match what you did in that PR!

service_grpc_catalog/src/lib.rs
```diff
@@ -210,10 +210,10 @@ mod tests {
             let metrics = Arc::new(metric::Registry::default());
             let catalog = Arc::new(MemCatalog::new(metrics));
             let mut repos = catalog.repositories().await;
-            let topic = repos.topics().create_or_get("iox_shared").await.unwrap();
+            let topic = repos.topics().create_or_get("iox_-shared").await.unwrap();
```
crepererum, line 213:
```suggestion
            let topic = repos.topics().create_or_get("iox-shared").await.unwrap();
```

crepererum (changes requested):

querier/src/cache/read_buffer.rs
crepererum: This one is fine. It's a placeholder name since we don't have per-DB/per-namespace metrics anymore.

crepererum (approved):

querier/src/cache/read_buffer.rs
lukebond: same for this one @crepererum ?

crepererum: yes

lukebond:
thanks @gulio-python for doing this work, sorry it took so long to get merged

alamb:
🎉  thanks all!

PR #5562, opened 2022-09-05
refactor: avoid double-hash in `AddessableHeap::insert`

crepererum:
Instead of a naive "remove + insert", use a proper insertion routine
that touches the hash map only once.

Note that in case of an override (i.e. the entry with this key already
existed) we need to touch the heap twice, because the sort order likely
changed (we don't optimize the "identical order" case here because it is
pretty unlikely that this will happen in practice).

**Perf results:**

```text
insert_n_elements/0     time:   [16.489 ns 16.497 ns 16.506 ns]
                        change: [-8.1154% -7.9967% -7.8990%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
insert_n_elements/1     time:   [59.806 ns 59.839 ns 59.875 ns]
                        change: [-14.241% -14.160% -14.086%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  7 (7.00%) high mild
  1 (1.00%) high severe
insert_n_elements/10    time:   [601.58 ns 602.26 ns 603.09 ns]
                        change: [-20.870% -20.714% -20.565%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  4 (4.00%) high mild
  5 (5.00%) high severe
insert_n_elements/100   time:   [6.9096 µs 6.9161 µs 6.9246 µs]
                        change: [-18.759% -18.667% -18.553%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
insert_n_elements/1000  time:   [107.71 µs 107.76 µs 107.82 µs]
                        change: [-14.564% -14.427% -14.295%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
insert_n_elements/10000 time:   [2.8642 ms 2.8700 ms 2.8765 ms]
                        change: [-11.079% -10.860% -10.605%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  15 (15.00%) high severe
````

Note that the results are even better for keys with more expansive hash
functions (we have few in the querier).

### CODE REVIEW

File: cache_system/src/addressable_heap.rs [hunk 1]
```diff
@@ -56,15 +56,26 @@ where
...
-        assert!(
-            self.key_to_order_and_value
-                .insert(k.clone(), (v, o.clone()))
-                .is_none(),
-            "entry should have been removed by now"
-        );
+        let result = match self.key_to_order_and_value.entry(k.clone()) {
+            hash_map::Entry::Occupied(mut entry_o) => {
+                // `entry_o.replace_entry(...)` is not stabel yet, see https://github.com/rust-lang/rust/issues/44286
+                let mut tmp = (v, o.clone());
```
alamb, line 62: Why is this `o` cloned (also in the `hash_map::Entry::Vacant` branch)? Maybe I missed it but it seems like since this function gets an owned `o` it could avoid calling clone.

Also, if you want to further optimize, you might want to consider https://crates.io/crates/hashbrown which has these lower level APIs and that we have had good luck using in arrow and datafusion

alamb (approved):
👍

cache_system/src/addressable_heap.rs
crepererum: Because both `o` and `k` are used in two data structures: the hashmap and the queue/heap (see `self.queue.insert(...)` a few lines down). So they are cloned. We could wrap them into an `Rc` or something to avoid that, but in reality `o` is mostly `Time` (which even implements `Copy` and is pretty cheap to clone) and `k` is either some ID (implementing `Copy`) or some `Arc` (not super cheap but still cheap enough I think). So I think this is good enough for now.

alamb: I agree

Issue #5055, opened 2022-07-06
parquet files marked "to_delete" in the catalog are never actually deleted


alamb:
Originally posted by @carols10cents in https://github.com/influxdata/influxdb_iox/issues/4577#issuecomment-1176263221

We have implemented code that marks catalog entries of compacted files with `delete = true`. There is code implemented in https://github.com/influxdata/influxdb_iox/pull/4131 to actually delete those entries from the catalog and the corresponding files, but that code isn't actually called anywhere yet:

```
$ git checkout main # e193913
$ grep garbage_collector
compactor/src/lib.rs
15:pub mod garbage_collector;
$
```

@carols10cents  notes that there are still questions about [if the granularity of removing files is correct, how and how often the task cleaning up to_delete=true files should run](https://github.com/influxdata/influxdb_iox/pull/4131#pullrequestreview-921855435), and [what "old" means](https://github.com/influxdata/influxdb_iox/blob/425b8a63cf6347ef232e255e63cdd84378f0f25d/compactor/src/garbage_collector.rs#L55).

# Definition of Done:
- [x] Resolve outstanding questions about frequency and where to run this
- [ ] Implement the plan

alamb:
I propose running the code implemented in https://github.com/influxdata/influxdb_iox/pull/4131 to actually delete those entries from the catalog and the corresponding files from object store once when the `garbage_collector` service starts up. 

Namely https://github.com/influxdata/influxdb_iox/tree/main/garbage_collector/src

So each time it gets deployed it would clear out old catalog entries first and then proceed to delete unreferenced object store files

alamb:
@carols10cents  / @NGA-TRAN  / @pauldix / @crepererum   do you have any issues with removing parquet files that were marked "to delete" from the catalog / object store now?

crepererum:
@alamb I don't fully understand your question TBH. The files shall be deleted eventually but they MUST be kept around for a while because there might be in-flight queries that use them. This is the whole reason for marking them instead of pruning them from the catalog instantly.

alamb:
@crepererum  there are, confusingly,  two "garbage collectors"

1. The [first](https://github.com/influxdata/influxdb_iox/tree/main/garbage_collector/src) that deletes anything in object store that does not have a catalog reference but does not touch any file that is marked `to_delete` in the catalog
[...]

PR #5567, opened 2022-09-06
feat: parquetfile size trigger

lukebond:
Part of https://github.com/influxdata/conductor/issues/1127

This creates a Postgres trigger to keep track of the total filesize of parquet files per namespace. On insert the total file size per namespace is added to, on delete it is subtracted from. I chose this route instead of querying it because the PF table is huge, there is no index on namespace ID and materialized views on this would also be slow to refresh for this reason.

This has the caveat of ignoring the `to_delete` flag, but I think it's a start we can fine-tune over the next month. I do not yet have information on whether or not we'll charge for those. we have to make a decision one way or another because at present [these are never deleted](https://github.com/influxdata/influxdb_iox/issues/5055).

If you try this locally make sure you run the integration tests as described in the comment, with a postgres instance available with the latest migrations.

NOTE: after implementing this @stuartcarnie suggested that the query could be made to work efficiently with the right indexes. I'm going to explore that alongside using our now-unused (since we switched to RDS), very large prod database volume. I still want this PR reviewed so we can make a proper assessment of the two approaches. I won't merge it right away if approved.


cannonpalms (approved):

alamb:
If you want a query to go really fast, in my experience, you need to get rid of the join.

NGA-TRAN:
@lukebond 
I do not know how we bill but parquet_files are created/inserted by both Ingester and Compactor which will start your trigger. Every time a compactor inserts a new file, it does not delete the compacted files immediately. It only makes `to_delete` to not null. The file will be deleted by garbage collector at some point later.

Regarding index, I agree with you creating index on `parquet_file.namespace_id` won't hep because we do not filter on that column.
We may consider to create index on `billing_summary.namespace_id`. I think it will help with your update. Try EXPLAIN on you insert/update query (explain is not always helpful on insert/update queries). Or we can wait a day or two until we ingest a lot of data in and measure this insert/update query performance.

As an extra note that related to the work in Ingester, Compactor, this trigger, and all indexes on parquet_file that affect the issue of huge Catalog we are discussing. We do a lot of ingestion and compaction, everytime we insert a row in parquet_file, we have to also do the work of this trigger and change all indexes created on parquet file. 
@domodwyer and @pauldix : this is something for us to think about during our catalog discussion

lukebond:
> @lukebond I do not know how we bill but parquet_files are created/inserted by both Ingester and Compactor which will start your trigger. Every time a compactor inserts a new file, it does not delete the compacted files immediately. It only makes `to_delete` to not null. The file will be deleted by garbage collector at some point later.

this is accounted for in the latest commit i pushed just before your comment. we won't bill for the files flagged for deletion.

> Regarding index, I agree with you creating index on `parquet_file.namespace_id` won't hep because we do not filter on that column. We may consider to create index on `billing_summary.namespace_id`. I think it will help with your update. Try EXPLAIN on you insert/update query (explain is not always helpful on insert/update queries). Or we can wait a day or two until we ingest a lot of data in and measure this insert/update query performance.

yes, good idea to add an index on billing_summary.namespace_id

> 
> As an extra note that related to the work in Ingester, Compactor, this trigger, and all indexes on parquet_file that affect the issue of huge Catalog we are discussing. We do a lot of ingestion and compaction, everytime we insert a row in parquet_file, we have to also do the work of this trigger and change all indexes created on parquet file. @domodwyer and @pauldix : this is something for us to think about during our catalog discussion

lukebond:
> If you want a query to go really fast, in my experience, you need to get rid of the join.

i'm not sure what this refers to? when experimenting with the index i was mostly using a query without a join. in the end i didn't go down the route Stuart suggested and this PR introduces no join.

alamb:
> i'm not sure what this refers to? when experimenting with the index i was mostly using a query without a join. in the end i didn't go down the route Stuart suggested and this PR introduces no join.

I was referring to the query with a join in https://github.com/influxdata/influxdb_iox/pull/5567#issuecomment-1238033426. It is clear from re-reading this PR you had already removed the join by the time I commented, so I apologize for introducing more noise into the process

stuartcarnie:
Looks fine @lukebond! 👍🏻 

**TIL**: PostgreSQL doesn't allow you to specify index hints (I used this type of feature on MySQL and MSSQL previously).

## Other musings

I'm surprised that PG didn't use the index. I wonder if 

```sql
ANALYZE parquet_file
```

might have helped, if table statistics were out of date 🤷🏻 

Another debugging feature is the following, to disable sequential scans, which might force it to try the index (only interesting to see if the index is faster)

```
SET enable_seqscan = off;
```

lukebond:
thanks for those tips, @stuartcarnie.

i tried running analyze and it still didn't use the index. tried turning off seq scan and it did use the index i created, but the cost was worse:

```
iox_catalog=# EXPLAIN SELECT namespace_id, SUM(parquet_file.file_size_bytes) AS namespace_bytes FROM parquet_file GROUP BY namespace_id;
                                                QUERY PLAN
-----------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=2187890.31..2188463.92 rows=2199 width=40)
   Group Key: namespace_id
   ->  Gather Merge  (cost=2187890.31..2188403.45 rows=4398 width=40)
         Workers Planned: 2
         ->  Sort  (cost=2186890.29..2186895.78 rows=2199 width=40)
               Sort Key: namespace_id
               ->  Partial HashAggregate  (cost=2186740.73..2186768.21 rows=2199 width=40)
                     Group Key: namespace_id
                     ->  Parallel Seq Scan on parquet_file  (cost=0.00..2074095.15 rows=22529115 width=16)
 JIT:
   Functions: 7
   Options: Inlining true, Optimization true, Expressions true, Deforming true
(12 rows)

iox_catalog=# SET enable_seqscan = off;
SET
iox_catalog=# EXPLAIN SELECT namespace_id, SUM(parquet_file.file_size_bytes) AS namespace_bytes FROM parquet_file GROUP BY namespace_id;
                                                                      QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize GroupAggregate  (cost=1000.59..35973388.73 rows=2199 width=40)
   Group Key: namespace_id
   ->  Gather Merge  (cost=1000.59..35973328.26 rows=4398 width=40)
         Workers Planned: 2
         ->  Partial GroupAggregate  (cost=0.56..35971820.59 rows=2199 width=40)
               Group Key: namespace_id
               ->  Parallel Index Only Scan using parquet_file_namespace_filesize_idx on parquet_file  (cost=0.56..35859147.53 rows=22529115 width=16)
 JIT:
   Functions: 6
   Options: Inlining true, Optimization true, Expressions true, Deforming true
(10 rows)
```

that seals it for me. i think this is how billing needs to be done.

PR #5568, opened 2022-09-06
refactor: do not always box `FunctionEstimator` function

crepererum:
This is similar to #5530. Do not box the function because the user will
likely use it as `Box<dyn ResourceEstimator<...>>` or
`Arc<dyn ResourceEstimator<...>>`.

The slightly larger code change is because the clean-up around `TestSize`
which helps with the doctests (and also we had two copies of that type).


cache_system/src/resource_consumption.rs
```diff
@@ -135,3 +170,17 @@ mod tests {
...
+mod tests {
+    use crate::resource_consumption::test_util::TestSize;
+
+    use super::*;
+
+    #[test]
+    fn test_function_estimator() {
+        let estimator =
+            FunctionEstimator::new(|k: &u8, v: &u16| TestSize((*k as usize) * 10 + (*v as usize)));
+        assert_eq!(estimator.consumption(&3, &2), TestSize(32));
```
alamb, line 184: 👍

alamb (approved):

PR #5571, opened 2022-09-06
chore: `cargo update` (w/ some help)

crepererum:
`workspace-hack` sometimes lists features that are not available in
newer version, preventing `cargo update` and dependabot from updating
dependencies. So from time to time we need some manual intervention:

1. remove all deps from `workspace-hack/Cargo.toml`
2. `cargo update`
3. `cargo hakari generate`
4. `cargo check --workspace --all-targets --all-features`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "0.7.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


crepererum:
Failure is reproducible, will debug that later...

crepererum:
Current hypothesis: some dependency update results in the tests actually being run in parallel and now we start multiple IOx processes listening on the same port. TBH this should never have worked, but I'm going to split up this PR just to figure out which dependency is the root cause.

PR #5573, opened 2022-09-06
feat: two input catalogs to compactor for read and write

NGA-TRAN:
Closes https://github.com/influxdata/conductor/issues/1139


NGA-TRAN:
@carols10cents  Let me catch up with you why we need this.
Test `compact_and_get_remote_partition` fails and I would like to brainstorm with you about a fix.

test_helpers_end_to_end/src/mini_cluster.rs
```diff
@@ -308,8 +308,14 @@ impl MiniCluster {
 
         log_command(command);
 
-        command.ok().unwrap();
-        dump_log_to_stdout("compactor run-once", &log_path);
```
carols10cents, line 312: I noticed it was hard to see why the command was failing in the CI output because this `unwrap` was preventing the log from being printed, so I added a commit to fix that. Next will be actually fixing the issue; I'm researching the best options in Clap we have for that :)

carols10cents (commented):
One tiny suggestion

NGA-TRAN:
From [slack conversation](https://influxdata.slack.com/archives/CRK9M8L5Q/p1662565300695079)  and the fact that Compactors' request are not much now, let us not consider this yet

PR #5579, opened 2022-09-07
refactor: disable read buffer usage in querier

crepererum:
Let's try again to disable the read buffer usage within the querier and only rely on parquet files. In contrast to the previous experiment #5252 (rolled back in #5268) we've made the following changes:

- **predicate pushdown:** make the parquet deserialization aware of predicates (#5531)
- **object store cache:** add a cache layer for the object store (initial impl. in #5527, actual usage in this PR here)

The PR is deliberately minimal. The actual removal of the read buffer code paths will follow later if the performance experiment is successful.

A quick&dirty local tests suggested that we might be slightly slower than the baseline but within an acceptable range.


crepererum:
This is a draft because I want to get prod a bit more stable before merging this, so the performance change is easier to spot.

querier/src/chunk/mod.rs
```diff
@@ -359,10 +359,10 @@ impl ChunkAdapter {
     /// Create new adapter with empty cache.
     pub fn new(
         catalog_cache: Arc<CatalogCache>,
-        store: ParquetStorage,
         metric_registry: Arc<metric::Registry>,
         load_settings: HashMap<ParquetFileId, QuerierChunkLoadSetting>,
     ) -> Self {
+        let store = ParquetStorage::new(Arc::clone(catalog_cache.object_store().object_store()));
```
alamb, line 365: Can you explain the rationale for moving the creation of `ParquetStorage` down here rather than passing it down. I don't see anything wrong with that, but I don't understand the need for it either.

querier/src/chunk/mod.rs
```diff
@@ -189,19 +189,19 @@ impl From<Arc<RBChunk>> for ChunkStage {
...
     /// This forces the querier to load the read buffer for this chunk.
     ReadBufferOnly,
 
-    /// Default "on-demand" handling.
+    /// "On-demand" handling.
     ///
     /// When the chunk is created, it will look up if there is already read buffer data loaded (or loading is already in
     /// progress) and is that. Otherwise it will parquet. If the actual row data is requested (via
     /// [`QueryChunk::read_filter`](iox_query::QueryChunk::read_filter)) then it will force-load the read buffer.
-    #[default]
```
alamb, line 204: This is the change to stop using the read buffer 👍

alamb (approved):
> object store cache: add a cache layer for the object store (initial impl. in https://github.com/influxdata/influxdb_iox/pull/5527, actual usage in this PR here)

I may have missed it but I didn't see any code in this PR that would enable object_store caching

I also looked at the querier dashboard in production, and it didn't seem to be caching object_store data (yet): https://grafana.prod01-eu-central-1.aws.influxdata.io/d/HhX2h_wnz/querier?orgId=1&viewPanel=38

I wonder if that is intentional? Or will skipping the read buffer cache reduce RAM pressure sufficiently to allow object_store data to be cached?


I love this PR in general. ❤️  I also wanted to note that, there is some exciting ongoing performance improvements to the parquet reader cooking in DataFusion: https://github.com/apache/arrow-datafusion/pull/3380 so I am feeling good about this direction in general

querier/src/chunk/mod.rs
alamb: Ah, that makes sense -- so this change makes the read from the cache rather than directly from the object store 👍

PR #5582, opened 2022-09-07
fix: support `u64` in `query_functions::selectors`

crepererum:
This this to be used for InfluxRPC and is currently panicking the prod querier from time to time.


alamb (approved):
Thank you

Issue #5511, opened 2022-08-30
Avoid creating empty splits during compaction


domodwyer:
When running compaction, we can simultaneously split multiple chunks at multiple timestamps to produce the desired number of output files:

```
                      ────────────── Time ────────────▶
              
                               │              │
              ┌────────────────────────┐
              │        Chunk 1 │       │      │
              └────────────────────────┘
                            ┌──┼──────────────┼──────┐
                            │        Chunk 2         │
                            └──┼──────────────┼──────┘
              
                               │              │
              
                            Split T1       Split T2
```

In the above example, we create 3 output streams (resulting in files):

* Before T1
* After T1, before T2
* After T2

However theoretically we can currently can generate plans that are guaranteed to produce at least one empty output stream:

```
                               │               │
    ┌────────────────────────┐                   ┌────────────────────────┐
    │        Chunk 1         │ │               │ │        Chunk 2         │
    └────────────────────────┘                   └────────────────────────┘
                               │               │

                            Split T1       Split T2
```

In the above, we can determine that one of the split points is redundant and can be removed by examining the min/max chunk timestamp statistics:

```
                               │
    ┌────────────────────────┐                   ┌────────────────────────┐
    │        Chunk 1         │ │                 │        Chunk 2         │
    └────────────────────────┘                   └────────────────────────┘
                               │

                            Split T1
```

This results in the same output (2 files) and avoids wasted `SplitExec` work + overhead, though I don't believe the overhead is prohibitively large - this is probably more a nice-to-have improvement.

Joe-Blount:
I suspect this is also possible:
```
                        │               │
    ┌──────────────┐  ┌─────────────────────────┐
    │   Chunk 1    │  │ │ Chunk 2       │       │
    └──────────────┘  └─────────────────────────┘
                        │               │

                    Split T1       Split T2
```
Where `Chunk 2` by chance has a gap in its data, so it has nothing between `Split T1` and `Split T2`.  But this can't be identified until we look at the data in the chunks.

PR #5583, opened 2022-09-07
fix: avoid splitting compaction output for time ranges with no chunks

Joe-Blount:
Closes https://github.com/influxdata/influxdb_iox/issues/5511

When computing split times for compaction output, it is assumed that data is dispersed evenly across the time range.  If large gaps exist between chunks, there may be time splits that have no chunks covering them.  This produces empty parquet files.  This PR checks the computed split times against the time ranges of the chunks to be included in the compaction.  If no chunks exist within the time range of split, the split is omitted.

It is still possible to have empty parquet files when chunks have large gaps within them, rather than between them.

### DIFF
--- a/compactor/src/utils.rs
+++ b/compactor/src/utils.rs
@@ -179,7 +179,7 @@ impl ParquetFileWithTombstone {
 ///     13 = 7 (previous time) + 6 (time range)
 ///     19 = 13 (previous time) + 6 (time range)
 pub(crate) fn compute_split_time(
-    chunk_times: Vec<(i64, i64)>,   // Vec of (min_time, max_time)
+    chunk_times: Vec<(i64, i64)>, // Vec of (min_time, max_time)
     min_time: i64,
     max_time: i64,
     total_size: u64,
@@ -214,14 +214,14 @@ pub(crate) fn compute_split_time(
 
 // time_range_present returns true if the given time range is included in any of the chunks.
 fn time_range_present(
-    chunk_times: Vec<(i64, i64)>,   // Vec of (min_time, max_time)
+    chunk_times: Vec<(i64, i64)>, // Vec of (min_time, max_time)
     min_time: i64,
     max_time: i64,
 ) -> bool {
     for chunk in chunk_times {
         // if chunk.max >= min_time and chunk.min <= max_time, this chunk is within the given time range.
         if chunk.1 >= min_time && chunk.0 <= max_time {
-            return true
+            return true;
         }
     }
     false
@@ -240,20 +240,38 @@ mod tests {
         let chunk_times = vec![(min_time, max_time)];
 
         // no split
-        let result = compute_split_time(chunk_times.clone(), min_time, max_time, total_size, max_desired_file_size);
+        let result = compute_split_time(
+            chunk_times.clone(),
+            min_time,
+            max_time,
+            total_size,
+            max_desired_file_size,
+        );
         assert_eq!(result.len(), 1);
         assert_eq!(result[0], max_time);
 
         // split 70% and 30%
         let max_desired_file_size = 70;
-        let result = compute_split_time(chunk_times.clone(), min_time, max_time, total_size, max_desired_file_size);
+        let result = compute_split_time(
+            chunk_times.clone(),
+            min_time,
+            max_time,
+            total_size,
+            max_desired_file_size,
+        );
         // only need to store the last split time
         assert_eq!(result.len(), 1);
         assert_eq!(result[0], 8); // = 1 (min_time) + 7
 
         // split 40%, 40%, 20%
         let max_desired_file_size = 40;
-        let result = compute_split_time(chunk_times, min_time, max_time, total_size, max_desired_file_size);
+        let result = compute_split_time(
+            chunk_times,
+            min_time,
+            max_time,
+            total_size,
+            max_desired_file_size,
+        );
         // store first and second split time
         assert_eq!(result.len(), 2);
         assert_eq!(result[0], 5); // = 1 (min_time) + 4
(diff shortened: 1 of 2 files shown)


compactor/src/utils.rs
```diff
@@ -179,6 +179,7 @@ impl ParquetFileWithTombstone {
 ///     13 = 7 (previous time) + 6 (time range)
 ///     19 = 13 (previous time) + 6 (time range)
 pub(crate) fn compute_split_time(
```
NGA-TRAN, line 181: Since this function now has different behavior, you may want to update the comments to reflect the changes.

An extra note for us to think after this PR is merged. Our goal is to split this into n times but with this change, we will split it into <=n times. I wonder how hard it will be if we still be able to split this into n times if we also consider sizes of the chunks 🤔

NGA-TRAN (approved):
Looks good, @Joe-Blount . Just a minor suggestion to use struct instead of tuple

NGA-TRAN:
@carols10cents : can you have quick review?

compactor/src/utils.rs
```diff
@@ -179,6 +179,7 @@ impl ParquetFileWithTombstone {
...
+    chunk_times: Vec<(i64, i64)>, // Vec of (min_time, max_time)
```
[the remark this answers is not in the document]
carols10cents: Or does using the [`TimestampMinMax` type](https://github.com/influxdata/influxdb_iox/blob/e01b64569607bcf1996150cda6554d4b9b60685a/data_types/src/lib.rs#L2204-L2214) make sense? If so, you can call [`timestamp_min_max().unwrap()`](https://github.com/influxdata/influxdb_iox/blob/e01b64569607bcf1996150cda6554d4b9b60685a/compactor/src/query.rs#L140-L145) rather than `min_time()` `max_time()` above

compactor/src/utils.rs
```diff
@@ -199,17 +200,33 @@ pub(crate) fn compute_split_time(
...
     loop {
         let split_time = min + ((max_time - min_time) as f64 * percentage).ceil() as i64;
-        if split_time < max_time {
-            split_times.push(split_time);
-            min = split_time;
-        } else {
+
+        if split_time >= max_time {
             break;
+        } else if time_range_present(chunk_times.clone(), min, split_time) {
```
carols10cents, line 206: Why is this `clone` needed? While it *shouldn't* be too expensive to clone a bunch of `i64`s, I don't think it's *needed*. I think this can be a reference:

```suggestion
        } else if time_range_present(&chunk_times, min, split_time) {
```

and then more changes coming below...

compactor/src/utils.rs
Joe-Blount: >I wonder how hard it will be if we still be able to split this into n times if we also consider sizes of the chunks 

When detecting a gap, and deciding to skip a split, it would be easy to increment n to split into smaller ranges and start over evaluating the splits.  But that will move all the split times so the gap may not line up with the smaller splitting, so we may not be able to skip a split after incrementing n.

I think what's best here depends on the intent of the max file size. Is the max truly a max we'd like to stay under, or is it a target we're trying to approach?

For example, we might call `compute_split_time` expecting it to split the time range into 4 ranges, but because of a gap we split it into 5 ranges (hoping to return 4).  But then maybe we can't skip any because the gap doesn't line up anymore, so we return all 5.  In this situation, would you rather have 3 splits (knowing at least one is >max), or 5 splits (knowing at least one is probably very small)?

compactor/src/utils.rs
```diff
@@ -199,17 +200,33 @@ pub(crate) fn compute_split_time(
...
+    max_time: i64,
+) -> bool {
+    for chunk in chunk_times {
+        // if chunk.max >= min_time and chunk.min <= max_time, this chunk is within the given time range.
+        if chunk.1 >= min_time && chunk.0 <= max_time {
+            return true;
+        }
+    }
+    false
+}
```
carols10cents, line 228: So this function can borrow a slice of `chunk_times` to eliminate the `clone` above, and then [`Iterator::any`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) does exactly what you've implemented here:

```suggestion
fn time_range_present(
    chunk_times: &[(i64, i64)],
    min_time: i64,
    max_time: i64,
) -> bool {
    chunk_times.iter().any(|&(chunk_min, chunk_max)| {
        chunk_max >= min_time && chunk_min <= max_time
    })
}
```

Again, please let me know if this doesn't make sense! I highly recommend to folks new to Rust to skim through the `Iterator` methods so that you have an idea of all the capabilities it provides for you :)

carols10cents (commented):
Just a few tips that are totally normal new-to-Rust things :)

compactor/src/utils.rs
NGA-TRAN: >  I think what's best here depends on the intent of the max file size. Is the max truly a max we'd like to stay under, or is it a target we're trying to approach?

Our goals is to have the output files each is around the size of  `max_desired_file_size_bytes` and we want to use the available `split_plan` to do so. `split_plan` builds a plan to split the output into n+1 streams based on n input times.

> For example, we might call compute_split_time expecting it to split the time range into 4 ranges, but because of a gap we split it into 5 ranges (hoping to return 4). But then maybe we can't skip any because the gap doesn't line up anymore, so we return all 5. In this situation, would you rather have 3 splits (knowing at least one is >max), or 5 splits (knowing at least one is probably very small)?

If we do not want any file > `max_desired_file_size_bytes`, then we will go with 5. But I do not think we have to strictly follow that. Maybe a good design discussion after you propose your approach

compactor/src/utils.rs
Joe-Blount: Thanks!  I tried to get rid of the clone using an Arc, but didn't I didn't get that to compile.

NGA-TRAN (approved):

PR #5588, opened 2022-09-08
feat: garbage collector now cleans up old parquet files

### LINKED ISSUE #5055: parquet files marked "to_delete" in the catalog are never actually deleted

lukebond:
Closes #5055

This PR introduces catalog parquet file cleanup to the garbage collector, and refactors it into two modules (one for object storage GC, plus this new one for parqet files).

I couldn't think of a way to test it. It does one thing and sleeps, and that one thing is tested elsewhere. :woman_shrugging:  Happy to take ideas.

*NOTE:* this changes command-line args and env vars, so I won't merge it until a corresponding PR is created for k8s-idpe, which I will link to here. Even if I merge both simultaneously there will likely be a moment where the GC is failing with bad k8s-idpe config, but it won't be a problem and won't be for more than a few minutes.


garbage_collector/src/parquetfile/deleter.rs
```diff
@@ -0,0 +1,46 @@
...
+            .repositories()
+            .await
+            .parquet_files()
+            .delete_old(older_than)
+            .await
+            .context(DeletingSnafu)?;
+        info!(delete_count = %deleted.len(), "iox_catalog::delete_old()");
+
+        // until next time
+        sleep(Duration::from_secs(60 * sleep_interval_minutes)).await;
```
crepererum, line 32: You probably wanna `select` the shutdown token here as well, otherwise we'll wait quite a long time before the compactor stops.

lukebond: thank you, fixed now

garbage_collector/src/objectstore/checker.rs
```diff
@@ -28,15 +28,16 @@ pub(crate) type Result<T, E = Error> = std::result::Result<T, E>;
 
 pub(crate) async fn perform(
     catalog: Arc<dyn Catalog>,
-    cutoff: DateTime<Utc>,
+    cutoff_days: usize,
```
crepererum, line 31: Wouldn't `cutoff: Duration` be the better typed solution?

garbage_collector/src/parquetfile/deleter.rs
```diff
@@ -0,0 +1,51 @@
+use data_types::Timestamp;
+use iox_catalog::interface::Catalog;
+use observability_deps::tracing::*;
+use snafu::prelude::*;
+use std::{sync::Arc, time::Duration};
+use tokio::{sync::broadcast, time::sleep};
+
+const DAY_SECONDS: u64 = 24 * 60 * 60;
```
crepererum, line 8: You could avoid magic constants by using a typed `cutoff` type.

iox_catalog/src/interface.rs
```diff
@@ -543,6 +543,10 @@ pub trait ParquetFileRepo: Send + Sync {
     /// Returns the deleted records.
     async fn delete_old(&mut self, older_than: Timestamp) -> Result<Vec<ParquetFile>>;
 
+    /// Delete all parquet files that were marked to be deleted earlier than the specified time.
+    /// Returns the deleted IDs only.
+    async fn delete_old_ids_only(&mut self, older_than: Timestamp) -> Result<Vec<ParquetFileId>>;
```
crepererum, line 548: This needs a test within `iox_catalog` (see this very module).

crepererum (commented):
Nits regarding testing and typing.

garbage_collector/src/objectstore/lister.rs
```diff
@@ -28,7 +30,11 @@ pub(crate) async fn perform(
                         checker.send(item).await?;
                     }
                     None => {
-                        break;
+                        // sleep for the configured time, then list again and go around the loop
+                        // again
+                        sleep(Duration::from_secs(60 * sleep_interval_minutes)).await;
```
crepererum, line 35: Note for other reviewers: this sleep can be cancelled due to the outer select statement that contains a `shutdown.recv()` arm.

crepererum (approved):
Let's ship this!

alamb:
So great to see this finally land. Thanks @lukebond and @crepererum

PR #5593, opened 2022-09-08
fix: adjust the split count when gaps are detected in the chunks

Joe-Blount:
This is a followup to https://github.com/influxdata/influxdb_iox/pull/5583, based on [this comment](https://github.com/influxdata/influxdb_iox/pull/5583#discussion_r965254805).

This PR makes two functional changes:
1) When a gap is detected in the chunks, causing a split time range to be omitted, this will increase the number of splits making it more likely to return the desired number of splits, and more likely they remain at or under the maximum file size.  Note that it is possible the gap will not line up to the shifted time splits, in which case it may cause smaller files.
2) The split times are now computed to create splits of equal sized time ranges.  For example, prior to this PR, if asked to split 102MB based on a 100MB max file size, the split time would be chosen to create a 100MB file and a 2MB file.  With this PR, the split time will be chose to create two 51MB files.

### DIFF
--- a/compactor/src/utils.rs
+++ b/compactor/src/utils.rs
@@ -164,8 +164,11 @@ impl ParquetFileWithTombstone {
 /// time range [min_time, max_time] contains.
 /// The split times assume that the data is evenly distributed in the time range and if
 /// that is not the case the resulting files are not guaranteed to be below max_desired_file_size
-/// Hence, the range between two contiguous returned time is percentage of
-/// max_desired_file_size/total_size of the time range
+/// Once the desired split count is computed, the time range will be divided evenly among those splits.
+/// Hence, the range between two contiguous returned times is the percentage of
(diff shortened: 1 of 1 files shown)


NGA-TRAN:
@Joe-Blount 
I think after we have discovered that  https://github.com/influxdata/conductor/issues/1130 is just a false alarm and https://github.com/influxdata/influxdb_iox/pull/5594 has removed those unnecessary warnings, I think we do not need this. More specific:

1. We only need this if we see  warning messages `parquet serialisation encoded 0 rows` and `SplitExec produced an empty result stream` but we have never seen them sine they were introduced hence we very rarely hit this use case. 
2. If we compact 102MB and max out desired size is 100MB, we prefer to split them 100MB and 2MB then next cycle we likely only need to merge the 2MB with the new data.

PR #5594, opened 2022-09-08
fix: remove unnecessary warning

NGA-TRAN:
Closes https://github.com/influxdata/conductor/issues/1130

See https://github.com/influxdata/conductor/issues/1130#issuecomment-1241222955 for the reason of this fix


parquet_file/src/serialize.rs
```diff
@@ -118,22 +118,14 @@ where
...
-            warn!("parquet serialisation stream yielded empty record batch");
-        } else {
+        if batch.num_rows() != 0 {
             writer.write(&batch)?;
         }
     }
 
     let meta = writer.close().map_err(CodecError::from)?;
     if meta.num_rows == 0 {
+        // throw warning if all input batches are empty
```
NGA-TRAN, line 128: This is the right warning that input data (all the batches together) is empty

parquet_file/src/serialize.rs
```diff
@@ -118,22 +118,14 @@ where
...
-        if batch.num_rows() == 0 {
-            // It is likely this is a logical error, where the execution plan is
-            // producing no output, and therefore we're wasting CPU time by
-            // running it.
-            //
-            // Unfortunately it is not always possible to identify this before
-            // executing the plan, so this code MUST tolerate empty RecordBatch
-            // and even entire files.
-            warn!("parquet serialisation stream yielded empty record batch");
-        } else {
```
NGA-TRAN, line 130: DF plan produces many batches and someone of them can be empty. So no need to throw warning here.

crepererum: Shouldn't we prune out empty batches before creating the parquet file than? I kinda seems pointless to create potentially empty row groups. CC @tustvold

NGA-TRAN: This is the work to improve in DF. Let us not let it prevent us from bombarding our logs unnecessary

NGA-TRAN: There are highly no empty batches in parquet files. These batches are **the results of running a DF query plan**. During the plan, we deduplicate data that may lead to this. Also, DF may partition data and do something else that may lead to this.

NGA-TRAN: | pointless to create potentially empty row groups

As @domodwyer and I discussed earlier, this row group is a part of streaming and we will catch this empty and won't write empty parquet file later. See `upload` function

alamb: The ArrowWriter (that is writing parquet) creates row groups based on a configuration setting rather than the size of the passed RecordBatches (this was not always the case, but @tustvold  fixed this a while ago). Thus in my opinion it would also be fine to simple call `write()` with zero length record batches and not have this special case

Here is the code in question

https://docs.rs/parquet/22.0.0/src/parquet/arrow/arrow_writer/mod.rs.html#133-149

alamb (approved):
make sense to me -- I also think we could possibly make the code even simpler

tustvold (approved):

parquet_file/src/serialize.rs
NGA-TRAN: Thanks @alamb . 

@crepererum  Maybe I can explain a bit clearer. Even if we can write empty row group, we do not use it here either that is the `if` condition is for. The fix for this batch is to not throwing a log warning if some of the batches generated by DF plan are empty which can be the case. We even do extra check below that if all the batches generated by DF plan is empty, we will throw warning. The warning is just for us to know that some of the streams generated by `split_plan` are empty which is legitimate, too, due to the way we split the time.

So the answer is "yes, we already pruned out all empty chunk". This is just stop throwing an unnecessary warning. If you are unclear, I can explain how Compactor plans work with DF plans

domodwyer: I think @crepererum was suggesting that we filter out the empty row groups before calling this fn / at the source, rather than removing the warning, as it's not really the responsibility of the parquet serialisation to decide what gets serialised and what doesn't (which I agree with, and would have preferred).

Practically speaking it doesn't make much difference in the end though 🤷

NGA-TRAN: Yeah, either way we have to iter over the batches and remove empty ones. We can do that before calling this function but as you say `Practically speaking it doesn't make much difference in the end though`.  I hope @alamb and @tustvold will improve this in DF to not return empty batches

alamb: I propose removing the check entirely: https://github.com/influxdata/influxdb_iox/pull/5620 which I think makes it clear that the parquet_file crate doesn't have to do anything special with/for empty record batches

crepererum: Wanna point out that we should aim for clean Parquet files that are not empty and have no empty row groups to minimize the overhead the querier has to deal with, esp. when we remove the RB. I think the "empty row group" issue is now solved upstream, the "empty file" issue could be solved by IOx (although it's the less important one).

PR #5600, opened 2022-09-09
chore: Allow running all-in-one with external object store

mkmik:
I wanted to test some interactions with a real object-store and I noticed that we don't expose the object store configuration block to the all-in-one command.

This PR exposes the whole object-store clap block and not only `--data-dir`. It remains backward compatible with the behavior of `--data-dir` in the `all-in-one` mode.

```console
$ minio serve &
$ echo "create token"
$ export AWS_ACCESS_KEY_ID="..."
$ export AWS_SECRET_ACCESS_KEY="...."
$ cargo run --features aws -- run all-in-one --object-store s3 --persist-partition-cold-threshold-seconds 10 --aws-endpoint http://localhost:9000 --bucket iox   --aws-allow-http
```


influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -361,7 +359,16 @@ impl Config {
             querier_max_table_query_bytes,
         } = self;
 
-        let object_store_config = ObjectStoreConfig::new(database_directory.clone());
+        let database_directory = object_store_config.database_directory.clone();
+        // If dir location is provided and no object store type is provided, use it also as file object store.
```
crepererum, line 363: Isn't this what `ObjectStoreConfig` is doing by default when creating the actual object cstore?

mkmik: I tried and it wasn't. It was complaining that when you pass the dir you also have to pass obiectstore type file

crepererum: Should we default the type to `file` then?

mkmik: By default it's memory. It becomes file only if you specify a directory. Or at least that's the thinking for the all-in-one mode, where it's helpful to provide a quick shortcut.

For the proper server mode, I like the idea that we should force all flags to be explicit and minimize the "magic activations" between flags.

alamb (approved):
This makes sense to me -- thanks @mkmik

PR #5603, opened 2022-09-09
revert: disable read buffer usage in querier (#5579)

crepererum:
# gRPC perf
This results in a 2x-3x slow down. It's not horrible, but also not good.

image

# CPU Usage
Also it increases CPU usage (which would be OK, because our queriers have loads of spare CPU power, but maybe we can improve on that front as well):

image

# Cache Usage
One nice side-effect of using parquet would be that the data is more compressed which would make our cache more efficient:
image


tustvold (approved):
Super-stoked by these results, we're almost there! :tada:

PR #5613, opened 2022-09-12
fix: apply selection in `TestChunk::read_filter`

crepererum:
TBH I have no idea how this worked so well before, but the chunks are
expected to apply the given selection. This is because
`IOxReadFilterNode::execute` will wrap the `QueryChunk::read_filter`
output into a `SchemaAdapterStream` and this one expects that there are
no input columns that are absent in the output schema (i.e. it will only
add null columns, it won't remove any). Funnily the `SchemaAdapterStream`
error will blame DataFusion for the mess.

**This doesn't fix any tests that fail right now, but an upcoming PR would fail w/o it and it took me quite a while to debug this.**


iox_query/src/test.rs
```diff
@@ -919,14 +920,24 @@ impl QueryChunk for TestChunk {
         &self,
         _ctx: IOxSessionContext,
         predicate: &Predicate,
-        _selection: Selection<'_>,
+        selection: Selection<'_>,
```
alamb, line 923: the fact this is called selection when it is really a projection definitely doesn't help things. I'll make a note to try and clean up the naming

alamb (approved):

iox_query/src/test.rs
```diff
@@ -919,14 +920,24 @@ impl QueryChunk for TestChunk {
...
+            None => self.table_data.clone(),
+            Some(projection) => self
+                .table_data
+                .iter()
+                .map(|batch| {
+                    let batch = batch.project(&projection)?;
+                    Ok(Arc::new(batch))
+                })
+                .collect::<std::result::Result<Vec<_>, ArrowError>>()?,
+        };
```
NGA-TRAN, line 940: This is nice 👍 . It has been a while and I forget how this work but I did ask the same question before "why we do not use `_selection` but be able to filter data?". IIRC, I found in a different some place that did this work but again I forget where. I think if we move this filter here, it is clearer and better.

PR #5614, opened 2022-09-12
refactor: make `QueryChunk::partition_id` non-optional

crepererum:
In our data model, a chunk always belongs to a partition[^1], so let's not make this attribute optional. The optional value only leads to -- mostly surprising -- conditional behavior, ranging from "do not equalize the partition sort key" (querier) to "always consider the chunk overlapping" (iox_query when dealing with ingester chunks).

[^1]: This is even true when the chunk belongs to a parquet file that is not
      yet added to the catalog, contrary to what a comment in the ingester
      stated. The catalog and data model used by the querier are two totally
      different things.


NGA-TRAN (commented):
Yeah, it is nice that you can always attach a chunk to a partition now. IIRC, at the early stage when we started building the Ingester, we were not able to attach it to a partition yet. We have evolved and this is a great way to go.

PR #5617, opened 2022-09-12
refactor: do not store chunks in `Deduplicator`

crepererum:
Only store context, settings (if any) and the schema interner within the de-duplicator. Extract a new `Chunks` type that handles the chunk classification and can passed around in a somewhat clean fashion.


NGA-TRAN (approved):
Looks very good. Will reduce memory usage 👍 . I have carefully gone over the code, the assert and the tests. But I have to rely on the tests to catch bugs if any 😄

Issue #5602, opened 2022-09-09
Avoid running de-dup within the ingester for querier<>ingester requests


crepererum:
The ingesters are somewhat a bottleneck because we cannot just scale them out. So they should do as little work as possible, esp. for querier requests. So for these requests they should NOT run de-duplication (the queriers can easily do that).

PR #5618, opened 2022-09-12
refactor: dedicated query path for querier tombstone materialization

crepererum:
I would like to remove de-dup from the querier<>ingester query path in #5602, but the tombstone application and parquet write path can still do the full work.


ingester/src/data/query_dedup.rs
```diff
@@ -0,0 +1,160 @@
...
+        .with_chunks([data as _])
+        .build()
+        .context(FrontendSnafu)?
+        .plan_builder
+        .build()
+        .context(LogicalPlanSnafu)?;
+
+    debug!(%table_name, plan=%logical_plan.display_indent_schema(),
+           "created single chunk scan plan");
+
```
NGA-TRAN, line 71: Question: you can remove the below block here because you converted `_selection` to `selection` in previous PR, tight?

```rust
 // Now, restrict to all columns that are relevant
    let logical_plan = match selection {
        Selection::All => logical_plan,
        Selection::Some(cols) => {
            // filter out columns that are not in the schema
            let schema = Arc::clone(logical_plan.schema());
            let cols = cols.iter().filter_map(|col_name| {
                schema
                    .index_of_column_by_name(None, col_name)
                    .ok()
                    .map(|_| datafusion::prelude::col(col_name))
            });

            LogicalPlanBuilder::from(logical_plan)
                .project(cols)
                .context(LogicalPlanSnafu)?
                .build()
                .context(LogicalPlanSnafu)?
        }
    };
```

NGA-TRAN (approved):
Quetion: is this refactor to move the plan with deduplication to query_dedup.rs for you to modify the function in querier_hanlder.rs to not do the deduplication work in next PR?

If so, it looks good to me

crepererum:
> Quetion: is this refactor to move the plan with deduplication to query_dedup.rs for you to modify the function in querier_hanlder.rs to not do the deduplication work in next PR?

Exactly 🙂

ingester/src/data/query_dedup.rs
crepererum: Yes, or in other worlds: the "internal" data paths in the ingested (tombstone materialization, Parquet file generation) always use all columns – in contrast to the querier data path that can select columns. So splitting the two even makes the internal data path simpler (also the internal path never applies predicates apart from delete predicates).

PR #5622, opened 2022-09-12
fix: Use `ColumnType` as an enum in more places; make `estimate_arrow_bytes_for_file` infallible

carols10cents:
I swear I'm going somewhere with this, this is just the first step to some refactoring to improve code reuse, encapsulation, and testing of the compactor code.

`estimate_arrow_bytes_for_file` had the possibility of failing if the column type `i16` couldn't be converted to a `ColumnType` value. This really shouldn't be the case, as we're controlling what goes into and out of the database. And even if there is a failure, it should be at the database interaction level, not computing estimated memory usage.

sqlx can take care of the transformations between int and enum for us if we add `#[repr(i16)]` to `ColumnType` (the `column_type` column in postgres is `SMALLINT`, [which corresponds to Rust `i16`](https://docs.rs/sqlx/latest/sqlx/postgres/types/index.html)).

There's only one spot where we're sending a bunch of records to an INSERT and sqlx isn't doing that quite right-- there's one little `as i16` needed there.

### DIFF
--- a/compactor/src/cold.rs
+++ b/compactor/src/cold.rs
@@ -331,15 +331,15 @@ mod tests {
         table.create_column("time", ColumnType::Time).await;
         let table_column_types = vec![
             ColumnTypeCount {
-                col_type: ColumnType::Tag as i16,
+                col_type: ColumnType::Tag,
                 count: 3,
             },
             ColumnTypeCount {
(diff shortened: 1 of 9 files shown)


NGA-TRAN (approved):
Nice

Issue #5624, opened 2022-09-13
Querier->Ingester projection pushdown


crepererum:
Currently we don't push down projection to the ingester and always request ALL columns. We shall fix that. 

We will do this in 2 steps:
- [ ] Projection pushdown for normal SQL queries #5740
- [ ] Projection pushdown for InfluxRPC-based queries #5741

crepererum:
Some more details:

This is where we would need a sensible subset of columns:

https://github.com/influxdata/influxdb_iox/blob/7e00426d493cc1e0a552ca3bdb7b9de7883d3c23/querier/src/table/mod.rs#L407-L417

For SQL-based queries, we actually have this information (it's just not wired through):

https://github.com/influxdata/influxdb_iox/blob/7e00426d493cc1e0a552ca3bdb7b9de7883d3c23/querier/src/table/query_access/mod.rs#L47

For InfluxRPC-based queries, we currently DON'T request chunks with a specific projection:

https://github.com/influxdata/influxdb_iox/blob/7e00426d493cc1e0a552ca3bdb7b9de7883d3c23/querier/src/namespace/query_access.rs#L39-L44

I guess for some query types this should be possible. So this is mostly a wiring exercise.[^1]

[^1]: I think, it might turn out to be more complicated though.

PR #5626, opened 2022-09-13
refactor: do not run de-dup in ingester for querier requests

### LINKED ISSUE #5602: Avoid running de-dup within the ingester for querier<>ingester requests

crepererum:
This removes the entire de-dup logic from the inegster for querier requests. Furthermore, it even removes the entire datafusion execution from the querier and just dumps the in-memory record batches as quickly as possible. No filters are applied. Note that even prior to this PR, we've never applied projections (tracked by #5624).

**Pros:**

- speed up query planning within the querier (since we need the ingester response for state reconciling)
- lowered ingester CPU load

**Cons:**

- more querier<>ingester network traffic

Closes #5602.


alamb:
Can you please explain a little more of the rationale to stop applying filtering in the ingester (other than that it will reduce latency)?

I ask mostly because I think @pauldix  and @carols10cents  had specific reasons for implementing a plan rather than just sending along the batches, and before we spend time ripping that all out (and the reviews, etc) I wanted to know why

For example, do you have data that shows the relative query performance / time added by running a DF plan on the ingester compared to the expected size increase in sending data? 

Or is the idea to gather this information with this PR and then revert the PR if it turns out to be a bad idea?

crepererum:
> Or is the idea to gather this information with this PR and then revert the PR if it turns out to be a bad idea?

Basically that.

mkmik:
For reference this is the latency heatmap of the ingester requests

image


on a first look I didn't think this was significantly impacting our performance; is the bad tail latency of the ingester impacting query performance more badly than what it seems?

query_tests/cases/in/duplicates.sql
```diff
@@ -1,5 +1,5 @@
 -- Test for predicate push down explains
--- IOX_SETUP: OneMeasurementFourChunksWithDuplicates
+-- IOX_SETUP: OneMeasurementFourChunksWithDuplicatesParquetOnly
```
alamb, line 2: Can you explain why you changed the setup for this query? Does the plan now change depending on the setup?

alamb (approved):
Looks good to me . It looks like there are some more end to end tests that need updating

query_tests/cases/in/duplicates.sql
NGA-TRAN: Looking at the test data, I think the new plans here look more reasonable. I wonder why the the EXPLAIN does not work well for the non-parquet file.

NGA-TRAN:
> we've never applied projections

We either missed it or we did it on purpose to return all for deduplication with existing data, IIRC

NGA-TRAN (approved):
LGTM
I also wonder if this will work better because we have to do deduplicate in Querier side for the data and we have to send extra data. But let's run it and see

crepererum:
> is the bad tail latency of the ingester impacting query performance more badly than what it seems?

That's my estimate. That said, this change is an easy roll-back, so the risk is rather low (esp. since it means less work for the ingester, so prod ain't going to explode with this change).

query_tests/cases/in/duplicates.sql
crepererum: So parquet files are internally de-duplicated (i.e. don't contain PK duplicates). However ingester data is NOT de-dupped now. So if you run the same `EXPLAIN` for both cases, the output is different. If extended the test setup to capture both scenarios (in separate tests!).

alamb: > So parquet files are internally de-duplicated

Yes this makes sense -- the deuplication is done in the persist plan run in the ingester

alamb:
> One other thought is that we should probably update the ingesters to always persist after some set number of rows. At least that'll put an upper bound on filtering time. Given the compactor can pick things up quickly, we could just always persist after 100k rows or over size (this latter is for wide tables). But then we're creating more catalog load.

@pauldix  -- I think @domodwyer implemented this feature in  https://github.com/influxdata/influxdb_iox/pull/5521 (`INFLUXDB_IOX_PERSIST_PARTITION_MAX_ROWS`)

crepererum:
The ingester latency is noticeably worse now:

image

Also the current implementation has a bug which we see in prod: #5711 (also fixed by this PR).

PR #5627, opened 2022-09-13
docs: Add comments about how to see debug logs via `cargo test`

alamb:
# Rationale:

While debugging an issue in https://github.com/influxdata/influxdb_iox/pull/5606 I wanted to see debug logs from the optimizer. I had a bunch of trouble (where the logs I wanted seemed not to appear in stdout) until I added `--no-capture` to the `cargo test` command line invocation. The code last appears to have changed in #3536 from @crepererum  which means I haven't looked for logs in tests for a while 😆 

# Changes
Leave some comments / hints to other readers (including my future self)


test_helpers/src/lib.rs
```diff
@@ -93,6 +95,10 @@ pub fn start_logging() {
 
         let subscriber = FmtSubscriber::builder()
             .with_env_filter(EnvFilter::from_default_env())
+            // Note `with_test_writer` allows cargo-test to capture
+            // logging (which may reorder / hide input) if you don't
+            // see your expected logs, try passing `--no-capture` to
+            // `cargo test`
```
crepererum, line 101: I'm not aware of reordering (within the same test). What the capturing does is to present you the output for failed tests in a "summary". This is super handy for CI systems, but also locally when you run multiple tests (remember: test execution is multi-threaded, so you won't have a good way to disentangle direct stdouts without capturing). I agree that this may conflict w/ debugging. Also your explanation is technically a bit off (don't wanna be too picky, but it may confuse a future reader), so how about:

```suggestion
            // Note `with_test_writer` allows libtest (used for all 
            // our tests and invoked by `cargo test`) to capture
            // per-test logging. The captured data will only be 
            // shown for failed tests. Pass `--no-capture` to 
            // disable that feature (but only try to run a single 
            // test to prevent output interleaving).
```

test_helpers/src/lib.rs
```diff
@@ -78,6 +78,8 @@ static LOG_SETUP: Once = Once::new();
 /// Enables debug logging regardless of the value of RUST_LOG
 /// environment variable. If RUST_LOG isn't specifies, defaults to
 /// "debug"
+///
+/// Hint: try running your test with `--no-capture` if you don't see expected logs
```
crepererum, line 82:
```suggestion
/// Hint: Try running your test with `--no-capture` if you don't see expected logs.
///
///       This should only be used when debugging a single tests, otherwise outputs 
///       will be interleaved because test execution is multi-threaded.
```

crepererum (approved):

test_helpers/src/lib.rs
```diff
@@ -78,6 +78,11 @@ static LOG_SETUP: Once = Once::new();
 /// Enables debug logging regardless of the value of RUST_LOG
 /// environment variable. If RUST_LOG isn't specifies, defaults to
 /// "debug"
+///
+/// Hint: Try running your test with `--no-capture` if you don't see expected logs.
```
alamb, line 82:
```suggestion
/// Hint: Try running your test with `--no-capture --test-threads=1` if you don't see expected logs.
```

test_helpers/src/lib.rs
```diff
@@ -78,6 +78,11 @@ static LOG_SETUP: Once = Once::new();
...
+///
+///       This should only be used when debugging a single tests, otherwise outputs 
+///       will be interleaved because test execution is multi-threaded.
```
alamb, line 85:
```suggestion
/// This is likely useful only when debugging a single tests or when running
/// with `--test-threads=1` , otherwise outputs will be interleaved because 
/// test execution is multi-threaded.
```

alamb: This is a good point @crepererum  -- I also updated your suggestion to include `--test-threads=1` as a way to avoid interleaved output too

Issue #5528, opened 2022-09-01
Support calling influxql/flux 'selector' aggregates from IOx SQL


alamb:
# background

InfluxQL and Flux have specialized timeseries functions called selectors: `first` and `last`
https://docs.influxdata.com/flux/v0.x/stdlib/universe/first/
https://docs.influxdata.com/flux/v0.x/stdlib/universe/last/
https://docs.influxdata.com/influxdb/v1.8/query_language/functions/#first
https://docs.influxdata.com/influxdb/v1.8/query_language/functions/#last

These selectors behave like aggregates in the sense that they take many rows in and produce a single row out

However, they are somewhat specialized as they rely on the value of the timestamp and return the value associated with the earliest and latest time, respectively

In order to support influxrpc ReadAggregate and ReadWindowAggregate IOx implements datafusion extension aggregate functions for this functionality. However, they can not be called from SQL directly (only programatically by the influxrpc --> LogicalPlan conversion logic). 

The flux team (@Marwes  and @wolffcm specifically) are working on implementing Flux in terms of IOx SQL, and to properly push down these selectors, it would be very helpful to call them via SQL

# Details

Basically to get first(host) the query made by the influxrpc frontend looks like

```sql
SELECT selector_first_value(host, time) as host,
       selector_first_time(time) as time
  FROM table;
```

And then there is a custom implementation of `selector_first_value` that tracks the values corresponding to the first timestamp.

IOx implementation:

https://github.com/influxdata/influxdb_iox/blob/b820f69b0552ad4225044d84300edcae4301056d/query_functions/src/selectors.rs#L37-L51

Internal slack context: https://influxdata.slack.com/archives/CRK9M8L5Q/p1661957034784189

jacobmarble:
The cited code mentions a JIRA feature request, which was migrated to GH/arrow-df as https://github.com/apache/arrow-datafusion/issues/600 . Does that Arrow feature request block this IOx feature request?

alamb:
@jacobmarble 

The issue is that a "selector" must return two values: the field's value as well as the corresponding timestamp. SQL aggregates return a single value. 

# Option 1
One possible way of modeling this is to return a single "struct" value that has two fields, aka what https://github.com/apache/arrow-datafusion/issues/600 describes

This would allow you to write SQL like this (accessing `.value` and `.time` fields)
```sql
SELECT 
  f.value, 
  f.time 
FROM 
  (select first(col, time) as f FROM ..)
```

[...]

PR #5628, opened 2022-09-13
feat: Support calling influxql/flux selector aggregates from IOx SQL

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/5528

Note the vast majority of this PR is tests. The rest of basically plumbing.

# Rationale:
The flux team needs (https://github.com/influxdata/idpe/issues/15748 https://github.com/influxdata/idpe/pull/15664) this to be able to push aggregates down in  flux sql

# Changes
- [x] Refactor selectors to provide their output as `structs` in addition to the individual fields / values
- [x] Register `first`, `last`, `min` and `max`  as two function (e.g. `first_value` and `first_time`)
- [x] SQL based tests

# TODO
- [x] Merge https://github.com/apache/arrow-datafusion/pull/3428
- [x] Update datafusion pin to include https://github.com/apache/arrow-datafusion/pull/3428
- [x] Finish testing for other types
- [x] Port over other selectors to new format
- [x] Remove old selector implementations
- [x] Remove code duplication
- [x] File upstream DataFusion ticket for better API to register user defined functions: https://github.com/apache/arrow-datafusion/issues/3465


query_functions/src/selectors.rs
```diff
@@ -147,15 +251,129 @@ pub fn selector_max(data_type: &DataType, output: SelectorOutput) -> AggregateUD
...
-        DataType::Utf8 => make_uda::<Utf8MaxSelector>(name, output),
-        DataType::Boolean => make_uda::<BooleanMaxSelector>(name, output),
-        _ => unimplemented!("max not supported for {:?}", data_type),
+    make_uda(
+        name,
+        FactoryBuilder::new(SelectorType::Max, output).with_value_type(data_type.clone()),
+    )
+}
+
+#[derive(Debug, Clone, Copy)]
```
alamb, line 263: This was largely a plumbing exercise to move instantiation of the selectors into code that didn't have too much duplication

query_functions/src/selectors.rs
```diff
@@ -35,6 +41,113 @@ use internal::{
...
 use schema::TIME_DATA_TYPE;
 
+/// registers selector functions so they can be invoked via SQL
+pub fn register_selector_aggregates(mut state: SessionState) -> SessionState {
+    let first = struct_selector_first();
+    let last = struct_selector_last();
+    let min = struct_selector_min();
+    let max = struct_selector_max();
+
+    //TODO make a nicer api for this in DataFusion
```
alamb, line 51:
```suggestion
    // TODO make a nicer api for this in DataFusion
    // https://github.com/apache/arrow-datafusion/issues/3465
```

query_tests/cases/in/selectors.sql
```diff
@@ -0,0 +1,66 @@
...
+select selector_first(float_field, time) from m;
+select selector_first(int_field, time) from m;
+select selector_first(uint_field, time) from m;
+select selector_first(string_field, time) from m;
+select selector_first(bool_field, time) from m;
+
+-- Test extracting value and time subfields
+select selector_first(float_field, time)['value'], selector_first(float_field, time)['time'] from m;
+-- Also with subqueries
+select f['value'], f['time'] from (select selector_first(float_field, time) as f from m) as sq;
```
alamb, line 21: here are the SQL tests showing how to use these selector functions and their struct output

query_functions/src/selectors.rs
```diff
@@ -54,16 +167,13 @@ pub fn selector_first(data_type: &DataType, output: SelectorOutput) -> Aggregate
...
-        DataType::Int64 => make_uda::<I64FirstSelector>(name, output),
-        DataType::UInt64 => make_uda::<U64FirstSelector>(name, output),
-        DataType::Utf8 => make_uda::<Utf8FirstSelector>(name, output),
-        DataType::Boolean => make_uda::<BooleanFirstSelector>(name, output),
-        _ => unimplemented!("first not supported for {:?}", data_type),
-    }
+    make_uda(
+        name,
+        FactoryBuilder::new(SelectorType::First, output).with_value_type(data_type.clone()),
+    )
```
NGA-TRAN, line 176: 🎉

query_functions/src/selectors.rs
```diff
@@ -147,15 +251,129 @@ pub fn selector_max(data_type: &DataType, output: SelectorOutput) -> AggregateUD
...
+
+/// Builder to create the appropriate typed factory functions for selectors
+#[derive(Debug)]
+struct FactoryBuilder {
+    selector_type: SelectorType,
+
+    output_type: SelectorOutput,
+
+    // If the selector output is "time" we can't determine the
+    // accumuator type from the return type, so hold we pass the data type explicitly
```
NGA-TRAN, line 279: question: can the return type be the same`time` or it must be different?

query_tests/cases/in/selectors.expected
```diff
@@ -0,0 +1,182 @@
...
++--------------------------------------------------------+
+| {"value": true, "time": 1970-01-01 00:00:00.000000600} |
++--------------------------------------------------------+
+-- SQL: select selector_last(float_field, time)['value'], selector_last(float_field, time)['time'] from m;
++--------------------------------------------+-------------------------------------------+
+| selector_last(m.float_field,m.time)[value] | selector_last(m.float_field,m.time)[time] |
++--------------------------------------------+-------------------------------------------+
+| 66                                         | 1970-01-01T00:00:00.000000600Z            |
++--------------------------------------------+-------------------------------------------+
+-- SQL: select f['value'], f['time'] from (select selector_last(float_field, time) as f from m) as sq;
```
NGA-TRAN, line 93: This query is very interesting

NGA-TRAN (approved):
Very nice. Now I know how to use the selector function. I went over all the functions and tests carefully.

alamb:
Thank you for the careful review @NGA-TRAN 👍

query_tests/cases/in/selectors.expected
alamb: I was (very) pleasantly surprised to see that the `['field']`syntax worked and that enough of DataFusion supported these struct types to use.

query_functions/src/selectors.rs
alamb: The return type can be the same as time (aka `DataType::Timestamp(TimeUnit::Nanoseconds, None)`) . However the accumulator state needs to hold both the `value` and the `time`

The factory function that figures out the accumulator's state type is passed the return type. When the output contains the actual column value (e.g. the `i64`/ `u64`, etc) then the internal state can be recovered from the return type. However, when the output type is just `time` then there needs to be some other way to know the column value's type

I actually tried to change the selectors to *always* return a struct (that is what https://github.com/influxdata/influxdb_iox/pull/5606 does) but I ran into trouble making the influxrpc interface use them.

PR #5632, opened 2022-09-14
feat: Improved InfluxQL error messages

### LINKED ISSUE #5537: InfluxQL: Add custom error type to InfluxQL parser
This issue is to add custom error type support to the `influxdb_influxql_parser` crate to improve the public error messages. It is dependent on #5416, in order to provide an initial set of APIs to explore the error reporting space.

stuartcarnie:
Closes #5537 

This PR introduces several improvements to the `influxdb_influxql_parser` to establish consistent error handling and user-facing messages.

##### Add `internal` module 

* Add a specialised `Error` type for handling nom errors and an alternative for capturing InfluxQL syntax error messages
* Add `map_fail` to function to map an `Err` from the provided closure to a `nom::Err::Failure` with a custom error message. A `nom::Err::Failure` terminates parsing immediately.
* Add `expect` function, which behaves similarly to nom's `cut` function, to force parsing to terminate. `expect` adds the custom syntax error message for context.
* Defines a specialised `nom::IResult`, `ParseResult` to be used by combinator functions throughout the crate. This type specifies the default error as the specialised `Error` struct.

##### Refactor to use `internal::ParseResult`

All existing combinator functions across the remaining modules were updated to use the `ParseResult` type.

##### Improve error messages

Parser combinator functions were updated to use the `expect` and `map_fail` functions, where appropriate. The goal of this refactoring was to ensure that once we reached unrecoverable points in parsing, we fail with an appropriate error message. There is no intention to match the existing Go implementation error messages, as these are not treated as values by clients.

##### Refactor `show_measurements` combinator

There are many InfluxQL statements that start with `SHOW`, so we have broken this into separate combinators. The first is the `show_statement`, which is responsible for parsing a `SHOW` token and then any one of the other supported `SHOW` statements. Currently there is only `show_measurements`, but future changes will add additional combinators. This change allows us to fail if we do not recognise the next token after the `SHOW` statement, which is unrecoverable.

##### Add `parse_statements` public function

Add the public APIs to the crate to support parsing InfluxQL statements separated by the `;` statement terminator. We also added the public error type to be consumed by clients, which is intentionally only useful for `Display`. This public API serves as the foundation for developing the remainder of the parsers.


influxdb_influxql_parser/src/common.rs
```diff
@@ -84,30 +84,39 @@ pub fn measurement_name_expression(i: &str) -> IResult<&str, MeasurementNameExpr
...
 }
 
 /// Parse a LIMIT <n> clause.
-pub fn limit_clause(i: &str) -> IResult<&str, u64> {
-    preceded(pair(tag_no_case("LIMIT"), multispace1), unsigned_number)(i)
+pub fn limit_clause(i: &str) -> ParseResult<&str, u64> {
+    preceded(
+        pair(tag_no_case("LIMIT"), multispace1),
+        expect(
+            "invalid LIMIT clause, expected unsigned integer",
```
stuartcarnie, line 96: `expect` always fails if the inner parser fails, returning an error with the specified message.

influxdb_influxql_parser/src/parameter.rs
```diff
@@ -99,8 +99,7 @@ mod test {
         // └─────────────────────────────┘
 
         // missing `$` prefix
-        let res = parameter("cpu");
-        assert!(res.is_err());
+        parameter("cpu").unwrap_err();
```
stuartcarnie, line 102: Drive-by cleanup 🧹

influxdb_influxql_parser/src/show.rs
```diff
@@ -0,0 +1,41 @@
...
+use nom::character::complete::multispace1;
+use nom::combinator::map;
+use nom::sequence::{pair, preceded};
+
+/// Parse a SHOW statement.
+pub fn show_statement(i: &str) -> ParseResult<&str, Statement> {
+    preceded(
+        pair(tag_no_case("SHOW"), multispace1),
+        expect(
+            "invalid SHOW statement, expected MEASUREMENTS",
```
stuartcarnie, line 14: We will update this message as we add more parsers

influxdb_influxql_parser/src/statement.rs
```diff
@@ -0,0 +1,27 @@
...
+        };
+
+        Ok(())
+    }
+}
+
+/// Parse a single InfluxQL statement.
+pub fn statement(i: &str) -> ParseResult<&str, Statement> {
+    // NOTE: This will become an alt(()) once more statements are added
+    show_statement(i)
```
stuartcarnie, line 26: We only have one statement, so we can't use the `alt` combinator as it doesn't compile. Once we add more statements, we'll switch to that combinator.

crepererum: Yeah, seems like an oversight in `nom`, `Alt` is not implemented for 0 and 1-tuples, see https://docs.rs/nom/latest/nom/branch/trait.Alt.html

crepererum (approved):
I really like the error handling and how it integrates with the new test macro :+1:

influxdb_influxql_parser/src/common.rs
```diff
@@ -280,7 +315,7 @@ mod tests {
         // Fallible cases
 
         // Must be "time" identifier
-        assert_failure!(order_by_clause("ORDER by foo"));
+        assert_expect_error!(order_by_clause("ORDER by foo"), "invalid ORDER BY");
```
NGA-TRAN, line 318: Question: Is it possible for us to even have a clear message here saying what should be following the `ORDER BY`?

influxdb_influxql_parser/src/lib.rs
```diff
@@ -11,14 +11,137 @@
...
+
+        // Parse multiple statements with a terminator in quotes, ensuring it is not interpreted as
+        // a terminator
+        let got = parse_statements(
+            "SHOW MEASUREMENTS WITH MEASUREMENT = \";\";SHOW MEASUREMENTS LIMIT 1",
+        )
+        .unwrap();
+        assert_eq!(
+            format!("{}", got[0]),
+            "SHOW MEASUREMENTS WITH MEASUREMENT = \";\""
```
NGA-TRAN, line 135: Question: is this a valid query that does not need to specify the name of a measurement? 
Can you also add a test `SHOW MEASUREMENTS WITH MEASUREMENT = a_whatever_name`?

influxdb_influxql_parser/src/common.rs
```diff
@@ -246,10 +269,22 @@ mod tests {
...
+            offset_clause("OFFSET "),
+            "invalid OFFSET clause, expected unsigned integer"
+        );
 
         // overflow
-        offset_clause("OFFSET 34593745733489743985734857394").unwrap_err();
+        assert_expect_error!(
+            offset_clause("OFFSET 34593745733489743985734857394"),
+            "unable to parse unsigned integer"
+        );
```
NGA-TRAN, line 287: 🎉  Nice capturing invalid syntax. I wonder whether we have tests for all supported syntax

influxdb_influxql_parser/src/common.rs
stuartcarnie: Done!

influxdb_influxql_parser/src/lib.rs
stuartcarnie: The name of the measurement is enclosed in double quotes, such that it is a single semicolon, `;`, and therefore valid syntax. I don't expect that IOx would allow a user to create a table as a single semicolon, so this query would fail to execute. 

Incidentally, these tests are intended to validate parsing multiple statements, and not the `SHOW MEASUREMENTS` statement. I have specific `SHOW MEASUREMENTS` tests in the `show_measurements.rs` module, including one for your suggestion, which includes various whitespace elements:

https://github.com/influxdata/influxdb_iox/blob/1b2ba222673b3131183001c501f5275ee554cabb/influxdb_influxql_parser/src/show_measurements.rs#L218

influxdb_influxql_parser/src/common.rs
stuartcarnie: I plan on adding a suite of SQL statements in a file that can be validated. I expect we will also add fuzzing tests.

stuartcarnie:
@NGA-TRAN and @crepererum thank you again for the reviews 🙇🏻  I've improved the `ORDER BY` error message, as suggested

PR #5634, opened 2022-09-14
refactor(ingester): split data.rs into submodules

domodwyer:
All the key data structures for the ingester lived in one giant `data` module / file - this PR breaks them into submodules/files. **There are no functional changes in this PR.**

The ingester buffer structures are effectively a tree - `Shard` has many `Namespace` has many `Tables` has many `Partitions` which has one `DataBuffer`. Unfortunately with everything living in one mod, there's no API boundaries / separation of concerns enforced and various levels of this hierarchical structure peek into their child nodes, directly manipulating their field state - this makes the code hard to follow, because all the code can modify one field from anywhere (and when multiplied by the number of fields and this quickly gets impossible to track).

As a result of this (and other reasons) many functions duplicate behaviours of other functions - some are tested but never used, and some are used but never tested. Some code calls `buffer.get_persisting_data()` but other code just reads `buffer.persisting` and manipulates that directly. It's a mess! I intend to tidy it up a bit before adding more logic to these structures, mostly because I don't understand the code paths as-is.

---

* refactor(ingester): split data.rs into modules (074722eb3)

      Breaks the gigantic data.rs file into sub-modules for Shard, Namespace, Table,
      Partition, and finally the actual data buffer used to store writes.

* style(ingester): fmt imports & long strings (ee8cdb48a)

      Rewrite the imports to be a consistent order; std, external, crate and merge
      all crate-level imports into one use statement.


ingester/src/data/partition/buffer.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use mutable_batch_lp::test_helpers::lp_to_mutable_batch;
+
+    use super::*;
+
+    #[test]
+    fn snapshot_empty_buffer_adds_no_snapshots() {
```
alamb, line 286: 👍

mkmik (approved):

alamb (approved):
I didn't review all changes in this PR carefully, but I trust that the end to end tests ensure we didn't break anything in connection and I spot checked that the existing tests in the ingester are still present (Today I learned there are only three 🤯 )

:thumbsup: nice work @domodwyer

NGA-TRAN (commented):
Nice splits 👍

PR #5639, opened 2022-09-14
refactor: Move fetching of table columns, extra partition info into the method

carols10cents:
This is another step on a refactoring journey.

Advantages of this refactor:

- Less repeated code outside of `Compactor`; hot and cold compaction had to call `compactor.table_columns` and `compactor.add_info_to_partitions` to get usable information after calling `compactor.hot_partitions_to_compact` and `compactor.cold_partitions_to_compact`. Now they only have to call that one method rather than 3 and they get everything they need.
- Tests that need this extra info will also get it without needing the extra setup steps of calling `table_columns` and `add_info_to_partitions`

Disadvantages:

- These functions now return a tuple of (candidates, table_columns) and those table columns aren't always needed, especially in tests-- that will be improved with my next PR :)
- The time to fetch this extra info recorded in `partitions_extra_info_reading_duration` is now also included in the `candidate_selection_duration` time, as reading the extra info is now part of the candidate selection function. Is this a problem? If so, I can move the `candidate_selection_duration` recording within the `compactor.hot_partitions_to_compact` and `compactor.cold_partitions_to_compact` methods and record it before those methods call `table_columns` or `add_info_to_partitions`.

### DIFF
--- a/compactor/src/compact.rs
+++ b/compactor/src/compact.rs
@@ -18,7 +18,7 @@ use parquet_file::storage::ParquetStorage;
 use schema::sort::SortKey;
 use snafu::{OptionExt, ResultExt, Snafu};
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet},
     sync::Arc,
     time::Duration,
 };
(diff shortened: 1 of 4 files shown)


compactor/src/compact.rs
```diff
@@ -326,13 +330,41 @@ impl Compactor {
...
+        );
+        let candidates = self.add_info_to_partitions(&candidates).await?;
+
+        if let Some(delta) = self.time_provider.now().checked_duration_since(start_time) {
+            let attributes = Attributes::from(&[("partition_type", compaction_type)]);
+            let duration = self
+                .partitions_extra_info_reading_duration
+                .recorder(attributes);
+            duration.record(delta);
+        }
```
NGA-TRAN, line 365: 👍 for moving this here

compactor/src/compact.rs
```diff
@@ -776,36 +840,24 @@ mod tests {
...
+        assert_eq!(*candidates[0].table, table);
+        assert_eq!(candidates[0].partition_key, partition3.partition_key);
+        assert_eq!(candidates[0].sort_key, partition3.sort_key()); // this sort key is None
+
+        assert_eq!(candidates[1].id(), another_partition.id);
+        assert_eq!(candidates[1].shard_id(), another_shard.id);
+        assert_eq!(*candidates[1].namespace, namespace);
+        assert_eq!(*candidates[1].table, another_table);
+        assert_eq!(candidates[1].partition_key, another_partition.partition_key);
+        assert_eq!(candidates[1].sort_key, another_partition.sort_key()); // this sort key is Some(tag1, time)
```
NGA-TRAN, line 860: 🎉

NGA-TRAN (approved):
Nice

PR #5642, opened 2022-09-15
chore: Remove variants from Identifier and BindParameter types

### LINKED ISSUE #5538: InfluxQL: Simplify Identifier and BindParameter AST elements
Refactor `Identifer` and `BindParameter` to a struct of `String` rather than `Quoted` and `Unquoted` sum type. This is intended to simplify their usage in the AST. It should be noted that the implementation of `Display` should correctly quote the identifier if it contains characters that requiring escaping.

## Problem

Currently, both the `Identifier`:

https://github.com/influxdata/influxdb_iox/blob/24978d5df27a7d98f7e802cfabb865a4b0c29105/influxdb_influxql_parser/src/identifier.rs#L46-L52

and `BindParameter` types:

https://github.com/influxdata/influxdb_iox/blob/d219f93241838387e3803a6a0b6d1bd721882a4d/influxdb_influxql_parser/src/parameter.rs#L35-L41

use a sum type to distinguish whether the previously parsed syntactical element was quoted or unquoted. This was originally intended so that when parsed, they would round-trip back to their original form when displayed. This would be useful if the AST retained whitespace, so as to ensure fidelity with the parsed source, however, as this is not the case, the `Quoted` vs `Unquoted` variants only result in more complex usage without any benefit.

stuartcarnie:
Closes #5538 

This PR removes the `Quoted` and `Unquoted` variants from the `BindParameter` and `Identifier` types. This simplifies their usage for clients. The `Display` traits for both types has been updated to properly quote and escape the output, when necessary.


stuartcarnie:
@crepererum here is the improvement we discussed to remove the `Quoted` and `Unquoted` variants from the `Identifier` and `BindParameter` types. It is a nice improvement 👍🏻

influxdb_influxql_parser/src/identifier.rs
```diff
@@ -109,24 +102,21 @@ mod test {
     fn test_identifier() {
         // quoted
         let (_, got) = identifier("\"quick draw\"").unwrap();
-        assert!(matches!(got, Identifier::Quoted(s) if s == "quick draw"));
+        assert_eq!(got, "quick draw".into());
```
crepererum, line 105: The code is gone now, but my 2cents regarding `assert!(matches!(...));`:

- in other places of our stack we uses [`assert_matches!`](https://crates.io/crates/assert_matches) (not stdlib!) which provides nice panic messages if the match fails (i.e. it prints out the data, not just a message saying the condition was false)
- In general I would use `assert_eq!` and `#[derive(PartialEq, Eq)]` on the struct/enum in question, even when this equality functionality is only needed for tests (there are only a very few cases where I would derive functionality only for tests, e.g. [when it requires a special test crate](https://github.com/influxdata/rskafka/blob/3208e4742f08048bbab4e8fc4e0a775507fe3e66/src/protocol/primitives.rs#L22)).

crepererum (approved):
I like it :)

influxdb_influxql_parser/src/identifier.rs
stuartcarnie: 👏🏻  thanks for the advice! I'll use that as I'm implementing the remainder of the InfluxQL statements.

Issue #5633, opened 2022-09-14
InfluxQL: Implement statement parsers


stuartcarnie:
Teach the InfluxQL parser how to parse the following statements, which are supported in Cloud 2

### Schema exploration 

See #5643 

- [x] `SHOW DATABASES`
- [x] `SHOW MEASUREMENTS`
- [x] `SHOW RETENTION POLICIES`
- [x] `SHOW TAG KEYS`
- [x] `SHOW TAG VALUES`
- [x] `SHOW FIELD KEYS`

### Data query

See #5692

- [x] `SELECT`

### DML Statements

See #5656

- [x] `DELETE` to delete series
- [x] `DROP MEASUREMENT`

### Debugging statements

See #5763

InfluxQL supports `EXPLAIN` statements. The output is not documented, so we can leverage the existing DataFusion `EXPLAIN` output:

- [x] `EXPLAIN`
- [x] `EXPLAIN ANALYZE`

This is not supported by InfluxQL but is worth adding:

- [x] `EXPLAIN VERBOSE`
- [x] `EXPLAIN ANALYZE VERBOSE`

NGA-TRAN:
All nice syntax.
Question: Is this ticket is to `parse the statement` only or to `parse and run them to return to results`?

stuartcarnie:
@NGA-TRAN this is only to develop the parser and parse the statements. Future issues will be created to define how we'll transform these ASTs into a logical plan that can be executed by DataFusion.

PR #5643, opened 2022-09-15
feat: Teach InfluxQL parser how to parse SHOW statements

stuartcarnie:
Helps #5633

This PR teaches the InfluxQL parser crate how to parse the following `SHOW` statements:

- [x] `SHOW DATABASES`
- [x] `SHOW MEASUREMENTS`
- [x] `SHOW RETENTION POLICIES`
- [x] `SHOW TAG KEYS`
- [x] `SHOW TAG VALUES`
- [x] `SHOW FIELD KEYS`

In addition to these new statements, `SHOW MEASUREMENTS` parsing was improved to ensure whitespace can be omitted under certain conditions, such as between operators in binary expressions. Additional tests were added to proved these improvements.


influxdb_influxql_parser/src/expression.rs
```diff
@@ -460,6 +460,20 @@ mod test {
...
 
+        // Expressions are still valid when unnecessary whitespace is omitted
+
+        let (_, got) = conditional_expression("5+51").unwrap();
+        assert_eq!(got, *binary_op!(5, Add, 51));
+
+        let (_, got) = conditional_expression("5+$foo").unwrap();
+        assert_eq!(got, *binary_op!(5, Add, param!("foo")));
+
+        let (_, got) = conditional_expression("5--(3|2)").unwrap();
```
NGA-TRAN, line 471: This is interesting. Do you really want to support `5--3`? I think we may want to stay with supporting `5 - -3` only unless we accidentally supported it before

influxdb_influxql_parser/src/show.rs
```diff
@@ -1,21 +1,154 @@
...
+///
+/// A `FROM` clause for a `SHOW` statements differs from a `FROM` in a
+/// `SELECT`, as it can only contain measurement name or regular expressions.
+///
+/// It is defined by the following EBNF notation:
+///
+/// ```text
+/// from_clause ::= "FROM" measurement_selection ("," measurement_selection)*
+/// measurement_selection ::= three_part_measurement_name | regex
+/// ```
```
NGA-TRAN, line 80: If you can add an example what a SHOW stmt with FROM clause looks like here, it will be helpful for reviewers and future coders

influxdb_influxql_parser/src/show_field_keys.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn test_show_field_keys() {
+        // No optional clauses
+        let (_, got) = show_field_keys("FIELD KEYS").unwrap();
+        assert_eq!(format!("{}", got), "SHOW FIELD KEYS");
+
+        let (_, got) = show_field_keys("FIELD KEYS ON db").unwrap();
+        assert_eq!(format!("{}", got), "SHOW FIELD KEYS ON db");
+
+        // measurement selection using name
+        let (_, got) = show_field_keys("FIELD KEYS FROM db..foo").unwrap();
```
NGA-TRAN, line 106: Question: is `db..foo` a valid name or it means something?

NGA-TRAN (approved):
LGTM

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Thanks @NGA-TRAN – glad you spotted that, as I re-rested what is significant whitespace by the existing Go implementation, and turns out unary operators require preceding whitespace, such that the following is invalid

```
5--(3|2)
```

and the following is valid:

```
5- -(3|2)
```

✅ I will make the correction and update the unit tests accordingly

influxdb_influxql_parser/src/show_field_keys.rs
stuartcarnie: Yes, in a number of places, InfluxQL accepts a 3-part name for a measurement:

https://github.com/influxdata/influxdb_iox/blob/a5418de43359f076478af5aaf3203a27316cc0f9/influxdb_influxql_parser/src/common.rs#L14-L20

A fully qualified name is therefore:

```
database.retention_policy.measurement_name
```

In this test case, `db..foo` refers to the measurement `foo` stored in the database, `db` using the default retention policy[^1]. The default retention policy can be changed by the user, which is normally named `autogen`. The equivalent of `db..foo` would therefore be `db.autogen.foo`, if the default retention policy was `autogen`.


[^1]: https://docs.influxdata.com/influxdb/v1.8/concepts/glossary/#retention-policy-rp

influxdb_influxql_parser/src/expression.rs
stuartcarnie: Further investigation revealed the issue. It turns out `--` is a line comment, and whitespace is not significant in the following example:

```
5-+(3|2)
```

I have created #5663 to address comment support in a future PR.

PR #5645, opened 2022-09-15
feat: split "pruned" metric into "early" and "late"

crepererum:
Took the liberty to clean up the code a bit:

- splits the metrics from `query_access`
- create a concept of a group (e.g. "pruned early", "not pruned", ...) to simplify metric creation and access


querier/src/table/query_access/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /// Called when pruning a chunk before fully creating the chunk structure
+    pub(crate) fn was_pruned_early(&self, row_count: u64, size_estimate: u64) {
+        self.metrics.pruned_early.inc(1, row_count, size_estimate);
+    }
+}
+
+impl PruningObserver for MetricPruningObserver {
+    fn was_pruned(&self, chunk: &dyn QueryChunk) {
```
alamb, line 164: Maybe this could be renamed `was_pruned_late` so it matches the metric. It would also help to add a docstring describing the difference between early / late again

querier/src/table/query_access/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            chunk_rows(chunk) as u64,
+            chunk_estimate_size(chunk) as u64,
+        );
+    }
+}
+
+fn chunk_estimate_size(chunk: &dyn QueryChunk) -> usize {
+    let chunk = chunk.as_any();
+
+    if let Some(chunk) = chunk.downcast_ref::<IngesterChunk>() {
```
alamb, line 200: Maybe worth (in some follow on PR perhaps) to add `estimate_size` and `chunk_rows` to the `QueryChunk` trait

querier/src/table/query_access/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    metrics: Arc<PruneMetrics>,
+}
+
+impl MetricPruningObserver {
+    pub(crate) fn new(metrics: Arc<PruneMetrics>) -> Self {
+        Self { metrics }
+    }
+
+    /// Called when pruning a chunk before fully creating the chunk structure
+    pub(crate) fn was_pruned_early(&self, row_count: u64, size_estimate: u64) {
```
alamb, line 158: I probably missed it, but where is `was_pruned_early` called?

querier/src/namespace/query_access.rs
```diff
@@ -348,23 +348,47 @@ mod tests {
...
+                .unwrap(),
+            &Observation::U64Counter(0),
+        );
+        assert_eq!(
+            reporter
+                .metric("query_pruner_rows")
+                .unwrap()
+                .observation(&[("result", "pruned_late")])
+                .unwrap(),
+            &Observation::U64Counter(0),
```
alamb, line 385: I wonder if it would be possible to have a test with an observation other than zero (to make sure the metric plumbing is hooked up correctly)

alamb (approved):
Looks good to me -- I had some style suggestions and one question about where `was_pruned_early`

Also a test showing that the metrics are all hooked up correctly might be valuable

querier/src/table/query_access/mod.rs
crepererum: The `QueryChunk` trait is also used by the Ingester and Compactor, and I think they might not have this data available easily (or even care about it).

querier/src/table/query_access/mod.rs
crepererum: `querier::table::QuerierTable::chunks_inner`, i.e. pruning based solely on very cheap data. This was implemented by @lesam some weeks ago.

alamb: Ah, I see -- https://github.com/influxdata/influxdb_iox/blob/f7b6f81fe12eb61797a6dc51900512061bc6a2e9/querier/src/table/mod.rs#L306

querier/src/table/query_access/mod.rs
alamb: A docstring would be helpful -- thank you

querier/src/namespace/query_access.rs
crepererum: I've tried, but time comparisons w/ SQL produce casts that DF cannot prune at the moment and I'm too lazy to write yet another testing route that construct the request manually just to get some monitoring tested.

alamb: I think there have been some significant improvements in pruning on predicates with casts in the last few days in DataFusion -- specifically https://github.com/apache/arrow-datafusion/pull/3422

PR #5646, opened 2022-09-15
feat: concurrent chunk creation

crepererum:
Create chunks in querier concurrently after we've pre-filtered them. Chunk creation still may require a bit of cached information (e.g. the partition sort key) and we can easily fetch these concurrently instead of in order.


querier/src/table/mod.rs
```diff
@@ -294,26 +299,32 @@ impl QuerierTable {
...
                             .new_chunk(
                                 cached_table,
                                 Arc::clone(self.table_name()),
-                                Arc::clone(cached_parquet_file),
+                                cached_parquet_file,
                                 span,
                             )
                             .await
                     })
+                    .buffer_unordered(CONCURRENT_CHUNK_CREATION_JOBS)
```
alamb, line 326: Love it

alamb (approved):

PR #5647, opened 2022-09-15
refactor: concurrent table planning in InfluxRPC

crepererum:
Some InfluxRPC can scan multiple tables. Prior to this PR we were always scanning the tables in sequence, adding up potential latencies (catalog, ingester, object store). There is no reason we need to do this, "ordinary" SQL queries would not serialize this way either.

So let's scan tables concurrently. This add concurrency to:

- read filter
- read group
- read window aggregate

There are other query types that could benefit from a similar treatment. They will be changed in a follow-up.


iox_query/src/frontend/influxrpc.rs
```diff
@@ -1354,6 +1303,82 @@ impl InfluxRpcPlanner {
     }
 }
 
+/// Stream of chunks for table predicates.
```
crepererum, line 1306: I've decided to pull out helpers so I don't need to repeat the code. `table_chunk_stream` is currently only used once (in `create_series_set_plans`) but will be used by the other query types that use a different execution "mode" (mostly the ones that iterate chunks manually).

iox_query/src/frontend/influxrpc.rs
```diff
@@ -36,6 +37,8 @@ use std::{
     sync::Arc,
 };
 
+const CONCURRENT_TABLE_JOBS: usize = 4;
```
alamb, line 40: Any reason this isn't 10 (to match the concurrent chunk jobs)? I realize it is a somewhat arbitrary number, I was just curious about its value

iox_query/src/frontend/influxrpc.rs
```diff
@@ -791,55 +779,37 @@ impl InfluxRpcPlanner {
...
-                continue;
-            }
-
-            let schema = database
-                .table_schema(table_name)
-                .context(TableRemovedSnafu { table_name })?;
-
-            let ss_plan = match agg {
-                Aggregate::None => self.read_filter_plan(
+        let plan = create_series_set_plans(
```
alamb, line 790: this is a nice refactor

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1354,6 +1303,82 @@ impl InfluxRpcPlanner {
...
+                    .await
+                    .context(GettingChunksSnafu { table_name })?;
+
+                Ok((table_name.as_str(), predicate, chunks))
+            }
+        })
+        .buffered(CONCURRENT_TABLE_JOBS)
+}
+
+/// Create series set plans.
```
alamb, line 1329:
```suggestion
/// Create series set plans that fetch the data specified in rpc_predicate.
///
/// `f(ctx, table_name, table_predicate, chunks, table_schema)` is
///  invoked on the chunks for each table to produce the final `SeriesSetPlan`
```

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1354,6 +1303,82 @@ impl InfluxRpcPlanner {
...
+            let ctx = &ctx;
+            let database = Arc::clone(&database);
+            let f = f.clone();
+
+            async move {
+                let schema = database
+                    .table_schema(table_name)
+                    .context(TableRemovedSnafu { table_name })?;
+
+                f(ctx, table_name, predicate, chunks, schema)
```
alamb, line 1373: It is very cool that you got this to work

predicate/src/rpc_predicate.rs
```diff
@@ -398,4 +398,8 @@ mod tests {
 
         Arc::new(schema)
     }
+
+    #[allow(dead_code)]
+    const fn assert_send<T: Send>() {}
+    const _: () = assert_send::<InfluxRpcPredicate>();
```
alamb, line 404: Perhaps it is worth a comment here about *why* this is here and why `InfluxRpcPredicate` needs to be `Send`

alamb (approved):
LGTM -- I also feel quite good about the coverage in `query_tests` so I don't worry about this kind of refactor messing up the plumbing.

Thanks @crepererum -- this is pretty sweet and a nice example of ["Fearless Concurrency"](https://doc.rust-lang.org/book/ch16-00-concurrency.html) in action.

iox_query/src/frontend/influxrpc.rs
crepererum: No, this was rather a random choice. Changed it to 10.

predicate/src/rpc_predicate.rs
crepererum: done

PR #5650, opened 2022-09-15
refactor: Clear up responsibilities of different parts of the compactor

carols10cents:
This is some more refactoring that I think clears up and separates some responsibilities:

- Estimating bytes is now done by the `PartitionCompactionCandidateWithInfo` because it keeps track of the column type counts.
- When `CompactorParquetFile`s are created, they give their number of rows to the partition and ask for their estimated bytes, then that gets carried with them.
- This makes it easier and clearer to test filtering based on the memory budget.
- The `FilteredFiles`' `FilterResult` now combines its data and its states, so that, for example, `NothingToCompact` doesn't need any data.
- There's a new type `ReadyToCompact` that gets created from a `FilterResult::Proceed` because the compaction after that point doesn't need to know about the memory budget at all.

I'm hoping to eventually get a clear pipeline of states like `Partition` -> `PartitionCompactionCandidateWithInfo` -> `FilteredFiles` -> `ReadyToCompact` that's clearer to follow, the code isn't quite there yet but this gets us closer.

Next, I think I'm going to be able to generalize some functions to work for L0+L1 -> L1 compaction as well as L1+L2 -> L2 compaction.

### DIFF
--- a/compactor/src/cold.rs
+++ b/compactor/src/cold.rs
@@ -186,7 +186,9 @@ fn group_by_size(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{compact_one_partition, handler::CompactorConfig, parquet_file_filtering};
+    use crate::{
+        compact_one_partition, handler::CompactorConfig, parquet_file_filtering, ReadyToCompact,
+    };
     use ::parquet_file::storage::ParquetStorage;
     use arrow_util::assert_batches_sorted_eq;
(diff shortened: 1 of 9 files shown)


NGA-TRAN (approved):
LGTM. I like the idea  `Partition -> PartitionCompactionCandidateWithInfo -> FilteredFiles -> ReadyToCompact`

PR #5651, opened 2022-09-15
refactor: concurrent table scan in "field columns"

crepererum:
Similar to #5647 and #5649.


crepererum:
Oh now, this time I broke it :D Will have a look later.

crepererum:
Fixed.

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1313,33 +1327,30 @@ fn table_chunk_stream<'a>(
         .buffered(CONCURRENT_TABLE_JOBS)
 }
 
-/// Create series set plans that fetch the data specified in rpc_predicate.
+/// Create plans that fetch the data specified in rpc_predicate.
```
alamb, line 1330:
```suggestion
/// Create plans that fetch the data specified in table_predicates.
///
/// table_predicates contains `(table_name, predicate_specialized_for_that_table)`
///
/// The plans are created in parallel as different `async` Tasks to reduce 
/// query latency due to planning
```

alamb (approved):
Looks very nice to me. I can't wait to see how query performance improves after this work

PR #5652, opened 2022-09-15
refactor(db): NULLable persisted_sequence_number

domodwyer:
0 is a valid sequence number (the first write) and it's therefore a valid persist marker: a parquet file could contain only the first write for a shard.

---

* refactor(db): NULLable persisted_sequence_number (8eebd1fb0)

      Makes the partition.persisted_sequence_number column in the catalog DB 
      NULLable. 0 is a valid persisted sequence number.

### DIFF
--- a/data_types/src/lib.rs
+++ b/data_types/src/lib.rs
@@ -853,7 +853,9 @@ pub struct Partition {
     /// It is a system invariant that this value monotonically increases over
     /// time - wrote another way, it is an invariant that partitions are
     /// persisted (or at least made visible) in sequence order.
-    pub persisted_sequence_number: SequenceNumber,
+    ///
+    /// If [`None`] no data has been persisted for this partition.
+    pub persisted_sequence_number: Option<SequenceNumber>,
 }
 
 impl Partition {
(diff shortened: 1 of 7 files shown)


iox_catalog/migrations/20220915155941_nullable-partition-persist-offset.sql
```diff
@@ -0,0 +1,11 @@
...
+--
+-- NULL == no persisted data for this partition
+ALTER TABLE
+    "partition" DROP COLUMN "persisted_sequence_number";
+
+-- Remove implicit NULL / default of 0.
+ALTER TABLE
+    "partition"
+ADD
+    COLUMN "persisted_sequence_number" BIGINT NULL DEFAULT NULL;
```
domodwyer, line 11: I don't of a way to change a column to NULLable without dropping the rows that are not NULL....

Either way, this have no consequence for new clusters, and the existing clusters are not using the values yet / no problem dropping them.

jacobmarble (approved):

NGA-TRAN (approved):
LGTM

data_types/src/lib.rs
```diff
@@ -853,7 +853,9 @@ pub struct Partition {
     /// It is a system invariant that this value monotonically increases over
     /// time - wrote another way, it is an invariant that partitions are
     /// persisted (or at least made visible) in sequence order.
-    pub persisted_sequence_number: SequenceNumber,
+    ///
+    /// If [`None`] no data has been persisted for this partition.
```
alamb, line 857: 👍 

Can this feld also be `None` also happen if the partition was persisted prior to the code change that added it?

alamb (commented):
👍

PR #5657, opened 2022-09-16
perf: O(1) partition persist mark discovery

domodwyer:
This is the read-side change to use the per-partition persist markers added in https://github.com/influxdata/influxdb_iox/pull/5648 (described/tracked in https://github.com/influxdata/influxdb_iox/issues/5638).

This should significantly help https://github.com/influxdata/conductor/issues/971 for clusters that have been running more than a couple of weeks.

---

* perf: O(1) partition persist mark discovery (2b8985071)

      Changes the ingest code path to eliminate scanning the parquet_files table to
      discover the last persisted offset per partition, instead utilising the new
      persisted_sequence_number field on the Partition itself to read the same
      value.

      This lookup blocks ingest for the shard, so removing the expensive query from
      the ingest hot path should improve catch-up time after a restart/deployment.

### DIFF
--- a/ingester/src/data/table.rs
+++ b/ingester/src/data/table.rs
@@ -161,26 +161,17 @@ impl TableData {
         shard_id: ShardId,
         catalog: &dyn Catalog,
     ) -> Result<(), super::Error> {
-        let mut repos = catalog.repositories().await;
-        let partition = repos
+        let partition = catalog
+            .repositories()
+            .await
             .partitions()
             .create_or_get(partition_key, shard_id, self.table_id)
             .await
             .context(super::CatalogSnafu)?;
 
-        // get info on the persisted parquet files to use later for replay or for snapshot
-        // information on query.
-        let files = repos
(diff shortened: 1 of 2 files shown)


ingester/src/data/table.rs
```diff
@@ -161,26 +161,17 @@ impl TableData {
...
-        // get info on the persisted parquet files to use later for replay or for snapshot
-        // information on query.
-        let files = repos
-            .parquet_files()
-            .list_by_partition_not_to_delete(partition.id)
-            .await
-            .context(super::CatalogSnafu)?;
-        // for now we just need the max persisted
-        let max_persisted_sequence_number = files.iter().map(|p| p.max_sequence_number).max();
-
```
domodwyer, line 180: :wave:

Removing this query was the intent of this series of PRs - we no longer block ingest and scan the giant `parquet_files` table in the hot path.

mkmik (approved):
Awesome!

Issue #5008, opened 2022-06-30
Catalog queries should be retried


domodwyer:
## TL;DR

All Catalog queries should retry when Postgres returns any error.

For example, Ingester can lose data due to transient errors; simply retry logic would fix this.

## Technical Description

Currently the ingester [does not](https://github.com/influxdata/influxdb_iox/blob/b8e7071f8229ec729369816430ac5c23d46b2f2e/ingester/src/data.rs#L143-L146) retry catalog queries (one example but there's many instances).

Once an error is returned from the internal buffer, the stream handler drops the write and continues consuming (incrementing the `potential_data_loss=true` metric):

https://github.com/influxdata/influxdb_iox/blob/b8e7071f8229ec729369816430ac5c23d46b2f2e/ingester/src/stream_handler/handler.rs#L343-L353

The buffer should retry non-fatal errors, and only return fatal ("this write will never succeed") errors to the stream handler.

There have been no incidence of errors causing data loss in the last 24h:

![image](https://user-images.githubusercontent.com/9275968/176700471-def2c914-91a4-4801-90c8-579262d80774.png)

Catalog errors are fortunately fairly rare, but induced when the catalog is under excessive load.

crepererum:
We do that within the querier, so I would label this as https://github.com/influxdata/influxdb_iox/labels/Ingester as originally filed.

jacobmarble:
@domodwyer please update status

domodwyer:
Low risk and alerted on when it does happen - I intend to fix this by making the queries unnecessary via #4880. We could retry all query errors but we risk live-locking the ingester should a namespace/table/column not exist in the catalog.

#4880 is sufficiently close to being implemented, and this issue sufficiently rare, that it's not worth a dodgy intermediate fix.

domodwyer:
This was resolved by #6069 🎉

PR #5658, opened 2022-09-16
refactor(ingester): partition provider abstraction

domodwyer:
This PR is a stepping-stone to implementing a partition cache, in order to significantly reduce hot-path queries at ingester startup (https://github.com/influxdata/influxdb_iox/issues/5638). 

It also improves testability of the ingester, which is something I intend to continue to push - it removes a test-only constructor, driving everything through the same code paths (less paths == less problems, though minor in this case). It also eliminates the need to construct a replica catalog with the various cross-linking references - often a source of excessive verbosity in the ingester tests, and fragile (it hit me multiple times when adding a new field in #5648). 

This also **fixes a potential data loss** - we've know about it for a while (https://github.com/influxdata/influxdb_iox/issues/5008) and have observed this happening in prod when the catalog was down. This now no longer happens for partition lookups, and I'll implement something similar for the other queries, or add simple retry logic as appropriate.

Relates to #5638, and #5008.

---

* refactor: decouple partition init from table (c95a5fb74)

      Removes the "how" of initialising a per-partition buffer structure
      (PartitionData) from the per-table buffer (TableData).

      This is a cleaner separation of concerns - a table buffer is responsible for
      addressing and initialising per-table partitions as necessary, and buffering
      of ops for them. It does not have to be concerned with the series of steps
      necessary to look up the various bits of data in order to construct a
      PartitionData.

      This abstract provider can be layered up to provide more complex behaviours -
      I intend to add a read-through cache impl that decorates the catalog impl in
      this commit, which should eliminate most partition queries at ingester startup
      utilising the indirection added here.

* refactor: ignore dead snapshot() code (b23182317)

      This method SHOULD be used by the prod code, but currently is only used by
      tests - prod code "peeks" through to the DataBuffer fields directly.

      I will resolve this peeking in a future PR.


domodwyer:
The failing tests construct every bit of ingester state from the bottom up - it populates each buffer, partition, table, namespace and shard and is therefore very tightly coupled with the current code - it makes a lot of assumptions about how the ingester works internally, rather than driving the ingester and asserting on the result.

I am untangling these tests - instead of setting the various bits of state to reflect what the test **believes** should be the state after buffering a write/delete, I am instead writing the test to simply buffer the writes/deletes through the ingester using the same code paths as actual writes/deletes and **test the resulting state** that the ingester builds up.

This is a large (test code only) change, and I will PR it separately - this PR is on hold until then :+1:

alamb:
Marking PR as draft per https://github.com/influxdata/influxdb_iox/pull/5658#issuecomment-1249464240

domodwyer:
This PR was rebased onto the changes in #5697 and is now ready for review.

NGA-TRAN (approved):
The PR does it is described

PR #5660, opened 2022-09-16
refactor(ingester): remove redundant fn args

domodwyer:
All this PR does is change a function like this:

```rust
partition_data.snapshot_to_persisting_batch(
  partition_info.partition.shard_id,
  partition_info.partition.table_id,
  partition_info.partition.id,
  &partition_info.table_name,
)
```

which has a risk of being passed the wrong thing, instead removing the redundant args (a partition should know it's own ID):

```rust
partition_data.snapshot_to_persisting_batch()
```

It replicates the same pattern where necessary for the shard, namespace, table & partition.

There catalog could be pulled out of fn args in a similar way, but I have grander plans for it that will help simplify testing so it can live there for a bit longer.

---

* refactor: snapshot_to_persisting redundant ID (85d6efafe)

      Partition::snapshot_to_persisting() passes the ID of the partition it is 
      calling `snapshot_to_persisting()` on. The partition already knows what its ID
      is, so at best it's redundant, and at worst, inconsistent with the actual ID.

* refactor: ShardId & TableId in PartitionData (c7ba0bea9)

      When we construct a PartitionData we have the ShardId and TableId. This commit
      stores them in the PartitionData for later use, rather than repeatedly passing
      them in again when constructing snapshots, at the risk of passing the wrong
      IDs.

* refactor: add table name in PartitionData (07b08fa9c)

      A partition belongs to a table - this commit stores the table name in the
      PartitionData (which was readily available at construction time) instead of
      redundantly passing it into various functions at the risk of getting it wrong.

* refactor: store ShardId in child nodes (b0eb85ddd)

      Instead of passing the ShardId into each function for child nodes of the 
      Shard, store it. This avoids the possibility of mistakenly passing the wrong
      value.

### DIFF
--- a/ingester/src/data.rs
+++ b/ingester/src/data.rs
@@ -267,7 +267,12 @@ impl Persister for IngesterData {
             .await
             .expect("retry forever");
 
-        let persisting_batch = namespace.snapshot_to_persisting(&partition_info).await;
+        let persisting_batch = namespace
+            .snapshot_to_persisting(
+                &partition_info.table_name,
+                &partition_info.partition.partition_key,
+            )
+            .await;
 
         if let Some(persisting_batch) = persisting_batch {
             // do the CPU intensive work of compaction, de-duplication and sorting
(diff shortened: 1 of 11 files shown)


jacobmarble (approved):
Nice cleanup 👍

PR #5661, opened 2022-09-16
chore: reduce number of histogram buckets

NGA-TRAN:
Now we know which buckets data most fall into, let us reduce the number of buckets to avoid overwhelming our metric collection


alamb (approved):
LGTM

PR #5669, opened 2022-09-19
chore: replace `ansi_style` with `nu-ansi-style`

crepererum:
See <https://rustsec.org/advisories/RUSTSEC-2021-0139.html>.

This follows the migration that `tracing-subscriber` will perform: https://github.com/tokio-rs/tracing/pull/2287

Note that `tracing-subscriber` is not released yet, so for the time being we will have `ansi_style` and `nu-ansi-style` in our dependency list. It's likely that `tracing-subscriber` will backport and and release this change rather soon though, because people are getting annoyed by RUSTSEC warnings.


mkmik (approved):

alamb:
An alternative would be to avoid ANSI styling at all (and just leave the messages as plain text). I sometimes find the styling quite annoying when working with the logs locally (as the control codes are still present when piping them to `less` or  a file(

alamb:
re ANSI coloring: https://github.com/influxdata/influxdb_iox/pull/5674

PR #5673, opened 2022-09-19
refactor: improve "remove if" cache checks

crepererum:
Currently "remove if" checks count as "key was used" in front of the LRU system. This has two problems:

- technically: it is unnessarily expensive to update the LRU addressable heap
- semantically: the value wasn't really used. "remove if" does not leak the actual value[^1]

So let's not treat these checks as "key was used".

[^1]: We technically do not prevent leaking the value, but it would require quite some effort and the code would look pretty weird and (hopefully) this would ring some alarm bells.


cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -208,6 +209,32 @@ async fn test_lru_learns_about_ttl_evictions() {
...
+    backend.set(1, String::from("a"));
+    time_provider.inc(Duration::from_secs(1));
+
+    backend.set(2, String::from("b"));
+    time_provider.inc(Duration::from_secs(1));
+
+    remove_if_handle.remove_if(&1, |_| false);
+    backend.set(3, String::from("c"));
+
+    assert_eq!(backend.get(&1), None);
```
alamb, line 235:
```suggestion
    // adding "c" totals 12 size, but backend has room for only 10
    // so "least recently used" (in this case 1, not 2) should be removed
    assert_eq!(backend.get(&1), None);
    assert!(backend.get(&2).is_some();
```

cache_system/src/backend/policy/mod.rs
```diff
@@ -599,6 +598,22 @@ where
     records: Vec<Record<K, V>>,
 }
 
+impl<K, V> Recorder<K, V>
```
alamb, line 601: I think `Observer` or `Peeker` might be more common terms here for this concept. But this name is fine too in my opinion

alamb (approved):
LGTM!

cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -208,6 +209,32 @@ async fn test_lru_learns_about_ttl_evictions() {
...
+    size_estimator.mock_size(2, String::from("b"), TestSize(4));
+    size_estimator.mock_size(3, String::from("c"), TestSize(4));
+
+    backend.set(1, String::from("a"));
+    time_provider.inc(Duration::from_secs(1));
+
+    backend.set(2, String::from("b"));
+    time_provider.inc(Duration::from_secs(1));
+
+    remove_if_handle.remove_if(&1, |_| false);
```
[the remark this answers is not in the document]
alamb: I suggested some comments to make it more clear to future readers what functionality the test was ensuring

PR #5675, opened 2022-09-19
refactor: Add helper methods for computing times to TimeProvider

carols10cents:
There are a bunch of places this sort of math is done in the compactor and catalog, and rather than copy-pasting this everywhere, it can be a method on `TimeProvider`.

### DIFF
--- a/compactor/src/cold.rs
+++ b/compactor/src/cold.rs
@@ -116,7 +116,7 @@ mod tests {
     use iox_query::exec::Executor;
     use iox_tests::util::{TestCatalog, TestParquetFileBuilder};
     use iox_time::{SystemProvider, TimeProvider};
-    use std::{collections::HashMap, time::Duration};
+    use std::collections::HashMap;
 
     #[tokio::test]
     async fn test_compact_remaining_level_0_files_many_files() {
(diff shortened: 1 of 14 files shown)


iox_time/src/lib.rs
```diff
@@ -190,6 +190,18 @@ pub trait TimeProvider: std::fmt::Debug + Send + Sync + 'static {
 
     /// Sleep until given time.
     fn sleep_until(&self, t: Time) -> Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+
+    /// Given a number of minutes, return the number of nanoseconds the specified number of minutes
+    /// in the past relative to this provider's `now`.
+    fn minutes_ago_in_ns(&self, minutes_ago: u64) -> i64 {
```
crepererum, line 196: Why don't we directly return `Timestamp` or `Time` here? It seems weird to me that we convert one untyped integer into another.

carols10cents: Good call! Added :)

compactor/src/cold.rs
```diff
@@ -178,7 +178,7 @@ mod tests {
 
         let partition = table.with_shard(&shard).create_partition("part").await;
         let time = Arc::new(SystemProvider::new());
-        let time_38_hour_ago = (time.now() - Duration::from_secs(60 * 60 * 38)).timestamp_nanos();
+        let time_38_hour_ago = time.hours_ago(38);
```
NGA-TRAN, line 181: 🎉

NGA-TRAN (approved):
Love this

PR #5676, opened 2022-09-19
refactor: improve interaction of LRU and cache refreshs

crepererum:
~On top of #5673.~

We should not treat refreshs as "was used" actions, i.e. it should be possible that entries that are refreshed are still kicked out by the LRU.


cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -95,6 +95,49 @@ async fn test_refresh_sets_new_expiration_after_it_finishes() {
...
+    time_provider.inc(Duration::from_secs(1));
+    backend.set(2, String::from("b"));
+
+    time_provider.inc(Duration::from_secs(1));
+
+    notify_idle.notified_with_timeout().await;
+    barrier.wait().await;
+    notify_idle.notified_with_timeout().await;
+
+    backend.set(3, String::from("c"));
```
alamb, line 136:
```suggestion
    // add a third item to the cache, forcing LRU to evict one of the items
    backend.set(3, String::from("c"));
```

cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -95,6 +95,49 @@ async fn test_refresh_sets_new_expiration_after_it_finishes() {
...
+
+    let barrier = loader.block_next(1, String::from("foo"));
+    backend.set(1, String::from("a"));
+
+    // trigger refresh
+    time_provider.inc(Duration::from_secs(1));
+    assert_eq!(backend.get(&1), Some(String::from("a")));
+
+    time_provider.inc(Duration::from_secs(1));
+    backend.set(2, String::from("b"));
```
alamb, line 128: I was expecting `2` to be set before the refresh was triggered -- otherwise since it was added after `1` already I would expect that it would be treated as least recently reused even without `1` refreshing. 

Or perhaps that is what the second `    time_provider.inc(Duration::from_secs(1));` does (causes `1` to refresh after `2` was added). But if that is the case, why is this refreshing `1` twice 🤔

cache_system/src/backend/policy/integration_tests.rs
```diff
@@ -95,6 +95,49 @@ async fn test_refresh_sets_new_expiration_after_it_finishes() {
...
+
+    time_provider.inc(Duration::from_secs(1));
+
+    notify_idle.notified_with_timeout().await;
+    barrier.wait().await;
+    notify_idle.notified_with_timeout().await;
+
+    backend.set(3, String::from("c"));
+
+    assert_eq!(backend.get(&1), None);
```
alamb, line 138:
```suggestion
    // Should evict `1` even though it was refreshed after `2` was added
    assert_eq!(backend.get(&1), None);
```

cache_system/src/backend/policy/lru.rs
```diff
@@ -653,10 +657,11 @@ where
         // maybe clean from pool
         {
             let mut inner = self.inner.lock();
-            if let Some((consumption, _last_used)) = inner.last_used.remove(&k) {
+            if let Some((consumption, last_used_already)) = inner.last_used.remove(&k) {
```
alamb, line 660:
```suggestion
            if let Some((consumption, last_used_previously)) = inner.last_used.remove(&k) {
```

Minor naming suggestion I would have found easier to understand

cache_system/src/backend/policy/lru.rs
```diff
@@ -940,6 +945,39 @@ mod tests {
...
+        backend.set(String::from("b"), 1usize);
+        time_provider.inc(Duration::from_millis(1));
+
+        // does NOT count as "used"
+        backend.set(String::from("a"), 1usize);
+        time_provider.inc(Duration::from_millis(1));
+
+        backend.set(String::from("c"), 1usize);
+
+        assert_eq!(backend.get(&String::from("a")), None);
```
alamb, line 978: 👍

alamb (approved):
Makes sense to me

cache_system/src/backend/policy/integration_tests.rs
crepererum: So the "GET" request that starts the refresh will be seen by the LRU as "used". However the refresh itself "takes a while" (technically: it is blocked by the `barrier`). When the refresh finishes, the LRU should NOT treat this as another "used" signal.

PR #5677, opened 2022-09-19
feat: Improve compactor code organization and test coverage

carols10cents:
This PR's goal is to move logic specific to hot compaction within the "hot" module, make it easier to test with less test setup, and add unit tests for the specific behavior.

This uses the code in https://github.com/influxdata/influxdb_iox/pull/5675 so that needs to be merged first.

I highly recommend reviewing commit-by-commit.


NGA-TRAN (approved):
These make sense to me. Thanks @carols10cents

PR #5678, opened 2022-09-19
feat: send only needed projection columns from querier to ingester in…

### LINKED ISSUE #5740: querier -> ingester: only send needed columns to Ingester in the case of SQL queries
SQL case of projection pushdown in `querier -> ingester` reported in https://github.com/influxdata/influxdb_iox/issues/5624

NGA-TRAN:
… case of normal SQL queries

Closes #5740  
First part of #5624 


influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -293,25 +293,45 @@ async fn ingester_panic() {
...
-                        .arg(log_data["predicate_binary"].as_str().unwrap())
-                        .assert();
-
-                    // The debug query should work.
-                    assert.success().stdout(
-                        predicate::str::contains(
-                            "| A    | B    | 1970-01-01T00:00:00.000123456Z | 42  |",
-                        )
-                    );
+                    for i in 0..2 {
```
NGA-TRAN, line 296: This is a merge conflict. I tried to remove this loop but I still get C
@crepererum : Do I missed anything?

NGA-TRAN: Hmm the test in circle CI disagrees with me. I will try this again tomorrow to have the test pass

crepererum: The loop is gone now, just leave the version where the CLI gets the result. My change also removed the "not" part of the predicate because the ingester no longer applies filters

querier/src/namespace/query_access.rs
```diff
@@ -57,6 +57,7 @@ impl QueryDatabase for QuerierNamespace {
             .chunks(
                 predicate,
                 ctx.span().map(|span| span.child("querier table chunks")),
+                &None, // todo: pushdown projection to chunks
```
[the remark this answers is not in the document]
crepererum: I agree that this is the beefy part 😅

querier/src/ingester/test_util.rs
```diff
@@ -32,15 +39,84 @@ impl IngesterConnection for MockIngesterConnection {
...
+                .lock()
+                .take()
+                .unwrap_or_else(|| Ok(vec![]));
+        }
+
+        // no partitions
+        let partitions = self.next_response.lock().take();
+        if partitions.is_none() {
+            return Ok(vec![]);
+        }
```
crepererum, line 71: this is unnecessary

querier/src/table/mod.rs
```diff
@@ -667,6 +694,94 @@ mod tests {
...
+            )
+            .unwrap()
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .map(Result::unwrap)
+            .collect::<Vec<_>>();
+
+        // Return value should have foo (a part of projectpushown), and other PK columns (tag1, tag1, time)
+        // The result looks like this but indeterministic
```
crepererum, line 765: What's nondeterministic here? The column order? Then I think you should sort the projection (see other comment).

querier/src/table/mod.rs
```diff
@@ -416,18 +427,33 @@ impl QuerierTable {
...
+                    .iter()
+                    .map(|i| self.schema().field(*i).1.name().to_string())
+                    .collect::<HashSet<_>>();
+
+                // Add missing PK columnns (tags and time) as they are needed for deduplication
+                let pk = self.schema.primary_key();
+                for col in pk {
+                    columns.insert(col.to_string());
+                }
+                columns.into_iter().collect::<Vec<String>>()
```
crepererum, line 447: You probably wanna wort this to get a deterministic behavior.

querier/src/table/mod.rs
```diff
@@ -667,6 +694,94 @@ mod tests {
...
+
+        let chunk = &chunks[0];
+
+        // verify chunk schema
+        let schema = chunk.schema();
+        let fields = schema
+            .fields_iter()
+            .map(|i| i.name().to_string())
+            .collect::<Vec<_>>();
+        assert_eq!(fields, vec!["foo"]); // mpte: no bar
```
crepererum, line 734: mpte?

querier/src/ingester/test_util.rs
NGA-TRAN: Two tests `test_chunk_size_limit` and `test_query` need this condition because they provided no partition which will panic when I do unwrap below

querier/src/table/mod.rs
NGA-TRAN: 😄 Fixed

querier/src/table/mod.rs
NGA-TRAN: Great hint. Done. Sorted the columns

querier/src/table/mod.rs
NGA-TRAN: Yeah, sorting the columns make This deterministic. I have updated the test accordingly

influxdb_iox/tests/end_to_end_cases/querier.rs
NGA-TRAN: Thanks for the hint. I have updated the test using your version in which ingester no longer applies filters

NGA-TRAN:
@alamb : since Marco is on FTO, can you help review this? I have addressed all his comments

querier/src/table/mod.rs
```diff
@@ -416,18 +427,35 @@ impl QuerierTable {
...
-        // As a future optimization, might be able to fetch only
-        // fields that are needed in query
-        let columns: Vec<String> = self
-            .schema
-            .iter()
-            .map(|(_, f)| f.name().to_string())
-            .collect();
+        // If the projection is provided, use it. Otherwise, use all columns of the table
+        // The provided projection should include all columns needed by the query
+        let columns: Vec<String> = match projection {
```
alamb, line 434: What do you think about moving this logic into Schema. e.g. here:

https://github.com/influxdata/influxdb_iox/blob/main/schema/src/lib.rs#L307

I think it would be easier to unit test there as well as being more discoverable

querier/src/table/mod.rs
```diff
@@ -416,18 +427,35 @@ impl QuerierTable {
...
+                    .map(|i| self.schema().field(*i).1.name().to_string())
+                    .collect::<HashSet<_>>();
+
+                // Add missing PK columnns (tags and time) as they are needed for deduplication
+                let pk = self.schema.primary_key();
+                for col in pk {
+                    columns.insert(col.to_string());
+                }
+                let mut columns = columns.into_iter().collect::<Vec<String>>();
+                columns.sort();
```
alamb, line 448: Is it correct to sort by column name here? Don't we need to keep the same order as the `projection` passed in ?

querier/src/table/mod.rs
```diff
@@ -667,6 +697,82 @@ mod tests {
...
+
+        // verify chunk schema
+        let schema = chunk.schema();
+        let fields = schema
+            .fields_iter()
+            .map(|i| i.name().to_string())
+            .collect::<Vec<_>>();
+        // only foo column. No bar
+        assert_eq!(fields, vec!["foo"]);
+        let tags = schema
```
alamb, line 739:
```suggestion
        // all tags should be present
        let tags = schema
```

querier/src/table/mod.rs
```diff
@@ -667,6 +697,82 @@ mod tests {
...
+            .tags_iter()
+            .map(|i| i.name().to_string())
+            .collect::<Vec<_>>();
+        assert_eq!(tags.len(), 2);
+        assert!(tags.contains(&"tag1".to_string()));
+        assert!(tags.contains(&"tag2".to_string()));
+        //
```
alamb, line 746: Seems like a missing comment?

querier/src/table/mod.rs
```diff
@@ -667,6 +697,82 @@ mod tests {
...
+        let times = schema
+            .time_iter()
+            .map(|i| i.name().to_string())
+            .collect::<Vec<_>>();
+        assert_eq!(times, vec!["time"]);
+
+        // verify chunk data
```
alamb, line 753: Given the data has the schema implicitly in it, I think simply checking the resulting batch (maybe with comments about both tags being present along with the field) should be good enough. 

No need to change, just a comment

alamb (approved):
Other than the column ordering thing looks good to me. I had some small refactorings suggestions but nothing required

querier/src/table/mod.rs
NGA-TRAN: @alamb : since we add tags and time, the code before the sort I think already in different order as the original projection. 

I am trying to think if there are any reason to keep the same order as the projection 🤔 . It seems the order is not important here. We only need to know the columns we want data. I think the returned schema keep all needed information for us to map the column index and name correctly

querier/src/table/mod.rs
NGA-TRAN: @alamb : do you mean we create a new function inside Schema for this purpose?
I am working on that.

schema/src/lib.rs
```diff
@@ -314,6 +314,32 @@ impl Schema {
         })
     }
 
+    /// Return names of the columns of given indexes with all PK columns (tags and time)
```
alamb, line 317: ❤️

PR #5680, opened 2022-09-19
fix: Make us log logs from the log crate again

### LINKED ISSUE #5679: Logs sent to the `log` crate are sent to dev/null
Our developer doc states:

```
## Developer Guide

IOx makes use of Rust's [tracing](https://docs.rs/tracing) ecosystem to output application logs to stdout. It
additionally makes use of [tracing-log](https://docs.rs/tracing-log) to ensure that crates writing events to
the [log](https://docs.rs/log/) facade are also rendered to stdout.
```

__Steps to reproduce:__

1. Add this snippet somewhere where it gets executed:

```rust
    error!("BEFORE log::error");
    log::error!("DEMO LOG");
    error!("AFTER log::error");
```

__Expected behaviour:__

```
2022-09-19T21:48:27.978623Z ERROR iox_catalog::postgres: BEFORE log::error
2022-09-19T21:48:27.978623Z ERROR iox_catalog::postgres: DEMO LOG
2022-09-19T21:48:27.978667Z ERROR iox_catalog::postgres: AFTER log::error
```

__Actual behaviour:__

```
2022-09-19T21:48:27.978623Z ERROR iox_catalog::postgres: BEFORE log::error
2022-09-19T21:48:27.978667Z ERROR iox_catalog::postgres: AFTER log::error
```

__Environment info:__

* `cargo run` (all-in-one)
* System info: `Darwin 21.6.0 arm64`

mkmik:
Don't merge before merging https://github.com/influxdata/influxdb_iox/pull/5681

---

Closes #5679

Regression introduced in #4279.

We have two code paths to initialize logging. The "simple" code path uses a Builder and is used by the client side commands. A more complex init code path is used for the server side commands.

When we introduced unified logging in https://docs.rs/tracing-log/latest/tracing_log/ we performed the initalization in the trogging `Builder`. When we introduced support for `tokio-console` we streamlined the initialization code to no longer use the `Builder` but that meant that we didn't share tracing_log crate init call.

Ideally we should unify those two setup paths once again.

This PR performs a lazy take on this unification by only factoring out some common code into an `install_global<S: Subscriber>` function. This solves #5679 and provides for a natural place to add more common initialization of the tracing  subscriber.


jacobmarble (approved):

alamb:
This also explains where the datafusion logging (which also uses the log crate) messages went. Thanks @mkmik

PR #5684, opened 2022-09-19
fix: Close old connection pool after swap

### LINKED ISSUE #5683: Catalog DB connections leak on hotswap
Symptom: postgres connection count keeps growing as time passes

image

When DB credentials change and the DSN file gets updated, IOx "hot reloads" the DB connection pool, but the old connection pool doesn't get properly disposed.

__Steps to reproduce:__

0. checkout https://github.com/influxdata/influxdb_iox/pull/5682 if not merged yet
1. create a local postgres DB, e.g. `iox_shared`
2. run all-in-one with a `dsn-file://` DSN: `cargo run -- run all-in-one --data-dir /tmp/iox_data --catalog-dsn dsn-file:///tmp/test-dsn.url --catalog-hotswap-poll-interval 1s --catalog-max-connections 10`
3. update the `for i in $(seq 1 600) ; do  echo -n "postgres:///iox_shared$(printf ' %.0s' {1..$i})" >/tmp/test-dsn.url; sleep 1; wc -c /tmp/test-dsn.url; done`

__Expected behaviour:__

```
iox_shared=# select application_name,state,count(*) from pg_stat_activity group by application_name,state having application_name = 'iox-all-in-one';
 application_name | state | count
------------------+-------+-------
 iox-all-in-one   | idle  |     3
```

__Actual behaviour:__

```
iox_shared=# select application_name,state,count(*) from pg_stat_activity group by application_name,state having application_name = 'iox-all-in-one';
 application_name | state | count
------------------+-------+-------
 iox-all-in-one   | idle  |     84
```

__Environment info:__

* Please provide the command you used to build the project, including any `RUSTFLAGS`.
* System info: Run `uname -srm` or similar and copy the output here (we want to know your OS, architecture etc).
* If you're running IOx in a containerised environment then details about that would be helpful.
* Other relevant environment details: disk info, hardware setup etc.

__Config:__
Copy any non-default config values here or attach the full config as a gist or file.

__Logs:__
Include snippet of errors in logs or stack traces here.
Sometimes you can get useful information by running the program with the `RUST_BACKTRACE=full` environment variable.
Finally, the IOx server has a `-vv` for verbose logging.

mkmik:
Closes #5683

(if you want to test this locally, you may need to cherry-pick https://github.com/influxdata/influxdb_iox/pull/5682 or wait for it to be merged)

### DIFF
--- a/iox_catalog/src/postgres.rs
+++ b/iox_catalog/src/postgres.rs
@@ -431,7 +431,14 @@ async fn new_pool(
                         Ok(None)
                     } else {
                         let new_pool = new_raw_pool(options, &new_dsn).await?;
-                        pool.replace(new_pool);
+                        let old_pool = pool.replace(new_pool);
+                        info!("replaced hotswap pool");
+                        info!(?old_pool, "closing old DB connection pool");
+                        // The pool is not closed on drop. We need to call `close`.
+                        // It will close all idle connections, and wait until acquired connections
+                        // are returned to the pool or closed.
+                        old_pool.close().await;
+                        info!(?old_pool, "closed old DB connection pool");
                         Ok(Some(new_dsn))
                     }
                 }
@@ -439,7 +446,6 @@ async fn new_pool(
                 match try_update(&options, &current_dsn, &dsn_file, &pool).await {
                     Ok(None) => {}
                     Ok(Some(new_dsn)) => {
-                        info!("replaced hotswap pool");
                         current_dsn = new_dsn;
                     }
                     Err(e) => {


cannonpalms:
Is it possible for there to be connections in the old pool in active use (mid-query perhaps) when this code to close the pool is executed?

Vault-agent, by default, will renew its lease on the dynamic database secret somewhere in the 70-90% of total lifetime range. This code, if prompted by a change to the DSN file, can probably be called in the middle of another async context. 

Is there a difference between closing the pool and closing the underlying connections?

mkmik:
Yes, it's possible there are existing in-use connection. Closing the pool doesn't not close those connections. Instead the close call waits until the connections are either returned to the pool or otherwise closed.

This PR contains a comment that explains it:

```
  // The pool is not closed on drop. We need to call `close`.
  // It will close all idle connections, and wait until acquired connections
  // are returned to the pool or closed.
```

Any suggestions to how to improve the comment are appreciated.

cannonpalms:
Thanks. I didn't catch the comment. 

What if the connections are never returned to the pool? Are they forcibly closed after some timeout? Or would a true connection leak result in the pool never being closed and this problem reoccurring?

mkmik:
If for some reason the `.close` call never terminates, the periodic credentials refresh will not cause new pools to be hot swapped, and eventually after 1h the pod will crash because the db connections won't authenticate.

This "good" because the current behavior is to overload the DB and enter a state that requires manual intervention to recover from.

cannonpalms:
Agreed, this behavior is more desirable.

cannonpalms (approved):

mkmik:
Even if the pod doesn't crash when it can no longer authenticate to the db (e.g. because our liveness check is broken); it still cannot overshoot the connection limit wildly as it currently does.

That said, the reasoning around the connection pool limit needs to be adjusted in light of all this:

If max connections is N, in the worst case it's possible that you have one pool use N connections, then a hotswap event creating a new pool, and while the old N connections are used, the new pool quickly scales up to its limit, consuming 2N connections in total. 

This is worst case, but I think it's a good idea to size the db instance and connections limits with a decent leeway.

mkmik:
The info logs emitted by this PR will help us see the number of the active connections in the "old pool" (idle connections are closed immediately). This will help us figure out how likely is the worst case.

Hopefully we'll "soon" have a catalog service, where it will be easier to size connection pools properly.

PR #5691, opened 2022-09-20
fix: GC `parquet_file` table in batches

crepererum:
Otherwise this transaction will never finish in prod.


garbage_collector/src/parquetfile/deleter.rs
```diff
@@ -23,11 +23,13 @@ pub(crate) async fn perform(
             .context(DeletingSnafu)?;
         info!(delete_count = %deleted.len(), "iox_catalog::delete_old()");
 
-        select! {
-            _ = shutdown.recv() => {
-                break
-            },
-            _ = sleep(Duration::from_secs(60 * sleep_interval_minutes)) => (),
+        if deleted.is_empty() {
```
alamb, line 26: Does this change mean the garbage collector won't respond to shutdowns until it has no files to delete?

iox_catalog/src/postgres.rs
```diff
@@ -1719,10 +1719,17 @@ RETURNING *;
     }
 
     async fn delete_old_ids_only(&mut self, older_than: Timestamp) -> Result<Vec<ParquetFileId>> {
+        // see https://www.crunchydata.com/blog/simulating-update-or-delete-with-limit-in-postgres-ctes-to-the-rescue
         let deleted = sqlx::query(
             r#"
+WITH parquet_file_ids as (
```
alamb, line 1725: 👍

iox_catalog/src/postgres.rs
```diff
@@ -1719,10 +1719,17 @@ RETURNING *;
...
+    SELECT id
+    FROM parquet_file
+    WHERE to_delete < $1
+    LIMIT 1000
```
alamb, line 1729: Is it worth making this 1000 a named constant somewhere?

alamb (approved):
makes sense to me. I don't really have any good ideas for testing sadly

garbage_collector/src/parquetfile/deleter.rs
crepererum: good catch, fixed

iox_catalog/src/postgres.rs
crepererum: done

lukebond:
:+1: thank you @crepererum

Issue #5663, opened 2022-09-19
InfluxQL: Add comment support


stuartcarnie:
InfluxQL supports single line and multi-line comments, as defined by [the documentation](https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#comments).

### Usage

Inspecting over 2.5 million queries revealed no usage

### Line comments

Line comments begin with `--` and run to the end of the line

#### Usage

Inspecting over 2.5 million Cloud 2 query logs revealed no usage of the `--` comment token.

### Multi-line comments

Multi-line comments begin with `/*` and terminate at the next `*/`, which may span multiple lines. **Multi-line comments do not support nesting.**

#### Usage

Inspecting over 2.5 million Cloud 2 query logs revealed roughly 360 uses of the multi-line comment token, all to variations of the following query:

```SQL
SELECT SUM("foo")  / (90000 * 30.65111111) * 100  FROM
(
  SELECT non_negative_difference(max("value")) as "foo"
  FROM "counter"
  WHERE
  (
          "col1" = 'val1' AND "col2" = 'val2' AND "col3" = 'val3' AND   time >= '2022-08-08 17:00:00' and time <=  now()   /*   '2022-08-09 23:39:05'   */
  )
  GROUP BY time(1m)
)
```

PR #5692, opened 2022-09-20
feat: Add `SELECT` statement parsing

stuartcarnie:
Helps #5633

This PR teaches the InfluxQL parser how to recognise a `SELECT` statement. It is accompanied by a number of unit tests, however, #5738 is intended to broaden the testing to a larger corpus of statements.

### Other notable changes

Refactors expression parsing into two modules to distinguish arithmetic expressions from conditional expressions. InfluxQL universally allows conditional expressions in a `WHERE` clause of any statement, but only arithmetic expressions in other places, such as the projection list of a `SELECT` statement.

Adds a new generic type, `OneOrMore` and accompanying trait, `Parser` to the `common` module to allow generic parsing of a list of parsers. This is used in both the `expression::arithmetic` and `expression::conditional` modules to permit customisation of the valid set of operands which may be parsed. This was necessary, as there are different set of operands for a field in a `SELECT` projection vs in a `WHERE` clause.


influxdb_influxql_parser/src/common.rs
```diff
@@ -83,29 +85,24 @@ pub fn measurement_name_expression(i: &str) -> ParseResult<&str, MeasurementName
...
-
-/// Parse a LIMIT <n> clause.
+/// Parse a `LIMIT <n>` clause.
 pub fn limit_clause(i: &str) -> ParseResult<&str, u64> {
     preceded(
         pair(tag_no_case("LIMIT"), multispace1),
         expect(
             "invalid LIMIT clause, expected unsigned integer",
-            unsigned_number,
+            unsigned_integer,
```
stuartcarnie, line 94: Reuse some existing code

influxdb_influxql_parser/src/delete.rs
```diff
@@ -46,7 +46,7 @@ pub fn delete_statement(i: &str) -> ParseResult<&str, DeleteStatement> {
     preceded(
         tag_no_case("DELETE"),
         expect(
-            "invalid DELETE statement, must be followed by FROM or WHERE",
+            "invalid DELETE statement, expected FROM or WHERE",
```
stuartcarnie, line 49: Followed this pattern everywhere of 

> invalid _something_, expected _list of symbols / tokens_

influxdb_influxql_parser/src/identifier.rs
```diff
@@ -11,8 +11,6 @@
 //! [identifier]: https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#identifiers
 //! [keywords]: https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#keywords
 
-#![allow(dead_code)]
-
```
stuartcarnie, line 15: All of these have been removed, as we no longer have any dead code 🥳

influxdb_influxql_parser/src/literal.rs
```diff
@@ -121,7 +120,7 @@ fn integer(i: &str) -> ParseResult<&str, i64> {
 /// ```text
 /// INTEGER ::= [0-9]+
 /// ```
-fn unsigned_integer(i: &str) -> ParseResult<&str, u64> {
+pub fn unsigned_integer(i: &str) -> ParseResult<&str, u64> {
```
stuartcarnie, line 123: Reuse this in several places

influxdb_influxql_parser/src/show_measurements.rs
```diff
@@ -321,13 +330,10 @@ mod test {
         );
 
         let (_, got) = on_clause("ON *").unwrap();
-        assert!(matches!(got, OnExpression::AllDatabases));
+        assert_matches!(got, OnExpression::AllDatabases);
```
stuartcarnie, line 333: Per your recommendation, @crepererum – thanks!

influxdb_influxql_parser/src/show_tag_values.rs
```diff
@@ -130,37 +128,20 @@ impl Display for WithKeyExpression {
...
+            Self::In(list) => write!(f, "IN ({})", list)?,
         };
 
         Ok(())
     }
 }
 
 /// Parse an identifier list, as expected by the `WITH KEY IN` clause.
-fn identifier_list(i: &str) -> ParseResult<&str, (Identifier, Option<Vec<Identifier>>)> {
+fn identifier_list(i: &str) -> ParseResult<&str, InList> {
```
stuartcarnie, line 139: `InList` leverages the new `OneOrMany` type

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        // Expressions are still valid when unnecessary whitespace is omitted
+
+        let (_, got) = arithmetic_expression("5+51").unwrap();
+        assert_eq!(got, binary_op!(5, Add, 51));
+
+        let (_, got) = arithmetic_expression("5+$foo").unwrap();
+        assert_eq!(got, binary_op!(5, Add, param!("foo")));
+
+        let (_, got) = arithmetic_expression("5- -(3|2)").unwrap();
```
NGA-TRAN, line 476: I wonder what happens with the expressions "5--(3|2)", "2|5%3"

influxdb_influxql_parser/src/expression/conditional.rs
```diff
@@ -0,0 +1,395 @@
...
+use nom::bytes::complete::{tag, tag_no_case};
+use nom::character::complete::{char, multispace0};
+use nom::combinator::{cut, map, value};
+use nom::multi::many0;
+use nom::sequence::{delimited, preceded, tuple};
+use std::fmt;
+use std::fmt::{Display, Formatter, Write};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConditionalOperator {
```
NGA-TRAN, line 17: It will be consistent and informative if you add a comment to link to the document of these operators as you did for data type

influxdb_influxql_parser/src/expression/conditional.rs
```diff
@@ -0,0 +1,395 @@
...
+        let (_, got) = conditional_expression("5 AND 6 OR 7").unwrap();
+        assert_eq!(got, *cond_op!(cond_op!(5, And, 6), Or, 7));
+
+        let (_, got) = conditional_expression("5 > 3 OR 6 = 7 AND 7 != 1").unwrap();
+        assert_eq!(
+            got,
+            *cond_op!(
+                cond_op!(5, Gt, 3),
+                Or,
+                cond_op!(cond_op!(6, Eq, 7), And, cond_op!(7, NotEq, 1))
```
NGA-TRAN, line 351: Nice

influxdb_influxql_parser/src/select.rs
```diff
@@ -0,0 +1,1039 @@
...
+    /// A value to specify an offset to start retrieving series.
+    pub series_offset: Option<u64>,
+
+    /// The timezone for the query, specified as [`tz('<time zone>')`][time_zone_clause].
+    ///
+    /// [time_zone_clause]: https://docs.influxdata.com/influxdb/v1.8/query_language/explore-data/#the-time-zone-clause
+    pub timezone: Option<String>,
+}
+
+impl Display for SelectStatement {
```
NGA-TRAN, line 66: This statement looks good and no need to change anything. I am just wondering whether we can extend sqlx parser somehow or we must go with this new SelectStatement 🤔 
@alamb may knows this better

NGA-TRAN (approved):
Very nice

influxdb_influxql_parser/src/common.rs
```diff
@@ -185,6 +180,64 @@ pub fn order_by_clause(i: &str) -> ParseResult<&str, OrderByClause> {
...
+}
+
+/// `OneOrMore` is a container for representing a minimum of one `T`.
+///
+/// `OneOrMore` provides a default implementation of [`fmt::Display`],
+/// which writes `first`, followed by `rest` separated with commas.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct OneOrMore<T: Display + Parser> {
+    pub first: T,
+    pub rest: Option<Vec<T>>,
```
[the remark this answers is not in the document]
carols10cents: Ahhh I meant to click "start a review". I'm still reviewing this PR.

influxdb_influxql_parser/src/literal.rs
```diff
@@ -244,8 +289,8 @@ fn duration(i: &str) -> ParseResult<&str, Duration> {
 
 /// Parse an InfluxQL literal, except a [`Regex`].
 ///
-/// See [`literal_regex`] for parsing literal regular expressions.
-pub fn literal(i: &str) -> ParseResult<&str, Literal> {
+/// Use [`literal`] for parsing any literals, including regular expressions.
+pub fn literal_no_regex(i: &str) -> ParseResult<&str, Literal> {
```
carols10cents, line 293: I'm confused, this comment says "including regular expressions" but the function is named `literal_no_regex`?

influxdb_influxql_parser/src/select.rs
```diff
@@ -0,0 +1,1039 @@
...
+                expect("invalid TIME call, expected duration", duration),
+            ),
+            |v| Expr::Literal(Literal::Duration(v)),
+        )(i)
+    }
+}
+
+/// Used to parse the offset argument of the TIME function
+///
+/// The offset argument accepts either a duration
```
carols10cents, line 232: This comment seems unfinished... either a duration or...?

influxdb_influxql_parser/src/select.rs
```diff
@@ -0,0 +1,1039 @@
...
+    Wildcard,
+}
+
+impl Display for Dimension {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Time {
+                interval,
+                offset: Some(offset),
+            } => write!(f, "TIME({}, {})", interval, offset)?,
```
carols10cents, line 283: Same comment as other Display impls

influxdb_influxql_parser/src/select.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+    /// Empty aggregate windows will be filled with a value that is the linear interpolation of
+    /// the prior two non-null window values and is specified as `fill(linear)`
+    Linear,
+}
+
+impl Display for FillOption {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Null => f.write_str("NULL")?,
```
carols10cents, line 385: Same comment as other Display impls

influxdb_influxql_parser/src/select.rs
```diff
@@ -0,0 +1,1039 @@
...
+
+        // Parse a call with a wildcard
+        let (_, got) = Field::parse("COUNT(*)").unwrap();
+        assert_eq!(
+            got,
+            Field {
+                expr: call!("COUNT", wildcard!()),
+                alias: None,
+            }
+        );
```
carols10cents, line 771: These two cases look the same, should one be removed or changed?

carols10cents (changes requested):
Overall, looks great, just a few cleanups/improvements!

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+        /// ```text
+        /// SELECT foo::field, host::tag, usage_idle::integer, idle::boolean FROM cpu
+        /// ```
+        ///
+        /// Specifies the following:
+        ///
+        /// * `foo::field` will return a field of any data type named `foo`
+        /// * `host::tag` will return a tag named `host`
+        /// * `usage_idle::integer` will return either a float or integer field named `usage_idle`,
+        ///   and casting it to an `integer`
```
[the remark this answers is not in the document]
stuartcarnie: As you presumed, `float` and `integer` conversions are supported in both directions, per [our docs](https://docs.influxdata.com/influxdb/v1.8/query_language/explore-data/#cast-operations):

> The :: syntax allows users to perform basic cast operations in queries. Currently, InfluxDB supports casting [field values](https://docs.influxdata.com/influxdb/v1.8/concepts/glossary/#field-value) from integers to floats or from floats to integers.

influxdb_influxql_parser/src/expression/arithmetic.rs
stuartcarnie: For the expression `5--(3|2)`, `--` will be treated as a comment, once I resolve #5663, which is the same as Go implementation.

For `2|5%3`, `|` and `%` have the same precedence and are left associative, so the result will be the same as `(2|5)%3`

influxdb_influxql_parser/src/common.rs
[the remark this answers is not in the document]
stuartcarnie: I was originally thinking that would avoid an allocation of `Vec<T>` for the many cases where it will have a single element. However, now looking at the internals, it appears an empty Vec does not allocate on the heap, so I'll switch this out:

https://github.com/rust-lang/rust/blob/b30c88623cf1f8f3e67b189fb2c5df5a0d89aaed/library/alloc/src/raw_vec.rs#L122-L123

You are correct that I was conveying that this type must be one or more elements. I'll refactor it to use your suggested approach and hide the `contents` behind appropriate APIs

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+            }
+            Self::BindParameter(v) => write!(f, "{}", v)?,
+            Self::Literal(v) => write!(f, "{}", v)?,
+            Self::UnaryOp(op, e) => write!(f, "{}{}", op, e)?,
+            Self::Binary { lhs, op, rhs } => write!(f, "{} {} {}", lhs, op, rhs)?,
+            Self::Nested(e) => write!(f, "({})", e)?,
+            Self::Call { name, args } => {
+                write!(f, "{}(", name)?;
+                match args {
+                    Some(args) if !args.is_empty() => {
```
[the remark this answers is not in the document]
stuartcarnie: This too was me thinking I was avoiding the allocation of an empty `Vec<T>`, figuring it started with some default capacity on the heap, but it turns out it does not, so I'll also remove `Option<Vec<Expr>>` here too.

Thanks!

influxdb_influxql_parser/src/common.rs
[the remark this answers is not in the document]
stuartcarnie: Addressed and greatly simplified in de62d18fac99f4e1ef77f52cee6af5fb7c0527d5.

Thanks for the feedback, @carols10cents 🙇🏻

influxdb_influxql_parser/src/common.rs
```diff
@@ -185,6 +180,64 @@ pub fn order_by_clause(i: &str) -> ParseResult<&str, OrderByClause> {
...
+}
+
+impl<T: Display + Parser> OneOrMore<T> {
+    /// Parse a list of one or more `T`, separated by commas.
+    ///
+    /// Returns an error using `msg` if `separated_list1` fails to parse any elements.
+    pub fn separated_list1<'a>(
```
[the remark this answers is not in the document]
stuartcarnie: Once I addressed your [earlier comment](https://github.com/influxdata/influxdb_iox/pull/5692#discussion_r982527476)[^1], this was pretty simple:

https://github.com/influxdata/influxdb_iox/blob/de62d18fac99f4e1ef77f52cee6af5fb7c0527d5/influxdb_influxql_parser/src/common.rs#L273-L282

[^1]: Addressed in de62d18fac99f4e1ef77f52cee6af5fb7c0527d5

influxdb_influxql_parser/src/expression/arithmetic.rs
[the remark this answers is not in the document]
stuartcarnie: Addressed in 4fbf03683c9dae65f245671a39886b338f18a10a

influxdb_influxql_parser/src/expression/conditional.rs
```diff
@@ -0,0 +1,395 @@
...
+pub fn is_valid_now_call(expr: &Expr) -> bool {
+    match expr {
+        Expr::Call { name, args: None } => name.to_lowercase() == "now",
+        Expr::Call {
+            name,
+            args: Some(args),
+        } => name.to_lowercase() == "now" && args.is_empty(),
+        _ => false,
+    }
+}
```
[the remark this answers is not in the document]
stuartcarnie: Addressed in 4fbf03683c9dae65f245671a39886b338f18a10a

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Float => f.write_str("float")?,
+            Self::Integer => f.write_str("integer")?,
+            Self::String => f.write_str("string")?,
+            Self::Boolean => f.write_str("boolean")?,
+            Self::Tag => f.write_str("tag")?,
+            Self::Field => f.write_str("field")?,
+        }
+        Ok(())
```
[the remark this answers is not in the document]
stuartcarnie: Ooh, nice catch, thank you!

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+    #[test]
+    fn test_var_ref() {
+        let (_, got) = var_ref("foo").unwrap();
+        assert_eq!(got, var_ref!("foo"));
+
+        // with cast operator
+        let (_, got) = var_ref("foo::tag").unwrap();
+        assert_eq!(got, var_ref!("foo", Tag));
+
+        // Fallible cases
```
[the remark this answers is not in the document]
stuartcarnie: Great catch - test case added for invalid data type.

influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -0,0 +1,644 @@
...
+            format!(
+                "{}",
+                Expr::VarRef {
+                    name: "foo".into(),
+                    data_type: Some(VarRefDataType::Field)
+                }
+            ),
+            "foo::field"
+        );
+    }
```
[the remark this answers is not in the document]
stuartcarnie: Done!

influxdb_influxql_parser/src/literal.rs
stuartcarnie: Oops, missed the comment when I removed regular expression parsing from this combinator. Great catch, thanks!

influxdb_influxql_parser/src/select.rs
stuartcarnie:
```
/// The offset argument accepts either a duration, datetime-like string or `now`.
```

influxdb_influxql_parser/src/select.rs
stuartcarnie: "are you really sure you can parse a call with a wildcard" 😆 

I've removed the duplicate – good catch!

influxdb_influxql_parser/src/expression/arithmetic.rs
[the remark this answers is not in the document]
stuartcarnie: I've reviewed all `Display` implementations in the crate and simplified where possible.

influxdb_influxql_parser/src/expression/conditional.rs
stuartcarnie: Unfortunately the documentation doesn't really have a separate link for conditional operators. It has a binary_op description here:

https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#expressions

But that may be confusing, as it includes arithmetic operators also.

influxdb_influxql_parser/src/select.rs
stuartcarnie: We did consider amending the sqlparser-rs crate, but determined InfluxQL would have been too invasive for an existing SQL parser and we also wanted to produce an AST that more closely modelled the existing Go implementation.

We only intend to use the `SelectStatement` for InfluxQL support.

influxdb_influxql_parser/src/expression/arithmetic.rs
stuartcarnie: I had also added a test case that will be uncommented when #5663 is resolved:

https://github.com/influxdata/influxdb_iox/blob/0753df8ae1cd543a1b08439e5c35c37db96d421a/influxdb_influxql_parser/src/expression/arithmetic.rs#L475-L476

stuartcarnie:
@carols10cents thank you for the thorough review, it is very much appreciated. I have implemented all your feedback in this PR.

@NGA-TRAN thank you too for your review – I've replied to your questions and linked any issues

carols10cents (approved):
The updates look great, thank you!!! 🎉

PR #5695, opened 2022-09-20
refactor: retry querier->ingester requests

crepererum:
Esp. for InfluxRPC requests that scan multiple tables, it may be that one ingester requests fails. We shall retry that request instead of failing the entire query.


backoff/src/lib.rs
```diff
@@ -48,13 +48,16 @@ impl Default for BackoffConfig {
...
 #[derive(Debug, Snafu, PartialEq, Eq)]
 #[allow(missing_copy_implementations, missing_docs)]
-pub enum BackoffError {
-    #[snafu(display("Deadline exceeded: {deadline:?}"))]
-    DeadlineExceeded { deadline: Duration },
+pub enum BackoffError<E>
+where
+    E: std::error::Error + 'static,
+{
+    #[snafu(display("Deadline {deadline:?} exceeded: {source}"))]
```
alamb, line 55:
```suggestion
    #[snafu(display("Retry did not exceed within {deadline:?}: {source}"))]
```

backoff/src/lib.rs
```diff
@@ -109,7 +112,7 @@ impl Backoff {
     }
 
     /// Returns the next backoff duration to wait for
```
alamb, line 114:
```suggestion
    /// Returns the next backoff duration to wait for, if any
```

backoff/src/lib.rs
```diff
@@ -139,11 +140,11 @@ impl Backoff {
         &mut self,
```
alamb, line 140: I think the `TODO` comment for this function needs to be updated (as I believe it will return an error now)

querier/src/ingester/mod.rs
```diff
@@ -333,6 +335,7 @@ impl IngesterConnectionImpl {
             shard_to_ingesters,
             Arc::new(FlightClientImpl::new()),
             catalog_cache,
+            BackoffConfig::default(),
```
alamb, line 338: https://github.com/influxdata/influxdb_iox/blob/main/backoff/src/lib.rs#L41

The default backoff of 500 seconds seems too long because even if  the querier gets an answer in that timeframe the client making the query request will have timed out

Maybe we could make it 30 seconds or something closer to the default gRPC timeout

querier/src/ingester/mod.rs
```diff
@@ -698,25 +703,38 @@ impl IngesterConnection for IngesterConnectionImpl {
...
             async move {
-                let res = execute(request.clone(), measure_me.span_recorder()).await;
+                let span_recorder = measure_me
+                    .span_recorder()
+                    .child("ingester request (retry block)");
+
+                let res = Backoff::new(&backoff_config)
+                    .retry_all_errors("ingester request", move || {
+                        let request = request.clone();
+                        let span_recorder = span_recorder.child("foo");
```
alamb, line 733: Is the name "foo" intentional here? Perhaps it should be "ingester request" or something more descriptive?

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -251,20 +253,22 @@ async fn ingester_panic() {
...
                     )
-                    .await
-                    .unwrap_err();
-                    if let influxdb_iox_client::flight::Error::GrpcError(status) = err {
-                        assert_eq!(status.code(), tonic::Code::Internal);
-                    } else {
-                        panic!("wrong error type");
-                    }
+                    .await;
+                    let expected = [
```
alamb, line 264: 👍  Nice

alamb: I wonder if there is some way to verify that the panic actually happened on the ingester (otherwise for example we may break the panic plumbing by accident in the future but not see any test failures)

alamb (approved):

querier/src/ingester/mod.rs
crepererum: The MAX backoff default is 500s, not the actual first backoff. It's growing exponentially though. I think the max doesn't matter much here, because if it takes too long, than the incoming gRPC request (Flight or InfluxRPC) will be cancelled anyways.

querier/src/ingester/mod.rs
crepererum: fixed

backoff/src/lib.rs
crepererum: done

influxdb_iox/tests/end_to_end_cases/querier.rs
crepererum: There was an indirect test that searched through the querier logs for `Failed to perform ingester query`. However I've just added (last commit) another assertion that goes the the ingester logs and searches for a panic message.

querier/src/ingester/mod.rs
alamb: See also https://github.com/influxdata/influxdb_iox/pull/5708

PR #5696, opened 2022-09-20
chore: temporarily turn off cold compaction to investigate an oom

NGA-TRAN:
First step to identify the culprit of the OOM in https://github.com/influxdata/conductor/issues/1167


lukebond (approved):
not sure if this is a common thing we might need, if so might be nice to have another way to do it, e.g. an envvar or cmdline arg. but lgtm for debugging!

PR #5699, opened 2022-09-20
chore: Improve debug logging when parquet files are created

alamb:
re https://github.com/influxdata/conductor/issues/1075

# Rationale
I am trying to to track down where large amounts of memory is being buffered in the ingester and I plan to use debug logging to do so

# Changes
1. Improve logging (will comment inline(


parquet_file/src/serialize.rs
```diff
@@ -111,24 +111,36 @@ where
...
-    if meta.num_rows == 0 {
+    let writer_meta = writer.close().map_err(CodecError::from)?;
+    if writer_meta.num_rows == 0 {
         // throw warning if all input batches are empty
         warn!("parquet serialisation encoded 0 rows");
         return Err(CodecError::NoRows);
     }
 
-    Ok(meta)
+    debug!(num_batches,
```
alamb, line 135: One theory I have is that the datafusion plan being run produces a single giant record batch for encoding (aka buffers the entire input) and I expect this debug logging to confirm/deny that hypothesis -- specifically `num_batches` and `num_rows`

parquet_file/src/storage.rs
```diff
@@ -152,6 +152,13 @@ impl ParquetStorage {
         let file_size = data.len();
         let data = Bytes::from(data);
 
+        debug!(
```
alamb, line 155: This is the corresponding log message I can use to figure out how large the parquet file was

parquet_file/src/serialize.rs
```diff
@@ -153,7 +165,7 @@ where
     let meta = to_parquet(batches, meta, &mut bytes).await?;
     bytes.shrink_to_fit();
 
-    debug!(?partition_id, ?meta, "generated parquet file metadata");
+    trace!(?partition_id, ?meta, "generated parquet file metadata");
```
alamb, line 168: If you are not familiar with that type of message this line generates it is terrible (I don't want to post one from our production environment here)

NGA-TRAN: Yeah, I see this a lot too.

NGA-TRAN (approved):

alamb:
I had a last minute thought here too that I wanted to know how long it took to write the parquet file. Will adjust in a moment

PR #5713, opened 2022-09-21
refactor: use cheaper hash keys for projected schemas

crepererum:
Instead of using the column names to address projected schemas, let's use the column IDs.


querier/src/cache/projected_schema.rs
```diff
@@ -17,27 +17,27 @@ use cache_system::{
...
+use super::{namespace::CachedTable, ram::RamSize};
 
 const CACHE_ID: &str = "projected_schema";
 
 /// Cache key.
 #[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
 struct CacheKey {
     table_id: TableId,
-    projection: Vec<String>,
+    projection: Vec<ColumnId>,
```
alamb, line 33: Not only is this more efficient, it is now more strongly typed too 👍

alamb (approved):
Moar `Arc` more better 👍

PR #5715, opened 2022-09-21
feat: add minimum row_count per file in estiumating compacting memory…

NGA-TRAN:
… budget and limit number files per compaction

Closes #


compactor/src/compact.rs
```diff
@@ -503,6 +516,12 @@ fn estimate_arrow_bytes_for_file(columns: &[ColumnTypeCount], row_count: i64) ->
 
     let average_cardinality = row_count / AVERAGE_ROW_COUNT_CARDINALITY_RATIO;
 
+    // Since there is a minimum number of rows per batch, we will use that nimum number of rows if the row_count is smaller
```
carols10cents, line 519: Typo:

```suggestion
    // Since there is a minimum number of rows per batch, we will use that minimum number of rows if the row_count is smaller
```

compactor/src/lib.rs
```diff
@@ -149,6 +153,22 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
                 FilterResult::NothingToCompact => {
                     debug!(?partition_id, compaction_type, "nothing to compact");
                 }
+                FilterResult::OverLimitFileNum { file_num } => {
+                    // We cannot compact this partition becasue it first set of overlapped files
```
carols10cents, line 157: Typo:

```suggestion
                    // We cannot compact this partition because its first set of overlapped files
```

compactor/src/lib.rs
```diff
@@ -149,6 +153,22 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
...
+                    // are over the limit of file num
+                    warn!(
+                        ?partition_id,
+                        ?table_id,
+                        compaction_type,
+                        file_num_limit = compactor.config.max_num_compacting_files,
+                        "skipped; over limit of number of files"
+                    );
```
carols10cents, line 165: Should this warn log include the `file_num` like the skipped reason does?

compactor/src/lib.rs
```diff
@@ -228,6 +242,21 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
     }
 }
 
+async fn record_skipped_compaction(
```
carols10cents, line 245: Nice function extraction! 🎉

compactor/src/parquet_file_filtering.rs
```diff
@@ -129,8 +139,20 @@ fn filter_parquet_files_inner(
         let estimated_file_bytes =
             ln_estimated_file_bytes + current_ln_plus_1_estimated_file_bytes.iter().sum::<u64>();
 
-        // Over budget
-        if total_estimated_budget + estimated_file_bytes > max_bytes {
+        // Over limit of num files
+        if files_to_return.len() + 1 /* L0 file */ + overlaps.len() > max_num_files {
```
carols10cents, line 143:
```suggestion
        if files_to_return.len() + 1 /* LN file */ + overlaps.len() > max_num_files {
```

clap_blocks/src/compactor.rs
```diff
@@ -140,6 +140,30 @@ macro_rules! gen_compactor_config {
...
+            )]
+            pub min_num_rows_allocated_per_record_batch_to_datafusion_plan: u64,
+
+            /// Max number of files to compact per partition
+            /// 
+            /// Due to limit in fan-in of datafusion plan, we need to limit the number of files to compact per partition.
+            #[clap(
+                long = "--compaction_max_num_compacting_files",
+                env = "INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES",
+                default_value = "20",
```
NGA-TRAN, line 163: @alamb and @tustvold: what do you think the hard limit number of files should be? is 20 reasonable or two small? 
Note, this is a hard limit in case the memory estimation does not work well. Or we can try with larger values when this works well

compactor/src/parquet_file_filtering.rs
```diff
@@ -129,8 +139,20 @@ fn filter_parquet_files_inner(
...
+            if files_to_return.is_empty() {
+                // Cannot compact this partition because its first set of overallaped files
```
carols10cents, line 145: Typo:

```suggestion
                // Cannot compact this partition because its first set of overlapped files
```

compactor/src/parquet_file_filtering.rs
```diff
@@ -432,15 +462,74 @@ mod tests {
...
+        );
+
+        assert_eq!(
+            filter_result,
+            FilterResult::OverLimitFileNum { file_num: 1 }
+        );
+    }
+
+    #[test]
+    fn file_num_litmit_1_return_over_file_limit() {
```
carols10cents, line 502: Typo:

```suggestion
    fn file_num_limit_1_return_over_file_limit() {
```

compactor/src/parquet_file_filtering.rs
```diff
@@ -432,15 +462,74 @@ mod tests {
...
+            1,
+            &files_metric,
+            &bytes_metric,
+        );
+
+        assert_eq!(
+            filter_result,
+            FilterResult::OverLimitFileNum { file_num: 2 }
+        );
+    }
```
carols10cents, line 531: I think this could use a test for a case that would have created a larger group of files except for the file num limit-- that is, to test the `break` in the `else` case here:

```
        if files_to_return.len() + 1 /* L0 file */ + overlaps.len() > max_num_files {
            if files_to_return.is_empty() {
                ...
            } else {
                // Only compact files that are under limit number of files
                break;
            }
        } 
```

The tests added here look like they're only testing the ` if files_to_return.is_empty() {` case.

carols10cents (changes requested):
Some thoughts!

compactor/src/parquet_file_filtering.rs
NGA-TRAN: Added

compactor/src/lib.rs
NGA-TRAN: Done

NGA-TRAN:
@carols10cents I have addressed all your comments

clap_blocks/src/compactor.rs
alamb: I think 20 is a good place to start

clap_blocks/src/compactor.rs
```diff
@@ -140,6 +140,30 @@ macro_rules! gen_compactor_config {
...
+                long = "--compaction-min-rows-allocated-per-record-batch-to-plan",
+                env = "INFLUXDB_IOX_COMPACTION_MIN_ROWS_PER_RECORD_BATCH_TO_PLAN",
+                default_value = "8192",
+                action
+            )]
+            pub min_num_rows_allocated_per_record_batch_to_datafusion_plan: u64,
+
+            /// Max number of files to compact per partition
+            /// 
+            /// Due to limit in fan-in of datafusion plan, we need to limit the number of files to compact per partition.
```
alamb, line 159:
```suggestion
            /// Due to limitations of the implementation of the compaction plans
            /// there is a hard maximum on the number of files that can be compacted
            /// at once. This avoids a wide fan-in multi-way merge in the DataFusion plan
```

clap_blocks/src/compactor.rs
```diff
@@ -140,6 +140,30 @@ macro_rules! gen_compactor_config {
...
             )]
             pub memory_budget_bytes: u64,
+
+            /// Minimum number of rows allocated for each record batch fed into DataFusion plan
+            /// 
+            /// We will use max(parquet_file's row_count, min_num_rows_allocated_per_record_batch_to_datafusion_plan)
+            /// to estimate number of rows allocated for each record batch fed into DataFusion plan.
+            /// 
+            #[clap(
+                long = "--compaction-min-rows-allocated-per-record-batch-to-plan",
```
alamb, line 150: I think this is fine, though as we discussed with @carols10cents  yesterday, I would personally try to make the cost model simpler and make this knob something like "COMPACTION_MEMORY_FACTOR" -- and then model the memory use in the compactor like:

```
memory_per_row * num_files * COMPACTION_MEMORY_FACTOR
```

Rather than going through the indirection of "number of rows"

compactor/src/compact.rs
```diff
@@ -503,6 +516,12 @@ fn estimate_arrow_bytes_for_file(columns: &[ColumnTypeCount], row_count: i64) ->
 
     let average_cardinality = row_count / AVERAGE_ROW_COUNT_CARDINALITY_RATIO;
 
+    // Since there is a minimum number of rows per batch, we will use that minimum number of rows if the row_count is smaller
```
alamb, line 519: I think @tustvold  and I are of the opinion that a better model of the memory use of the plan is a constant. 

Specifically, if the parquet file has 500,000 rows, this will likely be a large over estimate of the memory required for the plan.

I recommend using a constant like `row_count = 8000` or something here

compactor/src/handler.rs
```diff
@@ -127,6 +127,17 @@ pub struct CompactorConfig {
     /// The number of candidates compacted concurrently is also decided using this estimation and
     /// budget.
     pub memory_budget_bytes: u64,
+
+    /// Minimum number of rows allocated for each record batch fed into DataFusion plan
```
alamb, line 131: recommend same changes here

alamb (approved):
I didn't review the logic or tests in super detail as I assumed that @carols10cents did, but the basic idea looks great 

I think the current cost model will likely overly constrain the compactor from compacting large files (as in it will think those plans will use much more memory than they actually do) but perhaps the model can be refined as a follow on project

clap_blocks/src/compactor.rs
NGA-TRAN: @alamb We actually do something similar/same to what you stay above. Our formula for **each file** is

      `memory_per_row * max(the file's row_count, COMPACTION_MEMORY_FACTOR)`

- COMPACTION_MEMORY_FACTOR in our formula is this config param `min_num_rows_allocated_per_record_batch_to_datafusion_plan`.
- memory_per_row is computed based on number of columns and their types.
- Since we compute that for each file, we jus sum them and it is more accurate than `* num_files`.

compactor/src/compact.rs
NGA-TRAN: Since I have seen memory usage is way up more than I can imagine, let us go with this `max(row_count, 8192)`. I will do a lot more experience to adjust this value

clap_blocks/src/compactor.rs
carols10cents: We don't actually know if it's more accurate, and I think @tustvold said yesterday that it doesn't actually matter if there are more rows than `min_num_rows_allocated_per_record_batch_to_datafusion_plan`, so I think we *can* simplify this, but because it's implemented now, we may as well try it?

carols10cents: Also another point that @alamb @tustvold and I talked about yesterday is that the exact memory that DataFusion needs will likely change over time, but due to the work that needs to be done, it will always be linear in the number of files and linear in the number/types of columns. So those two factors are what our estimation needs to be linear in as well.

compactor/src/parquet_file_filtering.rs
```diff
@@ -530,6 +530,68 @@ mod tests {
...
+            ParquetFileBuilder::level_1()
+                .id(101)
+                .min_time(301)
+                .max_time(500)
+                .build(),
+            // overlaps in time with first and second L0s
+            ParquetFileBuilder::level_1()
+                .id(102)
+                .min_time(150)
+                .max_time(350)
```
carols10cents, line 568: This setup violates the assumption that level 1 files don't overlap each other, does that matter?

carols10cents (approved):
I think this is fine to merge now and see how it works.

compactor/src/parquet_file_filtering.rs
NGA-TRAN: Hah, let me fix it

clap_blocks/src/compactor.rs
NGA-TRAN: When this runs smoothly, I will make row_count always 8192

alamb: Sounds like a good plan

NGA-TRAN: Done in https://github.com/influxdata/influxdb_iox/pull/5733

Issue #1169, opened 2021-04-09
add database wide system table 


alamb:
__Use case:__
I think I saw the database on tools restart but can't find evidence of that in the logs. Adding a system table to show the current state of the overall database would help

There are also several database wide numbers / values that would be nice to expose in system tables:
* uptime
* total memory consumed e.g. via https://github.com/influxdata/influxdb_iox/pull/1152 from @tustvold  / @e-dard 

__Proposal:__
Add in a system table `system.database` with coulmns:
*  "uptime" 
* "total memory used" 
* writer id
* object store configuration details, perhaps
others?

__Alternatives considered:__
Remain in our unlight world of quasi-ignorance and  subserviance to `kubectl`

pauldix:
Sounds like a good addition to me

jacobmarble:
@alamb system tables are implemented now, yes?

alamb:
@jacobmarble  system tables are implemented, but they are per database (not per sever/process), so they can not contain per server information (like uptime)

However, since per server info like uptime, etc can most often be found by kubernetes or logs, I don't think this is particularly pressing feature to add

PR #5729, opened 2022-09-22
feat: Start of a tool to generate files to compact

carols10cents:
Connects to #1169.

Submitting this to start getting these pieces reviewed and make sure I'm headed in the right direction; this should not interfere with anything else.

## In this PR

These pieces are complete and can be tested:

- A new command, runnable with `cargo run -- compactor generate`. Help text and args available with `cargo run -- compactor generate --help`.
- If you specify any `--object-store` besides `file`, it will panic because this is expected to run locally and create files on your local file system. `--object-store file` requires that you specify a directory with `--data-dir` that will hold the files generated by this command.
- If you specify `--num-files` greater than the default of 1, in your `--data-dir`, you will get that number of `.txt` files each containing one line of line protocol (`--num-rows` currently has no effect and is in the unimplemented list below)
- If you specify a different `--num-cols` value than the default of 6, the generated line protocol will contain that number of columns. The first column will always be time, then for `--num-cols` 2-infinity it will add columns by cycling through the types [i64, f64, string, bool, tag] repeating. 

## Still unimplemented

These pieces will be coming in future PRs:

- Changing behavior when `--type` hot or cold is specified (however, those two values *are* currently validated as the only ones accepted for `--type` for this command)
- Actually generating files for `--num-partitions` specified
- Actually generating `--num-rows` specified
- Writing parquet files
- Adding entries to the catalog


carols10cents:
Whew. Ok, that was fun, figuring out my mistake causing the test failures (I cleared the whole env in the test context, which also cleared my `PATH`, which made the command not able to find `cargo`, but the error message wasn't clear)

Aaaanyway, this is now ready for your review @NGA-TRAN , in the meantime Jake and I will continue building on this!

influxdb_iox/src/commands/compactor/generate.rs
```diff
@@ -0,0 +1,346 @@
...
+pub struct Config {
+    #[clap(flatten)]
+    object_store_config: ObjectStoreConfig,
+
+    #[clap(flatten)]
+    catalog_dsn: CatalogDsnConfig,
+
+    /// The type of compaction to be done on the files. If `hot` is specified, the generated
+    /// files will have compaction level 0. If `cold` is specified, the generated files will
+    /// have compaction level 1 and will be marked that they were created at least 8 hours ago.
```
NGA-TRAN, line 23: To give us time between data generation and run the compaction, I think we can increase this at least 8 hours to at least 10 hours to ensure those generated files will be always picked up by the compactor. This makes me think, in the document for this, we need to mention this maximum 2 hours gap time between generation and compaction.

Question: Since we want to use this for large tests, do you think the time to generate the data will be a lot? We can also test it after the tool available and adjust as we go.

influxdb_iox/src/commands/compactor/generate.rs
```diff
@@ -0,0 +1,346 @@
...
+        .as_ref()
+        .expect("--data-dir is required and has already been checked")
+        .into();
+    let subdir = "compactor_data/line_protocol";
+    let lp_dir = root_dir.join(subdir);
+
+    let spec_location = format!("{subdir}/spec.toml");
+    let spec_in_root = root_dir.join(&spec_location);
+
+    for file_id in 0..config.num_files.get() {
```
NGA-TRAN, line 102: Silly question: is the variable after `--num-files` automatically named `num_files` if we do not specify?

NGA-TRAN (approved):
This is very new and interesting to me. A lot to learn from this PR.

influxdb_iox/src/commands/compactor/generate.rs
carols10cents: I'm happy to answer whatever question, but I'm not sure I understand? We *have* to specify `num_files` for the Rust struct field name? https://github.com/influxdata/influxdb_iox/pull/5729/files#diff-71a0a80fff96b77ea6c5559f22e077c19f6194106209ba203c379abb01944e92R51 Do you mean what if we don't specify `long = "--num-files"` for a field named `num_files`?

influxdb_iox/src/commands/compactor/generate.rs
carols10cents: > To give us time between data generation and run the compaction, I think we can increase this at least 8 hours to at least 10 hours to ensure those generated files will be always picked up by the compactor. 

Sure, I'm happy to do that, if that's what you'd like this tool to do. I think 8 hours and 1 min ago should be sufficient, though? Because the time ago will only get larger after the files are generated as time marches on :) Just curious about how you picked 10; I trust the compactor to pick up files even if they're marked 9 hours ago :) 

> This makes me think, in the document for this, we need to mention this maximum 2 hours gap time between generation and compaction.

What should the documentation say exactly? This sentence here makes it sound like the compactor can't compact files created by this tool if we ran this tool more than 2 hours ago, but I don't think that's the case, so I think I'm misunderstanding... what should the user of this tool be careful/aware of, exactly?

carols10cents:
> This is very new and interesting to me. A lot to learn from this PR.

Oh I had to remember how the data generator works at all :) Did you try running the command at all, and did it do what you expected it to do given the PR description?

influxdb_iox/src/commands/compactor/generate.rs
NGA-TRAN: Ah, silly me. Somehow I did not see the `num_files` definition. That was why I asked 😄

influxdb_iox/src/commands/compactor/generate.rs
NGA-TRAN: > Because the time ago will only get larger after the files are generated as time marches on

LOL. I though the opposite. Silly me. 8 hours is always fine then and no need to change anything.

This makes me think the L-0 case. We need them within 4 hours (or 24 hours). Which means we want to generate them within 30 minutes or so

NGA-TRAN:
>  Because the time ago will only get larger after the files are generated as time marches on

I wanted to try but when I pulled your branch, I got merge conflict with the current main. When I merged locally, it got a bit ugly. I will try after your main merge

influxdb_iox/src/commands/compactor/generate.rs
carols10cents: > This makes me think the L-0 case. We need them within 4 hours (or 24 hours). Which means we want to generate them within 30 minutes or so

👍🏻 I can do that, working on that for my next PR.

carols10cents:
> I wanted to try but when I pulled your branch, I got merge conflict with the current main. When I merged locally, it got a bit ugly. I will try after your main merge

Hm I'm not sure what git commands you're running, but you should be able to try this branch out without doing any merges? If you fetch the latest from GitHub by doing `git fetch origin` (or something else if your remote isn't called `origin`), then you can check out this branch by running `git checkout -t origin/cn/generate-parquet-to-compact`, which will create a new local branch named `cn/generate-parquet-to-compact` that tracks (what the `-t` stands for) the remote branch `origin/cn/generate-parquet-to-compact`, and doesn't do any merges.

NGA-TRAN:
I have tried it, it is very nice. The parameters  `--compaction-type`,  `--data-dir`,  `--num-cols`, and `num_files` work as you described. Each file includes a row (I know you are working on number of rows).

It seems if I run the command again on he same `--data-dir`, data  is appended to the existing files which will mix old with new data. I do not think we want that. We want to re-generate new data.

Here are a few of the commands I have tried:

```command
cargo run -- compactor generate --help
cargo run -- compactor generate --compaction-type cold --data-dir /tmp/iox2 --num-cols 3 --num-files 3 --num-partitions 1 --num-rows 100 --object-store file
INFLUXDB_IOX_CATALOG_DSN=postgres:postgres://$USER@localhost/iox_shared cargo run -- compactor generate --compaction-type hot --data-dir /tmp/iox --num-cols 10 --num-files 1 --num-partitions 1 --num-rows 20 --object-store file

ls /tmp/iox/compactor_data/line_protocol 
data_0.txt      spec.toml
```

carols10cents:
> It seems if I run the command again on he same `--data-dir`, data is appended to the existing files which will mix old with new data. I do not think we want that. We want to re-generate new data.

Great catch!! That's exactly the sort of thing I knew wouldn't be discovered without trying it, thank you :)

I've added a commit to clear out the whole directory every run, and added a warning about this to the command's help text.

PR #5732, opened 2022-09-23
refactor: avoid clones for `policy::Subscriber::set`

crepererum:
Policy subscribers basically never store `V` (the cached value), so we should not clone that one unconditionally. But even `K` (the cache key) is not stored in all cases. So let's pass a reference for both and let the policies decide if they wanna clone the data or not.

### DIFF
--- a/cache_system/src/backend/policy/lru.rs
+++ b/cache_system/src/backend/policy/lru.rs
@@ -635,12 +635,12 @@ where
 
     fn set(
         &mut self,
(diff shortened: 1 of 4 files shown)


alamb (approved):
seems fine to me though I am not sure how much we are saving by not `clone`ing the key/value

alamb:
I am going to merge this in so it doesn't bitrot while @crepererum  is on vacation

PR #5736, opened 2022-09-23
fix: a silly bug that did not capture file limit if a lot of L0 files and very few or non overlapped L1

NGA-TRAN:
Closes https://github.com/influxdata/conductor/issues/1178

Such a silly bug


compactor/src/parquet_file_filtering.rs
```diff
@@ -141,8 +141,12 @@ fn filter_parquet_files_inner(
             ln_estimated_file_bytes + current_ln_plus_1_estimated_file_bytes.iter().sum::<u64>();
 
         // Over limit of num files
-        if files_to_return.len() + 1 /* LN file */ + overlaps.len() > max_num_files {
-            if files_to_return.is_empty() {
+        // At this stage files_to_return only includes LN+1 files. To get both returning LN+1 and LN files,
+        // we need to consider both files_to_return and level_n_to_return
+        if files_to_return.len() + level_n_to_return.len() + 1 /* LN file */ + overlaps.len()
+            > max_num_files
```
NGA-TRAN, line 147: This is the bug that we miss to count level-n files. The limit won't work well if most of the files are L0 (because we do not count them). The limit will work well if there are a lot of overlapped level n+1

jacobmarble (approved):

PR #5734, opened 2022-09-23
feat: Implement `debug parquet_to_lp` command to convert parquet to line protocol

alamb:
Closes https://github.com/influxdata/conductor/issues/1175

Sorry this is large -- A substantial amount of it is a plumbing / test exercise. 

# Rationale
I want to locally reproduce some pathological cases we are seeing on our production system by ingesting the same data from production. 

However, today IOx only supports *ingesting* line protocol but there is no way to get the data back out as line protocol.

In the long term, it is likely we would want to offer users a way to get data out of IOx in the same form it went in (aka line protocol). While this is not my near term goal I will write the code so it could support this usecase without major rework. 

# Changes
1. Add a `parquet_to_line_protocol` crate with the core logic for reading parquet data and converting it to line protocol
1. Add a `influxdb_iox debug parquet_to_lp` command that cover
2.  (MAYBE FOLLOW ON) Nice to have -- implement a "LineProtocol" output format option for printing the results of queries as line protocol

# Example Runs
```shell
influxdb_iox debug parquet-to-lp /Users/alamb/.influxdb_iox//1/1/1/7/f1a1be58-bc3a-4c94-9c10-5ab634a4e7a2.parquet
net,host=MacBook-Pro-8.local,interface=awdl0 bytes_recv=112019i,bytes_sent=145642i,drop_in=0i,drop_out=0i,err_in=0i,err_out=0i,packets_recv=522i,packets_sent=798i 1658178310000000000
net,host=MacBook-Pro-8.local,interface=awdl0 bytes_recv=112019i,bytes_sent=145642i,drop_in=0i,drop_out=0i,err_in=0i,err_out=0i,packets_recv=522i,packets_sent=798i 1658178320000000000
net,host=MacBook-Pro-8.local,interface=bridge0 bytes_recv=0i,bytes_sent=0i,drop_in=0i,drop_out=0i,err_in=0i,err_out=0i,packets_recv=0i,packets_sent=0i 1658178310000000000
net,host=MacBook-Pro-8.local,interface=bridge0 bytes_recv=0i,bytes_sent=0i,drop_in=0i,drop_out=0i,err_in=0i,err_out=0i,packets_recv=0i,packets_sent=0i 1658178320000000000
net,host=MacBook-Pro-8.local,interface=en0 bytes_recv=40757412682i,bytes_sent=9745871936i,drop_in=0i,drop_out=14272i,err_in=0i,err_out=0i,packets_recv=56289964i,packets_sent=21883749i 1658178310000000000
[...]

PR #5739, opened 2022-09-26
feat: RFC improver CLI Error reporting

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/5734 (last commit)~

# Rationale
Inspired by @carols10cents 's comment in https://github.com/influxdata/influxdb_iox/pull/5734#pullrequestreview-1119028922 I tried to make error reporting better

# Changes
Using the Snafu [`iter_chain`](https://docs.rs/snafu/0.7.1/snafu/trait.ErrorCompat.html#method.iter_chain) I prototyped what it would look like to print out errors in this way

# Example output:

```shell
influxdb_iox debug parquet-to-lp influxdb_iox/src/main.rs

Error in parquet_to_lp subcommand: Error converting: Error inferring IOx schema: Parquet error: Parquet error: Invalid Parquet file. Corrupt footer
  Caused by: Error converting: Error inferring IOx schema: Parquet error: Parquet error: Invalid Parquet file. Corrupt footer
  Caused by: Error inferring IOx schema: Parquet error: Parquet error: Invalid Parquet file. Corrupt footer
  Caused by: Parquet error: Parquet error: Invalid Parquet file. Corrupt footer
```

While I do think this helps see some of the error structure better, it is still quite repetitive.

# Question / Musing
I could remove the redundancy by remove the `{source}` in the Error::Display "Error inferring IOx schema: {source}", but if I did so it would remove that information from other places the error is reported (e.g. in logs).

I wonder if it would be possible to implement the "Caused by" formatting automatically for the implementations of `Error::Display` so that I could simply do `println!("Error: {}", e)` and get a source chain? This would also allow the error to be logged as well

So for example, my ideal output would be:

```text
Error in parquet_to_lp subcommand: 
  Caused by: Error converting
  Caused by: Error inferring IOx schema
  Caused by: Parquet error: 
  Caused by: Invalid Parquet file
  Caused by: Corrupt footer
```


influxdb_iox/src/main.rs
```diff
@@ -194,6 +194,21 @@ enum Command {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use snafu::ErrorCompat;
+
+        write!(f, "{}", self.0)?;
+        for cause in ErrorCompat::iter_chain(&self.0).skip(1) {
+            write!(f, "  Caused by: {cause}")?;
+        }
+        Ok(())
+    }
+}
```
shepmaster, line 217: This is pretty much the same as [what I'm working on adding to SNAFU in the next release](https://github.com/shepmaster/snafu/commit/8730358de8f7bb5b506058485d05564f671d162f#diff-667bdcc3d21a418a8806ad64a3cb61f2ca3e778361aa79e07dfb70a709dd5a46R203-R228)

The only comment I have now is that you could probably avoid using `snafu::ErrorCompat` altogether here by using `ChainCompat` directly, something like `ChainCompat::new(&self.0).skip(1)`.

influxdb_iox/src/main.rs
```diff
@@ -194,6 +194,21 @@ enum Command {
...
 }
 
+/// Prints the error in a heirarchy to stderr
+struct FormattedError<E>(E);
+
+impl<E: snafu::ErrorCompat + std::error::Error + 'static> Display for FormattedError<E> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use snafu::ErrorCompat;
+
+        write!(f, "{}", self.0)?;
```
shepmaster: Why the choice of `write!` instead of `writeln!`?

shepmaster: 🙈

alamb: I think it is an oversight. I will fix

alamb: Sorry I got too eager asking for feedback

alamb:
Superceded by https://github.com/influxdata/influxdb_iox/pull/6050

PR #5742, opened 2022-09-26
feat: add columns in skipped_compactions catalog table

NGA-TRAN:
instead of adding num_files and memory budget into the reason text column, let us create different columns for them. We will be able to filter them easily if there are data in this table and we want to remove them


carols10cents (approved):
Looks great!

PR #5743, opened 2022-09-26
feat: Generate the requested num rows and compaction type

carols10cents:
Connects to https://github.com/influxdata/conductor/issues/1169.

Builds on https://github.com/influxdata/influxdb_iox/pull/5729.

## In this PR

- The `--num-rows` argument will now control how many rows end up in each line protocol file.
- When `--compaction-type hot` is specified (or if `--compaction-type` isn't specified, because hot is the default), the files will have data from about the last 30 minutes until now, and the data in each file will overlap its adjacent files' data by a minute on either side.
- When `--compaction-type cold` is specified, the files will have data from 24 hours ago to 8 hours ago, and the data in the files will NOT overlap their adjacent files' data.

## Still unimplemented

These pieces will be coming in future PRs:

- Actually generating files for `--num-partitions` specified
- Writing parquet files from these line protocol files
- Adding entries to the catalog

### DIFF
--- a/influxdb_iox/src/commands/compactor/generate.rs
+++ b/influxdb_iox/src/commands/compactor/generate.rs
@@ -23,8 +23,8 @@ pub struct Config {
     /// The type of compaction to be done on the files. If `hot` is specified, the generated
     /// files will have compaction level 0, will overlap with each other slightly, and will be
     /// marked that they were created within the last (approximately) 30 minutes. If `cold` is
-    /// specified, the generated files will have compaction level 1 and will be marked that they
-    /// were created at least 8 hours ago.
+    /// specified, the generated files will have compaction level 1, won't overlap with each other,
+    /// and will be marked that they were created between 8 and 24 hours ago.
     #[clap(
         arg_enum,
         value_parser,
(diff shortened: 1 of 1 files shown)


influxdb_iox/src/commands/compactor/generate.rs
```diff
@@ -273,10 +313,278 @@ bool = true"#
...
+            let num_files = 1;
+            let TimeValues {
+                sampling_interval_ns,
+                start_end_args,
+            } = TimeValues::new(COMPACTION_TYPE, num_rows, num_files);
+
+            assert_eq!(sampling_interval_ns, 1_921_921_921);
+            assert_eq!(
+                start_end_args,
+                vec![StartEndMinutesAgo { start: 32, end: 0 }]
```
NGA-TRAN, line 441: 👍

influxdb_iox/src/commands/compactor/generate.rs
```diff
@@ -273,10 +313,278 @@ bool = true"#
...
+                start_end_args,
+            } = TimeValues::new(COMPACTION_TYPE, num_rows, num_files);
+
+            assert_eq!(sampling_interval_ns, 720_000_000_000);
+            assert_eq!(
+                start_end_args,
+                vec![
+                    StartEndMinutesAgo { start: 34, end: 22 },
+                    StartEndMinutesAgo { start: 23, end: 11 },
+                    StartEndMinutesAgo { start: 12, end: 0 },
```
NGA-TRAN, line 460: 👍

influxdb_iox/src/commands/compactor/generate.rs
```diff
@@ -273,10 +313,278 @@ bool = true"#
...
+            assert_eq!(sampling_interval_ns, 19_200_000_000_000);
+            assert_eq!(
+                start_end_args,
+                vec![
+                    StartEndMinutesAgo {
+                        start: 1440,
+                        end: 1120
+                    },
+                    StartEndMinutesAgo {
+                        start: 1120,
```
NGA-TRAN, line 547: Question: Are these `start` and `end` times exclusive? If they are not, the generated data may have a chance of overlaps

NGA-TRAN (approved):
Except a question, everything else looks great. I have tired to generate a few different ones. They all look good

influxdb_iox/src/commands/compactor/generate.rs
carols10cents: Great catch! They're inclusive-- I'll fix.

PR #5745, opened 2022-09-27
refactor(ingester): limit visibility of internals

domodwyer:
**This PR has no functional changes**.

The only real code change is in de0be877b, where the `ShardData` initialisation is DRYed and internalised - this allows the `ShardData` (an internal data structure) to be non-pub.

The following commit 72e02ee9e capitalises on this reduced scope which cascades through the lower tiers of the buffer tree, marking child/related structures as also non-pub, which in turn highlighted various bits of dead code. I also got some drive-by non-pub :tada:

---

* refactor: internalise ShardData init (72e02ee9e)

      Move the initialisation of ShardData (an internal ingester data structure)
      into the ingester itself.

      Previously callers would initialise the ingester state, and pass it into the
      IngesterData constructor.

* refactor(ingester): limit visibility (de0be877b)

      Marks many internal data structures as non-pub.

      Many remain as they're used across tests / from multiple callers
      "peeking", but this limits the scope of false sharing in the future.

### DIFF
--- a/ingester/src/handler.rs
+++ b/ingester/src/handler.rs
@@ -32,8 +32,8 @@ use crate::{
     poison::PoisonCabinet,
     querier_handler::prepare_data_to_querier,
     stream_handler::{
-        sink_adaptor::IngestSinkAdaptor, sink_instrumentation::SinkInstrumentation,
-        PeriodicWatermarkFetcher, SequencedStreamHandler,
+        handler::SequencedStreamHandler, sink_adaptor::IngestSinkAdaptor,
+        sink_instrumentation::SinkInstrumentation, PeriodicWatermarkFetcher,
     },
 };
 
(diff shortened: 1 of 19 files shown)


ingester/src/handler.rs
```diff
@@ -135,18 +135,10 @@ impl IngestHandlerImpl {
...
-            shards.insert(
-                s.id,
-                ShardData::new(s.shard_index, s.id, Arc::clone(&metric_registry)),
-            );
-        }
         let data = Arc::new(IngesterData::new(
             object_store,
             catalog,
-            shards,
+            shard_states.clone().into_iter().map(|(idx, s)| (s.id, idx)),
```
alamb, line 141: FWIW I would like to say for the record that I think the change you made to `IngesterData::new` to use generics makes the code easier to read in this case 👍  (I wanted to say this as I also have highlighted cases when I think that is not true :) )

alamb (approved):
Looks like a very nice incremental change to me. Well done 👏

PR #5746, opened 2022-09-27
feat(catalog): most recent N partitions

domodwyer:
Discovering the most recently created partitions can be used as a proxy for partitions that are currently hot for writes - those that were just created, are likely being wrote to. 

Using a proxy / this query is very cheap:

```
                                                QUERY PLAN
----------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..79.61 rows=10 width=116)
   ->  Index Scan Backward using partition_pkey on partition  (cost=0.42..302094.18 rows=38150 width=116)
         Filter: (shard_id = ANY ('{0,1,2,3,4,5}'::bigint[]))
```

This can be used to identify which partitions to pre-load into an ingester at startup, ref #5638.

---

* feat(catalog): most recent N partitions (c3cd31a83)

      Adds a Partition::most_recent_n() method to the catalog interface, returning
      the N most recent partitions for a given set of shards.

      The most recently created partitions are likely to be currently "hot" for
      writes, and are cheap to list.

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -494,6 +494,9 @@ pub trait PartitionRepo: Send + Sync {
         partition_id: PartitionId,
         sequence_number: SequenceNumber,
     ) -> Result<()>;
+
+    /// Return the N most recently created partitions for the specified shards.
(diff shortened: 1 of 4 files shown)


NGA-TRAN (approved):
LGTM

Issue #5748, opened 2022-09-27
Improve `influxdb_iox write` to write using HTTP API and upload in parallel batches


alamb:
I now have a bunch of line protocol and I need to upload it quickly to IOx, ideally in smallish batches to mirror what the actual clients are doing.

In order to do so I would like to use the `influxdb_iox write` command, however it has a few issues:
1. Uses a different write path (via grpc rather than http)
2. Only sends one giant chunk
3. only sends a single upload at time

Thus, I need a way to load data much more quickly into IOx (and in smaller pieces) in order to generate my reproducer for OOMs I am seeing in https://github.com/influxdata/conductor/issues/1172

I realize I could script this but I would rather make a more general purpose tool.

Therefore I plan to do the following:

- [x] Change `influxdb_iox write` to use the HTTP (rather than the gRPC API)
- [ ] Add `--max-lines (defaults to 10000)` to limit the size of each individual upload
- [ ]  Add `--concurrent-uploads (defaults to $NCORES)` to send that many batches of line protocol in parallel

Bonus points for:
- [ ] `gz` support

alamb:
cc @domodwyer  as I remember you have mentioned several times a desire to remove the grpc write API

PR #939, opened 2021-03-08
feat: Add Service + CLI command to write line protocol to IOx

alamb:
Re https://github.com/influxdata/influxdb_iox/issues/942

# Rationale:
I want to be able to do the basic administration stuff from the IOx client directly (not a mix of curl and the CLI)

# Usage
```
./influxdb_iox database write --data=/path/to/line_protocol.lp
```

Example use:

```
cargo run -- writer set 43
Ok
cargo run -- database create foo
Ok
cargo run -- database write foo tests/fixtures/lineproto/metrics.lp
1000 Lines OK
```

# Changes
* Define and implement a Write gRPC service in IOx
* Define and implement a Write service client in `influxdb_iox_client`
* Implement the write CLI using those pieces
* tests for same

# Implementation Notes

After discussions with @tustvold, I propose adding a new gRPC write endpoint rather than using the existing http `/api/v2/write` endpoint) for the following reason:

1. The CLI currently only has the gRPC endpoint url (the http url is on a different port)
2. The influxdb2 write endpoint is in terms of org and bucket ‚ I need a command in terms of databases; We don't have an http write endpoint in terms of database (yet)

I ~cargo-culted from~ took a friendly look at the management service and tried to follow the same patterns

Note this mirrors the http endpoint pretty closely and does the line protocol parsing on the server side. It would be cool to make `ReplicatedWrite` structures directly on the client, but they have embedded sequence numbers which would we would need to reassign or something.

- [x] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #5749, opened 2022-09-27
feat: use /api/v2 upload for write command rather than grpc `write_service`

alamb:

# Rationale
I originally added the gRPC write API (https://github.com/influxdata/influxdb_iox/pull/939) so we could use the same HTTP endpoint for all client interactions with IOx OG. However, IOx NG now has multiple roles (ingester, querier, etc) so we have to interact with different endpoints anyways and thus my original rationale for a gRPC write api is no longer present.

The  fact that the line protocol write API is only used via the CLI makes the rest of the ingester/router more confusing

And there was no test coverage 

# Changes:
1. Change write command to use HTTP API rather than gRPC
2. Update docs
3. Add tests

Notes:
1. I did not remove the grpc write service in this PR -- I plan to do so in a follow on.
2. I plan to add support for smaller chunks and parallel upload in a follow on PR

Example output:
```shell

$ influxdb_iox --host=http://localhost:8080 write my_namespace test_fixtures/lineproto/air_and_water.lp 
1243 Bytes OK

$ influxdb_iox  query my_namespace 'show tables'
../target-iox/debug/influxdb_iox  query my_namespace 'show tables'
+---------------+--------------------+-----------------+------------+
| table_catalog | table_schema       | table_name      | table_type |
+---------------+--------------------+-----------------+------------+
| public        | iox                | air_temperature | BASE TABLE |
| public        | iox                | h2o_temperature | BASE TABLE |
| public        | system             | queries         | BASE TABLE |
| public        | information_schema | tables          | VIEW       |
| public        | information_schema | views           | VIEW       |
| public        | information_schema | columns         | VIEW       |
| public        | information_schema | df_settings     | VIEW       |
+---------------+--------------------+-----------------+------------+

### CODE REVIEW

File: influxdb_iox/Cargo.toml [hunk 1]
```diff
@@ -14,6 +14,7 @@ datafusion = { path = "../datafusion" }
 generated_types = { path = "../generated_types" }
 import = { path = "../import" }
 influxdb_iox_client = { path = "../influxdb_iox_client", features = ["flight", "format", "write_lp"] }
+influxdb2_client = { path = "../influxdb2_client" }
```
alamb, line 17: Chose to use the influxdb2_client rather than raw HTTP (so I can add chunking there in a few follow on PRs)

influxdb_iox/src/commands/write.rs
```diff
@@ -1,54 +1,154 @@
...
-    })?;
+    let mut file = File::open(file_name).context(ReadingFileSnafu { file_name })?;
 
     let mut lp_data = String::new();
     file.read_to_string(&mut lp_data)
-        .map_err(|e| Error::ReadingFile {
-            file_name: config.file_name.clone(),
-            source: e,
-        })?;
+        .context(ReadingFileSnafu { file_name })?;
```
alamb, line 47: here is the actual code change to use a new client. The rest of the PR is plumbing and tests

influxdb_iox/src/commands/write.rs
```diff
@@ -1,54 +1,154 @@
...
+        .await
+        .context(ClientSnafu)?;
 
-    println!("{} Lines OK", lines_written);
+    println!("{} Bytes OK", total_bytes);
 
     Ok(())
 }
+
+/// Splits up the strings into org_id and namespace
```
mkmik, line 70: do you mean org_id and bucket_id ?

mkmik (approved):
LGTM (minor typo in comment)

domodwyer (approved):
Changes look good, but given we know we are removing the gRPC API it might be nicer to switch the IOx client to use HTTP instead and continue using the IOx client here - it is a prerequisite to unblock multiple changes I have planned, but I understand you are just trying to get a thing working!

alamb:
> Changes look good, but given we know we are removing the gRPC API it might be nicer to switch the IOx client to use HTTP instead and continue using the IOx client here - it is a prerequisite to unblock multiple changes I have planned, but I understand you are just trying to get a thing working!

I will do this (add a write command to the IOx client) as a follow on PR 👍 It makes sense

influxdb_iox/src/commands/write.rs
alamb: yes; thanks

alamb:
```suggestion
/// Splits up the strings into org_id and bucket_id
```

alamb:
> Changes look good, but given we know we are removing the gRPC API it might be nicer to switch the IOx client to use HTTP instead and continue using the IOx client here 

@domodwyer  done in  https://github.com/influxdata/influxdb_iox/pull/5756

PR #5750, opened 2022-09-27
refactor: Remove protobuf based write service

alamb:
draft as it 
~builds on https://github.com/influxdata/influxdb_iox/pull/5749~
builds on https://github.com/influxdata/influxdb_iox/pull/5756#

closes https://github.com/influxdata/influxdb_iox/issues/4866

# Rationale
1. this service is only used in the CLI and not in any production setting. While we have plans to offer a better write api, keeping around one that is half baked and not supported just makes working with the rest of the code more difficult
2. It exposes some internal details (mutable batch) that makes changing the router and ingester harder,

# Changes
- [x] Remove `WriteService` service

cc @jacobmarble and @tustvold who worked on this originally. Not sure if we still want to keep this code around


jacobmarble:
> cc @jacobmarble and @tustvold who worked on this originally. Not sure if we still want to keep this code around

We do not want to keep this code around.

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
```diff
@@ -118,9 +118,12 @@ message Column {
...
 
-service WriteService {
-    rpc Write (WriteRequest) returns (WriteResponse);
-}
+// Note there used to be a service that would load this internal protobuf format.
+// See https://github.com/influxdata/influxdb_iox/pull/5750 for rationale of why it was removed
+
+// service WriteService {
+//     rpc Write (WriteRequest) returns (WriteResponse);
+// }
```
domodwyer, line 126: I think this :ghost: might be unnecessary - `git log`!

domodwyer (approved):
Fantastic - this will unblock a bunch of improvements for me :heart:

Thank you for putting in the effort - you could have left it after #5749 and called your task done, but this has cleaned up the code and pushed me forwards - teamwork!

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
alamb: I thought the comment about why there was not a service was the most valuable to try and inform our future selves of the history and thus avoid repeating the past.

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
```diff
@@ -118,9 +118,12 @@ message Column {
     bytes null_mask = 4;
 }
 
-service WriteService {
-    rpc Write (WriteRequest) returns (WriteResponse);
-}
+// Note there used to be a service that would load this internal protobuf format.
+// See https://github.com/influxdata/influxdb_iox/pull/5750 for rationale of why it was removed
```
alamb, line 122:
```suggestion
// See https://github.com/influxdata/influxdb_iox/pull/5750 and 
// https://github.com/influxdata/influxdb_iox/issues/4866
// for rationale of why it was removed
```

PR #5754, opened 2022-09-28
fix(write_buffer): spurious watermark < read offset panic

domodwyer:
Fixes a panic:

```
thread 'tokio-runtime-workerthread 'tokio-runtime-worker' panicked at '' panicked at 'Shard Index ShardIndex(23) stream for topic iox-shared has a high watermark BEFORE the sequence number we want. This is either a bug (see https://github.com/influxdata/rskafka/issues/147 for example) or means that someone re-created the shard and data is lost. In both cases, it's better to panic than to try something clever.Shard Index ShardIndex(16) stream for topic iox-shared has a high watermark BEFORE the sequence number we want. This is either a bug (see https://github.com/influxdata/rskafka/issues/147 for example) or means that someone re-created the shard and data is lost. In both cases, it's better to panic than to try something clever.', ', /influxdb_iox/ingester/src/stream_handler/handler.rs/influxdb_iox/ingester/src/stream_handler/handler.rs::320:32021
```

This is caused by the response not having a valid high watermark value (note `high_watermark: -1`):

```
WriteBufferError(SequenceNumberAfterWatermark): Server error OffsetOutOfRange with message "n/a", request: Fetch { topic_name: "iox-shared", partition_id: 23, offset: 1 }, response: Some(PartitionFetchState { high_watermark: -1, last_stable_offset: Some(-1) }), virtual: false
```

Which in turn was incorrectly mapped to a `SequenceNumberAfterWatermark` error:

https://github.com/influxdata/influxdb_iox/blob/5f2f735c7ef342b8af1a9248a091bf81cb3e6265/write_buffer/src/kafka/mod.rs#L297-L299

This now maps to an unknown error as we cannot definitely map it to a `SequenceNumberNoLongerExists` error state.

---

* fix: spurious watermark < read offset panic (5f2f735c7)

      In staging we observed an ingester panic due to the write buffer stream 
      yielding an WriteBufferErrorKind::SequenceNumberAfterWatermark, suggesting the
      ingester was attempting to read from an offset that exceeds the current max
      write offset in Kafka (high watermark offset).

      This turned out not to be the case - the partition had a single write at 
      offset 2, and the ingester was attempting to seek to offset 1. The first read
      would fail (offset 1 does not exist) and the error handling did not account
      for the high watermark not being correctly set (-1 in the response).

      I have no idea why rskafka returns this watermark / doesn't retry / etc but
      this change will allow the ingesters to recover.

### DIFF
--- a/write_buffer/src/kafka/mod.rs
+++ b/write_buffer/src/kafka/mod.rs
@@ -280,6 +280,14 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
                     let kind = match e {
                         RSKafkaError::ServerError {
                             protocol_error: ProtocolError::OffsetOutOfRange,
+                            response:
+                                Some(ServerErrorResponse::PartitionFetchState {
+                                    high_watermark, ..
+                                }),
+                            ..
+                        } if high_watermark < 0 => WriteBufferErrorKind::Unknown,
+                        RSKafkaError::ServerError {
+                            protocol_error: ProtocolError::OffsetOutOfRange,
                             request: RequestContext::Fetch { offset, .. },
                             response:
                                 Some(ServerErrorResponse::PartitionFetchState {
--- a/write_buffer/src/kafka/mod.rs
+++ b/write_buffer/src/kafka/mod.rs
@@ -348,6 +348,15 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
 
     async fn seek(&mut self, sequence_number: SequenceNumber) -> Result<(), WriteBufferError> {
         let offset = sequence_number.get();
+        let current = self.partition_client.get_offset(OffsetAt::Latest).await?;
+        if offset > current {
+            return Err(WriteBufferError::sequence_number_after_watermark(format!(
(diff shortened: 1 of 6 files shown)


domodwyer:
Huh, the seeking-past-data test fails - perhaps this functionality never worked at all...

write_buffer/src/kafka/mod.rs
```diff
@@ -278,6 +278,14 @@ impl WriteBufferStreamHandler for RSKafkaStreamHandler {
...
                     terminated.store(true, Ordering::SeqCst);
                     let kind = match e {
+                        RSKafkaError::ServerError {
+                            protocol_error: ProtocolError::OffsetOutOfRange,
+                            response:
+                                Some(ServerErrorResponse::PartitionFetchState {
+                                    high_watermark, ..
+                                }),
+                            ..
+                        } if high_watermark < 0 => WriteBufferErrorKind::Unknown,
```
NGA-TRAN, line 288: 🎉

NGA-TRAN (approved):

mkmik (approved):

PR #5756, opened 2022-09-28
refactor: Change influxdb_iox client to use http rather than grpc for write

alamb:
# Rationale
re https://github.com/influxdata/influxdb_iox/issues/4866

I am planning to remove the `WriteService` that allows sending internal protobuf format to IOx (see https://github.com/influxdata/influxdb_iox/pull/5750)

Suggested by @domodwyer  here: https://github.com/influxdata/influxdb_iox/pull/5749#pullrequestreview-1123215463

This change will both make removing the write service easier and sets me up to make the upload faster (concurrent and in chunks).

Changes
- [x] Add a reqwest client in `Connection` and `ConnectionBuilder`
- [x] Move namespace parsing into `client_util`
- [x] Remove mutable batch feature from client
- [x] Change `influxdb_iox write` command to use the `influxdb_iox_client` (rather than the` influxdb2_client`)
- [x] Better / tested HTTP --> client error conversion
- [x] Add a basic write client test 
- [x] Test http headers setting

Follow on work (I don't have any specific plans to do this at the moment):
- [ ] Update end to end tests to use the http client rather than a raw `reqwest` (I would need to port over the write info to do so, and this PR is already pretty big)


client_util/src/connection.rs
```diff
@@ -1,14 +1,70 @@
...
 use http::{uri::InvalidUri, HeaderValue, Uri};
 use std::convert::TryInto;
 use std::time::Duration;
 use thiserror::Error;
 use tonic::transport::{Channel, Endpoint};
 use tower::make::MakeConnection;
 
-/// The connection type used for clients
-pub type Connection = SetRequestHeadersService<tonic::transport::Channel>;
+/// The connection type used for clients. Use [`Builder`] to create
```
alamb, line 11: This is the biggest API change in the whole PR -- previously `influxdb_iox_client::connection::Connection`  could be used both with the Client structs in `influxdb_iox_client` as well as directly with the tonic generated protobuf clients.

This was a bit of a leaky abstraction as I couldn't figure out how to make raw http requests using  `SetRequestHeadersService<tonic::transport::Channel>;`

Thus I made the Connection have both http and grpc clients and added methods to obtain them

I also (briefly) tried to implement `tonic::Service` for `Connection` so that we could continue to use `Connection` directly with tonic generated client,  but I couldn't figure out how to make the inner tonic connection traits work and went with this more explicit (and verbose) approach instead.

client_util/src/connection.rs
```diff
@@ -128,11 +184,23 @@ impl Builder {
...
+        let headers_map: HeaderMap = self.headers.iter().cloned().collect();
+
         // Compose channel with new tower middleware stack
-        tower::ServiceBuilder::new()
+        let grpc_connection = tower::ServiceBuilder::new()
             .layer(SetRequestHeadersLayer::new(self.headers))
-            .service(channel)
+            .service(channel);
+
+        let http_client = reqwest::Client::builder()
```
alamb, line 195: Here is the construction of the http client

client_util/src/namespace_translation.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn new(namespace: impl Into<String>, reason: impl Into<String>) -> Self {
+        Self::InvalidNamespace {
+            namespace: namespace.into(),
+            reason: reason.into(),
+        }
+    }
+}
+
+/// Splits up the namespace name into org_id and bucket_id
+pub fn split_namespace(namespace: &str) -> Result<(&str, &str), Error> {
```
alamb, line 23: This function was moved out of `influxdb_iox`

influxdb_iox/src/commands/write.rs
```diff
@@ -46,109 +44,14 @@ pub async fn command(url: String, config: Config) -> Result<()> {
...
 
-    debug!(url, total_bytes, org_id, bucket_id, "Writing data");
-
-    // IOx's v2 api doesn't validate auth tokens so pass an empty one
-    let auth_token = "";
-    let client = influxdb2_client::Client::new(url, auth_token);
-
-    client
-        .write_line_protocol(org_id, bucket_id, lp_data)
+    let total_bytes = client
```
alamb, line 49: Now we use the write client directly

influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -118,9 +118,9 @@ pub struct Client {
 
 impl Client {
     /// Creates a new client with the provided connection
-    pub fn new(channel: Connection) -> Self {
+    pub fn new(connection: Connection) -> Self {
```
alamb, line 121: drive by cleanups to make the parameter names consistent

influxdb_iox_client/src/client/write.rs
```diff
@@ -19,68 +22,73 @@ use crate::error::Error;
...
 
-        let database_batch = mutable_batch_pb::encode::encode_write(db_name.as_ref(), &write);
+        let (org_id, bucket_id) = split_namespace(namespace.as_ref()).map_err(Error::Namespace)?;
 
-        self.inner
-            .write(generated_types::WriteRequest {
-                database_batch: Some(database_batch),
-            })
-            .await?;
+        let response = self
```
alamb, line 69: Here is the implementation of write in `influxdb_iox_client`

alamb:
I think @tustvold and @domodwyer  may have some thoughts on API design so asking for their review

client_util/src/connection.rs
tustvold: Trying to share the connection sounds like a world of pain, this seems sensible to me

influxdb_iox_client/src/client/error.rs
```diff
@@ -139,3 +148,38 @@ impl From<tonic::Status> for Error {
         }
     }
 }
+
+/// Translates a reqwest response to an Error
+pub(crate) async fn translate_response(response: reqwest::Response) -> Result<(), Error> {
```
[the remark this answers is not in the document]
tustvold: My preference would be to keep the same subset of error variants and map the HTTP responses to them, I think the gRPC authors already did a good job identifying meaningful semantic categories of error and adding more is not especially helpful for downstreams.

influxdb_iox_client/src/client/error.rs
```diff
@@ -114,6 +114,15 @@ pub enum Error {
 
     #[error("An unexpected error occurred in the client library: {0}")]
     Client(StdError),
+
+    #[error("Error making http request: {0}")]
+    Http(reqwest::Error),
+
+    #[error("Error making http request, unexpected response: {0}")]
+    UnexpectedHttpStatus(reqwest::StatusCode),
```
tustvold, line 122: I think this overlaps with the Unknown variant?

influxdb_iox_client/src/client/error.rs
```diff
@@ -114,6 +114,15 @@ pub enum Error {
...
+
+    #[error("Namespace parsing error: {0}")]
+    Namespace(client_util::namespace_translation::Error),
```
tustvold, line 125: This overlaps with the InvalidArgument variant

influxdb_iox_client/src/client/error.rs
```diff
@@ -114,6 +114,15 @@ pub enum Error {
 
     #[error("An unexpected error occurred in the client library: {0}")]
     Client(StdError),
+
+    #[error("Error making http request: {0}")]
+    Http(reqwest::Error),
```
tustvold, line 119: I would much prefer we break out the actual error responses into semantic categories, than leaking the client implementation out

influxdb_iox_client/src/client/write.rs
```diff
@@ -19,68 +22,91 @@ use crate::error::Error;
...
+        let lp_data = lp_data.into();
+        let data_len = lp_data.len();
 
-        let meta = dml::DmlMeta::unsequenced(None);
-        let write = dml::DmlWrite::new(db_name.as_ref().to_string(), tables, None, meta);
-        let lines = write.tables().map(|(_, table)| table.rows()).sum();
+        let write_url = format!("{}api/v2/write", self.inner.uri());
 
-        let database_batch = mutable_batch_pb::encode::encode_write(db_name.as_ref(), &write);
+        let (org_id, bucket_id) = split_namespace(namespace.as_ref()).map_err(Error::Namespace)?;
```
tustvold, line 67: I suspect @pauldix may have some thoughts about this, effectively this will hard-code namespaces must consist of org and bucket

tustvold (commented):
I've had a brief pass through and this looks plausible, I have to confess to not fully understanding the motivation for moving away from gRPC, but I'm more than willing to defer to the people who will be owning this going forward. 

My only suggestion would be as I believe the intent is for this to be a public client (i.e. beyond IOx), to keep the errors as opaque error categories. Otherwise it becomes very difficult to avoid leaking implementation details, etc... Whilst the error model is modelled after that of gRPC, it is not inherently specific to gRPC and for the avoidance of bike-shedding I would suggest just sticking with it :sweat_smile:

influxdb_iox_client/src/client/write.rs
alamb: That is actually the current requirement (as the only supported way to get data into IOx is via the /v2/api/write HTTP API)

alamb:
> Otherwise it becomes very difficult to avoid leaking implementation details, etc... Whilst the error model is modelled after that of gRPC, it is not inherently specific to gRPC and for the avoidance of bike-shedding I would suggest just sticking with it 😅

Right, I tried to do this with most of the errors. See `translate_error`. There were some gRPC specific things in `Error`, such as:

```rust
    InvalidArgument(ServerError<FieldViolation>),
```

Where FieldViolation is an error that is part of the generated protobuf code, for example

tustvold:
> Where FieldViolation is an error that is part of the generated protobuf code, for example

It isn't actually generated, it is just a struct that identifies a field and what was wrong with it. It just lives in generated_types as there was no other obvious place to put it at the time

```
#[derive(Debug, Default, Clone, PartialEq)]
pub struct FieldViolation {
    pub field: String,
    pub description: String,
}
```

influxdb_iox_client/src/client/write.rs
tustvold: Perhaps things have changed, but the pbdata encoding didn't use to have org and bucket concepts, just a database name and table name, and would support writing to arbitrarily named databases as a result

alamb: I don't think things have changed

alamb:
> It isn't actually generated, it is just a struct that identifies 

Right, my point is that the current incarnation of `Error` has some `generated_types` bleed through -- it is certainly fixable, but I figured I had already shaved enough

yak_hair_overflow

influxdb_iox_client/src/client/write.rs
pauldix: For now it's fine to have org and bucket as a requirement because as @alamb mentions, the only way to write data in is via the v2 write API, which requires it.

In the future, there will be a v3 write API, but that will come with a new version of Line Protocol and have a different logical data model from v2. Instead of measurements, tags, and fields, there will likely just be tables and columns with some set of those columns specified as primary key and logical types specified for the columns (gauge, counter, rate, state, event). Ideally, it will also support native summary types (histograms, etc.).

tustvold:
I guess I was saying I don't see the issue with this, the whole CLI does. We can just use the same error abstraction and move on with more important things 😂

influxdb_iox_client/src/client/write.rs
domodwyer: We have [this ticket](https://github.com/influxdata/influxdb_iox/issues/4866) to add `namespace` as an alternative option to the org/bucket pair in the API (previously discussed).

I actually have a code that adds support for this in the router - I can PR it later if suitable :+1:

I think we have a service that makes use of writing directly to namespaces (influxrpc_compare?) when I asked around last time.

alamb: And to be clear, the client write code takes a `namespace` parameter, consistent with the other IOx APIs. Internally (to the client) it is split into org/bucket so it can use the v2 write API.

I can take care of updating influxrpc_compare, if needed to write to a namespace that has a `_` in it.

domodwyer (approved):
The error handling stuff has always been wonky for how I like my errors, but I am not sure it's worth the time/effort to refactor.

I left a comment suggesting a path to supporting arbitrary namespaces (rather than org/bucket pairs) but happy either way - thanks for cleaning this all up :heart:

alamb:
> I left a comment suggesting a path to supporting arbitrary namespaces (rather than org/bucket pairs) but happy either way - thanks for cleaning this all up ❤️

No problem. I will file a ticket for the v2 api extension to support arbitrary namespaces

influxdb_iox_client/src/client/error.rs
alamb: I will try and consolidate

influxdb_iox_client/src/client/write.rs
alamb: filed https://github.com/influxdata/influxdb_iox/issues/5765 for follow on work

influxdb_iox_client/src/client/error.rs
alamb: Done.

The messages now look like the following

```
Error connecting to http://localhost:8080: Connection error: transport error (error trying to connect: tcp connect error: Connection refused (os error 61)) 

and 

Client error: Unknown server error: (status 413 Payload Too Large): {"code":"request too large","message":"max request size (10485760 bytes) exceeded"}
```

influxdb_iox_client/src/client/error.rs
```diff
@@ -139,3 +139,68 @@ impl From<tonic::Status> for Error {
         }
     }
 }
+
```
alamb, line 142: FYI @tustvold  I have removed all new `Error` variants from this PR and used the existing ones

PR #5766, opened 2022-09-29
ci: add dbg!() lint

domodwyer:
I left a `dbg!()` macro in the changes in #5764 - a5e5dfda6fb1c1af3873cba830f9ce7aff1594c9 removes it.

I thought we linted for this, but it was not in all crates, so I added it. `dbg!()` should not be in prod code (it is not logfmt and will mess with the log parsers, as well as likely being a unintended addition to the codebase)

---

* refactor: remove dbg! (a5e5dfda6)

      Left over from debugging.

* style: add no todo!() or dbg!() lints (cd4087e00)

      Some crates had theme, some not - lets be consistent and have the compiler
      spot dbg!() and todo!() macro calls - they should never be in prod code!

* refactor(read_buffer): todo!() -> unimplemented!() (5a5f9d104)

      todo!() is useful for incremental work (telling the compiler to shut up during
      dev work).

      unimplemented!() should be used for prod code that is not yet implemented but
      expected to be deployed. It signifies "this isn't implemented but doesn't need
      doing now".


service_grpc_catalog/src/lib.rs
```diff
@@ -8,7 +8,9 @@
     clippy::explicit_iter_loop,
     clippy::future_not_send,
     clippy::use_self,
-    clippy::clone_on_ref_ptr
+    clippy::clone_on_ref_ptr,
```
alamb, line 11: what we really need is a lint for the lints 😆

alamb (approved):

PR #5769, opened 2022-09-29
chore: use 'error' label in log instead of 'e'

mkmik:
It's hard to search for errors using graylog/loki if the error message is not always in the same field.

The backoff module logs the error in the `e` field, while elsewhere we seem to consistently use the `error` field:

```
2022-09-29T01:51:47Z level=info msg="request encountered non-fatal error - backing off" e="unhandled sqlx error: pool timed out while waiting for an open connection" task_name="get namespace schema" backoff_secs=0 target="backoff" location="backoff/src/lib.rs:166" time=1664416307775697166
```

```
2022-09-29T01:53:11Z level=warn msg="failed to retrieve namespace schema" error="unhandled sqlx error: pool timed out while waiting for an open connection" req.namespace=5ea750151dd410d3_edbded071ffa14e4 target="service_grpc_schema" location="service_grpc_schema/src/lib.rs:35" time=1664416391332721908
```

I did a quick grep through the codebase and it _appears_ that we're using `error` elsewhere.

---

I'm also not sure that the correct level for that particular messages is `info` but we may address that another time.


domodwyer (approved):
I always thought this should be `WARN` - something went wrong, it's not catastrophic, but we'll try again - it's a warning!

edit: I posted this before I read you too thing info is weird!

mkmik:
cool - updated to warning

PR #5772, opened 2022-09-29
feat: always pick cold partitions in next cycle even if it has been pa…

### LINKED ISSUE #5755: Compactor (optimization): New cycle of cold compaction does not pick the partitions that have just got partially compacted
# Rationale

The function `most_cold_files_partitions` only selects partitions with `created_at` older than certain time (8 hours ago). However, if there are a lot of L-0 and we can only compact a part of them, there will be a lot of L-0 and L1 left for us to compact in the next cycles but we do not select that partition because that partition has new write of compacted files (L1) or new upgrade from L0 to L1 that update the `create_at` to the latest time. This will lead us to wait another 8 hours to continue compact this partition. And it repeats waiting for another 8 hours. It may take days/weeks for finish compacting this partition.

If the Compactors are catching up well, we do not have many L0s and do not have to worry about this. However, if for some reasons, a partition does not get compacted immediately (e.g. this can be the case if it needs more memory budget and we push it in the `skipped_compactions` table to wait for manual action), it will accumulate a lot of L0s and leads to this situation.

We need to think a way to continue compacting this partition.

Currently (2202.09.28), partition `1685212` is in exactly the situation. When I remove more partitions from the skipped_compactions , we will hit more

NGA-TRAN:
…rtially compacted recently

Closes #5755 


iox_catalog/src/interface.rs
```diff
@@ -3046,6 +3047,10 @@ pub(crate) mod test_helpers {
         let file_params = ParquetFileParams {
             object_store_id: Uuid::new_v4(),
             partition_id: third_partition.id,
+            // L1 file
+            compaction_level: CompactionLevel::FileNonOverlapped,
+            // L1 created recently but since no L0, this partition is still cold
+            created_at: time_five_hour_ago,
```
NGA-TRAN, line 3053: The is the change of the test: instead of a L-0 created long a go, make it a L1 created recently. This partition is still a cold partition

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -375,7 +375,7 @@ async fn compact_and_get_remote_partition() {
                             predicate::str::contains(r#""id": "1""#)
                                 .and(predicate::str::contains(r#""shardId": "1","#))
                                 .and(predicate::str::contains(r#""partitionId": "1","#))
-                                .and(predicate::str::contains(r#""compactionLevel": 1"#)),
+                                .and(predicate::str::contains(r#""compactionLevel": 2"#)),
```
NGA-TRAN, line 378: Since now we have cold compaction that will compact everything to level-2, this is the expected output now. I am not sure why it still passed with 1 recently

iox_catalog/src/postgres.rs
```diff
@@ -1873,21 +1873,24 @@ LIMIT $4;
...
 LEFT OUTER JOIN skipped_compactions ON parquet_file.partition_id = skipped_compactions.partition_id
 WHERE  (compaction_level = $4 OR compaction_level = $5)
 AND    to_delete IS NULL
 AND    shard_id = $1
 AND    skipped_compactions.partition_id IS NULL
 GROUP BY 1, 2, 3, 4
-HAVING max(created_at) < $2
-ORDER BY 5 DESC
+HAVING max(case when compaction_level=0 then parquet_file.created_at end) < $2  OR
+       max(case when compaction_level=0 then parquet_file.created_at end) is null
```
carols10cents, line 1892: Could you change all the instances of `compaction_level=0` in the query to `compaction_level = $4`? I've been parameterizing compaction levels so that the Rust enum is the source of truth about what enum values equal what integers and there's only one place to update if we decide to change them in the future.

iox_catalog/src/interface.rs
carols10cents: Isn't the case of a L-0 file created long ago still a valid test case though? That is, if hot compaction isn't keeping up and a level 0 file gets old, then we want to choose it for cold compaction?

In other words, why is this changing the parquet file in the test setup rather than creating a new test setup scenario?

influxdb_iox/tests/end_to_end_cases/cli.rs
carols10cents: This is suspicious... this makes me think that before this PR, the partition in this test was being treated by the compactor as a hot partition and going from 0 -> 1, and now with these changes it's being treated as a cold partition and going from 0 -> 1 -> 2? We should figure out why that is.

carols10cents (changes requested):
Close, I think!

iox_catalog/src/interface.rs
NGA-TRAN: I have gone over the tests and we have 3 partitions, all with L-0 files which means they are duplicates. Actual they are `one` and `three` partitions are similar. The reason we created 3 partitions because we want to test num_partitions =2. Thus, instead of adding a new test, we change the partition `three` to make it include no-L0 but still a cold partition.

I also add num_partitions = 5 test to make sure we return all 3 of them

iox_catalog/src/postgres.rs
NGA-TRAN: Good idea. Working on it

influxdb_iox/tests/end_to_end_cases/cli.rs
NGA-TRAN: Found the reason. The `Step::Compact` invokes `run_compaction` that will call `run_compactor_once`.

The `run_compactor_once` invokes hot that will compact all L0s into L1 and then cold. Before the change if there are only L1 files, the cold will check for their `created_at` and if recently, do nothing. With this change, if there are only L1s which is the case, the cold will work.

So actually, this is a test for the change, too 🎉

NGA-TRAN:
@carols10cents I have addressed your comment. One code change and 2 other explanations.

iox_catalog/src/postgres.rs
NGA-TRAN: Done

NGA-TRAN:
I actually want to change a bit more. No new write the last N hours means we need to consider create_time of deleted files, too

iox_catalog/src/postgres.rs
```diff
@@ -1873,21 +1873,24 @@ LIMIT $4;
...
 FROM   parquet_file
 LEFT OUTER JOIN skipped_compactions ON parquet_file.partition_id = skipped_compactions.partition_id
 WHERE  (compaction_level = $4 OR compaction_level = $5)
-AND    to_delete IS NULL
 AND    shard_id = $1
 AND    skipped_compactions.partition_id IS NULL
 GROUP BY 1, 2, 3, 4
-HAVING max(created_at) < $2
-ORDER BY 5 DESC
+HAVING count(case when to_delete is null then 1 end) > 0
```
NGA-TRAN, line 1890: Make sure we only return data of non-deleted files

iox_catalog/src/postgres.rs
```diff
@@ -1873,21 +1873,24 @@ LIMIT $4;
...
-        // We have index on (shard_id, comapction_level, to_delete)
+        // This query returns partitions with most L0+L1 files and all L0 files (both deleted and non deleted) are either created
+        // before the given time ($2) or not available (removed by garbage collector)
         sqlx::query_as::<_, PartitionParam>(
             r#"
 SELECT parquet_file.partition_id, parquet_file.shard_id, parquet_file.namespace_id,
-       parquet_file.table_id, count(parquet_file.id), max(parquet_file.created_at)
+       parquet_file.table_id, 
+       count(case when to_delete is null then 1 end) total_count,
+       max(case when compaction_level= $4 then parquet_file.created_at end)
```
NGA-TRAN, line 1883: make sure we consider created-time of deleted L0, too

NGA-TRAN:
@carols10cents : I have committed the code to also consider `create_at` of deleted L0. This is now ready for re-review. I have added more tests to cover all cases I think but this is a bit tricky and hope you give it a stare to see if I miss anything

influxdb_iox/tests/end_to_end_cases/cli.rs
carols10cents: Good!!!!!

Issue #5767, opened 2022-09-29
Ingester performs many needless queries per persist op


domodwyer:
The ingester performs many queries when it persists a partition to parquet/object store. Most of this data gathered is redundant, or can be obtained at other times.

Removing these queries will prevent a large spike in catalog requests which currently impacts the availability / performance of our catalog.

PR #5778, opened 2022-09-30
refactor: LifecycleStats tracks Namespace/TableId

domodwyer:
This is a pre-requisite to eliminating a bunch of queries for #5767.

---

* refactor: LifecycleStats tracks Namespace/TableId (e84186763)

      Changes the lifecycle handle to also track the namespace + table ID in 
      addition to the existing shard ID.

      Adds asserts to ensure the values never vary for a given partition.


NGA-TRAN (approved):
LGTM

PR #5781, opened 2022-09-30
feat: not split output files in the first step of cold compaction

### LINKED ISSUE #5779: Compactor (optimization): Do not split output of first step of cold compaction
# Rationale

For hot compaction, we split the output data into many L1 files to avoid re-compacting them with new L0s.

For cold compaction, the first step is to do the same as hot compaction that compacts L0s with L1s  to produce L1, and we did the same as the hot compaction to split the output. We knew we did not have to split them but we did not want to have many changes in our first cold compaction delivery. The reason we do not have to split the output at this stage is the partition is cold and there is no longer coming L0s for us to worry about re-compacting L1s with L0s

By not splitting them, we will save memory because of no extra output streams and no extra parquet files to save. The performance will be better too because we do not need to do the split calculation

NGA-TRAN:
Closes #5779 


compactor/src/cold.rs
```diff
@@ -700,49 +700,206 @@ mod tests {
...
     #[tokio::test]
     async fn full_cold_compaction_many_files() {
         test_helpers::maybe_start_logging();
+
+        let TestDb {
+            catalog,
+            compactor,
+            table,
+        } = make_db_with_4_l0s_2_l1s().await;
+
```
NGA-TRAN, line 809: GitHub diff does not do a good job here. I have refactored moving all the test setup right here into the function `make_db_with_4_l0s_2_l1s` to reuse in the 2 new tests above

compactor/src/cold.rs
```diff
@@ -700,49 +700,206 @@ mod tests {
...
+        assert_eq!(files.len(), 1, "{files:?}");
+        let files_and_levels: Vec<_> = files
+            .iter()
+            .map(|f| (f.id.get(), f.compaction_level))
+            .collect();
+
+        // The initial files are: L0 1-4, L1 5-6. The first step of cold compaction took files 1-5
+        // and compacted them into a l-1 file 7. The second step of cold compaction
+        // took 6 and 7 and combined them all into file 8.
+        assert_eq!(files_and_levels, vec![(8, CompactionLevel::Final)]);
```
NGA-TRAN, line 829: again, GitHub diff does not do a good job here. Except the refactored test setup, the rest of the content from `// compact` down stays the same. The only change is the comment right above this assert and `9` was replaced `8` because we now no longer split output in step 1

compactor/src/cold.rs
```diff
@@ -700,49 +700,206 @@ mod tests {
...
+                "| 421       |      | OH   | 21   | 1970-01-01T00:00:00.000091Z    |",
+                "| 70        | UT   |      |      | 1970-01-01T00:00:00.000020Z    |",
+                "| 81601     |      | PA   | 15   | 1970-01-01T00:00:00.000090Z    |",
+                "+-----------+------+------+------+--------------------------------+",
+            ],
+            &batches
+        );
+    }
+
+    #[tokio::test]
```
NGA-TRAN, line 859: again, all of the diffs below are very weird. I did not change them. Maybe the tests are similar (but not the same) and confuse GitHub diff tools.

In summary, what changes in this PR:

1. I moved test setup inside `full_cold_compaction_many_files` to its own function named `make_db_with_4_l0s_2_l1s`.
2. Change 9 to 8 as explained above
3. Created 2 new tests for step 1 with split and without split and reused `make_db_with_4_l0s_2_l1s` for them

compactor/src/cold.rs
```diff
@@ -45,7 +45,7 @@ pub async fn compact(compactor: Arc<Compactor>, do_full_compact: bool) -> usize
         compaction_type,
         CompactionLevel::Initial,
         compact_in_parallel,
-        true, // split
+        false, // no split
```
carols10cents, line 48: Well the code change is easy to review 🥳

carols10cents (approved):
Looks good!

PR #5782, opened 2022-09-30
feat: Projection pushdown for querier -> ingester for rpc queries

### LINKED ISSUE #5741: querier -> ingester: only send needed columns to Ingester in the case of gRPC queries
gRPC-query case of projection pushdown in `querier -> ingester` reported in https://github.com/influxdata/influxdb_iox/issues/5624

NGA-TRAN:
Closes #5741


iox_query/src/test.rs
```diff
@@ -108,18 +108,50 @@ impl QueryDatabase for TestDatabase {
...
+                }
+                assert!(!new_chunk.table_data.is_empty(), "Error: empty batches");
+                let new_schema = Schema::try_from(new_chunk.table_data[0].schema()).unwrap();
+                new_chunk.add_schema_to_table(new_schema, true, None);
+                new_chunk
+            })
+            .collect::<Vec<_>>();
+
+        let chunks = futures::future::join_all(chunks).await;
+        Ok(chunks)
```
NGA-TRAN, line 154: @tustvold : here is the function that hits compile problem

alamb:
If I applied this diff I got the code to compile. Not sure if that is correct or not

```diff
 diff --git a/iox_query/src/test.rs b/iox_query/src/test.rs
index ee4124fbb..2202c76a5 100644
--- a/iox_query/src/test.rs
+++ b/iox_query/src/test.rs
@@ -117,13 +117,12 @@ impl QueryDatabase for TestDatabase {
         // todo: apply projction to chunks
         println!("==== projection: {:?}", projection);
 
-        let partitions = self.partitions.lock();
+        let partitions = self.partitions.lock().clone();
         let chunks = partitions
             .values()
             .flat_map(|x| x.values())
             .filter(|x| x.table_name == table_name)
-            .map(|x| Arc::clone(x) as Arc<dyn QueryChunk>)
-            .collect::<Vec<_>>();
+            .map(|x| Arc::clone(x) as Arc<dyn QueryChunk>);
 
         let chunks = chunks
             .into_iter()
@@ -133,20 +132,19 @@ impl QueryDatabase for TestDatabase {
                 let cols = cols.iter().map(|c| c.as_str()).collect::<Vec<_>>();
                 let selection = Selection::Some(&cols);
 
-                let mut batches: Vec<RecordBatch> = vec![];
                 let mut stream = c
                     .read_filter(IOxSessionContext::with_testing(), predicate, selection)
                     .expect("Error in read_filter");
 
-                let mut new_chunk = Arc::new(TestChunk::new(c.table_name()));
+                let mut new_chunk = TestChunk::new(c.table_name());
                 while let Some(b) = stream.next().await {
                     let b = b.expect("Error in stream");
                     new_chunk.table_data.push(Arc::new(b));
                 }
                 assert!(!new_chunk.table_data.is_empty(), "Error: empty batches");
                 let new_schema = Schema::try_from(new_chunk.table_data[0].schema()).unwrap();
-                new_chunk.add_schema_to_table(new_schema, true, None);
-                new_chunk
+                let new_chunk = new_chunk.add_schema_to_table(new_schema, true, None);
+                Arc::new(new_chunk) as _
             })
             .collect::<Vec<_>>();
```

iox_query/src/test.rs
NGA-TRAN: Thanks @alamb who has found the issue

iox_query/src/test.rs
```diff
@@ -938,7 +971,14 @@ impl QueryChunk for TestChunk {
...
                 .collect::<std::result::Result<Vec<_>, ArrowError>>()?,
         };
-        Ok(stream_from_batches(batches))
+
+        if batches.is_empty() {
+            Ok(Box::pin(EmptyRecordBatchStream::new(
+                self.schema().as_arrow(),
+            )))
+        } else {
+            Ok(stream_from_batches(batches))
```
NGA-TRAN, line 980: The `stream_from_batches` only works with non-empty batches. Need to add the empty stream to have all tests that now need to invoke this function work

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1314,6 +1334,7 @@ impl InfluxRpcPlanner {
 /// Stream of chunks for table predicates.
 fn table_chunk_stream<'a>(
```
NGA-TRAN, line 1335: The main implementation is in this function. It has a new parameter `all_cols` and if true, nothing is pushed down. When it is false, only columns in the `rpc_predicate` will be read. Since we always read PK( tags and time) for deduplication, we only need to extract columns of the rpc_predicate's `exprs` and `value_expr`. 

1. Queries that has projection pushed down in this PR: `table_name`, `tag_keys`, `tag_values`
2. Queries I know that we cannot push anything down:  `field_columns`, `read_filter`
3. I am not sure about these 2 and still keep reading all columns for them are: `read_group` and `read_window_aggregate`.  @alamb : can you tell me whether we can eliminate some columns in these cases?

NGA-TRAN (commented):
@alamb and @crepererum: the PR is ready for review. I have a couple of questions whether we should have projection pushdown for `read_group` or `read_window_aggregate`

iox_query/src/test.rs
crepererum: Then `stream_from_batches` should be fixed (it probably needs an additional `schema` parameter for that).

iox_query/src/frontend/influxrpc.rs
crepererum: Why can we not push down projections for `read_filter`? This seems to be the major use case (all the cases you list under point 1 are metadata queries and ideally we would remove payload scan for these entirely at some point, if politics allows it).

crepererum: Also, the parameter `all_cols` should probably be documented.

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
 
             let database = Arc::clone(&database);
 
+            let projection = if all_cols {
+                None
+            } else {
+                // Build projection that includes all columns in the predicate's exprs and val_expr
+                // Note that since PK cols (tags and time) will be added in later steps, inside the function
+                // ingester_partitions_inner,  we don't need to consider them here.
+                let mut exprs_columns = StdHashSet::new();
```
crepererum, line 1354: I think you can use the `hashbrown` hashset here as well. It should provide you all the features you need (IIRC its interface is mostly a superset of the std variant).

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
 
             let database = Arc::clone(&database);
 
+            let projection = if all_cols {
+                None
+            } else {
+                // Build projection that includes all columns in the predicate's exprs and val_expr
+                // Note that since PK cols (tags and time) will be added in later steps, inside the function
+                // ingester_partitions_inner,  we don't need to consider them here.
```
crepererum, line 1353: I understand this explanation but the argument is extremely fragile. You argue with a currently existing, concrete implementation of an abstract interface that is used here. Instead, you should document in `QueryDatabase::chunks` that the returned data ALWAYS must includes PK columns, not matter if the projection removes them or not.

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
...
+                    .iter()
+                    .map(|e| Expr::from((*e).clone()))
+                    .collect();
+                let val_exprs_cols_result =
+                    exprlist_to_columns(&exprs, &mut exprs_columns).context(ReadColumnsSnafu);
+                let table_schema = database.table_schema(table_name);
+                // convert the column names into their corresponding indexes in the schema
+                let projection = {
+                    if expr_cols_result.is_err() || val_exprs_cols_result.is_err() {
+                        // cannot determine columns, use all columns
```
crepererum, line 1370: please add a debug log to this "read all columns" fallback.

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
...
+                                if cols.contains(&None) || cols.is_empty() {
+                                    // At least one column has no matching index, we do not know which
+                                    // columns to filter. Read all columns
+                                    None
+                                } else {
+                                    // We know which columns to filter, read only those columns
+                                    Some(cols.into_iter().flatten().collect::<Vec<_>>())
+                                }
+                            }
+                            _ => None,
```
crepererum, line 1389: IIRC the `table_schema` can only be `None` if the table doesn't exist. I guess you could exit (for this one table, not for all) early in this case and don't return any chunks at all.

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
...
+                    } else {
+                        match table_schema {
+                            Some(table_schema) => {
+                                let cols = exprs_columns
+                                    .iter()
+                                    .map(|c| table_schema.find_index_of(&c.name))
+                                    .collect::<Vec<_>>();
+
+                                if cols.contains(&None) || cols.is_empty() {
+                                    // At least one column has no matching index, we do not know which
```
crepererum, line 1381: As above: please add a debug log (which should also say which case -- contains None or cols are empty -- was hit).

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1324,9 +1345,62 @@ fn table_chunk_stream<'a>(
 
             let database = Arc::clone(&database);
 
+            let projection = if all_cols {
```
crepererum, line 1348: Maybe this whole block can be moved into a new function? This would also help testing because you could test the "build a projection" code without building the entire stream (i.e. calling `table_chunk_stream`). Looking at your test code, I think this would be a net win.

iox_query/src/test.rs
```diff
@@ -108,18 +108,56 @@ impl QueryDatabase for TestDatabase {
         &self,
         table_name: &str,
         predicate: &Predicate,
+        projection: &Option<Vec<usize>>,
         _ctx: IOxSessionContext,
     ) -> Result<Vec<Arc<dyn QueryChunk>>, QueryDatabaseError> {
+        // return Err("Sugar we are going down".into());
```
crepererum, line 114: what's that?

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1765,12 +1840,285 @@ mod tests {
...
+        let test_db = Arc::new(TestDatabase::new(Arc::clone(&executor)));
+        test_db.add_chunk("my_partition_key", Arc::clone(&chunk0));
+        let ctx = test_db.new_query_context(None);
+
+        // predicate on a tag column `foo`
+        let expr = col("foo").eq(lit("some_thing"));
+        let predicate = Predicate::new().with_expr(expr);
+        let table_predicates = vec![("h2o".to_string(), predicate)];
+
+        let result = table_chunk_stream(test_db, true /* all cols */, &table_predicates, &ctx)
```
crepererum, line 1872: If you find yourself documenting your boolean flags, maybe this is a warning sign and you shall use an enum instead :wink:

crepererum:
@alamb if you review this, hiding whitespaces in the resulting diff helps a lot (since a large chunk of the changes is rustfmt re-arranging the code).

alamb (approved):
I found whitespace blind diff https://github.com/influxdata/influxdb_iox/pull/5782/files?w=1 much easer to see what had changed

Another test I think would add value is a basic "is this plumbing hooked up" type test -- that runs a grpc query and then validates that not all columns were fetched. 

In general there are some structural changes that I think would help with readability, and I have some concerns about the specifics of implementation. However, I feel very good about the coverage in `query_tests` so the fact that this PR passes means I feel confident in approving

This is a nice improvement

iox_query/src/test.rs
NGA-TRAN: Good catch. It was a simulation for a failed test that I fixed later

NGA-TRAN (commented):
Very helpful comments @crepererum and @alamb . Let me address them in this PR

NGA-TRAN:
I convert this to draft because even though I have addressed the main comments, there are are few minor things I need to address tomorrow

iox_query/src/frontend/influxrpc.rs
NGA-TRAN: The `exprs_columns` is the mutable parameter of DF's `eprlist_to_columns` and it is defined as std::HashSet

iox_query/src/frontend/influxrpc.rs
NGA-TRAN: Done

iox_query/src/frontend/influxrpc.rs
NGA-TRAN: It works. Since we return projection here, I make the projection None if table does not exist

iox_query/src/frontend/influxrpc.rs
NGA-TRAN: Done

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1314,6 +1334,7 @@ impl InfluxRpcPlanner {
 /// Stream of chunks for table predicates.
 fn table_chunk_stream<'a>(
     database: Arc<dyn QueryDatabase>,
+    all_cols: bool,
```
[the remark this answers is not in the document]
NGA-TRAN: @alamb : I have used `field_columns` per your suggestion to filter more data. However, I still need extra parameter in case this field_coumns is empty.  I renamed `all_cols` to `need_fields` for this purpose. More specifically, if need_fields is false (in case we only need to return tags or table name), we will do nothing if field_coumns is empty. When need_fields is true (in the case of field_columns, read_filter, ...) and field_coumns is empty, we have to read all field columns

NGA-TRAN:
@alamb and @crepererum : I have addressed all your comments and have a question. I will merge this shortly. If there are extra stuff you think we need to do, I will do it in next PR

PR #5786, opened 2022-10-03
fix(ingester): incorrect memory tracking of failed writes

domodwyer:
A quick fix for the op buffering / memory tracking where a write was recorded in the memory tracker before being applied. Because applying the op is fallible, this could lead to writes being recorded as applied, but not actually applied, and thus slowly draining the ingester of usable memory budget.

I don't think this has ever been hit, but it is in the area I'm working on today and was a quick fix 🚀 

---

* test: MockLifecycleHandle captures calls (20451921d)

      Changes the NoopLifecycleHandle to MockLifecycleCall, and adds code causing it
      to log all calls made to the log_write() method.

      This will allow tests to assert calls and their values in DML buffering tests.

* fix: spurious memory accounting for failed write (b23ad3171)

      Fixes a case where the ingester may incorrectly record a write as having been
      buffered in memory, when in fact the buffering failed.

      This could cause the effective buffer size to be reduced over time as more and
      more data is spuriously "added" to the buffer, but never released back to the
      memory tracker as it is never persisted.

### DIFF
--- a/ingester/src/data/table.rs
+++ b/ingester/src/data/table.rs
@@ -125,6 +125,11 @@ impl TableData {
         let rows = batch.rows();
         partition_data.buffer_write(sequence_number, batch)?;
 
+        // Record the write as having been buffered.
+        //
+        // This should happen AFTER the write is applied, because buffering the
+        // op may fail which would lead to a write being recorded, but not
+        // applied.
         let should_pause = lifecycle_handle.log_write(
             partition_data.id(),
             self.shard_id,
--- a/ingester/src/data/table.rs
+++ b/ingester/src/data/table.rs
@@ -121,16 +121,19 @@ impl TableData {
             }
         }
 
+        let size = batch.size();
+        let rows = batch.rows();
+        partition_data.buffer_write(sequence_number, batch)?;
+
         let should_pause = lifecycle_handle.log_write(
             partition_data.id(),
             self.shard_id,
             self.namespace_id,
             self.table_id,
             sequence_number,
-            batch.size(),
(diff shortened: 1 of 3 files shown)


ingester/src/data/table.rs
```diff
@@ -121,16 +121,19 @@ impl TableData {
             }
         }
 
+        let size = batch.size();
+        let rows = batch.rows();
+        partition_data.buffer_write(sequence_number, batch)?;
+
         let should_pause = lifecycle_handle.log_write(
```
alamb, line 128:
```suggestion
        // Log the write *after* trying to apply it so that failed writes
        // are not erroneously included in the tracking statistics
        let should_pause = lifecycle_handle.log_write(
```

ingester/src/data/table.rs
```diff
@@ -207,3 +210,146 @@ impl TableData {
...
+                    }
+                },
+            } => { assert_eq!(col_name, "value") }
+        );
+
+        // And the lifecycle handle should not be called.
+        //
+        // It still contains the first call, so the desired length is 1
+        // indicating no second call was made.
+        assert_eq!(handle.get_log_calls().len(), 1);
```
alamb, line 353: 👍

ingester/src/data/table.rs
alamb: I suggest adding a commnent explaining this subtlety as it may not be obvious to a future reader (though the tests cover the situation nicely 👍 )

alamb (approved):
🏅  very nice tests

Issue #4880, opened 2022-06-16
Refer to objects by catalog ID in kafka message, not their string names


domodwyer:
Ideally when pushing writes from the router, to the ingester, via kafka, we would avoid sending (many) copies of the following strings and instead pass their catalog IDs:

* database name
* table name (many)
* column names (many many)

This would decrease message size, and avoid needing the ingester to perform catalog queries to resolve the name strings to IDs. We may be able to drop additional indexes in postgres too.

cc @pauldix

PR #5791, opened 2022-10-04
refactor: reference buffer tree nodes by ID

domodwyer:
This PR accomplishes two things:

* Makes all the ID lookup queries at `persist()` time redundant, with the exception of acquiring the `SortKey`. In a follow-up PR I will remove this requirement and significantly reduce the query load at persist time (part of #5767).

* Changes the buffer tree nodes (shards -> namespaces -> tables -> partitions) to support being addressed by ID, in addition to string names. This enables the above, but is also in preparation for #4880.

There will be a slight memory overhead for maintaining the additional ID indexes, but I've been careful to minimise the amount of memory these will use (through use of existing Arc's, storing ref-counted keys and making two lookups in the cold ID path, etc).

This also documents a TOCTOU bug tracking the memory usage of partitions, and releasing that memory back during persist ops: https://github.com/influxdata/influxdb_iox/issues/5777. The effect of this bug is the usable amount of buffer memory **decreases over time** so is not likely to be a contributor to OOMs.

---

* refactor: ref TableData by name & ID (9c0e4e98c)

      Changes the NamespaceData to hold a map of table name -> TableData, and table
      ID -> TableData simultaneously.

      This allows for cheap lookups when the caller holds an ID, and is part of
      preparatory work to transition away from using string names in the ingester
      for tables.

      This commit also switches from a BTreeMap to a HashMap as the backing 
      collection, as maintaining key ordering doesn't appear to be necessary.

* refactor: ref NamespaceData by name & ID (66e05b5ea)

      Changes the ShardData to hold a map of namespace name -> NamespaceData, and
      namespace ID -> NamespaceData simultaneously.

      This allows for cheap lookups when the caller holds an ID, and is part of
      preparatory work to transition away from using string names in the ingester
      for tables.

      This commit also switches from a BTreeMap to a HashMap as the backing 
      collection, as maintaining key ordering doesn't appear to be necessary.

* refactor: PartitionData::id() -> partition_id() (0847cc545)

      Consistent naming is consistent - all the others are thing_id().

* refactor: ref PartitionData by key & ID (f9bf86927)

      Changes the TableData to hold a map of partition key -> PartitionData, and
      partition ID -> PartitionData simultaneously. This allows for cheap lookups
      when the caller holds an ID.

      This commit also manages to internalise the partition map within the TableData
      - one less pub / peeking!

      This commit also switches from a BTreeMap to a HashMap as the backing 
      collection, as maintaining key ordering doesn't appear to be necessary.

* refactor: persist() passes all necessary IDs (1a7eb47b8)

      This commit changes the persist() call so that it passes through all relevant
      IDs so that the impl can locate the partition in the buffer tree - this will
      enable elimination of many queries against the catalog in the future.

      This commit also cleans up the persist() impl, deferring queries until the
      result will be used to avoid unnecessary load, improves logging & error
      handling, and documents a TOCTOU bug in code:

          https://github.com/influxdata/influxdb_iox/issues/5777

### DIFF
--- a/ingester/src/compact.rs
+++ b/ingester/src/compact.rs
@@ -86,11 +86,8 @@ pub(crate) async fn compact_persisting_batch(
     namespace_id: i64,
     partition_info: &PartitionInfo,
     batch: Arc<PersistingBatch>,
-) -> Result<Option<CompactedStream>> {
-    // Nothing to compact
-    if batch.data.data.is_empty() {
-        return Ok(None);
-    }
+) -> Result<CompactedStream> {
+    assert!(!batch.data.data.is_empty());
 
     let namespace_name = &partition_info.namespace_name;
     let table_name = &partition_info.table_name;
@@ -141,11 +138,11 @@ pub(crate) async fn compact_persisting_batch(
         sort_key: Some(metadata_sort_key),
     };
 
-    Ok(Some(CompactedStream {
+    Ok(CompactedStream {
         stream,
         iox_metadata,
         sort_key_update,
-    }))
+    })
 }
 
 /// Compact a given Queryable Batch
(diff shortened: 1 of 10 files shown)


NGA-TRAN (approved):
The logic looks good

PR #5794, opened 2022-10-05
fix: Correct representation of 3-part measurement name

### LINKED ISSUE #5662: InfluxQL: 3-part measurement name must support regex
The existing implementation of `MeasurementNameExpression` is defined as:

```ebnf
measurement      ::= measurement_name |
                   ( policy_name "." measurement_name ) |
                   ( db_name "." policy_name? "." measurement_name )

db_name          ::= identifier
measurement_name ::= identifier
policy_name      ::= identifier
```
However, a 3-part measurement reference is defined by the following EBNF:

```ebnf
measurement      ::= measurement_name |
                   ( policy_name "." measurement_name ) |
                   ( db_name "." policy_name? "." measurement_name )

db_name          ::= identifier
measurement_name ::= identifier | regex_lit
policy_name      ::= identifier
```

This definition means that InfluxQL will accept and execute queries using the type for `measurement_name`, rather than
the type of operator.

More specifically, in the following example, InfluxQL[^2] executes an exact match for the `disk` measurement name, despite
the use of a regular expression equals operator `=~`[^1]

```text
> show measurements with measurement =~ db..disk
name: measurements
name
----
disk
```

[^1]: The expected result is an error that the `disk` token is not a regular expression
[^2]: Tested with InfluxDB OSS v1.8

Conversely, specifying despite the use of the `=` operator, because the `measurement_name` is a regular expression
reference, `/disk/`, InfluxQL returns all matching measurements, rather than an expected error:

```text
> show measurements with measurement = db../disk/
name: measurements
name
----
disk
diskio
```

stuartcarnie:
Closes #5662

This PR teaches the InfluxQL parser how to correctly represent a 3-part measurement name.


influxdb_influxql_parser/src/common.rs
```diff
@@ -290,35 +300,107 @@ mod tests {
...
+                name,
+            }
+        }
+    }
+
     #[test]
-    fn test_measurement_name_expression() {
-        let (_, got) = measurement_name_expression("diskio").unwrap();
+    fn test_qualified_measurement_name() {
+        use MeasurementName::*;
```
crepererum, line 345: TIL that this is possible :)

crepererum (approved):

PR #5796, opened 2022-10-05
feat: AST traversal using Visitor pattern

### LINKED ISSUE #5783: InfluxQL: AST should support traversal
We would like to implement transformations of the InfluxQL AST, such as to a DataFusion logical plan. A well understood approach is to utilise the [Visitor Pattern](https://en.wikipedia.org/wiki/Visitor_pattern).

Existing Rust implementations for inspiration include:

* https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_ast/visit.rs.html#102-111
* https://github.com/apache/arrow-datafusion/blob/master/datafusion/expr/src/expr_visitor.rs

## [derive_visitor crate](https://docs.rs/derive-visitor/latest/derive_visitor/)

@alamb introduced me to this crate via [a PR](https://github.com/sqlparser-rs/sqlparser-rs/pull/601). I pushed #5790 to demonstrate its use.

The benefits of the `derive_visitor` crate become evident when implementing visitors. For example,

https://github.com/influxdata/influxdb_iox/blob/293bfb0aa663c0f92fe41f6e966afe15588c87ca/influxdb_influxql_parser/src/delete.rs#L109-L117

demonstrates mutating the existing AST to convert the names in the `FROM` clause to uppercase.

This second example shows implementing a visitor with explicit enter / exit handlers for specific AST nodes:

https://github.com/influxdata/influxdb_iox/blob/293bfb0aa663c0f92fe41f6e966afe15588c87ca/influxdb_influxql_parser/src/delete.rs#L119-L149

stuartcarnie:
Closes #5783 

## AST / Visitor support

This PR adds a `visit` module to the `influxdb_influxql_parser` crate, to provide new API for walking parsed InfluxQL statements in a depth-first manner. This API has been modelled after the `expr_visitor` module in DataFusion:

https://github.com/apache/arrow-datafusion/blob/409ab1443ba1060d6d07998d77a9397e30b87f4d/datafusion/expr/src/expr_visitor.rs#L23-L53

Clients implement the `Visitor` trait and then override select `pre_visit` and `post_visit` functions based on the nodes to be interrogated. The `pre_visit` hook returns a `Recursion` value:

https://github.com/influxdata/influxdb_iox/blob/c84ac77467da0d6f0937919180109d05464e4e62/influxdb_influxql_parser/src/visit.rs#L28-L35

which provides two enumerations to allow the `Visitor` to stop or continue traversal deeper into the tree.

For example, a client may wish to inspect the `WHERE` clause and print it out:

https://github.com/influxdata/influxdb_iox/blob/63abe65990a46286ac5bbaa4de587470a17b5260/influxdb_influxql_parser/src/visit.rs#L5-L23

## Public API

Access modifiers were updated, `pub` vs `pub(crate)`, to permit exporting the desired types external to the crate. As a result, a significant amount of additional documentation was added.

## insta snapshot tests

The `insta` crate was added to utilise snapshot tests to validate all `pre` and `post` functions of the `Visitor` trait are called and in the correct order.


influxdb_influxql_parser/src/common.rs
```diff
@@ -122,49 +127,95 @@ pub fn qualified_measurement_name(i: &str) -> ParseResult<&str, QualifiedMeasure
...
 
+/// Represents the `WHERE` clause of a statement.
+#[derive(Debug, Clone, PartialEq)]
+pub struct WhereClause(pub ConditionalExpression);
+
+impl Display for WhereClause {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "WHERE {}", self.0)
+    }
+}
```
domodwyer, line 185: I think providing context-specific wrapping types is a fantastically valuable addition beyond being able to impl `Display` - being able to punt checking functions are being given the correct `ConditionalExpression` to the compiler is a fantastic maintenance & code readability improvement :heart:

(TL;DR:)

```rust
fn do_stuff_with_where(x: ConditionalExpression)
```

vs.

```rust
fn do_stuff_with_where(x: WhereClause)
```

influxdb_influxql_parser/src/common.rs
```diff
@@ -408,15 +448,15 @@ mod tests {
     #[test]
     fn test_limit_clause() {
         let (_, got) = limit_clause("LIMIT 587").unwrap();
-        assert_eq!(got, 587);
+        assert_eq!(got.0, 587);
```
domodwyer, line 451: You could also implement [`Deref`] on the type wrappers, which lets the wrapper be coerced into the inner type by the compiler and hides the indexing / lets you make it not-pub. It's a common pattern for these wrappers - here's an example for our `DatabaseName` type wrapper that adds context type-safety to the bare `String` identifier of a database name:

https://github.com/influxdata/influxdb_iox/blob/b5973b557515e77839ee5462a69416c372d6e42a/data_types/src/lib.rs#L1605-L1611

It's important to note that the wrapper is dereferenced to a reference of the inner value (that is, `&self.0`) though. I knocked this up to play with: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a987f8a2557650acd3f04613a9f765b4

[`Deref`]:https://doc.rust-lang.org/std/ops/trait.Deref.html

influxdb_influxql_parser/src/common.rs
```diff
@@ -122,49 +127,95 @@ pub fn qualified_measurement_name(i: &str) -> ParseResult<&str, QualifiedMeasure
...
+
+impl Display for LimitClause {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "LIMIT {}", self.0)
+    }
+}
+
 /// Parse a `LIMIT <n>` clause.
-pub fn limit_clause(i: &str) -> ParseResult<&str, u64> {
+pub(crate) fn limit_clause(i: &str) -> ParseResult<&str, LimitClause> {
```
domodwyer, line 141: I've recently gone on a "don't export everything" crusade in the ingester, so lots of :heart: for this!

domodwyer (approved):
Looks good! Love the snapshot tests 👍

NGA-TRAN (approved):
What a PR! Very nice, Stuart.

influxdb_influxql_parser/src/common.rs
stuartcarnie: Thanks Dom – totally agree!

influxdb_influxql_parser/src/common.rs
stuartcarnie: Awesome advice! I'm going to do this before merging – thank you.

influxdb_influxql_parser/src/common.rs
stuartcarnie: I'm hugely in favour of being explicit also – hopefully makes it clearer for others (and my future self 😆)

PR #5800, opened 2022-10-05
feat: Swap Kafka Producer implementation back to rdkafka as diagnosis of latency problem

carols10cents:
Connects to https://github.com/influxdata/conductor/issues/1149

As an aid in diagnosing the producer latency, swap the Kafka producer implementation back to rdkafka, mostly as it was in https://github.com/influxdata/influxdb_iox/pull/3625.

This is set to tell kodiak to squash, so it should be easily revertable, but I left the multiple commits for ease of reviewing.

@domodwyer I tried to make the `KafkaProducerMetrics` more general to be able to reuse it directly for the `write_buffer_client_produce_duration` metric, but that was... too complex and not really necessary. Please check that where I put the time statements is where you'd like them to get equivalent metrics!


write_buffer/src/kafka/rdkafka.rs
```diff
@@ -0,0 +1,499 @@
...
+        }
+
+        // Only send writes with known shard indexes to Kafka.
+        if !self.partitions.contains(&shard_index) {
+            return Err(format!("Unknown shard index: {}", shard_index).into());
+        }
+
+        let kafka_partition_id = shard_index.get();
+
+        let enqueue_start = self.time_provider.now();
```
carols10cents, line 98: Here's the metric start time...

write_buffer/src/kafka/rdkafka.rs
```diff
@@ -0,0 +1,499 @@
...
+        database_name,
+        n_sequencers.get() as i32,
+        TopicReplication::Fixed(1),
+    );
+    for (k, v) in cfg {
+        topic = topic.set(k, v);
+    }
+
+    let opts = AdminOptions::default();
+    let mut results = admin.create_topics([&topic], &opts).await?;
```
domodwyer, line 348: Is there any danger of this overwriting the current config or otherwise messing things up if args/counts/etc don't match?

Topics are created externally in prod, but I guess this is used for local devs / other users.

write_buffer/src/kafka/rdkafka.rs
```diff
@@ -0,0 +1,499 @@
...
+                    tmp
+                };
+
+                self.consumer_lag
+                    .recorder(attributes)
+                    .set(partition.consumer_lag as u64);
+            }
+        }
+    }
+}
```
domodwyer, line 499: We can probably do away with this and avoid creating a bunch more metrics in prometheus for the test :+1:

It's already angry enough!

domodwyer (commented):
I think this is probably good - I didn't give it too hard a look-over given it's existing code and temporary and I trust you - I hear you can rust 🦀

write_buffer/src/kafka/rdkafka.rs
carols10cents: Yes, the docs for this function say "This will NOT fail if the topic already exists!" which I take to mean it *will* overwrite the current config. 

The only place this function is called is in `maybe_auto_create_topics`, which only calls `create_kafka_topic` if no Kafka partitions are found. Production should always have Kafka partitions created externally, so this should never run in production.

I'll make these comments a bit clearer-- if we end up staying with this code long term, I'm happy to rearrange so that this really super for sure only called in non production environments.

carols10cents:
@domodwyer I think I've addressed everything! I'm not going to add the "automerge" label-- you do it when you're ready :)

domodwyer:
FYI I plan to merge this tomorrow (tuesday) as there has a been an outage / lots of noise that may mess up any comparisons.

domodwyer (approved):

domodwyer:
Thanks for this @carols10cents 🙏

PR #5801, opened 2022-10-05
fix: query error propagation

### LINKED ISSUE #5799: Querier OOM protection should result in proper gRPC status code
Currently the querier OOM protection results in "internal" or "invalid argument" error codes -- depending on the interface you use. It should return a proper "resource exhausted" error code.

crepererum:
- treat OOM protection as "resource exhausted"
- use `DataFusionError` in more places instead of opaque `Box<dyn Error>`
- improve conversion from/into `DataFusionError` to preserve more semantics

Overall, this improves our error handling. DF can now return errors like "resource exhausted" and gRPC should now automatically generate a sensible status code for it.

Fixes #5799.


mkmik (approved):
nice!

domodwyer (approved):

PR #5810, opened 2022-10-06
feat: build a query plan without deduplication

### LINKED ISSUE #5751: Filter data without deduplication
# Rationale
We recently remove deduplication  for Ingester data before sending it to Querier. This leads to no filter data as well. This ticket is to support a query plan that filter data but not apply deduplication

### LINKED ISSUE #5815: Build scan plan without deduplication
This is a task of #5751 that supports a scan plan without deduplication

NGA-TRAN:
Closes #5815 


iox_query/src/frontend.rs
```diff
@@ -57,6 +59,221 @@ mod test {
         };
     }
 
+    #[tokio::test]
```
NGA-TRAN, line 62: Tests creating logical plans are in this file

iox_query/src/provider.rs
```diff
@@ -1349,6 +1423,16 @@ mod test {
             chunk_ids(&chunks.no_duplicates_chunks),
             "00000000-0000-0000-0000-000000000001"
         );
+
```
NGA-TRAN, line 1426: Tests for physical plans are in this file

iox_query/src/frontend.rs
```diff
@@ -57,6 +59,221 @@ mod test {
...
+        assert_batches_eq!(&expected, &batches0);
+    }
+
+    #[tokio::test]
+    async fn test_scan_plan_without_deduplication() {
+        // Create 2 overlapped chunks
+        let (schema, chunks) = get_test_chunks();
+        let ctx = IOxSessionContext::with_testing();
+
+        // Build a logical plan without deduplication
```
NGA-TRAN, line 139: @domodwyer and @crepererum The code that hooks the scan with filter but without deduplication into the Ingester will look similar to this test. I am happy to do it after all the refactoring in the Ingester or feel free to hook them in at your convenience

iox_query/src/provider.rs
```diff
@@ -466,13 +518,20 @@ impl Deduplicater {
...
         // Building plans
         let mut plans: Vec<Arc<dyn ExecutionPlan>> = vec![];
-        if chunks.no_duplicates() {
-            // Neither overlaps nor duplicates, no deduplicating needed
-            debug!("All chunks neither overlap nor duplicate. Build only one scan node for all of them.");
+        if self.disable_deduplication || chunks.no_duplicates() {
+            // Either we do not want to do deduplication even if there are duplicates in the chunks or
+            // the chunks have neither overlaps nor duplicates
+            if self.disable_deduplication {
+                debug!(%table_name, "Deduplication is disable. Build only one scan node for all of them.");
```
crepererum, line 529: I'm only somewhat familiar w/ this code, but do you split the filter expression somewhere into parts that can be evaluated on non-duplicated data and stuff that requires de-dup?

Imagine the following chunk:

```
table foo=1,bar=1 0
table foo=2,bar=2 0
```

If you evaluate `bar=1` on the duplicated data and run de-dup (within the querier), you'll get `table foo=1,bar=1 0` as a result, however if you de-dup first and than evaluate the predicate, you (correctly) will filter out everything.

alamb: This is an excellent point -- basically we need to apply deduplication prior to any filtering. Thus, if we are skipping deduplication in the ingester, we also can't apply filtering prior to combining / deduplication again in the querier

In terms of the specific example, there is always a filter applied after all the scanning etc made by IOx, so I think it will correctly always filter out all the data

iox_query/src/provider.rs
```diff
@@ -2289,13 +2375,29 @@ mod test {
...
+                None,
+            )
             .unwrap();
         let batch = test_collect(plan).await;
         // No duplicates so no sort at all. The data will stay in their original order
         assert_batches_sorted_eq!(&expected, &batch);
+
+        // ----------------------------------
+        // Test with disable_deduplication on
+        let deduplicator = Deduplicater::new(IOxSessionContext::with_testing(), true);
```
alamb, line 2394: same thing here, stylistically this might be easier to read if it was like

```suggestion
        let deduplicator = Deduplicater::new(IOxSessionContext::with_testing())
          .with_deduplication(false);
```

iox_query/src/provider.rs
```diff
@@ -2366,6 +2474,16 @@ mod test {
...
+        let plan = deduplicator
+            .build_scan_plan(Arc::from("t"), schema, chunks, Predicate::default(), None)
+            .unwrap();
+        let batch = test_collect(plan).await;
+        // Deduplication is disabled, the output shoudl be the same as the original data
+        assert_batches_sorted_eq!(&original_expected, &batch);
```
alamb, line 2486: Should the sort be the same too (aka should this be `asset_batches_eq`?

iox_query/src/provider.rs
```diff
@@ -2948,6 +3133,57 @@ mod test {
...
+        let count = plan_str.matches("DeduplicateExec").count();
+        assert_eq!(count, 0);
+
+        // Verify 1 UnionExec
+        let count = plan_str.matches("UnionExec").count();
+        assert_eq!(count, 1);
+
+        // Verify 1 SortPreservingMergeExec
+        let count = plan_str.matches("SortPreservingMergeExec").count();
+        assert_eq!(count, 1);
```
alamb, line 3182: I am not sure how much value verifying these plan nodes adds

alamb (approved):
🏅  amazing tests, as always!

I have some suggestions on style ut on substance I think this is great. Nice work @NGA-TRAN

iox_query/src/provider.rs
crepererum: > In terms of the specific example, there is always a filter applied after all the scanning etc made by IOx, so I think it will correctly always filter out all the data

Different example that has the same issue and is not solved by yet-another filter step:

```
table foo=1,bar=1 0
table bar=2 0
```

Predicate `foo=1 AND bar=2` will filter out both lines prior to de-dup, so this predicate MUST NOT be applied early.

> This is an excellent point -- basically we need to apply deduplication prior to any filtering. 

In general I think just skipping filters when there's any de-dup to run kinda defeats the purpose, because than the ingester won't filter at all in many cases. Running de-dup on the other hand is exactly what we wanna avoid for perf reasons. I think we should isolate from expressions -- which are mostly OR/AND-combinations -- the part that can safely be applied prior to de-dup (that needs some careful thinking).

NGA-TRAN: > table foo=1,bar=1 0
table bar=2 0
Predicate foo=1 AND bar=2 will filter out both lines prior to de-dup, so this predicate MUST NOT be applied early.

@crepererum : This is a good example. This means if we apply filter before deduplication, we need to see which cases we can do it

iox_query/src/provider.rs
NGA-TRAN: The original plan is to test the plan and I just add the case of without deduplication. The plans are different and I think they are useful

iox_query/src/frontend.rs
NGA-TRAN: @crepererum has a good example about cases we cannot apply filter prior to deduplication. If we want to use this, we must be carefully to avoid those cases

https://github.com/influxdata/influxdb_iox/pull/5810/files#r1004392310

iox_query/src/frontend.rs
```diff
@@ -57,6 +59,221 @@ mod test {
...
+
+        // Build physical plan
+        let executor = Executor::new(1);
+        let physical_plan = executor
+            .new_context(ExecutorType::Reorg)
+            .create_physical_plan(&logical_plan)
+            .await
+            .unwrap();
+
+        // Verify plan: the plan has sort and merge but no deduplication
```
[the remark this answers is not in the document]
NGA-TRAN: I will remove the asserting the plans

iox_query/src/frontend/common.rs
```diff
@@ -106,6 +108,8 @@ impl<'a> ScanPlanBuilder<'a> {
             chunks: vec![],
             output_sort_key: None,
             predicate: None,
+            // always do deduplication in query
+            disable_deduplication: false,
```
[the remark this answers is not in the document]
NGA-TRAN: Will use this suggestions everywhere

iox_query/src/provider.rs
alamb: Filed https://github.com/influxdata/influxdb_iox/issues/6066 to track

PR #5818, opened 2022-10-07
feat(influxrpc): rewrite missing column references to NULL

alamb:
# Rationale
Broken out of the main datafusion upgrade PR https://github.com/influxdata/influxdb_iox/pull/5792

This is needed as part of the great "coercion" change where DataFusion changed how it did type coercion -- so specifically, things like `int64_col = int32_constant` no longer work

The issue problem is that DataFusion now does type coercion (e.g. casting `col = constant` so the arguments are the same type) as part of logical planning where it did so previously as part of physical planning

This causes issues because the expression simplification logic in DataFusion runs a physical plan, and therefore now requires proper type coercion to have occurred to complete successfully

# Changes
1. Add an explicit "rewrite missing columns" pass in the influxrpc_predicate predicate normalization


predicate/src/rpc_predicate.rs
```diff
@@ -187,31 +191,46 @@ fn normalize_predicate(
...
                 // augment field projections with conditions using `CASE` statements.
                 .and_then(|e| rewrite_field_value_references(&mut field_value_exprs, e))
+                .map(|e| log_rewrite(e, "rewrite_field_value_references"))
                 // Rewrite any references to `_field` with a literal
                 // and keep track of referenced field names to add to
                 // the field column projection set.
                 .and_then(|e| field_projections.rewrite_field_exprs(e))
+                .map(|e| log_rewrite(e, "field_projections"))
+                // remove references to columns that don't exist in this schema
+                .and_then(|e| e.rewrite(&mut missing_columums))
```
alamb, line 217: This is the actual logic change -- the calls to `log_rewrite` were just for debugging this issue

predicate/src/rpc_predicate.rs
```diff
@@ -290,9 +314,27 @@ mod tests {
...
     use arrow::datatypes::DataType;
-    use datafusion::logical_plan::{col, lit};
+    use datafusion::{
+        logical_plan::{col, lit},
+        scalar::ScalarValue,
+    };
     use test_helpers::assert_contains;
 
+    #[test]
+    fn test_normalize_predicate_coerced() {
```
alamb, line 324: hinting at other changes to come 😄

predicate/src/rpc_predicate.rs
```diff
@@ -336,6 +378,20 @@ mod tests {
...
+    fn test_normalize_predicate_field_non_tag() {
+        // should treat
+        let predicate = normalize_predicate(
+            "table",
+            schema(),
+            &Predicate::new().with_expr(col("not_a_tag").eq(lit("blarg"))),
+        )
+        .unwrap();
+
+        let expected = Predicate::new().with_expr(lit(ScalarValue::Boolean(None)));
```
alamb, line 391: Converted to a `NULL` correctly!

predicate/src/rpc_predicate/column_rewrite.rs
```diff
@@ -0,0 +1,99 @@
...
+            Ok(false)
+        }
+    }
+}
+
+fn lit_null() -> Expr {
+    lit(ScalarValue::Utf8(None))
+}
+
+impl ExprRewriter for MissingColumnRewriter {
```
alamb, line 38: Here is the rewrite logic

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -205,12 +205,12 @@ async fn test_read_filter_invalid_predicate_case() {
...
 async fn test_read_filter_unknown_column_in_predicate() {
     let predicate = Predicate::new()
-        // mystery_region is not a real column, so this predicate is
+        // mystery_region and bar are not real columns, so this predicate is
         // invalid but IOx should be able to handle it (and produce no results)
         .with_expr(
-            col("baz")
-                .eq(lit(4i32))
-                .or(col("bar").and(col("mystery_region").gt(lit(5i32)))),
+            col("baz").eq(lit(4i32)).or(col("bar")
```
alamb, line 211: This test previously was using `bar` directly as a boolean column which does not occur in rpc predicates

Instead the predicates are always `<col> <op> <literal>` so I changed it (otherwise this test fails because the rewrite puts a NULL string rather than a null literal)

predicate/src/lib.rs
```diff
@@ -12,7 +12,6 @@
 
 pub mod delete_expr;
 pub mod delete_predicate;
-pub mod rewrite;
```
alamb, line 15: I just moved this module into `rpc_predicate` to make it clearer that it was a rewrite for rpc_predicates

service_grpc_influxrpc/src/service.rs
```diff
@@ -1802,11 +1802,13 @@ mod tests {
         // Note multiple tables / measureemnts:
         let chunk0 = TestChunk::new("m1")
             .with_id(0)
+            .with_tag_column("state")
```
alamb, line 1805: These tests have a predicate that refers to `"state"` but the schema did not have it (and thus the column reference was rewritten to NULL). Thus I added the missing columns

predicate/src/rpc_predicate.rs
NGA-TRAN: 🎉

NGA-TRAN (approved):
Looks good. The functions are short 👍

alamb (commented):
Thank you for the review @NGA-TRAN

Issue #5005, opened 2022-06-30
Gossip to converge router schema caches faster


domodwyer:
## TL;DR

We should do our best to ensure all the routers schema cache are kept up to date / in sync. This is because:

* The schema API is used by the query team to discover the schema of a table for querying
  * They need the correct schema, not a partial schema that hasn't converged yet
* Each schema cache miss results in a catalog query
  * This causes a catalog write amplification of N where N=number of routers that need to discover and cache the column
* We need a cheap way to enforce column table / column limits ("service protection limits")
  * See https://github.com/influxdata/influxdb_iox/issues/4062

Current behaviour is probably acceptable for the first two points. The third requirement is likely low priority for the time being.

## Technical Description of Current State

The routers cache the schemas of all tables. When a router starts up, it pre-warms this cache with all the schema elements in the catalog and then incrementally adds new schema elements to the cache as each router observes them in the writes it processes.

When a router sees a column that is not within its schema cache, it attempts to insert it to the catalog. This either succeeds if the column has never been observed by the overall system, or the insert fails if another router has previously observed the column. In either case, the router then adds this column to it's schema cache and uses it to avoid hitting the catalog for subsequent requests. 

This behaviour causes the routers to be eventually consistent, synchronising their state through the catalog: in the absence of new columns being wrote, and in the presence of writes driving the convergence, all the routers will eventually cache the full set of schema elements.

In practice, we see a low amount of entropy with the current system:

![image](https://user-images.githubusercontent.com/9275968/176688498-72128bed-1282-4532-84ee-30db1881f8a0.png)

## Proposal

Gossip schema additions between routers, providing a cheap, low latency way of converging their caches without hitting the catalog at all.

This will remove the catalog write amplification that drives convergence today.

jacobmarble:
This idea would reduce catalog load, but we should hold off on catalog performance improvements while the catalog service is being developed.

PR #5819, opened 2022-10-07
feat: Enforce table column limits from the schema cache

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/4349.

Adding many columns at once, which `write` does through `validate_or_insert_schema`, was not checking the namespace `max_columns_per_table` limit because [the query was complicated and expensive](https://github.com/influxdata/influxdb_iox/pull/4351).

This PR adds a function `validate_column_limits` that takes the mutable batches in a write and the in-memory NamespaceSchema cache and checks that, as far as the cache knows, the columns in all the batches won't cause the table to go over the namespace's limit.

In addition to [the race condition I mentioned](https://github.com/influxdata/influxdb_iox/issues/4349#issuecomment-1270622632), this solution has another limitation that updating the `max_columns_per_table` limit on the namespace in the catalog doesn't update the namespace cache value, so if we change the limits at runtime, they won't take effect until that namespace schema is entirely reloaded into the cache, likely at router startup. If that's not ok, let me know and I can work on a solution!


router/src/dml_handlers/schema_validation.rs
```diff
@@ -253,17 +259,92 @@ where
...
+        table_column_names.append(&mut batch_column_names);
+        let num_batch_columns = table_column_names.len();
+
+        if num_batch_columns > schema.max_columns_per_table {
+            return Err(OverColumnLimit {
+                table_name: table_name.into(),
+                num_batch_columns,
+                max_columns_per_table: schema.max_columns_per_table,
+            });
+        }
```
domodwyer, line 292: I think the variable `table_column_names` actually being the full set of columns after line 283 caused the error message inconsistency above! If we were to use the proposed error, then `OverColumnLimit` would be something like:

```rust
struct OverColumnLimit {
    table_name: String,

    // Number of resultant columns after merging the write with existing
    // columns.
    merged_column_count: usize,

    // Number of columns already in the table.
    existing_column_count: usize,

    // The configured limit.
    max_columns_per_table: usize,
}
```

And this block could be:

```suggestion
        let mut existing_columns = schema
            .tables
            .get(table_name)
            .map(|t| t.column_names())
            .unwrap_or_default();

        let existing_column_count = existing_columns.len();
        let merged_column_count = {
            existing_columns.append(&mut batch.column_names());
            existing_columns.len()
        };

        if merged_column_count > schema.max_columns_per_table {
            return Err(OverColumnLimit {
                table_name: table_name.into(),
                merged_column_count,
                existing_column_count,
                max_columns_per_table: schema.max_columns_per_table,
            });
        }
```

This test case will highlight the bad var usage, where `num_batch_columns` is currently asserting as being 5, despite the write containing only one new column:

```rust
        // Create a table that is at the column limit.
        {
            let table = namespace.create_table("bananas").await;
            table.create_column("greatness", ColumnType::I64).await;
            table.create_column("tastiness", ColumnType::I64).await;
            table.create_column("timestamp", ColumnType::Time).await;
            let schema = namespace.schema().await;
            // Adding columns over the limit is an error
            let batches = lp_to_writes("bananas i_got_music=42i 123456");
            assert_matches!(
                validate_column_limits(&batches, &schema),
                Err(OverColumnLimit {
                    table_name: _,
                    num_batch_columns: 1, // i_got_music
                    max_columns_per_table: 3
                })
            );
        }
```

router/src/dml_handlers/schema_validation.rs
```diff
@@ -253,17 +259,92 @@ where
...
+    #[tokio::test]
+    async fn validate_limits() {
+        let (_catalog, namespace) = test_setup().await;
+
+        namespace.update_column_limit(3).await;
+
+        // Table not found in schema,
+        let schema = namespace.schema().await;
+        // Columns under the limit is ok
+        let batches = lp_to_writes("nonexistent val=42i 123456");
```
domodwyer, line 318: Nice 42 👋

router/src/dml_handlers/schema_validation.rs
```diff
@@ -253,17 +259,92 @@ where
...
+        // Columns under the limit is ok
+        let batches = lp_to_writes("no_columns_in_schema val=42i 123456");
+        assert!(validate_column_limits(&batches, &schema).is_ok());
+        // Columns over the limit is an error
+        let batches = lp_to_writes("no_columns_in_schema,tag1=A,tag2=B val=42i 123456");
+        assert!(validate_column_limits(&batches, &schema).is_err());
+
+        // Table exists with a column in the schema,
+        let table = namespace.create_table("i_got_columns").await;
+        table.create_column("i_got_music", ColumnType::I64).await;
```
domodwyer, line 336: Looks like a missing shadowing of `schema` to use the new table/columns:

```suggestion
        table.create_column("i_got_music", ColumnType::I64).await;
        let schema = namespace.schema().await;
```

Perhaps worth scoping the tests to avoid re-use:

```rust
        // Table exists with a column in the schema,
        {
            let table = namespace.create_table("i_got_columns").await;
            table.create_column("i_got_music", ColumnType::I64).await;
            let schema = namespace.schema().await;
            // Columns already existing is ok
            let batches = lp_to_writes("i_got_columns i_got_music=42i 123456");
            assert!(validate_column_limits(&batches, &schema).is_ok());
            // Adding columns under the limit is ok
            let batches = lp_to_writes("i_got_columns,tag1=A i_got_music=42i 123456");
            assert!(validate_column_limits(&batches, &schema).is_ok());
            // Adding columns over the limit is an error
            let batches = lp_to_writes("i_got_columns,tag1=A,tag2=B i_got_music=42i 123456");
            assert_matches!(
                validate_column_limits(&batches, &schema),
                Err(OverColumnLimit {
                    table_name: _,
                    num_batch_columns: 4, // tag1, tag2, i_got_music & timestamp
                    max_columns_per_table: 3
                })
            );
        }
```

domodwyer (changes requested):
Looking good! 

Might be worth documenting the column limit enforcement behaviour in the `SchemaValidator` [docs](https://github.com/influxdata/influxdb_iox/blob/755453385c5ebc749b7d17b7c93e648c89cdf23f/router/src/dml_handlers/schema_validation.rs#L41-L85), especially the expected raciness of this impl (which is a fine trade-off 👌) so we can point people at it when they ask "wat do".

Worth noting it's more racy than it first appears, because when requests are spread over 6 routers, it is calculated across 6 caches which are eventually consistent - the "overshoot" of this impl increases with the number of routers. https://github.com/influxdata/influxdb_iox/issues/5005 will help reduce this overshoot.

router/src/dml_handlers/schema_validation.rs
carols10cents: Yuuuuup, and laziness on my part just asserting `is_err` 😪

iox_catalog/src/lib.rs
```diff
@@ -144,43 +144,48 @@ where
...
-                    column_type: ColumnType::from(col.influx_type()),
-                });
             }
-        };
-    }
+        })
+        .collect::<Result<Vec<_>>>()?
+        .into_iter()
+        .flatten()
+        .collect();
```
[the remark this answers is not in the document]
carols10cents: Oh good call! My brain gets lost in type soup sometimes...

carols10cents:
@domodwyer Great comments, thank you!! I think this is ready for rereview :)

carols10cents:
Putting this back to draft-- Dom and I chatted and I'm going to change the "broken table" behavior in this PR :)

carols10cents:
@domodwyer two new commits for you to get rid of the broken table behavior!

domodwyer:
Awesome! Thank you - I'll review this tomorrow morning my time :+1:

router/src/dml_handlers/schema_validation.rs
```diff
@@ -420,13 +425,68 @@ mod tests {
...
+                .expect("request should succeed");
+            let writes2_add_column = lp_to_writes("dragonfruit,tag2=B val=43i 123457");
+            handler2
+                .write(&*NAMESPACE, writes2_add_column, None)
+                .await
+                .expect("request should succeed");
+
+            let schema = namespace.schema().await;
+
+            // Columns already existing is an error, this table is in a broken state
```
domodwyer: Nice TDD 👍

domodwyer (approved):
Fantastic stuff 🎉 

Thanks for the PR etiquette, pushing a couple of commits instead of rewriting the old ones - makes reviewing additive changes much easier :+1:

PR #5826, opened 2022-10-10
refactor: use TableName, not Arc<str>

domodwyer:
A quick dev life change - use a typed table name, rather than passing around a string.

This reflects the `NamespaceName` equivalent.

---

* refactor: use TableName, not Arc<str> (97c6e0f8c)

      Adds a type wrapper TableName, internally an Arc<str> to leverage the
      type system instead of passing around untyped strings.


alamb (approved):

cannonpalms (approved):
I'm finally starting to be able to understand these. I seek out your commits because of how helpful the messages are. ♥️

PR #5831, opened 2022-10-11
refactor(ingester): remove tombstone support and delete tests from `query_tests`

### LINKED ISSUE #5825: Deprecate tombstone support in Ingester
As a group, we have decided to temporarily de-prioritise predicate-based delete functionality (retention-based deletes will be a supported feature, and is not affected by this ticket).

As a group we also agreed to remove the temporarily unused tombstone/delete code paths.

This ticket tracks me removing tombstones in the ingester.

domodwyer:
This is a big one!

This PR removes the `DmlDelete` handling code in the ingester, and any tests that depended on it. This is currently unused, adds significant complexity, and the impl would be changing significantly (like, completely) in the short-term future anyhow.

Please note **the code is not gone** - it lives on in the git history - it's just no longer a maintenance and complexity burden.

Closes #5825.

---

* refactor(ingester): remove tombstone support (c4f542bbe)

      This commit removes tombstone support from the ingester, and deletes
      associated code/helpers/tests. This commit does NOT remove tombstone
      support from any other service, but MAY include removing overlapping
      test coverage.
      
      This also removes the tombstone support from the Ingester -> Querier RPC
      response message.
      
      This has the nice side effect of removing a whole lot of thread spawning
      in the ingester tests for the Executor, speeding everything up!


alamb (commented):
I plan to review the changes to query_tests in detail later today

jacobmarble (approved):
LGTM. It's not an easy decision to remove code that supports delete operations, but after several conversations, I've concluded that it's the best decision.

NGA-TRAN (approved):
Since we remove the code to support delete, it makes sense to remove all the tests. We just have to accept we need to rewrite all the tests when we add deletes back in. 
I do not think we will reuse the removed tests even though they are still in GitHub, they are useful and we want to reuse them.

PR #5833, opened 2022-10-11
Revert "chore(deps): Bump object_store from 0.5.0 to 0.5.1"

alamb:
@NGA-TRAN has seen some weirdness on our production system that may be correlated to the upgrade in https://github.com/influxdata/influxdb_iox/pull/5827

I propose reverting  933493fab3b07ada4005f4783d820bd5952cca94 to see if that stops the issue


alamb:
We are going to try another approach here;

PR #5837, opened 2022-10-11
feat: add file size into estimated memory

### LINKED ISSUE #5835: Count file size in memory budget
# Rationale

Per discussion with @alamb recently, when we compact a file, we first read all of the file into memory. Then when we run DF plan, for each column, DF/arrow allocates `8K * arrow-col-size` for that column. Hence the memory estimate should include both those memory. Our recent change includes only the latter one. This ticket us to add the former

NGA-TRAN:
Closes #5835 


compactor/src/cold.rs
```diff
@@ -937,7 +937,7 @@ mod tests {
         let mut config = make_compactor_config();
 
         // Set the memory budget such that only some of the files will be compacted in a group
-        config.memory_budget_bytes = 1050;
+        config.memory_budget_bytes = 10000;
```
NGA-TRAN, line 940: Since we add file size into the estimation, we need to increase the budget for the test to work as expected

compactor/src/cold.rs
```diff
@@ -962,11 +959,10 @@ mod tests {
             .with_max_time(20)
             .with_creation_time(time_38_hour_ago)
             .with_compaction_level(CompactionLevel::FileNonOverlapped);
+        // p1 file size: 1757
         let pf1 = partition.create_parquet_file(builder).await;
-        size_overrides.insert(
-            pf1.parquet_file.id,
-            compactor.config.max_desired_file_size_bytes as i64 + 10,
-        );
+        println!("=== p1 file size: {:#?}", pf1.parquet_file.file_size_bytes);
```
NGA-TRAN, line 964: I keep this here on purpose in case the size is changed later

compactor/src/cold.rs
```diff
@@ -951,9 +951,6 @@ mod tests {
             Arc::clone(&metrics),
         ));
 
-        // parquet files that are all in the same partition
-        let mut size_overrides = HashMap::<ParquetFileId, i64>::default();
```
NGA-TRAN, line 955: Found that this is redundant in this test

compactor/src/parquet_file_lookup.rs
```diff
@@ -88,17 +88,20 @@ impl ParquetFilesForCompaction {
         let mut level_2 = Vec::with_capacity(parquet_files.len());
 
         for parquet_file in parquet_files {
+            // Estimate the bytes DataFusion needs when scan this file
             let estimated_arrow_bytes = partition.estimated_arrow_bytes(
                 min_num_rows_allocated_per_record_batch_to_datafusion_plan,
                 parquet_file.row_count,
             );
+            // Add bytes need to read this file into memory
+            let total_estimated_bytes = estimated_arrow_bytes + parquet_file.file_size_bytes as u64;
```
alamb, line 97: Since we are trying to be conservative, I suggest we consider estimating the RAM required to store the output parquet file as well. 

Ideally we'll make the plans fully streaming at some point, but for now they need to buffer the input parquet files in RAM as well as buffer the output files in RAM.

Thus, perhaps we might want to change this estimate to 

```suggestion
            // as well as the bytes needed to buffer the output
            let total_estimated_bytes = estimated_arrow_bytes + (2 * parquet_file.file_size_bytes as u64);
```

alamb (approved):
Looks good to me. I have another suggestion for tuning the estimate but this seems better than what is on `main` so 👍

NGA-TRAN:
Due to the increase in estimate memory, I need to update the tests

PR #5840, opened 2022-10-12
chore: remove vestigal DML handling from `ioxd_common`

alamb:
# Rationale 
The router handles all DML today (e.g. https://github.com/influxdata/influxdb_iox/blob/main/router/src/server/http.rs)  This DML handling logic appears to be left over from OG.

I found this code while trying to implement a TODO in predicate.rs I ran into while upgrading datafusion -- see https://github.com/influxdata/influxdb_iox/pull/5792


domodwyer (approved):
🧹

alamb:
Thank you for the reviews @domodwyer

PR #5841, opened 2022-10-12
refactor: Move http delete parsing code into router along side other http code

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/5840~

# Rationale
Implement a TODO from long ago and move HTTP specific code next to the http handlers


router/Cargo.toml
```diff
@@ -26,9 +26,11 @@ parking_lot = "0.12"
 predicate = { path = "../predicate" }
 schema = { version = "0.1.0", path = "../schema" }
 serde = "1.0"
+serde_json = "1.0.86"
 serde_urlencoded = "0.7"
 service_grpc_schema = { path = "../service_grpc_schema" }
 service_grpc_object_store = { path = "../service_grpc_object_store" }
+snafu = "0.7"
```
alamb, line 33: I think this is the most potentially controversial change in this PR

The rest of the router code uses a different error handling pattern / library (`thiserror`) than other parts of the system (`snafu`). So when I ported the code from elsewhere now router uses snafu

If this is a problem, I can rework the code to use thiserror, though I would rather not as I don't think the value of doing so is very high compared to the benefit.

predicate/src/delete_predicate.rs
```diff
@@ -35,30 +33,6 @@ pub enum Error {
     /// Predicate include non supported expression
     #[snafu(display("Delete predicate must be conjunctive expressions of binary 'column_name = literal' or 'column_name != literal': ({})", value))]
     NotSupportPredicate { value: String },
-
```
alamb, line 38: This code and test is just all mover

domodwyer (approved):
We could also just gift it to the history of git, given recent developments.

alamb:
> We could also just gift it to the history of git, given recent developments.

Yeah, if I had to do anything more than move it around I will do so... Sometimes it is so hard to https://www.youtube.com/watch?v=FnpJBkAMk44

PR #5844, opened 2022-10-12
revert: "revert: rdkafka/rskafka swapping (#5800)"

carols10cents:
This reverts commit b77c354.

The free-form options accepted for `INFLUXDB_IOX_WRITE_BUFFER_CONNECTION_CONFIG` are different between rdkafka and rskafka. I reproduced the error that deploying rdkafka caused, then translated the options k8s-idpe sends from what rskafka used to what rdkafka expects so that no changes to k8s-idpe should be needed.


clap_blocks/src/write_buffer.rs
```diff
@@ -203,4 +203,40 @@ mod tests {
...
+            "kafka",
+            "--write-buffer-addr",
+            "localhost:1234",
+            "--write-buffer-connection-config",
+            "max_message_size=10485760,\
+            producer_max_batch_size=2621440,\
+            consumer_min_batch_size=1048576,\
+            consumer_max_batch_size=5242880,\
+            consumer_max_wait_ms=10",
+        ])
```
domodwyer, line 221: Nice idea - love 👏

domodwyer:
I have rebased this PR onto origin/main :+1:

PR #5845, opened 2022-10-12
refactor: split memory estimation into bytes to store and bytes to stream

NGA-TRAN:
No changes in functionality of this PR. I need to split memory estimation into 2 parts: one for storing the file and one for streaming it in a query plan. The second part will be used to add estimation each output stream needs for the work in https://github.com/influxdata/conductor/issues/1199 that I am doing next


alamb (approved):
Looks good 👍 

As this memory estimate gets more sophisticated, I think we should at some point consider breaking it out into its own crate (so that the querier and ingester can reuse it)

alamb:
To keep the merge queue down I added the "automerge" label -- hope that is ok

PR #5847, opened 2022-10-12
feat: add memory need for output streams into our estimation

NGA-TRAN:
Closes https://github.com/influxdata/conductor/issues/1199


compactor/src/parquet_file_filtering.rs
```diff
@@ -1204,7 +1401,7 @@ mod tests {
                 column_set: ColumnSet::new(std::iter::empty()),
             };
             // Estimated arrow bytes for one file with a tag, a time and 11 rows = 1176
-            CompactorParquetFile::new(f, 1176, 0)
+            CompactorParquetFile::new(f, ESTIMATED_STREAM_BYTES, (file_size_bytes * 2) as u64)
```
NGA-TRAN, line 1404: This is the setup of the tests above
- Replace 0 with `(file_size_bytes * 2)` to have the test include the bytes to store file in memory
- ESTIMATED_STREAM_BYTES is just a constant of 1176

compactor/src/parquet_file_filtering.rs
```diff
@@ -450,19 +508,83 @@ mod tests {
         let level_1 = vec![];
         let (files_metric, bytes_metric) = metrics();
 
+        let expected_estimated_budget = (2 * level_0[0].file_size_bytes())  as  u64 // bytes stored the file in memory
+            + 2 * ESTIMATED_STREAM_BYTES; // one input stream, one output stream
```
NGA-TRAN, line 512: This is the new output of the tests. Before this PR, it is simply `n * 1176`
 This PR does 2 things for these set of tests:
1. Change the setup to test the bytes stored in memory added in earlier PR (It was tested with that PR but I want to have more thorough tests here).  The `2 * level_0[0].file_size_bytes()` is for that.
2. Include output stream estimation added in this PR.

compactor/src/parquet_file_filtering.rs
```diff
@@ -42,6 +42,7 @@ pub(crate) enum FilterResult {
 ///
 /// The returned files will be ordered with the level N + 1 files first, then the level N files
 /// in the same order as they were in the input.
+#[allow(clippy::too_many_arguments)]
```
alamb, line 45: This logic is getting complicated enough it might be worth considering refactoring it into smaller structures somehow

compactor/src/parquet_file_filtering.rs
```diff
@@ -425,6 +479,9 @@ mod tests {
 
     const MEMORY_BUDGET: u64 = 1024 * 1024 * 10;
     const FILE_NUM_LIMIT: usize = 20;
+    const MAX_OUTPUT_SIZE: u64 = 1024 * 1024 * 1024;
```
alamb, line 482: I suggest calling this `DEFAULT_MAX_OUTPUT_SIZE` as a way to help describe it more (I was a little confused at first before I read the tests in detail)

alamb (approved):
makes sense to me -- thanks @NGA-TRAN

compactor/src/parquet_file_filtering.rs
NGA-TRAN: Agree. Let me have this in to see whether we can well control the current OOM and millwork on this

Issue #5836, opened 2022-10-11
Optimize full compaction


NGA-TRAN:
# Rationale

Since we now also limit the size of L2 ([See 1198](https://github.com/influxdata/conductor/issues/1198#event-7564937309)), the full compaction should 

1. upgrade L1 to L2 if the L1 is large enough
2. limit number of L1 input files if their total > limit L2 size

PR #5849, opened 2022-10-13
test: Add verification that aspects of L1 -> L2 compaction are working as expected

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/5836.

@NGA-TRAN as far as I can tell, the code is already doing the two items listed in #5836: 

1. If there is only one L1 file in a group because by itself it takes up the memory budget, upgrade its level to L2 without running compaction
2. Limit the L1 files chosen as input to a compaction round if their memory budget is greater than the limit

I've added tests to verify these, am I missing something or misunderstanding the issue description?

### DIFF
--- a/compactor/src/cold.rs
+++ b/compactor/src/cold.rs
@@ -910,7 +910,7 @@ mod tests {
         let mut config = make_compactor_config();
 
         // Set the memory budget such that only one of the files will be compacted in a group
-        config.memory_budget_bytes = 4600;
+        config.memory_budget_bytes = 4_900;
 
         let metrics = Arc::new(metric::Registry::new());
         let compactor = Arc::new(Compactor::new(
@@ -1015,7 +1015,7 @@ mod tests {
         let mut config = make_compactor_config();
 
         // Set the memory budget such that two of the files will be compacted in a group
-        config.memory_budget_bytes = 9_200;
+        config.memory_budget_bytes = 9_800;
 
         let metrics = Arc::new(metric::Registry::new());
         let compactor = Arc::new(Compactor::new(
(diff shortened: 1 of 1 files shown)


NGA-TRAN:
@carols10cents I am reviewing this because they are good tests but the idea of the ticket is we do not want to reply on memory budget for this.

Let say if our memory budget is 20GB (to compact many partitions in parallel) but our max_desired_file_size is 100MB. We want to:
1. upgrade all L1 that are >  `X% of max_desired_file_size` to L2 without compacting. X can be 90% or so.
2. If we have many contiguous L1s each < `X% of max_desired_file_size`, we only need to get the **first L1s** whose total size just > max_desired_file_size and compact them. 

The idea is since max size of L2 is now also around max_desired_file_size which is also the limit of L1, we do not want to waste time to compact a lot of L1 and then split them to have small L2s. To answer your specific questions:


> If there is only one L1 file in a group because by itself it takes up the memory budget, upgrade its level to L2 without running compaction

We want to change the code to replace `by itself it takes up the memory budget` with `by itself it takes up nearly max_desired_file_size`

> Limit the L1 files chosen as input to a compaction round if their memory budget is greater than the limit

We want to replace this `greater than the limit` with `greater than max_desired_file_size` 

I am reviewing the code now

carols10cents:
Ahaaa, I think my confusion came from #5836 not saying `max_desired_file_size` anywhere 😅 I understand now, thank you!

NGA-TRAN:
Sorry I should have described it clearer 😄

NGA-TRAN (approved):
The tests look good

PR #5855, opened 2022-10-13
chore: Update datafusion again

alamb:
# Rationale
* keep up with dependencies (I want to be ready for the predicate pruning soon -- https://github.com/influxdata/conductor/issues/1192 )

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/0aa050a1a02438c67bb85f7e95fde521e80c640b
3. Updates for API changes (mostly update include locations due to https://github.com/apache/arrow-datafusion/pull/3816)


compactor/src/parquet_file_combining.rs
```diff
@@ -6,7 +6,7 @@ use data_types::{
     CompactionLevel, ParquetFile, ParquetFileId, ParquetFileParams, PartitionId, SequenceNumber,
     TableSchema, TimestampMinMax,
 };
-use datafusion::{error::DataFusionError, logical_plan::LogicalPlan};
+use datafusion::{error::DataFusionError, logical_expr::LogicalPlan};
```
alamb, line 9: Most of this PR's changes is that https://github.com/apache/arrow-datafusion/pull/3816 removed this stub and thus we have to pay some tech debt that has accumulated when DataFusion split into several crates

query_tests/cases/in/pushdown.expected
```diff
@@ -135,21 +135,21 @@
...
-|               |                                                                                                                                                                                                                                                                                                                                                                             |
-+---------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
++---------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
alamb, line 138: This is the non `use` change -- the change is that some internal aliasing (within an expr) have been removed in the filter, which is reasonable.

NGA-TRAN: Yeah, the alias in the filter is unnecessary

NGA-TRAN (approved):
LGTM

alamb:
> Yeah, the alias in the filter is unnecessary

Was removed in https://github.com/apache/arrow-datafusion/pull/3796

PR #5856, opened 2022-10-13
feat: Limit L1 -> L2 compaction based on file size

### LINKED ISSUE #5836: Optimize full compaction

carols10cents:
Closes #5836.

If we're filtering files for L1 -> L2 compaction, take the size of files selected so far into account and compare against the `max_desired_file_size_bytes` configuration.

If one L1 file by itself is over `max_desired_file_size_bytes`, return only that file from filtering, then when it goes to [`compact_one_partition` its compaction level will be upgraded in the catalog](https://github.com/influxdata/influxdb_iox/blob/9134ccd6c3c043e67d1193d3d852bee1d431f965/compactor/src/lib.rs#L371-L379) rather than read in and written out.

When considering each L1 in turn, stop accumulating the group once the sum of their file sizes exceeds `max_desired_file_size_bytes`.


compactor/src/parquet_file_filtering.rs
```diff
@@ -162,8 +162,9 @@ fn filter_parquet_files_inner(
             ln_estimated_file_bytes + current_ln_plus_1_estimated_file_bytes.iter().sum::<u64>();
 
         // Estimate memory needed for the output streams if the files are large enough to split
-        // Get total size of ln files (level_n_file and level_n_to_return) and ln+1 files (overlaps and files_to_return)
-        let total_fize_size_bytes = level_n_file.file_size_bytes()
+        // Get total size of ln files (level_n_file and level_n_to_return) and ln+1 files (overlaps
+        // and files_to_return)
+        let total_file_size_bytes = level_n_file.file_size_bytes()
```
NGA-TRAN, line 167: Good catch 😄

compactor/src/parquet_file_filtering.rs
```diff
@@ -212,9 +213,15 @@ fn filter_parquet_files_inner(
                 // Only compact the ones under the given budget
                 break;
             }
+        } else if compaction_level_n == CompactionLevel::FileNonOverlapped
+            && total_file_size_bytes as u64 > max_desired_file_size_bytes
```
NGA-TRAN, line 217: I think if this happens to the very first L1, break here will lead to returning empty `files_to_return` which is not what we want. We want to return that first L1 and its overlapped L2s. I think there must be some work here before the break

compactor/src/parquet_file_filtering.rs
```diff
@@ -1320,6 +1337,100 @@ mod tests {
...
+
+        let (files_metric, bytes_metric) = metrics();
+        let filter_result = filter_parquet_files_inner(
+            level_1,
+            vec![],
+            // enough to not limit based on this criteria
+            (ESTIMATED_STREAM_BYTES * 4) + (FILE_SIZE_BYTES as u64 * 3 * 2) + 1,
+            FILE_NUM_LIMIT,
+            // Limit to only 1 of the 2 files based on file size
+            FILE_SIZE_BYTES as u64 + 1,
```
NGA-TRAN, line 1365: I think if you add one more test that sets this value to `FILE_SIZE_BYTES as u64 - 1`, you will hit the bug above that returns nothing while the correct behavior is returning the first L1

NGA-TRAN (commented):
I think there is a bug if the first L1 (and its overlaped L2s) larger than max_desire_file_size

carols10cents:
@NGA-TRAN One new commit ready for you to review!

NGA-TRAN (approved):
Looks good

PR #5857, opened 2022-10-14
fix: Address InfluxQL parsing issues

### LINKED ISSUE #5848: InfluxQL: Fails to parse statements from Cloud 2 query log
A log of existing production queries was used to validate the completeness of the Rust parser. 

| Status | Count | Comment |
| :----- | :---- | :---- |
| success | 2251910 | Statements successfully parsed |
| parse errors | 563 | Queries that failed to parse (some are expected, such as `CREATE DATABASE` or `SELECT INTO`) |

This issue records the statements that failed to parse, but are expected to succeed for compatibility.

# Failed statements

- [x] `<>` is an alternative for the not equal operator `!=` 

- [x] `COUNT(DISTINCT(val))` fails

Example query:

```SQL
SELECT COUNT(DISTINCT(value)) FROM measurement
```

The InfluxQL parser fails to parse a `SELECT` statement containing a call with a nested `DISTINCT` call, such as `COUNT(DISTINCT(value))`. The issue can be replicated with the following patch:

```patch
diff --git a/influxdb_influxql_parser/src/select.rs b/influxdb_influxql_parser/src/select.rs
--- a/influxdb_influxql_parser/src/select.rs	(revision fcec161697d75e29650cb27593e6e3994b7052c1)
+++ b/influxdb_influxql_parser/src/select.rs	(date 1665622960197)
@@ -789,6 +789,16 @@
             }
         );
 
+        // Parse a call with a nested distinct call
+        let (_, got) = Field::parse("COUNT(DISTINCT(foo))").unwrap();
+        assert_eq!(
+            got,
+            Field {
+                expr: call!("COUNT", call!("DISTINCT", var_ref!("foo"))),
+                alias: None
+            }
+        );
+
         // Parse a wildcard
         let (_, got) = Field::parse("*").unwrap();
         assert_eq!(
```

- [x] `TIME(5s + 10)` fails to parse

InfluxQL parses any expression within the `TIME` function, however, it is re-validated during query processing to ensure it results in a valid duration expression. The following demonstrates the types of queries that are valid vs invalid.

```
> select count(value) from cpu group by time(10)
ERR: time dimension must have duration argument
> select count(value) from cpu group by time(10s + 10)
ERR: time dimension must have duration argument
> select count(value) from cpu group by time(10s * 10) FILL(none)
name: cpu
time                count
----                -----
1661502000000000000 1
1661503400000000000 1
1661834000000000000 2
1663572700000000000 1
> select count(value) from cpu group by time(10s + 10ns) FILL(none)
name: cpu
time                count
----                -----
1661501991661501990 1
1661503421661503420 1
1661834021661834020 1
1661834031661834030 1
1663572781663572780 1
>
```

- [x] Conditional followed by `ORDER BY` fails to parse

Example query:

```SQL
SELECT "processEventHistory" FROM "PACS@E4751E045670" WHERE valuestreamname = 'PacsValueStream' AND time >= $0 AND time <= $1 ORDER BY time ASC LIMIT 10000
```

- [x] No spaces preceding `GROUP BY`

Example query:

```sql
SELECT non_negative_difference(max("value")) AS "total_cans" FROM "counter" WHERE ("line_name" = 'Line C' AND "machine_name" = 'Filler' AND "tag_name" = 'Counter_per_shift') AND time >= '2022-07-31 17:00:00' and time <= now()GROUP BY time(1m)
```

- [x] Quoted `time` in ORDER BY fails to parse

Example query

```sql
SELECT value FROM cpu ORDER BY "time" DESC
```

- [x] Digits preceding decimal point for float literals are optional

Example query

```sql
SELECT value - .3 FROM cpu
```

# Unsupported statements

* `CREATE DATABASE`
* `SELECT .. INTO`

# Remaining that fail

* Comments, which will be addressed in #5663

stuartcarnie:
Closes #5848

Teach the InfluxQL parser numerous fixes to address the parsing issues identified in #5848 

# `keyword` module improvements

## Better keyword matching

Replaced the `tag_no_case` combinator to a specialised `keyword` combinator, which addresses the eager matching of shorter tokens, such as `OR` from `ORDER`, when parsing a conditional expressions. This does so by always peeking after matching the token that a valid keyword separator token is matched, such as whitespace or a parenthesis.

## Cleaner `sql_keyword` matching

Replaced the existing multitude of `alt` combinators to match any keyword with a more efficient and easier to read combinator that looks up the keyword in a `BTreeSet`. A specialised `Token` type performs case-insensitive matches when looking up items in the `BTreeSet` without requiring allocations (such as is necessary when calling `to_lowercase`):

https://github.com/influxdata/influxdb_iox/blob/679818c488529d96b8387323df257591aca38e04/influxdb_influxql_parser/src/keywords.rs#L33-L34

The previous implementation, using `alt` combinators, resulted in a sequential execution of each `tag_no_case` combinator, until a match was found or through to the end. This is terribly inefficient, and occurs each time an `identifier` is parsed.

It was noted that with this improvement, parsing the 2.25 million SQL statements improved by about 15%, or 13 seconds down to 11 seconds. 

> ❗**NOTE:** 
> The original implementation was inspired by the [`nom-sql`](https://github.com/ms705/nom-sql/blob/23e2b18c600a3537dcc553f7efdda8fd2cd5f20b/src/keywords.rs#L26) crate, before I had learnt more about `nom`.


influxdb_influxql_parser/src/select.rs
```diff
@@ -1013,19 +1092,23 @@ mod test {
...
-            "invalid TIME call, expected duration"
-        );
+        // assert_expect_error!(
+        //     time_call_expression("TIME(3)"),
+        //     "invalid TIME call, expected duration"
+        // );
+        // assert_expect_error!(
+        //     time_call_expression("TIME(5m + 3)"),
+        //     "invalid TIME call, expected duration"
+        // );
```
domodwyer, line 1111: Accidental?

influxdb_influxql_parser/src/keywords.rs
```diff
@@ -20,126 +26,144 @@ fn keyword_follow_char(i: &str) -> ParseResult<&str, &str> {
...
+impl<'a> Ord for Token<'a> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        for (l, r) in self.0.chars().zip(other.0.chars()) {
+            match l.to_ascii_uppercase().cmp(&r.to_ascii_uppercase()) {
+                Ordering::Equal => continue,
+                v => return v,
+            }
+        }
+        self.0.len().cmp(&other.0.len())
+    }
```
domodwyer, line 64: If this impl is purely to satisfy the `Ord` bounds so they can be placed into a `BTreeSet` you could use the [`HashSet`](https://doc.rust-lang.org/std/collections/struct.HashSet.html) which doesn't require an `Ord` bound on the items within it.

I doubt it'll make any meaningful performance difference though.

domodwyer (approved):
This looks like a great improvement to me!

How are you measuring the perf. changes? I didn't see any benchmarks checked-in within the `influxdb_influxql_parser` crate. You should check out [`criterion`](https://bheisler.github.io/criterion.rs/book/index.html)!

influxdb_influxql_parser/src/select.rs
stuartcarnie: Ahh yes, good spot! I'll remove

influxdb_influxql_parser/src/keywords.rs
stuartcarnie: Good tip, thanks! 

`HashSet` requires the elements implement the `Hash` trait, but overall is simpler, so I've switch to that and updated the tests accordingly.

stuartcarnie:
> How are you measuring the perf. changes? I didn't see any benchmarks checked-in within the `influxdb_influxql_parser` crate. You should check out [`criterion`](https://bheisler.github.io/criterion.rs/book/index.html)!

I didn't use benchmarks at this point, just measured a program I wrote to parse some InfluxQL query logs, which contain InfluxQL queries embedded in JSON, which is embedded in CSV:

```rust
use influxdb_influxql_parser::parse_statements;
use rayon::prelude::*;
use serde::Deserialize;
use serde_json::{Map, Value};
use std::error::Error;
use std::sync::atomic::Ordering;

#[derive(Deserialize)]
struct QueryRequest {
    query: String,
}

#[derive(Deserialize)]
struct QueryResult {
    #[serde(rename = "Error")]
    error: Option<Map<String, Value>>,
    #[serde(rename = "QueryRequest")]
    query_request: QueryRequest,
}

fn main() -> Result<(), Box<dyn Error>> {
    let f = std::fs::File::open(
        "/Users/stuartcarnie/projects/go/play/2022-08-10T10:36-1h prod_influxql_log.csv.gz",
    )?;
    let r = flate2::read::GzDecoder::new(f);

    let mut rdr = csv::ReaderBuilder::new().from_reader(r);
    let headers = rdr.headers()?;
    println!("{:?}", headers);

    let success = std::sync::atomic::AtomicU32::new(0);
    let query_errors = std::sync::atomic::AtomicU32::new(0);
    let parse_errors = std::sync::atomic::AtomicU32::new(0);

    rdr.records().par_bridge().for_each(|op| {
        if let Ok(record) = op {
            let val = record.get(4).unwrap();
            if let Ok(qr) = serde_json::from_str::<QueryResult>(val) {
                if let Some(_) = qr.error {
                    query_errors.fetch_add(1, Ordering::AcqRel);
                    return;
                }

                match parse_statements(qr.query_request.query.as_str()) {
                    Ok(statements) => {
                        success.fetch_add(statements.len() as u32, Ordering::AcqRel);
                    }
                    Err(err) => {
                        println!("parse failed: {}", qr.query_request.query);
                        println!("error: {}", err);
                        parse_errors.fetch_add(1, Ordering::AcqRel);
                    }
                }
            }
        }
    });

    println!("success      : {}", success.load(Ordering::Acquire));
    println!("parse errors : {}", parse_errors.load(Ordering::Acquire));
    println!("query errors : {}", query_errors.load(Ordering::Acquire));

    Ok(())
}
```

PR #5860, opened 2022-10-14
refactor(iox_catalog): reduce default column limit

### LINKED ISSUE #5858: Change default column service protection limit
For a long time the default (but unenforced) limit on the number of columns in a single table has been 1,000.

As of https://github.com/influxdata/influxdb_iox/pull/5819 this limit is now enforced.

However the current DataFusion performance degrades (and resource utilisation increases) as the number of columns increases. In [this comment](https://github.com/influxdata/influxdb_iox/issues/4349#issuecomment-1277386300) @alamb recommended:

> I would personally recommend keeping it to 200 columns or less. I have seen IOx work (but struggle) with ~700 columns but for the best UX keep it lower.

This ticket is to change the default limit from 1,000 columns to 200 (or some other acceptable number).

domodwyer:
This PR changes the default value for the per-column service limit from 1,000 to 200.

**This change only affects new namespaces / org&buckets**.

It also beefs up some of the testing to make sure the code + database default stays in sync.

Closes #5858.

---

* refactor: reduce default column limit (46bbee542)

      Reduces the default number of columns allowed per-table, from 1,000 to
      200.

* test: assert default service limit values (e4179605d)

      Adds tests that assert the default service limit values.

### DIFF
--- a/iox_catalog/migrations/20221014122742_lower-default-per-table-column-limit.sql
+++ b/iox_catalog/migrations/20221014122742_lower-default-per-table-column-limit.sql
@@ -1,0 +1,7 @@
+-- Lower the defualt per-table column limit.
+--
+-- https://github.com/influxdata/influxdb_iox/issues/5858
+ALTER TABLE
+    namespace ALTER max_columns_per_table
+SET
+    DEFAULT 200;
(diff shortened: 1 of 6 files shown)


lukebond (approved):
thanks for improving those tests :+1:

mkmik (approved):

alamb (approved):

Issue #5716, opened 2022-09-21
calling mean() on a string column returns "internal error" instead of a meaningful message


rickspencer3:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behaviour.

1. You need data in a bucket where the _value is a string
2. Run code that uses mean() on that data

This example users Flux generated from the Data Explorer:

```javascript
// _value is a string, specifically a time stamp
from(bucket: "operating")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "task")
  |> aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)
//  querying a tsm bucket, the error message is: "unsupported input type for mean aggregate: string"
// querying an IOx bucket, the error message is: " An internal error has occurred"
```

__Expected behaviour:__
Users get back a sensible error message that they can action, such as  "unsupported input type for mean aggregate: string"

__Actual behaviour:__
Users get back a vague unactionable error

rickspencer3:
Note that, as expected, if you do not hit the storage engine, you get back a sensible error from Flux:

```import "array"javascript

array.from(rows: [{_value: "1"},{_value:"2"}])
  |> mean()
```

yields: ``` runtime error @4:6-4:12: mean: unsupported aggregate column type string```

alamb:
Added to the IOx project / lighthouse milestone
cc @jacobmarble

wolffcm:
So I have dug in on this at @aanthony1243's request. Here's what I have found.

When we make a call to `ReadAggregateWindow` (the storage RPC that corresponds to the query in the example), running with queryd connected to IOx in remocal, I see these errors from the `queryd` process in `iox-sticky-querier`:
```
ts=2022-10-13T20:18:39.567603Z
lvl=info 
msg="client returned stream that returned error" 
log_id=0dWdrbYl000 
service=storageclient 
method=WindowAggregate 
request-tag=1583838706 
partitionID=0 
node_id=0 
retryable=true 
error="rpc error: code = InvalidArgument desc = Error creating group plans for database '9c5955fc99a60b8f_95b31df6d1143287': read_window_aggregate: gRPC planner got error building plan: Error during planning: The function Sum does not support inputs of type Utf8.
caused by
Error during planning: The function Sum does not support inputs of type Utf8."
```
The code that generates this log is here:
[...]

PR #5864, opened 2022-10-14
fix: improve tonic status codes returned for query errors

alamb:
# Rationale

The cause of errors reported to users is often hidden for errors that are reported to be "internal". Thus we should make all errors that the user can do something about (e.g. bad sql) non internal so they can see what went wrong

Part of https://github.com/influxdata/influxdb_iox/issues/5716

# Changes
1. Translate more `DataFusion` errors into "bad request" rather than "internal"
2. Use consistent logic for influxrpc and SQL


service_common/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+        | DataFusionError::SchemaError(_)
+        | DataFusionError::Execution(_)
+        | DataFusionError::External(_)
+        | DataFusionError::ArrowError(_)
+        | DataFusionError::ParquetError(_)
+        | DataFusionError::NotImplemented(_)
+        | DataFusionError::Plan(_) => tonic::Code::InvalidArgument,
+        DataFusionError::Context(_, inner) => {
+            // traverse context chain
+            datafusion_error_to_tonic_code(inner.as_ref())
```
crepererum, line 20: I would prefer not to traverse the pointer chain using recursing -- which may blow up the stack.

service_common/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+    match e {
+        DataFusionError::ResourcesExhausted(_) => tonic::Code::ResourceExhausted,
+        // Map as many as possible back into user visible (non internal) errors
+        DataFusionError::SQL(_)
+        | DataFusionError::SchemaError(_)
+        | DataFusionError::Execution(_)
+        | DataFusionError::External(_)
+        | DataFusionError::ArrowError(_)
+        | DataFusionError::ParquetError(_)
+        | DataFusionError::NotImplemented(_)
```
crepererum, line 16: Shouldn't this be "unimplemented"?

service_common/src/error.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Converts a [`DataFusionError`] into the appropriate [`tonic::Code`]
+pub fn datafusion_error_to_tonic_code(e: &DataFusionError) -> tonic::Code {
+    match e {
+        DataFusionError::ResourcesExhausted(_) => tonic::Code::ResourceExhausted,
+        // Map as many as possible back into user visible (non internal) errors
+        DataFusionError::SQL(_)
+        | DataFusionError::SchemaError(_)
+        | DataFusionError::Execution(_)
+        | DataFusionError::External(_)
```
crepererum, line 13: "External" is defined as:

> Errors originating from outside DataFusion’s core codebase. For example, a custom S3Error from the crate datafusion-objectstore-s3

(https://docs.rs/datafusion/13.0.0/datafusion/error/enum.DataFusionError.html#variant.External)

So this is NOT always "invalid argument" and was the reason I've mapped it to "internal" (which would IMHO be less wrong).

Also, why are the following "invalid argument" errors?

- execution
- arrow
- parquet

crepererum (commented):
This error mapping interpretation looks a bit like a hack to me.

service_common/src/error.rs
alamb: Will change

service_common/src/error.rs
alamb: What I really want to achieve is "user sees the error message rather than an opaque message" -- in an ideal world this could be distinguished from the error code, but not always and so I would like to default to "show the user the message"  -- as I think allowing some internal error out is less of a problem than preventing some useful error out

> Also, why are the following "invalid argument" errors?

Basically because I wasn't sure they were all internal errors -- for example, you can get an Arrow error if you try and divide a column by zero, depending on the data.

alamb:
Thanks for the review @crepererum 

I think this PR needs some significant documentation on rationale which I will add. However, before I do any more work on this PR I am going to sort out #5716  with @wolffcm  to make sure messages do get back to the user, and then I will revisit this PR based on that knowledge.

crepererum:
To clarify: I'm OK with using a somewhat "weird" mapping (e.g. mapping "external" to "invalid argument") if the rational is documented.

service_common/src/error.rs
alamb: I reviewed all uses of `External` in the DataFusion codebase and they all seem to come from object_store access; Thus I changed `DataFusionError::External` to be classified as an Internal error as well

service_common/src/error.rs
alamb: Based on our conversations I have added a significant number of comments to explain the rational for these categorizations

service_common/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+        // that the client called an API endpoint that wasn't
+        // implemented.
+        //
+        // See examples in:
+        // https://github.com/apache/arrow-datafusion/search?q=NotImplemented
+        | DataFusionError::NotImplemented(_)
+        | DataFusionError::Plan(_) => tonic::Code::InvalidArgument,
+        e @ DataFusionError::Context(_,_) => {
+            // traverse context chain without recursion
+            datafusion_error_to_tonic_code(leaf_error(e))
```
alamb, line 56: I chose this construction so I didn't have to have a `unimplemented!()` match arm  but still avoid finding the leaf error via unbounded recursion (the recursion in this function is now limited to 2 stack frames)

carols10cents (approved):
This looks good-- either this PR or mine (https://github.com/influxdata/influxdb_iox/pull/5876) will need to be modified, whichever gets merged in second :)

service_common/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+        DataFusionError::ResourcesExhausted(_) => tonic::Code::ResourceExhausted,
+        // Map as many as possible back into user visible (non internal) errors
+        DataFusionError::SQL(_)
+        | DataFusionError::SchemaError(_)
+        // Execution, ArrowError and ParquetError might be due to an
+        // internal error (e.g. some sort of IO error or bug) or due
+        // to a user input error (e.g. you can get an Arrow error if
+        // you try and divide by a column and it has zeros).
+        //
+        // Since we are not sure they are all internal errors we
```
crepererum, line 35: Potential future TODO: we could inspect the error and decide. For `Box<dyn ...>` we could downcast the type and decide that IOx only puts a single concrete enum in there.

crepererum (approved):
thank you

service_common/src/error.rs
alamb: added in be5a1ba87

PR #5868, opened 2022-10-14
docs: compactor document and how to resolve OOMs

NGA-TRAN:
Document about how compactor works and how to resolve OOMs


docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+
+Compaction process in IOx is a background task that repeats selecting and compacting partitions with new files. Each repeat is called a `compaction cycle` that in turn includes N `hot cycles` and one `cold cycle`. N is configurable and the default is 4. 
+
+- A **Hot cycle** is responsible for compacting level-0 files and their overlapped level-1 files into non-overlapped level-1 files. Only partitions with at least two level-0 files that are ingested within 24 hours hours are selected to compact. In order to deduplicate data correctly, level-0 files with smaller sequence numbers are always compacted before the ones with larger sequence numbers. The compacted output will go to one or many non-time-overlapped level-1 each is estimated around the size of a configurable max-desired-file-size.
+- A **cold cycle** is responsible for compacting partitions that have no new level-0 files ingested the last 8 hours. A partition selected in cold cycle will go through two steps of compaction:
+    i. Step 1: similar to hot cycle, level-0 files will be compacted with their overlapped level-1 files to produce non-overlapped level-1 files.
```
lukebond, line 25: these sub dotpoints are not indenting. i think if you swap the `i.` and `ii.` for just `-` it will indent

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+    ii. Step 2: level-1 files will be compacted with their overlapped level-2 files to produce non-overlapped level-2 files. Max size of level-2 files is also configurable.
+
+The number of files in a partition selected to compact depends on the available memory. The number of partitions to be compacted concurrently in each cycle also depends on the compactor memory and is different in each cycle depending on the number of partitions needed to compact and the number and sizes of their compacting files. See below for how to estimate memory needed for each partition.
+
+In the case the Compactor cannot compact the smallest set of files of a partition, that partition will be put into a `skipped_compactions` table and won't get selected to compact again. A parttion that is not compacted will provide bad query performance or even cannot be queried if the Querier does not have enough memory to read and process their files. See below for the details of when this can happen, how to avoid it, and how to solve if that happens.
```
lukebond, line 30: s/parttion/partition/

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+
+# What to do if a Compactor hits OOMs
+
+Even though we have tried to avoid OOMs by estimating needed memory, it still happens in extreme cases. Currently, OOMs in compactor won't be resolved by themselves without human actions. The easiest way to stop OOMs is to increase memory. Doubling memory or even more is recommended. We can bring the memory down after all the high volumn partitions are compacted.
```
lukebond, line 34: perhaps add "this is because the compactor will restart and select the same files again that led to the OOM" after "without human action"?

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+
+If increasing memory a lot does not help, consider changing one or a combination of config parameters below but please be aware that by chooing doing this, you tell the compactor that it has a lot less memory budget to use and it will push one or many partitions into the `skipped_compactions`. Moreover, reducing memory budget also means reducing the concurrency capacity of the compactor. It is recommended that you do not try this unless you know the workload of the compactor very well.
```
lukebond, line 36: this is really good advice!

lukebond (changes requested):
i think this is a good start, looking forward to the rest! although i think it should be moved to our docs site where our runbooks live, see my message on slack.

docs/compactor.md
alamb, line 30:
```suggestion
In the case the Compactor cannot compact the smallest set of files of a partition due to memory limits, that partition will be put into a `skipped_compactions` table and won't be considered for compaction again. A partition that is not compacted will provide bad query performance or even cannot be queried if the Querier does not have enough memory to read and process their files. See below for the details of when this can happen, how to avoid it, and how to solve if that happens.
```

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+- INFLUXDB_IOX_COMPACTION_MIN_ROWS_PER_RECORD_BATCH_TO_PLAN: double or triple this value. This will tell the commpactor the compaction plans need more memory.
```
alamb, line 37:
```suggestion
- INFLUXDB_IOX_COMPACTION_MIN_ROWS_PER_RECORD_BATCH_TO_PLAN: double or triple this value. This will tell the compactor the compaction plans need more memory, and will reduce the number of files that can be compacted at once
```

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+- INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES: reduce this value in half or more. This tells the compactor to limit further the maximun number of files of a partition it can compact.
```
alamb, line 38:
```suggestion
- INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES: reduce this value in half or more. This puts a hard cap on the maximun number of files of a partition it can compact, even if its memory budget estimate would allow for more.
```

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+- INFLUXDB_IOX_COMPACTION_MEMORY_BUDGET_BYTES: reduce this value in half or more. This tells the compact its total budget is less so it will reduce the number of partition it can compact concurrenly or reduce number of files to be comapcted for a partition.
```
alamb, line 39:
```suggestion
- INFLUXDB_IOX_COMPACTION_MEMORY_BUDGET_BYTES: reduce this value in half or more. This tells the compact its total budget is less so it will reduce the number of partition it can compact concurrenly or reduce number of files to be compacted for a partition.
```

docs/compactor.md
```diff
@@ -0,0 +1,49 @@
...
+
+# Avoid and deal with files in skipped_compactions
+todo
+
+# Memory Estimation
+todo
+
+# Usual SQL to verify compaction status
+todo
```
alamb, line 48: These are great things to fill in

alamb (approved):
All of this content is `influxdb_iox` specific, and not specific to our cloud environment (for example, it was very applicable when I was running local performance tests), and thus I disagree with @lukebond that we should put it into the internal runbooks. Instead, I suggest we keep this content here and put a link in the runbooks to this content.

lukebond:
> All of this content is influxdb_iox specific, and not specific to our cloud environment (for example, it was very applicable when I was running local performance tests), and thus I disagree with @lukebond that we should put it into the internal runbooks. Instead, I suggest we keep this content here and put a link in the runbooks to this content.

as long as a responder has access to this info without having to hunt for it i don't have feelings about how. either the link to this MD goes in the alert or the runbook links to this. fine with me :+1:

alamb:
> as long as a responder has access to this info without having to hunt for it

100% agree

lukebond (approved):

PR #5869, opened 2022-10-15
chore: update datafusion

alamb:
# Rationale
* keep up with dependencies (I want to be ready for the predicate pruning soon)
* get https://github.com/apache/arrow-datafusion/pull/3810 and remove same code from IOx

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/fc5081d48ef59e39c1b353dd45fcd13af6186676


datafusion_util/src/lib.rs
```diff
@@ -42,33 +41,6 @@ use tokio::sync::mpsc::{Receiver, UnboundedReceiver};
 use tokio_stream::wrappers::{ReceiverStream, UnboundedReceiverStream};
 use watch::WatchedTask;
 
-/// Split an Expr up into its constituent parts
```
alamb, line 45: Use the one in datafusion instead

predicate/src/rpc_predicate.rs
```diff
@@ -229,11 +229,7 @@ fn normalize_predicate(
                 .and_then(|e| simplifier.coerce(e, Arc::clone(&df_schema)))
                 .map(|e| log_rewrite(e, "coerce_expr"))
                 // Call DataFusion simplification logic
-                .and_then(|e| {
-                    // simplify twice to ensure "full" cleanup
-                    let e = simplifier.simplify(e)?;
-                    simplifier.simplify(e)
-                })
+                .and_then(|e| simplifier.simplify(e))
```
alamb, line 232: DataFusion simplifier has gotten better (we don't need to call it twice)

Though to be honest, what I really did was upstream this change:  https://github.com/apache/arrow-datafusion/pull/3811 😆

crepererum (approved):

Issue #5720, opened 2022-09-22
Querier: improve refresh policy


crepererum:
# What
Improve refresh policy to use some exponential backoff model.

# Why
Currently we refresh a key after a `GET` operation with a fixed wait time. This has a number of drawbacks:

- **wait time estimation:** It is hard to estimate the wait time correctly, esp. for mixed throughput / access deployments.
- **single wait time:** There is only a single wait time and a single refresh. However a good rule of thumb is: the longer the time since the last `GET`, the less often we need to update the key.
- **correlation:** Updates tend to be correlated due to the fixed wait time.

We currently only refresh the `namespace` cache. However with #5718 the issues described here will be more prominent.

# How
Use the timing model that we already implemented in `backoff` and trigger recurring refreshs after a `GET` operation. Reset the backoff timer after a `GET`.

PR #5873, opened 2022-10-17
feat: harden and polish backoff system

crepererum:
I plan to use the backoff system for #5720 so I figured out that it needs some love. See commits.


domodwyer (approved):

backoff/src/lib.rs
```diff
@@ -201,6 +212,40 @@ impl Backoff {
     }
 }
 
+impl Iterator for Backoff {
```
alamb, line 215: This is nice

Issue #5802, opened 2022-10-05
Should return an error on unimplemented SQL statements


mkmik:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behaviour.

1. run the `drop table this_table_doesnt_exist` query

__Expected behaviour:__

An error.

__Actual behaviour:__

```console
$ ./target/debug/influxdb_iox query internal_test 'drop table this_table_doesnt_exist'
++
++
```

__Logs:__

Nothing gets logged

alamb:
I plan to fix this next week

carols10cents:
Ok, I have [tests reproducing this behavior](https://github.com/influxdata/influxdb_iox/compare/cn/error-on-unsupported-query) with both the `influxdb_iox query` CLI and with the Flight `perform_query` gRPC request, and I'm indeed not getting an error, but I'm not sure at what point the error should be raised.

When I run these tests with `LOG_FILTER=debug`, I see this from the datafusion query planning:

```
2022-10-17T15:20:18.000761Z  INFO service_grpc_flight: flight do_get db_name=0956239270112659_9864155902343535 sql_query=drop table this_table_doesnt_exist trace=
2022-10-17T15:20:18.028629Z DEBUG iox_query::exec::context: planning SQL query text=drop table this_table_doesnt_exist
2022-10-17T15:20:18.029891Z DEBUG iox_query::exec::context: logical plan plan=// Begin DataFusion GraphViz Plan (see https://graphviz.org)
digraph {
  subgraph cluster_1
  {
    graph[label="LogicalPlan"]
    2[shape=box label="DropTable: _this_table_doesnt_exist_ if not exist:=false"]
  }
  subgraph cluster_3
  {
    graph[label="Detailed LogicalPlan"]
    4[shape=box label="DropTable: _this_table_doesnt_exist_ if not exist:=false\nSchema: []"]
  }
}
// End DataFusion GraphViz Plan

2022-10-17T15:20:18.030354Z DEBUG iox_query::exec::context: create_physical_plan: initial plan text=DropTable: "this_table_doesnt_exist" if not exist:=false []
2022-10-17T15:20:18.030443Z DEBUG datafusion_optimizer::optimizer: Optimizer input:
DropTable: "this_table_doesnt_exist" if not exist:=false

// ...snip...

2022-10-17T15:20:18.031354Z DEBUG datafusion_optimizer::optimizer: Optimized plan:
[...]

PR #5876, opened 2022-10-17
fix: Return an error for unsupported SQL queries

carols10cents:
Closes https://github.com/influxdata/influxdb_iox/issues/5802.

Before, if you tried to query with the SQL `drop table foo`, you would see this:

```
$ ./influxdb_iox query company_sensors 'drop table foo'
++
++
$ echo $?
0
```

With this PR, you'll see this:

```
$ ./influxdb_iox query company_sensors 'drop table foo'
Error querying: Error while planning query: This feature is not implemented: DropTable
$ echo $?
1
```

I also took a look at the code where datafusion turns a logical plan into a physical plan, and saw that where it turns the `LogicalPlan::DropTable` variant into a `PhysicalPlan::EmptyExec` there were 3 other variants that have the same behavior: https://github.com/apache/arrow-datafusion/blob/b654fdea697b9aec1cb487e292dd288e5c9d09e3/datafusion/core/src/physical_plan/planner.rs#L1051-L1057

As far as I can tell IOx doesn't currently support create memory table, drop table, or drop view yet either, so I decided to handle them in the same way here as `DropTable`.

I also changed the `Display` implementation for `influxdb_iox_client::flight::Error::GrpcError`; without that change, the error would have looked like this:

```
Error querying: status: Unimplemented, message: "Error while planning query: This feature is not implemented: 
DropTable", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", 
"date": "Mon, 17 Oct 2022 20:32:39 GMT", "content-length": "0"} }
```

TMI, right???

I also chose to return `datafusion::error::DataFusionError::NotImplemented` rather than the `datafusion::error::DataFusionError::Internal` variant DataFusion returns in some other cases for other `LogicalPlan` variants, like `CREATE DATABASE` https://github.com/apache/arrow-datafusion/blob/b654fdea697b9aec1cb487e292dd288e5c9d09e3/datafusion/core/src/physical_plan/planner.rs#L1042-L1050

that we are relying on: https://github.com/influxdata/influxdb_iox/blob/b3b22cff3786b5a7bbe6776caef3674b81a34135/query_tests/src/sql.rs#L342-L364

but that would create different incorrect error messages that suggest it's a DataFusion bug that `DropTable` isn't implemented:

```
Error querying: Error while planning query: Internal error: Unsupported logical plan: DropTable. 
This was likely caused by a bug in DataFusion's code and we would welcome that you file 
an bug report in our issue tracker
```

but it's not, it's an IOx error. Maybe we should even match on `LogicalPlan::CreateCatalog` (and other variants that also return errors from datafusion) here too and return an IOxey error? I wasn't sure though, and I suspect this might interact with https://github.com/influxdata/influxdb_iox/pull/5864, I'm going to take a look at that next and maybe make a new PR to discuss those.


alamb:
> I also chose to return datafusion::error::DataFusionError::NotImplemented rather than the datafusion::error::DataFusionError::Internal variant DataFusion returns in some other cases for other LogicalPlan variants, like CREATE DATABASE https://github.com/apache/arrow-datafusion/blob/b654fdea697b9aec1cb487e292dd288e5c9d09e3/datafusion/core/src/physical_plan/planner.rs#L1042-L1050

I think this is correct

> I also took a look at the code where datafusion turns a logical plan into a physical plan, and saw that where it turns the LogicalPlan::DropTable variant into a PhysicalPlan::EmptyExec there were 3 other variants that have the same behavior: https://github.com/apache/arrow-datafusion/blob/b654fdea697b9aec1cb487e292dd288e5c9d09e3/datafusion/core/src/physical_plan/planner.rs#L1051-L1057


Looking at those references I think the right solution in DataFusion (eventually) to upstream this change to return errors (rather than an empty exec) by default. I can file that ticket if you don't disagree and I bet someone in the community will code it up

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -584,6 +584,45 @@ async fn write_and_query() {
     .await
 }
 
+/// Test error handling for the query CLI command
```
alamb, line 587: ❤️

iox_query/src/exec/context.rs
```diff
@@ -324,6 +321,24 @@ impl IOxSessionContext {
         debug!(text=%sql, "planning SQL query");
         let logical_plan = ctx.inner.create_logical_plan(sql)?;
         debug!(plan=%logical_plan.display_graphviz(), "logical plan");
+
+        // Handle unsupported SQL
+        match &logical_plan {
```
alamb, line 326: I double checked the list of variants here https://docs.rs/datafusion/13.0.0/datafusion/logical_plan/enum.LogicalPlan.html

I think we should be erroring on all of these: 
```
    CreateExternalTable(CreateExternalTable),
    CreateMemoryTable(CreateMemoryTable),
    CreateView(CreateView),
    CreateCatalogSchema(CreateCatalogSchema),
    CreateCatalog(CreateCatalog),
    DropTable(DropTable),
    DropView(DropView),
```

Though it looks like the others already make an error in DataFusion https://github.com/apache/arrow-datafusion/blob/b654fdea697b9aec1cb487e292dd288e5c9d09e3/datafusion/core/src/physical_plan/planner.rs#L1024-L1057

influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -46,7 +46,7 @@ pub enum Error {
 
     /// An unknown server error occurred. Contains the `tonic::Status` returned
     /// from the server.
-    #[error(transparent)]
+    #[error("{}", .0.message())]
```
alamb, line 49: 👍

alamb (approved):
Thanks @carols10cents  -- this is perfect. I think the longer term solution is to move the error checks into DataFusion, and I will file a ticket to do so (probably tomorrow).

Really nice work on the tests 🏅

alamb:
I filed https://github.com/apache/arrow-datafusion/issues/3873 to track the issue in DataFusion

iox_query/src/exec/context.rs
alamb: Tracked in https://github.com/apache/arrow-datafusion/issues/3873

alamb:
I am working on updating this PR to resolve logical conflicts with https://github.com/influxdata/influxdb_iox/issues/5802

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -455,6 +455,55 @@ async fn issue_4631_b() {
...
+    test_helpers::maybe_start_logging();
+    let database_url = maybe_skip_integration!();
+
+    // Set up the cluster  ====================================
+    let mut cluster = MiniCluster::create_shared(database_url).await;
+
+    StepTest::new(
+        &mut cluster,
+        vec![
+            Step::WriteLineProtocol("this_table_does_exist,tag=A val=\"foo\" 1".into()),
```
alamb, line 469: 👩‍🍳 👌 


Very nice

PR #5879, opened 2022-10-17
feat: read from parquet and only parquet

alamb:
Closes https://github.com/influxdata/conductor/issues/1002

As discussed internally, this PR  switces IOx immediately to query directly (and only) from (cached) parquet, rather than first loading into the RUB. This is the change we tested in [influxdb_iox/#/5866](https://github.com/influxdata/influxdb_iox/pull/5866). 

This change will:
* Allow more queries to run, especially those on large datasets
* Avoid the massive initial delay on first query (aka the “large tail latency”) while data is loaded into the RUB
* Decrease performance slightly, for pre-warmed queries with subsecond latency. 

We plan to regain (and then some) the lost performance in several follow on mini-projects, for which I will be filing tickets tomorrow.


tustvold (approved):

crepererum (approved):

crepererum:
:shipit:

PR #5888, opened 2022-10-18
refactor: remove querier chunk load settings

crepererum:
We no longer use dual-state ReadBuffer/Parquet chunks.

re https://github.com/influxdata/conductor/issues/1002


querier/src/cache/mod.rs
```diff
@@ -49,6 +49,7 @@ pub struct CatalogCache {
     tombstone_cache: TombstoneCache,
 
     /// Read buffer chunk cache
+    #[allow(dead_code)]
```
alamb, line 52: are you planning to remove these as part of a follow on PR?

alamb (approved):

querier/src/cache/mod.rs
crepererum: yes

PR #5890, opened 2022-10-18
refactor: remove querier dual chunk stage

crepererum:
We're parquet-only now.


querier/Cargo.toml
```diff
@@ -26,7 +26,6 @@ pin-project = "1.0"
 predicate = { path = "../predicate" }
 iox_query = { path = "../iox_query" }
 rand = "0.8.3"
-read_buffer = { path = "../read_buffer" }
```
alamb, line 29: oh boy

alamb (approved):

querier/Cargo.toml
crepererum: :drum:

PR #5891, opened 2022-10-18
refactor: remove read buffer

crepererum:
:wave: 

re https://github.com/influxdata/conductor/issues/1002


alamb (approved):

aierui:
@crepererum Why? and can you add a more detailed context here.

I think loading data into memory reduces the latency of queries, which is a common space-for-time practice.

In the NG architecture, in order for memory to be fully utilized, it is necessary to ensure that SQL follow static routing rules to queries instances.

crepererum:
@aierui The short version is: we cache data in-mem, but we'll use the parquet format for that instead of hand-rolling our own system. Parquet deserialization should be fast enough once @tustvold improves/enables predicate pushdown. This also makes the code more future proof and more maintainable. Plus: loading data into the read buffer (from parquet) is expensive, which complicates cache management.

alamb:
👍 

basically we believe we can get basically the same performance using https://github.com/apache/arrow-datafusion/issues/3462 without the need to read the entire file into a custom in-memory format

And by contributing / helping the project upstream we help the overall community as well as benefit from those changes ourselves

PR #5895, opened 2022-10-18
refactor: use histogram to record TTBR

### LINKED ISSUE #5878: TTBR should be reported as a histogram
__Use case:__
Alerting when a given shard experiences extended periods of high TTBR.

__Proposal:__
There is an existing metric, `ingester_ttbr`, that is tracked as a `Gauge`. I would like to propose that this metric be removed and replaced with an identically named `Histogram`. 

This will have the effect of "time boxing" the metric such that periods of high TTBR immediately followed by zero writes result in a `histogram_quantile` query for the time range with zero writes resulting in a value of `0s` at all quantiles.

It will also cause TTBR to spike appropriately when an ingester restarts and begins replaying from the stream of unpersisted operations.

__Current behaviour:__
The `ingester_ttbr` metric is set each time a write is _applied_. This means that periods of zero inbound writes results in the `ingester_ttbr` metric maintaining the observed TTBR of the last applied write. Even if the ingester is operating normally and TTBR _would have_ recovered (were there writes to be applied), TTBR appears to remain at its previous elevated level until a new write for the given shard comes in.

__Desired behaviour:__
TTBR should be measured over a sliding time window such that even in the absence of incoming writes, our metrics do not give the impression that TTBR has remained at its last observed value.

__Alternatives considered:__
We considered simply changing our alert to also assert that there are indeed inbound writes during a given time window, but this problem will be encountered by all users who would like to query/alert on TTBR metrics, and we do not currently publish our alerting rules.

domodwyer:
Closes #5878.

---

* refactor: use histogram to record TTBR (c63312ce1)

      Changes the TTBR metric from a gauge to a histogram so that observations
      maintain a time dimension.

### DIFF
--- a/ingester/src/stream_handler/handler.rs
+++ b/ingester/src/stream_handler/handler.rs
@@ -6,7 +6,7 @@ use data_types::{SequenceNumber, ShardId, ShardIndex};
 use dml::DmlOperation;
 use futures::{pin_mut, FutureExt, StreamExt};
 use iox_time::{SystemProvider, TimeProvider};
-use metric::{Attributes, DurationCounter, DurationGauge, U64Counter};
+use metric::{Attributes, DurationCounter, DurationHistogram, U64Counter};
 use observability_deps::tracing::*;
 use tokio_util::sync::CancellationToken;
 use write_buffer::core::{WriteBufferErrorKind, WriteBufferStreamHandler};
@@ -55,7 +55,7 @@ pub(crate) struct SequencedStreamHandler<I, O, T = SystemProvider> {
 
     // Metrics
     time_provider: T,
-    time_to_be_readable: DurationGauge,
+    time_to_be_readable: DurationHistogram,
 
     /// Duration of time ingest is paused at the request of the LifecycleManager
     pause_duration: DurationCounter,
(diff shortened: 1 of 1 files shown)


domodwyer:
@cannonpalms has kindly offered / agreed to update the dashboards + alerts for this change ❤️

ingester/src/stream_handler/handler.rs
```diff
@@ -99,10 +99,10 @@ impl<I, O> SequencedStreamHandler<I, O> {
     ) -> Self {
         // TTBR
         let time_to_be_readable = metrics
-            .register_metric::<DurationGauge>(
+            .register_metric::<DurationHistogram>(
                 "ingester_ttbr",
-                "duration of time between producer writing to consumer putting into queryable \
-                 cache",
+                "distribution of duration between producer writing \
+                to consumer putting into queryable cache",
```
alamb, line 105:
```suggestion
                "distribution of duration between a write to the router \
                and being available for query in the ingester",
```

alamb (approved):
makes sense to me

ingester/src/stream_handler/handler.rs
alamb: Proposed in https://github.com/influxdata/influxdb_iox/pull/5909

PR #5896, opened 2022-10-18
refactor: avoid some clones while caching ns schema

crepererum:
Found while reviewing the code.


alamb (approved):
Which clones are avoided? On initial inspection it appears to me that the site of the clone just got moved

crepererum:
Currently this only avoids cloning the table names. I should really fix arrow to use Arc strings instead of owned strings for the schemas to avoid the rest.

PR #5908, opened 2022-10-19
test: ingester integration tests

domodwyer:
This PR serves three purposes:

* Create a reusable, easy-to-use test harness for ingester integration tests that allows easy writing and querying of data, and asserting of alert-able metrics so that adding new tests is low-friction - the easier they are to write, the more that will be wrote!
* Add sufficient tests to guard against any breakage in upcoming PRs that significantly refactors the way data is buffered/snapshot'ed and bubbled up to queries by driving all external APIs.
* Replace some tests that are very tightly coupled to how data is currently stored in the ingester, which will change with the above referenced PRs anyway.

---
* refactor: impl converting IngesterQueryResponse (d0b546109)

      An existing function to map the complex IngesterQueryResponse type to a
      simple set of RecordBatch existed in test code - this has been lifted
      onto an inherent method on the response type itself for reuse.

* test(ingester): add integration TestContext (b12d472a1)

      Adds a test helper type that maintains the in-memory state for a single
      ingester integration test, and provides easy-to-use methods to
      manipulate and inspect the ingester instance.

* test: write, query & progress API coverage (7729494f6)

      This commit adds a new test that exercises all major external APIs of
      the ingester:
      
          * Writing data via the write buffer
          * Waiting for data to be readable via the progress API
          * Querying data and and asserting the contents
      
      This should provide basic integration coverage for the Ingester
      internals. This commit also removes a similar test (though with less
      coverage) that was tightly coupled to the existing buffering structures.

* test: write buffer seeking tests (40f1937e6)

      Asserts write buffer seeking behaviour, including:
      
          * Seeking past already persisted data correctly
          * Skipping to next available op in non-contiguous offset stream
          * Skipping to next available op for dropped ops due to retention
          * Panics when seeking beyond available data (into the future)
      
      Removes a pair of tests that covered some of the above due to their
      tight coupling with ingester internals.

* refactor: more verbose shard reset logs (0c0a38c48)

      Adds a little more context to the "shard reset" logs.

### DIFF
--- a/ingester/src/handler.rs
+++ b/ingester/src/handler.rs
@@ -445,7 +445,6 @@ mod tests {
     use write_buffer::mock::{MockBufferForReading, MockBufferSharedState};
 
     use super::*;
-    use crate::data::{partition::SnapshotBatch, table::TableName};
 
     #[tokio::test]
     async fn test_shutdown() {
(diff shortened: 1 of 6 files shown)


ingester/src/handler.rs
```diff
@@ -435,192 +435,20 @@ impl<T> Drop for IngestHandlerImpl<T> {
...
     use mutable_batch_lp::lines_to_batches;
     use object_store::memory::InMemory;
     use test_helpers::maybe_start_logging;
     use write_buffer::mock::{MockBufferForReading, MockBufferSharedState};
 
     use super::*;
-    use crate::data::{partition::SnapshotBatch, table::TableName};
-
-    #[tokio::test]
-    async fn read_from_write_buffer_write_to_mutable_buffer() {
```
domodwyer, line 452: This test was replaced with a more thorough integration test `test_write_query`

ingester/src/handler.rs
```diff
@@ -754,88 +582,6 @@ mod tests {
         (ingester, shard, namespace)
     }
 
-    async fn verify_ingester_buffer_has_data(
```
domodwyer, line 757: This method was very tightly coupled to the existing data buffering logic, and needed to be removed.

The test cases it supported now use the progress API instead of polling the actual buffers.

ingester/src/handler.rs
```diff
@@ -754,88 +582,6 @@ mod tests {
...
-
-                tokio::time::sleep(Duration::from_millis(200)).await;
-            }
-        })
-        .await
-        .expect("timeout");
-    }
-
-    #[tokio::test]
-    async fn seeks_on_initialization() {
```
domodwyer, line 799: This test case is covered in the various new seeking tests.

ingester/src/handler.rs
```diff
@@ -912,135 +658,24 @@ mod tests {
             .unwrap();
     }
 
-    #[tokio::test]
-    async fn skip_to_oldest_available() {
```
domodwyer, line 916: This test case is also covered in the various new seeking tests.

ingester/src/querier_handler.rs
```diff
@@ -168,10 +168,63 @@ impl IngesterQueryResponse {
...
+
+    /// Convert [`IngesterQueryResponse`] to a set of [`RecordBatch`]es.
+    ///
+    /// If the response contains multiple snapshots, this will merge the schemas into a single one and create
+    /// NULL-columns for snapshots that miss columns.
+    ///
+    /// # Panic
+    /// Panics if there are no batches returned at all. Also panics if the snapshot-scoped schemas do not line up with
+    /// the snapshot-scoped record batches.
+    pub async fn into_record_batches(self) -> Vec<RecordBatch> {
```
domodwyer, line 180: This helper fn turns an `IngesterQueryResponse` into a set of `RecordBatch` which is what the caller really wants.

This makes comparing the data from the query API easier.

ingester/tests/common/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+    // namespaces.
+    namespaces: HashMap<String, NamespaceSchema>,
+
+    catalog: Arc<dyn Catalog>,
+    object_store: Arc<DynObjectStore>,
+    write_buffer_state: MockBufferSharedState,
+    metrics: Arc<metric::Registry>,
+}
+
+impl TestContext {
```
domodwyer, line 65: This is the int. test helper type that supports:

* Init an ingester
* Restart the ingester
* Create namespaces on demand
* Enqueue DML writes to the write buffer
* Enqueue line protocol to the write buffer (helper decorating the above)
* Query the progress API
* Wait for a write to become readable
* Perform a query via the Query API
* Fetch ingester metrics for assertions

ingester/tests/write.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    let expected = vec![
+        "+-------+-----------+--------------------------------+",
+        "| count | greatness | time                           |",
+        "+-------+-----------+--------------------------------+",
+        "|       | unbounded | 1970-01-01T00:00:00.000000010Z |",
+        "| 42    |           | 1970-01-01T00:00:00.000000200Z |",
+        "+-------+-----------+--------------------------------+",
+    ];
+    assert_batches_sorted_eq!(&expected, &data);
```
domodwyer, line 73: This example shows how to write LP (driving ingest), wait for it to be readable (driving the progress API), and query the buffered data and assert the contents (driving the query API).

lukebond (approved):
a thing of beauty!

PR #5909, opened 2022-10-19
fix: improve ttbr histogram metric description

alamb:
# Rationale:
TTBR is a key metric, so having a clear description of what it is will help operators run IOx.

Also, implement a suggestion on https://github.com/influxdata/influxdb_iox/pull/5895#discussion_r998337962 so I can remove it from my email based todo list

# Changes
Update description of TTBR metric


domodwyer (approved):
These are not surfaced in any meaningful way 🤷

PR #5910, opened 2022-10-19
refactor: replace `croaring` with `arrow`

crepererum:
With the read buffer gone, roaring bitmaps are only used to calculate series sets and these calculations are pretty much possible with the pure-Rust version. Also I don't deem that that performance-critical (compared to the roaring bitmaps in the read buffer core).

This removes a bunch of dependencies, mostly because `bindgen` is gone. This also removes our "croaring architecture detection" hack.


domodwyer (commented):
Have you run any benchmarks with this? Last time I checked the `roaring` implementation was incomplete due to lack of an RLE impl - see https://github.com/influxdata/influxdb_iox/issues/234.

crepererum:
# Perf
That bit isn't covered by benches, but the operation we perform here is basically just ORing all the set, which should be around as fast as creating them in the first place (we actually iterate over all the rows, once per tag, w/o any vectorization).

# Size
Given the fact that we have all the record batches at hand when we calculate the sets (they are NOT pulled / computed / streamed lazily), we have at least "number of rows" in bytes allocated. So the bitset won't be the issue here.

domodwyer:
It sounds like adding benches might be worthwhile, though I don't really understand your reasoning. Without the RLE containers it's effectively the same as using a regular old bitmap, which if you're arguing is fine, why use an incomplete `roaring` lib instead? 

I think you're saying it might be OK currently because of how we use it today, but it's pretty easy to use `roaring` later somewhere else, expecting it to be performant and it not. It just seems like a bit of a set-up to me.

crepererum:
> Without the RLE containers it's effectively the same as using a regular old bitmap, which if you're arguing is fine, why use an incomplete `roaring` lib instead?

I agree that a regular bitmap (or even a `Vec<bool>`) is good enough for this specific use case we have left. Will change that.

alamb (approved):
As the original author of this code, here is some context:

I needed to do boolean set intersection and we had `RoaringBitmap`s in the codebase which I heard was fast

I neither thought about nor benchmarked it carefully with other possibilities (like `Vec<bool>` or `BooleanArray` from the arrow kernels)

While profiling queries I have never seen this code anywhere (which makes sense as its output is next converted into a very inefficient protobuf `DataFrame` format which I would expect would totally dominate any time spent on the transition calculation itself)

So TLDR is I think this PR is fine, though I do not have any numerical justification for that opinion

domodwyer:
@alamb I think you have missed my point - bringing in an incomplete `roaring` dependency that has known performance pitfalls will surely only lead to it being used elsewhere in the code base (like how you used it here!) where the performance characteristics may actually be important.

I propose we do not depend on `roaring` until it is complete / performant.

crepererum:
Alright, I'll try to just use an arrow boolean array and call it a day...

crepererum:
@alamb I've replaced the manual set stuff with arrow arrays. Could you please double-check the logic (and my somewhat naive arrow kernel usage)?

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -176,92 +184,25 @@ impl SeriesSetConverter {
     ///
     /// Note: This may return false positives in the presence of dictionaries
     /// containing duplicates
-    fn compute_transitions(batch: &RecordBatch, col_idx: usize) -> Bitmap {
+    fn compute_transitions(batch: &RecordBatch, col_idx: usize) -> ArrayRef {
```
alamb, line 187: If you returned `BooleanArray` here directly you might be able to save some `as_any` casting above

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -176,92 +184,25 @@ impl SeriesSetConverter {
...
-                    .expect("Casting column");
-                let keys = col.keys();
-                let get_key = |idx| {
-                    if col.is_valid(idx) {
-                        return Some(keys.value(idx));
-                    }
-                    None
-                };
 
-                let col_values = col.values();
```
alamb, line 219: yeah, this old code was a bit of a mess -- @tustvold  has made working with DictionaryArrays much easier recently. Also in my previous self's defense, eq_dyn didn't exist yet

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -176,92 +184,25 @@ impl SeriesSetConverter {
...
-            "row transitions for results"
-        );
-        bitmap
+        arrow::compute::concat(&[
+            &{
+                let mut b = BooleanArray::builder(1);
+                b.append_value(false);
+                b.finish()
+            },
+            &arrow::compute::neq_dyn(&col.slice(0, col.len() - 1), &col.slice(1, col.len() - 1))
```
alamb, line 202: I think `concat` is ok here -- it bothers me for some reason that we have an extra copy of the data (like maybe we could append the result of `neq_dyn` to the builder) but given how fast `concat` is likely to be and small the arrays are I am fairly sure it won't matter 

And the builder might even be slower 🤔

iox_query/src/exec/seriesset/converter.rs
alamb: I'll try it as a follow on PR

alamb (approved):
LGTM -- thank you @crepererum . This code is executed for almost all grpc queries so I spent extra time looking at this one. and I think it looks good. 

I am fairly confident in the test coverage of this code as well.

Thank you

alamb:
I am going to to take the liberty of merging this one in so I can try a small follow up improvement

iox_query/src/exec/seriesset/converter.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/5927  -- turns out I could not figure out how to do it

crepererum: Yeah, if you figure out how to get an owned `BooleanArray` or `Arc<BooleanArray>` out of the `ArrayRef` (this is what `concat` returns). I couldn't and gave up :sweat_smile:

alamb: Filed https://github.com/apache/arrow-rs/issues/2901 for our arrow expert friends :)

Issue #5897, opened 2022-10-18
Expose parquet files directly to DataFusion


crepererum:
# What
Expose `ParquetExec` directly to the DF physical plan.

# Why
With #5397 we now use the DF parquet reader implementation. However we still hide it within our `QueryChunk::read_filter` that returns an opaque record batch stream. We should instead insert the `ParquetExec` directly into the phys. plan to allow the DF optimizer to improve the plan.

# How
Extend `QueryChunk::read_filter` to return an enum ("stream", "parquet exec") and extend `iox_query` to deal with it.

Note that this will also affect the compactor but I don't expect this to make a large difference there.

alamb:
As I understand it, one rationale for making this change is to reuse the DataFusion plan's pruning and predicate pushdown, and fan-in limiting functionality

For example you can give datafusion 5000 parquet files divided into 16 partitions and it will only ever be reading from 16 of them at a time 

Making this change would also make it easier to cache at a lower level than entire parquet file (e.g. object_store ranges / pages) I think

alamb:
🎉

PR #5911, opened 2022-10-19
refactor: use `SendableRecordBatchStream` to write parquets

crepererum:
Use a proper typed stream instead of peeking the first element. This is more in line with our remaining stack and shall also improve error handling.

Helps with my work on #5897.


NGA-TRAN (approved):
Makes sense

PR #5912, opened 2022-10-19
fix: use structured errors for logging influxrpc errors

alamb:
re https://github.com/influxdata/conductor/issues/1208

# Rationale
I would like to search greylog using the message text but I can not because the error text is included in the message rather than as its own field 

# Changes
Make the error reporting consistent and report error message in the `e` field


service_grpc_influxrpc/src/service.rs
```diff
@@ -177,7 +177,7 @@ impl From<Error> for tonic::Status {
     /// Converts a result from the business logic into the appropriate tonic
     /// status
     fn from(err: Error) -> Self {
-        error!("Error handling gRPC request: {}", err);
+        error!(e=%err, "Error handling gRPC request");
```
alamb, line 180: This is now consistent with how SQL / flight errors are logged: https://github.com/influxdata/influxdb_iox/blob/main/service_grpc_flight/src/lib.rs#L89

domodwyer (approved):
I strongly advocate static string messages with fields for errors/contextual data (I never compose strings together in logs!)

It makes manual log searching and automated log analysis much, much, much easier.

PR #5914, opened 2022-10-19
docs: update sql doc to replace database with namespace

NGA-TRAN:
Also update observer


alamb (approved):
Thanks @NGA-TRAN  -- the `observer` mode I think is pretty outdated / bitrotted at this point (as we can now query the postgres catalog directly). Make I should just remove it 🤔

alamb:
Proposed removal of observer mode: https://github.com/influxdata/influxdb_iox/pull/5918

PR #5917, opened 2022-10-19
fix: add storage-type=iox header to influxrpc responses

alamb:
Draft until we get consensus on  https://github.com/influxdata/conductor/issues/1208#issuecomment-1283994760

# Rationale
As explained on https://github.com/influxdata/conductor/issues/1208 the error handling system on influxcloud only displays error to the user if there is certain metadata on the response (serialized Go structures) that IOx does not return. Thus there needs to be a way to determine that the response came from IOx so errors can be properly displayed to the user.

# Changes
1. return "StorageType=iox" HTTP header in influxrpc responses (both success and failure)
2. Tests for same


service_grpc_influxrpc/src/service.rs
```diff
@@ -1564,6 +1543,16 @@ impl<T, E: std::fmt::Debug> ErrorLogger for Result<T, E> {
     }
 }
 
+/// Return the stream of results as a gRPC (tonic) response
+pub fn make_response<S>(
```
alamb, line 1547: this just reduces some boiler plate and gives me a single location to add the header to successes

service_grpc_influxrpc/src/service.rs
```diff
@@ -214,10 +214,19 @@ impl Error {
...
         };
 
-        tonic::Status::new(code, msg)
+        let mut status = tonic::Status::new(code, msg);
+        add_headers(status.metadata_mut());
+        status
     }
 }
 
+/// Add IOx specific headers to the response
```
alamb, line 223:
```suggestion
/// Add IOx specific headers to the response
///
/// storage-type: iox (needed so IDPE can show the errors to users)
///   see <https://github.com/influxdata/conductor/issues/1208>
```

alamb:
@mkmik  are you ok with this proposed change?

mkmik (approved):

domodwyer (approved):

service_grpc_influxrpc/src/service.rs
alamb: in https://github.com/influxdata/influxdb_iox/pull/5926

PR #5920, opened 2022-10-19
feat: Pass trace id through Flight requests from querier to ingester 

### LINKED ISSUE #5723: Trace querier->ingester requests
# What
Propagate tracing through querier->ingester requests.

# Why
Let's figure out why the ingester requests are so slow. Also see #5722.

# How
Tracing spans are already propagated down. They just need to flow through the InfluxDB IOx client.

carols10cents:
Fixes #5723.

I'm not at all sure that I've used `Span`, `SpanContext`, and `SpanRecorder` in the right way...

This is what I saw in Jaeger for a trace **before** this PR:

trace of a query that does not include any ingester request info

And this is what I see in Jaeger **after** this PR:

trace of a query that includes ingester request info added in this PR


crepererum:
Just looking at the screenshot, something is wrong with the trace. The new tree should be a subtree of the querier "ingester request" span, not a new root. Looking at the small warning sign of the new root, I think we fail to emit a span and therefore Jaeger cannot stitch/link the whole thing. Let me have a look at the code if I can find the issue...

influxdb_iox_client/src/client/flight/low_level.rs
```diff
@@ -59,9 +66,11 @@ impl ClientMetadata for IngesterQueryRequest {
 /// Low-level flight client.
 ///
 /// # Request and Response Metadata
-/// The type parameter `T` -- which must implement [`ClientMetadata`] describes the request and response metadata that
-/// is send and received during the flight request. The request is encoded as protobuf and send as the Flight "ticket",
-/// the response is received via the so called "app metadata".
+///
+/// The type parameter `T` -- which must implement [`ClientMetadata`] -- describes the request and
```
crepererum, line 70: Just for my own interest: do you use an auto-formatter for docstrings?

querier/src/ingester/mod.rs
```diff
@@ -418,7 +418,13 @@ async fn execute(
     };
 
     let query_res = flight_client
-        .query(Arc::clone(&ingester_address), ingester_query_request)
+        .query(
+            Arc::clone(&ingester_address),
+            ingester_query_request,
+            span_recorder
+                .child_span("IngesterQuery")
```
crepererum, line 425: This is the place where the linking breaks (I think). Let me explain:

A *span* is just an ID (actually it's three IDs, one for the trace, one for the span, one for the parent span, but let's ignore that for now). By itself it won't do anything, it's just data. A *span recorder* is a wrapper around a _span_ and ensures that the time range of the _span_ is set (using the `SpanRecorder::new` and `SpanRecorder::drop` methods, the latter is just the normal Rust drop) and that the *span* is sent to the collector when the *span recorder* is dropped. This is why I usually pass the *span* to a function saying "this is where this function should be in the tree" and use a *span recorder* right at the function entry point saying "start recording the actual time range of this function and emit the span once we exit".

The gRPC/HTTP trace IDs are slightly differently: there we sent the *span* of the caller/client, not the *span* of the first server function. So the server will always create a new child of the sent *span*. 

Your call here will create a new *span* but this will never be sent to Jaeger (because the server will create yet another child and you don't wrap this very span here into a *span recorder*). This is what Jaeger complains about.

A more background: *spans* in Jaeger and OpenTelemetry only link to their direct parent, not to the the whole "path" from the root. So if you loose one level, the tree reconstruction will not work and Jaeger will just create a new root for you (plus the warning you see in your screenshot).

So my recommendation would be to just use the `span_recorder.span()` here and don't create yet another level.

query_tests/src/scenarios/util.rs
```diff
@@ -945,11 +946,13 @@ impl IngesterFlightClient for MockIngester {
         &self,
         _ingester_address: Arc<str>,
         request: IngesterQueryRequest,
+        _span_context: Option<SpanContext>,
     ) -> Result<Box<dyn IngesterFlightClientQueryData>, IngesterFlightClientError> {
+        let span = None;
```
crepererum, line 951: The correct mocked span propagation is:

```
let span_recorder = SpanRecorder::new(span_context.child("ingester"));

...

let response = prepare_data_to_querier(&self.ingester_data, &request, span_recorder.child_span("prepare_data_to_querier"))
...
```

Background: This is how our magic HTTP/gRPC middleware does it:

https://github.com/influxdata/influxdb_iox/blob/ac431984883d5209f6563e1bf8ead8abf661fb77/trace_http/src/tower.rs#L113-L139

crepererum (approved):
With the span propagation fixed (and maybe the mocking, although we don't really test tracing in our query tests), this is perfect. Thank you :)

influxdb_iox_client/src/client/flight/low_level.rs
carols10cents: No, it bothers me enough that I just do it by hand 🙈

querier/src/ingester/mod.rs
carols10cents: Thanks! That explanation helps some... With the change from `child_span` to `span` here that I pushed in a new commit, here's what I'm seeing now. It does look like it's in the right subtree, but I'm not sure why it says `error = true`? 

 
jaeger traces with ingester request at a different point in the tree

crepererum: I think if you hover over the warning icon, Jaeger tells you what the problem is.

carols10cents: I'm not getting any pop-up when I hover over the ❗️

crepererum: Let me check if I can replicate this.... (currently in a meeting, so might take a while)

crepererum: Works for me in all-in-one mode. Can you share your exact testing steps?

Also, jaeger can be started with debug logs enabled, maybe that tells you something:

```console
$ docker run -it --rm -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one --log-level=debug
```

carols10cents: Yup, I'm running jaeger with:

```
docker run -it --rm -p 6831:6831/udp -p 16686:16686 jaegertracing/all-in-one
```

Then running all-in-one with:

```
cargo run -- run all-in-one --traces-exporter jaeger
```

Then running a query with:

```
cargo run -- query company_sensors "select count(*) from processes" --gen-trace-id
```

carols10cents: Here's the jaeger debug logs, nothing's standing out to me: https://gist.github.com/carols10cents/eac97479711ee2dba95d1d0da2780147

crepererum: Ahhh wait, I think I know what's going on. It might be that the ingester doesn't have any data for the requests and returns a 404/not-found, so the trace is marked as "error". IMHO this is OK for now. So :shipit: :)

Issue #3816, opened 2022-02-22
Add High Watermark to DmlMeta


tustvold:
**Problem**

As pointed out by @crepererum [here](https://github.com/influxdata/influxdb_iox/pull/3806#discussion_r812069581) the current way we poll to obtain the high watermark in order to compute the offset lag is not ideal. We must periodically poll the broker in order to obtain the high watermark for the partition.

**Proposal**

Kafka's [Fetch](https://kafka.apache.org/protocol#The_Messages_Fetch) API already returns a high watermark, and rskafka's `StreamConsumer` exposes this. However, currently the write buffer interface just discards this information.

In order to avoid potentially significant churn, I'd like to propose adding an optional high watermark field to `DmlMeta` that can be populated by write buffer implementations and used to compute lag metrics. This would be in additional to the existing optional sequence number.

**Alternative Considered**

* The write buffer stream type could be changed to have an ok payload that wraps `DmlOperation` and adds a high watermark
* The write buffer stream type could implement a trait, with a member function to retrieve the high watermark

PR #5922, opened 2022-10-19
feat: add high watermark to DMLMeta (#3816)

tustvold:
Part of #3816

A future PR can then remove the polling based offset tracking


tustvold:
Run out of time to work on this, will come back to later

PR #5927, opened 2022-10-20
refactor: arrow API usage cleanup

alamb:
Here are some minor cleanups to the arrow API usage in https://github.com/influxdata/influxdb_iox/pull/5910

I had hoped to use `BooleanArray` throughout but I couldn't figure out how to get a `BooleanArray` back from an `ArrayRef` / `Arc<dyn Array>`(I tried using `try_unwrap()` but that failed as `Array` is not `Sized`)


iox_query/src/exec/seriesset/converter.rs
```diff
@@ -193,13 +182,13 @@ impl SeriesSetConverter {
...
 
-        arrow::compute::concat(&[
+        compute::concat(&[
             &{
                 let mut b = BooleanArray::builder(1);
                 b.append_value(false);
                 b.finish()
             },
-            &arrow::compute::neq_dyn(&col.slice(0, col.len() - 1), &col.slice(1, col.len() - 1))
+            &compute::neq_dyn(&col.slice(0, col.len() - 1), &col.slice(1, col.len() - 1))
```
alamb, line 191: I was really hoping I could return `BooleanArray` from this function rather than `ArrayRef` but could not figure out how to downcast the `ArrayRef` 😞

crepererum: I had the same issue. :shrug:

crepererum (approved):

iox_query/src/exec/seriesset/converter.rs
alamb: https://github.com/apache/arrow-rs/issues/2901 -- let's see if something clever emerges

alamb:
Filed https://github.com/influxdata/influxdb_iox/issues/5929 for intermittent test failure

Issue #5887, opened 2022-10-18
Enable parquet "filter pushdown" / "predicate pushdown" by default


alamb:
__Use case:__
As part of making [queries on larger datasets work better](https://docs.google.com/document/d/1_8VThLvfeK5Ci5OjHNjzRgDEKs59XgM5jaNzcURsaoY/edit#)

We started running queries directly from parquet files in https://github.com/influxdata/influxdb_iox/pull/5879 to fix https://github.com/influxdata/conductor/issues/1002. As part of this, some queries got slower and we have identified the parquet filter pushdown feature of DataFusion as one way to win back the this performance (and then some). See https://github.com/apache/arrow-datafusion/issues/3462 for details. 

__Proposal:__
- [x] validating predicate pushdown is faster and more performant that not using (see  https://github.com/influxdata/idpe/issues/16227 for one way to measure this)
- [x] Ensure the feature has sufficient test coverage
- [ ] Turn on the predicate pushdown feature in IOx (e.g. what was done in https://github.com/influxdata/influxdb_iox/pull/5867), 

__Current behaviour:__
parquet predicate pushdown is not enabled

PR #5928, opened 2022-10-20
Alamb/upgrade df44444

alamb:
Closes #

Describe your proposed changes here.

- [ ] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #5930, opened 2022-10-20
feat: Enable parquet predicate pushdown in IOx by default

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/5887

~Draft as it~ 
- [x] ~builds on https://github.com/influxdata/influxdb_iox/pull/5928~
- [x] upstream https://github.com/apache/arrow-datafusion/pull/3989
- [x] Builds on https://github.com/influxdata/influxdb_iox/pull/6010
- [x] Builds on https://github.com/influxdata/influxdb_iox/pull/6011
- [x] Complete upstream testing https://github.com/apache/arrow-datafusion/pull/3976

# Rationale:
Towards https://github.com/influxdata/idpe/issues/16227

While we have plans to make predicate pushdown even better, we would like to start with it on and then iterate from there. 

See more context on https://github.com/apache/arrow-datafusion/issues/3462 and https://github.com/influxdata/k8s-idpe/issues/11978#issuecomment-1284466216

# Changes:
1. Enable parquet predicate pushdown by default
2. Add tests

# Questions
~I am not sure how to test this~
Update: Testing using parquet metrics


alamb:
I wonder if we can/should add metrics to the DataFusion parquet reader that shows how many rows (if any) were pruned? If we had that we could write a proper test for this

crepererum:
Yeah, metrics for this would definitely be nice, also for debugging.

One quick win could be to instrument the `SendableRecordBatchStream` that reads from the `ParquetExec` and let this one count rows. However I would kindly ask you to NOT do that because I'm in the process of lifting the `ParquetExec` up so that DataFusion can "see" it (i.e. not hiding it behind some opaque stream).

alamb:
I will look into adding some appropriate metric, etc next week

alamb:
Update here is I am marking this as a draft while I work on some tests / visibility upstream

alamb:
I have a proposed PR to add metrics upstream in DataFusion: https://github.com/apache/arrow-datafusion/pull/3989

alamb:
Updated to test using metrics.

alamb:
Hmm, well I found a few bugs upstream in DataFusion https://github.com/apache/arrow-datafusion/issues/4006 and https://github.com/apache/arrow-datafusion/issues/4005.

I think it needs some more testing before enabling in IOx

query_tests/src/sql_metrics.rs
```diff
@@ -0,0 +1,99 @@
...
+        );
+
+        // verify that pushdown was enabled and that it filtered rows
+        let metrics = parquet_metrics(physical_plan);
+        assert_eq!(
+            metric_value_sum(&metrics, "pushdown_rows_filtered"),
+            8,
+            "Unexpected number of rows filtered in:\n\n{:#?}",
+            metrics
+        );
```
alamb, line 72: This is the actual test showing that rows were filtered

crepererum (approved):

alamb:
🎉

alamb:
cc @tustvold  I am turning this on

Issue #5894, opened 2022-10-18
Add gRPC API to see the content of skipped_compactions and remove partitions from it


NGA-TRAN:
# Rationale

We have documented how [to see and remove data from `skipped_compactions` directly from the Catalog](https://github.com/influxdata/influxdb_iox/blob/main/docs/compactor.md#avoid-and-deal-with-files-in-skipped_compactions). Since we do not want anyone to manually touch the catalog, let us provide gRPC API for them instead.

We can provide 2 commands like:

1. Show content of skipped_compactions 2. 2. 

```SQL
SHOW skipped_compactions;

-- This will work like
 SELECT * FROM skipped_compactions;
```

2.  Remove specified partitions from the skipped_compactions

```SQL
REMOVE <partition_id> FROM skipped_compactions;

-- This will work like
DELETE FROM skipped_compactions WHERE partition_id in ([your_partitions_ids]);
```

PR #5931, opened 2022-10-20
feat: Add gRPC and CLI APIs to list skipped compactions

carols10cents:
Connects to #5894.

This adds a gRPC API to list all skipped compactions, which is the first part of the issue.

Because I find gRPC hard to use, I also added a CLI. So if you have a compactor running with this code and the gRPC listener is at `127.0.0.1:8082`, you can run:

```
cargo run -- debug skipped-compactions list
```

and you'll see either empty braces if there are no skipped compactions:

```
[]
```

or JSON  describing the skipped compactions if there are some:

```
[
  {
    "partitionId": "51",
    "reason": "because",
    "skippedAt": "3",
    "numFiles": "10",
    "limitNumFiles": "8",
    "estimatedBytes": "10000",
    "limitBytes": "500"
  }
]
```

If a table would be a more useful format for this output, I'm happy to look into that next!

### DIFF
--- a/compactor/src/server/grpc.rs
+++ b/compactor/src/server/grpc.rs
@@ -3,7 +3,7 @@
 use crate::handler::{CompactorHandler, ListSkippedCompactionsError};
 use generated_types::influxdata::iox::compactor::v1::{
     self as proto,
-    skipped_compaction_service_server::{SkippedCompactionService, SkippedCompactionServiceServer},
+    compaction_service_server::{CompactionService, CompactionServiceServer},
 };
 use std::sync::Arc;
 use tonic::{Request, Response};
@@ -21,22 +21,20 @@ impl<I: CompactorHandler + Send + Sync + 'static> GrpcDelegate<I> {
         Self { compactor_handler }
     }
 
-    /// Acquire a SkippedCompaction gRPC service implementation.
-    pub fn skipped_compaction_service(
-        &self,
-    ) -> SkippedCompactionServiceServer<impl SkippedCompactionService> {
-        SkippedCompactionServiceServer::new(SkippedCompactionServiceImpl::new(Arc::clone(
-            &self.compactor_handler,
-        ) as _))
+    /// Acquire a Compaction gRPC service implementation.
+    pub fn compaction_service(&self) -> CompactionServiceServer<impl CompactionService> {
+        CompactionServiceServer::new(CompactionServiceImpl::new(
+            Arc::clone(&self.compactor_handler) as _,
+        ))
     }
 }
 
 /// Implementation of skipped compaction
-struct SkippedCompactionServiceImpl {
+struct CompactionServiceImpl {
     handler: Arc<dyn CompactorHandler + Send + Sync + 'static>,
 }
 
-impl SkippedCompactionServiceImpl {
+impl CompactionServiceImpl {
     pub fn new(handler: Arc<dyn CompactorHandler + Send + Sync + 'static>) -> Self {
         Self { handler }
     }
@@ -54,7 +52,7 @@ impl From<ListSkippedCompactionsError> for tonic::Status {
 }
 
 #[tonic::async_trait]
-impl SkippedCompactionService for SkippedCompactionServiceImpl {
+impl CompactionService for CompactionServiceImpl {
     async fn list_skipped_compactions(
         &self,
         _request: Request<proto::ListSkippedCompactionsRequest>,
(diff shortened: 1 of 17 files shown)


NGA-TRAN:
@carols10cents 
This looks good. I think in addition to gRPC and CLI, if you add this command in our SQL tool

```SQL
  SHOW   skipped_compactions;   -- or LIST skipped_compactions
```
 to display a table of all rows in skipped_partitions, it will easier for us and users to see the data.

It will work similar to `SHOW namespaces` documented in this  [doc](https://github.com/influxdata/influxdb_iox/blob/main/docs/sql.md) has example of `show namespaces`. 

Regarding to removing rows from the skipped_compactions, I am not sure if we want to provide a statement in our SQL though.

I add @alamb to give us advice here

NGA-TRAN (approved):

generated_types/protos/influxdata/iox/compactor/v1/skipped_compaction.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+package influxdata.iox.compactor.v1;
+option go_package = "github.com/influxdata/iox/compactor/v1";
+
+service SkippedCompactionService {
```
alamb, line 5: My only major suggestion is to name this `CompactorService` rather than `SkippedCompactionService` as I can imagine we will be adding other requests to the compactor over time. 

We can do so as a follow on PR as well when we want to add more requests, I just figured it might be easier to get the code into the final shape at this time.

generated_types/protos/influxdata/iox/compactor/v1/skipped_compaction.proto
```diff
@@ -0,0 +1,31 @@
...
+}
+
+message ListSkippedCompactionsRequest {}
+
+message ListSkippedCompactionsResponse {
+  // A list of skipped compactions
+  repeated SkippedCompaction skipped_compactions = 1;
+}
+
+message SkippedCompaction {
```
alamb, line 17: It might be nice to add some comments here explaining what the fields mean

compactor/src/server/grpc.rs
```diff
@@ -0,0 +1,76 @@
...
+use tonic::{Request, Response};
+
+/// This type is responsible for managing all gRPC services exposed by `compactor`.
+#[derive(Debug, Default)]
+pub struct GrpcDelegate<I: CompactorHandler> {
+    compactor_handler: Arc<I>,
+}
+
+impl<I: CompactorHandler + Send + Sync + 'static> GrpcDelegate<I> {
+    /// Initialise a new [`GrpcDelegate`] passing valid requests to the specified
```
alamb, line 18: I see you have gone 🇬🇧  on us 😆  (probably more following the pattern from @domodwyer )

Initialise vs Initialize -- still gets me every time

alamb (approved):

alamb:
> This looks good. I think in addition to gRPC and CLI, if you add this command in our SQL tool

In terms of the sql repl tool -- right now it needs to be connected to the querier service whereas the compactor API is now exposed on the compactor service. Thus I am not sure how useful a SHOW COMPACTION would be in the SQL REPL because you would have to connect to the compactor and the only command you could run would be `SHOW COMPACTIONS` -- you couldn't send it sql or `SHOW NAMESPACES`

compactor/src/server/grpc.rs
carols10cents: LOL 🇬🇧 copypasta 😂

carols10cents:
@NGA-TRAN It sounds like you'd rather have the results in a table, so I changed the display. When there are skipped compactions, it will look like this:

```
+--------------+---------+-------------------------------------+-----------------+-------------+-----------+-----------------+
| partition_id | reason  | skipped_at                          | estimated_bytes | limit_bytes | num_files | limit_num_files |
+--------------+---------+-------------------------------------+-----------------+-------------+-----------+-----------------+
| 51           | because | 1970-01-01T00:00:00.000000003+00:00 | 10000           | 500         | 10        | 8               |
+--------------+---------+-------------------------------------+-----------------+-------------+-----------+-----------------+
```

And when there aren't any, the table will be empty:

```
$ cargo run -- debug skipped-compactions list
+--------------+--------+------------+-----------------+-------------+-----------+-----------------+
| partition_id | reason | skipped_at | estimated_bytes | limit_bytes | num_files | limit_num_files |
+--------------+--------+------------+-----------------+-------------+-----------+-----------------+
+--------------+--------+------------+-----------------+-------------+-----------+-----------------+
```

I'm going to merge this since both of you approved, let me know if you'd like changes and I'll make them in a future PR!

NGA-TRAN:
Looks nice. Thanks @carols10cents

PR #5942, opened 2022-10-21
refactor: do NOT use fake DF context for parquet reading

crepererum:
Use the proper top-level DataFusion context and register the object store there.

Note that we still hide the `ParquetExec` behind an opaque record batch stream. Fixing that is next on my list.

Helps with #5897.


querier/src/chunk/mod.rs
```diff
@@ -410,13 +399,12 @@ pub mod tests {
     }
 
     /// collect data for the given chunk
-    async fn collect_read_filter(chunk: &dyn QueryChunk) -> Vec<RecordBatch> {
+    async fn collect_read_filter(
+        chunk: &dyn QueryChunk,
+        ctx: IOxSessionContext,
```
alamb, line 404: ❤️

query_tests/src/scenarios/util.rs
```diff
@@ -643,18 +643,14 @@ struct MockIngester {
...
 /// Query-test specific executor with static properties that may be relevant for the query optimizer and therefore may
 /// change `EXPLAIN` plans.
-static GLOBAL_EXEC: Lazy<Arc<Executor>> = Lazy::new(|| {
-    Arc::new(Executor::new_with_config(ExecutorConfig {
-        num_threads: 1,
-        target_query_partitions: 4,
-    }))
-});
+static GLOBAL_EXEC: Lazy<Arc<DedicatedExecutors>> =
+    Lazy::new(|| Arc::new(DedicatedExecutors::new(1)));
```
alamb, line 647: 🤔  I wonder if we should run `4` here? It probably doesn't matter really

alamb (approved):
This is a marvelous PR -- thank you @crepererum

parquet_file/src/storage.rs
```diff
@@ -223,19 +248,11 @@ impl ParquetStorage {
         schema: SchemaRef,
         path: &ParquetFilePath,
         file_size: usize,
+        session_ctx: &SessionContext,
     ) -> Result<SendableRecordBatchStream, ReadError> {
         let path = path.object_store_path();
         trace!(path=?path, "fetching parquet data for filtered read");
 
-        // set up "fake" DataFusion session (TODO thread the real one
```
alamb, line 230: for anyone else reading this PR -- the major goal was to remove this here (and thread the `session_ctx` all the way down)

query_tests/src/scenarios/util.rs
crepererum: Since tests run in multiple threads already, I would try not to use multiple threads per test because IMHO it's a bit hard to reason about that.

Issue #5872, opened 2022-10-17
Ingester memory tracker does not account for ongoing queries


domodwyer:
The ingester tracks the memory of data it has buffered. Once the data has been persisted, it returns the size of the data to the memory tracker. This tracker limits memory usage, which is supposed to prevent OOMs.

The ingester returns the data's byte size back to the memory tracker immediately after persistence, assuming it is dropped immediately. However ongoing queries retain a reference to the "persisting" data during query execution, and therefore prevent the data from being dropped.

The impact of this bug is that the memory tracker can (significantly) under-estimate the memory being used by the system. This would lead to OOMs.

PR #5943, opened 2022-10-21
feat(ingester): partition buffer state machine

### LINKED ISSUE #5805: Ingester does not incrementally build snapshots via query path
__Expected behaviour:__
For each query, the ingester's `MutableBatch` should be converted into a `RecordBatch` (creating a "snapshot"), stored, and then all snapshots should be returned to the querier.

Subsequent queries should re-use existing snapshots.

__Actual behaviour:__
The `MutableBatch` is converted, but the resulting `RecordBatch` is discarded after the query is complete.

Data is repeatedly converted into `RecordBatch` instead of being incrementally converted as necessary.

NOTE: data _is_ incrementally converted as tombstones are applied (as expected).

domodwyer:
**This PR adds code that isn't used yet.** A follow-up PR will stitch it in to minimise the PR size.

---

This PR aims to set the stage to replace the implicit state machine used to buffer writes in a partition, with the code in this PR.

The aim is to create an implementation that has a lower complexity burden for callers, and is easier/quicker to optimise & build additional functionality on top of.

When this is actually used (next PR) ~this will fix #5805 to decrease query latency~ (see comments below), and provides a clear path forwards for improved memory tracking in the ingester (to fix bugs such as #5872 and #5777).

---

* feat(ingester): partition buffer state machine (fbd262dd2)

      This commit introduces code that is intended to replace the current
      implicit state machine used by PartitionData. The existing code is still
      in use, the new code is NOT used in this commit. A follow-up commit will
      switch over to minimise the diff.
      
      This change has two main goals;
          * encapsulation & simplification for callers
          * robust implementation so developing correct additions is easier
      
      This is a significant refactor of the partition buffering logic to
      encapsulate the various states of data (buffering, snapshot, persisting
      and the mixed states between them) within the Partition. This alleviates
      the rest of the system from having to be concerned with the differences
      between "buffering" data, and "unpersisted data", "snapshot data",
      "persisting data", "persisting with snapshots" etc - callers now invoke
      a method called get_query_data() and they are provided with all the
      relevant data for a partition. This abstraction change alone
      significantly reduces code and test complexity in the rest of the
      ingester.
      
      For the second goal, the new implementation leverages an explicit state
      machine, encoded using typestates. Typestate ensures compile-time
      correctness of transitions and method calls, and the explicit FSM itself
      helps ensure the system progresses in the desired manner - this fixes
      and helps prevent bugs caused by implicit states such as:
      
          https://github.com/influxdata/influxdb_iox/issues/5805
      
      This state machine makes the system states explicit and
      self-descriptive, helping to reduce the cost of developer on-boarding
      (no prior knowledge of "how this bit works") and reduces ongoing
      developer burden. This explicit nature also de-risks adding new
      functionality - it should be relatively easy to add concurrent snapshot
      generation or incremental compaction without introducing bugs. The state
      transition logic is abstracted away from callers, minimising the
      overhead of this strategy.

### DIFF
--- a/ingester/src/data/partition/buffer/always_some.rs
+++ b/ingester/src/data/partition/buffer/always_some.rs
@@ -1,0 +1,99 @@
+//! A helper type that ensures an `Option` is always `Some` once the guard is
+//! dropped.
+
+/// A guard through which a value can be placed back into the [`AlwaysSome`].
+#[derive(Debug)]
+#[must_use = "Guard must be used to restore the value"]
+pub(super) struct Guard<'a, T>(&'a mut Option<T>);
+
+impl<'a, T> Guard<'a, T> {
+    /// Store `value` in the [`AlwaysSome`] for subsequent
+    /// [`AlwaysSome::take()`] calls.
+    pub(super) fn store(self, value: T) {
+        assert!(self.0.is_none());
+        *self.0 = Some(value);
+    }
+}
+
+/// A helper type that aims to ease working with an [`Option`] that must always
+/// be restored in a given scope.
+///
+/// Accessing the value within an [`AlwaysSome`] returns a [`Guard`], which MUST
+/// be used to store the value before going out of scope. Failure to store a
+/// value cause a subsequent [`Self::take()`] call to panic.
+///
+/// Failing to store a value in the [`Guard`] causes a compiler warning, however
+/// this does not prevent failing to return a value to the [`AlwaysSome`] as the
+/// warning can be falsely silenced by using it within one conditional code path
+/// and not the other.
+#[derive(Debug)]
+pub(super) struct AlwaysSome<T>(Option<T>);
+
+impl<T> Default for AlwaysSome<T>
+where
+    T: Default,
+{
+    fn default() -> Self {
+        Self::new(T::default())
+    }
+}
+
+impl<T> std::ops::Deref for AlwaysSome<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        self.0.as_ref().unwrap()
+    }
+}
+
+impl<T> AlwaysSome<T> {
+    /// Wrap `value` in an [`AlwaysSome`].
+    pub(super) fn new(value: T) -> Self {
+        Self(Some(value))
+    }
+
+    /// Read the value.
+    pub(super) fn take(&mut self) -> (Guard<'_, T>, T) {
+        let value = std::mem::take(&mut self.0);
+
+        (
+            Guard(&mut self.0),
+            value.expect("AlwaysSome value is None!"),
+        )
+    }
+
+    /// Deconstruct `self`, returning the inner value.
+    pub(crate) fn into_inner(self) -> T {
+        self.0.unwrap()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_always_some() {
+        let mut a = AlwaysSome::<usize>::default();
+
+        let (guard, value) = a.take();
+        assert_eq!(value, 0);
+        guard.store(42);
+
+        let (guard, value) = a.take();
+        assert_eq!(value, 42);
+        guard.store(24);
+
+        assert_eq!(a.into_inner(), 24);
+    }
+
+    #[test]
+    #[should_panic = "AlwaysSome value is None!"]
+    fn test_drops_guard() {
+        let mut a = AlwaysSome::<usize>::default();
+        {
+            let _ = a.take();
+        }
+        let _ = a.take();
+    }
+}
--- a/ingester/src/data/partition/buffer/mutable_buffer.rs
+++ b/ingester/src/data/partition/buffer/mutable_buffer.rs
@@ -1,0 +1,53 @@
+use std::sync::Arc;
+
+use arrow::record_batch::RecordBatch;
+use mutable_batch::MutableBatch;
+use schema::selection::Selection;
+
+/// A [`Buffer`] is an internal mutable buffer wrapper over a [`MutableBatch`]
+/// for the [`BufferState`] FSM.
+///
+/// A [`Buffer`] can contain no writes.
+///
+/// [`BufferState`]: super::super::BufferState
+#[derive(Debug, Default)]
+pub(super) struct Buffer {
+    buffer: Option<MutableBatch>,
+}
+
+impl Buffer {
+    /// Apply `batch` to the in-memory buffer.
+    ///
+    /// # Data Loss
+    ///
+    /// If this method returns an error, the data in `batch` is problematic and
+    /// has been discarded.
+    pub(super) fn buffer_write(&mut self, batch: MutableBatch) -> Result<(), mutable_batch::Error> {
+        match self.buffer {
+            Some(ref mut b) => b.extend_from(&batch)?,
+            None => self.buffer = Some(batch),
+        };
+
+        Ok(())
+    }
+
+    /// Generates a [`RecordBatch`] from the data in this [`Buffer`].
+    ///
+    /// If this [`Buffer`] is empty when this method is called, the call is a
+    /// NOP and [`None`] is returned.
+    ///
+    /// # Panics
+    ///
+    /// If generating the snapshot fails, this method panics.
+    pub(super) fn snapshot(self) -> Option<Arc<RecordBatch>> {
+        Some(Arc::new(
+            self.buffer?
+                .to_arrow(Selection::All)
+                .expect("failed to snapshot buffer data"),
+        ))
+    }
+
+    pub(super) fn is_empty(&self) -> bool {
+        self.buffer.is_none()
+    }
+}
--- a/ingester/src/data/partition/buffer/state_machine.rs
+++ b/ingester/src/data/partition/buffer/state_machine.rs
@@ -1,0 +1,289 @@
+#![allow(dead_code)]
+use std::sync::Arc;
+
+use arrow::record_batch::RecordBatch;
+use data_types::SequenceNumber;
+use mutable_batch::MutableBatch;
+
+mod buffering;
+mod buffering_with_snapshot;
+mod persisting;
+mod snapshot;
+
+pub(in crate::data::partition::buffer) use buffering::*;
+pub(in crate::data::partition::buffer) use buffering_with_snapshot::*;
+pub(crate) use persisting::*;
+pub(in crate::data::partition::buffer) use snapshot::*;
+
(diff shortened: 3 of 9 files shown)


ingester/src/data/partition/buffer/always_some.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Accessing the value within an [`AlwaysSome`] returns a [`Guard`], which MUST
+/// be used to store the value before going out of scope. Failure to store a
+/// value cause a subsequent [`Self::take()`] call to panic.
+///
+/// Failing to store a value in the [`Guard`] causes a compiler warning, however
+/// this does not prevent failing to return a value to the [`AlwaysSome`] as the
+/// warning can be falsely silenced by using it within one conditional code path
+/// and not the other.
+#[derive(Debug)]
+pub(super) struct AlwaysSome<T>(Option<T>);
```
domodwyer, line 30: This is a helper type internal to the FSM.

ingester/src/data/partition/buffer/mutable_buffer.rs
```diff
@@ -0,0 +1,53 @@
...
+use schema::selection::Selection;
+
+/// A [`Buffer`] is an internal mutable buffer wrapper over a [`MutableBatch`]
+/// for the [`BufferState`] FSM.
+///
+/// A [`Buffer`] can contain no writes.
+///
+/// [`BufferState`]: super::super::BufferState
+#[derive(Debug, Default)]
+pub(super) struct Buffer {
```
domodwyer, line 14: This serves as the "mutable buffer", currently backed by a `MutableBatch` instance.

In the future we can easily swap the `MubtableBatch` storage for something else without impacting the rest of the code.

ingester/src/data/partition/buffer/state_machine.rs
```diff
@@ -0,0 +1,289 @@
...
+/// Boxes with dashed lines indicate immutable, queryable states that contain
+/// data in an efficient data format for query execution ([`RecordBatch`]).
+///
+/// Boxes with solid lines indicate a mutable state to which further writes can
+/// be applied.
+///
+/// A [`BufferState`] tracks the bounding [`SequenceNumber`] values it has
+/// observed, and enforces monotonic writes (w.r.t their [`SequenceNumber`]).
+#[derive(Debug)]
+pub(crate) struct BufferState<T> {
```
domodwyer, line 81: This defines the FSM, and documents the stats & transitions.

Note that a persisting state is terminal - this prevents any accidental mutation once the data is marked as persisting. 

Because this state is a distinct type, it also allows us to define code that works with persisting data as operating on a `BufferState<Persisting>` type only, eliminating a bunch of error classes and making the compiler do the hard verification work :+1:

ingester/src/data/partition/buffer/state_machine.rs
```diff
@@ -0,0 +1,289 @@
...
+    pub(crate) fn write(
+        &mut self,
+        batch: MutableBatch,
+        n: SequenceNumber,
+    ) -> Result<(), mutable_batch::Error> {
+        self.state.write(batch)?;
+        self.sequence_range.observe(n);
+        Ok(())
+    }
+}
```
domodwyer, line 123: A FSM in a mutable state (one that implements `Writeable`) captures the sequence number to store the `(min, max)` range, and delegates the write to the actual FSM state.

ingester/src/data/partition/buffer/state_machine.rs
```diff
@@ -0,0 +1,289 @@
...
+where
+    T: Queryable,
+{
+    /// Returns the current buffer data.
+    ///
+    /// This is always a cheap method call.
+    fn get_query_data(&self) -> &[Arc<RecordBatch>] {
+        self.state.get_query_data()
+    }
+}
```
domodwyer, line 137: An immutable / query-able state (one that implements `Queryable`) delegates data reading to the state itself.

ingester/src/data/partition/buffer/always_some.rs
domodwyer, line 30: This is a helper type internal to the FSM.

ingester/src/data/partition/buffer/state_machine.rs
```diff
@@ -0,0 +1,289 @@
...
+    use std::ops::Deref;
+
+    use arrow_util::assert_batches_eq;
+    use mutable_batch_lp::test_helpers::lp_to_mutable_batch;
+    use schema::selection::Selection;
+
+    use super::*;
+
+    #[test]
+    fn test_buffer_lifecycle() {
```
domodwyer, line 150: This test enumerates the entire lifecycle of a FSM, and asserts various properties it should uphold.

ingester/src/data/partition/buffer/state_machine/buffering.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl BufferState<Buffering> {
+    /// Attempt to generate a snapshot from the data in this buffer.
+    ///
+    /// This returns [`Transition::Unchanged`] if this buffer contains no data.
+    pub(crate) fn snapshot(self) -> Transition<Snapshot, Buffering> {
+        if self.state.buffer.is_empty() {
+            // It is a logical error to snapshot an empty buffer.
+            return Transition::unchanged(self);
+        }
```
domodwyer, line 33: A snapshot is never empty, and therefore all code that has one of the states `Snapshot` / `BufferingWithSnapshot` / `Persisting` is guaranteed to contain data.

ingester/src/data/partition/buffer/traits.rs
```diff
@@ -0,0 +1,17 @@
...
+/// A state that can accept writes.
+pub(crate) trait Writeable: Debug {
+    fn write(&mut self, batch: MutableBatch) -> Result<(), mutable_batch::Error>;
+}
+
+/// A state that can return the contents of the buffer as one or more
+/// [`RecordBatch`] instances.
+pub(crate) trait Queryable: Debug {
+    fn get_query_data(&self) -> &[Arc<RecordBatch>];
+}
```
domodwyer, line 17: These traits define the behaviours of the various FSM states.

* Mutable states implement `Writeable`
* Immutable states implement `Queryable`

ingester/src/data/sequence_range.rs
```diff
@@ -0,0 +1,141 @@
+use data_types::SequenceNumber;
+
+/// A range of sequence numbers, both inclusive [min, max].
+#[derive(Debug, Default, PartialEq, Eq, Clone)]
+pub(crate) struct SequenceNumberRange {
```
domodwyer, line 5: A tracker of the `(min, max)` sequence number ranges it observes. This is used in the FSM to record the offsets the buffer contains incrementally.

ingester/src/data/sequence_range.rs
```diff
@@ -0,0 +1,141 @@
...
+    #[test]
+    fn test_ranges() {
+        let mut r = SequenceNumberRange::default();
+
+        r.observe(SequenceNumber::new(0));
+        r.observe(SequenceNumber::new(2));
+        r.observe(SequenceNumber::new(3));
+
+        assert_eq!(r.inclusive_min(), Some(SequenceNumber::new(0)));
+        assert_eq!(r.inclusive_max(), Some(SequenceNumber::new(3)));
```
domodwyer, line 64: The `inclusive_min()` naming ensures the caller is always certain of the bounds (inclusive/exclusive).

ingester/src/data/partition/buffer/state_machine/buffering_with_snapshot.rs
```diff
@@ -0,0 +1,62 @@
+//! A writfield1 buffer, with one or more snapshots.
```
carols10cents, line 1: What's "writfield1"?

ingester/src/data/partition/buffer/state_machine/persisting.rs
```diff
@@ -0,0 +1,37 @@
+//! A writfield1 buffer, with one or more snapshots.
```
carols10cents, line 1: This comment is the same as `buffering_with_snapshot`'s description? I'd highlight how they're different here.

ingester/src/data/partition/buffer/state_machine/snapshot.rs
```diff
@@ -0,0 +1,49 @@
+//! A writfield1 buffer, with one or more snapshots.
```
carols10cents, line 1: Same comment here too that I would expect to explain the differences :)

ingester/src/data/partition/buffer/always_some.rs
carols10cents, line 30: I'm interested to see how `AlwaysSome` is going to be used, I'm a little skeptical about whether this is actually necessary but I'll withhold judgment ;) Usage is coming in the next PR then?

carols10cents (approved):

domodwyer:
I will be removing the incremental snapshot generation (https://github.com/influxdata/influxdb_iox/issues/5805) due to both the [performance of datafusion on plans with many `RecordBatch`](https://github.com/influxdata/conductor/issues/1173) and the memory amplification of storing a small number of rows in a `RecordBatch` which always allocates memory for 8k rows.

I will restore this incremental snapshot behaviour alongside incremental compaction to ensure we coalesce snapshots instead of storing many until persist time. This may also help avoid the persist time memory explosions we've observed with the current code when compacting all the data in a partition at once.

ingester/src/data/partition/buffer/state_machine/buffering_with_snapshot.rs
domodwyer: It is a trap to ensure someone is paying attention.

NGA-TRAN (approved):
Except a question about sequence number range, the rest looks very good. Your explanation how it works make it easy to review. The tests are very good 👍

ingester/src/data/sequence_range.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// each instance, and the maximum of [`Self::inclusive_max()`].
+    ///
+    /// If both `self` and `other` contain no [`SequenceNumber`] observations,
+    /// the returned instance contains no observations.
+    pub(crate) fn merge(&self, other: &Self) -> Self {
+        let merged_range = self
+            .range
+            .into_iter()
+            .chain(other.range.clone())
+            .reduce(|a, b| (a.0.min(b.0), a.1.max(b.1)));
```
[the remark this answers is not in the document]
domodwyer: This type is unaware of any invariants on the values it contains - it simply merges.

Code that makes use of this type ensures they never overlap (coming in the next PR) :+1:

PR #5944, opened 2022-10-21
refactor: disable incremental snapshot generation

domodwyer:
This removes the `BufferingWithSnapshot` state from the partition buffer FSM, and makes the `Buffering` state queryable (discarding the generated `RecordBatch` after the query has executed).

This mimics the current non-incremental-snapshot behaviour (and leaves the state machine a little bare!)

See https://github.com/influxdata/influxdb_iox/issues/5805 for more context.

---

* refactor: disable incremental snapshot generation (7b3fa4320)

      This commit removes the on-demand, incremental snapshot generation
      driven by queries.
      
      This functionality is "on hold" due to concerns documented in:
      
          https://github.com/influxdata/influxdb_iox/issues/5805
      
      Incremental snapshots will be introduced alongside incremental
      compactions of those same snapshots.


ingester/src/data/partition/buffer/state_machine.rs
```diff
@@ -54,19 +51,14 @@ impl<A, B> Transition<A, B> {
...
-///               └──│ BufferingWithSnapshot │
-///                  └───────────────────────┘
+///                  ┌──────────────┐
+///                  │  Buffering   │
+///                  └───────┬──────┘
+///                          │
+///                          ▼
+///                  ┌ ─ ─ ─ ─ ─ ─ ─       ┌ ─ ─ ─ ─ ─ ─ ─
+///                      Snapshot   ├─────▶   Persisting  │
+///                  └ ─ ─ ─ ─ ─ ─ ─       └ ─ ─ ─ ─ ─ ─ ─
```
NGA-TRAN, line 61: Does it mean in the future, this immutable `Snapshot` will become mutable?

NGA-TRAN (approved):

ingester/src/data/partition/buffer/state_machine.rs
domodwyer: No, the snapshot remains immutable - it is now just a transient state / a step to get from `buffering` -> to `persisting` :+1:

I could have removed it, but I will be adding back `BufferingWithSnapshot` at some point soon hopefully so left it as-is to minimise the code deletion 🤞

NGA-TRAN: Yeah, makes more sense to have `BufferingWithSnapshot` instead

PR #5945, opened 2022-10-21
feat: Add gRPC and CLI APIs to delete skipped compactions 

shepmaster:
Follow up to #5931; part of #5894

### DIFF
--- a/iox_catalog/src/interface.rs
+++ b/iox_catalog/src/interface.rs
@@ -126,6 +126,9 @@ pub enum Error {
 
     #[snafu(display("could not list skipped compactions: {source}"))]
     CouldNotListSkippedCompactions { source: sqlx::Error },
+
+    #[snafu(display("could not delete skipped compactions: {source}"))]
+    CouldNotDeleteSkippedCompactions { source: sqlx::Error },
 }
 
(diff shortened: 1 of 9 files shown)


iox_catalog/src/interface.rs
```diff
@@ -1582,6 +1591,31 @@ pub(crate) mod test_helpers {
...
+            .partitions()
+            .delete_skipped_compactions(other_partition.id)
+            .await
+            .unwrap();
+
+        assert!(
+            not_deleted_skipped_compaction.is_none(),
+            "There should be no skipped compation",
+        );
+
```
carols10cents, line 1618: Could you please add an assertion that the list of skipped compactions here is empty again? 🙏🏻

NGA-TRAN (approved):
LGTM

iox_catalog/src/interface.rs
shepmaster: [Added!](https://github.com/influxdata/influxdb_iox/compare/9eff299b0b9d0af41f5acc8fbb313787d9f4789e..fa7fe2e9cf3b129da4b6a6ef0f106c2f81a13215)

carols10cents (approved):

PR #5946, opened 2022-10-21
fix: cardinality of each batch should use row count of the batch

### LINKED ISSUE #5899: More accurate cardinality estimation while estimating compaction memory
# Rationale

Currently, to estimate memory needed for dictionary for a tag, we need to estimate the file cardinality and [that value is 1/2 of the file row_count ](https://github.com/influxdata/influxdb_iox/blob/bb7df22aa1783e040ea165153876f1fe36838d4e/compactor/src/compact.rs#L516). If the file has a lot of row (> 100M), the cardinality will be huge which leads to memory needed is very large. The consequence of this is Compactor may think this partition is too large to compact and skipped it.

In the case of a lot of files with a lot of rows, we might reduce the ratio of cardinality estimation to have better memory estimation

NGA-TRAN:
Closes #5899 


compactor/src/lib.rs
```diff
@@ -677,52 +679,63 @@ pub mod tests {
         partition2.create_parquet_file_catalog_record(pf2_2).await;
 
         // P3: bytes >= 90% of full budget = 90% * 14,350 = 12,915
-        //   L0 40 rows. bytes: 2050 + 100 * 50 = 7,050
-        //   L1 24 rows. bytes: 2050 + 100 * 40 = 6,050
-        // total = 7,050 + 6,050 = 13,100
+        //   L0 40 rows. bytes: 2,250
+        //   Five L1s. bytes: 2,250 each
```
NGA-TRAN, line 683: Since the estimation does not rely on the file row count, I need to create many files to have the budget I need for the test

NGA-TRAN:
@carols10cents and @pauldix : this is the issue we talked this morning.

@alamb : this is the one I mentioned yesterday and it turns out to be a kind of bug. The cardinality for each batch should use the batch row count instead of the file row count

compactor/src/compact.rs
```diff
@@ -513,12 +510,12 @@ fn estimate_arrow_bytes_for_file(
     const BOOL_BYTE: i64 = 1;
     const AVERAGE_ROW_COUNT_CARDINALITY_RATIO: i64 = 2;
 
-    let average_cardinality = row_count / AVERAGE_ROW_COUNT_CARDINALITY_RATIO;
-
     // Since DataFusion streams files and allocates a fixed (configurable) number of rows per batch,
     // we always use that number to estimate the memory usage per batch.
     let row_count_per_batch = min_num_rows_allocated_per_record_batch_to_datafusion_plan as i64;
 
+    let average_cardinality = row_count_per_batch / AVERAGE_ROW_COUNT_CARDINALITY_RATIO;
```
alamb, line 517: this makes much more sense to me 👍

alamb (approved):
Looks good to me 👍

pauldix (approved):

domodwyer (approved):

PR #5947, opened 2022-10-21
docs: Add skipped-compaction CLI to the compactor documentation

carols10cents:
Closes https://github.com/influxdata/influxdb_iox/issues/5894.

Plus some cleanups because my brain is broken and notices inconsequential things like double spaces and then can't un-see them 🙃


NGA-TRAN (approved):
Nice. Thanks for fixing the typos. Previous reviewers have captured a few but there are still many 😆

PR #5948, opened 2022-10-21
fix: Compactor skipped_at is in seconds, not nanoseconds oops

carols10cents:
@NGA-TRAN just sent me some production output from the new skipped-compactions command and... I forgot that `skipped_at` hasn't been put through line protocol/etc and therefore is [postgres epoch seconds](https://github.com/influxdata/influxdb_iox/blob/95853aed1a0955e580fc9b96debc940c7eaf3e56/iox_catalog/src/postgres.rs#L1257) rather than nanoseconds 🤦🏻‍♀️ This should bring all the dates displayed back to the present day 😜 

doc brown saying "to the future"


NGA-TRAN (approved):
Now we are to the future

PR #5954, opened 2022-10-24
refactor: `IOxReadFilterNode` can always accumulate statistics

crepererum:
`IOxReadFilterNode` used to not emit statistics if one chunk has duplicates or delete predicates. This is wrong (or at least overly conservative), because the node itself (or the chunks themselves) do NOT perform dedup or delete predicate filtering. Instead this is done is done by parent nodes (`DeduplicateExec` and `FilterExec`) and its their job to propagate statistics correctly.

Helps w/ #5897.


influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -164,7 +164,7 @@ async fn query_after_persist_sees_new_files() {
             ],
         },
         // write another parquet file
-        Step::WriteLineProtocol(setup.lp_to_force_persistence()),
+        Step::WriteLineProtocol(setup.lp_to_force_persistence().replace("tag=A", "tag=B")),
```
crepererum, line 167: While I thought this would be a no-op change, this actually fixes a query bug. The line protocol is identical and hence should be de-duplicated. However `count(*)` w/o a filter can use row count stats if they are exact. And we wrongly reported that our stats are correct and hence got "2 rows" while there was actually only one.

I've added two more query tests that hit the same issue (the "fixed" count is 18, but with the bug on `main` it is 20).

iox_query/src/provider/deduplicate.rs
```diff
@@ -237,15 +237,9 @@ impl ExecutionPlan for DeduplicateExec {
     }
 
     fn statistics(&self) -> Statistics {
-        // TODO: we should acount for overlaps at this point -- if
```
alamb, line 240: 🎉

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -164,7 +164,7 @@ async fn query_after_persist_sees_new_files() {
             ],
         },
         // write another parquet file
```
alamb, line 166:
```suggestion
        // write another parquet file
        // that has non duplicated data 
```

alamb (approved):
Nice find -- and now that @NGA-TRAN  has the compactor tuned well, typically the parquet files are not overlapped 👍

NGA-TRAN (commented):
It is always nice that we know our system so well and find bugs before customers hit them 🎉

PR #5955, opened 2022-10-24
revert: use histogram to record TTBR

domodwyer:
Histogram bucketing makes messes with our ability to measure changes in TTBR, or correctly measure values that exceed the maximum bucket. We can increase the maximum bucket size, but ultimately TTBR is unbounded and can always exceed the max.

image

Reverts #5895.

---

* revert: use histogram to record TTBR (39f826518)

      This reverts commit c63312ce125d39415e78c9322f721ea322cf2869.
      
      This change fixed a low-priority alert when there was no traffic flowing
      through the system. The loss in TTBR value fidelity due to bucketing is
      a greater concern as it affects live, high-volume clusters and hinders
      operational insight.

### DIFF
--- a/ingester/src/stream_handler/handler.rs
+++ b/ingester/src/stream_handler/handler.rs
@@ -6,7 +6,7 @@ use data_types::{SequenceNumber, ShardId, ShardIndex};
 use dml::DmlOperation;
 use futures::{pin_mut, FutureExt, StreamExt};
 use iox_time::{SystemProvider, TimeProvider};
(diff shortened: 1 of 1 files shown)


cannonpalms (approved):
This was quite painful indeed. Thanks for reverting it.

domodwyer:
I think this was more of a hacky way of fixing the TTBR measurement rather than a useful independent metric.

A proper fix would be to return TTBR=0 when there are no outstanding messages to consume from Kafka, but changing the metric was quicker.

PR #5960, opened 2022-10-24
refactor: do not project chunks in `TestDatabase::chunks`

crepererum:
Databases are NOT required to project chunks (in practice this is only done by the querier for ingester-based chunks). Instead `iox_query` should (and already does) add the right stream adapters to project chunks or to create NULL-columns. Removing the special handling from the test setup makes it easier to understand and also less likely that `iox_query` starts to rely on this behavior.

Helps with #5897.


iox_query/src/test.rs
```diff
@@ -108,54 +108,27 @@ impl QueryDatabase for TestDatabase {
         &self,
         table_name: &str,
         predicate: &Predicate,
-        projection: &Option<Vec<usize>>,
+        _projection: &Option<Vec<usize>>,
```
alamb, line 111: is your eventual plan to remove this option from `read_filter` in all places?

alamb (approved):
This makes sense to me -- I wonder if we should remove the `projection` argument from the chunk interface entirely 🤔

iox_query/src/test.rs
crepererum: No, because the table CAN still use it if it wants. This is helpful if we only wanna pull a subset of columns from the ingester. However it is NOT mandatory.

PR #5961, opened 2022-10-24
refactor: simplify `iox_query::provider::overlap`

crepererum:
- remove generic that is basically unused (`group_potential_duplicates` is always called w/ `Arc<dyn QueryChunk>`)
- remove half-baked `impl` that is unused

Helps w/ #5897.


NGA-TRAN (approved):
Nice discovery 😆

PR #5963, opened 2022-10-24
refactor: remove `QueryChunkMeta::timestamp_min_max`

crepererum:
On top of #5961.

Use the table summary instead. This allows us to have a single mechanism
that both IOx and DataFusion understand. This basically lifts the "basic
table summary" mechanism that the querier uses to `iox_query` and let
the compactor and ingester use the same mechanism.

While not strictly necessary, simplifying the `QueryChunk[Meta]`
interface helps with #5897.


NGA-TRAN (approved):
I like this refactor. Using TableSummary everywhere is definitely better

NGA-TRAN:
But btw, this will conflict with @Dom's recent ingested refactoring that may not be merged soon.

alamb:
This is a good simplification -- thank you 👍

ingester/src/query_adaptor.rs
```diff
@@ -162,24 +180,14 @@ impl QueryChunkMeta for QueryAdaptor {
         None // Ingester data is not sorted
     }
 
-    fn timestamp_min_max(&self) -> Option<TimestampMinMax> {
```
alamb, line 165: Thank you -- this is a good cleanup

PR #5964, opened 2022-10-24
refactor: make `SchemaBuilder::field` fallible

crepererum:
It would be nice if the IOx data type would not be optional and this is a prep clean-up to achieve that.


schema/src/builder.rs
```diff
@@ -70,34 +70,34 @@ impl SchemaBuilder {
...
-            .map(InfluxColumnType::Field)
-            .ok();
-
-        self.add_column(column_name, true, influxdb_column_type, arrow_type)
+    pub fn field(
+        &mut self,
+        column_name: &str,
+        arrow_type: ArrowDataType,
+    ) -> Result<&mut Self, &'static str> {
+        let influxdb_column_type = arrow_type.clone().try_into().map(InfluxColumnType::Field)?;
```
alamb, line 86: 👍

alamb (approved):
makes sense to me

PR #5968, opened 2022-10-24
feat: Use workspace dep inheritance for arrow crates

carols10cents:
This is a low-priority enhancement that will save future work.

[Rust 1.64 stabilized workspace inheritance](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds), and one of the pieces of that is dependency inheritance. [The top-level Cargo.toml can specify a set of dependencies, their versions, and a universal set of features](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table), then [each crate in the workspace that uses that dependency need only specify `workspace = true`, any additional dependencies, and whether or not that dependency is optional](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace). 

I decided to start with just the arrow-related crates to see what folks thought of this feature, and because being able to specify the version numbers for the arrow-related crates in one spot would let us get rid of:

- The `datafusion` shim crate
- The `update_arrow_deps` script

The dependency inheritance feature does *not* remove the need for the `workspace-hack` crate because:

- The `workspace-hack` crate also manages indirect dependencies
- The `workspace-hack` crate prevents rebuilds of crates with different feature sets, but each crate should still only specify the features they need

Especially wanted to get @alamb's thoughts as he usually handles the arrow/parquet/datafusion upgrades.

I could see doing this for other crates that need to be updated all at once, like tonic or tokio, but decided to start small.


NGA-TRAN (approved):
LGTM but I think @alamb should review this

Cargo.toml
```diff
@@ -108,6 +107,13 @@ authors = ["IOx Project Developers"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
 
+[workspace.dependencies]
```
alamb, line 110: This is now the place that we would change the datafusion and arrow dependencies, right?

alamb (approved):
LGTM -- thanks @carols10cents 

I believe this PR will conflict with https://github.com/influxdata/influxdb_iox/pull/5976

Cargo.toml
carols10cents: Yup, you got it!!

domodwyer (approved):

PR #5972, opened 2022-10-25
refactor: enforce nullability for (known) IOx data types

crepererum:
We don't support non-null tags, non-null fields, or nullable timestamps. Let's just remove this from `schema` so that this never happens on accident.


schema/src/merge.rs
```diff
@@ -459,17 +460,23 @@ mod tests {
 
     #[test]
     fn test_merge_incompatible_schema_nullability() {
```
crepererum, line 462: This only works because we currently allow columns to not have an IOx data type at all. I'll try to fix this next, because this doesn't make any sense.

lukebond (approved):
i can't really appreciate the full impact of this but i can see it does what you set out to do

NGA-TRAN (approved):
LGTM

schema/src/merge.rs
alamb: As some historical context, I think I wrote this code when the requirements of the data model were much less clear to me (e.g would we allow inserting non line protocol formatted data into IOx). Ensuring the schema mapping makes sense for LineProtocol sounds good to me

alamb (commented):
LGTM 👍

PR #5973, opened 2022-10-25
chore: temporarily disable circle filter to build & push PRs

lukebond:
We will be deploying a canary image to one ingester only in one environment and rather than build the image locally this PR will enable image building and pushing for PRs temporarily so we can get an image for a branch that we don't want to merge ([this one](https://github.com/influxdata/influxdb_iox/pull/5958)).

I'll revert it once we have that image built.

UPDATE: i changed this on request to be parameterised and not temporary. in fact that's what `build_perf` already did, so i renamed that.


crepererum:
Could we have a parameter-based, non-temporary solution? I think we had this a while back which allowed team members to issue a circleCi REST call to build and push images. Let me check where this is gone...

crepererum:
Ah, it's still there:

https://github.com/influxdata/influxdb_iox/blob/99f36f88897c5a3f1fce05881cf891ce4a7f9191/.circleci/config.yml#L512-L523

lukebond:
> Ah, it's still there:

i think it's for the CI base image. but i'll do something similar for this build and deploy step :+1:

domodwyer (approved):
Sounds legit - thanks for this :+1:

.circleci/config.yml
```diff
@@ -509,6 +521,13 @@ workflows:
     jobs:
       - ci_image
 
+  # Force build and push of container image for non-main branch
+  release_branch:
```
crepererum, line 525: Isn't this what `build_perf` below does? Also `deploy_release` requires `build_release` to work.

crepererum (changes requested):

crepererum (approved):
Thanks :)

PR #5978, opened 2022-10-26
refactor: make influx column type mandatory in Schema

crepererum:
We basically assume everywhere that a column falls into one of the three known categories (time, tag, field), so lets encode this in our type system instead of defining "unknown" as "undefined behavior, may or may not crash".


schema/src/lib.rs
```diff
@@ -187,7 +200,7 @@ impl Schema {
     /// used only by the SchemaBuilder.
     pub(crate) fn new_from_parts(
         measurement: Option<String>,
-        fields: impl Iterator<Item = (ArrowField, Option<InfluxColumnType>)>,
+        fields: impl Iterator<Item = (ArrowField, InfluxColumnType)>,
```
alamb, line 203: So the rationale behind modeling the type as `Option<InfluxColumnType>` was so that IOx could eventually support data other than what was written as line protocol (and thus was tag / field / time). 

However, as we don't actually support such other data yet, making the code easier to work with rather than keeping it less easy for some possible future use is a good idea

alamb (approved):

NGA-TRAN (commented):
I like this refactor. We want to be clear rather than ambiguous code

schema/src/lib.rs
alamb: Yes, I agree 👍

Issue #5975, opened 2022-10-25
Log aborted gRPC requests


crepererum:
Log some information about aborted gRPC requests because this helps to diagnose "querier stuck" issues.

PR #5979, opened 2022-10-26
refactor: parse HTTP trace header even w/o collector

crepererum:
Since we log trace IDs to allow easier correlation of logs no matter what the `sampled` flag says, we should also parse these logs if we don't have a tace collector at all.

In practice, this won't make a difference since we always deploy with a trace collector, but it also makes the code easier to reason about.

Helps with #5975.


trace_http/src/tower.rs
```diff
@@ -99,24 +99,16 @@ where
...
+            .trace_header_parser
+            .parse(self.collector.as_ref(), request.headers())
+        {
             Ok(Some(ctx)) => {
                 request
                     .extensions_mut()
                     .insert(RequestLogContext::new(ctx.clone()));
 
-                ctx.sampled.then(|| {
+                (ctx.sampled && ctx.collector.is_some()).then(|| {
```
crepererum, line 111: This is the core change: the `RequestLogContext` (used for query logging) will be created even when no trace collector is configured. However the root span (that we emit to Jaeger&Co) is still only created when we have collector.

mkmik (approved):

PR #5980, opened 2022-10-26
feat: log cancelled HTTP/gRPC requests

### LINKED ISSUE #5975: Log aborted gRPC requests

crepererum:
This will be helpful to see when the querier or router is too slow and we timeout. In contrast to the existing metrics, this also helps w/ log correlation (i.e. "when did we get stuck").

Closes #5975.


alamb (approved):
looks good to me -- I think it would be good if @tustvold  gave this a review too (post merge would be fine)

PR #5982, opened 2022-10-26
refactor: enforce name of the one-and-only time column

crepererum:
We currently only support a single time dimension and some parts of other stack rely on the name of the time column. So lets enforce the name (note that `schema::try_from_arrow` already checks for duplicate column, so we are now left with a single dimension).


schema/src/lib.rs
```diff
@@ -89,6 +89,13 @@ pub enum Error {
         column_name: String,
         md: Option<String>,
     },
+
+    #[snafu(display(
+        "Time column should be named '{}' but is named '{}'",
```
alamb, line 94:
```suggestion
        "Internal Error: Time column should be named '{}' but is named '{}'",
```

alamb: maybe it is worth a hint to users that there is no way they are going to work around this lol

alamb (approved):
THERE CAN BE ONLY ONE

schema/src/lib.rs
crepererum: done

PR #5989, opened 2022-10-27
feat: Log how long each persist plan takes to complete

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/5958 from @domodwyer~

# Rationale
I think the time required to persist individual partitions can be quite large, thus increasing the overall time required for recovery. This is acutely important during "long catchup" replays when the ingester must work through the kafka backlog as quickly as possible.

I believe at least part of the challenge is related to the data itself and also that it is related to the time taken to sort the data in the DataFusion plan. 

I would like to be able to find and gather some exemplar data set to validate this hypothesis and so that we can evaluate how well @tustvold's ongoing changes to sorting in Arrow and DataFusion will improve the situation in IOx. 

Having information about what plans are taking so long, will allow me to more easily locate data to reproduce the syptoms

# Changes
Log how many seconds it takes to complete a persist operation


domodwyer:
This is a good candidate for exposing as a metric too, so we can graph it over time. I've already got it on my TODO list if you're happy with just logs for now though :+1:

domodwyer (approved):

alamb:
I think the logs and metrics serve different purposes -- metrics will show us how these operations are doing in aggregate and the logs will let us find the details of particular outliers that we want to examine more closely (aka so I can go pull their data locally and figure out what is being so slow). 

Of course knowing we should look in the logs for outliers will likely be triggered by the metrics 😆

PR #5984, opened 2022-10-26
feat: MVP of remote store get-table command

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/5967.

This PR adds the new command `influxdb_iox remote store get-table <DATABASE> <TABLE>` that will fetch all Parquet files for the specified database's table to the local filesystem.

I tested this out by running a local all-in-one server, which has http://localhost:8081 as the router gRPC port. I set `INFLUXDB_IOX_PERSIST_PARTITION_AGE_THRESHOLD_SECONDS=10` to get lots of Parquet files.  I wrote some data into the database `test_db`, table `disk`. Then I ran:

```
influxdb_iox -h http://localhost:8081 remote store get-table test_db disk
```

and I saw this output:

```
found 2 Parquet files, downloading...
downloading file 1 of 2 (a1fb45f9-d04f-4aa1-a058-0a75c846264c.parquet)...
downloading file 2 of 2 (54e03fa9-01fa-4da6-90b1-c92fa034d0a0.parquet)...
Done.
```

and I had those two files in a directory named `disk` in my current directory.

I'm going to work on the bonus points next but wanted to get feedback on the main logic/behavior before I got too far!

Issue #5967, opened 2022-10-24
CLI Command to fetch all parquet files for a table


alamb:
__Use case:__
I want to more easily be able to copy the parquet files for a table from our production cluster locally for reproduction. 

Currently I can fetch individual files using the `influxdb_iox remote store get <UUID>` command but there is no way to download but I need to know the UUID to use this

To find the UUID I need to go poke in the postgres catalog 

__Proposal:__
I would really like a command that worked like this

```shell
# fetch all files from the database 12345_6788 and table 'sensors' to the directory named sensors
influxdb_iox remote store get-table 12345_6788 'sensors'
```

After running this command, I would expect to see a directory like this with the relevant files

```shell
$ ls -l sensors/
total 3376
-rw-r--r--  1 alamb  staff   6.7K Oct 24 13:40 28bf025a-5175-408d-a380-18c4358844b6.parquet
-rw-r--r--  1 alamb  staff   258K Oct 24 13:40 4fd5f739-b76c-42bf-9635-98da4ae209a7.parquet
-rw-r--r--  1 alamb  staff   376K Oct 24 13:40 67798467-9b72-43f2-8980-cab0f7b853ff.parquet
-rw-r--r--  1 alamb  staff   332K Oct 24 13:40 8e6887cb-6b0a-4487-8531-c61d555ccbff.parquet
-rw-r--r--  1 alamb  staff   604K Oct 24 13:40 efad01a2-f510-47b4-ae6e-613485478ce3.parquet
```

Bonus points:
- [ ] Allow the output directory to be specified (`influxdb_iox remote store get-table 12345_6788 'sensors' -o output_directory`)
- [ ] Add the partition_id in the file names  (`<UUID>`.`<PARTITION_ID>`.`parquet`)
- [ ] Support incremental downloads / restarting (check to see if the parquet file already exists with the correct size and skip downloading it if so)
- [ ] Support streaming file reads (right now all data is buffered in memory)
 

I think this would be a matter of:
- [ ] Extending the catalog service so it could provide information on a table (like what parquet files are there)
- [ ] Adding the new cli command to use the new catalog service and existing remote service

pauldix:
You can do this with an individual partition already. Look at `remove partition`.  You should be able to specify the namespace, table name, and partition key (e.g. 2022-10-27) and the local catalog DSN (Postgres running on your laptop) and object store info (your local filesystem).

alamb:
[...]

PR #5990, opened 2022-10-27
feat: Some bonus features for `remote store get-table`

carols10cents:
Connects to https://github.com/influxdata/influxdb_iox/issues/5967

This PR implements 2 of the bonus points:

- [x] Allow the output directory to be specified (`influxdb_iox remote store get-table 12345_6788 'sensors' -o output_directory`)
- [x] Add the partition_id in the file names  (`<UUID>`.`<PARTITION_ID>`.`parquet`)

And [feedback from the previous PR](https://github.com/influxdata/influxdb_iox/pull/5984#discussion_r1006216160), at least for the code that I just added (the rest of the codebase I'll save for later): 

- [x] I think we should use the word "namespace" consistently throughout the codebase rather than sometimes using database and sometimes use namespace

I also made what I think are improvements to some error message displays:

Before this change, if the namespace specified wasn't found, the error message would be:

```
Error in the store subcommand: Client error: Some requested entity was not found: Namespace nacho-namespace 
not found
```

Which is:

- Unhelpful, I know I ran the `store` subcommand
- Confusing, am I the client? (no) Does "Client error" mean I did something wrong? (also no)
- Repetitive, why tell me something wasn't found and then tell me what wasn't found

After this change, the error message is:

```
IOx request failed: Namespace nacho-namespace not found
```

which I find much clearer.

### DIFF
--- a/influxdb_iox/src/commands/remote/store.rs
+++ b/influxdb_iox/src/commands/remote/store.rs
@@ -2,6 +2,7 @@
 
 use futures::StreamExt;
 use influxdb_iox_client::{catalog, connection::Connection, store};
+use std::path::PathBuf;
 use thiserror::Error;
 use tokio::{
     fs::{self, File},
@@ -51,6 +52,11 @@ struct GetTable {
     /// The name of the table to get the Parquet files for
     #[clap(action)]
     table: String,
+
+    /// The output directory to use. If not specified, files will be placed in a directory named
+    /// after the table in the current working directory.
+    #[clap(action, short)]
+    output_directory: Option<PathBuf>,
 }
 
 /// All possible subcommands for store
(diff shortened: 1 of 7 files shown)


influxdb_iox/tests/end_to_end_cases/remote.rs
```diff
@@ -45,26 +46,93 @@ async fn remote_store_get_table() {
...
+                        .arg(&router_addr)
+                        .arg("remote")
+                        .arg("store")
+                        .arg("get-table")
+                        .arg("nacho-namespace")
+                        .arg(&table_name)
+                        .assert()
+                        .failure()
+                        .stderr(predicate::str::contains(
+                            "Namespace nacho-namespace not found",
```
alamb, line 134: love those nacho namespaces !

influxdb_iox/src/commands/remote/store.rs
```diff
@@ -40,17 +41,21 @@ struct Get {
...
+    /// The namespace to get the Parquet files for
     #[clap(action)]
-    database: String,
+    namespace: String,
 
     /// The name of the table to get the Parquet files for
     #[clap(action)]
     table: String,
+
+    /// The output directory to use. If not specified, files will be placed in a directory named
```
alamb, line 55: 👍

alamb (approved):
Looks great @carols10cents  -- I again tried to use this to pull my favorite table from production but https://github.com/influxdata/influxdb_iox/pull/5984 appears to have not yet been deployed.

Issue #4897, opened 2022-06-17
Make PartitionKey required in DmlWrite


domodwyer:
All `DmlWrite` instances within the IOx service boundary should contain the partition key, but currently we have to make the partition key an optional value to enable the gRPC write API client to construct an non-partitioned DML write:

https://github.com/influxdata/influxdb_iox/blob/0fbff981ecb5b1deb152ae6b5b6a13a9fe9d6078/influxdb_iox_client/src/client/write.rs#L65

No other code needs this optionality property.

This means we can't statically assert the partition key is set within IOx, because the public API uses the same wire format as kafka, uses the same `DmlWrite` type to construct the request, and therefore has to be optional. This API is destined for removal (#4866) which should allow us to change the `PartitionKey` to be required.

See https://github.com/influxdata/influxdb_iox/pull/4864#discussion_r898095113.

PR #5996, opened 2022-10-28
refactor(dml): PartitionKey required for writes

domodwyer:
A `PartitionKey` is always required, but for reasons it had to be an `Option`. Those reasons are no longer valid :+1:

Closes https://github.com/influxdata/influxdb_iox/issues/4897.

---

* refactor(dml): PartitionKey required for writes (72a358e52)

      Changes the DmlWrite type to require a PartitionKey be specified,
      instead of accepting an Option.
      
      This requirement was already in place - the write buffer upheld an
      invariant that all writes contained a partition key value (was not
      "None") or it panicked at runtime when attempting to enqueue the write.
      
      It is now possible to encode this invariant in the type system, which is
      what this change does.


crepererum (approved):
Always a big fan of removing `Option<...>` from the code base :muscle:

PR #5997, opened 2022-10-28
refactor: mandatory `QueryChunkMeta::summary`

crepererum:
With #5963 merged, all chunks now provide a summary (even though it may not contain data for all columns). So let's make it mandatory, which also removes a few 🙈-style `.except(...)` calls.


alamb (approved):
looks reasonable to me

NGA-TRAN (commented):
Nice

alamb:
Merging this PR in to keep the queue backlog down

PR #5991, opened 2022-10-27
feat: Last 2 bonus features of remote store get-table

carols10cents:
Closes https://github.com/influxdata/influxdb_iox/issues/5967

This PR implements:

- [x] Support incremental downloads / restarting (check to see if the parquet file already exists with the correct size and skip downloading it if so)
- [x] Support streaming file reads (right now all data is buffered in memory)

and I think completes the tool, unless you've discovered further enhancements that you'd like :)

PR #5999, opened 2022-10-28
feat: comments and hints about what service to use with `remote` command

alamb:
Follow on to @carols10cents  https://github.com/influxdata/influxdb_iox/pull/5991 😍  PR
related to https://github.com/influxdata/influxdb_iox/issues/5967

# Rationale
Both @NGA-TRAN  and I have been confused when we tried to run the `remote` command against the wrong service (querier). One needs to run it when connected to the router.

# Changes
Add hints and docs that nudge my future self in this direction

# Example

Before this change:
```shell
$ ~/Software/target-iox/debug/influxdb_iox --host http://localhost:8083 remote store get-table foo bar
IOx request failed: Operation is not implemented or supported:
```

After this change:
```shell
$ influxdb_iox remote store get-table dab3b3b2e54dc4a4_ab038f883697864a vibrations -o /tmp/vibrations`
Error connecting to http://127.0.0.1:8082: Connection error: transport error (error trying to connect: tcp connect error: Connection refused (os error 61))
```
```shell
$ influxdb_iox remote store get-table dab3b3b2e54dc4a4_ab038f883697864a vibrations -o /tmp/vibrations`
HINT: are you connected to the router service?
IOx request failed: Operation is not implemented or supported:
```


domodwyer (approved):

carols10cents:
The other solution is we could add the catalog service to the other servers? Is there a reason we don't want to do that?

alamb:
> The other solution is we could add the catalog service to the other servers? Is there a reason we don't want to do that?

🤔  this is a good point -- we would also have to add the object store service to the other servers.

Since we already interact with the querier server a bunch, maybe we should just move all those services to the querier server?

NGA-TRAN:
I agree to have this work on all servers @carols10cents . There is no reason to not do so

alamb:
Closing in favor of https://github.com/influxdata/influxdb_iox/pull/6005

PR #6000, opened 2022-10-28
refactor(router): pass NamespaceId through handler stack

domodwyer:
The first in a series of PRs to push namespace + table IDs over Kafka (#4880).

This PR changes the router to propagate the `NamespaceId` through the request handlers, so it is available when serialising the DML operation into the Kafka wire format.

---

* style: format imports (0c5eb3f70)

      Re-order and re-format the imports so that they follow a consistent
      pattern.
      
      This helps eliminate conflicts due to imports.

* refactor: resolve namespace before DML dispatch (d166de931)

      This commit introduces a new (composable) trait; a NamespaceResolver is
      an abstraction responsible for taking a string namespace from a user
      request, and mapping to it's catalog ID.
      
      This allows the NamespaceId to be injected through the DmlHandler chain
      in addition to the namespace name.
      
      As part of this change, the NamespaceAutocreation layer was changed from
      an implementator of the DmlHandler trait, to a NamespaceResolver as it
      is a more appropriate abstraction for the functionality it provides.

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -1,14 +1,96 @@
-//! IOx router role implementation.
+//! IOx router implementation.
 //!
 //! An IOx router is responsible for:
 //!
 //! * Creating IOx namespaces & synchronising them within the catalog.
 //! * Handling writes:
-//!     * Receiving IOx write/delete requests via HTTP and gRPC endpoints.
+//!     * Receiving IOx write/delete requests via HTTP.
 //!     * Enforcing schema validation & synchronising it within the catalog.
 //!     * Deriving the partition key of each DML operation.
 //!     * Applying sharding logic.
-//!     * Push resulting operations into the appropriate shards (Kafka partitions if using Kafka).
+//!     * Push resulting operations into the appropriate shards (Kafka
+//!       partitions if using Kafka).
+//!
+//! The router is composed of singly-responsible components that each apply a
+//! transformation to the request:
+//!
+//! ```text
+//!                           ┌──────────────┐
+//!                           │   HTTP API   │
+//!                           └──────────────┘
+//!                                   │
+//!                                   │
+//!                                   ▼
+//!                      ╔═ NamespaceResolver ════╗
+//!                      ║                        ║
+//!                      ║  ┌──────────────────┐  ║
+//!                      ║  │    Namespace     │  ║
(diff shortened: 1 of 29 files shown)


router/src/lib.rs
```diff
@@ -1,14 +1,96 @@
...
+//!                           ┌──────────────┐
+//!                           │ Write Buffer │
+//!                           └──────────────┘
+//!                                   │
+//!                                   │
+//!                          ┌────────▼─────┐
+//!                          │    Kafka     ├┐
+//!                          └┬─────────────┘├┐
+//!                           └┬─────────────┘│
+//!                            └──────────────┘
```
domodwyer, line 70: This documents the structure of the router, and the various handlers a request progresses through.

router/src/lib.rs
```diff
@@ -1,14 +1,96 @@
...
+//!                      ║  │   Autocreation   │ ─║─ ─ ─ ─ ─ ─
+//!                      ║  └──────────────────┘  ║           │
+//!                      ║            │           ║
+//!                      ║            ▼           ║           │
+//!                      ║  ┌──────────────────┐  ║
+//!                      ║  │ NamespaceSchema  │  ║           │
+//!                      ║  │     Resolver     │ ─║─ ─ ─ ─ ─ ─
+//!                      ║  └──────────────────┘  ║           │
+//!                      ║            │           ║
+//!                      ╚════════════│═══════════╝           │
```
domodwyer, line 37: This is the subject of this PR - the namespace autocreator has moved from a `DmlHandler` to a `NamespaceResolver` and a new `NamespaceSchemaResolver` uses the namespace cache to discover the `NamespaceId` for the request without querying the catalog.

router/src/namespace_resolver.rs
```diff
@@ -0,0 +1,209 @@
...
+                // overwrites, but an entry will always exist once inserted, and
+                // the schemas will eventually converge.
+                self.cache
+                    .put_schema(namespace.clone(), Arc::clone(&schema));
+
+                trace!(%namespace, "schema cache populated");
+                Ok(schema.id)
+            }
+        }
+    }
```
domodwyer, line 97: This is the entirety of the new logic this PR adds - the rest of the diff is moving things around.

mkmik (approved):
love the diagrams

PR #6005, opened 2022-10-28
feat: Add the catalog service to ingester, querier, and compactor

carols10cents:
So that `remote get` that uses the catalog service can work no matter what kind of server you contact.

I believe this moots https://github.com/influxdata/influxdb_iox/pull/5999.

Closes https://github.com/influxdata/influxdb_iox/pull/5999.


alamb (approved):
LGTM -- thank you @carols10cents  -- we can probably consolidate the "namespace" service into CatalogService as well now that it is everywhere

Issue #5863, opened 2022-10-14
Document all HTTP error messages


domodwyer:
The IOx router returns responses to end-users. These look like:

```
{"code":"too many requests","message":"dml handler error: service limit reached: couldn't create column time in table 330331; limit reached on namespace"}
```

@rickspencer3 ~demands~ would like a list of all error strings in the `message` field for documentation purposes.

The errors returned are enumerated here: https://github.com/influxdata/influxdb_iox/blob/0044cd0f81ab1df6b6147194e6a1d2e5e4647c82/router/src/server/http.rs#L84-L136

I've created this ticket here as it also necessitates a code change in IOx; we should create fixture tests that ensure the string error messages are not accidentally changed in the future.

PR #6006, opened 2022-10-28
test: Ensure router's HTTP error messages are stable

shepmaster:
If you change the text of an error, the tests will fail.

If you add a new error variant to the `Error` enum but don't add it to the test, test compilation will fail with a "non-exhaustive patterns" message.

If you remove an error variant, test compilation will fail with a "no variant named `RemovedError`" message.

You can get the list of error variants and their current text via `cargo test -p router -- print_out_error_text --nocapture`.

A step towards accomplishing #5863

Co-authored-by: Carol (Nichols || Goulding) <carol.nichols@gmail.com>
Co-authored-by: Jake Goulding <jake.goulding@integer32.com>


carols10cents:
I was kind of intrigued by the challenges of this issue, and I nerdsniped @shepmaster with it too :)

shepmaster:
While @carols10cents and I were working on this, we noticed that some of the current error messages are... _suboptimal_.  Should we change some of these before documenting / committing to them? It doesn't necessarily have to be a part of this PR, but it could be.

/cc @rickspencer3

router/src/server/http.rs
```diff
@@ -1307,4 +1310,143 @@ mod tests {
...
+                    #[allow(unreachable_patterns)]
+                    match dummy_error {
+                        $(Error::$variant { .. } => {},)*
+                    }
+                }
+            }
+
+            #[tokio::test]
+            async fn error_messages_match() {
+                $(
```
domodwyer, line 1333:
```suggestion
            // A test that covers all errors given to this macro.
            async fn error_messages_match() {
                // Generate an assert for each error given to this macro.
                $(
```

router/src/server/http.rs
```diff
@@ -1307,4 +1310,143 @@ mod tests {
...
+                    let e = Error::$variant $(($data))?;
+                    assert_eq!(e.to_string(), $msg);
+                )*
+            }
+
+            #[test]
+            fn print_out_error_text() {
```
domodwyer, line 1340: Is this actually necessary? Wouldn't the failed asserts above spit out all the necessary strings?

domodwyer (commented):
Code is 👌 just some readability / comment suggestions.

Seems to have missed some important inner-error variants that are worth filling it - they happen to be the most likely errors a user will see!

I am totally cool with merging this and having fixture tests (helps avoid bad errors!) but I'm not sure we should commit to stable error messages this early on.

router/src/server/http.rs
carols10cents: This is actually for the case when all the asserts *pass* and we want to generate all the error messages to be transferred to some other documentation. If this doesn't sound useful, I'm happy to remove it!

domodwyer (approved):
Beautiful! Thanks for this 🙏

domodwyer:
Merging this so it doesn't get conflicted :+1:

PR #6011, opened 2022-10-31
refactor: move `SessionConfig` creation into `datafusion_utils`

alamb:
# Rationale:
Part of https://github.com/influxdata/influxdb_iox/pull/5930 / enabling parquet predicate pushdown.

Enabling parquet predicate pushdown should be done at session scope, but currently the creation of `SessionConfig` is done in `iox_query` (which is not, by design, shared with `parquet_file`)

# Changes:
Hoist the common session configuration into `datafusion_utils` which is shared by both so I can enable pushdown in one place and have it reflected in parquet_file


datafusion_util/src/config.rs
```diff
@@ -0,0 +1,24 @@
+use datafusion::{config::OPT_COALESCE_TARGET_BATCH_SIZE, prelude::SessionConfig};
```
alamb, line 1: All of this code is simply moved from elsewhere -- no changes are intended

NGA-TRAN (approved):
LGTM

PR #6013, opened 2022-11-01
docs: Document new CLI commands

alamb:
We have added some neat features recently like https://github.com/influxdata/influxdb_iox/issues/5967  and https://github.com/influxdata/influxdb_iox/pull/5734 -- they should be documented

Kudos to @NGA-TRAN  for reminding me to do this

I also spent some time reorganizing the CLI docs while I was at it

You can see the rendered version here: https://github.com/influxdata/influxdb_iox/blob/alamb/docs/docs/cli.md


NGA-TRAN (approved):
NIce. I have tried many of them. Thanks @alamb

PR #6014, opened 2022-11-01
feat: Add object_store handler to querier so `remote get-table` works

alamb:
# Rationale
When working on https://github.com/influxdata/influxdb_iox/pull/6013 I tried to run `remote get-table` against the querier and it didn't work. It turns out https://github.com/influxdata/influxdb_iox/pull/6005 from @carols10cents added the catalog service but not the object store service to the querier. 

So when I tried to do so it fails like:

```shell
$ influxdb_iox remote store get-table 26f7e5a4b7be365b_917b97a92e883afc mem 
found 6 Parquet files, downloading...
downloading file 1 of 6 (e194364e-3618-44f0-9b78-3bcf596dd28f.2.parquet)...
IOx request failed: Operation is not implemented or supported: 
```

# Changes
1. Add object store service to querier
2. Add basic end to end test

After this change ✅ 💰 

```shell
$ influxdb_iox remote store get-table 26f7e5a4b7be365b_917b97a92e883afc mem 
found 3 Parquet files, downloading...
downloading file 1 of 3 (1ce7e327-7b48-478f-b141-96e8d366ca12.5.parquet)...
downloading file 2 of 3 (fa45a0db-5e9e-4374-b3d3-8294b5e7ade0.5.parquet)...
downloading file 3 of 3 (ad5e47f6-b984-400b-99c2-f562151985d6.5.parquet)...
Done.
```


querier/src/handler.rs
```diff
@@ -28,16 +31,15 @@ pub enum Error {}
 /// The [`QuerierHandler`] does nothing at this point
 #[async_trait]
 pub trait QuerierHandler: Send + Sync {
-    /// Acquire a [`SchemaService`] gRPC service implementation.
-    ///
-    /// [`SchemaService`]: generated_types::influxdata::iox::schema::v1::schema_service_server::SchemaService.
```
alamb, line 33: this extra level of indirection is both unnecessary and confusing (as the returned type is not the service, it is the server)

crepererum (approved):

PR #6015, opened 2022-11-01
refactor: simplify `QueryChunk` data access

crepererum:
We have only two types for chunks (now that the RUB is gone):

1. In-memory RecordBatches
2. Parquet files

Loads of logic is duplicated in the different `read_filter` implementations. Also `read_filter` hides a solid amount of logic from DataFusion, which will prevent certain (future) optimizations. To enable #5897 and to simplify the interface, let the chunks return the data (batches or metadata for parquet files) directly and let `iox_query` perform the actual heavy-lifting.


iox_query/src/provider/physical.rs
```diff
@@ -125,32 +128,79 @@ impl ExecutionPlan for IOxReadFilterNode {
...
+                let stream = RecordBatchStreamAdapter::new(
+                    schema,
+                    futures::stream::once(execute_stream(Arc::new(exec), context)).try_flatten(),
+                );
+
+                // Note: No SchemaAdapterStream required here because `ParquetExec` already creates NULL columns for us.
+
+                Box::pin(stream)
+            }
+        };
```
NGA-TRAN, line 200: I agree we have `read_filter` everywhere which I think it was a good design for different types of chunks before but we started abusing this function name and found myself difficult to debug sometimes.
If you have better control and optimization here, removing the duplicate read filters makes sense

NGA-TRAN (approved):

iox_query/src/provider/physical.rs
```diff
@@ -125,32 +128,79 @@ impl ExecutionPlan for IOxReadFilterNode {
...
+                    object_store_url: exec_input.object_store_url,
+                    file_schema: Arc::clone(&schema),
+                    file_groups: vec![vec![PartitionedFile {
+                        object_meta: exec_input.object_meta,
+                        partition_values: vec![],
+                        range: None,
+                        extensions: None,
+                    }]],
+                    statistics: Statistics::default(),
+                    projection: None,
```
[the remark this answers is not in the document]
alamb: If not, then we can probably move the projection calculation code into the 
```rust
            QueryChunkData::RecordBatches(batches) => {
``` 

branch rather than always computing it

iox_query/src/provider/physical.rs
```diff
@@ -125,32 +128,79 @@ impl ExecutionPlan for IOxReadFilterNode {
...
+                let stream = Box::pin(MemoryStream::try_new(
+                    batches,
+                    incomplete_output_schema,
+                    projection,
+                )?);
+                let adapter = SchemaAdapterStream::try_new(stream, schema, baseline_metrics)
+                    .map_err(|e| DataFusionError::Internal(e.to_string()))?;
+                Box::pin(adapter) as SendableRecordBatchStream
+            }
+            QueryChunkData::Parquet(exec_input) => {
```
alamb, line 165: the eventual plan will be to take multiple `exec_input`s per `IOxReadFilterNode`, right?

parquet_file/src/storage.rs
```diff
@@ -107,6 +74,68 @@ impl std::fmt::Display for StorageId {
...
+    pub object_store_url: ObjectStoreUrl,
+
+    /// Object metadata.
+    pub object_meta: ObjectMeta,
+}
+
+impl ParquetExecInput {
+    /// Read parquet file into [`RecordBatch`]es.
+    ///
+    /// This is mostly for testing purposes.
```
alamb, line 95:
```suggestion
    /// This should only be used for testing purposes.
```

alamb (approved):
LGTM except for maybe the projection handling (about which I left a comment)

iox_query/src/provider/physical.rs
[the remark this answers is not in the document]
crepererum: No, because the `ParquetExec` doesn't need it. It will auto-project (and null-fill) files to the desired output schema. This is even more relevant once we stick multiple files into a single exec since files can have different schemas (for IOx: a different set of columns).

I think the projection field is a bit of a convenience thing. It can be used, but we don't need it here.

To clarify how this worked before: this particular phys. node here (`IOxReadFilter`) has a fixed schema. It then goes through the chunks and requests a projected schema from them and fills NULL cols that chunks cannot provide. This is still done for the read buffer case, but it's not required for the parquet version since `ParquetExec` does all of this internally.

iox_query/src/provider/physical.rs
crepererum: Yes. My plan is two mostly have two node types fused with some union in the end:

- `IOxReadFilter`: just contains read buffers and performs projection and null-filling
- `ParquetExec`: reads multiple parquet files

PR #6019, opened 2022-11-01
fix: be more conservative w/ "request canceled" messages

crepererum:
Slack context: https://influxdata.slack.com/archives/CRK9M8L5Q/p1667315716813859

We were seeing this message a lot on production when there were no canceled requests

### DIFF
--- a/trace_http/src/tower.rs
+++ b/trace_http/src/tower.rs
@@ -14,6 +14,7 @@
 
 use std::future::Future;
 use std::pin::Pin;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


cannonpalms:
@crepererum Can you explain the change you've made here and what problem the original implementation had?

cannonpalms:
Dumping the context I had pulled when looking at this in case it's ever relevant again in the future: we were seeing this ~2x per request, regardless of whether the request was being cancelled.
image
image

cannonpalms (approved):
Thanks for explaining! It helps me learn.

I tested this out and no longer see `request cancelled` being logged erroneously.

PR #6034, opened 2022-11-02
fix: InfluxQL parser incompatibilities

### LINKED ISSUE #6033: InfluxQL: Parser incompatibilties
A log of existing production queries was used to validate the completeness of the Rust parser. A few cases failed that should parse successfully and have been listed below.

# Failed statements

- [x] field references may be parsed as a 3-part name, however, they are treated as a quoted-identifier
  - For example

  ```sql
  SELECT a.b FROM b WHERE c.d = 'bar'`
  ```

  is parsed as

  ```sql
  SELECT "a.b" FROM b WHERE "c.d" = 'bar'`
  ```

- [x] Regular expressions that begin with a special character

  - `/\w.*/` is valid, but fails

- [x] Unable to parse microsecond duration literal

  - `218808603u` should be parsed as a microsecond, not `218808603us`
  - See https://github.com/influxdata/influxql/blob/7e7d61973256ffeef4b99edd0a89f18a9e52fa2d/parser.go#L3034-L3035

stuartcarnie:
Closes #6033

This PR fixes the incompatibilities identified in #6033


influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -539,6 +595,44 @@ mod test {
...
+        // Only parses 3 segments
+        let (rem, id) = segmented_identifier("part2.part1.part0.foo").unwrap();
+        assert_eq!(rem, ".foo");
+        assert_eq!(format!("{}", id), "\"part2.part1.part0\"");
+
+        // Quoted
+        let (rem, id) = segmented_identifier("\"part0\"").unwrap();
+        assert_eq!(rem, "");
+        assert_eq!(format!("{}", id), "part0");
+    }
```
NGA-TRAN, line 634: What is the expected of `r#""part" 2"."part 1".part0"#`, `r#""part" 2."part 1".part0"#`, `r#""part "2"."part 1".part0"#`,  `r#""part ""2"."part 1".part0"#`

Have a headache reviewing this 🤣

influxdb_influxql_parser/src/literal.rs
```diff
@@ -249,15 +249,15 @@ fn single_duration(i: &str) -> ParseResult<&str, i64> {
...
-                value(Microsecond, tag("us")),  // microseconds
-                value(Millisecond, tag("ms")),  // milliseconds
-                value(Second, tag("s")),        // seconds
-                value(Minute, tag("m")),        // minutes
-                value(Hour, tag("h")),          // hours
-                value(Day, tag("d")),           // days
-                value(Week, tag("w")),          // weeks
+                value(Nanosecond, tag("ns")),  // nanoseconds
+                value(Microsecond, tag("µ")), // microseconds
+                value(Microsecond, tag("u")),  // microseconds
```
NGA-TRAN, line 254: This is interesting

NGA-TRAN (approved):
Hats off to you for working on this work. So much detail

influxdb_influxql_parser/src/expression/arithmetic.rs
stuartcarnie: It does, doesn't it!

A `segmented_identifier` is defined as:

https://github.com/influxdata/influxdb_iox/blob/a356c465e0691060815c69d382c4a0b7ada1738f/influxdb_influxql_parser/src/expression/arithmetic.rs#L319-L325

and an `identifier` is defined as:

https://github.com/influxdata/influxdb_iox/blob/a356c465e0691060815c69d382c4a0b7ada1738f/influxdb_influxql_parser/src/identifier.rs#L3-L9

Therefore, each of the examples you provided would fail to parse, which I confirmed via 541c30667312a47982829ff7fe469b1bd9c73abb and also with the [InfluxQL Go parser](https://go.dev/play/p/k2150CJocVl)

stuartcarnie:
Thank you for the review, @NGA-TRAN  – I added those test cases you mentioned, and proved their compatibility with [this Go playground](https://go.dev/play/p/k2150CJocVl) of our Go parser.

PR #6035, opened 2022-11-02
fix: actually enable zstd compression for write write buffer

### LINKED ISSUE #5981: Enable write buffer ZSTD compression
Enable ZSTD compression for the write buffer.

While doing so, please disable the rskafka `snappy` feature since this is unused and all other cluster no longer have snappy-compressed data in them.

crepererum:
Fixes #5981.


cannonpalms (approved):
🎉

PR #6036, opened 2022-11-02
refactor(write_buffer): pass IDs in wire format

domodwyer:
This PR changes the router -> ingester write protocol message format to include the IDs of the table & namespaces for the write, as part of https://github.com/influxdata/influxdb_iox/issues/4880.

Note that this commit only **pushes the IDs on the sender-side** and the consumer **does not make use of them**. This will allow us a transition period where messages will contain both the names, and their IDs (see various large and shouty code comments).

In a follow-up PR I will remove the extra map added to the `DmlWrite` in this PR when switching the consumer to use IDs, so the size inflation is temporary.

---

* refactor(write_buffer): pass IDs in wire format (ddd6ab0ba)

      This commit is part of a two-part change in order to add the table &
      namespace IDs to the write buffer wire format. This commit forms the
      first half; changing the producer to send the IDs.
      
      In this commit the new ID values are never read on the consumer side,
      ensuring there is no consumer dependency on them. This ensures they
      remain operational during a rollout, where the consumer may be updated
      to the latest code dependent on the IDs before the producer is updated
      to send them. This also ensures we have a window of time where where the
      consumers can be rolled back after being updated, and still handle
      replaying messages in Kafka.

### DIFF
--- a/dml/src/lib.rs
+++ b/dml/src/lib.rs
@@ -15,7 +15,10 @@
 
 use std::time::Duration;
 
-use data_types::{DeletePredicate, NonEmptyString, PartitionKey, Sequence, StatValues, Statistics};
+use data_types::{
+    DeletePredicate, NamespaceId, NonEmptyString, PartitionKey, Sequence, StatValues, Statistics,
+    TableId,
+};
 use hashbrown::HashMap;
 use iox_time::{Time, TimeProvider};
 use mutable_batch::MutableBatch;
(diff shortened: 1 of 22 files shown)


dml/src/lib.rs
```diff
@@ -182,6 +185,32 @@ pub struct DmlWrite {
...
+    // consumer must not make use of them.
+    //
+    // In a follow-up PR, this consumer will be switched to make use of the
+    // TableIds, at which point the table map will change from the current
+    // `Table name -> Data` to `TableId -> Data`, and the second map can be
+    // removed from the DmlWrite.
+    #[allow(dead_code)]
+    namespace_id: NamespaceId,
+    // Used to resolve the table ID for a given table name during serialisation.
+    table_ids: HashMap<String, TableId>,
```
domodwyer, line 213: This describes the producer-only behaviour of this PR and the reasoning for it.

dml/src/lib.rs
```diff
@@ -293,6 +334,24 @@ impl DmlWrite {
...
+    }
+
+    /// Return the [`NamespaceId`] to which this [`DmlWrite`] should be applied.
+    ///
+    /// Marked unsafe because of the critical invariant; Kafka conumers MUST NOT
+    /// utilise this method until this warning is removed. See [`DmlWrite`]
+    /// docs.
+    pub unsafe fn namespace_id(&self) -> NamespaceId {
+        self.namespace_id
+    }
```
domodwyer, line 354: These methods are marked unsafe to ensure they are not accidentally used in a consumer - you have to explicitly opt into it, knowing there's an invariant at play.

write_buffer/src/codec.rs
```diff
@@ -204,7 +205,14 @@ pub fn decode(
...
                     Ok(DmlOperation::Write(DmlWrite::new(
                         headers.namespace,
+                        // Decoding MUST NOT make use of the (potentially empty)
+                        // namespace ID during the Kafka wire format change transition period.
+                        NamespaceId::new(0),
                         tables,
+                        // Decoding MUST NOT make use of the (potentially empty)
+                        // table IDs during the Kafka wire format change
+                        // transition period.
+                        HashMap::with_capacity(0),
```
domodwyer, line 215: When deserialising the `DmlWrite`, it is explicitly populated with nonsense and does not rely on the actual values in the message.

This ensures the consumer does not utilise the values until the follow-up PR.

write_buffer/src/codec.rs
```diff
@@ -235,7 +243,13 @@ pub fn encode_operation(
 ) -> Result<(), WriteBufferError> {
     let payload = match operation {
         DmlOperation::Write(write) => {
-            let batch = mutable_batch_pb::encode::encode_write(db_name, write);
+            // Safety: this code path is only invoked in the Kafka producer, so
+            // it is safe to utilise the ID.
+            //
+            // See DmlWrite docs for context.
+            let namespace_id = unsafe { write.namespace_id().get() };
```
domodwyer, line 250: This is an example of explicitly opting into using the IDs in the producer, knowing the underlying invariant is upheld.

mkmik (approved):

PR #6037, opened 2022-11-02
refactor: Rename `schema::selection::Selection` to `schema::projection::Projection`

alamb:
# Rationale:

In relational algebra:
1. Selection means "filter rows"
2. Projection means "select columns"

IOx was using `Selection` to represent `Projection` which was confusing (I was responsible for the original naming 😭 )

# Changes

1. Rename `schema::selection::Selection` to `schema::projection::Projection`
2. Add a `pub use`
3. Remove `HalfOwnedSelection` (which I think was used only by ReadBuffer)


schema/src/projection.rs
```diff
@@ -0,0 +1,137 @@
+use std::{collections::HashMap, fmt::Display, sync::Arc};
```
alamb, line 1: This is all the same as `selection.rs` except that:
1. The name is changed
2. A free function `select_schema` is moved to `Projection::project_schema`
3. `HalfOwnedSelection` et al are removed

NGA-TRAN (approved):
Massive but easy to review

crepererum (approved):

alamb:
CI appears to be stuck on this one

PR #6041, opened 2022-11-02
feat: grpc for updating namespace retention period

### LINKED ISSUE #6023: Support gRPC for setting its retention period
# Rationale

This is a task of #3053 and should be done after #6022 to provide 2 gRPCs:

1. Create name space with a specified hours of data retention

```shell
influxdb_iox namespace create <namespace_name> <num_hours>
```

2. Modify retention hours of a give namespace 

```shell
influxdb_iox namespace retention <namespace_name> <num_hours>
```
'num_hours` is an integer number >=0.  We will convert this num_hours into nanoseconds and record it into the `retention_period_nano` field of the `namespace` table.

If num_hours is 0, we will record `retention_period_nano`  as NULL to say it is infinite.

This task also:
- creates a default configuration parameter INFLUXDB_IOX_RETENTION_HOURS at set it to 0 which means infinite.
- modifies the current create namespace in Router (or Ingester?) to use NFLUXDB_IOX_RETENTION_HOURS to set the value of the namespace `retention_period_nano`

NGA-TRAN:
Connects to #6023

This PR support.  `influxdb_iox namespace retention --retention-hour <num_hours> <namespace>`
I will update the document after I do the second gRPC that create the namespace


generated_types/protos/influxdata/iox/schema/v1/service.proto
```diff
@@ -5,6 +5,9 @@ option go_package = "github.com/influxdata/iox/schema/v1";
 service SchemaService {
   // Get the schema for a namespace
   rpc GetSchema(GetSchemaRequest) returns (GetSchemaResponse);
+
+  // update retention period
+  rpc UpdateNamespaceRetention(UpdateNamespaceRetentionRequest) returns (UpdateNamespaceRetentionResponse);
```
alamb, line 10: I think this is reasonable for now. 

Longer term I recommend a  CRUD interfaces where you would pass back an entire namespace for update

That might look like 

```
  rpc GetNamespace(GetNamespaceRequest) returns (GetNamespaceResponse);
  rpc UpdateNamespace(UpdateNamespaceRequest) returns (UpdateNamespaceResponse);
...

message GetNamespaceRequest {
  // the name of the namespace
  string name = 1;
}


message UpdateNamespaceRequest {
  // the new namespace
  Namespace namespace = 1;
}
```

influxdb_iox/src/commands/namespace/mod.rs
```diff
@@ -25,17 +27,22 @@ pub struct Config {
 enum Command {
     /// Fetch namespaces
     List,
+
+    /// Create a new namespace
```
alamb, line 31: Maybe this should be "Update retention of an existing namespace"?

iox_catalog/src/interface.rs
```diff
@@ -295,6 +295,13 @@ pub trait NamespaceRepo: Send + Sync {
         query_pool_id: QueryPoolId,
     ) -> Result<Namespace>;
 
+    /// Update retention period for a namespace
+    async fn update_retention_period(
+        &mut self,
+        name: &str,
+        retention_hours: i64,
```
alamb, line 302: I find it somewhat confusing that the request contains retention in hours and the response contains retention in nanoseconds

alamb (approved):

data_types/src/lib.rs
```diff
@@ -442,6 +442,8 @@ pub struct Namespace {
     /// never drop data).
     #[sqlx(default)]
     pub retention_duration: Option<String>,
```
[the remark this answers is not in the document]
NGA-TRAN: This is the unused field and string type. I plan to remove this after we use the new one

generated_types/protos/influxdata/iox/schema/v1/service.proto
NGA-TRAN: Make sense. If we need to update other fields in the future, I will make it this way

influxdb_iox/src/commands/namespace/mod.rs
NGA-TRAN: Good catch

iox_catalog/src/interface.rs
NGA-TRAN: 😄 I wanted to make it easy for user to enter in hours but for us to compute in ns. No oneobjected at the design review. If we want to change this in the future, it is very easy to do so.

PR #6042, opened 2022-11-02
feat: push down all predicates

### LINKED ISSUE #6040: Allow more predicates to be pushed down
Currently we call:

https://github.com/influxdata/influxdb_iox/blob/3ba0458653dfa884ddb74c5c7ee2921dee046537/predicate/src/lib.rs#L487-L489

while building our scan plans here:

https://github.com/influxdata/influxdb_iox/blob/3ba0458653dfa884ddb74c5c7ee2921dee046537/iox_query/src/provider.rs#L255-L265

IIRC "execution engines" likely once referred to "ReadBuffer" -- which is now gone. Since we now only use DataFusion to implement filters (either via `FilterExec` or `ParquetExec`), we can push down more predicates.

While reading through the code I was wondering if we currently push down predicates through the deduplicator. We shouldn't do this for all predicates. Instead we must be a bit more careful.[^1]

[^1]: My current assumption is that we can push single and multi-column predicates that only affect primary-key columns (tags + time) but NOT fields throw de-dup. @alamb wasn't convinced by my hand-waving argument though :grin:

alamb:
~Draft until we fix https://github.com/influxdata/influxdb_iox/issues/6066~

# Rationale

Closes #6040 

With the advent of #5930 we are now pushing predicates down to the lowest level of the parquet scan. Previously we only tried to push down predicates that the ReadBuffer could handle. Now however, we can push them all down 

# Changes
1. Push all predicates down to the scan rather than just single column exprs (including, for example, IN list, etc)

# Open Questions
I am somewhat worried that in some cases this will actually slow queries down as the filters won't be very selective.

However, I want to start with gathering the data so that we can adjust heuristics of what filters are evaluated


alamb:
I am curious of reviewer opinion here 

Are there any other questions or tests I should look into running?

query_tests/cases/in/pushdown.expected
```diff
@@ -45,7 +45,7 @@
 |               |   CoalesceBatchesExec: target_batch_size=4096                                                                                    |
 |               |     FilterExec: CAST(count@0 AS Int64) > 200                                                                                     |
 |               |       RepartitionExec: partitioning=RoundRobinBatch(4)                                                                           |
-|               |         IOxReadFilterNode: table_name=restaurant, chunks=1 predicate=Predicate                                                   |
+|               |         IOxReadFilterNode: table_name=restaurant, chunks=1 predicate=Predicate exprs: [CAST(count AS Int64) > Int64(200)]        |
```
alamb, line 48: Here is some nice examples of more predicates pushed down more. 

These expressions should probably be "unwrapped" like `count > Float64(200)` -- I will file an upstream datafusion ticket

alamb: I believe the issue is that since `count` is unsigned, it is affected by https://github.com/apache/arrow-datafusion/issues/3702

Adding https://github.com/apache/arrow-datafusion/issues/3702 to https://github.com/influxdata/idpe/issues/16283

NGA-TRAN:
I start reviewing this

NGA-TRAN (commented):
It is very nice to see the expression down to IOxReadFilterNode.  🎉 

However, this will be perfect if there are no deduplication (chunks without time overlaps). If there are deduplication operators, we can only push predicates all the way down for certain cases.

Marco has example to show wrong results in the case of deduplication and predicate push down below it. I let him reproduce and then we may want to not pushdown for those cases before merging this

crepererum:
I think we should fix (or add a test) for "do not push through de-dup" first.

alamb:
Filed https://github.com/influxdata/influxdb_iox/issues/6066 to track filter issue. Will update the description of this ticket to match

query_tests/cases/in/pushdown.expected
```diff
@@ -37,34 +37,34 @@
...
+|               |     TableScan: restaurant projection=[count, system, time, town], partial_filters=[CAST(restaurant.count AS Int64) > Int64(200)]                           |
+| physical_plan | ProjectionExec: expr=[count@0 as count, system@1 as system, time@2 as time, town@3 as town]                                                                |
+|               |   CoalesceBatchesExec: target_batch_size=4096                                                                                                              |
+|               |     FilterExec: CAST(count@0 AS Int64) > 200                                                                                                               |
+|               |       RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                                     |
+|               |         ParquetExec: limit=None, partitions=[1/1/1/1/00000000-0000-0000-0000-000000000000.parquet], predicate=true, projection=[count, system, time, town] |
```
alamb, line 50: this "prediate = true" is due to https://github.com/apache/arrow-datafusion/issues/4020 (both a bad display and another missed chance to push filters -- will try and work on that in the next few days)

specifically "true" is the pruning predicate meaning the predicate (with cast(count...)) could not be pushed down as a pruning predicate. 

I think the reason that predicate can't be used as a pruning predicate is because it is unsigned and there is this issue with https://github.com/apache/arrow-datafusion/pull/4149 (which is fixed upstream, and should be fixed in the next version of datafusion)

alamb:
>  I think we should fix (or add a test) for "do not push through de-dup" first.

Given that @crepererum  has fixed https://github.com/influxdata/influxdb_iox/issues/6066 I think this PR could be merged now

NGA-TRAN (approved):
Yes ,let us merge it

PR #6044, opened 2022-11-03
feat: config param to set when partition is cold

### LINKED ISSUE #6038: Add config param for when to do cold compaction on partition
# Rationale

Currently we hard code 8 as the num of hours without new writes before we start doing cold compaction on that partition. Let us make it config param so we can change it easily.

Thought: do we also want to create config params for when we do hot compaction? Currently we hard code within 4 hours and then 24 hours

NGA-TRAN:
Closes #6038 


compactor/src/compact.rs
```diff
@@ -776,7 +778,9 @@ pub mod tests {
 
         // Create a compactor
         let time_provider = Arc::new(SystemProvider::new());
-        let config = make_compactor_config();
+        let mut config = make_compactor_config();
+        // 8 hours to be cold
+        config.minutes_without_new_writes_to_be_cold = 8 * 60;
```
NGA-TRAN, line 783: This was the test for it. I need to make it 8 hours (the hard code before this PR) to keep the tests the same

domodwyer (approved):

clap_blocks/src/compactor.rs
```diff
@@ -166,6 +166,15 @@ macro_rules! gen_compactor_config {
                 action
             )]
             pub max_num_compacting_files: usize,
+
+            /// Nmmber of minutes without a write to a partition before it is considered cold
```
alamb, line 170:
```suggestion
            /// Number of minutes without a write to a partition before it is considered cold
            /// and thus a candidate for compaction
```

clap_blocks/src/compactor.rs
```diff
@@ -166,6 +166,15 @@ macro_rules! gen_compactor_config {
...
+            #[clap(
+                long = "compaction-mimutes-without-new-writes-to-be-cold",
```
alamb, line 172:
```suggestion
                long = "compaction-minutes-without-new-writes-to-be-cold",
```

clap_blocks/src/compactor.rs
```diff
@@ -166,6 +166,15 @@ macro_rules! gen_compactor_config {
...
+                env = "INFLUXDB_IOX_COMPACTION_MINUTE_WITHOUT_NEW_WRITE_TO_BE_COLD",
```
alamb, line 173:
```suggestion
                env = "INFLUXDB_IOX_COMPACTION_MINUTES_WITHOUT_NEW_WRITE_TO_BE_COLD",
```

compactor/src/compact.rs
```diff
@@ -275,14 +275,15 @@ impl Compactor {
                 ("partition_type", compaction_type.into()),
             ]);
 
-            let time_8_hours_ago = Timestamp::from(self.time_provider.hours_ago(8));
```
alamb, line 278: 👍

clap_blocks/src/compactor.rs
```diff
@@ -166,6 +166,15 @@ macro_rules! gen_compactor_config {
...
+                default_value = "4 * 60",
```
alamb, line 174: I wonder why this the default value is 4 hours and not 8 hours  (what the hard coded value was) ?

alamb (approved):
I think there are a few typos in the parameter names

Thanks @NGA-TRAN 
cc @adrian-thurston

clap_blocks/src/compactor.rs
NGA-TRAN: I wanted to make it shorter as it is the desired now but I agree let us make it the same 8 hours and change the settings in IDPE for productions instead

NGA-TRAN:
I am looking into the test failure 
`end_to_end_cases::cli::compact_and_get_remote_partition`

PR #6047, opened 2022-11-03
feat: add ArcMap map type

domodwyer:
This is a helper type to DRY the frequent exactly-once initialisation of concurrency-safe map entries within the ingester, with an emphasis on minimising lock time.

It also contains some methods that are not currently used, because it's from "the future" (an ongoing branch for future PR-ing).

---

* feat: add ArcMap map type (17890a990)

      Implements a map of K -> Arc<V> with exactly-once initialisation
      semantics.
      
      This map can be used to ensure a given key maps to singleton instances
      of V; exactly what all the nodes in the ingester "buffer tree" of shard
      -> namespace -> table -> partition require.
      
      This impl contains unused funcs (silenced with an allow(dead_code)) due
      to it being picked from a future branch.

### DIFF
--- a/ingester/src/arcmap.rs
+++ b/ingester/src/arcmap.rs
@@ -153,6 +153,19 @@ where
         }
     }
 
+    /// Return a state snapshot of all the values in this [`ArcMap`] in
+    /// arbitrary order.
+    ///
+    /// # Concurrency
+    ///
+    /// The snapshot generation is serialised w.r.t concurrent calls to mutate
+    /// `self` (that is, a new entry may appear immediately after the snapshot
+    /// is generated). Calls to [`Self::values`] and other "read" methods
+    /// proceed in parallel.
+    pub(crate) fn values(&self) -> Vec<Arc<V>> {
+        self.map.read().values().map(Arc::clone).collect()
+    }
+
     fn compute_hash<Q: Hash + ?Sized>(&self, key: &Q) -> u64 {
         let mut state = self.hasher.build_hasher();
         key.hash(&mut state);
@@ -228,6 +241,23 @@ mod tests {
         let got = map.get_or_else(key, || Arc::new(13));
         assert_eq!(*got, 42);
         assert!(Arc::ptr_eq(&got, &other));
+    }
+
+    #[test]
+    fn test_values() {
+        let map = ArcMap::<usize, String>::default();
+
+        map.insert(&1, Arc::new("bananas".to_string()));
+        map.insert(&2, Arc::new("platanos".to_string()));
+
+        let mut got = map
+            .values()
+            .into_iter()
+            .map(|v| String::clone(&*v))
+            .collect::<Vec<_>>();
+        got.sort_unstable();
+
+        assert_eq!(got, &["bananas", "platanos"]);
     }
 
     #[test]
--- a/ingester/src/arcmap.rs
+++ b/ingester/src/arcmap.rs
@@ -1,0 +1,301 @@
+//! A map key-value map where values are always wrapped in an [`Arc`], with
+//! helper methods for exactly-once initialisation.
+
+#![allow(dead_code)]
+
+use std::{
+    borrow::Borrow,
+    hash::{BuildHasher, Hash, Hasher},
+    sync::Arc,
+};
+
+use hashbrown::{
+    hash_map::{DefaultHashBuilder, RawEntryMut},
+    HashMap,
+};
+use parking_lot::RwLock;
+
+/// A key-value map where all values are wrapped in [`Arc`]'s and shared across
+/// all readers of a given key.
+///
+/// Each key in an [`ArcMap`] is initialised exactly once, with subsequent
+/// lookups being handed an [`Arc`] handle to the same instance.
+#[derive(Debug)]
+pub(crate) struct ArcMap<K, V, S = DefaultHashBuilder> {
+    map: RwLock<HashMap<K, Arc<V>, S>>,
+    hasher: S,
+}
+
+impl<K, V, S> std::ops::Deref for ArcMap<K, V, S> {
+    type Target = RwLock<HashMap<K, Arc<V>, S>>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.map
+    }
+}
+
+impl<K, V> Default for ArcMap<K, V> {
+    fn default() -> Self {
+        Self {
+            map: Default::default(),
+            hasher: Default::default(),
+        }
+    }
+}
+
+impl<K, V, S> ArcMap<K, V, S>
+where
+    K: Hash + Eq,
+    S: BuildHasher,
+{
+    /// Fetch an [`Arc`]-wrapped `V` for `key`, or if this is the first lookup
+    /// for `key`, initialise the value with the provided `init` closure.
+    ///
+    /// # Concurrency
+    ///
+    /// This call is thread-safe - if two calls race, a value will be
+    /// initialised exactly once (one arbitrary caller's `init` closure will be
+    /// executed) and both callers will obtain a handle to the same instance of
+    /// `V`. Both threads will eagerly initialise V and race to "win" storing V
+    /// in the map.
+    ///
+    /// # Performance
+    ///
+    /// This method is biased towards read-heavy workloads, with many readers
+    /// progressing in parallel. If the value for `key` must be initialised, all
+    /// readers are blocked while `init` executes and the resulting `V` is
+    /// memorised.
+    pub(crate) fn get_or_else<Q, F>(&self, key: &Q, init: F) -> Arc<V>
+    where
+        Q: Hash + PartialEq<K> + ToOwned<Owned = K> + ?Sized,
+        F: FnOnce() -> Arc<V>,
+    {
+        // Memorise the hash outside of the lock.
+        //
+        // This allows the hash to be re-used (and not recomputed) if the value
+        // has to be inserted into the map after the existence check. It also
+        // obviously keeps the hashing outside of the lock.
+        let hash = self.compute_hash(key);
+
+        // First check if the entry exists already.
+        //
+        // This does NOT use an upgradable read lock, as readers waiting for an
+        // upgradeable read lock block other readers wanting an upgradeable read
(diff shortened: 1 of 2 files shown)


ingester/src/arcmap.rs
```diff
@@ -0,0 +1,301 @@
...
+        assert_eq!(*got, 42);
+
+        // And subsequent calls observe the same value, regardless of the init
+        // closure
+        let got = map.get_or_else(key, || Arc::new(13));
+        assert_eq!(*got, 42);
+
+        let got = map.get_or_default(key);
+        assert_eq!(*got, 42);
+    }
```
domodwyer, line 207: This test captures the gist of this type - you can add things to the RW locked map exactly once without having to do the `read_lock(), check if entry exists, drop lock, write_lock(), check if entry exists, maybe insert value or maybe use existing value, drop lock` dance; a frequent pattern in the ingester.

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+    S: BuildHasher,
+{
+    /// Fetch an [`Arc`]-wrapped `V` for `key`, or if this is the first lookup
+    /// for `key`, initialise the value with the provided `init` closure.
+    ///
+    /// # Concurrency
+    ///
+    /// This call is thread-safe - if two calls race, a value will be
+    /// initialised exactly once (one arbitrary caller's `init` closure will be
+    /// executed) and both callers will obtain a handle to the same instance of
```
alamb, line 58: Can you please make it clear here that the other caller's closure will never be called and simply dropped (if that is indeed the case)?

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// `V`. Both threads will eagerly initialise V and race to "win" storing V
+    /// in the map.
+    ///
+    /// # Performance
+    ///
+    /// This method is biased towards read-heavy workloads, with many readers
+    /// progressing in parallel. If the value for `key` must be initialised, all
+    /// readers are blocked while `init` executes and the resulting `V` is
+    /// memorised.
+    pub(crate) fn get_or_else<Q, F>(&self, key: &Q, init: F) -> Arc<V>
```
alamb, line 68: I would personally find a name like `get_or_insert_with` clearer to understand the intent (as map can be modified and `self` is not `mut`)  though I can see the rationale to use `or_else` to mirror `Option::unwrap_or_else` and related functions

```suggestion
    pub(crate) fn get_or_insert_with<Q, F>(&self, key: &Q, init: F) -> Arc<V>
```

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+    where
+        Q: Hash + PartialEq<K> + ToOwned<Owned = K> + ?Sized,
+        F: FnOnce() -> Arc<V>,
+    {
+        // Memorise the hash outside of the lock.
```
alamb, line 73:
```suggestion
        // Compute the hash outside of the lock.
```

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+        // has to be inserted into the map after the existence check. It also
+        // obviously keeps the hashing outside of the lock.
+        let hash = self.compute_hash(key);
+
+        // First check if the entry exists already.
+        //
+        // This does NOT use an upgradable read lock, as readers waiting for an
+        // upgradeable read lock block other readers wanting an upgradeable read
+        // lock. If all readers do that, it's effectively an exclusive lock.
+        if let Some((_, v)) = self.map.read().raw_entry().from_hash(hash, |q| key == q) {
```
alamb, line 85: 👍  this is the fast path and looks good

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+            RawEntryMut::Occupied(v) => Arc::clone(v.get()),
+            RawEntryMut::Vacant(v) => {
+                Arc::clone(v.insert_hashed_nocheck(hash, key.to_owned(), init()).1)
+            }
+        }
+    }
+
+    /// A convenience method over [`Self::get_or_else()`] that initialises `V`
+    /// to the default value when `key` has no entry.
+    pub(crate) fn get_or_default<Q>(&self, key: &Q) -> Arc<V>
```
alamb, line 103:
```suggestion
    pub(crate) fn get_or_insert_default<Q>(&self, key: &Q) -> Arc<V>
```

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+            .from_hash(hash, |q| key == q)
+            .map(|(_k, v)| Arc::clone(v))
+    }
+
+    /// Insert `value` indexed by `key`.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if a value already exists for `key`.
+    pub(crate) fn insert<Q>(&self, key: &Q, value: Arc<V>)
```
alamb, line 137: It is not clear why this method is `pub(crate)` -- maybe it could remain private to the struct

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        let key: &str = "bananas";
+
+        // Assert the value is initialised from the closure
+        let got = map.get_or_else(key, || Arc::new(42));
+        assert_eq!(*got, 42);
+
+        // And subsequent calls observe the same value, regardless of the init
+        // closure
+        let got = map.get_or_else(key, || Arc::new(13));
```
alamb, line 214: 👍

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    #[test]
+    #[allow(clippy::needless_collect)] // Only needless if you like deadlocks.
+    fn test_init_once_parallel() {
+        let map = Arc::new(ArcMap::<String, usize>::default());
+
+        const NUM_THREADS: usize = 10;
+
+        let barrier = Arc::new(Barrier::new(NUM_THREADS));
+        let init_count = Arc::new(AtomicUsize::new(0));
```
alamb, line 281: 👍

ingester/src/arcmap.rs
```diff
@@ -0,0 +1,331 @@
...
+                    });
+
+                    *got == i
+                })
+            })
+            .collect::<Vec<_>>();
+
+        let winners = handles
+            .into_iter()
+            .fold(0, |acc, h| if h.join().unwrap() { acc + 1 } else { acc });
```
alamb, line 313:
```suggestion
            .reduce(|acc, h| if h.join().unwrap() { acc + 1 } else { acc });
```

alamb: maybe - I didn't test it

alamb (approved):
Makes sense to me and I think the code is very well tested 🏅 

I had a few minor style suggestions that might help to make it better.

Nice work @domodwyer

ingester/src/arcmap.rs
domodwyer: I think you missed the lines above the bit you selected:

https://github.com/influxdata/influxdb_iox/blob/30f69ce4f6a0db1f25dc3f9b79887747af9ac4ff/ingester/src/arcmap.rs#L51-L52

The value in the closure is not necessarily dropped, so I don't think we should say that - they might have retained a ref to the `Arc<T>` they are inserting.

ingester/src/arcmap.rs
domodwyer: As you point out, [this](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) [is](https://doc.rust-lang.org/std/option/enum.Option.html#method.or_else) [an](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) [established](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else) [naming](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_or_else) [convention](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else) - I think established naming conventions should beat out personal preference as the latter is... personal! I find the consistent `<action>_or_else()` usage in the rust code base a blessing compared to the mish-mash of names in other languages.

ingester/src/arcmap.rs
domodwyer: Likewise this is an established naming convention

ingester/src/arcmap.rs
domodwyer: I'm not sure I understand - all the methods are `pub(crate)`, except the private ones.

ingester/src/arcmap.rs
domodwyer: This unfortunately won't work which is a little sad - `reduce()` takes the first element **of the iterator** as the starting value (which is a thread handle) rather than the default of the accumulator type.

domodwyer:
(merging because I need this to open the next PR)

ingester/src/arcmap.rs
alamb: I read that line and thought it could be clarified. I meant the closure (not whatever was bound inside it) would be dropped. If you think it is clear enough that is fine with me.

ingester/src/arcmap.rs
alamb: I guess I was thinking of following the convention of `std::HashMap`: https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#method.or_insert_with

ingester/src/arcmap.rs
alamb: Yeah, sorry I should have been clearer -- I meant that this seemed like an internal implementation that might be hard to use (aka panics under concurrent update)  but now I see there is no equivalent of `insert_or_else` 

It will probably be clear when I see the usage of this structuer

ingester/src/arcmap.rs
mkmik: Option and HashSet have `get_or_insert_with`.

domodwyer: This is a better reason! https://github.com/influxdata/influxdb_iox/pull/6065

PR #6048, opened 2022-11-03
refactor(router): pass NamespaceId for deletes

domodwyer:
This PR changes the router to resolve the `NamespaceId` for each delete request, and push it down the `DmlHandler` stack. It's basically the delete counterpart to #6000.

This value is currently unused.

---

* refactor(router): pass NamespaceId for deletes (a7835009d)

      Pushes the NamespaceId down the DML handler stack like we do for writes.


alamb (approved):
This is a neat example of incremental development 👍

PR #6052, opened 2022-11-03
refactor(write_buffer): propagate NamespaceId across Kafka boundary for DML deletes

domodwyer:
This is the delete counterpart to #6036, part of https://github.com/influxdata/influxdb_iox/issues/4880.

---

* feat: NamespaceId in DmlDelete (6fa48731a)

      Changes the DmlDelete to contain the NamespaceId for which it should be
      applied, propagating this value over the wire.
      
      Like the existing IDs within the DmlWrite, these values are marked
      unsafe to use due to avoid the consumers utilising them accidentally
      during deployment. Unlike DmlWrite, the DmlDelete is completely unused,
      so this is less of an issue.

* test(router): assert DmlHandler ID propagation (906645c5b)

      Adds an integration test asserting that the NamespaceId and Table IDs
      are propagated through the router.

### DIFF
--- a/dml/src/lib.rs
+++ b/dml/src/lib.rs
@@ -158,6 +158,20 @@ impl DmlOperation {
             Self::Delete(d) => d.namespace(),
         }
     }
+
+    /// Namespace catalog ID associated with this operation
+    ///
+    /// # Safety
+    ///
+    /// Marked unsafe because of the critical invariant; Kafka conumers MUST NOT
+    /// utilise this method until this warning is removed. See [`DmlWrite`]
+    /// docs.
+    pub unsafe fn namespace_id(&self) -> NamespaceId {
+        match self {
+            Self::Write(w) => w.namespace_id(),
+            Self::Delete(d) => d.namespace_id(),
+        }
+    }
 }
 
 impl From<DmlWrite> for DmlOperation {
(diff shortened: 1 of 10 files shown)


influxdb_iox_client/src/client/delete.rs
```diff
@@ -53,6 +53,7 @@ pub mod generated_types {
 /// client
 ///     .delete(
 ///         "my_db",
+///         42,
```
carols10cents, line 56: oh, hai. i leveled up ur warlock

dml/src/lib.rs
```diff
@@ -158,6 +158,20 @@ impl DmlOperation {
             Self::Delete(d) => d.namespace(),
         }
     }
+
+    /// Namespace catalog ID associated with this operation
+    ///
+    /// # Safety
+    ///
+    /// Marked unsafe because of the critical invariant; Kafka conumers MUST NOT
```
carols10cents, line 166: 🤣 I don't think I've seen `unsafe` used in quite this way before, but... ok ;)

crepererum (approved):

influxdb_iox_client/src/client/delete.rs
domodwyer: Amazing on many levels - thanks for fixing the tests 🙏

PR #6053, opened 2022-11-03
fix: use `ORDER BY` to prevent `INSERT` deadlocks

carols10cents:
Based on this: https://dba.stackexchange.com/a/195220/27897

Connects to https://github.com/influxdata/idpe/issues/16298

I have some cleanup to do and probably tests to fix, but @alamb could you try this out with your setup and see if it gets rid of the deadlocks for you?


alamb:
@carols10cents  I will test this tomorrow morning

aierui:
Is it the solution to this problem?

```log
{"timestamp":"2022-11-03T07:00:36.193402Z","level":"WARN","fields":{"message":"INSERT INTO column_name ( …; rows affected: 0, rows returned: 0, elapsed: 1.014s\n\nINSERT INTO\n  column_name (name, table_id, column_type)\nSELECT\n  name,\n  $1,\n  column_type\nFROM\n  UNNE
ST($2, $3) as a(name, column_type) ON CONFLICT ON CONSTRAINT column_name_unique DO\nUPDATE\nSET\n  name = column_name.name RETURNING *;\n","log.target":"sqlx::query","log.module_path":"sqlx::query"},"target":"sqlx::query"}
{"timestamp":"2022-11-03T07:00:36.820250Z","level":"WARN","fields":{"message":"request cancelled","trace":"","when":"before fully returning body data"},"target":"trace_http::tower"}
{"timestamp":"2022-11-03T07:00:37.170341Z","level":"WARN","fields":{"message":"error occurred while testing the connection on-release: error returned from database: deadlock detected","log.target":"sqlx_core::pool::connection","log.module_path":"sqlx_core::pool::connection","log.file":"/root/.cargo/registry/src/rsproxy.cn-8f6827c7555bfaf8/sqlx-core-0.6.2/src/pool/connection.rs","log.line":242},"target":"sqlx_core::pool::connection"}
```

carols10cents:
@aierui Yes, hopefully. Can you try this branch and see if it fixes the deadlock for you?

alamb:
So I ran the reproducer on https://github.com/influxdata/idpe/issues/16298#issue-1427259604 against this branch

In all cases I no longer saw the write rejected by IOx. 👍 

Interestingly, I did see this in the postgres log:

```
2022-11-04 14:03:42.113 UTC [242] ERROR:  duplicate key value violates unique constraint "namespace_name_unique"
2022-11-04 14:03:42.113 UTC [242] DETAIL:  Key (name)=(test_db2) already exists.
2022-11-04 14:03:42.113 UTC [242] STATEMENT:  
        INSERT INTO namespace ( name, retention_duration, topic_id, query_pool_id )
        VALUES ( $1, $2, $3, $4 )
        RETURNING *;
```

But it does not appear to have affected the load. 

Nice work @carols10cents

carols10cents:
@alamb Yes, I saw those namespace name conflicts as well, and thought about fixing them to do upserts instead, but on investigation [it looks intentional](https://github.com/influxdata/influxdb_iox/blob/8c8e607dcaf041448c18b89cac821fdadd1d42bf/iox_catalog/src/postgres.rs#L611-L614). 

I'll squash this down and document a bit more then submit for review. 🎉

carols10cents:
Ok, I think this is ready for review!

iox_catalog/src/postgres.rs
```diff
@@ -947,19 +946,25 @@ WHERE table_id = $1;
...
+        let (v_name, v_column_type): (Vec<&str>, Vec<i16>) = columns
+            .iter()
+            .map(|(&name, &column_type)| (name, column_type as i16))
+            .unzip();
+
+        // The `ORDER BY` in this statement is important to avoid deadlocks during concurrent
+        // writes to the same IOx table that each add many new columns. See:
+        //
+        // - <https://rcoh.svbtle.com/postgres-unique-constraints-can-cause-deadlock>
+        // - <https://dba.stackexchange.com/a/195220/27897>
```
alamb, line 961:
```suggestion
        // - <https://dba.stackexchange.com/a/195220/27897>
        // - <https://github.com/influxdata/idpe/issues/16298>
```

iox_catalog/src/interface.rs
```diff
@@ -369,7 +359,7 @@ pub trait ColumnRepo: Send + Sync {
     async fn create_or_get_many_unchecked(
         &mut self,
         table_id: TableId,
-        columns: &[ColumnUpsertRequest<'_>],
+        columns: HashMap<&str, ColumnType>,
```
alamb, line 362: what was the rationale for using a HashMap here? Is it so that the validation after catalog insert could be made faster?

```rust
            let want = columns.get(existing.name.as_str()).unwrap();
```

I was just wondering if we had somehow masked the error by simply changing the order of columns passed to the SQL statement (as now it is the hash of the column name rather than the order in the write) 🤔

iox_catalog/src/postgres.rs
```diff
@@ -947,19 +946,26 @@ WHERE table_id = $1;
...
+        // - <https://github.com/influxdata/idpe/issues/16298>
         let out = sqlx::query_as::<_, Column>(
             r#"
 INSERT INTO column_name ( name, table_id, column_type )
-SELECT name, $1, column_type FROM UNNEST($2, $3) as a(name, column_type)
+SELECT name, $1, column_type
+FROM UNNEST($2, $3) as a(name, column_type)
+ORDER BY name
```
alamb, line 968: I have no idea how much, if any, of an impact adding `ORDER BY` will have on our production catalog. I did some quick tests and it seems like adding order by slows it down by about 1.5-2x. 

I don't know if this will cause problems but it seems ok to me

```sql
iox_catalog_prod=> select * from column_name limit 10;
REDACTED

Time: 141.647 ms
iox_catalog_prod=> select * from column_name order by name limit 10; 
REDACTED
(10 rows)

Time: 320.348 ms
iox_catalog_prod=> select * from column_name order by name desc limit 10; 
REDACTED
(10 rows)

Time: 218.647 ms
iox_catalog_prod=> 
```

alamb (approved):
Looks good to me -- thank you @carols10cents . Previously I have validated it fixes my reproducer and I also tried to quantify its impact on the production catalog, which seems minimal 

I think @domodwyer should also approve this prior to merge

iox_catalog/src/interface.rs
carols10cents: I tried to document my rationale in the 2nd commit's message, does this explain it enough?

> The checks for whether a column already exists with a different type
were relying on ordering of the input matching the ordering of the
columns returned from inserting the columns in Postgres.
>
>Rather than trying to match the new ordering that is required to avoid
Postgres deadlocks, switch from a Vec to a HashMap and look up the
column type from the name.
>
> This also reduces some allocations that weren't really needed.

iox_catalog/src/postgres.rs
carols10cents: This `SELECT` isn't selecting from a table, though, it's selecting from the `UNNEST` of the parameters passed in from Rust, so I would hope they would perform better than selects from the whole `column_name` table?

iox_catalog/src/interface.rs
alamb: yes, thank you, sorry I do not typically look at messages on individual commits during code review -- I look at the description on the PR and comments in the code

iox_catalog/src/postgres.rs
alamb: > This SELECT isn't selecting from a table, though, it's selecting from the UNNEST of the parameters passed in from Rust, so I would hope they would perform better than selects from the whole column_name table?

That is an excellent point

iox_catalog/src/lib.rs
```diff
@@ -149,47 +150,41 @@ where
...
+                    new: col.influx_type(),
                 }
+                .fail();
             }
-        })
-        .collect::<Result<Vec<_>>>()?;
+            None => {
+                // The column does not exist in the cache, add it to the column
+                // batch to be bulk inserted later.
+                column_batch.insert(name.as_str(), ColumnType::from(col.influx_type()));
```
domodwyer, line 179: Perhaps this is a bit excessive, but I usually assert this wasn't an _update_ despite the method name (it always gets me!):

```suggestion
                let old = column_batch.insert(name.as_str(), ColumnType::from(col.influx_type()));
                assert!(old.is_none(), "duplicate column name in new column batch");
```

Replacing an existing entry **should** never happen with the current code 🤞, and the old code perhaps should have checked there were no duplicates too. With the new hashmap approach this check is _cheap_ though 🎉 

(this could also be applied to the TSM merging logic if you fancy)

iox_catalog/src/postgres.rs
```diff
@@ -947,19 +946,26 @@ WHERE table_id = $1;
...
-        let mut v_column_type = Vec::new();
-        for c in columns {
-            v_name.push(c.name.to_string());
-            v_column_type.push(c.column_type as i16);
-        }
-
+        let num_columns = columns.len();
+        let (v_name, v_column_type): (Vec<&str>, Vec<i16>) = columns
+            .iter()
+            .map(|(&name, &column_type)| (name, column_type as i16))
```
domodwyer, line 954: I like that this is reusing the `str` ref - I dunno why I had this copying the string? 👍

iox_catalog/src/postgres.rs
```diff
@@ -947,19 +946,26 @@ WHERE table_id = $1;
...
+            .unzip();
+
+        // The `ORDER BY` in this statement is important to avoid deadlocks during concurrent
+        // writes to the same IOx table that each add many new columns. See:
+        //
+        // - <https://rcoh.svbtle.com/postgres-unique-constraints-can-cause-deadlock>
+        // - <https://dba.stackexchange.com/a/195220/27897>
```
domodwyer, line 961: Thank you for including the relevant reference material ❤️ 

I learnt something new today! In hindsight it makes perfect sense (doesn't it always?)

iox_catalog/src/postgres.rs
domodwyer: I suspect this will be fine. Thanks for checking though @alamb, this was the first thing that I set out to check during PR when I saw the title :+1:

Fortunately we have good call latency instrumentation of all catalog calls, so it should be easy to validate the perf in prod too :+1:

iox_catalog/src/postgres.rs
```diff
@@ -2820,19 +2827,6 @@ mod tests {
...
-        intra_request_type_conflict,
-        calls = {
-            [
-                "test1" => ColumnType::String,
-                "test1" => ColumnType::Bool,
-            ]
-        },
-        want = Err(Error::SqlxError{ .. })
-    );
-
```
domodwyer, line 2835: I love that this error state is no longer representable because of the choice of types / type system - this is my favourite type of maintainability win :+1:

domodwyer (approved):
Thanks for tracking this down 🙏

PR #6056, opened 2022-11-04
feat: make logging clearer when parquet files upload is retried

alamb:
# Rationale

While working on https://github.com/influxdata/idpe/issues/16373 I saw errors like this in greylog:

```
level=error msg="failed to upload parquet file to object storage" error="Generic S3 error: Error performing put request 3997/39652/34/13213104/50c9679c-9003-4e0f-99b7-adbe7de82c1e.parquet:...
```

This made me temporarily think that there was some aws s3 error causing issues. After examining the code I think this is just normal s3 hiccup and the upload succeeded on retry

# Changes
1. Make the log message into a warning and then also log when a retried request subsequently succeeds


parquet_file/src/storage.rs
```diff
@@ -242,9 +242,15 @@ impl ParquetStorage {
...
+        let mut retried = false;
         while let Err(e) = self.object_store.put(&path, data.clone()).await {
-            error!(error=%e, ?meta, "failed to upload parquet file to object storage");
+            warn!(error=%e, ?meta, "failed to upload parquet file to object storage, retrying");
             tokio::time::sleep(Duration::from_secs(1)).await;
+            retried = true;
+        }
+
+        if retried {
+            info!(?meta, "Succeed uploading files to object storage on retry");
```
carols10cents, line 253: Hm, I expected this message to be at the same level as the `warn!` in the retry loop. If the log level specified is `warn` (does anyone ever do that? I don't know, I feel like I only use info, debug, or trace), the logs will show the "failed to upload" but not this message.

Also, nit, but I think it should be in past tense?

```suggestion
            info!(?meta, "Succeeded uploading files to object storage on retry");
```

alamb: > Hm, I expected this message to be at the same level as the warn! in the retry loop. 

It is a good point -- I don't think it is common to use `LOG_FILTER=warn` 

>  If the log level specified is warn  the logs will show the "failed to upload" but not this message.

Yes, that is what is happening now actually (though the logs are actually reporting "error") and it is very confusing

alamb: Perhaps @crepererum  would have some insight next week

crepererum: I don't have a strong opinion on this topic. First of all, I think using `warn!` instead of `error!` is reasonable. Regarding the "Succeeded" bit, there are two ways to read / reason here:

- if you enable the warning log for failed uploads (via filters), you may also wanna see the "Succeeded" message
- the "Succeeded" is an info because there is nothing to warn about anymore.

From a shifter/debugger PoV, I would find the first the variant more helpful.

alamb: What did during my debugging was 
1. Notice there were errors in the log
2. Find one of the relevant parquet uuids that had an error and search for other log entries with it

Since there were no other log entries, I then had to go look at the code to figure out if the error was a problem or not.

I don't have a super strong opinion on the relative levels here -- I mildly prefer the way I have it here (which would have worked for my debugging procedure) but I am also happy to have them both be warn (or info) -- whatever you both will accept 🙏

alamb:
@crepererum  are you ok with the changes in this PR? If not, what shall I change to get it approved?

PR #6060, opened 2022-11-04
chore: expose compactor's hot compaction hours thresholds as cfg

### LINKED ISSUE #6059: Add config param for when to do hot compaction on a partition
See also #6038

Currently the number of hours before doing hot compactions is hardcoded. Make it configurable.

lukebond:
Closes #6059 

In #6044 Nga made it possible to configure how old a partition needs to be in order to be considered cold. This does the same for hot partitions.


clap_blocks/src/compactor.rs
```diff
@@ -176,6 +176,24 @@ macro_rules! gen_compactor_config {
...
+            pub hot_compaction_hours_threshold_1: u64,
+
+            /// When querying for partitions with data for hot compaction, how many hours to look
+            /// back for a second pass if we found nothing in the first pass.
+            #[clap(
+                long = "compaction-hot-partition-hours-threshold-2",
+                default_value = "24",
+                action
+            )]
+            pub hot_compaction_hours_threshold_2: u64,
```
NGA-TRAN, line 196: You want to add env = "INFLUXDB_IOX_COMPACTION..." for these 3 config param so we can add them in k8s_idpe

compactor/src/compact.rs
```diff
@@ -258,7 +258,8 @@ impl Compactor {
     /// Return a list of partitions that:
     ///
     /// - Have not received any writes in 8 hours (determined by all level 0 and level 1 parquet
-    ///   files having a created_at time older than 8 hours ago)
+    ///   files having a created_at time older than 8 hours ago). Note that 8 is the default but
+    ///   it's configurable
```
NGA-TRAN, line 262: ❤️

compactor/src/compact.rs
```diff
@@ -779,7 +785,7 @@ pub mod tests {
         // Create a compactor
         let time_provider = Arc::new(SystemProvider::new());
         let mut config = make_compactor_config();
-        // 8 hours to be cold
+        // 8 hours to be cold (this is the default)
```
NGA-TRAN, line 788: For tests, I created a config with time to be cold to 10 minutes. See here https://github.com/influxdata/influxdb_iox/blob/654ed98d1f5bb77243417e1e5de0738d5bd44d98/compactor/src/compact.rs#L692

In this test, I make it 8 hours so all behaviors in the tests are the same as it was with previous default 8 hours. I think it is appropriate not to mention the default here. We just test with 8 hours
```suggestion
        // 8 hours to be cold
```

compactor/src/compact.rs
```diff
@@ -848,7 +854,7 @@ pub mod tests {
         assert!(candidates.is_empty());
 
         // --------------------------------------
-        // Case 3: no new recent writes (within the last 8 hours) --> return that partition
+        // Case 3: no new recent writes (within the last 8 hours, which is the default) --> return that partition
```
NGA-TRAN, line 857:
```suggestion
        // Case 3: no new recent writes (within the last 8 hours) --> return that partition
```

Same here

compactor/src/hot.rs
```diff
@@ -84,7 +85,11 @@ pub(crate) async fn hot_partitions_to_compact(
 
     // Get the most recent highest ingested throughput partitions within the last 4 hours. If not,
     // increase to 24 hours.
-    let query_times = query_times(compactor.time_provider());
+    let query_times = query_times(
+        compactor.time_provider(),
+        compactor.config.hot_compaction_hours_threshold_1,
+        compactor.config.hot_compaction_hours_threshold_2,
```
NGA-TRAN, line 91: 🎉

NGA-TRAN (commented):
I think if you add the 2 env for the new params it will be good. And minor comment changes

clap_blocks/src/compactor.rs
lukebond: i thought it was done automatically but think i'm confusing it with our golang codebase

NGA-TRAN (approved):
Thanks @lukebond . If we add these 2 env params into idpe, we can adjust them now. I guess so far there are no demands to change them. We may just keep the settings 4 and 24 for now

PR #6061, opened 2022-11-04
chore: Update datafusion and arrow/arrow-flight/parquet to `26.0.0`

alamb:
~Draft until https://github.com/apache/arrow-datafusion/pull/4039 is merged~

# Rationale
* keep up with dependencies
* get fix for timezone handling https://github.com/apache/arrow-datafusion/issues/3938

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/dd081d64a2fba8574e63bdd0662c14aec5852b48
1. Update arrow/arrow-flight/parquet to `26.0.0`
2. Update flatbuffers to `"22.9.29"` (and resolves "RUSTSEC-2021-0122") 🎉 
3. Run `cargo update` and check in the results
4. Will comment inline on changes


alamb:
🤔  workspace hacks seem to be failing due to the workspace hack issue identified by @carols10cents  on https://github.com/influxdata/influxdb_iox/pull/6032#issuecomment-1303987709

I tried to go back to `0.8.0` with

```shell
cargo update   --precise 'ahash@0.8.0'
```

and 

```shell
cargo update   -p 'ahash@0.8.0'
```

But neither of them worked 😭

carols10cents:
I pushed a commit that changes `arrow_util`'s `ahash` requirement to `=0.8.0` and that seems to do the trick?

arrow_util/Cargo.toml
```diff
@@ -7,7 +7,9 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
-ahash = { version = "0.8.0", default-features = false, features = ["runtime-rng"] }
+# Locking ahash to 0.8.0 because of a cargo hakari problem, see:
```
alamb, line 10: My hero 🦸‍♀️ !

iox_query/src/test.rs
```diff
@@ -601,7 +601,7 @@ impl TestChunk {
                 DataType::Int64 => Arc::new(Int64Array::from(vec![1000])) as ArrayRef,
                 DataType::Utf8 => Arc::new(StringArray::from(vec!["MA"])) as ArrayRef,
                 DataType::Timestamp(TimeUnit::Nanosecond, _) => {
-                    Arc::new(TimestampNanosecondArray::from_vec(vec![1000], None)) as ArrayRef
+                    Arc::new(TimestampNanosecondArray::from(vec![1000])) as ArrayRef
```
alamb, line 604: these changes are needed due to changes upstream in DataFusion

query_functions/src/lib.rs
```diff
@@ -190,8 +190,8 @@ mod test {
             "+----------------------------+-------------------------------+",
             "| time                       | bound                         |",
             "+----------------------------+-------------------------------+",
-            "| 1970-01-01 00:00:00.000001 | 1970-01-01 00:00:00.000001100 |",
-            "| 1970-01-01 00:00:00.000002 | 1970-01-01 00:00:00.000002100 |",
+            "| 1970-01-01T00:00:00.000001 | 1970-01-01T00:00:00.000001100 |",
```
alamb, line 193: DataFusion now formats timestamps (almost) like InfluxData (doesn't yet have the trailing`Z`)

deny.toml
```diff
@@ -11,11 +11,6 @@ ignore = [
     # why needed: used by `chrono`
     # upstream issue: https://github.com/chronotope/chrono/issues/553
     "RUSTSEC-2020-0071",
-
```
alamb, line 14: 🎉  @tustvold  fixed the upstream thing in flatbuffers and we have finally gotten it into IOx

query_functions/src/lib.rs
tustvold: The trailing Z would actually be incorrect - as we don't specify a timezone for our timestamps.

alamb: > The trailing Z would actually be incorrect - as we don't specify a timezone for our timestamps.

"not yet" for sure

Since all the timestamps in IOx are UTC I think we  eventually will make them match the arrow types. 

For now we use 'None' for the timezone and get away with it mostly -- I think as DataFusion gets more sophisticated it will be necessary to tighten up our handling as well

tustvold (approved):

PR #6064, opened 2022-11-07
refactor: NS+table ID (instead of name) in querier<>ingester

### LINKED ISSUE #6049: Use table / namespace IDs when communicating between Querier and Ingester
The querier and ingester communicate using a [custom protocol] that references [tables] and [namespaces] by their name, rather than their catalog ID.

Currently the ingester has to keep the string identifiers in memory exclusively to satisfy this RPC interface, which forces us to perform many catalog queries, or keep the strings in the write message (which we would rather not do #4880).

## Proposal

Change this RPC format to use IDs instead of strings.

I do not know if this is possible, but a brief chat with @crepererum suggests it should be.

[namespaces]:https://github.com/influxdata/influxdb_iox/blob/f20a0648ec30f61208f8c5b6a4e8bd21092951c1/generated_types/protos/influxdata/iox/ingester/v1/query.proto#L33
[tables]:https://github.com/influxdata/influxdb_iox/blob/f20a0648ec30f61208f8c5b6a4e8bd21092951c1/generated_types/protos/influxdata/iox/ingester/v1/query.proto#L12
[custom protocol]:https://github.com/influxdata/influxdb_iox/blob/f20a0648ec30f61208f8c5b6a4e8bd21092951c1/docs/ingester_querier_protocol.md

crepererum:
This is the protocol and querier half of the change. Handing over to @domodwyer for the ingester bit.

Closes #6049.


domodwyer:
I have picked this commit into an upcoming PR.

Issue #5725, opened 2022-09-22
Ingester: remove async locks from query path


crepererum:
# What
Remove all long-held async write locks from the querier request path.

# Why
Currently we perform catalog IO and datafusion plan execution while holding write locks, blocking the querier request handler and adding unnecessary query latency. Also see #5722.

# How
Improve separation between execution and the actual state data structures.

PR #6068, opened 2022-11-07
perf(ingester): granular per-partition locking

domodwyer:
Prior to this PR there was a per-table mutex that serialised operations across all the partitions within it. 

This change pushes the mutex down into the partitions themselves, therefore increasing intra-table query parallelism in the presence of multiple partitions, and reduces the contention for the write path due to various persist-time operations on individual partitions.

Closes https://github.com/influxdata/influxdb_iox/issues/5725.

---

* refactor: push down per-partition op skipping (425fd46de)

      This moves the logic that skips operations that do not need to be
      applied to a partition during shard replay from the table level, to the
      partition level.

* perf(ingester): granular per-partition locking (79d24fa35)

      This commit pushes the existing table-level mutex down to the partition.
      
      This allows the ingester to gather data from multiple partitions within
      a single table in parallel, and reduces contention between ingest/query
      workloads.

### DIFF
--- a/ingester/src/data/namespace.rs
+++ b/ingester/src/data/namespace.rs
@@ -23,26 +23,26 @@ use crate::{data::DmlApplyAction, lifecycle::LifecycleHandle};
 #[derive(Debug, Default)]
 struct DoubleRef {
     // TODO(4880): this can be removed when IDs are sent over the wire.
-    by_name: HashMap<TableName, Arc<tokio::sync::RwLock<TableData>>>,
-    by_id: HashMap<TableId, Arc<tokio::sync::RwLock<TableData>>>,
+    by_name: HashMap<TableName, Arc<TableData>>,
+    by_id: HashMap<TableId, Arc<TableData>>,
 }
 
 impl DoubleRef {
-    fn insert(&mut self, t: TableData) -> Arc<tokio::sync::RwLock<TableData>> {
+    fn insert(&mut self, t: TableData) -> Arc<TableData> {
         let name = t.table_name().clone();
         let id = t.table_id();
 
-        let t = Arc::new(tokio::sync::RwLock::new(t));
+        let t = Arc::new(t);
         self.by_name.insert(name, Arc::clone(&t));
         self.by_id.insert(id, Arc::clone(&t));
         t
     }
 
-    fn by_name(&self, name: &TableName) -> Option<Arc<tokio::sync::RwLock<TableData>>> {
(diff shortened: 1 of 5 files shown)


ingester/src/data/table.rs
```diff
@@ -157,34 +158,33 @@ impl TableData {
...
-        if let Some(max) = partition_data.max_persisted_sequence_number() {
-            if max >= sequence_number {
-                trace!(
-                    shard_id=%self.shard_id,
-                    op_sequence_number=?sequence_number,
-                    "skipping already-persisted write"
-                );
-                return Ok(DmlApplyAction::Skipped);
-            }
-        }
```
domodwyer, line 175: This check moved into the `partition.buffer_write()` so that it can be correctly serialised in the partition lock.

It is also arguably cleaner - the table doesn't need to gate writes to individual partitions based on the partition's state. The table writes, and the partition is responsible for buffering writes (or not).

ingester/src/data/namespace.rs
```diff
@@ -23,26 +23,26 @@ use crate::{data::DmlApplyAction, lifecycle::LifecycleHandle};
 #[derive(Debug, Default)]
 struct DoubleRef {
     // TODO(4880): this can be removed when IDs are sent over the wire.
-    by_name: HashMap<TableName, Arc<tokio::sync::RwLock<TableData>>>,
-    by_id: HashMap<TableId, Arc<tokio::sync::RwLock<TableData>>>,
+    by_name: HashMap<TableName, Arc<TableData>>,
```
domodwyer, line 26: The `Mutex<_>` goes from here...

ingester/src/data/table.rs
```diff
@@ -1,48 +1,48 @@
...
 
 /// A double-referenced map where [`PartitionData`] can be looked up by
 /// [`PartitionKey`], or ID.
 #[derive(Debug, Default)]
 struct DoubleRef {
     // TODO(4880): this can be removed when IDs are sent over the wire.
-    by_key: HashMap<PartitionKey, PartitionData>,
-    by_id: HashMap<PartitionId, PartitionKey>,
+    by_key: ArcMap<PartitionKey, Mutex<PartitionData>>,
+    by_id: ArcMap<PartitionId, Mutex<PartitionData>>,
```
domodwyer, line 22: ...to here - that's basically the whole PR :+1:

crepererum (approved):

PR #6072, opened 2022-11-07
refactor: expose `ParquetExec` directly to DataFusion phys. plan

### LINKED ISSUE #5897: Expose parquet files directly to DataFusion

crepererum:
Closes #5897.


iox_query/src/provider/record_batch_exec.rs
```diff
@@ -0,0 +1,173 @@
...
+        expressions::PhysicalSortExpr,
+        memory::MemoryStream,
+        metrics::{BaselineMetrics, ExecutionPlanMetricsSet, MetricsSet},
+        DisplayFormatType, ExecutionPlan, Partitioning, SendableRecordBatchStream, Statistics,
+    },
+};
+use observability_deps::tracing::trace;
+use std::{collections::HashSet, fmt, sync::Arc};
+
+/// Implements the DataFusion physical plan interface for [`RecordBatch`]es with automatic projection and NULL-column creation.
```
alamb, line 19: I think you may be able to use https://docs.rs/datafusion/13.0.0/datafusion/physical_plan/memory/struct.MemoryExec.html directly here rather than creating a new `RecordBatchesExec`

Though the MemoryExec re-creates statistics each time. Maybe we could implement a `MemoryExec::with_statistics()` if that turns out to matter. 

No change needed in this PR -- but it might be a nice follow on PR

querier/src/namespace/query_access.rs
```diff
@@ -487,15 +488,15 @@ mod tests {
...
+                "|               |   TableScan: cpu projection=[foo, host, load, time]                                                                                                                                                                                                                                                                                                      |",
+                "| physical_plan | ProjectionExec: expr=[foo@0 as foo, host@1 as host, load@2 as load, time@3 as time]                                                                                                                                                                                                                                                                      |",
+                "|               |   ParquetExec: limit=None, partitions=[1/1/1/1/<uuid>.parquet, 1/1/1/1/<uuid>.parquet, 1/1/1/1/<uuid>.parquet, 1/1/2/2/<uuid>.parquet, 1/1/1/3/<uuid>.parquet], projection=[foo, host, load, time] |",
```
alamb, line 497: 🥳  Very nice -- it is also cool to see it showing multiple parquet files

query_tests/src/runner.rs
```diff
@@ -290,13 +294,28 @@ impl<W: Write> Runner<W> {
...
 
-            let current_results = pretty_format_batches(&results)
+            let mut current_results = pretty_format_batches(&results)
                 .unwrap()
                 .trim()
                 .lines()
                 .map(|s| s.to_string())
                 .collect::<Vec<_>>();
 
+            // normalize UUIDs, if requested
```
alamb, line 304: 👍  Love it -- maybe in the future we can keep track of uuids to turn them into `00000000-0000-0000-0000-000000000001`, `00000000-0000-0000-0000-000000000002`, etc -- but for now this is great

query_tests/src/runner.rs
```diff
@@ -198,6 +199,9 @@ impl<W: Write> Runner<W> {
             if q.sorted_compare() {
                 output.push("-- Results After Sorting".into())
             }
+            if q.sorted_uuids() {
```
alamb, line 202: Very clever

iox_query/src/provider/metrics.rs
```diff
@@ -12,18 +10,6 @@ pub fn parquet_metrics(plan: Arc<dyn ExecutionPlan>) -> Vec<MetricsSet> {
 }
 
 fn parquet_metrics_impl(plan: Arc<dyn ExecutionPlan>, output: &mut Vec<MetricsSet>) {
-    // Temporarily need to special case `IoxReadFilter` as it
```
alamb, line 15: 🎉

query_tests/cases/in/pushdown.expected
```diff
@@ -284,7 +297,7 @@
...
 |               |       RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                                                                                                                                                                                                                                             |
-|               |         IOxReadFilterNode: table_name=restaurant, chunks=0 predicate=Predicate exprs: [Float64(5) < system, town != Dictionary(Int32, Utf8("tewsbury")), system < Float64(7), time > TimestampNanosecond(130, None)]                                                                                                                                               |
+|               |         RecordBatchesExec: batches_groups=0 batches=0                                                                                                                                                                                                                                                                                                              |
```
alamb, line 300: Not pushing filters down here is an improvement I think 👍

iox_query/src/provider/physical.rs
```diff
@@ -1,280 +1,163 @@
...
-    fn as_any(&self) -> &dyn std::any::Any {
-        self
-    }
-
-    fn schema(&self) -> SchemaRef {
-        self.iox_schema.as_arrow()
+    context: Arc<TaskContext>,
+) -> Arc<dyn ExecutionPlan> {
+    if chunks.is_empty() {
+        return Arc::new(RecordBatchesExec::new(
```
alamb, line 53: There is also https://docs.rs/datafusion/13.0.0/datafusion/physical_plan/empty/struct.EmptyExec.html but other than existing upstream I don't think it is any better / worse than this (no changes needed)

iox_query/src/provider/physical.rs
```diff
@@ -1,280 +1,163 @@
...
-                    "IOxReadFilterNode: table_name={}, chunks={} predicate={}",
-                    self.table_name,
-                    self.chunks.len(),
-                    self.predicate,
-                )
-            }
-        }
+/// Distribute items from the given iterator into `n` containers.
+///
+/// This will produce less than `n` containers if the input has less than `n` elements.
```
alamb, line 130: 👍  this is neat -- it is effectively the same thing datafusion does https://github.com/apache/arrow-datafusion/blob/master/datafusion/core/src/datasource/listing/helpers.rs#L147-L161

Longer it might be valuable to try and normalize the groups based on number of rows or file size or something but for now this is probably good enough

iox_query/src/provider/physical.rs
```diff
@@ -1,280 +1,163 @@
...
+};
 
-/// Implements the DataFusion physical plan interface
-#[derive(Debug)]
-pub(crate) struct IOxReadFilterNode {
-    table_name: Arc<str>,
-    /// The desired output schema (includes selection)
-    /// note that the chunk may not have all these columns.
+/// How many concurrent Parquet read stream should be created?
+const CONCURRENT_PARQUET_STREAMS: usize = 10;
```
alamb, line 24: I suggest using `target_partitions` here rather than a fixed size of 10 as it is what the rest of DataFusion does and should scale up / down with the other operators. 


It is available from the `context` via `SessionConfig`: https://docs.rs/datafusion/13.0.0/datafusion/execution/context/struct.SessionConfig.html

Something like `context.session_config().target_partitions()` perhaps

alamb (approved):
giphy

This is amazing @crepererum  very nice 👏

alamb:
I took the liberty of merging up from master and fixing the tracing tests in7967771a5 . Can't wait for this PR 🤞 

cc @tustvold

iox_query/src/provider/physical.rs
crepererum: TBH I was expecting to find something like this in `itertools` but didn't. May contribute this to upstream one day.

iox_query/src/provider/record_batch_exec.rs
crepererum: Does `MemoryExec` support NULL-column creation and projection? (i.e. this is what the adapter stream that `RecordBatchesExec` uses does)

iox_query/src/provider/physical.rs
crepererum: done

query_tests/src/runner.rs
crepererum: done

iox_query/src/provider/record_batch_exec.rs
alamb: Probably not. Good point.

PR #6074, opened 2022-11-07
chore: Upgrade to Rust 1.65

carols10cents:


influxrpc_parser/src/predicate.rs
```diff
@@ -305,7 +305,7 @@ mod test {
         // remove quoting from literal - whilst we need the quoting to parse
         // the sql statement correctly, the RPCNode for the literal would not
         // have the quoting present.
-        let literal = parts[2].replace('\'', "").replace('"', "");
+        let literal = parts[2].replace(['\'', '"'], "");
```
crepererum, line 308: TIL :)

ingester/src/data/table.rs
```diff
@@ -429,7 +429,7 @@ mod tests {
                 namespace_id: ns_id,
                 table_id,
                 sequence_number: SequenceNumber::new(42),
-                bytes_written: 1131,
+                bytes_written: 1107,
```
crepererum, line 432: why did that change?

crepererum (approved):
One small question, otherwise looks good. Thank you!

domodwyer (approved):

influxrpc_parser/src/predicate.rs
carols10cents: Me too!

ingester/src/data/table.rs
carols10cents: I documented this in the commit message for the change:

> I tracked down the source of the size difference to the difference in
> `mem::size_of::<mutable_batch::column::ColumnData>`. I believe this enum
> is now able to take advantage of this niche-filling optimization:
> 
> <https://github.com/rust-lang/rust/pull/94075>

domodwyer: Nice - thanks for taking the time to look that up! I was intrigued when I saw it change with `+nightly` testing locally.

Issue #6025, opened 2022-11-01
Soft delete partitions that are outside the retention period


NGA-TRAN:
# Rationale

This is a task of #3053. If a partition is outside the retention period, we should mark it to_delete (update its `to_delete` to current time). We may need to only soft delete a partition all of its files are soft deleted (see #6024)

NGA-TRAN:
I will implement this similar to Luke's PR https://github.com/influxdata/influxdb_iox/pull/6070

PR #6075, opened 2022-11-07
feat: flag partitions for delete

NGA-TRAN:
First part of #6025 to flag partitions to delete
Next PR I will invoke this in GC


iox_catalog/src/interface.rs
```diff
@@ -1689,6 +1694,90 @@ pub(crate) mod test_helpers {
...
+        let shard = repos
+            .shards()
+            .create_or_get(&topic, ShardIndex::new(1))
+            .await
+            .unwrap();
+
+        let mut created = BTreeMap::new();
+
+        // Test flagged for deletion
+        // No retention period set, nothing should be flagged for deletion
```
lukebond, line 1726: suggest prefixing comment with `1. ` so that the `2. ` and `3. ` further down makes sense

iox_catalog/src/interface.rs
```diff
@@ -1689,6 +1694,90 @@ pub(crate) mod test_helpers {
...
+            .flag_for_delete_by_retention()
+            .await
+            .unwrap();
+        assert!(ids.len() > 1); // fail PG
+                                // two_days_ago partition should be flagged for deletion
+        assert!(ids.contains(&partition_two_days_ago.id));
+        // today partition should not be flagged for deletion
+        assert!(!ids.contains(&partition_today.id)); // fail mem
+                                                     //
+                                                     // 3. flag for select again and shold not get anything return becasue the partitions are already flagged for deletion
```
lukebond, line 1772: fix alignment here. think you need a newline in between to get around the auto formatter

iox_catalog/src/interface.rs
```diff
@@ -1689,6 +1694,90 @@ pub(crate) mod test_helpers {
...
+            .await
+            .expect("failed to create partition");
+        created.insert(partition_two_days_ago.id, partition_two_days_ago.clone());
+        // Should have at least one partition deleted
+        let ids = repos
+            .partitions()
+            .flag_for_delete_by_retention()
+            .await
+            .unwrap();
+        assert!(ids.len() > 1); // fail PG
```
lukebond, line 1766: this `// fail PG` is a note for you perhaps? if it's to remain in the PR when not a draft then i don't know what it means. same with `'// fail mem` below

iox_catalog/src/mem.rs
```diff
@@ -931,6 +933,66 @@ impl PartitionRepo for MemTxn {
...
+    async fn flag_for_delete_by_retention(&mut self) -> Result<Vec<PartitionId>> {
+        let now = Timestamp::from(self.time_provider.now());
+        let stage = self.stage();
+
+        Ok(stage
+            .partitions
+            .iter_mut()
+            .filter_map(|p| {
+                // already flagged for delete
+                if p.to_delete.is_some() {
```
lukebond, line 946: wouldn't this return `None` from the `flag_for_delete_by_retention` function? i don't see the compiler complaining about this but it surprised me. wonder if it would be nicer to do something like:

```rust
        Ok(stage
            .partitions
            .iter_mut()
            // don't flag if already flagged for deletion
            .filter(|f| f.to_delete.is_none())
            .filter_map(|p| {
                stage
                    .tables
                    .iter()
                    .find(|t| t.id == p.table_id)
                    .and_then(|t| {
                      ...
```

iox_catalog/src/mem.rs
```diff
@@ -931,6 +933,66 @@ impl PartitionRepo for MemTxn {
...
+                                // convert p.partition_key to nano seconds
+                                // From string to a NaiveDate
+                                println!(" ======= partition key is {}", p.partition_key);
+                                let naive_date = NaiveDate::parse_from_str(
+                                    &p.partition_key.to_string(),
+                                    "%Y-%m-%d",
+                                );
+
+                                // naive_date is not a error due to non-date tsting partition key such as foo
+                                if naive_date.is_ok() {
```
lukebond, line 969: i think the error handling of `parse_from_str` would be nicer with something like this:

```rust
NaiveDate::parse_from_str(
  &p.partition_key.to_string(),
  "%Y-%m-%d",
)
.ok() // convert result to option
.and_then(|naive_date| { // return None if None, else call closure with existing logic & no need to unwrap
  ...
})
```

lukebond (changes requested):

NGA-TRAN:
@lukebond The pR is ready for review

iox_catalog/src/mem.rs
```diff
@@ -931,6 +933,61 @@ impl PartitionRepo for MemTxn {
...
+                    .tables
+                    .iter()
+                    .find(|t| t.id == p.table_id)
+                    .and_then(|t| {
+                        stage
+                            .namespaces
+                            .iter()
+                            .find(|n| n.id == t.namespace_id)
+                            .and_then(|ns| {
+                                if let Some(retention_ns) = ns.retention_period_ns {
```
lukebond, line 957: these `if let Some(...) { } else { None }` blocks could be cleaner like so:

```rust
ns.retention_period_ns.and_then(|retention_ms| {
})
```

which will return `None` if it's `None` and call the close if `Some`. there are three places in this block where you could do that.

not enough to make me decline the review of course!

iox_catalog/src/postgres.rs
```diff
@@ -1377,6 +1377,30 @@ WHERE id = $2;
         .await
         .map_err(|e| Error::SqlxError { source: e })
     }
+
+    async fn flag_for_delete_by_retention(&mut self) -> Result<Vec<PartitionId>> {
+        let flagged_at = Timestamp::from(self.time_provider.now());
+        let flagged = sqlx::query(
+            r#"
+            update partition p
```
lukebond, line 1385: just cosmetic of course but it'd be nice if the SQL keywords were uppercased like all the others in the file

lukebond (approved):
a couple of cosmetic and style things but otherwise LGTM!

lukebond:
@NGA-TRAN i'm going to make the changes i suggested above and merge, then rebase #6090 because it's difficult to review when it includes these changes as well

NGA-TRAN (commented):
Thanks for the suggestions and commit @lukebond

PR #6084, opened 2022-11-08
test: use `EXPLAIN ANALYZE` for SQL metric tests

crepererum:
Needs a bit more infra (due to normalization), but this seems to be worth it so we can easily hook up more metrics in the future.


query_tests/cases/in/duplicates_parquet.sql
```diff
@@ -16,3 +16,7 @@ EXPLAIN select state as name from h2o UNION ALL select city as name from h2o;
 
 -- count(*) plan that ensures that row count statistics are not used (because we don't know how many rows overlap)
 select count(*) from h2o;
+
+-- test pushdown metrics
+-- IOX_COMPARE: uuid, metrics
```
alamb, line 21: ❤️

query_tests/cases/in/duplicates_parquet.expected
```diff
@@ -76,3 +76,23 @@
...
+|    |                 UnionExec, metrics=[elapsed_compute=1.234ms, mem_used=0, output_rows=7, spill_count=0, spilled_bytes=0]    |
+|    |                   ParquetExec: limit=None, partitions=[1/1/1/1/00000000-0000-0000-0000-000000000000.parquet], predicate=state_min@0 <= MA AND MA <= state_max@1, projection=[area, city, max_temp, min_temp, state, time], metrics=[bytes_scanned{filename=1/1/1/1/00000000-0000-0000-0000-000000000000.parquet}=474, elapsed_compute=1.234ms, mem_used=0, num_predicate_creation_errors=0, output_rows=4, predicate_evaluation_errors{filename=1/1/1/1/00000000-0000-0000-0000-000000000000.parquet}=0, pushdown_eval_time{filename=1/1/1/1/00000000-0000-0000-0000-000000000000.parquet}=1.234ms, pushdown_rows_filtered{filename=1/1/1/1/00000000-0000-0000-0000-000000000000.parquet}=0, row_groups_pruned{filename=1/1/1/1/00000000-0000-0000-0000-000000000000.parquet}=0, spill_count=0, spilled_bytes=0, time_elapsed_opening=1.234ms, time_elapsed_processing=1.234ms, time_elapsed_scanning=1.234ms]    |
```
alamb, line 94: This will be great when I fix https://github.com/apache/arrow-datafusion/issues/4144 this should be so much better ❤️

query_tests/src/runner.rs
```diff
@@ -327,6 +330,31 @@ impl<W: Write> Runner<W> {
...
             }
 
+            // normalize metrics, if requested
+            if query.normalized_metrics() {
+                let regex_metrics = Regex::new(r#"metrics=\[([^\]]*)\]"#).expect("metrics regex");
+                let regex_timing = Regex::new(r#"[0-9]+(\.[0-9]+)?.s"#).expect("timing regex");
+                let regex_linesep = Regex::new(r#"[+-]{6,}"#).expect("linesep regex");
+                let regex_col = Regex::new(r#"\s+\|"#).expect("col regex");
+
+                current_results = current_results
```
alamb, line 340: Sometimes I find including example before/afters for regex manipulation helps me remember / understand what it is trying to do when I come on the regex again

alamb (approved):
This is great -- thank you @crepererum

query_tests/src/runner.rs
crepererum: added some comments

query_tests/cases/in/duplicates_parquet.expected
alamb: > Do you think it would be sensible for DF to sort by key

query_tests/src/runner.rs
```diff
@@ -340,10 +342,37 @@ impl<W: Write> Runner<W> {
                 current_results = current_results
                     .into_iter()
                     .map(|s| {
+                        // Replace timings with fixed value, e.g.:
```
alamb, line 345: ❤️  this is so nice. My future self thanks you

PR #6086, opened 2022-11-08
fix: cross-thread ArcMap::get()

domodwyer:
This PR fixes an issue with the `ArcMap` that caused an `insert()` in one thread to not be visible in another thread because different thread-local hasher instances were being used.

Also included is a small refactor to DRY the key checking, and avoid cloning a key when calling `insert()` that I noticed while working on this.

---

* test: cross-thread hashmap entry visibility (e9bda70c7)

      At the time of this commit, this test fails. Performing a get() on a key
      previously inserted by another thread should not fail.

* fix: cross-thread map entry visibility (1fb9cfba8)

      This commit changes the ArcMap HashBuilder to use the same instance as
      the underlying HashMap hasher.
      
      This prevents divergent hashing across threads that MAY initialise a
      hasher with a different seed.

* refactor: extract key equality checking (e50a854e6)

      Creates a shared fn for checking key equality to DRY the various
      chaining checks.

* refactor: accept owned key for insert() (08f82786a)

      Changes the bounds on the ArcMap to accept an owned key, avoiding an
      extra allocation.
      
      Cleans up the bounds on other fn to ensure the borrowed key impl Eq and
      is the ref type of K.

* refactor: inline helpers (b3c2ae72b)

      Inline the hash generation & key comparator.

### DIFF
--- a/ingester/src/arcmap.rs
+++ b/ingester/src/arcmap.rs
@@ -172,12 +172,14 @@ where
         self.map.read().values().map(Arc::clone).collect()
     }
 
+    #[inline]
     fn compute_hash<Q: Hash + ?Sized>(&self, key: &Q) -> u64 {
         let mut state = self.hasher.build_hasher();
         key.hash(&mut state);
         state.finish()
     }
 
+    #[inline]
     fn key_equal<Q>(q: &Q) -> impl FnMut(&'_ K) -> bool + '_
     where
         K: Borrow<Q>,
--- a/ingester/src/arcmap.rs
+++ b/ingester/src/arcmap.rs
@@ -69,7 +69,8 @@ where
     /// memorised.
     pub(crate) fn get_or_insert_with<Q, F>(&self, key: &Q, init: F) -> Arc<V>
     where
-        Q: Hash + PartialEq<K> + ToOwned<Owned = K> + ?Sized,
+        K: Borrow<Q>,
+        Q: Hash + Eq + ToOwned<Owned = K> + ?Sized,
         F: FnOnce() -> Arc<V>,
     {
         // Memorise the hash outside of the lock.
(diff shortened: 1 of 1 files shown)


ingester/src/arcmap.rs
```diff
@@ -316,6 +332,31 @@ mod tests {
...
+
+        for h in handles {
+            h.join().unwrap();
+        }
+
+        for i in 0..N_THREADS {
+            let v = refs.get(&i).unwrap();
+            assert_eq!(i, *v);
+        }
+    }
```
domodwyer, line 358: This test **could** fail (rarely for me!) prior to 1fb9cfba8c3bc6628e683b6147af88819c0f146d.

ingester/src/arcmap.rs
```diff
@@ -36,9 +36,11 @@ impl<K, V, S> std::ops::Deref for ArcMap<K, V, S> {
 
 impl<K, V> Default for ArcMap<K, V> {
     fn default() -> Self {
+        let map: HashMap<K, Arc<V>> = Default::default();
+        let hasher = map.hasher().clone();
         Self {
-            map: Default::default(),
-            hasher: Default::default(),
+            map: RwLock::new(map),
+            hasher,
```
domodwyer, line 43: This is the fix - simple but painful!

The same hasher should be used by everything that hashes for a consistent result.

mkmik (approved):

ingester/src/arcmap.rs
alamb, line 43: I suggest a comment as it is so subtle to try and help future readers. Something like

```suggestion
            // The same hasher should be used by everything that hashes for a consistent result.
            // otherwise some threads may not see the same contents
            // see https://github.com/influxdata/influxdb_iox/pull/6086 for details
            hasher,
```

Not sure if there is a better ticket reference

ingester/src/arcmap.rs
alamb: I think fuzz testing is likely to be the best possible given the situation

alamb (approved):

PR #6090, opened 2022-11-08
feat: call soft delete partitions from garbage collector

NGA-TRAN:
Second part of #6025. This depends on #6075 


garbage_collector/src/retention/flagger.rs
```diff
@@ -11,16 +11,25 @@ pub(crate) async fn perform(
...
 
-        if flagged.is_empty() {
+        let partition_flagged = catalog
+            .repositories()
+            .await
+            .partitions()
+            .flag_for_delete_by_retention()
+            .await
+            .context(PartitionFlaggingSnafu)?;
+        info!(partition_flagged_count = %partition_flagged.len(), "iox_catalog::partition::flag_for_delete_by_retention()");
```
NGA-TRAN, line 30: @lukebond : I am debating whether to create a new module named `partition_retention` for the work or simply to invoke the partitions' flag for delete right here like I do in this PR. While I want to do it in a separate module and make a new environment to just call it a few times a day, I feel calling it directly here may make more sense because they both rely on retention period.
Let me know what you think.

lukebond: i think it's good like this :+1:  lucky i called it `retention` and not something parquet file specific, a happy accident.

lukebond (approved):

PR #6093, opened 2022-11-09
feat: use IDs in ingester query API

domodwyer:
This PR builds off of #6064, changing the `querier -> ingester` communication wire format to utilise IDs for namespace & table exclusively, instead of the previous use of string names.

This alleviates the ingester of the need to discover the namespace & table names in the (ingest or query) hot path in order to support queries. After this PR, the ingesters can switch to a deferred name lookup to smear the queries over a duration of time.

**Breaking change:** the `query-ingester` CLI command now takes IDs, not names (as that is what the API now uses).

---

* refactor: NS+table ID (instead of name) in querier<>ingester (6e63024ca)

* feat(ingester): use IDs for query API (c162b1e25)

      Changes the ingester to utilise the ID fields (instead of names) sent
      over the query wire message wrapped within the Flight API.
      
      BREAKING: this changes the "query-ingester" CLI command arguments which
      now expects the namespace & table IDs, rather than their names.

* refactor(ingester): add more query logging context (a76094922)

      Updates the log messages during query execution to include more context
      fields.


test_helpers_end_to_end/src/mini_cluster.rs
```diff
@@ -236,6 +242,60 @@ impl MiniCluster {
...
+            .into_inner()
+            .schema
+            .unwrap()
+            .tables
+            .get(name)
+            .expect("table not found")
+            .id;
+
+        TableId::new(id)
+    }
```
domodwyer, line 297: End-to-end tests now make use of the Schema API in the router to map names to IDs where necessary - this gives extra coverage to the schema API as a nice side effect :tada:

test_helpers_end_to_end/src/mini_cluster.rs
```diff
@@ -236,6 +242,60 @@ impl MiniCluster {
         self.namespace.as_ref()
     }
 
+    /// Get a reference to the mini cluster's namespace ID.
+    pub async fn namespace_id(&self) -> NamespaceId {
+        *self
+            .namespace_id
+            .get_or_init(|| async {
```
domodwyer, line 249: For the test namespace ID, the lookup is only performed once per test.

ingester/src/querier_handler.rs
```diff
@@ -486,19 +482,24 @@ mod tests {
...
-            let scenario = Arc::new(make_ingester_data(two_partitions).await.0);
-            scenarios.push(scenario);
+            let (scenario, ns, table) = make_ingester_data(two_partitions).await;
+
+            let old = *table_id.get_or_insert(table);
+            assert_eq!(old, table);
+            let old = *ns_id.get_or_insert(ns);
+            assert_eq!(old, ns);
+
+            scenarios.push(Arc::new(scenario));
```
domodwyer, line 496: This just ensures a stable table ID is returned for all invocations of `make_ingester_data()` (a bit of a test monster!) within this test.

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -706,16 +713,23 @@ async fn query_ingester() {
...
-                        .arg("my_awesome_table2")
+                        .arg(state.cluster().namespace_id().await.get().to_string())
+                        .arg(
+                            state
+                                .cluster()
+                                .table_id("my_awesome_table2")
+                                .await
+                                .get()
+                                .to_string(),
+                        )
```
domodwyer, line 732: This shows the breaking CLI change: the `query-ingester` command now takes the respective IDs instead of names.

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -706,16 +713,23 @@ async fn query_ingester() {
                     // something like "wrong query protocol" or
                     // "invalid message" as the querier requires a
                     // different message format Ticket in the flight protocol
-                    let expected = "Unknown namespace: my_awesome_table2";
+                    let expected = "Unknown namespace: ";
```
domodwyer, line 716: The comment indicates this error string was wrong / not handled correctly prior to these changes. 

The code that emits the message doesn't have access to the ID and the existing output was wrong before, so I didn't sink too much time into it 🤷

generated_types/protos/influxdata/iox/ingester/v1/query.proto
```diff
@@ -29,8 +33,12 @@ message IngesterQueryRequest {
   reserved "greater_than_sequence_number";
   reserved 6;
 
+  // Was namespace indexed as a string (instead of via ID)
+  reserved "namespace";
+  reserved 7;
+
   // Namespace to search
-  string namespace = 7;
+  int64 namespace_id = 10;
```
domodwyer, line 41: This is the gist of the PR.

domodwyer:
I am marking this PR to squashed as the commits do not contain atomic changes.

alamb:
> Breaking change: the query-ingester CLI command now takes IDs, not names (as that is what the API now uses).

It would be nicer UX if the name can be used on the CLI. I think this would be possible by having the CLI command use the CatalogService to find the correct id from a name.

However, since the `query-ingester` CLI is basically an internal debugging tool at the moment,  I think the loss of UX in this case is fine for the time being

ingester/src/handler.rs
```diff
@@ -358,11 +358,13 @@ impl IngestHandler for IngestHandlerImpl {
         // TEMP(alamb): Log details about what was requested
```
alamb, line 358: lol -- gotta love the TEMP comments that are still in the code years later

alamb (approved):
Looks good to me. Thanks @domodwyer

Issue #4445, opened 2022-04-27
Implement streaming SeriesSet conversions (Querier performance / memory improvement for influxrpc)


alamb:
__Use case:__
Most of the IOx execution engine is "streaming" meaning it can process batches of rows and doesn't need to have all the data buffered contiguously in memory. This means less memory is required for query processing

One area where we are not yet streaming fully is series set conversion (the format that IOx uses to talk with Flux and InfluxQL at the time of this writing). 

You can see more context on https://github.com/influxdata/influxdb_iox/pull/4444 and https://github.com/influxdata/influxdb_iox/issues/4401 where the input stream of batches is concatenated into a single large `RecordBatch`

__Proposal:__
Modify the SeriesSet conversion logic to handle streams of `RecordBatch`es

__Current behaviour:__
RecordBatches are concatenated together

__Desired behaviour:__
Don't buffer any more than necessary

Issue #503, opened 2020-11-30
Support "streaming" results during query execution


alamb:
__Use case:__
At the time of writing, when responding to gRPC queries, IOx will effectively produce a single `ReadResponse` per measurement. This means that all the series / data for that measurement is buffered in memory and sent in a big chunk back to the client

This has several negative properties:
1. Increased resource usage: consumes more memory than necessary, both to buffer the intermediate results of the plan as well as the data coverson / network mesage.  
2. Increased Latency / loss of streaming: It delays any data from being sent until the entire result is converted

__Proposal:__
The high level proposal is to "stream" results back to the client as they are created. The underlying DataFusion executor supports streaming but we buffer the results (twice, actually) before sending them back, for implementation convenience. 

__Current behaviour:__
The buffering happens at two levels today:
1. The results of the DataFusion plan (`RecordBatch`es are collected -- calls to `collect` in exec.rs -- for example,  https://github.com/influxdata/influxdb_iox/blob/main/query/src/exec.rs#L363)

2. When converting RecordBatches --> gRPC data frames, data for all series (aka all columns) is collected into a `Vec` before being converted

Namely, 
```
pub fn series_set_item_to_read_response(series_set_item: SeriesSetItem) -> Result<ReadResponse> {
```
in https://github.com/influxdata/influxdb_iox/blob/main/src/server/rpc/data.rs#L88

__Desired behaviour:__
As data flows back from the DataFusion plan, it should be converted to SeriesSets, and as soon as the next group is seen, the series set should be produced. 

Also, once that SeriesSet is produced, then it should immediately be converted to the gRPC data frames and returned to the client before the next chunk of data (`RecordBatch`) from DataFusion is requested. 

Note this ticket might be best implemented in several chinks

__Alternatives considered:__
Describe other solutions or features you considered.

ming535:
It seems that this issue was already solved?
`Executor::to_series_set` spawns one task for each logical plan. 
For each of the above task, it execute the plan and streams `SendableRecordBatchStream` to `SeriesSetItem`.
`Executor::to_series_set` receives each task's `SeriesSetItem` and sends it to `rps::service`.
`rpc::service::convertt_seriess_set` streams `SeriesSetItem` to `ReadResponse`

alamb:
Hi @ming535  -- 

The specific buffering I am thinking about in this task is here:

[...]

PR #6094, opened 2022-11-09
refactor: chunk InfluxRPC `ReadResponses`

crepererum:
Currently we see some prod panics:

```
'assertion failed: len <= std::u32::MAX as usize', tonic/src/codec/encode.rs:127:5
```

This is due to an upstream bug in tonic:
https://github.com/hyperium/tonic/issues/1141

However the fix will only turn this into an error instead of panicking. We should instead NOT return such overlarge results, esp. because InfluxRPC supports streaming.

While we currently don't perform streaming conversion (like streaming the data out of the query stack into the gRPC layer), the 4GB size limit can easily be triggered (in prod) w/ enough RAM. So let's re-chunk our in-memory responses so that they stream nicely to the client.

We may later implement proper streaming conversion, see #4445 and #503.


service_grpc_influxrpc/src/service.rs
```diff
@@ -1586,6 +1595,58 @@ where
...
+
+    let mut frames = vec![];
+    let mut size = 0;
+    for frame in it {
+        let fsize = frame_size(&frame);
+
+        // flush?
+        if size + fsize > size_limit {
+            let mut tmp = vec![];
+            std::mem::swap(&mut tmp, &mut frames);
```
alamb, line 1619: I think you can write this pattern using `take`, which may be slightly easier to read

```suggestion
            let frames =  std::mem::take(&mut frames);
```

alamb (approved):
Looks good to me -- writing some tests might be good

service_grpc_influxrpc/src/service.rs
crepererum: done

crepererum:
> writing some tests might be good

Added unit tests for the chunking routine.

alamb (commented):
Thanks for the tests!

PR #6095, opened 2022-11-09
refactor: generic deferred loader helper

domodwyer:
**This PR makes no functional change the system.**

Instead, it splits apart the `DeferredSortKey` so the "deferred load" bit of it can be reused.

---

* refactor: generic deferred loader helper (3e1103a8e)

      Splits the DeferredSortKey loader introduced in #5807 into two parts - a
      generic helper type that implements deferred/background loading of
      values, and SortKey specific logic for use with it.
      
      As this will be more widley used, this implementation features improved
      behaviour of the deferred loader under concurrent demand requests
      (multiple calls to get() do not attempt to concurrently resolve the
      value), as well as complete cancellation safety (cancelling the get()
      doesn't affect the liveness of the background task).


ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// The value is being actively resolved by the background task.
+    ///
+    /// Callers can subscribe to a completion event by waiting on the
+    /// [`Notify`].
+    Loading(Arc<Notify>),
+    /// The value was fetched by the background task and is read to be consumed.
+    ///
+    /// Only the background task ever sets this state.
+    Resolved(T),
+}
```
domodwyer, line 33: It's a state machine!

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+                // Sleep for the random duration, or until a demand call is
+                // made.
+                tokio::select! {
+                    _ = tokio::time::sleep(wait_for) => {
+                        trace!("timeout woke loader task");
+                    }
+                    _ = rx => {
+                        trace!("demand call woke loader task");
+                    }
+                }
```
domodwyer, line 109: When initialising a `DeferredLoad` a background task is spawned and it waits until the random timer fires, or it is asked to resolve the value.

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+            let waker = match state.take().unwrap() {
+                // This caller is the first to demand the value - wake the
+                // background task, initialise the notification mechanism and
+                // wait for the task to complete.
+                State::Unresolved(task_waker) => {
+                    // Wake the running background task, ignoring any send error
+                    // as the background task may have concurrently woken up due
+                    // to the sleep timer and stopped listening on the waker
+                    // channel.
+                    let _ = task_waker.send(());
```
domodwyer, line 187: A user can wake up the background task and get it to resolve the value (by executing the future) by calling `get()`. 

If it is the first call to `get()` and the value hasn't yet been loaded in the background, the caller triggers the background task to execute the future immediately, and then waits for the result.

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+            // This check happens before the state lock is released, ensuring
+            // the background task doesn't concurrently finish (it would be
+            // blocked waiting to update the state).
+            assert!(!self.handle.is_finished());
+
+            waker
+        };
+
+        // Wait for the background task to complete resolving the value.
+        waker.notified().await;
```
domodwyer, line 219: All other concurrent callers to `get()` also wait for the result in the same way as the triggering caller.

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+                            // caller demanded the value.
+                            *guard = Some(State::Resolved(v));
+                            None
+                        }
+                        State::Loading(callers) => {
+                            // At least one caller is demanding the value, and
+                            // must be woken after the lock is released.
+                            *guard = Some(State::Resolved(v));
+                            Some(callers)
+                        }
```
domodwyer, line 130: Once the future execution is complete, the background task stores the result.

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+                        State::Resolved(_) => unreachable!(),
+                    }
+                };
+
+                // Wake the waiters, if any, outside of the lock to avoid
+                // unnecessary contention. If there are >1 threads waiting for
+                // the value, they make contend for the value lock however.
+                if let Some(callers) = callers {
+                    callers.notify_waiters();
+                }
```
domodwyer, line 140: And wakes up any `get()` callers that are still waiting for the value.

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+    /// DeferredLoad delay to this value, encouraging the background load to
+    /// happen a long time in the future.
+    ///
+    /// Because the delay before the background task runs is selected from a
+    /// uniform distribution, it is certainly possible the background task runs
+    /// before the demand call, invalidating the test - it's just statistically
+    /// very unlikely. Over multiple test runs, the odds of the background task
+    /// running before the demand call is so low, that if you see it happen
+    /// multiple times, I'd suggest taking out life insurance and staying
+    /// indoors during lightening storms.
```
carols10cents, line 255: 1. 🤣 
2. Typo, "lightening" should be "lightning"
3. This is a bit dark, isn't it? How about:

```suggestion
    /// multiple times, I'd suggest entering a lottery and/or checking changes for bugs.
```

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+    const LONG_LONG_TIME: Duration = Duration::from_secs(100_000_000);
+
+    #[tokio::test]
+    async fn test_demand() {
+        let d = DeferredLoad::new(LONG_LONG_TIME, async { 42 });
+
+        assert_eq!(d.get().await, 42);
```
carols10cents, line 262: If something about `get` is wrong and doesn't resolve the future immediately, instead waiting for the background task, isn't it likely that this test will hang rather than fail? That is, shouldn't this test be using `with_timeout_panic` like some of the other tests are with a value like 5s?

carols10cents (changes requested):
I have some questions and thoughts of various importance :)

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+/// background task exists once it has successfully resolved `T`.
+///
+/// # Stale Cached Values
+///
+/// This is effectively a cache that is pre-fetched in the background - this
+/// necessitates that the caller can tolerate, or detect, stale values.
+pub(crate) struct DeferredLoad<T> {
+    /// The inner state of the [`DeferredLoad`].
+    ///
+    /// The option facilitates taking ownership of the state to transition, and
```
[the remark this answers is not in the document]
domodwyer: I am totally fine with comments like this! I try to doclink to types as much as possible because a) it's nice, b) good docs are nice, and c) it helps keep our docs in-sync when we change the names of types/remove things. Admittedly (c) is unlikely to happen for `Option`, but ya never know 😆

ingester/src/deferred_load.rs
```diff
@@ -0,0 +1,369 @@
...
+where
+    T: Send + Sync + 'static,
+{
+    /// Construct a [`DeferredLoad`] instance that fetches `T` for the specified
+    /// after at most `deadline` duration of time, by executing `F`.
+    ///
+    /// The background task will wait a uniformly random duration of time
+    /// between `[0, deadline)` before attempting to pre-fetch `T` by executing
+    /// the provided future.
+    pub(crate) fn new<F>(deadline: Duration, loader: F) -> Self
```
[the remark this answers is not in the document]
domodwyer: Yeah and it's not actually really a deadline if we're putting a pedantic hat on - saying "it must complete in 5 seconds" is not something we can uphold; if the task is scheduled and starts the sleep 10 seconds later, then 💩

ingester/src/deferred_load.rs
domodwyer: Yes, truthfully I remembered I should be adding timeouts about half way through the tests and didn't bother going back over what I had already wrote 🦡

carols10cents (approved):
Thanks, this LGTM now!

ingester/src/data/partition/resolver/sort_key.rs
```diff
@@ -1,181 +1,60 @@
...
 }
 
-impl DeferredSortKey {
-    /// Construct a [`DeferredSortKey`] instance that fetches the [`SortKey`]
-    /// for the specified `partition_id`.
-    ///
-    /// The background task will wait a uniformly random duration of time
-    /// between `[0, max_smear)` before attempting to pre-fetch the [`SortKey`]
-    /// from `catalog`.
+impl SortKeyResolver {
```
alamb, line 18: FWIW this looks nontrivially similar to the `cache_system` crate from @crepererum  -- https://github.com/influxdata/influxdb_iox/tree/main/cache_system that is used in the querier

As the ingester grows more sophisticated caching it might be worth investing time to see if we can unify the caching mechanisms in the two (and get access to the common metrics, etssting, etc)

crepererum: I agree.

Issue #6026, opened 2022-11-01
Hard delete partitions


NGA-TRAN:
# Rationale

This is a task of #3053 .
A soft deleted partition should be hard deleted after all of its files are hard deleted from both catalog and objection store.

lukebond:
ah sorry i misread that it's partitions not parquet files :facepalm:

Issue #3053, opened 2021-11-05
Implement retention policies


jacobmarble:
Automatically delete data after a per-database configurable time.

lukebond:
i think we are missing a sub-task to add `retention_period_ns` to the `table` table as well? in order to satisfy this part of the requirements (from above):

> A table retention period, if it exists, supersedes its database retention period

NGA-TRAN:
@lukebond 

> i think we are missing a sub-task to add retention_period_ns to the table table as well? in order to satisfy this part of the requirements (from above):

>> A table retention period, if it exists, supersedes its database retention period

For the first version, we stay with per-bucket retention. When it works well, we will add table-level retention

PR #6099, opened 2022-11-09
feat: catalog delete old partitions

lukebond:
Part of #6026 

Adds a function to the catalog that will hard delete partitions that have been flagged for deletion. Will add this to the GC once merged.


iox_catalog/src/interface.rs
```diff
@@ -1720,7 +1744,24 @@ pub(crate) mod test_helpers {
             .await
             .unwrap();
 
-        let mut created = BTreeMap::new();
```
lukebond, line 1723: this wasn't being used and seems a copy-pasta artefact from another test that did assert against it

iox_catalog/src/interface.rs
```diff
@@ -1720,7 +1744,24 @@ pub(crate) mod test_helpers {
...
+        // Create some partitions
```
lukebond, line 1747: i moved this block up from a few lines below because otherwise test step marked `1.` is meaningless as there is nothing to delete

iox_catalog/src/interface.rs
```diff
@@ -1777,6 +1799,69 @@ pub(crate) mod test_helpers {
...
+            .partitions()
+            .flag_for_delete_by_retention()
+            .await
+            .unwrap();
+        assert_eq!(ids.len(), 1);
+        assert!(ids.contains(&partition_two_days_ago.id));
+        // 4. run the hard delete and verify that nothing was deleted
+        let ids = repos.partitions().delete_old_ids_only(now).await.unwrap();
+        assert!(ids.is_empty());
+        // 5. delete the parquet file because keeping it around makes later tests fail. this isn't
```
lukebond, line 1847: we should really have a separate schema or whatever it's called for each test so we don't have to worry about this

iox_catalog/src/mem.rs
```diff
@@ -984,6 +984,27 @@ impl PartitionRepo for MemTxn {
             })
             .collect())
     }
+
+    async fn delete_old_ids_only(&mut self, older_than: Timestamp) -> Result<Vec<PartitionId>> {
+        let stage = self.stage();
+
+        println!("parquet files: {:?}", stage.parquet_files);
```
lukebond, line 991: whoops, removing this debug...

crepererum:
The querier currently never fades out / deletes partitions. I wonder if this would be an issue (it only stores the partition sort key and only uses it when it has an ingester partition or a parquet file referencing it).

lukebond:
> The querier currently never fades out / deletes partitions. I wonder if this would be an issue (it only stores the partition sort key and only uses it when it has an ingester partition or a parquet file referencing it).

is that captured here? https://github.com/influxdata/influxdb_iox/issues/3053 there is mention of querier but doesn't sound like what you said. sounds like we might need to add another requirement/task to the design?

crepererum:
Yeah, I didn't think about the partition aspect of the catalog TBH. The main question is: can partitions for the same partition key be re-created somehow? If not, I think we're fine (due to "only stores the partition sort key and only uses it when it has an ingester partition or a parquet file referencing it").

iox_catalog/src/postgres.rs
```diff
@@ -1401,6 +1403,39 @@ WHERE id = $2;
...
+            r#"
+WITH partition_ids as (
+    SELECT id
+    FROM partition
+    WHERE to_delete < $1
+    LIMIT $2
+)
+DELETE FROM partition
+WHERE id IN (SELECT id FROM partition_ids)
+RETURNING id;
```
NGA-TRAN, line 1427: Using WITH is fine but if you want to eliminate an SELECT, I wonder if be low statement is clearer:

```SQL
DELETE FROM partition
WHERE id IN (
    SELECT p.id
    FROM partition p
    WHERE p.to_delete is not null and
          p.to_delete < $1
    LIMIT $2
)
RETURNING id;
```

iox_catalog/src/interface.rs
```diff
@@ -1777,6 +1799,69 @@ pub(crate) mod test_helpers {
             .unwrap();
         assert!(ids.is_empty());
 
+        // verify that hard delete will only delete flagged partitions
+        let now = Timestamp::new((catalog.time_provider().now()).timestamp_nanos());
+        let ids = repos.partitions().delete_old_ids_only(now).await.unwrap();
+        assert_eq!(ids.len(), 1);
+        assert!(ids.contains(&partition_two_days_ago.id));
+
```
NGA-TRAN, line 1807: I think to make sure we do not delete other partitions by accident (or bug), right here, we may want to verify that those partitions are still in the catalog. I am thinking to invoke `list_by_namespace` to get all partitions of this namespace and verify that `partition_two_days_ago` are no longer there but others are still

NGA-TRAN (approved):
Looks good, @lukebond . I have a suggestion to verify the results of the test further and a minor  SELECT suggestion

NGA-TRAN:
@crepererum 

> can partitions for the same partition key be re-created somehow?

Router will reject loading data outside retention period. It will be unlikely the same partition will be re-created. But if the retention period is modified to make it longer, it can happen.

crepererum:
>  But if the retention period is modified to make it longer, it can happen.

So let's play this scenario through:

- router will re-create partition but with different partition ID
- new partition may have different sort key compared to old one (which is OK within the querier, we address caches by partition ID not by partition key)
- the old and new partitions are considered distinct, so we will NOT de-dup between them

The last point really worries me TBH. To avoid that I would propose that either we do never modify the retention policy (unlikely to work, because it's a product requirement IIRC) or we add a proper generation ID/number to the retention policy so a querier has at least some chance to detect that mess.

PR #6101, opened 2022-11-09
chore: delete dead code `PartitionAddr`

alamb:
I ran across some code that is no longer used


carols10cents (approved):
I'm only mad I didn't find this first. 👍🏻

Issue #6029, opened 2022-11-01
Remove field retention_duration from namespace


NGA-TRAN:
# Rationale 

Currently the namespace table has field `retention_duration` that was designed to store the retention period. However, its data type is `character varying` that is not good for us to store nanosecond. This field is initiated and used in the code so we do not want to modify its type. Let us stick to our design and create a new field `retention_period_nano` as a bigint. After `retention_period_nano`  works well, we will use this ticket to remove `retention_duration`

NGA-TRAN:
I am giving this a shot before continuing using `retention_period_ns`. I hope I will have more insight and make the code cleaner before playing with the new `retention_period_ns`

PR #6103, opened 2022-11-09
feat: gRPC that creates namespace

### LINKED ISSUE #6096: Support gRPC to create a namespace
# Rationale

This is a task of #3053. We will use config topic and query_pool for this as done here https://github.com/influxdata/influxdb_iox/blob/30c353e9c0af203d13cb3385c31c357f7efecd99/ioxd_router/src/lib.rs#L218-L262

NGA-TRAN:
Closes #6096 


clap_blocks/src/write_buffer.rs
```diff
@@ -60,6 +55,32 @@ pub struct WriteBufferConfig {
...
+    /// Query pool name to dispatch writes to.
+    #[clap(
+        long = "query-pool",
+        env = "INFLUXDB_IOX_QUERY_POOL_NAME",
+        default_value = "iox-shared",
+        action
+    )]
+    pub query_pool_name: String,
+}
+
```
NGA-TRAN, line 83: @domodwyer I have refactored this to reuse in both router and create namespace as you suggested. Their CLI commands are different and I do not want to use router's config for namespace

iox_catalog/src/interface.rs
```diff
@@ -1061,13 +1116,15 @@ pub(crate) mod test_helpers {
         assert_eq!(NEW_COLUMN_LIMIT, modified.max_columns_per_table);
 
         const NEW_RETENTION_PERIOD: i64 = 5;
+        const NEW_RETENTION_PERIOD_NS: i64 = 5 * 60 * 60 * 1000 * 1000 * 1000;
```
domodwyer, line 1119: Perhaps we ensure this stays in sync:

```suggestion
        const NEW_RETENTION_PERIOD_NS: i64 = NEW_RETENTION_PERIOD * 60 * 60 * 1000 * 1000 * 1000;
```

iox_catalog/src/interface.rs
```diff
@@ -295,6 +304,16 @@ pub trait NamespaceRepo: Send + Sync {
...
+    /// Create namespace with retention_period
+    async fn create_with_retention(
+        &mut self,
+        name: &str,
+        retention_duration: &str,
+        retention_period_hours: Option<i64>,
+        topic_id: TopicId,
+        query_pool_id: QueryPoolId,
+    ) -> Result<Namespace>;
+
```
domodwyer, line 316: What is the difference between `retention_duration` and `retention_period_hours`? Can we not reuse the existing `create()` method to keep the catalog interface small & concise?

domodwyer (changes requested):
I think I understand some of the confusion I've had in slack! I left some comments but we can chat on zoom later today if you want :+1:

Thanks for building this out 🙏

iox_catalog/src/interface.rs
lukebond: `retention_duration` is old and unused and will be removed in https://github.com/influxdata/influxdb_iox/issues/6029 - i think Nga is holding off removing it until the replacement field is working, which is a small epic here: https://github.com/influxdata/influxdb_iox/issues/3053

removing it as part of this PR will touch a lot of places (i tried in an earlier PR and recoiled!).

domodwyer: I see - perhaps we can keep using the existing method / fields in this API in this PR, and change everything in one go? It seems like it will be confusing if the user says "retention: 4 days", and the API creates a database entry with "infinite" and "4 hours", and all the existing code ignores the "4 hours" bit - ideally we'd do an atomic change over to add the new field and use it, but looks like it's already in the database schema.

A clean changeover might be better where add the namespace API in one (this) PR, and then open another PR for #6029 after that changes all the code to use the new retention in one go rather than leaving it inconsistent.

generated_types/protos/influxdata/iox/schema/v1/service.proto
```diff
@@ -64,6 +67,24 @@ message ColumnSchema {
...
+  string name = 1;
+
+  // Number of hours of the retention period
+  optional int64 retention_hours = 2;
+
+  // topic name
+  string topic = 3;
+
+  // query pool name
+  string query_pool = 4;
```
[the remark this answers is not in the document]
NGA-TRAN: Silly me to interpret it that if we use the same parameter in the client request as done above https://github.com/influxdata/influxdb_iox/pull/6103/files#r1018465409, when we remove the parameter, we will remove them everywhere.

Using the one populated on the server side makes more sense. Let me do it then.

generated_types/protos/influxdata/iox/schema/v1/service.proto
```diff
@@ -6,6 +6,9 @@ service SchemaService {
   // Get the schema for a namespace
   rpc GetSchema(GetSchemaRequest) returns (GetSchemaResponse);
 
+  // Create a namespace
+  rpc CreateNamespace(CreateNamespaceRequest) returns (CreateNamespaceResponse);
+
   // update retention period
   rpc UpdateNamespaceRetention(UpdateNamespaceRetentionRequest) returns (UpdateNamespaceRetentionResponse);
```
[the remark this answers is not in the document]
NGA-TRAN: @alamb suggested me to use Schema API for namespace (update and create). Moving them somewhere else is not that tricky. Where  do you suggest them @alamb and @domodwyer ?

iox_catalog/src/interface.rs
NGA-TRAN: @domodwyer : let me try to see if I can remove `retention_duration` first and then come back to this PR. I agree it will be cleaner. I hoped to understand  `retention_duration` better as we proceed with `retention_period_hour` and then remove `retention_duration` but I think by doing the opposite way (removing `retention_duration` first), I might get a lot better insight to implement `retention_period_hour`

NGA-TRAN:
I have converted to draft. I will come back after #6029

iox_catalog/src/interface.rs
domodwyer: That makes sense too! I think today everything everywhere uses the hard-coded const (or hard-codes "inf" themselves):

https://github.com/influxdata/influxdb_iox/blob/93e11d4c9172fb3ae87d66d9f8a84f2a496df587/iox_catalog/src/lib.rs#L38-L39

So removing it should be fairly painless - nothing cares about it / makes use of the value today as far as I know.

generated_types/protos/influxdata/iox/schema/v1/service.proto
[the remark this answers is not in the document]
NGA-TRAN: @domodwyer: Let us do this in a separate PR after we have a decision of this discussion https://influxdata.slack.com/archives/CRK9M8L5Q/p1668193721306879

generated_types/protos/influxdata/iox/schema/v1/service.proto
```diff
@@ -64,6 +67,20 @@ message ColumnSchema {
     }
 }
 
+message CreateNamespaceRequest {
+  reserved 3, 4;
```
NGA-TRAN, line 71: Let me know if you think I should remove these reserved? I remove them in this PR but this PR is not yet merged

querier/src/handler.rs
```diff
@@ -106,7 +106,7 @@ impl QuerierHandlerImpl {
 #[async_trait]
 impl QuerierHandler for QuerierHandlerImpl {
     fn schema_service(&self) -> SchemaServiceServer<SchemaService> {
-        SchemaServiceServer::new(SchemaService::new(Arc::clone(&self.catalog)))
+        SchemaServiceServer::new(SchemaService::new(Arc::clone(&self.catalog), None, None))
```
NGA-TRAN, line 109: We will return `unimplemented` if the create namespace is invoked from Querier. Provide no topic and query pool here will do the job

router/src/server/grpc.rs
```diff
@@ -20,6 +21,8 @@ use crate::shard::Shard;
 /// This type is responsible for managing all gRPC services exposed by `router`.
 #[derive(Debug)]
 pub struct GrpcDelegate<S> {
+    topic_id: TopicId,
+    query_pool_id: QueryPoolId,
```
NGA-TRAN, line 25: Store topic id and query pool id in this struct when the router is created

router/src/server/grpc.rs
```diff
@@ -48,9 +55,11 @@ where
...
     /// [`SchemaService`]: generated_types::influxdata::iox::schema::v1::schema_service_server::SchemaService.
     pub fn schema_service(&self) -> schema_service_server::SchemaServiceServer<SchemaService> {
-        schema_service_server::SchemaServiceServer::new(SchemaService::new(Arc::clone(
-            &self.catalog,
-        )))
+        schema_service_server::SchemaServiceServer::new(SchemaService::new(
+            Arc::clone(&self.catalog),
+            Some(self.topic_id),
+            Some(self.query_pool_id),
+        ))
```
NGA-TRAN, line 62: And then they are provided to the Schema service

NGA-TRAN:
@domodwyer and @lukebond : The PR is ready for rereviewing

generated_types/protos/influxdata/iox/schema/v1/service.proto
mkmik: Did you add these fields in a previous commit of this PR and later removed them?

If yes, then we don't have to declare them as reserved here

NGA-TRAN: Yeah, the field #3 and #4 were added in a previous commit but not merged yet. I will remove this line

generated_types/protos/influxdata/iox/schema/v1/service.proto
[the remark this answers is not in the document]
alamb: I believe the consensus at the meeting was to put this API into the existing namespace service https://github.com/influxdata/influxdb_iox/blob/main/generated_types/protos/influxdata/iox/namespace/v1/service.proto and also run that gRPC service in the router server/process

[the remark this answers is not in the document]
NGA-TRAN: Yeah @lukebond is working on this

lukebond:
i have rebased this PR on main after merging #6151, which moved the namespace API to the router and put the update retention call in there. this PR now adds the create API call to the namespace API after said refactor. i also kept retention in nanoseconds except where it interfaces the user: the CLI.

NGA-TRAN (commented):
Thanks so much @lukebond for splitting namespace service and merge it here

iox_catalog/src/postgres.rs
```diff
@@ -731,26 +741,15 @@ RETURNING *;
...
-            )
-            .bind(rentenion_period_ns) // $1
-            .bind(name) // $2
-            .fetch_one(&mut self.inner)
-            .await
-        };
+        let rec = sqlx::query_as::<_, Namespace>(
+            r#"UPDATE namespace SET retention_period_ns = $1 WHERE name = $2 RETURNING *;"#,
+        )
+        .bind(retention_period_ns) // $1
```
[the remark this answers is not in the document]
domodwyer: that's right - sqlx maps `Option` to a nullable field 🎉

influxdb_iox/src/commands/namespace/create.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Write data into the specified database
+#[derive(Debug, clap::Parser)]
+pub struct Config {
+    /// The namespace to to be created
+    #[clap(action)]
+    namespace: String,
+
+    /// Num of hours of the retention period of this namespace.
+    /// If not specified, the default retention period will be used.
```
domodwyer, line 22:
```suggestion
    /// If not specified, an infinite retention period will be used.
```

iox_catalog/src/interface.rs
```diff
@@ -287,18 +287,28 @@ pub trait QueryPoolRepo: Send + Sync {
...
     ) -> Result<Namespace>;
 
+    /// Create namespace with retention_period
+    async fn create_with_retention(
+        &mut self,
+        name: &str,
+        retention_period_ns: Option<i64>,
+        topic_id: TopicId,
+        query_pool_id: QueryPoolId,
+    ) -> Result<Namespace>;
```
domodwyer, line 305: We should merge this with `create()` as they're functionally identical if you pass `None` as the retention period.

iox_catalog/src/interface.rs
```diff
@@ -717,23 +727,6 @@ where
     get_schema_internal(namespace, repos).await
 }
 
-/// Update retention for a namespace
-pub async fn update_namespace_retention<R>(
```
domodwyer, line 721: Thank you for removing this :+1: I am keen on keeping the API surface area as small as possible in the catalog because otherwise it becomes (more of!) a monster!

service_grpc_namespace/src/lib.rs
```diff
@@ -38,6 +48,35 @@ impl namespace_service_server::NamespaceService for NamespaceService {
         }))
     }
 
+    // create a namespace
+    async fn create_namespace(
+        &self,
+        request: Request<CreateNamespaceRequest>,
+    ) -> Result<Response<CreateNamespaceResponse>, Status> {
+        if self.topic_id.is_none() || self.query_id.is_none() {
+            return Err(Status::unimplemented("topic_id or query_id not set"));
```
domodwyer, line 57: This perhaps should be a "you gave us an invalid argument" response:

```suggestion
            return Err(Status::invalid_argument("topic_id or query_id not set"));
```

That signifies that the user-provided input is bad/insufficient/broken, whereas the `unimplemented` would mean "this action is not available", suggesting a code/server issue.

domodwyer (approved):
Approving because code looks great - thank you for going the extra mile on this 🙏

Minor comments, but good to merge after :+1:

iox_catalog/src/interface.rs
lukebond: have done this now

lukebond:
merging namespace `create` and `create_with_retention` touched a lot of files (mostly adding `, None` to test code). also found a constant only used in tests but not limited to `#[cfg(test)]` which proved too difficult to limit to `#[cfg(test])` because of how the atalog interface/mem/postgres setup works w.r.t. tests. i did tidy up its usage a little and rename the `create_namespace` helper that was silently setting retention to 1hr in all tests that used it.

lukebond:
i modified the behaviour introduced in #6148 that gave the mem catalog impl a default retention period of 1hr. touched a lot of places but avoids a trap that i walked right into and resulted in all the extra commits above until i realised what was going on

PR #6104, opened 2022-11-10
feat: hard partition delete in gc

### LINKED ISSUE #6026: Hard delete partitions

lukebond:
DO NOT MERGE until the team has discussed the impact of deleting and recreating partitions on the querier.

Closes #6026

Follows on from #6099 and calls the partition-deleting functionality from the garbage collector.


lukebond:
deleting partitions can have a strange side-effect in the querier if we allow retention periods to be changed. for now we won't delete them until we have adjusted the design to consider this.

Issue #5875, opened 2022-10-17
Queries with time ranges are problematic


mkmik:
__Steps to reproduce:__

Queries involving `now()` and time arithmetics blow up unless explicitly casted to `timestamp`


1. Load enough data in a table so that it's above the `INFLUXDB_IOX_MAX_TABLE_QUERY_BYTES` limit.
2. run this query: `SELECT count(*) FROM vibrations WHERE time BETWEEN (now() - interval '1 hour') AND now()`
 

(In our internal test environment this is `./target/debug/influxdb_iox query -h http://localhost:8083 dab3b3b2e54dc4a4_ab038f883697864a "SELECT count(*) FROM vibrations WHERE time BETWEEN (now() - interval '1 hour') AND now()" )

__Expected behaviour:__

Some result like:

```
+-----------------+
| COUNT(UInt8(1)) |
+-----------------+
| 2340000         |
+-----------------+
```

__Actual behaviour:__

```
Error querying: status: ResourceExhausted, message: "Error while planning query: Resources exhausted: Query would scan at least 10247432978 bytes, more than configured maximum 5368709120 bytes. Try adjusting your compactor settings or increasing the per query memory limit.", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 17 Oct 2022 14:24:22 GMT", "content-length": "0"} }
```

Changing the query slightly changes the behaviour:

```console
$ ./target/debug/influxdb_iox query -h http://localhost:8083 dab3b3b2e54dc4a4_ab038f883697864a "SELECT count(*) FROM vibrations WHERE time BETWEEN (now() - interval '1 hour')::timestamp AND now()";
+-----------------+
| COUNT(UInt8(1)) |
+-----------------+
| 2110000         |
+-----------------+
```

__Environment info:__

reproduced in our eu-central-1 prod-like test environment.

__More details:__

Due to https://github.com/influxdata/influxdb_iox/issues/5776 I can't view the plan by running `EXPLAIN SELECT`, but I can run similar queries on a smaller dataset:

```console
$ curl -i -X POST "http://localhost:8080/api/v2/write?bucket=test&org=internal&precision=s" --data-binary @./test_fixtures/lineproto/air_and_water.lp
$ ./target/debug/influxdb_iox query internal_test "select time,state from air_temperature where time > (now() - interval '3 years') and time < now()::timestamp"
+----------------------+-------+
| time                 | state |
+----------------------+-------+
[...]

Issue #6098, opened 2022-11-09
More efficient IOx scans with overlapping files/ingester data (rework physical plan construction)


crepererum:
# What
Update 2023-01-27: the way we create physical nodes leads to plans that refer to many parquet files to use too much memory than we have, leading to OOMs and other bad behavior. 

Currently we create physical nodes out of `QueryChunk`s in an imperative style to hopefully produce some optimal (or at least: not horrible) physical plan. Instead I would rather construct an naive plan and use optimizer passes to improve them.

# Why
The current code is:
- deeply-if-nested imperative code and hard to understand
- debug (no intermediate steps visible)
- it's unclear how well it plays with the remaining DF optimization framework
- hard to extend (see #6028)

# How

## Naive Plan
Produce a plan like this:

```text
         +------------+
         | FilterExec |
         +------------+
                ^
                |
          +----------+
          | SortExec |
          +----------+
                ^
                |
       +-----------------+
       | DeduplicateExec |
       +-----------------+
                ^
                |
   +-------------------------+
   | SortPreservingMergeExec |
   +-------------------------+
                ^
                |
          +-----------+
          | UnionExec |
          +-----------+
                ^
                |
      +---------+-------+---------------...
      |                 |
 +----------+      +----------+
 | SortExec |      | SortExec |         ...
 +----------+      +----------+
      ^                  ^
      |                  |
+-------------+  +-------------------+
| ParquetExec |  | RecordBatchesExec |  ...
+-------------+  +-------------------+
```

It is important that the plan is technically correct and easy to understand. All easy optimization like predicate pushdown or sort elimination should be modeled as optimizer passes!

## Optimizer Passes
TODO

alamb:
I think this is a great idea in the long run. I am not sure how critical it is in the short term

Other than possibly easier to maintain code, we might also be able to parallelize the plan better (aka repartition the data into disjoin partitions and do deduplication in parallel, for example)

 @mingmwang is introducing a pass that has similarities in DataFusion  https://github.com/apache/arrow-datafusion/pull/4122

Brainstorm: in order to pull off this design we would perhaps need optimizer passs that knew
1.  how/when to push deduplicate through sort preserving merge (probably can't without partitioning)
2. How to eliminate redundant sorts 
3. How to add repatitioning without destrying the sort order

crepererum:
[...]

PR #6108, opened 2022-11-10
chore: Update datafusion again

alamb:
# Rationale
* Keep up with changes in DataFusion.
* Test if https://github.com/apache/arrow-datafusion/pull/4122 causes us any problems
* Get fix for https://github.com/influxdata/influxdb_iox/issues/5875 (will test after this is merged)
* Get fix for https://github.com/apache/arrow-datafusion/issues/3968

# Changes:
1. Update DataFusion pin to https://github.com/apache/arrow-datafusion/commit/d2814c960168b45c4a0f5d7bbb72d9f412cb08bd
2. Update APIs for differences

# Required parts
- [x] Update  https://github.com/apache/arrow-datafusion/issues/4169
- [x] Write tests for ParquetChunkList
- [x] Fixed what adding an unecessary sort (very likely https://github.com/apache/arrow-datafusion/pull/4122)
Screenshot 2022-11-10 at 10 23 22 AM


query_tests/cases/in/duplicates_ingester.expected
```diff
@@ -1,91 +1,95 @@
...
+|               |     ProjectionExec: expr=[time@5 as time, state@4 as state, city@1 as city, min_temp@3 as min_temp, max_temp@2 as max_temp, area@0 as area]                            |
+|               |       RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                                                 |
+|               |         UnionExec                                                                                                                                                      |
+|               |           DeduplicateExec: [state@4 ASC,city@1 ASC,time@5 ASC]                                                                                                         |
+|               |             SortPreservingMergeExec: [state@4 ASC,city@1 ASC,time@5 ASC]                                                                                               |
+|               |               SortExec: [state@4 ASC,city@1 ASC,time@5 ASC]                                                                                                            |
```
alamb, line 17: This diff doesn't make it easy to see, but the difference is that a new SortExec is added here

The reason this happens is that the new `BasicEnforcement` rule added in https://github.com/apache/arrow-datafusion/pull/4122 by @mingmwang is (correctly) deciding that since the `ParquetExec` declares its output is not sorted, then the output of the `UnionExec` is also therefore not sorted, and thus needs to be resorted prior to the sort preserving merge.

 I think the way to avoid this is to teach DataFusion that the ParquetExec is actually sorted (which is is) and then everything will work out.


As a bonus, I think at least some part of the logic in the plan construction of IOx (that adds SortExec's in to sort the data) could potentially be removed as it is now covered by the DataFusion optimizer. I think this is very much in line with  @crepererum 's ideas from https://github.com/influxdata/influxdb_iox/issues/6098

crepererum (approved):

query_tests/cases/in/duplicates_ingester.expected
alamb: Filed upstream issue: https://github.com/apache/arrow-datafusion/issues/4169

iox_query/src/provider/physical.rs
```diff
@@ -86,31 +137,42 @@ pub fn chunks_to_physical_nodes(
...
                 partition_values: vec![],
                 range: None,
                 extensions: None,
             }),
             target_partitions,
         );
+
+        // Tell datafusion about the sort key, if any
+        let file_schema = iox_schema.as_arrow();
+        let output_ordering =
```
alamb, line 159: Creating this output ordering is important to tell  DataFusion the data is already sorted so that it doesn't put another sort in our plan

querier/src/namespace/query_access.rs
```diff
@@ -508,24 +508,24 @@ mod tests {
...
+                "|               |     ProjectionExec: expr=[host@0 as host, perc@1 as perc, time@2 as time]                                                                                                       |",
+                "|               |       UnionExec                                                                                                                                                                 |",
+                "|               |         CoalesceBatchesExec: target_batch_size=4096                                                                                                                             |",
+                "|               |           FilterExec: time@2 < 1 OR time@2 > 13 OR NOT host@0 = CAST(d AS Dictionary(Int32, Utf8))                                                                              |",
+                "|               |             ParquetExec: limit=None, partitions=[1/2/1/4/<uuid>.parquet], output_ordering=[host@0 ASC, time@2 ASC], projection=[host, perc, time] |",
```
alamb, line 523: the difference in these plans is that the output ordering is now specified

alamb:
@crepererum  FYI this PR has changed substantially since you last reviewed it

iox_query/src/provider/physical.rs
```diff
@@ -15,12 +18,58 @@ use datafusion::{
...
+    if let (Some(existing_sort_key), Some(chunk_sort_key)) = (existing_sort_key, chunk_sort_key) {
+        let combined_sort_key = SortKey::try_merge_key(&existing_sort_key, chunk_sort_key);
+
+        match combined_sort_key {
+            Some(combined_sort_key) if combined_sort_key == &existing_sort_key => {
+                Some(existing_sort_key)
+            }
+            Some(combined_sort_key) => Some(combined_sort_key.clone()),
+            None => None,
+        }
```
NGA-TRAN, line 66: So the whole purpose of this match is to not clone the key when possible?

iox_query/src/provider/physical.rs
```diff
@@ -63,12 +112,14 @@ pub fn chunks_to_physical_nodes(
                 let url_str = parquet_input.object_store_url.as_str().to_owned();
                 match parquet_chunks.entry(url_str) {
                     Entry::Occupied(mut o) => {
-                        o.get_mut().1.push(parquet_input.object_meta);
+                        o.get_mut()
+                            .add_parquet_file(chunk.as_ref(), parquet_input.object_meta);
```
NGA-TRAN, line 116: I asked myself why the `combine_sort_key` returns `None` when one of the input is none. But when I come to this line I see it makes sense 👍

iox_query/src/provider/physical.rs
```diff
@@ -156,4 +220,48 @@ mod tests {
...
+
+        assert_eq!(
+            combine_sort_key(Some(skey_t1_t2.clone()), Some(&skey_t1)),
+            Some(skey_t1_t2)
+        );
+
+        assert_eq!(
+            combine_sort_key(Some(skey_t2_t1.clone()), Some(&skey_t1)),
+            Some(skey_t2_t1)
+        );
```
NGA-TRAN, line 265: You may want to add a test to combine `skey_t1_t2` with `skey_t2_t1` and I think it will return None, right?

parquet_file/src/storage.rs
```diff
@@ -123,6 +123,8 @@ impl ParquetExecInput {
             table_partition_cols: vec![],
             // TODO avoid this `copied_config` when config_options are directly available on context
             config_options: session_ctx.copied_config().config_options(),
+            // This is not correct for production: they are ordered
+            output_ordering: None,
```
NGA-TRAN, line 127: I am not sure I understand the comment

query_tests/cases/in/duplicates_parquet.expected
```diff
@@ -91,8 +91,8 @@
...
...l_time{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=1.234ms, page_index_eval_time{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=1.234ms, page_index_rows_filtered{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=0, page_index_rows_filtered{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=0, predicate_evaluation_errors{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=0, predicate_evaluation_errors{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=0, pushdown_eval_time{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=1.234ms, pushdown_eval_time{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=1.234ms, pushdown_rows_filtered{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=2, pushdown_rows_filtered{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=3, row_groups_pruned{filename=1/1/1/1/00000000-0000-0000-0000-000000000002.parquet}=0, row_groups_pruned{filename=1/1/1/1/00000000-0000-0000-0000-000000000003.parquet}=0, spill_count=0, spilled_bytes=0, time_elapsed_opening=1.234ms, time_elapsed_processing=1.234ms, time_elapsed_scanning=1.234ms]    |
```
NGA-TRAN, line 96: Even though there are no highlight here, the differences are the output-ordering 👍

NGA-TRAN (approved):
So nice that we can see the output ordering in the plan now and that we upgrade DF but keep our query plan optimal. Thanks @alamb
Those query plans are worth keeping

iox_query/src/exec/non_null_checker.rs
```diff
@@ -218,8 +218,8 @@ impl ExecutionPlan for NonNullCheckerExec {
         None
     }
 
-    fn required_child_distribution(&self) -> Distribution {
-        Distribution::UnspecifiedDistribution
+    fn required_input_distribution(&self) -> Vec<Distribution> {
```
crepererum, line 221: Is the distribution "per input node" now instead of "for all inputs"?

parquet_file/src/storage.rs
```diff
@@ -123,6 +123,8 @@ impl ParquetExecInput {
             table_partition_cols: vec![],
             // TODO avoid this `copied_config` when config_options are directly available on context
             config_options: session_ctx.copied_config().config_options(),
+            // This is not correct for production: they are ordered
```
crepererum, line 126:
```suggestion
            // Parquet files ARE actually sorted but we don't care here since we just construct a `collect` plan.
```

querier/src/ingester/flight_client.rs
```diff
@@ -224,17 +224,19 @@ mod tests {
         // see https://github.com/influxdata/influxdb_iox/issues/5974
```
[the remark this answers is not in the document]
crepererum: No, we still need this test because there might be a different expression construct hitting this limit. I can construct one for you (using some ugly nested case statements for example).

iox_query/src/provider/physical.rs
```diff
@@ -15,12 +18,58 @@ use datafusion::{
...
+
+/// Combines the existing sort key with the sort key of the chunk, returning the new sort key.
+fn combine_sort_key(
+    existing_sort_key: Option<SortKey>,
+    chunk_sort_key: Option<&SortKey>,
+) -> Option<SortKey> {
+    if let (Some(existing_sort_key), Some(chunk_sort_key)) = (existing_sort_key, chunk_sort_key) {
+        let combined_sort_key = SortKey::try_merge_key(&existing_sort_key, chunk_sort_key);
+
+        match combined_sort_key {
```
crepererum, line 60: `combined_sort_key.map(|k| if ... {...} else {...})` is the more Rusty way, but that's optional

iox_query/src/provider/physical.rs
```diff
@@ -15,12 +18,58 @@ use datafusion::{
...
+            Some(combined_sort_key) if combined_sort_key == &existing_sort_key => {
+                Some(existing_sort_key)
+            }
+            Some(combined_sort_key) => Some(combined_sort_key.clone()),
+            None => None,
+        }
+    } else {
+        // no existing sort key means the data wasn't consistently sorted so leave it alone
```
crepererum, line 68: This will happen in prod. The issue is that we have chunks for different partitions here (I think) and partitions have different sort orders. You may chose to NOT address this now, since this requires changes to the chunk grouping, but then we at least need to track this in a somewhat-urgent (= next week) follow-up ticket.

crepererum (changes requested):
Needs some more work regarding the overflow test (see comment) and the parquet chunk grouping due to inter-partition sort key inconsistencies.

iox_query/src/exec/non_null_checker.rs
alamb: Yes, exactly, it is per-child, to handle cases like joins where the children may / may not be partitioned correctly to perform a parallel equijoin without redistributing the inputs

iox_query/src/provider/physical.rs
alamb: yes, that is correct -- I expect the sort key will commonly be the same so avoiding the cloning will be substantial. I will make this clear in comments.

iox_query/src/provider/physical.rs
alamb: It does sound somewhat concerning, though I don't fully follow what the concern is -- if we are merging chunks from different partitions with different sort orders together that will result in wrong results.

iox_query/src/provider/physical.rs
alamb: I tried doing this and the borrow checker was not happy:

```
error[E0505]: cannot move out of `existing_sort_key` because it is borrowed
  --> iox_query/src/provider/physical.rs:65:31
   |
63 |         let combined_sort_key = SortKey::try_merge_key(&existing_sort_key, chunk_sort_key);
   |                                                        ------------------ borrow of `existing_sort_key` occurs here
64 |
65 |         combined_sort_key.map(|combined_sort_key| {
   |                           --- ^^^^^^^^^^^^^^^^^^^ move out of `existing_sort_key` occurs here
   |                           |
   |                           borrow later used by call
...
69 |                 existing_sort_key
   |                 ----------------- move occurs due to use in closure
```

iox_query/src/provider/physical.rs
alamb: I added some comments to `combine_sort_key` to try and make this clearer

iox_query/src/provider/physical.rs
alamb: Agreed. Added in e5c1365cd

parquet_file/src/storage.rs
alamb: I think @crepererum 's rewording is much better -- does that make more sense?

```rust
 // Parquet files ARE actually sorted but we don't care here since we just construct a `collect` plan.
```

?

querier/src/ingester/flight_client.rs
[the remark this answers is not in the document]
alamb: Ah, I see -- I will see if I can construct such a thing

querier/src/ingester/flight_client.rs
```diff
@@ -231,28 +252,54 @@ mod tests {
...
-                    predicate: Some(predicate),
-                };
+                // build a deeply recursive nested expression:
+                //
+                // CASE
+                //  WHEN TRUE
+                //  THEN (WHEN ...)
+                // ELSE FALSE
+                //
+                let expr = (0..n).fold(lit(false), |expr, _|{
```
alamb, line 263: I have made another deeply nested expression as suggested by @crepererum  in addition to a second test that verifies deeply nested `AND`s now work

querier/src/ingester/flight_client.rs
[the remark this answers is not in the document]
alamb: done

iox_query/src/provider/physical.rs
crepererum: Ah, because this is only relevant for de-dup, right? Well, then let's leave it. If it is a performance issue in some way (because we insert merges that shouldn't be there), we can address this later.

crepererum (approved):

iox_query/src/provider/physical.rs
alamb: yeah -- I hope once we have completed the predicate pushdown work (we are very close) and "run on any data size" then we can move on to profiling and improving performance for large numbers of parquet files

PR #6113, opened 2022-11-10
refactor: first step to remove unused retention_duration

NGA-TRAN:
Connects to #6029 

This is the first PR to remove code that inserts `retention_duration` into the `namespace` catalog table.
My next PR will remove the field `retention_duration` from `namespace` catalog table and all of its corresponding variable in the code, too


iox_catalog/src/mem.rs
```diff
@@ -302,7 +301,7 @@ impl NamespaceRepo for MemTxn {
             name: name.to_string(),
             topic_id,
             query_pool_id,
-            retention_duration: Some(retention_duration.to_string()),
+            retention_duration: Some("inf".to_string()), // temporary until the field retention_durantion is removed in the catalgo table in next PR
```
NGA-TRAN, line 304: Since I remove the function parameter, I want to use hardcode "inf" here. In my next PR, after I remove the `retention_duration` from the namespace catalog table, I will remove this

alamb (approved):
Makes sense to me

domodwyer (approved):
Beautiful 🎉

lukebond (approved):

PR #6124, opened 2022-11-11
refactor: remove retention_duration field from namespace catalog table

### LINKED ISSUE #6029: Remove field retention_duration from namespace

NGA-TRAN:
Closes #6029 

Second and last PR for #6029


iox_catalog/migrations/20221111133939_drop_retention_duration.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE namespace DROP COLUMN IF EXISTS retention_duration;
```
alamb, line 1: I ran this on a local catalog

```sql
postgres=# ALTER TABLE namespace DROP COLUMN IF EXISTS retention_duration;
ALTER TABLE
postgres=# select * from namespace limit 1;
 id |               name                | topic_id | query_pool_id | max_tables | max_columns_per_table | retention_period_ns 
----+-----------------------------------+----------+---------------+------------+-----------------------+---------------------
  1 | 1443098387584974_7862845850241514 |        1 |             1 |      10000 |                   200 |                    
(1 row)

postgres=# 
```

And it seems to have worked well!

alamb (approved):

Issue #6125, opened 2022-11-11
Potential incorrect results when scanning many parquet files


alamb:
I thought of this problem while working on teaching DataFusion about how the parquet files are sorted in IOx: https://github.com/influxdata/influxdb_iox/pull/6108. I don't think that PR makes this problem any better / worse so I have filed a separate ticket

# Background

The [iox provider](https://github.com/influxdata/influxdb_iox/blob/main/iox_query/src/provider.rs#L511) makes physical plans like this:

```text
                                     ┌───────────────────────┐
                                     │SortPreservingMergeExec│   <-- This is added if sort_output = true
                                     └───────────────────────┘
                                              ▲
                                              │
                                     ┌─────────────────┐
                                     │    UnionExec    │
                                     │                 │
                                     └─────────────────┘
                                              ▲
                                              │
                       ┌──────────────────────┴───────────┬─────────────────────┐
                       │                                  │                     │
                       │                                  │                     │
              ┌─────────────────┐                ┌─────────────────┐   ┌─────────────────┐
              │ DeduplicateExec │                │ DeduplicateExec │   │     SortExec    │  <-- This is added if output_sort_key.is_some()
              └─────────────────┘                └─────────────────┘   │    (Optional)   │
                       ▲                                  ▲            └─────────────────┘
                       │                                  │                     ▲
           ┌───────────────────────┐                      │                     │
           │SortPreservingMergeExec│                      │             ┌─────────────────┐
           └───────────────────────┘                      │             │   ParquetExec   │
                       ▲                                  │             │    (Chunk 4)    │
                       │                                  │             └─────────────────┘
           ┌───────────────────────┐                      │
           │       UnionExec       │                      │
           └───────────────────────┘                      │
                      ▲                                   |
                      │                                   |
          ┌───────────┴───────────┐                       │
          │                       │                       │
 ┌─────────────────┐     ┌─────────────────┐    ┌─────────────────┐
 │    SortExec     │     │    SortExec     │    │    SortExec     │
 │   (optional)    │     │   (optional)    │    │   (optional)    │
 └─────────────────┘     └─────────────────┘    └─────────────────┘
          ▲                       ▲                      ▲
          │                       │                      │
          │                       │                      │
 ┌─────────────────┐     ┌─────────────────┐    ┌─────────────────┐
 │   ParquetExec   │     │   ParquetExec   │    │   ParquetExec   │
[...]

PR #6126, opened 2022-11-11
fix: workaround potential incorrect results in sort

alamb:
Workaround a potential bug by ensuring only one parquet file is in each partition. Related to https://github.com/influxdata/influxdb_iox/issues/6125


alamb:
🤔  it is not good there is no test failing as it means there is not enough test coverage for this feature

PR #6127, opened 2022-11-11
fix: Rename more instances of "database" to "namespace"

carols10cents:
Follow-on from https://github.com/influxdata/influxdb_iox/pull/6100. There are still more instances of the word "database" and "db", but this is a decent chunk.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,10 +57,9 @@ ENV PACKAGE=$PACKAGE
 COPY --from=build "/root/$PACKAGE" "/usr/bin/$PACKAGE"
 COPY docker/entrypoint.sh /usr/bin/entrypoint.sh
 
-ENV INFLUXDB_IOX_SERVER_MODE=database
 
 EXPOSE 8080 8082
 
 ENTRYPOINT ["/usr/bin/entrypoint.sh"]
 
-CMD ["run", "$INFLUXDB_IOX_SERVER_MODE"]
+CMD ["run"]
--- a/iox_query/src/test.rs
+++ b/iox_query/src/test.rs
@@ -1,5 +1,4 @@
-//! This module provides a reference implementation of
-//! [`QueryDatabase`] for use in testing.
+//! This module provides a reference implementation of [`QueryNamespace`] for use in testing.
 //!
 //! AKA it is a Mock
 
@@ -9,7 +8,7 @@ use crate::{
         ExecutionContextProvider, Executor, ExecutorType, IOxSessionContext,
     },
     Predicate, PredicateMatch, QueryChunk, QueryChunkData, QueryChunkMeta, QueryCompletedToken,
-    QueryDatabase, QueryText,
+    QueryNamespace, QueryText,
 };
 use arrow::{
     array::{
@@ -100,7 +99,7 @@ impl TestDatabase {
 }
 
 #[async_trait]
-impl QueryDatabase for TestDatabase {
+impl QueryNamespace for TestDatabase {
     async fn chunks(
         &self,
         table_name: &str,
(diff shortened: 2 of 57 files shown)


Dockerfile
```diff
@@ -57,10 +57,9 @@ ENV PACKAGE=$PACKAGE
 COPY --from=build "/root/$PACKAGE" "/usr/bin/$PACKAGE"
 COPY docker/entrypoint.sh /usr/bin/entrypoint.sh
 
-ENV INFLUXDB_IOX_SERVER_MODE=database
```
alamb, line 60: that is an embarrassing left over

generated_types/protos/influxdata/iox/write_buffer/v1/write_buffer.proto
```diff
@@ -56,8 +56,11 @@ message WriteBufferCreationConfig {
 
 // A write payload for the write buffer
 message WriteBufferPayload {
+  reserved 1;
```
alamb, line 59: I think this structure is actually serialized in kafka and because of that we need to be careful about changing the field names. 

To pull off a rename like this, I believe we would have to add (and populate) the new `write_batch` as well as the existing `write deploy the changes, wait for all message, and then remove the write once we are sure all data in kafka has `write_batch` and all older messages have been discarded due to retention

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
```diff
@@ -4,12 +4,12 @@ package influxdata.pbdata.v1;
 
 option go_package = "github.com/influxdata/influxdb-pb-data-protocol/golang;influxdbpbdataprotocol";
 
-message DatabaseBatch {
```
alamb, line 7: I am not enough of a protobuf expert, perhaps @mkmik  / @tustvold  /  @domodwyer  know more -- we may simply be able to rename all these fields and keep the names the same and have all existing data be compatible

iox_query/src/test.rs
```diff
@@ -100,7 +99,7 @@ impl TestDatabase {
 }
 
 #[async_trait]
-impl QueryDatabase for TestDatabase {
+impl QueryNamespace for TestDatabase {
```
alamb, line 102: next PR maybe we can rename `TestDatabase` to `TestNamespace` :)

alamb (approved):
I salute you for your attention to detail 👍 

I am worried about the .proto changes -- we should get confirmation  from others that changes to those definitions won't mess up existing data

The rest of the PR looks very good! THANK YOU

generated_types/protos/influxdata/iox/write_buffer/v1/write_buffer.proto
carols10cents: I can totally undo this especially since we're reworking the whole write buffer/Kafka stuff.

generated_types/protos/influxdata/iox/delete/v1/service.proto
```diff
@@ -35,11 +35,17 @@ message DeleteResponse {
 
 // A delete payload
 message DeletePayload {
-  // The name of the database
```
alamb, line 38: Looks like this proto file still has non comment changes

generated_types/protos/influxdata/pbdata/v1/influxdb_pb_data_protocol.proto
carols10cents: aaaaaand it's gone :)

generated_types/protos/influxdata/iox/delete/v1/service.proto
carols10cents: oooops fixed!

alamb:
👩‍🍳 👌

PR #6128, opened 2022-11-12
fix: Revert "fix: Rename database to namespace in service_grpc_flight type…s, vars, and errors"

alamb:
The change in name in the client protocol in https://github.com/influxdata/influxdb_iox/pull/6127 caused clients to break. 

I don't know how to backout the whole PR as it was merged with individual commits. I think this PR contains the commit that is the problem but I would prefer to revert the whole thing
 
This reverts commit 51afe53d23ac9a4b728595041c4230bed65a2d4e.


alamb:
This is no longer needed and instead @jacobmarble  has fixed the problem by updating the clients 🙏 thank you

PR #6133, opened 2022-11-14
fix: Revert database to namespace renames that are broken

carols10cents:
This reverts commit e7ac4eb7fea05068fe6dbd4d4c73d29b57eaad4b, reversing changes made to a3f2fe489c25791d576943e4117a04ab0876b3fc.

Connects to https://github.com/influxdata/idpe/issues/16465


alamb (changes requested):
Sadly we have already changed some downstream clients to use the new field name `namespace_name` so we can't revert #6127 (this PR) without breaking them

I think @carols10cents  is thinking / planning of adding some more backwards compatibility code to accept either `namespace_name` or `database_name`  -- please confirm @carols10cents

PR #6134, opened 2022-11-14
fix: Accept either database_name or namespace_name

carols10cents:
Connects to https://github.com/influxdata/idpe/issues/16465

As far as I can tell, this is only a problem when using JSON -- I did not change the field in the proto in https://github.com/influxdata/influxdb_iox/pull/6127.


jacobmarble (changes requested):
Thanks for this. Let's give a chance to the IDPE change. If it doesn't work, then we'll merge this.

service_grpc_flight/src/lib.rs
```diff
@@ -127,22 +132,60 @@ struct ReadInfo {
...
+            sql_query: read_info.sql_query,
+        })
+    }
+}
+
+#[derive(Deserialize, Debug, Clone)]
+/// Body of the `Ticket` serialized and sent to the do_get endpoint.
+pub struct JsonReadInfo {
+    namespace_name: Option<String>,
+    // Deprecated; use `namespace_name` instead. This is needed until all clients are updated.
```
alamb, line 150: 👍

alamb (approved):
The code looks great -- thank you @carols10cents 

I like that we have this as a backup if the additional updates don't work

alamb:
@jacobmarble  since the IDPE PR seems to be having issues, I would like to propose we merge this one asap to unbreak prod sql queries ASAP. Is that ok?

jacobmarble (approved):

alamb:
https://github.com/influxdata/idpe/pull/16466 appears to have fixed our customer issues. I think this PR is still good to have ready to go in case we see any other issues.

Thank you again for the quick work @carols10cents  🙏

carols10cents:
Thank you to @jacobmarble for cleaning up after the mess I made!! I'll close this PR but leave the branch around just in case.

PR #6135, opened 2022-11-14
fix: Improve testing and documentation to clarify Go client behavior

carols10cents:
Related to https://github.com/influxdata/idpe/issues/16465.

A bunch of the names and comments in this area seemed to imply that the JSON `ReadInfo` ticket for Flight queries was no longer being used-- but it very much is, by the Go clients, and I broke queries because I didn't appreciate that 😮‍💨

So! Add a test and links to the Go clients to better document that dependency, and hopefully prevent some future breakage.

I think this should go in even if https://github.com/influxdata/influxdb_iox/pull/6134 doesn't.

### DIFF
--- a/service_grpc_flight/src/lib.rs
+++ b/service_grpc_flight/src/lib.rs
@@ -127,6 +127,10 @@ struct ReadInfo {
 }
 
 impl ReadInfo {
+    /// The Go clients still use JSON tickets. See:
+    ///
+    /// - <https://github.com/influxdata/influxdb-iox-client-go/commit/2e7a3b0bd47caab7f1a31a1bbe0ff54aa9486b7b>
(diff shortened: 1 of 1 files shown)


jacobmarble (approved):

service_grpc_flight/src/lib.rs
```diff
@@ -127,8 +127,12 @@ struct ReadInfo {
 }
 
 impl ReadInfo {
+    /// The Go clients still use JSON tickets. See:
```
alamb, line 130: Thank you @carols10cents 

FYI @stuartcarnie I think this may be relevant as you plot out FlightSQL

alamb:
Thank you @carols10cents

PR #6136, opened 2022-11-14
test: test coverage for sorting and merging in compactor

alamb:
re https://github.com/influxdata/influxdb_iox/issues/6125

Covers the case of overlapping chunks in the compactor


compactor/src/parquet_file_combining.rs
```diff
@@ -949,6 +1053,79 @@ mod tests {
...
+
+            // ------------------------------------------------
+            // Verify the parquet file content
+
+            // Compacted file has all 6 rows (none are lost in the sort)
+            let batches = table.read_parquet_file(file).await;
+            // Ensure the data is sorted the same way
+            assert_batches_eq!(expected, &batches);
+        }
+    }
```
NGA-TRAN, line 1127: Nice coverage tests

NGA-TRAN (approved):

compactor/src/parquet_file_combining.rs
```diff
@@ -827,6 +827,109 @@ mod tests {
...
+        table.create_column("time", ColumnType::Time).await;
+        let table_schema = table.catalog_schema().await;
+
+        let partition = table
+            .with_shard(&shard)
+            .create_partition("2022-07-13")
+            .await;
+
+        // The sort key comes from the catalog and should be the union of all tags the
+        // ingester has seen
```
crepererum, line 848:
```suggestion
        // The sort key comes from the catalog and should be the union of all tags the
        // ingester has seen and the timestamp
```

compactor/src/parquet_file_combining.rs
```diff
@@ -949,6 +1053,79 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn many_sorted_files_do_not_lose_data() {
+        test_helpers::maybe_start_logging();
+
+        //try several permutations of the files to try and cover as many merge related issues as we can
```
crepererum, line 1060: Instead of doing this manually and makes everybody's brain hurt, just use [the force](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.permutations).

crepererum (approved):
Small nits.

compactor/src/parquet_file_combining.rs
alamb: done in 4bbd0ac78

PR #6137, opened 2022-11-14
test: tests in the reorg planner and query tests for merging parquet files

alamb:
Closes https://github.com/influxdata/influxdb_iox/issues/6125

Covers the case of overlapping chunks in the reorg planner and querier.

These cases are purposely constructed to hit some of the corner cases I was worried about in https://github.com/influxdata/influxdb_iox/issues/6125


query_tests/cases/in/duplicates_parquet_many.expected
```diff
@@ -0,0 +1,78 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
alamb, line 21: This plan has a combination of overlapped and non overlapped chunks

query_tests/src/cases.rs
```diff
@@ -115,19 +171,27 @@ async fn test_cases_two_chunks_sql() {
 
     let input_path = Path::new("cases").join("in").join("two_chunks.sql");
     let mut runner = Runner::new();
-    runner.run(input_path).await.expect("test failed");
-    runner.flush().expect("flush worked");
+    runner
```
alamb, line 174: I don't know why fmt changed this

query_tests/cases/in/duplicates_parquet_many.expected
```diff
@@ -0,0 +1,38 @@
...
++---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
alamb, line 12: This plan has a combination of overlapped and non overlapped chunks

query_tests/cases/in/duplicates_parquet_many.expected
```diff
@@ -0,0 +1,38 @@
...
+|               |                     ParquetExec: limit=None, partitions=[1/1/1/1/00000000-0000-0000-0000-000000000009.parquet], projection=[f, tag, time]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
```
crepererum, line 35: So technically we could group the files by sort key and reduce the number of execs here, right?

crepererum (approved):

query_tests/cases/in/duplicates_parquet_many.expected
alamb: As long as the `ParquetExec`s each had only one file per partition we could put them together and maintain the property that their output was sorted, as required

PR #6138, opened 2022-11-14
fix: Remove namespace string from router -> write buffer -> ingester communication

### LINKED ISSUE #6049: Use table / namespace IDs when communicating between Querier and Ingester

carols10cents:
I highly recommend reviewing this PR commit-by-commit.

## Summary

This PR:

 - Removes the namespace (database) name from `IoxHeaders`, which is metadata that goes into the write buffer but is never read as far as I can tell. CHECK ME ON THIS BEFORE MERGING!!!
 - Deprecates the protobuf `DeletePayload` `db_name` field and the `DatabaseBatch` `database_name` field, which I believe are unused as far as I can tell. CHECK ME ON THIS BEFORE MERGING!!!
 - Removes the namespace (database) name from `Dml` operations as part of https://github.com/influxdata/influxdb_iox/issues/4880
 - Fixes https://github.com/influxdata/influxdb_iox/issues/6049 by updating the documentation in docs/ingester_querier_protocol.md; everything else looks done for that issue.

## Commit descriptions

* fix: Update docs on the ingester-querier protocol.
 I believe this fixes #6049.
* docs: Remove outdated comments about safety; the temporary unsafe code is gone
* fix: Remove namespace name from IoxHeaders in write messages
* fix: Remove database_name field from DatabaseBatch
 It was only being used in one error message.
* fix: Remove now-unused db_name argument to encode_write
* fix: Remove db_name field from DeletePayload
 Doesn't seem to be used anywhere.
* fix: Remove db_name from encode_operation and FileBufferProducer
* test: Keep track of namespaces by ID in ingester TestContext
* fix: Remove namespace name field from DmlWrite and DmlDelete
 But leave the argument in their constructors for now.
 Not all numbers in tests can be 42, Dom.
* fix: Remove namespace from DmlWrite and DmlDelete constructors
* fix: Remove namespace argument from test helper function
* fix: Remove namespace argument from write test helper fn
* fix: Remove namespace argument from kafka test helper fn

### DIFF
--- a/dml/src/lib.rs
+++ b/dml/src/lib.rs
@@ -151,14 +151,6 @@ impl DmlOperation {
         }
     }
 
-    /// Namespace associated with this operation
-    pub fn namespace(&self) -> &str {
-        match self {
-            Self::Write(w) => w.namespace(),
-            Self::Delete(d) => d.namespace(),
-        }
-    }
-
     /// Namespace catalog ID associated with this operation
     pub fn namespace_id(&self) -> NamespaceId {
         match self {
@@ -184,7 +176,6 @@ impl From<DmlDelete> for DmlOperation {
 #[derive(Debug, Clone)]
 pub struct DmlWrite {
     /// The namespace being written to
-    namespace: String,
     namespace_id: NamespaceId,
     /// Writes to individual tables keyed by table name
     tables: HashMap<String, MutableBatch>,
@@ -208,7 +199,7 @@ impl DmlWrite {
     /// - a MutableBatch is empty
     /// - a MutableBatch lacks an i64 "time" column
     pub fn new(
-        namespace: impl Into<String>,
+        _namespace: impl Into<String>,
         namespace_id: NamespaceId,
         tables: HashMap<String, MutableBatch>,
         table_ids: HashMap<String, TableId>,
@@ -236,7 +227,6 @@ impl DmlWrite {
         }
 
         Self {
-            namespace: namespace.into(),
             tables,
             table_ids,
             partition_key,
@@ -245,11 +235,6 @@ impl DmlWrite {
             max_timestamp: stats.max.unwrap(),
             namespace_id,
         }
-    }
-
-    /// Namespace associated with this write
-    pub fn namespace(&self) -> &str {
-        &self.namespace
     }
 
     /// Metadata associated with this write
@@ -343,7 +328,6 @@ impl DmlWrite {
 /// A delete operation
 #[derive(Debug, Clone, PartialEq)]
 pub struct DmlDelete {
-    namespace: String,
     namespace_id: NamespaceId,
     predicate: DeletePredicate,
     table_name: Option<NonEmptyString>,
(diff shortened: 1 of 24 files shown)


dml/src/lib.rs
```diff
@@ -151,21 +151,7 @@ impl DmlOperation {
...
-        }
-    }
-
     /// Namespace catalog ID associated with this operation
-    ///
-    /// # Safety
-    ///
-    /// Marked unsafe because of the critical invariant; Kafka conumers MUST NOT
-    /// utilise this method until this warning is removed. See [`DmlWrite`]
-    /// docs.
```
domodwyer, line 168: Thank you :pray:

router/tests/http.rs
```diff
@@ -179,7 +179,6 @@ async fn test_write_ok() {
     let writes = ctx.write_buffer_state().get_messages(ShardIndex::new(0));
     assert_eq!(writes.len(), 1);
     assert_matches!(writes.as_slice(), [Ok(DmlOperation::Write(w))] => {
-        assert_eq!(w.namespace(), "bananas_test");
```
domodwyer, line 182: As this is an e2e test / last line of defence, it would be good if we could assert the correct ID was embedded in the write. If you add this snippet I have from the future to the `TestContext` impl:

```rust
    /// Return the [`TableId`] in the catalog for `name` in `namespace`, or panic.
    pub async fn table_id(&self, namespace: &str, name: &str) -> TableId {
        let mut repos = self.catalog.repositories().await;
        let namespace_id = repos
            .namespaces()
            .get_by_name(namespace)
            .await
            .expect("query failed")
            .expect("namespace does not exist")
            .id;

        repos
            .tables()
            .get_by_namespace_and_name(namespace_id, name)
            .await
            .expect("query failed")
            .expect("no table entry for the specified namespace/table name pair")
            .id
    }
```

We can change this (and the two other) lines to:

```rust
        let table_id = ctx.table_id("bananas_test", "platanos").await;
        assert!(w.table(table_id).is_some());
```

And then we have coverage of the namespace resolution + id plumbing (though I think it might always be ID 1, perhaps we can call `ensure_namespace()` a few times at the start of the test to advance the ID counter a bit) :+1:

PR #6144, opened 2022-11-15
feat: compactor ignores max file count for first file

### LINKED ISSUE #6091: Limit number of files to compact per partition even if we have budget to do more
# Rationale

Currently, we have this config param `INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES` that limits the number of files to compact to avoid OOM. If the minimum number of overlapped files  exceed that number, we put its partition into the `skipped_compactions` table to avoid considering it again.

Recently, since we improve our memory usage, we are able to increase that limit to have the compact to compact many files at once. However, it seems to me from the compactor dashboard https://grafana.prod01-eu-central-1.aws.influxdata.io/d/z7-FhLsnk/compactor?orgId=1, compacting 20 files at once may not be faster than compacting them twice 10 files each. 

This ticket is to create a different limit, INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES_PER_PARTITION, to limit number of files per partition even if we have more resource to do more. With this support, we will compact a smaller number of files per partition (which I predict DF handles better) but we will concurrently compact more partitions.

lukebond:
Closes #6091

This makes the following change in the compactor's `filter_parquet_files` function, which selects a subset of a candidate partition's parquet files and is used for cold and hold compaction: if the first `LN` parquet file, plus its overlapping `LN+1` files, overshoot the max file count per partition (from existing envvar `INFLUXDB_IOX_COMPACTION_MAX_COMPACTING_FILES`), then take those files for processing anyway, but don't proceed to take any further files for this partition.

The rationale being that a partition with many small overlapping files may get perpetually skipped even if we have the memory budget for it.


compactor/src/lib.rs
```diff
@@ -156,33 +156,6 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
                 FilterResult::NothingToCompact => {
                     debug!(?partition_id, compaction_type, "nothing to compact");
                 }
-                FilterResult::OverLimitFileNum {
```
lukebond, line 159: whilst we still have a max file count limit, we no longer need this variant because it's only used in the scenario we're changing: the first file in a partition

compactor/src/parquet_file_filtering.rs
```diff
@@ -187,20 +183,17 @@ fn filter_parquet_files_inner(
         // Over limit of num files
         // At this stage files_to_return only includes LN+1 files. To get both returning LN+1 and
         // LN files, we need to consider both files_to_return and level_n_to_return
+        // The file limit only applies after the first file, because want to avoid the situation
+        // where a file (and its overlapping files) will never get processed. In this scenario,
+        // however, we do not proceed to pick up any more files other than this first one (and its
+        // overlaps)
         if files_to_return.len() + level_n_to_return.len() + 1 /* LN file */ + overlaps.len()
             > max_num_files
+            && !(files_to_return.is_empty() && level_n_to_return.is_empty())
```
lukebond, line 192: this is the only logic change in this PR, the rest is removing unused code and adding tests

PR #6145, opened 2022-11-15
feat: define router -> ingester write protocol

domodwyer:
This PR defines a simple request/response placeholder for a direct RPC write from a router, to an ingester.

---

* feat: define router -> ingester write protocol (dcc7b10bc)

      Specify a gRPC service and request/response message formats to push
      writes directly from a router to an ingester.

### DIFF
--- a/generated_types/build.rs
+++ b/generated_types/build.rs
@@ -54,6 +54,7 @@ fn generate_grpc_types(root: &Path) -> Result<()> {
         ingester_path.join("parquet_metadata.proto"),
         ingester_path.join("query.proto"),
         ingester_path.join("write_info.proto"),
+        ingester_path.join("write.proto"),
         namespace_path.join("service.proto"),
         object_store_path.join("service.proto"),
         predicate_path.join("predicate.proto"),
--- a/generated_types/protos/influxdata/iox/ingester/v1/write.proto
+++ b/generated_types/protos/influxdata/iox/ingester/v1/write.proto
@@ -1,0 +1,15 @@
+syntax = "proto3";
+package influxdata.iox.ingester.v1;
+option go_package = "github.com/influxdata/iox/ingester/v1";
+
+import "influxdata/pbdata/v1/influxdb_pb_data_protocol.proto";
+
+service WriteService {
+  rpc Write(WriteRequest) returns (WriteResponse);
+}
+
+message WriteRequest {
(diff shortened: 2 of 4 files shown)


generated_types/protos/influxdata/iox/ingester/v1/write.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package influxdata.iox.ingester.v1;
+option go_package = "github.com/influxdata/iox/ingester/v1";
```
domodwyer, line 3: This defines the proto package + import path.

generated_types/protos/influxdata/iox/ingester/v1/write.proto
```diff
@@ -0,0 +1,15 @@
...
+
+import "influxdata/pbdata/v1/influxdb_pb_data_protocol.proto";
```
domodwyer, line 5: This imports the existing write protocol data type (`DatabaseBatch`)

generated_types/build.rs
```diff
@@ -54,6 +54,7 @@ fn generate_grpc_types(root: &Path) -> Result<()> {
         ingester_path.join("parquet_metadata.proto"),
         ingester_path.join("query.proto"),
         ingester_path.join("write_info.proto"),
+        ingester_path.join("write.proto"),
```
domodwyer, line 57: And this configures the rust build step to generate rust code from the proto file.

test_helpers_end_to_end/src/client.rs
```diff
@@ -51,18 +50,6 @@ pub fn get_write_token(response: &Response<Body>) -> String {
         .to_string()
 }
 
-/// Extracts the write token from the specified response (to the gRPC write API)
-pub fn get_write_token_from_grpc(response: &tonic::Response<WriteResponse>) -> String {
```
domodwyer, line 55: This is just dead code conflicting with names / leftover from removal of the old external gRPC write API in the router.

pauldix (approved):

pauldix:
@aierui yes, we're removing Kafka from our setup. We've found that we don't need all the extra guarantees and functionality it provides. It also adds quite a bit of cost, so we're optimizing for a more efficient and dynamically scalable ingestion pipeline. I'll probably write a more public blog post about it once we've tested this at scale in our mirrored production environment.

PR #6148, opened 2022-11-15
feat: reject writes that are outside the retention period

### LINKED ISSUE #6027: Router rejects data outside the retention period
# Rationale

This is a task of #3053. If client loads data outside of retention period (time < now - namespace's retention_period_nano), router will return a 400 error

NGA-TRAN:
Closes #6027


router/src/server/http.rs
```diff
@@ -140,6 +140,12 @@ impl From<&DmlError> for StatusCode {
 
             DmlError::Internal(_) | DmlError::WriteBuffer(_) => StatusCode::INTERNAL_SERVER_ERROR,
             DmlError::Partition(PartitionError::BatchWrite(_)) => StatusCode::INTERNAL_SERVER_ERROR,
+            DmlError::Retention(RetentionError::NamespaceLookup(_)) => {
+                StatusCode::INTERNAL_SERVER_ERROR
+            }
+            DmlError::Retention(RetentionError::OutsideRetention) => {
+                StatusCode::RANGE_NOT_SATISFIABLE
```
NGA-TRAN, line 147: Let if know if this is there right error code or we should create a new one for this

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Time is outside the retention period.
+    #[error("the write is outside of the retention period")]
+    OutsideRetention,
+}
+
+/// A [`DmlHandler`] implementation that validates that the write is within the
+/// retention period of the  namespace
+#[derive(Debug)]
+pub struct RetentionValidator<C = Arc<InstrumentedCache<MemoryNamespaceCache>>, P = SystemProvider>
```
NGA-TRAN, line 31: Is SystemProvider the right one to use here?

NGA-TRAN (commented):
@domodwyer : The only missing in this PR is tests. Can you quickly skimmed if I am doing what you expected?
I will modify the  `//! IOx router implementation.`  to `add RetentionValidator` into the diagram, too

NGA-TRAN:
I am working on tests

router/src/dml_handlers/retention_validator.rs
domodwyer: It is! 🎉

ioxd_router/src/lib.rs
```diff
@@ -204,6 +204,11 @@ pub async fn create_router_server_type(
     let schema_validator =
         InstrumentationDecorator::new("schema_validator", &metrics, schema_validator);
 
+    // Add a retention validator into handler stack to reject data outside the retention period
+    let retention_validator = RetentionValidator::new(Arc::clone(&catalog), Arc::clone(&ns_cache));
+    let retention_validator =
+        InstrumentationDecorator::new("retention_validator", &metrics, retention_validator);
```
domodwyer, line 210: Thank you for including this wrapper - it means we automatically measure the call latency and success/error rates for this handler, and emit tracing spans without any more work.

The metrics can be added to the metrics dashboard after this PR is merged 🎉

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Errors emitted during retention validation.
+#[derive(Debug, Error)]
+pub enum RetentionError {
+    /// The requested namespace could not be found in the catalog.
+    #[error("failed to read namespace schema from catalog: {0}")]
+    NamespaceLookup(iox_catalog::interface::Error),
+
+    /// Time is outside the retention period.
+    #[error("the write is outside of the retention period")]
+    OutsideRetention,
```
domodwyer, line 25: Maybe we add the name to help the user discover which table contains the data outside the retention window:

```suggestion
    #[error("data in table {0} is outside of the retention period")]
    OutsideRetention(String),
```

ingester/tests/common/mod.rs
```diff
@@ -194,6 +194,7 @@ impl TestContext {
                         self.topic_id,
                         self.query_id,
                         iox_catalog::DEFAULT_MAX_COLUMNS_PER_TABLE,
+                        None,
```
domodwyer, line 197: It might be good to make this a parameter of the function - then the tests can specify a retention period for the namespace, which means we can write a router end-to-end test that ensures the retention check runs properly, and returns the correct error message 👍

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,131 @@
...
+            for (_table_id, (_table_name, batch)) in &batch {
+                let time_summary = batch.timestamp_summary();
+                if let Some(time_summary) = time_summary {
+                    let min = time_summary.stats.min;
+                    if let Some(min) = min {
+                        if min < min_retention {
+                            return Err(RetentionError::OutsideRetention);
+                        }
+                    }
+                }
```
domodwyer, line 113: We can simplify this block for the reader:

```suggestion
                                if let Some(min) = batch.timestamp_summary().and_then(|v| v.stats.min) {
                    if min < min_retention {
                        return Err(RetentionError::OutsideRetention);
                    }
                }
```

That removes some redundant nesting and value matching which makes it a little easier to read :+1:

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,131 @@
...
+                self.cache
+                    .put_schema(namespace.clone(), Arc::clone(&schema));
+
+                trace!(%namespace, "schema cache populated");
+                schema
+            }
+        };
+
+        // retention is not infinte, validate all lines of a write are within the retention period
+        // schema.retention_period_ns.map(|retention_period_ns| {
```
domodwyer, line 100: 👀

router/src/server/http.rs
domodwyer: This has a defined meaning that we probably shouldn't use (it is not applicable for us) - I find https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors to be helpful when deciding.

Personally I would use `StatusCode::FORBIDDEN` because it is defined as:

> The request contained valid data and was understood by the server, but the server is refusing action. This may be due to the user not having the necessary permissions for a resource or needing an account of some sort, or attempting a prohibited action (e.g. creating a duplicate record where only one is allowed).

We would be rejecting the request because the user is forbidden from writing the data that is outside their account's retention period :+1:

router/src/namespace_cache/memory.rs
```diff
@@ -46,6 +46,7 @@ mod tests {
             query_pool_id: QueryPoolId::new(1234),
             tables: Default::default(),
             max_columns_per_table: 50,
+            retention_period_ns: None,
```
domodwyer, line 49: Perhaps we set this to `Some(value)` so that the test ensures it returns the correct value in the assert below (instead of using the "default value" which is `None`) 👍

domodwyer:
(I "request changes" only to make sure we add the tests / they get reviewed)

ingester/tests/common/mod.rs
NGA-TRAN: So I made it s parameter but so far all its tests values are `None`. I need to make one not none but with today OOM incident, I have not had much time for this. Will continue tomorrow

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,208 @@
...
+mod tests {
+    use iox_tests::util::{TestCatalog, TestNamespace};
+    use once_cell::sync::Lazy;
+    use std::sync::Arc;
+
+    use super::*;
+
+    static NAMESPACE: Lazy<NamespaceName<'static>> = Lazy::new(|| "bananas".try_into().unwrap());
+
+    #[tokio::test]
```
NGA-TRAN, line 138: @domodwyer : Here are the unit tests

I have addressed all your comments except this one https://github.com/influxdata/influxdb_iox/pull/6148#discussion_r1024593727 which I think may be used as an integration test to test a chain of write (retention validation, schema validation and partitioning)?

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,208 @@
...
+        // Create the validator whse retention period is 1 hour
+        let handler =
+            RetentionValidator::new(catalog.catalog(), Arc::new(MemoryNamespaceCache::default()));
+
+        // Make time now to be inside the retention period
+        let now = SystemProvider::default()
+            .now()
+            .timestamp_nanos()
+            .to_string();
+        let line = "bananas,tag1=A,tag2=B val=42i ".to_string() + &now;
```
domodwyer, line 154: I...didn't know you could concatenate strings like this!

router/src/dml_handlers/retention_validator.rs
```diff
@@ -0,0 +1,208 @@
...
+        let writes = lp_to_writes(&line);
+
+        let result = handler
+            .write(&NAMESPACE, NamespaceId::new(42), writes, None)
+            .await;
+
+        // error means the time is outside the retention period
+        assert!(result.is_err());
+        let message = result.unwrap_err().to_string();
+        assert!(message.contains("data in table bananas is outside of the retention period"));
```
domodwyer, line 190: Very cool :+1:

router/src/dml_handlers/retention_validator.rs
domodwyer: They look very good - very easy tests to write with the handler approach right?! 😄 

It might be worth adding two more tests because I think it is an important behaviour we discussed as a team: a write that partially overlaps the retention period.

* One table, which has timestamps both inside, and outside of the retention period limit
* Two tables, one which has timestamps inside, and one that has timestamps outside of the retention period limit

That way we cover the "partial retention" situation we discussed :+1:

---

> I have addressed all your comments except this one https://github.com/influxdata/influxdb_iox/pull/6148#discussion_r1024593727 which I think may be used as an integration test to test a chain of write (retention validation, schema validation and partitioning)?

Yes we should have an integration test that ensures a write is rejected and nothing is placed into kafka - you can copy the `test_schema_conflict` integration test and change the LP to be inside the retention period:

https://github.com/influxdata/influxdb_iox/blob/main/router/tests/http.rs#L237-L288

If you need help get me on zoom! :+1:

domodwyer (approved):
I think the code looks really good! A nice addition to the router - nice work @NGA-TRAN.

It would be good to add the test coverage described below before merging too :+1:

router/src/dml_handlers/retention_validator.rs
NGA-TRAN: I have added your suggested unit tests and integration tests

router/src/dml_handlers/retention_validator.rs
NGA-TRAN: After I installed git autopilot, I have been a lot more productive with Rust 🙂

NGA-TRAN (commented):
Thanks so much for all suggestions @domodwyer . Super helpful to me. Since you have approved it, I will merge it. If you have any concerns or suggestions for my last commit, I am happy to change them in a separate PR

domodwyer:
I reviewed the changes/tests - perfect! Thank you @NGA-TRAN 🙏

iox_catalog/src/mem.rs
```diff
@@ -303,7 +306,7 @@ impl NamespaceRepo for MemTxn {
             query_pool_id,
             max_tables: DEFAULT_MAX_TABLES,
             max_columns_per_table: DEFAULT_MAX_COLUMNS_PER_TABLE,
-            retention_period_ns: DEFAULT_RETENTION_PERIOD,
+            retention_period_ns: MEM_DEFAULT_RETENTION_PERIOD,
```
domodwyer, line 309: I should have spotted this but didn't - the mem catalog + postgres catalog should have the same defaults, otherwise it makes testing very confusing (because they behave differently from the "real" catalog). This should have done the same as the postgres impl, and I think it has made life tricky for @lukebond in https://github.com/influxdata/influxdb_iox/pull/6103 (my apologies).

I'm just leaving this comment so we all agree going forwards the mem catalog & postgres catalog should have identical behaviours / defaults.

lukebond: fixing this as part of #6103

NGA-TRAN: Thanks @domodwyer and @lukebond. Sorry for the trouble and thanks for the fix
I should have left a comment here that I have planned to create 2 tickets: 

1. Make the Mem and the Postgres value identical after the grpc create namespace PR is merged. It was hard to do it without create namespace because I have to change a lot of tests unnecessary. Sorry I did not think it would break the create namespace PR, too. 
2. Update router cache when we update the retention policy. Otherwise, we will do wrong rejection

It seems Luke has fixed the first one in the create namespace  PR already. Let me create a ticket for the second one only

PR #6156, opened 2022-11-16
fix: only push safe select expression through de-dup

### LINKED ISSUE #6066: Wrong answers when data was updated and filters applied
__Steps to reproduce:__
Run this in one shell
```shell
influxdb_iox run -vv --persist-partition-size-threshold-bytes=1
```

Then run all of these commands quickly:
```shell
echo "table foo=1,bar=1 0" > /tmp/foo1.lp
echo "table bar=2 0" > /tmp/foo2.lp
influxdb_iox write --host http://localhost:8080  my_db /tmp/foo1.lp
influxdb_iox write --host http://localhost:8080  my_db /tmp/foo2.lp
influxdb_iox query my_db 'select * from "table" where foo=1 AND bar=2'
```

You'll see no rows 

__Expected behaviour:__
Rather than the expected result of
```
+-----+-----+----------------------+
| bar | foo | time                 |
+-----+-----+----------------------+
| 2   | 1   | 1970-01-01T00:00:00Z |
+-----+-----+----------------------+
```

__Actual behaviour:__
```
++
++
```

Note this actual behavior only occurs when the data are in two different chunks (in this case two different write commands)

__Details__
As described by @crepererum  and @NGA-TRAN on https://github.com/influxdata/influxdb_iox/pull/5810 https://github.com/influxdata/influxdb_iox/pull/5810#discussion_r1004492275 the issue is due to applying filtering prior to deduplication

For the example above:

```
table foo=1,bar=1 0
table bar=2 0
```

Predicate `foo=1 AND bar=2` will filter out both lines prior to de-dup (when it should only be applied to the deduplicated data which will pass), so this predicate MUST NOT be applied early.

crepererum:
Fixes #6066.


predicate/src/lib.rs
```diff
@@ -484,6 +489,45 @@ impl Predicate {
         self
     }
 
+    /// Push this predicate down through deduplication.
```
alamb, line 492:
```suggestion
    /// Remove any clauses of this predicate that can not be run before deduplication.
    ///
    /// See https://github.com/influxdata/influxdb_iox/issues/6066 for more details
```

predicate/src/lib.rs
```diff
@@ -484,6 +489,45 @@ impl Predicate {
...
+                    // bail out, do NOT include this weird expression
+                    return false;
+                }
+
+                // check if all columns are part of the primary key
+                if !columns.into_iter().all(|c| pk.contains(c.name.as_str())) {
+                    return false;
+                }
+
+                expr.accept(RowBasedVisitor::default())
```
alamb, line 515: for what it is worth, I would never expect to see an `AggregateFunction`, GroupingSet or WindowFunction pushed down to a table scan. That would indicate to me a pretty serious bug in DataFusion

So I wonder if we could simply make this return an error if it found a non row based predicate 🤔

predicate/src/lib.rs
```diff
@@ -484,6 +489,45 @@ impl Predicate {
         self
     }
 
+    /// Push this predicate down through deduplication.
+    ///
+    /// For this to work we will only allow some selected expressions, mainly the row-based ones that only act on
+    /// primary keys (and constants).
```
alamb, line 495: I spent a while convincing myself that it was ok to push through predicates that only referred to primary keys.

I did eventually convince myself  that any predicate that operates solely on primary key columns will either pass or filter all rows with that primary key. Thus you can't construct some case where one of the rows is filtered out prior to deduplicate. I tried to suggest some comments (mostly as a help to my future self when I wonder about this in the future when reading this code)

alamb:
```suggestion
    /// Only expressions that are row-based and refer to primary key columns (and constants) 
    /// can be evaluated prior to deduplication. 
    ///
    /// If a predicate can filter out some but not all of the rows with
    /// the same primary key, it may filter out the row that should have been updated
    /// allowing the original through, producing incorrect results.  
    ///
    /// Any predicate that operates solely on primary key columns will either pass or filter 
    /// all rows with that primary key and thus is safe to push through.
```

alamb (approved):
Thank you @crepererum  -- this looks great

I spent quite a while convincing myself this was correct. Among other things I reverted the code changes and verified that the tests fail with wrong results, as expected. 

Screenshot 2022-11-16 at 3 35 56 PM

predicate/src/lib.rs
crepererum: Are there no operations like "rolling window" or "prev row" that can get pushed down? If the planner knows that some data is sorted, I thought it could come up with "prev row" or something.

alamb: the only place the previous row can be referred to that I know of is in window functions (like `SELECT LAG(x, 1) OVER ()...`) but semantically those all happen after filtering 

I suppose it would be theoretically possible to push them down to a scan if the optimizer knew the data was sorted but Datafusion definitely doesn't do that now and I think there are a bunch of details that would have to be sorted out, such as what you do at the RecordBatch boundaries

predicate/src/lib.rs
```diff
@@ -934,4 +1044,100 @@ mod tests {
...
+
+        // disassemble AND
+        assert_eq!(
+            Predicate {
+                field_columns: None,
+                range: None,
+                exprs: vec![col("tag1")
+                    .eq(lit("foo"))
+                    .and(col("field1").eq(lit(1.0)))
+                    .and(col("time").eq(lit(1))),],
```
NGA-TRAN, line 1111: I do not remember if we allow `OR` in the expression at this low level but I think with your implementation of `push_through_dedup`, it does not matter if we allow or not. Maybe add a test with OR?

NGA-TRAN (approved):
It is so nice to see the detail of the predicates pushed down or not in plans.

Everything looks great. I just wonder how it works with OR expressions and a combination of AND and OR. Maybe add unit tests for `push_through_dedup` and SQL and their explains for them?

predicate/src/lib.rs
crepererum: will add a test, but we won't split them.

crepererum: done

alamb:
🎉

Issue #6157, opened 2022-11-16
Put a cap on number of files for the minimal must-compact number of files


NGA-TRAN:
# Rationale

With recent change https://github.com/influxdata/influxdb_iox/pull/6144#event-7822870930, we do not cap the max number of files for the first/minimal number of files to compact but only rely on memory cap. We soon hit OOM and the investigation shows the memory needs actually can be very small https://influxdata.slack.com/archives/C040Z1T53RB/p1668624115812019. I think we need to cap number of files for this case, too.

@lukebond Maybe you and I should brainstorm on this

We want this code to increase this temporarily settings which are too low
https://github.com/influxdata/k8s-idpe/pull/13786

NGA-TRAN:
Let us go over this to discuss why we need a cap for this

```SQL
SELECT shard_index, table_id, partition_id, partition_key, left(reason, 25),
  num_files, limit_num_files, estimated_bytes, limit_bytes, to_timestamp(skipped_at) skipped_at
FROM skipped_compactions, partition, shard
WHERE partition.id = skipped_compactions.partition_id and partition.shard_id = shard.id
   and table_id not in (42, 188)
ORDER BY skipped_at desc;
 shard_index | table_id | partition_id | partition_key |      left      | num_files | limit_num_files | estimated_bytes | limit_bytes |    skipped_at
-------------+----------+--------------+---------------+---------------------------+-----------+-----------------+-----------------+-------------+------------------------
     59 |    22 |   19233159 | 2022-11-12  | over limit of num_files  |    45 |       40 |   3773528272 | 32212254720 | 2022-11-14 10:27:32+00
     52 |    32 |   19233158 | 2022-11-12  | over limit of num_files  |    41 |       40 |   3498023096 | 32212254720 | 2022-11-14 10:27:32+00
     59 |    22 |   13711168 | 2022-11-07  | over limit of num_files  |    64 |       40 |   1338558970 | 32212254720 | 2022-11-07 22:09:18+00
     52 |    32 |   13710553 | 2022-11-07  | over limit of num_files  |    63 |       40 |   1463606446 | 32212254720 | 2022-11-07 22:07:08+00
```

lukebond:
if i'm reading this correctly, we had a few examples of this feature compacting 60+ files and needing very little memory (and wasn't the culprit of the OOMs). but you think it's prudent to have a limit regardless?

if so i would guess around `80` might be a good limit. what do you think?

[...]

PR #6158, opened 2022-11-16
chore: Revert "feat: compactor ignores max file count for first file (#6144)"

NGA-TRAN:
This reverts commit bf1681f4fe99f70edb3b1e98a09396c64f7ccbd7.

This is the revert of #6144 
@lukebond : Sorry I need to revert this because of OOM that need big changes in settings in  to avoid and we do not want to do so (it will affect other partitions). I think we need a cap on this thi. Memory estimation for large number of files and small number of columns needs more thinking. 
I have created #6157 for us to work with this #6091


cannonpalms (approved):

NGA-TRAN:
@lukebond It turns out this one sis not the culprit. The culprit is unnecessary sorts that allocate more memory after upgrading DF. We have reverted it and open the bug here https://github.com/influxdata/idpe/issues/16500

However, let us add a cap for this before merging it back. I have added SQL results in #6157 and for us to discuss

PR #6159, opened 2022-11-16
chore: Revert "chore: Update datafusion again (#6108)"

alamb:
This reverts commit fbe9f27f107434623862c87626d8ad691eae495d.

re https://github.com/influxdata/idpe/issues/16500

We suspect this upgrade is causing the compactor to OOM so plan to back it out to test the hypothesis


NGA-TRAN (approved):
Thanks @alamb

PR #6166, opened 2022-11-17
feat: Introduce InfluxQL to Flight

### LINKED ISSUE #6114: InfluxQL: Create skeleton for parsing and processing InfluxQL queries
# Summary

Establish a minimal framework for teaching IOx how to execute an InfluxQL query.

# Overview

IOx will be able to execute an InfluxQL query, parse it and then return a `not implemented` error when attempting to execute it. This will establish the framework necessary to implement the supported InfluxQL queries. The following tasks will establish that framework:

- [ ] Extend Flight RPC to allow `do_get` to distinguish SQL from InfluxQL queries[^1]
- [ ] Add InfluxQL frontend to `iox_query` crate to parse InfluxQL queries
- [ ] Add InfluxQL planner to `iox_query` that rejects all parsed InfluxQL queries with a custom error for each query type
- [ ] Extend `influxdb_iox query` command to optionally parse a `--language|-l` flag to distinguish SQL from InfluxQL[^2]

[^1]: The interface must be extended in such a way that it is backwards compatible.
[^2]: The default behaviour when omitting the flag is to assume SQL

stuartcarnie:
Closes #6114

With this PR, we have taught IOx a few new tricks. 

### IOx CLI

The IOx CLI can now execute a SQL or InfluxQL query via the CLI via a new `--lang` argument. The default behaviour when the argument is omitted is to treat the query as SQL.

### IOx Flight gRPC API

The `ReadInfo` structure has a new enumerated field that specifies the type of query text contained in the `sql_query` field. We tried using an enum, but that broke the compatibility checks. As it stands, an existing client could still execute using a prior version of the `ReadInfo` struct and it would correctly be treated as SQL.

#### `frontend` module

This crate has a new InfluxQL planner in the frontend module. The planner is responsible for parsing the query and then handing off to the ~`InfluxQLToRel`~ `InfluxQLToLogicalPlan` type for transforming the InfluxQL AST into a relational / logical plan.

#### `plan` module

The `plan` module contains an implementation of the InfluxQL -> LogicalPlan planner. This planner returns a not implemented error for all InfluxQL statement types. It will serve as the foundation for implementing the planner for `SELECT` statements to begin with.


generated_types/protos/influxdata/iox/querier/v1/flight.proto
```diff
@@ -7,8 +7,14 @@ message ReadInfo {
   // Namespace name.
   string namespace_name = 1;
 
-  // SQL query.
-  string sql_query = 2;
+  reserved "sql_query";
```
crepererum, line 10: We normally evolve field names and IDs at the same time, because we don't wanna deal with the differences of our clients (some use text-based interfaces which use field names, other use binary interfaces which use IDs). So I would reserve `2` as well and start at `3` down below.

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -741,3 +741,39 @@ async fn repeat_ingester_request_based_on_querier_logs(state: &StepTestState<'_>
         .arg(log_data["predicate_binary"].as_str().unwrap())
         .assert()
 }
+
+// region InfluxQL Tests
```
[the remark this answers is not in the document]
stuartcarnie: Done

generated_types/protos/influxdata/iox/querier/v1/flight.proto
```diff
@@ -9,6 +9,17 @@ message ReadInfo {
 
   // SQL query.
   string sql_query = 2;
+
+  QueryType query_type = 3;
+
+  enum QueryType {
+    // An unspecified query type. IOx may choose how to interpret sql_query.
+    QUERY_TYPE_UNSPECIFIED = 0;
```
stuartcarnie, line 17: This is required by the `bufbuild` linter, presumably for being able to identify an older client that has not specified a value for the `query_type` field. 0 is the default value of an enum when decoding a protobuf, so using that for a specific enumeration appears to be an anti-pattern.

influxdb_iox/src/commands/query.rs
```diff
@@ -31,14 +41,20 @@ pub struct Config {
     /// Optional format ('pretty', 'json', or 'csv')
     #[clap(short, long, default_value = "pretty", action)]
     format: String,
+
+    /// Query type used
+    #[clap(short = 'l', long = "lang", default_value = "sql")]
+    query_lang: QueryLanguage,
```
stuartcarnie, line 47: Support specifying a language type for the SQL text. Defaults to `sql` when omitted.

influxdb_iox/src/commands/sql/repl.rs
```diff
@@ -398,6 +400,7 @@ async fn scrape_query(
         .perform_query(ReadInfo {
             namespace_name: db_name.to_string(),
             sql_query: query.to_string(),
+            query_type: read_info::QueryType::Sql as i32,
```
stuartcarnie, line 403: It's unfortunate that enumerated fields are not strongly typed on the generated types 😭

influxdb_iox/tests/end_to_end_cases/influxql.rs
```diff
@@ -0,0 +1,33 @@
+use test_helpers_end_to_end::{maybe_skip_integration, MiniCluster, Step, StepTest};
+
+#[tokio::test]
```
crepererum, line 3: I would like to see an end2end test for the CLI as well at some point, but this can be done in a later PR.

Prior art:
https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/tests/end_to_end_cases/cli.rs

crepererum (approved):
Small test nitpick, rest looks great. I'm esp. happy about how clean the wiring is :+1:

service_grpc_flight/src/lib.rs
```diff
@@ -134,19 +155,32 @@ impl ReadInfo {
...
 
         Ok(Self {
-            namespace_name: read_info.namespace_name,
-            sql_query: read_info.sql_query,
+            namespace_name: read_info.namespace_name.clone(),
+            query: match read_info.query_type {
+                0 /* unspecified */ | 1 /* SQL */ => Query::Sql(read_info.sql_query),
+                2 => Query::InfluxQL(read_info.sql_query),
+                _ => return Err(Error::InvalidQueryType),
+            },
```
alamb, line 183: There is a function query_type() that is strongly typed -- I recommend you could use

```suggestion
            query: match read_info.query_type() {
                QueryType::Unspecified | QueryType::Sql => Query::Sql(read_info.sql_query),
                QueryType::InfluxQl  => Query::InfluxQL(read_info.sql_query),
            },
```

service_common/src/planner.rs
```diff
@@ -43,6 +44,22 @@ impl Planner {
...
+    pub async fn influxql(
+        &self,
+        database: Arc<dyn QueryNamespace>,
+        query: impl Into<String> + Send,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let planner = InfluxQLQueryPlanner::new();
+        let query = query.into();
+        let ctx = self.ctx.child_ctx("planner influxql");
+
+        self.ctx
```
alamb, line 58: 👍

iox_query/src/frontend/influxql.rs
```diff
@@ -0,0 +1,46 @@
...
+        let mut statements = parse_statements(query)
+            .map_err(|e| DataFusionError::External(format!("{}", e).into()))?;
+
+        if statements.len() != 1 {
+            return Err(DataFusionError::NotImplemented(
+                "The context currently only supports a single SQL statement".to_string(),
+            ));
+        }
+
+        let planner = InfluxQLToRel::new(&ctx, database);
```
alamb, line 38: The `ToRel` naming is a legacy hold over from DataFusion -- maybe this would be better called `InfluxQLToLogicalPlan` or something (as a follow on PR)

service_grpc_flight/src/lib.rs
```diff
@@ -486,7 +534,70 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_read_info_decoding() {
+        let mut buf = Vec::with_capacity(1024);
+        proto::ReadInfo::encode(
+            &proto::ReadInfo {
+                namespace_name: "<foo>_<bar>".to_string(),
+                sql_query: "SELECT 1".to_string(),
+                query_type: QueryType::Unspecified as i32,
```
alamb, line 547: You can write this slighty more idomatically like

```suggestion
                query_type: QueryType::Unspecified.into(),
```

test_helpers_end_to_end/src/client.rs
```diff
@@ -190,15 +191,13 @@ pub fn all_persisted(res: &GetWriteInfoResponse) -> bool {
 }
 
 /// Runs a query using the flight API on the specified connection.
-///
-/// This is similar ot [`run_query`] but does NOT unwrap the result.
 pub async fn try_run_query(
-    sql: impl Into<String>,
+    sql_query: String,
```
alamb, line 195: An `impl into<String>` allows the caller to pass in a `&str` or `String` -- I wonder why you have changed to require passing in a String?

I doubt it matters, I am just curious

alamb (approved):
Very impressive @stuartcarnie  -- I left some small stylistic comments (especially around protobuf generated code handling) but nothing that would prevent merging this PR

influxdb_iox/src/commands/sql/repl.rs
alamb: I think a more idomatic way to do this is

```suggestion
            query_type: read_info::QueryType::Sql.into(),
```

So at least you don't have `as` sprinkled around

stuartcarnie: Thanks! I thought I had to implement one of the Into / From traits to do that

stuartcarnie: Updated all references 👍🏻

service_grpc_flight/src/lib.rs
stuartcarnie: Ahh yes, thank you! Cleans up the code nicely too, and removes the possibility of an error.

iox_query/src/frontend/influxql.rs
stuartcarnie: Agree – the `InfluxQLToRel` name is 👎🏻 .  Updated, thanks!

test_helpers_end_to_end/src/client.rs
stuartcarnie: I've restored the `impl Into<String>` argument, which is better. Likely an unintended change, when I was transitioning those functions.

influxdb_iox/tests/end_to_end_cases/influxql.rs
stuartcarnie: I've added it to this PR for completeness – thanks for the tip!

stuartcarnie:
@crepererum and @alamb thank you 🙏🏻  for the reviews. I've completed the suggestions and recommendations from both of you.

Issue #6067, opened 2022-11-07
Feature request: support `influxql write` from stdin


alamb:
__Use case:__

While working on https://github.com/influxdata/influxdb_iox/issues/6066 I wanted to write a  small amount of data into iox via the command line

__Proposal:__
would have liked to write a command like the following to write a small amount of data into iox via the command line:

```shell
echo "table bar=2 0" | influxdb_iox write --host http://localhost:8080  my_db
```

__Current behaviour:__

I had to write to a file first 

```shell
echo "table foo=1,bar=1 0" > /tmp/foo1.lp
influxdb_iox write --host http://localhost:8080  my_db /tmp/foo1.lp
```

While stdin would be much nicer.

__Desired behaviour:__
Describe what you want.

__Alternatives considered:__

It might make sense to use the semi-standard `-` argument to explicitly request reading from stdin:

```shell
echo "table bar=2 0" | influxdb_iox write --host http://localhost:8080  my_db -
```

alamb:
FYI @aierui -- I am not sure if you are interested in small projects like this, but I think it would be a great way to contribute

PR #6167, opened 2022-11-17
chore: support write from stdin

aierui:
#6067 

PartI:  only support simple stdin.

Next will support file input pipeline stream.

### DIFF
--- a/influxdb_iox/src/commands/write.rs
+++ b/influxdb_iox/src/commands/write.rs
@@ -4,6 +4,7 @@ use observability_deps::tracing::info;
 use snafu::{ensure, OptionExt, ResultExt, Snafu};
 use std::{
     fs::File,
+    io,
     io::{BufReader, Read},
     num::NonZeroUsize,
     path::PathBuf,
@@ -91,6 +92,7 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
     // opened and if not fail fast.
     let file_open_errors: Vec<_> = file_names
         .iter()
+        .filter(|file_name| file_name != &&PathBuf::from("-"))
         .filter_map(|file_name| {
             File::open(file_name)
                 .context(ReadingFileSnafu { file_name })
@@ -194,6 +196,24 @@ async fn slurp_file(file_name: PathBuf) -> Result<String> {
         }
         // anything else, treat as line protocol
         Some(_) | None => {
+            if file_name == &PathBuf::from("-") {
+                let mut buffer = String::new();
+                let lines = io::stdin().lines();
+                for line in lines {
+                    buffer.push_str(line.unwrap().as_str());
+                    buffer.push_str("\n");
+                }
+                buffer = buffer.to_owned().strip_suffix("\n").unwrap().to_string();
+
+                info!(
+                    ?file_name,
+                    file_size_bytes = buffer.len(),
+                    "Buffered line protocol from stdin"
+                );
+
+                return Ok(buffer);
+            }
+
             let lp_data =
                 std::fs::read_to_string(file_name).context(ReadingFileSnafu { file_name })?;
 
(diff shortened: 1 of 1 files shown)


alamb:
I think the CI failure https://app.circleci.com/pipelines/github/influxdata/influxdb_iox/25911/workflows/e8b2d4fd-4fec-4fb0-8aa1-885a6eb2f5bc/jobs/229572 wasn't related to changes in this PR and I have restarted it

aierui:
Hi, @alamb 
Now the implementation is very tricky and require use the semi-standard `-` argument to explicitly.

```
echo "table bar=2 0" | influxdb_iox write --host http://localhost:8080  my_db -
```

Next I will split stdin from other input, support the import of large file, instead of read the data in the file and store in memory, and then write.

eg:
```
cat text.data | influxdb_iox write --host http://localhost:8080  my_db
```

alamb:
Thanks @aierui ! 

I tried it locally, and it worked!

```shell
$  echo "table bar=2 0" | ../target-iox/debug/influxdb_iox write --host http://localhost:8080  my_db -
13 Bytes OK in 29.500259ms. 0.00 MB/sec
$ ../target-iox/debug/influxdb_iox query my_db 'show tables'
+---------------+--------------------+-------------+------------+
| table_catalog | table_schema       | table_name  | table_type |
+---------------+--------------------+-------------+------------+
| public        | iox                | table       | BASE TABLE |
| public        | system             | queries     | BASE TABLE |
| public        | information_schema | tables      | VIEW       |
| public        | information_schema | views       | VIEW       |
| public        | information_schema | columns     | VIEW       |
| public        | information_schema | df_settings | VIEW       |
+---------------+--------------------+-------------+------------+
```


> Next I will split stdin from other input, support the import of large file, instead of read the data in the file and store in memory, and then write.

Sounds great -- I think streaming read for all inputs would be a great feature (rather than gathering them into a single large string to send).

alamb (commented):
Thanks again @aierui  !

I think this PR needs two things prior to merging:
1. A test (maybe we can follow the model here): https://github.com/influxdata/influxdb_iox/blob/4d55efe6558eb09d1ba03a8a5cbfcf48a3425c83/influxdb_iox/tests/end_to_end_cases/cli.rs#L397-L415
2. Update the documentation to mention the special `-` case: https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox/src/commands/write.rs#L66-L69

Do you think you have time to try that out? If not I can help.

influxdb_iox/src/commands/write.rs
```diff
@@ -91,6 +92,7 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
     // opened and if not fail fast.
     let file_open_errors: Vec<_> = file_names
         .iter()
+        .filter(|file_name| file_name != &&PathBuf::from("-"))
```
alamb, line 95: Another approach might be to check `file_names.empty()` and then call a function like `slurp_stdin`. That might be a little tricky to make work with the stream approach below

aierui: How to convert stdin to futures async stream? unified call [write_lp_stream](https://github.com/influxdata/influxdb_iox/blob/033d9ac7ee6e595397cd6b0f918a305364b0e1be/influxdb_iox/src/commands/write.rs#L138) function.



https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=befb059deb89af518669c5b0840ab385

I try, but I always encounter some strange error.

alamb: Perhaps you can use `tokio::io::stdin`: https://docs.rs/tokio/1.22.0/tokio/io/fn.stdin.html

alamb:
Marking as draft to signify this PR is not quite ready for merging

PR #6170, opened 2022-11-17
fix: Remove table names from DmlWrite

carols10cents:
Leaving only table IDs.

Connects to https://github.com/influxdata/influxdb_iox/issues/4880

Again there's certainly more deduplication especially in tests that could be done, but I think this is an improvement.


ingester/tests/common/mod.rs
```diff
@@ -230,17 +227,6 @@ impl TestContext {
             .expect("write must be sequenced")
             .sequence_number;
 
-        // Perform schema validation, populating the catalog.
```
domodwyer, line 233: Integration tests that involve persistence (of which there are currently none, admittedly!) will need the catalog to be populated (specifically the table's columns must exist). Ideally we keep this call in the tests because that accurately reflects the expected catalog state.

I am guessing the problem here is that the validation needs the table names and this fn only has the DmlWrite that doesn't know them? Currently the tests don't use this fn directly, but always call through `write_lp()` so we can probably collapse them and run the validation in `write_lp()` as it knows the names :+1:

mutable_batch_pb/src/decode.rs
```diff
@@ -57,24 +56,20 @@ pub enum Error {
...
     for table_batch in &database_batch.table_batches {
-        let (_, batch) = name_to_data
+        let (_, batch) = id_to_data
             .raw_entry_mut()
-            .from_key(table_batch.table_name.as_str())
-            .or_insert_with(|| (table_batch.table_name.clone(), MutableBatch::new()));
-
-        id_to_name.insert(table_batch.table_name.clone(), table_batch.table_id);
+            .from_key(&table_batch.table_id)
+            .or_insert_with(|| (table_batch.table_id, MutableBatch::new()));
```
domodwyer, line 67: I think this can be simplified to:

```rust
id_to_data.entry(table_batch.table_id).or_default()
```

Previously it used `from_key()` that takes a ref to avoid cloning the `String` key for each iteration, but now we use `Copy`-able IDs it's simpler/less code to avoid all that.

mutable_batch_pb/src/encode.rs
```diff
@@ -16,29 +16,16 @@ pub fn encode_write(database_id: i64, write: &DmlWrite) -> DatabaseBatch {
     DatabaseBatch {
         table_batches: write
             .tables()
-            .map(|(table_name, batch)| {
-                // Temporary code.
-                //
-                // Once only IDs are pushed over the network this extra lookup
-                // can be removed.
```
domodwyer, line 23: :wave:

router/tests/http.rs
```diff
@@ -179,7 +199,8 @@ async fn test_write_ok() {
     let writes = ctx.write_buffer_state().get_messages(ShardIndex::new(0));
     assert_eq!(writes.len(), 1);
     assert_matches!(writes.as_slice(), [Ok(DmlOperation::Write(w))] => {
-        assert!(w.table("platanos").is_some());
+        let table_id = ctx.table_id("bananas_test", "platanos").await;
+        assert!(w.table(&table_id).is_some());
```
domodwyer, line 203: Thank you for validating this 👍

write_buffer/src/codec.rs
```diff
@@ -179,8 +179,10 @@ pub fn decode(
 
                     Ok(DmlOperation::Write(DmlWrite::new(
                         NamespaceId::new(write.database_id),
-                        tables,
-                        ids.into_iter().map(|(k, v)| (k, TableId::new(v))).collect(),
+                        tables
+                            .into_iter()
+                            .map(|(k, v)| (TableId::new(k), v))
+                            .collect(),
```
domodwyer, line 185: I'm hoping this will get compiled out to nothing right? It's a type wrapper over a primitive, so it shouldn't actually need any conversion in the generated code. Not a problem (obviously!), just this caught my interest!

domodwyer (approved):
Thank you! I am very very very very very grateful for this 🙏

write_buffer/src/codec.rs
carols10cents: Yeah I think this should be fine. I haven't studied the ASM, and we can certainly measure if we run into performance problems, but this feels like it shouldn't be an issue because it's only a type wrapper, like you said.

domodwyer: Ah I am not worried about the perf at all, I was just curious if you new :D

ingester/tests/common/mod.rs
carols10cents: You're exactly right, I've fixed 👍🏻

Issue #6173, opened 2022-11-18
Improve router RPC write retry logic


domodwyer:
The current router RPC write to ingester uses an extremely simple retry mechanism (a for loop with a timeout).

This should be expanded to, at a minimum:

* Send a request to each ingester in turn, trying the full set of ingesters before looping around and trying the first again (avoid sending to the same ingester twice in succession)
* Record success/error metrics & RPC latency per ingester
* Include the ingester name/endpoint in the RPC error messages

Follow-on work should enhance this further to include:

* Circuit breaking ingesters to avoid returning a likely dead ingester connection to subsequent RPC write requests (and the associated recovery/probing logic)
  * Metrics for the circuit breaker state, probes, etc

domodwyer:
In order to fix https://github.com/influxdata/influxdb_iox/issues/6503 quickly, I chose to avoid using our custom gRPC connection wrappers, and instead use tonic directly, pushing load-balancing down to the gRPC handler (while keeping the round-robin sharder)

1. Evaluate what, if anything, the IOx wrappers over tonic Channels provides. Potentially remove.
2. Pull load-balancing back up out of the gRPC Channel, so that we have more flexibility in  building circuit breaking / latency/load based balancing strategies / log details of unreachable ingesters / etc.
3. Try to avoid lazily loading connections for the first writes a router handles, to eliminate the latency spike for those requests (P99 impacting)
4. Only bring a router on-line if at least one ingester is up? (ease of debugging, avoid invalid configurations)

domodwyer:
In https://github.com/influxdata/influxdb_iox/pull/6509 I added a horrible hack to work around what appears to be buggy handling of static endpoints in the upstream tonic/tower load balancing implementation (see https://github.com/influxdata/influxdb_iox/issues/6508)

By pulling the load-balancing back up and out of the gRPC layer, we should eliminate the need for this hack.

PR #6174, opened 2022-11-18
feat(router): handler for direct write to ingester

domodwyer:
This adds an [RPC write](#6145) client to the router, to push DML operations to ingesters instead of going through Kafka.

**Note**: this impl is currently unused.

---

* feat(router): handler for direct write to ingester (9da5b7963)

      This commit adds the (unused) RpcWrite implementation of the DmlHandler
      trait that implements pushing a write over gRPC to a single, arbitrary
      ingester. Requests are round-robin'ed across all available ingesters.
      
      This DmlHandler implementation can be swapped out with the
      ShardedWriteBuffer to change how writes are propagated to the ingester.

### DIFF
--- a/router/src/dml_handlers/rpc_write.rs
+++ b/router/src/dml_handlers/rpc_write.rs
@@ -90,21 +90,16 @@ where
         // Extract the partition key & DML writes.
         let (partition_key, writes) = writes.into_parts();
 
-        // Extract the TableName -> Data map and TableName -> TableId map.
-        let (data, ids) = writes
+        // Drop the table names from the value tuple.
+        let writes = writes
             .into_iter()
-            .map(|(id, (name, data))| {
-                let name_data = (name.clone(), data);
-                let name_id = (name, id);
-                (name_data, name_id)
-            })
-            .unzip();
+            .map(|(id, (_name, data))| (id, data))
+            .collect();
 
         // Build the DmlWrite
         let op = DmlWrite::new(
             namespace_id,
-            data,
-            ids,
+            writes,
             partition_key.clone(),
             DmlMeta::unsequenced(span_ctx.clone()),
         );
@@ -161,7 +156,7 @@ where
 
 #[cfg(test)]
 mod tests {
-    use std::sync::Arc;
+    use std::{collections::HashSet, sync::Arc};
 
     use assert_matches::assert_matches;
     use data_types::PartitionKey;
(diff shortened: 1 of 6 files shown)


router/src/dml_handlers/rpc_write.rs
```diff
@@ -0,0 +1,296 @@
...
+        // (#6173).
+        tokio::time::timeout(RPC_TIMEOUT, async {
+            loop {
+                match self.endpoints.next().write(req.clone()).await {
+                    Ok(()) => break,
+                    Err(e) => warn!(error=%e, "failed ingester rpc write"),
+                };
+            }
+        })
+        .await?;
```
domodwyer, line 129: This stupid simple retry loop needs improving in the future - I documented some desirable improvements in https://github.com/influxdata/influxdb_iox/issues/6173.

router/src/dml_handlers/rpc_write/client.rs
```diff
@@ -0,0 +1,64 @@
...
+use tonic::transport::Channel;
+
+use super::RpcWriteError;
+
+/// An abstract RPC client that pushes `op` to an opaque receiver.
+#[async_trait]
+pub(super) trait WriteClient: Send + Sync + std::fmt::Debug {
+    /// Write `op` and wait for a response.
+    async fn write(&self, op: WriteRequest) -> Result<(), RpcWriteError>;
+}
```
domodwyer, line 14: This trait allows the DmlHandler to delegate calls to an abstract "rpc client" - this lets us mock out the generated gRPC client for testing purposes.

lukebond (approved):

router/src/dml_handlers/rpc_write.rs
```diff
@@ -0,0 +1,296 @@
...
+                let name_id = (name, id);
+                (name_data, name_id)
+            })
+            .unzip();
+
+        // Build the DmlWrite
+        let op = DmlWrite::new(
+            namespace_id,
+            data,
+            ids,
```
carols10cents, line 107: 👋🏻 I just broke this by merging https://github.com/influxdata/influxdb_iox/pull/6170 😂

domodwyer: God damn, I was hoping to win this race 🔥 😂

router/src/dml_handlers/rpc_write.rs
carols10cents: As a tiny little improvement, should this loop back off rather than retrying as fast as possible? That is, should this use `backoff::Backoff`? Just to avoid continual hammering ingesters in the case they're failing because they're overloaded? That would be my first bet on how this would fail.

router/src/dml_handlers/rpc_write.rs
carols10cents: 🤣 but I made this code shorter!!!

router/src/dml_handlers/rpc_write.rs
carols10cents: Ah, i missed the little `self.endpoints.next()` bit! I agree this is fine for now.

Issue #6028, opened 2022-11-01
Filter data outside the retention period at query time


NGA-TRAN:
# Rationale

This is a task of #3053. 
Querier will filter data outside the retention period by adding this predicate in all queries
    time < now - retention_period_nano of the table

alamb:
I will take a shot at this later this week

alamb:
# High level proposal

My proposal is to model the retention period using a `DeletePredicate` added in the `querier` and then passed the `iox_query` crate

So if we have a retention of 1 week, the querier will effectively add a predicate like

```sql
where time < now() - interval '1 week'
```

To any chunk that contains rows outside the retention interval.

We can determine if a chunk is may contain data past the retention interval by checking the min/max stats for the time column.


# Implementation Sketch

For chunks that may straddle retention bounds, the querier will attach a  `DeletePredicate` like:

```rust
DeletePredicate {
  range: TimestampRange {
         start: now() - retention_period,
         end: now()
  },
  exprs: None,
}
```

I would likely make a wrapper of a QuerierChunk that has the retention interval / new delete pedicate and wrap any chunks provided by the querier here: https://github.com/influxdata/influxdb_iox/blob/main/querier/src/namespace/query_access.rs#L81

I would also need to record and remember the retenion interval from the catatlog on `QuerierNamespace`


The benefit of this design is that it follows the established pattern of delete predicates and so can resused most of the code and test infrastructure

# Alternatives considered:
None really


cc @crepererum  and @NGA-TRAN

NGA-TRAN:
Using `DelerePredicate` sounds good to me. And the Namespace now has a `retention_period_ns` field which should be a part of the `QuerierNamespace` as you suggested.

crepererum:
From a current technical PoV this makes sense, however I was hoping to get rid of the chunk-bound delete predicates altogether since I may have convinced someone that this is actually a batch job :wink: 

IIRC the query optimization will suffer HEAVILY from this as well, since chunks with delete predicates are NOT grouped at the moment (i.e. you'll get 1 `ParquetExec` per file). So w/o adding grouping or any other optimization to the query path, I don't think we should do this and rather insert an extra `FilterExec` node above the whole `UnionExec` group that we produce.

NGA-TRAN:
One more thing we may want to consider is we can/want to delete this data before deduplication. Chunk-level filter will enable that

crepererum:
[...]

PR #6180, opened 2022-11-18
feat: Prune chunks fully outside retention period in querier

NGA-TRAN:
This is the PR implement the first half of #6028  that prunes chunks fully outside of the retention period.
My next PR will apply predicate to remaining chunks to filter specific rows that are outside the retention period


alamb (approved):
I think this looks great -- thank you @NGA-TRAN 

It would be great if @crepererum looked at this as well prior to merge as he is the querier expert.

querier/src/cache/namespace.rs
```diff
@@ -243,6 +243,7 @@ impl From<TableSchema> for CachedTable {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CachedNamespace {
     pub id: NamespaceId,
+    pub retention_period_ns: Option<i64>,
```
crepererum, line 246: Can we type this?

```suggestion
    pub retention_period: Option<Duration>,
```

querier/src/table/mod.rs
```diff
@@ -197,10 +203,28 @@ impl QuerierTable {
...
+                    - retention_period_ns;
+                let predicate = predicate.clone().with_retention(retention_ns);
+                self.chunks_inner(&predicate, &span_recorder, projection)
+                    .await
+            }
+            None => {
+                self.chunks_inner(predicate, &span_recorder, projection)
+                    .await
+            }
+        };
```
crepererum, line 225: This should go to `chunks_inner`. The only reason `chunks` is split into `chunks` and `chunks_inner` is the span scope and error propagation. All actual logic lives in `chunks_inner`.

querier/src/namespace/query_access.rs
```diff
@@ -588,6 +590,110 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_prune_chunks_outside_retention() {
+        test_helpers::maybe_start_logging();
+
+        let catalog = TestCatalog::new();
+
+        // namespace with 1-hour retention policy
+        let ns = catalog.create_namespace_1hr_retention("ns").await;
+        let inside_retention = catalog.time_provider.now().timestamp_nanos(); // now
+        let outside_retention = inside_retention - 2 * 60 * 60 * 1_000_000_000; // 2 hours ago
```
crepererum, line 602: If the retention would be properly typed (also see my comment in the cache section), we wouldn't need this arcane math here.

querier/src/namespace/query_access.rs
```diff
@@ -588,6 +590,110 @@ mod tests {
             .await;
     }
 
+    #[tokio::test]
```
crepererum, line 593: I would have expected a `query_test` (including `EXPLAIN`) for this. I'm also bringing this up because now you're adding filter predicates to EVERY chunk (in contrast to what your PR description says it does), even the ones that are fully within the retention window. This will come with some major overhead.

The fact that this also filters our entire chunks early is a mere side effect (this will work with every time-based predicate).

crepererum (changes requested):

querier/src/table/mod.rs
```diff
@@ -225,6 +232,26 @@ impl QuerierTable {
...
+                    - retention_period.as_nanos() as i64;
+                Some(predicate.clone().with_retention(retention_time_ns))
+            }
+            None => None,
+        };
+        let predicate = if predicate_with_retention.is_some() {
+            predicate_with_retention.as_ref().unwrap()
+        } else {
+            predicate
+        };
```
alamb, line 253:
```suggestion
        let predicate = match self.namespace_retention_period {
            // Add expression to only keep chunks inside the retention period: time >= now - retention_period
            Some(retention_period) => {
                let retention_time_ns = self
                    .chunk_adapter
                    .catalog_cache()
                    .time_provider()
                    .now()
                    .timestamp_nanos()
                    - retention_period.as_nanos() as i64;
                predicate.clone().with_retention(retention_time_ns)
            }
            None => predicate,
        };
```

querier/src/namespace/query_access.rs
NGA-TRAN: @crepererum  I have added the 2 explains for this test. Are they what you expected here or somewhere else?

Questions: 
1. Does the `chunks_inner`do early chunk pruning and decide which chunks to read? I think that is what I meant for this PR. Sorry for the confusion. 
2. I also see the `chunks_inner` asks Ingester for data which will also now filter data outside of the retention period because we add the retention predicate, right?
3. Do you think this is the right approach?

predicate/src/lib.rs
```diff
@@ -442,6 +443,14 @@ impl Predicate {
         self
     }
 
+    /// Add an  exprestion "time >= retention_time"
+    pub fn with_retention(mut self, retention_time: i64) -> Self {
+        let time = ScalarValue::TimestampNanosecond(Some(retention_time), None);
```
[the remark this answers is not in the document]
NGA-TRAN: Thanks.

querier/src/namespace/query_access.rs
```diff
@@ -588,6 +590,110 @@ mod tests {
...
+            .with_max_seq(2)
+            .with_min_time(inside_retention)
+            .with_max_time(inside_retention);
+        partition.create_parquet_file(builder).await;
+
+        // C3: fully outside retention
+        let lp = format!(
+            "
+            cpu,host=z load=0 {}\n
+            cpu,host=zz load=01 {}\n
```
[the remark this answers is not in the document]
NGA-TRAN: Done

querier/src/cache/namespace.rs
NGA-TRAN: Done

querier/src/table/mod.rs
NGA-TRAN: Done

querier/src/namespace/query_access.rs
crepererum: > I have added the 2 explains for this test. Are they what you expected here or somewhere else?

I was more aiming for tests in the `query_test` crate since there we already have loads of infrastructure instead of spraying the tests over the whole repo.


> 1. Does the `chunks_inner`do early chunk pruning and decide which chunks to read? I think that is what I meant for this PR. Sorry for the confusion.

It does. :+1: 


> 2. I also see the `chunks_inner` asks Ingester for data which will also now filter data outside of the retention period because we add the retention predicate, right?

Yes. I think this is correct for consistency reasons. I would be more worried if we would use different mechanisms for parquet and ingester data.


> 3. Do you think this is the right approach?

TBH I'm slightly worried about the execution-time overhead, but I think it's in the same as `select ... from ... where time > X`, so this is OK. Let's walk this route for now and let's rather invest into some predicate-pruning later where we can remove time-based predicates from chunks where the outcome is clear (e.g. chunks which are completely within the requested period and hence don't need any row-based filtering).

crepererum:
I really would like to see the query tests within the `query_test` crate instead of in `querier`. I leave it up to @NGA-TRAN if she wants to do this in this PR here or in a follow-up PR.

crepererum (approved):

querier/src/namespace/query_access.rs
NGA-TRAN: Thanks @crepererum. I will move the tests in the follow-up PR

alamb:
I believe the code in this PR has been merged under the aegis of other PRs

NGA-TRAN:
Close this as it was merged in #6198

PR #6188, opened 2022-11-21
refactor: move DmlSink

domodwyer:
Nothing crazy - this just moves the `DmlSink` trait out of the Kafka `stream_handler` mod.

---

* refactor: move DmlSink (8ce48b9b9)

      Extracts the DmlSink trait into its own module - it is independent of
      the Kafka handler and will be reused.


crepererum (approved):

ingester/src/dml_sink/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod r#trait;
```
crepererum, line 1: maybe call it `interface` instead of using a reserved keyword.

SimonRichardson (approved):
Took me a quick second to get the `mod r#trait`, but otherwise seems reasonable.

PR #6189, opened 2022-11-21
refactor: decouple DmlSink error type

domodwyer:
Allow the `DmlSink` implementation to specify what error type it returns, requiring that error type to be convertable into a `DmlError`.

This allows concise impl-specific error types, but still allow the API handlers that interact with one or more `DmlSink` to convert the impl error into a type they can enumerate and map to a suitable response. 

---

* refactor: decouple DmlSink error type (1938c18c5)

      Allows different DmlSink implementations to return different error
      types. This allows for small, concise errors that are local to the
      DmlSink implementation and specific to it. This helps avoid bloated
      "kitchen sink" error types.

### DIFF
--- a/ingester/src/dml_sink/mock_sink.rs
+++ b/ingester/src/dml_sink/mock_sink.rs
@@ -4,13 +4,13 @@ use async_trait::async_trait;
 use dml::DmlOperation;
 use parking_lot::Mutex;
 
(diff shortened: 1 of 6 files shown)


SimonRichardson (approved):
:+1: on targeted error messages.

PR #6192, opened 2022-11-21
feat: warm compaction

### LINKED ISSUE #6179: Proposal: "warm" mode for compactor
# Background

Currently the compactor has two modes:

1. "hot" - finds partitions with recently-written L0 files and compacts as many of those files L1 files that it can do in the memory budget
2. "cold" - finds partitions without recent writes, compacts any remaining L0 files into L1 files, and then will compact those L1 files into L2

Currently the system is configured to run four hot compaction cycles and then one cold compaction cycle.

# Hypothesis

This compaction model is leaving around a lot of small L1 files that could be combined which would improve query performance. They are too "warm" to be picked up by the cold compaction, and they don't necessarily overlap hence won't be picked up by hot compaction.

# Proposal

Introduce a new "warm" mode that looks for small [^1], adjacent L1 files that, if combined, would be less than `large_cutoff_bytes`

We could adjust the compaction schedule to do something like:

1. 1x hot compaction cycle
2. 1x warm compaction cycle
3. 2x hot compaction cycle
4. 1x cold compaction cycle

...the theory being that it's important to combine these to improve query performance, not so important as hot compaction, but maybe more important than the last round or two of hot compaction.

---
[^1]: small is defined as between `small_cutoff_bytes` and `(large_cutoff_bytes - small_cutoff_bytes) / 2`

lukebond:
Closes #6179

This is a work in progress, I don't even know if it works yet. Writing tests now. Raised the PR for interested parties to get an early look.


iox_catalog/src/postgres.rs
```diff
@@ -1906,6 +1906,43 @@ LIMIT $3;
...
+        min_small_file_count: usize,
+        num_partitions: usize,
+    ) -> Result<Vec<PartitionParam>> {
+        // This query returns partitions with at least `min_small_file_count` small L1 files,
+        // where "small" means no bigger than `small_size_threshold_bytes`, limited to the top `num_partitions`.
+        sqlx::query_as::<_, PartitionParam>(
+            r#"
+SELECT parquet_file.partition_id, parquet_file.shard_id, parquet_file.namespace_id,
+       parquet_file.table_id,
+       COUNT(*) AS l1_file_count,
```
NGA-TRAN, line 1922:
```suggestion
       COUNT(1) AS l1_file_count,
```
We are not clear what Postgres does with `count(*)` but `count(1)` is a count on a constant value 1 which we think will do the minimum work

NGA-TRAN (commented):
The draft looks good, @lukebond. I think the tests are the fun part. Hope the available tests will provide you some idea. The most tricky part is to create them in the right time ranges and sequence number plus good combination of L0, 1 and L2

clap_blocks/src/compactor.rs
```diff
@@ -120,6 +120,19 @@ macro_rules! gen_compactor_config {
...
+            /// partitions equally.
+            ///
+            /// Default is 1
+            #[clap(
+                long = "compaction-warm-multiple",
+                env = "INFLUXDB_IOX_COMPACTION_WARM_MULTIPLE",
+                default_value = "1",
+                action
+            )]
+            pub warm_multiple: usize,
```
lukebond, line 134: one of the goals in https://github.com/influxdata/influxdb_iox/issues/6179 was to make the scheduling more configurable. this doesn't really meet that goal but given that [this issue came up](https://github.com/influxdata/influxdb_iox/issues/6194) i figure it's best kicked down the road to when we do that.

compactor/src/warm.rs
```diff
@@ -0,0 +1,588 @@
...
+
+        assert_eq!(candidates[0].id(), partition6.partition.id);
+        // this sort key is None
+        assert_eq!(candidates[0].sort_key, partition6.partition.sort_key());
+
+        assert_eq!(candidates[1].id(), partition7.partition.id);
+        // this sort key is Some(tag1, time)
+        assert_eq!(candidates[1].sort_key, partition7.partition.sort_key());
+    }
+}
```
NGA-TRAN, line 588: Very nice tests here. I think we want to add tests to verify the compacting output, too.  With my comment above that the warm compaction  need to compacts adjacent L1s to fewer but larger L1s, we need the output verifying tests

NGA-TRAN (commented):
It is very nice @lukebond but I just realized warm compaction only compacts L1s and output are also L1s. This  means we need to refactor `compact_candidates_with_memory_budget` to do so.

compactor/src/warm.rs
```diff
@@ -0,0 +1,775 @@
...
+                "| 61    | greyhound | 1970-01-01T00:00:00.000000106Z |",
+                "| 62    | greyhound | 1970-01-01T00:00:00.000000104Z |",
+                "| 63    | greyhound | 1970-01-01T00:00:00.000000102Z |",
+                "| 64    | greyhound | 1970-01-01T00:00:00.000000100Z |",
+                "+-------+-----------+--------------------------------+",
+            ],
+            &batches
+        );
+    }
+}
```
NGA-TRAN, line 775: This is very nice test, Luke. I think if you want better coverage, you can add one L0 and two L2, one does not overlap with any L1 and one overlaps, in this test. and verify L0 and non-overlapped L2 stay. The overlapped L2 will get compacted with the L1.

You may want another test to have the compacted result split into 2 files

NGA-TRAN (approved):
Looks good. I suggest 2 more tests. I know you also plan to add small, big an don't contiguous files

compactor/src/warm.rs
```diff
@@ -0,0 +1,775 @@
...
+            .unwrap();
+        candidates.sort_by_key(|c| c.candidate);
+        assert_eq!(candidates.len(), 2);
+
+        assert_eq!(candidates[0].id(), partition6.partition.id);
+        // this sort key is None
+        assert_eq!(candidates[0].sort_key, partition6.partition.sort_key());
+
+        assert_eq!(candidates[1].id(), partition7.partition.id);
+        // this sort key is Some(tag1, time)
```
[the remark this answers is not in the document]
lukebond: should we make it impossible to create a partition without a sort key if it's required?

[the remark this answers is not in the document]
NGA-TRAN: I do not remember exactly but it may just be this test setup that requires that.
For real data, I think the partition has to be created before data is loaded which means no data to determine sort key at creation time and it has to be empty.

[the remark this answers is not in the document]
lukebond: have added it, copying from hot compaction tests. i must admit i don't fully understand why

compactor/src/lib.rs
```diff
@@ -121,9 +123,17 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
...
-                let (level_n, level_n_plus_1) = match initial_level {
-                    CompactionLevel::Initial => (level_0, level_1),
-                    CompactionLevel::FileNonOverlapped => (level_1, level_2),
+                let (level_n, level_n_plus_1) = match (initial_level, target_level) {
+                    (CompactionLevel::Initial, CompactionLevel::FileNonOverlapped) => {
+                        (level_0, level_1)
+                    }
+                    (CompactionLevel::FileNonOverlapped, CompactionLevel::FileNonOverlapped) => {
+                        // warm compaction; don't try to compact L2 files into L1
+                        (level_1, vec![])
```
lukebond, line 132: this change prevents the compactor from trying to compact L2s into L1, which will cause a panic.

compactor/src/lib.rs
```diff
@@ -53,6 +54,7 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
     compactor: Arc<Compactor>,
     compaction_type: &'static str,
     initial_level: CompactionLevel,
+    target_level: CompactionLevel,
```
lukebond, line 57: previously this function assumed target level was initial level + 1; but for warm compaction it's the same level, so this parameter was added

compactor/src/warm.rs
```diff
@@ -0,0 +1,1320 @@
...
+                },
+                Timestamp::new(0),
+                Timestamp::new(100),
+            )
+            .await;
+        assert_eq!(l1_count, 4);
+        let files = catalog.list_by_table_not_to_delete(table.table.id).await;
+        let total_count = files.len();
+        // by deduction this is l2 count (wasn't worth adding a count fn to the catalog!)
+        assert_eq!(total_count - l1_count, 2);
```
NGA-TRAN, line 1279: 👍

compactor/src/warm.rs
```diff
@@ -0,0 +1,1320 @@
...
+            partition_candidates.into(),
+        )
+        .await;
+
+        // this compaction outputs three files: the two L2 files are left alone, regardless of whether
+        // or not they overlap with L1 files, and the remaining L1s are compacted into one file.
+        // the important thing for us here is that it didn't do anything with the L2's- the current
+        // compactor algorithm won't compact L2s down into L1s. the L2 that overlaps with L1 will
+        // be dealt with in a round of cold compaction later.
+        let files = catalog.list_by_table_not_to_delete(table.table.id).await;
```
NGA-TRAN, line 1305: 👍

NGA-TRAN (commented):
This looks really good @lukebond . Thanks for all the details.

The last tiny piece is we want to ensure everything works as expected by having the test `test_compact_small_l1_files_becomes_two_l1s` to have deterministic number of output files and verify their data. I think your test is already deterministic, you just need to verify data of the compacted files

lukebond:
> @lukebond : do you plan to merge this PR or you want to add something?

i was working on something else for a while, coming back to fix this up and merge today

lukebond:
@NGA-TRAN if you're happy with this test now we can merge it

compactor/src/warm.rs
```diff
@@ -881,7 +882,9 @@ mod tests {
         // max file size is small enough for the third, larger file to not qualify.
         // (note that the small file threshold is only used for filtering candidate partitions, not
         // for the actual compaction)
-        let config = make_compactor_config(4_000, 5_000, 4);
+        // since we have such a small max file size, to make the tests easier to write, bump up the
+        // percentage value so it doesn't try to split everything really small
+        let config = make_compactor_config(4_000, 10_000, 4, 90);
```
NGA-TRAN, line 887: 👍

NGA-TRAN (approved):
Thanks Luke. The tests look good

PR #6197, opened 2022-11-21
feat: Apply retention delete predicate to eliminate data outside retention

### LINKED ISSUE #6028: Filter data outside the retention period at query time

NGA-TRAN:
Closes #6028 
This is the second and last PR for #6028 to filter rows on remaining chunks


querier/src/table/mod.rs
```diff
@@ -225,6 +234,38 @@ impl QuerierTable {
             "Fetching all chunks"
         );
 
+        let (predicate_with_retention, retention_delete_pred) =
```
alamb, line 237: I just doubled checked what @crepererum  mentioned this on a previous version of this PR. I think  updating `predicate` in this code will effectively will apply the retention predicate to data queried from the ingester as well but not to all the actual chunk scans, is that right?

querier/src/table/mod.rs
```diff
@@ -225,6 +234,38 @@ impl QuerierTable {
...
+                    let retention_delete_pred = Some(DeletePredicate::retention_delete_predicate(
+                        retention_time_ns,
+                    ));
+
+                    (predicate, retention_delete_pred)
+                }
+                // inifite retention, no need to add predicate
+                None => (None, None),
+            };
+        let predicate = if predicate_with_retention.is_some() {
```
alamb, line 263: I think you can make some of this code simpler by directly passing `predicate` back

```rust
        let (predicate, retention_delete_pred) =
...


                    (predicate, retention_delete_pred)
                }
                // inifite retention, keep original predicate
                None => (predicate, None),
        };
```

querier/src/namespace/query_access.rs
```diff
@@ -588,6 +592,168 @@ mod tests {
...
+                "| a    | 1    | 1970-01-01T00:00:00Z |",
+                "| b    | 2    | 1970-01-01T00:00:00Z |",
+                "| bb   | 21   | 1970-01-01T00:00:00Z |",
+                "+------+------+----------------------+",
+            ],
+            Some(span_ctx.clone()),
+        )
+        .await;
+
+        // Should see only see 2 chunks (2 ParquetExecs)
```
alamb, line 682:
```suggestion
        // Should see only see 2 chunks (2 ParquetExecs) with retention filters
```

alamb (approved):
Thank you @NGA-TRAN  -- I went over the PR carefully and it makes sense to me.

I think the order of operations would be merge https://github.com/influxdata/influxdb_iox/pull/6198 (into this PR) and then merge this PR into main.

I also suggest a follow on PR to push the retention filters into the ParquetExec scans.

crepererum:
How's this different from #6180? I'm confused.

NGA-TRAN:
@crepererum : hah, I should have closed this yesterday. I created this for second task but then thought it would be difficult to review so I created different PR to compare it with previous branch

NGA-TRAN:
Close this as the work is merged in https://github.com/influxdata/influxdb_iox/pull/6209

querier/src/table/mod.rs
NGA-TRAN: @alamb The `predicate_with_retention` will be applied to data from ingester and will be used to early prune chunks in querier.  `retention_delete_pred` will be applied to the chunks we scan after the chunk pruning

querier/src/table/mod.rs
alamb: Implemented in https://github.com/influxdata/influxdb_iox/pull/6211

Issue #1133, opened 2021-04-06
Large volumes of data being transferred using `influxdb_iox database query` command


alamb:
Broken off from https://github.com/influxdata/influxdb_iox/issues/1132

__Steps to reproduce:__
@mkmik  noticed when running the following command:

```
$ while true; do ./target/debug/influxdb_iox --host http://localhost:1234  database query 844910ece80be8bc_7be09b71c487d5d3 "select * from http_api_requests_total limit 10;"; done
```

there was 3Mbps of data while being downloaded while doing. Wireshark shows a conversation of 22Mb:

![image](https://user-images.githubusercontent.com/490673/113778020-dfc22d80-96f9-11eb-8e72-0ce7acdb03e2.png)

When the results displayed to the user are on the order of 1K

__Expected behaviour:__
Traffic size in proportion to the size of the results returned

__Actual behaviour:__
22MB is several orders of magnitude greater than the results returned

mkmik:
the wireshark gRPC dissector works well once you add the .proto files. However Apache Flight is all 'bytes' containing flatbuffer payloads, which can be decoded with `flatc -t`.

Turns out `select * from http_api_requests_total limit 10;` returns a single big recordbatch response of 22Mb.

While `select nodename from http_api_requests_total limit 1;` returns a single big recordbatch resposne of 3Mb.

Passing the `data_header` payload to `flatc -t` we can see:

```
{
  version: "V5",
  header_type: "RecordBatch",
  header: {
    length: 1,
    nodes: [
      {
        length: 1,
        null_count: 0
      }
    ],
    buffers: [
      {
        offset: 0,
        length: 8752
      },
      {
        offset: 8752,
        length: 280008
      },
      {
        offset: 288760,
        length: 2848784
      }
    ]
  },
  bodyLength: 3137544
}
```

by changing the limit, the record batch `length` field changes, but the size of the overall body stays about 3Mb.

Once the `limit` crosses about 70000, we start receiving new result batches, each about 3mb in size, no matter what's the length of the actual record batch. Some printf debugging:

```
Received response:
data_header.len = 144
data_body.len = 3131920
batch = RecordBatch { length: 70000, nodes: Some([FieldNode { length: 70000, null_count: 0 }]), buffers: Some([Buffer { offset: 0, length: 8752 }, Buffer { offset: 8752, length: 280008 }, Buffer { offset: 288760, length: 2843160 }]), compression: None }

Received response:

data_header.len = 144
[...]

PR #6205, opened 2022-11-22
fix: slice flight response batches

crepererum:
Same as #6094 but for the Apache Flight interface.

Ref https://github.com/influxdata/idpe/issues/16073.


arrow_util/src/optimize.rs
```diff
@@ -186,15 +186,63 @@ pub fn optimize_schema(schema: &Schema) -> Schema {
...
+        + usize::from(size % MAX_GRPC_RESPONSE_BATCH_SIZE != 0))
+    .max(1);
+    let rows_per_batch = batch.num_rows() / n_batches;
+    let mut out = Vec::with_capacity(n_batches + 1);
+
+    let mut offset = 0;
+    while offset < batch.num_rows() {
+        let length = (offset + rows_per_batch).min(batch.num_rows() - offset);
+        out.push(
+            RecordBatch::try_new(
```
alamb, line 219: I think you can use RecordBatch::slice here  https://docs.rs/arrow/27.0.0/arrow/record_batch/struct.RecordBatch.html#method.slice

arrow_util/src/optimize.rs
```diff
@@ -501,4 +549,29 @@ mod tests {
...
+        // no split
+        let c = UInt32Array::from(vec![1, 2, 3, 4, 5, 6]);
+        let batch = RecordBatch::try_from_iter(vec![("a", Arc::new(c) as ArrayRef)])
+            .expect("cannot create record batch");
+        let split = split_batch_for_grpc_response(batch.clone());
+        assert_eq!(split.len(), 1);
+        assert_eq!(batch, split[0]);
+
+        // split once
+        let n_rows = MAX_GRPC_RESPONSE_BATCH_SIZE + 1;
```
alamb, line 564: I think the test would be clearer if you hard coded the number of rows here -- it is somewhat confusing that a batch size of bytes is used as the number of rows in a test

ingester/src/querier_handler.rs
```diff
@@ -142,11 +142,22 @@ impl IngesterQueryResponse {
...
                                 });
 
-                                let tail = snapshot.map(move |batch_res| match batch_res {
-                                    Ok(batch) => Ok(FlatIngesterQueryResponse::RecordBatch {
-                                        batch: optimize_record_batch(&batch, Arc::clone(&schema))?,
-                                    }),
-                                    Err(e) => Err(e),
+                                let tail = snapshot.flat_map(move |batch_res| match batch_res {
+                                    Ok(batch) => {
+                                        match optimize_record_batch(&batch, Arc::clone(&schema)) {
```
alamb, line 147: Seeing this code I wonder if optimize_record_batch  is still needed 
https://github.com/influxdata/influxdb_iox/blob/caa595a6fcf06fd33e8b50edcbc1b3243643c2ba/arrow_util/src/optimize.rs#L118

-- it seems as if the upstream issues are fixed: https://github.com/influxdata/influxdb_iox/issues/1133 https://github.com/apache/arrow-rs/issues/208

alamb: (We could do as a follow on PR)

alamb (approved):
LGTM

mkmik (approved):

arrow_util/src/optimize.rs
crepererum: done

arrow_util/src/optimize.rs
alamb: I missed the point that num rows ~ num bytes because it is making a `UInt8Array` 👍  sorry for the confusion

```suggestion
        // Choose n_rows ~ n_bytes by creating an Int8Array
        let n_rows = MAX_GRPC_RESPONSE_BATCH_SIZE + 1;
```

arrow_util/src/optimize.rs
```diff
@@ -186,15 +186,53 @@ pub fn optimize_schema(schema: &Schema) -> Schema {
...
+    let n_batches = (size / MAX_GRPC_RESPONSE_BATCH_SIZE
+        + usize::from(size % MAX_GRPC_RESPONSE_BATCH_SIZE != 0))
+    .max(1);
+    let rows_per_batch = batch.num_rows() / n_batches;
+    let mut out = Vec::with_capacity(n_batches + 1);
+
+    let mut offset = 0;
+    while offset < batch.num_rows() {
+        let length = (offset + rows_per_batch).min(batch.num_rows() - offset);
+        out.push(batch.slice(offset, length));
```
alamb, line 218: ❤️

PR #6206, opened 2022-11-22
refactor(ingester): split out BufferTree from Shard

domodwyer:
This **looks** big, but it isn't, I promise!

Previously a [`Shard`] contained the set of `NamespaceData` within it (which in turn holds the map of `TableData`, and so on). This PR splits the data-carrying portion of the `Shard` into a new type called a `BufferTree` (because... it's a tree and it buffers :+1:)

Almost all of this diff is simply moving files - I will highlight the important changes/new code below.

[`Shard`]:https://github.com/influxdata/influxdb_iox/blob/e4c12fa6a5e9b607673d545f0a4bce4232e6a2f5/ingester/src/data/shard.rs#L20-L52

---

* refactor: decouple Shard & BufferTree (ee8b728c3)

      Splits out the nested tree of namespace -> tables -> partitions
      (referred to as the "buffer tree") from the Shard which previously held
      the namespace map.
      
      This allows the BufferTree to exist without a shard, or many trees to
      exist within a shard, etc.

* style: consistent import ordering (9eafa9dbe)

      Reorder all imports in the ingester to match a consistent order:
      
          * stdlib
          * external crates
          * intra-crate imports
      
      This helps prevent merge conflicts & keeps everything tidy.


ingester/src/buffer_tree/root.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use super::{
+    namespace::{name_resolver::NamespaceNameProvider, NamespaceData},
+    partition::resolver::PartitionProvider,
+    table::name_resolver::TableNameProvider,
+};
+use crate::{arcmap::ArcMap, data::DmlApplyAction, lifecycle::LifecycleHandle};
+
+#[derive(Debug)]
+pub(crate) struct BufferTree {
```
domodwyer, line 16: This file is new, and defines the `BufferTree` type.

The fields in a `BufferTree`, and the logic in the impl below, are simply moved from the [`Shard`](https://github.com/influxdata/influxdb_iox/blob/e4c12fa6a5e9b607673d545f0a4bce4232e6a2f5/ingester/src/data/shard.rs#L20-L52) without any changes.

ingester/src/data/shard.rs
```diff
@@ -4,51 +4,25 @@ use std::sync::Arc;
...
-    namespace_name_resolver: Arc<dyn NamespaceNameProvider>,
-    /// The [`TableName`] provider used by [`NamespaceData`] to initialise a
-    /// [`TableData`].
-    ///
-    /// [`TableName`]: crate::data::table::TableName
-    /// [`TableData`]: crate::data::table::TableData
-    table_name_resolver: Arc<dyn TableNameProvider>,
-
-    metrics: Arc<metric::Registry>,
-    namespace_count: U64Counter,
```
domodwyer, line 51: These fields are now part of the `BufferTree`, and this `Shard` now contains a `BufferTree` instance instead.

ingester/src/data/shard.rs
```diff
@@ -86,42 +55,19 @@ impl ShardData {
...
-                self.namespace_name_resolver.for_namespace(namespace_id),
-                Arc::clone(&self.table_name_resolver),
-                self.shard_id,
-                Arc::clone(&self.partition_provider),
-                &self.metrics,
-            ))
-        });
-
-        namespace_data
+        self.buffer_tree
```
domodwyer, line 58: All the methods on the `Shard` now delegate the actual work to the `BufferTree` within it.

SimonRichardson (approved):

ingester/src/buffer_tree/root.rs
```diff
@@ -0,0 +1,121 @@
...
+                self.shard_id,
+                Arc::clone(&self.partition_provider),
+                &self.metrics,
+            ))
+        });
+
+        namespace_data
+            .buffer_operation(dml_operation, lifecycle_handle)
+            .await
+    }
```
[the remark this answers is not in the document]
domodwyer: It's a good question, and yes, you're absolutely correct :)

PR #6212, opened 2022-11-22
fix: ignore fields when considering tag predicates

wolffcm:
Closes influxdata/idpe#16238

This PR makes a change to the `MissingColumnRewriter` used to transform Influx RPC predicates into equivalent SQL predicates. It turns out that any column references that are not `_value` must be tags according to Influx RPC conventions. The change here is to treat a column as missing (transform it to a `NULL`) even if it exists but is not a tag.


wolffcm:
Hmm, I thought I ran all the tests locally, investigating the failures now.

alamb (commented):
Looking good 👍 


I ran the tests that were failing like this to get more detailed logging:
```
cd /Users/alamb/Software/influxdb_iox2 && RUST_LOG=trace RUST_BACKTRACE=1 CARGO_TARGET_DIR=/Users/alamb/Software/target-iox2 cargo test -p iox_query -- test_predicate_read_group
```

And the logs show: 
```
2022-11-22T22:03:25.820375Z DEBUG iox_query::frontend::influxrpc: planning read_group rpc_predicate=InfluxRpcPredicate { table_names: None, inner: Predicate { field_columns: None, range: None, exprs: [(_measurement = Utf8("foo") OR _measurement = Utf8("h2o")) AND time > TimestampNanosecond(5, None)], value_expr: [] } } agg=None
2022-11-22T22:03:25.820461Z DEBUG predicate::rpc_predicate: rewriting expr e=(_measurement = Utf8("foo") OR _measurement = Utf8("h2o")) AND time > TimestampNanosecond(5, None)
2022-11-22T22:03:25.820484Z TRACE predicate::rpc_predicate: After rewrite expr=(Utf8("h2o") = Utf8("foo") OR Utf8("h2o") = Utf8("h2o")) AND time > TimestampNanosecond(5, None) description=rewrite_measurement_references
2022-11-22T22:03:25.820503Z TRACE predicate::rpc_predicate: After rewrite expr=(Utf8("h2o") = Utf8("foo") OR Utf8("h2o") = Utf8("h2o")) AND time > TimestampNanosecond(5, None) description=rewrite_field_value_references
2022-11-22T22:03:25.820523Z TRACE predicate::rpc_predicate: After rewrite expr=(Utf8("h2o") = Utf8("foo") OR Utf8("h2o") = Utf8("h2o")) AND time > TimestampNanosecond(5, None) description=field_projections
2022-11-22T22:03:25.820544Z TRACE predicate::rpc_predicate: After rewrite expr=(Utf8("h2o") = Utf8("foo") OR Utf8("h2o") = Utf8("h2o")) AND Utf8(NULL) > TimestampNanosecond(5, None) description=missing_columums
2022-11-22T22:03:25.820562Z TRACE predicate::rpc_predicate: After rewrite expr=(Utf8("h2o") = Utf8("foo") OR Utf8("h2o") = Utf8("h2o")) AND Utf8(NULL) > TimestampNanosecond(5, None) description=rewrite
2022-11-22T22:03:25.820598Z DEBUG predicate::rpc_predicate: rewritten expr e=Err(Plan("'Utf8 > Timestamp(Nanosecond, None)' can't be evaluated because there isn't a common type to coerce the types to"))
thread 'frontend::influxrpc::tests::test_predicate_read_group' panicked at 'creating plan: CreatingPredicates { source: Plan("'Utf8 > Timestamp(Nanosecond, None)' can't be evaluated because there isn't a common type to coerce the types to") }', iox_query/src/frontend/influxrpc.rs:2424:22
stack backtrace:
```

It looks to me like the `time` column is being rewritten to null --

```
AND Utf8(NULL) > TimestampNanosecond(5, None)
```

I bet if you stopped that the test would pass 🤔

predicate/src/rpc_predicate/column_rewrite.rs
```diff
@@ -4,30 +4,32 @@ use datafusion::{
...
         // todo a real error here (rpc_predicates shouldn't have table/relation qualifiers)
         assert!(col.relation.is_none());
 
-        if self.schema.find_index_of(&col.name).is_some() {
-            Ok(true)
-        } else {
-            Ok(false)
+        match self.schema.find_index_of(&col.name) {
+            Some(i) => match self.schema.field(i).0 {
+                InfluxColumnType::Tag => Ok(true),
```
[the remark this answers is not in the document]
alamb: BTW you can write this in a more functional style if you are into that sort of thing

Something like

```rust
        self.schema.find_index_of(&col.name)
            .map(|i| self.schema.field(i).0)
            .map(|influx_column_type| {
                matches!(influx_column_type, InfluxColumnType::Tag | InfluxColumnType::Timestamp)
            })
            .unwrap_or(false)
```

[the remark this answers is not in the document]
wolffcm: The thing is, there will never actually be a reference to the time column coming in from Influx RPC. There is no way to express it, since it is neither a field nor a tag. Any restrictions on `time` in the results are expressed on the `range`.

My plan was to update the test to use a different predicate, and still show that it can be simplified. But now I am wondering if I should relax this function. Do you have a preference?

wolffcm:
@alamb I've updated the branch to use a more realistic predicate for the test that was failing. Influx RPC remains frustrating to me in its design. I am open to the change you suggested as well, but I think the current behavior is more consistent with TSM. Thank you for the thorough review and the pointers!

There are still other tests failing though, it looks like they are referencing fields in the RPC predicate, inadvertently relying on the buggy behavior this PR intends to fix. I'll need to have a closer look tomorrow.

predicate/src/rpc_predicate/column_rewrite.rs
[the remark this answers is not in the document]
alamb: I do not have a preference -- you might have fewer tests to update if you ignore time as well

wolffcm:
@alamb I am finding this test case baffling:
https://github.com/influxdata/influxdb_iox/blob/52d70b060aed28f9c44454129fea1ec83d10c1cb/query_tests/src/influxrpc/read_filter.rs#L387-L406

The predicate
```SQL
(STATE = 'CA') OR (READING > 0)
```
isn't something that we would see coming from Influx RPC, since `READING` is a field, it would have to be something like:
```SQL
(STATE = 'CA') OR (_field = 'reading' AND _value > 0)
```
When I make this change to the test, it says that this predicate is unsupported since the conjunct mixes a `_field` column reference with another column---that's specifically the case we don't support. It also wouldn't get quite the same answer since `_field` = 'reading'` implies a projection which the original predicate does not.

The test seems to be verifying that the predicate is evaluated in a `FilterExec` in the DF plan, instead of just in `ParquetExec` using statistics (I'm guessing here). Would ORing two tag predicates together be enough to ensure the test covers what it is supposed to?

alamb:
> Would ORing two tag predicates together be enough to ensure the test covers what it is supposed to?

Yes I think so

I think the test case as written is likely due to my misunderstanding of how predicates on field values are supposed to work. Rewriting them is fine as long as flux doesn't make such predicates 👍

iox_query/src/frontend/influxrpc.rs
```diff
@@ -2491,16 +2494,20 @@ mod tests {
...
             col("_measurement")
                 .eq(lit("foo"))
                 .or(col("_measurement").eq(lit("h2o")))
-                .and(col("time").gt(lit_timestamp_nano(5))),
+                .and(col("foo").eq(lit("bar"))),
         );
 
-        // verify that the predicate was rewritten to time > 5
-        let expr = col("time").gt(lit_timestamp_nano(5));
+        // verify that the predicate was rewritten to foo = 'bar'
```
alamb, line 2505: 👍

predicate/src/rpc_predicate.rs
```diff
@@ -192,7 +192,7 @@ fn normalize_predicate(
     let mut predicate = predicate.clone();
 
     let mut field_projections = FieldProjectionRewriter::new(Arc::clone(&schema));
-    let mut missing_columums = MissingColumnRewriter::new(Arc::clone(&schema));
+    let mut missing_tag_columns = MissingTagColumnRewriter::new(Arc::clone(&schema));
```
alamb, line 195: Thank you for renaming this struct to match its new features

query_tests/src/influxrpc/read_filter.rs
```diff
@@ -390,19 +411,20 @@ async fn test_read_filter_data_pred_unsupported_in_scan() {
     // These predicates can't be pushed down into chunks, but they can
     // be evaluated by the general purpose DataFusion plan
 
-    // (STATE = 'CA') OR (READING > 0)
-    let predicate =
-        Predicate::default().with_expr(col("state").eq(lit("CA")).or(col("reading").gt(lit(0))));
+    // (STATE = 'CA') OR (CITY = 'Boston')
+    let predicate = Predicate::default()
```
alamb, line 415: looks good -- thank you

query_tests/src/scenarios/library.rs
```diff
@@ -644,6 +644,25 @@ impl DbSetup for TwoMeasurementsMultiSeries {
...
+#[derive(Debug)]
+pub struct TwoMeasurementsMultiTagValue {}
+#[async_trait]
+impl DbSetup for TwoMeasurementsMultiTagValue {
+    async fn make(&self) -> Vec<DbScenario> {
+        let partition_key = "1970-01-01T00";
+
+        let lp_lines = vec![
+            "h2o,state=MA,city=Boston temp=70.4 100",
+            "h2o,state=MA,city=Lowell temp=75.4 100",
```
alamb, line 656: Love the hometown reference

alamb (approved):
I think this looks beautiful @wolffcm  -- very nice 👍

Issue #6210, opened 2022-11-22
Querier: Push retention-predicate down to ParquetExec


NGA-TRAN:
# Rationale

This is a follow-up of https://github.com/influxdata/influxdb_iox/pull/6198/files#r1028537773 to push retention predicate down into ParquetExec

NGA-TRAN:
Close this because it will be a part of #6241

PR #6209, opened 2022-11-22
feat: querier filters data outsude retnetion period

NGA-TRAN:
Closes #6028 

- [ ] I've read the contributing section of the project [README](https://github.com/influxdata/influxdb_iox/blob/main/README.md).
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed).

PR #6213, opened 2022-11-22
feat: push retention predicate down to ParquetExec

NGA-TRAN:
Connect to #6210 


NGA-TRAN:
@alamb and @crepererum 
This PR has a different approach for filtering data outside retention period. In #6209, we did 2 steps: (i) early pruning chunks by adding `time >= retention_time` to early pruning predicate and then (ii) add `time < retention_time` into delete predicate to filter data after scanning the remaining chunks

We have found after the above PR that we can do better than just simply adding `time >= retention_time` into where predicate and then everything will be taken care for us. So this PR does:
1. Remove code did in previous PR #6209
2. Adding `time >= retention_time` to normal predicate. I found this must be done in 2 places, one for normal SQL and one for grpc.

I found this is more efficient but I will wait for both you you to review before making any move

querier/src/table/mod.rs
```diff
@@ -1253,46 +1235,6 @@ mod tests {
...
-        let max_tombstone_sequence_number = Some(SequenceNumber::new(2));
-        let ingester_partition =
-            builder.build(max_parquet_sequence_number, max_tombstone_sequence_number);
-
-        let querier_table = querier_table.with_ingester_partition(ingester_partition);
-
-        // There must be two delete predicates: one from the tombstone, one from retention
-        let deletes = num_deletes(querier_table.chunks().await.unwrap());
-        assert_eq!(&deletes, &[2, 0]);
-    }
```
NGA-TRAN, line 1294: This test is no longer needed because we do not add retention predicate into delete predicate anymore

querier/src/namespace/query_access.rs
```diff
@@ -684,26 +703,21 @@ mod tests {
...
+                "|               |           ParquetExec: limit=None, partitions=[1/1/1/1/<uuid>.parquet], predicate=time_max@0 >= -3600000000000, output_ordering=[host@0 ASC, time@2 ASC], projection=[host, load, time] |",
+                "|               |           ParquetExec: limit=None, partitions=[1/1/1/1/<uuid>.parquet], predicate=time_max@0 >= -3600000000000, output_ordering=[host@0 ASC, time@2 ASC], projection=[host, load, time] |",
+                "|               |                                                                                                                                                                                                                       |",
+                "+---------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+",
```
NGA-TRAN, line 720: Since we now have tests in query_tests, I am happy to remove this test here if you think it is better that way

alamb: yes I think removing the explain plans here would make maintaining the code easier

alamb (changes requested):
Actually, something doesn't look right upon more careful examination of the plans

querier/src/namespace/query_access.rs
```diff
@@ -56,9 +57,27 @@ impl QueryNamespace for QuerierNamespace {
             }
         };
 
+        let predicate = if let Some(retention_period) = table.namespace_retention_period() {
```
crepererum, line 60: Why duplicate this logic here and in the other `query_access` part (there are two because one is used for InfluxRPC, one for SQL) instead of making this part of `QuerierTable::chunks`?

crepererum (commented):
So the overall change is to encode the predicate as part of the predicate instead of the delete predicate, right?

The reason that this leads to a different query plan is clearly a weakness in our plan construction. They are semantically identical. I think we should fix the phys. plan construction, also see https://github.com/influxdata/influxdb_iox/issues/6098#issuecomment-1326153144 .

query_tests/cases/in/retention.expected
```diff
@@ -9,30 +9,23 @@
...
+|               |         DeduplicateExec: [host@0 ASC,time@2 ASC]                                                                                                                                                                          |
+|               |           SortPreservingMergeExec: [host@0 ASC,time@2 ASC]                                                                                                                                                                |
+|               |             UnionExec                                                                                                                                                                                                     |
+|               |               ParquetExec: limit=None, partitions=[1/1/1/1/00000000-0000-0000-0000-000000000000.parquet], predicate=time_max@0 >= -3600000000000, output_ordering=[host@0 ASC, time@2 ASC], projection=[host, load, time] |
+|               |               ParquetExec: limit=None, partitions=[1/1/1/1/00000000-0000-0000-0000-000000000001.parquet], predicate=time_max@0 >= -3600000000000, output_ordering=[host@0 ASC, time@2 ASC], projection=[host, load, time] |
```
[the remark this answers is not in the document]
crepererum: That's a bug in the phys. plan construction than. When building it, I was under the impression that the `ParquetExec` filter is exact.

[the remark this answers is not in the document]
crepererum: Oh, I think I know what's going on:

The outer `FilterExec` is created by DF itself because that's the predicate we actually want to apply (from an SQL user PoV). Now we mess with the predicate during phys. plan expansion and hence it's somewhat out-of-sync.

crepererum:
OK, so thinking about this and the bug it introduces, I would say let's rather improve the phys. plan construction and group chunks by delete predicate to avoid the excessive subtrees.

crepererum:
Also: why does `iox_query` not push down delete predicates to the chunks? I think we really should fix `iox_query` instead of adding more hacks to `querier`.

NGA-TRAN:
I have opened https://github.com/influxdata/influxdb_iox/issues/6241 to implement delete predicate pushdown and close this

PR #6223, opened 2022-11-24
chore: identifier parser handles preceding whitespace

### LINKED ISSUE #6214: InfluxQL: Fails to parse valid 3-part measurement name
Parser fails to interpret the following valid InfluxQL

```sql
SELECT usage_idle FROM "telegraf".. "cpu"
```

Per the canonical InfluxQL parser, identifiers can be preceded by optional whitespace:

https://github.com/influxdata/influxql/blob/7e7d61973256ffeef4b99edd0a89f18a9e52fa2d/parser.go#L433

Therefore an identifier could be specified as the following EBNF

```text
identifier          ::= whitespace? ( quoted_identifier | unquoted_identifier )

unquoted_identifier ::= [_a..zA..Z] [_a..zA..Z0..9]*

quoted_identifier   ::= '"' [^"\n] '"'
```

`whitespace` is any valid space or comment

stuartcarnie:
Closes #6214

## Issues resolved

This PR teaches the InfluxQL parser that identifiers may be preceded by whitespace. This addresses an issue that a 3-part name in the `FROM` clause may have whitespace following the period, such as

```sql
SELECT usage_idle FROM "telegraf".. "cpu"
```

## Features added

These APIs were developed whilst prototyping the InfluxQL DataFusion planner, and will be used in future PRs.

* Add an API to the `Identifier` type to determine if it must be surrounded in quotes. 
* `OneOrMore<T>` type implements the `Deref` trait, to obtain access to the underlying vector.


influxdb_influxql_parser/src/common.rs
```diff
@@ -427,6 +427,14 @@ impl<T> OneOrMore<T> {
     }
 }
 
+impl<T> Deref for OneOrMore<T> {
+    type Target = Vec<T>;
```
crepererum, line 431: Since you only implement `Deref` and not `DerefMut`, I think you could directly deref to `[T]` which would be more Rusty (I think).

influxdb_influxql_parser/src/select.rs
```diff
@@ -450,7 +450,7 @@ impl Parser for Field {
             pair(
                 arithmetic::<FieldExpression>,
                 opt(preceded(
-                    delimited(ws0, keyword("AS"), ws1),
+                    preceded(ws0, keyword("AS")),
```
crepererum, line 453: Is that a drive-by clean-up? Same goes for the other parser changes outside of `fn identifier`.

crepererum (approved):

influxdb_influxql_parser/src/select.rs
stuartcarnie: Yes, as the `keyword` combinator always peeks ahead one character for a separator, which includes whitespace, parens, etc. We don't need to force it to be whitespace in this instance, as it is followed by an `identifier`, which checks for that.

stuartcarnie: I made similar changes in other combinators.

influxdb_influxql_parser/src/common.rs
stuartcarnie: Good call! Makes sense to me, as `[T]` is more generic. `[T]` conveys that is borrowing a view into some sort of vector.

PR #6228, opened 2022-11-24
feat(ingester): ingester2 init

domodwyer:
This is actually pretty big - I am sorry. It is almost entirely a `cp` copy though.

## What the?!

This creates a new crate `ingester2`, that will serve as the development area for the ~kahless~ "kafkaless" design in isolation from the current `ingester` crate. This enables parallel development on the `main` branch for a little longer (yay no rebases!) but more importantly, isolates the `ingester2` code of any existing dependants that currently link with the `ingester` crate (of which there is a substantial amount of coupling). This decoupling is the primary motivator.

## What is?!

This commit contains a copy of many types that are in the `ingester` crate that can be re-used independently (such as the `ArcMap`, or the `BufferTree`, etc) with the concept of a "shard" removed throughout. It does not plumb them together - this will come in follow-up commits. 

This commit contains the following functionality

* In memory buffering (`BufferTree`) of operations and the partition FSM
* The `DmlSink` abstraction
* A new `query` abstraction, instrumentation, and types **but no query logic or Flight handlers**
* `ArcMap`
* `DeferredLoad`
* `QueryAdaptor` to present a query-able interface over ingested data
* `SequenceNumberRange` necessary to track the now internal offsets in a partition

Notably this does NOT include:

* gRPC handlers
* HTTP handlers (which appear to be plumbed in, but unused in the ingester!)
* Lifecycle management
* Anything to do with Kafka / write buffer / stream handling / etc
* Any shard related items
* Compaction code
* Any sort of way to initialise the types or use them

I will drag across other bits of code as they become necessary. 

## Next

I will start adding init code that glues these components together & gRPC handlers next.

Note that during the development of `ingester2` we MUST currently provide a `ShardId` for some methods - I have exposed a const `TRANSITION_SHARD_ID` in the crate root to use as a stub value for this purpose.

I will do my best to call out changes from the `ingester` crate below. Note there is no logic change in this commit, only moving things around, adding abstractions, and removing shards.

---

* feat(ingester): ingester2 init (a66fc0b64)

      Adds an ingester2 crate to hold the MVP of the Kafkaless project.
      
      This was necessary due to the tight coupling of the ingester internals
      with tests in external crates, and eases the parallel development of two
      version of the ingester.
      
      This commit contains various changes from the "ingester" crate, mostly
      removing the concept/references to a "shard" or "ShardId" where
      possible.
      
      This commit does not copy over all of the "ingester" crate - only those
      components that are definitely needed. I will drag across more as
      functionality is implemented.

### DIFF
--- a/ingester2/src/buffer_tree/namespace/name_resolver.rs
+++ b/ingester2/src/buffer_tree/namespace/name_resolver.rs
@@ -1,0 +1,138 @@
+use std::{sync::Arc, time::Duration};
+
+use backoff::{Backoff, BackoffConfig};
+use data_types::NamespaceId;
+use iox_catalog::interface::Catalog;
+
+use super::NamespaceName;
+use crate::deferred_load::DeferredLoad;
+
+/// An abstract provider of a [`DeferredLoad`] configured to fetch the
+/// [`NamespaceName`] of the specified [`NamespaceId`].
+pub(crate) trait NamespaceNameProvider: Send + Sync + std::fmt::Debug {
+    fn for_namespace(&self, id: NamespaceId) -> DeferredLoad<NamespaceName>;
+}
+
+#[derive(Debug)]
+pub(crate) struct NamespaceNameResolver {
+    max_smear: Duration,
+    catalog: Arc<dyn Catalog>,
+    backoff_config: BackoffConfig,
+}
+
+impl NamespaceNameResolver {
+    pub(crate) fn new(
+        max_smear: Duration,
+        catalog: Arc<dyn Catalog>,
+        backoff_config: BackoffConfig,
+    ) -> Self {
+        Self {
+            max_smear,
+            catalog,
+            backoff_config,
+        }
+    }
+
+    /// Fetch the [`NamespaceName`] from the [`Catalog`] for specified
+    /// `namespace_id`, retrying endlessly when errors occur.
+    pub(crate) async fn fetch(
+        namespace_id: NamespaceId,
+        catalog: Arc<dyn Catalog>,
+        backoff_config: BackoffConfig,
+    ) -> NamespaceName {
+        Backoff::new(&backoff_config)
+            .retry_all_errors("fetch namespace name", || async {
+                let s = catalog
+                    .repositories()
+                    .await
+                    .namespaces()
+                    .get_by_id(namespace_id)
+                    .await?
+                    .expect("resolving namespace name for non-existent namespace id")
+                    .name
+                    .into();
+
+                Result::<_, iox_catalog::interface::Error>::Ok(s)
+            })
+            .await
+            .expect("retry forever")
+    }
+}
+
+impl NamespaceNameProvider for NamespaceNameResolver {
+    fn for_namespace(&self, id: NamespaceId) -> DeferredLoad<NamespaceName> {
+        DeferredLoad::new(
+            self.max_smear,
+            Self::fetch(id, Arc::clone(&self.catalog), self.backoff_config.clone()),
+        )
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod mock {
+    use super::*;
+
+    #[derive(Debug)]
+    pub(crate) struct MockNamespaceNameProvider {
+        name: NamespaceName,
+    }
+
+    impl MockNamespaceNameProvider {
+        pub(crate) fn new(name: impl Into<NamespaceName>) -> Self {
+            Self { name: name.into() }
+        }
+    }
+
+    impl Default for MockNamespaceNameProvider {
+        fn default() -> Self {
+            Self::new("bananas")
+        }
+    }
+
+    impl NamespaceNameProvider for MockNamespaceNameProvider {
+        fn for_namespace(&self, _id: NamespaceId) -> DeferredLoad<NamespaceName> {
+            let name = self.name.clone();
+            DeferredLoad::new(Duration::from_secs(1), async { name })
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use data_types::ShardIndex;
+    use test_helpers::timeout::FutureTimeout;
+
+    use super::*;
+    use crate::test_util::populate_catalog;
+
+    const SHARD_INDEX: ShardIndex = ShardIndex::new(24);
+    const TABLE_NAME: &str = "bananas";
+    const NAMESPACE_NAME: &str = "platanos";
+
+    #[tokio::test]
+    async fn test_fetch() {
+        let metrics = Arc::new(metric::Registry::default());
+        let backoff_config = BackoffConfig::default();
+        let catalog: Arc<dyn Catalog> =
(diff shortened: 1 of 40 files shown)


ingester2/src/query/trait.rs
```diff
@@ -0,0 +1,47 @@
...
+#[async_trait]
+pub(crate) trait QueryExec: Send + Sync + Debug {
+    async fn query_exec(
+        &self,
+        namespace_id: NamespaceId,
+        table_id: TableId,
+        columns: Vec<String>,
+        span: Option<Span>,
+    ) -> Result<Response, QueryError>;
+}
```
domodwyer, line 29: This defines the interface we will use between the gRPC query API (currently a bit of an incomplete Flight API) and the actual query execution. This definitely needed some love!

This decouples the API handler, and the API types, from the actual query execution & internals within the ingester - this was what I was trying to do in the `ingester` crate but kept hitting wall after wall of coupling :(

ingester2/src/query/exec.rs
```diff
@@ -0,0 +1,38 @@
...
+        let mut _span_recorder = SpanRecorder::new(span);
+
+        info!(
+            namespace_id=%namespace_id,
+            table_id=%table_id,
+            columns=?columns,
+            "executing query"
+        );
+
+        unimplemented!();
```
domodwyer, line 36: This is a stub for the actual query execution (this is what the query API will call into).

ingester2/src/query/response.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Stream of partitions in this response.
+pub(crate) type PartitionStream =
+    Pin<Box<dyn Stream<Item = Result<PartitionResponse, ArrowError>> + Send>>;
+
+/// A response stream wrapper for ingester query requests.
+///
+/// The data structure is constructed to allow lazy/streaming/pull-based data
+/// sourcing..
+pub(crate) struct Response {
```
domodwyer, line 21: This defines the internal query result type an ingester will return to the query API handler - it allows pull-based streaming and is similar to what we do today in `ingester`. 

It just wraps up some otherwise complex types and provides convenience methods.

domodwyer (commented):
Everything else is pretty much the same!

ingester2/src/deferred_load.rs
```diff
@@ -0,0 +1,512 @@
...
+                // Sleep for the random duration, or until a demand call is
+                // made.
+                tokio::select! {
+                    _ = tokio::time::sleep(wait_for) => {
+                        trace!("timeout woke loader task");
+                    }
+                    _ = rx => {
+                        trace!("demand call woke loader task");
+                    }
+                }
```
SimonRichardson, line 194: :+1:

ingester2/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+// !!! PLEASE DO NOT EXPORT !!!
+//
+// Please be judicious when deciding if something should be visible outside this
+// crate. Overzealous use of `pub` exposes the internals of an Ingester and
+// causes tight coupling with other components. I know it's tempting, but it
+// causes problems in the long run.
+//
+// Ideally anything that NEEDS to be shared with other components happens via a
+// shared common crate. Any external components should interact with an Ingester
+// through its public API only, and not by poking around at the internals.
```
SimonRichardson, line 34: :+1:

SimonRichardson (approved):

ingester2/src/buffer_tree/partition/buffer.rs
```diff
@@ -0,0 +1,111 @@
...
+use self::{always_some::AlwaysSome, traits::Queryable};
+use crate::sequence_range::SequenceNumberRange;
+
+/// The current state of the [`BufferState`] state machine.
+///
+/// NOTE that this does NOT contain the [`Persisting`] state, as this is a
+/// immutable, terminal state that does not accept further writes and is
+/// directly queryable.
+#[derive(Debug)]
+#[must_use = "FSM should not be dropped unused"]
```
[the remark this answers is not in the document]
domodwyer: The actual data lifecycle of buffered within a single IOx partition is enforced through a finite state machine (of which rust lets you build pretty reliable/hard-to-use-wrong implementations!)

Say a caller transitions FSM states like this:

```rust
let new_state = fsm.transition();
```

This annotation will emit a lint warning if you do this:

```rust
fsm.transition();
```

Which is helpful at catching bugs - the caller should dropped the FSM state and the data within it!

ingester2/src/buffer_tree/partition/resolver/cache.rs
```diff
@@ -0,0 +1,408 @@
...
+
+        // Look up the partition key provided by the caller.
+        //
+        // If the partition key is a hit, clone the key from the map and return
+        // it instead of using the caller-provided partition key - this allows
+        // effective reuse of the same partition key str across all hits for it
+        // and is more memory efficient than using the caller-provided partition
+        // key in the PartitionData.
+        let key = entries.get_key_value(partition_key)?.0.clone();
+        let partition = entries.get_mut(partition_key).unwrap();
```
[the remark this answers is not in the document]
domodwyer: This is safe to `unwrap()` because of line 148 that checks for the existence of a value for the key `partition_key`, and returns an error if there isn't one (note the `?` in there!)

Once we know the entry exists, it is safe to grab the map entry and `unwrap()` it (which only fails if the get op refers to an entry that doesn't exist.

This is also _concurrency safe_ because we hold an exclusive lock over the map entries the whole time (line 139) :+1:

ingester2/src/buffer_tree/namespace/name_resolver.rs
```diff
@@ -0,0 +1,138 @@
...
+                    .get_by_id(namespace_id)
+                    .await?
+                    .expect("resolving namespace name for non-existent namespace id")
+                    .name
+                    .into();
+
+                Result::<_, iox_catalog::interface::Error>::Ok(s)
+            })
+            .await
+            .expect("retry forever")
```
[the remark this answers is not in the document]
domodwyer: This is called in a background thread that can happily sit there spinning and retrying (with backoff delays of course!)

ingester2/src/buffer_tree/partition/buffer/state_machine/buffering.rs
```diff
@@ -0,0 +1,90 @@
...
+            // It is a logical error to snapshot an empty buffer.
+            return Transition::unchanged(self);
+        }
+
+        // Generate a snapshot from the buffer.
+        let snap = self
+            .state
+            .buffer
+            .snapshot()
+            .expect("snapshot of non-empty buffer should succeed");
```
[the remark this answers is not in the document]
domodwyer: 0c4366b59 :+1:

SimonRichardson (approved):

alamb:
FWIW I think the idea of "develop in parallel on the main branch" served us very well during the NG->OG transition. As always there are tradeoffs between this and a long lived feature branch, but I think this approach will allow incremental migration which will be important to roll this thing out.

PR #6229, opened 2022-11-24
refactor: avoid `tokio::spawn` for cache requests

crepererum:
During the happy-path, we don't need any tokio task to drive a cache loader requests because the future issuing the request just act as a driver. If that future is cancelled, we place the cache request in an extra task. This avoid latencies due to task overhead and (task) context switches for most requests. This may remove a millisecond or two from latency but also makes the whole thing easier to analyze/profile because we don't spawn a truckload of tasks.

This trick was borrowed from rskafka.


cache_system/src/cache/driver.rs
```diff
@@ -63,7 +65,7 @@ where
     ) -> (Self::V, CacheGetStatus) {
         // place state locking into its own scope so it doesn't leak into the generator (async
         // function)
-        let (receiver, status) = {
+        let (fut, receiver, status) = {
```
alamb, line 68: This closure is getting substantial -- maybe it could be factored into its own function (mostly so there is a better place to document what the returned values are / should be used for)

cache_system/src/cache/driver.rs
```diff
@@ -91,12 +94,13 @@ where
                 let tag = state.tag_counter;
                 state.tag_counter += 1;
 
-                // need to wrap the query into a tokio task so that it doesn't get cancelled when
+                // need to wrap the query into a `CancellationSafeFuture` so that it doesn't get cancelled when
```
alamb, line 97: I wonder what is the rationale for this `CancellationSafeFuture` as in why avoid the drop when this request is dropped? Is it so that any work done to fetch the backing data is not wasted?

alamb (approved):
LGTM

cache_system/src/cache/driver.rs
```diff
@@ -45,6 +45,97 @@ where
             loader,
         }
     }
+
+    fn start_new_query(
```
alamb, line 49: 👍

PR #6233, opened 2022-11-25
feat(ingester2): gRPC methods & type-erased init

domodwyer:
This PR adds the gRPC handlers, and starts to build out the method of initialising and exposing them.

The RPC direct-write (from a router) code is largely unchanged from the `ingester` crate. The query RPC handler logic is unchanged, but I have tidied up the types and introduced an abstraction (`QueryExec` added in #6228) to separate the API handler and the actual internal query execution (they were a bit of a mess!). This should make development less annoying.

---

* feat(ingester2): gRPC methods & type-erased init (a0ab78298)

      This commit implements the gRPC direct-write RPC interface (largely
      copied from the ingester crate), and adds a much improved RPC query
      handler.
      
      Compared to the ingester crate, the query API is now split into two
      defined halves - the API handler side, and types necessary to support it
      (server/grpc/query.rs) and the Ingester query execution side (a stub in
      query/exec.rs). These two halves maintain a separation of concerns, and
      are interfaced by an abstract QueryExec trait (in query/trait.rs).
      
      I also added the catalog RPC interface as it is currently exposed on the
      ingester, though I am unsure if it is used by anything.
      
      This commit also introduces the "init" module, and the
      IngesterRpcInterface trait within it. This trait forms the public
      ingester2 crate API, defining the complete set of methods external
      crates can expect to utilise in a stable, unchanging and decoupled way.
      
      The IngesterRpcInterface trait also serves as a method of type-erasure
      on the underlying handler implementations, avoiding the need to
      expose/pub the types, abstractions, and internal implementation details
      of the ingester to external crates.


ingester2/src/init.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Acquire an opaque handle to the Ingester's Arrow Flight
+    /// [`FlightService`] RPC handler implementation, allowing at most
+    /// `max_simultaneous_requests` queries to be running at any one time.
+    fn query_service(
+        &self,
+        max_simultaneous_requests: usize,
+        metrics: &metric::Registry,
+    ) -> FlightServiceServer<Self::FlightHandler>;
+}
```
domodwyer, line 46: This is my plan for exporting a stable interface to the `ingester2` crate that can be used to drive an actual `ingester2` instance, or implement alternative behaviours (such as error injection) for external crates that depend on testing against an `ingester2` instance.

Crucially this abstraction keeps the internals of the ingester **internal** - we don't have to `pub` the internal types and traits just so the callers can initialise the gRPC handlers (which inevitably leads to type leakage, false dependency and tight coupling).

This file will soon contain a method like this:

```rust
pub fn new(catalog: Arc<dyn Catalog>, metrics, etc ...) -> impl IngesterRpcInterface {
    unimplemented!()
}
```

The `ioxd_ingester` create will call this to fully initialise an `ingester2` instance, and bind the RPC handlers - this forms the API boundary & `ingester2` entrypoint.

ingester2/src/server/grpc.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Implement the type-erasure trait to hide internal types from crate-external
+/// callers.
+impl<D, Q> IngesterRpcInterface for GrpcDelegate<D, Q>
+where
+    D: DmlSink + 'static,
+    Q: QueryExec + 'static,
+{
+    type CatalogHandler = CatalogService;
+    type WriteHandler = RpcWrite<Arc<D>>;
+    type FlightHandler = query::FlightService<Arc<Q>>;
```
domodwyer, line 54: This shows the type erasure of the `IngesterRpcInterface` in action - the internal traits (`DmlSink` and `QueryExec`, and the types they use such as error types and arguments) do not have to be publicly exported as they form bounds on a trait implementation, not the trait itself.

The actual concrete implementations are associated types, allowing decoupling / mocking / replacing / decorating / etc and erasing their type too - they also do not have to be `pub` :tada:

ingester2/src/server/grpc/query.rs
```diff
@@ -0,0 +1,634 @@
...
+    /// ingester services.
+    ///
+    /// This allows the ingester to drop a portion of requests when experiencing
+    /// an unusual flood of requests
+    request_sem: Semaphore,
+
+    /// Number of queries rejected due to lack of available `request_sem`
+    /// permit.
+    query_request_limit_rejected: U64Counter,
+}
```
domodwyer, line 108: This is a significantly cleaned-up Flight RPC implementation (kinda) that handles only the type parsing and logic of the API interface - the actual business logic of query execution is decoupled from the API implementation via the `query_handler` type which is an implementation of the `QueryExec` trait.

ingester2/src/server/grpc/query.rs
```diff
@@ -0,0 +1,634 @@
...
+        query_handler: Q,
+        max_simultaneous_requests: usize,
+        metrics: &metric::Registry,
+    ) -> Self {
+        let query_request_limit_rejected = metrics
+            .register_metric::<U64Counter>(
+                "query_request_limit_rejected",
+                "number of query requests rejected due to exceeding parallel request limit",
+            )
+            .recorder(&[]);
```
domodwyer, line 121: We retain the simultaneous query request limit & metrics :+1:

ingester2/src/server/grpc/query.rs
```diff
@@ -0,0 +1,634 @@
...
+        let request = proto::IngesterQueryRequest::decode(&*ticket.ticket).map_err(Error::from)?;
+
+        // Extract the namespace/table identifiers
+        let namespace_id = NamespaceId::new(request.namespace_id);
+        let table_id = TableId::new(request.table_id);
+
+        // Predicate pushdown is part of the API, but not implemented.
+        if let Some(p) = request.predicate {
+            warn!(predicate=?p, "ignoring query predicate (unsupported)");
+        }
```
domodwyer, line 186: Predicate pushdown forms part of the querier <-> ingester API, but is not currently supported in the `ingester` crate, so I added this warning and clarifying comment.

ingester2/src/server/grpc/query.rs
```diff
@@ -0,0 +1,634 @@
...
+
+        let response = self
+            .query_handler
+            .query_exec(
+                namespace_id,
+                table_id,
+                request.columns,
+                span_ctx.child_span("ingester query"),
+            )
+            .await?;
```
domodwyer, line 196: This passes the necessary information (using non-API types) into the abstract query delegate for execution.

ingester2/src/server/grpc/query.rs
```diff
@@ -0,0 +1,634 @@
...
+                Err(e) => futures::stream::once(async { Err(e) }).boxed(),
+            })
+            .boxed()
+    }
+}
+
+/// A mapping decorator over a [`FlatIngesterQueryResponseStream`] that converts
+/// it into [`arrow_flight`] response frames.
+#[pin_project]
+struct FlightFrameCodec {
```
domodwyer, line 355: This type is called `GetStream` in the `ingester` crate, but I figured out what it was doing and gave it a facelift :joy:

ingester2/src/server/grpc/rpc_write.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+}
+
+/// A gRPC [`WriteService`] handler.
+///
+/// This handler accepts writes from an upstream, and applies them to the
+/// provided [`DmlSink`].
+#[derive(Debug)]
+pub(crate) struct RpcWrite<T> {
```
domodwyer, line 87: This file is largely the same as the `ingester` implementation, modulo some minor error cleanup.

SimonRichardson (approved):

ingester2/src/init.rs
alamb: The basic idea looks great -- thank you.

As you work out additional interfaces, something else that would be super helpful  for query tests would be an equivalent interface for writing data (and eventually for controlling major externally observable data lifecycle actions like 'write data to parquet and remove from memory')

ingester2/src/server/grpc/query.rs
alamb: This is a good idea. For what it is worth this code inspired some of https://github.com/apache/arrow-rs/issues/3371

TLDR is I think that once I get the flight encoding/decoding upstreamed this code can be dramatically simplified (hopefully using `futures::Stream` combinators. 

Other than less code in IOx, I hope upstreaming would also allow us to sort out proper dictionary encoding. 

I don't plan any changes into ingester2 for a while, but since I ran across this code I figured I would telegraph my intentions

PR #6236, opened 2022-11-28
feat: InfluxQL planner learns how to normalise InfluxQL AST

stuartcarnie:
This PR teaches IOx how to normalise a `SELECT` statement AST.

**What is normalisation?**

Normalisation, with respect to InfluxQL, is about rewriting the AST such that it is easier to transform into a DataFusion LogicalPlan by

* expanding wildcard and regular expressions in the `SELECT` projection list
  * e.g. `SELECT * FROM ...` or `SELECT /foo/ FROM ...`
* expanding wildcard and regular expressions in function calls
  * e.g. `SELECT COUNT(*) ...` or `SELECT SUM(/foo/) ...`
* expanding regular expressions in the `FROM` clause to their matching measurements
  * e.g. `SELECT .. FROM /foo/`
* expanding wildcard and regular expressions in `GROUP BY` clauses
  * e.g. `SELECT ... GROUP BY *` or `SELECT ... GROUP BY /foo/`
* determining field types for variable references
  * e.g. `SELECT a_field, a_tag FROM foo` becomes `SELECT a_field::float, a_tag::tag FROM foo`
* performing such expansion recursively into subqueries
  * e.g. `SELECT .. FROM (SELECT foo FROM  bar)`

For the greatest compatibility, the Rust implementation is largely modelled on the existing InfluxQL implementation, which starts with the `RewriteFields` method[^1]:

https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/ast.go#L1181-L1185

[^1]: A notable deviation is that the Rust implementation also expands the `FROM` clause.

With this step complete, we will be able to expand the implementation of the `InfluxQLToLogicalPlan` type:

https://github.com/influxdata/influxdb_iox/blob/a32db32f7adcfbcbc54f417d58924ba0c4e84a9c/iox_query/src/plan/influxql.rs#L24

to first normalise the AST. 

**Next steps**

The next step will be to transform the AST into a `LogicalPlan`. In the Go implementation, much of the validation of the AST is performed by the `Compile` function:

https://github.com/influxdata/influxdb/blob/37088e8f5330bec0f08a376b2cb945d02a296f4e/influxql/query/compile.go#L107

This code will for a significant part of the inspiration for the AST → `LogicalPlan` transformation phase.


influxdb_influxql_parser/src/expression/arithmetic.rs
```diff
@@ -781,4 +796,34 @@ mod test {
...
+        // Ensure ordering of data types relative to one another.
+
+        assert!(Float < Integer);
+        assert!(Integer < Unsigned);
+        assert!(Unsigned < String);
+        assert!(String < Boolean);
+        assert!(Boolean < Field);
+        assert!(Field < Tag);
+
+        assert!(Float.is_field_type());
```
alamb, line 813: 👍

iox_query/src/exec/context.rs
```diff
@@ -333,6 +337,33 @@ impl IOxSessionContext {
         ctx.create_physical_plan(&logical_plan).await
     }
 
+    /// Prepare an InfluxQL statement for execution. This assumes that any
+    /// tables referenced in the InfluxQL have been registered with this context
+    pub async fn prepare_influxql(
```
alamb, line 342: Why was this moved out of InfluxQLQueryPlanner ? I thought the encapsulation in the InfluxQLPlanner was nice 🤷  not that it really matters

iox_query/src/exec/context.rs
```diff
@@ -333,6 +337,33 @@ impl IOxSessionContext {
...
+        query: &str,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let ctx = self.child_ctx("prepare_influxql");
+        debug!(text=%query, "planning InfluxQL query");
+        let mut statements = parse_statements(query)
+            .map_err(|e| DataFusionError::External(format!("{}", e).into()))?;
+
+        if statements.len() != 1 {
+            return Err(DataFusionError::NotImplemented(
+                "The context currently only supports a single SQL statement".to_string(),
```
alamb, line 354:
```suggestion
                "The context currently only supports a single InfluxQL statement".to_string(),
```

iox_query/src/plan/influxql/expr_type_evaluator.rs
```diff
@@ -0,0 +1,278 @@
+#![allow(dead_code)]
```
alamb, line 1: I wonder is this clippy lint still needed?

iox_query/src/plan/influxql/field.rs
```diff
@@ -0,0 +1,155 @@
...
+/// of rows.
+pub(crate) fn field_by_name(select: &SelectStatement, name: &str) -> Option<Field> {
+    select.fields
+        .iter()
+        .find(|f| {
+            field_name(f) == name || match &f.expr {
+                Expr::Call { name: func_name, args } if (func_name.eq_ignore_ascii_case("top")
+                    || func_name.eq_ignore_ascii_case("bottom"))
+                    && args.len() > 2 =>
+                    args[1..].iter().any(|f| matches!(f, Expr::VarRef { name: field_name, .. } if field_name.as_str() == name)),
```
alamb, line 49: 🤯

iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -0,0 +1,829 @@
...
+use influxdb_influxql_parser::string::Regex;
+use influxdb_influxql_parser::visit::{Recursion, Visitable, Visitor, VisitorResult};
+use itertools::Itertools;
+use std::borrow::Borrow;
+use std::collections::HashSet;
+use std::ops::Deref;
+use std::sync::Arc;
+
+fn parse_regex(re: &Regex) -> Result<regex::Regex> {
+    regex::Regex::new(re.as_str()).map_err(|e| {
```
alamb, line 25: FWIW we found that the go regex and the rust regex libraries are not quite the same -- here is what we did: 
https://github.com/influxdata/influxdb_iox/blob/5b9e5b4a922080cf9d8eca910d79a59e95b8e823/query_functions/src/regex.rs#L138-L149

iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -0,0 +1,829 @@
...
+            _ => unreachable!(),
+        }
+    }
+
+    #[test]
+    fn test_rewrite_statement() {
+        // Exact, match
+        let stmt = parse_select("SELECT usage_user FROM cpu");
+        let stmt = rewrite_statement(&stmt, MockSchemaProvider::new_schema_provider()).unwrap();
+        assert_eq!(format!("{}", stmt), "SELECT usage_user::float FROM cpu");
```
alamb, line 526: A style suggestion is that you can get the string form using `to_string()` rather than `format!()` like

```suggestion
        assert_eq!(stmt.to_string(), "SELECT usage_user::float FROM cpu");
```

I don't think it matters but I wanted to point it out

iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -0,0 +1,829 @@
...
+        // Subquery, exact, no match
+        let stmt = parse_select("SELECT usage_idle FROM (SELECT usage_idle FROM foo)");
+        let stmt = rewrite_statement(&stmt, MockSchemaProvider::new_schema_provider()).unwrap();
+        assert_eq!(
+            format!("{}", stmt),
+            "SELECT usage_idle FROM (SELECT usage_idle )"
+        );
+
+        // Subquery, regex, no match
+        let stmt = parse_select("SELECT bytes_free FROM (SELECT bytes_free FROM /^d$/)");
```
alamb, line 647: reading these testcases is a fascinating education in InfluxQL

alamb (approved):
Nice work @stuartcarnie 🏅 

I read this PR, though I will freely admit that I did not carefully understand each and every piece of it as I am not an InfluxQL expert. 

That being said, I think in my review I believe the code is very well structured / commented / tested and I would personally enjoy working on this code in the future (as 👨‍🍳 👌 )

Some thoughts:
It might make sense to refactor all of the influxql query planning into its own crate (as a follow on PR) -- `influxdb_influxql_planner` perhaps?

iox_query/src/exec/context.rs
stuartcarnie: I thought I was being consistent with the `prepare_sql` implementation by moving it back here, but I do agree that it is better encapsulated as it was, so I'll move it back.

Thanks!

iox_query/src/plan/influxql/expr_type_evaluator.rs
stuartcarnie: You wonder correctly - it is not, so I've removed it. Good catch!

💡 Be nice if `clippy` identified redundant attributes like this.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: Great! I'll use that function and thanks for the heads up.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: Definitely cleaner and thanks for pointing it out. I see that the implementation of `ToString` is generic for any type that implements `fmt::Display`, which is nice.

stuartcarnie:
Thanks for the review, Andrew. I've amended the PR with all your great suggestions 🙇🏻

alamb:
it is so exciting to see this progress

Issue #6216, opened 2022-11-23
Error rather than panic / segfault when trying to convert overly large series sets


alamb:
# Background
@tustvold  has found the root cause of a segfault we are seeing in production: https://github.com/apache/arrow-rs/pull/3157

However the fix in arrow (eta in IOx mid next week) will turn the segfault into a panic, which is good !

# Proposed changes

We should add some sort of  limit in IOx to error (rather than segfault or panic) if the query would try to convert too much data for the non streaming converter. 

# Background
The root cause of the segfault is that several RecordBatch's are being concatenated together (243 as I recall) and this causes one of the output string arrays to exceed 2GB in length (and overflowing the i32 used to index them)
The concat is happening here: https://github.com/influxdata/influxdb_iox/blob/9fb1de04287e2078011c78286ca42a5d3b3ea030/iox_query/src/exec/seriesset/converter.rs#L93-L104

The way to fix the problem for real would is streaming conversion to series sets

So by streaming the conversion https://github.com/influxdata/influxdb_iox/issues/4445 we will avoid the `concat` and thus avoid trying to make such a large string array and avoid hitting the overflow (edited)

PR #6239, opened 2022-11-28
refactor: make `ReadResponse` chunking stream-based

crepererum:
First step towards https://github.com/influxdata/influxdb_iox/issues/6216. I need to change the implementation "backwards", i.e. there are a few processing steps are the record batch => series set stuff that are not streaming which need to be adjusted first. This is the first PR to do that.


service_grpc_influxrpc/src/response_chunking.rs
```diff
@@ -0,0 +1,302 @@
...
+where
+    E: Send + 'static,
+{
+    /// Create new stream wrapper.
+    ///
+    /// # Panic
+    /// Panics if `size_limit` is 0.
+    pub fn new<S>(inner: S, size_limit: usize) -> Self
+    where
+        S: Stream<Item = Result<ReadResponse, E>> + Send + 'static,
```
crepererum, line 33: The funny thing is that the processing step before actually produces `Frame`s instead of `ReadResponse`s internally. The next PR will change the interface to avoid this additional flattening step.

service_grpc_influxrpc/src/response_chunking.rs
```diff
@@ -0,0 +1,302 @@
...
+            return Poll::Ready(None);
+        }
+
+        loop {
+            match ready!(this.inner.poll_next_unpin(cx)) {
+                Some(Ok(frame)) => {
+                    let fsize = frame_size(&frame);
+
+                    if fsize > this.size_limit {
+                        warn!(
```
alamb, line 74: Would it be better to return an error in this case? Or maybe make it a `debug!` message? I was thinking, if we see this message in the logs, is it clear what we would do about it 🤔

service_grpc_influxrpc/src/response_chunking.rs
```diff
@@ -0,0 +1,302 @@
...
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use futures::{ready, stream::BoxStream, Stream, StreamExt};
+use generated_types::{read_response::Frame, ReadResponse};
+use observability_deps::tracing::warn;
+
+/// Chunk given [`ReadResponse`]s -- while preserving the [`Frame`] order -- into responses that shall at max have the
+/// given size.
```
alamb, line 11:
```suggestion
/// given `size_limit`, in bytes.
```

service_grpc_influxrpc/src/response_chunking.rs
```diff
@@ -0,0 +1,302 @@
...
+pub struct ChunkReadResponses<E>
+where
+    E: Send + 'static,
+{
+    inner: BoxStream<'static, Result<Frame, E>>,
+    size_limit: usize,
+    finished: bool,
+    frames: Vec<Frame>,
+    frames_size: usize,
```
alamb, line 20:
```suggestion
    /// Current size of `frames`, in bytes
    frames_size: usize,
```

alamb (approved):
Looks great -- thank you @crepererum

service_grpc_influxrpc/src/response_chunking.rs
crepererum: Yeah, I think we should error in this case. Let me change that.

crepererum: done

PR #6247, opened 2022-11-28
feat: Add a feature flag to switch to the router RPC write path

### LINKED ISSUE #6242: Feature flag for RPC write router
Add a feature flag to be able to compile and run a router that uses the `RpcWrite` DML handler.

These routers should take an array of ingester addresses to set up a round robin sharder.

carols10cents:
Fixes #6242.

## Guiding philosophy

MVP! MVP! There's some unwraps, duplication, and no tests yet as being able to run an ingester2 to connect this to is next on my TODO list 😅 

I tried to use the feature flag in the least invasive way I could think of, at the CLI level. I opted to add new types that do have a lot of duplication with existing router types, but I didn't want to have to add irrelevant aspects like topic, query pool, etc to the new design. This way, hopefully, removing the old design code should mean deleting whole types/impls rather than picking through all places a `cfg` is used.

This also keeps as much code as possible compiling with the default feature set, so that there isn't a case of someone testing with the wrong feature flag set breaking code and not realizing it. When I add tests for this in a future PR, the tests will do `Command::cargo_bin` with feature flags, which will catch the few bits that only get compiled with the feature flag, and thus catch in CI any other breakage.

## How the feature flag hides this router mode

<details>
<summary>Long Details</summary>

If you run `cargo run -- run --help` (without the feature flag), you'll see:

```
$ cargo run -- run --help
Run the InfluxDB IOx server

Usage: influxdb_iox run [OPTIONS] [COMMAND]

Commands:
  compactor
          Run the server in compactor mode
  querier
          Run the server in querier mode
  router
          Run the server in router mode
  ingester
          Run the server in ingester mode
  all-in-one
          Run the server in "all in one" mode (Default)
  test
          Run the server in test mode
  garbage-collector
          Run the server in garbage collecter mode
  help
          Print this message or the help of the given subcommand(s)
```

and be none the wiser about the new way of running the router. However, if you run `cargo run --features rpc_write -- run --help`, you'll see:

```
$ cargo run --features rpc_write -- run --help
Run the InfluxDB IOx server

Usage: influxdb_iox run [OPTIONS] [COMMAND]

Commands:
  compactor
          Run the server in compactor mode
  querier
          Run the server in querier mode
  router
          Run the server in router mode
  router-rpc-write
          Run the server in router mode using the RPC write path
  ingester
          Run the server in ingester mode
  all-in-one
          Run the server in "all in one" mode (Default)
  test
          Run the server in test mode
  garbage-collector
          Run the server in garbage collecter mode
  help
          Print this message or the help of the given subcommand(s)
```

Note the new `router-rpc-write` command.

If you try to run that command but forget to turn the feature on, you'll get:

```
cargo run -- run router-rpc-write --help
error: The subcommand 'router-rpc-write' wasn't recognized

  Did you mean 'router'?
```

But with the feature, you'll see:

```
cargo run --features rpc_write -- run router-rpc-write --help
Run the server in router mode using the RPC write path

Usage: influxdb_iox run router-rpc-write [OPTIONS] --ingester-addresses <INGESTER_ADDRESSES>
```

</details>


ioxd_router/src/lib.rs
```diff
@@ -144,6 +145,81 @@ where
...
+    }
+
+    /// Registers the services exposed by the router [`RpcWriteGrpcDelegate`] delegate.
+    ///
+    /// [`RpcWriteGrpcDelegate`]: router::server::grpc::RpcWriteGrpcDelegate
+    async fn server_grpc(self: Arc<Self>, builder_input: RpcBuilderInput) -> Result<(), RpcError> {
+        let builder = setup_builder!(builder_input, self);
+        add_service!(builder, self.server.grpc().schema_service());
+        add_service!(builder, self.server.grpc().catalog_service());
+        add_service!(builder, self.server.grpc().object_store_service());
```
carols10cents, line 208: In contrast to the existing router server, this does not have the shard service (because it's not relevant) nor does it have the namespace service (because that currently requires a query ID and topic ID, and I don't think the new design requires that). I chanted MVP! MVP! when deciding to skip the namespace service; let me know if it should actually be implemented now instead!

router/src/dml_handlers/rpc_write.rs
```diff
@@ -75,7 +78,7 @@ where
     C: client::WriteClient,
 {
     type WriteInput = Partitioned<HashMap<TableId, (String, MutableBatch)>>;
-    type WriteOutput = ();
+    type WriteOutput = Vec<DmlMeta>;
```
carols10cents, line 81: This worked better with the existing chain of DML handlers, even though this isn't quite the same as the way the old design worked-- is this ok for now?

domodwyer: Yep! No problem :+1:

ioxd_router/src/lib.rs
domodwyer: We will need the namespace autocreation for the MVP too I'm afraid.

It is a bit messier, but I think using the same init fn (`create_router_server_type`) for both modes and `cfg()`-ing within the fn (and the ingester addr arg) might be the least worse route.  I'd like to avoid having two different inits if possible, partly so it is clearer that there is "The One True Way" and also because with the reintroduction of the namespace autocreate, it's almost entirely the same.

Just use the query & topic IDs that are passed through for the existing code now - we can set them via env/flags I think if we need to. We will have to rip it all out later!

domodwyer (commented):
The behaviour described in `How the feature flag hides this router mode` is perfect - I am mostly aiming to avoid any accidental interaction with the existing prod code path, and this looks like it nails that requirement 👌

ioxd_router/src/lib.rs
carols10cents: Done-- there are 2 new commits for you to review!

domodwyer (approved):
Thanks Carol - top commenting - hopefully that stops any potential mistakes in adding new code to one half but not the other.

PR #6259, opened 2022-11-29
perf(ingester2): streaming queries

domodwyer:
This PR changes the query reponse code path in `ingester2`, diverging it from how queries are streamed in the current ingester code.

This started out as an attempt to make the code/types more readable, and then once I understood what was happening optimising it for a fully streaming response was easy :tada:

**NOTE:** This PR changes the way queries _will_ be run - currently the query path is not connected up in `ingester2` (soon!)

---

* refactor(ingester2): explicit PartitionStream type (2ed9780f6)

      Simplify the streaming types by introducing explicitly named wrappers to
      improve visibility.

* perf(ingester2): streaming query data sourcing (1a379f5f1)

      Changes the query code (taken from the ingester crate) to stream data
      for query execution, tidy up unnecessary Result types and removing
      unnecessary indirection/boxing.
      
      Previously the query data sourcing would collect the set of RecordBatch
      for a query response during execution, prior to sending the data to the
      caller. Any data that was dropped or modified during this time meant the
      underlying ref-counted data could not be released from memory until all
      outstanding queries referencing it completed. When faced with multiple
      concurrent queries and ongoing ingest, this meant multiple copies of
      data could be held in memory at any one time.
      
      After this commit, data is streamed to the user, minimising the duration
      of time a reference to specific partition data is held, and therefore
      eliminating the memory overhead of holding onto all the data necessary
      for a query for as long as the client takes to read the data.
      
      When combined with an upcoming PR to stream RecordBatch out of the
      BufferTree, this should provide performant query execution with minimal
      memory overhead, even for a maliciously slow reading client.

### DIFF
--- a/ingester2/src/query/response.rs
+++ b/ingester2/src/query/response.rs
@@ -11,24 +11,33 @@ use futures::{Stream, StreamExt, TryStreamExt};
 use super::partition_response::PartitionResponse;
 
 /// Stream of partitions in this response.
-pub(crate) type PartitionStream =
-    Pin<Box<dyn Stream<Item = Result<PartitionResponse, ArrowError>> + Send>>;
+pub(crate) struct PartitionStream(
+    Pin<Box<dyn Stream<Item = Result<PartitionResponse, ArrowError>> + Send>>,
+);
+
+impl std::fmt::Debug for PartitionStream {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("PartitionStream").finish()
+    }
+}
+
+impl PartitionStream {
+    pub(crate) fn new<T>(s: T) -> Self
+    where
+        T: Stream<Item = Result<PartitionResponse, ArrowError>> + Send + 'static,
+    {
+        Self(s.boxed())
+    }
+}
 
 /// A response stream wrapper for ingester query requests.
 ///
 /// The data structure is constructed to allow lazy/streaming/pull-based data
-/// sourcing..
+/// sourcing.
+#[derive(Debug)]
 pub(crate) struct QueryResponse {
     /// Stream of partitions.
     partitions: PartitionStream,
-}
-
-impl std::fmt::Debug for QueryResponse {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("Response")
-            .field("partitions", &"<PARTITION STREAM>")
-            .finish()
-    }
 }
 
 impl QueryResponse {
@@ -38,8 +47,10 @@ impl QueryResponse {
     }
 
     /// Return the stream of [`PartitionResponse`].
-    pub(crate) fn into_partition_stream(self) -> PartitionStream {
-        self.partitions
+    pub(crate) fn into_partition_stream(
+        self,
+    ) -> Pin<Box<dyn Stream<Item = Result<PartitionResponse, ArrowError>> + Send>> {
+        self.partitions.0
     }
 
     /// Reduce the [`QueryResponse`] to a set of [`RecordBatch`].
--- a/ingester2/src/query/response.rs
+++ b/ingester2/src/query/response.rs
@@ -2,18 +2,15 @@
 //!
 //! [`QueryExec::query_exec()`]: super::QueryExec::query_exec()
 
-use std::{pin::Pin, sync::Arc};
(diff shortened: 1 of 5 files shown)


ingester2/src/query/response.rs
```diff
@@ -2,61 +2,55 @@
...
-                    .into_record_batch_stream()
-                    .map_ok(|snapshot| {
-                        let schema = Arc::new(optimize_schema(&snapshot.schema()));
-                        snapshot
-                            .map(move |batch| optimize_record_batch(&batch?, Arc::clone(&schema)))
-                    })
-                    .try_flatten()
-            })
-            .try_flatten()
-            .try_collect()
```
domodwyer, line 59: Previously the full set of partition data would be collected into a `Vec` and then sent to the client because of these `flatten()`/`collect() calls.

ingester2/src/query/response.rs
```diff
@@ -2,61 +2,55 @@
...
-    pub(crate) fn into_partition_stream(self) -> PartitionStream {
-        self.partitions
+    pub(crate) fn into_partition_stream(self) -> impl Stream<Item = PartitionResponse> {
+        self.partitions.0
     }
 
-    /// Reduce the [`QueryResponse`] to a set of [`RecordBatch`].
-    pub(crate) async fn into_record_batches(self) -> Result<Vec<RecordBatch>, ArrowError> {
+    /// Reduce the [`QueryResponse`] to a stream of [`RecordBatch`].
+    pub(crate) fn into_record_batches(self) -> impl Stream<Item = Result<RecordBatch, ArrowError>> {
```
domodwyer, line 52: Now the gRPC handler is given a stream of partition data, which performs pull-based streaming all the way down to the data sources.

ingester2/src/server/grpc/query.rs
```diff
@@ -293,57 +293,50 @@ pub enum FlatIngesterQueryResponse {
...
+                            let head = futures::stream::once(async {
+                                Ok(FlatIngesterQueryResponse::StartSnapshot {
+                                    schema: schema_captured,
+                                })
+                            });
+
+                            // TODO: these optimize calls may be redundant
+                            //
+                            // See: https://github.com/apache/arrow-rs/issues/208
+                            let tail = match optimize_record_batch(&snapshot, Arc::clone(&schema)) {
```
domodwyer, line 323: These "optimise" fns were called _twice_ in the original code, once here, and once when collecting the `Vec<RecordBatch>`. They appear to copy some data around to work around an old issue which appears to have since been resolved: https://github.com/apache/arrow-rs/issues/208

I left them here out of an abundance of caution (I don't have time to be debugging query differences!) but perhaps @alamb can confirm this is no longer necessary?

SimonRichardson (approved):
This makes the stream logic a lot more readable.

ingester2/src/server/grpc/query.rs
alamb: Yes, they should no longer be necessary. Given they are confusing (this is the second time in a week) I will remove them.

ingester2/src/query/response.rs
```diff
@@ -2,61 +2,55 @@
...
-                        let schema = Arc::new(optimize_schema(&snapshot.schema()));
-                        snapshot
-                            .map(move |batch| optimize_record_batch(&batch?, Arc::clone(&schema)))
-                    })
-                    .try_flatten()
-            })
-            .try_flatten()
-            .try_collect()
-            .await
+            .flat_map(|partition| partition.into_record_batch_stream())
```
alamb, line 54: I think that this construction means that only a single partition will be read at a time, one after another (which is fine). If we ever need more parallelism the way you have modeled it as `async` `Stream`s is very nice and can be parallelized with `buffered`. Nice work

domodwyer: That is one benefit of structuring it this way - it is easy to control the _amount_ of parallelism; the client API sends data sequentially, so buffering everything into memory in the server is a absolute waste of resources.

By structuring it this way we can choose to _pre-fetch_ the data of the next partition asynchronously to sending data to the current partition to the client. Any parallelism after that is useless (assuming we can pre-fetch faster than we can send, which is likely the case). I'm holding off on that until it is necessary :+1:

ingester2/src/server/grpc/query.rs
alamb: To complete the thread: https://github.com/influxdata/influxdb_iox/pull/6262

PR #6262, opened 2022-11-29
refactor: Remove unnecessary `optimize_record_batch`

alamb:
# Rationale
The underlying problem of serializing too far too much data been fixed in Arrow for quite some time: https://github.com/apache/arrow-rs/pull/2040

However, we still have the old workaround in IOx. Its presence in IOx sometimes causes some confusion , as most recently seen by @domodwyer,  https://github.com/influxdata/influxdb_iox/pull/6259/files#r1034572191. It also came up recently in a discussion with @crepererum  and myself: https://github.com/influxdata/influxdb_iox/pull/6205#discussion_r1029176076

# Changes
1. Remove `optimize_record_batch`
2. Rename other functions and update docstrings to better explain what they do

# Notes:
The tests for avoiding huge serialization are still present and pass 👍 

Without the optimize dictionaries code, the end to end tests fail like this:

```
---- end_to_end_cases::all_in_one::ephemeral_mode stdout ----
thread 'end_to_end_cases::all_in_one::ephemeral_mode' panicked at 'Error executing query: ArrowError(InvalidArgumentError("dictionary id not found in schema"))', /Users/alamb/Software/influxdb_iox/test_helpers_end_to_end/src/client.rs:240:10
stack backtrace:
   0: rust_begin_unwind
```


arrow_util/src/optimize.rs
```diff
@@ -465,7 +429,7 @@ mod tests {
 
         let big_batch = truncate_batch(&batch, batch.num_rows() - 1);
         let optimized_big_batch =
-            optimize_record_batch(&big_batch, Arc::clone(&schema)).expect("failed to optimize");
+            prepare_batch_for_flight(&big_batch, Arc::clone(&schema)).expect("failed to optimize");
```
alamb, line 432: this is the test for the original large flight data issue, and it still passes

ingester/src/querier_handler.rs
```diff
@@ -133,7 +135,8 @@ impl IngesterQueryResponse {
                         .snapshots
                         .flat_map(|snapshot_res| match snapshot_res {
                             Ok(snapshot) => {
-                                let schema = Arc::new(optimize_schema(&snapshot.schema()));
+                                let schema =
```
alamb, line 138: Even though this special schema / unpacking dictionaries is still required, I think the new name better explains what it is for

arrow_util/src/optimize.rs
```diff
@@ -140,35 +141,12 @@ pub fn optimize_record_batch(batch: &RecordBatch, schema: SchemaRef) -> Result<R
     RecordBatch::try_new(schema, columns?)
 }
 
-fn deep_clone_array(array: &ArrayRef) -> ArrayRef {
```
alamb, line 143: This copy is no longer needed!

crepererum (approved):

alamb:
> Hydrating the dictionaries is unfortunate but I see there is an open issue for the improvement 👍

I agree it is unfortunatel

I added https://github.com/apache/arrow-rs/issues/1206  to my list of DataFusion tracking issues: https://github.com/influxdata/idpe/issues/16283

If you ever see this show up in a performance profile or we need to lower the latency of queries in general it would be a good thing to pursue. @tustvold  do you know of any work / reasons flight can't support dictionaries (or is it just  a "simple matter of code")?

tustvold:
I seem to remember there were complexities caused by the lack of consistent handling of dictionary IDs. The IPC format handles dictionaries differently from the C data model which arrow-rs follows

PR #6263, opened 2022-11-29
perf(ingester2): streaming buffer tree queries

domodwyer:
This PR teaches the (previously write-only) `BufferTree` to return arrow `RecordBatch` instances for a given table (across any and all partitions within it).

It does this in a different way to the `ingester` crate query path by deferring the partition read until the client is ready to consume the partition data (pull-based streaming) - this reduces the memory usage of a query, and avoids doing unnecessary work if the client disconnects. This also lets us define the deterministic read consistency of the `BufferTree`, which was harder to specify/not so clear before.

---

* refactor: associated QueryExec return type (de6f0468d)

      Allow the return type of the QueryExec trait's query_exec() method to be
      parametrised by the implementer.
      
      This allows the trait to be reused across different data sources that
      return differing concrete types.

* perf(ingester2): stream BufferTree partition data (95216055d)

      This commit implements the QueryExec trait for the BufferTree, allow it
      to be queried for the partition data it contains. With this change, the
      BufferTree now provides "read your writes" functionality.
      
      Notably the implementation streams the contents of individual partitions
      to the caller on demand (pull-based execution), deferring acquiring the
      partition lock until actually necessary and minimising the duration of
      time a strong reference to a specific RecordBatch is held in order to
      minimise the memory overhead.
      
      During query execution a client sees a consistent snapshot of
      partitions: once a client begins streaming the query response, incoming
      writes that create new partitions do not become visible. However
      incoming writes to an existing partition that forms part of the snapshot
      set become visible iff they are ordered before the acquisition of the
      partition lock when streaming that partition data to the client.

### DIFF
--- a/ingester2/src/buffer_tree/root.rs
+++ b/ingester2/src/buffer_tree/root.rs
@@ -1,17 +1,71 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use data_types::NamespaceId;
+use data_types::{NamespaceId, TableId};
 use dml::DmlOperation;
 use metric::U64Counter;
+use trace::span::Span;
 
 use super::{
     namespace::{name_resolver::NamespaceNameProvider, NamespaceData},
     partition::resolver::PartitionProvider,
     table::name_resolver::TableNameProvider,
 };
-use crate::{arcmap::ArcMap, dml_sink::DmlSink};
+use crate::{
+    arcmap::ArcMap,
+    dml_sink::DmlSink,
+    query::{response::QueryResponse, tracing::QueryExecTracing, QueryError, QueryExec},
+};
 
+/// A [`BufferTree`] is the root of an in-memory tree of many [`NamespaceData`]
+/// containing one or more child [`TableData`] nodes, which in turn contain one
+/// or more [`PartitionData`] nodes:
+///
+/// ```text
+///
+///                        ╔════════════════╗
+///                        ║   BufferTree   ║
+///                        ╚═══════╦════════╝
+///                                ▼
+///                         ┌────────────┐
+///                         │ Namespace  ├┐
+///                         └┬───────────┘├┐
+///                          └┬───────────┘│
+///                           └────┬───────┘
+///                                ▼
+///                         ┌────────────┐
+///                         │   Table    ├┐
+///                         └┬───────────┘├┐
+///                          └┬───────────┘│
+///                           └────┬───────┘
+///                                ▼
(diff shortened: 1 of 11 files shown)


ingester2/src/buffer_tree/root.rs
```diff
@@ -1,17 +1,71 @@
...
+/// the table's current set of partitions is created and the data within these
+/// partitions will be streamed to the client as they consume the response. New
+/// partitions that are created concurrently to the query execution do not ever
+/// become visible.
+///
+/// Concurrent writes during query execution to a partition that forms part of
+/// this snapshot will be visible iff the write has been fully applied to the
+/// partition's data buffer before the query stream reads the data from that
+/// partition. Once a partition has been read, the data within it is immutable
+/// from the caller's perspective, and subsequent writes DO NOT become visible.
```
domodwyer, line 65: This documents the primary in-memory buffering structure, the `BufferTree`, and the read consistency of queries executed against it.

ingester2/src/buffer_tree/root.rs
```diff
@@ -95,3 +149,701 @@ impl DmlSink for BufferTree {
...
+    ///    has been initialised.
+    ///  * Concurrent writes to partitions that form part of the read snapshot
+    ///    become visible if they are ordered/applied before the acquisition of
+    ///    the partition data by the query stream. Writes ordered after the
+    ///    partition lock acquisition do not become readable.
+    ///
+    /// All writes use the same write timestamp as it is not a factor in
+    /// ordering of writes.
+    #[tokio::test]
+    async fn test_read_consistency() {
```
domodwyer, line 737: This test _asserts_ the read consistency behaviour.

ingester2/src/buffer_tree/table.rs
```diff
@@ -203,6 +210,57 @@ impl TableData {
...
+                )),
+                id,
+                None,
+            );
+
+            span.ok("read partition data");
+            Some(ret)
+        });
+
+        Ok(PartitionStream::new(futures::stream::iter(partitions)))
```
domodwyer, line 260: Partitions within the queried table are evaluated lazily, avoiding lock contention by spreading the lock acquisitions out over the duration of the client reading the results, and reducing both the amount, and the duration of time partition data has to be kept in memory.

domodwyer (commented):
The logic change in this PR is pretty small - most of the diff is docs & tests 🎉

SimonRichardson (approved):

PR #6266, opened 2022-11-29
refactor: make `IOxSessionContext::exec` non-optional

crepererum:
`None` was only used for testing and even than we should probably have a proper executor instead of panicking for some methods.

Found while working on #6216.


alamb (approved):
LGTM

PR #6268, opened 2022-11-29
test: (DO NOT MERGE) Check new pushdown filter code in datafusion

alamb:
I am using this PR to check proposed changes to the pushdown filter code in https://github.com/apache/arrow-datafusion/pull/4365


iox_query/src/frontend/influxrpc.rs
```diff
@@ -640,7 +640,7 @@ impl InfluxRpcPlanner {
                     .with_predicate(predicate)
                     .build()?;
 
-                let tag_name_is_not_null = Expr::Column(tag_name.into()).is_not_null();
+                let tag_name_is_not_null = tag_name.as_expr().is_not_null();
```
alamb, line 643:
```rust
Expr::Column(tag_name.into()).is_not_null();
```

ends up parsing `.` into relation.fieldname

Which causes errors downstream after some changes in https://github.com/apache/arrow-datafusion/pull/4365

iox_query/src/frontend/influxrpc.rs
```diff
@@ -1431,10 +1431,7 @@ fn columns_in_predicates(
     match &predicate.field_columns {
         Some(field_columns) => {
             for field in field_columns {
-                columns.insert(Column {
-                    relation: None,
-                    name: (*field).clone(),
-                });
+                columns.insert(Column::from_name(field));
```
alamb, line 1434: This is just a drive by cleanup -- the semantics are the same

PR #6275, opened 2022-11-30
refactor: stream-based(TM) `to_series_and_groups`, part 3

crepererum:
For #6216.


iox_query/src/exec/context.rs
```diff
@@ -430,11 +432,13 @@ impl IOxSessionContext {
...
-        let handles = plans
-            .into_iter()
-            .map(|plan| {
-                let ctx = self.child_ctx("to_series_set");
-                self.run(async move {
+        let ctx = self.child_ctx("to_series_set");
+        let exec = self.exec.clone();
+        let data = futures::stream::iter(plans)
+            .then(move |plan| {
+                // let exec = exec.clone();
```
alamb, line 439:
```suggestion
```

iox_query/src/exec/context.rs
```diff
@@ -448,45 +452,37 @@ impl IOxSessionContext {
...
                 let series: Vec<Series> = series_set
                     .try_into()
                     .map_err(|e| Error::Execution(format!("Error converting to series: {}", e)))?;
-                data.extend(series);
-            }
-        }
-        let data = futures::stream::iter(data).map(Ok);
+                Ok(Some(futures::stream::iter(series).map(Ok)))
+            })
+            .try_flatten();
```
alamb, line 485: I wonder if there is any difference in parallelism here -- previously `try_join_all` was used, but with this change the stream is read item by item, right? Conveniently, now that you have modeled this with `Stream`s we could run them in parallel again using `buffered()` if that turns out to be needed

I don't think we should change this PR now, but I wanted to point it out

alamb (approved):
Some most excellent `async` wizardry

iox_query/src/exec/context.rs
crepererum: I think it may pull the stream in-order now (i.e. fully drain the first stream, then the 2nd, and so on). We could make this "flatten unordered" (there's even a ready-to-use method for it) but I think this might mess up the conversion later and we end up with a trunk-load of series-sets due to the interleaving record batches (also I think this may be semantically wrong). So let's leave this for now and let's do something smarter later if it's really needed.

alamb: I agree the ordering of the series is critical for correctness

> So let's leave this for now and let's do something smarter later if it's really needed.

Agreed 👍

PR #6278, opened 2022-11-30
feat: add router config parameter for retention

lukebond:
Closes https://github.com/influxdata/idpe/issues/16576

Adds a config param to the router to set the retention period used for autocreated namespaces. Previously it was hardcoded as `None` (infinite). Now it defaults to `None`, so should be no functional change until we set it.

Also a couple of drive-by refactors:
- the router had no clap block of its own (i guess because it isn't reused anywhere), but i moved its own config params out into one to be more in line with other components, e.g. the ingester - i then added the new parameter there
- the `ns_autocreation` module was moved from being a dml handler to a namespace resolver, but the old file wasn't deleted. i deleted it


ioxd_router/src/lib.rs
```diff
@@ -258,13 +257,21 @@ pub async fn create_router_server_type(
         });
     txn.commit().await?;
 
+    // retention_hours = 0 means infinite retention. It's in nanoseconds internally
+    let retention: Option<i64> = if router_config.new_namespace_retention_hours == 0 {
+        None
+    } else {
```
domodwyer, line 263: You could also make the clap argument an option, then if not specified it is none already, defaulting to infinite (which is a sensible default behaviour for everyone I suspect - you wouldn't like to discover there is a default retention duration by finding your data has gone!)

lukebond: yes that's better, thanks

lukebond: as ever with rust, leaning on the idioms makes things neater. no need of that cumbersome if/else anymore!

Issue #6270, opened 2022-11-29
WAL reinitialisation fails


domodwyer:
When I do:

* Initialise a `Wal`
* Get a `WalWriter`
* Write a `DmlWrite` to the `WalHandle`
* Drop the `Wal`
* Reinitialise a `Wal` with the same dir 💥 

```
---- init::wal_replay::tests::test_replay stdout ----
thread 'init::wal_replay::tests::test_replay' panicked at 'failed to initialise WAL: InvalidUuid { filename: "b364f1c7-392a-4436-bba8-324456d5d369.dat", source: Error(Char { character: '.', index: 37 }) }', ingester2/src/init/wal_replay.rs:270:14
```

PR #6280, opened 2022-11-30
feat(ingester2): commit DML ops to WAL

domodwyer:
This commit adds the functionality to write incoming DML operations to a write-ahead log.

There is currently no functionality to rotate or replay this log (coming next).

---

* refactor(Wal): remove SequenceNumberNg (30e2dd7fb)

      This actually starts getting more confusing than passing the bare u64
      around.

* refactor(wal): rename next_ops -> next_op (628e0a713)

      It only returns one op, so remove the plural.

* feat(ingester2): commit writes to write-ahead log (4f4687569)

      Adds WalSink, an implementation of the DmlSink trait that commits DML
      operations to the write-ahead log before passing the DML op into the
      decorated inner DmlSink chain.
      
      By structuring the WAL logic as a decorator, the chain of inner DmlSink
      handlers within it are only ever invoked after the WAL commit has
      successfully completed, which keeps all the WAL commit code in a
      singly-responsible component. This also lets us layer on the WAL commit
      logic to the DML sink chain after replaying any existing WAL files,
      avoiding a circular WAL mess.
      
      The application-logic level WAL code abstracts over the underlying WAL
      implementation & codec through the WalAppender trait. This decouples the
      business logic from the WAL implementation both for testing purposes,
      and for trying different WAL implementations in the future.

* docs: reference WAL cancellation safety ticket (2fb04e80b)

      Reference https://github.com/influxdata/influxdb_iox/issues/6281

### DIFF
--- a/ingester2/src/wal/wal_sink.rs
+++ b/ingester2/src/wal/wal_sink.rs
@@ -39,6 +39,8 @@ where
     async fn apply(&self, op: DmlOperation) -> Result<(), Self::Error> {
         // TODO: cancellation safety
         //
+        // See https://github.com/influxdata/influxdb_iox/issues/6281.
+        //
         // Once an item is in the WAL, it should be passed into the inner
         // DmlSink so that is becomes readable - failing to do this means writes
         // will randomly appear after replaying the WAL.
--- a/ingester2/src/wal/wal_sink.rs
+++ b/ingester2/src/wal/wal_sink.rs
@@ -1,0 +1,169 @@
+use async_trait::async_trait;
+use dml::DmlOperation;
+use generated_types::influxdata::iox::wal::v1::sequenced_wal_op::Op;
+use mutable_batch_pb::encode::encode_write;
+use wal::SequencedWalOp;
+
+use crate::dml_sink::{DmlError, DmlSink};
+
+use super::traits::WalAppender;
+
+/// A [`DmlSink`] decorator that ensures any [`DmlOperation`] is committed to
+/// the write-ahead log before passing the operation to the inner [`DmlSink`].
+#[derive(Debug)]
+pub(crate) struct WalSink<T, W = wal::WalWriter> {
+    /// The inner chain of [`DmlSink`] that a [`DmlOperation`] is passed to once
+    /// committed to the write-ahead log.
+    inner: T,
+
+    /// The write-ahead log implementation.
+    wal: W,
+}
+
+impl<T, W> WalSink<T, W> {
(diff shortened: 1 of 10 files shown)


ingester2/src/wal/wal_sink.rs
```diff
@@ -0,0 +1,171 @@
...
+        // TODO: cancellation safety
+        //
+        // See https://github.com/influxdata/influxdb_iox/issues/6281.
+        //
+        // Once an item is in the WAL, it should be passed into the inner
+        // DmlSink so that is becomes readable - failing to do this means writes
+        // will randomly appear after replaying the WAL.
+        //
+        // This can happen If the caller stops polling just after the WAL commit
+        // future completes and before the inner DmlSink call returns Ready.
```
domodwyer, line 49: This is low risk / not a big deal - I'll come back to it later: https://github.com/influxdata/influxdb_iox/issues/6281

wal/src/lib.rs
```diff
@@ -115,22 +114,6 @@ pub enum Error {
 /// A specialized `Result` for WAL-related errors
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-/// `SequenceNumber` is metadata provided by users of the WAL for their tracking purposes of data
-/// getting written into a segment. No properties of `SequenceNumber` are verified in the WAL.
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Copy, Clone)]
-#[serde(transparent)]
-pub struct SequenceNumberNg(u64);
```
domodwyer, line 122: I removed this as I realised later on that this might not be a good thing to use for ordering in an unordered / eventually consistent system.

domodwyer (commented):
https://github.com/influxdata/influxdb_iox/issues/6270 needs fixing before this will pass CI.

SimonRichardson (approved):

carols10cents:
I corrected the replay test, and it'll fail intermittently on this branch until file ordering is correct/consistent, which I fixed in https://github.com/influxdata/influxdb_iox/pull/6287.

wal/src/lib.rs
carols10cents: Should the sequence number be removed entirely then?

carols10cents (approved):

wal/src/lib.rs
domodwyer: TL;DR: no

Long: https://github.com/influxdata/influxdb_iox/pull/6293 - I think we'll need a sequence number regardless (and there's no need for it to be a i64) and I think it should be internal to the ingester. But probably later :+1:

PR #6277, opened 2022-11-30
refactor: simplify `SeriesSet`

crepererum:
`RecordBatch` offers zero-copy slicing, so there is no need to store the row range manually. This makes #6216 simpler.

PR #6282, opened 2022-11-30
refactor: stream-based `SeriesSetConvert::convert` interface

crepererum:
On top of #6277.

Change the interface of `SeriesSetConvert::convert` to be stream-based.
This is the final interface-prep step before actually implementing #6216.


iox_query/src/exec/seriesset/converter.rs
```diff
@@ -93,26 +63,54 @@ impl SeriesSetConverter {
...
         // https://github.com/influxdata/influxdb_iox/issues/4445
-        let batches = collect(it).await.context(ReadingSnafu)?;
-
-        let batch = if !batches.is_empty() {
-            compute::concat_batches(&batches[0].schema(), &batches).context(ConcatenatingSnafu)?
-        } else {
-            return Ok(vec![]);
-        };
+        let batches = collect(it).await.map_err(|e| {
+            DataFusionError::Context(
```
alamb, line 85: nice use of error context 👍

alamb (approved):

PR #6285, opened 2022-11-30
refactor: stream-based series-set conversion

### LINKED ISSUE #6216: Error rather than panic / segfault when trying to convert overly large series sets

crepererum:
~On top of #6282.~

Closes #6216
Closes https://github.com/influxdata/influxdb_iox/issues/4445


alamb:
I believe this also closes https://github.com/influxdata/influxdb_iox/issues/4445 so I updated the PR description to say so

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -76,28 +78,6 @@ impl SeriesSetConverter {
 
         let schema = it.schema();
 
-        // for now, this logic only handles a single `RecordBatch` so
```
alamb, line 79: 😍

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -112,17 +92,30 @@ impl SeriesSetConverter {
...
+            schema,
+            it,
+            tag_indexes,
+            field_indexes,
+            table_name,
+            tag_columns,
+        })
+    }
+
+    fn compute_changepoints(batch: &RecordBatch, tag_indexes: &[usize]) -> Vec<usize> {
```
alamb, line 110:
```suggestion
    /// Returns the row indexes in `batch` where all of the values in the `tag_indexes` columns
    /// take on a new value.
    ///
    /// For example:
    /// tags A, B
    ///
    /// If the input is 
    /// A | B | C
    ///  1 | 2 | x
    ///  1 | 2 | y
    ///  2 | 2 | z
    ///  3 | 3 | q
    ///  3 | 3 | r
    ///
    /// Then this function will return [3, 4]:
    /// * The row at index 3 has values for A and B (2,2) different than the previous row (1,2). 
    /// * Similarly the row at index 4 has values (3,3) which are different than (2,2).
    /// * However, the row at index 5 has the same values (3,3) so is not a transition point 
    fn compute_changepoints(batch: &RecordBatch, tag_indexes: &[usize]) -> Vec<usize> {
```

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -257,6 +226,183 @@ impl SeriesSetConverter {
...
+
+                    (batch, true)
+                }
+            };
+
+            // compute changepoints
+            let mut changepoints =
+                SeriesSetConverter::compute_changepoints(&batch, &this.tag_indexes);
+            if this.it_finished {
+                // need to finish this
```
alamb, line 324:
```suggestion
                // need to finish last SeriesSet
```

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -257,6 +226,183 @@ impl SeriesSetConverter {
...
+                1 => (
+                    this.open_batches.last().expect("checked length").clone(),
+                    false,
+                ),
+                _ => {
+                    let batch_last = &this.open_batches[this.open_batches.len() - 2];
+                    let batch_current = &this.open_batches[this.open_batches.len() - 1];
+                    assert!(batch_last.num_rows() > 0);
+
+                    let batch = match compute::concat_batches(
```
alamb, line 302: I don't understand why this is concatenating only the last row from `batch_last` -- I would have expected we would need the entire batch

I also don't understand why we are concatening the prior batches here as well as below

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -257,6 +226,183 @@ impl SeriesSetConverter {
...
+                    Ok(batch) => batch,
+                    Err(e) => {
+                        this.we_finished = true;
+                        return Poll::Ready(Some(Err(DataFusionError::ArrowError(e))));
+                    }
+                };
+
+                let last_cp = *changepoints.last().expect("checked length");
+                if last_cp == batch.num_rows() {
+                    // fully drained open batches
```
alamb, line 360: How can we know we fully drained the open batches? Wouldn't last_cp mean that the last transition was the last row in the batch (and thus it may be part of the subsequent series)

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -257,6 +226,183 @@ impl SeriesSetConverter {
...
+    field_indexes: FieldIndexes,
+    table_name: Arc<str>,
+    tag_columns: Arc<Vec<Arc<str>>>,
+    it: SendableRecordBatchStream,
+}
+
+impl Stream for SeriesSetConverterStream {
+    type Item = Result<SeriesSet, DataFusionError>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
alamb, line 246: I am probably missing something, and it doesn't have to be implemented this way, but I was expecting something that had only a single concat_batches call

Like:
1. Read next batch
2. Combine with any previous batch (via concat_batches)
3. Compute transitions and emit any SeriesSets
4. Leave any remaining rows in `self.open_batches` (but would always be only a single RecordBatch)

alamb (commented):
Thank you @crepererum  -- it may be that I am near the end of my day, but something about this logic doesn't seem quite correct to me. This is so exciting to see

I think this feature needs some extra unit tests -- specifically to test the transition logic and the handling of 0, 1 and multiple previous batches. Perhaps it is already tested elsewhere but I don't remember writing such tests. 

Perhaps you could implement a quasi-fuzz test using code in datafusion as a possible useful inspiration: https://github.com/apache/arrow-datafusion/blob/3fe542f2afcd5360edc9abb7ad1e8243b560a6b2/test-utils/src/lib.rs#L70-L92

crepererum:
> I think this feature needs some extra unit tests -- specifically to test the transition logic and the handling of 0, 1 and multiple previous batches. Perhaps it is already tested elsewhere but I don't remember writing such tests.

This is what `parse_to_iterators` in the tests does. It takes the input and produces MULTIPLE test scenarios w/ different batch sizes. I can however extend it to be even more mean.

iox_query/src/exec/seriesset/converter.rs
crepererum: This is close to what we do here, however I don't store the whole batch in `open_batches` (point 4) because otherwise you get `O(n^2)` complexity because you copy the data again and again (in worst case). I'll add some comments to make this clearer.

crepererum:
Done (extended docs and made tests even fuzziers but adding WAY more possible streams). @alamb PTAL

alamb (approved):
I love it -- thank you @crepererum  -- I forgot about parse_to_iterators.

The test coverage has convinced me -- I understand now the multiple open_batches optimization. Thank you for the documentation -- I found it much easier to follow this time. 

I think we may want to wait to merge this in until we have sorted out https://github.com/influxdata/idpe/issues/16583 but otherwise 👍

PR #6292, opened 2022-12-01
feat: Add mutable visitor to InfluxQL parser crate

### LINKED ISSUE #6291: InfluxQL: Provide a `VisitorMut` type for the parser
## Rationale

In order to execute an InfluxQL query, the AST must first be transformed into a DataFusion `LogicalPlan`. To simplify this process, it would be preferable to perform some normalisation and / or expansion of the parsed AST. Examples include 

* expanding wildcards in the `SELECT` projection list to their respective column names, 
* generating column aliases whilst applying InfluxQL rules for resolving conflicting column names,
* expanding wildcards in a `GROUP BY` clause, etc

## Proposal

Add a `VisitorMut` type as a compliment the existing AST `Visitor`, which is capable of mutating both the visitor and the AST.

## Alternatives

Whilst it is feasible to potentially use the existing `Visitor` and create intermediate data structures for this normalised form, the existing AST already provides the constructs needed to represent transformed SQL. Cloning the AST each time it is mutated could become costly, so a mutating form is preferred.

Another alternative is to perform the normalisation in the planner as it walks the AST. The downside of this approach is it results in a complicated planner implementation that has multiple responsibilities. The proposed approach allows us to validate the normalised form before it is handed to the planner.

stuartcarnie:
Closes #6291 

This PR teaches the InfluxQL parser crate how to visit the AST using mutable data structures. Allows the use case where a visitor may wish to mutate the AST to perform various forms of normalisation or rewriting.


domodwyer (approved):
Looks good! I am interested to see this in action 🙏

crepererum (approved):

PR #6293, opened 2022-12-01
refactor(ingester2): document reordering / remove seqnum ranges

domodwyer:
### Internal Reordering 

I spent a good bit of time yesterday thinking about internal reordering within the ingester, mostly with respect to persistence, [how we can commit to the WAL][commit], and how to know a WAL segment can be dropped (which should happen when all the writes in it have been persisted, but how can you know that in a system that allows internally out-of-order writes?)

[commit]:https://github.com/influxdata/influxdb_iox/issues/6276

The easy way out would be to serialise (to some degree) the write path, but I am convinced this will be painful for performance of both writes, and queries. If we can construct the system so that internal reordering is allowed, we'll reap the benefit of increased parallelism and minimal/very granular locking. Because the ingester is now in the hot / latency sensitive path, performance is more important than the previous kafka-pull model (any ingester latency is now visible to the user as write latency, before it would show up as TTBR).

The good news is **I believe I have a plan**! I am working on persistence today, and it does not require serialisation of writes / etc, and it is fairly simple too. The thinking time was worth it!

### This PR / The Plan

The above brought on this PR - I wanted to both document the reordering behaviour in the source code, and remove the `SequenceNumberRange` type which is flawed in a reordering system (or at least, less useful and not useful today).

In the `ingester(1)` implementation we would use the bounding values of sequence number ranges to determine progress. When a partition was persisted that had a range of `[0, 10)` we could say "all writes up to `10` have been persisted.

That is no longer true in `ingester2` due to reordering of writes - there can be gaps within that range, and the writes could turn up later. Persisting a range of `[0, 10)` only concretely states that two operations (with sequence numbers 0 and 9) were persisted, and maybe some others within it. After persisting that range, you could get a late write with sequence number 5 that was incorrectly believed to have been persisted!

Therefore the sequence number ranges have been removed, and instead we'll rely on sequence number bitmaps for equality checks: a range of `[0, 10)` equals a range of `[0, 10)`, but a bitmap containing `{0,1,2,3,4,5,6,7,8,9}` does not equal a bitmap of `{0, 9}`. If we incrementally mark WAL segment entries as persisted using their IDs in a bitmap, we can drop a segment file once we're certain all the writes have been persisted without relying on ordering/ranges/range bounds (effectively reference counting a segment file).

---

* docs: arbitrarily reordering & sequence numbers (50d5e4a6f)

      Document the arbitrary reordering of concurrent writes in an ingester,
      and the potential divergence of WAL entries / buffered state.
      
      Also documents that in ingester2, sequence numbers only identify writes,
      not their ordering.

* refactor: remove ordering asserts, sequence ranges (464242ebc)

      This commit removes the invariant asserts of monotonicity carried over
      from the "ingester" crate - ingester2 does not define any ordering of
      writes within the system.
      
      This commit also removes the SequenceNumberRange as it is no longer
      useful to indirectly check the equality of two sets of ops -
      non-monotonic writes means overlapping ranges does not guarantee a full
      set of overlapping operations (gaps may be present). Likewise bounding
      values (such as "min persisted") provide no useful meaning in an
      out-of-order system.


ingester2/src/lib.rs
```diff
@@ -1,4 +1,34 @@
...
+//! performance of both.
+//!
+//! Because of re-ordering, ranges of [`SequenceNumber`] cannot be used to
+//! indirectly equality match (nor prefix match) the underlying data;
+//! non-monotonic writes means overlapping ranges do not guarantee equality of
+//! the set of operations they represent (gaps may be present). For example, a
+//! range of sequence numbers bounded by `[0, 2)` for thread 1 may not contain
+//! the same op data as another thread T2 with range with the same bounds due to
+//! reordering causing T1 to observe `{0, 1, 2}` and T2 observing `{0, 2}` and
+//! after a faulty range comparison, `{1}` to converge.
```
domodwyer, line 28: I wanted to explicitly call out using `SequenceNumber` ranges gives us different (incomplete) information compared to how we use them in `ingester1`.

ingester2/src/sequence_range.rs
```diff
@@ -1,141 +0,0 @@
-use data_types::SequenceNumber;
-
-/// A range of sequence numbers, both inclusive [min, max].
-#[derive(Debug, Default, PartialEq, Eq, Clone)]
-pub(crate) struct SequenceNumberRange {
```
domodwyer, line 5: I removed this so we are not tempted to use it expecting a behaviour it does not provide in `ingester2`. Nothing in `ingester2` makes use of it today.

I will replace it with a bitmap of sequence numbers instead later.

SimonRichardson (approved):

PR #6300, opened 2022-12-01
feat(ingester2): WAL entry replay at startup

domodwyer:
This PR adds write-ahead log replay functionality to the ingester2 startup sequence.

Startup will block while the WAL is replayed (the gRPC write service will not be available until it completes).

**Currently WAL files are never deleted** - that code is coming, but for testing purposes you'll need to periodically delete WAL files or every startup will take longer and longer.

I've done my best to make WAL replay _observable_ from the logs, including the progress of replay, as it is often a source of operational surprise / concern.

---

* feat(ingester2): WAL entry replay at startup (cb248c75d)

      Replay the WAL log, if any, at startup.
      
      Op replay is performed synchronously, during initialisation of the
      ingester2 instance, and passes all ops through the "normal" write path
      that the system uses once replay is complete, minus the WAL writer layer
      - this helps to DRY the write path and minimise different behaviours.

### DIFF
--- a/ingester2/src/init/wal_replay.rs
+++ b/ingester2/src/init/wal_replay.rs
@@ -1,0 +1,280 @@
+use data_types::{NamespaceId, PartitionKey, Sequence, SequenceNumber, TableId};
+use dml::{DmlMeta, DmlOperation, DmlWrite};
+use generated_types::influxdata::iox::wal::v1::sequenced_wal_op::Op;
+use mutable_batch_pb::decode::decode_database_batch;
+use observability_deps::tracing::*;
+use thiserror::Error;
+use wal::Wal;
+
+use crate::{
+    dml_sink::{DmlError, DmlSink},
+    TRANSITION_SHARD_INDEX,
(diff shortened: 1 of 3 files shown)


ingester2/src/init/wal_replay.rs
```diff
@@ -0,0 +1,280 @@
...
+                .collect(),
+            partition_key,
+            // The tracing context should be propagated over the RPC boundary.
+            DmlMeta::sequenced(
+                Sequence {
+                    shard_index: TRANSITION_SHARD_INDEX, // TODO: remove this from DmlMeta
+                    sequence_number,
+                },
+                iox_time::Time::MAX, // TODO: remove this from DmlMeta
+                // TODO: A tracing context should be added for WAL replay.
```
SimonRichardson, line 154: :+1:

SimonRichardson (approved):
I had a question about delete operations not being handled, but @domodwyer enlightened me about the reason.

PR #6304, opened 2022-12-01
wip: initial poc to add WASM backed UDFs

nathanielc:
Closes #

### DIFF
--- a/iox_query/src/frontend/wasm.rs
+++ b/iox_query/src/frontend/wasm.rs
@@ -1,0 +1,24 @@
+use std::sync::Arc;
+
+use crate::exec::context::IOxSessionContext;
+use datafusion::{error::Result, physical_plan::ExecutionPlan};
+
+/// This struct can create plans for running SQL queries against databases
+#[derive(Debug, Default)]
+pub struct WASMQueryPlanner {}
+
+impl WASMQueryPlanner {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Plan a SQL query against the catalogs registered with `ctx`, and return a
+    /// DataFusion physical execution plan that runs on the query executor.
+    pub async fn query(
+        &self,
+        query: &str,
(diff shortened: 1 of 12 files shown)


crepererum:
I think we have two ways to implement UDFs in WASM:

1. **scalar function calls:** Call the WASM function for every scalar/tuple/row. That's what you do here. Easy to implement (esp. for the user), but will probably not scale very well because you have a method call overhead per row.
2. **vector function calls:** Call the WASM function with arrays for every column. That's slightly more difficult to implement for the user (requires an additional loop), but the overhead will be way smaller. Also the WASM runtime could perform vectorization of loops.
3. **arrow function calls:** Allow the WASM function to retrieve/send arrow data structures directly. This has max. complexity for the user, but we could provide frameworks/tools to simplify this (maybe even adapters that call variant 1 and 2). See [WASM Arrow](https://github.com/domoritz/arrow-wasm/) for prior art.

iox_query/src/frontend/wasm.rs
```diff
@@ -0,0 +1,24 @@
+use std::sync::Arc;
+
+use crate::exec::context::IOxSessionContext;
+use datafusion::{error::Result, physical_plan::ExecutionPlan};
+
+/// This struct can create plans for running SQL queries against databases
```
alamb, line 6: You probably already know this but I don't think you need this specific frontend -- you can probably just use the normal SQL frontend

wasm_demo/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+#[no_mangle]
+pub extern "C" fn wasm_mul(a: f64, b: f64) -> f64 {
```
alamb, line 2: 👍  nice

alamb (commented):
This is very cool -- I won't say I followed all the wasm trampoline code but the basic idea looks awesome

PR #6305, opened 2022-12-01
feat: Add cap limit on number of partitions to be compacted in parallel

NGA-TRAN:
Closes https://github.com/influxdata/idpe/issues/16584


compactor/src/lib.rs
```diff
@@ -640,161 +647,11 @@ pub mod tests {
...
-            .with_creation_time(hot_time_one_hour_ago);
-        partition6.create_parquet_file_catalog_record(pf6_1).await;
-
-        let pf6_2 = TestParquetFileBuilder::default()
-            .with_min_time(4) // overlapped with pf6_1
-            .with_max_time(6)
-            .with_row_count(2)
-            .with_compaction_level(CompactionLevel::FileNonOverlapped)
-            .with_creation_time(hot_time_one_hour_ago);
-        partition6.create_parquet_file_catalog_record(pf6_2).await;
```
NGA-TRAN, line 797: This block is moved into a function to be called in 2 places, one here and one in a new test

compactor/src/lib.rs
```diff
@@ -241,15 +241,18 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
         // 4. Almost hitting max budget (only 10% left)
         //    OR no more candidates
         //    OR already considered all remaining candidates.
+        //    OR hit the max number of partitions to compact in parallel
         if (!parallel_compacting_candidates.is_empty())
             && ((remaining_budget_bytes <= (compactor.config.memory_budget_bytes / 10) as u64)
                 || (candidates.is_empty())
-                || (count == num_remaining_candidates))
+                || (count == num_remaining_candidates)
```
alamb, line 248: I don't understand why this is in the same clause as the "almost hitting max memory budget" -- like why would parallel compaction be happening if we are running out of memory *or* of we were close to the budget.

compactor/src/lib.rs
```diff
@@ -241,15 +241,18 @@ async fn compact_candidates_with_memory_budget<C, Fut>(
...
+                || (count as u64 == compactor.config.max_parallel_partitions))
         {
             debug!(
                 num_parallel_compacting_candidates = parallel_compacting_candidates.len(),
                 total_needed_memory_budget_bytes =
                     compactor.config.memory_budget_bytes - remaining_budget_bytes,
+                config_max_parallel_partitions = compactor.config.max_parallel_partitions,
                 compaction_type,
                 "parallel compacting candidate"
             );
```
alamb, line 258: I was expecting the max_parallel_compactions passed into `compact_function` and  then 

```rust
async fn compact_in_parallel(
```


https://github.com/influxdata/influxdb_iox/blob/main/compactor/src/lib.rs#L315

Would be modified to limit the concurrency but yet run all the plans

I think it we were to update the code in `compact_in_parallel` to be in terms of a `stream` then we could call `buffered_unorderd(max_parallel_partitions)` https://docs.rs/futures/0.3.25/futures/stream/trait.StreamExt.html#method.buffer_unordered on it to limit the number of plans that were run at once

clap_blocks/src/compactor.rs
```diff
@@ -208,6 +208,18 @@ macro_rules! gen_compactor_config {
                 action
             )]
             pub hot_compaction_hours_threshold_2: u64,
+
+            /// Max number of partitions that can be compacted in parallel at once
```
alamb, line 212: I wonder if we still need `INFLUXDB_IOX_COMPACTION_MAX_NUMBER_PARTITIONS_PER_SHARD` after this change? maybe we could remove that setting and instead just use this new limit `INFLUXDB_IOX_COMPACTION_MAX_PARALLEL_PARTITIONS`?

alamb (commented):
I don't feel I understand how the compactor's parallelism story at the moment to properly approve this PR. It doesn't seem quite right to me but that may be due to my lack of knowledge. Maybe other reviewers will feel better about approving or maybe I need to be instructed better

clap_blocks/src/compactor.rs
NGA-TRAN: These 2 are different. 

- `INFLUXDB_IOX_COMPACTION_MAX_NUMBER_PARTITIONS_PER_SHARD` will make sure we do not starve compacting certain shards if their partitions happen to have smaller number of L0/L1 files.
- `INFLUXDB_IOX_COMPACTION_MAX_PARALLEL_PARTITIONS` ensures the number or partitions to compact in parallel

In each cycle, there will be `num_shards * INFLUXDB_IOX_COMPACTION_MAX_NUMBER_PARTITIONS_PER_SHARD` partitions to get compacted. Among them, number of `INFLUXDB_IOX_COMPACTION_MAX_PARALLEL_PARTITIONS`  will be compacted in parallel until all are compacted

compactor/src/lib.rs
NGA-TRAN: - `count == num_remaining_candidates` simply means we are at the end of the list and we should compact them (we hit bug without checking this and add them into compacting list)
- `remaining_budget_bytes <= (compactor.config.memory_budget_bytes / 10)` means we are running out of budget, let compact them before consider other candidates.
- `count as u64 == compactor.config.max_parallel_partitions` means we hit limit parallel job cap, let compact them.

compactor/src/lib.rs
NGA-TRAN: We actually use `buffered_unorderd(max_parallel_partitions)` before but we can no longer use it because we need to know ahead how many files in each partition can compact. If partition only consume small memory, we can use the rest of memory to compact more partitions; or we can also tell a partition it can only compact certain number of files due to the remaining budget. This is quite dynamic and the `buffered_ordered`can only be used if each partitions consume similar amount of memory

compactor/src/lib.rs
alamb: I see -- makes sense. Maybe we could make that clearer in the comments somehow -- sorry for my confusion

compactor/src/lib.rs
alamb: I see -- that makes sense. That the limit is not just the number of parallel compactions, but also the amount of memory those compactions are expected to consume

alamb (approved):
Make sense now -- thank you @NGA-TRAN

compactor/src/lib.rs
NGA-TRAN: I have made the comments clearer. Thanks @alamb

PR #6309, opened 2022-12-02
fix: check schemas in `pretty_print_batches`

crepererum:
I think most users of this function (and `assert_batches_eq`) assume that all batches have the same schema. If not, `pretty_print_batches` may either fail producing an actual table (some rows may have more or less columns) or silently produce a table that looks "alright".


iox_query/src/provider.rs
```diff
@@ -2283,27 +2288,30 @@ mod test {
...
-            "| 70        | CT   | AL   | 1970-01-01T00:00:00.000000100Z |",
-            "| 100       | AL   | AL   | 1970-01-01T00:00:00.000000050Z |",
-            "| 5         | MT   | MT   | 1970-01-01T00:00:00.000005Z    |",
-            "| 1000      | 1000 | CT   | 1970-01-01T00:00:00.000001Z    |",
-            "| 10        | 10   | MT   | 1970-01-01T00:00:00.000007Z    |",
-            "| 70        | 70   | AL   | 1970-01-01T00:00:00.000000100Z |",
-            "| 100       | 100  | AL   | 1970-01-01T00:00:00.000000050Z |",
-            "| 5         | 5    | MT   | 1970-01-01T00:00:00.000005Z    |",
-            "+-----------+------+------+--------------------------------+",
+            "+-----------+------+------+--------------------------------+------+------------+",
```
crepererum, line 2291: This is a good example how broken the printing was when schemas were NOT equal.

alamb:
related to https://github.com/influxdata/influxdb_iox/issues/5804

crepererum:
> related to #5804

I would argue that #5804 is an anti-feature. It shouldn't silently iron over different schemas. This is why I extracted an explicit helper (`equalize_batch_schemas`) that can be used if a user EXPLICITLY want's that.

arrow_util/src/display.rs
```diff
@@ -86,7 +86,16 @@ fn create_table(results: &[RecordBatch]) -> Result<Table> {
     }
     table.set_header(header);
 
-    for batch in results {
+    for (i, batch) in results.iter().enumerate() {
```
alamb, line 89: 👍

ingester/src/querier_handler.rs
```diff
@@ -187,7 +190,6 @@ impl IngesterQueryResponse {
     /// do not line up with the snapshot-scoped record batches.
     pub async fn into_record_batches(self) -> Vec<RecordBatch> {
         let mut snapshot_schema = None;
-        let mut schema_merger = SchemaMerger::new();
```
alamb, line 190: I am not sure it matters, but by not using SchemaMerger any IOx specific metadata is not checked for inconsistencies

iox_query/src/provider.rs
```diff
@@ -2170,27 +2172,30 @@ mod test {
...
-            "| 70        | CT   | 1970-01-01T00:00:00.000000100Z |                                |",
-            "| 100       | AL   | 1970-01-01T00:00:00.000000050Z |                                |",
-            "| 5         | MT   | 1970-01-01T00:00:00.000005Z    |                                |",
-            "| 1000      | MT   | 1970-01-01T00:00:00.000001Z    |                                |",
-            "| 10        | MT   | 1970-01-01T00:00:00.000007Z    |                                |",
-            "| 70        | CT   | 1970-01-01T00:00:00.000000100Z |                                |",
-            "| 100       | AL   | 1970-01-01T00:00:00.000000050Z |                                |",
-            "| 5         | MT   | 1970-01-01T00:00:00.000005Z    |                                |",
-            "+-----------+------+--------------------------------+--------------------------------+",
+            "+-----------+------+------+--------------------------------+-----------------+",
```
alamb, line 2175: 😮

alamb (approved):
Looks good to me -- thank you @crepererum

alamb:
cc @domodwyer as you may have some interest in this code / https://github.com/influxdata/influxdb_iox/issues/5804

ingester/src/querier_handler.rs
crepererum: I think I can live with that :D

Issue #6310, opened 2022-12-02
Switch querier mem limit from ahead-of-time to on-demand


crepererum:
See https://github.com/apache/arrow-datafusion/issues/3941 and <https://docs.google.com/document/d/1QMTTkL5R2lRuxUGfVgvpN9Ix37ErwrfeS7fzdAedVGA/edit?usp=sharing>.

PR #6311, opened 2022-12-02
refactor: improve DF error handling

crepererum:
This is required to extract "resource exhausted" errors in more cases.

For #6310 (the final end2end test will fail otherwise, just wanted to extract this for easier review).


service_common/src/error.rs
```diff
@@ -21,27 +20,8 @@ use datafusion::{arrow::error::ArrowError, error::DataFusionError};
...
-                e = inner;
-                continue;
-            }
-        }
-
-        // no more traversal
-        break;
-    }
+pub fn datafusion_error_to_tonic_code(e: &DataFusionError) -> tonic::Code {
+    let e = e.find_root();
```
alamb, line 24: ❤️

alamb (approved):
LGTM -- thank you

PR #6312, opened 2022-12-02
test: simplify test executors

crepererum:
Have a single global test executor w/ reasonable defaults. Also don't require tests to join/await executor shutdowns (most tests forget this anyways and will get a runtime warning).

For #6310 (because we'll need to pass the memory pool size down and I want to avoid to have this specified in every single test).


executor/src/lib.rs
```diff
@@ -150,6 +156,10 @@ impl std::fmt::Debug for DedicatedExecutor {
     }
 }
 
+/// [`DedicatedExecutor`] for testing purposes.
+static TESTING_EXECUTOR: Lazy<DedicatedExecutor> =
+    Lazy::new(|| DedicatedExecutor::new_inner("testing", 1, true));
```
alamb, line 161: I wonder if it would be wise to give the test executor more than one thread? Otherwise we will end up serializing the work tests do in the QueryExecutor to this single thread 🤔 

We can do it as a follow on PR if needed

alamb (approved):
LGTM -- thanks @crepererum

Issue #5776, opened 2022-09-30
explain fails with error if the query would scan more than configured maximum limit


mkmik:
__Steps to reproduce:__

1. create a table with a lot of data.
2. execute a an `explain select ....` with a query that would not fit in memory.

__Expected behaviour:__

When the query fits in memory, I get a plan:

```console
; ./target/debug/influxdb_iox query dab3b3b2e54dc4a4_ab038f883697864a "explain select count(*) from vibrations where time > '2022-09-30 09:04:00'::timestamp;"
+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                                                                                         |
+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------+
| logical_plan  | Projection: #COUNT(UInt8(1))                                                                                                                                 |
|               |   Aggregate: groupBy=[[]], aggr=[[COUNT(UInt8(1))]]                                                                                                          |
|               |     Filter: #vibrations.time > TimestampNanosecond(1664528640000000000, None)                                                                                |
|               |       TableScan: vibrations projection=[time], partial_filters=[#vibrations.time > TimestampNanosecond(1664528640000000000, None)]                           |
| physical_plan | ProjectionExec: expr=[COUNT(UInt8(1))@0 as COUNT(UInt8(1))]                                                                                                  |
|               |   AggregateExec: mode=Final, gby=[], aggr=[COUNT(UInt8(1))]                                                                                                  |
|               |     CoalescePartitionsExec                                                                                                                                   |
|               |       AggregateExec: mode=Partial, gby=[], aggr=[COUNT(UInt8(1))]                                                                                            |
|               |         CoalesceBatchesExec: target_batch_size=4096                                                                                                          |
|               |           FilterExec: time@0 > 1664528640000000000                                                                                                           |
|               |             UnionExec                                                                                                                                        |
|               |               ProjectionExec: expr=[time@3 as time]                                                                                                          |
|               |                 RepartitionExec: partitioning=RoundRobinBatch(4)                                                                                             |
|               |                   DeduplicateExec: [plant@1 ASC,sensor@2 ASC,device@0 ASC,time@3 ASC]                                                                        |
|               |                     SortExec: [plant@1 ASC,sensor@2 ASC,device@0 ASC,time@3 ASC]                                                                             |
[...]

PR #6313, opened 2022-12-02
refactor: DF-driven on-demand mem limit instead of ahead-of-time heuristics

### LINKED ISSUE #6310: Switch querier mem limit from ahead-of-time to on-demand

crepererum:
On top of #6311 and #6312.

Closes #6310.


clap_blocks/src/querier.rs
```diff
@@ -216,19 +225,6 @@ pub struct QuerierConfig {
     )]
     pub max_concurrent_queries: usize,
 
-    /// Maximum bytes to scan for a table in a query (estimated).
```
alamb, line 219: 🥳 🦜

clap_blocks/src/querier.rs
```diff
@@ -51,6 +51,15 @@ pub struct QuerierConfig {
     )]
     pub num_query_threads: Option<usize>,
 
+    /// Size of memory pool used during query exec, in bytes.
```
alamb, line 54:
```suggestion
    /// Size of memory pool used during query exec, in bytes.
    /// If queries attempt to allocate more than this many bytes
    /// during execution, they will error with "ResourcesExhausted"
```

clap_blocks/src/querier.rs
```diff
@@ -51,6 +51,15 @@ pub struct QuerierConfig {
...
+    #[clap(
+        long = "query-exec-mem-pool-bytes",
+        env = "INFLUXDB_IOX_QUERY_EXEC_MEM_POOL_BYTES",
+        default_value = "1073741824",  // 1GB
```
alamb, line 58: I wonder if we should increase the default to 4GB or 8GB 🤔

influxdb_iox/src/commands/compactor.rs
```diff
@@ -43,6 +43,15 @@ pub enum Command {
             action
         )]
         query_exec_thread_count: usize,
+
+        /// Size of memory pool used during query exec, in bytes.
```
alamb, line 47: FYI @NGA-TRAN 

This will almost certainly need to be adjusted up for many of the wide compaction jobs. Maybe initially on the production clusters we can set it super high (aka "no limit", the same as today)

In theory the runtime memory limit might  also be able to protect the compactor from OOMs -- we would have to teach the compactor how to detect and respond to "resource exhausted" errors while running the plans.

influxdb_iox/src/commands/run/ingester.rs
```diff
@@ -75,6 +75,15 @@ pub struct Config {
         action
     )]
     pub query_exec_thread_count: usize,
+
+    /// Size of memory pool used during query exec, in bytes.
```
alamb, line 79: FYI @domodwyer  -- we are in the process of teaching the DataFusion executor how to avoid OOMing by living within its means. 

When we roll this change out, I initially suggest cranking this value quite high (effectively no enforced limit), the same as today, and then we can figure out if/when we want to start using this feature to avoid OOMs in the ingester

influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -635,6 +638,14 @@ async fn oom_protection() {
                     } else {
                         panic!("Not a gRPC error: {err}");
                     }
+
+                    // EXPLAIN should work though
+                    run_sql(
```
alamb, line 643: 🎉

alamb (commented):
This is amazing @crepererum  -- really wonderful work. The last PR in the chain almost makes it look easy but I know how much work was involved getting to this point. 


One suggestion is to call the setting  `exec_mem_pool_bytes` rather than `query_exec_mem_pool_bytes` everywhere (even the querier) so as not to confuse people with "why am I setting something on the ingester / comapctor related to query"?

I also want to wait to approve this PR until we have had a chance to discuss / plan for its potential impacts on the compactor and querier. I think we may want to pre-position memory limit settings for the ingester and compactor that are very large (to maintain the existing behavior) and then figure out how best to use the limits in those two systems

alamb:
This PR  also helps with https://github.com/influxdata/idpe/issues/16493 https://github.com/influxdata/idpe/issues/16227 and https://github.com/influxdata/influxdb_iox/issues/5776 (probably fixes them, but I want to verify myself before I close them)

alamb:
Also helps with https://github.com/influxdata/idpe/issues/16557

influxdb_iox/src/commands/run/ingester.rs
domodwyer: Can we make sure this is done prior to the PR merging/limit landing in prod? Perhaps by pre-setting the env var to a high value in IDPE, and overriding it specifically for the queriers before we merge this PR.

I think the current ingesters (ingester1) will have a very horrible time if not.

domodwyer (commented):
I think we should do some preparation in IDPE before landing this.

influxdb_iox/src/commands/run/ingester.rs
alamb: I agree 100%

influxdb_iox/src/commands/compactor.rs
NGA-TRAN: Sounds good @alamb . As you chatted, I think if we do not panic due this over limit for each plan, the compactor will just ignore the failure and continue compacting other partitions. Then will recompact this partition in the next cycle. However, we need to explore the code to see if that is the case. I agree to make it very large first

crepererum:
Updated config names, set the max for ingester1/2 and compactor to `usize::MAX/inf`. Should be ready to go :rocket:

influxdb_iox/src/commands/run/compactor.rs
```diff
@@ -76,6 +76,15 @@ pub struct Config {
         action
     )]
     pub query_exec_thread_count: usize,
+
+    /// Size of memory pool used during query exec, in bytes.
+    #[clap(
+        long = "exec-mem-pool-bytes",
+        env = "INFLUXDB_IOX_EXEC_MEM_POOL_BYTES",
+        default_value = &USIZE_MAX[..],
```
alamb, line 84: Note the default limit has been increased to USIZE_MAX ✅

influxdb_iox/src/commands/run/ingester2.rs
```diff
@@ -70,6 +70,15 @@ pub struct Config {
         action
     )]
     pub exec_thread_count: usize,
+
+    /// Size of memory pool used during query exec, in bytes.
+    #[clap(
+        long = "exec-mem-pool-bytes",
+        env = "INFLUXDB_IOX_EXEC_MEM_POOL_BYTES",
+        default_value = &USIZE_MAX[..],
```
alamb, line 78: Note the default limit has been increased to USIZE_MAX ✅

alamb (approved):
Let's do it 👍

PR #6314, opened 2022-12-02
fix: gRPC errors regarding group cols

crepererum:
- missing group col prev. produced an "internal error" but should be "invalid argument"
- duplicate group cols produced a panic but should also be "invalid argument"


crepererum:
This needs another iteration, because some names are treated specially (e.g. `_field`).

crepererum:
> This needs another iteration, because some names are treated specially (e.g. `_field`).

done.

predicate/src/rpc_predicate.rs
```diff
@@ -57,6 +57,24 @@ pub const FIELD_COLUMN_NAME: &str = "_field";
...
 pub const VALUE_COLUMN_NAME: &str = "_value";
 
+/// Special group key for `read_group` requests.
+///
+/// Treat these specially and use value "" to mirror what TSM does
+/// see <https://github.com/influxdata/influxdb_iox/issues/2693#issuecomment-947695442>
+/// for more details.
+///
+/// See also [`GROUP_KEY_SPECIAL_STOP`].
+pub const GROUP_KEY_SPECIAL_START: &str = "_start";
```
alamb, line 67: The comments say use value "" but then the constant is "_start" which I found confusing

influxdb_iox/tests/end_to_end_cases/querier/influxrpc/read_group.rs
```diff
@@ -207,6 +207,90 @@ async fn test_read_group_periods() {
...
+            Step::Custom(Box::new(move |state: &mut StepTestState| {
+                async move {
+                    let grpc_connection = state
+                        .cluster()
+                        .querier()
+                        .querier_grpc_connection()
+                        .into_grpc_connection();
+                    let mut storage_client = StorageClient::new(grpc_connection);
+
+                    let group_keys = match variant {
```
alamb, line 258: nice

alamb (approved):
Seems reasonable to me -- thank you @crepererum

predicate/src/rpc_predicate.rs
crepererum: improved.

PR #6316, opened 2022-12-02
perf(ingester2): partition persist queue

domodwyer:
This PR changes the `PartitionData` to support an (unbounded) number of buffers in the "persisting" state - this lets us queue up partition work without having to wait for existing persist tasks to complete; see commit message.

I think this fixes https://github.com/influxdata/influxdb_iox/issues/4928 if I understand the ticket correctly.

I think @aierui might be interested in this too! :)

---

* perf(ingester2): partition persist queue (4f928fbd0)

      This commit swaps the existing single "persisting batch" slot (field) in
      a PartitionData for an ordered queue of outstanding partitions.
      
      This decouples marking a partition buffer for persistence from the
      actual persistence operation, allowing them to proceed at different
      rates. This reduces the complexity of persistence management, but also
      allows us to gracefully handle "hot" partitions; for example, this
      problematic scenario in the ingester(1) implementation during recovery:
      
          * Writes come into a partition, reaching a size/row/hotness limit
          * Partition is enqueued for persistence
          * More writes come into the new buffer, exceeding the same limit
          * Cannot persist the hot buffer because of outstanding persist op
      
      Without this change the only possibilities in this situation are:
      
          * stop ingest for the partition and error all writes that
            (partially!) write to the partition, or
          * continue accepting writes, allowing the partition to exceed the
            limit that marked it for persistence in the first place
      
      The latter is what the ingester(1) implementation does today, which
      results in partitions exceeding their row/size/age limits, which exist
      to limit the cost of generating the Parquet file from the buffer - this
      is a significant contributor to instability during recovery.
      
      This strategy enforces configured the limits on the partition buffer,
      but does not block / slow down recovery while persistence is completed.

### DIFF
--- a/ingester2/src/buffer_tree/partition.rs
+++ b/ingester2/src/buffer_tree/partition.rs
@@ -1,17 +1,21 @@
 //! Partition level data buffer structures.
 
-use std::sync::Arc;
+use std::{collections::VecDeque, sync::Arc};
 
 use data_types::{NamespaceId, PartitionId, PartitionKey, SequenceNumber, TableId};
 use mutable_batch::MutableBatch;
 use observability_deps::tracing::*;
 use schema::sort::SortKey;
 
-use self::buffer::{traits::Queryable, BufferState, DataBuffer, Persisting};
+use self::{
+    buffer::{traits::Queryable, BufferState, DataBuffer, Persisting},
+    persisting::{BatchIdent, PersistingData},
+};
 use super::{namespace::NamespaceName, table::TableName};
 use crate::{deferred_load::DeferredLoad, query_adaptor::QueryAdaptor};
 
 mod buffer;
+mod persisting;
 pub(crate) mod resolver;
 
 /// The load state of the [`SortKey`] for a given partition.
@@ -69,8 +73,18 @@ pub(crate) struct PartitionData {
     /// A [`DataBuffer`] for incoming writes.
     buffer: DataBuffer,
 
-    /// The currently persisting [`DataBuffer`], if any.
-    persisting: Option<BufferState<Persisting>>,
+    /// The currently persisting [`DataBuffer`] instances, if any.
+    ///
+    /// This queue is ordered from newest at the head, to oldest at the tail -
+    /// forward iteration order matches write order.
+    ///
+    /// The [`BatchIdent`] is a generational counter that is used to tag each
+    /// persisting with a unique, opaque identifier.
+    persisting: VecDeque<(BatchIdent, BufferState<Persisting>)>,
+
+    /// The number of persist operations started over the lifetime of this
+    /// [`PartitionData`].
+    started_persistence_count: BatchIdent,
 }
 
 impl PartitionData {
@@ -94,7 +108,8 @@ impl PartitionData {
             table_id,
             table_name,
             buffer: DataBuffer::default(),
-            persisting: None,
+            persisting: VecDeque::with_capacity(1),
+            started_persistence_count: BatchIdent::default(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


ingester2/src/buffer_tree/partition.rs
```diff
@@ -69,8 +73,18 @@ pub(crate) struct PartitionData {
...
+    /// This queue is ordered from newest at the head, to oldest at the tail -
+    /// forward iteration order matches write order.
+    ///
+    /// The [`BatchIdent`] is a generational counter that is used to tag each
+    /// persisting with a unique, opaque identifier.
+    persisting: VecDeque<(BatchIdent, BufferState<Persisting>)>,
+
+    /// The number of persist operations started over the lifetime of this
+    /// [`PartitionData`].
+    started_persistence_count: BatchIdent,
```
domodwyer, line 87: We tag each persisting buffer with a unique generation counter number to make for cheap equality comparisons when marking the buffer as persisted.

ingester2/src/buffer_tree/partition.rs
```diff
@@ -217,16 +235,30 @@ impl PartitionData {
...
+            .pop_back()
+            .expect("no currently persisting batch");
+
+        // Currently in ingester1 there is an invariant that partition buffers
+        // not be persisted out-of-order - the assert below enforces that.
+        assert_eq!(
+            old_ident,
+            batch.batch_ident(),
+            "out-of-order persist notification received"
+        );
```
domodwyer, line 251: I believe this is no longer an invariant we must uphold @pauldix?

This used to be enforced so that the queriers could use the lower sequence number bound / min-persisted to know what they should look in the object store for. All writes < N where in the ingester and all writes > N were in object storage.

With the new parquet discovery plan for the queriers, they react to the ingester saying "there are 3 new parquet files in object storage for this partition" causing them to pre-fetch (or know to fetch) the new files, as opposed to using the lower bound of the sequence number to know what to expect to be in object storage - that should let us remove this old invariant from ingester1 I believe? 

If so, removing it would have its benefits.

ingester2/src/buffer_tree/partition/persisting.rs
```diff
@@ -0,0 +1,84 @@
...
+        Self(self.0)
+    }
+
+    /// Only for tests, this allows reading the opaque identifier to assert the
+    /// value changing between persist ops.
+    #[cfg(test)]
+    pub(super) fn get(&self) -> u64 {
+        self.0
+    }
+}
```
domodwyer, line 24: This is the opaque persist batch identifier that does not leak outside of the persist code / modules.

ingester2/src/buffer_tree/partition/persisting.rs
```diff
@@ -0,0 +1,84 @@
...
+///
+/// Cloning this type is relatively cheap.
+///
+/// [`PartitionData::mark_persisting()`]: super::PartitionData::mark_persisting
+/// [`PartitionData::mark_persisted()`]: super::PartitionData::mark_persisted
+#[derive(Debug, Clone)]
+pub(crate) struct PersistingData {
+    data: QueryAdaptor,
+    batch_ident: BatchIdent,
+}
```
domodwyer, line 44: And this is the wrapper type that holds the persisting data, and the batch ident.

This wrapper gives us static type checking to ensure we only pass persisting data to `mark_persisted()` which is a nice, free, avoidance of potential bugs.

aierui:
@domodwyer  highly efficient 👍

SimonRichardson (approved):

ingester2/src/buffer_tree/partition.rs
pauldix: That's right, this isn't an invariant we need anymore. We can persist multiple Parquet files from a single partition at the same time.

PR #6318, opened 2022-12-02
feat(ingester2): parallel partition persistence

domodwyer:
Implements a persist system that buffers outstanding persist requests and provides controllable parallelism across partitions. All persist ops for a given partition are serialised.

This is very much MVP code - there are few tests, and currently no handling of back-pressure when the persist queue is full. I will return to this when I am back from holiday (unless someone writes the tests before then! :)

---

* feat(ingester2): parallel partition persistence (f52468760)

      Implements actor-based, parallel persistence in ingester2 with
      controllable fan-out parallelism and queue depths.
      
      This implementation encapsulates the complexity of persistence, queuing
      and parallelism - the caller simply uses the handle to persist a
      partition, while the actor handles fan-out to a set of persistence
      workers, compaction in a separate thread-pool, and optional completion
      notifications.
      
      By consistently hashing persist jobs onto workers, parallelism is
      achieved across partitions, but serialisation of partition persists is
      enforced so that the sort key update is correctly serialised.


ingester2/src/persist/handle.rs
```diff
@@ -0,0 +1,173 @@
...
+/// blocking tasks from being passed through to workers with spare capacity.
+///
+/// # Parallelism & Partition Serialisation
+///
+/// Persistence jobs are parallelised across partitions, with up to at most
+/// `n_worker` parallel persist executions at once.
+///
+/// Because updates of a partition's [`SortKey`] are not commutative, they must
+/// be serialised. For this reason, persist operations for given partition are
+/// always placed in the same worker queue, ensuring they execute sequentially.
```
domodwyer, line 110: This describes the implementation and properties it provides.

ingester2/src/persist/handle.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// broadcasts a notification.
+    ///
+    /// # Panics
+    ///
+    /// Panics (asynchronously) if the [`PartitionData`]'s sort key is updated
+    /// between persistence starting and ending.
+    ///
+    /// This will panic (asynchronously) if `data` was not from `partition` or
+    /// all worker threads have stopped.
+    pub(crate) async fn queue_persist(
```
domodwyer, line 157: This is how a user pushes persist tasks into the queue.

This will need improving to support back-pressure when the queue is full, and depending on the outcome of https://github.com/influxdata/influxdb_iox/pull/6316#discussion_r1038139759, it may need changing to make acquiring the persist data and enqueuing atomic to preserve ordering. Today only one thread persists, so this isn't an issue.

domodwyer (commented):
Note this isn't wired up - this just adds the persist code.

SimonRichardson (approved):

PR #6320, opened 2022-12-02
fix: Use client_util GrpcConnection instead of tonic Channel

carols10cents:
@domodwyer this appears to let the router send gRPC requests to the ingester for me, and no, I'm not sure why yet 😅


router/src/dml_handlers/rpc_write.rs
```diff
@@ -14,12 +14,17 @@ use observability_deps::tracing::*;
...
 /// Create a client to the ingester's write service.
-pub fn write_service_client(ingester_addr: &str) -> WriteServiceClient<Channel> {
-    WriteServiceClient::new(Channel::builder(ingester_addr.parse().unwrap()).connect_lazy())
+pub async fn write_service_client(
+    ingester_addr: &str,
+) -> WriteServiceClient<client_util::connection::GrpcConnection> {
+    let connection = client_util::connection::Builder::default()
+        .build(format!("http://{}", ingester_addr))
+        .await
+        .expect("failed to connect to server");
```
domodwyer, line 26:
```suggestion
        .unwrap_or_else(|e| panic!("failed to connect to server {ingester_addr}: {e}"));
```

I think we need to support some ingesters being down when a router starts up (otherwise availability is tied to a single ingester being offline) - I guess this is why there was a deferred connect thing? Perhaps we can error only if all are unreachable or something.

Either way, MVP! MVP! - can you create a ticket for it?

domodwyer (approved):
Awesome/thank you.

router/src/dml_handlers/rpc_write.rs
carols10cents: Done: https://github.com/influxdata/influxdb_iox/issues/6321

PR #6323, opened 2022-12-02
feat(ingester2): persist data when rotating WAL segments

domodwyer:
Trigger persistence when a WAL segment is rotated.

---

* feat(ingester2): run persistence task (66aab5553)

      Configures the initialisation of an ingester2 instance to spawn a
      persistence task (currently unused) and plumbs in various configuration
      parameters.

* feat(ingester2): persist partitions on WAL rotate (f145d6415)

      Changes the WAL rotation code to cause the current set of partitions to
      be submitted for persistence.
      
      Because rotating the WAL segment and triggering persistence is not
      atomic (not under an exclusive lock preventing writes) the persisted
      buffers MAY contain writes that appear in the new segment file.
      
      In the happy/non-crash path, this will have no effect, however if the
      ingester crashes and replays the WAL files, these writes will be
      duplicated into object storage (where compaction will resolve the issue)
      - this seems like a good trade-off, allowing us to avoid blocking write
      requests for as long as it takes to rotate the WAL and mark all
      partitions as persisting.
      
      (though currently WAL files are not dropped and thus everything is
      replayed all the time.)


NGA-TRAN (approved):
The PR does what it describes. I am not the expert in this area and someone else may want to review it, too

PR #6324, opened 2022-12-02
feat(ingester2): drop WAL segments after persist

domodwyer:
Once persistence is complete, drop the old WAL segment.

**There is a race in this logic** described (at length) in the code. I also propose a plan to fix it, and have synced with @carols10cents to discuss implementing it.

It is low risk, low impact, and definitely not a blocker for MVP testing next week - I will put together a proper solution when I am back :+1:

---

* feat(ingester2): drop WAL segments after persist (26bf54d04)

      Changes the WAL rotation task to drop the WAL segment after the
      partition data has completely persisted.
      
      This logic contains a race condition outlined (at length) in the code
      comments, along with a plan to resolve it once I'm back from holiday.
      For now, the extremely low likelihood and minor impact of the race
      occurring is likely acceptable for testing purposes.

### DIFF
--- a/ingester2/src/wal/rotate_task.rs
+++ b/ingester2/src/wal/rotate_task.rs
@@ -29,6 +29,49 @@ pub(crate) async fn periodic_rotation(
             "rotated wal"
         );
 
+        // TEMPORARY HACK: wait 5 seconds for in-flight writes to the old WAL
+        // segment to complete before draining the partitions.
+        //
+        // This can occur because writes to the WAL & buffer tree are not atomic
+        // (avoiding a serialising mutex in the write path).
+        //
+        // A flawed solution would be to have this code read the current
+        // SequenceNumber after rotation, and then wait until at least that
+        // sequence number has been buffered in the BufferTree. This may work in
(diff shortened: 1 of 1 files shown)


ingester2/src/wal/rotate_task.rs
```diff
@@ -29,6 +29,49 @@ pub(crate) async fn periodic_rotation(
...
+        // WAL for it to do a set difference to derive the remaining sequence
+        // IDs, and therefore number of references to the WAL segment. Once the
+        // set of remaining IDs is empty (all data is persisted), the segment is
+        // safe to delete. This content-addressed reference counting technique
+        // has the added advantage of working even with parallel / out-of-order
+        // / hot partition persists that span WAL segments, and means there's no
+        // special code path between "hot partition persist" and "wal rotation
+        // persist" - it all works the same way!
+        //
+        // TODO: this properly as described above.
```
domodwyer, line 71: This is what I discussed with @carols10cents and what plan to implement upon my return.

SimonRichardson (approved):
I approve the documentation, lovely.

Issue #6317, opened 2022-12-02
Querier support for new write path


alamb:
This is a place holder ticket to track work in the querier to support the new write path / ingester2

Possible work items:
- [x] Integrate ingester2 into querier (including tests)
- [x] https://github.com/influxdata/influxdb_iox/issues/6617
- [ ] integrate new DataFusion memory limits into ingester2
- [x] https://github.com/influxdata/influxdb_iox/issues/6336
- [x] https://github.com/influxdata/influxdb_iox/issues/6333
- [x] https://github.com/influxdata/influxdb_iox/issues/6334
- [x] https://github.com/influxdata/influxdb_iox/issues/6335

alamb:
Closing this as we have completed all items here except for 

>  integrate new DataFusion memory limits into ingester2

I think this is a means to an end (a possible way to handle if the ingester2 is hitting OOMs) and we can evaluate using that feature of datafusion if we need to handle memory more gracefully in the ingester2

PR #6337, opened 2022-12-05
feat: Make a mode for the querier to use ingester2 instead, behind the rpc_write feature flag

### LINKED ISSUE #6336: Queriers should get a list of ingester2 gRPC addresses via CLI args
Behind the `rpc_write` feature flag, the querier CLI will have a simpler `--ingester2-addresses` arg than the current `--shard-to-ingesters` config because there aren't shards in the RPC write path design.

This will match the router rpc_write args.

carols10cents:
Closes #6336

This enables a querier to connect to an ingester2 **but does not do any of the catalog cache reloading yet!!!!**

I am, however, able to run a router in RPC write mode with:

```
cargo run --features=rpc_write -- run router-rpc-write --api-bind=127.0.0.1:8081 --ingester-addresses=127.0.0.1:8042
```

And an ingester2 instance with:

```
cargo run --features=rpc_write -- run ingester2 --grpc-bind=127.0.0.1:8042
```

And a querier with: (**NOTE THAT THE INGESTER ADDRESS NEEDS THE `http://` OR IT WON'T WORK RIGHT NOW** does this need to be changed in this PR?)

```
cargo run --features rpc_write -- run querier --ingester-addresses=http://127.0.0.1:8042 --grpc-bind 127.0.0.1:8043  --api-bind 127.0.0.1:8044
```

And I'm pretty sure the querier contacts the ingester2 and gets results, I'm not sure that I have the right persistence settings to ensure I'm not actually seeing data the querier is getting from the object store, but I do see the ingester get a request!

Also note that the `--ingester-addresses` are not accessible or visible without the `--features rpc_write` flag enabled.


crepererum:
> but does not do any of the catalog cache reloading yet!!!!

Is this work tracked somewhere?

querier/src/ingester/mod.rs
```diff
@@ -170,6 +170,42 @@ pub fn create_ingester_connections_by_shard(
...
+pub fn create_ingester2_connection(
+    ingester_addrs: &[String],
+    catalog_cache: Arc<CatalogCache>,
+    open_circuit_after_n_errors: u64,
+) -> Option<Arc<dyn IngesterConnection>> {
+    if ingester_addrs.is_empty() {
+        return None;
+    }
+
+    // This backoff config is used to retry requests for a specific table-scoped query.
```
crepererum, line 183: Copied configs. Every change to the existing code (or to this one) will automatically de-sync this. Please move these configs into a const/lazy/method/whatever.

querier/src/ingester/mod.rs
```diff
@@ -406,6 +442,160 @@ impl IngesterConnectionImpl {
...
+        }
+    }
+}
+
+#[async_trait]
+impl IngesterConnection for IngesterRpcWriteConnection {
+    /// Retrieve chunks from the ingester for the particular table, shard, and predicate
+    async fn partitions(
+        &self,
+        // Shard indexes aren't relevant for the RPC write path. This should always be `None`, but
```
crepererum, line 490: > This should always be `None`

Then please `assert!` this.

querier/src/ingester/mod.rs
```diff
@@ -406,6 +442,160 @@ impl IngesterConnectionImpl {
...
+        expected_schema: Arc<Schema>,
+        span: Option<Span>,
+    ) -> Result<Vec<IngesterPartition>> {
+        let mut span_recorder = SpanRecorder::new(span);
+
+        let metrics = Arc::clone(&self.metrics);
+
+        let measured_ingester_request = |ingester_address: Arc<str>| {
+            let metrics = Arc::clone(&metrics);
+            let request = GetPartitionForIngester {
```
crepererum, line 507: So what's different in this copy-pasta? The construction of the `request`?

crepererum (changes requested):
This is too much copy-paste. In vein of the long duration of this switch-over, this ain't gonna fly (code de-sync, half-fixed bugs, maintenance burden, code complexity). Please abstract the relevant bit. I offer you to help with this if you just point me out to the low-level RPC interface that I shall call (is this still flight-based?).

carols10cents:
> Is this work tracked somewhere?

Indeed it is!! See the linked issues in this tracking issue: https://github.com/influxdata/influxdb_iox/issues/6317

pauldix:
@crepererum one thing to keep in mind is that we're trying to get to the basic working prototype as quickly as possible. Hopefully the duration of the switchover is going to be 4 weeks or less. So that's the scope of how long a little mess is going to exist where we clean things up after.

clap_blocks/src/querier.rs
```diff
@@ -297,8 +343,21 @@ impl QuerierConfig {
...
+    #[cfg(feature = "rpc_write")]
+    pub fn rpc_write(&self) -> bool {
+        true
+    }
+
+    /// Whether the querier is contacting ingesters that use the RPC write path or not.
+    #[cfg(not(feature = "rpc_write"))]
+    pub fn rpc_write(&self) -> bool {
+        false
+    }
```
carols10cents, line 357: I was going to have the switch be only based on which variant of the `IngesterAddresses` enum that `ioxd_querier` gets, but the querier also needs to know whether to create a sharder using the shards in the catalog or not, so this boolean is for that purpose.

querier/src/database.rs
```diff
@@ -128,9 +130,13 @@ impl QuerierDatabase {
...
-        let sharder = Arc::new(
-            create_sharder(catalog_cache.catalog().as_ref(), backoff_config.clone()).await?,
-        );
+        let sharder = if rpc_write {
+            None
+        } else {
+            Some(Arc::new(
+                create_sharder(catalog_cache.catalog().as_ref(), backoff_config.clone()).await?,
+            ))
+        };
```
carols10cents, line 139: This is where the sharder either needs to be created or not, but the ingester connections have already been configured, so this needed the boolean.

querier/src/ingester/mod.rs
```diff
@@ -161,13 +162,29 @@ pub fn create_ingester_connections_by_shard(
...
+            circuit_breaker_backoff_config,
+            open_circuit_after_n_errors,
+        )),
+        (None, Some(ingester_addresses)) => Arc::new(IngesterConnectionImpl::by_addrs(
+            ingester_addresses,
+            catalog_cache,
+            retry_backoff_config,
+            circuit_breaker_backoff_config,
+            open_circuit_after_n_errors,
+        )),
```
carols10cents, line 186: This is a consequence of sharing the most code and having some irrelevant parts carried along in the types-- to me, this is "messier", because it'll be more involved to remove the old code, but the overall amount of copy-pasted code to stand up a querier is indeed less. Tradeoffs everywhere! 🤷🏻‍♀️

querier/src/ingester/mod.rs
```diff
@@ -726,20 +773,57 @@ impl IngesterConnection for IngesterConnectionImpl {
...
+                                return ShardNotMappedSnafu {
+                                    shard_index: *shard_index,
+                                }
+                                .fail()
+                            }
+                        },
+                    }
+                }
+                relevant_ingester_addresses
+            }
```
carols10cents, line 824: The inside of this arm is moved without changes from below.

carols10cents:
@crepererum @pauldix This is ready for re-review! Please note in the description that the command for the querier has changed-- there is no `querier-rpc-write` command anymore, just `querier`.

crepererum (approved):
Looks good, thanks a lot for the clean up :+1:

PR #6343, opened 2022-12-06
fix: support InfluxRPC OR-chains w/ arbitrary child nodes

crepererum:
Also convert another assertion regarding child nodes of Eq-nodes into a proper error.

See https://github.com/influxdata/idpe/issues/16582 .


service_grpc_influxrpc/src/expr.rs
```diff
@@ -1232,6 +1248,92 @@ mod tests {
         assert!(predicate.range.is_none());
     }
 
+    #[test]
+    fn test_convert_predicate_multiple_field_selection_flat_node1() {
```
alamb, line 1252: I think we should have test coverage for:
1.  multiple `AND` nodes
2. Single child `AND`  and `OR` node (should error probably)

service_grpc_influxrpc/src/expr.rs
```diff
@@ -381,30 +382,35 @@ impl InListBuilder {
...
         // lhs OR rhs
         else if Some(RPCValue::Logical(RPCLogical::Or as i32)) == node.value {
-            assert_eq!(node.children.len(), 2);
-
-            let lhs = &node.children[0];
-            let rhs = &node.children[1];
-
-            // recurse down both sides
-            self.append(lhs).and_then(|s| s.append(rhs))
+            node.children
```
alamb, line 401: should we really allow only one child?

alamb (approved):
I think this code is an improvement, though I think it could use some additional error checking / test cases

service_grpc_influxrpc/src/expr.rs
crepererum: done

service_grpc_influxrpc/src/expr.rs
```diff
@@ -381,30 +382,35 @@ impl InListBuilder {
...
+    fn append(self, node: &RPCNode) -> Result<Self, String> {
         // lhs = rhs
         if Some(RPCValue::Comparison(RPCComparison::Equal as i32)) == node.value {
+            if node.children.len() != 2 {
+                return Err(format!(
+                    "Eq nodes should have 2 children but found {}",
+                    node.children.len()
+                ));
+            }
             assert_eq!(node.children.len(), 2);
```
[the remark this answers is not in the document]
crepererum: done

PR #6345, opened 2022-12-07
fix: account for memory allocations in InfluxRPC group outputs

crepererum:
This should prevent the querier from OOMing.

See https://github.com/influxdata/idpe/issues/16614 .


iox_query/src/exec/seriesset/converter.rs
```diff
@@ -491,32 +502,29 @@ impl Stream for SeriesSetConverterStream {
...
+    pub fn new(group_columns: Vec<Arc<str>>, memory_manager: Arc<MemoryManager>) -> Self {
+        Self {
+            group_columns,
+            memory_manager,
+        }
     }
 
     /// groups the set of `series` into SeriesOrGroups
     ///
     /// TODO: make this truly stream-based
```
alamb, line 518:
```suggestion
    /// TODO: make this truly stream-based
    /// https://github.com/influxdata/influxdb_iox/issues/6347
```

I filed a ticket to track the truly streaming solution. I don't think we should pursue it unless we get people explicitly asking for it / hitting this ResourcesExhausted limit

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -658,12 +666,170 @@ impl SortableSeries {
...
+    /// Memory usage in bytes, including `self`.
+    fn size(&self) -> usize {
+        std::mem::size_of_val(self) + self.series.size() - std::mem::size_of_val(&self.series)
+            + (std::mem::size_of::<Arc<str>>() * self.tag_vals.capacity())
+            + self.tag_vals.iter().map(|s| s.len()).sum::<usize>()
+    }
+}
+
+/// [`Future`] that collects [`Series`] objects into a [`SortableSeries`] vector while registering/checking memory
+/// allocations with a [`MemoryManager`].
```
alamb, line 688:
```suggestion
/// allocations with a [`MemoryManager`].
///
/// This avoids unbounded memory growth when merging multiple `Series` in memory
```

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -658,12 +666,170 @@ impl SortableSeries {
...
+            inner,
+            group_columns,
+            collected: Vec::with_capacity(0),
+            buffered_size: 0,
+            mem_proxy: Some(mem_proxy),
+            mem_proxy_alloc_fut: None,
+        }
+    }
+
+    /// Start a [`MemoryConsumerProxy::alloc`] future.
```
alamb, line 748: BTW @tustvold  proposes removing `MemoryConsumerProxy` in https://github.com/apache/arrow-datafusion/pull/4522

though I think if that PR is merged this whole logic will become simpler

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -658,12 +666,170 @@ impl SortableSeries {
...
+            match ready!(this.inner.poll_next_unpin(cx)) {
+                Some(Ok(series)) => match SortableSeries::try_new(series, &this.group_columns) {
+                    Ok(series) => {
+                        // Note: the size of `SortableSeries` itself is already included in the vector allocation
+                        this.buffered_size += series.size() - std::mem::size_of_val(&series);
+                        this.collected
+                            .push_accounted(series, &mut this.buffered_size);
+
+                        // should we clear our allocation buffer?
+                        if this.buffered_size > 1024 * 1024 {
```
alamb, line 806: I recommend making the 1MB a named constant (so it can be documented as the incremental allocation size)

iox_query/src/exec/seriesset/converter.rs
```diff
@@ -658,12 +666,170 @@ impl SortableSeries {
...
+                            this.alloc();
```
alamb, line 807: the rationale that `alloc()` will not panic is that it there was an outstanding future it would have been handled by the 

```rust
            if let Some(fut) = this.mem_proxy_alloc_fut.as_mut() {
```
clause above?

alamb (approved):
Shall we say this closes https://github.com/influxdata/idpe/issues/16614 ?

I think this is a reasonable solution to the immediate problem. 👍  thank you @crepererum

iox_query/src/exec/seriesset/converter.rs
crepererum: I know, I just don't wanted to wait til this lands in IOx.

iox_query/src/exec/seriesset/converter.rs
crepererum: Yeah, `alloc` itself won't panic, but the `ready!(fut.poll_unpin(cx))` block in the upper section of this function can. But TBH: if you really poll a future after a panic, you cannot expect reasonable behavior and this will just panic because the internal state is so messed up. This happens to many IOx methods BTW. (which is also why I'm a big fan of lock poisoning).

PR #6346, opened 2022-12-07
feat: allow enabling/disabling ns autocreation in router

lukebond:

Before this PR one could set the retention period (infinite or otherwise) for namespaces when they are autocreated.

After this PR one can enable/disable namespace autocreation entirely via a CLI arg / envvar. It is enabled by default, so no behaviour change without some config change in k8s-idpe.

I did a little refactoring along the way because the retention arg was duplicated in two clap blocks. It's now only in the generic router clap block, plus this new boolean enable/disable arg. 


jacobmarble (approved):

carols10cents:
> I did a little refactoring along the way because the retention arg was duplicated in two clap blocks. It's now only in the generic router clap block, plus this new boolean enable/disable arg.

Hey, this is my fault for not having tests for the router-rpc-write command yet, but this refactoring caused problems because some of the clap args in RouterRpcConfig and RouterConfig have the same names:

```
$ cargo run --features rpc_write -- run router-rpc-write --help
thread 'main' panicked at 'Command router-rpc-write: Argument names must be unique, but 'query_pool_name' is in use by more than one argument or group', /Users/carolnichols/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-4.0.29/src/builder/debug_asserts.rs:95:13
```

[I have a fix for this that duplicates the new arg onto RouterRpcWriteConfig instead](https://github.com/influxdata/influxdb_iox/pull/6355).

 I did this on purpose so that it will be easier to delete the whole RouterConfig file when we switch over to the RPC write path, rather than having to pick through and remove unused pieces, but yes, it is going to be a bit of duplication for a while.

Just wanted you to know!

PR #6348, opened 2022-12-07
Wasm uda

nathanielc:
POC of WASM + Datafusion UDAs


alamb:
It is not clear how we are proceeding with this feature, so closing this PR for now

PR #6349, opened 2022-12-07
chore: Upgrade datafusion to get correct support for multi-part identifiers

alamb:
Resolves https://github.com/influxdata/idpe/issues/16612 (customer reported issue querying measurements with periods in names via SQL).

# Rationale
1. Get https://github.com/apache/arrow-datafusion/pull/4530 and various other DataFusion improvements

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/fbadebb894672f61327a30f77cda2ee88a343b2a
2. Update explain plans for plan output
2. Add regression tests for sql querying of identifiers with periods


parquet_to_line_protocol/src/lib.rs
```diff
@@ -16,6 +15,7 @@ use datafusion::{
     },
     prelude::{SessionConfig, SessionContext},
 };
+use datafusion_util::config::iox_session_config;
```
alamb, line 18: I cleaned up this code to use the same configuration settings that are used elsewhere in IOx

querier/src/namespace/query_access.rs
```diff
@@ -487,15 +487,15 @@ mod tests {
...
+                "|               |   TableScan: cpu projection=[foo, host, load, time]                                                                                                                                                                                                                                                                                                                   |",
+                "| physical_plan | ProjectionExec: expr=[foo@0 as foo, host@1 as host, load@2 as load, time@3 as time]                                                                                                                                                                                                                                                                                   |",
+                "|               |   ParquetExec: limit=None, partitions={1 group: [[1/1/1/1/<uuid>.parquet, 1/1/1/1/<uuid>.parquet, 1/1/1/1/<uuid>.parquet, 1/1/2/2/<uuid>.parquet, 1/1/1/3/<uuid>.parquet]]}, projection=[foo, host, load, time] |",
```
alamb, line 496: All the explain plans have the same difference due to improved display of partitioned files added in  https://github.com/apache/arrow-datafusion/pull/4467. The actual plans are not changed

iox_query/src/provider.rs
```diff
@@ -221,7 +221,7 @@ impl TableProvider for ChunkTableProvider {
     async fn scan(
         &self,
         _ctx: &SessionState,
-        projection: &Option<Vec<usize>>,
+        projection: Option<&Vec<usize>>,
```
alamb, line 224: API became more idiomatic in https://github.com/apache/arrow-datafusion/pull/4458

Cargo.lock
```diff
@@ -5075,6 +5094,25 @@ dependencies = [
  "unicode_categories",
 ]
 
+[[package]]
+name = "sqllogictest"
```
alamb, line 5098: 🤔  this probably doesn't need to be a dependency of datafusion -- will fix upstream

query_tests/cases/in/duplicates_parquet.expected
```diff
@@ -91,8 +91,8 @@
...
+|    |             ParquetExec: limit=None, partitions={2 groups: [[1/1/1/1/00000000-0000-0000-0000-000000000002.parquet], [1/1/1/1/00000000-0000-0000-0000-000000000003.parquet]]}, predicate=state = Dictionary(Int32, Utf8("MA")), pruning_predicate=state_min@0 <= MA AND MA <= state_max@1, projection=[area, city, max_temp, min_temp, state, time], metrics=[bytes_scanned=1219, elapsed_compute=1.234ms, mem_used=0, num_predicate_creation_errors=0, output_rows=5, page_index_eval_time=1.234ms, page_index_rows_filtered=0, predicate_evaluation_errors=0, pushdown_eval_time=1.234ms, pushdown_rows_filtered=5, row_groups_pruned=0, spill_count=0, spilled_bytes=0, time_elapsed_opening=1.234ms, time_elapsed_processing=1.234ms, time_elapsed_scanning=1.234ms]    |
```
alamb, line 96: here is a good example of how the new display is easier to see -- it shows that there are two groups (each read concurrently) rather than one big blob

query_tests/cases/in/periods.expected
```diff
@@ -0,0 +1,21 @@
+-- Test Setup: PeriodsInNames
```
alamb, line 1: Here is the new regression test for SQL on measurement names and identifiers with periods

alamb (commented):
https://github.com/influxdata/influxdb_iox/pull/6349/files?w=1 shows the differences more easily I think

Cargo.lock
alamb: https://github.com/apache/arrow-datafusion/pull/4544

crepererum (approved):

PR #6351, opened 2022-12-08
fix: `GroupGenerator`/`Converter` panic

crepererum:
Do not poll a ready future.


iox_query/src/exec/seriesset/converter.rs
```diff
@@ -786,6 +814,7 @@ where
                 let (mem_proxy, res) = ready!(fut.poll_unpin(cx));
                 assert!(this.mem_proxy.is_none());
                 this.mem_proxy = Some(mem_proxy);
+                this.mem_proxy_alloc_fut = None;
```
crepererum, line 817: This is the actual fix. The rest is just to allow for easier testing.

alamb (approved):

PR #6352, opened 2022-12-08
test: "not found" end2end tests for querier

crepererum:
I couldn't find any end2end tests for these cases and I was kinda worried that our error codes were wrong. Turns out they are correct, but let's have some nice tests for this behavior.


influxdb_iox/tests/end_to_end_cases/querier.rs
```diff
@@ -665,13 +740,7 @@ async fn oom_protection() {
...
                         .await
                         .unwrap_err();
-                    assert_eq!(
-                        status.code(),
-                        tonic::Code::ResourceExhausted,
-                        "Wrong status code: {}\n\nStatus:\n{}",
-                        status.code(),
-                        status,
-                    );
+                    check_tonic_status(status, tonic::Code::ResourceExhausted, None);
```
alamb, line 743: nice

alamb (approved):
Looks great to me -- thanks @crepererum

PR #6353, opened 2022-12-08
refactor: add additional span around chunk spans

crepererum:
# Before
image

# After
image

image


querier/src/table/mod.rs
```diff
@@ -330,6 +330,9 @@ impl QuerierTable {
         // create parquet files
         let parquet_files: Vec<_> = match cached_table {
             Some(cached_table) => {
+                // use nested scope because we span many child scopes here and it's easier to digest
```
alamb, line 333:
```suggestion
                // use nested scope because we span many child scopes here and it's easier to 
                // aggregate / collapse in the UI
```

alamb (approved):
Thank you! I swear I filed a ticket about this exact request, but I can't find it now

crepererum:
> I swear I filed a ticket about this exact request, but I can't find it now

Ditto :sweat_smile:

PR #6354, opened 2022-12-08
feat: rate-limit Jaeger UDP messages

### LINKED ISSUE #5446: Throttle Jaeger tracing exporter
# Why
This is related to https://github.com/influxdata/influxdb_iox/issues/5445. We often see that prod drops tracing spans and the Jaeger UDP protocol provides us no way to detect this. We should be a better citizen and not overload Jaeger.

# What
Limit the tracing Jaeger exporter by packages/s and/or bytes/s.

# How
The tracing spans are queued in a bound/limited queue anyways and we even log a warning when the queue is full (we don't block in this case, we just throw away the span). Not sure how to come up with reasonable limits for packages/s and bytes/s though (maybe we find some prod network measure that provides us a useful hint).

crepererum:
The Jaeger UDP protocol provides no way to signal backpressure / overload. In certain situations, we are emitting that many tracing spans in a short period of time that the OS, the network, or Jaeger drop them. While a rate limit is not a perfect solution, it for sure helps a lot (tested locally).

Note that the limiter does NOT lead to unlimited buffering because we already have a limited outbox queue in place (see
`trace_exporters::export::CHANNEL_SIZE`).

Fixes #5446.


trace_exporters/src/export.rs
```diff
@@ -64,10 +77,16 @@ impl TraceCollector for AsyncExporter {
...
         match self.sender.try_send(Some(span)) {
             Ok(_) => {
+                // sending worked again, so re-enable warning
+                self.warned_sender_full.store(false, Ordering::SeqCst);
+
                 //TODO: Increment some metric (#2613)
             }
             Err(TrySendError::Full(_)) => {
-                warn!("exporter cannot keep up, dropping spans")
+                // avoid spamming the log system (there might be thousands of traces incoming)
```
alamb, line 86: 👍

trace_exporters/src/export.rs
```diff
@@ -64,10 +77,16 @@ impl TraceCollector for AsyncExporter {
         use mpsc::error::TrySendError;
         match self.sender.try_send(Some(span)) {
             Ok(_) => {
+                // sending worked again, so re-enable warning
+                self.warned_sender_full.store(false, Ordering::SeqCst);
```
alamb, line 81: this is nice to reset it

alamb (approved):
Looks like a reasonable change to me -- 

I don't think this PR actually "rate limits" the warnings as the log could still get many warnings if one trace works and then a subsequent doesn't and then another works, ...

However, this code seems like it is a definite improvement (it will not cause *more* messages)

👍

crepererum:
> I don't think this PR actually "rate limits" the warnings as the log could still get many warnings if one trace works and then a subsequent doesn't and then another works, ...

Yeah, if you flip-flow between open and closed state of the warning circuit (atomic bool), you don't rate limit. I think it's OK for now.

PR #6357, opened 2022-12-08
feat: Identify each run of an ingester with a Uuid 

### LINKED ISSUE #6333: Ingester2 should generate a UUID for itself on startup
And send its UUID in the response to querier requests.

carols10cents:
~~**Blocked by/contains https://github.com/influxdata/influxdb_iox/pull/6355**, merge that first please!~~ Done!

And send that UUID in the Flight response for queries to that ingester run.

Fixes #6333.

Please see each commit for more explanation of the test setup.


carols10cents:
Ok @pauldix I think this is really ready for review now-- this should also add a new, not-yet-required job to CI to start running a few tests against the rpc_write path.

domodwyer:
Hey @carols10cents - I plan on looking at this today :+1:

ingester2/src/init.rs
```diff
@@ -157,6 +157,25 @@ pub async fn new(
...
+        .await
+        .expect("get topic")
+        .unwrap();
+    let s = txn
+        .shards()
+        .create_or_get(&topic, TRANSITION_SHARD_INDEX)
+        .await
+        .expect("create transition shard");
+    assert_eq!(s.id, TRANSITION_SHARD_ID);
+    txn.commit().await.expect("commit transition shard");
```
domodwyer, line 177: I am unsure if this might have implications during deployment - what needs it? tests?

carols10cents: Right now there's a database constraint that partitions have a shard_id foreign key in the `shard` table:

```
\d+ partition
                                                               Table "iox_catalog.partition"
          Column           |       Type        | Collation | Nullable |           Default            | Storage  | Compression | Stats target | Description
---------------------------+-------------------+-----------+----------+------------------------------+----------+-------------+--------------+-------------
 id                        | bigint            |           | not null | generated always as identity | plain    |             |              |
 shard_id                  | bigint            |           | not null |                              | plain    |             |              |
 table_id                  | bigint            |           | not null |                              | plain    |             |              |
 partition_key             | character varying |           | not null |                              | extended |             |              |
 sort_key                  | text[]            |           | not null |                              | extended |             |              |
 persisted_sequence_number | bigint            |           |          |                              | plain    |             |              |
 to_delete                 | bigint            |           |          |                              | plain    |             |              |
Indexes:
    "partition_pkey" PRIMARY KEY, btree (id)
    "partition_key_unique" UNIQUE CONSTRAINT, btree (table_id, partition_key)
Foreign-key constraints:
    "partition_sequencer_id_fkey" FOREIGN KEY (shard_id) REFERENCES shard(id) NOT VALID
```

so if [we insert partitions with the transition shard ID](https://github.com/influxdata/influxdb_iox/blob/e49ffc02f8e51dba70ca2a10cd0d29ee88da3d3c/ingester2/src/buffer_tree/partition/resolver/catalog.rs#L51), the transition shard ID needs to exist in the `shard` table. 

Is there a better way to fix this both for the transition period and after? Maybe allowing NULL for partition.shard_id and not use TRANSITION_SHARD_ID at all?

Issue #6358, opened 2022-12-09
Faster query planning


crepererum:
Let's shave off a few milliseconds from the query planning phase. This is mostly:

- avoid a few cache requests that we don't really need
- avoid string processing if we can (and use IDs instead)

crepererum:
Brain dump of outstanding tasks:

- **cache lookup grouping:** Cache lookups are not super slow but also not exactly super cheap since they involve the whole cache system maintenance like LRU. Change parquet chunk creation to lookup partition sort key and interned/cached/projected schema once per cache key instead of once per parquet chunk.
- **concurrency during ingester request:** During ingester RPC, we already fetch the parquet file cache value. The resulting value may be outdated (and we don't invalidate it til we have a watermark from the ingester; we also don't wanna invalidate unconditionally) but at least for the parquet chunks that we already know, we can create parquet chunks (and while doing so perform the per-partition cache lookups). It's unlikely that the new chunks that we discover after the ingester RPC actually span a new relevant partition (this probably only happens once an hour).
- **change partition sort key cache:** We may want to fetch partition sort keys per table, not per partition. If someone aggregates a year of compacted data, we would otherwise end up with nearly 10k (24*365=8760) catalog requests for a cold-cache querier. On eu-central-1, we have one querier that has 200k cached entries just for this data type (for <20k object store files that it uses).

alamb:
Thank you for this list @crepererum  -- from my perspective (and it is only one perspective) I think the current querier planning path is quite fast ("good enough for now").

Shaving off a few more milliseconds when the tall pole for "fully cached" planning is querier/ingester communication will have diminishing benefits.

Here are some performance I think would make a more direct impact to customers in the short term:
1. Faster metadata queries https://github.com/influxdata/influxdb_iox/issues/6315
3. Faster influxrpc performance: https://github.com/influxdata/idpe/issues/16602 (reported by customers)
1. Faster "from cold" end to end query (aka time to results from when the caches are empty) - partly covered by (https://github.com/influxdata/influxdb_iox/issues/5892)

crepererum:
> 3\. Faster "from cold" end to end query (aka time to results from when the caches are empty) - **partly covered by #5892**

[...]

PR #6359, opened 2022-12-09
refactor: avoid string-hashing of parquet file column names

crepererum:
For #6358.


querier/src/chunk/mod.rs
```diff
@@ -227,7 +227,7 @@ impl ChunkAdapter {
     ) -> Option<ChunkParts> {
         let span_recorder = SpanRecorder::new(span);
 
-        let parquet_file_cols: HashMap<&str, ColumnId> = parquet_file
+        let parquet_file_cols: HashMap<ColumnId, &str> = parquet_file
```
crepererum, line 230: So I've moved some processing into a per-table cache instead of doing it per-chunk here.

alamb (approved):
Makes sense to me 👍 

I think a PR title of "cache primary key column name to column id" would also be correct as the basic idea here is to cache the translation once rather than per query 👍

PR #6360, opened 2022-12-09
refactor: store reverse column ID map for cached tables

crepererum:
~On top of #6359.~

For #6358.

Doesn't do anything on it's own, but I need it to avoid some strings in this cache request here (which is per-chunk):

https://github.com/influxdata/influxdb_iox/blob/0221820123893ca43737ea9fa4fc03d00b7fe7fc/querier/src/cache/partition.rs#L120

Pulled this out to simplify the review process.


alamb (approved):
I reviewed the second commit https://github.com/influxdata/influxdb_iox/pull/6360/commits/35236f6dfc0ff1fe7d393d0daeed94007f4d0163 and it looks good to me 👍

PR #6361, opened 2022-12-09
refactor: re-use cached table

crepererum:
On top of #6360.

Saves a cache request but also allows us to pass down `CachedTable` to some other subroutines that can use more cached/pre-computed state.

For #6358.


crepererum:
This is not going to fly because we need to refresh the cached partition after we learn about new columns from the ingester (or from any of the parquet files).

PR #6365, opened 2022-12-09
feat: Keep track of & report # of Parquet files persisted, invalidate querier cache

### LINKED ISSUE #6334: Ingester2 instances should track number of parquet files persisted to object storage per partition
And send this number in the response to querier requests.

carols10cents:
~~**This builds on https://github.com/influxdata/influxdb_iox/pull/6357, please merge that one first so I don't have to un-resolve and re-resolve merge conflicts with myself! :)**~~ Done!

Per partition and starting over each time the ingester restarts.

Fixes #6334.
Fixes #6335.

I highly recommend reviewing commit-by-commit.

There are some unit tests; I tried to write some integration tests but I started to get bogged down because ingester2 doesn't implement the write info service yet, so there's no way for integration tests to check the status of a write (such as whether it's been persisted or not). So I decided to push it for now.


domodwyer (approved):
Looks like what I would expect it to look like! Thank you 🙏

alamb:
cc @crepererum

PR #6368, opened 2022-12-11
Fixes persistence in ingester2, simplifies and optimizes WAL

pauldix:
Just getting this up so @domodwyer can have a look.

The persistence mechanism in ingester2 was broken. It would kick off a persist but then get locked after a few. I found this pretty quickly testing with the data generator. I simplified things without the hashing as there's no need for that with the new write semantics. We can handle sort key update conflicts by re-doing persist jobs if we see it afterwards.

I simplified the WAL significantly and made it actually batch ops into the WAL. I just fixed it to a 5ms timer to check, but we can play around with that. 

I also made some updates to the data generator so it outputs more information and can also deal with failed requests and keep running.

Obviously this all needs tests and I need to also fix the replay logic (which should be pretty quick for me to do tomorrow).

But overall the performance on this looks pretty good so far. Tested with 100 agents simulating load writing data as fast as they can request by request with each write request having 8k rows (lines). Some info:

```
made 60100 requests at 165/sec with 480800000 rows at 1324517/sec in 363.249068s with 0 errors for a 0.00 error rate

This is ~346 MB/sec total being sent from 100 agents in batches of 8k lines per request into 20 separate measurements. Here's a sample line:
main_measurement_1,agent_id=1,child_tag=id_1_n4vSSmcQpC,extra_static=whatever-constant-value,measurement_id=1,rotation=id_1_b8a51a36-d6da-4881-8021-b36e2c565644,some_tag=id_1_lWU7DsmbV8MOdU4 intfield=13633i,floatfield=0.25732639401898805 1670776639102032000
```


ingester2/src/persist/queue.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+impl PersistQueue {
+    pub fn new(exec: Arc<Executor>, store: ParquetStorage, catalog: Arc<dyn Catalog>, worker_count: usize) -> Self {
+        let inner = Arc::new(Inner {
+            exec,
+            store,
+            catalog
+        });
+        let (tx, rx) = unbounded();
```
domodwyer, line 38: > The persistence mechanism in ingester2 was broken. It would kick off a persist but then get locked after a few. I found this pretty quickly testing with the data generator.

The only practical difference I can see between this code, and the existing code is that this uses unbounded channels - I'm not sure if holding an unbounded amount of data in memory is feasible, but even if it is, simply changing the existing queue depth to a high value should have provided the same effect. The current default gives us a queue depth of 10 per partition, which sounds like it matches with the "broken" behaviour you saw.

> I simplified things without the hashing as there's no need for that with the new write semantics. We can handle sort key update conflicts by re-doing persist jobs if we see it afterwards.

Because we can accurately and cheaply determine if a sort key update is necessary on a per-persist-task basis before running the persistence op, the existing code structure allows us to parallelise all other persist tasks for a given partition (which I expect to be the majority) and serialise only those that require it for correctness - it currently serialises all persist ops for a partition in the MVP. This will avoid the "do it again" method proposed in the description - persisting is easily the most expensive op we run so ideally we avoid redoing them at all.

I can implement parallel partition persistence for sort keys that do not need updating easily, with the existing code structure, today if necessary for the MVP.

domodwyer (commented):
I reviewed the changes to design, and not the code itself - I haven't looked at the WAL implementation yet.

ingester2/src/wal/rotate_task.rs
```diff
@@ -1,76 +1,38 @@
...
-        // special code path between "hot partition persist" and "wal rotation
-        // persist" - it all works the same way!
-        //
-        // TODO: this properly as described above.
-
-        tokio::time::sleep(Duration::from_secs(5)).await;
+        // ensure that all buffered writes have been flushed. The buffer tree waits for this
+        // notification after writes so all buffered writes should exist in the tree, making
+        // it safe to persist the data and be sure that the WAL segment can be removed
+        wal.notify_on_flush().changed().await.expect("wal flush failed");
```
domodwyer, line 35: I am not sure the issues in the deleted comment have been resolved, and I remain convinced this approach is flawed. As an example of a write losing the durable WAL entry; with `T1` being a thread handling a write, and `T2` being the WAL rotate / persist task:

| T1                                            | T2                                                                                           |
| --------------------------------------------- | -------------------------------------------------------------------------------------------- |
| Write to WalSink                              |                                                                                              |
| Apply to WAL successfully                     |                                                                                              |
| Call BufferTree::apply() and get parked[^1]   |                                                                                              |
|                                               | Rotate the WAL                                                                               |
|                                               | Drain the buffer, read partition data for persistence - **this does not contain T1's write** |
|                                               | Finish persistence, drop the "persisted" WAL file                                            |
| Unpark and apply the write to the buffer tree |                                                                                              |

At this point the write is in the buffer tree and queryable, was ACK'd to the user as durable, does not exist in the persisted data in S3, and has no WAL entry - a crash will cause this write to disappear. This is an extreme timing variation to show the issue at hand, but other, more likely timing/scheduling variations exist.

This strategy of timing the WAL rotation to accurately determine when WAL segments can be dropped becomes increasingly complex when you consider the planned "hot partition" persistence, which I assume happens in parallel with writes and WAL rotations (otherwise a sufficiently hot partition would starve either, let alone the latency/req-per-sec implications of serialising them).

This complexity is furthered when you consider that persist operations can be re-ordered after being enqueued - just because a persist task is enqueued first, doesn't mean a worker finishes it first.

I believe the only robust method to ensure WAL segments are retained as long as there's unpersisted data in the system is to reference count the segments and drop them when there is no unpersisted data within them. This works regardless of ordering, parallelism, method, or reason for persistence and can be implemented simply & cheaply too.

[^1]: Applying a write to the buffer tree involves querying the catalog once per partition. This causes the write task to be parked while the I/O happens asynchronously.

domodwyer (commented):
GitHub lost my other comment! 🔥

domodwyer:
@pauldix I rebased this onto the (yet to be merged) changes in #6377 that adds support for benchmarking WAL replay. Currently it is stupid simple - it writes 1 line of LP and then replays it - adjust as needed :+1:

The benchmark fails with what looks like a legit error though:

```console
% cargo bench -p ingester2 --features=benches
Benchmarking replay/1 write: Warming up for 3.0000 sthread 'main' panicked at 'WAL replay error: ReadEntry(UnableToReadNextOps { source: UnableToDeserializeData { source: DecodeError { description: "invalid wire type: LengthDelimited (expected Varint)", stack: [("SequencedWalOp", "sequence_number")] } } })', ingester2/benches/wal.rs:62:22
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   5: tokio::runtime::park::CachedParkThread::block_on
   6: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
   7: criterion::bencher::AsyncBencher<A,M>::iter_batched
   8: <criterion::routine::Function<M,F,T> as criterion::routine::Routine<M,T>>::warm_up
   9: criterion::routine::Routine::sample
  10: criterion::analysis::common
  11: criterion::benchmark_group::BenchmarkGroup<M>::bench_function
  12: wal::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

error: bench failed, to rerun pass `-p ingester2 --bench wal`
```

I also fixed the test compilation along the way! :+1:

pauldix:
The bits of this were pulled out into separate smaller PRs.

PR #6374, opened 2022-12-12
refactor: use column IDs for partition cache invalidation

crepererum:
This shall avoid a bunch of string hashing during query planning.

For #6358.


querier/src/chunk/mod.rs
```diff
@@ -93,7 +93,7 @@ pub struct QuerierChunk {
     delete_predicates: Vec<Arc<DeletePredicate>>,
 
     /// Partition sort key (how does the read buffer use this?)
-    partition_sort_key: Arc<Option<SortKey>>,
+    partition_sort_key: Option<Arc<SortKey>>,
```
crepererum, line 96: I took the opportunity to change the weird `Arc<Option<...>>` into a somewhat saner `Option<Arc<...>>` (the `None`-case doesn't need to be `Arc`ed).

querier/src/ingester/mod.rs
```diff
@@ -580,7 +577,7 @@ struct IngesterStreamDecoder {
     current_chunk: Option<(Schema, Vec<RecordBatch>)>,
     ingester_address: Arc<str>,
     catalog_cache: Arc<CatalogCache>,
-    expected_schema: Arc<Schema>,
+    cached_table: Arc<CachedTable>,
```
crepererum, line 580: we now pass a `CachedTable` around in a few more places instead of freestanding objects like `Schema`. This has the advantage that we can use that for a few other "lookup tables".

querier/src/table/mod.rs
```diff
@@ -487,15 +483,17 @@ impl QuerierTable {
             .as_ref()
             .map(|sharder| vec![**sharder.shard_for_query(&self.table_name, &self.namespace_name)]);
 
+        // get cached table w/o any must-coverage information
```
alamb, line 486: the formatting of this line seems quite strange but I suppose if `rustfmt` passes then 🤷

querier/src/ingester/mod.rs
```diff
@@ -632,18 +629,27 @@ impl IngesterStreamDecoder {
             let primary_keys: Vec<_> = schemas.iter().map(|s| s.primary_key()).collect();
             let primary_key: Vec<_> = primary_keys
                 .iter()
-                .flat_map(|pk| pk.iter().copied())
+                .flat_map(|pk| pk.iter())
+                // cache may be older then the ingester response status, so some entries might be missing
```
alamb, line 633: Is this a required change as part of this PR or is it a "drive by cleanup"? 

Or put another way, I don't understand why this additional filtering is now required 🤔

alamb (approved):
LGTM -- I had one question, but I don't see any reason not to merge the PR 👍

querier/src/ingester/mod.rs
crepererum: It's mostly because we need to map names back to IDs and this wasn't required before. So we now rely on another synced state (the schema) for this particular bit of code. The columns would be filtered out by DF anyways because they cannot be part of the schema (and the schema is the first thing that the whole query planning actually inspects, before creating any chunk or pulling ingester data).

PR #6375, opened 2022-12-12
refactor: faster sort key calculation

crepererum:
~On top of #6374.~

Avoid nasty string lookups to determine which columns make a parquet's
sort key.

For https://github.com/influxdata/influxdb_iox/issues/6358.


alamb (approved):
Makes sense to me

PR #6376, opened 2022-12-12
refactor: improve concurrency during parquet chunk creation

crepererum:
~On top of #6375.~

See commits.

For #6358.


querier/src/table/mod.rs
```diff
@@ -367,19 +368,27 @@ impl QuerierTable {
...
+                        None
                     }
-
-                    let keep = *keep;
-                    async move { keep }
                 })
-                .map(|(cached_parquet_file, _keep)| {
+                .collect::<Vec<_>>();
+
+            // de-correlate parquet files so that subsequent items likely don't block/wait on the same cache lookup
```
alamb, line 386: that is interesting 🤔 

I always worry about injecting randomness into a system because it makes it potentially harder to reproduce symptoms -- like what if sometimes the order is good and sometimes bad?

Some ideas for consideration
1. Make this approach deterministic is maybe shuffling using a random number generator seeded with a hash of the query txt or something. 
2. Remove the shuffle entirely -- given 100 parquet files are potentially being fetched concurrently, will you really actually start more catalog queries by shuffling requests? It seems more likely to me that you'll already be at the max concurrent catalog query limit so adding more concurrent request will just make the catalog queue deeper

querier/src/table/mod.rs
```diff
@@ -367,19 +368,27 @@ impl QuerierTable {
...
                 &MetricPruningObserver::new(Arc::clone(&self.prune_metrics));
 
-            futures::stream::iter(parquet_files.files.iter().cloned().zip(keeps))
-                .filter(|(cached_parquet_file, keep)| {
-                    if !keep {
-                        early_pruning_observer.was_pruned_early(
-                            cached_parquet_file.row_count as u64,
-                            cached_parquet_file.file_size_bytes as u64,
-                        );
+            let mut parquet_files = parquet_files
```
alamb, line 371:
```suggestion
            // Remove any unused parquet files up front to maximize the 
            // concurrent catalog requests that could be outstanding
            let mut parquet_files = parquet_files
```

alamb (approved):
I have a question about the file shuffle, but nothing that would prevent this PR from merging. thank you @crepererum

querier/src/table/mod.rs
crepererum: > Make this approach deterministic is maybe shuffling using a random number generator seeded with a hash of the query txt or something.

:+1: 

> Remove the shuffle entirely -- given 100 parquet files are potentially being fetched concurrently, will you really actually start more catalog queries by shuffling requests? It seems more likely to me that you'll already be at the max concurrent catalog query limit so adding more concurrent request will just make the catalog queue deeper

With the new chatty write path (and the compactor backfill issues), I'm afraid it is very much possible that the first/last 100 files are from the same partition.

crepererum: done

querier/src/table/mod.rs
```diff
@@ -387,7 +389,8 @@ impl QuerierTable {
 
             // de-correlate parquet files so that subsequent items likely don't block/wait on the same cache lookup
             // (they are likely ordered by partition)
-            parquet_files.shuffle(&mut rand::thread_rng());
+            let mut rng = StdRng::seed_from_u64(self.id().get() as u64);
```
alamb, line 392: 👍

PR #6386, opened 2022-12-13
feat(ingester2): log persist active & queue timings

domodwyer:
This adds a bit more logging to persist, and captures the time a persist job spent in the queue, and how long it actually took to persist once dequeued - this information is displayed in the existing "persisted partition" INFO log:

```
INFO persisted partition <snip> total_persist_duration=51.535291ms active_persist_duration=12.523375ms queued_persist_duration=39.011916ms
```

---

* feat(ingester2): persist active & queue timings (5fa4e4909)

      Adds more debug logging to the persist code paths, as well as capturing
      & logging (at INFO) timing information tracking the time a persist task
      spends in the queue, the active time spent actually persisting the data,
      and the total duration of time since the request was created (sum of
      both durations).

### DIFF
--- a/ingester2/src/persist/context.rs
+++ b/ingester2/src/persist/context.rs
@@ -11,7 +11,7 @@ use observability_deps::tracing::*;
 use parking_lot::Mutex;
 use parquet_file::metadata::IoxMetadata;
 use schema::sort::SortKey;
-use tokio::sync::Notify;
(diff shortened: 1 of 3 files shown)


pauldix (approved):
Nice, I was definitely looking for something like this during my ad hoc testing 😄

Issue #6391, opened 2022-12-13
FlightSQL Milestone 1: Groundwork


alamb:
The goal of this ticket is to get the existing Flight code in influxdb_iox ready documented and ready for expansion

I plan 
- [x] Client cleanup
- [ ] Server cleanup
- [ ] ~Fork the FlightSQL client into iox (temporarily)~ -- going to do this as part of https://github.com/influxdata/influxdb_iox/issues/6411 instead

PR #6392, opened 2022-12-13
feat: Document IOx Flight API and add convenience methods

alamb:
re https://github.com/influxdata/influxdb_iox/issues/6391

# Rationale
I spent a bunch of time yesterday studying this code, understanding how it works, and writing that down. I would like to encode my learnings into the IOx source code for my future self (and I hope it is helpful for other readers)

# Changes
1. Standardize naming on `IOx Flight API`
2. Add additional documentation on how it works
3. Improve the rust client's API bindings to hide some of the ticket creation

There are no changes to the actual messages exchanged or protocol


influxdb_iox/src/commands/query.rs
```diff
@@ -59,17 +52,10 @@ pub async fn command(connection: Connection, config: Config) -> Result<()> {
...
-            namespace_name: namespace,
-            sql_query: query,
-            query_type: match query_lang {
-                QueryLanguage::Sql => read_info::QueryType::Sql,
-                QueryLanguage::InfluxQL => read_info::QueryType::InfluxQl,
-            }
-            .into(),
-        })
-        .await?;
+    let mut query_results = match query_lang {
```
alamb, line 55: I moved creation of the `ReadInfo` structure into the client code as I expect the `ReadInfo `structure to become more sophisticated during the implementation of FlightSQL and would like to avoid changing the rest of the codebase as much as possible

I also think this encapsulation makes for an easier to use client.

influxdb_iox/src/commands/sql/repl.rs
```diff
@@ -389,31 +391,3 @@ fn history_file() -> PathBuf {
...
-            namespace_name: db_name.to_string(),
-            sql_query: query.to_string(),
-            query_type: read_info::QueryType::Sql.into(),
-        })
-        .await
-        .context(RunningRemoteQuerySnafu)?;
-
-    let mut batches = vec![];
-
-    while let Some(data) = query_results
```
alamb, line 410: this functionality appeared in the IOx flight API client at some point, so I switched to using that instead

influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -126,20 +135,59 @@ impl Client {
...
+    /// [`PerformQuery`] instance that streams Arrow [`RecordBatch`] results.
+    pub async fn influxql(
+        &mut self,
+        namespace_name: String,
+        influxql_query: String,
+    ) -> Result<PerformQuery, Error> {
+        let request = ReadInfo {
+            namespace_name,
+            sql_query: influxql_query,
+            query_type: QueryType::Sql.into(),
```
alamb, line 163: 🤦

PR #6393, opened 2022-12-13
fix(ingester2): persist deadlock

domodwyer:
Removes the "submission queue" to simplify configuring the persist system, and fixes the persist deadlock under load.

Most of this diff is moving things out of the deleted `actor.rs` file, and putting them in the `PersistHandle`. The actual change is minor.

---

* fix(ingester2): persist deadlock (1da9b63cc)

      Removes the submission queue from the persist fan-out, instead the
      PersistHandle now carries the shared state internally (cheaply cloned
      via ref counts).
      
      This also resolves the persist deadlock when under load.


ingester2/src/persist/handle.rs
```diff
@@ -113,7 +92,19 @@ pub(crate) enum PersistError {
...
+    /// A consistent hash implementation used to consistently map buffers from
+    /// one partition to the same worker queue.
+    ///
+    /// This ensures persistence is serialised per-partition, but in parallel
+    /// across partitions (up to the number of worker tasks).
+    persist_queues: Arc<JumpHash<mpsc::Sender<PersistRequest>>>,
+
+    /// Task handles for the worker tasks, aborted on drop of all
+    /// [`PersistHandle`] instances.
+    tasks: Arc<Vec<AbortOnDrop<()>>>,
```
domodwyer, line 107: This stuff used to live in the `PersistActor`, but is now shared in the handles themselves.

This eliminates the need for the queue to the actor.

ingester2/src/persist/handle.rs
```diff
@@ -122,67 +113,112 @@ impl PersistHandle {
...
+        let (tx_handles, tasks): (Vec<_>, Vec<_>) = (0..n_workers)
+            .map(|_| {
+                let inner = Arc::clone(&inner);
+                let (tx, rx) = mpsc::channel(worker_queue_depth);
+                (tx, AbortOnDrop(tokio::spawn(run_task(inner, rx))))
+            })
+            .unzip();
 
-        (Self { tx }, actor)
+        assert!(!tasks.is_empty());
```
domodwyer, line 147: Likewise this was in the actor, and now is initialised with the first `PersistHandle`.

pauldix (approved):

PR #6395, opened 2022-12-13
refactor: Improve Flight API server side code and comments

alamb:
# Rationale
This is the side partner for https://github.com/influxdata/influxdb_iox/pull/6392

I spent a bunch of time yesterday studying this code, understanding how it works, and writing it down. I would like to encode my learnings into the IOx source code for my future self (and I hope it is helpful for other readers)

# Changes
1. Standardize naming on `IOx Flight API`
2. Add additional documentation on how the server  works
3. Consolidate the Tickeet interpretation in its own crate, add some more tests

There are no changes to the actual messages exchanged or protocol


service_grpc_flight/src/lib.rs
```diff
@@ -123,68 +117,6 @@ impl Error {
 
 type TonicStream<T> = Pin<Box<dyn Stream<Item = Result<T, tonic::Status>> + Send + Sync + 'static>>;
 
-#[derive(Deserialize, Debug)]
```
alamb, line 126: I moved all the ticket interpretation into `request.rs` (as I expect to be adding significant functionality there as I implement FlightSQL)

service_grpc_flight/src/lib.rs
```diff
@@ -512,98 +447,13 @@ impl Stream for GetStream {
...
-    use assert_matches::assert_matches;
     use futures::Future;
-    use generated_types::influxdata::iox::querier::v1::read_info::QueryType;
     use metric::{Attributes, Metric, U64Gauge};
     use service_common::test_util::TestDatabaseStore;
     use tokio::pin;
 
     use super::*;
 
-    #[test]
```
alamb, line 524: tests are moved to `request.rs`

service_grpc_flight/src/request.rs
```diff
@@ -0,0 +1,227 @@
...
+        // Do not change this test without having first changed what the Go clients are sending!
+        let ticket = make_json_ticket(r#"{"namespace_name": "my_db", "sql_query": "SELECT 1;"}"#);
+        let ri = IoxGetRequest::try_new(ticket).unwrap();
+
+        assert_eq!(ri.namespace_name, "my_db");
+        assert_matches!(ri.query, RunQuery::Sql(query) => assert_eq!(query, "SELECT 1;"));
+    }
+
+    #[test]
+    fn json_ticket_decoding_error() {
```
alamb, line 144: new error test

service_grpc_flight/src/request.rs
```diff
@@ -0,0 +1,227 @@
...
+        // invalid json (database name rather than namespace name)
+        let ticket = make_json_ticket(r#"{"database_name": "my_db", "sql_query": "SELECT 1;"}"#);
+        let e = IoxGetRequest::try_new(ticket).unwrap_err();
+        assert_matches!(e, Error::Invalid);
+    }
+
+    #[test]
+    fn proto_ticket_decoding_unspecified() {
```
alamb, line 152: I broke these tests into individual rust tests as they were all independent. I also changed them to call `IoxGetRequest::try_new()` so they went through the same path as the actual server uses

service_grpc_flight/src/request.rs
```diff
@@ -0,0 +1,227 @@
...
+        });
+
+        // Reverts to default (unspecified) for invalid query_type enumeration, and thus SQL
+        let ri = IoxGetRequest::try_new(ticket).unwrap();
+        assert_eq!(ri.namespace_name, "<foo>_<bar>");
+        assert_matches!(ri.query, RunQuery::Sql(query) => assert_eq!(query, "SELECT 1"));
+    }
+
+    #[test]
+    fn proto_ticket_decoding_error() {
```
alamb, line 206: added error coverage

alamb:
@tustvold  is there any chance you could review this code (I am hoping you can help with reviews here, and in arrow-rs) over the next few weeks 🙏

service_grpc_flight/src/lib.rs
```diff
@@ -209,26 +141,26 @@ where
         &self,
         span_ctx: Option<SpanContext>,
         permit: InstrumentedAsyncOwnedSemaphorePermit,
-        query: Query,
-        namespace: String,
```
tustvold, line 213: Why this change given the `namespace.to_string()` in the lines below?

tustvold (approved):

service_grpc_flight/src/lib.rs
alamb: I am not sure which one you are referring to. This function needs an owned `String` and sometimes callers have one already. This change was to avoid a clone (via `namespace.to_string()`) in those cases.

I removed a namespace.to_string in the lines below 

I actually originally tried to have this be `namespace: impl Into<String>` but the compiler was getting all cranky that it wasn't `Send` even though I called `to_string` prior to an await point for some reason and I couldn't figure a workaround

tustvold: You call to_string - https://github.com/influxdata/influxdb_iox/pull/6395/files/8620f6a47c1172de8cae9b1c73c2549589265409#diff-e0cdaab404f239ec66e06ce90f980a67634c20de0aa6d30167657f5270ecffc0R176

Despite having changed it to &str

https://github.com/influxdata/influxdb_iox/pull/6395/files/8620f6a47c1172de8cae9b1c73c2549589265409#diff-e0cdaab404f239ec66e06ce90f980a67634c20de0aa6d30167657f5270ecffc0R176

alamb: I see -- I will put it back -- this was my ill fated attempt to make this work with `impl Into<String>` -- sorry for the confusion

alamb: Fixed in 818955dd6

PR #6397, opened 2022-12-13
feat: make data generator handle failed requests

pauldix:
Updates the data generator to handle failed requests. Adds some println output to show progress along the way.


iox_data_generator/src/agent.rs
```diff
@@ -71,6 +71,46 @@ pub struct Agent {
...
+            format!("made {} requests at {}/sec with {} rows at {}/sec in {:?} with {} errors for a {:.2} error rate",
+                self.request_count, reqs_per_sec, self.row_count, rows_per_sec, elapsed_time, self.error_count, self.error_rate())
+        }
+    }
+
+    fn error_rate(&self) -> f64 {
+        if self.error_count == 0 {
+            return 0.0;
+        }
+        self.error_count as f64 / self.request_count as f64 * 100.0
```
domodwyer, line 110: This might print an error rate of "inf" if there's been no requests because of the division by 0 - I don't think that's a big deal though.

domodwyer (approved):
Love it :+1:

PR #6398, opened 2022-12-13
feat: FlightSQL Milestone 2 basic FlightSQL client and FlightSQL server implementation and plumbing

alamb:
~Draft as it builds on~
- [x] https://github.com/influxdata/influxdb_iox/pull/6422
- [x] https://github.com/influxdata/influxdb_iox/pull/6427
- [x] https://github.com/influxdata/influxdb_iox/pull/6438

# Rationale
Closes https://github.com/influxdata/influxdb_iox/issues/6411 

This PR implements all the plumbing (including an end to end test) needed to run FlightSQL queries directly in IOx. It is no where near complete, but I think it is a strong foundation on which to build.

I plan to guide the next steps of the implementation by writing an integration test with the FilghtSQL JDBC driver and filling out the parts of the implementation it needs.


avantgardnerio:
Thanks for the heads up @alamb we're essentially doing the same. It'l be a fun merge when all this is ready :)

alamb:
> Thanks for the heads up @alamb we're essentially doing the same. It'l be a fun merge when all this is ready :)


Yeah, I agree -- I plan to start sending bits and pieces back up stream as I can -- for example https://github.com/apache/arrow-rs/blob/2749dcca50e6dd0ac72db7fe802552c2db742c3c/arrow-flight/src/sql/client.rs#L454-L530

Which I found we have basically the same thing in IOx:
https://github.com/influxdata/influxdb_iox/blob/main/influxdb_iox_client/src/client/flight/low_level.rs#L229-L305

There is also some code that handles the streaming send from the server which would be nice to send back upstream

influxdb_iox/tests/end_to_end_cases/flightsql.rs
```diff
@@ -0,0 +1,66 @@
...
+                        "+------+------+--------------------------------+-----+",
+                        "| A    | B    | 1970-01-01T00:00:00.000123456Z | 42  |",
+                        "| A    | C    | 1970-01-01T00:00:00.000123457Z | 43  |",
+                        "+------+------+--------------------------------+-----+",
+                    ];
+
+                    let connection = state.cluster().querier().querier_grpc_connection();
+                    let (channel, _headers) = connection.into_grpc_connection().into_parts();
+
+                    let mut client = iox_arrow_flight::FlightSqlClient::new(channel);
```
alamb, line 40: Here is the new end to end test showing how the FlightSQL client works

iox_arrow_flight/src/flightsql.rs
```diff
@@ -0,0 +1,161 @@
...
+
+/// A FlightSQLServiceClient handles details of interacting with a
+/// remote server using the FlightSQL protocol.
+#[derive(Debug)]
+pub struct FlightSqlClient {
+    inner: FlightClient,
+}
+
+/// An [Arrow Flight
+/// SQL](https://arrow.apache.org/docs/format/FlightSql.html) client
```
tustvold, line 43:
```suggestion
/// An [Arrow Flight SQL](https://arrow.apache.org/docs/format/FlightSql.html) client
```
A minor nit, but the line break makes this hard to read imo

service_grpc_flight/src/lib.rs
```diff
@@ -259,11 +313,43 @@ where
...
+        let cmd = match request.r#type() {
+            DescriptorType::Cmd => Ok(&request.cmd),
+            DescriptorType::Path => Err(Error::unsupported_message_type("FlightInfo with Path")),
+            DescriptorType::Unknown => Err(Error::unsupported_message_type(
+                "FlightInfo of unknown type",
+            )),
+        }?;
+
+        let message: prost_types::Any =
+            prost::Message::decode(cmd.as_slice()).context(DeserializationSnafu)?;
```
tustvold, line 349: We will eventually want to change this to reflect the upstream changes to use Bytes

service_grpc_flight/src/lib.rs
```diff
@@ -295,6 +381,88 @@ where
...
+    /// flight_descriptor: is the descriptor sent in the request (included in response)
+    ///
+    /// msg is the `cmd` field of the flight descriptor decoded as a protobuf  message
+    async fn dispatch(
+        &self,
+        namespace_name: &str,
+        flight_descriptor: FlightDescriptor,
+        msg: prost_types::Any,
+    ) -> Result<FlightInfo> {
+        fn try_unpack<T: ProstMessageExt>(msg: &prost_types::Any) -> Result<Option<T>> {
```
tustvold, line 405: FWIW this exists upstream as `ProstAnyExt` in arrow 29, and the custom `Any` in the future arrow 30.

Unfortunately I can't link to docs because of https://github.com/apache/arrow-rs/pull/3377

tustvold (approved):

service_grpc_flight/src/lib.rs
alamb: 👍  -- I just did that in https://github.com/apache/arrow-rs/pull/3378 as it turns out 😆 

My hope is that I will get this upstreamed ASAP rather than chasing dependencies in IOx

PR #6399, opened 2022-12-13
feat: optimize wal with batching

pauldix:
Simplified the wal writer so that it batches up write operations. Currently it waits 10ms between fsync calls. We can pull this out to a config variable later if we want, but I think this is good enough for now.

Also updated the reader to be a more simple blocking reader without the extra tasks and channels as that wasn't really getting us anything that I know of.


wal/src/lib.rs
```diff
@@ -192,7 +210,7 @@ impl Wal {
     ///
     /// Similarly, editing or deleting files within a `Wal`'s root directory via some other
     /// mechanism is not supported.
-    pub async fn new(root: impl Into<PathBuf>) -> Result<Self> {
+    pub async fn new(root: impl Into<PathBuf>) -> Result<Arc<Self>> {
         let root = root.into();
         tokio::fs::create_dir_all(&root)
             .await
```
domodwyer, line 216: After creating the directory structure with `create_dir_all()`, we **must** ensure the directory metadata is flushed to disk, otherwise we'll fall into a fantastic filesystem trap: the files exist, they were carefully `fsync()`ed to maintain data consistency, but they're not reachable because the _directory they're in_ doesn't exist 🤦 

Something like this here, after the `create_dir_all()` (but before we use the dir) will ensure the directory metadata is made crash safe:

```rust
std::fs::File::open(&dir)
    .expect("should be able to open just-created directory")
    .sync_all()
    .expect("fsync failure");
```

This isn't even a stupid legacy problem either - `btrfs` is known[^1] to allow reordering of file & directory ops in a way that would vanish the WAL files.

[^1]: https://www.usenix.org/system/files/conference/osdi14/osdi14-paper-pillai.pdf, table 1.

wal/src/lib.rs
```diff
@@ -241,113 +259,159 @@ impl Wal {
...
+        b.ops.push(op);
+        b.flush_notification.clone()
     }
 
-    /// Opens a reader for a given segment from the WAL
-    pub async fn reader_for_segment(&self, id: SegmentId) -> Result<ClosedSegmentFileReader> {
-        let path = build_segment_path(&self.0.root, id);
-        ClosedSegmentFileReader::from_path(path).await
+    /// Gets a watch channel that will notify when the buffered ops are flushed to disk
+    pub fn notify_on_flush(&self) -> watch::Receiver<Option<WriteResult>> {
```
domodwyer, line 306: I don't think this is used, and I think we should remove it - it's hard to use this correctly and it'll always race with rotation:

* call `notify_on_flush()`
* add op
* wait for flush

The flush might have happened before `add op`, leaving the caller with the incorrect belief that the op is durable. 

It can also let someone accidentally shoot themselves in the foot:

* add op
* call `notify_on_flush()`
* wait for flush
* add op
* wait for flush

But the notification is per-buffer - the second wait-for-flush will never happen, leading to a deadlock.

wal/src/lib.rs
```diff
@@ -241,113 +259,159 @@ impl Wal {
...
+                .collect();
+            let batch = ProtoWalOpBatch { ops };
+            let encoded = batch.encode_to_vec();
+
+            // now write the data and let everyone know how things went
+            let mut segments = self.segments.lock().expect("mutex poisoned");
+            let res = match segments.open_segment.write(&encoded) {
+                Ok(summary) => WriteResult::Ok(summary),
+                Err(e) => WriteResult::Err(e.to_string()),
+            };
```
domodwyer, line 361: Calling `send()` on `notify_flush` involves acquiring another mutex which might block. We can avoid holding the `segments` lock while we do that:

```suggestion
            let res = {
                let mut segments = self.segments.lock().expect("mutex poisoned");
                match segments.open_segment.write(&encoded) {
                    Ok(summary) => WriteResult::Ok(summary),
                    Err(e) => WriteResult::Err(e.to_string()),
                }
            };
```

wal/src/lib.rs
```diff
@@ -241,113 +259,159 @@ impl Wal {
...
+            filled_buffer
+                .notify_flush
+                .send(Some(res))
+                .expect("unable to send flush notification");
```
domodwyer, line 365: It's unlikely, but perfectly possible there will be no one listening to this broadcast:

* One write comes in
* It is added to the WAL buffer, returning the completion watch
* The writer closes their connection, dropping the receiver of the watch
* The WAL flushes, and there is no one around to see this send!

```suggestion
            // Do not panic if no thread is waiting for the flush notification -
            // this may be the case if all writes disconnected before the WAL
            // was flushed.
            let _ = filled_buffer.notify_flush.send(Some(res));
```

This will prevent a needless panic :+1:

domodwyer (approved):
This is an impressive improvement - solid code too!

wal/src/lib.rs
```diff
@@ -13,23 +13,34 @@
...
+        WalOpBatch as ProtoWalOpBatch,
     },
 };
 use prost::Message;
 use snafu::prelude::*;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+use std::sync::Mutex;
+use std::time::Duration;
```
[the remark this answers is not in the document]
pauldix: Bah, I'm using CLion and I thought it used to do this. But new laptop so who knows 🤷

wal/src/lib.rs
pauldix: This was meant for some part of the code that isn't writing ops to be notified when the buffer has been flushed. Can't quite remember why I put this in here, but it was related to the persisting code I hacked together which didn't make it into this PR. Will remove, the only channel writers should be using is the one returned when they write.

PR #6403, opened 2022-12-14
feat(ingester2): persist back pressure

domodwyer:
This teaches ingester2 to apply back-pressure when the persist system is saturated.

Once there are enough outstanding persist jobs that a persist worker's queue overflows, ingest is stopped (by returning a "resource exhausted" error to the user). 

With the current router logic, a router that sees an overloaded ingester will move on and try the other ingesters (if any) before retrying the overloaded ingester again. After 3 seconds of constant failures, the routers return an error to the user - this should help us maintain write availability while an overloaded ingester recovers (assuming sufficient number of ingesters are running).

Most of the diff is docs & tests.

---

* test: show caller in timeout panics (6c555600e)

      Changes the stack trace of the timeout panics to show the line that
      timed out, rather than the timeout implementation itself.

* feat(ingester2): persist back-pressure (e76b10733)

      This commit causes an ingester2 instance to stop accepting new writes
      when at least one persist queue is full. Writes continue to be rejected
      until the persist workers have processed enough outstanding persist
      tasks to drain the queues to half of their capacity, at which point
      writes are accepted again.
      
      When a write is rejected, the ingester returns a "resource exhausted"
      RPC code to the caller.
      
      Checking if the system is in a healthy state for writes is extremely
      cheap, as it is on the hot path for all writes.

* refactor(config): default persist queue depth=250 (c7e4bf3dd)

      Allow up to 250 persist jobs to be enqueued for any one worker before
      pausing.
      
      With 5 workers, this gives a maximum outstanding persist jobs of 2,500.

### DIFF
--- a/ingester2/src/persist/backpressure.rs
+++ b/ingester2/src/persist/backpressure.rs
@@ -1,0 +1,521 @@
+use std::{
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use crossbeam_utils::CachePadded;
+use observability_deps::tracing::*;
+use parking_lot::Mutex;
+use tokio::{
+    sync::mpsc,
+    task::JoinHandle,
+    time::{Interval, MissedTickBehavior},
+};
+
+/// The interval of time between evaluations of the state of the persist system
+/// when [`CurrentState::Saturated`].
+const EVALUATE_SATURATION_INTERVAL: Duration = Duration::from_secs(1);
+
+/// A state of the persist system.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum CurrentState {
+    /// The system is operating normally.
+    Ok,
+    /// The persist system is overloaded.
+    Saturated,
+}
+
+/// A handle to read (and set, within the persist module) the state of the
+/// persist system.
+///
+/// Clone operations are cheap, and state read operations are very cheap.
+///
+/// # Saturation Recovery
+///
+/// Once the persist system is marked as [`CurrentState::Saturated`], it remains
+/// in that state until the following conditions are satisfied:
+///
+///   * There are no outstanding enqueue operations (no thread is blocked adding
+///     an item to any work queue).
+///
+///   * All queues have at least half of their capacity free (being at most,
+///     half full).
+///
+/// These conditions are evaluated periodically, at the interval specified in
+/// [`EVALUATE_SATURATION_INTERVAL`].
+#[derive(Debug)]
+pub(crate) struct PersistState {
+    /// The actual state value.
+    ///
+    /// The value of this variable is set to the [`CurrentState`] discriminant
+    /// for the respective state that was [`PersistState::set()`] in it.
+    ///
+    /// This is cache padded due to the high read volume, preventing any
+    /// unfortunate false-sharing of cache lines from impacting the hot-path
+    /// reads.
+    state: CachePadded<AtomicUsize>,
+
+    /// Tracks the number of async tasks waiting within
+    /// [`PersistHandle::queue_persist()`], asynchronously blocking to enqueue a
+    /// persist job.
+    ///
+    /// This is modified using [`Ordering::SeqCst`] as performance is not a
+    /// priority for code paths that modify it.
+    ///
+    /// [`PersistHandle::queue_persist()`]:
+    ///     super::handle::PersistHandle::queue_persist()
+    waiting_to_enqueue: Arc<AtomicUsize>,
+
+    /// The handle to the current saturation evaluation/recovery task, if any.
+    recovery_handle: Mutex<Option<JoinHandle<()>>>,
+}
+
+/// Initialise a [`PersistState`] with [`CurrentState::Ok`].
+impl Default for PersistState {
+    fn default() -> Self {
+        let s = Self {
+            state: Default::default(),
+            waiting_to_enqueue: Arc::new(AtomicUsize::new(0)),
+            recovery_handle: Default::default(),
+        };
+        s.set(CurrentState::Ok);
+        s
+    }
+}
+
+impl PersistState {
+    /// Set the reported state of the [`PersistState`].
+    fn set(&self, s: CurrentState) -> bool {
+        // Set the new state, retaining the most recent state.
+        //
+        // SeqCst is absolute overkill, but is used here due to the strong
+        // ordering guarantees providing minimal risk of bugs. The low volume of
+        // writes to this variable means the overhead is more than acceptable.
+        let last = self.state.swap(s as usize, Ordering::SeqCst);
+
+        // If "s" does not match the old state, this is the first thread to
+        // switch the state from "last", to "s", since setting it to "last".
+        //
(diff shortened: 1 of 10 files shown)


ingester2/src/persist/backpressure.rs
```diff
@@ -0,0 +1,521 @@
...
+/// in that state until the following conditions are satisfied:
+///
+///   * There are no outstanding enqueue operations (no thread is blocked adding
+///     an item to any work queue).
+///
+///   * All queues have at least half of their capacity free (being at most,
+///     half full).
+///
+/// These conditions are evaluated periodically, at the interval specified in
+/// [`EVALUATE_SATURATION_INTERVAL`].
```
domodwyer, line 48: This documents what constitutes a successful recovery once the persist system is marked as saturated.

ingester2/src/persist/handle.rs
```diff
@@ -90,6 +89,28 @@ pub(crate) enum PersistError {
...
+/// generation of new persist tasks (for example, by blocking any further
+/// ingest) on a best-effort basis.
+///
+/// When the persist queue is saturated, the [`PersistState::is_saturated()`]
+/// returns true. Once the backlog of persist jobs is reduced, the
+/// [`PersistState`] is switched back to a healthy state and new persist jobs
+/// may be generated as normal.
+///
+/// For details of the exact saturation detection & recovery logic, see
+/// [`PersistState`].
```
domodwyer, line 113: This provides general information about the back-pressure implementation for users.

ingester2/src/server/grpc/rpc_write.rs
```diff
@@ -115,6 +134,27 @@ where
...
+        //   * Increase the persist queue depth if there is a decent headroom of
+        //     unused RAM allocated to the ingester.
+        //   * Increase the RAM allocation, and increase the persist queue
+        //     depth proportionally.
+        //   * Deploy more ingesters to reduce the request load on any single
+        //     ingester.
+        //
+        if self.persist_state.is_saturated() {
+            return Err(RpcError::PersistSaturated)?;
+        }
```
domodwyer, line 155: This comment is intended to help people who come looking for the source of the RPC error understand what is happening and how to solve it.

ingester2/src/persist/backpressure.rs
```diff
@@ -0,0 +1,521 @@
...
+        // INVARIANT: this increment MUST happen-before returning the guard, and
+        // waiting on the queue send(), and before starting the saturation
+        // monitor task so that it observes this waiter.
+        let _ = s.waiting_to_enqueue.fetch_add(1, Ordering::SeqCst);
+
+        // Attempt to set the system to "saturated".
+        let first = s.set(CurrentState::Saturated);
+        if first {
+            // This is the first thread to mark the system as saturated.
+            warn!("persist queues saturated, blocking ingest");
```
domodwyer, line 157: The first thread that sets the persist system state to "saturated" spawns the worker task that evaluates the state of the system and marks it as recovered when appropriate.

ingester2/src/persist/backpressure.rs
```diff
@@ -0,0 +1,521 @@
...
+    }
+}
+
+/// A task that monitors the `waiters` and `queues` to determine when the
+/// persist system is no longer saturated.
+///
+/// Once the system is no longer saturated (as determined according to the
+/// documentation for [`PersistState`]), the [`PersistState`] is set to
+/// [`CurrentState::Ok`].
+async fn saturation_monitor_task<T>(
```
domodwyer, line 215: This is the task that monitors the state of the persist system, and unblocks ingest once the backlog has reduced sufficiently.

ingester2/src/persist/backpressure.rs
```diff
@@ -0,0 +1,521 @@
...
+///   * There are no outstanding enqueue operations (no thread is blocked adding
+///     an item to any work queue).
+///
+///   * All queues have at least half of their capacity free (being at most,
+///     half full).
+///
+/// These conditions are evaluated periodically, at the interval specified in
+/// [`EVALUATE_SATURATION_INTERVAL`].
+#[derive(Debug)]
+pub(crate) struct PersistState {
```
pauldix, line 50: This feels like a ton of code and wrapping for something that, I think, is very simple. If you have a persist queue, the persist state is set on the depth of that queue. The state of saturated and not saturated only changes whenever the depth of the queue changes, which should set the state automatically. I'm not sure what the recovery task is for given that when workers pull from the queue, its depth (and hence the sate) should be updated.

Maybe this is a result of having multiple queues? This just seems like a ton of extra wrapping for just checking if queue depth is greater than some value.

domodwyer: It's not the result of having multiple queues - you could simply `sum()` all the queue depths. There's two things at play:

* Threads trying to push into the queues are not reflected in the queue capacity - you could have a queue of depth 10, but 100 writers waiting to push into it. Reading the queue capacity alone will lead to a flappy health check where you observe some space in the queue, decide that the system has recovered, only for it to be immediately filled by the outstanding writers.

* Reading the queue depths becomes increasingly expensive as the number of writers increases (more persist ops are enqueued), which happens on the (hottest of the) hot paths. 

I get that it appears like a lot of code, the diff is very big, but it's actually ~100 lines of code (including newlines) if you delete all the documentation and comments, which is the vast majority of the LOC - it's a relatively simple `is_saturated: bool` in effect.

carols10cents (approved):
I've tried to make this actually happen for me and I'm not setting up the scenario properly (or the system is just Too Fast!) Seems ok, let's try it out.

ingester2/src/persist/backpressure.rs
```diff
@@ -0,0 +1,521 @@
...
+        // different state is set.
+        s as usize != last
+    }
+
+    /// Get the current reported state of the [`PersistState`].
+    ///
+    /// Reading this value is extremely cheap and can be done without
+    /// performance concern.
+    ///
+    /// This value is eventually consistent, with a presumption of
```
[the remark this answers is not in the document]
domodwyer: [No joke.](https://gcc.gnu.org/wiki/Atomic/GCCMM/AtomicSync#Relaxed)

> There is also the presumption that relaxed stores from one thread are seen by relaxed loads in another thread within a reasonable amount of time.

It is genuinely fine though. We also use relaxed ordering for our metrics, for example - this isn't much different. The actual C++ memory model spec (which rust also uses) is a little more [_reasonable_](https://timsong-cpp.github.io/cppwp/n4868/atomics#order-11) 🥁 :

> Implementations should make atomic stores visible to atomic loads within a reasonable amount of time[.](https://timsong-cpp.github.io/cppwp/n4868/atomics#order-11.sentence-1)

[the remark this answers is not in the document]
carols10cents: I was trying to point out that this comment ends mid-sentence ;) Just a nit though!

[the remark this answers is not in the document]
domodwyer: Ha, brilliant - totally missed that 😂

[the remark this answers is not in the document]
domodwyer: https://github.com/influxdata/influxdb_iox/pull/6437

PR #6405, opened 2022-12-14
feat: Increase default all-in-one compactor memory to 300MB and expose it as a setting

alamb:
Closes https://github.com/influxdata/idpe/issues/16707

# Rationale
The default all in one configuration limits the compactor's memory_budget_bytes to 300K (**NOT MB**). Furthermore it is hard coded and can not be changed (increased). 

This means that the compactor does not really operate in the all in one mode on any reasonable sized data.

# Changes
1. Increase the default to 300M and allow it to be configured from the command line.
2. Expose `INFLUXDB_IOX_COMPACTION_MEMORY_BUDGET_BYTES` option to expand


influxdb_iox/src/commands/run/all_in_one.rs
```diff
@@ -429,7 +451,7 @@ impl Config {
             max_number_partitions_per_shard: 1,
             min_number_recent_ingested_files_per_partition: 1,
             hot_multiple: 4,
-            memory_budget_bytes: 300_000,
```
alamb, line 432: 300K!

NGA-TRAN (approved):
Thanks @alamb

PR #6406, opened 2022-12-14
fix: Remove the rpc_write feature flag and use INFLUXDB_IOX_MODE env var instead

### LINKED ISSUE #6402: Rip out rpc_write feature flag; instead use one env var for switching into iox2 mode
So that we can deploy and run either mode from the same docker image.

- Env var `INFLUXDB_IOX_MODE` presence allows running `router2`, `ingester2`, `querier` in RPC write mode, and (to be created soon) `compactor2`
- Specifying `INFLUXDB_IOX_MODE` and trying to run `router`, `ingester`, or `compactor` is an error. 
- Omitting `INFLUXDB_IOX_MODE` and trying to run `router2`, `ingester2`, or `compactor2` is an error.
- Querier:
  - Specifying  `INFLUXDB_IOX_MODE` and `--shard-to-ingesters` mapping is an error.
  - Omitting `INFLUXDB_IOX_MODE` and specifying `--ingester-addresses` is an error.

carols10cents:
And standardize on ingester2 and router2 for consistency.

**The `test_rpc_write` CI job is now exactly the same as the `test` job but I can't remove it because it's required; @domodwyer could you un-require it please?** ❤️

Closes #6402.

You will now see `router2` and `ingester2` in `influxdb_iox run --help`.

BUT it shouldn't be possible to *only* typo the command by adding `2`; you also need to set `INFLUXDB_IOX_RPC_MODE` (to any value, I've been setting it to `2`, but it's only set/unset that's checked).

Querier is still different as it shares more code; there is no `querier2`. But the `INFLUXDB_IOX_RPC_MODE` and setting `--ingester-addresses` instead of shard to ingester mappings is how you switch the querier into the RPC write path.

## Great, what do I copy-paste now?

So now you can run these commands to use the RPC write path all the way through:

```
# ingester2
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run ingester2 --grpc-bind=127.0.0.1:8042

# router2
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run router2 --api-bind=127.0.0.1:8081 --ingester-addresses=127.0.0.1:8042

# querier
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run querier --ingester-addresses=http://127.0.0.1:8042 --grpc-bind 127.0.0.1:8043
```

### DIFF
--- a/clap_blocks/src/querier.rs
+++ b/clap_blocks/src/querier.rs
@@ -278,9 +278,9 @@ impl QuerierConfig {
     // When we have switched to using the RPC write path only, this method can be changed to be
     // infallible as clap will handle failure to parse the list of strings.
     //
-    // Switching into the RPC write path mode requires *both* the `INFLUXDB_IOX_MODE` environment
+    // Switching into the RPC write path mode requires *both* the `INFLUXDB_IOX_RPC_MODE` environment
     // variable to be specified *and* `--ingester-addresses` to be set in order to switch. If the
-    // `INFLUXDB_IOX_MODE` is enabled and `--shard-to-ingesters*` is set, use the write buffer path
+    // `INFLUXDB_IOX_RPC_MODE` is enabled and `--shard-to-ingesters*` is set, use the write buffer path
     // instead.
     pub fn ingester_addresses(&self) -> Result<IngesterAddresses, Error> {
         if let Some(file) = &self.shard_to_ingesters_file {
--- a/clap_blocks/src/querier.rs
+++ b/clap_blocks/src/querier.rs
@@ -212,7 +212,6 @@ pub struct QuerierConfig {
     ///
     /// for multiple addresses.
     #[clap(long = "ingester-addresses", env = "INFLUXDB_IOX_INGESTER_ADDRESSES")]
-    #[cfg(feature = "rpc_write")]
     pub ingester_addresses: Vec<String>,
 
     /// Size of the RAM cache used to store catalog metadata information in bytes.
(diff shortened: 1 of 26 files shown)


clap_blocks/src/querier.rs
```diff
@@ -272,45 +271,17 @@ impl QuerierConfig {
...
+    // infallible as clap will handle failure to parse the list of strings.
     //
-    // For now, to enable turning on the `rpc_write` feature in tests but not necessarily switching
-    // into the RPC write path mode, require *both* the feature flag to be enabled *and*
-    // `--ingester-addresses` to be set in order to switch. If the `rpc_write` feature is enabled
-    // and `--shard-to-ingesters*` are set, use the write buffer path instead.
-    #[cfg(feature = "rpc_write")]
+    // Switching into the RPC write path mode requires *both* the `INFLUXDB_IOX_MODE` environment
+    // variable to be specified *and* `--ingester-addresses` to be set in order to switch. If the
+    // `INFLUXDB_IOX_MODE` is enabled and `--shard-to-ingesters*` is set, use the write buffer path
```
domodwyer, line 283: Perhaps setting `mode=2` and kafka settings should be an error? it's definitely a mix of two, unmixable desires. It's not a big deal though - you can't protect people from themselves all the time!

domodwyer: Actually it looks like this might be the case, and the comment is old? 

https://github.com/influxdata/influxdb_iox/pull/6406/files#diff-f39df79b58095b0137a251c8b8d8c8207560ea0eaaa214e589adcd0755f10aaaR194-R197

domodwyer (approved):
Looks cool - thank you :heart:

I have one suggestion; either make the code check the value of `INFLUXDB_IOX_MODE` (like `INFLUXDB_IOX_MODE=rpc` or `2` or something that is checked) or name it `INFLUXDB_IOX_RPC_MODE` to gate the RPC behaviour. That way it's clear what is happening from the context of the command (if it was "INFLUX_IOX_MODE=kafka" that would be confusing!)

> Great, what do I copy-paste now?

You certainly know your audience :joy:

clap_blocks/src/querier.rs
carols10cents: Whoops, yep, out of date, will fix the comment.

carols10cents:
I decided to go with changing the environment variable to be named `INFLUXDB_IOX_RPC_MODE` and I've updated the commands in this PR's description.

mkmik:
If you like copy pasta, I made a take-away version:

```console
mkdir -p /tmp/kafkaless
createdb iox_shared

# ingester2
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run ingester2 --api-bind=127.0.0.1:8081 --grpc-bind=127.0.0.1:8042 --wal-directory /tmp/kafkaless  --catalog-dsn postgres:///iox_shared  -v

# router2
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run router2 --api-bind=127.0.0.1:8083 --grpc-bind=127.0.0.1:8085 --ingester-addresses=127.0.0.1:8042 --catalog-dsn postgres:///iox_shared -v

# querier
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run querier --ingester-addresses=http://127.0.0.1:8042 --api-bind 127.0.0.1:8080 --grpc-bind 127.0.0.1:8082 --catalog-dsn postgres:///iox_shared -v
```

I picked the ports so that the router accepts writes at `8080`  and querier at `8082`, compatible with what all-in-one does and thus compatible with all my shell snippets I use to try out iox. YMMV

PR #6414, opened 2022-12-15
feat: delete namespace api

lukebond:

This is the naive portion of the namespace deletion feature. This will delete it and all associated resources (i.e. cascading), but issues remain that need to be address, with design and discussion:

- the querier and router maintain a namespace cache
- re-creation of a namespace with the same name will cause some undefined behaviour

I don't know if we want to merge this once approved and just not use it yet until we complete that design, or hold off. cc @crepererum @NGA-TRAN 


lukebond:
whoops forgot an important part of this PR, brb...

PR #6415, opened 2022-12-15
feat: delete namespace api

lukebond:

This is the naive portion of the namespace deletion feature. This will delete it and all associated resources (i.e. cascading), but issues remain that need to be address, with design and discussion:

the querier and router maintain a namespace cache
re-creation of a namespace with the same name will cause some undefined behaviour
I don't know if we want to merge this once approved and just not use it yet until we complete that design, or hold off. cc @crepererum @NGA-TRAN


alamb:
I plan to review this PR tomorrow morning (Eastern) if no one else has had a chance

iox_catalog/migrations/20221215145057_namespace_delete_cascade.sql
```diff
@@ -0,0 +1,48 @@
+ALTER TABLE parquet_file
+    DROP CONSTRAINT parquet_file_namespace_id_fkey;
+ALTER TABLE parquet_file
+    ADD CONSTRAINT parquet_file_namespace_id_fkey
+    FOREIGN KEY (namespace_id) REFERENCES namespace(id)
+    ON DELETE CASCADE;
```
NGA-TRAN, line 6: Oh, if we want to change all our FKs to `ON DELETE CASCADE` which I think we should, we have to do it before Jan 31 when we start accepting customer data. After that changing like this will be a headache consideration

iox_catalog/src/interface.rs
```diff
@@ -312,6 +315,9 @@ pub trait NamespaceRepo: Send + Sync {
     /// Gets the namespace by its unique name.
     async fn get_by_name(&mut self, name: &str) -> Result<Option<Namespace>>;
 
+    /// Delete a namespace by namDelete a namespace by name
```
NGA-TRAN, line 318: extra paste? "/// Delete a namespace by name"?

iox_catalog/src/interface.rs
```diff
@@ -4120,6 +4163,448 @@ pub(crate) mod test_helpers {
...
+        assert!(repos
+            .processed_tombstones()
+            .exist(p1_n2.id, t3_n2.id)
+            .await
+            .expect("processed tombstone exists check should succeed"));
+        assert!(repos
+            .processed_tombstones()
+            .exist(p2_n2.id, t3_n2.id)
+            .await
+            .expect("processed tombstone exists check should succeed"));
```
NGA-TRAN, line 4605: Nice test 👍

iox_catalog/src/postgres.rs
```diff
@@ -680,6 +680,28 @@ WHERE name = $1;
...
+        sqlx::query(
+            r#"
+WITH namespace_id as (
+    SELECT id
+    FROM namespace
+    WHERE name = $1
+    LIMIT 1
+)
+DELETE FROM namespace
+WHERE id = (SELECT id FROM namespace_id)
```
NGA-TRAN, line 694: I think we can simplify this
```SQL
DELETE FROM namespace WHERE name = $1
```

Usually the number of deleted rows will be returned. Do we need to get the id in returning list? We don't right?

NGA-TRAN (commented):
Very nice, Luke. Let us chat with the team to se if we need to consider any changes in behavior

NGA-TRAN:
@lukebond: I have a question. Since parquet files are gone from the catalog, are we able to delete the actual files in object store? Maybe we need to address this first before talking with the team about other behavior due to catching in Ingester and Querier?

iox_catalog/src/interface.rs
lukebond: thanks, good spot

iox_catalog/src/postgres.rs
lukebond: yes you're right. i wrote this before i realised there was a unique constraint on the name column, so i was trying to prevent a scan of the table to find all with that name. i'll fix it

lukebond:
> @lukebond: I have a question. Since parquet files are gone from the catalog, are we able to delete the actual files in object store? Maybe we need to address this first before talking with the team about other behavior due to catching in Ingester and Querier?

good point- they'll never get deleted in this case. i guess we could loosen a constraint and allow parquet files to have a `NULL` namespace id, and then we can mark them for deletion?

iox_catalog/src/postgres.rs
lukebond: can confirm it does use the unique name index, as one would expect:

```
postgres=# explain analyze delete from namespace where name = 'baz';
                                                              QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------
 Delete on namespace  (cost=0.15..8.17 rows=0 width=0) (actual time=0.010..0.011 rows=0 loops=1)
   ->  Index Scan using namespace_name_unique on namespace  (cost=0.15..8.17 rows=1 width=6) (actual time=0.009..0.010 rows=0 loops=1)
         Index Cond: ((name)::text = 'baz'::text)
 Planning Time: 0.094 ms
 Execution Time: 0.038 ms
(5 rows)
```

lukebond:
i've addressed your feedback @NGA-TRAN. there just remains the following:

- what to do about deleted parquet files in the catalog after a namespace deletion? they won't be cleaned up. i would suggest we mark them as not having a namespace, which requires loosening the constraint there. means we can keep the row in the DB marked with `to_delete` for the GC to clean up. feels a bit icky. open to other ideas
- if i'm adding `ON DELETE CASCADE` here, should we add it for everything else? seems to me it can be a separate PR if so
- once this PR is okay do we merge it without having designed a solution to namespace caching and re-creation of namespaces with the same name? seems to me we can deal with it separately

alamb:
> what to do about deleted parquet files in the catalog after a namespace deletion? they won't be cleaned up. i would suggest we mark them as not having a namespace, which requires loosening the constraint there. means we can keep the row in the DB marked with to_delete for the GC to clean up. feels a bit icky. open to other ideas

The gc also deletes parquet files from object store that are not in the catalog. https://github.com/influxdata/influxdb_iox/blob/main/garbage_collector/src/lib.rs#L88-L95

Thus I think it would be fine to simply delete the catalog entries and let the garbage collector clean them up eventually.

iox_catalog/migrations/20221215145057_namespace_delete_cascade.sql
alamb: Rather than say "ON DELETE CASCADE" in the DDL, can we instead add "CASCADE" to the delete command itself?

Like 

```sql
delete from namespace where id = 12345 CASCADE;
```

That way we would not have to change the DDL

lukebond:
> > what to do about deleted parquet files in the catalog after a namespace deletion? they won't be cleaned up. i would suggest we mark them as not having a namespace, which requires loosening the constraint there. means we can keep the row in the DB marked with to_delete for the GC to clean up. feels a bit icky. open to other ideas
> 
> The gc also deletes parquet files from object store that are not in the catalog. https://github.com/influxdata/influxdb_iox/blob/main/garbage_collector/src/lib.rs#L88-L95
> 
> Thus I think it would be fine to simply delete the catalog entries and let the garbage collector clean them up eventually.

of course! :+1:  we're good then :)

iox_catalog/src/interface.rs
```diff
@@ -4120,6 +4163,448 @@ pub(crate) mod test_helpers {
...
+            .unwrap();
+        assert!(!exist);
+        let exist = repos
+            .processed_tombstones()
+            .exist(p1_n2.id, t2_n2.id)
+            .await
+            .unwrap();
+        assert!(exist);
+
+        // now delete namespace_1 and assert it's all gone and none of namespace_2 is gone
```
alamb, line 4437: 👍

iox_catalog/src/postgres.rs
```diff
@@ -680,6 +680,21 @@ WHERE name = $1;
         Ok(Some(namespace))
     }
 
+    async fn delete(&mut self, name: &str) -> Result<()> {
+        // note that there is a uniqueness constraint on the name column in the DB
+        sqlx::query(
+            r#"
+DELETE FROM namespace
+WHERE name = $1;
```
alamb, line 688: I suggest considering putting a `CASCADE` at the end of this statement rather than changing the tables

ioxd_querier/src/rpc/namespace.rs
```diff
@@ -65,6 +65,15 @@ impl proto::namespace_service_server::NamespaceService for NamespaceServiceImpl
         ))
     }
 
+    async fn delete_namespace(
+        &self,
+        _request: tonic::Request<proto::DeleteNamespaceRequest>,
+    ) -> Result<tonic::Response<proto::DeleteNamespaceResponse>, tonic::Status> {
+        Err(tonic::Status::unimplemented(
```
alamb, line 72: 👍

iox_catalog/src/postgres.rs
lukebond: :exploding_head: did not know you could do that

alamb:
> once this PR is okay do we merge it without having designed a solution to namespace caching and re-creation of namespaces with the same name? seems to me we can deal with it separately

I agree it could be dealt with as a follow on,

Initially I suggest we make a follow on PR that 
1. creates a namespace and writes data into it
2. runs a query against the namespace
3.  and drops a namespace 
4. Tries to write and query it again

As long as nothing catastrophic happens (like the router getting wedged) we can probably make the recover semantics nicer as a follow on

iox_catalog/src/postgres.rs
alamb: 🤔  I just tried it locally and it turns out postgres doesn't actually support that 😢 

Sorry for my confusion

alamb:
```
postgres=# delete from namespace where id = 1;
ERROR:  update or delete on table "namespace" violates foreign key constraint "table_name_namespace_id_fkey" on table "table_name"
DETAIL:  Key (id)=(1) is still referenced from table "table_name".
postgres=# delete from namespace where id = 1 cascade;
ERROR:  syntax error at or near "cascade"
LINE 1: delete from namespace where id = 1 cascade;
```

lukebond: @alamb could you give a full example? simple adding it as you suggest is a syntax error. looking through the docs for postgres delete command it doesn't mention cascade.

lukebond: :+1: okay cool i'll just leave it as-is :smile:

lukebond: (sorry i didn't refresh before i posted the question!)

NGA-TRAN:
> if i'm adding ON DELETE CASCADE here, should we add it for everything else? seems to me it can be a separate PR if so

I think the `ON DELETE CASCADE` should be added to all FKs and can be done in a separate PR, maybe even before this PR is in. It is up to you

iox_catalog/migrations/20221215145057_namespace_delete_cascade.sql
lukebond: as discussed in another suggestion in this PR, this won't work in Postgres. however it's moot now because i need to soft-delete this anyway, instead.

alamb (approved):
I think it is worth considering @NGA-TRAN 's suggestion to add the constraints as part of another PR (so we can test its rollout individually)

After this PR is merged I do think this feature should have the end to end tests I described previously before we try to delete namespaces in production 👍

alamb:
I wonder if we ever implemented the test suggested in  https://github.com/influxdata/influxdb_iox/pull/6415#issuecomment-1354620546

PR #6416, opened 2022-12-15
feat(ingester2): optimal persist parallelism

### LINKED ISSUE #6146: Ingester:  Improve persist throughput  per-Partition
# perf issue

During my testing, I found that the Ingester module was very slow and the module did not resource bottlenecks(such as cpu, memory). when the Router had a lot of traffic, the Ingester TTBR metrics would skyrocket and pile up more and more data in the WriteBuffer(redpanda). Based on the inch test tool, some of my key metrics are as follows.

[benchmark data](https://docs.google.com/spreadsheets/d/1DdW0yu2x7Co91_OuFPDF0S2p76eDbYm2rFIEaSI-0Xs/edit?usp=sharing)


The current persistence core process is that the WriteBuffer data is pulled to the DataBuffer, and after certain conditions are met, it will be persisted in the Partition(IOx dimension), and the operation step is a single thread. The `persist` operation is very expensive, especially compact and write parquet_file, which is the reason why will pile up more and more data in the WriteBuffer.

```log
{"timestamp": "2022-10-31T10:20:31.188638Z", "level": "DEBUG", "fields":{"message": "Created parquet file", "num_batches":2, "num_rows": 6553700, "object_store_id": "90659f64-e53d-4879-9d3b-d0509082bf61", "partition_id": "PartitionId(85)", "write_batch_size":1024, "max_row _group_size":1048576}, "target": "parquet_file::serialize"}
{"timestamp": "2022-10-31T10:20:31.189402Z", "level": "DEBUG", "fields":{"message": "Uploading parquet to object store", "file_size": 91640446, "object_store_id": "90659f64-e53d-4879-9d3b-d0509082bf61", "partition_id": "PartitionId(85)", "total_time_to_create_parquet_ bytes": "69.42999872s"}, "target": "parquet_file::storage"}
```

Is it possible to optimize the current single-threaded processing flow into a multi-threaded concurrent process, with a concise flow as follows.

```log
   ┌───────────────┬───────────────┬───────────────┬───────────────┬───────────────┬───────────────┐              
   │               │               │               │               │               │               │              
   │    Buffer1    │    Buffer2    │    Buffer3    │   Buffer...   │     Pause     │    BufferN    │      pull    
   │     32MB      │     32MB      │     32MB      │     32MB      │               │     32MB      │  ◀───────────
   │               │               │               │               │               │               │              
   └───────┬───────┴───────┬───────┴───────┬───────┴───────────────┴───────────────┴───────────────┘              
           │               │               │                                                                      
           │               │               │                                                                      
           │               │               │         concurrent snapshot ...                                      
           │               │               │                                                                      
           │               │               │                                                                      
   ┌───────▼───────┬───────▼───────┬───────▼───────┐                                                              
   │               │               │               │                                                              
   │   Snapshot    │   Snapshot    │   Snapshot    │                                                              
   │  Persisting   │  Persisting   │  Persisting   │                                                              
   │               │               │               │                                                              
   └───────────────┴───────────────┴───────────────┘                                                              
           │               │               │                                                                      
           │               │               │                                                                      
           ▼               │               │                                                                      
max_sequence_number1       │               │                                                                      
                           ▼               │                                                                      
                 max_sequence_number2      │                                                                      
                                           ▼                                                                      
                                 max_sequence_number3                                                             
                                                                                                                  
                                                                                                                  
    ┌──────────────────────────────────────────────┐                                                              
    │                                              │                                                              
    │            Async update sort_key,            │                                                              
    │         (min)max_sequence_number,etc         │                                                              
    │                                              │                                                              
    └──────────────────────────────────────────────┘                                                              
                                                                                                                  
                                                                                                                  
                                                                                                                  
 ────────────────────────────────────────────────────────────────────────────────────────────▶  time              
```

- DataBuffer is split into multiple fixed size Buffer, so that when the maximum memory configuration is exceeded, pull data from WriteBuffer can be pause.
- A Buffer is a persist task, and the total number of persist tasks can be the number of CPU logical cores, with concurrent persistence operations to increase the consumption rate
- Synchronizing some data at regular intervals, eg: sort_key, max_sequence_number

domodwyer:
This PR changes the ingester2 persist system with the aim to perform optimal parallelisation of persist tasks within the parallelism bound; persists that can be parallelised are parallelised (across all workers), serialising only those jobs that **must** be serialised for correctness.

Some persist operations require updating the partition's sort key - this is not a commutative operation; updates must be applied one after another. We can cheaply and easily determine if a persist job will need to update the sort key before starting the persist job, and this PR takes advantage of that fact:

* Persist jobs that need to update a sort key are enqueued to a partition's assigned worker - serialising within the partition, but parallelising across partitions.

* Persist jobs that will not modify the partition's sort key are enqueued into a global work queue, from which all workers consume.

This provides optimal[^1] parallelisation; in the happy path, no sort key update is required and the work is parallelised (without head-of-line blocking) across N workers. In the case of a sort key update being necessary, the task is serialised **only with other persists that update the key for the same partition**. If a task doesn't update the sort key, it happens in parallel regardless of any ongoing, serialised persist jobs for the same partition.

### Back-pressure

Because there are multiple queues, and because we want to bound the total number of tasks in the system (not in any one queue) we give out a semaphore permit for each job, regardless of the queue it is placed in. This lets us effectively control the global number of jobs with a single number.

The back-pressure system introduced in #6403 also accounts for writers waiting to enqueue (waiting for a permit), in order to prevent an effectively unbounded number of persist operations.

All queues are unbounded to prevent them work from blocking when adding to a busy queue - the system itself is bounded with the semaphore.

### Multiple Queues?

Yes, multiple queues! An alternative solution would be to redo a persist operation when at the end you discover the sort key has been updated. It's expensive to redo a single persist operation (the most expensive operation the ingester does!), but it's actually worse than a simple "redo it" might first seem - the cost grows super-linearly with the number of parallel persist jobs!

Assuming the specified number of persist jobs are running in parallel for a single hot partition, you need this many retries:

<div align="center">

| Parallel Jobs | Retries Needed |
|--|--|
|5|15|
|6|21|
|7|27|
|..|..|
|10|51|

</div>

At 20 parallel persists for a single partition, you're looking at ~200 retries! It's easy to imagine a situation in which a user writing to a few hot partitions completely saturates the persist system with a retry strategy.

[^1]: Technically you could enqueue serialised work to an idle worker if the one it should go to is busy already, but this requires maintaining tracking state that constantly grows. I guess this is optimal for a **stateless** implementation you nit picker.

### Thanks

This closes #6146, opened by @aierui to discuss improvements to the persist system in the original ingester. This design is very similar to the one he proposed which was made possible by the recent work we've done on ingester2  ❤️

---

* refactor: inject persist started timestamp (e24d21255)

      Instead of recording the "enqueued_at" when initialising the
      PersistRequest, inject the value in.
      
      This lets us re-order the request construction while retaining accurate
      timing.

* feat(ingester2): optimal persist parallelism (933ab1f8c)

      This commit changes the behaviour of the persist system to enable
      optimal parallelism of persist operations, and improve the accuracy of
      the outstanding job bound / back-pressure.
      
      Previously all persist operations for a given partition were
      consistently hashed to a single worker task. This serialised persistence
      per partition, ensuring all updates to the partition sort key were
      serialised. However, this also unnecessarily serialises persist
      operations that do not need to update the sort key, reducing the
      potential throughput of the system; in the worst case of a single
      partition receiving all the writes, only one worker would be persisting,
      and the other N-1 workers would be idle.
      
      After this change, the sort key is inspected when enqueuing the persist
      operation and if it can be determined that no sort key update is
      necessary (the typical case), then the persist task is placed into a
      global work queue from which all workers consume. This allows for
      maximal parallelisation of these jobs, and the removes the per-worker
      head-of-line blocking.
      
      In the case that the sort key does need updating, these jobs continue to
      be consistently hashed to a single worker, ensuring serialised sort key
      updates only where necessary.
      
      To support these changes, the back-pressure system has been changed to
      account for all outstanding persist jobs in the system, regardless of
      type or assigned worker - a logical, bounded queue is composed together
      of a semaphore limiting the number of persist tasks overall, and a
      series of physical, unbounded queues - one to each worker & the global
      queue. The overall system remains bounded by the
      INFLUXDB_IOX_PERSIST_QUEUE_DEPTH value, and is now simpler to reason
      about (it is independent of the number of workers, etc).

### DIFF
--- a/clap_blocks/src/ingester2.rs
+++ b/clap_blocks/src/ingester2.rs
@@ -39,16 +39,15 @@ pub struct Ingester2Config {
     )]
     pub persist_max_parallelism: usize,
 
-    /// The maximum number of persist tasks that can be queued for each worker.
+    /// The maximum number of persist tasks that can be queued at any one time.
     ///
-    /// Note that each partition is consistently hashed to the same worker -
-    /// this can cause uneven distribution of persist tasks across workers in
-    /// workloads with skewed / hot partitions.
+    /// Once this limit is reached, ingest is blocked until the persist backlog
+    /// is reduced.
     #[clap(
-        long = "persist-worker-queue-depth",
-        env = "INFLUXDB_IOX_PERSIST_WORKER_QUEUE_DEPTH",
+        long = "persist-queue-depth",
+        env = "INFLUXDB_IOX_PERSIST_QUEUE_DEPTH",
         default_value = "250",
         action
     )]
-    pub persist_worker_queue_depth: usize,
+    pub persist_queue_depth: usize,
 }
--- a/ingester2/src/persist/backpressure.rs
+++ b/ingester2/src/persist/backpressure.rs
@@ -10,7 +10,7 @@ use crossbeam_utils::CachePadded;
 use observability_deps::tracing::*;
 use parking_lot::Mutex;
 use tokio::{
-    sync::mpsc,
+    sync::Semaphore,
     task::JoinHandle,
     time::{Interval, MissedTickBehavior},
 };
@@ -41,8 +41,8 @@ pub(crate) enum CurrentState {
 ///   * There are no outstanding enqueue operations (no thread is blocked adding
 ///     an item to any work queue).
 ///
-///   * All queues have at least half of their capacity free (being at most,
-///     half full).
+///   * The number of outstanding persist jobs is less than 50% of
+///     `persist_queue_depth`
 ///
 /// These conditions are evaluated periodically, at the interval specified in
 /// [`EVALUATE_SATURATION_INTERVAL`].
@@ -58,9 +58,9 @@ pub(crate) struct PersistState {
     /// reads.
     state: CachePadded<AtomicUsize>,
 
-    /// Tracks the number of async tasks waiting within
-    /// [`PersistHandle::queue_persist()`], asynchronously blocking to enqueue a
-    /// persist job.
+    /// Tracks the number of [`WaitGuard`] instances, which in turn tracks the
+    /// number of async tasks waiting within [`PersistHandle::queue_persist()`]
+    /// to obtain a semaphore permit and enqueue a persist job.
     ///
     /// This is modified using [`Ordering::SeqCst`] as performance is not a
     /// priority for code paths that modify it.
@@ -69,24 +69,40 @@ pub(crate) struct PersistState {
     ///     super::handle::PersistHandle::queue_persist()
     waiting_to_enqueue: Arc<AtomicUsize>,
 
+    /// The persist task semaphore with a maximum of `persist_queue_depth`
+    /// permits allocatable.
+    sem: Arc<Semaphore>,
+    persist_queue_depth: usize,
+
     /// The handle to the current saturation evaluation/recovery task, if any.
     recovery_handle: Mutex<Option<JoinHandle<()>>>,
 }
 
-/// Initialise a [`PersistState`] with [`CurrentState::Ok`].
-impl Default for PersistState {
-    fn default() -> Self {
+impl PersistState {
+    /// Initialise a [`PersistState`] with [`CurrentState::Ok`], with a total
+    /// number of tasks bounded to `persist_queue_depth` and permits issued from
+    /// `sem`.
+    pub(crate) fn new(persist_queue_depth: usize, sem: Arc<Semaphore>) -> Self {
+        // The persist_queue_depth should be the maximum number of permits
+        // available in the semaphore.
+        assert!(persist_queue_depth >= sem.available_permits());
+        // This makes no sense and later we divide by this value.
+        assert!(
+            persist_queue_depth > 0,
+            "persist queue depth must be non-zero"
+        );
+
         let s = Self {
             state: Default::default(),
             waiting_to_enqueue: Arc::new(AtomicUsize::new(0)),
             recovery_handle: Default::default(),
+            persist_queue_depth,
+            sem,
         };
         s.set(CurrentState::Ok);
         s
     }
-}
 
-impl PersistState {
     /// Set the reported state of the [`PersistState`].
     fn set(&self, s: CurrentState) -> bool {
         // Set the new state, retaining the most recent state.
@@ -134,27 +150,25 @@ impl PersistState {
     }
 
     /// Mark the persist system as saturated, returning a [`WaitGuard`] that
-    /// MUST be held during any subsequent async-blocking enqueue request
-    /// ([`mpsc::Sender::send()`] and the like).
+    /// MUST be held during any subsequent async-blocking to acquire a permit
+    /// from the persist semaphore.
     ///
-    /// Holding the guard over the `send()` await allows the saturation
+    /// Holding the guard over the `acquire()` await allows the saturation
     /// evaluation to track the number of threads with an ongoing enqueue wait.
-    pub(super) fn set_saturated<T>(s: Arc<Self>, persist_queues: Vec<mpsc::Sender<T>>) -> WaitGuard
-    where
-        T: Send + 'static,
-    {
-        // Increment the number of tasks waiting to push into a queue.
+    pub(super) fn set_saturated(s: Arc<Self>) -> WaitGuard {
+        // Increment the number of tasks waiting to obtain a permit and push
+        // into any queue.
         //
         // INVARIANT: this increment MUST happen-before returning the guard, and
-        // waiting on the queue send(), and before starting the saturation
-        // monitor task so that it observes this waiter.
+        // waiting on the semaphore acquire(), and before starting the
+        // saturation monitor task so that it observes this waiter.
         let _ = s.waiting_to_enqueue.fetch_add(1, Ordering::SeqCst);
 
         // Attempt to set the system to "saturated".
         let first = s.set(CurrentState::Saturated);
         if first {
             // This is the first thread to mark the system as saturated.
-            warn!("persist queues saturated, blocking ingest");
+            warn!("persist queue saturated, blocking ingest");
 
             // Always check the state of the system EVALUATE_SATURATION_INTERVAL
             // duration of time after the last completed evaluation - do not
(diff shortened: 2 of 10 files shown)


clap_blocks/src/ingester2.rs
```diff
@@ -39,16 +39,15 @@ pub struct Ingester2Config {
...
-    /// Note that each partition is consistently hashed to the same worker -
-    /// this can cause uneven distribution of persist tasks across workers in
-    /// workloads with skewed / hot partitions.
+    /// Once this limit is reached, ingest is blocked until the persist backlog
+    /// is reduced.
     #[clap(
-        long = "persist-worker-queue-depth",
-        env = "INFLUXDB_IOX_PERSIST_WORKER_QUEUE_DEPTH",
+        long = "persist-queue-depth",
+        env = "INFLUXDB_IOX_PERSIST_QUEUE_DEPTH",
```
domodwyer, line 48: This bounds the total number of persist jobs in the system, regardless of how they they get to a worker.

ingester2/src/persist/handle.rs
```diff
@@ -135,46 +157,63 @@ impl PersistHandle {
...
             exec,
             store,
             catalog,
         });
 
+        // Initialise the global queue.
+        //
+        // Persist tasks that do not require a sort key update are enqueued into
+        // this queue, from which all workers consume.
+        let (global_tx, global_rx) = async_channel::unbounded();
```
domodwyer, line 184: While the queue itself is unbounded, the system is not unbounded - we only allow the desired number of outstanding persist operations before marking the persist system as saturated and blocking ingest to clear the backlog. 

(there may be a few more jobs waiting to obtain a permit, so we technically do not bound the number of tasks exactly, it allows some overrun.)

ingester2/src/persist/backpressure.rs
```diff
@@ -206,60 +221,63 @@ impl Drop for WaitGuard {
...
+        // This check below ensures that the semaphore is at least half capacity
         // before marking the system as recovered.
-        let n_queues = queues
-            .iter()
-            .filter(|q| !has_sufficient_capacity(q.capacity(), q.max_capacity()))
-            .count();
-        if n_queues != 0 {
-            debug!(n_queues, "waiting for queues to drain");
+        let available = sem.available_permits();
+        let outstanding = persist_queue_depth.checked_sub(available).unwrap();
```
domodwyer, line 266: The saturation recovery task now only has to consider the single semaphore, and the number of outstanding waiters for a permit when deciding to mark the system as recovered.

ingester2/src/persist/handle.rs
```diff
@@ -206,48 +249,117 @@ impl PersistHandle {
...
+        // partitions.
+        //
+        // Likewise if it can be determined that a sort key is necessary, it
+        // must be serialised via the same mechanism.
+        //
+        // Do NOT attempt to fetch the sort key in this handler, as doing so
+        // would cause a large number of requests against the catalog in a short
+        // period of time (bounded by the number of tasks being enqueued).
+        // Instead, the workers fetch the sort key, bounding the number of
+        // queries to at most `n_workers`.
```
domodwyer, line 305: This is a more detailed/implementation orientated description of how the sort key handling w.r.t deferred loads works when choosing a queue.

ingester2/src/persist/handle.rs
```diff
@@ -206,48 +249,117 @@ impl PersistHandle {
...
+                }
+            }
+            None => {
+                // If no sort key is known (either because it was unresolved, or
+                // not yet set), the task must be serialised w.r.t other persist
+                // jobs for the same partition.
+                trace!(partition_id, "persist job has no known sort key");
+                self.assign_worker(r);
+            }
+        }
```
domodwyer, line 345: This is the entirety of the queue selection logic. It makes use of the existing `adjust_sort_key_columns()` logic in the schema crate used for compaction&persistence.

ingester2/src/persist/handle.rs
```diff
@@ -266,8 +378,44 @@ pub(super) struct Inner {
...
+            v = global_queue.recv() => {
+                match v {
+                    Ok(v) => v,
+                    Err(RecvError) => {
+                        // The global channel is closed.
+                        return
+                    },
+                }
+            }
+        };
```
domodwyer, line 417: This is the worker's job polling routine - it prefers the work it has been explicitly assigned, and then pulls from the global queue.

domodwyer (commented):
Almost all of the diff is tests / docs - the logic changes are:

* Queue selection in the persist handle
* Reading from two queues in the workers
* Using a semaphore to bound persist jobs

I highlighted both of those sections below.

ingester2/src/persist/handle.rs
```diff
@@ -206,48 +249,117 @@ impl PersistHandle {
...
-            Ok(()) => {} // Success!
-            Err(TrySendError::Closed(_)) => panic!("persist worker has stopped"),
-            Err(TrySendError::Full(r)) => {
-                // The worker's queue is full. Mark the persist system as being
-                // saturated, requiring some time to clear outstanding persist
-                // operations.
+        let enqueued_at = Instant::now();
+
+        // Try and acquire the persist task permit immediately.
+        let permit = match Arc::clone(&self.sem).try_acquire_owned() {
```
domodwyer, line 258: Here the semaphore is used to check for saturation instead of the queue (which is now unbounded) - the logic is otherwise the same (the queue is essentially a semaphore and an array, after all).

ingester2/src/persist/handle.rs
```diff
@@ -280,3 +424,392 @@ async fn run_task(inner: Arc<Inner>, mut rx: mpsc::Receiver<PersistRequest>) {
...
+    async fn test_persist_sort_key_no_update_necessary() {
+        let storage = ParquetStorage::new(Arc::new(InMemory::default()), StorageId::from("iox"));
+        let metrics = Arc::new(metric::Registry::default());
+        let catalog = Arc::new(MemCatalog::new(metrics));
+
+        let (mut handle, state) = PersistHandle::new(1, 2, Arc::clone(&EXEC), storage, catalog);
+        assert!(!state.is_saturated());
+
+        // Kill the workers, and replace the queues so we can inspect the
+        // enqueue output.
```
carols10cents, line 755: Ideally, some of this repetitive test setup code would get factored out into a test helper function or three, buuuuuut... MVP! MVP! 😁

carols10cents (approved):
Looks all right to me!

Issue #6411, opened 2022-12-15
FlightSQL Milestone 2: Support basic ad hoc queries


alamb:
It has three main parts:
- [ ]  Create a general Flight client by generalizing some of the clever code in the existing IOx Flight Client
- [ ]  Build a FlightSQL client on top of that client
- [ ]  Implement the server side of that

I think once I have the client part sorted out, building the server side will be straightforward and mechanical.

I intend to propose upstreaming the general flight client and the FightSQL client to the `arrow-filight` crate

In terms of a project plan, I hope to  have a PR up with this approach working in the next day or two, but then it will likely take me another ~ 5 working days to get it pieced up into PRs (with testing, documentation, etc)

PR #6422, opened 2022-12-16
feat: Implement creating Tickets in gRPC service

alamb:
re #6411 

# Rationale
Part of FlightSQL involves the server creating `Tickets` to return to the client. 

You can see how this code is used in https://github.com/influxdata/influxdb_iox/pull/6398

# Changes
Implement `IoxGetRequest::new()` and `IoxGetRequest::try_encode` to support programmatically creating tickets in the flight server


service_grpc_flight/src/request.rs
```diff
@@ -62,6 +71,39 @@ impl IoxGetRequest {
...
+        let read_info = match query {
+            RunQuery::Sql(sql_query) => proto::ReadInfo {
+                namespace_name,
+                sql_query,
+                query_type: QueryType::Sql.into(),
+            },
+            RunQuery::InfluxQL(influxql) => {
+                proto::ReadInfo {
+                    namespace_name,
+                    // field name is misleading
```
NGA-TRAN, line 91: 🤷 🙂 We support sql and its dialects

service_grpc_flight/src/request.rs
```diff
@@ -62,6 +71,39 @@ impl IoxGetRequest {
...
+                    sql_query: influxql,
+                    query_type: QueryType::InfluxQl.into(),
+                }
+            }
+        };
+
+        let mut buf = Vec::with_capacity(1024);
```
NGA-TRAN, line 98: What 1024 means? 1024 characters of SQL query? is it too short? Should we allow longer? Queries with subqueries will need a lot more

NGA-TRAN (approved):
LGTM. Thanks @alamb for making  the PR short

service_grpc_flight/src/request.rs
alamb: Yeah, I mostly didn't want to mess with the field name to avoid messing up the existing clients, though I *think* it would be ok to rename the field on protobuf as the field name isn't serialized 🤔

service_grpc_flight/src/request.rs
alamb: I am not sure why this has `1024` hard coded -- I was just cargo culting (following the existing pattern).
I removed it in febfb4e53

Issue #6420, opened 2022-12-16
ingester2 fails to start with empty catalog


mkmik:
Trying out #6406

```console
$ INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run ingester2 --grpc-bind=127.0.0.1:8042 --wal-directory /tmp/kafkaless  --catalog-dsn postgres:///iox_shared
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /Users/mkm/Projects/w/iox/ingester2/src/init.rs:170:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2022-12-16T04:57:17.582897Z  WARN iox_catalog::postgres: Dropping PostgresTxn w/o finalizing (commit or abort)
2022-12-16T04:57:17.583034Z  WARN executor: DedicatedExecutor dropped without calling shutdown()
2022-12-16T04:57:17.583211Z  WARN executor: DedicatedExecutor dropped without calling shutdown()
```

offending code: suggests it expects some existing data in the `topic` table https://github.com/influxdata/influxdb_iox/blob/a75e9aadf30d9336e1662cdb4ca91946ce1d1500/ingester2/src/init.rs#L165-L169

mkmik:
I ran iox standalone, inserted some data which populated the catalog db. It now says:

```console
$ INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run ingester2 --grpc-bind=127.0.0.1:8042 --wal-directory /tmp/kafkaless  --catalog-dsn postgres:///iox_shared
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `ShardId(13)`,
 right: `ShardId(1)`', /Users/mkm/Projects/w/iox/ingester2/src/init.rs:176:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2022-12-16T05:08:09.679405Z  WARN iox_catalog::postgres: Dropping PostgresTxn w/o finalizing (commit or abort)
2022-12-16T05:08:09.679538Z  WARN executor: DedicatedExecutor dropped without calling shutdown()
2022-12-16T05:08:09.679762Z  WARN executor: DedicatedExecutor dropped without calling shutdown()
```

mkmik:
A similar problem happens when pointing to an existing catalog (e.g. staging):

```
[...]

PR #6423, opened 2022-12-16
fix(ingester2): Make ingester2 work with empty or existing catalogs 

mkmik:
Make ingester2 work with empty or existing catalogs. Also adds a quickstart documentation that shows how to run it the i/r/q triplet locally.

Closes https://github.com/influxdata/influxdb_iox/issues/6420
I see that the `TRANSITION_SHARD_ID` is also asserted by the test utils, it seems that that's the place this assert belongs to, thus removing from the main init code.

The documentation is based on the gourmet copy pasta first cooked by @carols10cents in #6406

### DIFF
--- a/ingester2/README.md
+++ b/ingester2/README.md
@@ -1,0 +1,35 @@
+# ingester2
+
+## Quick run
+
+Set-up empty catalog db:
+
+```bash
+mkdir -p /tmp/iox/{wal,obj}
+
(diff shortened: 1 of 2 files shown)


pauldix (approved):

domodwyer:
We should restore the assert given the temporary catalog dance @mkmik :+1:

PR #6427, opened 2022-12-16
refactor: Extract common ArrowFlight client into iox_arrow_flight

alamb:
This is PR one of two left for https://github.com/influxdata/influxdb_iox/issues/6411. You can see how it goes together in https://github.com/influxdata/influxdb_iox/pull/6398

I am sorry it is so large

# Rationale:

The logic in the IOx flight client and the arrow-rs FlightSQL client is substantially duplicated. Specially the details of decoding the stream of FlightData that comes back into RecordBatches is very similar. 

I (strongly) believe most / all of the code in `iox_arrow_flight`  eventually belongs in the `arrow-flight` crate. However, I believe it will be faster to iterate on our own copy of this client and then contribute the pieces back upstream than to juggle PRs between repos.

Thus I plan to fork the code into IOx for some time while I work on it

https://github.com/apache/arrow-rs/issues/3301 tracks some of the improvements I will likely need to make

# Changes

 In order to upstream a useful Flight client, it needs to have general purpose features rather than IOx specific features (such as assuming Tickets are encoded protobuf, for example)

This PR extracts the common parts out of the iox flight client in the `influxdb_iox_client` crate into `FlightClient` in the `iox_arrow_flight` crate, and updates the `influxdb_iox_client` and `ingester` to use that refactored client.

cc @crepererum  (who I know is away but should probably review this when he returns)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ members = [
     "influxrpc_parser",
     "ingester",
     "ingester2",
+    "iox_arrow_flight",
```
alamb, line 29: See comments on this below -- this is a wrapped version of `arrow_flight` that contains items I intend to upstream to arrow-rs

influxdb_iox/src/commands/query_ingester.rs
```diff
@@ -15,7 +16,7 @@ pub enum Error {
     Formatting(#[from] influxdb_iox_client::format::Error),
 
     #[error("Error querying: {0}")]
-    Query(#[from] influxdb_iox_client::flight::Error),
+    Query(#[from] iox_arrow_flight::FlightError),
```
alamb, line 19: The ingester uses the lower level ArrowFlight client now directly

influxdb_iox_client/src/client/flight/low_level.rs
```diff
@@ -1,306 +0,0 @@
...
-//! - **Dictionary Batch:** A new dictionary for a given column is registered. An existing
-//!   dictionary for the same column will be overwritten. No app metadata is extracted. This
-//!   message is NOT visible to the user.
-//! - **Record Batch:** Record batch is created based on the current schema and dictionaries. This
-//!   fails if no schema was transmitted yet. App metadata is extracted is is presented -- together
-//!   with the record batch -- to the user.
-//!
-//! All other message types (at the time of writing: tensor and sparse tensor) lead to an error.
-
-use super::Error;
```
alamb, line 23: Most  of this code is moved into `iox_arrow_flight/src/client.rs`

influxdb_iox_client/src/client/flight/low_level.rs
```diff
@@ -1,306 +0,0 @@
...
-        {
-            Result::Ok(())
-        } else {
-            Result::Err(Error::HandshakeFailed)
-        }
-    }
-}
-
-#[derive(Debug)]
-struct PerformQueryState {
```
alamb, line 144: I separated out the "decode app_metadata as protobuf" and the streaming state machine but otherwise I mostly ported this and renamed the structs.

influxdb_iox_client/src/client/flight/low_level.rs
```diff
@@ -1,306 +0,0 @@
...
-        })
-    }
-
-    /// Returns next low-level message, or `None` if there are no further results available.
-    pub async fn next(&mut self) -> Result<Option<(LowLevelMessage, T)>, Error> {
-        let Self {
-            state, response, ..
-        } = self;
-
-        loop {
```
alamb, line 234: There is a partial copy of this code in arrow-rs: https://github.com/apache/arrow-rs/blob/89354ca7c3a66e05cefad77bd6350c8fad7b58bd/arrow-flight/src/sql/client.rs#L462-L529

querier/src/ingester/circuit_breaker.rs
```diff
@@ -206,13 +204,13 @@ enum Circuit {
...
 
-/// Wrapper around [`FlightClient`] that implements the [Circuit Breaker Design Pattern].
+/// Wrapper around [`IngesterFlightClient`] that implements the [Circuit Breaker Design Pattern].
 ///
 /// [Circuit Breaker Design Pattern]: https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern
 #[derive(Debug)]
 pub struct CircuitBreakerFlightClient {
     /// The underlying client.
-    inner: Arc<dyn FlightClient>,
+    inner: Arc<dyn IngesterFlightClient>,
```
alamb, line 214: There were too many things called "FlightClient" so I renamed this to `IngesterFlightClient`

influxdb_iox/tests/end_to_end_cases/cli.rs
```diff
@@ -476,9 +476,9 @@ async fn query_error_handling() {
                         .arg("drop table this_table_doesnt_exist")
                         .assert()
                         .failure()
-                        .stderr(predicate::eq(
-                            "Error querying: Error while planning query: This feature is not \
-                            implemented: DropTable\n",
+                        .stderr(predicate::str::contains(
```
alamb, line 479: I changed these because the output now contains more detail (not just the message but also the tonic status code)

```
The output now contains more detail (a tonic status, for example)
└── var as str: Error querying: Tonic(Status { code: InvalidArgument, message: "Error while planning query: This feature is not implemented: CREATE DATABASE", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 16 Dec 2022 18:59:25 GMT", "content-length": "0"} }, source: None })
```

influxdb_iox/tests/end_to_end_cases/ingester.rs
```diff
@@ -222,24 +230,56 @@ async fn ingester_flight_api_table_not_found() {
...
         .unwrap_err();
-    if let influxdb_iox_client::flight::Error::GrpcError(status) = err {
+    if let iox_arrow_flight::FlightError::Tonic(status) = err {
         assert_eq!(status.code(), tonic::Code::NotFound);
     } else {
         panic!("Wrong error variant: {err}")
     }
 }
+
+async fn next_message(
```
tustvold, line 256: I presume this is logic you intend to push into the mid-level flight client eventually?

influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -123,112 +127,139 @@ pub enum Error {
...
-/// resulting from executing an IOx Flight query.
-///
-/// Most users will not interact with this structure directly, but
-/// rather should use [`Client::sql`] or [`Client::influxql`]
-/// methods.
 #[derive(Debug)]
-pub struct PerformQuery {
-    inner: LowLevelPerformQuery<AppMetadata>,
-    got_schema: bool,
+/// Translates errors from FlightErrors to IOxErrors
```
tustvold, line 234: FWIW using `BoxStream` would allow writing this whole construct as

```
inner.map_err(Error::from).boxed()
```

iox_arrow_flight/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "iox_arrow_flight"
+description = "IOx specific parts of Apache Arrow Flight"
```
tustvold, line 3: I thought you intended to upstream this crate? Or is this describing what this crate will eventually just contain?

querier/src/ingester/flight_client.rs
```diff
@@ -161,14 +179,14 @@ impl SerializeFailureReason {
 
 /// Data that is returned by an ingester gRPC query.
 ///
-/// This is mostly the same as [`PerformQuery`] but allows some easier mocking.
+/// This is mostly the same as [`FlightDataStream`] but allows mocking in tests
```
tustvold, line 182: Not directly related to this PR, but this could simply be

```
type QueryData = BoxStream<'static, Result<(DecodedPayload, proto::IngesterQueryResponseMetadata), FlightError>
```

Which would additionally allow for using the futures combinators, and avoid some additional boxing

influxdb_iox/tests/end_to_end_cases/ingester.rs
```diff
@@ -222,24 +230,56 @@ async fn ingester_flight_api_table_not_found() {
...
+    decoded_data.map(|decoded_data| {
+        let DecodedFlightData { inner, payload } = decoded_data;
+
+        // extract the metadata from the underlying FlightData structure
+        let app_metadata = &inner.app_metadata[..];
+        let app_metadata: proto::IngesterQueryResponseMetadata =
+            Message::decode(app_metadata).unwrap();
+
+        (payload, app_metadata)
+    })
```
tustvold, line 270:
```suggestion
    let DecodedFlightData { inner, payload } = performed_query.next().await.unwrap()?;

    // extract the metadata from the underlying FlightData structure
    let app_metadata = &inner.app_metadata[..];
    let app_metadata: proto::IngesterQueryResponseMetadata =
    Message::decode(app_metadata).unwrap();

    Some((payload, app_metadata))
```

influxdb_iox/tests/end_to_end_cases/ingester.rs
```diff
@@ -222,24 +230,56 @@ async fn ingester_flight_api_table_not_found() {
...
+    performed_query: &mut FlightDataStream,
+) -> Option<(DecodedPayload, proto::IngesterQueryResponseMetadata)> {
+    let decoded_data = performed_query.next().await.unwrap();
+
+    decoded_data.map(|decoded_data| {
+        let DecodedFlightData { inner, payload } = decoded_data;
+
+        // extract the metadata from the underlying FlightData structure
+        let app_metadata = &inner.app_metadata[..];
+        let app_metadata: proto::IngesterQueryResponseMetadata =
```
tustvold, line 266: FWIW this type hint is unnecessary, as it can be inferred from the return type

querier/src/ingester/mod.rs
```diff
@@ -1350,14 +1351,20 @@ mod tests {
     use tokio::{runtime::Handle, sync::Mutex};
     use trace::{ctx::SpanContext, span::SpanStatus, RingBufferTraceCollector};
 
+    fn make_tonic_error(status: tonic::Status) -> influxdb_iox_client::flight::Error {
```
tustvold, line 1354: This feels like it could be a `From` conversion?

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+///   dictionary for the same column will be overwritten. This
+///   message is NOT visible.
+///
+/// - **Record Batch:** Record batch is created based on the current
+///   schema and dictionaries. This fails if no schema was transmitted
+///   yet.
+///
+/// All other message types (at the time of writing: e.g. tensor and
+/// sparse tensor) lead to an error.
+// TODO make this a real futures::Stream
```
tustvold, line 269: :+1: x :100:

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+                        ));
+                    }
+                    self.got_schema = true;
+                }
+                Some(DecodedPayload::RecordBatch(batch)) => return Ok(Some(batch)),
+                Some(DecodedPayload::None) => {
+                    // loop again
+                }
+            }
+        }
```
tustvold, line 231:
```suggestion
        while let Some(data) = self.inner.next().await? {
            match data.payload {
                DecodedPayload::Schema(_) => {
                    if self.got_schema {
                        return Err(FlightError::protocol(
                            "Unexpectedly saw multiple Schema messages in FlightData stream",
                        ));
                    }
                    self.got_schema = true;
                }
                DecodedPayload::RecordBatch(batch) => return Ok(Some(batch)),
                DecodedPayload::None => {
                    // loop again
                }
            }
        }
        Ok(None)
```

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /// Collect and return all `RecordBatch`es as a `Vec`
+    pub async fn collect(&mut self) -> Result<Vec<RecordBatch>> {
```
tustvold, line 235: Making this an actual Stream would allow eliminating this, and just using `TryStreamExt`

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+            state: None,
+            response,
+        }
+    }
+
+    /// Returns the result of decoding the next [`FlightData`] message
+    /// from the server, or `None` if there are no further results
+    /// available.
+    pub async fn next(&mut self) -> Result<Option<DecodedFlightData>> {
+        loop {
```
tustvold, line 291: This could also have the `while let Some` treatment as the method above

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+}
+
+/// A stream of [`RecordBatch`]es from from an Arrow Flight server.
+///
+/// To access the lower level Flight messages directly, consider
+/// calling [`Self::into_inner`] and using the [`FlightDataStream`]
+/// directly.
+#[derive(Debug)]
+pub struct FlightRecordBatchStream {
```
tustvold, line 183: Once we move to streams I think this can be written as

```
flight_data_stream.try_filter(|data| matches!(data.payload, DecodedPayload::RecordBatch))
```

iox_arrow_flight/src/client.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /// return the metadata field of the inner flight data
+    pub fn app_metadata(&self) -> &[u8] {
+        &self.inner.app_metadata
+    }
+}
+
+/// The result of decoding [`FlightData`]
+#[derive(Debug)]
```
tustvold, line 420: I wonder if this is lower-level than we actually need? Does anything actually care about receiving separate `Schema` and `RecordBatch`, or is just providing `RecordBatch` sufficient? Is this because of app_metadata?

tustvold (approved):
I think it would be really good to refactor this code to use `futures::BoxStream` as it would allow implementing a lot of the code using combinators instead of fairly verbose async functions. Approving as I don't think this mainly just moves pre-existing code around

influxdb_iox/tests/end_to_end_cases/ingester.rs
alamb: Maybe -- I would be curious what you think on that matteer

The distinction is that Arrow Flight itself does not imply the use of protobuf in the ticket or app data field. FlightSQL does put encoded protobuf messages there, but Flight itself just treats them as opaque.

Thus I was not sure how much the mid level flight client should assume the opaque bytes were protobuf encoded messages. 🤔

influxdb_iox_client/src/client/flight/mod.rs
alamb: 👍 Let me give that a try

iox_arrow_flight/Cargo.toml
alamb: Good point -- clarified to

> description = "Extended Apache Arrow Flight (intended to be upstreamed to arrow-rs)"

alamb (commented):
Thank you for the review @tustvold 

I will take a shot at trying to update this to use streams

iox_arrow_flight/src/client.rs
alamb: The querier <--> ingester link needs to get different schemas (because the different partitions may have different subsets of columns). So the reason that @crepererum  originally exposed this low level interface, as I recall, was to handle that usecase

querier/src/ingester/mod.rs
alamb: good point -- will try

iox_arrow_flight/src/client.rs
alamb: Done in 224f8c452

querier/src/ingester/mod.rs
alamb: In f4f864729

alamb:
My plan is to merge this PR and attempt to "streamify" it as a follow on

alamb:
Follow on PR for futures::Streams https://github.com/influxdata/influxdb_iox/pull/6438

crepererum (commented):
Looks good :)

Issue #6421, opened 2022-12-16
querier2 cannot query data after persist


mkmik:
__Steps to reproduce and actual behavior:__

spin up `ingester2`, `router2` and `querier` after applying a small diff to work around #6420:

```diff
--- ingester2/src/init.rs
+++ ingester2/src/init.rs
@@ -173,7 +173,7 @@ pub async fn new(
         .create_or_get(&topic, TRANSITION_SHARD_INDEX)
         .await
         .expect("create transition shard");
-    assert_eq!(s.id, TRANSITION_SHARD_ID);
+    //    assert_eq!(s.id, TRANSITION_SHARD_ID);
     txn.commit().await.expect("commit transition shard");
 
     // Initialise the deferred namespace name resolver.
```

Set-up empty catalog db:

```bash
mkdir -p /tmp/iox/{wal,obj}

createdb iox_shared
./target/debug/influxdb_iox catalog setup --catalog-dsn postgres:///iox_shared

# there has to exist one "topic", see https://github.com/influxdata/influxdb_iox/issues/6420
psql 'dbname=iox_shared options=-csearch_path=public,iox_catalog' -c "insert into topic (name) values ('iox-shared')"
```

Run ingester2:

```bash
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run ingester2 --api-bind=127.0.0.1:8081 --grpc-bind=127.0.0.1:8042 --wal-directory /tmp/iox/wal  --catalog-dsn postgres:///iox_shared --object-store=file --data-dir=/tmp/iox/obj -v 
```

Run router2:

```bash
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run router2 --api-bind=127.0.0.1:8080 --grpc-bind=127.0.0.1:8085 --ingester-addresses=127.0.0.1:8042 --catalog-dsn postgres:///iox_shared -v
```

Run querier:

```bash
INFLUXDB_IOX_RPC_MODE=2 ./target/debug/influxdb_iox run querier --ingester-addresses=http://127.0.0.1:8042 --api-bind 127.0.0.1:8083 --grpc-bind 127.0.0.1:8082 --catalog-dsn postgres:///iox_shared --object-store=file --data-dir=/tmp/iox/obj -v
```

Load some data:

```bash
[...]

PR #6428, opened 2022-12-16
fix: Correct *some* problems mkm found with ingester2/querier

carols10cents:
Connects to #6421. Does not completely solve yet.

# Commit 1

The first problem both Dom and I suspected was that the transition shard ID was not matching the hardcoded hack value. The first commit in this PR changes the code to still hardcode the transition shard *index* to 1, but to look up in the catalog what the transition shard *id* ends up being and use that value. This didn't seem to affect the problem in the issue at all as far as I could tell, but was incorrect, so good to change anyway. Ultimately, shards will completely go away when the transition is complete.

# Commit 2

The next problem I noticed is in the test mkm was doing, the steps were:

- Enqueue data
- Query, data is returned
- Ingester persists
- Query, data is not returned

I noticed the querier was not getting any metadata about the ingester UUID and number of files persisted from the second query when it queries the ingester. The ingester was just not returning anything if there were no record batches to return from its memory, but we need to always send back the metadata.

The second commit in this PR changes that, and I'm now able to see the querier get the metadata and the function I expect to say "should I refresh the cache?" is now returning `true` for the second request, but I'm still not seeing the data returned, so there's still another bug somewhere.

I can't write a test for this because ingester2 doesn't implement the write info service yet, so there's not a convenient way for tests to tell that data has been persisted or not.

# Commit 3

The third commit removes the `test_rpc_write` CI job; Dom already made it un-required again. Since the feature flag has been removed, it's exactly the same as the `test` job and is just wasting resources.

### DIFF
--- a/ingester2/src/buffer_tree/table.rs
+++ b/ingester2/src/buffer_tree/table.rs
@@ -267,7 +267,7 @@ where
         );
 
         // Gather the partition data from all of the partitions in this table.
-        let partitions = self.partitions().into_iter().filter_map(move |p| {
+        let partitions = self.partitions().into_iter().map(move |p| {
             let mut span = SpanRecorder::new(span.clone().map(|s| s.child("partition read")));
 
             let (id, completed_persistence_count, data) = {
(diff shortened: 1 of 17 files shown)


ingester2/src/buffer_tree/table.rs
```diff
@@ -263,38 +267,40 @@ where
...
+                    } else {
+                        Projection::Some(columns.as_ref())
+                    };
+
+                    let data = Box::pin(MemoryStream::new(
+                        data.project_selection(selection).into_iter().collect(),
+                    ));
+                    PartitionResponse::new(data, id, None, completed_persistence_count)
+                }
+                None => PartitionResponse::new_no_batches(id, None, completed_persistence_count),
```
carols10cents, line 299: Dom: [`datafusion_util::MemoryStream` must have at least one batch](https://github.com/influxdata/influxdb_iox/blob/97d90f5615a1370b43a1548c827dafde5f32c168/datafusion_util/src/lib.rs#L123-L125), so even though creating a `MemoryStream` with an empty vec *compiles*, it panics at runtime.

pauldix (approved):
Worth having @domodwyer take a look, but I think it's ok for merging.

ingester2/src/buffer_tree/table.rs
domodwyer: What a nice surprise!

domodwyer:
I took a look over - I think we probably didn't need the ShardId code churn but doesn't hurt - I hope we can remove it soon.

I opened a PR that refactors the constructor stuff to work around the `MemoryStream` stuff and some minor other bits https://github.com/influxdata/influxdb_iox/pull/6436

PR #6436, opened 2022-12-19
refactor(ingester2): query types

domodwyer:
Minor refactoring to the query path types.

---

* fix: show lack of partition data in query output (13ed3f9ac)

      Show that a PartitionResponse does not contain data in the Debug output.

* refactor: remove max seqnum in PartitionResponse (c1db76bf9)

      Removes the redundant max_persisted_sequence_number in
      PartitionResponse, which was functionally replaced with
      completed_persistence_count for the Querier's parquet file discovery
      instead.

* refactor: single PartitionResponse constructor (b87d572e4)

      Removes the PartitionResponse::new_no_batches() constructor, instead
      using an Option-wrapped data. Before that would have been confusing
      (many Option in the constructor signature) but now there's only one!

* refactor: avoid double flattening stream (371857399)

      Changes the into_record_batches() method to avoid creating an extra
      stream out of the Option that must be flattened (iterating over the
      option vs. filtering out all None first).


ingester2/src/query/partition_response.rs
```diff
@@ -23,7 +23,13 @@ pub(crate) struct PartitionResponse {
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("PartitionResponse")
-            .field("batches", &"<SNAPSHOT STREAM>")
+            .field(
+                "batches",
+                &match self.batches {
+                    Some(_) => "<SNAPSHOT STREAM>",
+                    None => "<NO DATA>,",
+                },
+            )
```
carols10cents, line 29: LOVE IT. I think I had something like this locally at some point but accidentally threw it away in some of my flailings.

ingester2/src/query/partition_response.rs
```diff
@@ -38,20 +38,12 @@ impl std::fmt::Debug for PartitionResponse {
 
 impl PartitionResponse {
     pub(crate) fn new(
-        batches: SendableRecordBatchStream,
+        data: Option<SendableRecordBatchStream>,
```
carols10cents, line 41: I swear I tried this exact thing and it wouldn't compile but maybe I actually had a different problem. I'm glad it works though!

ingester2/src/query/response.rs
```diff
@@ -51,7 +51,7 @@ impl QueryResponse {
     /// Reduce the [`QueryResponse`] to a stream of [`RecordBatch`].
     pub(crate) fn into_record_batches(self) -> impl Stream<Item = Result<RecordBatch, ArrowError>> {
         self.into_partition_stream()
-            .flat_map(|partition| futures::stream::iter(partition.into_record_batch_stream()))
+            .filter_map(|partition| future::ready(partition.into_record_batch_stream()))
```
carols10cents, line 54: THANK YOU. ❤️ I just stopped once I got to the first thing that compiled; this is much better.

carols10cents (approved):

PR #6438, opened 2022-12-19
refactor: Make arrow flight client return `futures::Streams`

alamb:
# Rationale
re https://github.com/influxdata/influxdb_iox/issues/6411

This is a follow on left over from https://github.com/influxdata/influxdb_iox/pull/6427

Recommend viewing with whitespace blind diff: https://github.com/influxdata/influxdb_iox/pull/6438/files?w=1

# Changes
1. use `futures::Stream` for client streams


influxdb_iox_client/src/client/flight/mod.rs
```diff
@@ -237,7 +240,8 @@ impl Client {
 }
 
 #[derive(Debug)]
-/// Translates errors from FlightErrors to IOxErrors
+/// Translates errors from FlightErrors to IOx client errors,
```
alamb, line 243: I still like the interface of "let the caller get the inner stream if needed" and for that I thought keeping a separate struct was simpler than using the stream combinator (`map_err`)

tustvold: MapErr has an into_inner method - https://docs.rs/futures/latest/futures/stream/struct.MapErr.html#method.into_inner

iox_arrow_flight/src/client.rs
```diff
@@ -276,93 +292,127 @@ impl FlightDataStream {
...
-                    return Err(FlightError::protocol(format!("Unexpected message: {name}")));
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<Self::Item>> {
+        if self.done {
+            return Poll::Ready(None);
+        }
+        loop {
+            let res = ready!(self.response.poll_next_unpin(cx));
```
tustvold, line 391: FWIW this can be written more concisely as

```
return Poll::Ready(match res {
    None => {
        self.done = true;
        None // inner is exhausted
    }
    Some(data) => Some(match data {
        Err(e) => Err(FlightError::Tonic(e)),
        Ok(data) => match self.extract_message(data) {
            Ok(Some(extracted)) => Ok(extracted),
            Ok(None) => continue, // Need next input message
            Err(e) => Err(e),
        },
    }),
});
```

Whether that is more legible is likely subjective

querier/src/ingester/flight_client.rs
```diff
@@ -184,7 +185,7 @@ impl SerializeFailureReason {
 pub trait QueryData: Debug + Send + 'static {
```
tustvold, line 185: Are you planning to replace this with `BoxStream` in a subsequent PR?

alamb: I tried (very briefly) https://github.com/influxdata/influxdb_iox/pull/6427#discussion_r1052334164 but have decided to give up for now 😢

PR #6443, opened 2022-12-19
fix: Disable parquet file filtering in the querier based on max seq num in RPC write mode

carols10cents:
Connects to #6421.

It seems to fix that issue for me locally, but I'm hesitant to say it's FIXED fixed til I get confirmation from @mkmik 🤞🏻 🤞🏻 🤞🏻 

Per a call with @pauldix, this changes the querier state reconciler to not call the `filter_parquet_files` function when the querier is running in RPC write mode.


pauldix (approved):

carols10cents:
Looks like this code was added in https://github.com/influxdata/influxdb_iox/pull/4355

mkmik:
@carols10cents I didn't try the reproducer locally yet.

But in our testing system on k8s, queries don't include recent results. Perhaps that's another issue though.

I'm just writing some data into a bucket and querying it back and I notice that sometimes I see some results, sometimes I see zero results. all the times I see stale results. There are 3 queriers deployed so I assume that depending on which querier the query lands to, I either get 0 or some results.

Issue #6439, opened 2022-12-19
Unserialised cross-node sort key updates


domodwyer:
## Sort Keys

A sort key defines the sort order of data in a parquet file. The ingester produces the sort key, the compactors use the sort key when compacting to optimise the data sort order for the querier, which uses the sort key during queries (chunk deduplication). We aim to avoid re-sorting data as much as possible for performance reasons.

When an ingester observes a new column, the sort key is updated according to the following logic:

https://github.com/influxdata/influxdb_iox/blob/94c2f94ea150d4ecd82e43d97ca15d2efa284f8a/schema/src/sort.rs#L478-L485

This is not commutative, requiring updates to the key be serialised, which in turn also requires persistence operations to be serialised (at least, those that update the sort key). This is handled in ingester2 by serialising persist operations per partition that [update the sort key](https://github.com/influxdata/influxdb_iox/pull/6416).

## The Problem

Previously (ingester1) all writes for a given `(namespace, table)` tuple were (effectively) consistently hashed to the same ingester instance. This meant that so long the instance serialised persist operations / sort key updates internally, updates to the sort key were serialised.

In ingester2, the current implementation of the router round-robins requests over all N routers without any consistency of data placement. This causes multiple ingester instances to need to update the sort key for the same partition, potentially at the same time:

| Ingester A| Ingester B|
|--|--|
| write A=1| |
| | write B=1|
| persist (`sort_key=A,time`)| |
| | persist (`sort_key=B,time`)|

This sort key update for B conflicts with A. This example could be mitigated as the column additions are disparate (you could merge them with "lexicographically greater wins" to consistently produce `A,B`). This doesn't hold for more complex examples, such as two instances deriving different orders for the same columns: `A,B,C` and `A,C,B`.

### Replication Complications

This also makes it difficult to support persisting from replicas should a primary be (believed to be) unavailable. It's impossible to know for sure that a primary is down, and therefore that the sort key will not be updated concurrently.

### Retries

We could read the sort key from the catalog, generate the parquet file in the ingester (sorting the data according to the sort key) and then open a transaction, check the sort key and abort if it has changed (and differs). Retry infinitely until the transaction succeeds. 

[...]

PR #6448, opened 2022-12-20
fix(ingester2): handle concurrent sort key updates

domodwyer:
Handle concurrent, cross-node catalog sort key updates as described in https://github.com/influxdata/influxdb_iox/issues/6439.

Closes https://github.com/influxdata/influxdb_iox/issues/6439.

---

* fix(ingester2): handle concurrent sort key updates (f64ffbe03)

      Allow an ingester2 instance to tolerate concurrent partition sort key
      updates in the catalog.
      
      A persist job is optimistically executed with the locally cached sort
      key. If an ingester2 instance observes a concurrent update, it aborts
      both the sort key update, and the overall persist operation (before
      making the parquet file visible) and retries the operation with the
      newly observed sort key. Concurrent sort key updates are theorised to be
      relatively rare overall.
      
      Any orphaned parquet files uploaded as part of a persist job that aborts
      due to a concurrent sort key update are eventually removed by the
      (external) object store GC task.
      
      See https://github.com/influxdata/influxdb_iox/issues/6439

* feat(persist): accept concurrent matching updates (e083f3276)

      As an optimisation, allow a persist task to progress if it observes a
      concurrent catalog sort key update that exactly matches the sort key it
      was committing.

### DIFF
--- a/ingester2/src/persist/context.rs
+++ b/ingester2/src/persist/context.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::{ops::ControlFlow, sync::Arc};
 
 use backoff::Backoff;
 use data_types::{
@@ -11,6 +11,7 @@ use observability_deps::tracing::*;
 use parking_lot::Mutex;
 use parquet_file::metadata::IoxMetadata;
 use schema::sort::SortKey;
+use thiserror::Error;
 use tokio::{
     sync::{oneshot, OwnedSemaphorePermit},
     time::Instant,
@@ -29,9 +30,17 @@ use crate::{
 
 use super::handle::Inner;
 
-/// An internal type that contains all necessary information to run a persist task.
-///
-/// Used to communicate between actor handles & actor task.
+/// Errors a persist can experience.
+#[derive(Debug, Error)]
+pub(super) enum PersistError {
+    /// A concurrent sort key update was observed and the sort key update was
+    /// aborted.
+    #[error("detected concurrent sort key update")]
+    ConcurrentSortKeyUpdate,
+}
+
+/// An internal type that contains all necessary information to run a persist
+/// task.
 #[derive(Debug)]
 pub(super) struct PersistRequest {
     complete: oneshot::Sender<()>,
@@ -179,6 +188,8 @@ impl Context {
     }
 
     pub(super) async fn compact(&self) -> CompactedStream {
+        let sort_key = self.sort_key.get().await;
+
         debug!(
             namespace_id = %self.namespace_id,
             namespace_name = %self.namespace_name,
@@ -186,6 +197,7 @@ impl Context {
             table_name = %self.table_name,
             partition_id = %self.partition_id,
             partition_key = %self.partition_key,
+            ?sort_key,
             "compacting partition"
         );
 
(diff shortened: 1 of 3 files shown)


ingester2/src/persist/handle.rs
```diff
@@ -378,6 +379,42 @@ pub(super) struct Inner {
...
+///               │                      │no
+///               └─────┬────────────────┘
+///               ┌─────▽─────┐
+///               │ADD PARQUET│
+///               │TO CATALOG │
+///               └─────┬─────┘
+///             ┌───────▽──────┐
+///             │NOTIFY PERSIST│
+///             │JOB COMPLETE  │
+///             └──────────────┘
```
domodwyer, line 416: This describes the new persist job execution flow.

ingester2/src/persist/handle.rs
```diff
@@ -418,13 +455,50 @@ async fn run_task(
...
+        // catalog, attempt to update the catalog with a compare-and-swap
+        // operation; if this update fails due to a concurrent sort key update,
+        // the compaction must be redone with the new sort key and uploaded
+        // before continuing.
+        let parquet_table_data = loop {
+            match compact_and_upload(&ctx).await {
+                Ok(v) => break v,
+                Err(PersistError::ConcurrentSortKeyUpdate) => continue,
+            };
+        };
```
domodwyer, line 471: This is the retry loop that keeps driving compaction & sort key updates until a sort key CAS operation succeeds.

ingester2/src/persist/context.rs
```diff
@@ -289,11 +301,118 @@ impl Context {
         (catalog_sort_key_update, parquet_table_data)
     }
 
-    pub(crate) async fn update_database(
-        self,
-        sort_key_update: Option<SortKey>,
+    pub(super) async fn update_catalog_sort_key(
```
domodwyer, line 304: This block of code was pulled out of the nested `if Some(update) = sort_key_update` and moved into it's own function.

ingester2/src/persist/context.rs
```diff
@@ -289,11 +301,118 @@ impl Context {
...
+                            // observed value for the next attempt.
+                            self.partition
+                                .lock()
+                                .update_sort_key(Some(SortKey::from_columns(observed)));
+                            // Stop the retry loop with an error.
+                            ControlFlow::Break(Err(PersistError::ConcurrentSortKeyUpdate))
+                        }
+                    }
+                }
+            })
```
domodwyer, line 377: Here we retry any error other than `CasFailure::ValueMismatch`, which causes the retry loop to stop and return control to the caller - this leads to the persist job being retried by the loop I have commented on above.

This "retry forever" matches the existing logic (also in ingester1).

ingester2/src/persist/context.rs
```diff
@@ -289,11 +301,143 @@ impl Context {
...
+                                table_name = %self.table_name,
+                                partition_id = %self.partition_id,
+                                partition_key = %self.partition_key,
+                                expected=?old_sort_key,
+                                ?observed,
+                                update=?new_sort_key_str,
+                                "detected matching concurrent sort key update"
+                            );
+                            ControlFlow::Break(Ok(()))
+                        }
```
domodwyer, line 367: Here we allow a concurrent sort key update that exactly matches the sort key this thread was trying to commit. Happy days!

pauldix (approved):

domodwyer:
CI seems wedged, pushing a NOP...

PR #6449, opened 2022-12-20
fix: query annotations and params

cole-abbeduto-particle:
**Fix** `query.dialect.annotations` should be a `Vec<String>`, not a `String`. In the current state if you try to send it as just a String you get a go error response.

**Add** `query.params` which utilizes the params feature of influx's query API to allow for using things like `params.param_name` in your query string. This uses a new Param enum to restrict the data types to `String` and `serde_json::Value::Number` which is what is supported accroding to the query API docs [here](https://docs.influxdata.com/influxdb/cloud/query-data/parameterized-queries/#supported-parameter-data-types).


wolffcm:
Hi @cole-abbeduto-particle, as you may have noticed the Rust implementation of the InfluxDB v2 driver is experimental and has not been updated in a while. Still, your changes seem reasonable, so I am going to merge this. Thank you for this contribution!

wolffcm (approved):

PR #6454, opened 2022-12-21
feat(ingester2): persist saturation metric

domodwyer:
When this happens we stop accepting writes, so it's a good starting point!

---

* feat(ingester2): persist saturation metric (679c6a789)

      Expose a metric ("ingester_persist_saturated_duration_ns") that records
      the cumulative duration of time the persist system has spent in the
      "saturated" state.

### DIFF
--- a/ingester2/src/persist/backpressure.rs
+++ b/ingester2/src/persist/backpressure.rs
@@ -80,7 +80,7 @@ pub(crate) struct PersistState {
 
     /// A counter tracking the number of nanoseconds the state value is set to
     /// [`CurrentState::Saturated`].
-    saturated_duration_ms: DurationCounter,
+    saturated_duration: DurationCounter,
 }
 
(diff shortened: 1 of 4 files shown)


ingester2/src/persist/backpressure.rs
```diff
@@ -92,12 +101,20 @@ impl PersistState {
             "persist queue depth must be non-zero"
         );
 
+        let saturated_duration_ms = metrics
```
carols10cents, line 104: The name of this variable contains `ms`, the metric name says `ns`, the comment above on line 81 says "nanoseconds"? And really it's counting `Duration`s, which can be expressed in whatever unit...?

carols10cents (approved):
This is good to go after removing the units (that conflict and are irrelevant anyway) from names and comments. Love the tests!

PR #6455, opened 2022-12-21
fix: Allow partitions from ingesters to overlap in RPC write mode

carols10cents:
Connects to #6421, [namely this comment](https://github.com/influxdata/influxdb_iox/issues/6421#issuecomment-1361580387).

### DIFF
--- a/querier/src/table/mod.rs
+++ b/querier/src/table/mod.rs
@@ -208,7 +208,8 @@ impl QuerierTable {
         );
 
         let (predicate, retention_delete_pred) = match self.namespace_retention_period {
-            // The retention is not fininte, add predicate to filter out data outside retention period
+            // The retention is not fininte, add predicate to filter out data outside retention
+            // period
             Some(retention_period) => {
                 let retention_time_ns = self
                     .chunk_adapter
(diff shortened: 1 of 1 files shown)


querier/src/table/mod.rs
```diff
@@ -442,7 +449,19 @@ impl QuerierTable {
             .map(|sharder| vec![**sharder.shard_for_query(&self.table_name, &self.namespace_name)]);
 
         // get cached table w/o any must-coverage information
-        let Some(cached_table) = self.chunk_adapter.catalog_cache().namespace().get(Arc::clone(&self.namespace_name), &[], span_recorder.child_span("get namespace")).await.and_then(|ns| ns.tables.get(&self.table_name).cloned()) else {return Ok(vec![])};
```
domodwyer, line 445: WRAP SOME LINES YEAH?

querier/src/table/mod.rs
```diff
@@ -459,19 +478,21 @@ impl QuerierTable {
...
-                    return Err(Error::IngestersOverlap {
-                        ingester1: Arc::clone(o.get()),
-                        ingester2: Arc::clone(partition.ingester()),
-                        partition: partition.partition_id(),
-                    })
-                }
-                Entry::Vacant(v) => {
-                    v.insert(Arc::clone(partition.ingester()));
+        if !self.rpc_write() {
+            // check that partitions from ingesters don't overlap
```
domodwyer, line 482: I wonder if this comment should have a "because if they do, X will explode when Y happens" type comment...

domodwyer (approved):

querier/src/table/mod.rs
carols10cents: Yeah I don't think rustfmt understands let/else yet :(

querier/src/table/mod.rs
carols10cents: Shoulda coulda woulda, but it should get deleted soon.... 🙏🏻

PR #6456, opened 2022-12-21
fix(persist): invalidate cached sort key

domodwyer:
Correctly handle sort key conflicts. Unblocks https://github.com/influxdata/influxdb_iox/issues/6421#issuecomment-1361627546.

Fortunately I was writing tests for the persist system this afternoon... I will PR soon.

---

* fix(persist): invalidate cached sort key (23b781f27)

      The sort-key conflict path invalidated the cached sort key in the
      PartitionData, but not the cached sort key in the persist's Context. Now
      both are invalidated.


carols10cents (approved):
This fixes the infinite persisting retry loop I was seeing locally!

Issue #6390, opened 2022-12-13
Implement Native FlightSQL in IOx


alamb:
Follow on on to the design hashed out in  https://github.com/influxdata/influxdb_iox/issues/6102

# Use Case

We would like to expand the tools available to our customers to execute queries using IOx. In particular, we would like IOx to support the [JDBC and ODBC drivers](https://arrow.apache.org/blog/2022/11/01/arrow-flight-sql-jdbc/), which are standard APIs for communicating with database servers. These drivers utilize the [FlightSQL](https://arrow.apache.org/blog/2022/02/16/introducing-arrow-flight-sql/) interface to interface with the database server.

# Design Doc
InfluxData only: https://docs.google.com/document/d/1MtasrcQ-iBGZ2WgiXK-Lk6DevygtuSewoihOWag0kGU/edit#

# Milestones
- [x] #6391
- [x] #6411
- [x] #6471
- [ ] ~#6789~
- [x] #7210

# Follow on work (moved out of scope from this initial project): 
- [ ] #6789
- [ ] #7265

# Post-Implementation Tasks
IOx supports both FlightSQL and its custom (1 round trip) Flight based SQL, and that will still be the case after this project
- [ ] ~Migrate Flight RPC clients (queryrouterd) in IDPE to FlightSQL~
- [x] Expose FlightSQL to customers
- [ ] ~Remove FlightRPC service from IOx~

alamb:
Initial PR to start upstreaming flight client is here https://github.com/apache/arrow-rs/pull/3378

alamb:
# Project Update

As of 2023-02-1 the `influxdb_iox` binary (as opposed to the IOx service exposed on InfluxDB cloud2) can run SQL via the FlightSQL protocol (for example, using the arrow flight JDBC drivers). The functionality is covered by [end to end tests](https://github.com/influxdata/influxdb_iox/blob/967450f0de6ba453f5c46477b795ac6f455bfa91/influxdb_iox/tests/end_to_end_cases/flightsql.rs#L183-L260)

The major known remaining  gaps in IOx's FlightSQL implementation are:
1. Support for parameterized queries (milestone 4)
2. Metadata API (milestone 5)

## Parameterized Queries

The term "parameterized query" means a query that has a placeholder like  `?` such as

```sql
SELECT * FROM cpu where time > ?
```

The values for these parameters are provided by the client in a separate API call.

[...]

PR #6460, opened 2022-12-22
feat: Extract encoding from `RecordBatch` --> `FlightData` from flight implementations

alamb:
re #6390 

# Rationale:
The details of converting `RecordBatch`es into streams of `FlightData` is common across all flight implementations (including flight SQL) and thus belongs upstream in arrow-rs.

In addition,  I also need something like this upstream in arrow-rs to write reasonable tests for do_get

Also, I think getting it into arrow-rs will allow more expert analysis of this code / improvement over time by the community (such as currently having to hydrate dictionaries - #1318, etc )

# Changes
1. Extracts the general purpose translation into a `iox_arrow_flight::encode` in preparation to upstream

# Notes
The flight server code in `ingester` and `ingester2` needs a some love, but I did not give it any in this PR. Once the new write path has settled down, I would like to look into cleaning up / generalizing that code more.


arrow_util/src/optimize.rs
```diff
@@ -95,127 +95,17 @@ fn optimize_dict_col(
     Ok(Arc::new(new_dictionary.to_arrow(new_keys, nulls)))
 }
 
-/// Hydrates a dictionary to its underlying type
```
alamb, line 98: moved into `iox_arrow_flight` in preparation for upstreaming

ingester/src/querier_handler.rs
```diff
@@ -150,15 +149,26 @@ impl IngesterQueryResponse {
...
-                                        match prepare_batch_for_flight(&batch, Arc::clone(&schema))
-                                        {
-                                            Ok(batch) => futures::stream::iter(
-                                                split_batch_for_grpc_response(batch),
-                                            )
-                                            .map(|batch| {
-                                                Ok(FlatIngesterQueryResponse::RecordBatch { batch })
-                                            })
-                                            .boxed(),
+                                        let batch =
```
alamb, line 152: this change is to map the error types.

The whole structure of the `ingester` / `ingester2` handlers could be improved significantly. I plan to do so after I get the basic flight code upstreamed into arrow

ingester2/src/server/grpc/query.rs
```diff
@@ -336,12 +338,17 @@ impl From<QueryResponse> for FlatIngesterQueryResponseStream {
                                     })
                                 });
 
-                                let tail = match prepare_batch_for_flight(
-                                    &snapshot,
-                                    Arc::clone(&schema),
-                                ) {
+                                let batch =
```
alamb, line 341: As in `ingester` these changes are to massage the error types (as `ingester2` has a mostly copy/paste version of the handler from `ingester`).  I plan to clean the this whole implementation up after `ingester` is gone and  I have upstreamed this code to arrow-rs.

cc @domodwyer  and @carols10cents

iox_arrow_flight/src/encode.rs
```diff
@@ -0,0 +1,375 @@
...
+        // to have that schema too
+        let schema = Arc::new(prepare_schema_for_flight(&schema));
+        let mut schema_flight_data: FlightData = SchemaAsIpc::new(&schema, &options).into();
+        schema_flight_data.app_metadata = app_metadata;
+
+        let schema_stream = futures::stream::once(async move { Ok(schema_flight_data) });
+
+        let options = Arc::new(options);
+
+        let data_stream = input
```
alamb, line 103: I will admit getting to this async stream code took a lot of finagling, but i am quite pleased with how it worked out

iox_arrow_flight/src/encode.rs
```diff
@@ -0,0 +1,375 @@
...
+/// which defines the mapping from dictionary IDs. It then sends these
+/// dictionaries over the wire.
+///
+/// This requires identifying the different dictionaries in use, assigning
+/// them IDs, and sending new dictionaries, delta or otherwise, when needed
+///
+/// This is tracked by <https://github.com/influxdata/influxdb_iox/issues/1318>
+///
+/// See also:
+/// * <https://github.com/influxdata/influxdb_iox/issues/4275>
```
alamb, line 236:
```suggestion
```

service_grpc_flight/src/lib.rs
```diff
@@ -463,145 +458,84 @@ where
...
 
-        // Add response metadata
-        let mut bytes = BytesMut::new();
-        let app_metadata = proto::AppMetadata {};
-        prost::Message::encode(&app_metadata, &mut bytes).context(SerializationSnafu)?;
-        schema_flight_data.app_metadata = bytes.to_vec();
+        let schema = physical_plan.schema();
 
-        let mut stream_record_batches = ctx
+        let query_results = ctx
```
alamb, line 486: Here is an example use of the new decoder in the flight API implementation. cc @crepererum 

I am quite pleased with how this looks now - and it is a preview of what I hope this code will look like eventually in the ingester.

alamb:
I don't plan to merge this until the holiday break is over, but I wanted to get it up in case anyone has time to review

iox_arrow_flight/src/encode.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+    };
+
+    let mut output = vec![];
+    for batch in split_batch_for_grpc_response(batch, max_batch_size) {
+        let (flight_dictionaries, flight_batch) = flight_data_from_arrow_batch(&batch, options);
+        output.extend(flight_dictionaries);
+        output.push(flight_batch)
+    }
+    let output = output.into_iter().map(Ok);
```
carols10cents, line 148: Feel free to punt this until profiling to know if this makes a difference, but I think some allocations can be saved with:

```suggestion
    let output: Vec<_> = split_batch_for_grpc_response(batch, max_batch_size)
        .into_iter()
        .map(|batch| {
            let (flight_dictionaries, flight_batch) = flight_data_from_arrow_batch(&batch, options);
            flight_dictionaries
                .into_iter()
                .chain(std::iter::once(flight_batch))
        })
        .flatten()
        .map(Ok)
        .collect();
```

service_grpc_flight/src/lib.rs
carols10cents: The new code is definitely much nicer than the code that was here 🎉

carols10cents (changes requested):
I'm confused about whether `max_batch_size` et al are in bytes or number of rows, and I'd like to see that resolved before merging (I suspect you'd like that too :))

Some doc nits and possible small performance suggestions that you can feel free to take or leave!

crepererum (commented):
Just had a quick look since @carols10cents already did an awesome job her. I really like the upstream movement and the clean up of the API. It kinda grew over the last months a bit too much and really needed some grooming.

iox_arrow_flight/src/encode.rs
```diff
@@ -0,0 +1,375 @@
...
+/// This can be used to implement [`FlightService::do_get`] in an
+/// Arrow Flight implementation.
+///
+/// # Caveats:
+///   1. Any `DictionaryArray`s are converted to their underlying types prior to transport
+///   See <https://github.com/apache/arrow-rs/issues/3389> for more details
+///
+/// [`FlightService::do_get`]: crate::flight_service_server::FlightService::do_get
+#[derive(Debug)]
+pub struct StreamEncoderBuilder {
```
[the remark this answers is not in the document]
alamb: FWIW I thought a bit more about this and I think with some judicious use of `flatten` this could actually be used for ingester / ingester2. But we'll get there eventually

iox_arrow_flight/src/encode.rs
```diff
@@ -0,0 +1,375 @@
...
+    pub options: IpcWriteOptions,
+    /// Metadata to add to the schema message
+    pub app_metadata: Vec<u8>,
+}
+
+/// Default target size for record batches to send.
+///
+/// Note this value would normally be 4MB, but the size calculation is
+/// somehwhat inexact, so we set it to 2MB.
+pub const GRPC_TARGET_MAX_BATCH_SIZE: usize = 2097152;
```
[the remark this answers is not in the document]
alamb: Indeed -- actually I just moved the code around (it used to be in arrow_util/src/optimize.rs) but I should have caught it during initial review as well 😆  Will fix here

crepererum (approved):

ingester2/src/server/grpc/query.rs
alamb: https://github.com/influxdata/influxdb_iox/pull/6623 is the promised cleanup

PR #6463, opened 2022-12-22
fix: drop WAL segments after replay

### LINKED ISSUE #6461: Ingester2 should delete WAL files after replay
When the ingester starts up, it replays any WAL files it finds in the configured directory.

Currently one replayed, the ingester resumes normal ingest and the WAL files remain.

Instead, once replayed, the data should be persisted and then the WAL files dropped before ingest resumes.

As an optimisation in the future we may allow concurrent ingest & persist operations.

domodwyer:
Fixes #6461, but may slow down startup as we now wait for persistence to complete for each file before being "ready". As a nice side-effect, crash loops that make some progress (a file or two each time) will eventually recover automatically.

This can be optimised to allow something like:

* Replay file 1
* Enqueue persist jobs but do not wait for completion
* Replay file
* Enqueue persist job
* (persists for job 1 complete) -> drop file 1

But this approach must respect the saturation of the persist system (otherwise we're just moving the OOM problem to persisting data, rather than buffered data). Given there should rarely be many files, this doesn't seem like a big deal.

Now we have a bit of time, we need to increase the test coverage before making WAL replay more complicated - I'd love to get #6366 merged and then increase the coverage ASAP :+1:

---

* fix: drop WAL segments after replay (66f162823)

      Changes the WAL replay logic to:
      
          * Replay a segment file
          * Persist all replayed data
          * Drop segment file
          * ...repeat...
      
      This ensures old WAL segments are removed once their contents have been
      made durable, fixing #6461.

### DIFF
--- a/ingester2/src/init/wal_replay.rs
+++ b/ingester2/src/init/wal_replay.rs
@@ -1,14 +1,16 @@
-use std::time::Instant;
 use data_types::{NamespaceId, PartitionKey, Sequence, SequenceNumber, TableId};
 use dml::{DmlMeta, DmlOperation, DmlWrite};
 use generated_types::influxdata::iox::wal::v1::sequenced_wal_op::Op;
 use mutable_batch_pb::decode::decode_database_batch;
 use observability_deps::tracing::*;
+use std::time::Instant;
 use thiserror::Error;
 use wal::{SequencedWalOp, Wal};
 
 use crate::{
     dml_sink::{DmlError, DmlSink},
+    persist::{drain_buffer::persist_partitions, queue::PersistQueue},
+    wal::rotate_task::PartitionIter,
     TRANSITION_SHARD_INDEX,
 };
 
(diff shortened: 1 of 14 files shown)


ingester2/src/init/wal_replay.rs
```diff
@@ -111,6 +118,30 @@ where
...
+            .await
+            .expect("failed to drop wal segment");
+
+        info!(
+            file_number,
+            n_files,
+            file_id = %file.id(),
+            size = file.size(),
+            "dropped persisted wal segment"
+        );
```
domodwyer, line 144: This is the entire logical fix - we replay a file, we persist the buffered data, and we delete the file.

domodwyer (commented):
Almost all of this diff is caused by testing and visibility changes to ensure code can be run by benchmarks without exporting everything under the sun by default.

ingester2/src/init/wal_replay.rs
pauldix: So will this cause a flood of persists if the ingesters keep getting bounced and get just some writes?

pauldix (approved):
It looks like this will just persist whatever was in the previous segments on startup, clearing them out. How does this fix the OOM problems? Seems like we'd still have a problem where the ingester gets to OOM and then persists on the restart, but then buffers up again until it again gets to OOM.

If it's OOMing, I think we should lower the persist cost threshold from 20M to something more conservative like 15M and see what happens. Unless I'm not quite understanding the issue?

domodwyer:
It's OOMing because it never cleaned up WAL files that had been replayed, so we had been accumulating more and more data to buffer at startup - see the ticket this closes (https://github.com/influxdata/influxdb_iox/issues/6461).

I'm not attempting to limit memory usage or guard against overly large WAL segments in this PR.

PR #6467, opened 2022-12-23
chore: Upgrade datafusion

alamb:
# Rationale
1. Keep up with upstream config changes from @tustvold

# Changes:
1. Update to https://github.com/apache/arrow-datafusion/commit/07f49803a3d7a9e9b3c2c9a7714c1bb08db71385
2. Update for upstream API


iox_query/src/exec/context.rs
```diff
@@ -310,26 +307,13 @@ impl IOxSessionContext {
     pub async fn prepare_sql(&self, sql: &str) -> Result<Arc<dyn ExecutionPlan>> {
         let ctx = self.child_ctx("prepare_sql");
         debug!(text=%sql, "planning SQL query");
+
+        // NOTE can not use ctx.inner.sql here as it also interprets DDL
+        // TODO: file ticket in DataFusion and fix
```
alamb, line 312:
```suggestion
        // https://github.com/apache/arrow-datafusion/issues/4720
```

alamb: The API needs cleaning up -- I proposed doing so in https://github.com/apache/arrow-datafusion/pull/4721

alamb: Update here is that @tustvold  has an alternate plan. Will commence after return from break

alamb:
First of several PRs I have planned today to update IOx to the latest and greatest datafusion

crepererum (approved):

PR #6469, opened 2022-12-27
refactor: use SessionState directly

tustvold:
This shows how to use non-deprecated APIs to plan SQL queries without introducing potential state inconsistency


iox_query/src/exec/context.rs
```diff
@@ -226,11 +230,12 @@ impl IOxSessionConfig {
 
         let inner = SessionContext::with_state(state);
 
+        // TODO: Clean up information schema handling on SessionContext
```
tustvold, line 233: This was already on my list, the way that we shim the catalog providers is just a little strange

tustvold: PR - https://github.com/apache/arrow-datafusion/pull/4740

iox_query/src/exec/context.rs
```diff
@@ -308,13 +315,17 @@ impl IOxSessionContext {
         let ctx = self.child_ctx("prepare_sql");
         debug!(text=%sql, "planning SQL query");
 
-        // NOTE can not use ctx.inner.sql here as it also interprets DDL
-        // TODO: file ticket in DataFusion and fix
-        #[allow(deprecated)]
-        let logical_plan = ctx.inner.create_logical_plan(sql)?;
+        let mut statements = DFParser::parse_sql(sql)?;
```
alamb, line 318: This API is reasonable to me. I think the `DFParser` is not super well documented at the moment, but I will go rectify that now 🏃

iox_query/src/exec/context.rs
```diff
@@ -308,13 +315,17 @@ impl IOxSessionContext {
...
+        if statements.len() != 1 {
+            return Err(DataFusionError::NotImplemented(
+                "The context currently only supports a single SQL statement".to_string(),
```
alamb, line 321:
```suggestion
                "IOx currently only supports a single SQL statement".to_string(),
```

alamb (approved):
makes sense to me 👍

tustvold:
I intend to keep this open for a bit, as the API is still going through flux (these methods are going to move onto DataFrame), but I wanted to get something up to show the vague idea

alamb:
I think the idea is good 👍

PR #6476, opened 2022-12-30
fix: make estimate memory of a query needs higher due to recent observation

NGA-TRAN:
Connects to https://github.com/influxdata/idpe/issues/16822


compactor/src/parquet_file_lookup.rs
```diff
@@ -92,7 +92,8 @@ impl ParquetFilesForCompaction {
             let estimated_arrow_bytes = partition
                 .estimated_arrow_bytes(min_num_rows_allocated_per_record_batch_to_datafusion_plan);
             // Estimated bytes to store this file in memory
-            let estimated_bytes_to_store_in_memory = 2 * parquet_file.file_size_bytes as u64;
+            // Since we ignore the size allocated in IOx and DF, let us make this large enough to cover those
+            let estimated_bytes_to_store_in_memory = 5 * parquet_file.file_size_bytes as u64;
```
NGA-TRAN, line 96: This is the change. The rest is test modification to work with this change

pauldix: Ultimately, this should take into account the width of the table. Without that it's going to vary wildly, but we can move forward with this for now.

pauldix (approved):

compactor/src/parquet_file_lookup.rs
alamb: I think any estimate of the memory requirements based on the size of parquet will always be somewhat inaccurate as parquet is so highly compressed and the compression ratio varies so much

The more dynamic runtime memory limits / checking I think will be much more effective -- can't wait to get that working

PR #6480, opened 2023-01-02
test: Lay the groundwork for porting query_tests to the RPC write path

carols10cents:
Connects to #6317.

This PR does not actually port any of the `query_tests`. That will come in a future PR.

It does, however, port some more of the end-to-end tests and also adds some infrastructure to the test helpers that I plan to use in the port of the query tests, namely:

- Creating a MiniCluster that contains a querier using the RPC write path (router2 and ingester2 existed already)
- Creating an ingester2 configured to effectively never persist data (`INFLUXDB_IOX_WAL_ROTATION_PERIOD_SECONDS=86400`) so that the tests can exercise the state of data being in the ingester's buffers but not in Parquet files yet. The default configuration for ingester2 in tests is to persist as fast as possible (`INFLUXDB_IOX_WAL_ROTATION_PERIOD_SECONDS=1`)
- Creating a function that verifies a change in the number of Parquet files in the catalog for a particular namespace and table, which allows tests to wait until data has been persisted.

### DIFF
--- a/influxdb_iox/tests/end_to_end_cases/ingester.rs
+++ b/influxdb_iox/tests/end_to_end_cases/ingester.rs
@@ -102,6 +102,7 @@ async fn ingester2_flight_api() {
     let table_name = "mytable";
 
     // Set up cluster
+    // Don't use a shared cluster because the ingester is going to be restarted
     let mut cluster = MiniCluster::create_non_shared2(database_url).await;
 
     // Write some data into the v2 HTTP API ==============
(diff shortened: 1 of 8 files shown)


crepererum (approved):
Looks good.

Just to clean up any confusion: this only concerns the end2end tests. The so called `query_tests` are in https://github.com/influxdata/influxdb_iox/tree/main/query_tests and are NOT truly end2end (to safe some test time and also to enable some more sophisticated state manipulation).

alamb:
> Yes, next I am going to be rewriting query_tests to instead be end-to-end because there are only 2 states chunks can be in: ingester and parquet. Using the capabilities added in this PR, they can use the MiniCluster gRPC API and avoid duplicating internal logic.

@carols10cents  -- i think it is quite important to be able to set up the scenarios like "two writes went in, one got persisted but the other is still in the ingester". 

As long as this is possible in the end to end tests I think it is ok to port query_tests to run against an actual cluster process. I know this is the approach that @domodwyer  prefers. 

Last time we tried this it got quite complicated in the end to end tests as the ingester would persist thing based on its own trigger policy (like time since write, or size buffered) rather than what the test scenario and thus it was quite hard to cover all the scenarios (delete again a persisted chunk, delete against an unpersisted chunk, etc).

If you want to have a brief discussion about it, I feel quite strongly about the importance of query_tests and the coverage they provide and would be happy to speak.

Issue #6112, opened 2022-11-10
InfluxQL: Teach IOx how to process a SELECT statement


stuartcarnie:
## Summary

Teach IOx how to process InfluxQL `SELECT` queries via the Flight RPC API.

## Overview

With EPIC #5353 complete, the next step in the _InfluxQL for IOx_ journey is to teach IOx how to process an InfluxQL query request. Upon completion of this epic, IOx will be able to

* receive an InfluxQL query request from the IOx `influxdb_iox query` command;
* process the request; and
* return the results to the client

In order to accomplish this, IOx will require changes in 3 high-level areas:

* Enhancing the `iox_query` crate to process InfluxQL requests, which includes changes to
  * Flight RPC interfaces to handle SQL and InfluxQL queries;
  * parsing and transforming InfluxQL queries, so they may be processed by DataFusion; and
  * streaming the response to the client
* The CLI `influxdb_iox query` will require an argument to override the query language, for example:

  ```sh
  influxdb_iox query -l influxql company_sensors "SELECT MEAN(usage_idle) FROM cpu WHERE time > now() - 5m GROUP BY TIME(10s)"
  ```

> **NOTE:**  
> See the section later **InfluxQL Query Engine** for developer notes regarding the high-level operation of the InfluxQL
> query engine in InfluxDB v2.4 OSS edition.

## Scope

As described in the summary, the scope of this epic is to teach IOx how to process InfluxQL queries. It does not describe changes to Cloud 2, however, this architecture is understood and linked to this issue.

This document intends to describe the behaviour of the InfluxQL query engine in sufficient detail in order
to translate to DataFusion `LogicalPlan`s and expressions.

> **NOTE:**  
> For the remainder of this document, any references to an InfluxQL query refer to an InfluxQL `SELECT` statement.

## `SELECT` behaviour

This section will break down the behaviour of a `SELECT` statement into distinct sections. These sections may be used to
drive implementation and testing for the implementation of InfluxQL in IOx. The order is not significant.

### `DISTINCT`

- [ ] `SELECT DISTINCT` support

* The `DISTINCT` keyword can be specified as a unary operator

```sql
select distinct i64 from m0
```
```text
name: m0
time                 distinct
----                 --------
1970-01-01T00:00:00Z 101
1970-01-01T00:00:00Z 211
```

* The timestamp of the values is the lower bounds of the time range

```sql
select distinct i64 from m0 where time >= '2004-04-09T12:13:14Z'
```
```text
[...]

PR #6481, opened 2023-01-03
feat: Teach InfluxQL rewriter how to name columns

stuartcarnie:
Helps #6112 

The primary purpose of this PR is to teach the InfluxQL rewriter how to name InfluxQL columns in the projection list, as to be largely compatible with the [`ColumnNames`](https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/ast.go#L1651) API in the original Go implementation. This is important for clients, as they will be expected column names to match. A number of additional tests were added to validate this behaviour.

The following minor fixes and improvements are included in the PR:

* Added a `Timestamp` case to the `VarRefDataType` enumeration
* Fix the `field_name` function, to walk the nested expression for the `UnaryOperator` case. Test cases were added to validate this new behaviour.
* Teach the InfluxQL parser that keywords may be immediately followed by a `!` token, such as `WHERE tag!='foo'`. Appropriate test cases were added.
* Improve the `test_utils` module, so that a future client may access the `TestChunk` vector to create a `TestDatabase`, which will be useful for testing the InfluxQL planner


iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -488,6 +488,49 @@ fn rewrite_field_list(stmt: &mut SelectStatement, schema: Arc<dyn SchemaProvider
     Ok(())
 }
 
+/// Resolve the outer-most `SELECT` projection list column names in accordance with the
+/// [original implementation]. The names are assigned to the `alias` field of the [`Field`] struct.
+///
+/// [original implementation]: https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/ast.go#L1651
+fn rewrite_field_list_aliases(field_list: &mut FieldList) -> Result<()> {
+    let mut column_aliases = HashMap::new();
```
crepererum, line 496: I see you clone the column names below, but with some type annotations I think you could get away w/:

```suggestion
    let mut column_aliases = HashMap::<&str, _>::new();
```

iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -488,6 +488,49 @@ fn rewrite_field_list(stmt: &mut SelectStatement, schema: Arc<dyn SchemaProvider
...
+    field_list.deref_mut().iter_mut().for_each(|field| {
+        field.alias = Some(
+            {
+                let name = match &field.alias {
+                    Some(v) => v.deref().clone(),
+                    None => field_name(field),
+                };
+                // Generate a new name if there is an existing alias
+                let name = if let Some(count) = column_aliases.get(&name) {
```
crepererum, line 505: For this if-else as well as the below: you double-hash the names and the whole "I need to reference the same map w/ the same key in the if-statement as well as in the two branches" is kinda bad code smell and can lead to subtle bugs. Rust provides you a so called "entry" API, which you may (optional) use here:

```rust
// assuming that `column_aliases` is `HashMap<&str, _>`!
let name = match column_aliases.entry(&name) {
    Entry::Occupied(o) => {
        ...
        // somewhere in here:
        o.insert(count + 1);
        ...
    }
    Entry::Vacant(_) => {
        name
    }
};
```

iox_query/src/plan/influxql/rewriter.rs
```diff
@@ -488,6 +488,49 @@ fn rewrite_field_list(stmt: &mut SelectStatement, schema: Arc<dyn SchemaProvider
     Ok(())
 }
 
+/// Resolve the outer-most `SELECT` projection list column names in accordance with the
+/// [original implementation]. The names are assigned to the `alias` field of the [`Field`] struct.
+///
+/// [original implementation]: https://github.com/influxdata/influxql/blob/1ba470371ec093d57a726b143fe6ccbacf1b452b/ast.go#L1651
```
crepererum, line 494: In general I find it a bit worrisome that we reference code instead of a spec here, but I guess this is not the point to fix that.

crepererum (approved):
Some smaller nits only.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: Agreed. 

Unfortunately, there is no spec for InfluxQL, so the implementation becomes the unofficial spec 😬  It also makes it much slower to implement, as it does require careful studying of the code to understand how various aspects of InfluxQL work.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: Great info, I'll revisit and update!

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: The problem I ran into with `&str` being slices, is they outlive their references created in the closure for the HashMap, which lives in the outer scope.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: Result of me trying to use `entry` is that this borrows a mutable reference to `column_aliases`, and so when the implementation tries to determine if the newly resolved name exists in the map, the borrow-checker complains, as we're also trying to obtain another immutable reference:

```
507 |                 match column_aliases.entry(name.clone()) {
    |                       ---------------------------------- mutable borrow occurs here
...
512 |                             if column_aliases.contains_key(&resolved_name) {
    |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ immutable borrow occurs here
513 |                                 o.insert(count + 1);
    |                                 ------------------- mutable borrow later used here
```

Still, I have simplified the code so there is only a single `match` vs two cases and also identified an edge case that has been resolved.

iox_query/src/plan/influxql/rewriter.rs
stuartcarnie: I resolved this during simplification, as I capture all the base fields names in a `Vec` first (to fix an edge case), which allows me to borrow slices from this to use in the `HashMap`.

PR #6484, opened 2023-01-03
fix: Fix max RecordBatch size limit calculation to avoid gRPC message too large errors

alamb:
~Draft as it builds on https://github.com/influxdata/influxdb_iox/pull/6460~

# Rationale
Closes  https://github.com/influxdata/idpe/issues/16834

There is code that tries to limit the size when serializing record batches over the network. However, the way it is implemented has a bug in it (the first batch is limited, but the next batch has 2x the limit and then next has 4x the limit, etc).

Embarrassingly, it turns out I actually fixed this issue while upstreaming (and writing more tests for this logic):

https://github.com/apache/arrow-rs/pull/3391/files#diff-e11595282032996fd43b9f4749bf13e67094fe5d76c9d7795a5233c763e2aadeR299-R324

# Changes
1. Correct splitting logic by copying upstreamed code
2. Add more tests

# Upstream
PR to upstream this test in https://github.com/apache/arrow-rs/pull/3444


iox_arrow_flight/src/encode.rs
```diff
@@ -366,11 +366,48 @@ mod tests {
...
             split.iter().map(|batch| batch.num_rows()).sum::<usize>(),
             n_rows
         );
         assert_eq!(concat_batches(&batch.schema(), &split).unwrap(), batch);
     }
+
+    #[test]
+    fn test_split_batch_for_grpc_response_sizes() {
+        // 2000 8 byte entries into 2k pieces: 8 chunks of 250 rows
+        verify_split(2000, 2 * 1024, vec![250, 250, 250, 250, 250, 250, 250, 250]);
```
alamb, line 380: On main without this code fix, the test fails like:

```
thread 'encode::tests::test_split_batch_for_grpc_response_sizes' panicked at 'assertion failed: `(left == right)`
  left: `[250, 500, 1000, 250]`,
 right: `[250, 250, 250, 250, 250, 250, 250, 250]`: mismatch for RecordBatch { schema: Schema { fields: [Field { name: "a", data_type: UInt64, nullable: false, dict_id: 0, d
```

iox_arrow_flight/src/encode.rs
```diff
@@ -189,12 +189,12 @@ pub fn split_batch_for_grpc_response(
 
     let n_batches =
         (size / max_batch_size_bytes + usize::from(size % max_batch_size_bytes != 0)).max(1);
-    let rows_per_batch = batch.num_rows() / n_batches;
+    let rows_per_batch = (batch.num_rows() / n_batches).max(1);
```
alamb, line 192: In my testing upstream I found that for very small batches (of zero rows) this code would loop forever without .max(1)

alamb (commented):
I plan to delete all this code from IOx eventually in favor of the version I have upstreamed in https://github.com/apache/arrow-rs/pull/3391 (which also has this fix)

crepererum (approved):

PR #6485, opened 2023-01-03
fix: add index and modify a catalog query  to have 2 catalog queries  run faster for the case of a lot of files and no shard

NGA-TRAN:
Connects to https://github.com/influxdata/idpe/issues/16836

# Rationale
The 2 common used hot and cold catalog queries timeout a lot because:
1. Query to collect hot partitions: we modified them not to filter `where shard_id = $n` anymore and, thus, index is not used.
2. Query to collect cold partition: index seems not working well for this and a lot more files make it worse 

# Changes in this PR:

1. Created a new index that I think will help query number 1
2. Simplified query number 2 to no longer order them by their number of files to eliminate expensive operators: order by and group by.
3. Since the meaning of query number 2 has changed a bit, tests need to get updated accordingly. 


iox_catalog/migrations/20230103172433_create_index_no_shard_id.sql
```diff
@@ -0,0 +1,2 @@
+-- Add index for query to select recent_highest_throughput_partitions for the case without shrad_id provided
+CREATE INDEX IF NOT EXISTS parquet_file_compaction_delete_created_idx ON parquet_file (compaction_level, to_delete, created_at);
```
NGA-TRAN, line 2: 1st change: this index is a new version of `(shard_id, compaction_level, to_delete, created_at)`

iox_catalog/src/postgres.rs
```diff
@@ -1900,6 +1900,7 @@ LIMIT $4;
                 .map_err(|e| Error::SqlxError { source: e })
             }
             None => {
+                // todo: create index on (compaction_level, to_delete, created_at)
```
NGA-TRAN, line 1903: Add this todo for me to verify if it works after the index is fully created. Index works differently on different dataset hence my local small data set won't be able to verify this

iox_catalog/src/postgres.rs
```diff
@@ -1969,22 +1970,38 @@ LIMIT $3;
...
+                // To avoid expensive group-by and order-by, this query will go thru 2 steps:
+                //    1. Get num_partitions partittions that needed to compact:
+                //       aka partitions with non-deleted L0 and/or L1 files and not in skipped_compactions
+                //    2. Get cold partitions among num_partitions returned in #1:
+                //       aka partitions without L0 files (they were all hard deleted) OR
+                //       all L0 (both deleted and not deleted) created before time_in_the_past
+                //
+                // Note that in this case, we do not prioritize partitions with more L0+L1 files,
+                // we just compact whatever cold partitions we hit first. With the kafkaless ingesting design, there
+                // will be a lot of files in each partittions and prioritize them is no longer necessary
```
NGA-TRAN, line 1982: This is a bit of change in behavior but it works better with our a-lot-of-files situation

iox_catalog/src/interface.rs
```diff
@@ -3300,20 +3300,26 @@ pub(crate) mod test_helpers {
             .await
             .unwrap();
         assert_eq!(partitions.len(), 2);
-        // They must be in order partition_2 (more files), partition
-        assert_eq!(partitions[0].partition_id, partition_2.id); // 2 files
-        assert_eq!(partitions[1].partition_id, partition_1.id); // 1 file
-                                                                // Across all shards
-                                                                // Must return 2 partitions
```
NGA-TRAN, line 3307: Because the partitions are no longer sorted on their number of files, the tests here are changed/removed accordingly

NGA-TRAN:
@carols10cents  and @pauldix I think you know this the most. can you help review it?

iox_catalog/src/postgres.rs
pauldix: Can you log this as an issue, rather than a todo?

iox_catalog/src/postgres.rs
```diff
@@ -1969,22 +1970,38 @@ LIMIT $3;
...
+
+                // To ensure the number of cold partittions is near given num_partitions,
+                // we will multiply num_partitions by cold_factor
+                let cold_factor = 2;
+                let num_partitions = num_partitions * cold_factor;
```
pauldix, line 1987: I don't understand why this multiplier is here. We're deliberately returning more cold partitions than are being requested in the function signature?

iox_catalog/src/postgres.rs
```diff
@@ -1969,22 +1970,38 @@ LIMIT $3;
...
+                        partition_id IN
+                        ( SELECT partition_id 
+                            FROM parquet_file 
+                            WHERE (compaction_level = $3 OR compaction_level= $4) 
+                                  AND to_delete IS NULL 
+                                  AND partition_id NOT IN (SELECT partition_id FROM skipped_compactions)
+                            LIMIT $2
+                         )
+                    GROUP BY partition_id, shard_id, namespace_id, table_id
+                    HAVING MAX(case when compaction_level = $3 then created_at end) IS NULL OR
```
pauldix, line 2003: why would the `created_at` be null? Seems like the files should all have this property set. Or is this a holdout from old catalog that didn't have this?

iox_catalog/src/postgres.rs
```diff
@@ -1969,22 +1970,38 @@ LIMIT $3;
...
+                    MAX(case when compaction_level = 0 then created_at end)  < $1;
```
pauldix, line 2004: Why does this have the compaction level hard coded at 0 when it seems like $3 is bound to this value. Should this be using $3 here?

pauldix (commented):
Have some questions in there, but pending that I think it's good.

iox_catalog/src/postgres.rs
NGA-TRAN: It just mean this partition has no more L0 files for us to verify if they are cold or not. Those L0 files were soft and hard deleted.

iox_catalog/src/postgres.rs
NGA-TRAN: Good catch.

iox_catalog/src/postgres.rs
NGA-TRAN: This is the reason I do not like this query at all (see below). We first choose number of partitions with non-deleted L0 and L1. Then among them, choose cold one. Hence if the former set do not include a lot of cold partitions, we will return nothing

iox_catalog/src/postgres.rs
NGA-TRAN: Can do. The index is created in this PR. I just need to verify this

NGA-TRAN:
We no longer need this PR

PR #6486, opened 2023-01-04
New ingest structure

pauldix:
This is a big draft to prototype something I wanted to try. In order to make the read replicas work, we need sequence numbers in the buffer. I created the `ingest_structure` crate to represent what the new WAL and RPC formats should look like (in the first commit).

The second commit is the prototype to pull MutableBatch and the DML wrapper out of the router, ingester, and WAL. This was a considerably large refactor given how intertwined everything was. But I wanted to get this working end-to-end so that I could benchmark the entire system as opposed to just some microbenchmark on a serialization format.

I wanted to prove out the memory size for the buffer using the new format, the performance of the whole ingestion pipeline, and the performance of the WAL replay. The last bit is important for the replicas as well since the WAL format is what they'll be consuming directly.

## Some Test Results
I ran this test on my M1 Macbook Pro with a single router, single ingester, local Postgres and file system object store. I ran this command:

```
./target/release/iox_data_generator --host 127.0.0.1:8080 --specification ./iox_data_generator/schemas/eu_central.toml --org paul --bucket db --token foo --start '60 seconds ago'
```

## New
* 7.7 GB RAM in Ingester
	* I later added an intermediate conversion of the proto into RecordBatch during ingestion. I set it to do that every 1k records in a partition, which reduced the memory size to 2.68 GB without negatively impacting performance.
* 2.2 GB WAL file
* 500% ingester CPU, 250% router CPU
* 100% CPU during replay (tied to single thread)
* File replay duration (just into RAM) 42.5s
* Replay duration with persist: 63.7s

```
./target/release/iox_data_generator --host 127.0.0.1:8080 --specification ./iox_data_generator/schemas/eu_central.toml --org paul --bucket db --token foo --start '60 seconds ago'

made 3415 requests at 53/sec with 55211140 rows at 862674/sec in 64.079260458s with 0 errors for a 0.00 error rate
Submitted 55211140 total points

# run 2
made 3415 requests at 54/sec with 55211140 rows at 876367/sec in 63.69987675s with 0 errors for a 0.00 error rate
Submitted 55211140 total points
```

## Old
* 1.67 GB RAM in Ingester
* 1.1 GB WAL file
* 500% router CPU 250% ingester CPU
* 100% CPU during replay (tied to single thread)
* File replay duration into RAM 265.2s
* Replay duration with persist: 274.3s

```
made 3419 requests at 16/sec with 55291140 rows at 273718/sec in 202.484022833s with 0 errors for a 0.00 error rate
Submitted 55291140 total points
```

## Summary & Thoughts
Ingest on the new structure was 3.2x faster than older structure. Replay took 6.3x less time, which would also be reflected in our Replicas as that's the same structure they would be replaying. More testing would be useful with different numbers of writers, smaller request sizes, and a more varied data set (all could be done with the generator).

As expected, the RAM and WAL footprint were larger with the new structure since it doesn't dictionary encode repeated values. In a real world scenario I don't think the WAL difference would be as large since this test had some very large requests in it with many repeated strings.

For the RAM, I mainly wanted to confirm that the amount of memory needed wouldn't explode to some amount that would invalidate this approach. I think for our uses, this memory profile is totally acceptable. I think it can be tuned a bit with intermediate conversion to RecordBatch during ingestion while sharing common schema (to avoid re-creating those).

The WAL replay on both approaches is limited by a single thread reading from the segment file. I think we can probably change this up a bit to create more smaller segment files and to replay them in parallel (particularly given we don't guarantee write sequencing). But I don't think we need to worry about that for a while.


pauldix:
Closing this out but keeping the branch around if we want to come back to any of it at some future point. It's just way too much code churn at this point with an upcoming release. We'll get to replicas another way and then take time to refactor the ingest structure when we don't have both the old and new running at the same time.

PR #6493, opened 2023-01-04
refactor: improve influxRPC warning logging

crepererum:
The current version is barely readable because the logged schema w/ all it's metadata is soooo long.

Also cleaned up the code a bit.


alamb (approved):
LGTM

PR #6494, opened 2023-01-04
feat(replication): define replication RPC API

domodwyer:
A minimal replication RPC proto definition. This WILL have to grow/change in the future to support replica lag/back-pressure/etc.

Makes use of a `byte` array to encode a CRoaring bitmap (docs for [serialise](https://docs.rs/croaring/latest/croaring/bitmap/struct.Bitmap.html#method.serialize), [deserialise](https://docs.rs/croaring/latest/croaring/bitmap/struct.Bitmap.html#method.deserialize)) using the format [defined here](https://github.com/RoaringBitmap/RoaringFormatSpec/) - not very abstract, but quick to implement.

---

* feat(replication): define replication RPC API (91680854c)

      Defines the rough outline of an replication RPC API. More details/docs
      to follow.


pauldix (approved):
Looks good. One question on the PartitionBufferService request. The assumption I have is that as a replica, once I send a request for the PartitionBuffers, I will also start getting ReplicationRequests from that ingester. Is that the case? Will the ingester just be able to use the info from the incoming GetPartitionBuffers request to be able to send out ReplicationRequests to the same endpoint?

domodwyer:
> Looks good. One question on the PartitionBufferService request. The assumption I have is that as a replica, once I send a request for the PartitionBuffers, I will also start getting ReplicationRequests from that ingester. Is that the case?

Truthfully I haven't thought about how they connect and do this dance at all - I'm interested to chat with @mkmik about what is easiest deployment wise (especially depending on how DNS can affect this discovery during deployments). 

This defines the "frames" sent over the wire in some way - I imagine the protocol will look something like what you describe (perhaps the replica sends it's peering address in the `GetParittionBuffers` request, and we start pushing to them immediately).

PR #6507, opened 2023-01-05
refactor: decouple influxql from SchemaProvider

tustvold:
Upstream changes have now made SchemaProvider async, as a result using it for this sort of planning would force planning itself to become async. Aside from the complexity, the extensive use of recursion makes this not a great idea.

Fortunately we can apply the same approach we do in DataFusion to handle this, namely extract a sync trait and plan based on this. If/when we make IOx's catalog async this will require some additional logic to pre-fetch the necessary tables, as is done in DF, but this shouldn't be hugely complex.

As an added bonus this create better encapsulation of the rewriter, as it no longer depends on all the physical planning plumbing inherent to SchemaProvider.


iox_query/src/test.rs
```diff
@@ -983,6 +983,10 @@ impl TestChunk {
     pub fn table_name(&self) -> &str {
         &self.table_name
     }
+
+    pub fn schema_ref(&self) -> &Schema {
```
tustvold, line 987: The way we are double-boxing arrow::Schema is a little strange imo, I intend to tweak this in a separate PR

stuartcarnie (approved):
Looks good.

How will I derive or obtain a `RewriterContext` from an `IOxSessionContext`, which is ultimately where this will come from when executing a query.

Issue #6508, opened 2023-01-05
Router fails to load balance across ingesters


domodwyer:
After merging https://github.com/influxdata/influxdb_iox/pull/6504 to switch to using tower/tonic's gRPC load balancing, I observed that only one ingester was handling requests. Router logs showed many "upstream ingester" errors.

Once I rebooted the routers, the load was correctly distributed across all upstream ingesters again, suggesting it was due to the router's gRPC balancer state.

![image](https://user-images.githubusercontent.com/9275968/210784616-4ade50ad-6401-4efe-9a7d-11173076902d.png)

I also confirmed all ingesters were reachable from a router prior to rebooting it (thus if it tried to connect, it would have succeeded).

When looking into the code [for the balance logic](https://tower-rs.github.io/tower/src/tower/balance/p2c/service.rs.html#234), I discovered that when a node fails [it is removed from the pool](https://docs.rs/tower-ready-cache/0.3.1/tower_ready_cache/cache/struct.ReadyCache.html):

> When ReadyCache::check_ready* returns false, it indicates that the specified service is not ready. If an error is returned, this indicats that the server failed and has been removed from the cache entirely.

Because the original code was using the [`Channel::balance_list()` constructor](https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html#method.balance_list) which [inserts the nodes into the pool once in the constructor itself](https://docs.rs/tonic/latest/src/tonic/transport/channel/mod.rs.html#113-117) the nodes that subsequently failed were permanently removed from the pool. Note that this is not documented, and I would argue is probably not the desired behaviour when using a static list (as eventually your service is guaranteed to be loading only one node).

As a strange quirk, the last node in the pool remains in the pool even when it fails. When this node recovered, it received all the writes. This has already been noticed and a upstream bug was opened [in the tonic repo](https://github.com/hyperium/tonic/issues/1147).

PR #6509, opened 2023-01-05
fix(router2): drive ingester node (re)-discovery

domodwyer:
I'm not sure the upstream tonic/tower load-balancing implementation works as it should for a static list of endpoints. See https://github.com/hyperium/tonic/issues/1147.

Hopefully fixes https://github.com/influxdata/influxdb_iox/issues/6508.

---

* fix(router2): drive ingester node (re)-discovery (9ab86fa15)

      The tonic / tower load-balance implementation discards failed nodes,
      even when using a static list - this causes nodes that fail once to
      never be retried.
      
      This doesn't happen for the last node for some reason, and leads to all
      the load from one router hitting a single ingester instead of load
      balancing across all ingesters.
      
      This commit adds a hack to constantly tell the load balancer to probe
      all nodes, hopefully causing them to re-discover previously failed
      nodes. I don't have the time to do this properly :(

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4626,6 +4626,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tonic",
+ "tower",
  "trace",
  "workspace-hack",
  "write_buffer",
--- a/router/Cargo.toml
+++ b/router/Cargo.toml
@@ -43,6 +43,7 @@ trace = { path = "../trace/" }
 workspace-hack = { path = "../workspace-hack"}
 write_buffer = { path = "../write_buffer" }
 write_summary = { path = "../write_summary" }
+tower = { version = "0.4.13", features = ["balance"] }
 
 [dev-dependencies]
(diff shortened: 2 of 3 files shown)


cannonpalms:
Use of an external load balancer (istio/envoy) is another option here that may obviate the need for this code. It may also result in more optimal load distribution.

domodwyer:
Yes, it could - unfortunately using the grpc balancer itself was a "quick hack" to fix deployments and unblock other work....

The problem with pushing balancing "down" to 3rd party code (either internal, like the tower grpc balancer, or external, like a routing proxy/LB) is that you loose the knowledge of which endpoints are doing what, which means you can't act on that information (even in the most basic sense, like logging "ingester-X.prod.network is unreachable!") 

https://github.com/influxdata/influxdb_iox/issues/6173 tracks pulling this back up into the application logic.

lukebond (approved):

PR #6511, opened 2023-01-05
refactor: cleanup schema boxing

tustvold:
This PR does the following:

* Removes `Arc<iox::Schema>` as `iox::Schema` is just a wrapper around `Arc<arrow::Schema>` and is therefore cheap to clone
* Changes most methods from needing `Arc<Schema>` to `&Schema` except for where they actually need an owned `Schema`
* Changes methods from returning `Arc<Schema>` to `&Schema` this allows the caller to clone if needed, and ensures lifetimes propagate correctly - see https://github.com/apache/arrow-rs/issues/313 for more context on why this is a code smell
* Changes `Arc<arrow::Schema>` to `SchemaRef` to help distinguish it from the IOx schema


ingester/src/query_adaptor.rs
```diff
@@ -80,13 +80,14 @@ impl QueryAdaptor {
         // partitions - if there is a QueryAdaptor, it contains data.
         assert!(data.iter().map(|b| b.num_rows()).sum::<usize>() > 0);
 
+        let schema = merge_record_batch_schemas(&data);
```
tustvold, line 83: This is a concrete behaviour change, the schema is computed eagerly. I struggle to conceive of a case where the schema would not be needed and so I think this is fine. Perhaps @domodwyer or @NGA-TRAN have some context on why this is this way?

domodwyer: It was evaluated every time the func was called, and there was a `TODO: generate it once` on this line so I did the legwork as instructed 😂

alamb: I agree this looks like a fine change to me

ingester2/src/query_adaptor.rs
```diff
@@ -59,13 +59,14 @@ impl QueryAdaptor {
         // partitions - if there is a QueryAdaptor, it contains data.
         assert!(data.iter().map(|b| b.num_rows()).sum::<usize>() > 0);
 
+        let schema = merge_record_batch_schemas(&data);
```
alamb, line 62: same "behavior change" comment above applies here (as this is a fork of the ingester code)

alamb (commented):
Looks good to me. Thank you for the cleanup @tustvold 

I also made https://github.com/influxdata/influxdb_iox/pull/6523 to try and make it clearer that `Schema` was cheaply copyable

alamb (approved):

PR #6512, opened 2023-01-05
fix: ensure ingester state tracked in querier cache is always in-sync

### LINKED ISSUE #6510: `parquet_file` cache locking
The querier `parquet_file` cache has a lock-gap under the new write path (=kafkaless). The data in this lookup table:

https://github.com/influxdata/influxdb_iox/blob/e1d0bcdaad46ad24b8e7f4b136761d77e8e6ffd6/querier/src/cache/parquet_file.rs#L104

Should be part of the cached values (`CachedParquetFiles`) and the `remove_if` check should use these. Otherwise concurrent cache accesses may de-sync these two data structures. The caching framework was deliberately designed to avoid this issue.

crepererum:
Fixes #6510.


querier/src/cache/parquet_file.rs
```diff
@@ -534,18 +539,6 @@ mod tests {
             .get(table_id, None, Some(HashMap::from([(new_uuid, 1)])), None)
             .await;
         assert_histogram_metric_count(&catalog.metric_registry, METRIC_NAME, 4);
-
-        // See the "new" UUID *and* the old UUID with the same counts as seen before: still use the
-        // cache
```
crepererum, line 539: This behavior is really hard to archive w/ proper, scalable locking. However this use case is also somewhat pointless: if there's really one ingester that is gone and comes back later, than what the hack is wrong with the system anyways?!

querier/src/cache/parquet_file.rs
```diff
@@ -34,19 +33,31 @@ pub enum Error {
...
 
+type IngesterCounts = Option<Arc<HashMap<Uuid, u64>>>;
+
 /// Holds catalog information about a parquet file
 #[derive(Debug)]
 pub struct CachedParquetFiles {
     /// Parquet catalog information
     pub files: Arc<Vec<Arc<ParquetFile>>>,
+
+    /// Number of persisted Parquet files per table ID per ingester UUID that ingesters have told
```
alamb, line 44: 👍

querier/src/cache/parquet_file.rs
```diff
@@ -64,12 +75,14 @@ impl CachedParquetFiles {
...
-        // size of the Arc itself
+        // size of the Arc+(Option+HashMap) itself
         mem::size_of_val(self) +
         // Vec overhead
             mem::size_of_val(self.files.as_ref()) +
         // size of the underlying parquet files
-            self.files.iter().map(|f| f.size()).sum::<usize>()
+            self.files.iter().map(|f| f.size()).sum::<usize>() +
+        // hashmap data
+            self.persisted_file_counts_from_ingesters.as_ref().map(|map| std::mem::size_of_val(map.as_ref()) + map.capacity() * mem::size_of::<(Uuid, u64)>()).unwrap_or_default()
```
alamb, line 85: it is unfortunate there is no way to make sure this calculation stays in sync with IngesterCounts (like if we change what is in `IngesterCounts` someone will have to remember to change this too.

The only way around it might be a newtype that wraps the IngesterCounts 🤔

alamb (approved):

PR #6513, opened 2023-01-05
feat(data_types): SequenceNumberSet

domodwyer:
A set of sequence numbers! I even wrote a test.

---

* feat(data_types): SequenceNumberSet (a184028a0)

      Adds a space-efficient encoding of a set of SequenceNumber, backed by
      roaring bitmaps.
      
      Memory utilisation will change as the number of elements changes,
      according to the underlying roaring bitmap design, but is intended to be
      "relatively" cheap.


data_types/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
+croaring = "0.7.0"
```
crepererum, line 10: do we really need this dependency w/ all it's buckload of hacks again? If so, you also need to bring back all arcane foo removed here:

https://github.com/influxdata/influxdb_iox/pull/5910/files#diff-f6009bd0d260464389ace37ab2f89adae993e1fa4a47f779e4c9859937005ced

crepererum: OK, at least version 0.7 doesn't require bindgen and clang anymore. That's good :)

crepererum (approved):

PR #6514, opened 2023-01-05
feat: trigger that updates partition table when a parquet file is created

NGA-TRAN:
Another PR for https://github.com/influxdata/idpe/issues/16836


iox_catalog/migrations/20230105120822_trigger_update_partition.sql
```diff
@@ -0,0 +1,22 @@
...
+END;
+$$;
+
+-- TRIGGER that fires the update_partition_on_new_file_at function when a new file is added to the parquet_file table
+CREATE TRIGGER update_partition 
+    AFTER INSERT ON parquet_file 
+    FOR EACH ROW 
+    EXECUTE PROCEDURE update_partition_on_new_file_at();
+    
+
```
NGA-TRAN, line 22: I have tested this locally. When I insert a new parquet file, the correspondent partition is updated

iox_catalog/migrations/20230105120822_trigger_update_partition.sql
```diff
@@ -0,0 +1,22 @@
+-- A new field in the partition that stores the latest time a new file is added to the partition
+ALTER TABLE partition ADD COLUMN IF NOT EXISTS new_file_at bigint;
+
+-- FUNTION that updates the new_file_at field in the partition table when the update_partition trigger is fired
+CREATE OR REPLACE FUNCTION update_partition_on_new_file_at()
+RETURNS TRIGGER 
+LANGUAGE PLPGSQL
+AS $$
+BEGIN 
+    UPDATE partition SET new_file_at = (SELECT (EXTRACT(EPOCH FROM (select now() ))) * 1000000000) WHERE id = NEW.partition_id;
```
pauldix, line 10: I think the selects here are extraneous. You should be able to just do: 
```
UPDATE partition SET new_file_at = extract (epoch from now()) * 1000000000 WHERE id = NEW.partition_id;
```

NGA-TRAN: Good idea

NGA-TRAN: Fixed @pauldix

pauldix (approved):

PR #6519, opened 2023-01-05
feat: function to get parttion candidates from partition table

NGA-TRAN:
One more step for https://github.com/influxdata/idpe/issues/16836

This PR does:
1. Main tasks: add the function `partitions_with_recent_created_files` to read data from  table `partition` instead2. 
2. While add test for the above function, I found data created in the catalog  from previous tests affect this test.  To fix this, I remove the test namespaces which will cascade to delete  all tables and partitions at the end of all catalog tests. 
3. As a result, I found assert bug in a test (found ticket and add comment next to the bug)
4. Created and attached a ticket to remove functions and indexes after we finish using the new function (need a couple more PRs)


iox_catalog/src/postgres.rs
```diff
@@ -1413,6 +1461,24 @@ WHERE id = $2;
...
+        &mut self,
+        time_in_the_past: Timestamp,
+    ) -> Result<Vec<PartitionParam>> {
+        sqlx::query_as(
+            r#"
+            SELECT p.id as partition_id, p.table_id, t.namespace_id, p.shard_id
+            FROM partition p, table_name t
+            WHERE p.new_file_at > $1
+                AND p.table_id = t.id;
+            "#,
```
NGA-TRAN, line 1475: This is the main change of this PR. Since we are trying to minimize the changes in hot/warm/cold compaction functions, we return `partition_id, table_id, namespace_id, and shard_id` here. This requires us to join `partition` with `table_name`. Since they are both small, this query is quite fast.
It will take me one or 2 more PRs to integrate this function into compaction functions. After the integration, I will modify the code to read `table_id, namespace_id`  after this step and  simplify this query to only return `partition_id` and do not do the join here

iox_catalog/src/interface.rs
```diff
@@ -447,6 +447,16 @@ pub trait PartitionRepo: Send + Sync {
         table_id: TableId,
     ) -> Result<Partition>;
 
+    /// create partition with new_file_at value
+    /// NOTE: this is only used for testing purpose becasue the mem catalog does not have trigger
```
pauldix, line 451: If this is the case, then it seems like it would be better to just update the MemCatalog to set this value on partitions when a parquet_file is inserted. Then this function could be removed and the MemCatalog will behave as we expect the Postgres catalog to behave. Which means we should probably have a test for that as well.

iox_catalog/migrations/20230105120822_trigger_update_partition.sql
```diff
@@ -7,7 +7,7 @@ RETURNS TRIGGER
 LANGUAGE PLPGSQL
 AS $$
 BEGIN 
-    UPDATE partition SET new_file_at = EXTRACT(EPOCH FROM now() ) * 1000000000 WHERE id = NEW.partition_id;
+    UPDATE partition SET new_file_at = NEW.created_at WHERE id = NEW.partition_id;
```
NGA-TRAN, line 10: Make the partitions `new_file_at` the same as its parquet_file's `created_at` to keep them consistent and also easy to test

iox_catalog/src/interface.rs
```diff
@@ -3740,6 +3841,13 @@ pub(crate) mod test_helpers {
             .await
             .unwrap();
         assert!(partitions.is_empty());
+        // get from partition table
+        let partitions = repos
+            .partitions()
+            .partitions_with_recent_created_files(time_two_hour_ago)
+            .await
+            .unwrap();
+        assert!(partitions.is_empty());
```
NGA-TRAN, line 3850: This test tests whether the parquet files is created recently or not. Before we use parquet_file's function to verify it. We only need to add the new partition's function which does the same thing here.

iox_catalog/src/interface.rs
NGA-TRAN: @pauldix This is addressed

iox_catalog/src/postgres.rs
pauldix: Ugh, ok, do you have an issue to track that? The compactor shouldn't need that other information at this stage of work. It should only need a list of partition IDs, which is all it needs to then do the query to get the parquet file records, which all have the namespace and table ids in them.

pauldix: It seems like this PR should reflect what we want and then the follow on compactor PRs should fix the issue that it requires more information than it should

pauldix (approved):

